PR #9328, opened 2022-11-08
fix(emqx_authn_api): return 404 for status of unknown authenticator

sstrigler:
This also makes sure we call the same code everytime we access an authenticator. Moreover we return a 500 in case a remote call fails due to technical issues.

Fixes [EMQX-7821](https://emqx.atlassian.net/browse/EMQX-7821)

## PR Checklist

### DIFF
--- a/apps/emqx_authn/src/emqx_authn_api.erl
+++ b/apps/emqx_authn/src/emqx_authn_api.erl
@@ -30,6 +30,7 @@
 -define(BAD_REQUEST, 'BAD_REQUEST').
 -define(NOT_FOUND, 'NOT_FOUND').
 -define(ALREADY_EXISTS, 'ALREADY_EXISTS').
+-define(INTERNAL_ERROR, 'INTERNAL_ERROR').
 
 % Swagger
 
@@ -224,7 +225,8 @@ schema("/authentication/:id/status") ->
                     hoconsc:ref(emqx_authn_schema, "metrics_status_fields"),
                     status_metrics_example()
                 ),
-                400 => error_codes([?BAD_REQUEST], <<"Bad Request">>)
+                404 => error_codes([?NOT_FOUND], <<"Not Found">>),
+                500 => error_codes([?INTERNAL_ERROR], <<"Internal Service Error">>)
             }
         }
     };
@@ -576,7 +578,11 @@ authenticator(delete, #{bindings := #{id := AuthenticatorID}}) ->
     delete_authenticator([authentication], ?GLOBAL, AuthenticatorID).
 
 authenticator_status(get, #{bindings := #{id := AuthenticatorID}}) ->
-    lookup_from_all_nodes(?GLOBAL, AuthenticatorID).
+    with_authenticator(
+        AuthenticatorID,
+        [authentication],
+        fun(_) -> lookup_from_all_nodes(?GLOBAL, AuthenticatorID) end
+    ).
 
 listener_authenticators(post, #{bindings := #{listener_id := ListenerID}, body := Config}) ->
     with_listener(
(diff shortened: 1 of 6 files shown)


sstrigler:
I decided to go with a `404` now to be consistent with calls to "parent" resource, ie 
```
GET /authentication/foo
> 404

GET /authentication/foo/status
> 404
```
both show the same result (as in HTTP status code). Anything else would feel weird.

If I understand things right though it could still happen that on the target node the authenticator exists, thus we yield a `200` but on a remote node it doesn't, in that case this would still show in `node_error` of the response body just as it did before. And I think that's fine.

apps/emqx_authn/src/emqx_authn_api.erl
```diff
@@ -576,7 +578,11 @@ authenticator(delete, #{bindings := #{id := AuthenticatorID}}) ->
     delete_authenticator([authentication], ?GLOBAL, AuthenticatorID).
 
 authenticator_status(get, #{bindings := #{id := AuthenticatorID}}) ->
-    lookup_from_all_nodes(?GLOBAL, AuthenticatorID).
+    with_authenticator(
+        AuthenticatorID,
+        [authentication],
+        fun(_) -> lookup_from_all_nodes(?GLOBAL, AuthenticatorID) end
```
id, line 584: Not sure if this code is equivalent.
Before we'd try to lookup authenticator on each node in the cluster, now we only lookup on other nodes when this authenticator is found in the configuration on the local node?
Or did I get it wrong?

sstrigler: We don't do it instead, we do it in addition. I explained why in the comment above. Alternative would be to also do a remote lookup in `/authentication/{name}` and then (for both resources) return `404` if just a single one of the remote calls returns a `not_found`?

sstrigler: I mentioned it in the ticket but I think the design of the API here is flawed, it should include node names, and maybe have an explicit `all` node for aggregates. Do we want to leak node names though? How does the cluster behave if nodes have different configurations? Is this by design and a wanted situation? If not, who should be in charge to resolve it (the API or the 'system' itself)?

apps/emqx_authn/src/emqx_authn_api.erl
```diff
@@ -901,7 +924,7 @@ lookup_from_all_nodes(ChainName, AuthenticatorID) ->
                 node_error => HelpFun(maps:map(Fun, ErrorMap), reason)
             }};
         {error, ErrL} ->
-            {400, #{
+            {500, #{
```
thalesmg, line 927: Just confirming: is there a test case that covers this?

apps/emqx_authn/test/emqx_authn_api_SUITE.erl
```diff
@@ -247,6 +247,15 @@ test_authenticator(PathPrefix) ->
         <<"connected">>,
         LookFun([<<"status">>])
     ),
+
+    ?assertMatch(
+        {ok, 404, _},
+        request(
+            get,
+            uri(PathPrefix ++ [?CONF_NS, "unknown_auth_chain", "status"])
```
thalesmg, line 255: Nit: maybe also add a test for an unknown authenticator status under a listener?

thalesmg (approved):

apps/emqx_authn/test/emqx_authn_api_SUITE.erl
sstrigler: That's implicit since `PathPrefix` is a param to the test and it is also called as `test_authenticator(["listeners", ?TCP_DEFAULT])`.
If that's what you mean.

apps/emqx_authn/src/emqx_authn_api.erl
sstrigler: I'll add one.

apps/emqx_authn/test/emqx_authn_api_SUITE.erl
thalesmg: Aha, I see.  That's what I meant.  :smiley_cat:

lafirest (approved):

thalesmg (approved):

PR #9342, opened 2022-11-10
feat: rule_engine_validator integer and num range support

JimMoen:
## PR Checklist

## More information
We need to add a `range` attribute to other number types, or fields that should be of type integer.

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_validator.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_validator.erl
@@ -32,7 +32,7 @@
     schema => spec()
 }.
 
--type data_type() :: string | password | number | boolean
+-type data_type() :: string | password | number | integer | boolean
     | object | array | file | binary_file | cfgselect.
 
 -type params_spec() :: #{name() => spec()} | any.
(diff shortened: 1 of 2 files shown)


apps/emqx_rule_engine/src/emqx_rule_validator.erl
```diff
@@ -164,6 +175,30 @@ validate_file(Val) when is_list(Val) -> Val;
...
+min_(Interval, Val) ->
+    throw({invalid_range_interval, {min, {Val, Interval}}}).
+
+max_(Val, any) -> Val;
+max_(Val, Max) when is_number(Max) andalso Val =< Max -> Val;
+max_(Val, #{open := any}) -> Val;
+max_(Val, #{open := Max}) when Val < Max -> Val;
+max_(Val, #{closed := Max}) when Val =< Max -> Val;
+max_(Val, Interval) ->
+    throw({invalid_range_interval, {max, {Val, Interval}}}).
```
terry-xiaoyu, line 200: invalid_range_interval -> invalid_range

terry-xiaoyu (dismissed):

zmstone:
this breaks compatibility ?
what happens to the users using mixed numbers ?

zmstone (changes requested):
link jira or provide more information about why the change.

questions:
does this break compatibility ?
what happens to uses who are already using integer and float in a mixed way?

PR #9347, opened 2022-11-10
refactor(crl): rename `s/enable_crl_cache/enable_crl_check/g`

thalesmg:
To avoid some confusion that may arise from the `enable_crl_cache` name.  Enabling this option is actually the way to enable the CRL check itself, and caching should always be on in that case.

## PR Checklist

### DIFF
--- a/priv/emqx.schema
+++ b/priv/emqx.schema
@@ -1702,7 +1702,7 @@ end}.
   {datatype, {duration, ms}}
 ]}.
 
-{mapping, "listener.ssl.$name.enable_crl_cache", "emqx.listeners", [
+{mapping, "listener.ssl.$name.enable_crl_check", "emqx.listeners", [
   {default, false},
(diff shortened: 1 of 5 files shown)


priv/emqx.schema
```diff
@@ -1702,7 +1702,7 @@ end}.
   {datatype, {duration, ms}}
 ]}.
 
-{mapping, "listener.ssl.$name.enable_crl_cache", "emqx.listeners", [
+{mapping, "listener.ssl.$name.enable_crl_check", "emqx.listeners", [
```
lafirest, line 1705: should we keep the old mapping for compatibility?

thalesmg: This feature is not released yet, so it's safe to rename now.  :smiley_cat:

lafirest (approved):

zmstone (approved):

PR #9367, opened 2022-11-15
feat(banned): kick session when it is banned by clientid

lafirest:
A banned client shouldn't continue to publish/receive, so it's better to kick its session when it is banned.

## PR Checklist

      [EMQX-8109](https://emqx.atlassian.net/browse/EMQX-8109)

### DIFF
--- a/apps/emqx/src/emqx_banned.erl
+++ b/apps/emqx/src/emqx_banned.erl
@@ -21,6 +21,7 @@
 -include("emqx.hrl").
 -include("logger.hrl").
 -include("types.hrl").
+-include_lib("snabbkaffe/include/snabbkaffe.hrl").
 
 %% Mnesia bootstrap
 -export([mnesia/1]).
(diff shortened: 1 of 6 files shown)


apps/emqx/src/emqx_banned.erl
```diff
@@ -266,3 +267,22 @@ expire_banned_items(Now) ->
...
+
+insert_banned(Banned) ->
+    mria:dirty_write(?BANNED_TAB, Banned),
+    on_banned(Banned).
+
+on_banned(#banned{who = {clientid, ClientId}}) ->
+    %% kick the session if the client is banned by clientid
+    ?tp(
+        warning,
+        try_to_kick_session,
```
zmstone, line 279:
```suggestion
        kick_session_due_to_banned,
```

apps/emqx/src/emqx_banned.erl
```diff
@@ -266,3 +267,22 @@ expire_banned_items(Now) ->
...
+        #{
+            reason => "client is banned",
```
zmstone, line 281: no need for this reason.
```suggestion
```

zmstone (approved):

Issue #9341, opened 2022-11-10
clients API too slow in emqx 5.0.8. 


siniusxcode:
<img width="836" alt="스크린샷 2022-11-10 오후 5 35 19" src="https://user-images.githubusercontent.com/1418182/201042155-59fc7dc7-6d10-48e1-803a-6afd61d3b380.png">
<img width="857" alt="스크린샷 2022-11-10 오후 5 35 42" src="https://user-images.githubusercontent.com/1418182/201042168-6d9451e4-e2a2-4875-8d91-871b34e37868.png">
#### Environment

- OS: CentOS Linux release 7.5.1804 (Core) 
- Erlang/OTP: 22.2.6-1
- EMQ: 5.0.8

#### Description

I have two emqx clusters(4.0.7 and 5.0.8) of two vm servers(same spec: 12Cores, 16G) each.
Using another test server, I tried connecting 50,000 clients to cluster and calling clients API.

Result of response time for client API
5.0.8 : over 3s
4.0.7 : under 300ms

There is too much difference in response time between the two versions.
I want to use the 5.x version, but it is difficult to use because of this part.

zhongwencool:
Thanks for reporting, We plan to fix this in the next version `5.0.10` 👍

siniusxcode:
Thank you for your answer. By the way, I tested the same in version 5.0.10 now and the same problem still occurs. And I found some additional problems, I will register them as other issue.

zhongwencool:
> Thank you for your answer. By the way, I tested the same in version 5.0.10 now and the same problem still occurs. And I found some additional problems, I will register them as other issue.

sorry for the wrong version. The next version is `5.0.11`

PR #9374, opened 2022-11-16
Improve Clients query performance

HJianBo:
Fixes https://github.com/emqx/emqx/issues/9341

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change [EMQX-8137](https://emqx.atlassian.net/browse/EMQX-8137)

## Backward Compatibility

1. When performing a fuzzy search, `count` is always returned as `#{meta => #{count := 0}}`, because if we want to get the total number of results, We have to query all rows out and perform a fuzzy matching function each of rows. It's maybe more expensive.

2. Change the `emqx_channel_info` ets table type from `set` to `ordered_set`

## More information

This PR contains 2 fixes:
1. Fix incorrect Swagger Schema for `GET ../clients` and `GET ../subscriptions`.
3. Refactored the implementation of the `emqx_mgmt_api:node_query` and `cluster_query` functions to improve query performance with large data sets.

In the previous implementation, there were two issues that resulted in significant performance overhead:
- **Problem1:** Formatting is performed for each page scanning, whether these pages are returned.
- **Problem2:** Each query traverses the entire dataset, whether it is the first page or last page needed.

## Benchmark
Generate 1M clients into the `emqx_channel_info` table by [emqx-clients-gen script](https://gist.github.com/HJianBo/a7f555e99025702eda0f41da88afad33) and compare the following scenarios

#### Case 1: No conditions
Scripts
```erlang
timer:tc(fun() ->
  emqx_mgmt_api_clients:clients(
    get,
     #{query_string => #{<<"limit">> => 100,<<"page">> => 1}})
end).
```
Report:
| Page   | Before    | Now    |
| ------ | --------- | ------ |
| 1          | 32.304s  | 5.264ms |
| 5000   | 32.935s  | 1.521s |
| 10000 | 31.533s   | 3.074s |

#### Case 2: Performing an exact match
Script:
```erlang
timer:tc(fun() ->
  emqx_mgmt_api_clients:clients(
    get,
    #{query_string => #{<<"limit">> => 100,<<"page">> => 1, <<"username">> => <<"user-999">>}})
 end).
```
Report:
| Page | Before | Now    |
| ---- | ------ | ------ |
| 1    | 1.721s | *3.469s |

`*` It is 2 times more expensive, as the implementation now queries the total and target data separately.

#### Case 3:  Performing a fuzzy search
Script:
```erlang
timer:tc(fun() ->
  emqx_mgmt_api_clients:clients(
    get,
    #{query_string => #{<<"limit">> => 100,<<"page">> => 12,<<"like_username">> => <<"-999">>}})
  end).
```
Report:
| Page | Before | Now    |
| ---- | ------ | ------ |
| 1    | 3.845s | 0.306s |
| 6    | 3.973s | 1.992s |
| 12   | 3.541s | 3.380s |

### DIFF
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -651,7 +651,7 @@ init([]) ->
     ok = emqx_tables:new(?CHAN_TAB, [bag, {read_concurrency, true} | TabOpts]),
     ok = emqx_tables:new(?CHAN_CONN_TAB, [bag | TabOpts]),
     ok = emqx_tables:new(?CHAN_INFO_TAB, [ordered_set, compressed | TabOpts]),
-    ok = emqx_tables:new(?CHAN_LIVE_TAB, [set, {write_concurrency, true} | TabOpts]),
+    ok = emqx_tables:new(?CHAN_LIVE_TAB, [ordered_set, {write_concurrency, true} | TabOpts]),
     ok = emqx_stats:update_interval(chan_stats, fun ?MODULE:stats_fun/0),
     State = #{chan_pmon => emqx_pmon:new()},
     {ok, State}.
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -650,7 +650,7 @@ init([]) ->
     TabOpts = [public, {write_concurrency, true}],
     ok = emqx_tables:new(?CHAN_TAB, [bag, {read_concurrency, true} | TabOpts]),
     ok = emqx_tables:new(?CHAN_CONN_TAB, [bag | TabOpts]),
-    ok = emqx_tables:new(?CHAN_INFO_TAB, [set, compressed | TabOpts]),
+    ok = emqx_tables:new(?CHAN_INFO_TAB, [ordered_set, compressed | TabOpts]),
     ok = emqx_tables:new(?CHAN_LIVE_TAB, [set, {write_concurrency, true} | TabOpts]),
     ok = emqx_stats:update_interval(chan_stats, fun ?MODULE:stats_fun/0),
     State = #{chan_pmon => emqx_pmon:new()},
--- a/apps/emqx_authn/src/enhanced_authn/emqx_enhanced_authn_scram_mnesia.erl
+++ b/apps/emqx_authn/src/enhanced_authn/emqx_enhanced_authn_scram_mnesia.erl
@@ -274,9 +274,12 @@ list_users(QueryString, #{user_group := UserGroup}) ->
 %%--------------------------------------------------------------------
 %% QueryString to MatchSpec
 
--spec qs2ms(atom(), {list(), list()}) -> {ets:match_spec(), fun() | undefined}.
+-spec qs2ms(atom(), {list(), list()}) -> emqx_mgmt_api:match_spec_and_filter().
 qs2ms(_Tab, {QString, Fuzzy}) ->
-    {ms_from_qstring(QString), fuzzy_filter_fun(Fuzzy)}.
+    #{
+        match_spec => ms_from_qstring(QString),
+        fuzzy_fun => fuzzy_filter_fun(Fuzzy)
+    }.
 
 %% Fuzzy username funcs
 fuzzy_filter_fun([]) ->
--- a/apps/emqx_authn/src/enhanced_authn/emqx_enhanced_authn_scram_mnesia.erl
+++ b/apps/emqx_authn/src/enhanced_authn/emqx_enhanced_authn_scram_mnesia.erl
@@ -262,7 +262,14 @@ lookup_user(UserID, #{user_group := UserGroup}) ->
 
 list_users(QueryString, #{user_group := UserGroup}) ->
     NQueryString = QueryString#{<<"user_group">> => UserGroup},
-    emqx_mgmt_api:node_query(node(), NQueryString, ?TAB, ?AUTHN_QSCHEMA, ?QUERY_FUN).
+    emqx_mgmt_api:node_query(
+        node(),
+        NQueryString,
+        ?TAB,
+        ?AUTHN_QSCHEMA,
+        ?QUERY_FUN,
+        fun ?MODULE:format_user_info/1
+    ).
 
 %%--------------------------------------------------------------------
 %% Query Functions
@@ -273,8 +280,7 @@ query(Tab, {QString, []}, Continuation, Limit) ->
         Tab,
         Ms,
         Continuation,
-        Limit,
-        fun format_user_info/1
+        Limit
     );
 query(Tab, {QString, FuzzyQString}, Continuation, Limit) ->
     Ms = ms_from_qstring(QString),
@@ -283,8 +289,7 @@ query(Tab, {QString, FuzzyQString}, Continuation, Limit) ->
         Tab,
         {Ms, FuzzyFilterFun},
         Continuation,
-        Limit,
-        fun format_user_info/1
+        Limit
     ).
 
 %%--------------------------------------------------------------------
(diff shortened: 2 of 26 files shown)


apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -855,6 +838,8 @@ ms(created_at, X) ->
 %%--------------------------------------------------------------------
 %% Match funcs
 
+fuzzy_filter_fun([]) ->
+    undefined;
```
thalesmg, line 842: Just confirming: do we have some test coverage for this clause? (and similar ones in the other modules)

apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
```diff
@@ -168,16 +172,12 @@ subscriptions(get, #{query_string := QString}) ->
...
-format({{Subscriber, Topic}, Options}) ->
-    format({Subscriber, Topic, Options});
-format({_Subscriber, Topic, Options}) ->
+format(WhichNode, {{_Subscriber, Topic}, Options}) ->
     maps:merge(
         #{
             topic => get_topic(Topic, Options),
             clientid => maps:get(subid, Options),
-            node => node()
+            node => WhichNode
```
thalesmg, line 180: This seems like a new behavior; maybe add a test?

apps/emqx_management/src/emqx_mgmt_api_clients.erl
HJianBo: Recently, I have added some test cases to all the new code added. thanks ❤️

apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
HJianBo: Done. I think it can be covered by `emqx_mgmt_api_SUITE:t_cluster_query/1` now

apps/emqx/src/emqx_cm.erl
```diff
@@ -650,7 +650,7 @@ init([]) ->
     TabOpts = [public, {write_concurrency, true}],
     ok = emqx_tables:new(?CHAN_TAB, [bag, {read_concurrency, true} | TabOpts]),
     ok = emqx_tables:new(?CHAN_CONN_TAB, [bag | TabOpts]),
-    ok = emqx_tables:new(?CHAN_INFO_TAB, [set, compressed | TabOpts]),
+    ok = emqx_tables:new(?CHAN_INFO_TAB, [ordered_set, compressed | TabOpts]),
     ok = emqx_tables:new(?CHAN_LIVE_TAB, [set, {write_concurrency, true} | TabOpts]),
```
zmstone, line 654:
```suggestion
    ok = emqx_tables:new(?CHAN_LIVE_TAB, [ordered_set, {write_concurrency, true} | TabOpts]),
```

apps/emqx_authn/src/enhanced_authn/emqx_enhanced_authn_scram_mnesia.erl
```diff
@@ -262,35 +261,25 @@ lookup_user(UserID, #{user_group := UserGroup}) ->
...
+        ?AUTHN_QSCHEMA,
+        fun ?MODULE:qs2ms/2,
+        fun ?MODULE:format_user_info/1
     ).
 
 %%--------------------------------------------------------------------
-%% Match funcs
+%% QueryString to MatchSpec
+
+-spec qs2ms(atom(), {list(), list()}) -> {ets:match_spec(), fun() | undefined}.
```
zmstone, line 276: to become more search friendly, also to avoid `badfun`,
we can change this return value to:
```
#{match_spec => ms_from_qstring(Qstrong),
    fuzzy_fun => {?MODULE, fuzzy_filter, [Fuzzy]}
}.
```

HJianBo: Good catch! how about a named function? i.g `fun ?MODULE:fuzzy_filter/1` It is more friendly for dialyzer?

apps/emqx_authn/src/simple_authn/emqx_authn_mnesia.erl
```diff
@@ -288,35 +287,25 @@ lookup_user(UserID, #{user_group := UserGroup}) ->
...
-query(Tab, {QString, FuzzyQString}, Continuation, Limit) ->
-    Ms = ms_from_qstring(QString),
-    FuzzyFilterFun = fuzzy_filter_fun(FuzzyQString),
-    emqx_mgmt_api:select_table_with_count(
-        Tab,
-        {Ms, FuzzyFilterFun},
-        Continuation,
-        Limit,
-        fun format_user_info/1
+    emqx_mgmt_api:node_query(
```
savonarola, line 290: Nit: (Not related to the PR directly) These parts of the authenticators' code have always made me confused: here in a "deeply internal and data-related module" we have to deal with both presentation-related (formatting) and controller-related  (qs -> ms) logic... and also call `emqx_mgmt_api` app. But I myself do not have any good ideas how to easily refactor this (and should we do this at all) 😅

lafirest: same for me, it is not easy to understand and use these API. I personally prefer to connect some atomic functions by pipe

apps/emqx/src/emqx_cm.erl
```diff
@@ -650,8 +650,8 @@ init([]) ->
     TabOpts = [public, {write_concurrency, true}],
     ok = emqx_tables:new(?CHAN_TAB, [bag, {read_concurrency, true} | TabOpts]),
     ok = emqx_tables:new(?CHAN_CONN_TAB, [bag | TabOpts]),
-    ok = emqx_tables:new(?CHAN_INFO_TAB, [set, compressed | TabOpts]),
-    ok = emqx_tables:new(?CHAN_LIVE_TAB, [set, {write_concurrency, true} | TabOpts]),
+    ok = emqx_tables:new(?CHAN_INFO_TAB, [ordered_set, compressed | TabOpts]),
```
lafirest, line 653: Although `ordered_set` is fast too, slower than set.
Is it worth changing its type for only the API search?

apps/emqx_authn/src/simple_authn/emqx_authn_mnesia.erl
HJianBo: It is mainly due to implementing the behavior defined in `emqx_authentication.erl`.
:thinking:  Perhaps the user management part can be detached from the behavior of emqx_authentication

apps/emqx/src/emqx_cm.erl
zmstone: it's easier to implement a externa-serializable query cursor using ordered_set

apps/emqx_authn/src/enhanced_authn/emqx_enhanced_authn_scram_mnesia.erl
zmstone: return a map instead of tuple from here?

zmstone (changes requested):
Update swagger schema description to explain the `count = 0` situations.

zmstone (approved):

apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
```diff
@@ -138,7 +138,14 @@ fields(limit) ->
     Meta = #{in => query, desc => Desc, default => ?DEFAULT_ROW, example => 50},
     [{limit, hoconsc:mk(range(1, ?MAX_ROW_LIMIT), Meta)}];
 fields(count) ->
-    Meta = #{desc => <<"Results count.">>, required => true},
+    Desc = <<
+        "Total count.<br/>"
+        "Note: This value returns 0 when there are no records or when a "
+        "fuzzy searching is performed.<br/>"
+        "Because it's very expensive to get the total count in a fuzzy "
+        "searching."
```
thalesmg, line 146:
```suggestion
        "Note: This value returns 0 when there are no records or when a "
        "fuzzy search is performed.<br/>"
        "Because it's very expensive to get the total count in a fuzzy "
        "search."
```

zmstone (dismissed):

zmstone (approved):

PR #9387, opened 2022-11-17
chore: make remsh node name away from the atom DOS attack

lafirest:
The remsh node name is generated to be unique, this may cause atom leakage, so we need to change the generation rule to limit the total of these names

## PR Checklist

- [x] ~~Added tests for the changes~~
- [x] ~~Changed lines covered in coverage report~~
- [x] ~~Change log has been added to `changes/` dir~~
      [EMQX-8221](https://emqx.atlassian.net/browse/EMQX-8221)

### DIFF
--- a/bin/emqx
+++ b/bin/emqx
@@ -396,7 +396,7 @@ remsh() {
 
 # Generate a random id
 relx_gen_id() {
-    od -t x -N 4 /dev/urandom | head -n1 | awk '{print $2}'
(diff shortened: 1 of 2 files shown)


zmstone:
Just to be clear, this is *NOT* any attack.
If an attacker gained access to run emqx shell commands, atom table overflow is perhaps the least to care about.

PR #9389, opened 2022-11-17
ci: release only the default packages

zmstone:
https://emqx.atlassian.net/browse/EMQX-8239
We no longer need to 'flavour' the packages with otp or elixir version numbers since we found a way forward for hot-upgrading 

Based this on v5.0.10 because we want to merge it to e5.0.0-beta.5
Will add release note change in the next PR after merge.

## PR Checklist

~~- [ ] Added tests for the changes~~
~~- [ ] Changed lines covered in coverage report~~
~~- [ ] For EMQX 4.x: `appup` files updated (execute `scripts/update-appup.sh emqx`)~~
~~- [ ] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up~~
~~- [ ] In case of non-backward compatible changes, reviewer should check this item as a write-off, and add details in~~ **Backward Compatibility** section

### DIFF
--- a/pkg-vsn.sh
+++ b/pkg-vsn.sh
@@ -11,11 +11,10 @@ help() {
     echo "$0 PROFILE [options]"
     echo
     echo "-h|--help:       To display this usage information"
-    echo "--default:       Print default vsn number. e.g. e.g. 5.0.0-ubuntu20.04-amd64"
-    echo "--long:          Print long vsn number. e.g. 5.0.0-otp24.2.1-1-ubuntu20.04-amd64"
+    echo "--long:          Print long vsn number. e.g. 5.0.0-ubuntu20.04-amd64"
     echo "                 Otherwise short e.g. 5.0.0"
     echo "--elixir:        Include elixir version in the long version string"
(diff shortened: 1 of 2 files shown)


pkg-vsn.sh
```diff
@@ -12,10 +12,10 @@ help() {
     echo
     echo "-h|--help:       To display this usage information"
     echo "--default:       Print default vsn number. e.g. e.g. 5.0.0-ubuntu20.04-amd64"
-    echo "--long:          Print long vsn number. e.g. 5.0.0-otp24.2.1-1-ubuntu20.04-amd64"
+    echo "--long:          Print long vsn number. e.g. 5.0.0-ubuntu20.04-amd64"
```
thalesmg, line 15: Now, it seems this option outputs the same as `--default`?

pkg-vsn.sh
```diff
@@ -12,10 +12,10 @@ help() {
     echo
     echo "-h|--help:       To display this usage information"
     echo "--default:       Print default vsn number. e.g. e.g. 5.0.0-ubuntu20.04-amd64"
```
thalesmg, line 14: Should remove this?  As it has been removed below?

thalesmg (approved):

id (approved):

PR #9240, opened 2022-10-26
feat: do not drop MQTTv5 properties in rule/bridge

kraftwerk28:
Fixes #9179

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [ ] Added tests for the changes
- [ ] Changed lines covered in coverage report
- [ ] Change log has been added to `changes/` dir
- [ ] For EMQX 4.x: `appup` files updated (execute `scripts/update-appup.sh emqx`)
- [ ] For internal contributor: there is a jira ticket to track this change, and another jira tickt to track doc updates (if any)
- [ ] In case of non-backward compatible changes, reviewer should check this item as a write-off, and add details in **Backward Compatibility** section

PR #9398, opened 2022-11-21
1121 rule bridge properties

zmstone:
this is a follow up for https://github.com/emqx/emqx/pull/9240

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_actions.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_actions.erl
@@ -110,8 +110,9 @@ republish(
     Payload = format_msg(PayloadTks, Selected),
     QoS = replace_simple_var(QoSTks, Selected, 0),
     Retain = replace_simple_var(RetainTks, Selected, false),
+    PubProps = format_pub_props(maps:get(<<"pub_props">>, Selected, #{})),
     ?TRACE("RULE", "republish_message", #{topic => Topic, payload => Payload}),
-    safe_publish(RuleId, Topic, QoS, Flags#{retain => Retain}, Payload);
+    safe_publish(RuleId, Topic, QoS, Flags#{retain => Retain}, Payload, PubProps);
 %% in case this is a "$events/" event
 republish(
     Selected,
@@ -129,8 +130,9 @@ republish(
     Payload = format_msg(PayloadTks, Selected),
     QoS = replace_simple_var(QoSTks, Selected, 0),
     Retain = replace_simple_var(RetainTks, Selected, false),
+    PubProps = maps:get(pub_props, Selected, #{}),
     ?TRACE("RULE", "republish_message_with_flags", #{topic => Topic, payload => Payload}),
-    safe_publish(RuleId, Topic, QoS, #{retain => Retain}, Payload).
+    safe_publish(RuleId, Topic, QoS, #{retain => Retain}, Payload, PubProps).
 
 %%--------------------------------------------------------------------
 %% internal functions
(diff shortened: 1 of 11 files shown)


apps/emqx_rule_engine/src/emqx_rule_actions.erl
```diff
@@ -110,8 +110,9 @@ republish(
     Payload = format_msg(PayloadTks, Selected),
     QoS = replace_simple_var(QoSTks, Selected, 0),
     Retain = replace_simple_var(RetainTks, Selected, false),
+    PubProps = maps:get(pub_props, OriginalMsg, #{}),
```
zmstone, line 113: This was originally (#9240) taken from `Selected`,
which seemed to be wrong because selected element keys are of binary format. (i.e. `<<"pub_props">>`).

Now if we take it from the original message, it may seem to work for republishes, however it also affects
other events like `message.dropped`, `message.acked`, in those events, a user may not necessary
want to keep the properties.

For example, if someone just want to republish a `message.dropped` event, but only interested in the topic name.

the "proper" approach is perhaps to treat message headers like qos, retain flag, and payload, 
meaning to add another argument here:
https://github.com/emqx/emqx/blob/966e6ddeeb9efb0d76c2f20fe3e7d1e44c4434a8/apps/emqx_rule_engine/src/emqx_rule_engine_schema.erl#L139-L177

@terry-xiaoyu @kjellwinblad @thalesmg wdyt ?

terry-xiaoyu: Yes, we need a 'properties' config in this case. 

- If the 'properties' is not configured and it's a PUBLISH message, we republish the message using the original properties. 
- If the 'properties' is not configured and it's an event like `$events/client_connected`, we republish the message without any property. 
- If the 'properties' is configured, then we republish using the configured properties.

kjellwinblad (commented):
Looks good to me (just a few minor comments about sleeps in test cases).

apps/emqx_rule_engine/src/emqx_rule_actions.erl
thalesmg: From my very limited knowledge about this, it seems like the props should indeed be a new config; otherwise, it's not obvious that a property should be dropped or not.

apps/emqx_rule_engine/src/emqx_rule_actions.erl
```diff
@@ -93,44 +97,33 @@ republish(
...
-            retain := RetainTks,
-            topic := TopicTks,
-            payload := PayloadTks
-        }
-    }
-) ->
-    Topic = emqx_plugin_libs_rule:proc_tmpl(TopicTks, Selected),
-    Payload = format_msg(PayloadTks, Selected),
-    QoS = replace_simple_var(QoSTks, Selected, 0),
-    Retain = replace_simple_var(RetainTks, Selected, false),
```
zmstone, line 131: these three lines duplicates line 113-116, so i merged the two function clauses.
the only special part is the `flags` field.

apps/emqx_rule_engine/src/emqx_rule_actions.erl
zmstone: Thank you for your suggestions.
I have added a `user_properties` republish action argument with default template `${user_properties}`.

Decided *NOT* to fill `user_properties` with original message's `pub_props.'User-Property'`,
because the code is shared for both message republish and for message events such as `message.acked` and `message.dropped`.

In order to republish message with `User-Property`, a user will have to explicitly select a variable,
or configure with the special value `${pub_props.'User-Property'}`.

terry-xiaoyu (approved):

kjellwinblad (approved):

PR #9416, opened 2022-11-24
fix(ctl): fix unsafe `list_to_atom`

lafirest:
A command can be called only when it exists.

## PR Checklist

- [x] ~~Added tests for the changes~~
        [EMQX-8044](https://emqx.atlassian.net/browse/EMQX-8044)


zhongwencool:
Good catch! Maybe we can merge this into `v5.0.11`, not v5.0.12`?

zhongwencool (approved):

Issue #9402, opened 2022-11-21
Sync data from cluster node fails


tom-tschiller:
### What happened?

I try to set up an EMQX cluster on Kubernetes (EKS) with EMQX Operator. While EMQX Operator works fine, the EMQX nodes cannot join the cluster. The node, that starts first gets healthy, but all subsequently started nodes fail to join this node. According to the logs it looks like the incoming configuration cannot be applied locally:

```
2022-11-21T14:49:23.789436+00:00 [emergency] line: 202, mfa: emqx_conf_app:sync_data_from_node/1, msg: sync_data_from_node_failed, node: 'emqx@emqx-core-0.emqx-headless.mqtt-broker-test.svc.cluster.local', reason: {badrpc,{'EXIT',{{badmatch,{error,erofs}},[{emqx_conf_app,sync_data_from_node,0,[{file,"emqx_conf_app.erl"},{line,65}]}]}}}
```

The complete log file can be found below.

### What did you expect to happen?

I'd expect, that the cluster node joins the cluster and applies the configuration it has received.

### How can we reproduce it (as minimally and precisely as possible)?

I tried to simplify the configuration as much as possible to this:

<details>

```
apiVersion: apps.emqx.io/v2alpha1
kind: EMQX
metadata:
  name: emqx
spec:
  image: internal-artifactory-docker-registry/emqx/emqx:5.0.9
  coreTemplate:
    spec:
      env:
        - name: EMQX_LOG__CONSOLE_HANDLER__LEVEL
          value: debug
      podSecurityContext:
        runAsUser: 1000
        runAsGroup: 1000
        fsGroup: 1000
      extraVolumes:
        - name: emqx-home-volume
          emptyDir: {}
      extraVolumeMounts:
        - name: emqx-home-volume
          mountPath: /home/emqx
```

</details>

### Anything else we need to know?

The problem might have something to do with some of our Kubernetes settings being too restrictive. As you can see in the configuration above, I had to define some settings due to this already (podSecurityContext, extraVolumes).

However in this case I don't see what's wrong. I don't see what EMQX wants to do but is not able or allowed to do. It would already help me, if you have more information about this.

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
sysdescr  : EMQX
version   : 5.0.9
datetime  : 2022-11-21T16:34:28.119752362+00:00
uptime    : 1 hours, 47 minutes, 40 seconds
```

</details>

### OS version

<details>

```console
# On Linux:
[...]

PR #9429, opened 2022-11-28
fix: don't save cert/authz zip on disk when emqx start

zhongwencool:
Fixes [<9402>](https://github.com/emqx/emqx/issues/9402)

Make the data.zip file only in memory, don't need to save this on disk.

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_app.erl
+++ b/apps/emqx_conf/src/emqx_conf_app.erl
@@ -62,10 +62,11 @@ sync_data_from_node() ->
     TargetDirs = lists:filter(fun(Type) -> filelib:is_dir(filename:join(Dir, Type)) end, [
         "authz", "certs"
     ]),
-    {ok, Zip} = zip:zip(atom_to_list(node()) ++ "_data.zip", TargetDirs, [{cwd, Dir}]),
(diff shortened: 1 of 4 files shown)


ieQu1 (approved):

apps/emqx_conf/src/emqx_conf_app.erl
```diff
@@ -195,8 +196,12 @@ conf_sort({ok, _}, {ok, _}) ->
 sync_data_from_node(Node) ->
     case emqx_conf_proto_v2:sync_data_from_node(Node) of
         {ok, DataBin} ->
-            {ok, Files} = zip:unzip(DataBin, [{cwd, emqx:data_dir()}]),
-            ?SLOG(debug, #{node => Node, msg => "sync_data_from_node_ok", files => Files}),
+            case zip:unzip(DataBin, [{cwd, emqx:data_dir()}]) of
+                {ok, []} ->
+                    ?SLOG(debug, #{node => Node, msg => "sync_data_from_node_ignore"});
+                {ok, Files} ->
+                    ?SLOG(debug, #{node => Node, msg => "sync_data_from_node_ok", files => Files})
```
qzhuyan, line 203: do we need to handle the error here?

PR #9434, opened 2022-11-28
refactor: return `204` instead of `200` for `PUT /authenticator/:id`

sstrigler:
Fixes [EMQX-7982](https://emqx.atlassian.net/browse/EMQX-7982)

## PR Checklist

- [x] Sync with dashboard team

### DIFF
--- a/apps/emqx_authn/src/emqx_authn_api.erl
+++ b/apps/emqx_authn/src/emqx_authn_api.erl
@@ -1006,11 +1006,10 @@ update_authenticator(ConfKeyPath, ChainName, AuthenticatorID, Config) ->
         )
     of
         {ok, #{
-            post_config_update := #{emqx_authentication := #{id := ID}},
-            raw_config := AuthenticatorsConfig
+            post_config_update := #{emqx_authentication := #{id := _ID}},
+            raw_config := _AuthenticatorsConfig
         }} ->
-            {ok, AuthenticatorConfig} = find_config(ID, AuthenticatorsConfig),
-            {200, maps:put(id, ID, convert_certs(fill_defaults(AuthenticatorConfig)))};
+            {204};
         {error, {_PrePostConfigUpdate, emqx_authentication, Reason}} ->
             serialize_error(Reason);
         {error, Reason} ->
--- a/changes/v5.0.12-zh.md
+++ b/changes/v5.0.12-zh.md
@@ -4,6 +4,8 @@
 
 - 通过 `node.global_gc_interval = disabled` 来禁用全局垃圾回收 [#9418](https://github.com/emqx/emqx/pull/9418)。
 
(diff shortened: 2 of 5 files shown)


changes/v5.0.12-zh.md
```diff
@@ -4,6 +4,8 @@
 
 - 通过 `node.global_gc_interval = disabled` 来禁用全局垃圾回收 [#9418](https://github.com/emqx/emqx/pull/9418)。
 
+- [FIXME] - Return `204` instead of `200` for `PUT /authenticator/:id` [#9434](https://github.com/emqx/emqx/pull/9434/).
```
lafirest, line 7: 现在，`PUT /authenticator/:id` 将会返回 `204` 而不再是 `200` [#9434](https://github.com/emqx/emqx/pull/9434/)。

qzhuyan (approved):

apps/emqx_authn/src/emqx_authn_api.erl
```diff
@@ -1006,11 +1006,10 @@ update_authenticator(ConfKeyPath, ChainName, AuthenticatorID, Config) ->
         )
     of
         {ok, #{
-            post_config_update := #{emqx_authentication := #{id := ID}},
-            raw_config := AuthenticatorsConfig
+            post_config_update := #{emqx_authentication := #{id := _ID}},
+            raw_config := _AuthenticatorsConfig
         }} ->
-            {ok, AuthenticatorConfig} = find_config(ID, AuthenticatorsConfig),
-            {200, maps:put(id, ID, convert_certs(fill_defaults(AuthenticatorConfig)))};
```
zmstone, line 1013: I'm afraid this has to be reverted.

zmstone: this is not an echo, the `convert_certs` will store files to disk, and return stored paths.
and `fill_defaults` will fill default values for the missing fields in the request.

sstrigler: I've been reading through the code a bit now. `convert_certs` does have side-effects, it *reads* `pem` files and checks if they are valid. But it doesn't look like it's creating anything anywhere. 

But `update_config` might trigger creating of ssl related files. Couldn't quite trace it back but there is `emqx_authentication_config:pre_config_update` which later calls some other `convert_certs` within its own module. And this one, deep down the line does a call to `save_pem_file` eventually. 

I just assume that through some callbacks `update_config` does call `emqx_authentication_config:pre_config_update` and as such create those `pem` file(s). Means it should happen with or without that call to that *local* `convert_certs`. 

It still might be necessary to do call those checks to trigger a crash just to indicate there's something wrong, but in that case a proper error handling would we preferable. And it should already happen within `update_config` I'd think.

PR #9438, opened 2022-11-28
build: detach 4.3 branches from 4.4 branches

zmstone:


scripts/rel/cut4x.sh
```diff
@@ -10,13 +10,13 @@ usage() {
     cat <<EOF
 $0 RELEASE_GIT_TAG [option]
 RELEASE_GIT_TAG is a 'v*' or 'e*' tag for example:
-  v4.3.21
+  v4.4.10
```
thalesmg, line 13: Shouldn't those changes target `main-v4.4`?

zmstone: that is: the `sync-remotes.sh` script in main-v4.4 is still the old version *before* this is commit is merged.
after that, running it again will not pull v43 any more.

thalesmg (approved):

PR #9446, opened 2022-11-29
1129 emqx 7958 add cover

zmstone:


thalesmg (approved):

zmstone:
this is an internal requirement, so no release note update.

PR #9458, opened 2022-11-30
feat(bridge): add Redis bridge

savonarola:
Add Redis egress bridge

## PR Checklist

- [ ] ~~Change log has been added to `changes/` dir~~
- [ ]  ~~For EMQX 4.x: `appup` files updated (execute `scripts/update-appup.sh emqx`)~~
- [ ] ~~In case of non-backward compatible changes, reviewer should check this item as a write-off, and add details in **Backward Compatibility** section~~

### DIFF
--- a/.ci/docker-compose-file/toxiproxy.json
+++ b/.ci/docker-compose-file/toxiproxy.json
@@ -22,5 +22,12 @@
     "listen": "0.0.0.0:3307",
     "upstream": "mysql-tls:3306",
     "enabled": true
+  },
+
+  {
+    "name": "redis_single_tcp",
+    "listen": "0.0.0.0:6379",
+    "upstream": "redis:6379",
+    "enabled": true
   }
 ]
--- a/lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
+++ b/lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
@@ -1,0 +1,493 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2022 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_ee_bridge_redis_SUITE).
+
+-compile(nowarn_export_all).
+-compile(export_all).
+
+-include_lib("eunit/include/eunit.hrl").
+-include_lib("common_test/include/ct.hrl").
+-include_lib("snabbkaffe/include/snabbkaffe.hrl").
+
+-include_lib("emqx_bridge/include/emqx_bridge.hrl").
+
+%%------------------------------------------------------------------------------
+%% CT boilerplate
+%%------------------------------------------------------------------------------
+
+-define(REDIS_TOXYPROXY_CONNECT_CONFIG, #{
+    <<"server">> => <<"toxiproxy:6379">>
+}).
+
+-define(COMMON_REDIS_OPTS, #{
+    <<"password">> => <<"public">>,
+    <<"command_template">> => [<<"RPUSH">>, <<"MSGS">>, <<"${payload}">>],
+    <<"local_topic">> => <<"local_topic/#">>
+}).
+
+-define(BATCH_SIZE, 5).
+
+-define(PROXY_HOST, "toxiproxy").
+-define(PROXY_PORT, "8474").
+
+all() -> [{group, redis_types}, {group, rest}].
+
+groups() ->
+    ResourceSpecificTCs = [t_create_delete_bridge],
+    TCs = emqx_common_test_helpers:all(?MODULE) -- ResourceSpecificTCs,
+    TypeGroups = [
+        {group, redis_single},
+        {group, redis_sentinel},
+        {group, redis_cluster}
+    ],
+    BatchGroups = [
+        {group, batch_on},
+        {group, batch_off}
+    ],
+    [
+        {rest, TCs},
+        {redis_types, [
+            {group, tcp},
+            {group, tls}
+        ]},
+        {tcp, TypeGroups},
+        {tls, TypeGroups},
+        {redis_single, BatchGroups},
+        {redis_sentinel, BatchGroups},
+        {redis_cluster, BatchGroups},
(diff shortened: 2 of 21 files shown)


lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
```diff
@@ -0,0 +1,412 @@
...
+    <<"command_template">> => [<<"RPUSH">>, <<"MSGS">>, <<"${payload}">>],
+    <<"local_topic">> => <<"local_topic/#">>
+}).
+
+-define(BATCH_SIZE, 5).
+
+-define(PROXY_HOST, "toxiproxy").
+-define(PROXY_PORT, "8474").
+
+all() -> emqx_common_test_helpers:all(?MODULE).
```
thalesmg, line 34: Nit: I wonder if we could repeat the tests in a matrix-like fashion for each of the 3 Redis types using groups, so it's easier to select only a single type for quicker test iteration, and when something fails it's easier to debug which config caused it?

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
```diff
@@ -0,0 +1,412 @@
...
+            config => #{
+                <<"query_mode">> => <<"sync">>,
+                <<"enable_batch">> => <<"false">>,
+                <<"enable_queue">> => <<"false">>
+            }
+        },
+        #{
+            batch => true,
+            config => #{
+                <<"query_mode">> => <<"async">>,
```
thalesmg, line 399: This probably won't take effect, since the callback mode is [`always_sync`](https://github.com/emqx/emqx/pull/9458/files?diff=split&w=1#diff-5de9b83ca2aad13ff49b95d4418ae2db3476366de33f4ed374bfe5a76eba7164R25)?

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
```diff
@@ -0,0 +1,412 @@
...
+        },
+
+        %% check export by rule
+        {ok, _} = emqx_rule_engine:create_rule(RuleConf),
+        _ = check_resource_queries(<<"t/test">>, IsBatch),
+        ok = emqx_rule_engine:delete_rule(RuleId),
+
+        %% check export through local topic
+        _ = check_resource_queries(<<"local_topic/test">>, IsBatch),
+
```
thalesmg, line 111: Suggestion: maybe it's good to also query Redis directly to check that the data is persisted there in the expected format?

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
savonarola: Yes, but want to see whath happens :)

scripts/ct/run.sh
```diff
@@ -111,6 +111,10 @@ for dep in ${CT_DEPS}; do
                      '.ci/docker-compose-file/docker-compose-redis-sentinel-tcp.yaml'
                      '.ci/docker-compose-file/docker-compose-redis-sentinel-tls.yaml' )
             ;;
+        redis_cluster)
+            FILES+=( '.ci/docker-compose-file/docker-compose-redis-cluster-tcp.yaml'
```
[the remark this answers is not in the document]
savonarola: We can't use toxiproxy with cluster in a natural way: we may only break connections to the seed nodes, but then the cluster tells the client real IPs
As far as I know, there is still a PR in Redis 7 to allow cluster nodes to identify themselves to a client by domain names.

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
savonarola: First I tried to organize the tests with groups 😅 like in influx tests, but things started to seem a bit too implicit for me, so I made an attempt to use a more explicit approach

.ci/docker-compose-file/toxiproxy.json
```diff
@@ -10,5 +10,12 @@
     "listen": "0.0.0.0:8087",
     "upstream": "influxdb_tls:8086",
     "enabled": true
+  },
+
+  {
```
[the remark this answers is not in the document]
savonarola: I aimed to test the bridge behaviour on recoverable errors in general, i.e. that the messages are batched, replayed, and finally get into the database after the resource is healthy; and in a lesser degree to test the restoration of the actual connection...
Additionally, it is not so straightforward to hide sentinel or cluster behind the proxy; this shuold bring much complexity to the setup.

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
savonarola: ✅

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
savonarola: ✅

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_redis_SUITE.erl
thalesmg: Good point.  =)

my-key-file.pem
```diff
@@ -0,0 +1,6 @@
+
```
thalesmg, line 1: Is this supposed to be commited?  :thinking:

thalesmg (approved):
There's just a couple more things to test (what happens when the [query returns an error](https://coveralls.io/builds/54877341/source?filename=lib-ee%2Femqx_ee_connector%2Fsrc%2Femqx_ee_connector_redis.erl#L115) and status checks), but they could be done in another moment.  :beers:

thalesmg (approved):

PR #9461, opened 2022-12-01
refactor: move metrics out of /rules(/:id) to /rules/:id/metrics

sstrigler:
Fixes [EMQX-7988](https://emqx.atlassian.net/browse/EMQX-7988)

Summary: 
 * you have to query `GET /rules/:id/metrics` explicitly in order to get metrics to a given rule. To reset metrics of a given you resource you `PUT /rules/:id/metrics/reset`. 
 * `/rules/:id/metrics/reset` now returns `204` instead of `200` to comply with rest of API
 * for all queries that include `:id` a `404` is returned in case rule does not exist

## PR Checklist

- [x] Sync with dasboard team

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
@@ -31,12 +31,18 @@
 -export([api_spec/0, paths/0, schema/1, namespace/0]).
 
 %% API callbacks
--export(['/rule_events'/2, '/rule_test'/2, '/rules'/2, '/rules/:id'/2, '/rules/:id/reset_metrics'/2]).
+-export([
+    '/rule_events'/2,
+    '/rule_test'/2,
+    '/rules'/2,
+    '/rules/:id'/2,
+    '/rules/:id/metrics'/2,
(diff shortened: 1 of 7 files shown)


apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
```diff
@@ -363,15 +395,36 @@ param_path_id() ->
...
+    end.
+
+'/rules/:id/metrics/reset'(put, #{bindings := #{id := Id}}) ->
+    case emqx_rule_engine:get_rule(Id) of
+        {ok, _Rule} ->
+            case emqx_rule_engine_proto_v1:reset_metrics(Id) of
+                ok ->
+                    {204};
+                Failed ->
+                    {400, #{
```
zmstone, line 421: hard to reason why this should be a 400,
isn't this 500 ?

sstrigler: You're right, doesn't make sense to be a 400. Also things get a lot easier by just letting it crash here.

zmstone (approved):

PR #9462, opened 2022-12-01
feat(docs): generate config docs divided in subsections

thalesmg:
Instead of generating a single huge Markdown file with all the configurations, which is then manually copied over to `emqx-docs.git` as `admin/cfg.md`, here we split the docs in a few subsections:

- EMQX
- Authentication
- Authorization
- Bridges
- Rule Engine
- And the rest

Only the "EMQX" subsection contains the preamble that explains the configuration (Hocon, Layered structure, Syntax, etc.).

Fixes [EMQX-8330](https://emqx.atlassian.net/browse/EMQX-8330)

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/doc_generation/emqx_conf_authn_section.erl
+++ b/apps/emqx_conf/src/doc_generation/emqx_conf_authn_section.erl
@@ -1,0 +1,59 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2022 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


thalesmg:
Added zips with sample output files in the Jira ticket comments section.

zmstone:
An alternative: if we take the `schema.json` as input and split it with another script, then it doesn't have to be Erlang.

zmstone:
> An alternative: if we take the `schema.json` as input and split it with another script, then it doesn't have to be Erlang.

Just created this repo: https://github.com/emqx/emqx-docgen

zmstone:
Discussion: the shared schema, such as `broker:ssl_client_opts` is repeatedly generated in each .md file. Does it make more sense to put them in a shared doc and cross-file reference from the places where they are used ?

zmstone:
Need a gateways.md

thalesmg:
> > An alternative: if we take the `schema.json` as input and split it with another script, then it doesn't have to be Erlang.
> 
> Just created this repo: https://github.com/emqx/emqx-docgen

So the idea is to don't touch the schema creation here in this repo, but just the slicing and dicing of the generated json in the new repo?

thalesmg:
> Discussion: the shared schema, such as `broker:ssl_client_opts` is repeatedly generated in each .md file. Does it make more sense to put them in a shared doc and cross-file reference from the places where they are used ?

I think that, if such schemas appear in multiple subsections, it might be convenient to refer to them in the same document than to follow a link.

thalesmg:
See https://github.com/emqx/emqx/pull/9485

PR #9474, opened 2022-12-04
fix: load bootstrap file when no bootstrap user

zhongwencool:
When doing a hot upgrade or rolling upgrade, the `mqtt_app` table is not empty, Leading to skip import bootstrap users.

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_auth.erl
+++ b/apps/emqx_management/src/emqx_mgmt_auth.erl
@@ -92,19 +92,21 @@ init_bootstrap_apps() ->
     end.
 
 need_bootstrap() ->
-    {atomic, Res} = mnesia:transaction(fun() -> bootstrap_apps() =:= [] end),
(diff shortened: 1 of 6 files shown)


apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -81,21 +83,39 @@ add_default_app() ->
...
+clear_bootstrap_apps() ->
+    {atomic, ok} =
+        mnesia:transaction(fun() ->
+            DeleteFun = fun(A) -> mnesia:delete_object(A) end,
+            lists:foreach(DeleteFun, bootstrap_apps())
+                           end),
+    ok.
+
+bootstrap_apps() ->
+    mnesia:match_object(mqtt_app, #mqtt_app{desc = ?BOOTSTRAP_TAG, _ = '_'}, read).
```
lafirest, line 107: nit: maybe mnesia:select/4 which can select with a limit will be better

PR #9478, opened 2022-12-05
fix: return 400 if node in query doesn't look like a known node

sstrigler:
Fixes [EMQX-7996](https://emqx.atlassian.net/browse/EMQX-7996)

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_trace.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_trace.erl
@@ -451,10 +451,10 @@ update_trace(put, #{bindings := #{name := Name}}) ->
 %% if HTTP request headers include accept-encoding: gzip and file size > 300 bytes.
 %% cowboy_compress_h will auto encode gzip format.
 download_trace_log(get, #{bindings := #{name := Name}, query_string := Query}) ->
-    case parse_node(Query, undefined) of
-        {ok, Node} ->
-            case emqx_trace:get_trace_filename(Name) of
-                {ok, TraceLog} ->
+    case emqx_trace:get_trace_filename(Name) of
+        {ok, TraceLog} ->
+            case parse_node(Query, undefined) of
+                {ok, Node} ->
                     TraceFiles = collect_trace_file(Node, TraceLog),
                     ZipDir = emqx_trace:zip_dir(),
                     Zips = group_trace_file(ZipDir, TraceLog, TraceFiles),
@@ -474,10 +474,10 @@ download_trace_log(get, #{bindings := #{name := Name}, query_string := Query}) -
                     },
                     {200, Headers, {file_binary, ZipName, Binary}};
                 {error, not_found} ->
-                    ?NOT_FOUND(Name)
+                    ?BAD_REQUEST('NODE_ERROR', <<"Node not found">>)
             end;
         {error, not_found} ->
-            ?BAD_REQUEST('NODE_ERROR', <<"Node not found">>)
+            ?NOT_FOUND(Name)
     end.
 
 group_trace_file(ZipDir, TraceLog, TraceFiles) ->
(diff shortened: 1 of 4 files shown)


zhongwencool (dismissed):

changes/v5.0.12-zh.md
```diff
@@ -22,3 +22,5 @@
 - 在遥测功能未开启时，通过 /telemetry/data 请求其数据，将会返回 404 [#9464](https://github.com/emqx/emqx/pull/9464)。
 
 - 修复了一些 MQTT 协议包的潜在解析错误 [#9477](https://github.com/emqx/emqx/pull/9477)。
+
+- [FIXME] Return `400` if query param `node` is not a known in `/trace/:id/download?node={node}` [#9478](https://github.com/emqx/emqx/pull/9478).
```
sstrigler, line 26: @lafirest please! 🙇

apps/emqx_management/src/emqx_mgmt_api_trace.erl
```diff
@@ -450,32 +451,37 @@ update_trace(put, #{bindings := #{name := Name}}) ->
...
-            %% TODO use file replace file_binary.(delete file after send is not ready now).
-            {ok, Binary} = file:read_file(ZipFile),
-            ZipName = filename:basename(ZipFile),
-            _ = file:delete(ZipFile),
-            Headers = #{
-                <<"content-type">> => <<"application/x-zip">>,
-                <<"content-disposition">> => iolist_to_binary("attachment; filename=" ++ ZipName)
-            },
-            {200, Headers, {file_binary, ZipName, Binary}};
+    case parse_node(Query, []) of
```
zmstone, line 454: default to `[]` ?

apps/emqx_management/src/emqx_mgmt_api_trace.erl
```diff
@@ -450,32 +451,37 @@ update_trace(put, #{bindings := #{name := Name}}) ->
...
+        {ok, Node} ->
+            case emqx_trace:get_trace_filename(Name) of
+                {ok, TraceLog} ->
+                    case collect_trace_file(Node, TraceLog) of
+                        {error, not_found} ->
+                            ?BAD_REQUEST('NODE_ERROR', <<"Node not found">>);
```
zmstone, line 460: is this really a "Node not found" error ?
the node is found, but failed to do RPC, should be a 500 instead?

sstrigler: That's misleading, the node was not "found" - at that point we just know it is a known atom. But it could be sth like `cowboy_rest` and would still be treated like a node name.

sstrigler: I just realised, in `parse_node` we could always check if `Node` is a member of `mria_mnesia:running_nodes()`. Would that make sense?

zmstone (approved):

apps/emqx_management/src/emqx_mgmt_api_trace.erl
```diff
@@ -633,8 +630,12 @@ read_file(Path, Offset, Bytes) ->
 parse_node(Query, Default) ->
     try
         case maps:find(<<"node">>, Query) of
-            error -> {ok, Default};
-            {ok, Node} -> {ok, binary_to_existing_atom(Node)}
+            error ->
+                {ok, Default};
+            {ok, NodeBin} ->
+                Node = binary_to_existing_atom(NodeBin),
+                true = lists:member(Node, mria_mnesia:running_nodes()),
```
thalesmg, line 637: This'll turn into a 500 if the node is not part of the cluster or offline.  Maybe it could return an error that becomes 400?

sstrigler: It's turned into `{error, not_found}` by try-catch and then later turned into `400`

thalesmg (approved):

apps/emqx_management/src/emqx_mgmt_api_trace.erl
sstrigler: Maybe bad style, lazy me trying to save a few lines since we have the try-catch anyway. :/

PR #9496, opened 2022-12-07
Fix `emqx_authentication` hook cooperation with other hooks

savonarola:
Fix `emqx_authenticator` cooperation with other `'client.authenticate'` hooks.

Also refactor authn metrics, because  there can be authenticators other than `emqx_authentication`.
Now authn/authz metrics look uniformly.

Fixes https://askemq.com/t/topic/2742/32

## PR Checklist

- [ ] ~~For EMQX 4.x: `appup` files updated (execute `scripts/update-appup.sh emqx`)~~
- [ ] ~~In case of non-backward compatible changes, reviewer should check this item as a write-off, and add details in **Backward Compatibility** section~~

### DIFF
--- a/apps/emqx/src/emqx_authentication.erl
+++ b/apps/emqx/src/emqx_authentication.erl
@@ -228,7 +228,6 @@ when
 -spec pre_hook_authenticate(emqx_types:clientinfo()) ->
     ok | continue | {error, not_authorized}.
 pre_hook_authenticate(#{enable_authn := false}) ->
-    inc_authenticate_metric('authentication.success.anonymous'),
     ?TRACE_RESULT("authentication_result", ok, enable_authn_false);
 pre_hook_authenticate(#{enable_authn := quick_deny_anonymous} = Credential) ->
     case maps:get(username, Credential, undefined) of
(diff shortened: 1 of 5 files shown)


apps/emqx/src/emqx_authentication.erl
```diff
@@ -649,7 +637,7 @@ handle_create_authenticator(Chain, Config, Providers) ->
     end.
 
 do_authenticate(_ChainName, [], _) ->
-    {stop, {error, not_authorized}};
+    {ok, {error, not_authorized}};
```
savonarola, line 640: These two changed lines in `do_authenticate/3` is the major part of the fix. We do not stop the hook chain fold, but return the new acc.
In case there are no other hooks, the behaviour is the same.
In case there are other hooks with lower priority, the can override the auth result (acc).

apps/emqx/test/emqx_authentication_SUITE.erl
```diff
@@ -500,6 +533,92 @@ t_convert_certs(Config) when is_list(Config) ->
...
+    ?assertAuthFailureForUser(ignore),
+
+    %% lower-priority hook can overrride auth result,
+    %% because emqx_authentication permits/denies with {ok, ...}
+    ?assertAuthSuccessForUser(hook_user_good),
+    ?assertAuthFailureForUser(hook_user_bad),
+    ?assertAuthSuccessForUser(hook_user_finally_good),
+    ?assertAuthFailureForUser(hook_user_finally_bad),
+
+    ok = unhook();
```
thalesmg, line 604: Nit: unhook in `end` callback?

thalesmg (dismissed):

apps/emqx/test/emqx_authentication_SUITE.erl
savonarola: We need to unhook once more during the test, so I left all the unhooks in the body.

zmstone (approved):

apps/emqx/src/emqx_authentication.erl
```diff
@@ -673,7 +661,7 @@ do_authenticate(
                 _ ->
                     ok
             end,
-            {stop, Result}
+            {ok, Result}
```
zmstone, line 664: shouldn't we stop if result is error ?

apps/emqx/src/emqx_authentication.erl
zmstone: should return ignore from here ?

apps/emqx/src/emqx_authentication.erl
savonarola: @zmstone Yes if we want to forbid to override `emqx_authentication` result. Changed the behaviour to stop.

apps/emqx/src/emqx_authentication.erl
savonarola: Seems that no, because it will change the behaviour. Currently if we have several providers all of which return `ignore`, authentication fails.
Returning `ignore` here will make authentication successful.

PR #9514, opened 2022-12-08
refactor: rename `/users/{name}/change_pwd` to `/users/{name}/password`

sstrigler:
Fixes [EMQX-8373](https://emqx.atlassian.net/browse/EMQX-8373)

## PR Checklist

- [ ] sync with dasboard team

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard.app.src
+++ b/apps/emqx_dashboard/src/emqx_dashboard.app.src
@@ -2,7 +2,7 @@
 {application, emqx_dashboard, [
     {description, "EMQX Web Dashboard"},
     % strict semver, bump manually!
-    {vsn, "5.0.8"},
+    {vsn, "5.0.9"},
     {modules, []},
     {registered, [emqx_dashboard_sup]},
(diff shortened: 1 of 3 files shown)


lafirest (commented):
Good job, do we need a change description?

sstrigler:
> Good job, do we need a change description?

I think that I need to sync with the dashboard team at least. Where would a change description go (or what does it mean)?

lafirest:
> > Good job, do we need a change description?
> 
> I think that I need to sync with the dashboard team at least. Where would a change description go (or what does it mean)?

I mean this one `Change log has been added to changes/ dir`, but the modification is small so that this may not be needed

sstrigler:
> 



> > > Good job, do we need a change description?
> > 
> > 
> > I think that I need to sync with the dashboard team at least. Where would a change description go (or what does it mean)?
> 
> I mean this one `Change log has been added to changes/ dir`, but the modification is small so that this may not be needed

Oh, of course. I'll add one once it's clear it's not going into 5.0.12. Which is basically clear now. But I've also posted some thoughts on slack, there might be more we want to do.

lafirest (approved):

PR #9548, opened 2022-12-14
fix(retainer_mnesia): used undefined mria api

JimMoen:


JimMoen:
@thalesmg I'll try re-build later, Thanks.

PR #9564, opened 2022-12-15
feat: implement kafka consumer (ee5.0)

thalesmg:
https://emqx.atlassian.net/browse/EMQX-8386

Also, updates `kafka_protocol`, `wolff` and `brod`.

# Testing locally

The consumer and producer tests works locally, by running Kafka in a container and the tests in the local machine.  Except for Kerberos authentication, which I couldn't make it work without running the tests inside a container.

```sh
{
  env \
    docker-compose \
    -f .ci/docker-compose-file/docker-compose-toxiproxy.yaml \
    -f .ci/docker-compose-file/docker-compose.yaml \
    -f .ci/docker-compose-file/docker-compose-kafka.yaml \
    up --remove-orphans

  env \
    docker-compose \
    -f .ci/docker-compose-file/docker-compose-toxiproxy.yaml \
    -f .ci/docker-compose-file/docker-compose.yaml \
    -f .ci/docker-compose-file/docker-compose-kafka.yaml \
    down --remove-orphans
}

env \
    PROFILE=emqx-enterprise \
    EMQX_LOG__CONSOLE_HANDLER__LEVEL=info \
    EMQX_BRIDGES__KAFKA_CONSUMER__kconsumer__ENABLE=true \
    EMQX_BRIDGES__KAFKA_CONSUMER__kconsumer__BOOTSTRAP_HOSTS='"localhost:9192"' \
    EMQX_BRIDGES__KAFKA_CONSUMER__kconsumer__KAFKA__TOPIC=test-topic-three-partitions \
    EMQX_BRIDGES__KAFKA_CONSUMER__kconsumer__MQTT__TOPIC="mqtt/topic" \
    make emqx-enterprise-elixir quickrun

env \
  ENABLE_COVER_COMPILE=1 \
  KAFKA_PLAIN_HOST=localhost \
  KAFKA_PLAIN_PORT=9192 \
  KAFKA_SASL_PLAIN_HOST=localhost \
  KAFKA_SASL_PLAIN_PORT=9193 \
  KAFKA_SSL_HOST=localhost \
  KAFKA_SSL_PORT=9194 \
  KAFKA_SASL_SSL_HOST=localhost \
  KAFKA_SASL_SSL_PORT=9195 \
  PROXY_HOST=localhost \
  PROXY_PORT=8474 \
  CI_SHARED_SECRET_PATH=/tmp/emqx-ci/emqx-shared-secret \
  LOGLEVEL=notice \
  make ct-suite \
  SUITE=lib-ee/emqx_ee_bridge/test/emqx_bridge_impl_kafka_consumer_SUITE \
  GROUP='[plain],[ssl],[sasl_plain,sasl_auth_plain],[sasl_ssl,sasl_auth_scram256]'
```

## PR Checklist

## Backward Compatibility

This changes changes the configuration for the Kafka Producer as well (they are now split).  This means that the frontend will have to be adapted for the new config.


apps/emqx_bridge/src/emqx_bridge.erl
```diff
@@ -115,12 +116,12 @@ load_hook(Bridges) ->
         maps:to_list(Bridges)
     ).
 
-do_load_hook(Type, #{local_topic := _}) when ?EGRESS_DIR_BRIDGES(Type) ->
+do_load_hook(Type, #{local_topic := LocalTopic}) when
+    ?EGRESS_DIR_BRIDGES(Type) andalso is_binary(LocalTopic)
+->
```
JimMoen, line 121: There may be multiple bridges enabled in an application.
Some of them may not have `local_topic` configuration, so it may not be necessary to determine whether to load in the `load_hook` stage.
The issue should be fixed in `e5.0.0` instead of postponed to `e5.0.1`.
see also #9686

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -39,24 +39,24 @@ emqx_ee_bridge_kafka {
...
-            zh: "本地 MQTT 数据源和 Kafka 桥接的配置。"
+            en: "Kafka Producer configuration."
+            zh: "Kafka Producer的配置。"
         }
         label {
-            en: "MQTT to Kafka"
-            zh: "MQTT 到 Kafka"
+            en: "Kafka Producer"
+            zh: "卡夫卡制片人"
         }
```
JimMoen, line 50:
```
        desc {
            en: "Kafka Producer configuration."
            zh: "Kafka Producer 配置。"
        }
        label {
            en: "Kafka Producer"
            zh: "Kafka Producer"
        }
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -468,4 +468,128 @@ emqx_ee_bridge_kafka {
...
     }
+
+    kafka_consumer {
+        desc {
+            en: "Kafka Consumer configuration."
+            zh: "Kafka Consumer的配置。"
+        }
+        label {
+            en: "Kafka Consumer"
+            zh: "卡夫卡消费者"
```
JimMoen, line 479:
```
en: "Kafka Consumer"
zh: "Kafka Consumer"
```
Btw, there are two same lables `Kafka Consumer`, see line: 498

thalesmg:
This will probably have lots of conflicts when the current `master` gets synced into `release-e501`, so I've put this in draft until that happens.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -0,0 +1,308 @@
...
+        children_ids := ChildrenIds,
+        kafka_client_id := ClientID
+    } = State,
+    stop_consumer_pool(ChildrenIds),
+    stop_client(ClientID),
+    ok.
+
+-spec on_get_status(manager_id(), state()) -> connected.
+on_get_status(_InstanceID, _State) ->
+    connected.
```
thalesmg, line 177: - [ ] FIXME: should check if all consumers are connected or if client can access the leaders of all partitions (how to list partitions?).

lib-ee/emqx_ee_bridge/rebar.config
```diff
@@ -2,7 +2,7 @@
 {deps, [ {wolff, {git, "https://github.com/kafka4beam/wolff.git", {tag, "1.7.5"}}}
        , {kafka_protocol, {git, "https://github.com/kafka4beam/kafka_protocol.git", {tag, "4.1.2"}}}
        , {brod_gssapi, {git, "https://github.com/kafka4beam/brod_gssapi.git", {tag, "v0.1.0-rc1"}}}
-       , {brod, {git, "https://github.com/kafka4beam/brod.git", {tag, "3.16.7"}}}
+       , {brod, {git, "https://github.com/thalesmg/brod.git", {tag, "3.16.8-tmg0"}}}
```
thalesmg, line 5: TODO: change to real tagged version

lib-ee/emqx_ee_bridge/test/emqx_bridge_impl_kafka_consumer_SUITE.erl
```diff
@@ -0,0 +1,1347 @@
...
+    Params = KafkaConfig#{<<"type">> => TypeBin, <<"name">> => Name},
+    Path = emqx_mgmt_api_test_util:api_path(["bridges", BridgeId]),
+    AuthHeader = emqx_mgmt_api_test_util:auth_header_(),
+    ct:pal("updating bridge (via http): ~p", [Params]),
+    Res =
+        case emqx_mgmt_api_test_util:request_api(put, Path, "", AuthHeader, Params) of
+            {ok, Res0} -> {ok, emqx_json:decode(Res0, [return_maps])};
+            Error -> Error
+        end,
+    ct:pal("bridge creation result: ~p", [Res]),
```
savonarola, line 600: Nit: update result?

lib-ee/emqx_ee_bridge/test/emqx_bridge_impl_kafka_consumer_SUITE.erl
```diff
@@ -0,0 +1,1347 @@
...
+            ?match_n_events(N, #{?snk_kind := kafka_consumer_subscriber_init}),
+            Timeout
+        ),
+    ok.
+
+%% kinda hacky, but for yet unknown reasons kafka/brod seem a bit
+%% flaky about when they decide truly consuming the messages...
+%% `Period' should be greater than the `sleep_timeout' of the consumer
+%% (default 1 s).
+ping_until_healthy(_Config, _Period, Timeout) when Timeout =< 0 ->
```
savonarola, line 688: Nit: `Period` -> `Interval`?

lib-ee/emqx_ee_bridge/test/emqx_bridge_impl_kafka_consumer_SUITE.erl
```diff
@@ -0,0 +1,1347 @@
...
+                begin
+                    X = emqx_guid:to_hexstr(emqx_guid:gen()),
+                    #{
+                        key => X,
+                        value => X
+                    }
+                end
+             || _ <- lists:seq(1, NumMsgs)
+            ],
+            %% emqx:subscribe(MQTTTopic),
```
savonarola, line 1107: Nit: remove?

lib-ee/emqx_ee_bridge/test/emqx_bridge_impl_kafka_producer_SUITE.erl
```diff
@@ -500,39 +513,49 @@ publish_helper(
...
         },
         Conf0
     ),
-
-    emqx_bridge_resource:create(kafka, erlang:list_to_atom(Name), Conf, #{}),
+    {ok, _} = emqx_bridge:create(
+        <<"kafka_producer">>, list_to_binary(Name), Conf
+    ),
     %% To make sure we get unique value
     timer:sleep(1),
```
savonarola, line 534: Just curious, why not `erlang:unique_integer()` 🤔

apps/emqx/test/emqx_common_test_helpers.erl
```diff
@@ -614,25 +615,62 @@ emqx_cluster(Specs0, CommonOpts) ->
...
+            ok;
+        {error, {already_running, _}} ->
+            ok;
+        Other ->
+            throw(Other)
+    end,
     pong = net_adm:ping(Node),
+    put_peer_mod(Node, SlaveMod),
     setup_node(Node, Opts),
+    ok = snabbkaffe:forward_trace(Node),
```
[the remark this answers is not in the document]
thalesmg: I think so, since this is a test node.  If we are checking traces, we probably want to know what happened in the peer nodes.

If we want only a single node's trace, one may use the `?of_node/2` macro.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -0,0 +1,378 @@
...
+            client => ClientID,
+            group_id => GroupID,
+            topics => [KafkaTopic],
+            cb_module => ?MODULE,
+            init_data => InitialState,
+            message_type => message,
+            consumer_config => ConsumerConfig,
+            group_config => GroupConfig
+        },
+    %% group_consumer_v2 spawns one worker for each partition
```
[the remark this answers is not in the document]
thalesmg: :speak_no_evil: 

Indeed, it's a bit obscure just looking at the following lines, sorry.

This is a justification for `emqx_ee_bridge_kafka_consumer_sup:start_child`.  It starts a single `brod_group_subscriber_v2` worker without an option to spawn a pool of those because these subscribers manage their sub-workers depending on the assigned partitions.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -0,0 +1,378 @@
...
+    on_start/2,
+    on_stop/2,
+    on_get_status/2
+]).
+
+%% `brod_group_consumer' API
+-export([
+    init/2,
+    handle_message/2
+]).
```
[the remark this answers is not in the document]
thalesmg: It's actually the `brod_group_subscriber_worker` "behavior".  But that module does not declare any `-callback` module attributes, so I think it wouldn't work currently :see_no_evil: 

I agree that we should add the `-behaviour` module attribute here, once we add the `-callback`s to `brod.

lib-ee/emqx_ee_bridge/test/emqx_bridge_impl_kafka_producer_SUITE.erl
thalesmg: Good question.  This was already like this before, but I suppose we could change it to use `unique_integer`.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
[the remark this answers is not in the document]
thalesmg: I'll improve the comment.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -0,0 +1,378 @@
...
+                        false
+                end,
+                Results
+            ),
+    WorkersAlive = are_subscriber_workers_alive(SubscriberId),
+    case AllLeadersOk andalso WorkersAlive of
+        true ->
+            connected;
+        false ->
+            disconnected
```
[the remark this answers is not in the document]
thalesmg: It's a bit confusing whether to choose `disconnected` and `connecting` in this case, I think.  :thinking: 

From the [docs](https://github.com/emqx/emqx/blob/43bedf1d4c1ab4aaa53d74092efaf5ed01c1bd67/apps/emqx_bridge/i18n/emqx_bridge_schema.conf#L38-L43), I understand that `connecting` should be used as the status before any health checks are made.  Here, we have checked a few things (connection to leader, worker pids), so I understand that returning `disconnected` seems to agree better with those descriptions.

apps/emqx_bridge/src/emqx_bridge.erl
thalesmg: Apparently fixed in https://github.com/emqx/emqx/pull/9687

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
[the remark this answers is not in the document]
thalesmg: Please check if the new comment is clearer now.  :smiley_cat:

lib-ee/emqx_ee_bridge/test/emqx_bridge_impl_kafka_consumer_SUITE.erl
thalesmg: I thought those were equivalent (like frequency <-> period/interval)?

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -0,0 +1,378 @@
...
+                        Message#kafka_message.value
+                end,
+            EncodedPayload = emqx_json:encode(Payload),
+            MQTTMessage = emqx_message:make(ResourceId, MQTTQoS, MQTTTopic, EncodedPayload),
+            _ = emqx:publish(MQTTMessage),
+            emqx:run_hook(Hookpoint, [FullMessage]),
+            emqx_resource_metrics:received_inc(ResourceId),
+            %% note: just `ack' does not commit the offset to the
+            %% kafka consumer group.
+            {ok, commit, State}
```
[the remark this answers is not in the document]
thalesmg: I'll that parameter as a configuration.

But just for my own understanding: what do you mean by duplicates?  The chance of having successfully processed a few Kafka messages and then the worker dying before commiting the offset?

I think the ack/commit is asynchronous in this case because of this sequence: [1](https://github.com/kafka4beam/brod/blob/ef97325545eed0ae406733bf177090b7a6a192b3/src/brod_group_subscriber_worker.erl#L80) -> [2](https://github.com/kafka4beam/brod/blob/ef97325545eed0ae406733bf177090b7a6a192b3/src/brod_topic_subscriber.erl#L497) -> [3](https://github.com/kafka4beam/brod/blob/ef97325545eed0ae406733bf177090b7a6a192b3/src/brod_topic_subscriber.erl#L521) -> [4](https://github.com/kafka4beam/brod/blob/ef97325545eed0ae406733bf177090b7a6a192b3/src/brod_consumer.erl#L269) .

But I'm not absolutely sure.  @zmstone , do you know if it's indeed async?

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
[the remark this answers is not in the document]
savonarola: Reading the code of some other connectors, it seems that some return "connecting" when there is a "chance to connect", i.e. there are connectivity issues, and so on.
And "disconncted" is returned when there is "no chance to connect", i.e. when the related connection pool is down.
May be we should revise the statuses in other connectors later.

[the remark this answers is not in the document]
thalesmg: Yes, I tried to consult other connectors for inspiration, and got confused by that as well.  :hear_no_evil:

apps/emqx/rebar.config
```diff
@@ -32,7 +32,7 @@
     {hocon, {git, "https://github.com/emqx/hocon.git", {tag, "0.36.0"}}},
     {pbkdf2, {git, "https://github.com/emqx/erlang-pbkdf2.git", {tag, "2.0.4"}}},
     {recon, {git, "https://github.com/ferd/recon", {tag, "2.5.1"}}},
-    {snabbkaffe, {git, "https://github.com/kafka4beam/snabbkaffe.git", {tag, "1.0.0"}}}
+    {snabbkaffe, {git, "https://github.com/kafka4beam/snabbkaffe.git", {tag, "1.0.5"}}}
```
ieQu1, line 35: Nit: 1.0.7 is available.

apps/emqx/test/emqx_common_test_helpers.erl
```diff
@@ -614,25 +615,62 @@ emqx_cluster(Specs0, CommonOpts) ->
...
     Node.
 
 %% Node stopping
-stop_slave(Node) ->
-    ct_slave:stop(Node).
+stop_slave(Node0) ->
+    Node = node_name(Node0),
+    SlaveMod = get_peer_mod(Node),
+    erase_peer_mod(Node),
+    case rpc:call(Node, ekka, leave, []) of
```
ieQu1, line 664: Is this step necessary? It will create a partitioned cluster before teardown, which may trigger a bunch of alarms, and restart EMQX apps (takes time).

apps/emqx_bridge/src/emqx_bridge_resource.erl
```diff
@@ -45,7 +45,12 @@
 ]).
 
 %% bi-directional bridge with producer/consumer or ingress/egress configs
--define(IS_BI_DIR_BRIDGE(TYPE), TYPE =:= <<"mqtt">>; TYPE =:= <<"kafka">>).
+-define(IS_BI_DIR_BRIDGE(TYPE),
+    TYPE =:= <<"mqtt">>
```
ieQu1, line 49: Nit: 
```suggestion
    (TYPE) =:= <<"mqtt">>
```

apps/emqx_bridge/src/emqx_bridge_resource.erl
```diff
@@ -45,7 +45,12 @@
...
+).
+-define(IS_INGRESS_BRIDGE(TYPE),
+    TYPE =:= <<"kafka_consumer">> orelse ?IS_BI_DIR_BRIDGE(TYPE)
```
ieQu1, line 52: Nit:
```suggestion
    (TYPE) =:= <<"kafka_consumer">> orelse ?IS_BI_DIR_BRIDGE(TYPE)
```

apps/emqx/test/emqx_common_test_helpers.erl
thalesmg: I took this from the 4.x branch.  Since the default behavior here is to join one of the slaves and not the master node, this might be redundant.  I'll remove it.

apps/emqx/test/emqx_common_test_helpers.erl
```diff
@@ -614,25 +615,57 @@ emqx_cluster(Specs0, CommonOpts) ->
...
+        end,
+    case DoStart() of
+        {ok, _} ->
+            ok;
+        {error, started_not_connected, _} ->
+            ok;
+        {error, already_started, _} ->
+            ok;
+        {error, {already_running, _}} ->
+            ok;
```
ieQu1, line 649: Q: Is idempotency desirable in this situation? Consider that there is a leftover slave from a previous testcase, that is in some weird state. Trying to run different testcase on it will lead to hard-to-decipher problem.

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge.erl
```diff
@@ -82,14 +83,6 @@ resource_type(clickhouse) -> emqx_ee_connector_clickhouse.
 
 fields(bridges) ->
     [
-        {kafka,
```
ieQu1, line 85: Q: Should it be deprecated instead of straight deletion?

changes/ee/feat-9564.zh.md
```diff
@@ -0,0 +1,2 @@
+实现了Kafka消费者桥。
```
zmstone, line 1:
```suggestion
实现了 Kafka 消费者桥接。
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -39,24 +39,24 @@ emqx_ee_bridge_kafka {
...
+            zh: "Kafka Producer"
         }
     }
-    producer_mqtt_opts {
+    producer_opts {
         desc {
-            en: "MQTT data source. Optional when used as a rule-engine action."
-            zh: "需要桥接到 MQTT 源主题。"
+            en: "Local MQTT data source and Kafka bridge configs."
+            zh: "本地 MQTT 数据源和 Kafka 桥接的配置。"
```
zmstone, line 55:
```suggestion
            en: "Local MQTT data source and Kafka bridge configs. Should not configure this if the bridge is used as a rule action."
            zh: "本地 MQTT 数据源和 Kafka 桥接的配置。若该桥接用于规则的动作，则必须将该配置项删除。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
             zh: "GSSAPI/Kerberos"
         }
     }
+
+    kafka_consumer {
+        desc {
+            en: "Kafka Consumer configuration."
+            zh: "Kafka Consumer的配置。"
```
zmstone, line 480:
```suggestion
            zh: "Kafka 消费者配置。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+        label {
+            en: "Kafka Consumer"
+            zh: "Kafka Consumer"
```
zmstone, line 484:
```suggestion
            zh: "Kafka 消费者"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+    }
+    consumer_opts {
+        desc {
+            en: "Local MQTT data sink and Kafka bridge configs."
+            zh: "本地MQTT数据汇和Kafka桥配置。"
```
zmstone, line 490:
```suggestion
            en: "Local MQTT publish and Kafka consumer configs."
            zh: "本地 MQTT 转发 和 Kafka 消费者配置。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+    }
+    consumer_kafka_opts {
+        desc {
+            en: "Kafka consumer configs."
+            zh: "Kafka消费者配置。"
+        }
+        label {
+            en: "Kafka Consumer"
+            zh: "卡夫卡消费者"
```
zmstone, line 504:
```suggestion
            zh: "Kafka 消费者"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+    }
+    consumer_mqtt_opts {
+        desc {
+            en: "MQTT data sink."
+            zh: "MQTT数据汇。"
```
zmstone, line 510:
```suggestion
            en: "Local MQTT message publish."
            zh: "本地 MQTT 消息转发。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+        label {
+            en: "MQTT data sink."
+            zh: "MQTT数据汇。"
```
zmstone, line 514: do not use . and 。 in labels.
```suggestion
            en: "MQTT publish"
            zh: "MQTT 转发"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+    }
+    consumer_mqtt_topic {
+        desc {
+            en: "Local topic to which consumed Kafka messages should be published to."
+            zh: "消耗的Kafka消息应该被发布到的本地主题。"
```
zmstone, line 520:
```suggestion
            en: "Local topic to which consumed Kafka messages should be published to."
            zh: "设置 Kafka 消息向哪个本地 MQTT 主题转发消息。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+        label {
+            en: "MQTT Topic"
+            zh: "MQTT主题"
+        }
+    }
+    consumer_mqtt_qos {
+        desc {
+            en: "MQTT QoS used to publish messages consumed from Kafka."
+            zh: "MQTT QoS用于发布从Kafka消耗的消息。"
```
zmstone, line 530:
```suggestion
            zh: "转发 MQTT 消息时使用的 QoS。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+        label {
+            en: "MQTT Topic QoS"
+            zh: "MQTT 主题服务质量"
```
zmstone, line 534:
```suggestion
            en: "QoS"
            zh: "QoS"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+    }
+    consumer_mqtt_payload {
+        desc {
+            en: "The payload of the MQTT message to be published.\n"
+                "<code>full_message</code> will encode all data available as a JSON object,"
```
zmstone, line 540:
```suggestion
                "<code>full_message</code> will encode available Kafka message attributes as a JSON object, including Key, Value, Timestamp and Headers"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+                "<code>message_value</code> will directly use the Kafka message value as the "
+                "MQTT message payload."
+            zh: "要发布的MQTT消息的有效载荷。"
+                "<code>full_message</code>将把所有可用数据编码为JSON对象，"
```
zmstone, line 544:
```suggestion
                "<code>full_message</code>将把所有可用数据编码为JSON对象，包括 Key，Value，Timestamp 和 Headers。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+                "<code>message_value</code>将直接使用Kafka消息值作为MQTT消息的有效载荷。"
```
zmstone, line 545:
```suggestion
                "<code>message_value</code>将直接使用 Kafka 消息值作为MQTT消息的 Payload。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+        label {
+            en: "MQTT Payload"
+            zh: "MQTT有效载荷"
```
zmstone, line 549:
```suggestion
            zh: "MQTT Payload"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+    }
+    consumer_kafka_topic {
+        desc {
+            en: "Kafka topic to consume from."
+            zh: "从Kafka主题消费。"
```
zmstone, line 555:
```suggestion
            zh: "指定从哪个 Kafka 主题消费消息。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+        label {
+            en: "Kafka topic"
+            zh: "卡夫卡主题 "
```
zmstone, line 559:
```suggestion
            zh: "Kafka 主题 "
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+    consumer_max_batch_bytes {
+        desc {
+            en: "Maximum bytes to fetch in a batch of messages."
+                "NOTE: this value might be expanded to retry when "
+                "it is not enough to fetch even a single message, "
+                "then slowly shrink back to the given value."
+            zh: "在一批消息中要取的最大字节数。"
+                "注意：这个值可能会被扩大，"
+                "当它甚至不足以取到一条消息时，就会重试，"
+                "然后慢慢缩回到给定的值。"
```
zmstone, line 571:
```suggestion
            en: "Maximum bytes to fetch in a batch of messages."
                "Please note that if the configured value is smaller than the message size in Kafka, it may negatively impact the fetch performance."
            zh: "在一批消息中要取的最大字节数。"
                "如果该配置小于 Kafka 中消息到大小，则可能会影响消费性能。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+    consumer_offset_reset_policy {
+        desc {
+            en: "Defines how the consumers should reset the start offset when "
+                "a topic partition has and invalid or no initial offset."
+            zh: "定义当一个主题分区的初始偏移量无效或没有初始偏移量时，"
+                "消费者应如何重置开始偏移量。"
+        }
+        label {
+            en: "Offset Reset Policy"
+            zh: "偏移重置政策"
```
zmstone, line 597:
```suggestion
            zh: "偏移重置策略"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+        }
+    }
+    consumer_offset_commit_interval_seconds {
+        desc {
+            en: "Defines the time interval between two <code>OffsetCommitRequest</code> messages."
+            zh: "定义了两个<code>OffsetCommitRequest</code>消息之间的时间间隔。"
```
zmstone, line 603:
```suggestion
            en: "Defines the time interval between two offset commit requests sent for each consumer group."
            zh: "指定 Kafka 消费组偏移量提交的时间间隔。"
```

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge.erl
zmstone: renamed to `kafka_producer` ?
We'll need to add a `aliases` feature in hocon schema so we can more freely rename rename fields without writing a lot of converters for backward compatibility

thalesmg: Yes, it was renamed to `kafka_producer`, and its schema was also changed.  That was because it had a structure that would be more suitable if the bridge were to behave bidirectionally, like the MQTT bridge.  But that doesn't work well for the Kafka Producer / Consumer duo.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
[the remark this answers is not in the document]
zmstone: sorry for the late reply.
yes, the offset commits are async.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -0,0 +1,403 @@
...
+            topic := _
+        },
+        mqtt := #{topic := _, qos := _, payload := _},
+        ssl := SSL
+    } = Config,
+    BootstrapHosts = emqx_bridge_impl_kafka:hosts(BootstrapHosts0),
+    KafkaType = kafka_consumer,
+    %% Note: this is distinct per node.
+    ClientID0 = emqx_bridge_impl_kafka:make_client_id(KafkaType, BridgeName),
+    ClientID = binary_to_atom(ClientID0),
```
zmstone, line 105: is this called by the probe APIs ? if yes, we'll need to avoid creating new atoms.

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,138 @@ emqx_ee_bridge_kafka {
...
+                "注意：这个值可能会被扩大，"
+                "当它甚至不足以取到一条消息时，就会重试，"
+                "然后慢慢缩回到给定的值。"
+        }
+        label {
+            en: "Max Bytes"
+            zh: "最大字节数"
+        }
+    }
+    consumer_max_rejoin_attempts {
```
zmstone, line 578: this is not something useful for EMQX user.
we can mark this filed has `hidden`, or pick a sane default value which can protect the consumer's supervisor from crash due to max intensity.
(there is a small delay between the re-join attempts, so if the max attempt is large enough, it should not cause supervisor restart.)

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge.erl
zmstone: we'll have to keep backward compatibility for this change.

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
thalesmg: We're using the default of `5`, which is the same default used in `brod`.  I'll make this field hidden then.

apps/emqx/test/emqx_common_test_helpers.erl
thalesmg: Good point, I've removed those 2 clauses.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
thalesmg: Excellent catch.  While adding a test for that, I noticed that there were several schema fields and types related to the API that were lacking updates.  :see_no_evil:

thalesmg:
Putting this back on draft while I think about the hocon alias + backwards compatibility thingy.

thalesmg:
Feedback from product team:

- Allow more Kafka-MQTT topic mappings per bridge.
- Support placeholders in `mqtt.payload`.
- Add key and value encoding options (utf8 / base64).

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_kafka.erl
```diff
@@ -30,14 +30,25 @@
...
 conn_bridge_examples(Method) ->
     [
         #{
+            %% TODO: rename this to `kafka_producer' after alias
+            %% support is added to hocon; keeping this as just `kafka'
+            %% for backwards compatibility.
             <<"kafka">> => #{
-                summary => <<"Kafka Bridge">>,
+                summary => <<"Kafka Producer Bridge">>,
+                value => values(Method)
```
thalesmg, line 46: TODO: improve those examples (those are just bootstrap hosts from before)

thalesmg:
@ysfscream @Kinplemelon These changes will require some changes in the frontend.

The schema for the Kafka Producer changed slightly.  I tried to fix it here: https://github.com/emqx/emqx-dashboard-web-new/pull/1051

I'm not sure if I covered all the necessary changes, though.  I'd appreciate if you could review that PR and indicate what I can improve.  :smiley_cat:

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -473,4 +473,175 @@ emqx_ee_bridge_kafka {
...
+    }
+    consumer_encoding_mode {
+        desc {
+            en: "Defines how the key or value from the Kafka message is"
+                " dealt with before being forwarded via MQTT.\n"
+                "<code>force_utf8</code> Uses UTF-8 encoding directly from the original message.\n"
+                "<code>base64</code> Uses base-64 encoding on the received key or value."
+            zh: "定义了在通过MQTT转发之前如何处理Kafka消息的键或值。"
+                "<code>force_utf8</code> 直接使用原始信息的UTF-8编码。\n"
+                "<code>base64</code> 对收到的密钥或值使用base-64编码。"
```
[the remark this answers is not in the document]
thalesmg: Sorry, this one changed after our discussion about changing `force_utf8` to `none`.  :speak_no_evil: 

Could you please have another look?

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -0,0 +1,499 @@
...
+            ?SLOG(error, Log#{
+                exception => C,
+                reason => E
+            })
+    end.
+
+-spec consumer_group_id(atom() | binary()) -> binary().
+consumer_group_id(BridgeName0) ->
+    BridgeName = to_bin(BridgeName0),
+    <<"emqx-kafka-consumer:", BridgeName/binary>>.
```
zmstone, line 435:
```suggestion
    <<"emqx-kafka-consumer-", BridgeName/binary>>.
```

thalesmg:
Performed a manual test for resiliency.

1. Created a Kafka cluster with 2 brokers.
2. Ensured that `__consumer_offsets` replication factor was 2, so that coordinator failover was possible.
3. Created a normal topic with replication factor 2 and configured a Kafka Consumer to consume from that.
4. Started a Kafka producer to produce sequential messages to such topic.
5. Found out which Kafka broker was that group's coordinator and stopped that Kafka broker.
6. Eventually Kafka elected the surviving node as the new group coordinator; the consumer group in EMQX eventually got notified and received the messages that were published between the broker going down and the election end/group rebalance.

zmstone (approved):

PR #9565, opened 2022-12-15
feat(cover): add covertool

zmstone:
part 2 of [EMQX-7985](https://emqx.atlassian.net/browse/EMQX-7958)

## PR Checklist


src/emqx_cover.erl
```diff
@@ -85,3 +200,9 @@ is_emqx_module(Module) ->
         _ ->
             false
     end.
+
+os_env(Name) ->
+    case os:getenv(Name) of
+        false -> "";
+        V -> V
+    end.
```
thalesmg, line 208: Nit:

```suggestion
    os:getenv(Name, "").
```

rebar.config.erl
```diff
@@ -253,6 +253,7 @@ relx_apps(ReleaseType) ->
     , {ekka, load}
     , {emqx_plugin_libs, load}
     , observer_cli
+    , covertool
```
thalesmg, line 256: Maybe we could just load this instead of always starting?

```suggestion
    , {covertool, load}
```

thalesmg (approved):

qzhuyan (approved):

JimMoen (approved):

PR #9579, opened 2022-12-19
fix(kafka_producer): cleanup client after failing to start producers

thalesmg:
https://emqx.atlassian.net/browse/EMQX-8547

If a Kafka Producer bridge is given bad configuration (e.g.: bad authn credentials), the Wolff client process is started successfully, as it does not attempt to connect, but when the producer process is attempted to be started, it fails (only then the client tries to connect to Kafka).  At this point, an error was thrown, but the supervised client process remained running.

If the configuration was later fixed and the bridge updated, which prompted its removal and recreation, the Wolff client would report to be "already started", so it would never pick up the new (fixed) configuration, and the producers would perpetually fail to start until the node would be restarted.

We simply ensure the client is stopped before throwing the error, unrolling the start-up procedure.

Fixes <issue-number>

## PR Checklist

### DIFF
--- a/lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
+++ b/lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
@@ -86,6 +86,19 @@ on_start(InstId, Config) ->
                 kafka_topic => KafkaTopic,
                 reason => Reason2
             }),
+            %% Need to stop the already running client; otherwise, the
+            %% next `on_start' call will try to ensure the client
+            %% exists and it will be already present and using the old
+            %% config.  This is specially bad if the original crash
+            %% was due to misconfiguration and we are trying to fix
+            %% it...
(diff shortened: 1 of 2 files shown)


lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
```diff
@@ -86,6 +86,18 @@ on_start(InstId, Config) ->
                 kafka_topic => KafkaTopic,
                 reason => Reason2
             }),
+            %% Need to stop the already running client; otherwise, the
+            %% next `on_start' call will try to ensure the client
+            %% exists and it will.  This is specially bad if the
```
zmstone, line 91: and it will ?

thalesmg: Fixed the comment, thanks

zmstone (approved):

PR #9587, opened 2022-12-21
ci: try latest erlef/setup-beam in build_slim_packages

id:


id:
https://github.com/emqx/emqx/pull/9590

PR #9592, opened 2022-12-21
1221 fix io format error when rpc from remsh

zmstone:
fixes [EMQX-8581](https://emqx.atlassian.net/browse/EMQX-8581)

### DIFF
--- a/src/emqx_misc.erl
+++ b/src/emqx_misc.erl
@@ -26,6 +26,7 @@
 -export([ merge_opts/2
         , maybe_apply/2
         , maybe_mute_rpc_log/0
+        , safe_io_device/0
         , compose/1
         , compose/2
         , run_fold/3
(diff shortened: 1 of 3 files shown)


id (approved):

src/emqx_misc.erl
```diff
@@ -26,6 +26,7 @@
 -export([ merge_opts/2
         , maybe_apply/2
         , maybe_mute_rpc_log/0
+        , safe_io_device/0
```
thalesmg, line 29: Q: shouldn't this be called in nodetool?

zmstone: no, it is to be called by the rpc handler node.
only one place found so far: emqx_license (will be sent in another PR to emqx-enterprise.git).

qzhuyan (approved):

PR #9593, opened 2022-12-22
fix(bridges): obfuscate the password in bridges API responses

lafirest:
port `redact` from 4.3 to master, and obfuscate sensitive data in the response when querying `bridges` information by API
fix [EMQX-8374](https://emqx.atlassian.net/browse/EMQX-8374)

### DIFF
--- a/apps/emqx/src/emqx_misc.erl
+++ b/apps/emqx/src/emqx_misc.erl
@@ -68,7 +68,7 @@
     nolink_apply/2
 ]).
 
--export([clamp/3]).
+-export([clamp/3, redact/1, redact/2, is_redacted/2, is_redacted/3]).
 
 -dialyzer({nowarn_function, [nolink_apply/2]}).
 
@@ -556,6 +556,75 @@ try_to_existing_atom(Convert, Data, Encoding) ->
         _:Reason -> {error, Reason}
     end.
 
+is_sensitive_key(token) -> true;
+is_sensitive_key("token") -> true;
+is_sensitive_key(<<"token">>) -> true;
+is_sensitive_key(password) -> true;
+is_sensitive_key("password") -> true;
+is_sensitive_key(<<"password">>) -> true;
+is_sensitive_key(secret) -> true;
+is_sensitive_key("secret") -> true;
+is_sensitive_key(<<"secret">>) -> true;
+is_sensitive_key(_) -> false.
+
+redact(Term) ->
+    do_redact(Term, fun is_sensitive_key/1).
+
+redact(Term, Checker) ->
+    do_redact(Term, fun(V) ->
+        is_sensitive_key(V) orelse Checker(V)
+    end).
+
+do_redact(L, Checker) when is_list(L) ->
+    lists:map(fun(E) -> do_redact(E, Checker) end, L);
+do_redact(M, Checker) when is_map(M) ->
+    maps:map(
+        fun(K, V) ->
+            do_redact(K, V, Checker)
+        end,
+        M
+    );
+do_redact({Key, Value}, Checker) ->
+    case Checker(Key) of
+        true ->
+            {Key, redact_v(Value)};
+        false ->
+            {do_redact(Key, Checker), do_redact(Value, Checker)}
+    end;
+do_redact(T, Checker) when is_tuple(T) ->
+    Elements = erlang:tuple_to_list(T),
+    Redact = do_redact(Elements, Checker),
+    erlang:list_to_tuple(Redact);
+do_redact(Any, _Checker) ->
+    Any.
+
+do_redact(K, V, Checker) ->
+    case Checker(K) of
+        true ->
+            redact_v(V);
+        false ->
+            do_redact(V, Checker)
+    end.
+
+-define(REDACT_VAL, "******").
+redact_v(V) when is_binary(V) -> <<?REDACT_VAL>>;
+redact_v(_V) -> ?REDACT_VAL.
+
+is_redacted(K, V) ->
+    do_is_redacted(K, V, fun is_sensitive_key/1).
+
+is_redacted(K, V, Fun) ->
+    do_is_redacted(K, V, fun(E) ->
+        is_sensitive_key(E) orelse Fun(E)
+    end).
+
+do_is_redacted(K, ?REDACT_VAL, Fun) ->
+    Fun(K);
+do_is_redacted(K, <<?REDACT_VAL>>, Fun) ->
+    Fun(K);
+do_is_redacted(_K, _V, _Fun) ->
+    false.
+
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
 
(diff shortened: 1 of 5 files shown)


apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -402,12 +402,12 @@ schema("/nodes/:node/bridges/:id/operation/:operation") ->
...
     {200,
         zip_bridges([
-            [format_resp(Data, Node) || Data <- emqx_bridge_proto_v1:list_bridges(Node)]
+            [redact(format_resp(Data, Node)) || Data <- emqx_bridge_proto_v1:list_bridges(Node)]
          || Node <- mria_mnesia:running_nodes()
         ])}.
 
 '/bridges/:id'(get, #{bindings := #{id := Id}}) ->
-    ?TRY_PARSE_ID(Id, lookup_from_all_nodes(BridgeType, BridgeName, 200));
+    ?TRY_PARSE_ID(Id, redact(lookup_from_all_nodes(BridgeType, BridgeName, 200)));
```
thalesmg, line 410: I think that those calls to `redact` would be better inserted at `format_resp` and `format_bridge_info`, since we'll probably want that effect every time we format data for a response.

apps/emqx/src/emqx_misc.erl
```diff
@@ -556,6 +556,60 @@ try_to_existing_atom(Convert, Data, Encoding) ->
         _:Reason -> {error, Reason}
     end.
 
+is_sensitive_key(token) -> true;
```
zmstone, line 559: hmm. i though we have ported this function long ago. there are more keys in 4.4 ?

zmstone (dismissed):
I'm afraid the situation is a bit more complicated than what it seems to be.
Since the GET response is used for dashboard to pre-fill the text boxes so the user can update some fields and click the "Save" or "Update" button to update bridge configuration, we'll need to handle the case in which the sensitive filed is provided with value <<"******">> then we do not update the password for this config.

Otherwise for whatever they want to update, they will have to re-fill the password box.
super inconvenient, but also not making it more secure than before because the user
would likely have to copy-paste the password over and over again, i.e. actually higher chance to leak it.

apps/emqx/src/emqx_misc.erl
lafirest: Yes, but some are not usually used, here are only common ones, and then the redact/2 allow to pass a parameter as a custom checker function

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -390,40 +390,50 @@ schema("/nodes/:node/bridges/:id/operation/:operation") ->
...
+                {ok, _} ->
+                    {400, error_msg('ALREADY_EXISTS', <<"bridge already exists">>)};
+                {error, not_found} ->
+                    case ensure_bridge_created(BridgeType, BridgeName, Conf) of
+                        ok -> lookup_from_all_nodes(BridgeType, BridgeName, 201);
+                        {error, Error} -> {400, Error}
+                    end
+            end;
+        _ ->
+            {400, error_msg('BAD_REQUEST', <<"There is obfuscated sensitive data">>)}
```
zmstone, line 405: if this POST is for creation only, then it's reasonable to return 400 if `******` is provided as password.
otherwise if this also used for updating existing resource, the handler should replace it with the currently configured password.

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -390,40 +390,50 @@ schema("/nodes/:node/bridges/:id/operation/:operation") ->
...
+                                lookup_from_all_nodes(BridgeType, BridgeName, 200);
+                            {error, Error} ->
+                                {400, Error}
+                        end;
+                    {error, not_found} ->
+                        {404, error_msg('NOT_FOUND', <<"bridge not found">>)}
+                end
+            );
+        _ ->
+            {400, error_msg('BAD_REQUEST', <<"There is obfuscated sensitive data">>)}
```
zmstone, line 435: if this PUT is used for creation, then it's reasonable to return 400 if ****** is provided as password.

AFAIK, PUTs are used for updating, in this caes the handler should replace it with the currently configured password.

apps/emqx_bridge/src/emqx_bridge_api.erl
lafirest: There is and only POST will create a bridge, so there is no need to handle with config updates here

apps/emqx_bridge/src/emqx_bridge_api.erl
lafirest: There is and only POST will create a bridge, so there is no need to handle creating here

apps/emqx_bridge/src/emqx_bridge_api.erl
zmstone: then there should be no need to check it here, because `<<"******">>` can be a legit password.

apps/emqx_bridge/test/emqx_bridge_api_SUITE.erl
```diff
@@ -514,6 +515,40 @@ t_reset_bridges(Config) ->
     {ok, 204, <<>>} = request(delete, uri(["bridges", BridgeID]), []),
     {ok, 200, <<"[]">>} = request(get, uri(["bridges"]), []).
 
+t_with_redact_update(Config) ->
+    ErrorMsg = <<"{\"code\":\"BAD_REQUEST\",\"message\":\"There is obfuscated sensitive data\"}">>,
```
thalesmg, line 519: This doesn't seem to be used?

apps/emqx_bridge/test/emqx_bridge_api_SUITE.erl
```diff
@@ -47,7 +47,8 @@
 }).
 
 all() ->
-    emqx_common_test_helpers:all(?MODULE).
+    [t_with_redact_update].
+%%    emqx_common_test_helpers:all(?MODULE).
```
thalesmg, line 51:
```suggestion
    emqx_common_test_helpers:all(?MODULE).
```

apps/emqx_bridge/test/emqx_bridge_api_SUITE.erl
```diff
@@ -514,6 +515,40 @@ t_reset_bridges(Config) ->
...
+    {ok, 200, _ResBin} = request(
+        put,
+        uri(["bridges", BridgeID]),
+        Conf
+    ),
+    RawConf = emqx:get_raw_config([bridges, Type, Name]),
+    Password = <<"password">>,
+    Value = maps:get(Password, RawConf),
+    ?assertNot(emqx_misc:is_redacted(Password, Value)),
+    ok.
```
thalesmg, line 550: Suggestion: assert the current value to make it clear it's not changed.

```suggestion
    ?assertEqual(<<"123456">>, Value),
    ok.
```

terry-xiaoyu (approved):

PR #9599, opened 2022-12-22
fix(banned): ensure the default expiration time of `banned` is large enough

lafirest:
fix [EMQX-8460](https://emqx.atlassian.net/browse/EMQX-8460)

### DIFF
--- a/apps/emqx/src/emqx_banned.erl
+++ b/apps/emqx/src/emqx_banned.erl
@@ -59,6 +59,10 @@
 
 -define(BANNED_TAB, ?MODULE).
 
+%% The default expiration time should be infinite
+%% but for compatibility, a large number (1 years) is used here to represent the 'infinite'
+-define(EXPIRATION_TIME, 31536000).
+
 -ifdef(TEST).
 -compile(export_all).
 -compile(nowarn_export_all).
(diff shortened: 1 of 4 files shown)


apps/emqx/src/emqx_banned.erl
```diff
@@ -126,7 +130,7 @@ parse(Params) ->
             By = maps:get(<<"by">>, Params, <<"mgmt_api">>),
             Reason = maps:get(<<"reason">>, Params, <<"">>),
             At = maps:get(<<"at">>, Params, erlang:system_time(second)),
-            Until = maps:get(<<"until">>, Params, At + 5 * 60),
+            Until = maps:get(<<"until">>, Params, At + ?EXPIRATION_TIME),
```
thalesmg, line 133: Q: can't the default value when `<<"until">>` is absent really be `infinity`?

It's a bit of a hack, but since atoms are always greater than numbers:

```elixir
iex(10)> :infinity > 100000000000000000
true
```

:joy_cat: 

Perhaps because this value has to be returned in some API, and must be a number?

thalesmg (approved):

apps/emqx/src/emqx_banned.erl
lafirest: I also think using `permanent` or `infinite` will be better, but they are not friendly for the external caller.

apps/emqx/src/emqx_banned.erl
```diff
@@ -59,6 +59,10 @@
 
 -define(BANNED_TAB, ?MODULE).
 
+%% The default expiration time should be infinite
+%% but for compatibility, a large number (100 years) is used here to represent the 'infinite'
+-define(EXPIRATION_TIME, 3153600000).
```
zmstone, line 64: haha
I think we can set it to 1 year.

thalesmg (approved):

zmstone (approved):

PR #9611, opened 2022-12-23
fix(upgrade): fix start script existence check (v4.4)

thalesmg:


thalesmg:
I wonder if we need this?  :thinking: 

https://github.com/emqx/emqx/blob/c1e5ec5f80060eed1ba19b3a4a36ec6782bafac0/bin/install_upgrade.escript#L189-L191

The code that calls this supposedly is run when `releases/$VERSION` is absent.  But `releases/RELEASES` is not necessarily absent, and calling this will wipe the existing version data, severing access to the node (i.e. `emqx any_command` will error).

zmstone (dismissed):

zmstone (approved):

PR #9612, opened 2022-12-23
feat(upgrade): add forward version check for upgrade script

thalesmg:
https://emqx.atlassian.net/browse/EMQX-6978

[Related EIP](https://github.com/emqx/eip/blob/c4864eeccb4f2897f42e5c9e2917133a0d3215bd/active/0022-forward-check-install-upgrade-script.md)

Currently, when performing a hot upgrade, the scripts that are run are those from the currently installed EMQX version. It has a validation that prevents upgrade between different minor versions. If we want to allow an upgrade from, say, 5.0.x to 5.1.y, then the scripts in already released packages will deny such operation. Also, if an upgrade installation script contains a bug in the current, it will never be able to execute properly without manual patching.

By attempting to execute the scripts from the target version, we may add fixes and new validations to new EMQX versions and have them executed by older versions.

### DIFF
--- a/bin/emqx
+++ b/bin/emqx
@@ -5,7 +5,13 @@
 set -euo pipefail
 
 DEBUG="${DEBUG:-0}"
-[ "$DEBUG" -eq 1 ] && set -x
+if [ "$DEBUG" -eq 1 ]; then
+  set -x
+fi
+if [ "$DEBUG" -eq 2 ]; then
+  set -x
+  export PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+fi
 
 # We need to find real directory with emqx files on all platforms
 # even when bin/emqx is symlinked on several levels
@@ -36,6 +42,7 @@ export RUNNER_ROOT_DIR
 export EMQX_ETC_DIR
 export REL_VSN
 export SCHEMA_MOD
+export IS_ENTERPRISE
 
 RUNNER_SCRIPT="$RUNNER_BIN_DIR/$REL_NAME"
 CODE_LOADING_MODE="${CODE_LOADING_MODE:-embedded}"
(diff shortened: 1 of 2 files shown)


bin/emqx
```diff
@@ -540,6 +547,121 @@ check_license() {
...
+  fi
+}
+
+# extract_from_package packege_path destination file1 file2
+extract_from_package() {
+  local package="$1"
+  local dest_dir="$2"
+  shift 2
+
+  tar -C "$dest_dir" -xf "$package" "$@"
```
zmstone, line 600: the mac release is unfortunately a .zip
we can perhaps simple drop relup support for macos.

bin/emqx
```diff
@@ -922,8 +1044,20 @@ case "${COMMAND}" in
...
+        if ! am_i_the_newest_script "$newest_vsn"; then
+          script_path="$(versioned_script_path emqx "$newest_vsn")"
+          exec "$script_path" "$COMMAND" "$@"
+        fi
+
+        upgrade_script_path="$(versioned_script_path install_upgrade.escript "$newest_vsn")"
+
         ERL_FLAGS="${ERL_FLAGS:-} $EPMD_ARGS" \
-        exec "$BINDIR/escript" "$RUNNER_ROOT_DIR/bin/install_upgrade.escript" \
+        exec "$BINDIR/escript" "$upgrade_script_path" \
```
zmstone, line 1060: log the path to console ?

bin/install_upgrade.escript
```diff
@@ -217,7 +219,12 @@ find_and_link_release_package(Version, RelName) ->
     %% we've found where the actual release package is located
     ReleaseLink = filename:join(["releases", Version,
                                  RelNameStr ++ ".tar.gz"]),
-    TarBalls = filename:join(["releases", RelNameStr ++ "-*" ++ Version ++ "*.tar.gz"]),
+    ReleaseNamePattern =
+        case IsEnterprise of
+            false -> RelNameStr;
+            true -> RelNameStr ++ "-enterprise"
+        end,
+    TarBalls = filename:join(["releases", ReleaseNamePattern ++ "-" ++ Version ++ "*.tar.gz"]),
```
zmstone, line 227: nit
```suggestion
    TarBalls = filename:join(["releases", ReleaseNamePattern, "-", Version, "*.tar.gz"]),
```

zmstone:
good job.
now we can make the install_upgrade.escript exit with a log "unsupported upgrade path" before we are ready.

bin/install_upgrade.escript
thalesmg: I'm afraid that won't have the desired output.  Using `filename:join` will introduce `/` between the parts of the pattern rather than concatenate then.

thalesmg: I'll refactor it to use `lists:flatten/1`.

thalesmg: :heavy_check_mark: 
https://github.com/emqx/emqx/pull/9612/commits/660d028ebb04c240324d00742cb6249292a9520f

bin/emqx
```diff
@@ -912,6 +1034,11 @@ case "${COMMAND}" in
         ;;
 
     upgrade|downgrade|install|unpack|uninstall)
+        ## For now, we don't have relup; remove once we are ready
+        ## TODO: always deny relup for macos (unsupported)
+        logerr "Unsupported upgrade path"
+        exit 1
```
zmstone, line 1040: Er, we can't release it like this.
What I meant is to do this in the `main` function in install_upgrade.escript.
Otherwise if we release `bin/emqx` this way for `e5.0.0`, there is no way to upgrade from it.

thalesmg: :hear_no_evil: :see_no_evil:

thalesmg: :heavy_check_mark: 
https://github.com/emqx/emqx/pull/9612/commits/465e73cac0176fb9af1837c8e93d2c423bc9be89

bin/install_upgrade.escript
```diff
@@ -18,22 +18,31 @@ main([Command0, DistInfoStr | CommandArgs]) ->
...
     end,
     F(DistInfo, Opts);
 main(Args) ->
     ?INFO("unknown args: ~p", [Args]),
     erlang:halt(1).
 
+%% temporary block for hot-upgrades; next release will just remove
+%% this and the new script version shall be used instead of this
+%% current version.
+%% TODO: always deny relup for macos (unsupported)
```
zmstone, line 37: please create a jira to follow up on this change.
the fix should be like this: https://github.com/emqx/emqx/commit/ba46be717ba7f4a906b1bec2313ed82b37c63dee

zmstone (approved):

bin/install_upgrade.escript
thalesmg: :heavy_check_mark: 
https://emqx.atlassian.net/browse/EMQX-8654

PR #9619, opened 2022-12-27
refactor(metrics): use absolute gauge values rather than deltas (v5.0)

thalesmg:
https://emqx.atlassian.net/browse/EMQX-8548

Currently, we face several issues trying to keep resource metrics
reasonable.  For example, when a resource is re-created and has its
metrics reset, but then its durable queue resumes its previous work
and leads to strange (often negative) metrics.

Instead using `counters` that are shared by more than one worker to
manage gauges, we introduce an ETS table whose key is not only scoped
by the Resource ID as before, but also by the worker ID.  This way,
when a worker starts/terminates, they should set their own gauges to
their values (often 0 or `replayq:count` when resuming off a queue).
With this scoping and initialization procedure, we'll hopefully avoid
hitting those strange metrics scenarios and have better control over
the gauges.

### DIFF
--- a/apps/emqx_bridge/i18n/emqx_bridge_schema.conf
+++ b/apps/emqx_bridge/i18n/emqx_bridge_schema.conf
@@ -195,7 +195,7 @@ emqx_bridge_schema {
 
     metric_sent_inflight {
                    desc {
-                         en: """Count of messages that were sent asynchronously but ACKs are not received."""
+                         en: """Count of messages that were sent asynchronously but ACKs are not yet received."""
                          zh: """已异步地发送但没有收到 ACK 的消息个数。"""
                         }
                    label: {
--- a/apps/emqx_resource/src/emqx_resource_metrics.erl
+++ b/apps/emqx_resource/src/emqx_resource_metrics.erl
@@ -24,11 +24,12 @@
 ]).
 
 -export([
-    batching_change/2,
+    batching_set/3,
+    batching_shift/3,
     batching_get/1,
-    inflight_change/2,
+    inflight_set/3,
     inflight_get/1,
-    queuing_change/2,
+    queuing_set/3,
     queuing_get/1,
     dropped_inc/1,
     dropped_inc/2,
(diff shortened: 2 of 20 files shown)


thalesmg:
Requires https://github.com/kafka4beam/wolff/pull/41 to have proper metrics for the Kafka Producer bridge.

ieQu1 (dismissed):

thalesmg:
@ieQu1 sorry, forgot to update [this clause](https://github.com/thalesmg/emqx/blob/77feb8366a99d252496875b059b9a4659a02ed29/lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl#L324) for `queuing` and `inflight` after Wolff's upgrade... :see_no_evil:

apps/emqx_resource/src/emqx_resource_metrics.erl
```diff
@@ -152,6 +147,34 @@ handle_telemetry_event(
...
+    end;
+handle_telemetry_event(
+    [?TELEMETRY_PREFIX, Event],
+    _Measurements = #{gauge_shift := Val},
+    _Metadata = #{resource_id := ID, worker_id := WorkerID},
+    _HandlerConfig
+) ->
+    case Event of
+        batching ->
+            emqx_metrics_worker:shift_gauge(?RES_METRICS, ID, WorkerID, 'batching', Val);
```
zmstone, line 174: we can delete the batching ?

apps/emqx_bridge/i18n/emqx_bridge_schema.conf
```diff
@@ -195,7 +195,7 @@ emqx_bridge_schema {
 
     metric_sent_inflight {
```
zmstone, line 196: nit: inflight is sent
```suggestion
    metric_inflight {
```

apps/emqx_resource/src/emqx_resource_metrics.erl
thalesmg: I did it in this other refactoring of the buffer workers: https://github.com/emqx/emqx/pull/9642

zmstone (approved):

apps/emqx_bridge/i18n/emqx_bridge_schema.conf
thalesmg: Will fix it here: https://github.com/emqx/emqx/pull/9642

thalesmg:
The spellcheck failure is unrelated to the changes here and was fixed [elsewhere](https://github.com/emqx/emqx/pull/9658).

PR #9622, opened 2022-12-27
ci: remove step to update description on docker hub

id:


id:
https://github.com/emqx/emqx/pull/9643

PR #9623, opened 2022-12-27
fix: set ERL_LIBS variable in bin/emqx commands

id:
Fixes https://emqx.atlassian.net/browse/EMQX-8544

When emqx is installed via Homebrew on MacOS, the following commands fail:
- emqx remote_console
- emqx versions

with this error (truncated):
```
    reason: {shutdown,
                {failed_to_start_child,ekka_epmd,
                    {'EXIT',
                        {undef,
                            [{ekka_epmd,start_link,[],[]},
```

### DIFF
--- a/bin/emqx
+++ b/bin/emqx
@@ -908,7 +908,7 @@ case "${COMMAND}" in
         assert_node_alive
 
         shift
-        remsh
+        ERL_LIBS="$ERTS_LIB_DIR" remsh
         ;;
 
     upgrade|downgrade|install|unpack|uninstall)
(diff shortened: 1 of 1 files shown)


lafirest (approved):

zmstone:
why didn't https://github.com/emqx/emqx/blob/9f403e421728a802f3266bfd5645a2c5561b5a4b/bin/nodetool#L338-L360 work?

bin/emqx
```diff
@@ -922,7 +922,7 @@ case "${COMMAND}" in
 
         assert_node_alive
 
-        ERL_FLAGS="${ERL_FLAGS:-} $EPMD_ARGS" \
+        ERL_FLAGS="${ERL_FLAGS:-} $EPMD_ARGS" ERL_LIBS="$ERTS_LIB_DIR" \
```
zmstone, line 925: we will stop supporting relup for macos, do not bother.

id:
> wonder why nodetool's add_libs didn't work:

remote_console == remsh, this one does not use nodetool

id:
The real problem seems to be is that we try to start vanilla erlang node to connect to emqx, but we also tell it to use custom `-epmd_module ekka` and `-proto_dist ekka` which it obviously does not have access to.

I'm now wondering why we don't have this issue anywhere, only on brew-packaged MacOS binaries.

id:
start_clean.boot has all the paths, still looking

id:
ok, the issue is `./rebar3 as "$PROFILE" tar` rewrites `*.boot` files to set correct relative paths, but I use the vanilla ones in the brew package /facepalm

PR #9632, opened 2022-12-29
fix(resource): fix resource auto_reconnect not working

JimMoen:


thalesmg (commented):
Is it possible to add a test for this change to avoid regressions?

JimMoen:
@thalesmg  Yes, but the process will be complicated.
This requires closing the corresponding resources during the test and then testing their reconnection, including but not limited to mysql, such as redis, pgsql, etc. This is difficult to do in common test, we need help of some external testing tools like jmeter.
In addition, some resources do not provide `auto_reconnect` option, we need to handle it separately, such as mongodb.

thalesmg:
> @thalesmg Yes, but the process will be complicated. This requires closing the corresponding resources during the test and then testing their reconnection, including but not limited to mysql, such as redis, pgsql, etc. This is difficult to do in common test, we need help of some external testing tools like jmeter. In addition, some resources do not provide `auto_reconnect` option, we need to handle it separately, such as mongodb.

I see.  Do you think that even using Toxiproxy it would be hard to check it?

For example: https://github.com/JimMoen/emqx/blob/3d1666e29a2d0c1c8c925b5881eeed2951c5ff08/lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_influxdb_SUITE.erl#L809-L811

Maybe we don't need to add a test for every single bridge indeed, maybe one could be enough, assuming Toxiproxy can arouse that behavior.  :thinking:

terry-xiaoyu:
Close this one, I will do some refactoring on the driver level config `auto_reconnect` and the emqx-resource level config `auto_restart_interval `.

PR #9637, opened 2022-12-29
fix(clients): fix expiry_interval unit error

HJianBo:
1. The unit of expiry_interval was changed in 0704cbc986c37ff507d39c9e9ff953ef76a17767 in the previous implementation of 5.0, but the corresponding HTTP API returns were not changed accordingly. So we fixed it in this PR.
2.  The unnecessary search parameter proto_name has been removed because all clients' proto_names must be MQTT, so this parameter is meaningless.

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -65,7 +65,6 @@
     {<<"ip_address">>, ip},
     {<<"conn_state">>, atom},
     {<<"clean_start">>, atom},
-    {<<"proto_name">>, binary},
     {<<"proto_ver">>, integer},
     {<<"like_clientid">>, binary},
     {<<"like_username">>, binary},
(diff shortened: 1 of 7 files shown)


zhongwencool (dismissed):

changes/v5.0.14-en.md
```diff
@@ -0,0 +1,9 @@
+# v5.0.13
```
id, line 1:
```suggestion
# v5.0.14
```

id:
@HJianBo please rebase on latest master and address changelog conflict

lafirest (dismissed):

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -145,14 +144,6 @@ schema("/clients") ->
                         required => false,
                         description => <<"Whether the client uses a new session">>
                     })},
-                {proto_name,
```
zmstone, line 148: deprecate => {since, "v5.0.14"}, please find an example in the code base,

zmstone: never mind, seems to be only used in GET response.

qzhuyan (approved):

PR #9642, opened 2022-12-29
feat(buffer_worker): refactor buffer/resource workers to always use queue and use offload mode

thalesmg:
https://emqx.atlassian.net/browse/EMQX-8623

Currently, we face several issues trying to keep resource metrics
reasonable.  For example, when a resource is re-created and has its
metrics reset, but then its durable queue resumes its previous work
and leads to strange (often negative) metrics.

Instead using `counters` that are shared by more than one worker to
manage gauges, we introduce an ETS table whose key is not only scoped
by the Resource ID as before, but also by the worker ID.  This way,
when a worker starts/terminates, they should set their own gauges to
their values (often 0 or `replayq:count` when resuming off a queue).
With this scoping and initialization procedure, we'll hopefully avoid
hitting those strange metrics scenarios and have better control over
the gauges.

This makes the buffer/resource workers always use `replayq` for
queuing, along with collecting multiple requests in a single call.
This is done to avoid long message queues for the buffer workers and
rely on `replayq`'s capabilities of offloading to disk and detecting
overflow.

Also, this deprecates the `enable_batch` and `enable_queue` resource
creation options, as: i) queuing is now always enables; ii) batch_size > 1 
<=> batch_enabled.  The corresponding metric
`dropped.queue_not_enabled` is dropped, along with `batching`.  The
batching is too ephemeral, especially considering a default batch time
of 20 ms, and is not shown in the dashboard, so it was removed.

Also, fixes a bug related to message loss in kafka producer when connection is down.  Currently, the kafka producer bridge will test the connection to kafka itself to say if the resource is connected.  However, if kafka or its connection is down, this’ll make messages to be lost, as there are no buffer workers for this bridge, and, being “down”, the resource won’t call the wolff producers, leading to both message loss and wrong metrics (as there won’t be failed counters bumped).


thalesmg:
Based/depends on https://github.com/emqx/emqx/pull/9619

apps/emqx_bridge/src/schema/emqx_bridge_schema.erl
```diff
@@ -113,12 +113,9 @@ fields(bridges) ->
     ] ++ ee_fields_bridges();
 fields("metrics") ->
     [
-        {"batching", mk(integer(), #{desc => ?DESC("metric_batching")})},
```
zmstone, line 116: deprecate it.

zmstone: ah, sorry, if this is only for 'GET' requests, then it's fine to delete.

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -80,11 +98,13 @@ start_link(Id, Index, Opts) ->
 sync_query(Id, Request, Opts) ->
     PickKey = maps:get(pick_key, Opts, self()),
     Timeout = maps:get(timeout, Opts, infinity),
+    emqx_resource_metrics:matched_inc(Id),
```
thalesmg, line 101: Matched moved here because otherwise retries would bump the `matched` metric as if they were new requests.

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -608,32 +760,24 @@ maybe_append_queue(Id, Index, Q, Items) ->
...
 
+-spec get_first_n_from_queue(replayq:q(), pos_integer()) ->
+    empty | {replayq:q(), replayq:ack_ref(), [?Q_ITEM(?QUERY(_From, _Request, _HasBeenSent))]}.
 get_first_n_from_queue(Q, N) ->
-    get_first_n_from_queue(Q, N, []).
-
-get_first_n_from_queue(_Q, 0, Acc) ->
-    lists:reverse(Acc);
-get_first_n_from_queue(Q, N, Acc) when N > 0 ->
-    case replayq:peek(Q) of
```
thalesmg, line 621: Changed to pop because, previously, the head of the queue would be duplicated N times.

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -674,12 +816,14 @@ inflight_size(Name) ->
 
 inflight_append(undefined, _Ref, _Query, _Id, _Index) ->
     ok;
-inflight_append(Name, Ref, [?QUERY(_, _, _) | _] = Batch, Id, Index) ->
-    ets:insert(Name, {Ref, [?QUERY(From, Req, true) || ?QUERY(From, Req, _) <- Batch]}),
+inflight_append(Name, Ref, [?QUERY(_, _, _) | _] = Batch0, Id, Index) ->
+    Batch = mark_as_sent(Batch0),
+    ets:insert(Name, {Ref, Batch}),
     emqx_resource_metrics:inflight_set(Id, Index, inflight_size(Name)),
```
thalesmg, line 822: Fixme: if there are batches in the inflight ETS, the inflight metric will track number of batches rather than number of messages.

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -405,64 +556,65 @@ batch_reply_caller(Id, BatchResult, Batch) ->
...
+    handle_query_result(Id, Result, HasBeenSent, BlockWorker).
 
-handle_query_result(Id, ?RESOURCE_ERROR_M(exception, Msg), HasSent, BlockWorker) ->
+handle_query_result(Id, ?RESOURCE_ERROR_M(exception, Msg), HasBeenSent, BlockWorker) ->
     ?SLOG(error, #{msg => resource_exception, info => Msg}),
-    inc_sent_failed(Id, HasSent),
+    inc_sent_failed(Id, HasBeenSent),
     BlockWorker;
-handle_query_result(_Id, ?RESOURCE_ERROR_M(NotWorking, _), _HasSent, _) when
+handle_query_result(_Id, ?RESOURCE_ERROR_M(NotWorking, _), _HasBeenSent, _) when
```
thalesmg, line 581: Fixme: if the resource is down and a simple sync query is done (e.g.: kafka producer), it'll not be enqueued (there're no buffer workers) nor sent to the producer process: the message will be lost, `matched` will have increased, but no failure metrics increase.

thalesmg: https://github.com/emqx/emqx/pull/9642/commits/7155ecc086a81a519411e97e62ae08217be92810

thalesmg:
Fixme: the telemetry handler ID is changed everytime the resource is created, leading to duplicate metrics.

changes/v5.0.14-zh.md
```diff
@@ -7,8 +7,17 @@
   `env EMQX_BRIDGES__MQTT__XYZ__SERVER='"localhost:1883"'`。
   此修复后，可以不使用引号，例如 `env EMQX_BRIDGES__MQTT__XYZ__SERVER='localhost:1883'`。
 
+- 废弃了桥梁/资源的`enable_batch`和`enable_queue`选项 [#9642](https://github.com/emqx/emqx/pull/9642)。
```
zmstone, line 10:
```suggestion
- 废弃了桥接的 `enable_batch` 和 `enable_queue` 配置项 [#9642](https://github.com/emqx/emqx/pull/9642)。
```

changes/v5.0.14-zh.md
```diff
@@ -7,8 +7,17 @@
...
+  在这一改变之后，队列总是被启用，而批处理由`batch_size`选项控制：`batch_size > 1`意味着批处理将被启用。
```
zmstone, line 11:
```suggestion
  在这一改变之后，桥接的工作进程总是启用缓存队列，而批处理由 `batch_size` 选项控制：`batch_size > 1` 则意味着启用批处理。
```

changes/v5.0.14-zh.md
```diff
@@ -7,8 +7,17 @@
...
+
 ## 修复
 
 - 修复了测试GCP PubSub可能泄露内存的问题，以及其JWT令牌第二次刷新失败的问题。 [#9640](https://github.com/emqx/emqx/pull/9640)
 
 - 修复 MySQL 驱动断开连接时出现的数据丢失和匹配错误的问题 [#9638](https://github.com/emqx/emqx/pull/9638)。
+
+- 修复一些可能导致错误桥接指标的问题[#9642](https://github.com/emqx/emqx/pull/9642)。
```
zmstone, line 19:
```suggestion
- 修复一些可能导致错误桥接指标的问题 [#9642](https://github.com/emqx/emqx/pull/9642)。
```

changes/v5.0.14-zh.md
```diff
@@ -7,8 +7,17 @@
...
+
+- 修复当Kafka或其连接中断时，可能导致Kafka Producer桥的消息丢失和错误指标的问题[#9642](https://github.com/emqx/emqx/pull/9642)。
```
zmstone, line 21:
```suggestion
- 修复当 Kafka 或其连接中断时，可能导致 Kafka 桥接的消息丢失和错误指标的问题 [#9642](https://github.com/emqx/emqx/pull/9642)。
```

changes/v5.0.14-zh.md
```diff
@@ -7,8 +7,17 @@
...
+
+- 修复一些问题，这些问题可能导致在为桥接使用批处理时，同一消息被多次传递，以及批处理被重试时[#9642](https://github.com/emqx/emqx/pull/9642)。
```
zmstone, line 23:
```suggestion
- 修复桥接启用批处理情况下，当重试发生时，同一个消息被投递多次的问题 [#9642](https://github.com/emqx/emqx/pull/9642)。
```

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
```diff
@@ -1,8 +1,10 @@
 %%--------------------------------------------------------------------
-%% Copyright (c) 2022-2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%% Copyright (c) 2022 EMQ Technologies Co., Ltd. All Rights Reserved.
```
zmstone, line 2:
```suggestion
%% Copyright (c) 2022-2023 EMQ Technologies Co., Ltd. All Rights Reserved.
```

apps/emqx_resource/i18n/emqx_resource_schema_i18n.conf
```diff
@@ -126,8 +126,8 @@ When disabled the messages are buffered in RAM only."""
 
   batch_size {
     desc {
-      en: """Maximum batch count."""
-      zh: """批量请求大小。"""
+      en: """Maximum batch count. If equal to 1, there's effectively no batching."""
+      zh: """批量请求大小。如果等于1，实际上就没有批处理。"""
```
qzhuyan, line 130: 批量请求大小。如果设为1，则无批处理。

qzhuyan (approved):

qzhuyan (approved):

id (approved):

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -110,7 +129,9 @@ simple_async_query(Id, Request, ReplyFun) ->
     %% would mess up the metrics anyway.  `undefined' is ignored by
     %% `emqx_resource_metrics:*_shift/3'.
     Index = undefined,
-    Result = call_query(async, Id, Index, ?QUERY(ReplyFun, Request, false), #{}),
+    QueryOpts = #{},
+    emqx_resource_metrics:matched_inc(Id),
```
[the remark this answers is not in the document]
thalesmg: Makes sense.  But searching the code base, I could not find any usage of `block/2` :thinking: 

I think I'll just remove it.  :smiley_cat:

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -252,145 +274,274 @@ code_change(_OldVsn, State, _Extra) ->
...
+            {next_state, running, Data0};
+        {Q1, QAckRef, [?QUERY(_, Request, HasBeenSent) = Query]} ->
+            QueryOpts = #{inflight_name => Name},
             Result = call_query(configured, Id, Index, Query, QueryOpts),
-            case reply_caller(Id, ?REPLY(undefined, Request, HasSent, Result)) of
+            Reply = ?REPLY(undefined, Request, HasBeenSent, Result),
+            case reply_caller(Id, Reply) of
                 true ->
-                    {keep_state, St, {state_timeout, ResumeT, resume}};
+                    {keep_state, Data0, {state_timeout, ResumeT, resume}};
```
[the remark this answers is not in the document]
thalesmg: Before this refactoring, the only requests that did bump the `retried.*` counters were the async requests, and they are the only ones that do get inserted into the inflight table and marked as sent.

I'm not entirely sure this was the intended behavior indeed.  But I think that this behavior is kept because, after this refactoring, the path that an async request does is:

1. The request gets enqueued into replayq.
2. It's then popped (possible in a batch).
3. It's appended to the inflight table (if there's room).
4. In the first ("outer") `handle_query_result`, since it'll return `{async_return, ok}`, it'll be ack'ed in replayq, removing it from there.
5. Eventually, when `{batch_,}reply_after_query` is called, then it'll bump either `retried.*` counter.

For sync requests, you are right: they won't bump the `retried.*` counters.  Do you think that's the original intention?  We would need to track what has been retried somehow.  

We can't change the replayq items without re-appending them to the queue, and that would change the order of the requests, unfortunately.  One way could be to keep a table of hashes of requests and then check if those have been sent or not.  :thinking: 

What do you think?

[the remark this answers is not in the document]
lafirest: Personally, I think we can just ignore this here, then create a track ticket to the product owner, and left the discussion to them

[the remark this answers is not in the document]
thalesmg: Sounds good to me: we can revisit this point in a future PR, if required.  :smiley_cat:

ieQu1 (approved):

PR #9643, opened 2022-12-29
ci: simplify build_and_push_docker_images workflow

id:
Fixes https://emqx.atlassian.net/browse/EMQX-8409

Reference build: https://github.com/emqx/emqx/actions/runs/3802854325/jobs/6468734209

- delete scripts/docker-create-push-manifests.sh and instead build multi-platform images via buildx
- consequently, delete docker-meta action since it was only needed to support the above
- move elixir docker image build back to the same job
- enable triggers on docker-latest-ce and docker-latest-ee tags
- fix our custom labels

```
$ docker inspect emqx/emqx:latest
...
            "Labels": {
               ...
              "org.opencontainers.image.edition": "Opensource\\norg.opencontainers.image.otp.version=24.3.4.2-1",
...
```


zmstone:
we moved away from buildx so we can have arm builders to build arm images then a final stage to merge the manifests.

Rory-Z (approved):

qzhuyan:
buildx for arm64 platform is too slow takes hours to finish due to emulation
 is this solved?

deploy/docker/Dockerfile.alpine
```diff
@@ -1,6 +1,7 @@
 ARG BUILD_FROM=ghcr.io/emqx/emqx-builder/5.0-26:1.13.4-24.3.4.2-1-alpine3.15.1
 ARG RUN_FROM=alpine:3.15.1
-FROM ${BUILD_FROM} AS builder
+ARG BUILDPLATFORM=linux/amd64
```
qzhuyan, line 3: why this is needed?

.github/workflows/build_and_push_docker_images.yaml
```diff
@@ -156,229 +167,48 @@ jobs:
...
+          type=raw,value=latest,enable=${{ needs.prepare.outputs.IS_LATEST }}
+        labels: |
+          org.opencontainers.image.otp.version=${{ matrix.otp }}
+          org.opencontainers.image.edition=${{ needs.prepare.outputs.EDITION }}
+          ${{ steps.pre-meta.outputs.extra_labels }}
 
     - uses: docker/build-push-action@v3
       with:
-        push: ${{ needs.prepare.outputs.IS_EXACT_TAG == 'true' || github.repository_owner != 'emqx' }}
+        push: ${{ needs.prepare.outputs.IS_EXACT_TAG == 'true' && github.repository_owner == 'emqx' }}
```
qzhuyan, line 205: `|| github.repository_owner != 'emqx' ` is for testing in forked repo.
like:
https://hub.docker.com/r/qzhuyan/emqx-enterprise

deploy/docker/Dockerfile.alpine
id: to pick correct platform from multiplatform image, I've seen some issues when building on arm64 platform

id:
> buildx for arm64 platform is too slow takes hours to finish due to emulation is this solved?

@qzhuyan take a look https://github.com/emqx/emqx/actions/runs/3802854325/jobs/6468734209
regular build ~20 min, elixir build ~40 min.
Another one is running right now https://github.com/emqx/emqx/actions/runs/3900335948/jobs/6660920842

Stone suggested to use zip packages for corresponding platforms instead of building everything again, I think that should help a lot. But it will be another change.

qzhuyan (approved):

qzhuyan:
<30 mins isn't bad at all.

We used to run into issue when arm64 build timeout and never get succeed.

PR #9648, opened 2022-12-30
build: skip qlc modules if there were no git diff

zmstone:
All the listed modules with qlc.hrl included, qlc is the only header.
so check git diff of the module is safe.

https://emqx.atlassian.net/browse/EMQX-8655

### DIFF
--- a/scripts/update_appup.escript
+++ b/scripts/update_appup.escript
@@ -66,9 +66,54 @@ ignored_apps() ->
      emqx_modules_spec %% generic appup file for all versions
     ] ++ otp_standard_apps().
 
+%% modules with qlc.hrl included are not deterministic
+%% beam md5 may change even though nothing in the code is changed
+%% this should not be applied to other modules in general because
+%% a module may include other hrl files,
+%% or if it has parse-transform etc, beam may implicitly change.
+qlc_modules0() ->
+    ["src/emqx_cm.erl",
+     "apps/emqx_management/src/emqx_mgmt.erl",
+     "apps/emqx_management/src/emqx_mgmt_api.erl",
+     "apps/emqx_auth_mnesia/src/emqx_auth_mnesia_api.erl",
+     "apps/emqx_auth_mnesia/src/emqx_acl_mnesia_db.erl",
+     "apps/emqx_rule_engine/src/emqx_rule_registry.erl",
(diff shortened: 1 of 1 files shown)


scripts/update_appup.escript
```diff
@@ -66,9 +66,50 @@ ignored_apps() ->
...
+
+is_git_diff(BaseVsn, File) ->
+    CMD = "git diff --name-only "++ BaseVsn ++ "...HEAD " ++ File,
+    case string:trim(os:cmd(CMD)) of
+        "" ->
+            false;
+        File ->
+            true;
+        Other ->
+            logerr("Failed to git diff ~s:~n", [File]),
```
zmstone, line 96:
```suggestion
            logerr("Failed to git diff: ~s~n", [File]),
```

scripts/update_appup.escript
```diff
@@ -620,7 +664,7 @@ diff_app(UpOrDown, App,
             case UpOrDown =:= up of
                 true ->
                     %% only log for the upgrade case because it would be the same result
-                    logerr("Application '~p' contains changes, but its version is not updated. ~s",
+                    logerr("Application '~p' contains changes, but its version is not updated. ~s~n",
```
thalesmg, line 667: nit: maybe add the `~n` to `logerr`'s definition?

scripts/update_appup.escript
```diff
@@ -66,9 +66,50 @@ ignored_apps() ->
...
+                      {Module, #{path => F,
+                                 is_git_diff => is_git_diff(BaseVsn, F)
+                                }}
+              end, qlc_modules()).
+
 main(Args) ->
     #{prev_tag := Baseline} = Options = parse_args(Args, default_options()),
-    init_globals(Options),
+    QlcModules = qlc_module_git_diff(Baseline),
+    log("Changedd modules which have qlc.hrl included: ~p", [QlcModules]),
```
thalesmg, line 111:
```suggestion
    log("Changed modules which have qlc.hrl included: ~p", [QlcModules]),
```

thalesmg (approved):
maybe we'd need to assert that the listed files with QLC headers indeed always have only that header to guarantee the git diff output to be safe?

thalesmg (approved):

PR #9663, opened 2023-01-02
0102 fix stopped nodes detection

zmstone:


apps/emqx_conf/test/emqx_cluster_rpc_SUITE.erl
```diff
@@ -48,11 +48,14 @@ init_per_suite(Config) ->
     meck:new(emqx_alarm, [non_strict, passthrough, no_link]),
     meck:expect(emqx_alarm, activate, 3, ok),
     meck:expect(emqx_alarm, deactivate, 3, ok),
+    meck:new(mria_mnesia, [non_strict, passthrough, no_link]),
+    meck:expect(mria_mnesia, running_nodes, 0, [?NODE1, {node(), ?NODE2}, {node(), ?NODE3}]),
```
thalesmg, line 52: As `mria_mnesia:running_nodes/0` returns a list of nodes, returning a single node or a tuple of nodes doesn't seem a real case?  Maybe the intention was:

```suggestion
    meck:expect(mria_mnesia, running_nodes, 0, meck:seq([[?NODE1], [node(), ?NODE2], [node(), ?NODE3]])),
```

zmstone: the nodes are not real in this test.

thalesmg: My point was that the return type is probably incorrect for that function.  I made a quick local test and the fn was return strange stuff rather than the desired sequence of list of nodes.

zmstone: it should return a consistent (static) list of nodes for all calls in this test SUITE.

thalesmg: In that case, I think this would have the correct return type:

```suggestion
    meck:expect(mria_mnesia, running_nodes, 0, [?NODE1FQDN, ?NODE2FQDN, ?NODE3FQDN]),
```

where `?NODEXFQDN` is basically the existing macros but with the form `node@host`.

thalesmg: If this is passing even returning the wrong return types, I guess it's not really testing those calls.

thalesmg (approved):

apps/emqx_conf/test/emqx_cluster_rpc_SUITE.erl
zmstone: it's actually not just the test case, the production code doesn't rely on it being a real node name either.
the node name in cluster_rpc is essentially a table key, associated to which is transaction apply progress of each "node".
i.e. it does not really use the node name to perform rpc or messaging passing.

PR #9670, opened 2023-01-03
chore: use @emqx/emqx-review-board team in codeowners

id:
Team members according to https://emqx.atlassian.net/wiki/spaces/EMQX/overview#Code-Review-Board.

One approval from any of that team members should be enough for any PR.


thalesmg (approved):

id:
merging as test failure does not have anything to do with the changes.

PR #9681, opened 2023-01-04
feat(gcp_pubsub): generate jwt tokens on demand without workers (rv4.4)

thalesmg:
https://emqx.atlassian.net/browse/EMQX-8683

Rather than relying on a JWT worker to produce and refresh tokens, we could just produce them on demand.  That can generate a bit of extra work (as multiple processes might realize it’s time to refresh the JWT and do so), but that shouldn’t be much.  In return, we avoid any possibility of not having a fresh JWT when pushing messages.

# Note

base branch is currently `main-v4.4`


apps/emqx_rule_engine/src/emqx_rule_engine_jwt.erl
```diff
@@ -54,3 +76,67 @@ delete_jwt(TId, ResourceId) ->
         error:badarg ->
             ok
     end.
+
+%% @doc Attempts to retrieve a valid JWT from the cache.  If there is
+%% none or if the cached token is expired, generates an caches a fresh
+%% one.
+-spec ensure_jwt(jwt_config()) -> jwt().
+ensure_jwt(JWTConfig) ->
```
savonarola, line 84: Just curious: seems that new fun `ensure_jwt` is still called only by jwt worker, while according the desc it is to be called directly somewhere else 🤔

thalesmg: In this repo, yes, no other module calls this function.  But in the GCP PubSub rule resource/action, the intention is to call this when publishing messages to generate the authn header, superseding the need for the JWT worker.

PR #9697, opened 2023-01-06
chore: add i18n file for ldap connector

lafirest:


thalesmg (approved):

thalesmg:
This seems to not be referenced in the code?  I expected CI to fail if a translation would be missing :thinking:

lafirest:
> This seems to not be referenced in the code? I expected CI to fail if a translation would be missing 🤔


I think the reason is that there was no `?DESC` used in the old code

PR #9703, opened 2023-01-09
fix: missing default value for qos field of API /clients/:clientid/subscribe

terry-xiaoyu:
Set the default value of qos to 0, which is the default value of the API `/mqtt/subscribe` and `/mqtt/subscribe_batch` in emqx 4.x.

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -549,7 +549,7 @@ fields(keepalive) ->
     ];
 fields(subscribe) ->
     [
-        {topic, hoconsc:mk(binary(), #{desc => <<"Topic">>})},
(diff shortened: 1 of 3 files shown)


qzhuyan (dismissed):

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -550,7 +550,7 @@ fields(keepalive) ->
 fields(subscribe) ->
     [
         {topic, hoconsc:mk(binary(), #{desc => <<"Topic">>})},
```
terry-xiaoyu, line 552: We may also set `topic` as required field.

terry-xiaoyu: fixed

thalesmg (approved):
Nit: maybe add a test case?

lafirest (approved):

PR #9705, opened 2023-01-09
fix: remove the default value of webhook body field

terry-xiaoyu:
- if the default value is set to "", then we send "" as the http body.
- if the default value is not set, then we send all the available fields as the http body.

### DIFF
--- a/apps/emqx_bridge/i18n/emqx_bridge_webhook_schema.conf
+++ b/apps/emqx_bridge/i18n/emqx_bridge_webhook_schema.conf
@@ -93,11 +93,20 @@ HTTP 请求的标头。<br/>
                    desc {
                          en: """
 The body of the HTTP request.<br/>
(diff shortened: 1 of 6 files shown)


apps/emqx_bridge/i18n/emqx_bridge_webhook_schema.conf
```diff
@@ -93,10 +93,19 @@ HTTP 请求的标头。<br/>
                    desc {
                          en: """
 The body of the HTTP request.<br/>
+If not provided, the body will be a JSON object of all the available fields.<br/>
+The <code>'all available fields'</code> will be the context of a MQTT message when
```
thalesmg, line 97:
```suggestion
Here, "all available fields" means the context of a MQTT message when
```

thalesmg (approved):

JimMoen (approved):

zhongwencool (approved):

PR #9708, opened 2023-01-09
EMQX 8588 refactor emqx resource manager

sstrigler:
Ticket: https://emqx.atlassian.net/browse/EMQX-8588

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_api.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_api.erl
@@ -485,7 +485,7 @@ schema("/bridges_probe") ->
     RequestMeta = #{module => ?MODULE, method => post, path => "/bridges_probe"},
     case emqx_dashboard_swagger:filter_check_request_and_translate_body(Request, RequestMeta) of
         {ok, #{body := #{<<"type">> := ConnType} = Params}} ->
-            case do_probe(ConnType, maps:remove(<<"type">>, Params)) of
+            case emqx_bridge_resource:create_dry_run(ConnType, maps:remove(<<"type">>, Params)) of
                 ok ->
                     {204};
                 {error, Error} ->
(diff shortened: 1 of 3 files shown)


sstrigler:
Does this one still need a changelog entry since it is a follow up to https://emqx.atlassian.net/browse/EMQX-8571

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -485,57 +485,22 @@ schema("/bridges_probe") ->
     RequestMeta = #{module => ?MODULE, method => post, path => "/bridges_probe"},
     case emqx_dashboard_swagger:filter_check_request_and_translate_body(Request, RequestMeta) of
         {ok, #{body := #{<<"type">> := ConnType} = Params}} ->
-            case do_probe(ConnType, maps:remove(<<"type">>, Params)) of
+            case emqx_bridge_resource:create_dry_run(ConnType, maps:remove(<<"type">>, Params)) of
                 ok ->
                     {204};
                 {error, Error} ->
-                    {400, error_msg('TEST_FAILED', Error)}
+                    {400, error_msg('TEST_FAILED', try_json_encode(Error))}
```
kjellwinblad, line 492: This might be very nit-picky but I'm thinking about the response code for the error case. When reading here https://developer.mozilla.org/en-US/docs/Web/HTTP/Status it seems like 4xx error codes are supposed to indicate client errors (which I interpret as the client has done something it is not really supposed to do). But this error case happens when the health check of the resource fails which I don't think is a client error. However, this response code was used before the PR as well so I guess this is already a part of the API so feel free to just ignore this comment.

kjellwinblad, line 492: This might be very nit-picky but I'm thinking about the response code for the error case. When reading here https://developer.mozilla.org/en-US/docs/Web/HTTP/Status it seems like 4xx error codes are supposed to indicate client errors (which I interpret as the client has done something it is not really supposed to do). But this error case happens when the health check of the resource fails which I don't think is a client error. However, this response code was used before the PR as well so I guess this is already a part of the API so feel free to just ignore this comment.

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -485,57 +485,22 @@ schema("/bridges_probe") ->
...
-        [Scheme, UrlRem] ->
-            case string:split(UrlRem, "/", leading) of
-                [HostPort, Path] ->
-                    {iolist_to_binary([Scheme, "//", HostPort]), Path};
-                [HostPort] ->
-                    {iolist_to_binary([Scheme, "//", HostPort]), <<>>}
-            end;
-        [Url] ->
-            error({invalid_url, Url})
+try_json_encode(Term) ->
```
kjellwinblad, line 498: It seems like error_msg/2 calls emqx:readable_error_msg/1 to try to make the Error term into a readable string. Why does it need to be json encoded?

kjellwinblad (approved):
Looks good to me. The comments I gave are just due to some small questions I got while looking at the code and not things that I believe need to be fixed.

apps/emqx_bridge/src/emqx_bridge_api.erl
sstrigler: You're totally right, making it as human readable as possible sure is the way to go.

sstrigler: The problem is that for all the cases I tested against `emqx_misc:readable_error_msg/1` apparently didn't produce something meaningful and instead just returned a representation of the erlang data structure. Which is less readable than JSON. I thought. But now we'd always have JSON without even trying. Maybe the wrong place to fix this. I'll stick to what we do in general. But to address that in a more generic way.

apps/emqx_bridge/src/emqx_bridge_api.erl
kjellwinblad: I think it would make more sense to give success responses in both cases and let the message body describe if the dry run was successful or not. This is really nit-picky and just my opinion and a change now would break the API and would at least require a change of the UI. So please just ignore my comments regarding this.

PR #9717, opened 2023-01-10
fix(emqx_resource_manager): do not start when disabled

zmstone:
do not start bridge after creation if 'enable' config is set to 'false'.

when testing with an unreachable host (which causes gen_tcp:connect to hang for a long time).
it's impossible to change other configs because the resource manager always tries to connect the server even when it is disabled due to below error (causing the config handler not to persist the updated config):

```
[warning] entrypoint: <<"emqx:update_config/3">>, kind: initiate, line: 500, mfa: emqx_cluster_rpc:log_and_alarm/3, msg: cluster_rpc_apply_result, result: {error,{post_config_update,emqx_bridge,timeout}}, tnx_id: 7
```


terry-xiaoyu (approved):

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -122,7 +122,10 @@ create_and_return_data(MgrId, ResId, Group, ResourceType, Config, Opts) ->
 -define(START_TIMEOUT, 5000).
 
 %% @doc Create a resource_manager and wait until it is running
-create(MgrId, ResId, Group, ResourceType, Config, Opts) ->
+create(MgrId, ResId, Group, ResourceType, Config, Opts0) ->
+    Enabled = maps:get(enable, Config, true),
```
terry-xiaoyu, line 126: It is good, but beaks the design that emqx-resource never depends on the details of the `Config`. What the emqx-resource knows are the `ResId`, `ResType` and`Opts`, but the `Config` should be transparent to the `emqx-resource` level.

terry-xiaoyu: In authz, we changed the opt `start_after_create` to false [here](https://github.com/emqx/emqx/blob/6bbb5edb536957425f11fc7f7ce38032cbf97208/apps/emqx_authz/src/emqx_authz_utils.erl#L43), and then start the resource only when enabled [here](https://github.com/emqx/emqx/blob/6bbb5edb536957425f11fc7f7ce38032cbf97208/apps/emqx_authz/src/emqx_authz_utils.erl#L62)

terry-xiaoyu (approved):

zhongwencool (approved):

PR #9718, opened 2023-01-10
0108 refactor authn schema union member selector

zmstone:
https://emqx.atlassian.net/browse/EMQX-8657

Also got rid of the `lazy` type which forced us to validate the config during node boot sequence, but not before.

Verified with below test script:
```
env EMQX_AUTHENTICATION__1='{mechanism="password_based",backend="mysql",server="localhost:3306",database="emqx",username="emqx",password="emqx",query="SELECT password_hash, salt, is_superuser FROM mqtt_user WHERE username = ${username} LIMIT 1",password_hash_algorithm={name=sha256,salt_position=prefix},enable=true}' \
    EMQX_DASHBOARD__default_password='ppp***' \
    EMQX_AUTHENTICATION__1__backend=yoursql \
    EMQX_NODE__COOKIE=foobar \
    ./_build/emqx/rel/emqx/bin/emqx check_config
```
## before the fix

`check_config` command does not find the error,
instead, it will crash when starting with quite a few lines of  rather big crash logs like:

```
Erlang/OTP 24 [erts-12.3.2.2] [emqx] [64-bit] [smp:20:20] [ds:20:20:8] [async-threads:4] [jit]

Listener ssl:default on 0.0.0.0:8883 started.
Listener tcp:default on 0.0.0.0:1883 started.
Listener ws:default on 0.0.0.0:8083 started.
Listener wss:default on 0.0.0.0:8084 started.
Listener http:dashboard on :18083 started.
2023-01-11T19:20:14.501565+01:00 [error] line: 71, mfa: emqx_authn_app:initialize/0, msg: failed_to_initialize_authentication, reason: #{error => unknown_backend,value => <<"yoursql">>}
2023-01-11T19:20:14.502160+01:00 [error] crasher: initial call: application_master:init/4, pid: <0.2513.0>, registered_name: [], exit: {{{failed_to_initialize_authentication,#{error => unknown_backend,value => <<"yoursql">>}},{emqx_authn_app,start,[normal,[]]}},[{application_master,init,4,[{file,"application_master.erl"},{line,142}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,226}]}]}, ancestors: [<0.2512.0>], message_queue_len: 1, messages: [{'EXIT',<0.2514.0>,normal}], links: [<0.2512.0>,<0.1705.0>], dictionary: [], trap_exit: true, status: running, heap_size: 376, stack_size: 29, reductions: 167; neighbours:
2023-01-11T19:20:14.502623+01:00 [critical] app: emqx_authn, line: 88, mfa: emqx_machine_boot:start_one_app/1, msg: failed_to_start_app, reason: {emqx_authn,{{failed_to_initialize_authentication,#{error => unknown_backend,value => <<"yoursql">>}},{emqx_authn_app,start,[normal,[]]}}}
......a lot of lines ....
stack_size: 29, reductions: 170; neighbours:
Stop listener http:dashboard on :18083 successfully.
Listener ssl:default on 0.0.0.0:8883 stopped.
Listener tcp:default on 0.0.0.0:1883 stopped.
Listener ws:default on 0.0.0.0:8083 stopped.
Listener wss:default on 0.0.0.0:8084 stopped.
[os_mon] memory supervisor port (memsup): Erlang has closed
[os_mon] cpu supervisor port (cpu_sup): Erlang has closed
{"Kernel pid terminated",application_controller,"{application_start_failure,emqx_machine,{{shutdown,....
.... a lot of lines ....
Crash dump is being written to: log/erl_crash.dump...done
```

## after fix:

```
2023-01-13T14:25:14.768426+01:00 [error] failed_to_check_schema: emqx_conf_schema
2023-01-13T14:25:14.771508+01:00 [error] #{backend => <<"yoursql">>,kind => validation_error,path => "authentication.1",reason => "unknown_backend"}
ERROR: call_hocon_failed: -v -t 2023.01.13.14.25.14 -s emqx_conf_schema -c /mnt/code/src/emqx/0/_build/emqx/rel/emqx/etc/emqx.conf -d /mnt/code/src/emqx/0/_build/emqx/rel/emqx/data/configs generate
```

## This PR also sunsets the never-documented, over-engineered authn extension mechanism

Develop a new Erlang app which
  * must use HOCON for configs
  * implement a `check_config/` callback for config validation
  * implement `create/2` callback called at boot
  * implement `update/2` callback to handle runtime config updates
  * register itself with authn providers at boot time

With all the above complexity, the gains are:
* Be able to configure the extended authn provider as a part of an authn chain
* Be able to re-configure from the HTTP API
* Maybe after some new feature to be add to dashboard, it's possible to configure the chain from UI

NOTE: the old way of extending authn is always supported: implement a plugin with hooks to 'client.authenticate' hook-point.

### DIFF
--- a/apps/emqx/src/emqx_authentication_config.erl
+++ b/apps/emqx/src/emqx_authentication_config.erl
@@ -136,7 +136,7 @@ do_pre_config_update({move_authenticator, _ChainName, AuthenticatorID, Position}
 ) ->
     ok | {ok, map()} | {error, term()}.
 post_config_update(_, UpdateReq, NewConfig, OldConfig, AppEnvs) ->
-    do_post_config_update(UpdateReq, check_configs(to_list(NewConfig)), OldConfig, AppEnvs).
+    do_post_config_update(UpdateReq, to_list(NewConfig), OldConfig, AppEnvs).
 
 do_post_config_update({create_authenticator, ChainName, Config}, NewConfig, _OldConfig, _AppEnvs) ->
     NConfig = get_authenticator_config(authenticator_id(Config), NewConfig),
@@ -174,56 +174,6 @@ do_post_config_update(
     _AppEnvs
 ) ->
     emqx_authentication:move_authenticator(ChainName, AuthenticatorID, Position).
-
-check_configs(Configs) ->
-    Providers = emqx_authentication:get_providers(),
-    lists:map(fun(C) -> do_check_config(C, Providers) end, Configs).
-
-do_check_config(Config, Providers) ->
-    Type = authn_type(Config),
-    case maps:get(Type, Providers, false) of
-        false ->
-            ?SLOG(warning, #{
-                msg => "unknown_authn_type",
-                type => Type,
-                providers => Providers
-            }),
-            throw({unknown_authn_type, Type});
-        Module ->
-            do_check_config(Type, Config, Module)
-    end.
-
-do_check_config(Type, Config, Module) ->
-    F =
-        case erlang:function_exported(Module, check_config, 1) of
-            true ->
-                fun Module:check_config/1;
-            false ->
-                fun(C) ->
-                    Key = list_to_binary(?EMQX_AUTHENTICATION_CONFIG_ROOT_NAME),
-                    AtomKey = list_to_atom(?EMQX_AUTHENTICATION_CONFIG_ROOT_NAME),
-                    R = hocon_tconf:check_plain(
-                        Module,
-                        #{Key => C},
-                        #{atom_key => true}
-                    ),
-                    maps:get(AtomKey, R)
-                end
-        end,
-    try
-        F(Config)
-    catch
-        C:E:S ->
-            ?SLOG(warning, #{
-                msg => "failed_to_check_config",
-                config => Config,
-                type => Type,
-                exception => C,
-                reason => E,
-                stacktrace => S
-            }),
-            throw({bad_authenticator_config, #{type => Type, reason => E}})
-    end.
 
 to_list(undefined) -> [];
 to_list(M) when M =:= #{} -> [];
--- a/apps/emqx/test/emqx_authentication_SUITE.erl
+++ b/apps/emqx/test/emqx_authentication_SUITE.erl
@@ -53,48 +53,8 @@
 ).
 
 %%------------------------------------------------------------------------------
-%% Hocon Schema
-%%------------------------------------------------------------------------------
-
-roots() ->
-    [
-        {config, #{
-            type => hoconsc:union([
-                hoconsc:ref(?MODULE, type1),
-                hoconsc:ref(?MODULE, type2)
-            ])
-        }}
-    ].
-
-fields(type1) ->
-    [
-        {mechanism, {enum, [password_based]}},
-        {backend, {enum, [built_in_database]}},
-        {enable, fun enable/1}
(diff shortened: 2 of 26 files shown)


apps/emqx/src/emqx_authentication_config.erl
```diff
@@ -136,7 +136,7 @@ do_pre_config_update({move_authenticator, _ChainName, AuthenticatorID, Position}
 ) ->
     ok | {ok, map()} | {error, term()}.
 post_config_update(_, UpdateReq, NewConfig, OldConfig, AppEnvs) ->
-    do_post_config_update(UpdateReq, check_configs(to_list(NewConfig)), OldConfig, AppEnvs).
```
zmstone, line 139: Since `authentication` root field's type is no loner 'lazy',
`NewConfig` is already type checked, so there is no need to check it again here.

apps/emqx/test/emqx_authentication_SUITE.erl
```diff
@@ -326,14 +286,14 @@ t_update_config(Config) when is_list(Config) ->
     ok = register_provider(?config("auth2"), ?MODULE),
     Global = ?config(global),
     AuthenticatorConfig1 = #{
-        <<"mechanism">> => <<"password_based">>,
-        <<"backend">> => <<"built_in_database">>,
-        <<"enable">> => true
+        mechanism => password_based,
+        backend => built_in_database,
+        enable => true
```
savonarola, line 291: Why do we need to change to atom? 🤔

savonarola:
Great UX improvement!

apps/emqx/test/emqx_authentication_SUITE.erl
zmstone: It's because this SUITE is now testing configs without type check.
Previously, there was a (hidden API contract) `check_config/1` callback (implemented in this module, see old code) which does the type check.
--- Having to support this callback type check was essentially the reason why we had to use LAZY type and caused all the complexity!

Now, we have to admit that this way of extending authn functionality is over-engineered:
it's way much more straightforward to just add another Erlang app (e.g. a plugin) which hooks to 'client.authenticate' hook point.

apps/emqx_authn/src/emqx_authn_schema.erl
```diff
@@ -45,24 +45,75 @@ enable(desc) -> ?DESC(?FUNCTION_NAME);
...
+                    try_select_union_member(Module, Value)
+            end
+    end;
+select_union_member(Value, _Providers) when is_map(Value) ->
+    throw(#{reason => "missing_mechanism_field"});
+select_union_member(Value, _Providers) ->
+    throw(#{reason => "not_a_struct", value => Value}).
+
+try_select_union_member(Module, Value) ->
+    %% some modules have union_member_selector/1 exported to help selectin the sub-types
```
savonarola, line 108:
```suggestion
    %% some modules have `union_member_selector/1` exported to help selecting the sub-types
```

savonarola (approved):
Looks good to me. I also experimented a bit starting node with different configs & updating it, got comprehensive error messages.

id (approved):
My approval has little value since it's only about log message in `bin/emqx`

zmstone:
> My approval has little value since it's only about log message in `bin/emqx`

Thanks.

PR #9726, opened 2023-01-11
fix(api): augment paged search responses with `hasnext` flag

keynslug:
This flag indicates whether there are more results available on the next pages. It is needed in cases when the total number of search results is not known in advance (e.g. when client requests fuzzy search).

Also, in such cases there's no `count` field in responses anymore because responding with `0` was confusing for clients: it's not possible to differentiate between "there are no results" and "we don't know how much".

---

[EMQX-8702](https://emqx.atlassian.net/browse/EMQX-8702)

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api.erl
@@ -210,6 +210,8 @@ cluster_query(Tab, QString, QSchema, MsFun, FmtFun) ->
     end.
 
 %% @private
+do_cluster_query([], QueryState, ResultAcc) ->
+    finalize_query(ResultAcc, mark_complete(QueryState));
 do_cluster_query(
     [Node | Tail] = Nodes,
     QueryState,
--- a/apps/emqx_management/src/emqx_mgmt_api.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api.erl
@@ -422,10 +422,11 @@ accumulate_query_rows(
         NCursor when NCursor < PageStart ->
             {more, ResultAcc#{cursor => NCursor}};
         NCursor when NCursor < PageEnd ->
+            SubRows = lists:nthtail(max(0, PageStart - Cursor - 1), Rows),
             {more, ResultAcc#{
                 cursor => NCursor,
-                count => Count + length(Rows),
-                rows => [{Node, Rows} | RowsAcc]
+                count => Count + length(SubRows),
+                rows => [{Node, SubRows} | RowsAcc]
             }};
         NCursor when NCursor >= PageEnd ->
             SubRows = lists:sublist(Rows, Limit - Count),
--- a/apps/emqx_management/src/emqx_mgmt_api.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api.erl
@@ -20,7 +20,6 @@
 
 -elvis([{elvis_style, dont_repeat_yourself, #{min_complexity => 100}}]).
 
--define(FRESH_SELECT, fresh_select).
 -define(LONG_QUERY_TIMEOUT, 50000).
 
 -export([
@@ -174,13 +173,12 @@ do_node_query(
     case do_query(Node, QueryState) of
         {error, {badrpc, R}} ->
             {error, Node, {badrpc, R}};
-        {Rows, NQueryState = #{continuation := ?FRESH_SELECT}} ->
-            {_, NResultAcc} = accumulate_query_rows(Node, Rows, NQueryState, ResultAcc),
-            NResultAcc;
-        {Rows, NQueryState} ->
+        {Rows, NQueryState = #{complete := Complete}} ->
             case accumulate_query_rows(Node, Rows, NQueryState, ResultAcc) of
                 {enough, NResultAcc} ->
-                    NResultAcc;
+                    finalize_query(NResultAcc, NQueryState);
+                {_, NResultAcc} when Complete ->
+                    finalize_query(NResultAcc, NQueryState);
                 {more, NResultAcc} ->
                     do_node_query(Node, NQueryState, NResultAcc)
             end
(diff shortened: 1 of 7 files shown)


apps/emqx_management/src/emqx_mgmt_api.erl
```diff
@@ -390,40 +403,51 @@ counting_total_fun(_QueryState = #{fuzzy_fun := FuzzyFun}) when FuzzyFun =/= und
...
                 count => Count + length(SubRows),
-                total => TotalAcc,
-                rows => [{Node, SubRows} | RowsAcc]
+                rows => [{Node, SubRows} | RowsAcc],
+                partial => (Limit - Count) < Len
             }}
     end.
 
+finalize_query(Result = #{partial := Partial}, QueryState = #{complete := Complete}) ->
+    HasNext = Partial orelse not Complete,
```
zmstone, line 442: If the last page is partial, doe it mean it will always imply `hasnext` even reached end of table ?

apps/emqx_management/src/emqx_mgmt_api.erl
```diff
@@ -212,8 +210,6 @@ cluster_query(Tab, QString, QSchema, MsFun, FmtFun) ->
     end.
 
 %% @private
-do_cluster_query([], _QueryState, ResultAcc) ->
-    ResultAcc;
```
thalesmg, line 216: I think we could keep this clause.  When `mria` stops (e.g.: when it restarts for a node to join a cluster), then `mria_mnesia:running_nodes/0` will return an empty list.

apps/emqx_management/src/emqx_mgmt_api.erl
keynslug: Yes. This naming is probably a bit confusing, `partial` here means "only part of the last set of rows ended up in the results". Maybe something like `overflow` would be less confusing?

apps/emqx_management/src/emqx_mgmt_api.erl
```diff
@@ -266,13 +261,14 @@ collect_total_from_tail_nodes(Nodes, QueryState, ResultAcc = #{total := TotalAcc
 %%--------------------------------------------------------------------
 
 %% QueryState ::
-%%  #{continuation := ets:continuation(),
+%%  #{continuation => ets:continuation(),
 %%    page := pos_integer(),
 %%    limit := pos_integer(),
-%%    total := [{node(), non_neg_integer()}],
+%%    total := #{node() => non_neg_integer()},
```
thalesmg, line 267: Nano nit: `total` is optional?

```suggestion
%%    total => #{node() => non_neg_integer()},
```

apps/emqx_management/src/emqx_mgmt_api.erl
```diff
@@ -266,13 +261,14 @@ collect_total_from_tail_nodes(Nodes, QueryState, ResultAcc = #{total := TotalAcc
 %%--------------------------------------------------------------------
 
 %% QueryState ::
```
thalesmg, line 263: Unrelated to this PR.  I wonder if this could be a proper typespec and have dialyzer check it.  :thinking:

apps/emqx_management/test/emqx_mgmt_api_subscription_SUITE.erl
```diff
@@ -84,40 +83,78 @@ t_subscription_api(_) ->
...
+
+path() ->
+    emqx_mgmt_api_test_util:api_path(["subscriptions"]).
+
+init_per_testcase(_TC, Config) ->
+    {ok, Client} = emqtt:start_link(#{username => ?USERNAME, clientid => ?CLIENTID, proto_ver => v5}),
+    {ok, _} = emqtt:connect(Client),
+    [{client, Client} | Config].
 
+end_per_testcaser(_TC, Config) ->
```
thalesmg, line 158:
```suggestion
end_per_testcase(_TC, Config) ->
```

apps/emqx_management/src/emqx_mgmt_api.erl
```diff
@@ -318,47 +328,50 @@ do_select(
...
-            _ ->
+            Continuation ->
                 %% XXX: Repair is necessary because we pass Continuation back
                 %% and forth through the nodes in the `do_cluster_query`
                 ets:select(ets:repair_continuation(Continuation, Ms))
         end,
     case Result of
-        '$end_of_table' ->
-            {[], QueryState#{continuation => ?FRESH_SELECT}};
+        {Rows, '$end_of_table'} ->
```
thalesmg, line 347: `ets:continuation()` says it's an opaque value.  Is it safe to match its value here?

apps/emqx_management/src/emqx_mgmt_api.erl
keynslug: Certainly. There's a followup planned where we want to redesign these APIs quite a bit. There I plan to refactor this code substantially and spice it up with typespecs.

apps/emqx_management/src/emqx_mgmt_api.erl
keynslug: Heh, [the code says](https://github.com/erlang/otp/blob/6edaed6e23779057965231abd4e4b0421d566ec7/lib/stdlib/src/ets.erl#L52-L54) it's not opaque (as opposed to the docs). Also, dialyzer seems to be happy with this line when run manually.

apps/emqx_management/src/emqx_mgmt_api.erl
keynslug: Brought the clause back.

zmstone (approved):

apps/emqx_management/src/emqx_mgmt_api.erl
thalesmg: I think the docs say it's opaque because, although it's defined in the module, it's not exported.

PR #9728, opened 2023-01-11
EMQX-8625: add bridges probe api for kafka and mysql

sstrigler:
https://emqx.atlassian.net/browse/EMQX-8625

### DIFF
--- a/apps/emqx_connector/src/emqx_connector_http.erl
+++ b/apps/emqx_connector/src/emqx_connector_http.erl
@@ -387,11 +387,6 @@ on_get_status(_InstId, #{pool_name := PoolName, connect_timeout := Timeout} = St
         ok ->
             {connected, State};
         {error, Reason} ->
-            ?SLOG(error, #{
-                msg => "http_connector_get_status_failed",
-                reason => Reason,
-                state => State
-            }),
             {disconnected, State, Reason}
     end.
 
@@ -404,7 +399,7 @@ do_get_status(PoolName, Timeout) ->
                     ok;
                 {error, Reason} = Error ->
                     ?SLOG(error, #{
-                        msg => "ehttpc_health_check_failed",
+                        msg => "http_connector_get_status_failed",
                         reason => Reason,
                         worker => Worker
                     }),
--- a/apps/emqx_connector/src/emqx_connector_http.erl
+++ b/apps/emqx_connector/src/emqx_connector_http.erl
@@ -384,14 +384,15 @@ on_query_async(
 
 on_get_status(_InstId, #{pool_name := PoolName, connect_timeout := Timeout} = State) ->
     case do_get_status(PoolName, Timeout) of
-        true ->
-            connected;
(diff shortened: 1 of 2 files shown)


apps/emqx_connector/src/emqx_connector_http.erl
```diff
@@ -384,14 +384,15 @@ on_query_async(
...
 on_get_status(_InstId, #{pool_name := PoolName, connect_timeout := Timeout} = State) ->
     case do_get_status(PoolName, Timeout) of
-        true ->
-            connected;
-        false ->
+        ok ->
+            {connected, State};
+        {error, Reason} ->
             ?SLOG(error, #{
                 msg => "http_connector_get_status_failed",
```
thalesmg, line 391: Tiny nit: this log seems repeated with l. 406?  Maybe we could keep only the latter and rename that message to `http_connector_get_status_failed`, since we have more context at that line?

apps/emqx_connector/src/emqx_connector_mysql.erl
```diff
@@ -106,8 +104,8 @@ on_start(
     Options = [
         {host, Host},
         {port, Port},
-        {user, User},
-        {password, Password},
+        {user, maps:get(username, Config, <<"root">>)},
+        maybe_password_opt(maps:get(password, Config, undefined)),
```
thalesmg, line 108: 1) Maybe we should rather set the default value for `username` in the MySQL connector schema?  That way, I think it would be reflected in the generated documentation.
2) `maybe_password_opt` will return an empty list if the password is `undefined`, so `Options` will be `[{_, _}, ..., [], ... {_, _}]`.  Perhaps the intention was to concat the options with the result of that fn (first wrapping `{password, _}` in a list)?

```erlang
Options = [maybe_password_opt(Password) | ...]
```

thalesmg (commented):
The title mentions kafka and mysql, but here the HTTP bridge has a fix and none for kafka?

apps/emqx_connector/src/emqx_connector_mysql.erl
terry-xiaoyu: Sorry I've changed the same code on the master branch, so you have to solve the conflict before merging to master.

sstrigler: You're right, the default value should be reflected at the documentation. It's not quite straight-forward to change that though since the definition is shared across multiple schemas. So unfortunately a bit ugly code.

As of `maybe_password_opt`, it's fine to have an empty list as an element in a proplist, even though the spec tells differently. I found, it isn't quite as readable to have the password coming first. That's why I chose the other approach but if this confuses people it's better to use your way then.

sstrigler: > Sorry I've changed the same code on the master branch, so you have to solve the conflict before merging to master.

Will deal with it when merging the integration branch.

PR #9730, opened 2023-01-11
fix: remove atom leaks

kjellwinblad:
Both emqx_resource_managers and emqx_resource_workers leaked atoms as they created unique atoms to use as registered names. This is fixed by removing the need to register the names.

Fixes: https://emqx.atlassian.net/browse/EMQX-8583


apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -429,6 +428,14 @@ read_cache(ResId) ->
         [] -> not_found
     end.
 
+read_manager_pid_from_cache(ResId) ->
+    case read_cache(ResId) of
+        not_found ->
+            erlang:error(badarg);
```
zmstone, line 434: nit: if we do `exit(noproc)` here, we can git rid of the `error:badarg` catch clause

zmstone: actually, we should add a test to cover this error instead.

thalesmg:
I think this'll conflict a lot with this refactoring: https://github.com/emqx/emqx/pull/9699 :see_no_evil:

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -649,10 +656,8 @@ do_wait_for_ready(ResId, Retry) ->
 
 safe_call(ResId, Message, Timeout) ->
     try
-        Module = atom_to_binary(?MODULE),
-        MgrId = get_owner(ResId),
-        ProcName = binary_to_existing_atom(<<Module/binary, "_", MgrId/binary>>, utf8),
-        gen_statem:call(ProcName, Message, {clean_timeout, Timeout})
+        ManagerPid = read_manager_pid_from_cache(ResId),
+        gen_statem:call(ManagerPid, Message, {clean_timeout, Timeout})
```
thalesmg, line 660: Since we now rely on the table, maybe we could just match on the lookup result rather than raising errors.  Ex:

```suggestion
        case read_cache(ResId) of
            not_found ->
                {error, not_found};
            ManagerPid ->
                gen_statem:call(ManagerPid, Message, {clean_timeout, Timeout})
        end
```

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -757,82 +756,85 @@ get_first_n_from_queue(Q, N) ->
...
+        emqx_resource_worker_inflight_tab,
+        [ordered_set, public, {write_concurrency, true}]
+    ),
+    inflight_append(TableId, ?MAX_SIZE_REF, {max_size, InfltWinSZ}, Id, Index),
     %% we use this counter because we might deal with batches as
     %% elements.
-    inflight_append(Name, ?SIZE_REF, 0, Id, Index),
-    ok.
+    inflight_append(TableId, ?SIZE_REF, 0, Id, Index),
+    {ok, TableId}.
```
thalesmg, line 768: Nit:

```suggestion
    TableId.
```

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -165,12 +164,12 @@ init({Id, Index, Opts}) ->
     emqx_resource_metrics:queuing_set(Id, Index, queue_count(Queue)),
     emqx_resource_metrics:inflight_set(Id, Index, 0),
     InfltWinSZ = maps:get(async_inflight_window, Opts, ?DEFAULT_INFLIGHT),
-    ok = inflight_new(Name, InfltWinSZ, Id, Index),
+    {ok, InflightTID} = inflight_new(InfltWinSZ, Id, Index),
```
thalesmg, line 167: Nit:

```suggestion
    InflightTID = inflight_new(InfltWinSZ, Id, Index),
```

thalesmg (approved):
Just minor suggestions.

Since this will conflict a lot with #9699 , maybe those minor changes could be done there...

kjellwinblad:
@thalesmg:
> Just minor suggestions.

> Since this will conflict a lot with https://github.com/emqx/emqx/pull/9699 , maybe those minor changes could be done there...

Thanks for the suggestions. I applied them anyway as I had to update the changelog and I found and fixed a flaky test case as well. I will try to merge this soon so you can rebase https://github.com/emqx/emqx/pull/9699 on top of it.

thalesmg (approved):

PR #9734, opened 2023-01-12
refactor: remove default value of timestamp field of influxdb

terry-xiaoyu:
Remove the default value of timestamp field of the influxdb data-bridge.

Before this change, the `timestamp` defaults to `${timestamp}`, but this requires the Rule has a `timestamp` field in its output.

I removed the default value so that if the `timestamp` is not provided, it will use the current time according to the `precision` field.

### DIFF
--- a/lib-ee/emqx_ee_connector/src/emqx_ee_connector_influxdb.erl
+++ b/lib-ee/emqx_ee_connector/src/emqx_ee_connector_influxdb.erl
@@ -142,7 +142,11 @@ fields(common) ->
     [
         {server, server()},
         {precision,
-            mk(enum([ns, us, ms, s, m, h]), #{
+            %% The influxdb only supports these 4 precision:
+            %% See "https://github.com/influxdata/influxdb/blob/
+            %% 6b607288439a991261307518913eb6d4e280e0a7/models/points.go#L487" for
+            %% more information.
+            mk(enum([ns, us, ms, s]), #{
                 required => false, default => ms, desc => ?DESC("precision")
             })}
     ];
@@ -210,9 +214,7 @@ start_client(InstId, Config) ->
 do_start_client(
     InstId,
     ClientConfig,
-    Config = #{
-        write_syntax := Lines
-    }
+    Config = #{write_syntax := Lines}
 ) ->
     case influxdb:start_client(ClientConfig) of
         {ok, Client} ->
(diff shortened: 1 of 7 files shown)


lib-ee/emqx_ee_connector/src/emqx_ee_connector_influxdb.erl
```diff
@@ -348,30 +352,33 @@ do_async_query(InstId, Client, Points, ReplyFunAndArgs) ->
...
+    Item = #{
+        measurement => emqx_plugin_libs_rule:preproc_tmpl(maps:get(measurement, Item0)),
+        timestamp => preproc_tmpl_timestamp(Ts, Precision),
+        tags => to_kv_config(maps:get(tags, Item0)),
+        fields => to_kv_config(maps:get(fields, Item0))
     },
-    to_config(Rest, [Res | Acc]).
+    to_config(Rest, [Item | Acc], Precision).
+
+preproc_tmpl_timestamp(undefined, <<"ns">>) ->
```
thalesmg, line 370: Not related to this PR in particular, but I noticed we do `atom_to_binary` on the precision.  :thinking: 

Do you know if there's a reason for that?

lib-ee/emqx_ee_connector/src/emqx_ee_connector_influxdb.erl
```diff
@@ -414,7 +421,7 @@ parse_batch_data(InstId, BatchData, SyntaxLines) ->
         fields := [{binary(), binary()}],
         measurement := binary(),
         tags := [{binary(), binary()}],
-        timestamp := binary()
+        timestamp := list() | integer()
```
thalesmg, line 424: Nit: if this `list()` means the processed template, maybe a more precise type would be:

```suggestion
        timestamp := emqx_plugin_libs_rule:tmpl_token() | integer()
```

thalesmg (approved):

lib-ee/emqx_ee_connector/src/emqx_ee_connector_influxdb.erl
terry-xiaoyu: I don't know either,  I guess it does `atom_to_binary()` for the influxdb to concatenate them into a query URL.

thalesmg: Aha, I see!  It was just my own curiosity.  :smiley_cat: 

I think that the client should perform that internally, but I digress :joy_cat:

thalesmg (dismissed):

zhongwencool (approved):

PR #9737, opened 2023-01-12
fix(limiter): message_in limiter only works on publish packets

HJianBo:
https://emqx.atlassian.net/browse/EMQX-8695


zmstone (approved):

qzhuyan:
I think we need some feature switch to control if we limit on publish or not  otherwise attacker could attack with flooding PINGREQ.

zmstone:
I agree with William

PR #9740, opened 2023-01-12
feat: enable periodic iterator refresh

keynslug:
This might be helpful during replays taking multiple tens of seconds so that underlying iterators won't hold onto in-memory / on-disk data structures for too long, preventing rocksdb from recycling them.

---

[EMQX-8622](https://emqx.atlassian.net/browse/EMQX-8622)


apps/emqx_replay/src/emqx_replay_message_storage.erl
```diff
@@ -274,41 +287,51 @@ store(DB = #db{handle = DBHandle, cf = CFHandle}, MessageID, PublishedAt, Topic,
     rocksdb:put(DBHandle, CFHandle, Key, Value, DB#db.write_options).
 
 -spec make_iterator(db(), emqx_topic:words(), time() | earliest) ->
+    {ok, iterator()} | {error, _TODO}.
+make_iterator(DB, TopicFilter, StartTime) ->
+    % TODO wire it up somehow to the upper level
```
ieQu1, line 292: Read directly from `emqx_replay_conf.erl` ?

ieQu1 (approved):

PR #9742, opened 2023-01-12
feat(mongo): expose buffer worker opts to the bridge frontend (5.0)

thalesmg:


lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_mongodb.erl
```diff
@@ -69,7 +69,26 @@ fields("get_sharded") ->
 fields("get_single") ->
     emqx_bridge_schema:metrics_status_fields() ++
         fields(mongodb_single) ++
-        type_and_name_fields(mongodb_single).
+        type_and_name_fields(mongodb_single);
+fields("creation_opts") ->
+    lists:map(
+        fun
+            ({query_mode, FieldSchema0}) ->
+                FieldSchema = FieldSchema0#{default => sync},
```
zmstone, line 77: maybe just `hoconsc:mk/2` a new one instead of manipulating an opaque term.

thalesmg: Ok, I'll do that.

But it feels dangerous that it's quite easy for the "upstream" type to change and then this new version will get out of date.  :speak_no_evil:

k32 (approved):

ieQu1 (approved):

ieQu1 (approved):

PR #9747, opened 2023-01-13
docs: improve kafka key and value field description

zmstone:
https://emqx.atlassian.net/browse/EMQX-8464

### DIFF
--- a/lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
+++ b/lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
@@ -250,9 +250,9 @@ emqx_ee_bridge_kafka {
     kafka_message_key {
         desc {
             en: "Template to render Kafka message key. "
-                "If the desired variable for this template is not found in the input data "
+                "If the template is rendered into an empty string, "
                 "<code>NULL</code> is used."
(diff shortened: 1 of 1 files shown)


lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -250,9 +250,9 @@ emqx_ee_bridge_kafka {
     kafka_message_key {
         desc {
             en: "Template to render Kafka message key. "
-                "If the desired variable for this template is not found in the input data "
+                "If the template is rendered into an empty string, "
                 "<code>NULL</code> is used."
```
thalesmg, line 254: Q: is that done somehow by `kpro` when it encounters an empty binary?  I couldn't find mentions of `null` in `wolff_producer` :thinking:

thalesmg (approved):

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
zmstone: it's done in `kafka_protocol`. empty strings are encoded as `NULL`.

zmstone: I actually need to check if it's `"undefined"` or `""` for unknown variables.

PR #8895, opened 2022-09-05
fix: cluster/local-override.conf does not take effect at vm.args/app.config

zhongwencool:
Config changes such as log levels etc can be applied from dashboard in realtime.
However due to the sys.config (or app.config) generation did not take cluster-override (nor local-override) into account.
such config change are not persisted.

This PR tries to include both cluster and local overrides into config generation stage.

- [x] Verify space is allowed in data_dir config (and env override)

Fix `emqx.cmd` for Windows in another PR.

PR #9750, opened 2023-01-13
fix: save app's env when bootstrapping and reloading

zhongwencool:
If we configure `logger` and `prometheus`  in `cluster-override.conf`
```
log {
  file_handlers {
    default {
      burst_limit {
        enable = true
        max_count = 10000
        window_time = "1s"
      ...
    }
prometheus {
  enable = true
  interval = "151s"
  job_name = "${name}/instance/${name}~${host}"
  mnesia_collector = "enabled"
  push_gateway_server = "http://127.0.0.1:9091"
  vm_dist_collector = "enabled"
  vm_memory_collector = "enabled"
  vm_msacc_collector = "enabled"
  vm_statistics_collector = "enabled"
  vm_system_info_collector = "enabled"
}
```
These configurations will have no effect. This is because `log` is an environment variable for `kernal` applications.
`*_collector` are the environment variable for the `prometheus` application.

we need to save those two applications' env after init_load.

close https://github.com/emqx/emqx/pull/8895

### DIFF
--- a/apps/emqx/src/config/emqx_config_logger.erl
+++ b/apps/emqx/src/config/emqx_config_logger.erl
@@ -37,9 +37,14 @@ remove_handler() ->
 %% so we need to refresh the logger config after this node starts.
 %% It will not affect the logger config when cluster-override.conf is unchanged.
 refresh_config() ->
-    Log = emqx:get_raw_config(?LOG),
-    {ok, _} = emqx:update_config(?LOG, Log),
-    ok.
+    case emqx:get_raw_config(?LOG, undefined) of
+        %% no logger config when CT is running.
+        undefined ->
+            ok;
+        Log ->
+            {ok, _} = emqx:update_config(?LOG, Log),
+            ok
+    end.
 
 post_config_update(?LOG, _Req, _NewConf, _OldConf, AppEnvs) ->
     Kernel = proplists:get_value(kernel, AppEnvs),
--- a/apps/emqx/src/config/emqx_config_logger.erl
+++ b/apps/emqx/src/config/emqx_config_logger.erl
@@ -18,7 +18,7 @@
 -behaviour(emqx_config_handler).
 
 %% API
--export([add_handler/0, remove_handler/0]).
+-export([add_handler/0, remove_handler/0, refresh_config/0]).
 -export([post_config_update/5]).
 
 -define(LOG, [log]).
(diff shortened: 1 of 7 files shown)


apps/emqx/src/config/emqx_config_logger.erl
```diff
@@ -31,6 +31,21 @@ remove_handler() ->
...
+%% but we only copy cluster-override.conf from other node after this node starts,
+%% so we need to refresh the logger config after this node starts.
+%% It will not affect the logger config when cluster-override.conf is unchanged.
+refresh_config() ->
+    case emqx:get_raw_config(?LOG, undefined) of
+        %% no logger config when CT is running.
+        undefined ->
+            ok;
+        Log ->
+            {ok, _} = emqx:update_config(?LOG, Log),
```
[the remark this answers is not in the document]
zmstone: the answer is no

apps/emqx/src/emqx_config.erl
```diff
@@ -534,23 +535,19 @@ get_root_names() ->
...
     %% We first try to save to override.conf, because saving to files is more error prone
     %% than saving into memory.
     ok = save_to_override_conf(OverrideConf, Opts),
-    %% We may need also support hot config update for the apps that use application envs.
-    %% If that is the case uncomment the following line to update the configs to app env
-    %save_to_app_env(_AppEnvs),
+    save_to_app_env(AppEnvs),
     save_to_config_map(Conf, RawConf).
 
+-define(IGNORE_APPS, [kernel]).
```
zmstone, line 545: add a comment here ? e.g. why ignore for kernel.

apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -463,7 +463,7 @@ fields("node") ->
             )},
         {"config_files",
             sc(
-                list(string()),
```
zhongwencool, line 466: when we config `node.config_files = ["test"]`
```
node {
  name = "emqx@127.0.0.1"
  cookie = "emqxsecretcookie"
  data_dir = "data"
  config_files = ["test"]
}
```
the config schema check will failed by:
```
exception throw: {emqx_conf_schema,[#{kind => validation_error,
                                      path => "node.config_files",
                                      reason =>
                                          #{expected_type => <<"[string()]">>},
                                      value => [<<"test">>]}]}
  in function  hocon_tconf:assert_no_error/2 (hocon_tconf.erl, line 1142)
  in call from hocon_tconf:map/4 (hocon_tconf.erl, line 295)
  in call from hocon_cli:get_values/3 (hocon_cli.erl, line 194)
  in call from hocon_cli:multi_get/2 (hocon_cli.erl, line 182)
  in call from escript:run/2 (escript.erl, line 750)
  in call from escript:start/1 (escript.erl, line 277)
  in call from init:start_em/1
  in call from init:do_boot/3
```
So we change the type to `hocon:array(string()).`

zmstone (approved):

PR #9761, opened 2023-01-14
feat: introduce 'this' concept for placeholder, and use it in Kafka bridge

zmstone:
Prior to this change there is no way to express 'the root' in value templates.
Now a hidden `this` is introduced, and `.` is allowed in front of each var (path)

`${.client}` means the same thing as `${client}` (for compatibility reasons).
most importantly `${.}` means `this`.

With `${.}` supported, there is no longer the ambiguity of configs like:
"if this template is left empty, then it means send the whole thing",
for which, the template parsing and rendering both need special function clauses.
e.g. https://github.com/emqx/emqx/pull/9705

Edit:
Prior to this change, "${.}" is a 'valid' placeholder, but it can never work as it's parsed into empty path segments like below.
so there is no compatibility issue introduced in this change.
```
> emqx_placeholder:preproc_tmpl("${.}").
[{var,{path,[{key,<<>>},{key,<<>>}]}}]
```

### DIFF
--- a/apps/emqx_plugin_libs/src/emqx_placeholder.erl
+++ b/apps/emqx_plugin_libs/src/emqx_placeholder.erl
@@ -298,7 +298,7 @@ replace_with(Tmpl, RE, '$n') ->
         ),
     Res.
 parse_nested(<<".", R/binary>>) ->
-    %% ignroe the root .
+    %% ignore the root .
     parse_nested(R);
 parse_nested(Attr) ->
     case string:split(Attr, <<".">>, all) of
--- a/apps/emqx_plugin_libs/src/emqx_placeholder.erl
+++ b/apps/emqx_plugin_libs/src/emqx_placeholder.erl
@@ -39,6 +39,8 @@
     sql_data/1
 ]).
 
+-include_lib("emqx/include/emqx_placeholder.hrl").
+
 -define(EX_PLACE_HOLDER, "(\\$\\{[a-zA-Z0-9\\._]+\\})").
 
 -define(EX_PLACE_HOLDER_DOUBLE_QUOTE, "(\\$\\{[a-zA-Z0-9\\._]+\\}|\"\\$\\{[a-zA-Z0-9\\._]+\\}\")").
@@ -233,9 +235,6 @@ proc_param_str(Tokens, Data, Quote) ->
         proc_tmpl(Tokens, Data, #{return => rawlist, var_trans => Quote})
     ).
 
-%% backward compatibility for hot upgrading from =< e4.2.1
-get_phld_var(Fun, Data) when is_function(Fun) ->
-    Fun(Data);
 get_phld_var(Phld, Data) ->
     emqx_rule_maps:nested_get(Phld, Data).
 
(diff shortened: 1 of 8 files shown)


apps/emqx_plugin_libs/src/emqx_placeholder.erl
```diff
@@ -298,9 +297,12 @@ replace_with(Tmpl, RE, '$n') ->
             Parts
         ),
     Res.
-
+parse_nested(<<".", R/binary>>) ->
+    %% ignroe the root .
```
terry-xiaoyu, line 301: ignore

apps/emqx_plugin_libs/src/emqx_placeholder.erl
```diff
@@ -298,9 +297,12 @@ replace_with(Tmpl, RE, '$n') ->
...
+    parse_nested(R);
 parse_nested(Attr) ->
     case string:split(Attr, <<".">>, all) of
+        [<<>>] -> {var, ?PH_VAR_THIS};
```
terry-xiaoyu, line 305: For placeholders that contain only `<<".">>` it is `{var,?PH_VAR_THIS}`, but for empty string placeholders (`<<>>`), it should be `{var, <<>>}`.

i.e. if empty string is provided, we should send the empty string as it is to the data-bridges.

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -250,9 +250,10 @@ emqx_ee_bridge_kafka {
     kafka_message_key {
         desc {
             en: "Template to render Kafka message key. "
-                "If the desired variable for this template is not found in the input data "
-                "<code>NULL</code> is used."
-            zh: "生成 Kafka 消息 Key 的模版。当所需要的输入没有时，会使用 <code>NULL</code>。"
+                "If the template is rendered into a NULL value (i.e. there is no such data field in Rule Engine context) "
+                "then Kafka's <code>NULL</code> (but not empty string) is used."
+            zh: "生成 Kafka 消息 Key 的模版。如果模版生成后为空值，"
+                "则会使用 Kafka 的 <code>NULL</code ，而非空字符串。"
```
terry-xiaoyu, line 256: `<code>NULL</code>`

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_kafka.conf
```diff
@@ -262,9 +263,11 @@ emqx_ee_bridge_kafka {
...
         desc {
             en: "Template to render Kafka message value. "
-                "If the desired variable for this template is not found in the input data "
-                "<code>NULL</code> is used."
-            zh: "生成 Kafka 消息 Value 的模版。当所需要的输入没有时，会使用 <code>NULL</code>。"
+                "If the template is rendered "
+                "into a NULL value (i.e. there is no such data field in Rule Engine context) "
+                "then Kafka's <code>NULL</code> (but not empty string) is used."
+            zh: "生成 Kafka 消息 Value 的模版。如果模版生成后为空值，"
+                "则会使用 Kafka 的 <code>NULL</code ，而非空字符串。"
```
terry-xiaoyu, line 270: `<code>NULL</code>`

apps/emqx_plugin_libs/src/emqx_placeholder.erl
zmstone: we should probably not allow it ?

zmstone: Actually, "${}" is not a valid placeholder today. as it's parsed into :
```
> emqx_placeholder:preproc_tmpl("${}").
[{str,<<"${}">>}]
```
if we want to do something about it, it should be a separate PR.

ieQu1 (approved):

terry-xiaoyu (approved):

zmstone:
the one failing test case has been fixed in master branch.

PR #9765, opened 2023-01-15
fix(schema): add password converter to ensure its binary() type

### LINKED ISSUE #9400: If the value of a binary config field can also be treated as an integer, EMQX will throw a type error
### What happened?

Set a value by the env variable for a binary config field, if the value can be converted to an integer(e.g. `123456`), EMQX will throw a type error

If the type of a field is binary/string, maybe we can use what we got instead of converting it

export EMQX_SYSMON__TOP__DB_PASSWORD=123456
make run

_No response_

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
# paste output here
In fact, it does not start at all
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux bookworm/sid"
NAME="Debian GNU/Linux"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
$ uname -a
Linux DESKTOP-LH9I5M8 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 GNU/Linux

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Log files

<details>
2022-11-21T22:48:51.858674+08:00 [error] failed_to_check_schema: emqx_conf_schema
2022-11-21T22:48:51.860889+08:00 [error] #{kind => validation_error,path => "sysmon.top.db_password",reason => #{expected_type => <<"binary()">>},value => 123456}
ERROR: call_hocon_failed: -v -t 2022.11.21.22.48.51 -s emqx_conf_schema -c /home/firest/labori/emqx/_build/emqx/rel/emqx/etc/emqx.conf -d /home/firest/labori/emqx/_build/emqx/rel/emqx/data/configs generate
make: *** [Makefile:209: quickrun] Error 1
</details>

zmstone:
fixes #9400
Now `EMQX_DASHBOARD__DEFAULT_PASSWORD=1234` is allowed.

With `EMQX_DASHBOARD__DEFAULT_PASSWORD='{a:1}'`
You get 

[error] #{kind => validation_error,path => "dashboard.default_password",reason => "must_quote"}
```

### DIFF
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -114,6 +114,7 @@
 -export([namespace/0, roots/0, roots/1, fields/1, desc/1, tags/0]).
 -export([conf_get/2, conf_get/3, keys/2, filter/1]).
 -export([server_ssl_opts_schema/2, client_ssl_opts_schema/1, ciphers_schema/1]).
+-export([password_converter/2]).
 -export([authz_fields/0]).
 -export([sc/2, map/2]).
 
@@ -1510,7 +1511,9 @@ fields("sysmon_top") ->
                 #{
                     mapping => "system_monitor.db_password",
                     default => "system_monitor_password",
-                    desc => ?DESC(sysmon_top_db_password)
+                    desc => ?DESC(sysmon_top_db_password),
+                    converter => fun password_converter/2,
+                    sensitive => true
                 }
             )},
         {"db_name",
@@ -1900,7 +1903,8 @@ common_ssl_opts_schema(Defaults) ->
                     required => false,
                     example => <<"">>,
                     format => <<"password">>,
-                    desc => ?DESC(common_ssl_opts_schema_password)
+                    desc => ?DESC(common_ssl_opts_schema_password),
+                    converter => fun password_converter/2
                 }
             )},
         {"versions",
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

File: apps/emqx/src/emqx_schema.erl [hunk 1]
```diff
@@ -1900,7 +1903,8 @@ common_ssl_opts_schema(Defaults) ->
                     required => false,
                     example => <<"">>,
                     format => <<"password">>,
-                    desc => ?DESC(common_ssl_opts_schema_password)
+                    desc => ?DESC(common_ssl_opts_schema_password),
+                    converter => fun password_converter/2
```
lafirest, line 1907: if the db_name also needs to use this function, maybe the name `try_converter_to_binary` will be better?

zmstone: if it's sensitive information, it'd be better to use the function name with a sensitive keyword.

zmstone: I mean, for db_name, is it possible to use integer?

zmstone: added another export: `bin_str_cvonverter/2`

lafirest (dismissed):

lafirest (approved):

thalesmg (approved):

qzhuyan (approved):

PR #9774, opened 2023-01-16
feat: add dashboard password validation

zhongwencool:
Port https://github.com/emqx/emqx/pull/9696 from v4.4 to v5.0.
A password format check has been added to the dashboard UI(`v1.1.5`). Now it needs to be added to the API as well.
Now dashboard password must :
- At least two different kinds of characters from groups of letters, numbers, and special characters.
- The range of password length is 8~64.


apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -92,21 +92,79 @@ add_default_user() ->
...
+            ok
+    end.
+
+-define(LOW_LETTER_CHARS, "abcdefghijklmnopqrstuvwxyz").
+-define(UPPER_LETTER_CHARS, "ABCDEFGHIJKLMNOPQRSTUVWXYZ").
+-define(LETTER, ?LOW_LETTER_CHARS ++ ?UPPER_LETTER_CHARS).
+-define(NUMBER, "0123456789").
+-define(SPECIAL_CHARS, "!@#$%^&*()_+-=[]{}\"|;':,./<>?`~ ").
+-define(INVALID_PASSWORD_MSG, <<
+    "Bad username."
```
zmstone, line 125:
```suggestion
    "Bad password."
```

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -92,21 +92,79 @@ add_default_user() ->
...
+    "At least two different kind of characters from groups of letters, numbers, and special characters."
```
zmstone, line 126:
```suggestion
    "At least two different kind of characters from groups of letters, numbers, and special characters. "
```

zmstone (approved):

zmstone (approved):

PR #9777, opened 2023-01-16
refactor: config_files is a readonly config

zmstone:
the `node.config_files` config is only used to map to app env,
but a user can never change it because EMQX has to boot from `etc/emqx.conf`


zhongwencool (approved):

zhongwencool (approved):

ieQu1 (dismissed):

changes/v5.0.15/fix-9750.en.md
```diff
@@ -0,0 +1,5 @@
+Reload overriding configs after boot.
+Prior to this change, two configs were allow to change from dashboard, but will not take effect after reboot:
```
thalesmg, line 2:
```suggestion
Prior to this change, two configs were allowed to be changed from dashboard, but would not take effect after reboot:
```

thalesmg (approved):
Q: this seems to be a doc and test only change.  Do we need changelogs for this?

zmstone:
> Q: this seems to be a doc and test only change. Do we need changelogs for this?

the change log is not for this pr

thalesmg (approved):

zmstone:
Sorry, the most important commit was lost during rebase...
Now just pushed after rebased again.

lafirest (approved):

zhongwencool (approved):

PR #9778, opened 2023-01-16
fix: atom leak when doing Kafka bridge dry-run

kjellwinblad:
A new atom was created every time one did a dry run of a Kafka bridge (that is, clicking the Test button in the settings dialog for the bridge).

After this fix, we will only create a new atom when a bridge with a new name is created. This should be acceptable as bridges with new names are created relatively rarely and it seems to be useful to have an unique atom for every Wolff producer.

Fixes: https://emqx.atlassian.net/browse/EMQX-8739

### DIFF
--- a/lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
+++ b/lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
@@ -75,7 +75,16 @@ on_start(InstId, Config) ->
             }),
             throw(failed_to_start_kafka_client)
     end,
-    WolffProducerConfig = producers_config(BridgeName, ClientId, ProducerConfig),
+    %% Check if this is a dry run
+    TestIdStart = string:find(InstId, ?TEST_ID_PREFIX),
+    IsDryRun =
+        case TestIdStart of
+            nomatch ->
+                false;
+            _ ->
+                string:equal(TestIdStart, InstId)
+        end,
+    WolffProducerConfig = producers_config(BridgeName, ClientId, ProducerConfig, IsDryRun),
     case wolff:ensure_supervised_producers(ClientId, KafkaTopic, WolffProducerConfig) of
         {ok, Producers} ->
             {ok, #{
@@ -241,7 +250,7 @@ ssl(#{enable := true} = SSL) ->
 ssl(_) ->
     [].
 
-producers_config(BridgeName, ClientId, Input) ->
+producers_config(BridgeName, ClientId, Input, IsDryRun) ->
     #{
         max_batch_bytes := MaxBatchBytes,
         compression := Compression,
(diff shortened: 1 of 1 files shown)


lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
```diff
@@ -293,12 +302,20 @@ make_client_id(BridgeName) ->
 
 %% Producer name must be an atom which will be used as a ETS table name for
 %% partition worker lookup.
-make_producer_name(BridgeName) when is_atom(BridgeName) ->
-    make_producer_name(atom_to_list(BridgeName));
-make_producer_name(BridgeName) ->
-    %% Woff needs atom for ets table name registration
-    %% The assumption here is bridge is not often re-created
-    binary_to_atom(iolist_to_binary(["kafka_producer_", BridgeName])).
```
thalesmg, line 301: I think this one is fixed even for dry-runs.  I made a couple quick tests, and it seems to be the user-defined bridge name (without any random suffix).

kjellwinblad: So was this fix after the base of my PR? I tested this and the bridge name was random...

kjellwinblad: (I might have rebased just before pushing so not sure if the base was what I tested with...)

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
```diff
@@ -293,12 +302,20 @@ make_client_id(BridgeName) ->
...
+    make_producer_name(atom_to_list(BridgeName), IsDryRun);
+make_producer_name(BridgeName, IsDryRun) ->
+    %% Woff needs an atom for ets table name registration. The assumption here is
+    %% that bridges with new names are not often created.
+    case IsDryRun of
+        true ->
+            %% It is a dry run and we don't want to leak too many atoms
+            %% so we use the default producer name instead of creating
+            %% an unique name.
+            wolff_producers;
```
zmstone, line 315:
```suggestion
            probing_wolff_producers;
```

kjellwinblad: Fixed in updated commit

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
kjellwinblad: I rebased on `release-50` and printed the BridgeName in this function, configured a bridge in the UI and pressed the test button three times and got: 
```
<<"bridges-create-dry-run-ef6ab73d">>
<<"bridges-create-dry-run-959ffc4e">>
<<"bridges-create-dry-run-0b915820">>
```
So I think this PR is still relevant.

thalesmg (approved):

PR #9792, opened 2023-01-17
chore(bridges): add more helpful descriptions

savonarola:
Add more helpful descriptions for `command_template` option of Redis bridge.


zmstone (commented):
there is no need for the change logs for doc change.

zmstone (approved):

PR #9816, opened 2023-01-19
0119 fix kafka rule counters

zmstone:


lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
```diff
@@ -143,7 +143,7 @@ on_query(_InstId, {send_message, Message}, #{message_template := Template, produ
     %% If the producer process is down when sending, this function would
     %% raise an error exception which is to be caught by the caller of this callback
     {_Partition, _Pid} = wolff:send(Producers, [KafkaMessage], {fun ?MODULE:on_kafka_ack/3, [#{}]}),
-    {async_return, ok}.
```
thalesmg, line 146: This should be kept.  Otherwise it'll always increase the success counter.  Wolff already handles that counter (and the failure).

zmstone: then it becomes a failure counter.
after this fix, it double counts success. 
fixed it in my force push.

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -123,11 +124,26 @@ simple_sync_query(Id, Request) ->
     QueryOpts = #{simple_query => true},
     emqx_resource_metrics:matched_inc(Id),
     Ref = make_message_ref(),
-    Result = call_query(sync, Id, Index, Ref, ?QUERY(self(), Request, false), QueryOpts),
+    %% no need to reply anything to simple callers
+    From = undefined,
+    Result = call_query(sync, Id, Index, Ref, ?QUERY(From, Request, false), QueryOpts),
     HasBeenSent = false,
```
thalesmg, line 130:
```suggestion
    HasBeenSent = false,
    Result = call_query(sync, Id, Index, Ref, ?QUERY(From, Request, HasBeenSent), QueryOpts),
```

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -123,11 +124,26 @@ simple_sync_query(Id, Request) ->
...
+-spec simple_async_query(id(), request()) -> term().
+simple_async_query(Id, Request) ->
+    Index = undefined,
+    QueryOpts = #{simple_query => true},
+    emqx_resource_metrics:matched_inc(Id),
+    Ref = make_message_ref(),
+    %% no need to reply anything to simple callers
+    From = undefined,
+    Result = call_query(async, Id, Index, Ref, ?QUERY(From, Request, false), QueryOpts),
+    HasBeenSent = false,
```
thalesmg, line 144:
```suggestion
    HasBeenSent = false,
    Result = call_query(async, Id, Index, Ref, ?QUERY(From, Request, HasBeenSent), QueryOpts),
```

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -680,9 +696,9 @@ reply_caller_defer_metrics(Id, ?REPLY(From, _, HasBeenSent, Result), QueryOpts)
     {ShouldAck, PostFn}.
 
 handle_query_result(Id, Result, HasBeenSent) ->
-    {ShouldBlock, PostFn} = handle_query_result_pure(Id, Result, HasBeenSent),
+    {_ShouldBlock, PostFn} = handle_query_result_pure(Id, Result, HasBeenSent),
     PostFn(),
-    ShouldBlock.
+    ok.
```
thalesmg, line 701: I think it can be confusing to ignore the result above?

apps/emqx_resource/src/emqx_resource_worker.erl
```diff
@@ -982,7 +1004,7 @@ inflight_new(InfltWinSZ, Id, Index) ->
     inflight_append(TableId, {?SIZE_REF, 0}, Id, Index),
     inflight_append(TableId, {?INITIAL_TIME_REF, erlang:system_time()}, Id, Index),
     inflight_append(
-        TableId, {?INITIAL_MONOTONIC_TIME_REF, erlang:monotonic_time(nanosecond)}, Id, Index
+        TableId, {?INITIAL_MONOTONIC_TIME_REF, make_message_ref()}, Id, Index
```
thalesmg, line 1007: Nit: I think it's more semantic/easy to understand if we don't couple it to `make_message_ref/0`, as the key name is pretty indicative of monotonic time, while `make_message_ref` could change to something else.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
```diff
@@ -131,19 +132,47 @@ on_stop(_InstanceID, #{client_id := ClientID, producers := Producers, resource_i
...
+    %% * Must be a single element batch because wolff books calls, but not batch sizes
+    %%   for counters and gauges.
+    Batch = [KafkaMessage],
     %% The retuned information is discarded here.
     %% If the producer process is down when sending, this function would
     %% raise an error exception which is to be caught by the caller of this callback
-    {_Partition, _Pid} = wolff:send(Producers, [KafkaMessage], {fun ?MODULE:on_kafka_ack/3, [#{}]}),
-    {async_return, ok}.
+    {_Partition, _Pid} = wolff:send(Producers, Batch, {fun ?MODULE:on_kafka_ack/3, [AsyncReplyFn]}),
+    ok.
```
thalesmg, line 175:
```suggestion
    {_Partition, Pid} = wolff:send(Producers, Batch, {fun ?MODULE:on_kafka_ack/3, [AsyncReplyFn]}),
    {ok, Pid}.
```

thalesmg (approved):

apps/emqx_resource/src/emqx_resource_worker.erl
zmstone: it's not used anywhere from this function's return.

apps/emqx_resource/src/emqx_resource_worker.erl
zmstone: if we change make_message_ref to something else, then we have a problem.

apps/emqx_resource/src/emqx_resource_worker.erl
thalesmg: But it's different from the behavior of the pure version, which could be misleading when touching this module/fn in the future.  If the result is not wanted at the call site now, maybe we can just ignore it there?

thalesmg (dismissed):

thalesmg (dismissed):

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_producer.erl
```diff
@@ -185,9 +215,14 @@ render_timestamp(Template, Message) ->
...
     end.
 
-on_kafka_ack(_Partition, _Offset, _Extra) ->
-    %% Do nothing so far.
-    %% Maybe need to bump some counters?
+%% Wolff producer never gives up retrying
+%% so there can only be 'ok' results.
+on_kafka_ack(_Partition, Offset, {ReplyFn, Args}) when is_integer(Offset) ->
+    %% the ReplyFn is emqx_resource_worker:reply_after_query/8
+    apply(ReplyFn, Args ++ [ok]);
```
thalesmg, line 222: Small nit, but just remembered that we have `emqx_resource:apply_reply_fun(ReplyFunAndArgs, Result)`.

ieQu1 (approved):

ieQu1 (approved):

PR #9818, opened 2023-01-19
Move APP var to header for emqx_management app

danfilip:
for consistency ;)

### DIFF
--- a/apps/emqx_management/include/emqx_mgmt.hrl
+++ b/apps/emqx_management/include/emqx_mgmt.hrl
@@ -14,6 +14,8 @@
 %% limitations under the License.
 %%--------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


apps/emqx_management/include/emqx_mgmt.hrl
```diff
@@ -14,6 +14,8 @@
 %% limitations under the License.
 %%--------------------------------------------------------------------
 
+-define(APP, emqx_management).
```
sstrigler, line 17: Some other modules in this app define `APP` on their own and it's different from this one. So you can't just bring it here to avoid collisions/redefinitions.

danfilip: you're right.. :1st_place_medal:

PR #9826, opened 2023-01-20
feat(ft): introduce simple filesystem storage backend + assembler

keynslug:
Right now no context provided to the frontend.

Assembling of a single file is done in 4 general steps:
1. Ask local node for known fragments of the file.
2. If local node doesn't have enough fragments, ask **all other nodes** for known fragments.
3. If it's still not enough, bail out.
4. Read file segments from coverage nodes sequentially, one-by-one, writing them down to the local file, bailing out on any failure.

Open questions.
* Process naming.
* Configuration.
* Need for retries / recovery.
* Communication with the channel process (e.g. how to say _file assembled_ or _we failed we're very sorry_).

---

[EMQX-8265](https://emqx.atlassian.net/browse/EMQX-8265)

### DIFF
--- a/apps/emqx_ft/src/emqx_ft_assembler_sup.erl
+++ b/apps/emqx_ft/src/emqx_ft_assembler_sup.erl
@@ -1,0 +1,44 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2020-2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


apps/emqx_ft/src/emqx_ft_assembler_sup.erl
```diff
@@ -0,0 +1,44 @@
...
+
+-module(emqx_ft_assembler_sup).
+
+-export([start_link/1]).
+-export([start_child/3]).
+
+-behaviour(supervisor).
+-export([init/1]).
+
+-define(REF(ID), {via, gproc, {n, l, {?MODULE, ID}}}).
```
thalesmg, line 25: For my own information: why do we use `gproc` to register the supervisor here?  It seems to be to allow using the tuple as the process alias, and something else?

keynslug: Honestly, we're not settled on what the IDs here are yet but I imagined that in the simplest case a directory name (a storage root) could be used as an `ID`, and `gproc` makes working with such identifiers a bit saner.

PR #9832, opened 2023-01-23
EMQX 8774 failure to handle timeout error in resource worker

sstrigler:
Fixes [EMQX-8775](https://emqx.atlassian.net/browse/EMQX-8774)

## PR Checklist


savonarola (approved):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -244,15 +243,17 @@ code_change(_OldVsn, State, _Extra) ->
 
 %%==============================================================================
 -define(PICK(ID, KEY, EXPR),
```
kjellwinblad, line 245: The macro is just called in two places. Maybe it is better to remove the macro and write out the code explicitly in these two places. This might increase code duplication but might also make the code easier to follow. We could also rewrite the macro to a function and pass the expression as a fun. Is the reason for using a macro here performance or anything else?

kjellwinblad (approved):
Looks good to me. I think we can merge this as it is but feel free to get rid of the macro if there is no good reason to use a macro here.

zmstone:
please send it to `release-50`

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
sstrigler: Can't be performance since the resulting syntax tree is the same.

zmstone:
If I understand it correctly, this only affects a `action_failed` (keyword) error message.

sstrigler:
> If I understand it correctly, this only affects a `action_failed` (keyword) error message.

There's more too it, it doesn't create an error log message, which could be plenty and potentially result in an overload scenario. Not sure, if it actually also changes behaviour or not, we didn't get that far with our testing and wanted to get rid of that problem first to be able to see beyond.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
kjellwinblad: Ok, I see. A function instead of a macro might still be an option but I think this is good enough.

kjellwinblad (dismissed):

changes/v5.0.16/fix-9832.zh.md
```diff
@@ -0,0 +1 @@
+[FIXME] failure to handle timeout error in resource worker.
```
kjellwinblad, line 1: You can use https://www.deepl.com/translator to make a first attempt at translating. So far I have not got any complainant for the translations that it has generated for me.

kjellwinblad (dismissed):
LGTM

zmstone (approved):

zmstone (approved):

zmstone (approved):

Issue #9829, opened 2023-01-23
Connection immediately closed only for port 1883


Shakerrry:
### What happened?

When I am trying to connect using the webclient or MQTTX, I get connected and then immediately disconnected and it is happening when using the 1883 port. The other ports are working fine, for example: ws 8083.
This is the error i get in broker logs (i replaced my ip on peername):
`2023-01-23T08:26:16.231584+00:00 [error] crasher: initial call: emqx_connection:init/4, pid: <0.4191.0>, registered_name: [], error: {badarith,[{emqx_connection,recvloop,2,[{file,"emqx_connection.erl"},{line,413}]},{proc_lib,wake_up,3,[{file,"proc_lib.erl"},{line,236}]}]}, ancestors: [<0.2191.0>,<0.2190.0>,esockd_sup,<0.1920.0>], message_queue_len: 1, messages: [{tcp,#Port<0.57>,<<16,48,0,4,77,81,84,84,4,194,0,60,0,14,109,113,116,116,120,95,99,55,101,54,98,53,102,57,0,6,115,116,101,102,97,110,0,12,84,117,114,110,105,110,97,108,116,49,52,33>>}], links: [<0.2191.0>,#Port<0.57>], dictionary: [{'$logger_metadata$',#{peername => "myip:51871"}}], trap_exit: false, status: running, heap_size: 610, stack_size: 29, reductions: 1131; neighbours:`

### What did you expect to happen?

I expect to connect to the EMQX broker on my instance. When using the cloud version I can connect on 1883 port too.

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
sysdescr  : EMQX
version   : 5.0.14
datetime  : 2023-01-23T08:43:40.084204712+00:00
uptime    : 43 minutes, 47 seconds
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
NAME="Ubuntu"
VERSION="20.04.5 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.5 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
[...]

PR #9834, opened 2023-01-23
fix(emqx_connection): crash when idle_timeout is set to infinity

zmstone:
Fixes https://github.com/emqx/emqx/issues/9829

### DIFF
--- a/apps/emqx/i18n/emqx_schema_i18n.conf
+++ b/apps/emqx/i18n/emqx_schema_i18n.conf
@@ -656,8 +656,15 @@ mqtt 下所有的配置作为全局的默认值存在，它可以被 <code>zone<
 
     mqtt_idle_timeout {
         desc {
-            en: """After the TCP connection is established, if the MQTT CONNECT packet from the client is not received within the time specified by <code>idle_timeout</code>, the connection will be disconnected."""
-            zh: """TCP 连接建立后，如果在 <code>idle_timeout</code> 指定的时间内未收到客户端的 MQTT CONNECT 报文，则连接将被断开。"""
+            en: """After the TCP connection is established, if the MQTT CONNECT packet from the client is
+not received within the time specified by <code>idle_timeout</code>, the connection will be disconnected.
+After the CONNECT packet has been accepted by EMQX, if the connection idles for this long time,
+then the Erlang process is put to hibernation to save OS resources. Note: long <code>idle_timeout</code>
+interval may impose risk at the system if large number of malicious clients only establish connections
+but do not send any data."""
+            zh: """TCP 连接建立后，如果在 <code>idle_timeout</code> 指定的时间内未收到客户端的 MQTT CONNECT 报文，则连接将被断开。
+如果连接在 CONNECT 报文被 EMQX 接受之后空闲超过该时长，那么服务这个连接的 Erlang 进程会进入休眠以节省系统资源。
+注意，该配置值如果设置过大的情况下，如果大量恶意客户端只连接，但不发任何数据，可能会导致系统资源被恶意消耗。"""
         }
         label: {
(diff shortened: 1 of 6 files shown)


apps/emqx/test/emqx_client_SUITE.erl
```diff
@@ -363,3 +390,14 @@ tls_certcn_as_clientid(TLSVsn, RequiredTLSVsn) ->
     #{clientinfo := #{clientid := CN}} = emqx_cm:get_chan_info(CN),
     confirm_tls_version(Client, RequiredTLSVsn),
     emqtt:disconnect(Client).
+
+wait(F, 1) ->
```
id, line 394: 0?

apps/emqx/test/emqx_client_SUITE.erl
```diff
@@ -43,6 +43,8 @@
     <<"TopicA/#">>
 ]).
 
+-define(WAIT(EXPR, ATTEMPTS), wait(fun() -> EXPR end, ATTEMPTS)).
```
ieQu1, line 46: Nit: https://github.com/kafka4beam/snabbkaffe/blob/master/include/test_macros.hrl#L180 ? (included in snabbkaffe.hrl)

apps/emqx/i18n/emqx_schema_i18n.conf
```diff
@@ -656,8 +656,13 @@ mqtt 下所有的配置作为全局的默认值存在，它可以被 <code>zone<
 
     mqtt_idle_timeout {
         desc {
-            en: """After the TCP connection is established, if the MQTT CONNECT packet from the client is not received within the time specified by <code>idle_timeout</code>, the connection will be disconnected."""
-            zh: """TCP 连接建立后，如果在 <code>idle_timeout</code> 指定的时间内未收到客户端的 MQTT CONNECT 报文，则连接将被断开。"""
+            en: """After the TCP connection is established, if the MQTT CONNECT packet from the client is
+not received within the time specified by <code>idle_timeout</code>, the connection will be disconnected.
+After the CONNECT packet has been accepted by EMQX, if the connection idles for this long time,
```
qzhuyan, line 661: I think set to `infinity` will be a risk opening for DDOS because no idle timer will be started for this connection process.

The attacker could just keep establishing new TCP connections without sending MQTT.CONNECT packet that will easily bump memory usage in EMQX. 

If we don't want hibernation, just disable it with another config switch?

apps/emqx/test/emqx_client_SUITE.erl
zmstone: 1 and do F() for the last time

apps/emqx/i18n/emqx_schema_i18n.conf
zmstone: We can set a maximum value, and translate infinity to max for backward compatibility. What would be the good max?

apps/emqx/test/emqx_client_SUITE.erl
```diff
@@ -85,6 +87,12 @@ init_per_suite(Config) ->
 end_per_suite(_Config) ->
     emqx_common_test_helpers:stop_apps([]).
 
+init_per_testcase(Config) ->
```
sstrigler, line 90: Isn't it `init_per_testcase(Case, Config)`? Same for `end_per_testcase/2`?

apps/emqx/i18n/emqx_schema_i18n.conf
qzhuyan: maybe just leave the info in this doc, and let the user decide it. they just need to know the risk.

qzhuyan (dismissed):

PR #9838, opened 2023-01-25
feat(redis): disable batching in redis_cluster bridges

keynslug:
Batching will inevitably fail otherwise: it's not generally possible to update different keys through the same cluster connection, one or more update will fail with `MOVED` status.


savonarola:
We still need to fix `on_batch_query` implementation https://github.com/emqx/emqx/blob/release-50/apps/emqx_connector/src/emqx_connector_redis.erl#L269 (mb simply reside to sequential `q`'s) – resources could be created not only from bridge configs.

savonarola (approved):

zmstone (approved):

PR #9839, opened 2023-01-25
fix: Authorization header leak in log entries for webhook

kjellwinblad:
The Authorization header contains sensitive information (e.g., password) so one typically don't want to print this to log files. This PR makes sure that the content of a Authorization header that users have specified for a webhook bridge is not printed to log files.

Jira issue: https://emqx.atlassian.net/browse/EMQX-8791

I don't know if release-50 is the right branch for this to be merged to so let me know if another branch is better and I will rebase it to that branch instead.

### DIFF
--- a/apps/emqx/src/emqx_misc.erl
+++ b/apps/emqx/src/emqx_misc.erl
@@ -609,7 +609,11 @@ do_redact(K, V, Checker) ->
 
 -define(REDACT_VAL, "******").
 redact_v(V) when is_binary(V) -> <<?REDACT_VAL>>;
-redact_v(_V) -> ?REDACT_VAL.
+%% The HOCON schema system may generate sensitive values with this format
+redact_v([{str, Bin}]) when is_binary(Bin) ->
+    [{str, <<?REDACT_VAL>>}];
+redact_v(_V) ->
+    ?REDACT_VAL.
 
 is_redacted(K, V) ->
     do_is_redacted(K, V, fun is_sensitive_key/1).
--- a/apps/emqx_connector/src/emqx_connector_http.erl
+++ b/apps/emqx_connector/src/emqx_connector_http.erl
@@ -592,3 +592,33 @@ is_sensitive_key(_) ->
 %% information (i.e., passwords)
 redact(Data) ->
     emqx_misc:redact(Data, fun is_sensitive_key/1).
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+
+redact_test_() ->
+    TestData1 = [
+        {<<"content-type">>, <<"application/json">>},
+        {<<"Authorization">>, <<"Basic YWxhZGRpbjpvcGVuc2VzYW1l">>}
+    ],
+
+    TestData2 = #{
+        headers =>
+            [
+                {[{str, <<"content-type">>}], [{str, <<"application/json">>}]},
+                {[{str, <<"Authorization">>}], [{str, <<"Basic YWxhZGRpbjpvcGVuc2VzYW1l">>}]}
+            ]
+    },
+    [
+        ?_assert(is_sensitive_key(<<"Authorization">>)),
+        ?_assert(is_sensitive_key(<<"AuthoriZation">>)),
+        ?_assert(is_sensitive_key('AuthoriZation')),
+        ?_assert(is_sensitive_key(<<"PrOxy-authoRizaTion">>)),
+        ?_assert(is_sensitive_key('PrOxy-authoRizaTion')),
+        ?_assertNot(is_sensitive_key(<<"Something">>)),
+        ?_assertNot(is_sensitive_key(89)),
+        ?_assertNotEqual(TestData1, redact(TestData1)),
+        ?_assertNotEqual(TestData2, redact(TestData2))
+    ].
+
+-endif.
--- a/apps/emqx_connector/src/emqx_connector_http.erl
+++ b/apps/emqx_connector/src/emqx_connector_http.erl
@@ -209,7 +209,7 @@ on_start(
     ?SLOG(info, #{
         msg => "starting_http_connector",
         connector => InstId,
-        config => emqx_misc:redact(Config)
+        config => redact(Config)
     }),
     {Transport, TransportOpts} =
         case Scheme of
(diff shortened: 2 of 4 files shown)


apps/emqx/src/emqx_misc.erl
```diff
@@ -557,16 +557,47 @@ try_to_existing_atom(Convert, Data, Encoding) ->
...
+    true;
+is_sensitive_key(<<"password">>) ->
+    true;
+is_sensitive_key(secret) ->
+    true;
+is_sensitive_key("secret") ->
+    true;
+is_sensitive_key(<<"secret">>) ->
+    true;
+is_sensitive_key([{str, Bin}]) when is_binary(Bin) ->
```
sstrigler, line 578: Where does this come from? Maybe a short comment?

apps/emqx/src/emqx_misc.erl
```diff
@@ -557,16 +557,47 @@ try_to_existing_atom(Convert, Data, Encoding) ->
...
+    is_sensitive_key(erlang:atom_to_binary(Atom));
+is_sensitive_key(Bin) when is_binary(Bin), (size(Bin) =:= 19 orelse size(Bin) =:= 13) ->
+    try
+        %% This is wrapped in a try-catch since we don't know that Bin is a
+        %% valid string so string:lowercase/1 might throw an exception.
+        %%
+        %% We want to convert this to lowercase since the http header fields
+        %% are case insensitve, which means that a user of the webhook bridge
+        %% can write this field name in many different ways.
+        LowercaseBin = iolist_to_binary(string:lowercase(Bin)),
```
sstrigler, line 590: Why is `iolist_to_binary` necessary?

apps/emqx/src/emqx_misc.erl
```diff
@@ -609,7 +640,12 @@ do_redact(K, V, Checker) ->
 
 -define(REDACT_VAL, "******").
 redact_v(V) when is_binary(V) -> <<?REDACT_VAL>>;
-redact_v(_V) -> ?REDACT_VAL.
+redact_v([{str, Bin}]) when is_binary(Bin) ->
+    [{str, <<?REDACT_VAL>>}];
+redact_v([{str, LString}]) when is_list(LString) ->
```
sstrigler, line 645: `is_sensitive_key` doesn't have this clause, am I missing sth?

apps/emqx/src/emqx_misc.erl
```diff
@@ -557,16 +557,47 @@ try_to_existing_atom(Convert, Data, Encoding) ->
...
+    true;
+is_sensitive_key("secret") ->
+    true;
+is_sensitive_key(<<"secret">>) ->
+    true;
+is_sensitive_key([{str, Bin}]) when is_binary(Bin) ->
+    is_sensitive_key(Bin);
+is_sensitive_key(Atom) when is_atom(Atom) ->
+    is_sensitive_key(erlang:atom_to_binary(Atom));
+is_sensitive_key(Bin) when is_binary(Bin), (size(Bin) =:= 19 orelse size(Bin) =:= 13) ->
```
sstrigler, line 582: Add to unit tests further down the module?

apps/emqx_connector/src/emqx_connector_http.erl
```diff
@@ -366,7 +370,11 @@ on_query_async(
     ?TRACE(
         "QUERY_ASYNC",
         "http_connector_received",
-        #{request => Request, connector => InstId, state => State}
+        #{
+            request => emqx_misc:redact(Request),
+            connector => InstId,
+            state => emqx_misc:redact(State)
```
sstrigler, line 376: I think we should rather use `emqx_misc:redact/2` here and use a custom `Checker` that adds support for those http headers rather then adding that check to the generic `is_sensitive_key/1`.
But probably the latter should also be exported from `emqx_misc`, then you could use it like
`emqx_misc:redacted(State, fun http_header_checker/1)` and 
```http_header_checker(Key) ->
   case emqx_misc:is_sensitive_key(Key) of 
     false -> 
      ...
      true -> 
           true
   end.
   ```

olcai: Yes, I did something similar [here](https://github.com/emqx/emqx/pull/9840/files#diff-9c8db1387f528a80196a9d0b371694e51d55ebb27bfa8f960abefc1ed05634c6R341).

apps/emqx/src/emqx_misc.erl
kjellwinblad: As we discussed `string:lowercase(Bin)` might always return a binary if the input is a Bin but this might change in the future if it is not documented.

apps/emqx/src/emqx_misc.erl
kjellwinblad: A short comment is a good idea

apps/emqx_connector/src/emqx_connector_http.erl
kjellwinblad: Yes, I think this makes sense unless sensitive data also have to be redacted at a higher level. For example, if the connector crashes, then maybe the state of the connector is logged in another module. I will try to investigate a little bit more how this works before I refactor to use an module internal is_sensitive_key function.

apps/emqx/src/emqx_misc.erl
kjellwinblad: Localized the redacting as you suggested so this module is not the right place anymore but maybe it is a good idea to have unit tests anyway. I will try to add something similar to what @olcai did  here: https://github.com/emqx/emqx/pull/9840/files#diff-9c8db1387f528a80196a9d0b371694e51d55ebb27bfa8f960abefc1ed05634c6R629

apps/emqx_connector/src/emqx_connector_http.erl
kjellwinblad: @olcai @sstrigler thanks I forced pushed a new commit where the redacting is localized.

apps/emqx/src/emqx_misc.erl
kjellwinblad: Added a commit with a unit test case as well. Not sure if such test cases are executed automatically in the CI runs or if one have to add it manually. Do you happen to know?

kjellwinblad:
There might be another possibility for sensitive info leakage. If the resource manager for the webhook resource crashes, OTP might log the spec for the resource manager which contains the Config and thus the Authorization header. This is probably an issue for other resources as well and should be fixed in another commit. The following issue has been created to fix this:

https://emqx.atlassian.net/browse/EMQX-8791

qzhuyan (approved):

kjellwinblad:
@qzhuyan I had forgot the change log entry. Can you approve again if you think the changelog entry is ok?

qzhuyan (approved):

PR #9845, opened 2023-01-26
0126 reply caller for buffer overflow queue items

zmstone:
fixes:
https://emqx.atlassian.net/browse/EMQX-8782

For calls discarded due to buffer overflow, an error should replied immediately instead of waiting for timeout.

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -242,7 +242,7 @@ blocked(cast, flush, Data) ->
     resume_from_blocked(Data);
 blocked(state_timeout, unblock, St) ->
     resume_from_blocked(St);
-blocked(info, ?SEND_REQ(_ReqFrom, {query, _Request, _Opts}) = Request0, Data0) ->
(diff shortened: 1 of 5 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1083,34 +1098,41 @@ queue_item_marshaller(Item) ->
...
+                {[], Q0};
+            OverflownBytes ->
+                PopOpts = #{bytes_limit => OverflownBytes, count_limit => 999999999},
                 {Q1, QAckRef, Items2} = replayq:pop(Q0, PopOpts),
                 ok = replayq:ack(Q1, QAckRef),
                 Dropped = length(Items2),
                 emqx_resource_metrics:dropped_queue_full_inc(Id),
-                ?SLOG(error, #{msg => drop_query, reason => queue_full, dropped => Dropped}),
-                Q1
+                ?SLOG(info, #{
```
zmstone, line 1118: this log level has been changed from error to info.
because there could be too much to log.

qzhuyan (approved):

PR #9849, opened 2023-01-27
0127 refactor buffer worker simplify caller reply

zmstone:
Turned the `From = {Pid, Ref}` to `{fun gen_statem:reply/2, [Ref]}`
so we can unify the code to only having to deal with callbacks when reply async callers.

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -210,7 +210,7 @@ running(cast, flush, Data) ->
     flush(Data);
 running(cast, block, St) ->
     {next_state, blocked, St};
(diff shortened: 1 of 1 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -210,7 +210,7 @@ running(cast, flush, Data) ->
     flush(Data);
 running(cast, block, St) ->
     {next_state, blocked, St};
-running(info, ?SEND_REQ(_From, _Req) = Request0, Data) ->
+running(info, ?SEND_REQ(_ReplyTo, _Req) = Request0, Data) ->
```
qzhuyan, line 213: Nit, honestly I think `From` is ideal name.

qzhuyan (approved):

PR #9853, opened 2023-01-27
0127 refactor buffer worker no need to keep request for reply callback

zmstone:
prior to this change, the Request body is kept in the async reply callback args,
the args are passed down to the async API handers (bridge driver).

The Request, which can be potentially very large,
is actually not needed (except for when testing) when the async callback is evaluated.
this PR is to replace this term with en empty list `[]`.

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -52,7 +52,7 @@
 
 -export([queue_item_marshaller/1, estimate_size/1]).
 
--export([reply_after_query/8, batch_reply_after_query/8]).
+-export([handle_async_reply/8, batch_reply_after_query/8]).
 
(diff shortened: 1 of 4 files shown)


keynslug (approved):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -73,9 +73,8 @@
 -type id() :: binary().
 -type index() :: pos_integer().
 -type expire_at() :: infinity | integer().
--type queue_query() :: ?QUERY(from(), request(), HasBeenSent :: boolean(), expire_at()).
+-type queue_query() :: ?QUERY(reply_fun(), request(), HasBeenSent :: boolean(), expire_at()).
```
thalesmg, line 76: The first argument should still be `from()`, right?  For example, when performing a sync request.

Issue #9353, opened 2022-11-11
Nodes API error sending and receiving messages in emqx 5.0.10


siniusxcode:
### What happened?

<img width="1726" alt="스크린샷 2022-11-11 오후 5 32 24" src="https://user-images.githubusercontent.com/1418182/201300622-813c382e-2767-42cf-8519-0f4f9fb45662.png">

Increasing the number of messages sent and received results in nodes API delays and errors.
I tested it as below using [emqtt-bench](https://github.com/emqx/emqtt-bench).

./emqtt_bench sub -c 2000 -i 1 -h {host} -p 1883 -k 0 -t bench/%i -q 2
./emqtt_bench pub -c 2000 -I 10 -h {host} -p 1883 -k 0 -t bench/%i -s 1024

As clients, subscriptions, and messages increase in version 5.0.10, the related APIs(clients, subscriptions, nodes) seem to suffer from poor performance.
This did not happen with my emqx 4.0.7 cluster.

### What did you expect to happen?

I hope that overall API performance will be improved and there will be no problem with the dashboard.

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
sysdescr  : EMQX
version   : 5.0.10
datetime  : 2022-11-11T17:50:14.310923503+09:00
uptime    : 1 hours, 39 minutes, 11 seconds
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Log files

<details>

</details>

zhongwencool:
Plan fix in v5.0.16

qzhuyan:
In the fix of #9856 we added a workaround with caching. 
Looks like your system is overloaded. You might need to scale up your EMQX node. 
You might hit other issue if you don't scale up the nodes.

PR #9856, opened 2023-01-29
fix: /api/nodes is timeout when emqx in high load

zhongwencool:
1. Fix https://github.com/emqx/emqx/issues/9353
When OS resources are limited (CPU utilization > 90%), it causes the node API to be slow (> 20 seconds), and the dashboard to report errors continuously.

After caching erlang version and adding a timeout to fetch sys_memory,
this API only takes 40~50ms even when CPU utilization > 95%.

2. Ensure that the precision of the alarm's usage is 2, and add more test cases.
3. Correct the type of `emqx_vm:loads().` to float, not string.

https://emqx.atlassian.net/browse/EMQX-8801


apps/emqx_management/src/emqx_mgmt_sys_memory.erl
```diff
@@ -0,0 +1,79 @@
...
+
+init([]) ->
+    {ok, #{last_time => 0}}.
+
+handle_call(get_sys_memory, _From, State = #{last_time := LastTime}) ->
+    Now = erlang:system_time(millisecond),
+    case Now - LastTime >= ?TIMEOUT of
+        true ->
+            Memory = load_ctl:get_sys_memory(),
+            persistent_term:put(?SYS_MEMORY_CACHE_KEY, Memory),
```
zmstone, line 54: Maybe don't use persisten_term for frequently updated cache value.
We can use ETS for this.

zhongwencool:
> > After caching erlang version and adding a timeout to fetch sys_memory,
> 
> How often do we fetch sys_memory? I am just afraid to return a cached value may mislead the user about what the real problem is.

Request every 2 seconds. 
In my opinion, it's better to return a cache value than to keep time_out.
but I can change to use cache value only when gen_server call timeout(2s), otherwise, use the real value.

qzhuyan:
We could have some timeout and return error or use `0` to indicate timeout but shall not return the cached value.
Reading system memory timeout indicates a serious resource issue, using cached value hides the problem.

During the performance troubleshooting the failed/timeout of memory usage reading combined with CPU usage could help us locates where the constraints are.

apps/emqx_management/src/emqx_mgmt_sys_memory.erl
zhongwencool: DONE

apps/emqx_management/src/emqx_mgmt_cache.erl
```diff
@@ -0,0 +1,104 @@
...
+get_sys_memory() ->
+    Now = now_millisecond(),
+    {CacheMem, ExpiredAt} = get_memory_from_cache(),
+    case Now > ExpiredAt of
+        true ->
+            erlang:send(?MODULE, fresh_sys_memory),
+            CacheMem;
+        %% stale cache value, try to recalculate
+        false ->
+            get_sys_memory_sync()
```
zmstone, line 38:
```suggestion
    case Now < ExpiredAt of
        true ->
            erlang:send(?MODULE, fresh_sys_memory),
            CacheMem;
        %% stale cache value, try to recalculate
        false ->
            get_sys_memory_sync()
```

apps/emqx_management/src/emqx_mgmt_cache.erl
```diff
@@ -0,0 +1,104 @@
...
+-define(REFRESH_MS, ?EXPIRED_MS - 100).
+-define(DEFAULT_BAD_MEMORY, {0, 0}).
+
+-export([start_link/0, get_sys_memory/0]).
+-export([init/1, handle_call/3, handle_cast/2, handle_info/2, terminate/2, code_change/3]).
+
+get_sys_memory() ->
+    Now = now_millisecond(),
+    {CacheMem, ExpiredAt} = get_memory_from_cache(),
+    case Now > ExpiredAt of
```
zmstone, line 32: maybe add a `is_expired` help function, or make get_memory_from_cache/1 return `{ok, CacheMem}`, or `stale`

zhongwencool: good idea

apps/emqx_management/src/emqx_mgmt_cache.erl
```diff
@@ -0,0 +1,108 @@
...
+    ok.
+
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
+
+%%%===================================================================
+%%% Internal functions
+%%%===================================================================
+
+fresh_sys_memory(State = #{fresh_at := LastFreshAt}) ->
```
zmstone, line 78: nit:
```suggestion
refresh_sys_memory(State = #{last_refresh := LastRefresh}) ->
```

qzhuyan (approved):

apps/emqx/src/emqx_os_mon.erl
```diff
@@ -236,5 +244,5 @@ do_update_mem_alarm_status(HWM0) ->
     ok.
 
 usage_msg(Usage, What) ->
-    %% devide by 1.0 to ensure float point number
+    %% divide by 1.0 to ensure float point number
```
qzhuyan, line 247: nit, should `float(Int)` help here?

apps/emqx_management/src/emqx_mgmt_cache.erl
```diff
@@ -0,0 +1,104 @@
...
+
+init([]) ->
+    ets:new(?MODULE, [set, named_table, public, {keypos, 1}]),
+    {ok, #{fresh_at => 0}}.
+
+handle_call(get_sys_memory, _From, State) ->
+    {Mem, NewState} = fresh_sys_memory(State),
+    {reply, Mem, NewState};
+handle_call(_Request, _From, State) ->
+    {reply, ok, State}.
```
qzhuyan, line 57: nit, better return something other than `ok`

PR #9870, opened 2023-01-31
feat(mqtt-bridge): avoid middleman process

keynslug:
Instead, supervise `emqtt` client process directly. New scheme should be much simpler than before.

This currently has some minor drawbacks.
* Bridge will survive intermittent network problems worse than before: the resource manager will try to restart the bridge only after a _ReconnectInterval_ number of seconds, during this window a number of messages will probably be lost.

---

[EMQX-8798](https://emqx.atlassian.net/browse/EMQX-8798)


apps/emqx_connector/src/emqx_connector_mqtt.erl
```diff
@@ -189,44 +187,49 @@ on_stop(_InstId, #{name := InstanceId}) ->
...
-) ->
+on_query_async(_InstId, {send_message, Msg}, Callback, #{name := InstanceId}) ->
     ?TRACE("QUERY", "async_send_msg_to_remote_node", #{message => Msg, connector => InstanceId}),
-    %% this is a cast, currently.
-    ok = emqx_connector_mqtt_worker:send_to_remote_async(InstanceId, Msg, {ReplyFun, Args}),
-    WorkerPid = get_worker_pid(InstanceId),
-    {ok, WorkerPid}.
+    case emqx_connector_mqtt_worker:send_to_remote_async(InstanceId, Msg, Callback) of
+        ok ->
+            % TODO this is racy
```
zmstone, line 201: `emqx_connector_mqtt_worker:send_to_remote_async/3` can return `{ok, Pid}` ?

apps/emqx_connector/src/mqtt/emqx_connector_mqtt_worker.erl
```diff
@@ -60,172 +60,252 @@
...
-        max_inflight => MaxInflightSize,
-        connection => undefined,
-        name => Name
-    }.
+trycall(Fun, Else) ->
+    try
+        Fun()
+    catch
+        error:badarg ->
+            maps:get(badarg, Else);
```
zmstone, line 286: it seems all `badarg`s are mapped to `disconnected`, maybe simply return `disconnected` when `badarg` happens.

apps/emqx_connector/src/mqtt/emqx_connector_mqtt_worker.erl
```diff
@@ -60,172 +60,252 @@
...
+        exit:{noproc, _} ->
```
zmstone, line 287: is `noproc` just a race condition of `badarg` ?

zmstone: if yes, it should return `disconnected` too ?

zmstone (approved):

zmstone:
minor comment fix can be fixed in follow-up PRs.

apps/emqx_connector/src/mqtt/emqx_connector_mqtt_worker.erl
keynslug: Not exactly, AFAICS. This is what `gen:call` will throw when there's no pid behind a name, even when the `gproc` act as a registry.

> if yes, it should return `disconnected` too ?

On a second glance the code seems a bit too complicated, so probably yes. I'll open a followup PR.

apps/emqx_connector/src/mqtt/emqx_connector_mqtt_worker.erl
```diff
@@ -245,238 +325,112 @@ pre_process_conf(Key, Conf) ->
...
-str(S) when is_list(S) ->
-    S.
+%% TODO
+% maybe_destroy_session(#{connect_opts := ConnectOpts = #{clean_start := false}} = State) ->
+%     try
+%         %% Destroy session if clean_start is not set.
+%         %% Ignore any crashes, just refresh the clean_start = true.
+%         _ = do_connect(State#{connect_opts => ConnectOpts#{clean_start => true}}),
+%         _ = disconnect(State),
+%         ok
```
keynslug, line 430: This also needs some attention, @zmstone. As discussed before I got an impression that this _should not_ be needed anymore? Does it also mean that we should drop the [`clean_start`](https://github.com/emqx/emqx/blob/78028a7fcf26a64fe10290348c853cac26b3521e/apps/emqx_connector/src/mqtt/emqx_connector_mqtt_schema.erl#L108-L115) field from the bridge config so that a user wouldn't be able to set it to `false`?

keynslug: #9889

PR #9886, opened 2023-02-02
fix: remove async mode from mongodb/redis/mysql/pgsql bridge

zhongwencool:
mongodb/mysql/pgsql/redis bridges only support sync
Fix: https://emqx.atlassian.net/browse/EMQX-8845


zmstone:
mongodb connector has no async API. maybe we should remove the async option.

zhongwencool:
image
image

zmstone (approved):

zmstone (approved):

PR #9893, opened 2023-02-02
feat(banned): filter out banned messages from mqueue

lafirest:
Fixes https://emqx.atlassian.net/browse/EMQX-8056


qzhuyan:
if the message is in the mqueue, that means the messages weren't banned while routing/forwarding? The ban of the client could be issued afterward?

If so, I am not sure this is a general solution for all. 
Is this behavior approved by the product?

lafirest:
> 

This PR is from a discussion, and I'm not sure about this modification either.
Now, when a client was banned, we will kick it out but won't clear its messages or other data and also won't check if the client was banned when delivering(routing/forwarding) a message.

qzhuyan (approved):

id:
@lafirest @qzhuyan confirmed with Ben and Stone, we can merge this.
@qzhuyan do you think it may lead to performance issues?

PR #9894, opened 2023-02-02
ci: always run static_checks

id:
[EMQX-8841](https://emqx.atlassian.net/browse/EMQX-8841)

### DIFF
--- a/.github/workflows/run_test_cases.yaml
+++ b/.github/workflows/run_test_cases.yaml
@@ -77,6 +77,7 @@ jobs:
             make ensure-rebar3
             # fetch all deps and compile
             make ${{ matrix.profile }}
+            make static_checks
             make test-compile
             cd ..
             zip -ryq source.zip source/* source/.[^.]*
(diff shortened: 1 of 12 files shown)


ieQu1 (approved):

.github/workflows/run_test_cases.yaml
```diff
@@ -77,6 +77,7 @@ jobs:
             make ensure-rebar3
             # fetch all deps and compile
             make ${{ matrix.profile }}
+            make static_checks
```
keynslug, line 80: I think having PLT cache here would be great, otherwise the CI feedback might be delayed noticeably. Also wondering why not separate workflow?

Makefile
```diff
@@ -79,7 +79,8 @@ ct: $(REBAR) merge-config
 
 .PHONY: static_checks
 static_checks:
-	@$(REBAR) as check do dialyzer, xref, ct --suite apps/emqx/test/emqx_static_checks --readable $(CT_READABLE)
+	@$(REBAR) as check do dialyzer, xref
```
keynslug, line 82: I'd probably swap them since `xref` is much cheaper.

PR #9900, opened 2023-02-02
test(ft): add some basic assembler tests

keynslug:
And ensure local happy path is working now.

---

[EMQX-8265](https://emqx.atlassian.net/browse/EMQX-8265)

### DIFF
--- a/apps/emqx_ft/src/emqx_ft_storage.erl
+++ b/apps/emqx_ft/src/emqx_ft_storage.erl
@@ -24,6 +24,14 @@
     ]
 ).
 
+-export([list_local/2]).
+-export([pread_local/4]).
+
+-export([local_transfers/0]).
+
+-type offset() :: emqx_ft:offset().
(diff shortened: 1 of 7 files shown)


apps/emqx_ft/src/emqx_ft_storage.erl
```diff
@@ -63,12 +71,53 @@ assemble(Transfer, Callback) ->
...
     Mod:assemble(storage(), Transfer, Callback).
 
+%%--------------------------------------------------------------------
+%% Local FS API
+%%--------------------------------------------------------------------
+
+-type filefrag() :: emqx_ft_storage_fs:filefrag().
+-type transferinfo() :: emqx_ft_storage_fs:transferinfo().
+
+-spec list_local(transfer(), fragment | result) ->
```
savonarola, line 81: 🤔 shouldn't these function belong to `emqx_ft_storage_fs`?

other storages (like s3) may not have a concept of locality...

PR #9902, opened 2023-02-02
0202 ci fix docker push

zmstone:
Do not trigger build and push at the event of docker-latest-* tag push.
Instead, make sure the tag is updated before tagging the release.


zmstone:
will try to get rid of the `docker-latest-*` tag all together.

PR #9922, opened 2023-02-06
fix(bufworker): do not avoid retry if inflight table is full

keynslug:
Otherwise there's no other piece of code that would retry the inflight queries in that case.

Also add the testcase that should verify that the buffer will retry all inflight queries failed with recoverable errors + flush all outstanding queries.

---

[EMQX-8876](https://emqx.atlassian.net/browse/EMQX-8876)


apps/emqx_resource/test/emqx_resource_SUITE.erl
```diff
@@ -1452,6 +1452,53 @@ t_retry_async_inflight(_Config) ->
...
+            query_mode => async,
+            async_inflight_window => AsyncInflightWindow,
+            batch_size => 1,
+            batch_time => 20,
+            worker_pool_size => 1,
+            resume_interval => ResumeInterval
+        }
+    ),
+    ?check_trace(
+        begin
```
ieQu1, line 1474:
```suggestion
    ?check_trace(
        #{timetrap => 10_000},
        begin
```

zmstone (approved):
Good catch! Thanks a lot for the fix.

PR #9932, opened 2023-02-08
feat(bridges): add TDengine

lafirest:


lib-ee/emqx_ee_connector/i18n/emqx_ee_connector_tdengine.conf
```diff
@@ -0,0 +1,22 @@
+emqx_ee_connector_tdengine {
+
+    server {
+        desc {
+          en: """
+The IPv4 or IPv6 address or the hostname to connect to.<br/>
+A host entry has the following form: `Host[:Port]`.<br/>
+The PostgreSQL default port 5432 is used if `[:Port]` is not specified.
```
thalesmg, line 8:
```suggestion
The TDengine default port 6041 is used if `[:Port]` is not specified.
```
?

lib-ee/emqx_ee_connector/i18n/emqx_ee_connector_tdengine.conf
```diff
@@ -0,0 +1,22 @@
...
+"""
+          zh: """
+将要连接的 IPv4 或 IPv6 地址，或者主机名。<br/>
+主机名具有以下形式：`Host[:Port]`。<br/>
+如果未指定 `[:Port]`，则使用 PostgreSQL 默认端口 5432。
```
thalesmg, line 13:
```suggestion
如果未指定 `[:Port]`，则使用 TDengine 默认端口 6041。
```
?

lib-ee/emqx_ee_connector/src/emqx_ee_connector_tdengine.erl
```diff
@@ -0,0 +1,235 @@
...
+fields(config) ->
+    [
+        {server, server()}
+        | add_default_username(emqx_connector_schema_lib:relational_db_fields(), [])
+    ].
+
+add_default_username([{username, OrigUsernameFn} | Tail], Head) ->
+    Head ++ [{username, add_default_fn(OrigUsernameFn, <<"root">>)} | Tail];
+add_default_username([Field | Tail], Head) ->
+    add_default_username(Tail, Head ++ [Field]).
```
thalesmg, line 50: Nit: maybe just a `lists:map`?

```suggestion
add_default_username(Fields) ->
    lists:map(
      fun({username, OrigUsernameFn}) ->
            {username, add_default_fn(OrigUsernameFn, <<"root">>)};
         (Field) ->
             Field
      end,
      Fields).
```

lib-ee/emqx_ee_connector/src/emqx_ee_connector_tdengine.erl
```diff
@@ -0,0 +1,235 @@
...
+        connector => InstanceId
+    }),
+    emqx_plugin_libs_pool:stop_pool(PoolName).
+
+on_query(InstanceId, {query, SQL}, State) ->
+    do_query(InstanceId, SQL, State);
+on_query(InstanceId, Request, State) ->
+    %% because the `emqx-tdengine` client only supports a whole SQL cmd
+    %% so the `on_query` and `on_batch_query` have the same process, that is:
+    %% we need to contact all data into one SQL cmd and then call the insert API
```
thalesmg, line 116: If I understood correctly:

```suggestion
    %% because the `emqx-tdengine` client only supports a single SQL cmd
    %% so the `on_query` and `on_batch_query` have the same process, that is:
    %% we need to sollect all data into one SQL cmd and then call the insert API
```

lafirest: Yes, you are right, I want to say a single query string without any params.

thalesmg (approved):

PR #9934, opened 2023-02-08
build: make sure run.sh with EE app is used with EE profile

kjellwinblad:
If one runs a test suite for an EE app using `scripts/ct/run.sh` without setting the EE profile, the test suite seems to just get stuck. This commit ensures that this will not happen and can hopefully save someone some time.

### DIFF
--- a/scripts/ct/run.sh
+++ b/scripts/ct/run.sh
@@ -80,6 +80,14 @@ if [ "${WHICH_APP}" = 'novalue' ]; then
     exit 1
 fi
 
+if [[ "${WHICH_APP}" == lib-ee* && (-z "${PROFILE+x}" || "${PROFILE}" != emqx-enterprise) ]]; then
+    echo 'You are trying to run an enterprise test case without the emqx-enterprise profile.'
+    echo 'This will most likely not work.'
+    echo ''
+    echo 'Run "export PROFILE=emqx-enterprise" and "make" to fix this'
+    exit 1
+fi
+
 ERLANG_CONTAINER='erlang'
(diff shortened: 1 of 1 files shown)


id (dismissed):

scripts/ct/run.sh
```diff
@@ -80,6 +80,17 @@ if [ "${WHICH_APP}" = 'novalue' ]; then
     exit 1
 fi
 
+beginswith() { case $2 in "$1"*) true;; *) false;; esac; }
+
+if beginswith lib-ee "${WHICH_APP}" && { [ -z "${PROFILE+x}" ] || [ "${PROFILE}" != emqx-enterprise ]; }
```
id, line 85:
```suggestion
if [[ "${WHICH_APP}" == lib-ee* && ( -z "${PROFILE+x}" || "${PROFILE}" != emqx-enterprise) ]]; then
```

then `beginswith` function is not needed

scripts/ct/run.sh
```diff
@@ -80,6 +80,16 @@ if [ "${WHICH_APP}" = 'novalue' ]; then
     exit 1
 fi
 
+beginswith() { case $2 in "$1"*) true;; *) false;; esac; }
```
id, line 83: this one is not needed

kjellwinblad: sorry, must have forgotten to delete this line.

will fix.

kjellwinblad: Fixed!

id (approved):

PR #9936, opened 2023-02-08
feat: disable disksup by default

id:
Fixes https://emqx.atlassian.net/browse/EMQX-8869

This is one way to solve the issue.
If there is a standard accepted way of configuring OTP apps in emqx that I don't know about, please let me know.


zmstone:
great!
please add changlogs.

lafirest (approved):

PR #9938, opened 2023-02-08
feat(mqtt-bridge): report recoverable errors of async queries

keynslug:
This should help to avoid delivery failures of messages which could be safely retried, in the event of intermittent connectivity loss for example.

It should now be safe since 73d5592b.


zmstone:
app vsn check failed.

zmstone (approved):

PR #9939, opened 2023-02-08
Mria replicant manual join

ieQu1:
Start `emqx_ctl` process before `emqx_machine` application, to get the CLI working, even if the rest of the EMQX is down. It may happen when on replicant node when it can't join to any core node.

Sadly, because emqx bootstrapping is done in a separate application, and because `apps/emqx` may be used standalone, I had no other way, but to split emqx_ctl module to a separate app to make all static checks happy. This is, admittedly, a pretty hacky solution, since this new repo is useless on its own.

### DIFF
--- a/apps/emqx/src/emqx_kernel_sup.erl
+++ b/apps/emqx/src/emqx_kernel_sup.erl
@@ -35,7 +35,6 @@ init([]) ->
             child_spec(emqx_hooks, worker),
             child_spec(emqx_stats, worker),
             child_spec(emqx_metrics, worker),
-            child_spec(emqx_ctl, worker),
             child_spec(emqx_authn_authz_metrics_sup, supervisor)
         ]
     }}.
--- a/apps/emqx_conf/src/emqx_conf.app.src
+++ b/apps/emqx_conf/src/emqx_conf.app.src
@@ -1,6 +1,6 @@
 {application, emqx_conf, [
     {description, "EMQX configuration management"},
-    {vsn, "0.1.12"},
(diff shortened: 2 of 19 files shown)


mix.exs
```diff
@@ -88,7 +88,8 @@ defmodule EMQXUmbrella.MixProject do
       {:ranch,
        github: "ninenines/ranch", ref: "a692f44567034dacf5efcaa24a24183788594eb7", override: true},
       # in conflict by grpc and eetcd
-      {:gpb, "4.19.5", override: true, runtime: false}
+      {:gpb, "4.19.5", override: true, runtime: false},
+      {:emqx_ctl, github: "emqx/emqx_ctl", tag: "0.1.0", override: true}
```
zmstone, line 92: does it have to be an external dependency ?
should it work if we add a `apps/emqx_ctl` ?

zmstone: emqx app should not need to depend on emqx_ctl?

zmstone:
apps/emqx/test/emqx_ctl_SUITE.erl should be moved?

mix.exs
ieQu1: emqx app registers some CLI commands, so it does depend on it.

zmstone (approved):

zmstone:
emqx_ctl app ct failed?

ieQu1:
@zmstone 
> emqx_ctl app ct failed?

Indeed, there was a typo. Fixed.

zmstone (approved):

PR #9943, opened 2023-02-09
fix(auth): add more detailed logs for auth/acl

lafirest:


apps/emqx_auth_http/src/emqx_acl_http.erl
```diff
@@ -44,7 +44,10 @@ check_acl(ClientInfo, PubSub, Topic, _AclResult, #{acl := ACLParams = #{path :=
     Username = maps:get(username, ClientInfo1, undefined),
     case check_acl_request(ACLParams, ClientInfo1) of
         {ok, 200, <<"ignore">>} -> ok;
```
terry-xiaoyu, line 46: Should we also print some debug msgs when the auth/acl plugins is ignored/skipped?

lafirest: Feel like this will make the log too much? I'm not sure.

savonarola (approved):

terry-xiaoyu (approved):

PR #9967, opened 2023-02-13
perf: add tls common conf hibernate after

qzhuyan:
Fixes <issue-or-jira-number>

## PR Checklist


qzhuyan:
do you think we should intro a value `disabled` or `undefined` to turn it off or just use some semi xlarge value like 1000 hours?

currently having issues letting hocon and emqx_config cope with the value `undefined`.
And `disabled` need some extra convert work here and there that leads to extra overhead on maintaining the code since this opt is common opt for all TLS servers and clients.

apps/emqx/src/emqx_schema.erl
```diff
@@ -1929,6 +1929,15 @@ common_ssl_opts_schema(Defaults) ->
                     default => Df("secure_renegotiate", true),
                     desc => ?DESC(common_ssl_opts_schema_secure_renegotiate)
                 }
+            )},
+
+        {"hibernate_after",
+            sc(
+                duration(),
+                #{
+                    default => Df("hibernate_after", "5s"),
```
qzhuyan, line 1938: here follows emqx 4.x that has default `5s` and personally I think it will be much easier to make it default to be disabled `undefined`

apps/emqx/i18n/emqx_schema_i18n.conf
```diff
@@ -1495,6 +1495,17 @@ In case PSK cipher suites are intended, make sure to configure
...
 }
 
+common_ssl_opts_schema_hibernate_after {
+    desc {
+        en: """ Hibernate the SSL process after idling for amount of time reducing its memory footprint. """
+        zh: """ 在闲置一定时间后休眠 SSL 进程，减少其内存占用。"""
+    }
+    label: {
+        en: "hibernate after"
+        zh: "闲置后休眠时间"
```
lafirest, line 1505: Maybe `多久后休眠` will be better? Sorry, I can't think of a better word

lafirest (approved):

HJianBo (approved):

apps/emqx/src/emqx_schema.erl
zhongwencool: Personally, I'd recommend using `infinity` instead of `undefined` for this.
Although their effect is the same.
But our other configurations are using `infinity` 
such as 
```
 {"idle_timeout",
            sc(
                hoconsc:union([infinity, duration()])
                ...
```
https://www.erlang.org/doc/man/erlang.html#type-timeout
```
hibernate_after() = timeout()
timeout() = infinity | integer() >= 0
```

By the way, what happens if we set it to 0, is 0 the same as infinity? 
Maybe we should also mention this in the document.

PR #9975, opened 2023-02-15
ci: export docker image in build-slim-packages

id:
Fixes https://emqx.atlassian.net/browse/EMQX-8961


thalesmg (approved):

build
```diff
@@ -329,13 +329,15 @@ make_docker() {
     local default_tag="emqx/${PROFILE%%-elixir}:${PKG_VSN}"
     EMQX_IMAGE_TAG="${EMQX_IMAGE_TAG:-$default_tag}"
 
+    echo '_build' >> ./.dockerignore
```
thalesmg, line 332: Nit: maybe we could add a `.dockerignore` to the repo and avoid deleting it as a side effect (people could have their own)?  That could be useful once we start producing docker images for Elixir as well, since that'll probably include `deps`.

thalesmg (approved):

zmstone (approved):

build
id: @thalesmg ideally yes, but I don't know how to pull it off.
`Dockerfile.ubuntu20.04.runner` in the root needs `_build`, while `deploy/docker/Dockerfile` doesn't. And there is no way to specify custom `.dockerignore` on command line.

id: Not deleting existing file is a good point, I'll fix this in the next PR.

PR #9983, opened 2023-02-16
feat(ft): make storage backend fully async-aware

keynslug:
Introduce an ad-hoc concept of tasks that need to be kicked off manually. Rework filesystem backend to accommodate for this change. Adapt responder logic for that "kickoff" protocol.

---

[EMQX-8913](https://emqx.atlassian.net/browse/EMQX-8913)

### DIFF
--- a/apps/emqx_ft/src/emqx_ft_responder_sup.erl
+++ b/apps/emqx_ft/src/emqx_ft_responder_sup.erl
@@ -1,0 +1,48 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2020-2022 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+-module(emqx_ft_responder_sup).
+
+-export([start_link/0]).
+-export([start_child/3]).
+
+-behaviour(supervisor).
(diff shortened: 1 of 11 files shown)


apps/emqx_ft/src/emqx_ft_responder_sup.erl
```diff
@@ -0,0 +1,48 @@
...
+init(_) ->
+    Flags = #{
+        strategy => simple_one_for_one,
+        intensity => 100,
+        period => 100
+    },
+    ChildSpec = #{
+        id => responder,
+        start => {emqx_ft_responder, start_link, []},
+        restart => temporary
```
savonarola, line 46: `permanent `?

keynslug: This thing became a `simple_one_for_one` supervisor, and each responder is now a short-lived process, it should not be a problem if it explodes and dies.

keynslug: Right now it serves as a mutex and translates DOWN signals into callback calls. If you have an idea how to get rid of it, I'm all ears.

savonarola: As far as I understood, responders are now added dynamically, on eash async request; while this one is a "predefined" one and its purpose is unclear... also `emqx_ft_responder` seems to not have `emqx_ft_responder:start_link/0` 🤔

savonarola: Oh, seems I get it, nevermind :)

savonarola (approved):

apps/emqx_ft/test/emqx_ft_responder_SUITE.erl
```diff
@@ -39,58 +38,58 @@ init_per_testcase(_Case, Config) ->
...
+%     DefaultAction = fun(K) -> error({oops, K}) end,
+%     ?assertWaitEvent(
+%         emqx_ft_responder:start(Key, DefaultAction, 10),
+%         #{?snk_kind := ft_timeout_action_applied, key := <<"test">>},
+%         1000
+%     ),
+%     ?assertEqual(
+%         {error, not_found},
+%         emqx_ft_responder:ack(Key, oops)
+%     ).
```
savonarola, line 89: Shouldn't we test an unexpected crash in the callback?

PR #9984, opened 2023-02-16
Add regression test for resource buffer worker flush

olcai:
This PR adds a regression test for recursive flushing in `emqx_resource_buffer_worker`. The actual crash was fixed earlier in #9835.

Fixes EMQX-8784.

## PR Checklist

- [X] Changed lines covered in coverage report
- [ ] ~Change log has been added to `changes/<version>/(feat|fix)-<PR-id>.en.md` and `.zh.md` files~ (no, it's just a test)
- [ ] ~If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up~
- [ ] ~Schema changes are backward compatible~

### DIFF
--- a/apps/emqx_resource/test/emqx_resource_SUITE.erl
+++ b/apps/emqx_resource/test/emqx_resource_SUITE.erl
@@ -2291,6 +2291,67 @@ t_expiration_retry_batch_multiple_times(_Config) ->
     ),
     ok.
 
+t_recursive_flush(_Config) ->
+    emqx_connector_demo:set_callback_mode(async_if_possible),
+    {ok, _} = emqx_resource:create(
+        ?ID,
+        ?DEFAULT_RESOURCE_GROUP,
+        ?TEST_RESOURCE,
+        #{name => test_resource},
+        #{
+            query_mode => async,
+            batch_size => 1,
+            batch_time => 10_000,
+            worker_pool_size => 1
+        }
+    ),
+    do_t_recursive_flush().
+
+t_recursive_flush_batch(_Config) ->
+    emqx_connector_demo:set_callback_mode(async_if_possible),
+    {ok, _} = emqx_resource:create(
+        ?ID,
+        ?DEFAULT_RESOURCE_GROUP,
+        ?TEST_RESOURCE,
+        #{name => test_resource},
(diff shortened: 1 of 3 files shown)


apps/emqx_resource/test/emqx_resource_SUITE.erl
```diff
@@ -2291,6 +2291,67 @@ t_expiration_retry_batch_multiple_times(_Config) ->
...
+            ok
+        end,
+        fun(Trace) ->
+            %% check that a recursive flush leads to a new call to flush/1
+            Pairs = ?find_pairs(
+                #{?snk_kind := buffer_worker_flush_ack_reflush},
+                #{?snk_kind := buffer_worker_flush},
+                Trace
+            ),
+            ?assert(lists:any(fun(E) -> E end, [true || {pair, _, _} <- Pairs]))
```
thalesmg, line 2350: Just a note: from what I understand, it seems that there are a few sources for flushing that could lead to a `buffer_worker_flush` event, such as the return from the async call and the flush timer, so these could also give the impression that the "reflush" causes the `flush_worker` call.  

But I can't see a stronger way to assert that causality at this moment. :thinking: 

Did the `{internal, flush}` event cause problems in the past?  That could be a way to differentiate the cause (by the way, `running/3` with `internal, flush, _` is now unreachable).

I remember there was a [situation](https://github.com/emqx/emqx/blob/056bc71af26d486b220cb5780151e1d8ecef3f76/apps/emqx_resource/src/emqx_resource_buffer_worker.erl#L199-L203) when returning the internal event could crash, but it seemed to be resolved as the flush was called later via a timer.  :thinking:

thalesmg (approved):

apps/emqx_resource/test/emqx_resource_SUITE.erl
olcai: Thank you for your pointers! It was hard finding a good way in general to test this.

Stone and I found a crash while testing the influx bridge when `[{next_event, internal, flush}]}` was triggered. We simply patched it up and I have not thought about any deeper implications.

If it is okay, I'll go ahead and merge it.

PR #9985, opened 2023-02-16
feat: add clickhouse database bridge

kjellwinblad:
This PR adds a Clickhouse bridge to EMQX 5. The bridge is similar to the Clickhouse bridge in the 4.4, but adds the possibility to use different formats (such as JSON) for values to be inserted.

@ysfscream We need a configuration user interface in the dashboard for this as well. I have added a ticket for this as well (https://emqx.atlassian.net/browse/ED-429).


.ci/docker-compose-file/docker-compose-clickhouse.yaml
```diff
@@ -0,0 +1,16 @@
+version: '3.9'
+
+services:
+  clickhouse:
+    container_name: clickhouse
+    image: clickhouse/clickhouse-server:latest
```
thalesmg, line 6: Suggestion: pin a specific version to avoid this breaking after a new version is released.

lib-ee/emqx_ee_connector/i18n/emqx_ee_connector_clickhouse.conf
```diff
@@ -0,0 +1,15 @@
+
+emqx_ee_connector_clickhouse {
+
+    base_url {
+        desc {
+          en: """The HTTP URL to the Clickhouse server that you want to connect to (for example http://myhostname:8123)"""
+          zh: """"""
```
thalesmg, line 7: Just a reminder to fill these

lib-ee/emqx_ee_connector/test/ee_connector_clickhouse_SUITE.erl
```diff
@@ -0,0 +1,199 @@
...
+            "    password = public\n"
+            "    pool_size = 8\n"
+            "    url = \"http://~s:~b\"\n"
+            "    "
+            "",
+            [?CLICKHOUSE_HOST, ?CLICKHOUSE_DEFAULT_PORT]
+        )
+    ),
+
+    {ok, Config} = hocon:binary(RawConfig),
```
thalesmg, line 192: Suggestion: since this does not check against any schema, creating a map directly could be simpler to read and modify.

lib-ee/emqx_ee_connector/test/ee_connector_clickhouse_SUITE.erl
```diff
@@ -0,0 +1,199 @@
...
+                clickhouse:start_link([
+                    {url, clickhouse_url()},
+                    {user, <<"default">>},
+                    {key, "public"},
+                    {pool, tmp_pool}
+                ]),
+            {ok, _, _} = clickhouse:query(Conn, <<"CREATE DATABASE IF NOT EXISTS mqtt">>, #{}),
+            Config;
+        false ->
+            {skip, no_clickhouse}
```
thalesmg, line 71: We need to blow up in CI:

```suggestion
            case os:getenv("IS_CI") of
                "yes" ->
                    throw(no_clickhouse);
                _ ->
                    {skip, no_clickhouse}
            end

```

lib-ee/emqx_ee_connector/test/ee_connector_clickhouse_SUITE.erl
```diff
@@ -0,0 +1,199 @@
...
+            %% Create the db table
+            {ok, Conn} =
+                clickhouse:start_link([
+                    {url, clickhouse_url()},
+                    {user, <<"default">>},
+                    {key, "public"},
+                    {pool, tmp_pool}
+                ]),
+            {ok, _, _} = clickhouse:query(Conn, <<"CREATE DATABASE IF NOT EXISTS mqtt">>, #{}),
+            Config;
```
thalesmg, line 69: Nit: close the connection at some point?

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_clickhouse.conf
```diff
@@ -0,0 +1,95 @@
...
+SQL template to form a batch request. The value specified with
+this parameter will be inserted between the values. The default
+value ',' works for the VALUES format but other values
+might be needed if you specify some other format with the 
+clickhouse FORMAT syntax.
+
+See https://clickhouse.com/docs/en/sql-reference/statements/insert-into/ and
+https://clickhouse.com/docs/en/interfaces/formats#formats for more information about
+the format syntax and the available formats."""
+            zh: """"""
```
thalesmg, line 45: Just a reminder to fill these.

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_clickhouse_SUITE.erl
```diff
@@ -0,0 +1,329 @@
...
+            snabbkaffe:fix_ct_logging(),
+            %% Create the db table
+            Conn = start_clickhouse_connection(),
+            % erlang:monitor,sb
+            {ok, _, _} = clickhouse:query(Conn, sql_create_database(), #{}),
+            {ok, _, _} = clickhouse:query(Conn, sql_create_table(), []),
+            clickhouse:query(Conn, sql_find_key(42), []),
+            [{clickhouse_connection, Conn} | Config];
+        false ->
+            {skip, no_clickhouse}
```
thalesmg, line 42: We need to blow up in CI:

```suggestion
            case os:getenv("IS_CI") of
                "yes" ->
                    throw(no_clickhouse);
                _ ->
                    {skip, no_clickhouse}
            end

```

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_clickhouse_SUITE.erl
```diff
@@ -0,0 +1,329 @@
...
+
+all() ->
+    [
+        t_make_delete_bridge,
+        t_send_message_query,
+        t_send_simple_batch,
+        t_send_simple_batch_alternative_format,
+        t_heavy_batching,
+        t_heavy_batching_alternative_format
+    ].
```
thalesmg, line 88:
```suggestion
    emqx_common_test_helpers:all(?MODULE).
```
?

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_clickhouse_SUITE.erl
```diff
@@ -0,0 +1,329 @@
...
+    reset_table(Config),
+    BridgeID = make_bridge(#{enable_batch => false}),
+    Key = 42,
+    Payload = #{key => Key, data => <<"clickhouse_data">>, timestamp => 10000},
+    %% This will use the SQL template included in the bridge
+    emqx_bridge:send_message(BridgeID, Payload),
+    %% Check that the data got to the database
+    %%timer:sleep(1000),
+    % {ok, 200, ResultString} = clickhouse:query(ClickhouseConnection, sql_find_key(42), []),
+    % <<"42">> = iolist_to_binary(string:trim(ResultString)),
```
thalesmg, line 242: Tiny nit: remove commented code?

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_clickhouse_SUITE.erl
```diff
@@ -0,0 +1,329 @@
...
+
+    check_key_in_clickhouse(Key, Config),
+    delete_bridge(),
+    reset_table(Config),
```
thalesmg, line 246: Suggestion: we could always delete all bridges and reset the table in `{init,end}_per_testcase` so we avoid inter-testcase dependencies.

lib-ee/emqx_ee_connector/src/emqx_ee_connector_clickhouse.erl
```diff
@@ -0,0 +1,439 @@
...
+    maps:merge(values(put), #{name => <<"connector">>});
+values(get) ->
+    values(post);
+values(put) ->
+    #{
+        database => <<"mqtt">>,
+        enable => true,
+        pool_size => 8,
+        type => clickhouse,
+        url => <<"http://127.0.0.1:6570">>
```
thalesmg, line 118:
```suggestion
        url => <<"http://127.0.0.1:8123">>
```
?

lib-ee/emqx_ee_connector/src/emqx_ee_connector_clickhouse.erl
```diff
@@ -0,0 +1,439 @@
...
+    SQL = objects_to_sql(ObjectsToInsert, Templates),
+    %% Do the actual query in the database
+    ResultFromClickhouse = execute_sql_in_clickhouse_server(PoolName, SQL),
+    %% Transform the result to a better format
+    transform_and_log_clickhouse_result(ResultFromClickhouse, ResourceID, SQL).
+
+ensure_keys_are_of_type_send_message(Keys) ->
+    case lists:all(fun is_send_message_atom/1, Keys) of
+        true -> ok;
+        false -> erlang:error(<<"Unexpected type for batch message (Expected send_message)">>)
```
thalesmg, line 374: Maybe we could return an error value like `{error, {unrecoverable_error, Msg}}` so that the buffer workers won't retry this batch again?

kjellwinblad: I think this is probably a good suggestion (but I don't understand the bridge interface well enough to be certain). One problem that I have noticed while doing this bridge is that the interface to the bridge is difficult to understand. For example, how would a bridge creator know this without studying the the code for buffer workers in detail? Is this documented somewhere already? I have in my backlog to create a tutorial for creating bridges so I will need to understand the interface to make a good tutorial :) The buffer worker already catches errors in the callback. Would it be a good idea to define some special error type that has the same effect as returning `{error, {unrecoverable_error, Msg}}`? I think this could make the code of some bridges cleaner.

.ci/docker-compose-file/docker-compose-clickhouse.yaml
kjellwinblad: Fixed

lib-ee/emqx_ee_connector/i18n/emqx_ee_connector_clickhouse.conf
kjellwinblad: Fixed with auto translate. It should be checked by a Chinese speaker @JimMoen

lib-ee/emqx_ee_connector/test/ee_connector_clickhouse_SUITE.erl
kjellwinblad: Thanks, fixed

lib-ee/emqx_ee_connector/test/ee_connector_clickhouse_SUITE.erl
kjellwinblad: Fixed

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_clickhouse_SUITE.erl
kjellwinblad: Fixed

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_clickhouse_SUITE.erl
kjellwinblad: I now reset table in these hooks. I think it is fine to create and delete the bridge in the cases as the create bridge take different parameters depending on the test case.

lib-ee/emqx_ee_connector/src/emqx_ee_connector_clickhouse.erl
kjellwinblad: Fixed by changing the bridge interface to handle special types of expectation. See the code change in the buffer worker. What do you think about this change?

kjellwinblad:
I'm trying to sneak in a change that perhaps should have its own PR but I though it was so small so that it could go in this one [5ec7bf7](https://github.com/emqx/emqx/pull/9985/commits/5ec7bf79968fb4779b07b7b0b6e27513830e2838).

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_clickhouse.conf
kjellwinblad: Fixed with auto translate. It should be checked by a Chinese speaker @JimMoen

thalesmg:
> I'm trying to sneak in a change that perhaps should have its own PR but I though it was so small so that it could go in this one [5ec7bf7](https://github.com/emqx/emqx/pull/9985/commits/5ec7bf79968fb4779b07b7b0b6e27513830e2838).

Awesome, it is a bit annoying and I was just ignoring it.  :see_no_evil:

lib-ee/emqx_ee_connector/src/emqx_ee_connector_clickhouse.erl
thalesmg: True, we indeed don't have "internal tutorials", and often I also get bogged down trying to understand some flows... :see_no_evil: 

About this `unrecoverable_error` thingy, I just know about that because I spent a lot of time trying to refactor some pieces of that code (and introduced this concept with the help of Stone).

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -875,6 +875,15 @@ do_call_query(_QM, _Id, _Index, _Ref, _Query, _QueryOpts, _Data) ->
...
     catch
+        %% For convenience and to make the code in the callbacks cleaner an
+        %% error exception with the two following formats are translated to the
+        %% corresponding return values. The receiver of the return values
+        %% recognizes these special return formats and use them to decided if a
+        %% request should be retried.
+        error:{unrecoverable_error, Msg} ->
+            {error, {unrecoverable_error, Msg}};
+        error:{recoverable_error, Msg} ->
+            {error, {recoverable_error, Msg}};
```
thalesmg, line 886: I think that resources that do want to signal a recoverable error should just return an error value (like `{error, {recoverable_error | unrecoverable_error, _}}`) instead of `error({recoverable_error, _})`, so this wouldn't be strictly necessary.  

In that case, I'd suggest we only treat that case (error values) rather than open up the possibility of raising errors with the same meaning, so we have only one way to reason and signal those intentions rather than multiple.  

What do you think?

scripts/ct/run.sh
```diff
@@ -188,7 +191,7 @@ if [[ -t 1 ]]; then
 fi
 
 function restore_ownership {
-    if ! sudo chown -R "$ORIG_UID_GID" . >/dev/null 2>&1; then
+    if [ ! -z ${EMQX_TEST_DO_NOT_RUN_SUDO+x} ] || ! sudo chown -R "$ORIG_UID_GID" . >/dev/null 2>&1; then
```
thalesmg, line 194: :heart:

lib-ee/emqx_ee_connector/src/emqx_ee_connector_clickhouse.erl
```diff
@@ -0,0 +1,444 @@
...
+%% This function is similar to emqx_plugin_libs_rule:split_insert_sql/1 but can
+%% also handle Clickhouse's SQL extension for INSERT statments that allows the
+%% user to specify different formats:
+%%
+%% https://clickhouse.com/docs/en/sql-reference/statements/insert-into/
+%%
+split_clickhouse_insert_sql(SQL) ->
+    ErrorMsg = <<"The SQL template should be an SQL INSERT statement but it is something else.">>,
+    case
+        re:split(SQL, "(\\s+(?i:values)|(?i:format\\s+(?:[A-Za-z0-9_])+)\\s+)", [{return, binary}])
```
thalesmg, line 228: I noticed that we also have `emqx_plugin_libs_rule:split_insert_sql`, which seems to only cover the first part of this regex more or less.  Should we improve that version and use that here?

kjellwinblad: The format syntax that clickhouse has is not standard SQL so I don't think that will be useful for the other databases. Therefore, I think it is fine to have a specialized version just for clickhouse. What do you think?

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
kjellwinblad: My thinking was that by being able to throw an exception to signal these special types of errors one could make the code in the callback cleaner. Especially if the error happen deep down in the call stack and not in the top level callback function the code will get more messy if one have to check for error return values after every function call. Also, we capture all exceptions anyway so this just add the possibility to tell the caller more precisely what type of error it is when one throws an exception. Essentially one could accomplish the same thing by wrapping the callback function in a try-catch but if this is done in several bridges there will be a lot of unnecessary repetition.  You are a more experienced Erlang programmer than me so I will let you have the final say in this.

thalesmg: I see, I understand that it's currently inconvenient when we want to abort some long expression chain.  Hopefully we'll be able to use [`maybe`](https://www.erlang.org/doc/reference_manual/expressions.html#maybe) when we've fully moved to OTP 25 :smile_cat: .  

Usually, I'd say that returning error values rather than using exceptions for the flow control is easier to reason about.  But since it's a bit inconvenient in this case, let's roll with the `error` approach anyways.  :beers: 

(BTW I don't think I'm more experienced than you :joy_cat: )

lib-ee/emqx_ee_connector/src/emqx_ee_connector_clickhouse.erl
thalesmg: Sure, I just wanted to bring that other function to attention so we'd avoid missing fixes between the bridges that use SQL, in case both used the same logic and a bug were to be found affecting both.

scripts/spellcheck/dicts/emqx.txt
```diff
@@ -267,3 +267,5 @@ keytab
 jq
 nif
 TDengine
+Clickhouse
```
lafirest, line 270: Q: why it has two name format

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge.erl
```diff
@@ -183,3 +187,18 @@ pgsql_structs() ->
...
+        {Type,
+            mk(
+                hoconsc:map(name, ref(emqx_ee_bridge_clickhouse, "config")),
+                #{
+                    desc => <<Name/binary, "Clickhouse Bridge Config">>,
+                    required => false
+                }
+            )}
+     || {Type, Name} <- [
+            {clickhouse, <<"Clickhouse">>}
```
lafirest, line 202: not necessary to use a tuple list here ?

rebar.config
```diff
@@ -74,6 +74,7 @@
     , {esasl, {git, "https://github.com/emqx/esasl", {tag, "0.2.0"}}}
     , {jose, {git, "https://github.com/potatosalad/erlang-jose", {tag, "1.11.2"}}}
     , {telemetry, "1.1.0"}
+    , {hackney, {git, "https://github.com/benoitc/hackney", {tag, "1.18.1"}}}
```
lafirest, line 77: Q: what this for, does anywhere used the `hackney`

apps/emqx_bridge/src/emqx_bridge.erl
```diff
@@ -57,7 +57,8 @@
     T == influxdb_api_v2;
     T == redis_single;
     T == redis_sentinel;
-    T == redis_cluster
+    T == redis_cluster;
+    T == clickhouse
```
lafirest, line 61: ❤️
Thanks for these codes I found I forgot to update here in my code.

rebar.config
kjellwinblad: Yes, hackney is used by clickhouse.

scripts/spellcheck/dicts/emqx.txt
kjellwinblad: It was to make sure that the spell script would accept the capitalization. I removed it now to test that it goes through event without the capitalization.

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge.erl
kjellwinblad: Thanks, fixed in latest commit.

thalesmg (approved):

PR #9990, opened 2023-02-17
chore: add clean-changelog script

lafirest:


zmstone:
it'd be easier to read like below:

```
is_released() {
  file="$1"
  prefix="$2"
  latest_commit=$(git log -1 --pretty=format:"%H" "$file")
  git tag --contains $latest_commit | grep -q -E "^${prefix}5\.[0-9]+\.[0-9]+$"
}

for $file in ${ce_files[@]}; do
  if is_released "$file" 'v' && is_released "$file" 'e'; then
    git rm $file
  fi
done

for $file in ${ee_files[@]}; do
  if is_released "$file" 'e'; then
    git rm $file
  fi
done
```

I did not test it though.

scripts/clean-changelog.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
qzhuyan, line 1: nit, add some README

scripts/clean-changelog.sh
```diff
@@ -0,0 +1,65 @@
...
+        [ -n "$snd_check_name" ] && [[ ! "$snd_merged" =~ $PR ]] && flag=0
+        if [[ $flag -eq 1 ]]; then
+            echo "deleted $target PR: #$PR"
+            find "$dir" -name "*$PR*.md" -type f -delete
+        fi
+    done
+    git add "$dir"
+}
+
+do_clean "$ee_name" "$ee_name"
```
qzhuyan, line 60: question, shall we handle both `ee` and `ce` per script execution?

ieQu1:
I am not sure if we need to automate this. This script should be somewhat aware of our branching strategy and whatnot, so it will add maintenance burden for us. Git keeps all deleted files in the repo forever anyway, so it won't help to speed up cloning of the repo. 

I think it will be sufficient to simply delete files that are "obviously very old" from time to time manually.

zmstone:
> I am not sure if we need to automate this. This script should be somewhat aware of our branching strategy and whatnot, so it will add maintenance burden for us. Git keeps all deleted files in the repo forever anyway, so it won't help to speed up cloning of the repo.
> 
> I think it will be sufficient to simply delete files that are "obviously very old" from time to time manually.

not intended for automation.

PR #9996, opened 2023-02-18
0217 fix raw defaults in schema

zmstone:
Fixes [EMQX-8987](https://emqx.atlassian.net/browse/EMQX-8987)

The default values should be parsed HOCON term (as if returned from `hocon:binary/1`).
Which means the default values defined in schema modules should use `binary()` representation for string values.

`hocon_tconf` type check functions actually accepts `string()` type values as input,
but when dumping the raw config values to JSON artifacts, `string()` are treated as integer arrays in JSON.

This PR fixes all default string values to `binary()`

## PR Checklist

- [~] Added tests for the changes
- [~] Changed lines covered in coverage report
- ~~Change log has been added to `changes/<version>/(feat|fix)-<PR-id>.en.md` and `.zh.md` files~~ This change does not affect end-user.

### DIFF
--- a/apps/emqx/src/emqx_limiter/src/emqx_limiter_schema.erl
+++ b/apps/emqx/src/emqx_limiter/src/emqx_limiter_schema.erl
@@ -110,11 +110,11 @@ fields(limiter) ->
         ];
 fields(node_opts) ->
     [
-        {rate, ?HOCON(rate(), #{desc => ?DESC(rate), default => "infinity"})},
+        {rate, ?HOCON(rate(), #{desc => ?DESC(rate), default => <<"infinity">>})},
         {burst,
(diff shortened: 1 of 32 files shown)


lafirest:
does this mean we must use binary instead of string in the default field?

savonarola (dismissed):

zmstone:
> does this mean we must use binary instead of string in the default field?

* Default values MUST be parsed HOCON values (the output of `hocon:binary/1`), (e.g. `#{<<"key">> => <<"value">>}`)
* Default values are NOT hocon literals (e.g. `"{key = value}"`), (but environment variable overrides must be).
* Default values are NOT type-checked Erlang terms (e.g. `#{key => value}`)

id (approved):

PR #10006, opened 2023-02-20
feat(ft): add segment checksum validation

keynslug:
Also downgrade validation errors to mere info messages.

---

[EMQX-8916](https://emqx.atlassian.net/browse/EMQX-8916)


apps/emqx_ft/src/emqx_ft.erl
```diff
@@ -349,15 +359,26 @@ validate(Validations, Fun) ->
...
-                msg => "validate: invalid $file command",
-                reason => Reason
-            }),
+            ?tp(info, "client_violated_protocol", #{reason => Reason}),
             ?RC_UNSPECIFIED_ERROR
     end.
 
 do_validate([], Parsed) ->
     {ok, lists:reverse(Parsed)};
+do_validate([{fileid, FileId} | Rest], Parsed) ->
```
savonarola, line 368: 👍 validations are really better place for these checks

savonarola (approved):

PR #10020, opened 2023-02-22
fix(bridge): fix dropped counter and inflight gauge

zmstone:
Prior to this fix, there were three metrics-related issues when a bridge is configured to use batch + async mode:

1. When receiving a reply for a batch in which all requests had expired, the "late_reply" counter was only incremented by one instead of the batch size. This resulted in a counter discrepancy where "late_reply + dropped + succeeded" did not equal "matched".

2. When a batch was partially delivered or expired, the "dropped" requests were not decremented from the "inflight size" gauge. This caused the "inflight" gauge to never return to zero.

3. Upon receiving a batch reply, if the batch was partially expired, the "succeeded" counter was over-booked. This was because "late_reply" was counted as "succeeded".

Fixes [EMQX-8991](https://emqx.atlassian.net/browse/EMQX-8991)

## PR Checklist


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1341,9 +1378,9 @@ mark_inflight_items_as_retriable(Data, WorkerMRef) ->
     ok.
 
 %% used to update a batch after dropping expired individual queries.
-update_inflight_item(InflightTID, Ref, NewBatch) ->
+update_inflight_item(InflightTID, Ref, NewBatch, NumExpired) ->
     _ = ets:update_element(InflightTID, Ref, {?ITEM_IDX, NewBatch}),
-    ?tp(buffer_worker_worker_update_inflight_item, #{ref => Ref}),
+    _ = ets:update_counter(InflightTID, ?SIZE_REF, {2, -NumExpired, 0, 0}),
```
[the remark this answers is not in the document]
zmstone: Add a function guard `when NumExpired > 0`

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -327,19 +327,22 @@ resume_from_blocked(Data) ->
                     {next_state, running, Data}
             end;
         {expired, Ref, Batch} ->
-            IsAcked = ack_inflight(InflightTID, Ref, Id, Index),
-            IsAcked andalso emqx_resource_metrics:dropped_expired_inc(Id, length(Batch)),
+            %% assert true
+            true = ack_inflight(InflightTID, Ref, Id, Index),
```
thalesmg, line 331: Q: if it's inflight, it might also get ack'ed from the async return fn as a late reply?

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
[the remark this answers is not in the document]
thalesmg: I mean, maybe we should guard just this line?  Otherwise, trying to update an element without any expired items might blow up.

```suggestion
    NumExpired > 0 andalso (_ = ets:update_counter(InflightTID, ?SIZE_REF, {2, -NumExpired, 0, 0})),
```
?

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
zmstone: true, I forgot about this race condition.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -715,17 +714,20 @@ batch_reply_caller_defer_metrics(Id, BatchResult, Batch, QueryOpts) ->
...
+    {Action, PostFn1} = reply_caller_defer_metrics(Id, hd(Replies), QueryOpts),
+    PostFns =
         lists:foldl(
-            fun(Reply, {_ShouldAck, PostFns}) ->
-                {ShouldAck, PostFn} = reply_caller_defer_metrics(Id, Reply, QueryOpts),
-                {ShouldAck, [PostFn | PostFns]}
+            fun(Reply, PostFns) ->
+                {Action1, PostFn} = reply_caller_defer_metrics(Id, Reply, QueryOpts),
+                %% assert
+                true = (Action1 =:= Action),
```
thalesmg, line 723: Maybe we don't need this split and assert.  As the result is the same for the whole batch, the verdict should be the same for all replies.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1047,19 +1043,59 @@ handle_async_batch_reply(
...
+            %% the batch from reply context is minimized, so it cannot be used
+            %% to update the inflight items, hence discard Batch and lookup the RealBatch
+            ?tp(handle_async_reply_expired, #{expired => _Expired}),
+            handle_async_batch_reply2(ets:lookup(InflightTID, Ref), ReplyContext, Result, Now)
+    end.
+
+handle_async_batch_reply2([], _, _, _) ->
+    %% e.g. if the driver evaluates the callback more than once
+    %% which should really be a bug
+    ok;
```
keynslug, line 1062: Wondering if it's worth some form of rate-limited logging? Or test-only assertion?

keynslug (approved):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
zmstone: will add a log (unfortunately we don't have throttled-logger so far).

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1020,21 +1019,19 @@ do_handle_async_reply(
...
+    ok = maybe_flush_after_async_reply(IsFullBefore).
 
 handle_async_batch_reply(
     #{
-        buffer_worker := Pid,
-        resource_id := Id,
-        worker_index := Index,
         inflight_tid := InflightTID,
         request_ref := Ref,
         batch := Batch
```
savonarola, line 1037: M.b. we could explicitly express the fact that in reply contexts we have minified batches, not real ones?

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1317,13 +1361,18 @@ ack_inflight(InflightTID, Ref, Id, Index) ->
...
                 0
         end,
-    IsAcked = Count > 0,
-    IsAcked andalso ets:update_counter(InflightTID, ?SIZE_REF, {2, -Count, 0, 0}),
-    emqx_resource_metrics:inflight_set(Id, Index, inflight_num_msgs(InflightTID)),
-    IsAcked.
+    IsKnownRef = (Count > 0),
+    case IsKnownRef of
+        true ->
+            ets:update_counter(InflightTID, ?SIZE_REF, {2, -Count, 0, 0}),
```
savonarola, line 1370: Nit: here we have quite the same pattern as @thalesmg spotted, and the check got a bit separated from the update.
Maybe it is worth considering to hide the whole `ets:update_counter(..)` in an inline fun with a guard.

savonarola (dismissed):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1020,47 +1034,99 @@ do_handle_async_reply(
...
             %% Keep retrying.
-            mark_inflight_as_retriable(InflightTID, Ref),
+            ok = mark_inflight_as_retriable(InflightTID, Ref),
             ?MODULE:block(Pid);
         ack ->
-            do_ack(InflightTID, Ref, Id, Index, PostFn, Pid, QueryOpts)
-    end.
+            do_async_ack(InflightTID, Ref, Id, Index, PostFn, QueryOpts)
+    end,
+    ok = maybe_flush_after_async_reply(IsFullBefore).
```
thalesmg, line 1046: This should only run in the `ack` branch, right?  Otherwise, we will cast a `block` and a `flush` at the same time if the inflight is true?

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1020,47 +1034,99 @@ do_handle_async_reply(
...
+            %% because the batch can be quite big
+            ok = do_handle_async_batch_reply(ReplyContext, Result);
+        {_NotExpired, _Expired} ->
+            %% at least one is expired
+            %% the batch from reply context is minimized, so it cannot be used
+            %% to update the inflight items, hence discard Batch and lookup the RealBatch
+            ?tp(handle_async_reply_expired, #{expired => _Expired}),
+            ok = handle_async_batch_reply2(ets:lookup(InflightTID, Ref), ReplyContext, Result, Now)
+    end,
+    ok = maybe_flush_after_async_reply(IsFullBefore).
```
thalesmg, line 1089: Should do this only if there's at least one item in the batch that was acked?

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1101,9 +1166,44 @@ do_ack(InflightTID, Ref, Id, Index, PostFn, WorkerPid, QueryOpts) ->
...
+%% 1. If the request was previously deleted from inflight table due to
+%%    either succeeded previously or expired, this function logs a
+%%    warning message and returns 'discard' instruction.
+%% 2. If the request was previously failed and now pending on a retry,
+%%    then this function will return 'continue' as there is no way to
+%%    tell if this reply is stae or not.
+maybe_handle_unknown_async_reply(InflightTID, Ref) ->
+    try ets:member(InflightTID, Ref) of
+        true ->
+            %% NOTE: this does not mean the
```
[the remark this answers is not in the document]
zmstone: removed.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
zmstone: it's fine to cast a `flush` after `block` because the message is discarded.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1341,9 +1446,17 @@ mark_inflight_items_as_retriable(Data, WorkerMRef) ->
     ok.
 
 %% used to update a batch after dropping expired individual queries.
-update_inflight_item(InflightTID, Ref, NewBatch) ->
+update_inflight_item(InflightTID, Ref, NewBatch, NumExpired) ->
     _ = ets:update_element(InflightTID, Ref, {?ITEM_IDX, NewBatch}),
-    ?tp(buffer_worker_worker_update_inflight_item, #{ref => Ref}),
+    ok = dec_inflight(InflightTID, NumExpired),
```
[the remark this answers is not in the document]
zmstone: it's a bug, so it *SHOULD* blow up (I mean the when `Count > 0` function guard)

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
zmstone: it's find to cast `flush` after `block` because the message is discarded.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -715,17 +714,18 @@ batch_reply_caller_defer_metrics(Id, BatchResult, Batch, QueryOpts) ->
...
-                {ShouldAck, [PostFn | PostFns]}
+            fun(Reply, PostFns) ->
+                {_, PostFn} = reply_caller_defer_metrics(Id, Reply, QueryOpts),
+                [PostFn | PostFns]
             end,
-            {ack, []},
-            Replies
+            [PostFn1],
+            tl(Replies)
         ),
```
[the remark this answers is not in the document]
zmstone: ok, reverted.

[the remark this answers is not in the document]
zmstone: added a comment, it caused some confusion .

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
zmstone: for this one (non-batch), there are not many branches, so I changed it per your suggestion.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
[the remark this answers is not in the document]
thalesmg: Isn't it possible that we just want to update the batch (e.g.: set as retriable) when there are no expired items?  :thinking:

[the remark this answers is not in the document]
thalesmg: Since the fn is called `update_inflight_item`, it would be confusing if it required `NumExpires > 0` for any of its usages.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1020,47 +1033,100 @@ do_handle_async_reply(
...
-            emqx_resource_metrics:late_reply_inc(Id, NumExpired),
-            NumExpired > 0 andalso
-                ?tp(handle_async_reply_expired, #{expired => Expired}),
-            do_handle_async_batch_reply(ReplyContext#{batch := NotExpired}, Result)
-    end.
+        _ ->
+            %% some queries are not expired, put them back to the inflight batch
+            %% so it can be either acked now or retried later
+            ok = update_inflight_item(InflightTID, Ref, RealNotExpired, NumExpired),
+            ok = do_handle_async_batch_reply(ReplyContext#{min_batch := RealNotExpired}, Result)
```
thalesmg, line 1127: Just a note: the new `min_batch` is not minimized, but it doesn't seem to be a problem because it'll only show up in the test trace.

thalesmg (approved):
LGTM

The only thing is that `update_inflight_item` seems a bit strange in that it can only be called with expired items.  Later I saw the comment, but the fn name could be something like `drop_expired_inflight_items` or something else. :thinking:

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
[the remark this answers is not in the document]
zmstone: I believe I have all the `Expored =:= []` covered without having to call `update_inflight_item`.

thalesmg (approved):

apps/emqx_resource/test/emqx_resource_SUITE.erl
```diff
@@ -1528,49 +1529,33 @@ t_async_reply_multi_eval(_Config) ->
...
+            matched := Matched,
+            success := Success,
+            dropped := Dropped,
+            late_reply := LateReply,
+            failed := Failed
+        } = Counters,
+        ?assertEqual(TotalQueries, Matched - 1),
+        ?assertEqual(Matched, Success + Dropped + LateReply + Failed)
+    end,
+    loop_wait(F, _Interval = 5, TotalTime).
```
thalesmg, line 1558: Nit: there's `?retry/3` from `snabbkaffe` that does this.

thalesmg (approved):

thalesmg (approved):

PR #10022, opened 2023-02-22
230222 ci build el9 and macos12 packages

id:


.github/workflows/check_deps_integrity.yaml
```diff
@@ -1,11 +1,12 @@
...
 
-on: [pull_request, push]
+on:
+  pull_request:
 
 jobs:
   check_deps_integrity:
-    runs-on: ubuntu-20.04
-    container: ghcr.io/emqx/emqx-builder/5.0-28:1.13.4-25.1.2-2-ubuntu20.04
+    runs-on: ubuntu-latest
```
savonarola, line 8: Just wondering, why do we have `latest` somewhere (deps integrity) and also `22.04` somewhere (code style check) while in both cases a container is used?

id: Not sure tbh.
Replaced with `-latest` in less important workflows, and bumped to 22.04 in more critical ones.

.github/workflows/build_slim_packages.yaml
```diff
@@ -30,12 +30,11 @@ jobs:
       matrix:
         profile:
         - ["emqx", "24.3.4.2-2", "el7"]
-        - ["emqx", "24.3.4.2-2", "ubuntu20.04"]
         - ["emqx", "25.1.2-2", "ubuntu22.04"]
-        - ["emqx-enterprise", "24.3.4.2-2", "ubuntu20.04"]
-        - ["emqx-enterprise", "25.1.2-2", "ubuntu22.04"]
+        - ["emqx-enterprise", "24.3.4.2-2", "amzn2"]
+        - ["emqx-enterprise", "25.1.2-2", "ubuntu20.04"]
```
thalesmg, line 35: Q: should this be 22.04?

thalesmg (approved):

.github/workflows/build_slim_packages.yaml
id: Good to build test packages for both 20.04 and 22.04. Enterprise is more likely to use older systems.

thalesmg (approved):

PR #10023, opened 2023-02-23
fix(docker): fix env conflict for `EMQX_NODE_NAME` and `EMQX_NODE__NAME`

Rory-Z:
`docker-entrypoint` will set `EMQX_NODE_NAME` in env, and `emqx forground` will use it. but when user running `./bin/emqx` in container, emqx maybe not found `EMQX_NODE_NAME`, because it just work in `docker-entrypount`, so print `node.name` to `emqx.conf` to resolve it.

### DIFF
--- a/deploy/docker/docker-entrypoint.sh
+++ b/deploy/docker/docker-entrypoint.sh
@@ -39,9 +39,8 @@ if [[ -z "$EMQX_HOST" ]]; then
     export EMQX_HOST
 fi
 
-if [[ -z "$EMQX_NODE_NAME" ]]; then
-    export EMQX_NODE_NAME="$EMQX_NAME@$EMQX_HOST"
-fi
+export EMQX_NODE_NAME=${EMQX_NODE__NAME:-$EMQX_NAME@$EMQX_HOST}
+echo "node.name = $EMQX_NODE_NAME" >> /opt/emqx/etc/emqx.conf
 
 # The default rpc port discovery 'stateless' is mostly for clusters
 # having static node names. So it's troulbe-free for multiple emqx nodes


id (approved):

deploy/docker/docker-entrypoint.sh
```diff
@@ -39,9 +39,8 @@ if [[ -z "$EMQX_HOST" ]]; then
     export EMQX_HOST
 fi
 
-if [[ -z "$EMQX_NODE_NAME" ]]; then
-    export EMQX_NODE_NAME="$EMQX_NAME@$EMQX_HOST"
-fi
+export EMQX_NODE_NAME=${EMQX_NODE__NAME:-$EMQX_NAME@$EMQX_HOST}
+echo "node.name = $EMQX_NODE_NAME" >> /opt/emqx/etc/emqx.conf
```
zmstone, line 43:
```suggestion
if [[ -z "$EMQX_NODE_NAME" ]] && [[ -z "$EMQX_NODE_NAME" ]]; then
    export EMQX_NODE__NAME="$EMQX_NAME@$EMQX_HOST"
fi
```

zmstone: we should try not to touch emqx.conf as it can be read-only.

zmstone: another thing:
Currently 'EMQX_NODE_NAME' has higher precedence over 'EMQX_NODE__NAME'.
Although it's unlikely that one will set both intentionally, but changing the order might affect some current users.
This is why we should not take `EMQX_NODE__NAME` for `EMQX_NODE_NAME` in docker entrypoint.

zmstone:
https://github.com/emqx/emqx/pull/10027

PR #10035, opened 2023-02-27
perf(ft-asm): express coverage through shortest path on graph

keynslug:
The current algorithm's complexity proved to be unacceptable with sufficiently large inputs (e.g. GiBs-worth files split into 100 KiB sized segments for example). Express and reimplement it as the shortest path problem on graph where nodes are offsets and edges are Erlang nodes' segments.

---

[EMQX-8916](https://emqx.atlassian.net/browse/EMQX-8916)

### DIFF
--- a/apps/emqx_ft/src/emqx_ft_assembly.erl
+++ b/apps/emqx_ft/src/emqx_ft_assembly.erl
@@ -137,6 +137,10 @@ append_filemeta(Asm, Node, Fragment = #{fragment := {filemeta, Meta}}) ->
         meta = orddict:store(Meta, {Node, Fragment}, Asm#asm.meta)
     }.
 
+append_segmentinfo(Asm, _Node, #{fragment := {segment, #{size := 0}}}) ->
+    % NOTE
+    % Empty segments are valid but meaningless for coverage.
+    Asm;
 append_segmentinfo(Asm, Node, Fragment = #{fragment := {segment, Info}}) ->
     Offset = maps:get(offset, Info),
     Size = maps:get(size, Info),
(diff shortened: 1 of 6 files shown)


apps/emqx_ft/src/emqx_ft_assembly.erl
```diff
@@ -137,42 +137,160 @@ append_filemeta(Asm, Node, Fragment = #{fragment := {filemeta, Meta}}) ->
...
         % be able to disambiguate them though.
-        segs = orddict:store({Offset, locality(Node), -End, Node}, Fragment, Asm#asm.segs)
+        segs = Segs
     }.
 
-coverage([{{Offset, _, _, _}, _Segment} | Rest], Cursor, Sz) when Offset < Cursor ->
-    coverage(Rest, Cursor, Sz);
-coverage([{{Cursor, _Locality, MEnd, Node}, Segment} | Rest], Cursor, Sz) ->
+coverage(Segs, Size) ->
+    find_shortest_path(Segs, 0, Size).
```
savonarola, line 158: Could we use https://www.erlang.org/doc/man/digraph.html#get_short_path-3?

apps/emqx_ft/src/emqx_ft_assembly.erl
```diff
@@ -137,42 +137,160 @@ append_filemeta(Asm, Node, Fragment = #{fragment := {filemeta, Meta}}) ->
...
+
+find_shortest_path(G1, From, To) ->
```
savonarola, line 160: Graph-related code seems to be quite unrelated to the assembly domain 🤔 
Maybe we move it into a separate module?

Also it seems to be quite important to describe how coverage task domain maps to the graph domain (what are vertices, edges, and how path is related to the coverage)

savonarola:
Great application of the graph algorithm 👍

apps/emqx_ft/src/emqx_ft_assembly.erl
keynslug: Tried it initially, was not happy with the performance, not much better than our naive algorithm. Not sure why, haven't looked closely into the implementation.

keynslug: Also it has no concept of cost / weights, though that was easy to overcome with extra nodes.

savonarola (approved):

PR #10051, opened 2023-03-01
fix(buffer_worker): add batch time automatic adjustment 

thalesmg:
Related ticket: https://emqx.atlassian.net/browse/EMQX-9029

To avoid message loss due to misconfigurations, we adjust `batch_time`
based on `request_timeout`.  If `batch_time` > `request_timeout`, all
requests will timeout before being sent if the message rate is low.
Even worse if `pool_size` is high.  We cap `batch_time` at
`request_timeout div 2` as a rule of thumb.

Also:
- Small refactoring: we now avoid starting timers when `batch_time` = 0.  If the user sets `batch_time = 0`, it's analogous to `TCP_NODELAY`: we don't perform (time-based) batching.  Batching might still occur if the concurrent message rate is high enough.
- Removed potentially confusing `request_timeout` from the root of the GCP PubSub bridge (affects the frontend @ysfscream @Kinplemelon )
    - I tried to do it here: https://github.com/emqx/emqx-dashboard-web-new/pull/1014
- Added a conversion to check for a similar inconsistency in the webhook bridge / HTTP connector.  Removing that option from this bridge is trickier because its connector is shared among authn/authz.  So we just ensure both values are the same to avoid confusion.

### DIFF
--- a/apps/emqx_bridge/src/schema/emqx_bridge_schema.erl
+++ b/apps/emqx_bridge/src/schema/emqx_bridge_schema.erl
@@ -239,13 +239,6 @@ do_convert_webhook_config(
     case {MReqTRoot, MReqTResource} of
         {{ok, ReqTRoot}, {ok, ReqTResource}} ->
             {_Parsed, ReqTRaw} = max({ReqTRoot, ReqTRootRaw}, {ReqTResource, ReqTResourceRaw}),
-            ?SLOG(
-                debug,
-                #{
-                    msg => adjusting_webhook_bridge_request_time,
-                    new_value => ReqTRaw
-                }
-            ),
             Conf1 = emqx_map_lib:deep_merge(
                 Conf0,
                 #{
--- a/apps/emqx_bridge/src/schema/emqx_bridge_schema.erl
+++ b/apps/emqx_bridge/src/schema/emqx_bridge_schema.erl
@@ -17,6 +17,7 @@
 
 -include_lib("typerefl/include/types.hrl").
 -include_lib("hocon/include/hoconsc.hrl").
+-include_lib("emqx/include/logger.hrl").
 
 -import(hoconsc, [mk/2, ref/2]).
 
@@ -140,11 +141,7 @@ fields(bridges) ->
                 #{
                     desc => ?DESC("bridges_webhook"),
                     required => false,
-                    converter => fun(X, _HoconOpts) ->
-                        emqx_bridge_compatible_config:upgrade_pre_ee(
-                            X, fun emqx_bridge_compatible_config:webhook_maybe_upgrade/1
-                        )
-                    end
+                    converter => fun webhook_bridge_converter/2
                 }
             )},
         {mqtt,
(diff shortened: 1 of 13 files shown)


apps/emqx_bridge/src/schema/emqx_bridge_schema.erl
```diff
@@ -212,3 +213,26 @@ status() ->
 
 node_name() ->
     {"node", mk(binary(), #{desc => ?DESC("desc_node_name"), example => "emqx@127.0.0.1"})}.
+
+webhook_bridge_validator(
```
savonarola, line 217: Could it be a bit unintuitive to have two different settings that should be aways set equal?
M.b. one setting should suppress the other one and emit a warning?

thalesmg: Another idea I had was to use the converter to copy the greater value to both fields automatically and log that change.

thalesmg: Changed to this second approach; now the bridge won't fail, just use the greatest of the two possible values.

savonarola (dismissed):

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_gcp_pubsub.erl
```diff
@@ -79,15 +79,6 @@ fields(bridge_config) ->
                     desc => ?DESC("max_retries")
                 }
             )},
-        {request_timeout,
```
zmstone, line 82: should deprecate? removing it will cause compatibility issue. (if it's released in e5.0.0)

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_gcp_pubsub.erl
```diff
@@ -79,15 +79,6 @@ fields(bridge_config) ->
...
-            sc(
-                emqx_schema:duration_ms(),
-                #{
-                    required => false,
-                    default => <<"15s">>,
-                    desc => ?DESC("request_timeout")
```
zmstone, line 88: update the desc ?

zmstone:
please change default batch time to `0`.
I'll talk to frontend team to maybe hide it from users.

zmstone:
and improve docs like:
```
  batch_time {
    desc {
      en: """Maximum waiting interval when accumulating a batch at a low message rate for more efficient resource use."""
      zh: """在较低消息率情况下尝试累积批量输出时的最大等待间隔，以提高资源的利用率。"""
    }
    label {
      en: """Max Batch Wait Time"""
      zh: """批量等待最大间隔"""
    }
  }
```

zmstone:
For the record, I talked to product team. 
We can drop the batch_time to zero (but keep the field in the schema for backward compatibility reasons).

batch time at application layer is mostly considered an over-design. 
and we have been always optimizing towards lowest latency in other cases. 
e.g. for socket, we always set `nodelay` to true,
for Kafka producer, we always set `max_linger` time to 0.

apps/emqx_bridge/src/schema/emqx_bridge_schema.erl
```diff
@@ -212,3 +209,55 @@ status() ->
...
+    } = Conf0
+) ->
+    %% different values; we set them to the same, if they are valid
+    %% durations
+    MReqTRoot = emqx_schema:to_duration_ms(ReqTRootRaw),
+    MReqTResource = emqx_schema:to_duration_ms(ReqTResourceRaw),
+    case {MReqTRoot, MReqTResource} of
+        {{ok, ReqTRoot}, {ok, ReqTResource}} ->
+            {_Parsed, ReqTRaw} = max({ReqTRoot, ReqTRootRaw}, {ReqTResource, ReqTResourceRaw}),
+            ?SLOG(
```
zmstone, line 242: should not do logging in converters.

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_gcp_pubsub.erl
thalesmg: It's released there, indeed.  I'll change it to deprecation.

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_gcp_pubsub.erl
thalesmg: :heavy_check_mark:

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_gcp_pubsub.erl
thalesmg: :heavy_check_mark:

apps/emqx_bridge/src/schema/emqx_bridge_schema.erl
thalesmg: :heavy_check_mark:

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1639,3 +1648,46 @@ do_minimize(?QUERY(_ReplyTo, _Req, _Sent, _ExpireAt) = Query) -> Query.
...
 do_minimize(?QUERY(ReplyTo, _Req, Sent, ExpireAt)) -> ?QUERY(ReplyTo, [], Sent, ExpireAt).
 -endif.
+
+%% To avoid message loss due to misconfigurations, we adjust
+%% `batch_time' based on `request_timeout'.  If `batch_time' >
+%% `request_timeout', all requests will timeout before being sent if
+%% the message rate is low.  Even worse if `pool_size' is high.
+%% We cap `batch_time' at `request_timeout div 2' as a rule of thumb.
+adjust_batch_time(Id, RequestTimeout, BatchTime0) ->
+    BatchTime = max(0, min(BatchTime0, RequestTimeout div 2)),
```
qzhuyan, line 1658: Do we need to handle `infinity` here?

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_gcp_pubsub.conf
```diff
@@ -77,8 +77,8 @@ emqx_ee_bridge_gcp_pubsub {
 
     request_timeout {
         desc {
-          en: "HTTP request timeout."
-          zh: "HTTP 请求超时。"
+          en: "Deprecated: Configure the request timeout in the buffer settings."
+          zh: "废弃的。在缓冲区设置中配置请求超时。"
```
zmstone, line 81:
```suggestion
          zh: "已废弃。请在缓冲区设置中配置请求超时。"
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_gcp_pubsub.conf
```diff
@@ -77,8 +77,8 @@ emqx_ee_bridge_gcp_pubsub {
 
     request_timeout {
         desc {
-          en: "HTTP request timeout."
-          zh: "HTTP 请求超时。"
+          en: "Deprecated: Configure the request timeout in the buffer settings."
```
zmstone, line 80:
```suggestion
          en: "Deprecated. Please configure `request_timeout` in the buffer settings."
```

ieQu1 (approved):

PR #10052, opened 2023-03-01
0301 optimize boot failure logs

zmstone:
Fixes [EMQX-9052](https://emqx.atlassian.net/browse/EMQX-9052)

Improve daemon mode startup failure logs.

Before this change, it was difficult for users to understand the reason for EMQX 'start' command failed to boot the node.
The only information they received was that the node did not start within the expected time frame,
and they were instructed to boot the node with 'console' command in the hope of obtaining some logs.
However, the node might actually be running, which could cause 'console' mode to fail for a different reason.

With this new change, when daemon mode fails to boot, a diagnosis is issued. Here are the possible scenarios:

* If the node cannot be found from `ps -ef`, the user is instructed to find information in log files `erlang.log.*`.
* If the node is found to be running but not responding to pings, the user is advised to check if the host name is resolvable and reachable.
* If the node is responding to pings, but the EMQX app is not running, it is likely a bug. In this case, the user is advised to report a Github issue.

## PR Checklist


Rory-Z (approved):

bin/emqx
```diff
@@ -969,6 +974,59 @@ if [[ "$IS_BOOT_COMMAND" == 'yes' && "$(get_boot_config 'node.db_backend')" == "
     fi
 fi
 
+diagnos_boot_failure_and_die() {
```
id, line 977: Nit
```suggestion
diagnose_boot_failure_and_die() {
```

bin/emqx
```diff
@@ -969,6 +974,59 @@ if [[ "$IS_BOOT_COMMAND" == 'yes' && "$(get_boot_config 'node.db_backend')" == "
...
+    local ps_line
+    ps_line="$(find_emqx_process)"
+    if [ -z "$ps_line" ]; then
+        echo "Find more information in the latest log file: ${RUNNER_LOG_DIR}/erlang.log.*"
+        exit 1
```
id, line 982: might be a good idea to use different exit codes for different detected reasons

changes/ce/fix-10052.zh.md
```diff
@@ -0,0 +1,11 @@
+优化 EMQX daemo 模式启动启动失败的日志。
```
id, line 1:
```suggestion
优化 EMQX daemon 模式启动启动失败的日志。
```

apps/emqx_machine/src/emqx_machine.erl
```diff
@@ -47,6 +48,11 @@ start() ->
 graceful_shutdown() ->
     emqx_machine_terminator:graceful_wait().
 
+%% only used when failed to boot
+brutal_shutdown() ->
+    init:stop(),
+    timer:sleep(10_000).
```
id, line 54: what's this for?

thalesmg: If `init:stop/0` is called, does it reach this point?

bin/emqx
```diff
@@ -969,6 +974,59 @@ if [[ "$IS_BOOT_COMMAND" == 'yes' && "$(get_boot_config 'node.db_backend')" == "
...
+    WAIT_TIME="${EMQX_WAIT_FOR_STOP:-120}"
+    if ! wait_for "$WAIT_TIME" 'is_down' "$PID"; then
+        msg="dangling after ${WAIT_TIME} seconds"
+        # also log to syslog
+        logger -t "${REL_NAME}[${PID}]" "STOP: $msg"
+        # log to user console
+        set +x
+        logerr "Stop failed, $msg"
+        echo "ERROR: $PID is still around"
+        ps -p "$PID"
```
thalesmg, line 1023: Since there's a `set +x` just above, this won't be printed?

bin/emqx
```diff
@@ -1014,33 +1072,15 @@ case "${COMMAND}" in
             echo "$EMQX_DESCRIPTION $REL_VSN is started successfully!"
             exit 0
         else
-            echo "$EMQX_DESCRIPTION $REL_VSN failed to start in ${WAIT_TIME} seconds."
-            echo "Please find more information in erlang.log.N"
-            echo "Or run 'env DEBUG=1 $0 console' to have logs printed to console."
-            exit 1
+            logerr "${EMQX_DESCRIPTION} ${REL_VSN} using node name '${NAME}' failed ${WAIT_TIME} probs."
```
thalesmg, line 1075:
```suggestion
            logerr "${EMQX_DESCRIPTION} ${REL_VSN} using node name '${NAME}' failed ${WAIT_TIME} probes."
```

thalesmg (approved):

bin/emqx
zmstone: it prints

apps/emqx_machine/src/emqx_machine.erl
zmstone: sorry. I have removed this sleep.
the pipe evaluation exits immediately after EOF anyway.

thalesmg (approved):

PR #10053, opened 2023-03-01
Chore ensure docker latest tag release 50

id:


.github/workflows/build_and_push_docker_images.yaml
```diff
@@ -57,38 +58,33 @@ jobs:
...
+            PROFILE=emqx-enterprise
           else
             echo "This is NOT an exact git tag, will not publish images"
             is_exact='false'
           fi
-          case $tag in
-            refs/tags/v*)
-              PROFILE='emqx'
+
+          case "${PROFILE:-$INPUTS_PROFILE}" in
```
thalesmg, line 75: Q: shouldn't it be the other way around?  Like `${INPUTS_PROFILE:-PROFILE}`?

That way, if we use a workflow dispatch, we can force the profile we specified.

thalesmg (approved):

.github/workflows/build_and_push_docker_images.yaml
id: If we trigger this manually, then it most likely going to be on an existing tag. Can't think of a scenario when we'd need to override profile.

Issue #10025, opened 2023-02-24
max_awaiting_rel config option is not honoured


biggms8x8:
### What happened?

Attempted to adjust the max_awaiting_rel option in emqx.conf but it is not picked up, the limit seems to only be taken from the default (100).  This shows on both the dashboard and when attempting to overwhelm the broker with QoS 2 messages.
The following options were added to emqx.conf (as shown in emqx-example.conf)

```
mqtt {
   max_awaiting_rel = infinity
}

zones.external.mqtt {
   max_awaiting_rel = infinity
}

zones.internal.mqtt {
   max_awaiting_rel = infinity
}
```
After a suggestion i also edited "data/configs/cluster-override.conf" to add the max_awaitng_rel option, this too is ignored

The setting isn't picked up if the listener is in a zone or default.

Additionally setting this in the dashboard in the MQTT->Session also has no effect.

### What did you expect to happen?

max_awaiting_rel to be read from emqx.conf
max_awaiting_rel to be honoured when set in the dashboard

### How can we reproduce it (as minimally and precisely as possible)?

Add the following to emqx.conf
```
mqtt {
   max_awaiting_rel = infinity
}

zones.external.mqtt {
   max_awaiting_rel = infinity
}

zones.internal.mqtt {
   max_awaiting_rel = infinity
}
```

Observe in the dashboard for a connection "QoS 2 Message Receive Queue" is still 100

On the dashboard change Configuration->MQTT->Session->Max Awaiting PUBREL to "infinity"
Observe in the dashboard for a connection "QoS 2 Message Receive Queue" is still 100

### EMQX version

5.0.16

tigercl:
Hi, @biggms8x8 

If you have modified this configuration item through the Dashboard, then you need to modify `data/configs/cluster-override.conf` when you want to modify it again through the configuration file.

You can refer to [Configuration Files](https:// www.emqx.io/docs/en/v5.0/configuration/configuration.html)

biggms8x8:
Thanks @tigercl .. Just tried modifying that, I've added that file to the original post.  So that's the `emqx.conf`, dashboard settings and `cluster-override.conf` that aren't honoured.

tigercl:
@biggms8x8 Sorry, I check it again. I found that it does not work, this may be a bug.

We are investigating, I will synchronize the results with you ASAP.

kedar-hatkar:
[...]

PR #10055, opened 2023-03-02
fix: update max_awaiting_rel in session

zhongwencool:
FIx: https://emqx.atlassian.net/browse/EMQX-9057 
update max_awaiting_rel in session.
close https://github.com/emqx/emqx/issues/10025

### DIFF
--- a/apps/emqx/src/emqx_session.erl
+++ b/apps/emqx/src/emqx_session.erl
@@ -200,7 +200,7 @@
 
 -spec init(options()) -> session().
 init(Opts) ->
-    MaxInflight = maps:get(max_inflight, Opts, 1),
+    MaxInflight = maps:get(max_inflight, Opts),
     QueueOpts = maps:merge(
         #{
             max_len => 1000,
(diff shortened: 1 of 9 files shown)


apps/emqx/src/emqx_session.erl
```diff
@@ -211,17 +211,17 @@ init(Opts) ->
     #session{
         id = emqx_guid:gen(),
         clientid = maps:get(clientid, Opts, <<>>),
-        is_persistent = maps:get(is_persistent, Opts, false),
-        max_subscriptions = maps:get(max_subscriptions, Opts, infinity),
+        is_persistent = maps:get(is_persistent, Opts),
```
lafirest, line 214: Could we use `maps:merge` instead of these `maps:get`?

JimMoen (approved):

apps/emqx/src/emqx_session.erl
zhongwencool: It‘s a record， not a map.

lafirest (approved):

PR #10056, opened 2023-03-02
Bridge API: do not return 403 in case of inconsistency in application logic

sstrigler:
Fixes [EMQX-9064](https://emqx.atlassian.net/browse/EMQX-9064)

## PR Checklist


changes/ce/fix-10056.en.md
```diff
@@ -0,0 +1 @@
+`/bridges` API: return `400` instead of `403` in case of inconsistency in application logic either because bridge is about to be deleted but active rules still depend on it or an operation (start|stop|restart) is called but the bridge is not enabled.
```
ieQu1, line 1:
```suggestion
`/bridges` API: return `400` instead of `403` in case of inconsistency in the application logic either because bridge is about to be deleted, but active rules still depend on it, or an operation (start|stop|restart) is called, but the bridge is not enabled.
``` 

English punctuation is not my strongest skill, but I feel like this sentence is missing commas, since its clauses refer to different subjects:

https://dictionary.cambridge.org/grammar/british-grammar/punctuation

> When main clauses are separated by and, or, but, we don’t normally use a comma if the clauses have the same subject. However, we sometimes use commas if the clauses have different subjects

ieQu1 (approved):

PR #10058, opened 2023-03-02
fix: deprecate unsupp quic TLS options

qzhuyan:
The idea was to remove unsupp QUIC TLS options to avoid confusion.
But removing fields can cause schema backward compatible issues, so we decide to mark them deprecated. 

Manually checked that the Swagger spec file has these fields deprecated.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -1280,7 +1280,18 @@ fields("listener_wss_opts") ->
         true
     );
 fields("listener_quic_ssl_opts") ->
-    server_ssl_opts_schema(#{}, false);
+    %% Mark unsupported TLS options deprecated.
+    lists:map(
+        fun({Name, Schema}) ->
+            case is_quic_ssl_opts(Name) of
+                true ->
+                    {Name, Schema};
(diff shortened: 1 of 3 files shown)


lafirest:
Q: Could we just ignore these and not mark them as deprecated? I don't familiar about QUIC, but I worry maybe these can be supported someday.

lafirest:
Q: Could we just ignore these and not mark them as deprecated? I don't familiar about QUIC, but I worry maybe these can be supported someday.

qzhuyan:
> Q: Could we just ignore these and not mark them as deprecated? I don't familiar about QUIC, but I worry maybe these can be supported someday.

I agree with you and that is exactly what I planned to do, add support in future releases but QA team doesn't like this idea they don't accept settings that takes no effect (updating doc stating it takes no effect on QUIC protocol is not accepted as well).

qzhuyan:
and I assume we could remove the `deprecated` flag when I want to add them back.
Just want some input/review if it is possible without any other pitfalls.

thalesmg (approved):

savonarola (approved):

PR #10059, opened 2023-03-02
rule engine api format errors in a human readable way

sstrigler:
Fixes [EMQX-8740](https://emqx.atlassian.net/browse/EMQX-8740)

## PR Checklist


id (dismissed):

apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
```diff
@@ -423,7 +423,12 @@ param_path_id() ->
 %% Internal functions
 %%------------------------------------------------------------------------------
 
-err_msg(Msg) -> emqx_misc:readable_error_msg(Msg).
+err_msg({RuleError, {_E, R, _S}}) when is_tuple(R) ->
+    emqx_misc:readable_error_msg(emqx_json:encode([{RuleError, element(1, R)}]));
```
qzhuyan, line 427: just question, what if encode crashes?

sstrigler:
Screenshot 2023-03-03 at 09 04 01
Screenshot 2023-03-03 at 09 03 35
Some examples of error messages as they would look like now.

apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
sstrigler: Yeah, good question, I went with "let's just risk it" in this case. It's more of an issue with line 429 though. I can wrap it in a try catch and default to the original behavior.

qzhuyan (dismissed):

apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
```diff
@@ -423,7 +423,22 @@ param_path_id() ->
 %% Internal functions
 %%------------------------------------------------------------------------------
 
-err_msg(Msg) -> emqx_misc:readable_error_msg(Msg).
+err_msg({RuleError, {_E, Reason, _S}}) ->
```
sstrigler, line 426: We _could_ move this functionality to `emqx_misc:readable_error_msg` because it seems to be only used from API modules. Unfortunately this is not obvious and also it would kinda leaking abstractions then.

id (approved):

PR #10060, opened 2023-03-02
feat(ft): add GC logic and process for the FS storage backend

keynslug:
Implementation is straightforward, and is probably prone to some race conditions (e.g. collecting an incomplete transfer while there's another incoming segment). These conditions demand thorough testing.

Rough TODOs.
* [x] More basic tests
* [ ] ~~Filemeta preservation~~ (will be part of _exporters_ effort)
* [ ] ~~Collecting assembled transfers proactively (new BPAPI probably)~~ (followup)
* [ ] ~~Randomized concurrent testing for race conditions~~ (followup)
* [x] Proper GC config

---

[EMQX-8271](https://emqx.atlassian.net/browse/EMQX-8271)

### DIFF
--- a/apps/emqx_ft/src/emqx_ft_storage_fs.erl
+++ b/apps/emqx_ft/src/emqx_ft_storage_fs.erl
@@ -14,6 +14,12 @@
 %% limitations under the License.
 %%--------------------------------------------------------------------
 
+%% Filesystem storage backend
+%%
+%% NOTE
+%% If you plan to change storage layout please consult `emqx_ft_storage_fs_gc`
+%% to see how much it would break or impair GC.
+
 -module(emqx_ft_storage_fs).
 
 -behaviour(emqx_ft_storage).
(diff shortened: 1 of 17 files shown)


apps/emqx_ft/src/emqx_ft_storage_fs_gc.erl
```diff
@@ -0,0 +1,354 @@
...
+    gen_server:call(mk_server_ref(Storage), {collect, erlang:system_time()}, infinity).
+
+-spec reset(emqx_ft_storage_fs:storage()) -> ok.
+reset(Storage) ->
+    gen_server:cast(mk_server_ref(Storage), reset).
+
+collect(Storage, Transfer, Nodes) ->
+    gen_server:cast(mk_server_ref(Storage), {collect, Transfer, Nodes}).
+
+mk_server_ref(Pid) when is_pid(Pid) ->
```
savonarola, line 66: A bit unclear why this is a single fun — the second case in only start_link?

keynslug: Also tests and assembler. The first clause though is kind of artificial though, will remove it.

apps/emqx_ft/src/emqx_ft_storage_fs.erl
```diff
@@ -88,13 +100,25 @@
     %% System limit (e.g. number of ports) reached.
     | system_limit.
 
+%% Related resources childspecs
+-spec child_spec(storage()) ->
```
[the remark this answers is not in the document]
keynslug: Makes sense, but I do not see completely how it should work with backend-specific resources (as in this case). Personally I see it more natural and easier to understand when a backend provides its own `child_spec/1`. Admittedly it's a bit far-fetched right now, so I might just drop it in this PR, but this is a kind of illustration of what we eventually need to decide on when we'll be dealing with per-zone ft configs.

PR #10065, opened 2023-03-03
ci: release packages for raspbian

id:
https://emqx.atlassian.net/browse/EMQX-9076

### DIFF
--- a/.github/workflows/build_packages.yaml
+++ b/.github/workflows/build_packages.yaml
@@ -213,6 +213,8 @@ jobs:
           - ubuntu18.04
           - debian11
           - debian10
+          - raspbian10
+          - raspbian9
           - el9
           - el8
           - el7
@@ -229,6 +231,10 @@ jobs:
           build_machine: ubuntu-22.04
         - arch: amd64
           build_machine: aws-arm64
+        - arch: amd64
+          os: raspbian9
+        - arch: amd64
+          os: raspbian10
         include:
           - profile: emqx
             otp: 25.1.2-2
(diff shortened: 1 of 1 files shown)


.github/workflows/build_packages.yaml
```diff
@@ -372,6 +374,10 @@ jobs:
         push "debian/buster" "packages/$PROFILE/$PROFILE-$VERSION-debian10-arm64.deb"
         push "debian/bullseye" "packages/$PROFILE/$PROFILE-$VERSION-debian11-amd64.deb"
         push "debian/bullseye" "packages/$PROFILE/$PROFILE-$VERSION-debian11-arm64.deb"
+        push "raspbian/stretch "packages/$PROFILE/$PROFILE-$VERSION-raspbian9-amd64.deb"
+        push "raspbian/stretch" "packages/$PROFILE/$PROFILE-$VERSION-raspbian9-arm64.deb"
+        push "raspbian/buster" "packages/$PROFILE/$PROFILE-$VERSION-raspbian10-amd64.deb"
```
thalesmg, line 379: I thought raspbian was arm64 only :see_no_evil:

thalesmg (dismissed):

.github/workflows/build_packages.yaml
id: oops

thalesmg (approved):

id:
@Nosferatu31 sorry about that, I did not do my homework when submitting this PR.
The problem is that raspbian is 32 bits, while we run everything on 64 bits. We don't have resources to support 32 bits arch just for raspbian as of now unfortunately.
Could you suggest which 64bit distro we could build packages for which is used for RaspbianPi?

PR #10066, opened 2023-03-03
fix: return human readable error message for most common cases

sstrigler:
Fixes [EMQX-8932](https://emqx.atlassian.net/browse/EMQX-8932)

## PR Checklist


apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -420,6 +420,9 @@ schema("/bridges/:id/:operation") ->
             ],
             responses => #{
                 204 => <<"Operation success">>,
+                400 => error_schema(
+                    'BAD_REQUEST', "Problem with configuration of external service"
```
kjellwinblad, line 424: Maybe it is possible to be more specific? In what situation do we give back a BAD_REQUEST error? If you get this when the operation does not exist, then maybe something like this would be more informative?

"The operation does not exist."

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -440,7 +443,10 @@ schema("/nodes/:node/bridges/:id/:operation") ->
             ],
             responses => #{
                 204 => <<"Operation success">>,
-                400 => error_schema('BAD_REQUEST', "Forbidden operation, bridge not enabled"),
+                400 => error_schema(
+                    'BAD_REQUEST',
+                    "Problem with configuration of external service or bridge not enabled"
```
kjellwinblad, line 448: Can one be more specific about what the configuration problem might be?

kjellwinblad (approved):
Looks good to me. I just had some questions about if it is possible to be more specific in the error message for the 400 responses.

apps/emqx_bridge/src/emqx_bridge_api.erl
sstrigler: I'm afraid that's a limitation of the schema descriptions (OpenAPI/swagger?), it doesn't allow for a list of possible outcomes (or does it?). That's why I tried to describe the nature of possible messages. The actual error message is different and rather specific.

apps/emqx_bridge/src/emqx_bridge_api.erl
sstrigler: That was part of some other ticket and has been added already, see line below. We then return a `404` now.

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -555,8 +561,8 @@ schema("/bridges_probe") ->
             case emqx_bridge_resource:create_dry_run(ConnType, maps:remove(<<"type">>, Params)) of
                 ok ->
                     {204};
-                {error, Error} ->
-                    {400, error_msg('TEST_FAILED', Error)}
+                {error, Reason} when not is_tuple(Reason); element(1, Reason) =/= 'exit' ->
+                    {400, error_msg('TEST_FAILED', to_hr_reason(Reason))}
```
savonarola, line 565: 🤔 Won't this fail with 500 and case clause on `{error, {exit, ...}}`? Is this intentional?

savonarola (approved):

apps/emqx_bridge/src/emqx_bridge_api.erl
sstrigler: Yes it will and it's by intention. Don't think it's worth its own handler. Yes, the error will be one level more nested but it shouldn't happen in the first place, so if we see it, we'll fix whatever caused it and we have less code to maintain. 🤷

id (approved):

PR #10067, opened 2023-03-03
feat: add ocsp stapling and crl support to mqtt ssl listener (5.0)

thalesmg:
https://emqx.atlassian.net/browse/EMQX-8816


apps/emqx/src/emqx_crl_cache.erl
```diff
@@ -0,0 +1,320 @@
...
+        URLs
+    ),
+    ?tp(crl_cache_refresh_config, #{
+        refresh_interval => RefreshIntervalMS,
+        http_timeout => HTTPTimeoutMS,
+        urls => URLs
+    }),
+    {noreply, State};
+handle_cast({refresh_config, Config}, State0) ->
+    case Config of
```
lafirest, line 161: nit: a little redundant, repeated the above cause

apps/emqx/src/emqx_listeners.erl
```diff
@@ -785,3 +790,24 @@ quic_listener_optional_settings() ->
...
+
+inject_sni_fun(ListenerId, Conf = #{ssl_options := #{ocsp := #{enable_ocsp_stapling := true}}}) ->
+    emqx_ocsp_cache:inject_sni_fun(ListenerId, Conf);
+inject_sni_fun(_ListenerId, Conf) ->
+    Conf.
+
+inject_crl_config(
+    Conf = #{ssl_options := #{crl := #{enable_crl_check := true} = CRLOpts} = SSLOpts}
+) ->
+    HTTPTimeout = maps:get(cache_refresh_http_timeout, CRLOpts, timer:seconds(15)),
```
lafirest, line 802: nit: use a macro instead of the 15

apps/emqx/src/emqx_ocsp_cache.erl
```diff
@@ -0,0 +1,521 @@
...
+    ok = ?MODULE:register_listener(ListenerID, Conf),
+    Conf.
+
+%%--------------------------------------------------------------------
+%% gen_server behaviour
+%%--------------------------------------------------------------------
+
+init(_Args) ->
+    logger:set_process_metadata(#{domain => [emqx, ocsp, cache]}),
+    _ = ets:new(?CACHE_TAB, [
```
savonarola, line 152: Do we consider OK to lose cache on process crash?

apps/emqx/test/emqx_crl_cache_SUITE_data/emqx_crl_cache_http_server.erl
```diff
@@ -0,0 +1,73 @@
...
+    {noreply, State}.
+
+handle_info(_Info, State) ->
+    {noreply, State}.
+
+terminate(_Reason, _State) ->
+    stop_http().
+
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
```
ieQu1, line 39: Nit: Can be omitted

apps/emqx/test/emqx_crl_cache_SUITE_data/emqx_crl_cache_http_server.erl
```diff
@@ -0,0 +1,73 @@
...
+
+handle_call(_Request, _From, State) ->
+    {reply, ignored, State}.
+
+handle_cast(_Msg, State) ->
+    {noreply, State}.
+
+handle_info(_Info, State) ->
+    {noreply, State}.
+
```
ieQu1, line 34: Nit: can be omitted

apps/emqx/src/emqx_crl_cache.erl
```diff
@@ -0,0 +1,320 @@
...
+        http_timeout = HTTPTimeoutMS
+    }
+) ->
+    case maps:get(URL, RefreshTimers, undefined) of
+        TRef ->
+            ?tp(crl_refresh_timer, #{url => URL}),
+            ?SLOG(debug, #{
+                msg => "refreshing_crl_response",
+                url => URL
+            }),
```
ieQu1, line 200: Nit: ?tp(debug, "refreshing_crl_response", ...)
?

apps/emqx/src/emqx_schema.erl
```diff
@@ -1289,6 +1317,75 @@ fields("listener_quic_ssl_opts") ->
...
+            sc(
+                boolean(),
+                #{
+                    default => false,
+                    desc => ?DESC("server_ssl_opts_schema_enable_ocsp_stapling")
+                }
+            )},
+        {"responder_url",
+            sc(
+                binary(),
```
ieQu1, line 1332: There is a common typerefl reflection for URLs: `emqx_connector_http:url()`

apps/emqx/i18n/emqx_schema_i18n.conf
```diff
@@ -1747,6 +1747,110 @@ server_ssl_opts_schema_gc_after_handshake {
     }
 }
 
+server_ssl_opts_schema_enable_ocsp_stapling {
+    desc {
+        en: "Whether to enable OCSP stapling for the listener.  If set to true,"
```
ieQu1, line 1752:
```suggestion
        en: "Whether to enable Online Certificate Status Protocol
 (OCSP) stapling for the listener.  If set to true,"
```
?

apps/emqx/i18n/emqx_schema_i18n.conf
```diff
@@ -1747,6 +1747,110 @@ server_ssl_opts_schema_gc_after_handshake {
...
+    }
+    label: {
+        en: "Enable CRL Check"
+        zh: "启用CRL检查"
+    }
+}
+
+server_ssl_opts_schema_crl_cache_urls {
+    desc {
+        en: "URLs for CRL servers to fetch and cache CRLs from."
```
ieQu1, line 1820:
```suggestion
        en: "URLs for Certificate Revocation List (CRL) servers to fetch and cache CRLs from."
```
I think it would be nice to explain the acronyms at least once.

apps/emqx/src/emqx_crl_cache.erl
```diff
@@ -0,0 +1,320 @@
...
+-spec gather_config() ->
+    #{
+        urls := [binary()],
+        refresh_interval := timer:time(),
+        http_timeout := timer:time()
+    }.
+gather_config() ->
+    TLSListeners = emqx_config:get([listeners, ssl], #{}),
+    URLs = collect_urls(TLSListeners),
+    RefreshIntervalMS0 = emqx_config:get([crl_cache, refresh_interval], timer:minutes(15)),
```
thalesmg, line 307: TODO: might need to add a special handler to refresh this, since the freshest value is not available during the `post_config_update` of the listener.

apps/emqx/src/emqx_crl_cache.erl
thalesmg: True, I've refactored this bit.  Thanks!

thalesmg:
> just a question, does this compatible with the old version configuration or other data?

Since this just adds new fields, it should be compatible with schemas without those new fields.

apps/emqx/src/emqx_schema.erl
thalesmg: This returns a map of type `emqx_http_lib:uri_map()`, right?  Is there a way to use that type just to validate, but to keep this URL as a (normalized) binary?  Because we just use this by concatenating a path to this fixed URL.

ieQu1 (approved):

apps/emqx/src/emqx_schema.erl
ieQu1: Ok. Maybe add a new reusable type for this usecase?

apps/emqx/src/emqx_ocsp_cache.erl
thalesmg: Excellent point :thinking: 

The effect of losing the cache would be that the response is not sent to the connecting client until it's restored.  So, if the client requires that, then it could lead to a period of clients not being able to connect (due to their own policy).

I think we could mitigate that risk by making the process supervisor the heir of this table and then making it `public`.  Does this make sense?

apps/emqx/src/emqx_schema.erl
```diff
@@ -1294,6 +1326,68 @@ fields("listener_quic_ssl_opts") ->
...
+    [
+        {"enable_crl_check",
+            sc(
+                boolean(),
+                #{
+                    default => false,
+                    desc => ?DESC("server_ssl_opts_schema_enable_crl_check")
+                }
+            )},
+        {"cache_urls",
```
zmstone, line 1382: maybe just `urls` ?

thalesmg:
Closing this; I'll open separate PRs for the OCSP Stapling and CRL Check parts.

Sorry for the inconvenience.  :speak_no_evil:

PR #10068, opened 2023-03-03
test: fix flaky mqtt bridge test

thalesmg:
Sometimes, this test fails because the metrics are still in the inflight phase.


qzhuyan:
Try to avoid using sleeps, maybe retry is better? 

I am also cleaning some flaky tests due to sleeps

thalesmg:
> Try to avoid using sleeps, maybe retry is better?
> 
> I am also cleaning some flaky tests due to sleeps

Ok, I'll change it to use `?retry`.

savonarola (approved):

qzhuyan (approved):

PR #10069, opened 2023-03-03
Ci/emqx 9055 configuration item flush time interval used for stats d integration not work

SergeTupchiy:
Closes: https://emqx.atlassian.net/browse/EMQX-9055

Note: preferably, we should also add 'sample_time_interval' to the dashboard (statsd integration).

### DIFF
--- a/apps/emqx_statsd/src/emqx_statsd_config.erl
+++ b/apps/emqx_statsd/src/emqx_statsd_config.erl
@@ -45,9 +45,9 @@ remove_handler() ->
     ok = emqx_config_handler:remove_handler(?STATSD),
     ok.
 
-post_config_update(?STATSD, _Req, #{enable := true}, _Old, _AppEnvs) ->
+post_config_update(?STATSD, _Req, #{enable := true} = New, _Old, _AppEnvs) ->
     emqx_statsd_sup:ensure_child_stopped(?APP),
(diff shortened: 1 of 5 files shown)


apps/emqx_statsd/src/emqx_statsd_config.erl
```diff
@@ -45,9 +45,9 @@ remove_handler() ->
     ok = emqx_config_handler:remove_handler(?STATSD),
     ok.
 
-post_config_update(?STATSD, _Req, #{enable := true}, _Old, _AppEnvs) ->
+post_config_update(?STATSD, _Req, #{enable := true} = New, _Old, _AppEnvs) ->
     emqx_statsd_sup:ensure_child_stopped(?APP),
-    emqx_statsd_sup:ensure_child_started(?APP);
+    emqx_statsd_sup:ensure_child_started(?APP, New);
```
SergeTupchiy, line 50: `post_config_update/5` cb is called before an updated config is saved: https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_config_handler.erl#L292, 
so it's was not possible for the restarted process to read the updated config in its `init/1` cb.

savonarola (approved):

id (approved):

PR #10076, opened 2023-03-06
fix: lost messages when HTTP connection times out

kjellwinblad:
When using async mode with the webhook bridge, queued messages that are not fully processed when the connection times out could be lost. This commit fixes this by letting the bridge return a recoverable_error when this happen. The message send will then be retried in sync mode by the emqx_resource_buffer_worker.

Fixes: https://emqx.atlassian.net/browse/EMQX-8974


apps/emqx_bridge/test/emqx_bridge_webhook_SUITE.erl
```diff
@@ -0,0 +1,252 @@
...
+    ok = emqx_common_test_helpers:start_apps([emqx_conf, emqx_bridge]),
+    ok = emqx_connector_test_helpers:start_apps([emqx_resource]),
+    {ok, _} = application:ensure_all_started(emqx_connector),
+    {ok, _} = application:ensure_all_started(emqx_ee_connector),
+    {ok, _} = application:ensure_all_started(emqx_ee_bridge),
+    snabbkaffe:fix_ct_logging(),
+    [].
+
+end_per_suite(_Config) ->
+    ok = emqx_common_test_helpers:stop_apps([emqx_conf]),
```
thalesmg, line 45:
```suggestion
    ok = emqx_common_test_helpers:stop_apps([emqx_conf, emqx_bridge]),
```

apps/emqx_bridge/test/emqx_bridge_webhook_SUITE.erl
```diff
@@ -0,0 +1,252 @@
...
+    _ = application:stop(emqx_connector),
+    _ = application:stop(emqx_ee_connector),
+    _ = application:stop(emqx_bridge),
+    ok.
+
+suite() ->
+    [{timetrap, {seconds, 60}}].
+
+%%------------------------------------------------------------------------------
+%% HTTP server for testing
```
thalesmg, line 55: No need to change here since it ended up a very short implementation, but, as an idea, another option would be to use `cowboy` to make a quick-and-dirty server (ex: `apps/emqx_authn/test/emqx_authn_http_test_server.erl`)

thalesmg (dismissed):

apps/emqx_bridge/test/emqx_bridge_webhook_SUITE.erl
```diff
@@ -0,0 +1,249 @@
...
+        Name,
+        BridgeConfig
+    ),
+    emqx_bridge_resource:bridge_id(Type, Name).
+
+%%------------------------------------------------------------------------------
+%% Testcases
+%%------------------------------------------------------------------------------
+
+%% This test ensures that https://emqx.atlassian.net/browse/CI-62 is fixed.
```
savonarola, line 203: Nit: Should we keep Jira links in the code? M.b. give a short description directly?

apps/emqx_bridge/test/emqx_bridge_webhook_SUITE.erl
kjellwinblad: Thanks for the suggestion. Yes, that sounds more robust. I followed the example from the emqx_bridge_api_SUITE and thought it was a good idea to make a small custom thing so one could have total control over what the server is doing but I ended up spending a lot of time debugging some problem with the test server so I think it would have been better to just use cowboy from the beginning (I think it would probably be possible to configure cowboy to do what we want as well).

apps/emqx_bridge/test/emqx_bridge_webhook_SUITE.erl
```diff
@@ -0,0 +1,249 @@
...
+    ok.
+
+receive_request_notifications(MessageIDs, _ResponseDelay) when map_size(MessageIDs) =:= 0 ->
+    ok;
+receive_request_notifications(MessageIDs, ResponseDelay) ->
+    receive
+        {http_server, received, Req} ->
+            RemainingMessageIDs = remove_message_id(MessageIDs, Req),
+            receive_request_notifications(RemainingMessageIDs, ResponseDelay)
+    after (30 * 1000) ->
```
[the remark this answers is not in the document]
kjellwinblad: This will not work. After one gives back a recoverable_error from the bridge there is an additional delay until the message is retried. This delay seems to be a few seconds but I'm not sure exactly which parameter controls this delay.

apps/emqx_bridge/test/emqx_bridge_webhook_SUITE.erl
kjellwinblad: I don't think it hurts to have a reference to the bug report that motivated the test case. If one wonders why this test case exists one can go an have a look at the bug report for additional details. I agree with you that it would be good with an explanation here as well and I thought I had written one but I see now that it is only half finished so I will fix that.

apps/emqx_bridge/test/emqx_bridge_webhook_SUITE.erl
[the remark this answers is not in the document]
thalesmg: So it seems that `ResponseDelay` is not used at all by this fn.  Not a big deal, just confirming my understanding.

thalesmg (dismissed):

zmstone:
changelog is missing.

id:
Could not push to Kjell's fork https://github.com/emqx/emqx/pull/10088

zmstone:
> Could not push to Kjell's fork #10088

there is also some flakyness in the test case. fixed in my commit.

changes/ce/fix-10076.en.md
```diff
@@ -0,0 +1,2 @@
+Fix webhook bridge error handling: connection timeout should be a retriable error.
+Prior to this fix, connection timeout was clasified as unrecoverable error hence lead to request being dropped.
```
id, line 2:
```suggestion
Prior to this fix, connection timeout was classified as unrecoverable error and led to request being dropped.
```

id (approved):

thalesmg (approved):

id (approved):

thalesmg (approved):

PR #10083, opened 2023-03-07
feat(bridges): integrate DynamoDB with data bridges

lafirest:


lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_dynamo.conf
```diff
@@ -0,0 +1,72 @@
+emqx_ee_bridge_dynamo {
+
+    local_topic {
+        desc {
+            en: """The MQTT topic filter to be forwarded to DynamoDB. All MQTT 'PUBLISH' messages with the topic
+matching the local_topic will be forwarded.</br>
+NOTE: if this bridge is used as the action of a rule (EMQX rule engine), and also local_topic is
+configured, then both the data got from the rule and the MQTT messages that match local_topic
+will be forwarded."""
```
ieQu1, line 9:
```suggestion
            en: """The MQTT topic filter to be forwarded to DynamoDB. All MQTT `PUBLISH` messages with the topic
matching the local_topic will be forwarded.</br>
NOTE: if this bridge is used as the action of a rule (EMQX rule engine), and also `local_topic` is
configured, then both the data got from the rule and the MQTT messages that match `local_topic`
will be forwarded."""
```

ieQu1 (approved):

PR #10095, opened 2023-03-08
fix(mysql): be explicit that batch queries are parameterless

keynslug:
So that mysql client won't attempt to prepare them automatically, thus trashing the server's prepared statements table and making whole interaction overall heavier. Also add tests to verify that we don't allocate more prepared statements than needed.

This PR also fixes an issue where some strings might have been inserted inconsistently, since escaping logic missed some edge cases.

---

[EEC-782](https://emqx.atlassian.net/browse/EEC-782)
[EMQX-9167](https://emqx.atlassian.net/browse/EMQX-9167)


lib-ee/emqx_ee_connector/src/emqx_ee_connector_tdengine.erl
```diff
@@ -160,7 +160,7 @@ status_result(_Status = false) -> connecting.
 %%========================================================================================
 
 do_batch_insert(InstanceId, BatchReqs, InsertPart, Tokens, State) ->
-    SQL = emqx_plugin_libs_rule:proc_batch_sql(BatchReqs, InsertPart, Tokens),
+    SQL = emqx_plugin_libs_rule:proc_batch_mysql(BatchReqs, InsertPart, Tokens),
```
thalesmg, line 163: Q: since this is also used in TDengine, maybe the function name could be something more generic (perhaps the old name)?

Either this or "fork" the old function so that TDengine uses the old version?

apps/emqx_plugin_libs/src/emqx_placeholder.erl
```diff
@@ -312,8 +314,53 @@ unwrap(<<"\"${", Val/binary>>, _StripDoubleQuote = true) ->
...
+    bin(Val).
+
+-spec escape_mysql(binary()) -> iolist().
+escape_mysql(S0) ->
+    % https://dev.mysql.com/doc/refman/8.0/en/string-literals.html
+    [$', escape_mysql(S0, 0, 0, S0), $'].
+
+%% NOTE
+%% This thing looks more complicated than needed because it's optimized for as few
+%% intermediate memory (re)allocations as possible.
```
thalesmg, line 344: :brain:

thalesmg (dismissed):
I'm still digesting the escaping logic, but don't want to block the fix.

apps/emqx_plugin_libs/src/emqx_placeholder.erl
```diff
@@ -28,7 +28,7 @@
     preproc_sql/1,
     preproc_sql/2,
     proc_sql/2,
-    proc_sql_param_str/2,
+    proc_mysql_param_str/2,
```
lafirest, line 31: I think adding a new function with this name may be better than replacing the generic function, there are many SQL-likes backend.

lafirest:
@keynslug 
awesome PR, could you share what happened and the reason for the performance improvement?
And a change file in English is required, I can translate it into Chinese.

keynslug:
Thanks, @lafirest.

> could you share what happened and the reason for the performance improvement?

This is actually a speculation if this will improve performance, but I suspect it will, slightly, because we now won't hit whole layer of [caching prepared statements](https://github.com/emqx/mysql-otp/blob/63e0be1d3f6d0074e50aa20c86a392ac68f54ef8/src/mysql_conn.erl#L312-L332) in the client.

> And a change file in English is required, I can translate it into Chinese.

Sure, will do shortly, along with other post-review changes.

keynslug:
> could you share what happened and the reason for the performance improvement?

Replied too quickly. 😅 What I meant is that while trying to hit some hot codepath and pinpoint where does a performance regression come from, under my workload I actually observed message loss due to prepared statements table exhaustion on the server. So this PR is not actually addressing a root cause for the performance regression, but theoretically, might contribute to it.

Also hit some concerning error from the server (about SQL syntax error) that forced me to reimplement escaping logic.

apps/emqx_plugin_libs/src/emqx_placeholder.erl
keynslug: Did that as part of https://github.com/emqx/emqx/pull/10095/commits/2544742b0cf02aed1707b068237bb3762651b663. I tried to raise some awareness there about `pgsql` in a comment, would be glad if you take another look.

lib-ee/emqx_ee_connector/src/emqx_ee_connector_tdengine.erl
keynslug: Did that as part of https://github.com/emqx/emqx/pull/10095/commits/2544742b0cf02aed1707b068237bb3762651b663. I tried to raise some awareness there about pgsql in a comment, would be glad if you take another look.

thalesmg (approved):

keynslug:
> And a change file in English is required, I can translate it into Chinese.

Hi @lafirest, I've attached a changelog entry with DeepL-ed Chinese translation, much appreciate if you could take a quick read-through.

changes/ee/fix-10095.zh.md
```diff
@@ -0,0 +1,3 @@
+阻止 mysql 客户端在每个批处理中用不必要的 `PREPARE` 查询反复轰炸mysql服务器，破坏服务器并耗尽其内部限制。当 mysql 桥处于批处理模式时，这种情况正在发生。
```
lafirest, line 1: 优化 MySQL 桥接在批量模式下能更高效的使用`预处理语句` ，减少了对 MySQL 服务器的查询压力， 并确保对 SQL 语句进行更安全和谨慎的转义。

lafirest: I didn't translate it word for word, but the meaning is correct.

changes/ee/fix-10095.en.md
```diff
@@ -0,0 +1,3 @@
+Stop mysql client from bombarding mysql server repeatedly with unnecessary `PREPARE` queries on every batch, trashing the server and exhausting its internal limits. This was happening when the mysql bridge was in the batch mode.
```
lafirest, line 1: nit: mysql -> MySQL

Rory-Z (approved):

zmstone (approved):

Issue #9783, opened 2023-01-16
emqx_authz_mongodb - Mongo 6 - InvalidNamespace


mattiabenin:
#### Environment
 
- EMQX K8S: 5.0.14
- MongoDB: 6.0.3-debian-11-r0

#### Description

I configure EMQX security authentication and authorization with MongoDB.

Authentication works correctly.

Authorization I see this error to logs when I try to subscribe an auth topic. I receive same error also when I try publish.

[...]

PR #10098, opened 2023-03-08
fix: mongodb authz crash

kjellwinblad:
When configuring mongodb authorization the mongodb connector crashed with the following error in the log file. The reason is that the collection name that was sent to the mongodb connection was an atom. This is fixed by making sure it is not an atom. An even better fix would probably be to to figure out why the configuration data for the connection has the collection stored as an atom and fix the issue at the source.

```
2023-03-08T17:16:34.215523+01:00 [error] msg: query_mongo_error, mfa: emqx_authz_mongodb:authorize/4, line: 95, peername: 127.0.0.1:53212,
clientid: client123, collection: mqtt_acl, filter: #{username =>
<<"emqx_u">>}, reason: {resource_error,#{msg => #{error =>
{error,{error_cannot_parse_response,{op_msg_response,#{<<"code">> => 73,<<"codeName">> => <<"InvalidNamespace">>,<<"errmsg">> => <<"Failed to parse namespace element">>,<<"ok">> => 0.0}}}},id => <<"emqx_authz_mongodb:3">>,name => call_query,request => {find,mqtt_acl,#{username => <<"emqx_u">>},#{}},stacktrace => [{mc_connection_man,reply,1,[{file,"mc_connection_man.erl"},{line,123}],
...]}, reason => exception}}, resource_id: <<"emqx_authz_mongodb:3">>
```

Fixes: https://github.com/emqx/emqx/issues/9783


thalesmg (dismissed):
:heavy_check_mark: 

Just changelogs are missing?

kjellwinblad:
@thalesmg 
> Just changelogs are missing?

Fixed

savonarola (dismissed):

changes/v5.0.20-en.md
```diff
@@ -0,0 +1,5 @@
+# v5.0.20
```
thalesmg, line 1: The way we generate changelogs has changed recently.  Now we create one markdown file per {fix,feat}x{en,md} under `changes/{ce,ee}/`, referencing the PR number only in the filename.

So, for this PR, it would go to `changes/ce/fix-10098.{en,zh}.md` with the contents like:

```
When configuring mongodb authorization the mongodb connector crashed with an error in the log file. The reason was that the collection name had the wrong type.
```

apps/emqx_connector/src/emqx_connector_mongo.erl
```diff
@@ -256,6 +258,14 @@ on_query(
             {ok, Result}
     end.
 
+%% For some reason the collection name is turned into an atom by the emqx_authz
+%% application which causes the mongo query to break. TODO: This should probably
+%% be fixed at another level.
+make_collection_bin(Atom) when is_atom(Atom) ->
```
zmstone, line 264: the field is of type `binary()` if it ended up as an atom, we should investigate (because it could be a potential atom leak here).

changes/v5.0.20-en.md
```diff
@@ -0,0 +1,5 @@
...
+
+## Bug Fixes
+
+- [#10098](https://github.com/emqx/emqx/pull/10098) When configuring mongodb authorization the mongodb connector crashed with an error in the log file. The reason was that the collection name had the wrong type.
```
zmstone, line 5:
```suggestion
- Fix mongodb authorization crash when `collection` is configured.
```

changes/v5.0.20-zh.md
```diff
@@ -0,0 +1,5 @@
+# v5.0.20
+
+## 修复
+
+- [#10098](https://github.com/emqx/emqx/pull/10098) 在配置mongodb授权时，mongodb连接器崩溃了，日志文件中出现了错误。原因是集合名称的类型不对。
```
zmstone, line 5: 修复 MongoDB 授权配置中设置 `collection` 后的一个异常。

changes/v5.0.20-en.md
kjellwinblad: ahh, yeah, of course. I don't understand how I could forget.

changes/v5.0.20-en.md
kjellwinblad: The crash does not happen when the collection is configured. It happens when the authorization system tries to query the mongodb database.

apps/emqx_connector/src/emqx_connector_mongo.erl
kjellwinblad: I will try to investigate this.

changes/v5.0.20-en.md
kjellwinblad: Fixed

apps/emqx_connector/src/emqx_connector_mongo.erl
kjellwinblad: Fixed by changing the authz schema (see new commit below). I don't understand why the authz schema don't piggyback on the mongodb connector schema but maybe there is a good reason for that?

kjellwinblad:
Squashed the commits after latest changes.

savonarola (approved):

PR #10100, opened 2023-03-08
Fix channel crash for slow clients with enhanced authentication

savonarola:
Also  gracefully handle `Authentication-Data` absence.


changes/ce/fix-10100.zh.md
```diff
@@ -0,0 +1 @@
+使用增强的身份验证修复慢速客户端的通道崩溃。
```
lafirest, line 1: 修复响应较慢的客户端在使用增强认证时可能出现崩溃的问题。

savonarola: ✅

lafirest (approved):

changes/ce/fix-10100.en.md
```diff
@@ -0,0 +1 @@
+Fix channel crash for slow clients with enhanced authentication.
```
ieQu1, line 1: Nit: It would be nice explain the implications of the error from users' perspective.

ieQu1 (approved):

PR #10103, opened 2023-03-09
Continue 10056

JimMoen:


sstrigler (commented):
Tried to solve the same problem at my current PR https://github.com/emqx/emqx/pull/10107

But my solution moves that check to `emqx_bridge_resource:start`. wdyt?

PR #10117, opened 2023-03-10
fix: copy plugins to a new node joining a cluster

SergeTupchiy:
Closes: https://emqx.atlassian.net/browse/EMQX-8889

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
@@ -323,8 +323,6 @@ get_plugins() ->
 upload_install(post, #{body := #{<<"plugin">> := Plugin}}) when is_map(Plugin) ->
     [{FileName, Bin}] = maps:to_list(maps:without([type], Plugin)),
     %% File bin is too large, we use rpc:multicall instead of cluster_rpc:multicall
-    %% TODO what happens when a new node join in?
-    %% emqx_plugins_monitor should copy plugins from other core node when boot-up.
     NameVsn = string:trim(FileName, trailing, ".tar.gz"),
     case emqx_plugins:describe(NameVsn) of
         {error, #{error := "bad_info_file", return := {enoent, _}}} ->
@@ -456,8 +454,8 @@ delete_package(Name) ->
 
 %% for RPC plugin update
 ensure_action(Name, start) ->
-    _ = emqx_plugins:ensure_enabled(Name),
     _ = emqx_plugins:ensure_started(Name),
+    _ = emqx_plugins:ensure_enabled(Name),
     ok;
 ensure_action(Name, stop) ->
     _ = emqx_plugins:ensure_stopped(Name),
--- a/apps/emqx_plugins/src/proto/emqx_plugins_proto_v1.erl
+++ b/apps/emqx_plugins/src/proto/emqx_plugins_proto_v1.erl
@@ -1,0 +1,35 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2022-2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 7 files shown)


apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -456,8 +456,8 @@ delete_package(Name) ->
 
 %% for RPC plugin update
 ensure_action(Name, start) ->
-    _ = emqx_plugins:ensure_enabled(Name),
     _ = emqx_plugins:ensure_started(Name),
+    _ = emqx_plugins:ensure_enabled(Name),
```
SergeTupchiy, line 460: The order is important here: first start a plugin then enable it in the conf.
Besides the case when config can be enabled successfully but then fail to be started, there is one more (specific) case when the order is important:

1. Node `A` has plugin `P` which is installed but not started, so it is not listed in the conf (`emx_conf:get([plugins,states])` is empty).
2. Node `B` joins the cluster (node `A`), it doesn't copy plugin `P` because it is not listed in the conf yet (since it's not started on any node).
3. Plugin `P` is started from the dashboard (on any node), and is now being started on all the nodes (via  `emqx_mgmt_api_plugins_proto_v1:ensure_action/2`).
4.  `emqx_plugins:ensure_started(Name)` must now be called before `ensure_enabled` on node `B`, because it does ensure that plugin exists (will copy it from node `A').
5. After this, `emqx_plugins:ensure_enabled/1` can be safely called on node `B`. It will read the plugin description and add it to the `[plugins,states]` conf as enabled.

changes/ce/fix-10117.en.md
```diff
@@ -0,0 +1,2 @@
+Fix an error occurring when a joining node doesn't have plugins that are installed on other nodes in the cluster.
```
ieQu1, line 1: Tip: deepl.com usually does a pretty ok job translating changelogs to Chinese.

apps/emqx_plugins/src/proto/emqx_plugins_proto_v1.erl
```diff
@@ -0,0 +1,37 @@
...
+-define(TIMEOUT, 25000).
+
+-type name_vsn() :: binary() | string().
+
+introduced_in() ->
+    "5.0.21".
+
+-spec get_tar(node(), name_vsn()) -> {ok, binary()} | {error, any}.
+get_tar(Node, NameVsn) ->
+    rpc:call(Node, emqx_plugins, get_tar, [NameVsn], ?TIMEOUT).
```
ieQu1, line 37: It would be nice to test this RPC with a very large file and with EMQX under traffic. OTP team did some optimizations to make sure large messages don't block up the distribution socket, but there still could be some performance pitfalls related to transfering very large binaries over RPC.

apps/emqx_plugins/src/proto/emqx_plugins_proto_v1.erl
```diff
@@ -0,0 +1,37 @@
...
+-behaviour(emqx_bpapi).
+
+-export([
+    introduced_in/0,
+    get_tar/2
+]).
+
+-include_lib("emqx/include/bpapi.hrl").
+
+-define(TIMEOUT, 25000).
```
ieQu1, line 28: Maybe infinity is a safer choice? Or, perhaps, make it configurable

changes/ce/fix-10117.en.md
SergeTupchiy: Thanks, I followed the latest instruction from @id to add only English changes and than have a batch translation before a release.

apps/emqx_plugins/src/proto/emqx_plugins_proto_v1.erl
SergeTupchiy: Fixed:
`get_tar(Node, NameVsn, Timeout)`
and call it:
`emqx_plugins_proto_v1:get_tar(Node, NameVsn, infinity)`

apps/emqx_plugins/src/proto/emqx_plugins_proto_v1.erl
SergeTupchiy: I agree, but this is essentially not different to what we already have in `emqx_mgmt_api_plugins.erl`:  
https://github.com/emqx/emqx/blob/master/apps/emqx_management/src/emqx_mgmt_api_plugins.erl#L324
Perhaps we should add validation for dashboard/CLI plugin uploads to check the max size of initially added plugin?

ieQu1: I see. If the existing code is already designed like that (I haven't studied it prior to the review), then this PR follows the design, however it is.

savonarola (approved):

ieQu1 (approved):

ieQu1 (approved):

PR #10119, opened 2023-03-13
fix: don't allow empty string in server_port schema

zhongwencool:
`statsd.server`(emqx_schema:server_sc/1) is not allowed to be empty.
we can't connect to an empty host even with the default port.


lafirest (dismissed):

apps/emqx/src/emqx_schema.erl
```diff
@@ -2816,20 +2817,17 @@ is_port_number(Port) ->
...
-    try
-        P = list_to_integer(string:strip(Port)),
-        true = (P > 0),
-        true = (P =< 65535),
-        P
-    catch
-        _:_ ->
-            throw("bad_port_number")
+    case string:to_integer(string:strip(Port)) of
+        {P, ""} when P < 0 -> throw("port_number_too_small");
```
zmstone, line 2821:
```suggestion
        {P, ""} when P =< 0 -> throw("port_number_must_be_positive");
```

changes/ce/fix-10119.en.md
```diff
@@ -0,0 +1 @@
+Fix crash when `statsd.server` is empty
```
zmstone, line 1:
```suggestion
Fix crash when `statsd.server` is set to an empty string.
```

changes/ce/fix-10119.zh.md
```diff
@@ -0,0 +1 @@
+修复 `statsd.server` 为空时启动崩溃的问题
```
zmstone, line 1:
```suggestion
修复 `statsd.server` 配置为空字符串时启动崩溃的问题。
```

zmstone (approved):

PR #10123, opened 2023-03-13
perf(bridge-api): ask nodes' bridge listings in parallel

keynslug:
Also make sure to reply `emqx_resource:get_instance/1` from cache.

The resource manager may be busy at times, so this change ensures that getting resource instance state will not block. Currently, no users of `emqx_resource:get_instance/1` do seem to be relying on resource state being "as-actual-as-possible" guarantee it was providing.

---

[EMQX-9136](https://emqx.atlassian.net/browse/EMQX-9136)

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_api.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_api.erl
@@ -483,11 +483,18 @@ schema("/bridges_probe") ->
             end
     end;
 '/bridges'(get, _Params) ->
-    {200,
-        zip_bridges([
-            [format_resp(Data, Node) || Data <- emqx_bridge_proto_v1:list_bridges(Node)]
-         || Node <- mria:running_nodes()
-        ])}.
+    Nodes = mria:running_nodes(),
+    NodeReplies = emqx_bridge_proto_v3:list_bridges_on_nodes(Nodes),
+    case is_ok(NodeReplies) of
+        {ok, NodeBridges} ->
+            AllBridges = [
+                format_resource(Data, Node)
+             || {Node, Bridges} <- lists:zip(Nodes, NodeBridges), Data <- Bridges
+            ],
+            {200, zip_bridges([AllBridges])};
+        {error, Reason} ->
+            {500, error_msg('INTERNAL_ERROR', Reason)}
+    end.
 
 '/bridges/:id'(get, #{bindings := #{id := Id}}) ->
     ?TRY_PARSE_ID(Id, lookup_from_all_nodes(BridgeType, BridgeName, 200));
@@ -591,7 +598,7 @@ lookup_from_all_nodes_metrics(BridgeType, BridgeName, SuccCode) ->
 
 do_lookup_from_all_nodes(BridgeType, BridgeName, SuccCode, FormatFun) ->
     Nodes = mria:running_nodes(),
-    case is_ok(emqx_bridge_proto_v1:lookup_from_all_nodes(Nodes, BridgeType, BridgeName)) of
+    case is_ok(emqx_bridge_proto_v3:lookup_from_all_nodes(Nodes, BridgeType, BridgeName)) of
         {ok, [{ok, _} | _] = Results} ->
             {SuccCode, FormatFun([R || {ok, R} <- Results])};
         {ok, [{error, not_found} | _]} ->
@@ -602,7 +609,7 @@ do_lookup_from_all_nodes(BridgeType, BridgeName, SuccCode, FormatFun) ->
 
 lookup_from_local_node(BridgeType, BridgeName) ->
     case emqx_bridge:lookup(BridgeType, BridgeName) of
-        {ok, Res} -> {ok, format_resp(Res)};
+        {ok, Res} -> {ok, format_resource(Res, node())};
         Error -> Error
     end.
 
@@ -802,10 +809,7 @@ aggregate_metrics(
 aggregate_metrics(#{}, Metrics) ->
     Metrics.
 
-format_resp(Data) ->
-    format_resp(Data, node()).
-
-format_resp(
+format_resource(
     #{
         type := Type,
         name := BridgeName,
@@ -974,7 +978,7 @@ do_bpapi_call(Node, Call, Args) ->
 do_bpapi_call_vsn(SupportedVersion, Call, Args) ->
     case lists:member(SupportedVersion, supported_versions(Call)) of
         true ->
-            apply(emqx_bridge_proto_v2, Call, Args);
+            apply(emqx_bridge_proto_v3, Call, Args);
         false ->
             {error, not_implemented}
     end.
@@ -984,9 +988,9 @@ maybe_unwrap({error, not_implemented}) ->
 maybe_unwrap(RpcMulticallResult) ->
     emqx_rpc:unwrap_erpc(RpcMulticallResult).
 
-supported_versions(start_bridge_to_node) -> [2];
-supported_versions(start_bridges_to_all_nodes) -> [2];
-supported_versions(_Call) -> [1, 2].
+supported_versions(start_bridge_to_node) -> [2, 3];
+supported_versions(start_bridges_to_all_nodes) -> [2, 3];
+supported_versions(_Call) -> [1, 2, 3].
 
 to_hr_reason(nxdomain) ->
     <<"Host not found">>;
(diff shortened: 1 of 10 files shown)


apps/emqx_resource/src/emqx_resource.erl
```diff
@@ -309,7 +311,7 @@ set_resource_status_connecting(ResId) ->
 -spec get_instance(resource_id()) ->
     {ok, resource_group(), resource_data()} | {error, Reason :: term()}.
 get_instance(ResId) ->
-    emqx_resource_manager:lookup(ResId).
+    emqx_resource_manager:ets_lookup(ResId, [metrics]).
```
lafirest, line 314: Personally, I prefer add a new function for API call, the `emqx_resource:lookup` not only used by HTTP API

keynslug: In general, I agree with your consideration, but in this case I think changing the existing function is reasonable.
1. All users of this function are either API handler related code or tests. API related code should benefit from this function no longer being blocking, and tolerate the slight loss of linearizability (e.g. read the state of resource which is in the process of removal).
2. Aforementioned loss is practically non-existent since the cache now is kept consistent with the state of a resource.

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -229,14 +231,25 @@ set_resource_status_connecting(ResId) ->
...
 -spec ets_lookup(resource_id()) -> {ok, resource_group(), resource_data()} | {error, not_found}.
 ets_lookup(ResId) ->
+    ets_lookup(ResId, []).
+
+%% @doc Lookup the group and data of a resource from the cache
+-spec ets_lookup(resource_id(), [Option]) ->
+    {ok, resource_group(), resource_data()} | {error, not_found}
+when
+    Option :: metrics.
+ets_lookup(ResId, Options) ->
```
savonarola, line 248: Nit: this is the inherited naming, but maybe we could use something like `cache_lookup` instead of `ets_lookup` for external usage, not exposing the internals in naming.

apps/emqx_bridge/test/emqx_bridge_mqtt_SUITE.erl
```diff
@@ -827,13 +819,6 @@ t_mqtt_conn_bridge_egress_async_reconnect(_) ->
         }
     ),
 
-    on_exit(fun() ->
-        %% delete the bridge
-        {ok, 204, <<>>} = request(delete, uri(["bridges", BridgeIDEgress]), []),
-        {ok, 200, <<"[]">>} = request(get, uri(["bridges"]), []),
-        ok
-    end),
-
```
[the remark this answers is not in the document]
keynslug: There's already this thing in `end_per_testcase/2`.

https://github.com/emqx/emqx/blob/b9377f2f26e555c326d403d2136ebff9c3daf469/apps/emqx_bridge/test/emqx_bridge_mqtt_SUITE.erl#L167-L168

So those requests seemed redundant.

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -291,26 +304,30 @@ start_link(MgrId, ResId, Group, ResourceType, Config, Opts) ->
...
     end.
 
+terminate({shutdown, removed}, _State, _Data) ->
+    ok;
 terminate(_Reason, _State, Data) ->
-    _ = stop_resource(Data),
-    _ = maybe_clear_alarm(Data#data.id),
-    delete_cache(Data#data.id, Data#data.manager_id),
+    _ = maybe_stop_resource(Data),
+    ok = delete_cache(Data#data.id, Data#data.manager_id),
```
savonarola, line 330: We don't need `maybe_clear_alarm` anymore?

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -406,6 +416,22 @@ handle_event(EventType, EventData, State, Data) ->
     ),
     keep_state_and_data.
 
+verify_state_consistency(State, #data{status = State} = Data) ->
```
savonarola, line 419: Nit:  mb? `log_state_consistency`

apps/emqx_resource/src/emqx_resource_manager.erl
keynslug: It actually should have already been cleaned in https://github.com/emqx/emqx/blob/c6f67c104389916683035b30dd2d6960f11a63d2/apps/emqx_resource/src/emqx_resource_manager.erl#L547-L548 at this point.

savonarola:
Greate cache improvement 👍 
But it seems that there were many cases when the cache was out of sync 🤔  How could things work before the fixes? Or the cache was used only for informational purposes?

keynslug:
> Or the cache was used only for informational purposes?

Mostly this. It was used either as a fallback for the `lookup/1` or as a means to access resource configuration which is more or less constant during resource lifetime. It might have been critical to track `stopped` status in cache consistently [for the buffer worker](https://github.com/emqx/emqx/blob/c6f67c104389916683035b30dd2d6960f11a63d2/apps/emqx_resource/src/emqx_resource_buffer_worker.erl#L888-L889) to work correctly.

apps/emqx_resource/src/emqx_resource_manager.erl
keynslug: Done in [d44e683](https://github.com/emqx/emqx/pull/10123/commits/d44e683088b9264ed0531f446c398d57924034bb).

thalesmg (dismissed):

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -229,14 +231,25 @@ set_resource_status_connecting(ResId) ->
 -spec lookup(resource_id()) -> {ok, resource_group(), resource_data()} | {error, not_found}.
 lookup(ResId) ->
     case safe_call(ResId, lookup, ?T_LOOKUP) of
-        {error, timeout} -> ets_lookup(ResId);
+        {error, timeout} -> lookup_cached(ResId, [metrics]);
```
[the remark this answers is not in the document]
keynslug: Well, no, because the PR you mentioned made them disappear only on the API level, where responses are being formatted. This change only makes sure that `lookup/1` semantics become more consistent (e.g. _always return metrics info_ instead _sometimes return metrics info_). Ideally I think we'd want to stop asking the backend for metrics when we don't want them, but that'll require few more changes, and smells like a followup PR material.

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -483,10 +483,11 @@ schema("/bridges_probe") ->
             end
     end;
 '/bridges'(get, _Params) ->
+    Nodes = mria:running_nodes(),
+    NodeBridges = lists:zip(Nodes, emqx_bridge_proto_v3:list_bridges_on_nodes(Nodes)),
     {200,
         zip_bridges([
-            [format_resp(Data, Node) || Data <- emqx_bridge_proto_v1:list_bridges(Node)]
-         || Node <- mria:running_nodes()
+            [format_resource(Data, Node) || {Node, {ok, Bridges}} <- NodeBridges, Data <- Bridges]
```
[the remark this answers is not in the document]
keynslug: Yeah. The previous behavior was presumably to crash on a failed RPC, but I'm not sure if this was better. Anyway, made handling consistent with the `do_lookup_from_all_nodes/4` behavior.

apps/emqx_resource/test/emqx_connector_demo.erl
```diff
@@ -243,7 +243,7 @@ on_get_status(_InstId, #{pid := Pid}) ->
     end.
 
 spawn_counter_process(Name, Register) ->
-    Pid = spawn_link(?MODULE, counter_loop, []),
```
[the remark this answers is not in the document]
keynslug: In the happy case `on_stop` callback should take care of that.

https://github.com/emqx/emqx/blob/d44e683088b9264ed0531f446c398d57924034bb/apps/emqx_resource/test/emqx_connector_demo.erl#L77-L79

But in general, it might not be enough. Brought that back, instead tweaked the `on_stop/2` behaviour.

apps/emqx_resource/src/emqx_resource_manager.erl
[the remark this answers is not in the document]
thalesmg: I see.  I thought that at this point we wouldn't want metrics since they now have their own endpoint where they're collected explicitly.

thalesmg (approved):

apps/emqx_resource/src/emqx_resource_manager.erl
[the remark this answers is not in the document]
keynslug: That's right. I attempted to make a commit which optimizes that out, but:
1. It became quite involved.
2. Only then I noticed that #10114 haven't been merged yet. 🫠

So I guess I'll open a followup PR once it's merged.

PR #10126, opened 2023-03-13
fix(helm-charts): add home volume when container uses readonly filesystem

tom-tschiller:
When an EMQX container runs with readonly filesystem, it needs a writable home directory to save the Erlang cookie

### DIFF
--- a/deploy/charts/emqx-enterprise/templates/StatefulSet.yaml
+++ b/deploy/charts/emqx-enterprise/templates/StatefulSet.yaml
@@ -74,6 +74,10 @@ spec:
         secret:
           secretName: {{ .Values.emqxLicenseSecretName }}
       {{- end }}
+      {{- if and (.Values.containerSecurityContext.enabled) (.Values.containerSecurityContext.readOnlyRootFilesystem) }}
+      - name: emqx-home
+        emptyDir: {}
+      {{- end }}
       {{- if .Values.podSecurityContext.enabled }}
       securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
@@ -138,6 +142,10 @@ spec:
             subPath: "emqx.lic"
             readOnly: true
           {{- end }}
+          {{- if and (.Values.containerSecurityContext.enabled) (.Values.containerSecurityContext.readOnlyRootFilesystem) }}
+          - name: emqx-home
+            mountPath: "/home/emqx"
+          {{- end}}
           readinessProbe:
             httpGet:
               path: /status
--- a/deploy/charts/emqx/templates/StatefulSet.yaml
+++ b/deploy/charts/emqx/templates/StatefulSet.yaml
@@ -74,6 +74,10 @@ spec:
         secret:
           secretName: {{ .Values.emqxLicenseSecretName }}
       {{- end }}
+      {{- if and (.Values.containerSecurityContext.enabled) (.Values.containerSecurityContext.readOnlyRootFilesystem) }}
+      - name: emqx-home
+        emptyDir: {}
+      {{- end }}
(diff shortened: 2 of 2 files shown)


Rory-Z:
Hi, @tom-tschiller could you please tell me why need Eralng cookie file ? The EMQX have cookie configure in `etc/emqx.conf`

@zmstone @id I am not an expert in Eralng, do you have any idea ?

id:
@tom-tschiller could you describe the circumstances when you have observed the issue you are trying to fix here?

tom-tschiller:
Hi @id and @Rory-Z. I checked it again. Starting a cluster with `readOnlyRootFilesystem: true` works in general. I had this problem, when I executed `emqx_ctl cluster status` on the command line. This results in an error:

```
emqx@mqtt-broker-test-helm-0:/opt/emqx$ emqx_ctl cluster status
escript: exception error: no match of right hand side value 
                 {error,
                     {{shutdown,
                          {failed_to_start_child,auth,
                              {"Failed to create cookie file '/home/emqx/.erlang.cookie': erofs",
                               [{auth,init_no_setcookie,0,
                                    [{file,"auth.erl"},{line,313}]},
                                {auth,init,1,[{file,"auth.erl"},{line,165}]},
                                {gen_server,init_it,2,
                                    [{file,"gen_server.erl"},{line,423}]},
                                {gen_server,init_it,6,
                                    [{file,"gen_server.erl"},{line,390}]},
                                {proc_lib,init_p_do_apply,3,
                                    [{file,"proc_lib.erl"},{line,226}]}]}}},
                      {child,undefined,net_sup_dynamic,
                          {erl_distribution,start_link,
                              [#{clean_halt => false,
                                 name =>
                                     'remsh_maint_emqx_test404@mqtt-broker-test-helm-0.mqtt-broker-test-helm-headless.mqtt-broker-test-helm.svc.cluster.local',
                                 name_domain => longnames,
                                 net_tickintensity => 4,net_ticktime => 60,
                                 supervisor => net_sup_dynamic}]},
                          permanent,false,1000,supervisor,
                          [erl_distribution]}}}
  in function  nodetool__escript__1678__868664__459130__3:'-do/1-fun-3-'/1 (/opt/emqx/bin/nodetool, line 40)
  in call from nodetool__escript__1678__868664__459130__3:do_with_ret/3 (/opt/emqx/bin/nodetool, line 169)
  in call from nodetool__escript__1678__868664__459130__3:do/1 (/opt/emqx/bin/nodetool, line 37)
  in call from escript:run/2 (escript.erl, line 750)
  in call from escript:start/1 (escript.erl, line 277)
  in call from init:start_em/1 
  in call from init:do_boot/3 
ERROR: node_is_not_running!
```
The cluster runs fine as far as I can see from the dashboard. Only the CLI tool seems to fail. As you say it shouldn't save the Erlang cookie in this directory. So do you think it's a problem in the CLI tool and should be fixed there?

id:
That's right, this looks like an issue with the cli tool. We'll look into it, thanks!

Rory-Z:
> That's right, this looks like an issue with the cli tool. We'll look into it, thanks!

Maybe the EMQX 4 has same issue

tom-tschiller:
Thanks for your replys! So I'll close this pull request and create an issue for this, it that's ok for you

id:
@tom-tschiller quick workaround is to create a file with some short alphanumeric string and mount it inside container as `/home/emqx/.erlang.cookie`. Then the cli should work.

id:
@Rory-Z most likely, yes.

id:
@Rory-Z yes, we override it in `nodetool` from command line argument.

Rory-Z:
> @Rory-Z yes, we override it in `nodetool` from command line argument.

I see, thx

PR #10128, opened 2023-03-13
feat: add ocsp stapling support to mqtt ssl listener (5.0) 

thalesmg:
https://emqx.atlassian.net/browse/EMQX-8816

Spin off of https://github.com/emqx/emqx/pull/10067

Since there were some fundamental changes requested to the CRL part of that PR, I've split the OCSP stapling bits for separate review to uncouple the merge.


changes/ce/feat-10067.zh.md
```diff
@@ -0,0 +1 @@
+为SSL MQTT监听器增加对OCSP订书和CRL检查的支持。
```
qzhuyan, line 1: 为 SSL MQTT 监听器增加对 OCSP Stapling 和 CRL 检查的支持。

apps/emqx/i18n/emqx_schema_i18n.conf
```diff
@@ -1753,6 +1753,63 @@ server_ssl_opts_schema_gc_after_handshake {
...
+}
+
+server_ssl_opts_schema_ocsp_refresh_http_timeout {
+    desc {
+        en: "The timeout for the HTTP request when checking OCSP responses."
+        zh: "检查OCSP响应时，HTTP请求的超时。"
+    }
+    label: {
+        en: "OCSP Refresh HTTP Timeout"
+        zh: "OCSP刷新HTTP超时"
```
qzhuyan, line 1809: same here

qzhuyan:
would be good if we could have PR template?

changes/ce/feat-10128.zh.md
```diff
@@ -0,0 +1 @@
+为 SSL MQTT 监听器增加对 OCSP Stapling 和 CRL 检查的支持。
```
zmstone, line 1:
```suggestion
为 SSL MQTT 监听器增加对 OCSP Stapling 的支持。
```

qzhuyan:
> > would be good if we could have PR template?
> 
> Ok, I'll get it and put it in the description.
> 
> I has been some time since the templates don't show up automatically anymore. 🤔
> 
> Maybe we could reconfigure them to be automatic again?

I couldn't find a way to let GitHub support templates for different target branches.
We need to manually append the following while creating the PR

`?template=v5.md`

id:
> I couldn't find a way to let GitHub support templates for different target branches.

Why do we need different templates for different target branches? What is the use case?

Right now we have
* ci.md
* doc.md
* v4.md
* v5.md

We are not going to keep v4 and v5 in sync, may be we can just have one standard template here?

apps/emqx/src/emqx_ocsp_cache.erl
```diff
@@ -0,0 +1,522 @@
...
+handle_call({http_fetch, ListenerID}, _From, State) ->
+    case do_lookup(ListenerID) of
+        {ok, DERResponse} ->
+            {reply, {ok, DERResponse}, State};
+        {error, invalid_listener_id} ->
+            {reply, error, State};
+        {error, not_cached} ->
+            Conf = undefined,
+            with_refresh_params(ListenerID, Conf, {reply, error, State}, fun(Params) ->
+                case do_http_fetch_and_cache(ListenerID, Params) of
```
qzhuyan, line 170: I think it is better to have a `try .... catch` here or in `with_refresh_params`, The code path handles external inputs.

qzhuyan (dismissed):

qzhuyan:
> > I couldn't find a way to let GitHub support templates for different target branches.
> 
> Why do we need different templates for different target branches? What is the use case?
> 
> Right now we have
> 
> * ci.md
> * doc.md
> * v4.md
> * v5.md
> 
> We are not going to keep v4 and v5 in sync, may be we can just have one standard template here?

yeah, I agree to merge them into one and manually strip the parts that are irrelevant.

apps/emqx/src/emqx_ocsp_cache.erl
```diff
@@ -0,0 +1,522 @@
...
+-define(CALL_TIMEOUT, 20_000).
+-define(RETRY_TIMEOUT, 5_000).
+-define(REFRESH_TIMER(LID), {refresh_timer, LID}).
+-ifdef(TEST).
+-define(MIN_REFRESH_INTERVAL, timer:seconds(5)).
+-else.
+-define(MIN_REFRESH_INTERVAL, timer:minutes(1)).
+-endif.
+
+-define(WITH_LISTENER_CONFIG(ListenerID, ConfPath, Pattern, ErrorResp, Action),
```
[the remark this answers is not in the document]
thalesmg: I was using a macro because `Pattern` is pattern to match on when fetching the config; if the pattern fails, then `ErrorResp` is returned.

Maybe we could do the same with a normal function and then catch a `function_clause` error.  Do you think that would be better?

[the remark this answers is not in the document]
thalesmg: Actually, since it's used in only 2 places now, it's not so bad to repeat.  I'll change it to a simple function, it's indeed better.  Thanks!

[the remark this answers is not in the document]
qzhuyan: although the code looks cleaner, It is really hard to trace or debug macros in erlang, I am just asking if it is possible to avoid that.
If you think macro helps in your case, just try to make it smaller.

qzhuyan (approved):

apps/emqx/src/emqx_schema.erl
```diff
@@ -1294,6 +1298,49 @@ fields("listener_quic_ssl_opts") ->
     );
 fields("ssl_client_opts") ->
     client_ssl_opts_schema(#{});
+fields("ocsp") ->
+    [
+        {"enable_ocsp_stapling",
+            sc(
+                boolean(),
```
zmstone, line 1305: there had been some discussion about abusing `enable` flag in many places.
is it really necessary for ocsp ?
e.g. if the parent field is set to `null` (`undefined` from HOCON map), doesn't that serve as disable ?

thalesmg: This struct is always present because of the fields with default values, right?

We _could_ interpret the absence of `responder_url` as a sign that it's disabled, but in my opinion being explicit with an `enable` flag is more clear than it being implicitly disabled by the absence of a certain field.  :thinking: 

If we want to avoid using explicit configs, then we can use the absence of that field to disable it, yes.

PR #10129, opened 2023-03-13
chore(ci): update mongodb evironment

savonarola:
Update mongodb to the latest version (6).
Also, test authn/authz against the previous version of the protocol (<5.1)


savonarola:
Too outdated chore

PR #10136, opened 2023-03-14
fix: schema default value so that it shows up correctly in UI

kjellwinblad:
This complements PR https://github.com/emqx/emqx/pull/10124. The default values for duration_ms() fields needs to be formatted as a binary string with unit to show up correctly in the dashboard UI.

No changelog entry needed since this is provided with https://github.com/emqx/emqx/pull/10124

### DIFF
--- a/apps/emqx_connector/src/emqx_connector_mongo.erl
+++ b/apps/emqx_connector/src/emqx_connector_mongo.erl
@@ -106,7 +106,14 @@ fields(topology) ->
         {socket_timeout_ms, duration("socket_timeout")},
         {server_selection_timeout_ms, duration("server_selection_timeout")},
         {wait_queue_timeout_ms, duration("wait_queue_timeout")},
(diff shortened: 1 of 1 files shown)


apps/emqx_connector/src/emqx_connector_mongo.erl
```diff
@@ -407,11 +414,10 @@ duration(Desc) ->
         desc => ?DESC(Desc)
     }.
 
-heartbeat_frequency_ms(type) -> emqx_schema:duration_ms();
-heartbeat_frequency_ms(desc) -> ?DESC("heartbeat_period");
-heartbeat_frequency_ms(default) -> 200000;
-heartbeat_frequency_ms(validator) -> [?MIN(1)];
-heartbeat_frequency_ms(_) -> undefined.
+% heartbeat_frequency_ms(type) -> emqx_schema:duration_ms();
```
zmstone, line 417: please delete this.

zhongwencool (approved):

Issue #10116, opened 2023-03-10
Feat: Add extraVolumeMounts to Helm chart


idelsink:
### What would you like to be added or enhanced?

I would like to see an `extraVolumeMounts` option get added to the Helm chart. 

### Why is this needed?

Having the ability to mount files into the EMQX instance allows us to add, for example, ACL rule files as mentioned in https://github.com/emqx/emqx/issues/9052

Rory-Z:
Sorry for delay, looks good idea, we will finish it in next release

idelsink:
@Rory-Z great that this feature has been added so quickly! But as I also noted in your PR (which unfortunately already closed :sweat_smile: ) there might be an issue with the indentation. 

When I'm using it I get the error:

> Error: YAML parse error on emqx/templates/StatefulSet.yaml: error converting YAML to JSON: yaml: line 84: did not find expected key

Changing the indentations for both volume mount entries (to 6 and 10 respectively) solves the issue.

On a side note, I am using Terraform to call this helm chart using the helm_release resource, but that is not *really* relevant I think.

idelsink:
@Rory-Z I've created the necessary changes and created a PR: https://github.com/emqx/emqx/pull/10193 Perhaps you can take a look and update it to fit with the internal project's requirements?

PR #10139, opened 2023-03-15
feat(helm): helm chart support extraVolumeMounts and extraVolumes

Rory-Z:
Done of https://github.com/emqx/emqx/issues/10116

### DIFF
--- a/deploy/charts/emqx-enterprise/templates/StatefulSet.yaml
+++ b/deploy/charts/emqx-enterprise/templates/StatefulSet.yaml
@@ -74,12 +74,15 @@ spec:
         secret:
           secretName: {{ .Values.emqxLicenseSecretName }}
       {{- end }}
+      {{- if .Values.extraVolumes }}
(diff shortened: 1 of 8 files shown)


id:
@Rory-Z you only need to add changelog once: `ee` picks up everything from `ce` + adds `ee` changes on top.

id (dismissed):

id (approved):

deploy/charts/emqx-enterprise/templates/StatefulSet.yaml
```diff
@@ -138,6 +141,9 @@ spec:
             subPath: "emqx.lic"
             readOnly: true
           {{- end }}
+          {{- if .Values.extraVolumeMounts }}
+            {{- toYaml .Values.extraVolumeMounts | nindent 12 }}
```
idelsink, line 145: @Rory-Z I'm testing it out locally and I think the indentation should be 10 here, to be in line with the YAML indentation for the other volume mounts

PR #10143, opened 2023-03-15
feat(bridges): integrate RocketMQ into data bridges 

lafirest:


lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
```diff
@@ -0,0 +1,348 @@
...
+    ?SLOG(info, #{
+        msg => "starting_rocketmq_connector",
+        connector => InstanceId,
+        config => redact(Config1)
+    }),
+    Config = maps:merge(default_security_info(), Config1),
+    {Host, Port} = emqx_schema:parse_server(Server, ?ROCKETMQ_HOST_OPTIONS),
+
+    Server1 = [{Host, Port}],
+    ClientId = client_id(InstanceId),
```
savonarola, line 111: Could we avoid converting to an atom here?

lafirest: No, the RocketMQ driver needs an atom clientid

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
```diff
@@ -0,0 +1,348 @@
...
+            ),
+            Error
+    end.
+
+on_stop(InstanceId, #{client_id := ClientId, producers_map_pid := Pid} = _State) ->
+    ?SLOG(info, #{
+        msg => "stopping_rocketmq_connector",
+        connector => InstanceId
+    }),
+    Pid ! ok,
```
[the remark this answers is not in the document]
lafirest: I agree with this, and I had tried it. But We can't update the resource state when doing a query, so I wrote this in the light of the v4.

changes/ee/feat-10143.en.md
```diff
@@ -0,0 +1 @@
+Add `RocketMQ` support for Data-Bridg.
```
HJianBo, line 1:
```suggestion
Add `RocketMQ` support for Data-Bridge.
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_rocketmq.conf
```diff
@@ -0,0 +1,72 @@
...
+        }
+        label {
+                en: "Local Topic"
+                zh: "本地 Topic"
+            }
+    }
+
+    template {
+        desc {
+            en: """Template, the default value is empty. When this value is empty the whole message will be stored in the database"""
```
HJianBo, line 21:
```suggestion
            en: """Template, the default value is empty. When this value is empty the whole message will be stored in the RocketMQ"""
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_rocketmq.conf
```diff
@@ -0,0 +1,72 @@
...
+            zh: """模板, 默认为空，为空时将会将整个消息存入数据库"""
```
HJianBo, line 22:
```suggestion
            zh: """模板, 默认为空，为空时将会将整个消息写入 RocketMQ"""
```

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
savonarola: 👍 
Maybe we could fix that later, e.g. using `gproc` instead of local process registering https://github.com/emqx/rocketmq-client-erl/blob/master/src/rocketmq_client.erl#L51

.ci/docker-compose-file/docker-compose-rocketmq.yaml
```diff
@@ -0,0 +1,34 @@
...
+    command: ./mqnamesrv 
+    networks:
+      - emqx_bridge
+
+  mqbroker:
+    image: apache/rocketmq:4.9.4 
+    container_name: rocketmq_broker
+    ports:
+      - 10909:10909
+      - 10911:10911
```
thalesmg, line 21: I suggest we keep the `ports` option commented out, so that when we run the containers locally they don't occupy those ports by default, but we still have the option to easily uncomment those when developing.

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_rocketmq.erl
```diff
@@ -0,0 +1,119 @@
...
+        {enable, mk(boolean(), #{desc => ?DESC("config_enable"), default => true})},
+        {template,
+            mk(
+                binary(),
+                #{desc => ?DESC("template"), default => ?DEFAULT_TEMPLATE}
+            )},
+        {local_topic,
+            mk(
+                binary(),
+                #{desc => ?DESC("local_topic"), default => undefined}
```
thalesmg, line 81:
```suggestion
                #{desc => ?DESC("local_topic"), required => false}
```
?

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_rocketmq.erl
```diff
@@ -0,0 +1,119 @@
...
+            )},
+        {resource_opts,
+            mk(
+                ref(?MODULE, "creation_opts"),
+                #{
+                    required => false,
+                    default => #{},
```
thalesmg, line 88: I think we should either define `required` or `default`:

```suggestion
                    required => false,
```

lib-ee/emqx_ee_bridge/i18n/emqx_ee_bridge_rocketmq.conf
```diff
@@ -0,0 +1,72 @@
...
+        }
+        label {
+            en: "Enable Or Disable Bridge"
+            zh: "启用/禁用桥接"
+        }
+        }
+
+    desc_config {
+        desc {
+            en: """Configuration for an RocketMQ bridge."""
```
thalesmg, line 42:
```suggestion
            en: """Configuration for a RocketMQ bridge."""
```

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_rocketmq_SUITE.erl
```diff
@@ -0,0 +1,292 @@
...
+                    {rocketmq_config, RocketMQConf},
+                    {rocketmq_bridge_type, BridgeType},
+                    {rocketmq_name, Name},
+                    {proxy_host, ProxyHost},
+                    {proxy_port, ProxyPort}
+                    | Config0
+                ],
+            Config;
+        false ->
+            {skip, no_rocketmq}
```
thalesmg, line 115: We need to explode in CI to avoid silent failures:

```suggestion
            case os:getenv("IS_CI") of
                false ->
                    {skip, no_rocketmq};
                _ ->
                    throw(no_rocketmq)
            end
```

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
```diff
@@ -0,0 +1,348 @@
...
+    Server1 = [{Host, Port}],
+    ClientId = client_id(InstanceId),
+    ClientCfg = #{acl_info => #{}},
+
+    TopicTks = emqx_plugin_libs_rule:preproc_tmpl(Topic),
+    ProducerOpts = make_producer_opts(InstanceId, Config),
+    Templates = parse_template(Config),
+    ProducersMapPID = create_producers_map(ClientId),
+    State = #{
+        poolname => InstanceId,
```
thalesmg, line 119:
```suggestion
        instnace_id => InstanceId,
```

Also, it seems to be unused?

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
thalesmg, line 144: Maybe we should monitor the pid and ensure it's really dead?

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
```diff
@@ -0,0 +1,348 @@
...
+        security_token := SecurityToken,
+        send_buffer := SendBuff,
+        refresh_interval := RefreshInterval
+    }
+) ->
+    ACLInfo = acl_info(Username, Password, SecurityToken),
+    #{
+        tcp_opts => [{sndbuf, SendBuff}],
+        ref_topic_route_interval => RefreshInterval,
+        name => to_atom(InstanceId),
```
thalesmg, line 297: Do we really need this to be an atom?

The instance ID will change each time a bridge is (re)created or even if the bridge is simply health probed, so this could lead to an atom leak.

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
```diff
@@ -0,0 +1,348 @@
...
+apply_template([{Key, _} | _] = Reqs, Templates) ->
+    case maps:get(Key, Templates, undefined) of
+        undefined ->
+            [emqx_json:encode(Msg) || {_, Msg} <- Reqs];
+        Template ->
+            [emqx_plugin_libs_rule:proc_tmpl(Template, Msg) || {_, Msg} <- Reqs]
+    end.
+
+client_id(InstanceId) ->
+    erlang:list_to_atom("bridge_rocketmq:" ++ to_str(InstanceId)).
```
thalesmg, line 265: Note: when a bridge is recreated or health probed, this ID will change.

Perhaps we could tie it to the bridge _name_, so that it's the same between restarts, and also check if the bridge is being probed and use a fixed id for such case.

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
lafirest: No,  According to here, https://github.com/emqx/rocketmq-client-erl/blob/e1975007ca55d60e1368853c15869edf6bf81beb/src/rocketmq_producers_sup.erl#L67, the `name` isn't necessary

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
lafirest: I think this process is enough simple, monitor maybe make thing complicated.

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
thalesmg: Sorry, I meant to check if we need `name` _to be an atom_, not if the parameter is required.

If it's possible to keep it as a binary, it would avoid atom leakage.

lafirest: Yes, I know. The name must be an atom, this is a limit for the driver, but the name not is required, so I delete it from the options map

changes/ee/feat-10143.en.md
```diff
@@ -0,0 +1 @@
+Add `RocketMQ` support for Data-Bridge.
```
zmstone, line 1:
```suggestion
Add `RocketMQ` data integration bridge.
```

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge_rocketmq.erl
```diff
@@ -0,0 +1,120 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2022 EMQ Technologies Co., Ltd. All Rights Reserved.
```
thalesmg, line 2:
```suggestion
%% Copyright (c) 2023 EMQ Technologies Co., Ltd. All Rights Reserved.
```

lib-ee/emqx_ee_connector/src/emqx_ee_connector_rocketmq.erl
```diff
@@ -0,0 +1,338 @@
...
+apply_template([{Key, _} | _] = Reqs, Templates) ->
+    case maps:get(Key, Templates, undefined) of
+        undefined ->
+            [emqx_json:encode(Msg) || {_, Msg} <- Reqs];
+        Template ->
+            [emqx_plugin_libs_rule:proc_tmpl(Template, Msg) || {_, Msg} <- Reqs]
+    end.
+
+client_id(InstanceId) ->
+    Name = emqx_resource_manager:manager_id_to_resource_id(InstanceId),
```
thalesmg, line 264: Note: when the probe API is called on the bridge, there's an extra prefix added, so this might not work as intended.

Ref: https://github.com/lafirest/emqx/blob/17e207cb71dd5247a2718ff0db6c043b1d038b2c/apps/emqx_bridge/src/emqx_bridge_resource.erl#L228

thalesmg (approved):
Suggestion: test that the probe API works as intended with the RocketMQ clientid generation.

PR #10197, opened 2023-03-21
fix: add inflight window setting to the clickhouse bridge

kjellwinblad:
This PR makes sure the inflight window setting is present for the clickhouse bridge. It also changes emqx_resource_schema that previously removed the inflight window setting from resources with query mode `always_sync`. We don't need to do that because all bridges that uses the buffer worker queue will get async call handling even if the bridge don't support the async callback.

@thalesmg: maybe this should be merged after async support for bridges with `always_sync` has been added to the `emqx_resource_buffer_worker` module?

### DIFF
--- a/apps/emqx_resource/i18n/emqx_resource_schema_i18n.conf
+++ b/apps/emqx_resource/i18n/emqx_resource_schema_i18n.conf
@@ -135,14 +135,14 @@ When disabled the messages are buffered in RAM only."""
     }
   }
 
-  async_inflight_window {
+  inflight_window {
     desc {
-      en: """Async query inflight window."""
-      zh: """异步请求飞行队列窗口大小。"""
+      en: """Query inflight window. When query_mode is set to async, this config has to be et to 1 if messages from the same MQTT client have to be stricktly ordered."""
+      zh: """请求飞行队列窗口大小。当请求模式为异步时，如果需要严格保证来自同一 MQTT 客户端的消息有序，则必须将此值设为 1。"""""
     }
     label {
-      en: """Async inflight window"""
-      zh: """异步请求飞行队列窗口"""
+      en: """Inflight window"""
+      zh: """请求飞行队列窗口"""
     }
   }
 
--- a/apps/emqx_resource/i18n/emqx_resource_schema_i18n.conf
+++ b/apps/emqx_resource/i18n/emqx_resource_schema_i18n.conf
@@ -137,7 +137,7 @@ When disabled the messages are buffered in RAM only."""
 
   inflight_window {
     desc {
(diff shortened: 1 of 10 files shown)


apps/emqx_resource/src/schema/emqx_resource_schema.erl
```diff
@@ -60,7 +59,7 @@ fields("creation_opts") ->
         {auto_restart_interval, fun auto_restart_interval/1},
         {query_mode, fun query_mode/1},
         {request_timeout, fun request_timeout/1},
-        {async_inflight_window, fun async_inflight_window/1},
+        {inflight_window, fun inflight_window/1},
```
thalesmg, line 62: Just a reminder to update this point: https://github.com/kjellwinblad/emqx/blob/27b84453371fe99c3a8b0e2e0e0aaed274b1d31c/apps/emqx_resource/src/emqx_resource_buffer_worker.erl#L196

thalesmg:
> maybe this should be merged after async support for bridges with always_sync has been added to the emqx_resource_buffer_worker module?

I have yet to make the changes for that, and I'll have to touch all bridges that do use that `"creation_opts_sync_only"` type.  I essentially will remove this type, so this should be merged before that, actually.  :smile_cat:

apps/emqx_resource/src/schema/emqx_resource_schema.erl
kjellwinblad: Not sure if I understand what you mean by "Just a reminder" but I have added a commit to this PR that renames all occurrences of `async_inflight_window` to `inflight_window`.

thalesmg: Sorry, I think I was seeing just one of the commits.

Maybe it's worth adding an entry to keep track of when this field was renamed, yes.

apps/emqx_resource/i18n/emqx_resource_schema_i18n.conf
```diff
@@ -135,14 +135,14 @@ When disabled the messages are buffered in RAM only."""
     }
   }
 
-  async_inflight_window {
+  inflight_window {
```
thalesmg, line 138: Note: I noticed the frontend seems to have some i18n config paths they expect to find translations in.

So maybe just confirm that the frontend items that refer to the inflight window when creating a bridge still show the label and tooltip correctly.

kjellwinblad: I have checked the dashboard UI for clickhouse and influxdb. The clickhouse bridge have a field labeled "Inflight window" that is always present. The influxdb bridge still have a field called "Async inflight window" that is only present when async mode is activated. I have yet to investigate why that is but at least it looks like the inflight labels are present in the dashboard.

thalesmg (approved):

PR #10201, opened 2023-03-22
fix(tdengine): remove the redundant table name in SQL template

lafirest:
remove the redundant table name in SQL template

Fixes [EMQX-9226 ](https://emqx.atlassian.net/browse/EMQX-9226)

## PR Checklist


lib-ee/emqx_ee_connector/src/emqx_ee_connector_tdengine.erl
```diff
@@ -49,6 +51,8 @@ add_default_username(Fields) ->
         fun
             ({username, OrigUsernameFn}) ->
                 {username, add_default_fn(OrigUsernameFn, <<"root">>)};
+            ({password, OrigUsernameFn}) ->
+                {password, add_default_fn(OrigUsernameFn, ?DEFAULT_PASSWORD)};
```
keynslug, line 55: Concept of _default password_ sounds somewhat strange to me, honestly. I think that if there's a passwordless auth, then connector should anticipate password being not defined. If there is no passwordless auth, then we should just mark the field required?

lafirest: Usually, The Tdengine use a default password, "data", users don't need to care about this password,  and the schema of `password` is not required, of course, we can change it to true, but in this way, users have to repeat the password which should have omitted.
So we should not force users to input this password again.

keynslug: I see, but I wouldn't really expect users to deploy TDengine DB with a default password in the wild.

thalesmg: https://github.com/emqx/emqx/pull/10269

changes/ee/fix-10201.en.md
```diff
@@ -0,0 +1 @@
+In TDengine, Remove the redundant table name from the SQL template.
```
keynslug, line 1: Isn't that _database name_, according to https://docs.tdengine.com/taos-sql/table/#create-table ?

lafirest: Yes, it is `database` not `table`.

keynslug (approved):

changes/ee/fix-10201.en.md
```diff
@@ -0,0 +1 @@
+In TDengine, Remove the redundant database name from the SQL template.
```
thalesmg, line 1:
```suggestion
In TDengine, removed the redundant database name from the SQL template.
```

JimMoen (approved):

HJianBo (approved):

PR #10208, opened 2023-03-22
test(dynamo): attempt to fix flaky dynamo tests

thalesmg:
Those tests in the `flaky` test are really flaky and require lots of CI retries.

Apparently, the flakiness comes from race conditions from restarting bridges with the same name too fast between test cases.  Previously, all test cases were sharing the same bridge name (the module name).

Fixes https://emqx.atlassian.net/browse/EMQX-9287

## PR Checklist

### DIFF
--- a/lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_dynamo_SUITE.erl
+++ b/lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_dynamo_SUITE.erl
@@ -83,9 +83,10 @@ end_per_suite(_Config) ->
     ok = emqx_common_test_helpers:stop_apps([emqx_bridge, emqx_conf]),
     ok.
 
-init_per_testcase(_Testcase, Config) ->
+init_per_testcase(TestCase, Config) ->
     create_table(Config),
-    Config.
+    ok = snabbkaffe:start_trace(),
+    [{dynamo_name, atom_to_binary(TestCase)} | Config].
 
 end_per_testcase(_Testcase, Config) ->
     ProxyHost = ?config(proxy_host, Config),
@@ -93,7 +94,7 @@ end_per_testcase(_Testcase, Config) ->
     emqx_common_test_helpers:reset_proxy(ProxyHost, ProxyPort),
     ok = snabbkaffe:stop(),
     delete_table(Config),
-    delete_bridge(Config),
+    delete_all_bridges(),
     ok.
 
 %%------------------------------------------------------------------------------
@@ -186,15 +187,22 @@ parse_and_check(ConfigString, BridgeType, Name) ->
     Config.
 
 create_bridge(Config) ->
-    BridgeType = ?config(dynamo_bridge_type, Config),
-    Name = ?config(dynamo_name, Config),
-    TDConfig = ?config(dynamo_config, Config),
-    emqx_bridge:create(BridgeType, Name, TDConfig).
+    create_bridge(Config, _Overrides = #{}).
 
-delete_bridge(Config) ->
+create_bridge(Config, Overrides) ->
     BridgeType = ?config(dynamo_bridge_type, Config),
     Name = ?config(dynamo_name, Config),
-    emqx_bridge:remove(BridgeType, Name).
+    DynamoConfig0 = ?config(dynamo_config, Config),
+    DynamoConfig = emqx_map_lib:deep_merge(DynamoConfig0, Overrides),
+    emqx_bridge:create(BridgeType, Name, DynamoConfig).
+
+delete_all_bridges() ->
+    lists:foreach(
+        fun(#{name := Name, type := Type}) ->
+            emqx_bridge:remove(Type, Name)
+        end,
+        emqx_bridge:list()
+    ).
 
 create_bridge_http(Params) ->
     Path = emqx_mgmt_api_test_util:api_path(["bridges"]),
@@ -327,10 +335,12 @@ t_setup_via_http_api_and_publish(Config) ->
     ok.
 
 t_get_status(Config) ->
-    ?assertMatch(
-        {ok, _},
-        create_bridge(Config)
-    ),
+    {{ok, _}, {ok, _}} =
+        ?wait_async_action(
+            create_bridge(Config),
+            #{?snk_kind := resource_connected_enter},
+            20_000
+        ),
 
     ProxyPort = ?config(proxy_port, Config),
     ProxyHost = ?config(proxy_host, Config),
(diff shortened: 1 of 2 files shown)


keynslug:
I wonder if [giving `self()` as client pid to the pool](https://github.com/emqx/emqx/blob/2ffe7e43b55c7a31ce4adc36a0c5036263afefdd/lib-ee/emqx_ee_connector/src/emqx_ee_connector_dynamo.erl#L263-L265) have something to do with flakiness in tests. It may subtly subvert pool's expectations given that `ecpool_worker` clearly expects some fresh pid, not `self()`.

thalesmg:
> I wonder if [giving `self()` as client pid to the pool](https://github.com/emqx/emqx/blob/2ffe7e43b55c7a31ce4adc36a0c5036263afefdd/lib-ee/emqx_ee_connector/src/emqx_ee_connector_dynamo.erl#L263-L265) have something to do with flakiness in tests. It may subtly subvert pool's expectations given that `ecpool_worker` clearly expects some fresh pid, not `self()`.

I'll have to check how it uses this pid. 

But, at least running locally, stability seemed to improve greatly by simply using different bridge names.  This makes me suspect it might be related to the pool name (which is the bridge id) being reused to stop/start new bridges too fast between the tests.  :thinking:

keynslug:
> stability seemed to improve greatly by simply using different bridge names

Yeah, it's perfectly fine solution.

My concern is that some bug is still there, just masked out with the help of different bridge names. Well, given the assumption that resource code is correct (i.e. it waits for the resource startup / shutdown before giving control back to the caller), but that assumption holds so far (for example `emqx_bridge_mqtt_SUITE`'s testcases often use same bridge names).

zmstone (approved):

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_dynamo_SUITE.erl
```diff
@@ -186,15 +187,22 @@ parse_and_check(ConfigString, BridgeType, Name) ->
...
 
 create_bridge(Config) ->
-    BridgeType = ?config(dynamo_bridge_type, Config),
-    Name = ?config(dynamo_name, Config),
-    TDConfig = ?config(dynamo_config, Config),
-    emqx_bridge:create(BridgeType, Name, TDConfig).
+    create_bridge(Config, _Overrides = #{}).
 
-delete_bridge(Config) ->
+create_bridge(Config, Overrides) ->
```
keynslug, line 192: Nit. Is there a code that uses `create_bridge/2`?

keynslug (approved):

thalesmg:
> > stability seemed to improve greatly by simply using different bridge names
> 
> Yeah, it's perfectly fine solution.
> 
> My concern is that some bug is still there, just masked out with the help of different bridge names. Well, given the assumption that resource code is correct (i.e. it waits for the resource startup / shutdown before giving control back to the caller), but that assumption holds so far (for example `emqx_bridge_mqtt_SUITE`'s testcases often use same bridge names).

Good point.

I'll try to investigate it (or maybe @lafirest could have a look as well), but I think we could merge this one to at least decrease the high number of failures in CI from this suite.

lib-ee/emqx_ee_bridge/test/emqx_ee_bridge_dynamo_SUITE.erl
thalesmg: During my attempts, I tried to customize a few settings, ended up not using this one yet :see_no_evil:

thalesmg:
Example of the strange post config update failure; after a successful test, just before a failed one: 

```
*** System report during emqx_ee_bridge_dynamo_SUITE:t_get_status/1 in flaky 2023-03-22 18:56:01.785 ***
[🔗](file:///home/thales/dev/emqx/emqx5-ubuntu/_build/test/logs/last/lib.emqx_ee_bridge.emqx_ee_bridge_dynamo_SUITE.groups.logs/run.2023-03-22_18.55.47/emqx_ee_bridge_dynamo_suite.t_get_status.html#e-546)

=WARNING REPORT==== 22-Mar-2023::18:56:01.785207 ===
    entrypoint: <<"emqx:remove_config/2">>
    kind: initiate
    msg: cluster_rpc_apply_result
    result: {error,{post_config_update,emqx_bridge,timeout}}
    tnx_id: 2
```

keynslug:
> Example of the strange post config update failure

Oh, good find! Came up with a theory why this happens.

1. Testcase code removes the bridge.
2. Resource manager calls `emqx_ee_connector_dynamo:on_stop/2`, which then calls `ecpool_sup:stop_pool/1`.
3. Pool supervisor orders its `ecpool_worker_sup` to terminate, which in turn orders pool workers to terminate, sequentially, one-by-one.
4. Each worker, not aware that it manages itself (because of [this](https://github.com/emqx/emqx/blob/2ffe7e43b55c7a31ce4adc36a0c5036263afefdd/lib-ee/emqx_ee_connector/src/emqx_ee_connector_dynamo.erl#L263-L265)), sends itself [an exit signal](https://github.com/emqx/ecpool/blob/d01b8cb99af90bc177eeeabe29075133db878fb3/src/ecpool_worker.erl#L316-L317).
5. Since worker process traps exits, this eventually results in timeout and [a suicide](https://github.com/emqx/ecpool/blob/d01b8cb99af90bc177eeeabe29075133db878fb3/src/ecpool_worker.erl#L323-L324).
6. Timeout is fixed at [300 ms](https://github.com/emqx/ecpool/blob/d01b8cb99af90bc177eeeabe29075133db878fb3/src/ecpool_worker.erl#L215), but sequential nature of the process of supervisor termination might result in this process taking long (e.g. almost 5 seconds with 16 workers).

Why this causes test flaps though isn't clear.

thalesmg:
> > Example of the strange post config update failure
> 
> Oh, good find! Came up with a theory why this happens.
> 
>     1. Testcase code removes the bridge.
> 
>     2. Resource manager calls `emqx_ee_connector_dynamo:on_stop/2`, which then calls `ecpool_sup:stop_pool/1`.
> 
>     3. Pool supervisor orders its `ecpool_worker_sup` to terminate, which in turn orders pool workers to terminate, sequentially, one-by-one.
> 
>     4. Each worker, not aware that it manages itself (because of [this](https://github.com/emqx/emqx/blob/2ffe7e43b55c7a31ce4adc36a0c5036263afefdd/lib-ee/emqx_ee_connector/src/emqx_ee_connector_dynamo.erl#L263-L265)), sends itself [an exit signal](https://github.com/emqx/ecpool/blob/d01b8cb99af90bc177eeeabe29075133db878fb3/src/ecpool_worker.erl#L316-L317).
> 
>     5. Since worker process traps exits, this eventually results in timeout and [a suicide](https://github.com/emqx/ecpool/blob/d01b8cb99af90bc177eeeabe29075133db878fb3/src/ecpool_worker.erl#L323-L324).
> 
>     6. Timeout is fixed at [300 ms](https://github.com/emqx/ecpool/blob/d01b8cb99af90bc177eeeabe29075133db878fb3/src/ecpool_worker.erl#L215), but sequential nature of the process of supervisor termination might result in this process taking long (e.g. almost 5 seconds with 16 workers).
> 
> 
> Why this causes test flaps though isn't clear.

Excellent observations!  I'll look into those later points you mentioned, thank you!  :beers: 

I was trying to figure out by brute force here (sprinkling debug logs everywhere), and found out it has some correlation with the `resource_manager` process being stuck on 1 or more `health_check` calls (each with 5 s timeout), when a `remove` call arrives (also with 5 s timeout).  This times out, and the removal of the bridge from the config (and its teardown) fails, wrecking havoc in the suite.

I'll check your tips tomorrow to see if I can make further progress.

PR #10237, opened 2023-03-24
fix: create consistent interface 'with_node' for API access

sstrigler:
Fixes [EMQX-9138](https://emqx.atlassian.net/browse/EMQX-9138)

## PR Checklist

- [-] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up

### DIFF
--- a/apps/emqx/include/emqx_api_lib.hrl
+++ b/apps/emqx/include/emqx_api_lib.hrl
@@ -1,0 +1,36 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2021-2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+-ifndef(EMQX_API_LIB_HRL).
+-define(EMQX_API_LIB_HRL, true).
+
+-define(ERROR_MSG(CODE, REASON), #{code => CODE, message => emqx_misc:readable_error_msg(REASON)}).
+
+-define(OK(CONTENT), {200, CONTENT}).
+
+-define(NO_CONTENT, 204).
+
+-define(BAD_REQUEST(CODE, REASON), {400, ?ERROR_MSG(CODE, REASON)}).
+-define(BAD_REQUEST(REASON), ?BAD_REQUEST('BAD_REQUEST', REASON)).
+
+-define(NOT_FOUND(REASON), {404, ?ERROR_MSG('NOT_FOUND', REASON)}).
+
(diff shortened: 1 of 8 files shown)


apps/emqx/include/emqx_api_lib.hrl
```diff
@@ -0,0 +1,36 @@
...
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+-ifndef(EMQX_API_LIB_HRL).
+-define(EMQX_API_LIB_HRL, true).
+
+-define(ERROR_MSG(CODE, REASON), #{code => CODE, message => emqx_misc:readable_error_msg(REASON)}).
+
+-define(OK(CONTENT), {200, CONTENT}).
```
ieQu1, line 22: Nit:
```suggestion
-define(OK(CONTENT), {200, (CONTENT)}).
```

apps/emqx/include/emqx_api_lib.hrl
```diff
@@ -0,0 +1,36 @@
...
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+-ifndef(EMQX_API_LIB_HRL).
+-define(EMQX_API_LIB_HRL, true).
+
+-define(ERROR_MSG(CODE, REASON), #{code => CODE, message => emqx_misc:readable_error_msg(REASON)}).
```
savonarola, line 20: (Maybe for future discussion): I always wondered why we should have all the responses as macros. We do not match against them anywhere, and the performance impact should be neglectable 🤔

savonarola (dismissed):

apps/emqx/include/emqx_api_lib.hrl
sstrigler: I think it vastly improves readability and consistency, my goal is rather to ONLY use those macros throughout rather than constructing all those responses by hand.

Or is it, that you would favor function calls instead?

sstrigler: (Also I match against them in my tests)

savonarola: >that you would favor function calls instead?
Yes, I mean 
```
error_msg(Code, Reason) ->
...
```
vs
```
-define(ERROR_MSG(CODE, REASON), ...).
```

apps/emqx_management/src/emqx_mgmt_api_nodes.erl
```diff
@@ -250,55 +247,46 @@ nodes(get, _Params) ->
...
-node_error() ->
-    emqx_dashboard_swagger:error_codes([?SOURCE_ERROR], <<"Node error">>).
+to_ok_result({error, _} = Error) ->
+    Error;
+to_ok_result({ok, _} = Ok) ->
+    Ok;
+to_ok_result(Result) ->
+    {ok, Result}.
+
+to_ok_result_fun(Fun) when is_function(Fun) ->
```
lafirest, line 286: just a note to say, because of the hot-upgrade(I know, we haven't supported this feature) we rarely use an anonymous function, instead with the {M,F,A} tuple.

sstrigler: Taken as is, but I'll keep it in mind for a future refactoring. Good point.

lafirest (approved):

ieQu1 (approved):

PR #10251, opened 2023-03-27
fix(rule_engine): count referenced bridges in `from` clauses as dependencies (rv5.0)

thalesmg:
## Note: targetting `release-50`

Fixes https://emqx.atlassian.net/browse/EMQX-9325

Currently, ingress bridges referenced in the `FROM` clause of rules are not being accounted as dependencies.

When we try to delete an ingress bridge that's referenced in a rule like `select * from "$bridges/mqtt:ingress"`, that bridge does not trigger an UI warning about dependent actions.

Fixes <issue-or-jira-number>

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_engine.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_engine.erl
@@ -213,11 +213,12 @@ get_rules_with_same_event(Topic) ->
     ].
 
 -spec get_rule_ids_by_action(action_name()) -> [rule_id()].
-get_rule_ids_by_action(ActionName) when is_binary(ActionName) ->
+get_rule_ids_by_action(BridgeId) when is_binary(BridgeId) ->
     [
         Id
-     || #{actions := Acts, id := Id} <- get_rules(),
-        lists:any(fun(A) -> A =:= ActionName end, Acts)
+     || #{actions := Acts, id := Id, from := Froms} <- get_rules(),
+        forwards_to_bridge(Acts, BridgeId) orelse
+            references_ingress_bridge(Froms, BridgeId)
     ];
 get_rule_ids_by_action(#{function := FuncName}) when is_binary(FuncName) ->
     {Mod, Fun} =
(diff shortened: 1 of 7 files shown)


apps/emqx_rule_engine/src/emqx_rule_engine.erl
```diff
@@ -216,8 +216,14 @@ get_rules_with_same_event(Topic) ->
 get_rule_ids_by_action(ActionName) when is_binary(ActionName) ->
     [
         Id
-     || #{actions := Acts, id := Id} <- get_rules(),
-        lists:any(fun(A) -> A =:= ActionName end, Acts)
+     || #{actions := Acts, id := Id, from := From} <- get_rules(),
+        lists:any(fun(A) -> A =:= ActionName end, Acts) orelse
+            lists:any(
```
sstrigler, line 221: Think it would help to wrap this in a function with a name that explains what it does

apps/emqx_rule_engine/src/emqx_rule_engine.erl
```diff
@@ -216,8 +216,14 @@ get_rules_with_same_event(Topic) ->
...
+                fun(BridgeID) ->
+                    BridgeID =:= ActionName
+                end,
+                [BridgeID || <<"$bridges/", BridgeID/binary>> <- From]
```
zmstone, line 225: nit: there should be a function to parse resource ID to bridge ID ?

thalesmg: Do you mean bridge hookpoint to bridge id?  Because the targeted bridge id is the input to this function, so we can just compare for equality when searching for dependencies.

apps/emqx_rule_engine/src/emqx_rule_engine.erl
thalesmg: :heavy_check_mark:

apps/emqx_rule_engine/src/emqx_rule_engine.erl
```diff
@@ -317,8 +318,14 @@ get_basic_usage_info() ->
...
-                    BridgeIDs0 = [BridgeID || <<"$bridges/", BridgeID/binary>> <- From],
+                fun(#{actions := Actions, from := Froms}, Acc) ->
+                    BridgeIDs0 =
+                        [
+                            BridgeID
+                         || From <- Froms,
+                            {ok, BridgeID} <-
+                                [emqx_bridge_resource:bridge_hookpoint_to_bridge_id(From)]
+                        ],
                     BridgeIDs1 = lists:filter(fun is_binary/1, Actions),
```
zmstone, line 329: This fold and filter can be extracted to a private func for better readability.

zmstone (approved):

apps/emqx_rule_engine/src/emqx_rule_engine.erl
thalesmg: I'll make this refactoring in a follow up PR.

apps/emqx_rule_engine/src/emqx_rule_engine.erl
```diff
@@ -478,3 +485,19 @@ contains_actions(Actions, Mod0, Func0) ->
         end,
         Actions
     ).
+
+forwards_to_bridge(Actions, BridgeId) ->
+    lists:any(fun(A) -> A =:= BridgeId end, Actions).
```
zmstone, line 490: nit: lists:member/1

PR #10253, opened 2023-03-27
chore: improve UI error message when using bad conn parameters

thalesmg:
## Note: targeting `release-50`

Fixes https://emqx.atlassian.net/browse/EMQX-9350

Helps the user by hinting at possibly wrong connection parameters when clicking "Reconnect" for Kafka Consumer.

## PR Checklist

- [~] Added tests for the changes
- [~] Changed lines covered in coverage report
- [~] Change log has been added to `changes/{ce,ee}/(feat|perf|fix)-<PR-id>.en.md` and `.zh.md` files

- [~] If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)
- [~] Change log has been added to `changes/` dir for user-facing artifacts update

### DIFF
--- a/lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
+++ b/lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
@@ -145,6 +145,16 @@ on_start(InstanceId, Config) ->
                 instance_id => InstanceId,
                 kafka_hosts => BootstrapHosts
             });
+        {error, already_present = Reason} ->
+            ?SLOG(error, #{
+                msg => "failed_to_start_kafka_consumer_client",
+                instance_id => InstanceId,
+                kafka_hosts => BootstrapHosts,
+                reason => emqx_misc:redact(Reason)
+            }),
+            throw(
+                "Kafka client failed to start properly.  Please check the connection parameters."
+            );
         {error, Reason} ->
             ?SLOG(error, #{
                 msg => "failed_to_start_kafka_consumer_client",
--- a/lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
+++ b/lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
@@ -153,7 +153,7 @@ on_start(InstanceId, Config) ->
                 reason => emqx_misc:redact(Reason)
             }),
             throw(
-                "Kafka client failed to start properly.  Please check the connection parameters."
+                "Failed to start Kafka client. Please check the connection parameters."
             );
         {error, Reason} ->
             ?SLOG(error, #{
--- a/lib-ee/emqx_ee_bridge/test/emqx_bridge_impl_kafka_consumer_SUITE.erl
+++ b/lib-ee/emqx_ee_bridge/test/emqx_bridge_impl_kafka_consumer_SUITE.erl
@@ -1635,7 +1635,11 @@ t_bridge_rule_action_source(Config) ->
                 },
                 emqx_json:decode(RawPayload, [return_maps])
             ),
-            ?assertEqual(1, emqx_resource_metrics:received_get(ResourceId)),
+            ?retry(
+                _Sleep0 = 100,
+                _Attempts0 = 50,
+                ?assertEqual(1, emqx_resource_metrics:received_get(ResourceId))
+            ),
             ok
         end
     ),


lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -145,6 +145,16 @@ on_start(InstanceId, Config) ->
                 instance_id => InstanceId,
                 kafka_hosts => BootstrapHosts
             });
+        {error, already_present = Reason} ->
+            ?SLOG(error, #{
+                msg => "failed_to_start_kafka_consumer_client",
+                instance_id => InstanceId,
+                kafka_hosts => BootstrapHosts,
+                reason => emqx_misc:redact(Reason)
```
zmstone, line 153: nit: no need to redact ?

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -145,6 +145,16 @@ on_start(InstanceId, Config) ->
...
+            }),
+            throw(
+                "Kafka client failed to start properly.  Please check the connection parameters."
```
zmstone, line 156:
```suggestion
                "Failed to start Kafka client. Please check the connection parameters."
```

also, should this be moved to line 165?

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
thalesmg: :joy_cat: 

https://github.com/emqx/emqx/pull/9564#discussion_r1128570431

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
thalesmg: You mean write this message for any error?  I was trying to be specific here, because I don't know what other errors might be returned here.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
zmstone: the `Reason` here in this clause is obviously an atom.
The next clause however might contain sensitive info.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -145,6 +145,16 @@ on_start(InstanceId, Config) ->
...
             });
+        {error, already_present = Reason} ->
+            ?SLOG(error, #{
+                msg => "failed_to_start_kafka_consumer_client",
+                instance_id => InstanceId,
+                kafka_hosts => BootstrapHosts,
+                reason => emqx_misc:redact(Reason)
+            }),
+            throw(
+                "Failed to start Kafka client. Please check the connection parameters."
```
zmstone, line 156: are you sure this throw is added to the right clause ?
it seems to me this clause means the client is already running.?

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -145,6 +145,16 @@ on_start(InstanceId, Config) ->
...
                 kafka_hosts => BootstrapHosts
             });
+        {error, already_present = Reason} ->
+            ?SLOG(error, #{
+                msg => "failed_to_start_kafka_consumer_client",
+                instance_id => InstanceId,
+                kafka_hosts => BootstrapHosts,
+                reason => emqx_misc:redact(Reason)
+            }),
+            throw(
```
zmstone, line 155: do we really need to throw here? if the client is `already_present`,
we should probably stop it and restart (given the config might have changed)?
otherwise if config is not changed, we can just return ok ?

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
thalesmg: I tested the reported scenario: the wrong port is used, so the bridge is created, but the client never gets to connect.  In such state, the user clicks "Reconnect" in the UI, which [internally means](https://github.com/emqx/emqx/blob/9b35aeec6da31c7ab93e770ffa5334f7efb5a65d/apps/emqx_bridge/src/proto/emqx_bridge_proto_v3.erl#L97-L106) to start the bridge again.

In that situation, we get `already_started` because the brod client process indeed started, as we're starting the same bridge with the same name and thus kafka client id, but keeps crashing while trying to connect to Kafka.

thalesmg: I've added a couple screenshots in the description to illustrate the before and after errors messages.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
thalesmg: When the user clicks "Reconnect" in the UI, there's no config change, just `emqx_bridge_resource:start/2` is called without any changes to the existing config.

On the one hand, I guess we could indeed just return ok, the status will stay "Disconnected", and the user will have to investigate the logs. 

On the other hand, `already_present` means that the client is in the supervision tree but not running, so it might not be exactly a successful start. :thinking:

thalesmg: A similar issue was raised by QA for the Kafka Producer bridge.

At this point, the returned error offers no hint about what went wrong, but the user can inspect the logs and check that the password is wrong, or the connection cannot be established. 

How about we return a hint for the user to check the logs?  Like `"Failed to start Kafka client. Please inspect the EMQX logs and your connection parameters."`

thalesmg:
Will be superseded by #10273

PR #10273, opened 2023-03-29
fix: return friendly message when kafka producer and consumer fails to start (rv5.0)

thalesmg:
## Note: targeting `release-50`

Fixes https://emqx.atlassian.net/browse/EMQX-9392

The returned information does not allow to diagnose the issue (i.e.: a connection issue due to the wrong host and port, the wrong password failing authn).  However, such information is printed to the logs.

This changes the returned error to the API so that the user is hinted at looking at the logs for further investigation of the error.

#### Before the fix

image

#### After the fix

image

When probing:

image

image

image

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource.erl
+++ b/apps/emqx_resource/src/emqx_resource.erl
@@ -356,7 +356,14 @@ is_buffer_supported(Module) ->
 -spec call_start(manager_id(), module(), resource_config()) ->
     {ok, resource_state()} | {error, Reason :: term()}.
 call_start(MgrId, Mod, Config) ->
-    ?SAFE_CALL(Mod:on_start(MgrId, Config)).
+    try
+        Mod:on_start(MgrId, Config)
+    catch
+        throw:{error, Error} ->
+            {error, Error};
+        Kind:Error:Stacktrace ->
+            {error, {Kind, Error, Stacktrace}}
+    end.
 
 -spec call_health_check(manager_id(), module(), resource_state()) ->
     resource_status()
--- a/apps/emqx_resource/src/emqx_resource.erl
+++ b/apps/emqx_resource/src/emqx_resource.erl
@@ -359,10 +359,10 @@ call_start(MgrId, Mod, Config) ->
     try
         Mod:on_start(MgrId, Config)
     catch
-        throw:{error, Error} ->
+        throw:Error ->
             {error, Error};
         Kind:Error:Stacktrace ->
-            {error, {Kind, Error, Stacktrace}}
+            {error, #{exception => Kind, reason => Error, stacktrace => Stacktrace}}
     end.
 
 -spec call_health_check(manager_id(), module(), resource_state()) ->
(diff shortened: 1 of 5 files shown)


zmstone:
for throws, we can log the stacktrace but no need to expose to user ?

thalesmg:
> for throws, we can log the stacktrace but no need to expose to user ?

Good point, I'll adapt the point where the throws are collected for such cases.

thalesmg:
I've made a similar change for the Kafka consumer error message here.

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -95,6 +95,11 @@
     commit_fun => brod_group_subscriber_v2:commit_fun()
 }.
 
+-define(CLIENT_DOWN_MESSAGE,
+    "Failed to start Kafka client. Please check the logs for errors and check"
+    " the connection parameters"
```
zmstone, line 100:
```suggestion
-define(CLIENT_DOWN_MESSAGE,
    "Failed to start Kafka client. Please check the logs for errors and check"
    " the connection parameters."
```

apps/emqx_resource/src/emqx_resource.erl
```diff
@@ -356,7 +356,14 @@ is_buffer_supported(Module) ->
...
     {ok, resource_state()} | {error, Reason :: term()}.
 call_start(MgrId, Mod, Config) ->
-    ?SAFE_CALL(Mod:on_start(MgrId, Config)).
+    try
+        Mod:on_start(MgrId, Config)
+    catch
+        throw:{error, Error} ->
+            {error, Error};
+        Kind:Error:Stacktrace ->
+            {error, {Kind, Error, Stacktrace}}
```
zmstone, line 365:
```suggestion
            {error, #{exception => Kind, reason => Error, stacktrace => Stacktrace}}
```

apps/emqx_resource/src/emqx_resource.erl
```diff
@@ -356,7 +356,14 @@ is_buffer_supported(Module) ->
 -spec call_start(manager_id(), module(), resource_config()) ->
     {ok, resource_state()} | {error, Reason :: term()}.
 call_start(MgrId, Mod, Config) ->
-    ?SAFE_CALL(Mod:on_start(MgrId, Config)).
+    try
+        Mod:on_start(MgrId, Config)
+    catch
+        throw:{error, Error} ->
+            {error, Error};
```
zmstone, line 363:
```suggestion
        throw:Reason ->
            {error, Reason};
```

zmstone (approved):

lib-ee/emqx_ee_bridge/src/kafka/emqx_bridge_impl_kafka_consumer.erl
```diff
@@ -173,7 +178,7 @@ on_get_status(_InstanceID, State) ->
         kafka_client_id := ClientID,
         kafka_topics := KafkaTopics
     } = State,
-    do_get_status(ClientID, KafkaTopics, SubscriberId).
+    do_get_status(State, ClientID, KafkaTopics, SubscriberId).
```
savonarola, line 181: Nit: we pass `State` just to get it back unchanged.. 
Mb we could decorate the result with state after the call 🤔

thalesmg: I'll refactor those points in a follow up PR.

PR #10296, opened 2023-03-30
feat(ft-fs): introduce fs iterators concept + forward seeks

keynslug:
In order to support paging over filesystem contents, to serve REST APIs effectively.

---

[EMQX-9416](https://emqx.atlassian.net/browse/EMQX-9416).


apps/emqx_ft/src/emqx_ft_storage_exporter_fs.erl
```diff
@@ -265,30 +293,35 @@ start_reader(Options, RelFilepath, CallerPid) ->
 
 %%
 
--spec list(options()) ->
+-spec list(options(), query()) ->
     {ok, [exportinfo(), ...]} | {error, [{node(), _Reason}]}.
-list(_Options) ->
+list(_Options, Query) ->
```
savonarola, line 298: Seems that much complexity comes from using the same function both for iteration and individual transfer fetching. Also, having N nodes, we seem to make (N-1)x extra RPC calls (if I understand correctly) to perform the listing.

But we are not forced to do this; here are some considerations:

* To get files for a transfer, we **must** make a multicall because files can be on any node. But in this case, we may suffice a simple glob lookup of a concrete folder (is this true?).
* Without having to support transfer lookup with the listing, we are not forced to do a multicall then. So we may treat a node as a part of the cursor (like a top level of a path) and iterate over each node sequentially. If I am not mistaken, this should reduce the number of RPC calls N times.

apps/emqx_ft/src/emqx_ft_storage_exporter_fs.erl
```diff
@@ -265,30 +293,35 @@ start_reader(Options, RelFilepath, CallerPid) ->
...
     ),
-    length(Errors) > 0 andalso
-        ?SLOG(warning, #{msg => "list_remote_exports_failed", errors => Errors}),
-    case Results of
-        [_ | _] ->
-            {ok, Results};
+    length(Failures) > 0 andalso
+        ?SLOG(warning, #{msg => "list_remote_exports_failed", failures => Failures}),
+    Limit = maps:get(limit, Query, -1),
+    case emqx_ft_util:sublist_mergesorted(NodeExports, Limit) of
```
savonarola, line 320: It is unobious part, and maybe we need some explanations here. 
`NodeExports` are maps(?), so the order they are sorted does not match the order of folder traversal in the file iterator 🤔 If true, this will cause some transfers to be skipped during the listing 🤔

apps/emqx_ft/src/emqx_ft_api.erl
```diff
@@ -57,29 +59,131 @@ schema("/file_transfer/files") ->
...
+                    ['BAD_REQUEST'], <<"Invalid cursor">>
+                ),
                 503 => emqx_dashboard_swagger:error_codes(
-                    ['SERVICE_UNAVAILABLE'], <<"Service unavailable">>
+                    ['SERVICE_UNAVAILABLE'], error_desc('SERVICE_UNAVAILABLE')
+                )
+            }
+        }
+    };
+schema("/file_transfer/files/:clientid/:fileid") ->
```
savonarola, line 77: Nit: the API method design is driven by the edge case: that we have multiple files exported with the same fileid.
But currently, I have no ideas how to improve either 😄

apps/emqx_ft/src/emqx_ft_storage_exporter_fs.erl
keynslug: `NodeExports` should be [list](https://github.com/keynslug/emqx/blob/e4356e503da15565a0c5d76b9efa46c598ca02f9/apps/emqx_ft/src/emqx_ft_storage_exporter_fs.erl#L308) of [lists](https://github.com/keynslug/emqx/blob/e4356e503da15565a0c5d76b9efa46c598ca02f9/apps/emqx_ft/src/emqx_ft_storage_exporter_fs.erl#L235) now. Also the [iteration order](https://github.com/keynslug/emqx/blob/e4356e503da15565a0c5d76b9efa46c598ca02f9/apps/emqx_ft/src/emqx_ft_storage_exporter_fs.erl#L209) [should be kept](https://github.com/keynslug/emqx/blob/e4356e503da15565a0c5d76b9efa46c598ca02f9/apps/emqx_ft/src/emqx_ft_fs_iterator.erl#L72). Though this still needs testing.

apps/emqx_ft/src/emqx_ft_api.erl
keynslug: Heh. 🙃 Well, as long as it stands for _collection of files, scoped by client id / file id_ I think that should be fine (where REST is concerned).

apps/emqx_ft/src/emqx_ft_storage_exporter_fs.erl
keynslug: Reimplemented with sequential iteration over sorted list of cluster nodes. While we no longer need mergesort-related stuff, I'd say that overall it didn't become much simpler than it was before (when comparing against naive and wasteful implementation though). Also it made cursor opaque and its structure became more complex, which wasn't really needed before since currently both local filesystem and S3 support transparent cursors (which are also a bit more useful to an API user).

savonarola (approved):

PR #10263, opened 2023-03-28
feat: add new command for Elixir expression evaluation

paulozulato:
Currently, Elixir expressions are evaluated on an EMQX node using the 'eval' command, which works for both Erlang and Elixir expressions. This commit adds the new command 'eval-ex' exclusively for evaluating Elixir expressions on Elixir nodes, similar to 'eval-erl' for Erlang.

Fixes https://emqx.atlassian.net/browse/EMQX-8947

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [~] Added tests for the changes
- [~] Changed lines covered in coverage report
- [X] Change log has been added to `changes/{ce,ee}/(feat|perf|fix)-<PR-id>.en.md` and `.zh.md` files
- [X] For internal contributor: there is a jira ticket to track this change
- [~] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up
- [X] Schema changes are backward compatible

## Checklist for CI (.github/workflows) changes

- [~] If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)
- [~] Change log has been added to `changes/` dir for user-facing artifacts update

PR #10297, opened 2023-03-30
fix: keep eval command backward compatible with v4

paulozulato:
Keeps "eval" command evaluating only Erlang expressions, even on Elixir node.

Fixes https://emqx.atlassian.net/browse/EMQX-8947

Make sure to target release-50 branch if this PR is intended to fix the issues for the release candidate.

## PR Checklist

- [ ] ~Added tests for the changes~
- [ ] ~Changed lines covered in coverage report~
- [ ] ~If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up~
- [ ] ~Schema changes are backward compatible~

- [ ] ~If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)~
- [ ] ~Change log has been added to `changes/` dir for user-facing artifacts update~

### DIFF
--- a/bin/emqx
+++ b/bin/emqx
@@ -159,9 +159,6 @@ usage() {
         echo "Print EMQX installation root dir"
         ;;
     eval)
-        echo "Evaluate an Erlang or Elixir expression in the EMQX node"
-        ;;
-    eval-erl)
         echo "Evaluate an Erlang expression in the EMQX node, even on Elixir node"
         ;;
     eval-ex)
@@ -231,7 +228,7 @@ usage() {
         echo "  Install Info:     ertspath | root_dir"
         echo "  Runtime Status:   pid | ping"
         echo "  Validate Config:  check_config"
-        echo "  Advanced:         console_clean | escript | rpc | rpcterms | eval | eval-erl | eval-ex"
+        echo "  Advanced:         console_clean | escript | rpc | rpcterms | eval | eval-ex"
         echo ''
         echo "Execute '$REL_NAME COMMAND help' for more information"
     ;;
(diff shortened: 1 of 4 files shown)


bin/emqx
```diff
@@ -159,9 +159,6 @@ usage() {
         echo "Print EMQX installation root dir"
         ;;
     eval)
-        echo "Evaluate an Erlang or Elixir expression in the EMQX node"
-        ;;
-    eval-erl)
```
thalesmg, line 164: I think it's better to keep this option as-is, for backward compatibility as well.

paulozulato: It's to be backward compatible with v4. On that version:
```
    eval)
        echo "Evaluate an Erlang expression in the EMQX node"
        ;;
```
So we're keeping this behavior (as suggested by @zmstone [here](https://github.com/emqx/emqx/pull/10263#issuecomment-1489777099)).

thalesmg: I mean for `eval-erl`, it's possible someone is using that since it's released.

If it's fine to drop it since Elixir support is still kinda experimental, then just ignore me.  :smile_cat:

thalesmg (dismissed):

bin/emqx
```diff
@@ -159,9 +159,6 @@ usage() {
...
         echo "Evaluate an Erlang expression in the EMQX node, even on Elixir node"
```
zmstone, line 162:
```suggestion
        echo "Evaluate an Erlang expression in the EMQX node."
```

bin/emqx
zmstone: I think its fine to drop it.

PR #10300, opened 2023-03-30
fix(plugins): create directory before uploading

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-9434

Fixes https://github.com/emqx/emqx-elixir-plugin/issues/23

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
@@ -428,6 +428,7 @@ update_boot_order(post, #{bindings := #{name := Name}, body := Body}) ->
 %% For RPC upload_install/2
 install_package(FileName, Bin) ->
     File = filename:join(emqx_plugins:install_dir(), FileName),
+    ok = filelib:ensure_dir(File),
     ok = file:write_file(File, Bin),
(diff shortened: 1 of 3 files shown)


apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -428,6 +428,7 @@ update_boot_order(post, #{bindings := #{name := Name}, body := Body}) ->
 %% For RPC upload_install/2
 install_package(FileName, Bin) ->
     File = filename:join(emqx_plugins:install_dir(), FileName),
+    ok = filelib:ensure_dir(File),
```
lafirest, line 431: I don't think this would cover all situations like the users maybe use a device which not exist in all nodes, so I think an error catch & report also is necessary.

thalesmg: This would mean that the plugins directory does not exist and cannot be created by EMQX.  Since that directory path should be writable by EMQX, I guess we should raise an error here. :thinking: 

I've fixed the Mix file so that the empty directory is added to the tarball, so this should be a no-op, unless it's a pathological case as I've described above.

SergeTupchiy (approved):

savonarola (approved):

PR #10306, opened 2023-03-31
feat(bridges): enable async query mode for all bridges with buffer workers

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-9130

Since buffer workers always support async calls ("outer calls"), we should decouple those two call modes (inner and outer), and avoid exposing the inner call configuration to user to avoid complexity.

For bridges that currently only allow sync query modes, we should allow them to be configured with async.  That means basically all bridge types except Kafka Producer.

### <samp>🤖 Generated by Copilot at 066f73f</samp>

This pull request simplifies and unifies the schema definitions for various bridge resources by reusing the common `resource_opts` field from the `emqx_resource_schema` module and removing unused or redundant fields. It also makes the `query_mode` field more flexible and consistent across different resources, allowing both sync and async modes.

## PR Checklist

### DIFF
--- a/apps/emqx_connector/src/emqx_connector_mysql.erl
+++ b/apps/emqx_connector/src/emqx_connector_mysql.erl
@@ -172,10 +172,15 @@ on_query(
                     %% not return result, next loop will try again
                     on_query(InstId, {TypeOrKey, SQLOrKey, Params, Timeout}, State);
                 {error, Reason} ->
-                    LogMeta = #{connector => InstId, sql => SQLOrKey, state => State},
-                    ?SLOG(
+                    ?tp(
                         error,
-                        LogMeta#{msg => "mysql_connector_do_prepare_failed", reason => Reason}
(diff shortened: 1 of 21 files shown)


thalesmg:
- [ ] TODO:

Noticed an issue with rule action statistics when using async actions.  They always count as successes, because the rule engine doesn't wait for the reply.  The bridge/connector returns just `ok` when async (it's just a cast), and that's immediately interpreted as a success.

zmstone:
> > Noticed an issue with rule action statistics when using async actions. They always count as successes, because the rule engine doesn't wait for the reply. The bridge/connector returns just `ok` when async (it's just a cast), and that's immediately interpreted as a success.
> 
> Actually, this issue exists **prior** to the present changes. So we could merge this one as it is and create a follow up issue for fixing those metrics.

yeah, it's a known "issue", and some argue it's not an issue.

zmstone (approved):

PR #10309, opened 2023-04-01
add statefulset for kubernetes

Statemood:
This pull request add deploy with emqx 5.0 for kubernetes statefulset.

### DIFF
--- a/deploy/kubernetes/emqx-configmap.yaml
+++ b/deploy/kubernetes/emqx-configmap.yaml
@@ -1,0 +1,95 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: emqx-config
+  namespace: default
+data:
+  acl.conf: |-
+    {allow, {username, "^dashboard?"}, subscribe, ["$SYS/#"]}.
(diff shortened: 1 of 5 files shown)


Rory-Z:
Hi @Statemood thanks for your contribution.
But the emqx/emqx already have helm chart, so why we need this ?

Statemood:
Provide another option for users who do not use helm.

Statemood:
> Hi @Statemood thanks for your contribution. But the emqx/emqx already have helm chart, so why we need this ?

Provide another option for users who do not use helm.

Rory-Z:
> Provide another option for users who do not use helm.

I think we just support helm chart is good, because if use YAML file, the user must be very familiar with the file structure, compared to helm, this is much more complicated.

@id @zmstone Any ideas ?

id:
This does not have to be part of emqx repo. May be it could be a separate community project for starters? What do you think @Statemood?

id:
@Statemood sorry for long radio silence.

Could you add CI tests similar to https://github.com/emqx/emqx/blob/master/.github/workflows/run_fvt_tests.yaml#L110 so that it will be possible for us to maintain this feature?

PR #10314, opened 2023-04-03
0403 fix monitor current api

id:
Fixes https://emqx.atlassian.net/browse/EMQX-9436

## Summary
copilot:summary

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_stats.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_stats.erl
@@ -129,7 +129,19 @@ list(get, #{query_string := Qs}) ->
         _ ->
             Data = [
                 maps:from_list(emqx_mgmt:get_stats(Node) ++ [{node, Node}])
-             || Node <- mria:running_nodes()
+             || Node <- running_nodes()
             ],
             {200, Data}
     end.
+
+%%%==============================================================================================
+%% Internal
+
+running_nodes() ->
+    Nodes = erlang:nodes([visible, this]),
+    RpcResults = erpc:multicall(Nodes, emqx, is_running, [], 15000),
+    [
+        Node
+     || {Node, IsRunning} <- lists:zip(Nodes, RpcResults),
+        IsRunning =:= {ok, true}
+    ].
--- a/apps/emqx_dashboard/src/emqx_dashboard.app.src
+++ b/apps/emqx_dashboard/src/emqx_dashboard.app.src
@@ -2,7 +2,7 @@
 {application, emqx_dashboard, [
     {description, "EMQX Web Dashboard"},
     % strict semver, bump manually!
-    {vsn, "5.0.15"},
+    {vsn, "5.0.16"},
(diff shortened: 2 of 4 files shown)


apps/emqx_management/src/emqx_mgmt_api_stats.erl
```diff
@@ -129,7 +129,19 @@ list(get, #{query_string := Qs}) ->
...
             ],
             {200, Data}
     end.
+
+%%%==============================================================================================
+%% Internal
+
+running_nodes() ->
+    Nodes = erlang:nodes([visible, this]),
+    RpcResults = erpc:multicall(Nodes, emqx, is_running, [], 15000),
```
thalesmg, line 142: I think this should be added to a `emqx_mgmt_api_stats_proto_v1` module for backplane communications?

thalesmg (dismissed):

apps/emqx_management/src/emqx_mgmt_api_stats.erl
id: @thalesmg like this?

thalesmg: Yes.  I think the bpapi static check will also probably ask to add this new module to `apps/emqx/priv/bpapi.versions`.

thalesmg (approved):

apps/emqx_management/src/proto/emqx_mgmt_api_stats_proto_v1.erl
```diff
@@ -0,0 +1,34 @@
...
+]).
+
+-include_lib("emqx/include/bpapi.hrl").
+
+introduced_in() ->
+    "5.0.22".
+
+-spec is_emqx_running([node()]) -> emqx_rpc:erpc_multicall(boolean()).
+is_emqx_running(Nodes) ->
+    erpc:multicall(Nodes, emqx, is_running, [], 15000).
```
ieQu1, line 34: There is `emqx_proto_v1` with `is_running` function, it's better to extend it instead.

id:
> I think it makes more sense to call `is_running` function on the remote side, so the check and the call happen in one "transaction". 

but we only need the result of `is_running` here

ieQu1 (approved):

apps/emqx/src/emqx_rpc.erl
```diff
@@ -27,6 +27,8 @@
     cast/5,
     multicall/4,
     multicall/5,
+    multicall_on_running/5,
+    on_running/3,
```
thalesmg, line 31: Just to confirm: this isn't used anywhere, right?

thalesmg (approved):

apps/emqx/src/emqx_rpc.erl
ieQu1: Yes, I guess it can be used only in the next release. But it's good to have it now.

PR #10327, opened 2023-04-04
fix(rule_engine): don't increment unknown counter on unrecoverable er…

SergeTupchiy:
Closes: EMQX-8786

Fixes <issue-or-jira-number>

### <samp>🤖 Generated by Copilot at 7134acd</samp>

Fix rule engine crash bug and refactor `is_ok_result` function in `emqx_rule_runtime.erl`

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_runtime.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_runtime.erl
@@ -514,6 +514,8 @@ inc_action_metrics({error, {recoverable_error, _}}, RuleId) ->
     emqx_metrics_worker:inc(rule_metrics, RuleId, 'actions.failed.out_of_service');
 inc_action_metrics(?RESOURCE_ERROR_M(R, _), RuleId) when ?IS_RES_DOWN(R) ->
     emqx_metrics_worker:inc(rule_metrics, RuleId, 'actions.failed.out_of_service');
+inc_action_metrics({error, {unrecoverable_error, _}}, RuleId) ->
+    emqx_metrics_worker:inc(rule_metrics, RuleId, 'actions.failed');
 inc_action_metrics(R, RuleId) ->
     case is_ok_result(R) of
         false ->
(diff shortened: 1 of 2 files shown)


apps/emqx_rule_engine/src/emqx_rule_runtime.erl
```diff
@@ -523,9 +525,7 @@ inc_action_metrics(R, RuleId) ->
             emqx_metrics_worker:inc(rule_metrics, RuleId, 'actions.success')
     end.
 
-is_ok_result(ok) ->
```
SergeTupchiy, line 526: `ok` is already matched on line 511:
```
inc_action_metrics(ok, RuleId) ->
    emqx_metrics_worker:inc(rule_metrics, RuleId, 'actions.success');
```

SergeTupchiy:
NOTE: this fix is applicable only for synchronous bridges. If a bridge is asynchronous, rule action statistics will count everything as success.

thalesmg (dismissed):
Nit: could we add simple test for this case?  Could be in a follow up PR to avoid blocking this fix for the release, of course.

Also, maybe we need a changelog here?

thalesmg (approved):

SergeTupchiy:
> > Nit: could we add simple test for this case? Could be in a follow up PR to avoid blocking this fix for the release, of course.
> > Also, maybe we need a changelog here?
> > Thanks, @thalesmg, I've added a changelog, working on a test case, I think I'll add it to this PR.

@thalesmg , I'm going to merge it without tests in order not to block the release, as adding them may take some time.
I've added a follow-up ticket for tests https://emqx.atlassian.net/browse/EMQX-9459

apps/emqx_rule_engine/src/emqx_rule_runtime.erl
zmstone: if ok results has to be leaked down here anyway we can probably delete line 511 instead.

PR #10347, opened 2023-04-06
refactor(kafka_bridge): move kafka bridge into its own app

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-9481

## Summary
copilot:summary

## PR Checklist


jinfahua:
I think we should have a README.md under the root of the app.

id:
Remove kafka from `lib-ee/emqx_ee_bridge/docker-ct`?

JimMoen:
The `$PROJ_ROOT/LICENSE` file need be updated.

thalesmg:
> > The `$PROJ_ROOT/LICENSE` file need be updated.
> 
> How so? I've moved only this EE-only app, so as far as I know the root license is still Apache, right? thinking

I see what you mean now.  I've updated the root license file to list this new application.  Thanks!

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.app.src
```diff
@@ -0,0 +1,15 @@
+{application, emqx_bridge_kafka, [
+    {description, "EMQX Enterprise Kafka Bridge"},
+    {vsn, "0.1.0"},
+    {registered, [emqx_bridge_kafka_consumer_sup]},
+    {applications, [
+        kernel,
+        stdlib,
+        emqx_ee_connector,
```
keynslug, line 8: A couple of concerns / questions.
1. Shouldn't we depend on `emqx_connector`? In the not so distant future `emqx_ee_connector` will be split into several different applications, and there'll be no `emqx_ee_connector` application as such. Also `emqx_ee_connector` doesn't depend on `emqx_connector` / `emqx_resource` which is kinda confusing.
2. Shouldn't we depend on the client libraries here (e.g. `wolff` / `brod`)? Also, there's no need to list them in `emqx_ee_connector.app.src`?

scripts/find-apps.sh
```diff
@@ -72,6 +72,9 @@ describe_app() {
         runner="docker"
     fi
     case "${app}" in
+        apps/emqx_bridge_kafka)
+            profile='emqx-enterprise'
+            ;;
```
keynslug, line 77: Q: I wonder why doesn't this app under `lib-ee/` then? Considering also the changes in `run.sh` there are a lot of things to keep track of in different scripts, which will eventually become quite hairy.

keynslug: Ah, thanks, missed that. 👍🏼

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.app.src
thalesmg: 1. I think I accidentally inherited the dependency on `emqx_ee_connector` from the ee_bridge app.  I tested locally, and apparently we don't need either connector app in the applications list.
2. Good catch!  Forgot to move those deps. :beers:

keynslug: > apparently we don't need either connector app in the applications list.

Just a small nit. Dependencies are transitive, so they might actually be used to simplify test setups (i.e. `application:ensure_all_loaded(emqx_ee_kafka)` instead of `application:load(emqx_connector), application:load(emqx_resource), ...`). Also, it would be harder to forget to also explicitly include [something in the release](https://github.com/emqx/emqx/blob/d16b5c40d658e8fbabc0774b53c1e615a9678eb6/rebar.config.erl#L375-L409).

Though I admit I might be missing some conventions on `app.src`'s dependencies.

thalesmg: I'm double checking locally that the release and the test suites run smoothly with those changes.

lib-ee/emqx_ee_bridge/src/emqx_ee_bridge.app.src
```diff
@@ -1,12 +1,13 @@
...
     {vsn, "0.1.9"},
-    {registered, [emqx_ee_bridge_kafka_consumer_sup]},
+    {registered, []},
     {applications, [
         kernel,
         stdlib,
         emqx_ee_connector,
-        telemetry
+        telemetry,
+        emqx_bridge_kafka
```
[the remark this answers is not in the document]
thalesmg: That's the release start order.  However, since there's this kinda of mutual dependency, the `emqx_bridge` app [must wait](https://github.com/emqx/emqx/blob/f01e2f358b6e8011f4d8ffd3ca8f69e2b7cbe326/apps/emqx_bridge/src/emqx_bridge_app.erl#L47) for the EE bridge apps to start.  Otherwise, the bridges might fail to start up because dependencies like `brod` and `wolff` are not running yet.

keynslug (approved):

savonarola (approved):

PR #10359, opened 2023-04-10
feat(resource): ask for metrics only when needed

keynslug:
Followup to [EMQX-9136](https://emqx.atlassian.net/browse/EMQX-9136), [EMQX-8375](https://emqx.atlassian.net/browse/EMQX-8375).

Followup to https://github.com/emqx/emqx/pull/10123#discussion_r1138285457.

### <samp>🤖 Generated by Copilot at a71ce2e</samp>

This pull request refactors the resource metrics management in the `emqx_resource` app and introduces a new version of the bridge protocol, `emqx_bridge_proto_v4`, in the `emqx_bridge` app. The new protocol improves the efficiency and scalability of bridge metrics retrieval and separates them from the bridge status and configuration data. The pull request also updates the `emqx_authn` and `emqx_authz` apps to use the new protocol and modifies some test cases accordingly.

## PR Checklist

- [ ] ~~For internal contributor: there is a jira ticket to track this change~~


changes/ce/feat-10359.en.md
```diff
@@ -0,0 +1 @@
+Metrics now are not implicitly collected in places where API handlers does not make any use of them. Instead, a separate BPAPI RPC gathers cluster-wide metrics.
```
ieQu1, line 1:
```suggestion
Metrics now are not implicitly collected in places where API handlers don't make any use of them. Instead, a separate BPAPI RPC gathers cluster-wide metrics.
```

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -990,7 +988,7 @@ do_bpapi_call(Node, Call, Args) ->
 do_bpapi_call_vsn(SupportedVersion, Call, Args) ->
     case lists:member(SupportedVersion, supported_versions(Call)) of
         true ->
-            apply(emqx_bridge_proto_v3, Call, Args);
+            apply(emqx_bridge_proto_v4, Call, Args);
```
savonarola, line 991: Nit: isn't this a kind of evading from some of the bpapi checks? 🤔

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -37,7 +37,6 @@
     list_all/0,
```
thalesmg, line 37: Just curiosity: `list_all/0` still collects the metrics, and it seems to be used only by `emqx_resource:list_instances_verbose`, which in turn seems to be for debugging purposes.  Maybe we can remove `data_record_to_external_map_with_metrics`?

thalesmg (approved):

apps/emqx_bridge/src/emqx_bridge_api.erl
keynslug: What scenario do you have in mind? I believe that piece of logic always (silently) assumed that the type (i.e. signature) of an existing RPC does not change across protocol versions. Which is general is not true, but I'm not aware of bpapi checks for that.

savonarola: I mean, dialyzer won't check that `Args` type matches the bpapi's wrapper type; only that bpapi's wrapper declared type matches the type of the corresponding remote call.

savonarola (dismissed):

savonarola:
Things got much more readable when they got separated.

Do I understand correctly that cluster calls in bridge API ware needed only for metrics?

apps/emqx_resource/src/emqx_resource_manager.erl
thalesmg: I think the latter would be better, so we could drop the `..._map_with_metrics` fn and use the same API to collect metrics when calling this debug function.  What do you think?

keynslug: I think mostly the same, did that in 9c9f39d0f7431ae7f17b711ff61c17135a223761.

savonarola (approved):

PR #10362, opened 2023-04-10
feat(resource): stop adding uniqueness to manager ids

keynslug:
Fixes [EMQX-9257](https://emqx.atlassian.net/browse/EMQX-9257)

## Summary
Before this change, a separate `manager_id` / `instance_id` was used as a resource manager id, which made connector interface somewhat inconsistent: part of function calls to connector implementation used instance id as first argument while the rest used resource id itself.

## PR Checklist

- [ ] ~~Added tests for the changes~~

### DIFF
--- a/apps/emqx_resource/src/emqx_resource.erl
+++ b/apps/emqx_resource/src/emqx_resource.erl
@@ -113,7 +113,10 @@
 
 -export([apply_reply_fun/2]).
 
--export_type([resource_data/0]).
+-export_type([
+    resource_id/0,
+    resource_data/0
+]).
 
 -optional_callbacks([
     on_query/3,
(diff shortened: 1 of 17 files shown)


apps/emqx_resource/src/emqx_resource.erl
```diff
@@ -365,15 +368,15 @@ call_start(MgrId, Mod, Config) ->
             {error, #{exception => Kind, reason => Error, stacktrace => Stacktrace}}
     end.
 
--spec call_health_check(manager_id(), module(), resource_state()) ->
+-spec call_health_check(resource_id(), module(), resource_state()) ->
     resource_status()
     | {resource_status(), resource_state()}
     | {resource_status(), resource_state(), term()}
     | {error, term()}.
 call_health_check(MgrId, Mod, ResourceState) ->
```
savonarola, line 376: Should we get rid of `MgrId` naming?

keynslug: Nice find!

ieQu1 (dismissed):

PR #10371, opened 2023-04-11
feat(respool): switch to `emqx_resource_pool`

keynslug:
Fixes [EMQX-9257](https://emqx.atlassian.net/browse/EMQX-9257).

## Summary
This module was previously known as `emqx_plugin_libs_pool`. This is part of the effort to get rid of `emqx_plugin_libs` application.

Also make naming across connectors / testsuites more consistent.

## PR Checklist

- [ ] ~~Added tests for the changes~~

### DIFF
--- a/apps/emqx_connector/src/emqx_connector_mysql.erl
+++ b/apps/emqx_connector/src/emqx_connector_mysql.erl
@@ -51,7 +51,7 @@
 -type sqls() :: #{atom() => binary()}.
 -type state() ::
     #{
-        poolname := atom(),
+        pool_name := binary(),
         prepare_statement := prepares(),
         params_tokens := params_tokens(),
         batch_inserts := sqls(),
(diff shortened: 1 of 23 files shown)


apps/emqx_connector/src/emqx_connector_mysql.erl
```diff
@@ -143,12 +140,12 @@ maybe_add_password_opt(undefined, Options) ->
...
     [{password, Password} | Options].
 
-on_stop(InstId, #{poolname := PoolName}) ->
+on_stop(InstId, #{pool_name := PoolName}) ->
     ?SLOG(info, #{
         msg => "stopping_mysql_connector",
         connector => InstId
     }),
-    emqx_plugin_libs_pool:stop_pool(PoolName).
+    emqx_resource_pool:stop(PoolName).
```
keynslug, line 148: Note. We can't get rid of the `pool_name` state bit altogether because of implications of #8247, see #10362.

thalesmg (approved):

thalesmg (approved):

apps/emqx_connector/src/emqx_connector_mysql.erl
keynslug: Ah. We probably can. But the latter is a good question. I'll address it in #10362.

lib-ee/emqx_ee_connector/src/emqx_ee_connector_gcp_pubsub.erl
```diff
@@ -155,75 +150,71 @@ on_stop(
...
+    resource_id(),
     {send_message, map()},
     state()
 ) ->
     {ok, status_code(), headers()}
     | {ok, status_code(), headers(), body()}
     | {error, {recoverable_error, term()}}
     | {error, term()}.
-on_query(BridgeId, {send_message, Selected}, State) ->
+on_query(ResourceId, {send_message, Selected}, State) ->
```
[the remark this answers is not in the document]
keynslug: That's the consequence of #8247. You probably mean that we need something in the code to tell contributors that explicitly? Again, probably better to tackle in the sibling PR.

PR #10392, opened 2023-04-13
freat: add date_to_unix_ts/3 function to the rule engine

kjellwinblad:
Fixes https://emqx.atlassian.net/browse/EMQX-9245

### <samp>🤖 Generated by Copilot at 8435571</samp>

Add a new rule engine function `date_to_unix_ts/3` to convert date strings to unix timestamps and a test case for it in `emqx_rule_funcs_SUITE.erl`.

## PR Checklist

- [X] Change log has been added to `changes/{ce,ee}/(feat|perf|fix)-<PR-id>.en.md` files
- [X] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up
- [X] Schema changes are backward compatible

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_funcs.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_funcs.erl
@@ -227,6 +227,7 @@
     now_timestamp/1,
     format_date/3,
     format_date/4,
+    date_to_unix_ts/3,
     date_to_unix_ts/4
 ]).
 
@@ -1083,6 +1084,14 @@ format_date(TimeUnit, Offset, FormatString, TimeEpoch) ->
             emqx_plugin_libs_rule:str(FormatString),
             TimeEpoch
         )
+    ).
+
+date_to_unix_ts(TimeUnit, FormatString, InputString) ->
+    emqx_rule_date:parse_date(
+        time_unit(TimeUnit),
+        "Z",
+        emqx_plugin_libs_rule:str(FormatString),
+        emqx_plugin_libs_rule:str(InputString)
     ).
 
 date_to_unix_ts(TimeUnit, Offset, FormatString, InputString) ->
--- a/apps/emqx_rule_engine/src/emqx_rule_funcs.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_funcs.erl
@@ -1087,12 +1087,7 @@ format_date(TimeUnit, Offset, FormatString, TimeEpoch) ->
     ).
 
 date_to_unix_ts(TimeUnit, FormatString, InputString) ->
-    emqx_rule_date:parse_date(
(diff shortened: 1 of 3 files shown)


apps/emqx_rule_engine/test/emqx_rule_funcs_SUITE.erl
```diff
@@ -1003,6 +1003,24 @@ prop_format_date_fun() ->
                         )
                     ]
             )
+    ),
+    %% When no offset is specified, the offset should be taken from the formatted time string
+    ArgsNoOffset = [<<"second">>, <<"%y-%m-%d-%H:%M:%S%Z">>],
+    ArgsOffset = [<<"second">>, <<"+08:00">>, <<"%y-%m-%d-%H:%M:%S%Z">>],
+    ?FORALL(
+        S,
+        erlang:system_time(second),
```
thalesmg, line 1012: I think this is called only once for all tests.  I tried locally with `nanosecond` and printing the value, it's always the same.

To achieve the desired generation of values, I think one way is to do:

```suggestion
        ?LAZY(erlang:system_time(second)),
```

Not sure if the most correct way to produce values, but does produce different values.

Also, note that the generation is very fast: it produces several values in a second.  So, even with `?LAZY`, this will basically always return the same value.

Perhaps we could instead define a range of epoch values?  Like

```suggestion
        range(0, 4000000000),
```

apps/emqx_rule_engine/src/emqx_rule_funcs.erl
```diff
@@ -1085,6 +1086,14 @@ format_date(TimeUnit, Offset, FormatString, TimeEpoch) ->
         )
     ).
 
+date_to_unix_ts(TimeUnit, FormatString, InputString) ->
+    emqx_rule_date:parse_date(
+        time_unit(TimeUnit),
+        "Z",
+        emqx_plugin_libs_rule:str(FormatString),
+        emqx_plugin_libs_rule:str(InputString)
+    ).
```
thalesmg, line 1095: Tiny nit:

```suggestion
    date_to_unix_ts(TimeUnit, "Z", FormatString, InputString).
```
?

apps/emqx_rule_engine/test/emqx_rule_funcs_SUITE.erl
kjellwinblad: Thanks, I think this problem existed in the other test cases in the same function as well. Will try to fix them too.

kjellwinblad: @thalesmg thanks for spotting this. I have updated the test case in the latest commit.

thalesmg (approved):

Issue #10357, opened 2023-04-10
Resource manager crashes due to too many alarms


terry-xiaoyu:
### What happened?

The original problem: https://askemq.com/t/topic/4583/1

The resource manager crashes due to failed to send alarms to emqx_alarm:

![The resource manager crashed and reached the max restart times](https://askemq.com/uploads/default/original/2X/5/502f4d65d70d21f02695961fcc6f87070b991d84.png)

![Too many alarms generated on the dashboard](https://askemq.com/uploads/default/original/2X/5/502f4d65d70d21f02695961fcc6f87070b991d84.png)

### What did you expect to happen?

Don't crash the resource manager. Maybe we should use `gen_server:cast/2` rather than `gen_server:call/3`, as we only need try our best when sending alarms. 

But this would overwhelm the `emqx_alarm`, so we need some overload protection mechanism.

### EMQX version

<details>

opensource emqx 5.0.17, 3-node cluster

</details>

### OS version

<details>

Centos 7.9

</details>

### Log files

<details>

</details>

PR #10407, opened 2023-04-14
fix(emqx_resource): call emqx_alarm safely and don't reactivate alarm…

SergeTupchiy:
emqx_alarm gen_server may not respond timely under high load, especially when there are also a relatively high number of failing resources, e.g. bridges. This fix avoids unnecessary calls to activate an alarm when it was already activated before. Additionally, it adds safe calls to emqx_alarm, so that alarm timeouts doesn't cause emqx_resource_manager to reach it restart intensity and crash completely.

Fixes EMQX-9529, #10357

## Summary
copilot:summary

## PR Checklist


apps/emqx_resource/include/emqx_resource_utils.hrl
```diff
@@ -14,10 +14,24 @@
...
+
 -define(SAFE_CALL(_EXP_),
     ?SAFE_CALL(_EXP_, {error, {_EXCLASS_, _EXCPTION_, _ST_}})
 ).
 
+-define(SAFE_CALL_WITH_LOG(_EXP_),
+    ?SAFE_CALL(
+        _EXP_,
+        ?SLOG(warning, #{
+            msg => "safe call error",
```
savonarola, line 27: Nit: maybe provide a message as an argument?

changes/ce/fix-10407.en.md
```diff
@@ -0,0 +1,6 @@
+Call emqx_alarm safely, and don't reactivate alarms on reoccurring errors.
+emqx_alarm gen_server may not respond timely under high load, especially
+when there is also a relatively high number of failing resources, e. g. bridges.
```
zmstone, line 3:
```suggestion
when there is also a relatively high number of failing resources, e.g. bridges.
```

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -293,7 +294,9 @@ start_link(MgrId, ResId, Group, ResourceType, Config, Opts) ->
         config = Config,
         opts = Opts,
         state = undefined,
-        error = undefined
+        %% `undefined` error may signify error state with undefined reason,
+        %% `none` explicitly means that no error occurred
+        error = none
```
savonarola, line 299: Nit: `ok | {error, Error}`?

apps/emqx_resource/include/emqx_resource_utils.hrl
zmstone: and we use snake_case_message for `msg`

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -605,21 +608,27 @@ update_state(Data, _DataWas) ->
...
         case Error of
             undefined -> <<"Unknown reason">>;
             _Else -> emqx_utils:readable_error_msg(Error)
         end,
-    emqx_alarm:activate(
-        ResId,
-        #{resource_id => ResId, reason => resource_down},
-        <<"resource down: ", HrError/binary>>
-    ).
+    ?SAFE_CALL_WITH_LOG(
```
zmstone, line 621: move the safety code to the callee module

apps/emqx_resource/src/emqx_resource_manager.erl
keynslug, line 299: Nit. Having a term of type `{error, _Reason :: term()} | undefined` here should unambiguous? Right now AFAICS it's ambiguous with respect to `{error, none}` error from a healthcheck.

savonarola (approved):

keynslug (commented):
The problem itself begs also a question: shouldn't we deal with it in the `emqx_alarm` module? For example, it's not really clear why it runs `mria:transaction/3` in the gen_server process, or why it needs mria transactions at all, given that Alarms API does cluster-wide queries by itself. Also it already does some kind of alarm deduplication, maybe we just need to improve its "hitrate" there somehow.

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -605,21 +608,27 @@ update_state(Data, _DataWas) ->
...
+        emqx_alarm:activate(
+            ResId,
+            #{resource_id => ResId, reason => resource_down},
+            <<"resource down: ", HrError/binary>>
+        )
+    ),
+    ?tp(resource_activate_alarm, #{resource_id => ResId});
+%% Assume that alarm is already active
+maybe_alarm(_Status, _ResId, _Error, _PrevError) ->
```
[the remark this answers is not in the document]
SergeTupchiy: You are right, we don't update alarm description when moving from one error state to another (new) error state. 
I think it also should be addressed.. It can be easily changed in `maybe_alarm` by comparing prev and current errors, but we need to update emqx_alarm accordingly.

apps/emqx_resource/test/emqx_resource_SUITE.erl
```diff
@@ -2751,6 +2750,42 @@ t_volatile_offload_mode(_Config) ->
...
+                #{?snk_kind := resource_activate_alarm, resource_id := ?ID}
+            ),
+            ?assertMatch([#{activated := true, name := ?ID}], emqx_alarm:get_alarms(activated)),
+            {ok, SubRef} = snabbkaffe:subscribe(
+                ?match_event(#{?snk_kind := SubscribeEvent}), 4, 7000
+            ),
+            ?assertMatch({ok, [_, _, _, _]}, snabbkaffe:receive_events(SubRef))
+        end,
+        fun(Trace) ->
+            ?assertMatch([_ | []], ?of_kind(resource_activate_alarm, Trace))
```
thalesmg, line 2785: Tiny nit:

```suggestion
            ?assertMatch([_], ?of_kind(resource_activate_alarm, Trace))
```

thalesmg:
> or why it needs mria transactions at all, given that Alarms API does cluster-wide queries by itself

Good point.

I think the transactions would be for activations/deactivations that happen internally (not triggered by the HTTP API).

I wonder if it could use just `dirty_write` or even `async_dirty` :thinking: 
The deduplication would be lost, I guess, but if we use the same name, then it could just be an "upsert" operation.  :thinking:

apps/emqx_resource/test/emqx_resource_SUITE.erl
SergeTupchiy: Thanks, fixed.. I wonder how and why I came up with `[_ | [ ]]`, quite ridiculous.. :see_no_evil:

apps/emqx_resource/include/emqx_resource_utils.hrl
SergeTupchiy: Fixed

apps/emqx_resource/src/emqx_resource_manager.erl
SergeTupchiy: Fixed as suggested by @keynslug (https://github.com/emqx/emqx/pull/10407#discussion_r1170223263)

changes/ce/fix-10407.en.md
SergeTupchiy: Fixed.

apps/emqx_resource/src/emqx_resource_manager.erl
SergeTupchiy: Done.

apps/emqx_resource/src/emqx_resource_manager.erl
SergeTupchiy: Thanks, fixed.

apps/emqx_resource/src/emqx_resource_manager.erl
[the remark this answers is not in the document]
SergeTupchiy: I've changed it to avoid activating alarm only if previous error is the same as the current one. 
For `emqx_alarm`, I've added a separate task: https://emqx.atlassian.net/browse/EMQX-9636, as it's not closely connected with this PR..

keynslug (dismissed):

apps/emqx/src/emqx_alarm.erl
```diff
@@ -474,3 +476,19 @@ normalize_message(Name, <<"">>) ->
     list_to_binary(io_lib:format("~p", [Name]));
 normalize_message(_Name, Message) ->
     Message.
+
+safe_call(Req) ->
```
SergeTupchiy, line 480: Given the introduced optimizations in resource_manager (avoid extra activate calls) and emqx_alarm itself (dirty ops instead of transaction), it seems safe to use `gen_server:cast/2` instead. 
However, I'm still reluctant to do it, as cast is easier to abuse.. :thinking:

SergeTupchiy:
> > or why it needs mria transactions at all, given that Alarms API does cluster-wide queries by itself
> 
> Good point.
> 
> I think the transactions would be for activations/deactivations that happen internally (not triggered by the HTTP API).
> 
> I wonder if it could use just `dirty_write` or even `async_dirty` thinking The deduplication would be lost, I guess, but if we use the same name, then it could just be an "upsert" operation. thinking

I've replaced transaction with dirty read/write operations. 
It's a local content shard and all the operations are effectively serialized, as they are called by one process (`emqx_alarm` gen_server), so it's safe to use dirty ops.
This gives a great performance boost, for example, transaction latency under high load degraded from sub millisecond to 
~500 ms. Dirty read/write latency was almost unaffected by the high load and stayed under a millisecond. 
The load was generated locally by `./emqtt_bench pub -p 1883 -c 1000 -I 10 -t bench/%i -s 25` against a single node EMQX on a very basic hardware  (i5-7200U laptop), OS: Ubuntu 20.04.
This alone should be enough to mitigate the issue, but I've kept all the other changes in this PR for now.

apps/emqx/src/emqx_alarm.erl
```diff
@@ -283,18 +285,18 @@ get_validity_period() ->
...
         [] ->
             Alarm = #activated_alarm{
                 name = Name,
                 details = Details,
                 message = normalize_message(Name, iolist_to_binary(Message)),
                 activate_at = erlang:system_time(microsecond)
             },
-            ok = mnesia:write(?ACTIVATED_ALARM, Alarm, write),
-            Alarm
+            ok = mnesia:dirty_write(?ACTIVATED_ALARM, Alarm),
```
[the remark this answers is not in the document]
SergeTupchiy: Fixed, thanks.

ieQu1 (approved):

ieQu1 (approved):

Issue #10411, opened 2023-04-16
Slash missing from actual request


rosaschafe:
### What happened?

`EMQX_AUTHENTICATION__1__URL: 'http://frontend/api/authn/'`

But error:

2023-04-16T18:51:18.177945+00:00 [error] msg: http connector do request, received error response., mfa: emqx_connector_http:on_query/3, line: 344, peername: 172.20.0.8:36027, clientid: MzEwMjEzNTU5OTQ4OTc4MjYzOTczMTU2MzM5NjMwNzM1MzG, connector: <<"emqx_authn_http:1">>, note: the body will be redacted due to security reasons, request: {"/api/authn",[{<<"accept">>,<<"application/json">>},{<<"cache-control">>,<<"no-cache">>},{<<"connection">>,<<"keep-alive">>},{<<"content-type">>,<<"application/json">>},{<<"keep-alive">>,<<"timeout=30, max=1000">>},{<<"x-request-source">>,<<"EMQX">>}],<<"******">>}, status_code: 301

While making the request, "/" was dropped and this cause 302 from Django.

### What did you expect to happen?

Request with a "/" was given

### EMQX version

Docker image: emqx:5.0.22

rosaschafe:
FYI Also doing: `EMQX_AUTHENTICATION__1__URL: 'http://frontend/api/authn/?'` didnt help either.

savonarola:
Hello! This seems to be a bug, this should work without workarounds. I created an internal issue for that.

PR #10420, opened 2023-04-17
Fix path joining and quoting in authn/authz

savonarola:
Fixes https://emqx.atlassian.net/browse/EMQX-9600, https://github.com/emqx/emqx/issues/10411

## PR Checklist

~~- [ ] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up~~

### DIFF
--- a/apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
+++ b/apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
@@ -285,9 +285,9 @@ parse_url(Url) ->
                     BaseUrl = iolist_to_binary([Scheme, "//", HostPort]),
                     case string:split(Remaining, "?", leading) of
                         [Path, QueryString] ->
-                            {BaseUrl, Path, QueryString};
+                            {BaseUrl, <<"/", Path/binary>>, QueryString};
                         [Path] ->
-                            {BaseUrl, Path, <<>>}
+                            {BaseUrl, <<"/", Path/binary>>, <<>>}
                     end;
                 [HostPort] ->
                     {iolist_to_binary([Scheme, "//", HostPort]), <<>>, <<>>}
@@ -328,7 +328,7 @@ generate_request(Credential, #{
     body_template := BodyTemplate
 }) ->
     Headers = maps:to_list(Headers0),
-    Path = emqx_authn_utils:render_str(BasePathTemplate, Credential),
+    Path = emqx_authn_utils:render_urlencoded_str(BasePathTemplate, Credential),
     Query = emqx_authn_utils:render_deep(BaseQueryTemplate, Credential),
     Body = emqx_authn_utils:render_deep(BodyTemplate, Credential),
     case Method of
@@ -343,9 +343,9 @@ generate_request(Credential, #{
     end.
 
 append_query(Path, []) ->
-    encode_path(Path);
+    Path;
 append_query(Path, Query) ->
-    encode_path(Path) ++ "?" ++ binary_to_list(qs(Query)).
+    Path ++ "?" ++ binary_to_list(qs(Query)).
 
 qs(KVs) ->
     qs(KVs, []).
@@ -406,10 +406,6 @@ parse_body(ContentType, _) ->
 
 uri_encode(T) ->
     emqx_http_lib:uri_encode(to_list(T)).
-
-encode_path(Path) ->
-    Parts = string:split(Path, "/", all),
-    lists:flatten(["/" ++ Part || Part <- lists:map(fun uri_encode/1, Parts)]).
 
 request_for_log(Credential, #{url := Url} = State) ->
     SafeCredential = emqx_authn_utils:without_password(Credential),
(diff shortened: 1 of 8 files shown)


changes/ce/fix-10420.zh.md
```diff
@@ -0,0 +1,3 @@
+在身份验证和授权模块中为 HTTP 请求编写 URL 时修复 HTTP 路径处理。
+* 避免不必要的 URL 规范化，因为我们不能假设外部服务器平等对待原始 URL 和规范化 URL。 这导致了像 [#10411](https://github.com/emqx/emqx/issues/10411) 这样的错误。
+* 修复路径段可能被 HTTP 编码两次的问题。
```
zmstone, line 3: no need for this anymore.

apps/emqx_connector/src/emqx_connector_http.erl
```diff
@@ -553,15 +553,41 @@ formalize_request(Method, BasePath, {Path, Headers, _Body}) when
...
+    lists:reverse(Path1) ++ [$/ | Path2];
+%% "abc" + "/cde"
+do_join_paths(Path1, [$/ | Path2]) ->
+    lists:reverse(Path1) ++ [$/ | Path2];
+%% "abc" + "cde"
+do_join_paths(Path1, Path2) ->
+    lists:reverse(Path1) ++ [$/ | Path2].
+
+to_list(List) when is_list(List) -> List;
+to_list(Bin) when is_binary(Bin) -> binary_to_list(Bin).
```
zmstone, line 584: we have to handle unicode here.

apps/emqx_connector/src/emqx_connector_http.erl
```diff
@@ -553,15 +553,41 @@ formalize_request(Method, BasePath, {Path, Headers, _Body}) when
...
+
+to_bin(Bin) when is_binary(Bin) ->
     Bin;
-bin(Str) when is_list(Str) ->
+to_bin(Str) when is_list(Str) ->
     list_to_binary(Str);
```
zmstone, line 589: unicode list.

apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
```diff
@@ -285,9 +285,9 @@ parse_url(Url) ->
                     BaseUrl = iolist_to_binary([Scheme, "//", HostPort]),
                     case string:split(Remaining, "?", leading) of
                         [Path, QueryString] ->
-                            {BaseUrl, Path, QueryString};
```
thalesmg, line 288: Not that it should be done in this PR, but I feel this `parse_url/1` fn could be replaced by `uri_string:parse` :thinking: 

(same for `apps/emqx_authz/src/emqx_authz_http.erl`)

apps/emqx_connector/src/emqx_connector_http.erl
savonarola: Valid URLs cannot contain wide characters anyway; a request with Unicode will either fail in the library or be rejected by the server.
However, checking URL validity on each request seems to be an overkill 🤔

apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
savonarola: Unfortunately not 😅 This is an old conceptual issue — that users provide _templates_ for urls `https://my.site/user/${username}?password=${password}` which are not valid urls and are treated as invalid by URL parsers.

savonarola: Seems that this problem was the reason for appearing of this loose parser.

apps/emqx_connector/src/emqx_connector_http.erl
savonarola: This was intended to be a no-op for joining HTTP paths.
`BasePath` is parsed by `emqx_http_lib` and is already a list.
The second argument comes from `on_query` arguments, we may add a guard there to force connector users to take care of converting path to a valid string themselves.

changes/ce/fix-10420.zh.md
savonarola: ✅

zmstone (approved):

apps/emqx_connector/src/emqx_connector_http.erl
```diff
@@ -642,4 +668,21 @@ redact_test_() ->
...
+        ?_assertEqual("abc/cde", join_paths("abc/", "/cde")),
+
+        ?_assertEqual("/", join_paths("", "")),
+        ?_assertEqual("/cde", join_paths("", "cde")),
+        ?_assertEqual("/cde", join_paths("", "/cde")),
+        ?_assertEqual("/cde", join_paths("/", "cde")),
+        ?_assertEqual("/cde", join_paths("/", "/cde")),
+
+        ?_assertEqual("//cde/", join_paths("/", "//cde/")),
+        ?_assertEqual("abc///cde/", join_paths("abc//", "//cde/"))
```
HJianBo, line 685: nit: why not `abc////cde/`? @savonarola

PR #10445, opened 2023-04-18
chore: delete old changelogs

id:


changes/ee/feat-10409.en.md
```diff
@@ -1 +0,0 @@
-Add support for [Protocol Buffers](https://protobuf.dev/) schemas in Schema Registry.
```
thalesmg, line 1: Is this "old"?  I think it's not included in the e5.0.3 alpha changelog, and was merged very recently to `master` only. :thinking:

id: Indeed, I just ran the script and didn't check if it's valid.

PR #10455, opened 2023-04-19
fix(buffer_worker): avoid sending late reply messages to callers

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-9635

During a sync call from process `A` to a buffer worker `B`, its call to the underlying resource `C` can be very slow.  In those cases, `A` will receive a timeout response and expect no more messages from `B` nor `C`.  However, prior to this fix, if `B` is stuck in a long sync call to `C` and then gets its response after `A` timed out, `B` would still send the late response to `A`, polluting its mailbox.

### <samp>🤖 Generated by Copilot at 8b5521b</samp>

Added a new function `reply_call/2` to `emqx_resource_buffer_worker` module to improve query request handling and tracing. Updated the version number of the `emqx_resource` application. Added test cases for timeout and late reply scenarios using the `emqx_connector_demo` module.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -52,7 +52,7 @@
 
 -export([queue_item_marshaller/1, estimate_size/1]).
 
--export([handle_async_reply/2, handle_async_batch_reply/2]).
+-export([handle_async_reply/2, handle_async_batch_reply/2, reply_call/2]).
 
 -export([clear_disk_queue_dir/2]).
 
@@ -293,10 +293,8 @@ code_change(_OldVsn, State, _Extra) ->
 
 pick_call(Id, Key, Query, Timeout) ->
     ?PICK(Id, Key, Pid, begin
-        Caller = self(),
         MRef = erlang:monitor(process, Pid, [{alias, reply_demonitor}]),
-        From = {Caller, MRef},
-        ReplyTo = {fun gen_statem:reply/2, [From]},
+        ReplyTo = {fun ?MODULE:reply_call/2, [MRef]},
         erlang:send(Pid, ?SEND_REQ(ReplyTo, Query)),
         receive
             {MRef, Response} ->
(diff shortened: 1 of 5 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -295,19 +295,19 @@ pick_call(Id, Key, Query, Timeout) ->
...
                 erlang:demonitor(MRef, [flush]),
                 Response;
             {'DOWN', MRef, process, Pid, Reason} ->
                 error({worker_down, Reason})
         after Timeout ->
             erlang:demonitor(MRef, [flush]),
             receive
-                {MRef, Response} ->
+                {[alias | MRef], Response} ->
                     Response
```
zmstone, line 311: we are not using gen_statem:call, it's unnecessary to use gen:reply here.
maybe simply 
```
From = Mref,
ReplyTo = {fun erlang:send/2, [Mref]},
erlang:send(Pid, ....)
receive
  {Mref, Response} ->
     ...

File: apps/emqx_resource/src/emqx_resource_buffer_worker.erl [hunk 2]
```diff
@@ -295,19 +295,19 @@ pick_call(Id, Key, Query, Timeout) ->
     ?PICK(Id, Key, Pid, begin
         Caller = self(),
         MRef = erlang:monitor(process, Pid, [{alias, reply_demonitor}]),
-        From = {Caller, MRef},
+        From = {Caller, [alias | MRef]},
```
ieQu1, line 298: This function is pretty complicated, it would be great to explain what's going on in a comment.
Q: why does From contain an improper list?

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
thalesmg: That's what I did originally.  I'll change it back to that.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
thalesmg: I'll drop the usage of `gen_statem:reply` and go back to my first solution of just sending to `MRef` the reply, after Stone's comment above.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
thalesmg: :heavy_check_mark:

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
thalesmg: :heavy_check_mark:

zmstone (approved):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
zmstone: Sorry, if it was me suggested to use gen_statem:reply, i probably misunderstood how alias works in the `gen` module.

thalesmg: No, I think it was using that since the first version of this module :thinking:

PR #10449, opened 2023-04-19
fix: always check authn_http's header and ssl_option

zhongwencool:
Fixes [EMQX-9631](https://emqx.atlassian.net/browse/EMQX-9631)

If the ssl_options and header configurations are set incorrectly when creating `authn_http`, it could result in the entire authn being unusable. 

### Before
we can create an `authn_http` with `{ssl_options.enable=false, url = "https://xxx"}`, but **https** must enable ssl_options.enable. 

### After
image

### <samp>🤖 Generated by Copilot at dfb7bbf</samp>

Added and improved validation functions and tests for the http authentication backend in `emqx_authn`. Fixed some minor issues and updated the version number.

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [x] Added tests for the changes
- [x] Changed lines covered in coverage report
- [x] Change log has been added to `changes/{ce,ee}/(feat|perf|fix)-<PR-id>.en.md` files
- [x] For internal contributor: there is a jira ticket to track this change
- [ ] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up
- [x] Schema changes are backward compatible

## Checklist for CI (.github/workflows) changes

- [ ] If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)
- [ ] Change log has been added to `changes/` dir for user-facing artifacts update

PR #10469, opened 2023-04-21
Port authn http validation

zhongwencool:
Port https://github.com/emqx/emqx/pull/10449

### <samp>🤖 Generated by Copilot at 1db38de</samp>

This pull request fixes a bug in the `http` authentication backend of EMQ X, and adds validation functions to ensure the correctness of the backend configuration. It also improves the testing of the configuration schema parsing, and removes an obsolete file. The changes are documented in `changes/ce/fix-10449.en.md`.

## PR Checklist

### DIFF
--- a/apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
+++ b/apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
@@ -38,6 +38,8 @@
     headers/1
 ]).
 
+-export([check_headers/1, check_ssl_opts/1]).
+
 -export([
     refs/0,
     union_member_selector/1,
@@ -106,8 +108,8 @@ common_fields() ->
 
 validations() ->
     [
-        {check_ssl_opts, fun check_ssl_opts/1},
-        {check_headers, fun check_headers/1}
+        {check_ssl_opts, fun ?MODULE:check_ssl_opts/1},
+        {check_headers, fun ?MODULE:check_headers/1}
     ].
 
 url(type) -> binary();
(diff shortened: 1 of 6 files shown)


apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
```diff
@@ -261,21 +263,47 @@ transform_header_name(Headers) ->
...
-    Headers = get_conf_val("headers", Conf),
-    Method =:= <<"post">> orelse (not maps:is_key(<<"content-type">>, Headers)).
+    case is_backend_http(Conf) of
+        true ->
+            Headers = get_conf_val("headers", Conf),
+            case to_bin(get_conf_val("method", Conf)) of
+                <<"post">> ->
+                    ok;
+                <<"get">> ->
+                    case maps:is_key(<<"content-type">>, Headers) of
```
zmstone, line 293: we should add a converter to just drop this header if provided.

apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
```diff
@@ -261,21 +263,47 @@ transform_header_name(Headers) ->
...
+        true ->
+            Url = get_conf_val("url", Conf),
+            {BaseUrl, _Path, _Query} = parse_url(Url),
+            case BaseUrl of
+                <<"https://", _/binary>> ->
+                    case get_conf_val("ssl.enable", Conf) of
+                        true ->
+                            ok;
+                        false ->
+                            <<"it's required to enable the TLS option to establish a https connection">>
```
zmstone, line 276: we can simply use a converter to enable ssl ?

zmstone:
add the converter to http connector schema so all modules can benefit from it.

apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
zhongwencool: If we help user to enabled ssl, how to configure `#{enable => true, cacertfile => ... certfile => ....keyfile => ...}`
I thinks it must provide by user.

apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
zhongwencool: Why we drop headers silently ? 
Maybe we should tell user, we don't carry this header, otherwise it's hard to find problem.

zmstone:
found more issues
1. base_url field in emqx_connector_http is never used, so we can delete it from there. (and remove the maps:without and proplists:delete calls).
2. There is a ssl check implementation in emqx_connector_http, it should be generic enough to be reused by both authn and authz (the function is exported), but IMO, we should export a converter instead of a validator.

apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
zmstone: the default is verify_none, so there is usually no need to configure cacerts.
and typically http servers do not validate client certificates, so the cert files typically not needed.

apps/emqx_authn/src/simple_authn/emqx_authn_http.erl
zmstone: if this was added from day one, then it's probably not an issue,
but we are now adding a more strict check on more lossy check older version. I'm not even sure if node will successfully boot after this change.

zhongwencool: I'm sure this change is ok for older version.
The older version always check config when fill default. 
https://github.com/emqx/emqx/blob/master/apps/emqx_authn/src/emqx_authn_api.erl#L1133-L1136
this check_config alway call those two validators.

zmstone (approved):

HJianBo (approved):

Issue #10474, opened 2023-04-21
Upgrading to EMQX 5.0.23 from 5.0.22 appears to lose previously configured settings and break.


ruifung:
### What happened?

When upgrading from EMQX 5.0.22 to 5.0.23, certain settings do not appear to be preserved.
For example. I had authorization rules and authentication accounts set up via admin interface.
After upgrading to 5.0.23, it lists no authentication sources and no authorization sources. Authentication does still appear to work (i.e. previously created username/password pairs still work)
however, authorization appears to be broken as I've observed services on my home network are unable to listen to messages.

### What did you expect to happen?

For existing settings and configuration to be carried across with no issues. Especially since according to the version number, this should be a patch release.

### How can we reproduce it (as minimally and precisely as possible)?

1. (From a pre-5.0.23 version), create username (internal DB), and authorization (file)
2. Upgrade to 5.0.23
3. Observe admin UI no longer shows them, and authorization rules seem to be wierd.

### Anything else we need to know?

For now, I've downgraded back to 5.0.22. Being able to control the lights is kind of important. For now, I do not have logs from 5.0.23.

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
NOT APPLICABLE. Running on Talos Linux v1.3.3 / Kubernetes, does not contain console.
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Log files

<details>

</details>

savonarola:
Hello!
* How many nodes does your setup have?
* What is the deployment and upgrade method used for Kubernetes?
* Did the authentication and authorization return back to normal after the downgrade back to 5.0.23?
* What mechanism is used to provide persistence for EMQX nodes?

nicandris:
yup had the same issue (docker). i couldnt login on the dashboard and no device could connect. I had to change all references in the config files of message_in to messages, bytes_in to bytes etc (check changelog)

but now i keep getting errors while navigating menus in the dashboard eg

[...]

PR #10484, opened 2023-04-23
fix: copy cluster-override.conf from old version

zhongwencool:
Fixes https://github.com/emqx/emqx/issues/10474
[EMQX-9680](https://emqx.atlassian.net/browse/EMQX-9680)

Previously, when we copied the initial configuration from an old node, if there was no 'has_deprecated_file' field, we treated it as a new mechanism using `cluster.hocon`. However, the old node never returned the `has_deprecated_file'` field before upgrading (due to a problem with the old code), so during rolling upgrade, it was always assumed that `cluster.hocon` was in effect when it should have been `cluster-override.conf`.

It's hard to write  rolling upgrade CT test. I have verify on `v5.0.21` and `v5.0.23`

### <samp>🤖 Generated by Copilot at 7af9c18</samp>

Improved cluster configuration synchronization for mixed-version nodes. Refactored `emqx_conf_app.erl` and bumped `emqx_conf` version to 0.1.18.

## PR Checklist


apps/emqx_conf/src/emqx_conf_app.erl
```diff
@@ -175,7 +175,7 @@ copy_override_conf_from_core_node() ->
                 _ ->
                     [{ok, Info} | _] = lists:sort(fun conf_sort/2, Ready),
                     #{node := Node, conf := RawOverrideConf, tnx_id := TnxId} = Info,
-                    HasDeprecatedFile = maps:get(has_deprecated_file, Info, false),
+                    HasDeprecatedFile = has_deprecated_file(Info),
```
zmstone, line 178: There is no need to introduce a new version bpapi.
All we need to do is to check `maps:is_key(has_deprecated_file, Info)`

zhongwencool: We will always get false, regardless of whether the old **version(v5.0.22) has this file. We can't change old code.

zhongwencool: Mix versions in the same cluster for a long time is not normal.

1. v5.0.22 has cluster-override.conf, v5.0.24 join in, v5.0.24 will copy from v5.0.22 without has_deprecated_file.
So the v5.0.24 will write to cluster.hocon. result: after all node upgrade to v5.0.24,  the config priority is changed.
So we need rpc to old node to check the file exist.

2. v5.0.22 don't have cluster-override.conf, v5.0.24 join in, v5.0.24 will copy from v5.0.22 without has_deprecated_file. So the v5.0.24 will write to cluster.hocon. after all node upgrade to v5.0.24, the config priority is changed, This level of change is normal because it had no priority on the old version node.
It is important to note that when rolling upgrades (when there are multiple versions) we should not change the configuration via the old version without cluster-override.conf.

zmstone (approved):
Writeoff for merge without test case.
Please make sure that:
1. The change has been tested (manually?)
2. There is a follow up task to add test case.

zhongwencool:
> Writeoff for merge without test case. Please make sure that:
> 
> 1. The change has been tested (manually?)
> 2. There is a follow up task to add test case.

follow up task: https://emqx.atlassian.net/browse/EMQX-9690

PR #10489, opened 2023-04-24
feat: warning unknown config root key

zhongwencool:
Fixes [EMQX-9496](https://emqx.atlassian.net/browse/EMQX-9496)
When configuring EMQX, it is possible to manually misspell configuration items, which can lead to configuration not taking effect. 
For example, if "log" is misspelled as "Log", we need to give a warning that this configuration item will not take effect.
Ignore bad config root key, and only give warning log.

`Hocon` will ignore the configuration of root keys that he doesn't recognize. I think it's a good chioce, we will not modify this behavior.

### <samp>🤖 Generated by Copilot at bcb66bb</samp>

Add a function and a test case to warn about unknown config keys in `emqx_config`. Use snabbkaffe for logging and tracing.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_config.erl
+++ b/apps/emqx/src/emqx_config.erl
@@ -18,6 +18,7 @@
 -compile({no_auto_import, [get/0, get/1, put/2, erase/1]}).
 -elvis([{elvis_style, god_modules, disable}]).
 -include("logger.hrl").
+-include_lib("snabbkaffe/include/snabbkaffe.hrl").
 
 -export([
     init_load/1,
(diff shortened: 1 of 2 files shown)


apps/emqx/src/emqx_config.erl
```diff
@@ -764,3 +767,19 @@ atom_conf_path(Path, ExpFun, OnFail) ->
                     error(Err)
             end
     end.
+
+warning_deprecated_rook_key(RootNames, RawConf) ->
```
lafirest, line 771: maybe `warning_unknown_root_key`, and `rook` is a typo

HJianBo (approved):

lafirest (approved):

PR #10501, opened 2023-04-24
feat(ft): properly propagate config updates

keynslug:
Fixes [EMQX-9521](https://emqx.atlassian.net/browse/EMQX-9521).

## Summary
Ensure that:
* Storage config might be removed.
* Local FS GC process is set up when Local FS storage is configured.
* Local FS GC process gets its timer reset on config updates.
* Storage / exporter gets chosen based on `type` only.
* Exporter config updates propagated as before.

Also employ `emqx_ft_schema:translate/1` instead of duplicating defaults where applicable.

## PR Checklist

- [ ] ~~Schema changes are backward compatible~~

### DIFF
--- a/apps/emqx_ft/src/emqx_ft_schema.erl
+++ b/apps/emqx_ft/src/emqx_ft_schema.erl
@@ -25,6 +25,8 @@
 
 -export([schema/1]).
 
+-export([translate/1]).
+
 -type json_value() ::
     null
     | boolean()
@@ -82,13 +84,25 @@ fields(file_transfer) ->
             )},
         {storage,
             mk(
-                hoconsc:union([
-                    ref(local_storage)
-                ]),
+                hoconsc:union(
+                    fun
+                        (all_union_members) ->
+                            [
+                                % NOTE: by default storage is disabled
+                                undefined,
+                                ref(local_storage)
+                            ];
+                        ({value, #{<<"type">> := <<"local">>}}) ->
+                            [ref(local_storage)];
+                        ({value, #{<<"type">> := _}}) ->
+                            throw(#{field_name => type, expected => "local"});
+                        (_) ->
+                            [undefined]
+                    end
+                ),
                 #{
                     required => false,
-                    desc => ?DESC("storage"),
-                    default => default_storage()
+                    desc => ?DESC("storage")
                 }
             )}
     ];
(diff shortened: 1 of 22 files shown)


apps/emqx_ft/src/emqx_ft_schema.erl
```diff
@@ -82,13 +84,25 @@ fields(file_transfer) ->
...
             mk(
-                hoconsc:union([
-                    ref(local_storage)
-                ]),
+                hoconsc:union(
+                    fun
+                        (all_union_members) ->
+                            [
+                                % NOTE: by default storage is disabled
+                                undefined,
```
keynslug, line 92: This feels hacky, yet I didn't found a clearer way to tell hocon "there's no default for this union". This is due to the need to be able to "unset" file transfer storage through config, because for me it felt strange that we're currently unable to "unset" the config after setting it up (see `t_remove_restore_config/1` testcase).

Q: Do we really need that?

apps/emqx_ft/test/emqx_ft_conf_SUITE.erl
```diff
@@ -85,3 +86,86 @@ t_update_config(_Config) ->
...
+                <<"$file/f2/init">>,
+                emqx_json:encode(emqx_ft:encode_filemeta(#{name => "f2", size => 42})),
+                1
+            )
+        ),
+        fun(Trace) ->
+            ?assertMatch(
+                [#{transfer := {ClientId, <<"f2">>}, reason := {error, disabled}}],
+                ?of_kind("store_filemeta_failed", Trace)
+            )
```
keynslug, line 130: This is also imperfect IMO. I'm currently thinking that ideally we'd want to give users an ability to "turn off" file transfer feature, which in turn would just remove relevant `emqx_ft` hooks.
Q: Do you think the same? How do we do that "turning off" given that `emqx_conf:remove([file_transfer])` is prohibited? Special `enabled` flag in the config schema?

savonarola:
With file transfer disabling, I think we may follow the common pattern, like in:

https://github.com/emqx/emqx/blob/v5.0.23/apps/emqx_prometheus/src/emqx_prometheus_schema.erl#L78
https://github.com/emqx/emqx/blob/v5.0.23/apps/emqx_retainer/src/emqx_retainer_schema.erl#L39

That is, to have `enable` field.
A downside seems to be that we can't specify a union schema for the root key (`file_transfer :: #{enable := false} | {enable := true, storage := ...}`). So we will have  `enable = false` together with some unused values, like we have for prometheus:
```
  prometheus =>
      #{enable => false,headers => [],interval => 15000,
        job_name => <<"${name}/instance/${name}~${host}">>,
        mnesia_collector => enabled,
        push_gateway_server => "http://127.0.0.1:9091",
        vm_dist_collector => enabled,vm_memory_collector => enabled,
        vm_msacc_collector => enabled,
        vm_statistics_collector => enabled,
        vm_system_info_collector => enabled},
```
But I think it os OK to follow the common pattern.

apps/emqx_ft/test/emqx_ft_conf_SUITE.erl
keynslug: Made a global killswitch, see https://github.com/emqx/emqx/pull/10501/commits/d190398b51861dc6d2a930cdb4f61b67b66ebcfd.

apps/emqx_ft/src/emqx_ft_schema.erl
savonarola: Having `enable`, do we still need this?

keynslug: I'm leaning to _no, we don't_. This will leave a couple of code paths untested since we'll no longer be able to change storage to something other than `local`, but I guess it's fine price to pay for making configuration schema simpler.

savonarola (approved):

PR #10517, opened 2023-04-25
chore(kafka_producer): rename `kafka` to `kafka_producer`

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-9717

# Note

This will require some change in the frontend, as it apparently defaults to Kafka Consumer when it receives a Kafka bridge of type `kafka_producer`.  https://github.com/emqx/emqx-dashboard5/pull/1293 seemed to work locally.

### <samp>🤖 Generated by Copilot at 8a32517</samp>

This pull request renames the `kafka` bridge type to `kafka_producer` in the `emqx_bridge_kafka` and `emqx_ee_bridge` applications, to distinguish it from the `kafka_consumer` type and to support Hocon aliases. It also updates the code, comments, tests, and configuration files accordingly.

## PR Checklist

- [X] Change log has been added to `changes/{ce,ee}/(feat|perf|fix)-<PR-id>.en.md` files
- [X] Schema changes are backward compatible

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge.erl
+++ b/apps/emqx_bridge/src/emqx_bridge.erl
@@ -55,10 +55,7 @@
     T == gcp_pubsub;
     T == influxdb_api_v1;
     T == influxdb_api_v2;
-    %% TODO: rename this to `kafka_producer' after alias support is
-    %% added to hocon; keeping this as just `kafka' for backwards
-    %% compatibility.
-    T == kafka;
+    T == kafka_producer;
     T == redis_single;
     T == redis_sentinel;
     T == redis_cluster;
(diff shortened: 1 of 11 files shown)


apps/emqx_bridge/src/emqx_bridge.erl
```diff
@@ -75,6 +72,16 @@
 
 load() ->
     Bridges = emqx:get_config([bridges], #{}),
+    %% When using Hocon aliases, the raw config still holds the legacy
```
savonarola, line 75: It is not quite clear why we should do this: it looks like hocon alias abstraction leakage 🤔  

>the raw config still holds the legacy type name
Isn't this always a case when hocon aliases are used? 🤔

savonarola: I mean, since aliases are a publicly available feature, any contributor may start to use them in the schemas. Trying to maintain some level of "raw config consistency" everywhere would be a nightmare 😓 
Shouldn't we better discourage raw config usage?

thalesmg: I suspect it's related to this (still looking into it):

https://github.com/emqx/emqx/blob/567413389cdd3e33354508ede0e7aef3a24526d1/apps/emqx_bridge/src/emqx_bridge.erl#L272

PR #10533, opened 2023-04-26
fix(buffer_worker): avoid sending late reply messages to callers (r5.0)

thalesmg:
Port of #10455  to `release-50`

Fixes https://emqx.atlassian.net/browse/EMQX-9635

During a sync call from process `A` to a buffer worker `B`, its call to the underlying resource `C` can be very slow.  In those cases, `A` will receive a timeout response and expect no more messages from `B` nor `C`.  However, prior to this fix, if `B` is stuck in a long sync call to `C` and then gets its response after `A` timed out, `B` would still send the late response to `A`, polluting its mailbox.

### <samp>🤖 Generated by Copilot at 50485b9</samp>

Fixed a noisy error log caused by late replies from slow synchronous calls to external resources. Exported and used `reply_call/2` function in `emqx_resource_buffer_worker` to handle late replies gracefully. Added a test case and a changelog entry for the fix.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -52,7 +52,7 @@
 
 -export([queue_item_marshaller/1, estimate_size/1]).
 
--export([handle_async_reply/2, handle_async_batch_reply/2]).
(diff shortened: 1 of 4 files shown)


zmstone (approved):
Thank you.

PR #10536, opened 2023-04-26
0118 for better test coverage

zmstone:


ieQu1 (dismissed):

apps/emqx_machine/src/emqx_machine.erl
```diff
@@ -61,17 +61,6 @@ set_backtrace_depth() ->
 is_ready() ->
     emqx_machine_terminator:is_running().
 
--if(?OTP_RELEASE > 22).
```
lafirest, line 64: So we won't warn users anymore even if it using an unsupported OTP version?

lafirest (approved):

apps/emqx_machine/src/emqx_machine.erl
zmstone: it won't even compile on 22 now.

thalesmg (approved):

PR #10547, opened 2023-04-27
0427 default configs in schema

zmstone:
Fixes EMQX-9770

Move all the default value of the configs which can be changed from the dashboard UI to schema source code.
Default values in schema have the lowest priority. including

* `listeners`
* `authorization.sources`
* `log`

### <samp>🤖 Generated by Copilot at 17d3b19</samp>

This pull request refactors the configuration system of EMQX to use HOCON format and environment variables, and to avoid duplication and inconsistency. It moves the default listener configurations from `emqx.conf` to `emqx_schema.erl`, improves the handling of SSL certificate files in `emqx_tls_lib.erl`, and updates the comment in `emqx_conf.conf` to explain the new mechanism and provide documentation links.

## PR Checklist


apps/emqx/test/emqx_schema_tests.erl
```diff
@@ -513,3 +513,81 @@ url_type_test_() ->
             typerefl:from_string(emqx_schema:url(), <<"">>)
         )
     ].
+
+env_test_() ->
+    Do = fun(Input) -> emqx_schema:naive_env_interpolation(Input) end,
```
thalesmg, line 518: Tiny nit:
```suggestion
    Do = fun emqx_schema:naive_env_interpolation/1,
```

apps/emqx/src/emqx_schema.erl
```diff
@@ -3083,3 +3089,104 @@ assert_required_field(Conf, Key, ErrorMessage) ->
...
+    case Value =/= false andalso Value =/= "" of
+        true ->
+            {ok, Value};
+        false ->
+            special_env(Name)
+    end.
+
+-ifdef(TEST).
+special_env("EMQX_ETC_DIR") ->
+    filename:join([code:lib_dir(emqx), etc]).
```
thalesmg, line 3189: It seems like this should be:

```suggestion
    {ok, filename:join([code:lib_dir(emqx), etc])}.
```

to keep the expected type in l. 3158

apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -1312,3 +1352,15 @@ validator_string_re(Val, RE, Error) ->
...
         _:_ -> {error, Error}
     end.
+
+default_log_handlers() ->
+    #{
+        <<"file_handlers">> =>
+            #{
+                <<"default">> =>
+                    #{
+                        <<"level">> => <<"info">>,
```
thalesmg, line 1362: Here, `default` has level `info`, but [here](https://github.com/emqx/emqx/pull/10547/files#diff-49e4178ff28f46d34760aa4181bc135529d8bbdcffedd491fd4927f85b9ace1dR880) it is `warning`.

Perhaps this inner map containing `default` could be extracted to another function and used in both places to avoid divergence?

zmstone: this was stale code forgot to delete. it's already deleted in my latest push.

apps/emqx/src/emqx_schema.erl
```diff
@@ -3083,3 +3089,105 @@ assert_required_field(Conf, Key, ErrorMessage) ->
...
+        true ->
+            {ok, Value};
+        false ->
+            special_env(Name)
+    end.
+
+-ifdef(TEST).
+%% when running tests, we need to mock the env variables
+special_env("EMQX_ETC_DIR") ->
+    {ok, filename:join([code:lib_dir(emqx), etc])}.
```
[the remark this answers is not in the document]
zmstone: hmm. this is not expected to happen.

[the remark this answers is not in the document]
thalesmg: I think one of the eunit tests is failing because of this :thinking:

zhongwencool (changes requested):

zhongwencool (approved):

apps/emqx/src/emqx_listeners.erl
```diff
@@ -105,19 +112,22 @@ do_list_raw() ->
...
-                end,
-            LConf4 = maps:put(<<"current_connections">>, CurrConn, LConf3),
-            {Type0, LName, LConf4}
+    lists:filtermap(
+        fun
+            ({LName, LConf0}) when is_map(LConf0) ->
+                Bind = parse_bind(LConf0),
+                Running = is_running(Type, listener_id(Type, LName), LConf0#{bind => Bind}),
+                LConf1 = maps:remove(<<"authentication">>, LConf0),
+                LConf3 = maps:put(<<"running">>, Running, LConf1),
```
savonarola, line 121: Nit: `LConf2` is missing

apps/emqx_management/src/emqx_mgmt_api_listeners.erl
```diff
@@ -642,7 +644,7 @@ create(Path, Conf) ->
     wrap(emqx_conf:update(Path, {create, Conf}, ?OPTS(cluster))).
 
 ensure_remove(Path) ->
-    wrap(emqx_conf:remove(Path, ?OPTS(cluster))).
+    wrap(emqx_conf:update(Path, emqx_schema:tombstone(), ?OPTS(cluster))).
```
savonarola, line 647: Nit: Maybe add `emqx_conf:tombstone(KeyPath, Opts)` (since we have e.g. emqx_conf:remove/2)

zmstone: that'll introduce a new bpapi ?

zmstone: sorry, i forgot it's already added, just forgot to call it.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -61,8 +64,12 @@
 -export([certs_dir/2]).
 -endif.
 
+-type listener_id() :: atom() | binary().
+
 -define(CONF_KEY_PATH, [listeners, '?', '?']).
 -define(TYPES_STRING, ["tcp", "ssl", "ws", "wss", "quic"]).
+-define(MARK_DEL, marked_for_deletion).
+-define(MARK_DEL_BIN, <<"marked_for_deletion">>).
```
savonarola: Should we emphasize that `?MARK_DEL` is a _command_ (that should never appear in a translated config), while `?MARK_DEL_BIN` is a raw config _value_ preventing population? That is, to have `mark_for_deletion` atom for command and `<<"marked_for_deletion">>` as a raw conf value?

zmstone: Make sense.
I'll update the macro in a follow up PR.

zmstone: Maybe:
`?CONFIG_COMMAND_MARK_DEL`
`?CONFIG_VALUE_MARK_DEL` ?

apps/emqx/src/emqx_schema.erl
```diff
@@ -3083,3 +3099,138 @@ assert_required_field(Conf, Key, ErrorMessage) ->
...
+special_env(_Name) ->
+    %% only in tests
+    error.
+-else.
+special_env(_Name) -> error.
+-endif.
+
+%% The tombstone atom.
+tombstone() ->
+    marked_for_deletion.
```
savonarola: Curious: why do we have tombstone _atom_ for a value that resides only in the raw config?
Can (should?) this value leak to translated configs?
Should we separate tombstone commands from tombstone values to avoid confusion 🤔 ?

savonarola: A related question to clarify things:
Conceptually, how do we want to deal with tombstones?
* to make users deal with tombstones themselves completely, returning tombstones as values `marked_for_deletion  = emqx_config:get([something, tombstoned])`
* to postprocess parsed config when returning, removing keys with tombstone values (that is not the case currently)
* not to have tombstone values at all in the translated config

zmstone: Sure. I'll use another atom for the config update request.
This atom here (in schema) however is used as a `singleton` type.

The value has to be kept for default listener in raw configs, but I have not found a good solution to avoid keeping it in checked configs (which seem to be harmless anyway).

apps/emqx/src/emqx_listeners.erl
savonarola: 😓 The confusion is that it is unclear from the naming, whether we consider `MARK_DEL` to be a special update command or a special value.
* If we treat it as a real value, then seems that the convention that `update_config` works with either commands or raw configs is violated
* If we treat it as a command, then it is a bit confusing to have command with the same name as value 😅

savonarola (approved):

PR #10555, opened 2023-04-27
chore(ft): add s3 exporter tests

savonarola:
Fixes [EMQX-9519](https://emqx.atlassian.net/browse/EMQX-9519)

I made a special suite for the s3 exporter finally because most of the `emqx_ft` suite actually does not relate to exporting.

Also:
* env helpers (`set_special_configs`) deal with _raw configs_ so I expressed that explicitly
* added missing stop actions on ft application stop
* also tried to make start/stop semantics a bit more straightforward in some places. It seemed to be a bit difficult to operate the conjunction of flags (storage defined/undefined + enable true/false + application started/stopped) without that 😓 

### <samp>🤖 Generated by Copilot at d61964d</samp>

This pull request adds and improves test cases and helpers for the emqx_ft (file transfer) application, especially for the S3 exporter feature. It also updates the emqx_ft configuration and the emqx_s3 test helpers to support the minio service as a mock S3 server. The affected files are `apps/emqx_ft/test/emqx_ft_test_helpers.erl`, `apps/emqx_ft/docker-ct`, `apps/emqx_ft/src/emqx_ft_conf.erl`, `apps/emqx_ft/test/emqx_ft_storage_exporter_s3_SUITE.erl`, `apps/emqx_ft/test/emqx_ft_storage_fs_SUITE.erl`, and `apps/emqx_s3/test/emqx_s3_test_helpers.erl`.

## PR Checklist


apps/emqx_ft/test/emqx_ft_conf_SUITE.erl
```diff
@@ -34,7 +34,9 @@ end_per_suite(_Config) ->
 init_per_testcase(_Case, Config) ->
     % NOTE: running each testcase with clean config
     _ = emqx_config:save_schema_mod_and_names(emqx_ft_schema),
-    ok = emqx_common_test_helpers:start_apps([emqx_conf, emqx_ft], fun(_) -> ok end),
+    ok = emqx_common_test_helpers:start_apps(
+        [emqx_conf, emqx_ft], emqx_ft_test_helpers:env_handler(Config)
+    ),
```
keynslug, line 39: This will change the meaning of the successful test passes here. Before they were specifically made to run without any initial `emqx_ft` configuration (see NOTE above), to show how things work when someone enables `emqx_ft` at runtime. I'd vote at least for keeping that in some form, say, in a separate test group.

apps/emqx_ft/src/emqx_ft_conf.erl
```diff
@@ -115,23 +119,26 @@ pre_config_update(_, Req, _Config) ->
...
+        #{enable := true, storage := Storage} ->
+            ok = emqx_ft_storage:on_config_update(undefined, Storage),
+            ok = emqx_ft:hook();
+        _ ->
+            ok
+    end.
+
+stop() ->
+    ok = emqx_ft:unhook(),
+    ok = emqx_ft_storage:on_config_update(emqx_config:get([file_transfer, storage]), undefined).
```
keynslug, line 144:
```suggestion
    ok = emqx_ft_storage:on_config_update(storage(), undefined).
```

apps/emqx_ft/src/emqx_ft_conf.erl
```diff
@@ -115,23 +119,26 @@ pre_config_update(_, Req, _Config) ->
...
+on_config_update(#{enable := false}, #{enable := true, storage := NewStorage}) ->
+    ok = emqx_ft_storage:on_config_update(undefined, NewStorage),
     ok = emqx_ft:hook();
-on_config_update(enable, _, false) ->
-    ok = emqx_ft:unhook().
+on_config_update(#{enable := true, storage := OldStorage}, #{enable := true, storage := NewStorage}) ->
+    ok = emqx_ft_storage:on_config_update(OldStorage, NewStorage).
+
+maybe_start() ->
+    case emqx_config:get([file_transfer]) of
```
keynslug, line 134: Wondering if we're sure it won't fail with an exception when `emqx_ft` is not configured (i.e. no `file_transfer` root in the config).

apps/emqx_ft/src/emqx_ft_conf.erl
```diff
@@ -115,23 +119,26 @@ pre_config_update(_, Req, _Config) ->
...
-    ).
-
-on_config_update(_, Config, Config) ->
+on_config_update(#{enable := false}, #{enable := false}) ->
     ok;
-on_config_update(storage, OldConfig, NewConfig) ->
-    ok = emqx_ft_storage:on_config_update(OldConfig, NewConfig);
-on_config_update(enable, _, true) ->
+on_config_update(#{enable := true, storage := OldStorage}, #{enable := false}) ->
+    ok = emqx_ft_storage:on_config_update(OldStorage, undefined),
```
keynslug, line 125: Then we also need to update `enable` field description, because with this change it also affects API availability and background activities (e.g. GC). And we also probably need to reply with `503`s in the API handlers when file transfer is disabled. Remotely related: we then also need to decide who is responsible for cleaning up the resources (namely, disk) when file transfer is disabled, and how they do that.

I kinda see the point but wanted to avoid giving the flag these semantics for the sake of simplicity of implementation.

apps/emqx_ft/test/emqx_ft_test_helpers.erl
```diff
@@ -41,32 +44,47 @@ stop_additional_node(Node) ->
...
 
 load_config(Config) ->
-    emqx_common_test_helpers:load_config(emqx_ft_schema, #{file_transfer => Config}).
+    emqx_common_test_helpers:load_config(emqx_ft_schema, #{<<"file_transfer">> => Config}).
+
+translated_config(Config) ->
+    #{file_transfer := TranslatedConf} = hocon_tconf:check_plain(
+        emqx_ft_schema, #{<<"file_transfer">> => Config}, #{atom_key => true}, [file_transfer]
+    ),
+    TranslatedConf.
```
keynslug, line 80: Looks like existing `emqx_ft_schema:translate/1` does the same thing?

apps/emqx_ft/test/emqx_ft_storage_exporter_s3_SUITE.erl
```diff
@@ -0,0 +1,132 @@
...
+    ClientId = ?config(clientid, Config),
+
+    FileId = <<"🌚"/utf8>>,
+    Name = "cool_name",
+    Data = <<"data"/utf8>>,
+
+    OldConfig = emqx_config:get([file_transfer]),
+    NewConfig = emqx_utils_maps:deep_put([storage, exporter, bucket], OldConfig, "invalid-bucket"),
+
+    ok = emqx_ft_conf:on_config_update(OldConfig, NewConfig),
```
keynslug, line 110: Q: Why do it like this, instead of `emqx_conf:update([file_transfer, storage, exporter, bucket], <<"invalid-bucket">>, #{})`?

keynslug (commented):
👍🏼

apps/emqx_ft/test/emqx_ft_conf_SUITE.erl
savonarola: Does starting with an empty env configuration handler model an existing situation? The app cannot work without `emqx_conf` (because only emqx_ft schema is injected into it), and when started with `emqx_conf`, the config will be filled either with default or provided schema.

One of the ways to achieve no initial configuration is to 1) stop the app 2) delete its config _explicitly_ 3) try to start

I don't think we should handle such a situation 😅 
I prefer not to handle "empty" config and fail explicitly, because this is a situation we shouldn't normally reach, and it is a bug if we reach it 🤔

keynslug: > Does starting with an empty env configuration handler model an existing situation?

Your answer suggests that this would model situation _emqx starts with default configuration_. Which on first look sounds testcase-worthy.

> this is a situation we shouldn't normally reach

I agree with the general idea. 👍🏼 But still wondering what we'll see if the user will put `file_transfer = null` in the config file.

savonarola: >this would model situation emqx starts with default configuration
Maybe this is the fault of our test helpers but this is not the case: starting with `emqx_common_test_helpers:start_apps` and empty handler _does not_ model starting EMQX with empty config because `emqx_common_test_helpers:start_apps` does not make population with default configs.

When we start EMQX without `file_transfer` at all or with `file_transfer = null` population with default values happens. So instead, some `env_handler` is needed to emulate this initial default values.

apps/emqx_ft/test/emqx_ft_storage_exporter_s3_SUITE.erl
savonarola: ✅ 
Didn't want to start `emqx_conf`

apps/emqx_ft/src/emqx_ft_conf.erl
savonarola: Regarding cleanup, it also seems unintuitive for a disabled app to have (potentially resource-consuming) background tasks.

apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
```diff
@@ -83,13 +84,14 @@
...
--type filter() :: fun((request(), request_meta()) -> filter_result()).
+-type filter() :: fun((request(), request_meta()) -> filter_result()) | undefined.
 
 -type spec_opts() :: #{
     check_schema => boolean() | filter(),
     translate_body => boolean(),
     schema_converter => fun((hocon_schema:schema(), Module :: atom()) -> map()),
-    i18n_lang => atom() | string() | binary()
+    i18n_lang => atom() | string() | binary(),
+    filter => filter()
```
keynslug, line 94: Also, I understand that this is preexisting naming, but it's a bit confusing now that it's part of module interface. Have no good ideas how to call it though, _interceptor_ comes into mind, or something along the lines of `before_request` / `on_request`.

apps/emqx_ft/src/emqx_ft_conf.erl
keynslug: Agree, hence former `enable` description. Now that is complete killswitch, what do we do with segments storage? Do we need to clean it up completely, or is it up to the user? What do we do with exports in the local fs? Since they are now unreachable through the API, do we also need to wipe them clean? If not, how users are supposed to get them? What to do with file expiration? I think we'd need some followup tickets for those.

apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
savonarola: `filter`s belong to `minirest` domain, and implement the very filter protocol defined by it; since we do not add any additional abstraction over it maybe it is better to preserve the name for intuitiveness and clarity

apps/emqx_ft/src/emqx_ft_conf.erl
savonarola: I don't think we should do any implicit cleanup on app disable/stop: a user may want to enable the app back soon.
Maybe we should introduce a cli method to make explicit cleanup even for a disabled app.

apps/emqx_ft/test/emqx_ft_storage_exporter_s3_SUITE.erl
```diff
@@ -0,0 +1,199 @@
...
+expected_page_count(PageSize, Total) ->
+    case Total rem PageSize of
+        0 -> Total div PageSize;
+        _ -> Total div PageSize + 1
+    end.
+
+file_pages(Query) ->
+    case emqx_ft_storage:files(Query) of
+        {ok, #{items := Items, cursor := NewCursor}} ->
+            [Items] ++ file_pages(Query#{cursor => NewCursor});
```
keynslug, line 181: Seems that I introduced an inconsistency in my last PR. According to types and existing fs implementation, here should be (see https://github.com/emqx/emqx/blob/ed3756ea0928965c628a518015b5c5d0c79f01ef/apps/emqx_ft/src/emqx_ft_storage.erl#L52-L53).
```suggestion
            [Items] ++ file_pages(Query#{following => NewCursor});
```

Yet S3 implementation of this thing violate this type. 😅 
https://github.com/emqx/emqx/blob/a9866fede4cb4feba41a78d79885843ade0ad729/apps/emqx_ft/src/emqx_ft_storage_exporter_s3.erl#L152

savonarola: fixed in the exporter and the tests

keynslug (approved):

PR #10556, opened 2023-04-28
fix(emqx_connector_http): wrap and unwrap auth headers

sstrigler:
This so we don't have sensitive information in a process' state

Fixes [EMQX-9795](https://emqx.atlassian.net/browse/EMQX-9795)

### <samp>🤖 Generated by Copilot at 97cc177</samp>

This pull request enhances the security and functionality of the emqx_connector application. It adds encryption and decryption of sensitive HTTP header values using `emqx_secret` in `emqx_connector_http.erl`. It also updates the version number in `emqx_connector.app.src`.

## PR Checklist

### DIFF
--- a/apps/emqx_connector/src/emqx_connector_http.erl
+++ b/apps/emqx_connector/src/emqx_connector_http.erl
@@ -473,7 +473,7 @@ preprocess_request(
         method => emqx_plugin_libs_rule:preproc_tmpl(to_bin(Method)),
         path => emqx_plugin_libs_rule:preproc_tmpl(Path),
         body => maybe_preproc_tmpl(body, Req),
-        headers => preproc_headers(Headers),
+        headers => wrap_auth_header(preproc_headers(Headers)),
         request_timeout => maps:get(request_timeout, Req, 30000),
         max_retries => maps:get(max_retries, Req, 2)
     }.
@@ -502,6 +502,36 @@ preproc_headers(Headers) when is_list(Headers) ->
         end,
         Headers
     ).
+
+wrap_auth_header(Headers) ->
+    lists:map(fun maybe_wrap_auth_header/1, Headers).
+
+maybe_wrap_auth_header({[{str, Key}] = StrKey, Val}) ->
+    {_, MaybeWrapped} = maybe_wrap_auth_header({Key, Val}),
+    {StrKey, MaybeWrapped};
+maybe_wrap_auth_header({Key, Val} = Header) when
+    is_binary(Key), (size(Key) =:= 19 orelse size(Key) =:= 13)
+->
+    %% We check the size of potential keys in the guard above and consider only
+    %% those that match the number of characters of either "Authorization" or
+    %% "Proxy-Authorization".
+    case try_bin_to_lower(Key) of
+        <<"authorization">> ->
+            {Key, emqx_secret:wrap(Val)};
+        <<"proxy-authorization">> ->
+            {Key, emqx_secret:wrap(Val)};
+        _Other ->
+            Header
+    end;
+maybe_wrap_auth_header(Header) ->
+    Header.
+
+try_bin_to_lower(Bin) ->
+    try iolist_to_binary(string:lowercase(Bin)) of
+        LowercaseBin -> LowercaseBin
+    catch
+        _:_ -> Bin
+    end.
 
 maybe_preproc_tmpl(Key, Conf) ->
     case maps:get(Key, Conf, undefined) of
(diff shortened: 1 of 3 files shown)


thalesmg (commented):
Maybe we could add a few testcases for this?

sstrigler:
> Maybe we could add a few testcases for this?

Hm, yeah, will do. Apparently it is already tested indirectly by other test suites, otherwise coveralls would have complained. But it's better to make it explicit.

qzhuyan (approved):

ieQu1 (approved):

apps/emqx_connector/test/emqx_connector_http_tests.erl
```diff
@@ -0,0 +1,76 @@
...
+
+-define(MY_SECRET, <<"my_precious">>).
+
+wrap_auth_headers_test_() ->
+    {setup,
+        fun() ->
+            meck:expect(ehttpc_sup, start_pool, 2, {ok, foo}),
+            [ehttpc_sup]
+        end,
+        fun meck:unload/1, fun(_) ->
```
zmstone, line 28: nit
```suggestion
        fun meck:unload/1,
        fun(_) ->
```

apps/emqx_connector/src/emqx_connector_http.erl
```diff
@@ -606,21 +651,13 @@ is_sensitive_key([{str, StringKey}]) ->
 is_sensitive_key(Atom) when is_atom(Atom) ->
```
zmstone, line 651: with the string wrapped, do we still need this ?

apps/emqx_connector/src/emqx_connector_http.erl
```diff
@@ -490,6 +490,48 @@ preproc_headers(Headers) when is_list(Headers) ->
...
+
+try_bin_to_lower(Bin) ->
+    try iolist_to_binary(string:lowercase(Bin)) of
+        LowercaseBin -> LowercaseBin
+    catch
+        _:_ -> Bin
+    end.
+
+maybe_wrap([{str, Val}]) ->
+    [{str, emqx_secret:wrap(Val)}];
```
zmstone, line 527: we are passing `{str, function()}` to `emqx_placeholder:proc_tmpl`, 
this violates the spec of `tmpl_token()`.
maybe update the spec (and add a comment there).

apps/emqx_connector/test/emqx_connector_http_tests.erl
sstrigler: nah, that's `make fmt` that insists on that.

apps/emqx_connector/src/emqx_connector_http.erl
sstrigler: Good catch

apps/emqx_connector/src/emqx_connector_http.erl
sstrigler: Yes, it's two different concepts, `redact`, which uses this, is called at various places where we want to log data and for most parts this happens either before or after wrapping/unwrapping - or even on raw data coming from the actual http client.

apps/emqx_connector/src/emqx_connector_http.erl
sstrigler: Turns out, this was actually a bug. We need to unwrap before calling proc_tmpl. Unlike what I believed no test uses auth-headers.

apps/emqx_connector/src/emqx_connector_http.erl
```diff
@@ -490,6 +490,39 @@ preproc_headers(Headers) when is_list(Headers) ->
...
 
+wrap_auth_header(Headers) ->
+    lists:map(fun maybe_wrap_auth_header/1, Headers).
+
+maybe_wrap_auth_header({[{str, Key}] = StrKey, Val}) ->
+    {_, MaybeWrapped} = maybe_wrap_auth_header({Key, Val}),
+    {StrKey, MaybeWrapped};
+maybe_wrap_auth_header({_Key, Val} = Header) when is_function(Val) ->
+    %% Already wrapped?
+    Header;
```
zmstone, line 501: this is not needed, double wrapping is ok by eqmx_secret

zmstone (approved):

apps/emqx_connector/test/emqx_connector_http_tests.erl
```diff
@@ -0,0 +1,90 @@
...
+is_wrapped(_Other) ->
+    false.
+
+untmpl([{_, V} | _]) -> V.
+
+is_unwrapped_headers(Headers) ->
+    lists:all(fun is_unwrapped_header/1, Headers).
+
+is_unwrapped_header({_, V}) when is_function(V) -> false;
+is_unwrapped_header({_, [{str, _V}]}) -> throw(unecpected_tmpl_token);
```
thalesmg, line 89:
```suggestion
is_unwrapped_header({_, [{str, _V}]}) -> throw(unexpected_tmpl_token);
```

zmstone (approved):

Issue #10587, opened 2023-05-04
The API Key limit does not match the documentation.


HJianBo:
### What happened?

In the document https://www.emqx.io/docs/en/v5.0/getting-started/restrictions.html#reference-list, the API Key limit is 1024, 
<img width="599" alt="image" src="https://user-images.githubusercontent.com/13825269/236116741-827776cd-fb5b-4b17-a4a2-4bd8ca4cdd3d.png">

but in code it is 100 (`emqx_mgmt_auth:create/4`)
<img width="735" alt="image" src="https://user-images.githubusercontent.com/13825269/236116760-ac5c98b4-71aa-4ef7-b645-356afa102127.png">

Reported by: https://askemq.com/t/topic/4712/2

### What did you expect to happen?

-

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
# paste output here
-
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Log files

<details>

</details>

PR #10593, opened 2023-05-04
chore(mgmt): modify the API Key limit to 1024.

HJianBo:
fixes https://github.com/emqx/emqx/issues/10587

Fixes https://emqx.atlassian.net/browse/CI-123

### <samp>🤖 Generated by Copilot at 1c258c4</samp>

Use a macro for API key limit in `emqx_mgmt_auth`. This change refactors the `create` function to use a macro for the maximum number of API keys allowed per user, instead of a hard-coded value. This makes the code more consistent and easier to modify.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_auth.erl
+++ b/apps/emqx_management/src/emqx_mgmt_auth.erl
@@ -45,6 +45,7 @@
 -endif.
 
 -define(APP, emqx_app).
+-define(API_KEY_LIMIT, 1024).
 
 -record(?APP, {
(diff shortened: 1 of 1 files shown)


zhongwencool:
Perhaps changing the document to **100** would be better.

- Users won't need  1024 API keys. Otherwise it's probably design issue. 
- We would also need to add pagination for the GET api_key  API.  Personally, I don't think it's necessary.

PS: This 100 has been confirmed by the product manager before.

PR #10597, opened 2023-05-04
ci: limit gitlint job git fetch depth

zmstone:


JimMoen (approved):

zmstone:
no need for this change.
the job is only triggered on pr, and only the added commits are checked.

PR #10616, opened 2023-05-05
ci: in 'prepare' step, make clean-all before fetch deps

zmstone:


.github/workflows/build_packages.yaml
```diff
@@ -77,7 +77,10 @@ jobs:
           echo "VERSION=$(./pkg-vsn.sh $PROFILE)" >> $GITHUB_OUTPUT
       - name: get_all_deps
         run: |
-          make -C source deps-all
+          pusd source
```
thalesmg, line 80:
```suggestion
          pushd source
```

.github/workflows/build_packages.yaml
```diff
@@ -77,7 +77,10 @@ jobs:
...
+          make clean-all
+          make deps-emqx deps-emqx-enterprise
```
thalesmg, line 82: Since emqx-enterprise is a superset of emqx, maybe only `deps-emqx-enterprise` would be enough?

.github/workflows/build_packages.yaml
```diff
@@ -175,6 +178,8 @@ jobs:
   linux:
     needs: prepare
     runs-on: ${{ matrix.build_machine }}
+    # always run in builder container because the host might have the wrong OPT version etc.
```
thalesmg, line 181:
```suggestion
    # always run in builder container because the host might have the wrong OTP version etc.
```

thalesmg (approved):

id (approved):

PR #10617, opened 2023-05-05
ci: use cd instead of pushd

zmstone:


thalesmg (approved):
For my own information: why pushd/popd is bad in this case?

zmstone:
> For my own information: why pushd/popd is bad in this case?

no such command ;(

PR #10618, opened 2023-05-05
feat(ft-conf): simplify schema of storage / exporter backends

keynslug:
~~Fixes [EMQX-9678](https://emqx.atlassian.net/browse/EMQX-9678)~~

Assumption is this changes will make `emqx_ft` config schema user-friendlier and also more future-proof.

### <samp>🤖 Generated by Copilot at 4995741</samp>

Refactor and simplify the schema and logic for the storage and exporter backends of the file transfer feature. Use references, custom validators, and helper functions to reduce code duplication and improve error handling. Update the test cases and the i18n file to match the new schema.

## PR Checklist

- [ ] ~~Schema changes are backward compatible~~

### DIFF
--- a/apps/emqx_ft/src/emqx_ft_schema.erl
+++ b/apps/emqx_ft/src/emqx_ft_schema.erl
@@ -237,6 +237,10 @@ desc(s3_exporter) ->
     "S3 Exporter settings for the File transfer local storage backend";
 desc(local_storage_segments_gc) ->
     "Garbage collection settings for the File transfer local segments storage";
+desc(local_storage_exporter_backend) ->
+    "Exporter for the local file system storage backend";
+desc(storage_backend) ->
+    "Storage backend settings for file transfer";
 desc(_) ->
     undefined.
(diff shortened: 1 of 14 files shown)


apps/emqx_ft/src/emqx_ft_schema.erl
```diff
@@ -93,36 +93,30 @@ fields(file_transfer) ->
...
+                        <<"local">> => #{}
+                    }
                 }
             )}
     ];
-fields(local_storage) ->
+fields(storage_backend) ->
     [
-        {type,
+        {local,
```
zmstone, line 109: do we plan to add another storage backend in the future ?

savonarola: We discussed a storage that uploads files to S3 on the fly.
We decided not to implement it in the first iterations because it would be quite complicated algorithmically

savonarola (approved):

PR #10640, opened 2023-05-08
ci: ensure git safe dir in build_packages

id:


.github/workflows/build_packages.yaml
```diff
@@ -35,6 +35,7 @@ jobs:
       - name: Get profile to build
         id: get_profile
         run: |
+          git config --global --add safe.directory "$GITHUB_WORKSPACE"
```
thalesmg, line 38: Maybe:

```suggestion
          git config --global --add safe.directory "*"
```
?

thalesmg (approved):

.github/workflows/build_packages.yaml
id: I try to follow the "Principle of least privilege" as much as possible.

id: If this one does not work, I'd like to know why it's not enough.

zmstone (approved):

PR #10641, opened 2023-05-08
chore: bump gproc -> 0.9.0.1

thalesmg:
Includes fix: https://github.com/uwiger/gproc/pull/193

Prior to the fix, when using the `random` pool strategy, one of the workers receives about double the load of other workers, which decreases throughput of bridges like webhook.

Related: https://emqx.atlassian.net/browse/EMQX-9637

### <samp>🤖 Generated by Copilot at 5a1cd01</samp>

Updated the `gproc` dependency to a forked version with a bug fix for the `gproc_dist` module. This change improved the stability and performance of the EMQX cluster feature.

## PR Checklist

### DIFF
--- a/apps/emqx/rebar.config
+++ b/apps/emqx/rebar.config
@@ -24,7 +24,7 @@
 {deps, [
     {emqx_utils, {path, "../emqx_utils"}},
     {lc, {git, "https://github.com/emqx/lc.git", {tag, "0.3.2"}}},
-    {gproc, {git, "https://github.com/uwiger/gproc", {tag, "0.8.0"}}},
+    {gproc, {git, "https://github.com/emqx/gproc", {tag, "0.9.0.1"}}},
     {cowboy, {git, "https://github.com/emqx/cowboy", {tag, "2.9.0"}}},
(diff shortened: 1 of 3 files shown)


apps/emqx/rebar.config
```diff
@@ -24,7 +24,7 @@
 {deps, [
     {emqx_utils, {path, "../emqx_utils"}},
     {lc, {git, "https://github.com/emqx/lc.git", {tag, "0.3.2"}}},
-    {gproc, {git, "https://github.com/uwiger/gproc", {tag, "0.8.0"}}},
+    {gproc, {git, "https://github.com/emqx/gproc", {tag, "0.9.1-emqx0"}}},
```
zmstone, line 27: is the vsn in app.src `0.9.1`, or with a suffix ?

zmstone: nah, it's using `git`. we'll need to use a better app vsn for it.

thalesmg: The problem here is that upstream has not tagged that version yet.  So, if we tag `0.9.1` in our fork, it might be strange when upstream does so.  So I used that suffix to differentiate it.

PR #10674, opened 2023-05-11
Add an experimental feature: support external rule-engine SQL functions

zmstone:
Often there are community users wish for functions that are hard to get prioritized.
This 'eval' func should allow them to hook in their functions in a patch module.

So far, there is no plan to widely announce this feature,
we'll try to see if it is helpful before documenting it.

### DIFF
--- a/apps/emqx_rule_engine/include/rule_engine.hrl
+++ b/apps/emqx_rule_engine/include/rule_engine.hrl
@@ -92,6 +92,8 @@
     ?RAISE(EXP, _ = do_nothing, ERROR)
 ).
 
+-define(RAISE_BAD_SQL(Detail), throw(Detail)).
+
 -define(RAISE(EXP, EXP_ON_FAIL, ERROR),
     fun() ->
         try
@@ -106,3 +108,16 @@
 
 %% Tables
 -define(RULE_TAB, emqx_rule_engine).
+
+%% Allowed sql function provider modules
+-define(DEFAULT_SQL_FUNC_PROVIDER, emqx_rule_funcs).
+-define(IS_VALID_SQL_FUNC_PROVIDER_MODULE_NAME(Name),
+    (case Name of
+        <<"emqx_rule_funcs", _/binary>> ->
+            true;
+        <<"EmqxRuleFuncs", _/binary>> ->
+            true;
+        _ ->
+            false
+    end)
+).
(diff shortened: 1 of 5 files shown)


apps/emqx_rule_engine/src/emqx_rule_funcs.erl
```diff
@@ -1167,3 +1170,25 @@ convert_timestamp(MillisecondsTimestamp) ->
...
+%% -module(my_module).
+%% -epxort([is_my_topic/1]).
+%% %% return true if the topic has 5 levels
+%% is_my_topic(Topic) ->
+%%   emqx_topic:levels(Topic) =:= 5.
+%%
+%% Use it in SQL:
+%% SELECT * FROM t/# WHERE eval('my_module:is_my_topic', topic)
+%%
+eval(ModFun, Args) ->
```
terry-xiaoyu, line 1190: The user may want to pass variable number of args to the customized function. See the implementation of the `sprintf` function at: https://github.com/emqx/emqx/blob/0617a9b11c227a70e48c6da9811ece60274f0322/apps/emqx_rule_engine/src/emqx_rule_funcs.erl#LL1124C12-L1124C12

kjellwinblad: I'm not convinced that this is preferable in this case:

```
eval('mymod:myfun', something, 34324, 24)
eval('mymod:myfun', [something, 34324, 24])
```

I think it is more clear in the second example what the list of parameters to myfun is.

zhongwencool:
We need a white list, otherwise execute init:stop() directly.

kjellwinblad:
I believe this is a useful idea. Here are some suggestions:

I think a more descriptive name would be better. For someone unfamiliar with Erlang, it would be difficult to guess the meaning of `eval('module_name:something', [asd, asd, sad])`.

Here are some alternative names:

* exec_erlang_function
* execute_erlang_function
* apply_erlang_function
* run_erlang_function

exec_beam_function and similar names could also work, but I believe it might be worse, despite being more formally correct, as there are probably more people who know what Erlang is compared to those who know what Beam is.

The name might not be a significant issue, so I'm fine with proceeding with `eval` if you prefer that.

Additionally, I think it would be better to use three parameters: module, function, and parameter list. I don't believe that including the string 'mod:fun' significantly improves readability, and using three parameters would make the function more flexible, particularly if the module or function is determined dynamically.

zmstone:
do we allow `[...]` composing?
used the 'bash' style instead, supports up to 9 args.

kjellwinblad:
> do we allow `[...]` composing? used the 'bash' style instead, supports up to 9 args.

Can you elaborate or give an example?

kjellwinblad (approved):
Looks good to me!

PR #10690, opened 2023-05-12
perf(webhook): add async retries and evaluate reply callback in fresh process

thalesmg:
(Partially) fixes https://emqx.atlassian.net/browse/EMQX-9637

There are two optimizations made to improve the throughput of the webhook bridge.

- `normal` and `{shutdown, normal}` requests are retried indefinetely.

  Since this bridge is called using `async` callback mode, and `ehttpc` frequently returns errors of the form `normal` and `{shutdown, normal}` that are retried "for free" by `ehttpc`, we add this behavior to async requests as well.  Other errors are retried too, but they are not "free": 3 attempts are made at a maximum.

  This is important because, when using buffer workers, we should avoid making them enter the `blocked` state, since that halts all progress and makes throughput plummet.

- The reply callback `emqx_connector_http:reply_delegator` is evaluated in a fresh process.
  
  This surprisingly simple change yields a big performance improvement in throughput.

  While the previous commit achieves ~ 55 k messages / s in throughput under some test conditions (100 k concurrent publishers publishing 1 QoS 1 message per second), the simple change in this commit improves it further to ~ 63 k messages / s.

  Benchmarks indicated that the evaluating one reply function is consistently quite fast (~ 20 µs), which makes this performance gain counterintuitive.  Perhaps, although each call is cheap, `ehttpc` calls several of these in a row when there are several sent requests, and those costs might add up in latency.

This also makes a slight change in the load balancing of requests to `ehttpc` from the webhook bridge: now the routing key is the message's clientid rather than the calling PID.  This should have the same behavior as before, except in the event of a client takeover.  In this case, requests will still be forwarded to the same HTTP connection for the same client (assuming that such connection is still alive when the takeover is complete) rather than a different one.

### <samp>🤖 Generated by Copilot at 4ed859c</samp>

This pull request enhances the HTTP connector with a retry mechanism and fixes a bug related to gproc counters. It also updates some dependencies in `mix.exs` and `rebar.config` to use more reliable sources and versions.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/test/emqx_bridge_webhook_SUITE.erl
+++ b/apps/emqx_bridge/test/emqx_bridge_webhook_SUITE.erl
@@ -23,6 +23,7 @@
 -compile(export_all).
 
 -import(emqx_mgmt_api_test_util, [request/3, uri/1]).
+-import(emqx_common_test_helpers, [on_exit/1]).
 
 -include_lib("eunit/include/eunit.hrl").
 -include_lib("common_test/include/ct.hrl").
@@ -51,6 +52,13 @@ end_per_suite(_Config) ->
 
 suite() ->
     [{timetrap, {seconds, 60}}].
+
+init_per_testcase(_TestCase, Config) ->
+    Config.
+
+end_per_testcase(_TestCase, _Config) ->
+    emqx_common_test_helpers:call_janitor(),
+    ok.
 
 %%------------------------------------------------------------------------------
 %% HTTP server for testing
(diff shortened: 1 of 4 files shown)


apps/emqx/rebar.config
```diff
@@ -24,7 +24,7 @@
 {deps, [
     {emqx_utils, {path, "../emqx_utils"}},
     {lc, {git, "https://github.com/emqx/lc.git", {tag, "0.3.2"}}},
-    {gproc, {git, "https://github.com/emqx/gproc", {tag, "0.9.0.1"}}},
+    {gproc, {git, "https://github.com/uwiger/gproc", {tag, "0.9.1"}}},
```
qzhuyan, line 27: better to sync and use our fork.

thalesmg: Our `0.9.0.1` was tagged just because `0.9.1` was not ready upstream yet.  Both are basically at the same state, just `0.9.0.1` has appup for v4.4.

I talked with @zmstone , and I understood that we wanted to use upstream if possible here. :thinking:

zmstone: Sorry.
For gproc, since we already forked (with appup etc in place), it makes more sense to use our fork instead.

qzhuyan (approved):

yanzhiemq:
_

yanzhiemq:
_

PR #10692, opened 2023-05-12
ci: update performance test

id:


.github/workflows/performance_test.yaml
```diff
@@ -77,38 +76,37 @@ jobs:
...
+      id: scenario_1on1
       working-directory: ./tf-emqx-performance-test
+      timeout-minutes: 60
+      env:
+        TF_VAR_bench_id: "${{ needs.prepare.outputs.BENCH_ID }}/1on1"
+        TF_VAR_use_emqttb: 1
+        TF_VAR_use_emqtt_bench: 0
+        TF_VAR_emqttb_instance_count: 2
+        TF_VAR_emqttb_instance_type: "c5.large"
+        TF_VAR_emqttb_scenario: "@pub --topic 't/%n' --pubinterval 10ms --qos 1 --publatency 50ms --size 16 --num-clients 25000 @sub --topic 't/%n' --num-clients 25000"
```
id, line 89: @ieQu1 how can I fix it so that each pub client publishes a message every second?

qzhuyan (approved):

.github/workflows/performance_test.yaml
thalesmg: I think it should be `--pubinterval 1s`

id: Dmitrii suggested this one:

```
./emqttb @pubsub_fwd -n 50_000 --pub-qos 1 --sub-qos 1 
```

I'll try out both tomorrow and will update the workflow.

PR #10705, opened 2023-05-15
feat(tls-partial-chains): update schema

qzhuyan:
followup review comments in #10673 

Since we now support cacert refresh and makes the logic a little complicated, 
detailed description of the feature is moved to emqx-docs PR:  https://github.com/emqx/emqx-docs/pull/1893


HJianBo (approved):
LGTM

PR #10717, opened 2023-05-16
fix(buffer_worker): avoid setting flush timer when inflight is full

thalesmg:
# Targeting `release-50`

Fixes https://emqx.atlassian.net/browse/EMQX-9902

When the buffer worker inflight window is full, we don’t need to set a timer to flush the messages again because there’s no more room, and one of the inflight windows will flush the buffer worker by calling `flush_worker`.

Currently, we do set the timer on such situation, and this fact combined with the default batch time of 0 yields a busy loop situation where the CPU spins a lot while inflight messages do not return.

This original change uncovered another bug that was hidden by it:
`maybe_flush_after_async_reply` was sending a message to the wrong
PID.  It was sending a message to `self()` meaning to target a buffer
worker, but `self()` in that context is never the buffer worker, it's
the connector's worker.

This change also revealed a race condition where the buffer workers
could stop flushing messages.  So we piggy-backed on the atomic update
of the table size count to check if the buffer worker should be poked
to continue flushing.  This allows us to get rid of
`maybe_flush_after_async_reply` altogether.

### <samp>🤖 Generated by Copilot at a77f6e5</samp>

Optimize resource buffer worker performance by removing redundant timer calls. This affects the file `emqx_resource_buffer_worker.erl` and the function `ensure_flush_timer/1`.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -70,18 +70,6 @@
 -define(RETRY_IDX, 3).
 -define(WORKER_MREF_IDX, 4).
 
--define(ENSURE_ASYNC_FLUSH(InflightTID, EXPR),
-    (fun() ->
-        IsFullBefore = is_inflight_full(InflightTID),
-        case (EXPR) of
-            blocked ->
-                ok;
-            ok ->
-                ok = maybe_flush_after_async_reply(IsFullBefore)
-        end
-    end)()
-).
-
 -type id() :: binary().
 -type index() :: pos_integer().
 -type expire_at() :: infinity | integer().
@@ -337,7 +325,8 @@ resume_from_blocked(Data) ->
                     {next_state, running, Data}
             end;
         {expired, Ref, Batch} ->
-            IsAcked = ack_inflight(InflightTID, Ref, Id, Index),
+            WorkerPid = self(),
+            IsAcked = ack_inflight(InflightTID, Ref, Id, Index, WorkerPid),
             IsAcked andalso emqx_resource_metrics:dropped_expired_inc(Id, length(Batch)),
             ?tp(buffer_worker_retry_expired, #{expired => Batch}),
             resume_from_blocked(Data);
@@ -389,7 +378,8 @@ retry_inflight_sync(Ref, QueryOrBatch, Data0) ->
             {keep_state, Data0, {state_timeout, ResumeT, unblock}};
         %% Send ok or failed but the resource is working
         {ack, PostFn} ->
-            IsAcked = ack_inflight(InflightTID, Ref, Id, Index),
+            WorkerPid = self(),
+            IsAcked = ack_inflight(InflightTID, Ref, Id, Index, WorkerPid),
             %% we need to defer bumping the counters after
             %% `inflight_drop' to avoid the race condition when an
             %% inflight request might get completed concurrently with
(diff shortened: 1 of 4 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -337,7 +325,8 @@ resume_from_blocked(Data) ->
                     {next_state, running, Data}
             end;
         {expired, Ref, Batch} ->
-            IsAcked = ack_inflight(InflightTID, Ref, Id, Index),
+            WorkerPid = self(),
+            IsAcked = ack_inflight(InflightTID, Ref, Id, Index, WorkerPid),
```
savonarola, line 329: m.b. an inline fun `ack_inflight_from_worker(InflightTID, Ref, Id, Index)`?

zmstone (approved):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
thalesmg: I have another PR underway that will also touch this function, I'll try to do it in that PR.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1441,7 +1422,11 @@ ack_inflight(InflightTID, Ref, Id, Index) ->
             [] ->
                 {0, false}
         end,
-    ok = dec_inflight_remove(InflightTID, Count, Removed),
+    FlushCheck = dec_inflight_remove(InflightTID, Count, Removed),
+    case FlushCheck of
+        continue -> ok;
+        flush -> ?MODULE:flush_worker(WorkerPid)
```
savonarola, line 1428: Nit: not quite clear how `continue` opposes to `flush` 😅

thalesmg: Would `no_flush` or `dont_flush` be better?  :thinking:

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -596,13 +585,14 @@ do_flush(
...
                 true when IsUnrecoverableError ->
-                    ack_inflight(InflightTID, Ref, Id, Index);
+                    ack_inflight(InflightTID, Ref, Id, Index, WorkerPid);
                 true ->
                     ok;
                 false ->
-                    ack_inflight(InflightTID, Ref, Id, Index)
+                    ack_inflight(InflightTID, Ref, Id, Index, WorkerPid)
             end,
             {Data1, WorkerMRef} = ensure_async_worker_monitored(Data0, Result),
```
savonarola, line 597: Q: `ensure_async_worker_monitored` we also have `WorkerPid`s,  couldn't be this confusing?

savonarola, line 597: Q: In `ensure_async_worker_monitored` we also have `WorkerPid`s, couldn't be that confusing?

thalesmg: Good point, there are workers all around :sweat_smile: 
I think the differentiating point is that we have _buffer_ workers (the current module) and _async_ workers (from the connectors).  I'll add the `Async` modifier to the variable name to help with that differentiation.

thalesmg: Fixed here: https://github.com/emqx/emqx/pull/10740

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
thalesmg: Fixed here: https://github.com/emqx/emqx/pull/10740

PR #10740, opened 2023-05-18
perf(buffer_worker): update metrics periodically rather than immediately (v5.0)

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-9905

Since calling `telemetry` is costly in a hot path, we instead collect
metrics inside the buffer workers state and periodically flush them,
rather than immediately as events happen.

## Summary
copilot:summary

## PR Checklist


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -778,91 +813,162 @@ handle_query_result(Id, Result, HasBeenSent) ->
...
+                emqx_resource_metrics:retried_failed_inc(Id, Val);
+            (success, Val) ->
+                emqx_resource_metrics:success_inc(Id, Val);
+            (retried_success, Val) ->
+                emqx_resource_metrics:retried_success_inc(Id, Val);
+            (dropped_resource_not_found, Val) ->
+                emqx_resource_metrics:dropped_resource_not_found_inc(Id, Val);
+            (dropped_resource_stopped, Val) ->
+                emqx_resource_metrics:dropped_resource_stopped_inc(Id, Val)
+        end,
```
zmstone, line 957: write it as a static fun should be easier to read (and more performant).

thalesmg: I think it'll still require a closure, since we're capturing `Id` from the context.  :thinking:

thalesmg: Alternatively, we could use `maps:iterator` and recurse over that.  I think I'll do that to avoid the closure.

zmstone (approved):

thalesmg:
Had to resolve a conflict.

qzhuyan (approved):

PR #10744, opened 2023-05-18
fix(evacuation): handle expire interval correctly

savonarola:
Fixes EMQX-9928

### <samp>🤖 Generated by Copilot at 03943ad</samp>

Use `EXPIRE_INTERVAL_INFINITE` macro for expiry interval of MQTT sessions. This pull request defines a macro `EXPIRE_INTERVAL_INFINITE` in `emqx_channel.hrl` and replaces the literal value `?UINT_MAX` with it in several modules that deal with the expiry interval of MQTT sessions. This improves code readability and consistency with the MQTT specification.

## PR Checklist

~- [ ] Change log has been added to `changes/{ce,ee}/(feat|perf|fix)-<PR-id>.en.md` files~
~- [ ] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up~
~- [ ] Schema changes are backward compatible~

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -2064,7 +2064,7 @@ maybe_resume_session(#channel{
 
 maybe_shutdown(Reason, Channel = #channel{conninfo = ConnInfo}) ->
     case maps:get(expiry_interval, ConnInfo) of
-        ?UINT_MAX ->
+        ?EXPIRE_INTERVAL_INFINITE ->
             {ok, Channel};
         I when I > 0 ->
             {ok, ensure_timer(expire_timer, I, Channel)};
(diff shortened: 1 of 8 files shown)


apps/emqx/src/emqx_channel.erl
```diff
@@ -2064,7 +2064,7 @@ maybe_resume_session(#channel{
 
 maybe_shutdown(Reason, Channel = #channel{conninfo = ConnInfo}) ->
     case maps:get(expiry_interval, ConnInfo) of
-        ?UINT_MAX ->
+        ?EXPIRE_INTERVAL_INFINITE ->
```
savonarola, line 2067: This seems to be an old bug: the inherited code wasn't changed.
However, it's hard to test this part naturally — we can't "wait" for such great intervals of time; and any kind mocking make test to have no sense 🤔

paulozulato (approved):
I believe we're not using `UINT_MAX` anymore. Maybe remove its definition on `emqx_mqtt.erl`?

zmstone (approved):

PR #10763, opened 2023-05-21
fix(rule_engine): support Tuple types as arrays

### LINKED ISSUE #9891: Rule engine doesn't handle protobuf variant types
### What happened?

When a protobuf schema contains fields declared as a variant types (i.e. oneof) , the protobuf decoder returns a value wrapped in a tuple, where the first element is a type name and the second one is the value.

Rule engine can’t handle such terms.

Consider a test message: 

22 Nov 2022, 01:12 PM
That complies with the following protobuf schema: [tahu/sparkplug_b.proto at master · eclipse/tahu](https://github.com/eclipse/tahu/blob/master/sparkplug_b/sparkplug_b.proto)

To reproduce the crash, create the following rule:

```
SELECT
  schema_decode('spB-spec3', payload, 'Payload') as msg, payload
FROM
  "a"
Republish with Action template
${msg}
```
Protobuf decoder returns the following term:

```
{<<"msg">> =>
      #{metrics =>
            [#{datatype => 12,is_null => false,name => <<"myTags/tag2">>,
               properties => #{keys => [],values => []},
               timestamp => 1669046712681,
               value => {string_value,<<"Hello There">>}}],
        seq => 4,timestamp => 1669046715695}
```
Note the value of `{string_value, <<"Hello There">>}` : json encoder used by rule engine can’t handle it.

The parsed message is successfully encoded as JSON

See schema from the description

_No response_

### EMQX version

<details>

4.*, 5.*
```

</details>


### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Log files

<details>

</details>

tfoldi:
SQL array functions such as nth or last was working on lists only. This adds Tuple support for array functions.

Fixes #9891

## Summary
copilot:summary

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_funcs.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_funcs.erl
@@ -852,25 +852,52 @@ jq(FilterProgram, JSONBin) ->
 %%------------------------------------------------------------------------------
 
 nth(N, L) when is_integer(N), is_list(L) ->
-    lists:nth(N, L).
+    lists:nth(N, L);
+nth(N, Tuple) when is_integer(N), is_tuple(Tuple) ->
+    element(N, Tuple).
 
 length(List) when is_list(List) ->
-    erlang:length(List).
+    erlang:length(List);
+length(Tuple) when is_tuple(Tuple) ->
+    tuple_size(Tuple).
 
 sublist(Len, List) when is_integer(Len), is_list(List) ->
-    lists:sublist(List, Len).
+    lists:sublist(List, Len);
+sublist(Len, Tuple) when is_integer(Len), is_tuple(Tuple), Len =< tuple_size(Tuple), Len > 0 ->
+    [element(N, Tuple) || N <- lists:seq(1, Len)];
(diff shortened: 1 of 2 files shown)


apps/emqx_utils/test/emqx_utils_json_SUITE.erl
```diff
@@ -111,7 +113,7 @@ t_decode_encode(_) ->
         <<"int">> => 10,
         <<"foo">> => <<"bar">>
     },
-    ?assertEqual(JsonText, encode({decode(JsonText, [])})),
```
tfoldi, line 114: When reviewing, please ensure I did everything correctly in the JSON encoder, as I had to modify one existing test to make it pass. 

In the original test, the decoded value is wrapped in a tuple before encoding. If this is intentional, I have to modify my PR to support this scenario. However, I assume the test is supposed to check if `decode`->`encode` will yield the same text.

PR #10779, opened 2023-05-22
chore: allow special chars in log dir

zmstone:
this pr reverts part of https://github.com/emqx/emqx/pull/10607

file path should allow space, ':' and maybe even unicode. the validator certainly will cause complains.


thalesmg (approved):

zhongwencool:
I try configure as "测试.log" and "测 试 空格.log" on my machine, It's work as expect :)

PR #10790, opened 2023-05-23
 perf(config): read zone conf once

qzhuyan:
Fixes EMQX-9801

## Summary
Prior to this change, the `default` zone could be referenced by the listeners and it is not necessarily an item in the CONF PT if it is not defined under root `zones` by user .
Reading config for the zone must be done through the `emqx_config:get_zone_conf/2` in which it first reads if the `default` zone exists in the config, otherwise it fallback to reading from the root of EMQX as the default fallback. If the `default` zone does exist but the field is undefined, it will fallback to read from default as well.

Generally, it has the following issues:
1.  Inconsistent return val of `emqx_config:get_zone_conf(default,[])` and `emqx_config:get([zones, default])` that one of them return some value another read failure. 
1.  Always read the config twice if no user-defined `default` zone or the field is undefined. (first read miss then fallback to default) which hurts overall performance if read is done in the hot code path.

1. Read a zone with a none existing zone name (not `default`) will also return values that may confuse the users. 

In this PR:
1. In CONF, The default zone which is referenced by the listeners is now explicitly inserted into the runtime config during emqx config initialization during emqx schema loading. the values are merged with zone global defaults and user-defined values.

1. in RAWCONF, it remains unchanged and it keeps a copy of user-defined configs for reference

1. When there is an update on the global defaults, it will be merged into the zones. 

1. When there is an update on the existing zone, it is just a regular update no merge with global defaults.

1. When there is an insertion of new zones, they will be merged with zone global defaults

1. During initialization, loading config order is enforced. load `zones` after the other ROOTs.
### Breaking changes:
1. Read the none-existing zone will fail now.
1. After `emqx_schema` is loaded, CONF PT stores the runtime config that has a `default` zone with complete children fields all set
1. CONF is now the runtime config for hot code path reading while RAWCONF has user-defined values and they are not in sync for root `zones` (<<"zones">>)

# PR Checklist


qzhuyan:
note, with many different attempts in the PR, need to squash the commits before the merge.

apps/emqx/src/emqx_config.erl
```diff
@@ -596,6 +587,12 @@ save_to_app_env(AppEnvs0) ->
 -spec save_to_config_map(config(), raw_config()) -> ok.
 save_to_config_map(Conf, RawConf) ->
     ?MODULE:put(Conf),
+    case emqx_config:get([zones], ?CONFIG_NOT_FOUND_MAGIC) of
+        ?CONFIG_NOT_FOUND_MAGIC ->
+            skip;
+        Zones when is_map(Zones) ->
+            init_default_zone()
```
qzhuyan, line 594: move to `init_load`

apps/emqx/src/emqx_config.erl
```diff
@@ -788,3 +785,113 @@ to_atom_conf_path(Path, OnFail) ->
...
+                        )
+                    of
+                        ?CONFIG_NOT_FOUND_MAGIC ->
+                            ZoneVal#{RootName => NewRootValue};
+                        RawUserZoneRoot ->
+                            ZoneVal#{
+                                RootName :=
+                                    emqx_utils_maps:deep_merge(
+                                        NewRootValue,
+                                        emqx_utils_maps:unsafe_atom_key_map(RawUserZoneRoot)
```
qzhuyan, line 868: check_config(SchemaMod, RawConf, #{}),

apps/emqx/test/emqx_shared_sub_SUITE.erl
```diff
@@ -46,12 +47,30 @@
 all() -> emqx_common_test_helpers:all(?SUITE).
 
 init_per_suite(Config) ->
-    net_kernel:start(['master@127.0.0.1', longnames]),
+    DistPid =
```
qzhuyan, line 50: remove

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -33,8 +33,6 @@
 all() -> emqx_common_test_helpers:all(?MODULE).
 
 init_per_suite(Config) ->
-    emqx_common_test_helpers:boot_modules(all),
```
qzhuyan, line 36: double check

apps/emqx/test/emqx_config_SUITE.erl
```diff
@@ -96,3 +97,284 @@ t_unknown_rook_keys(_) ->
...
+
+    %% Then user defined value is set
+    {MqttV, Others} = maps:take(mqtt, emqx_config:get([zones, myzone])),
+    {ZGDMQTT, ExpectedOthers} = maps:take(mqtt, zone_global_defaults()),
+    ?assertEqual(ZGDMQTT#{max_topic_alias := 1024}, MqttV),
+    %% Then others are defaults
+    ?assertEqual(ExpectedOthers, Others),
+    %% Then default zone still have the defaults
+    ?assertEqual(zone_global_defaults(), emqx_config:get([zones, default])).
+
```
qzhuyan, line 198: add duration value `10s`

apps/emqx/test/emqx_config_SUITE.erl
```diff
@@ -96,3 +97,284 @@ t_unknown_rook_keys(_) ->
...
+t_init_zones_with_cust_root_mqtt(Config) ->
+    emqx_config:erase_all(),
+    %% Given config file with mqtt user overrides
+    ConfFile = prepare_conf_file(?FUNCTION_NAME, <<"mqtt.retry_interval=600000">>, Config),
```
qzhuyan, line 202: s/600000/6s

apps/emqx/test/emqx_config_SUITE.erl
```diff
@@ -96,3 +97,284 @@ t_unknown_rook_keys(_) ->
...
+    ).
+
+t_myzone_is_updated_after_global_defaults_updated(Config) ->
+    emqx_config:erase_all(),
+    %% Given emqx conf file 3with user override in myzone (none default zone)
+    ConfFile = prepare_conf_file(?FUNCTION_NAME, <<"zones.myzone.mqtt.max_inflight=32">>, Config),
+    application:set_env(emqx, config_files, [ConfFile]),
+    ?assertEqual(ok, emqx_config:init_load(emqx_schema)),
+    ?assertNotEqual(900000, emqx_config:get([zones, myzone, mqtt, retry_interval])),
+    %% When emqx_schema is loaded
```
qzhuyan, line 236: update: change root conf with API put

qzhuyan:
also improve: emqx_mgmt_caps:get_caps

apps/emqx/src/emqx_config.erl
```diff
@@ -788,3 +781,138 @@ to_atom_conf_path(Path, OnFail) ->
                     V
             end
     end.
+
+-spec maybe_init_default_zone() -> skip | ok.
+maybe_init_default_zone() ->
+    case emqx_config:get([zones], ?CONFIG_NOT_FOUND_MAGIC) of
+        ?CONFIG_NOT_FOUND_MAGIC ->
+            skip;
+        Zones when is_map(Zones) ->
```
lafirest, line 790: Why not pass the `Zones` to `init_default_zone` so we don't need to call the `get([zones])` twice?

apps/emqx/src/emqx_config.erl
```diff
@@ -788,3 +781,138 @@ to_atom_conf_path(Path, OnFail) ->
...
+%%    when 1) zone updates, return *new* zones
+%%    when 2) zone global config updates, write to PT directly.
+%% Zone global defaults are always presented in the configmap (PT) when updating zone
+-spec maybe_update_zone(runtime_config_key_path(), RootValue :: map(), Val :: term()) ->
+    NewZoneVal :: map().
+maybe_update_zone([zones | T], ZonesValue, Value) ->
+    %% note, do not write to PT, return *New value* instead
+    NewZonesValue = emqx_utils_maps:deep_put(T, ZonesValue, Value),
+    ExistingZoneNames = maps:keys(?MODULE:get([zones], #{})),
+    case maps:keys(NewZonesValue) -- ExistingZoneNames of
```
lafirest, line 833: Or
```
maps:fold(fun(...) -> 
...
end,
maps:without(ExistingZoneNames, NewZonesValue))
```

apps/emqx/src/emqx_config.erl
qzhuyan: thanks, I will merge these two funs into one.

zhongwencool (approved):
LGTM

yanzhiemq:
_

yanzhiemq:
_

PR #10869, opened 2023-05-30
0530 prepare for 5.1.0 alpha.1

zmstone:
* Update `release-50` to `release-51` in release scripts.
* Finalize `master.bpapi` as `5.0.bpapi`


apps/emqx/include/emqx_release.hrl
```diff
@@ -32,10 +32,10 @@
 %% `apps/emqx/src/bpapi/README.md'
 
 %% Community edition
--define(EMQX_RELEASE_CE, "5.0.26").
+-define(EMQX_RELEASE_CE, "5.1.0-alpha.1").
```
id, line 35: ?

scripts/rel/cut.sh
```diff
@@ -270,7 +270,7 @@ else
         *beta*)
             true
             ;;
-        e5.0.0*)
+        e5.1.0*)
             # the first release has no change log
```
id, line 274: We don't want to show changes between 5.0 and 5.1?

scripts/rel/cut.sh
```diff
@@ -13,14 +13,14 @@ usage() {
...
-  e5.0.0-beta.6
+  v5.1.1
+  e5.1.0-beta.6
 
 options:
   -h|--help:         Print this usage.
 
   -b|--base:         Specify the current release base branch, can be one of
-                     release-50
+                     release-51
```
id, line 23: May be we can just use `release` at this point

zmstone: `release` is not quite search friendly. 
how about `release-candidate`?

apps/emqx/include/emqx_release.hrl
zmstone: yes, both will enter 5.1 from now

id (approved):

PR #10881, opened 2023-05-30
IoTDB test coverage 

sstrigler:
Fixes [EMQX-9355](https://emqx.atlassian.net/browse/EMQX-9355)

### <samp>🤖 Generated by Copilot at 767f7c5</samp>

This pull request enhances the `emqx_bridge_iotdb` application and its test suite, by adding more trace points, error handling, logging, and data processing features. It also fixes some bugs and improves the test case robustness and coverage. The version is updated to 0.1.2.

## PR Checklist

- [x] ~Change log has been added to `changes/{ce,ee}/(feat|perf|fix)-<PR-id>.en.md` files~
- [x] ~For internal contributor: there is a jira ticket to track this change~

### DIFF
--- a/apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_impl.erl
+++ b/apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_impl.erl
@@ -154,21 +154,9 @@ on_query_async(InstanceId, {send_message, Message}, ReplyFunAndArgs0, State) ->
 %%--------------------------------------------------------------------
 
 make_parsed_payload(PayloadUnparsed) when is_binary(PayloadUnparsed) ->
-    emqx_utils_json:decode(PayloadUnparsed, [return_maps]);
+    emqx_utils_json:decode(PayloadUnparsed);
 make_parsed_payload(PayloadUnparsed) when is_list(PayloadUnparsed) ->
-    lists:map(fun make_parsed_payload/1, PayloadUnparsed);
-make_parsed_payload(
-    #{
-        measurement := Measurement,
-        data_type := DataType,
-        value := Value
-    } = Data
-) ->
-    Data#{
-        <<"measurement">> => Measurement,
-        <<"data_type">> => DataType,
-        <<"value">> => Value
-    }.
+    lists:map(fun make_parsed_payload/1, PayloadUnparsed).
 
 preproc_data_list(DataList) ->
     lists:foldl(
@@ -270,6 +258,7 @@ replace_var(Val, _Data) ->
     Val.
 
 convert_bool(B) when is_boolean(B) -> B;
+convert_bool(null) -> null;
 convert_bool(1) -> true;
 convert_bool(0) -> false;
 convert_bool(<<"1">>) -> true;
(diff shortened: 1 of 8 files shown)


apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_impl.erl
```diff
@@ -104,83 +104,108 @@ on_get_status(InstanceId, State) ->
...
-        data_type := DataType,
-        value := Value
-    } = Data
-) ->
-    Data#{
-        <<"measurement">> => Measurement,
-        <<"data_type">> => DataType,
-        <<"value">> => Value
-    }.
+get_payload(#{payload := Payload}) ->
```
kjellwinblad, line 156: Using payload as the field name for the field might be convenient in some sense since it allows one to use use a `SELECT *`  to forward an MQTT message to the bridge. However, choosing a different field name might force users to get more aware of that they can use the rule to transform the data and one could make the name more descriptive (for example using the name `measurements` would make it more clear that the bridge can handle more than one measurement).

Maybe one could also allow users to specify `timestamp` and `device_id` on the top level and use the top level value for those fields if a measurement is lacking any of those fields?

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_impl.erl
```diff
@@ -104,83 +104,108 @@ on_get_status(InstanceId, State) ->
...
-        value => emqx_plugin_libs_rule:preproc_tmpl(Value)
-    }.
-
-preproc_data_list(DataList) ->
-    lists:map(
-        fun preproc_data/1,
-        DataList
-    ).
+    [
+        #{
```
kjellwinblad, line 184: After our discussion yesterday about that making it possible to do the same thing in too many different ways might be an antipattern, I'm not sure that treating timestamp and so forth as a template is a good idea. It seems to me as a little bit unnecessary complicated to send template strings from the rule that then get instantiated based on other data from the rule, when one could just as well assemble the data to have the desired content directly in the rule. Based on the same reasoning it seems unnecessary to support the special timestamp value `"now"` when the rule engine has build in functions for making timestamps.

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_impl.erl
```diff
@@ -104,83 +104,108 @@ on_get_status(InstanceId, State) ->
...
+    ),
+    Acc.
+
+maybe_preproc_tmpl(Value) when is_binary(Value) ->
+    emqx_plugin_libs_rule:preproc_tmpl(Value);
+maybe_preproc_tmpl(Value) ->
+    Value.
 
 proc_data(PreProcessedData, Msg) ->
     NowNS = erlang:system_time(nanosecond),
```
kjellwinblad, line 211: This probably have little impact on the overall performance but it seems a bit wasteful to do a system call and create the `Nows` map when we might not need any of it (since the user might have sent a timestamp from the rule). Perhaps it is better to do this computation only when we now that we need it? I understand that it is done in the beginning so that we only have to do it once if the message contains multiple measurements but we could make sure that it is only done once even if it is done on demand.

kjellwinblad (approved):
I have left some comments. Mostly based on the insight we had yesterday when discussing the bridge. It is up to you how much of it you think we should address. Most of the comments are about removing functionality and I don't think they are crucial to address.

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_impl.erl
sstrigler: On top of that, IoTDB would allow sending `now()` as a timestamp value itself

PR #10911, opened 2023-06-01
fix: friendly error message when creating bridges with too long names

kjellwinblad:
This commit makes the error message and log entry that appear when one tries to create a bridge with a name the exceeds 255 bytes (the max length for atoms) more friendly and easier to understand.

An even better fix would be to not store bridge names as atoms but this probably requires a more substantial change.

Fixes:
https://emqx.atlassian.net/browse/EMQX-9609

### <samp>🤖 Generated by Copilot at 75ff76a</samp>

This pull request improves the error handling, validation, and configuration logic for bridges and rules in emqx. It also updates the version numbers of the affected applications: `emqx_rule_engine`, `emqx_ee_bridge`.

## PR Checklist

- [] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up

### DIFF
--- a/apps/emqx/src/emqx_config_handler.erl
+++ b/apps/emqx/src/emqx_config_handler.erl
@@ -473,6 +473,17 @@ bin_path(ConfKeyPath) -> [bin(Key) || Key <- ConfKeyPath].
 bin(A) when is_atom(A) -> atom_to_binary(A, utf8);
 bin(B) when is_binary(B) -> B.
 
+atom(Bin) when is_binary(Bin), size(Bin) > 255 ->
+    erlang:throw(
+        iolist_to_binary(
+            io_lib:format(
+                "Name is is too long."
+                " Please provide a shorter name (<= 255 bytes)."
+                " The name that is too long: \"~s\"",
+                [Bin]
+            )
+        )
+    );
 atom(Bin) when is_binary(Bin) ->
     binary_to_atom(Bin, utf8);
 atom(Str) when is_list(Str) ->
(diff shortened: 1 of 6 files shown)


apps/emqx/src/emqx_config_handler.erl
```diff
@@ -473,6 +473,17 @@ bin_path(ConfKeyPath) -> [bin(Key) || Key <- ConfKeyPath].
...
 bin(B) when is_binary(B) -> B.
 
+atom(Bin) when is_binary(Bin), size(Bin) > 255 ->
+    erlang:throw(
+        iolist_to_binary(
+            io_lib:format(
+                "Name is is too long."
+                " Please provide a shorter name (<= 255 bytes)."
+                " The name that is too long: \"~s\"",
+                [Bin]
```
zmstone, line 483:
```suggestion
                "Name '~s' is is too long."
                " It should be less than 255 bytes."
                [Bin]
```

zmstone (commented):
please add a test case for regression purposes.

apps/emqx/src/emqx_config_handler.erl
kjellwinblad: Originally, I had the text written more or less exactly like you suggest but then I noticed that when you have a long name the text is cut off in the log entry so all you see is "Name 'nameeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee...". That is why I changed to its current form. Do you still prefer your version?

kjellwinblad:
> please add a test case for regression purposes.
Sure will do so.

ieQu1 (approved):

yanzhiemq:
_

PR #10927, opened 2023-06-02
feat(tlslib): add separate managed certfiles GC process

keynslug:
Which periodically inpects managed certificates directory and tries to collect "orphans" here, in other words files that aren't referenced anywhere in the current emqx config. At the same time stop explicitly deleting certfiles in `post_config_update`s throughout the code base.

Fixes [EMQX-9930](https://emqx.atlassian.net/browse/EMQX-9930)

### <samp>🤖 Generated by Copilot at 6c5b69c</samp>

Refactored and simplified the code related to TLS certificate management across different modules. Extracted the common logic for cleaning up orphaned TLS files into a separate module `emqx_tls_certfile_gc` and a worker process `emqx_tls_certfile_gc`. Added utility functions for file system operations in `emqx_utils_fs`. Removed unused or redundant code and functions from several modules.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_tls_lib.erl
+++ b/apps/emqx/src/emqx_tls_lib.erl
@@ -389,7 +389,7 @@ is_pem(MaybePem) ->
 %% Also a potentially half-written PEM file (e.g. due to power outage)
 %% can be corrected with an overwrite.
 save_pem_file(Dir, KeyPath, Pem, DryRun) ->
-    Path = pem_file_name(Dir, KeyPath, Pem),
+    Path = pem_file_name(Dir, KeyPath),
     case filelib:ensure_dir(Path) of
         ok when DryRun ->
             {ok, Path};
@@ -412,9 +412,8 @@ is_managed_ssl_file(Filename) ->
         _ -> false
     end.
 
-pem_file_name(Dir, KeyPath, Pem) ->
-    <<CK:8/binary, _/binary>> = crypto:hash(md5, Pem),
-    Suffix = binary:encode_hex(CK),
+pem_file_name(Dir, KeyPath) ->
+    Suffix = binary:encode_hex(crypto:strong_rand_bytes(8)),
     Segments = lists:map(fun ensure_bin/1, KeyPath),
     Filename0 = iolist_to_binary(lists:join(<<"_">>, Segments)),
     Filename1 = binary:replace(Filename0, <<"file">>, <<>>),


apps/emqx/src/emqx_tls_certfile_gc.erl
```diff
@@ -0,0 +1,335 @@
...
+        end,
+        [],
+        Config
+    ).
+
+is_file_reference([<<"keyfile">> | _]) -> true;
+is_file_reference([<<"certfile">> | _]) -> true;
+is_file_reference([<<"cacertfile">> | _]) -> true;
+is_file_reference([<<"issuer_pem">>, <<"ocsp">> | _]) -> true;
+is_file_reference(_) -> false.
```
savonarola, line 247: How could we keep in sync this list with possible newly added certs?

keynslug: Good question. Speaking of this PR, it'd probably be a good idea to move that stuff to `emqx_tls_lib`. Generally speaking, my (optimistic) outlook is that we won't need to mess with this "reifying" process anymore, instead keep configured certs as is: either filename or certificate itself (i.e. if the user pasted it in the dashboard).

apps/emqx/src/emqx_tls_certfile_gc.erl
```diff
@@ -0,0 +1,335 @@
...
+        #{},
+        collect(St, #{evhandler => {fun log_event/2, []}})
+    ),
+    {noreply, start_timer(StNext)}.
+
+start_timer(St = #{gc_interval := Interval}) ->
+    TRef = erlang:start_timer(Interval, self(), collect),
+    St#{next_gc_timer => TRef}.
+
+restart_timer(St = #{next_gc_timer := TRef}) ->
```
zmstone, line 147: ensure the old is cancelled before start a new one.
because the `col.lect` message can be sent by anyone.

apps/emqx/src/emqx_tls_certfile_gc.erl
```diff
@@ -0,0 +1,335 @@
...
+    ).
+
+is_file_reference(Stack) ->
+    lists:any(
+        fun(KP) -> lists:prefix(lists:reverse(KP), Stack) end,
+        emqx_tls_lib:ssl_file_conf_keypaths()
+    ).
+
+is_string(Value) ->
+    is_list(Value) orelse is_binary(Value).
```
zmstone, line 250:
```suggestion
    is_binary(Value).
```

apps/emqx/src/emqx_tls_certfile_gc.erl
```diff
@@ -0,0 +1,335 @@
...
+        end,
+        Certfiles,
+        emqx_config:get_root_names()
+    ).
+
+convicts(Orphans, OrphansLast) ->
+    maps:fold(
+        fun(AbsPath, #file_info{mtime = MTime, inode = Inode}, Acc) ->
+            case maps:get(AbsPath, Orphans, undefined) of
+                #file_info{mtime = MTime, inode = Inode} ->
```
zmstone, line 187: It seems `inode` is always `0` in windows.

apps/emqx/src/emqx_tls_certfile_gc.erl
```diff
@@ -0,0 +1,335 @@
...
+orphans() ->
+    Dir = emqx_utils_fs:canonicalize(emqx:mutable_certs_dir()),
+    orphans(Dir).
+
+orphans(Dir) ->
+    Certfiles = find_managed_files(fun is_managed_file/2, Dir),
+    lists:foldl(
+        fun(Root, Acc) ->
+            References = find_references(Root),
+            maps:without(References, Acc)
```
zmstone, line 177: if the certs directory is once symlinked and expanded to an abs path,
e.g. `foo.bar.ssl_options.cacertfile=/data1/certs/cacertfile-0000.pem`
then  `node.data_dir` changes to another link, e.g. `/data2/` (node restart), this will cause all the cert files to be deleted after 2 collection-ticks?

zmstone: `data1/certs/bridge1/cacertfile-0000.pem`
`data1/certs/bridge2/cacertfile-0000.pem`

zmstone: solution: use random suffix instead of cert hash hex string for file name suffix.

apps/emqx/src/emqx_tls_lib.erl
```diff
@@ -432,15 +406,15 @@ save_pem_file(Dir, KeyPath, Pem, DryRun) ->
...
 %% e.g. key-1234567890abcdef, cert-1234567890abcdef, and cacert-1234567890abcdef
-is_generated_file(Filename) ->
+is_managed_ssl_file(Filename) ->
     case string:split(filename:basename(Filename), "-") of
         [_Name, Suffix] -> is_hex_str(Suffix);
         _ -> false
     end.
 
 pem_file_name(Dir, KeyPath, Pem) ->
     <<CK:8/binary, _/binary>> = crypto:hash(md5, Pem),
```
zmstone, line 416: Maybe add the original full path together with pem content to make the md5 hash, so to make more unique suffix. Because file base name

keynslug: Oh, right, forgot about that. Made suffixes random in [d5aedaa](https://github.com/emqx/emqx/pull/10927/commits/d5aedaac7daa9237f3af71f3fe6532feac7c829d).

zmstone (approved):

PR #10954, opened 2023-06-06
feat: don't hide enable in flapping detect conf

zhongwencool:
Fixes <issue-or-jira-number>
don't hide flapping_detect on global zone.

### <samp>🤖 Generated by Copilot at 169ee4a</samp>

This pull request adds and exposes the flapping detection feature for zones, which allows users to configure different policies for detecting and handling unstable MQTT clients. It also refactors and simplifies the code related to flapping detection and configuration updates in `emqx_config.erl`, `emqx_flapping.erl`, `emqx_channel.erl`, and `emqx_schema.erl`, and updates the corresponding test suite.

## PR Checklist

### DIFF
--- a/apps/emqx/src/config/emqx_config_zones.erl
+++ b/apps/emqx/src/config/emqx_config_zones.erl
@@ -1,0 +1,35 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2020-2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+-module(emqx_config_zones).
+
+-behaviour(emqx_config_handler).
+
+%% API
+-export([add_handler/0, remove_handler/0, pre_config_update/3]).
+
+-define(ZONES, [zones]).
+
+add_handler() ->
+    ok = emqx_config_handler:add_handler(?ZONES, ?MODULE),
+    ok.
+
+remove_handler() ->
+    ok = emqx_config_handler:remove_handler(?ZONES),
+    ok.
+
+%% replace the old config with the new config
+pre_config_update(?ZONES, NewRaw, _OldRaw) ->
+    {ok, NewRaw}.
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -1632,10 +1632,9 @@ check_banned(_ConnPkt, #channel{clientinfo = ClientInfo}) ->
 %%--------------------------------------------------------------------
 %% Flapping
 
-count_flapping_event(_ConnPkt, Channel = #channel{clientinfo = ClientInfo = #{zone := Zone}}) ->
-    is_integer(emqx_config:get_zone_conf(Zone, [flapping_detect, window_time])) andalso
-        emqx_flapping:detect(ClientInfo),
-    {ok, Channel}.
+count_flapping_event(_ConnPkt, #channel{clientinfo = ClientInfo}) ->
+    _ = emqx_flapping:detect(ClientInfo),
+    ok.
 
 %%--------------------------------------------------------------------
 %% Authenticate
--- a/apps/emqx/src/emqx_config.erl
+++ b/apps/emqx/src/emqx_config.erl
@@ -637,11 +637,13 @@ save_to_override_conf(false, RawConf, _Opts) ->
 
 add_handlers() ->
     ok = emqx_config_logger:add_handler(),
+    ok = emqx_config_zones:add_handler(),
     emqx_sys_mon:add_handler(),
     ok.
 
 remove_handlers() ->
     ok = emqx_config_logger:remove_handler(),
+    ok = emqx_config_zones:remove_handler(),
     emqx_sys_mon:remove_handler(),
     ok.
 
@@ -914,7 +916,8 @@ rawconf_to_conf(SchemaModule, RawPath, RawValue) ->
     emqx_utils_maps:deep_get(AtomPath, RawUserDefinedValues).
 
 %% When the global zone change, the zones is updated with the new global zone.
-%% The zones config has no config_handler callback, so we need to update via this hook
+%% The global zone's keys is too many,
+%% so we don't choose to write a global zone change emqx_config_handler callback to hook
 post_save_config_hook(?PERSIS_KEY(?CONF, zones), _Zones) ->
     emqx_flapping:update_config(),
     ok;
--- a/apps/emqx/src/emqx_config.erl
+++ b/apps/emqx/src/emqx_config.erl
@@ -831,13 +831,14 @@ merge_with_global_defaults(GlobalDefaults, ZoneVal) ->
     NewZoneVal :: map().
 maybe_update_zone([zones | T], ZonesValue, Value) ->
     %% note, do not write to PT, return *New value* instead
-    NewZonesValue = emqx_utils_maps:deep_put(T, ZonesValue, Value),
     GLD = zone_global_defaults(),
+    NewZonesValue0 = emqx_utils_maps:deep_put(T, ZonesValue, Value),
+    NewZonesValue1 = emqx_utils_maps:deep_merge(#{default => GLD}, NewZonesValue0),
     maps:map(
         fun(_ZoneName, ZoneValue) ->
             merge_with_global_defaults(GLD, ZoneValue)
         end,
-        NewZonesValue
+        NewZonesValue1
     );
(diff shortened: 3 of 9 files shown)


apps/emqx/src/config/emqx_config_zones.erl
```diff
@@ -0,0 +1,35 @@
...
+    ok = emqx_config_handler:add_handler(?ZONES, ?MODULE),
+    ok.
+
+remove_handler() ->
+    ok = emqx_config_handler:remove_handler(?ZONES),
+    ok.
+
+%% replace the old config with the new config
+pre_config_update(?ZONES, NewRaw, _OldRaw) ->
+    {ok, NewRaw}.
```
zhongwencool, line 35: If we don't implement this pre_config_update/3, the zones is a map object.
the default behaivor is to merge not replace, such as :
we have a configuration 
```
zones {my1 {mqtt {idle_timeout= "5s"}}}.
```
We can't  remove `my1` zones with `emqx:update_config([zones], #{})` before.
now we implement replace behaivor, this update is working now.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1632,10 +1632,9 @@ check_banned(_ConnPkt, #channel{clientinfo = ClientInfo}) ->
 %%--------------------------------------------------------------------
 %% Flapping
 
-count_flapping_event(_ConnPkt, Channel = #channel{clientinfo = ClientInfo = #{zone := Zone}}) ->
-    is_integer(emqx_config:get_zone_conf(Zone, [flapping_detect, window_time])) andalso
-        emqx_flapping:detect(ClientInfo),
-    {ok, Channel}.
+count_flapping_event(_ConnPkt, #channel{clientinfo = ClientInfo}) ->
+    _ = emqx_flapping:detect(ClientInfo),
+    ok.
```
zhongwencool, line 1637: Channel is not changed in this pipeline, so we can return ok instead of `{ok, Channel}`

apps/emqx_management/test/emqx_mgmt_api_configs_SUITE.erl
```diff
@@ -209,7 +209,7 @@ t_zones(_Config) ->
     ?assertEqual(Mqtt1, NewMqtt),
     %% delete the new zones
     {ok, #{}} = update_config("zones", Zones),
-    ?assertEqual(undefined, emqx_config:get_raw([new_zone, mqtt], undefined)),
+    ?assertEqual(undefined, emqx_config:get_raw([zones, new_zone], undefined)),
```
zhongwencool, line 212: Check the wrong key before this PR...

apps/emqx/src/emqx_config.erl
```diff
@@ -828,16 +833,14 @@ merge_with_global_defaults(GlobalDefaults, ZoneVal) ->
...
+    NewZonesValue0 = emqx_utils_maps:deep_put(T, ZonesValue, Value),
+    NewZonesValue1 = emqx_utils_maps:deep_merge(#{default => GLD}, NewZonesValue0),
+    maps:map(
+        fun(_ZoneName, ZoneValue) ->
+            merge_with_global_defaults(GLD, ZoneValue)
         end,
-        NewZonesValue,
-        maps:without(ExistingZoneNames, NewZonesValue)
+        NewZonesValue1
     );
```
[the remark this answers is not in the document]
qzhuyan: Just a question about precondition, in which case we could get zone that has no defaults?
```
zones {my1 {mqtt {idle_timeout="5s"}}
```

Unless the PT is put manually, it should always have defaulted in runtime(atom key path) `config zones.my1.mqtt` isn't it?

apps/emqx/src/emqx_config.erl
```diff
@@ -828,16 +833,14 @@ merge_with_global_defaults(GlobalDefaults, ZoneVal) ->
...
     %% note, do not write to PT, return *New value* instead
-    NewZonesValue = emqx_utils_maps:deep_put(T, ZonesValue, Value),
-    ExistingZoneNames = maps:keys(?MODULE:get([zones], #{})),
-    %% Update only new zones with global defaults
     GLD = zone_global_defaults(),
-    maps:fold(
-        fun(ZoneName, ZoneValue, Acc) ->
-            Acc#{ZoneName := merge_with_global_defaults(GLD, ZoneValue)}
+    NewZonesValue0 = emqx_utils_maps:deep_put(T, ZonesValue, Value),
+    NewZonesValue1 = emqx_utils_maps:deep_merge(#{default => GLD}, NewZonesValue0),
```
qzhuyan, line 838: Why do we need prefill `#{default => GLD}`? the default one should be inserted during schema loading by design.

apps/emqx/src/emqx_config.erl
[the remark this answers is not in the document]
zhongwencool: We can always update `zones` at runtime via (`/configs/zones`)http api.
the http api will call `emqx_config:update([zones],Conf)` 
so we need ensure all zone has full fields to not retrive twice.

apps/emqx/src/emqx_flapping.erl
```diff
@@ -82,24 +93,13 @@ detect(#{clientid := ClientId, peerhost := PeerHost, zone := Zone}) ->
...
-        end,
-        #{},
-        Keys
-    );
-get_policy(Key, Zone) ->
-    #{Key := Conf} = get_policy([Key], Zone),
-    Conf.
+%% with default, if we delete Zone at running time. we should not crash.
+get_policy(Zone) ->
+    emqx_config:get_zone_conf(Zone, [flapping_detect], ?DEFAULT_POLICY).
```
qzhuyan, line 102: I think even user delete the`default` zone, the `default` zone should still present in run time config
 thus we don't need to provide default ?DEFAULT_POLICY ?

apps/emqx/src/emqx_config.erl
```diff
@@ -911,3 +914,12 @@ rawconf_to_conf(SchemaModule, RawPath, RawValue) ->
         ),
     AtomPath = to_atom_conf_path(RawPath, {raise_error, maybe_update_zone_error}),
     emqx_utils_maps:deep_get(AtomPath, RawUserDefinedValues).
+
+%% When the global zone change, the zones is updated with the new global zone.
+%% The global zone's keys is too many,
+%% so we don't choose to write a global zone change emqx_config_handler callback to hook
+post_save_config_hook(?PERSIS_KEY(?CONF, zones), _Zones) ->
+    emqx_flapping:update_config(),
```
qzhuyan, line 922: this is the side effect of updating flapping detect, so question is do we have other side effects for other configs?

qzhuyan:
just a bit concerned about when deleting the zone we must check the references, otherwise LGTM.

qzhuyan (approved):

apps/emqx/src/emqx_flapping.erl
```diff
@@ -190,18 +193,30 @@ terminate(_Reason, _State) ->
...
 
-start_timer(Zone) ->
-    case get_policy(window_time, Zone) of
-        WindowTime when is_integer(WindowTime) ->
-            emqx_utils:start_timer(WindowTime, {garbage_collect, Zone});
-        disabled ->
-            ok
-    end.
+start_timer(#{enable := true, window_time := WindowTime}, Zone) ->
+    emqx_utils:start_timer(WindowTime, {garbage_collect, Zone});
```
zmstone, line 197: if we keep allowing `disabled`, the is_integer check is still needed.

apps/emqx/src/emqx_flapping.erl
zmstone: there is currently no fallback to `default` zone, right ?
so if `Zone` is deleted, providing a default here indeed can prevent crashing, but it just hides the missing zone reference issue to later.

apps/emqx/src/emqx_config.erl
```diff
@@ -911,3 +914,12 @@ rawconf_to_conf(SchemaModule, RawPath, RawValue) ->
         ),
     AtomPath = to_atom_conf_path(RawPath, {raise_error, maybe_update_zone_error}),
     emqx_utils_maps:deep_get(AtomPath, RawUserDefinedValues).
+
+%% When the global zone change, the zones is updated with the new global zone.
+%% The global zone's keys is too many,
+%% so we don't choose to write a global zone change emqx_config_handler callback to hook
+post_save_config_hook(?PERSIS_KEY(?CONF, zones), _Zones) ->
```
zmstone, line 921: this is not a hook, at least not the current form.
maybe name it `zone_config_post_change_actions`

apps/emqx/src/emqx_config.erl
zhongwencool: after my examination, only this one is yes. :0

apps/emqx/src/emqx_schema.erl
```diff
@@ -33,6 +33,7 @@
 -define(MAX_INT_TIMEOUT_MS, 4294967295).
 %% floor(?MAX_INT_TIMEOUT_MS / 1000).
 -define(MAX_INT_TIMEOUT_S, 4294967).
+-define(DEFAULT_WINDOW_TIME, "1m").
```
zmstone, line 36:
```suggestion
-define(DEFAULT_WINDOW_TIME, <<"1m">>).
```

apps/emqx/src/emqx_flapping.erl
qzhuyan: right.
1. `default` zone must be present in the runtime after the schema is loaded.
2. right, fallback to default here just hides the issue. Without fallback, we could catch some error in the CI if there is unexpected behavior.

qzhuyan:
not targeting release-51?

qzhuyan (approved):

PR #10955, opened 2023-06-07
Fix(mqttsn): clean predefined topics once gateway unload

HJianBo:
Fixes https://emqx.atlassian.net/browse/EMQX-10131

### <samp>🤖 Generated by Copilot at 407c632</samp>

This pull request implements and tests the feature of deleting predefined topics for mqttsn clients when they disconnect or when the gateway stops. It adds a new function to `emqx_mqttsn_registry` and modifies `emqx_gateway_mqttsn` and `emqx_sn_protocol_SUITE` accordingly.

## PR Checklist


apps/emqx_gateway_mqttsn/src/emqx_mqttsn_registry.erl
```diff
@@ -118,6 +119,16 @@ lookup_element(Tab, Key, Pos) ->
 unregister_topic({_, Pid}, ClientId) ->
     gen_server:call(Pid, {unregister, ClientId}).
 
+-spec clean_predefined_topics(atom(), list()) -> ok.
+clean_predefined_topics(InstaId, PredefTopics) when is_list(PredefTopics) ->
+    Tab = name(InstaId),
+    F = fun(#{id := TopicId, topic := TopicName0}) ->
+        TopicName = iolist_to_binary(TopicName0),
+        mria:dirty_delete(Tab, {predef, TopicId}),
```
zmstone, line 127: this deletes data on all nodes.

HJianBo: I have just submitted a new commit to remove `emqx_mqttsn_registry` table from Mria. This registration relationship only takes effect for the client (similar to MQTT protocol's topic alias), so it doesn't need to be stored in Mria.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_registry.erl
```diff
@@ -17,64 +17,91 @@
...
+    id_to_name := map(),
+    %% The mapping from topic name to topic id
+    name_to_id := map()
+}.
 
--record(emqx_mqttsn_registry, {key, value}).
+%%-----------------------------------------------------------------------------
 
--type registry() :: {Tab :: atom(), RegistryPid :: pid()}.
+-spec persist_predefined_topics(list()) -> ok.
```
zmstone, line 49: `list()` is not a quite helpful spec.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_registry.erl
```diff
@@ -17,64 +17,91 @@
...
--spec start_link(atom(), list()) ->
-    ignore
-    | {ok, pid()}
-    | {error, Reason :: term()}.
-start_link(InstaId, PredefTopics) ->
-    gen_server:start_link(?MODULE, [InstaId, PredefTopics], []).
+-spec init() -> registry().
+init() ->
+    #{
+        next_topic_id => persistent_term:get(?PKEY_MAX_PREDEF_ID, 0),
```
zmstone, line 83: this makes the config change tricky.
e.g. if there are 10 predefined topics when booted, then there are 11-X registered at runtime,
later, if the sysadmin decided to expand the pre-defined topics list, 11 would have already been taken.
 
I would recommend to preserved 1 - N (N could be 100 or 1024 etc) for predefined, and not allow more than that from configuration.
And the reset (N+1 up to 16#FFFF) is for runtime registration.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_registry.erl
```diff
@@ -83,182 +110,64 @@ register_topic({_, Pid}, ClientId, TopicName) when is_binary(TopicName) ->
...
-name(InstaId) ->
-    list_to_atom(lists:concat([emqx_mqttsn_, InstaId, '_registry'])).
-
-init([InstaId, PredefTopics]) ->
-    %% {predef, TopicId}     -> TopicName
-    %% {predef, TopicName}   -> TopicId
-    %% {ClientId, TopicId}   -> TopicName
-    %% {ClientId, TopicName} -> TopicId
-    Tab = name(InstaId),
-    ok = mria:create_table(Tab, [
```
zmstone, line 135: please test if this will affect new version node join old version cluster.

zmstone:
just a question. 
if session migrate from one node to another, is the registry takeover ?

zmstone:
just to confirm:
topic registration is independent between clients.
topic `a` could be registered as `11` for client-1, and but it does have to be `11` for client-2?

HJianBo:
Both answers are yes. So we don't need a cluster-wide topic registry for the MQTT-SN gateway.  Just treat it as a part of the Session

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_registry.erl
HJianBo: Nice catch, I will fix it asap

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_registry.erl
HJianBo: I am unsure how to test it in CI, but I do not think it will cause compatibility issues during rolling upgrades. 🤔

zmstone: it might already covered in CI with this
https://github.com/emqx/emqx/blob/aea3ae4024e51af93365d6da17257bdf252b25b8/.github/workflows/build_slim_packages.yaml#L212-L212

zmstone: please confirm check

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_registry.erl
```diff
@@ -17,64 +17,92 @@
...
+    lookup_topic/2,
+    lookup_topic_id/2
 ]).
 
--export([lookup_name/1]).
+-define(PKEY(Id), {mqttsn, predef_topics, Id}).
 
--define(SN_SHARD, emqx_mqttsn_shard).
+-type registry() :: #{
+    %% The next topic id to be assigned to new registration
```
zmstone, line 38:
```suggestion
    %% The last topic id allocated.
```

zmstone (dismissed):

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_registry.erl
HJianBo: 😢 yes, I will fix it in following PR

zhongwencool (approved):

yanzhiemq:
_

yanzhiemq:
_

PR #10960, opened 2023-06-07
chore: add 'mcast' to spellcheck dict

zmstone:


ieQu1 (approved):

JimMoen:
https://mcast.edu.mt

it's a collage name.
maybe we don't need to add it into dict.

PR #11028, opened 2023-06-12
fix(tls): fix incompatible tls options and issue with tls version gap

id:
Fixes https://emqx.atlassian.net/browse/EMQX-10245.

See also https://github.com/emqx/ranch/pull/1.

Need some discussion on how to handle this best. With this patch + patched ranch it works fine, but the UI does not reflect version gap fix. I.e. if a user sets TLS versions to '1.3', '1.1', we will be using '1.3' only on backend, but the UI will still be showing '1.3', '1.1'.

### <samp>🤖 Generated by Copilot at 599367c</samp>

Improved TLS options validation and adjustment in `emqx_tls_lib.erl`. Renamed `filter` to `ensure_valid_options` and added warnings for possible issues.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_tls_lib.erl
+++ b/apps/emqx/src/emqx_tls_lib.erl
@@ -478,7 +478,7 @@ to_server_opts(Type, Opts) ->
     Versions = integral_versions(Type, maps:get(versions, Opts, undefined)),
     Ciphers = integral_ciphers(Versions, maps:get(ciphers, Opts, undefined)),
     Path = fun(Key) -> resolve_cert_path_for_read_strict(maps:get(Key, Opts, undefined)) end,
-    filter(
+    ensure_valid_options(
         maps:to_list(Opts#{
             keyfile => Path(keyfile),
             certfile => Path(certfile),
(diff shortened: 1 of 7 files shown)


apps/emqx/src/emqx_tls_lib.erl
```diff
@@ -556,33 +556,96 @@ resolve_cert_path_for_read_strict(Path) ->
...
 
-filter([], _) ->
-    [];
-filter([{_, undefined} | T], Versions) ->
-    filter(T, Versions);
-filter([{_, ""} | T], Versions) ->
-    filter(T, Versions);
-filter([{K, V} | T], Versions) ->
+ensure_valid_options(Options, Versions0) ->
+    Versions = validate_version_gap(Versions0),
```
SergeTupchiy, line 560: `Versions` are updated here and used in the next function call to validate other opts, but returned Options will still hold original `Versions`. 
I suggest moving version gap validation to emqx_schema, so we will disallow listener create/update with the version gap and return an error message to a user if it happens.

SergeTupchiy: moved to emqx_schema

HJianBo (approved):

zhongwencool (approved):

yanzhiemq:
_

yanzhiemq:
_

PR #11065, opened 2023-06-15
fix: avoid logging unnecessary errors in async cleanup functions

SergeTupchiy:
Cleanup functions that access ETS tables may fail with `badarg` error during EMQX shutdown. 
They are called asynchronously by `emqx_pool` workers and accessed ETS tables may be already destroyed as their owners are shut down. This fix catches ETS `badarg` errors silently and individually, before they can be caught and logged by `emqx_pool`.

Fixes EMQX-9992

### <samp>🤖 Generated by Copilot at 86e3cdb</samp>

This pull request refactors the code that uses ets tables in various modules to improve error handling and readability. It introduces a new module `emqx_utils_ets` that provides wrapper functions and a macro for ets operations. It also moves some broker-related functions to the `emqx_broker` module.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_broker_helper.erl
+++ b/apps/emqx/src/emqx_broker_helper.erl
@@ -153,13 +153,17 @@ code_change(_OldVsn, State, _Extra) ->
 %%--------------------------------------------------------------------
 
 clean_down(SubPid) ->
-    case ets:lookup(?SUBMON, SubPid) of
-        [{_, SubId}] ->
-            true = ets:delete(?SUBMON, SubPid),
-            true =
-                (SubId =:= undefined) orelse
-                    ets:delete_object(?SUBID, {SubId, SubPid}),
-            emqx_broker:subscriber_down(SubPid);
-        [] ->
-            ok
+    try
+        case ets:lookup(?SUBMON, SubPid) of
+            [{_, SubId}] ->
+                true = ets:delete(?SUBMON, SubPid),
+                true =
+                    (SubId =:= undefined) orelse
+                        ets:delete_object(?SUBID, {SubId, SubPid}),
+                emqx_broker:subscriber_down(SubPid);
+            [] ->
+                ok
+        end
+    catch
+        error:badarg -> ok
     end.
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -734,7 +734,11 @@ code_change(_OldVsn, State, _Extra) ->
 %%--------------------------------------------------------------------
 
 clean_down({ChanPid, ClientId}) ->
-    do_unregister_channel({ClientId, ChanPid}),
+    try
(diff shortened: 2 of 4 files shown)


apps/emqx_utils/src/emqx_utils_ets.erl
```diff
@@ -44,18 +59,18 @@ new(Tab, Opts) ->
             ok
     end.
 
+-spec lookup(ets:tab(), term()) -> [tuple()].
+lookup(Tab, Key) ->
+    ?safe_ets(ets:lookup(Tab, Key), []).
```
zhongwencool, line 64: Is it really necessary to add a try-catch block in every ets operation ? 
Maybe the cost too high? is it possible to only handle this during shutdown?

SergeTupchiy: I also thought about adding a generic `try catch` block directly in the cleanup but ultimately opted for this solution instead. 
As the goal is to avoid extra logging, it seems slightly risky to catch and any `badarg` error as it's quite generic (may be potentially triggered by other code besides ETS). 
As to the performance cost, I think `try catch` is cheap enough as long as stack trace is not inspected.

keynslug:
My 5 cents: sounds like startup order / dependency problem to me, that is usually generally solvable on the supervision tree level. For example, with `emqx_cm` the problem seems to be that common `emqx_pool` outlives `emqx_cm`, I'd say we need to start dedicated `emqx_cm_pool` and put it _after_ `emqx_cm` in the startup order (so it shuts down before). Nice thing is it would work with any resources that might be tied to a process lifetime (or its dependencies), not just ETS tables.

SergeTupchiy:
@keynslug,
You are totally right about the root problem and your idea must be a good alternative. 
Given all the feedback, I'll revisit this PR and implement something smarter (probably dedicated pools as you suggest).

apps/emqx_utils/src/emqx_utils_ets.erl
lafirest: IMO, I think this fix introduces a side effect for the general call, keeping the affected scope in the `clean up` would be better.
At the same time, I think try-catch is always expensive because it needs to save the calling environment for restoration when some exception happens

zmstone: try catch without attempting to retrieve the stacktrace should not impact performance in a observable way.
we can try to run some simple benchmark test to verify it.

zmstone: I agree with zhongwen and firest that hiding badarg error in a generic call path may hinder some potential issues especially during tests.

SergeTupchiy: Thanks for the feedback!
After additional discussion with @zmstone I've removed safe ETS wrappers and added `try .. catch` only in relevant shutdown functions that use async worker pool.

PR #11090, opened 2023-06-19
feat(gcp_pubsub_consumer): implement GCP PubSub Consumer bridge

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10281

This PR contains only basic tests, as it's already quite big.  There'll be a follow up PR adding more comprehensive tests.

It's recommended to review this by going through the commits.

Example config:

```hcl
bridges {
  gcp_pubsub_consumer {
    my_bridge {
      connect_timeout = "15s"
      consumer {
        ack_retry_interval = "5s"
        pull_max_messages = 10
        pull_worker_multiplier = 3
        topic_mapping = [
          {
            mqtt_topic = "pubsub/in/1"
            payload_template = "${.}"
            pubsub_topic = "my-topic"
            qos = 1
          },
          {
            mqtt_topic = "pubsub/in/2"
            payload_template = "${.value}"
            pubsub_topic = "my-topic2"
            qos = 2
          }
        ]
      }
      enable = true
      max_retries = 0
      pipelining = 1
      pool_size = 8
      resource_opts {
        health_check_interval = "15s"
        request_ttl = "45s"
        worker_pool_size = 16
      }
      service_account_json {
        # ...
      }
    }
  }
}
```

### <samp>🤖 Generated by Copilot at 8ede004</samp>

This pull request adds the consumer bridge functionality for the `emqx_bridge_gcp_pubsub` module, which allows EMQ X to subscribe to GCP PubSub topics and forward messages to MQTT topics. It also refactors and improves the existing producer bridge module, adds more tests and code quality checks, and updates some dependencies and configurations.

## PR Checklist


apps/emqx/test/emqx_static_checks_data/5.0.bpapi
thalesmg, line 1: I had to manually fix the `query_mode` type because there was a mismatch between the type, the typespecs, and the actual values used.

paulozulato (approved):

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_worker.erl
```diff
@@ -0,0 +1,552 @@
...
+            ?tp(
+                debug,
+                "gcp_pubsub_consumer_worker_subscription_ready",
+                #{instance_id => InstanceId}
+            ),
+            ?MODULE:pull_async(self()),
+            optvar:set(?OPTVAR_SUB_OK(self()), subscription_ok),
+            {noreply, State0};
+        error ->
+            %% FIXME: add delay if topic does not exist?!
```
paulozulato, line 179: will the worker be responsible while topic doesn't exist (or the process will keep looping here and ignoring external messages)?

apps/emqx_utils/src/emqx_utils_maps.erl
```diff
@@ -293,3 +294,12 @@ if_only_to_toggle_enable(OldConf, Conf) ->
...
     end.
+
+%% Like `maps:update_with', but does nothing if key does not exist.
+update_if_present(Key, Fun, Map) ->
+    case Map of
+        #{Key := Val} ->
+            Map#{Key := Fun(Val)};
+        _ ->
+            Map
+    end.
```
paulozulato, line 305:
```
update_if_present(Key, Fun, Map) when is_map_key(Key, Map) ->
    Map#{Key := Fun(Val)};
update_if_present(_Key, _Fun, Map) ->
    Map.
```

apps/emqx_bridge_gcp_pubsub/rebar.config
```diff
@@ -1,10 +1,26 @@
...
+{erl_opts, [
+    warn_unused_vars,
+    warn_shadow_vars,
+    warn_unused_import,
+    warn_obsolete_guard,
+    warnings_as_errors,
+    debug_info
+]}.
 {deps, [ {emqx_connector, {path, "../../apps/emqx_connector"}}
        , {emqx_resource, {path, "../../apps/emqx_resource"}}
```
savonarola, line 11: Nit: m.b. format `rebar.config` too?

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_impl_producer.erl
```diff
@@ -209,3 +215,71 @@ publish_path(
...
+
+reply_delegator(ReplyFunAndArgs, Response) ->
+    case Response of
+        {error, Reason} when
+            Reason =:= econnrefused;
+            %% this comes directly from `gun'...
+            Reason =:= {closed, "The connection was lost."};
+            Reason =:= timeout
+        ->
+            Result1 = {error, {recoverable_error, Reason}},
```
savonarola, line 281: Nit: Seems, there is no `Result` or `Result0` around 😅

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_connector.erl
```diff
@@ -76,11 +88,22 @@ on_start(
...
-    HostPort = os:getenv("PUBSUB_EMULATOR_HOST", "pubsub.googleapis.com:443"),
+    {Transport, HostPort} =
+        case os:getenv("PUBSUB_EMULATOR_HOST") of
+            false ->
+                {tls, "pubsub.googleapis.com:443"};
+            HostPort0 ->
+                %% The emulator is plain HTTP...
+                Transport0 = persistent_term:get({?MODULE, transport}, tcp),
+                {Transport0, HostPort0}
+        end,
```
savonarola, line 99: Nit: the code seems to be quite overloaded with different kinds injection (through env, through persistent term).

Maybe introduce some kind of injectable `emqx_constant`

```
{Transport, HostPort} = emqx_constant:get(gcp_pubsub, {tls, "pubsub.googleapis.com:443"})
```

or just move to a separate fun...

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_impl_consumer.erl
```diff
@@ -0,0 +1,200 @@
...
+
+start_consumers(InstanceId, ConnectorState, Config) ->
+    #{
+        bridge_name := BridgeName,
+        consumer := ConsumerConfig0,
+        hookpoint := Hookpoint
+    } = Config,
+    ConsumerConfig1 = maps:update_with(topic_mapping, fun convert_topic_mapping/1, ConsumerConfig0),
+    TopicMapping = maps:get(topic_mapping, ConsumerConfig1),
+    PullWorkerMultiplier = maps:get(pull_worker_multiplier, ConsumerConfig1),
```
savonarola, line 94: Nit: m.b. `pull_worker_multiplier ` -> `consumer_workers_per_topic`? The name was a bit unclear until this point.

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_worker.erl
```diff
@@ -0,0 +1,552 @@
...
+%%-------------------------------------------------------------------------------------------------
+%% `emqx_resource' API
+%%-------------------------------------------------------------------------------------------------
+
+connect(Opts0) ->
+    Opts = maps:from_list(Opts0),
+    #{
+        ack_retry_interval := AckRetryInterval,
+        bridge_name := BridgeName,
+        connector_state := ConnectorState,
```
savonarola, line 130: This is a bit disturbing 😅

Because the state is supposed to be controlled by the connector, but it is "multiplied" several times and passed to workers, where it is no more controlled by the connector.

We seem to use the implicit knowledge that the connector's state is actually some constant settings that are valid as long as the workers are alive. Maybe we should at least reflect this fact in the naming somehow?

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_worker.erl
```diff
@@ -0,0 +1,552 @@
...
+            case Reason of
+                #{status_code := 409} ->
+                    %% the subscription was not found; deleted?!
+                    ?MODULE:ensure_subscription(WorkerPid);
+                _ ->
+                    ?MODULE:pull_async(WorkerPid)
+            end;
+        {ok, #{status_code := 200, body := RespBody}} ->
+            ?MODULE:process_pull_response(WorkerPid, RespBody)
+    end.
```
savonarola, line 98: It's a bit difficult to trace that all paths finally renew `pull_async`. Also, we must ensure that a call to `reply_delegator` should never be lost.

Maybe we also add some timeout-based `pull_async`?

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_worker.erl
thalesmg: I think we should treat an nonexistent topic as in your PR: https://github.com/emqx/emqx/pull/11090/files#diff-e339f277c61f643708ab4199f7cc1d3af03beb7772ff6fcd0914d5c3e4ea20b9R105

apps/emqx_utils/src/emqx_utils_maps.erl
thalesmg: We still need to extract the value from the map, so I wonder if it's better with clauses? :thinking:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_worker.erl
thalesmg: :heavy_check_mark:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_worker.erl
thalesmg: I've experimented managing the connector as an individual resource and interacting with it via `emqx_resource:simple_{,a}sync_query`, but that introduced some additional difficulties:

- that call still logs "errors" like the 409 status code when we ensure that a subscription exists;
- that also could potentially bump some metrics.

We _could_ make some changes to avoid such side effects as well, but I'm more inclined to simply add a comment in this particular bridge warning to _not_ attempt to change the connector state value in the `on_get_status` callback, for simplicity.  :thinking: :see_no_evil:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_impl_consumer.erl
thalesmg: :heavy_check_mark:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_connector.erl
thalesmg: :heavy_check_mark:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_impl_producer.erl
thalesmg: :heavy_check_mark:

apps/emqx_bridge_gcp_pubsub/rebar.config
thalesmg: :heavy_check_mark:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_worker.erl
thalesmg: After our discussion, I've transformed the connector into an opaque client.  I think it's cleaner now.  :beers:

paulozulato (approved):
I'm just wondering if we could prevent the Client to be changed somehow...

thalesmg:
> I'm just wondering if we could prevent the Client to be changed somehow...

Now that the connector no longer implements `emqx_resource` behavior, it doesn't have an `on_get_status` to change its internal values.

savonarola (approved):

PR #11108, opened 2023-06-20
Durable storage

ieQu1:
Fixes EMQX-10332

### <samp>🤖 Generated by Copilot at 6191bfe</samp>

This pull request adds a new app `emqx_durable_storage` to the `emqx` repository, which provides a way of storing and replaying MQTT messages and sessions using rocksdb as the underlying database. The app consists of several modules that implement the application logic, the configuration options, the storage layer, the storage backends, and the API functions.

## PR Checklist


keynslug (approved):

apps/emqx_durable_storage/src/emqx_ds_message_storage_bitmask.erl
```diff
@@ -0,0 +1,731 @@
...
+    timestamp_bits := TimestampBits,
+    topic_bits_per_level := BitsPerLevel,
+    epoch := MaxEpoch
+}) ->
+    TimestampLSBs = min(TimestampBits, floor(math:log2(MaxEpoch))),
+    TimestampMSBs = TimestampBits - TimestampLSBs,
+    NLevels = length(BitsPerLevel),
+    {LevelBits, [TailLevelsBits]} = lists:split(NLevels - 1, BitsPerLevel),
+    Source = lists:flatten([
+        [{timestamp, TimestampLSBs, TimestampMSBs} || TimestampMSBs > 0],
```
thalesmg, line 262: Possibly a dumb question: if `TimestampLSBs` is an [offset](https://github.com/emqx/emqx/pull/11114#event-9605545761), shouldn't this be the other way around?  i.e.: `{timestamp, MSBs, LSBs}` (skip the MSBs and take the LSBs)?

thalesmg (commented):
Still going through :book:

apps/emqx_durable_storage/src/emqx_ds_message_storage_bitmask.erl
keynslug: Heh, I could probably understand the confusion. So, here we're saying "take `TimestampMSBs` bits starting for `TimestampLSBs`th one". But that bitsource actually denotes part of a bitstring laying in the MSBs region, i.e. for consistency we'd need to reverse this list. It's was just easier to deal with this list in reversed order IIRC.

apps/emqx_durable_storage/src/emqx_ds_message_storage_bitmask.erl
```diff
@@ -0,0 +1,731 @@
...
+%% (a.k.a epoch)                    (a.k.a time offset)
+%%
+%% Topic hash is level-aware: each topic level is hashed separately
+%% and the resulting hashes are bitwise-concatentated. This allows us
+%% to map topics to fixed-length bitstrings while keeping some degree
+%% of information about the hierarchy.
+%%
+%% Next important concept is what we call "epoch". It is time
+%% interval determined by the number of least significant bits of the
+%% timestamp found at the tail of the rocksdb key.
```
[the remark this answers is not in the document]
ieQu1: Technically correct, but I agree that the phrasing is horrible.
What it means is _duration_ of the epoch is determined by the number of LsB. But epoch itself is stored in MsB of the key.

apps/emqx_durable_storage/src/emqx_ds_message_storage_bitmask.erl
```diff
@@ -0,0 +1,731 @@
...
+}.
+
+%% Persistent configuration of the generation, it is used to create db
+%% record when the database is reopened
+-record(schema, {keymapper :: keymapper()}).
+
+-opaque schema() :: #schema{}.
+
+-record(db, {
+    shard :: emqx_ds:shard(),
```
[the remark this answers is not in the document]
ieQu1: Good catch (strage that dialyzer ignored that), moved these types to `emqx_ds` module.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -0,0 +1,505 @@
...
+schema_put_current(DBHandle, GenId) ->
+    rocksdb:put(DBHandle, ?CURRENT_GEN, integer_to_binary(GenId), ?SCHEMA_WRITE_OPTS).
+
+-spec schema_gen_key(integer()) -> binary().
+schema_gen_key(N) ->
+    <<"gen", N:32>>.
+
+-undef(CURRENT_GEN).
+-undef(SCHEMA_WRITE_OPTS).
+-undef(SCHEMA_READ_OPTS).
```
thalesmg, line 422: Q: why do we undefined those macros here?

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -0,0 +1,505 @@
...
+%% Internal functions
+%%================================================================================
+
+-record(db, {handle :: rocksdb:db_handle(), cf_iterator :: rocksdb:cf_handle()}).
+
+-spec populate_metadata(state()) -> ok.
+populate_metadata(S = #s{shard = Shard, db = DBHandle, cf_iterator = CFIterator}) ->
+    ok = meta_put(Shard, db, #db{handle = DBHandle, cf_iterator = CFIterator}),
+    Current = schema_get_current(DBHandle),
+    lists:foreach(fun(GenId) -> populate_metadata(GenId, S) end, lists:seq(0, Current)).
```
thalesmg, line 221: Q: this seems to potentially set the `current` gen multiple times for a shard while it's starting; since `persistent_term` may perform global GC for updates, wouldn't it be better to set the current generation after all generation metadata keys have been populated?

thalesmg (approved):
:wine_glass: 

(Still going through the tests :books: :moyai: )

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
keynslug: Overall, yes it would. However, `current` gen here is just a (short) integer, persistent term GC for such updates should be no-op.

apps/emqx_durable_storage/src/emqx_ds_message_storage_bitmask.erl
```diff
@@ -0,0 +1,731 @@
...
+    State = #{
+        v => 1,
+        cursor => Cursor
+    },
+    term_to_binary(State).
+
+-spec restore_iterator(db(), emqx_ds:replay(), binary()) ->
+    {ok, iterator()} | {error, _TODO}.
+restore_iterator(DB, Replay, Serial) when is_binary(Serial) ->
+    State = binary_to_term(Serial),
```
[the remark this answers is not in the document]
ieQu1: Not sure if it's necessary here, since iterators come from a trusted source.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
ieQu1: These macros are local, there's no need to expose them to the entire module.

PR #11117, opened 2023-06-21
fix(emqx_rule_engine): update metrics using gproc_ps

sstrigler:
Fixes [EMQX-8842](https://emqx.atlassian.net/browse/EMQX-8842)

## Summary
copilot:summary

## PR Checklist


apps/emqx_resource/src/emqx_resource_metrics.erl
```diff
@@ -209,6 +209,7 @@ dropped_inc(ID) ->
 dropped_inc(_ID, 0) ->
     ok;
 dropped_inc(ID, Val) ->
+    publish_event(dropped, {ID, Val}),
```
thalesmg, line 212: Since we're already executing a telemetry handler, perhaps the rule engine could its own handler for such events rather than having this separate function? :thinking:

thalesmg: I see.  Haven't looked too much into this, but perhaps the rule engine/actions could inject their own `{ReplyFun, [Args]}` callback when calling the bridge, so their metrics can be triggered. :thinking: 

I think that would also require a bit of adjustment in `buffer_worker` to read such `reply_to => {ReplyFun, [Args]}` callbacks from the query options.

sstrigler: Yes, that's the approach I wanted to avoid for better decoupling.

PR #11126, opened 2023-06-22
fix(emqx_rule_engine): set inc_action_metrics as async_reply_fun

sstrigler:
Fixes [EMQX-8842](https://emqx.atlassian.net/browse/EMQX-8842)

Next try. This time using a callback.

### <samp>🤖 Generated by Copilot at 5c23ed7</samp>

Added support for asynchronous replies from bridge resources in rule engine and resource buffer. This allows bridge actions to update metrics and tracing data after sending or receiving messages. Modified `emqx_bridge`, `emqx_resource_buffer_worker`, and `emqx_rule_runtime` modules to implement this feature.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge.erl
+++ b/apps/emqx_bridge/src/emqx_bridge.erl
@@ -220,15 +220,14 @@ send_to_matched_egress_bridges(Topic, Msg) ->
 send_message(BridgeId, Message) ->
     {BridgeType, BridgeName} = emqx_bridge_resource:parse_bridge_id(BridgeId),
     ResId = emqx_bridge_resource:resource_id(BridgeType, BridgeName),
-    send_message(BridgeType, BridgeName, ResId, Message, undefined).
+    send_message(BridgeType, BridgeName, ResId, Message, #{}).
 
-send_message(BridgeType, BridgeName, ResId, Message, ReplyTo) ->
+send_message(BridgeType, BridgeName, ResId, Message, QueryOpts0) ->
     case emqx:get_config([?ROOT_KEY, BridgeType, BridgeName], not_found) of
         not_found ->
             {error, bridge_not_found};
         #{enable := true} = Config ->
-            QueryOpts0 = query_opts(Config),
-            QueryOpts = QueryOpts0#{reply_to => ReplyTo},
+            QueryOpts = maps:merge(query_opts(Config), QueryOpts0),
             emqx_resource:query(ResId, {send_message, Message}, QueryOpts);
         #{enable := false} ->
             {error, bridge_stopped}
--- a/apps/emqx_bridge/src/emqx_bridge.erl
+++ b/apps/emqx_bridge/src/emqx_bridge.erl
@@ -228,7 +228,7 @@ send_message(BridgeType, BridgeName, ResId, Message, ReplyTo) ->
             {error, bridge_not_found};
         #{enable := true} = Config ->
             QueryOpts0 = query_opts(Config),
-            QueryOpts = QueryOpts0#{async_reply_fun => ReplyTo},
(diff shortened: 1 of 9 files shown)


kjellwinblad (commented):
LGTM. The only comments that I have are the things we discussed before. Maybe async_reply_fun is not the optimal name for the field as we use it even when the "query" is not async. Maybe on_complete_reply_fun would be better? Is async_reply_fun used by something else? Could it be "overwritten" somewhere in the call chain? Also as we discussed before, it would be good with a test case.

apps/emqx_bridge/src/emqx_bridge.erl
```diff
@@ -220,14 +220,15 @@ send_to_matched_egress_bridges(Topic, Msg) ->
 send_message(BridgeId, Message) ->
     {BridgeType, BridgeName} = emqx_bridge_resource:parse_bridge_id(BridgeId),
     ResId = emqx_bridge_resource:resource_id(BridgeType, BridgeName),
-    send_message(BridgeType, BridgeName, ResId, Message).
+    send_message(BridgeType, BridgeName, ResId, Message, undefined).
```
thalesmg, line 223: Nit:

```suggestion
    Opts = #{reply_to => undefined},
    send_message(BridgeType, BridgeName, ResId, Message, Opts).
```

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -158,7 +158,9 @@ simple_async_query(Id, Request, QueryOpts0) ->
     Ref = make_request_ref(),
     Result = call_query(async_if_possible, Id, Index, Ref, ?SIMPLE_QUERY(Request), QueryOpts),
     _ = handle_query_result(Id, Result, _HasBeenSent = false),
-    Result.
+    maybe_apply_async_reply_fun(
```
thalesmg, line 161: I think we should leave the calling of the reply function to `do_reply_caller`, and just inject the callback as the `From` here instead of `undefined`.

i.e., something like:

```erlang
  ReplyTo = maps:get(reply_to, QueryOpts, undefined),
  Query = ?QUERY(ReplyTo, Request, _HasBeenSeen = false, _ExpireAt = infinity),
  ...
```

and then I think `do_reply_caller` will call `{Fun, Args}` from `From` without further changes.

apps/emqx_rule_engine/test/emqx_rule_engine_SUITE.erl
```diff
@@ -2822,6 +2868,114 @@ t_get_rule_ids_by_action_reference_ingress_bridge(_Config) ->
...
+    RuleId = <<"rule:test_metrics_bridge_action">>,
+    {ok, #{id := RuleId}} =
+        emqx_rule_engine:create_rule(
+            #{
+                id => RuleId,
+                sql => <<"SELECT * FROM \"topic/#\"">>,
+                actions => [BridgeId]
+            }
+        ),
+    timer:sleep(100),
```
kjellwinblad, line 2954: minor comment: to avoid the risk of creating flake test and not making the tests slow to run one could loop until the the expected condition is met and fail if it is not met after a certain amount of time (see https://github.com/klarna/snabbkaffe/blob/master/src/snabbkaffe.erl#L313)

kjellwinblad (approved):

ieQu1 (approved):

thalesmg (approved):

sstrigler:
I’ll create a follow-up ticket

sstrigler:
https://emqx.atlassian.net/issues/?jql=issueKey%20in%20(EMQX-10470%2CEMQX-10471)

PR #11132, opened 2023-06-23
feat(authz): use extensible map format for actions in authz rules

savonarola:
Fixes [EMQX-9766](https://emqx.atlassian.net/browse/EMQX-9766)

## Problem

We have to control whether a client can write retain messages or subscribe/publish using certain QoS levels.

## Current model

In authz, we have
* *subjects* — "who" makes a request/action
* *permissions* — resolution whether to allow an action or not, `allow` and `deny`
* *resources* — "what" is allowed. We have resources split into `topic` and `action` (publish, subscribe, all)

## New model

The current model is not very extensible. We have hotwired primitive types for all resources. We need to extend the model to support more complex resources.

Since the topic is considered to be more or less "constant" data-like structure (e.g. different clients may subscribe the same topic in a different manner), we choose to extend the `action` part of the resource.
This also more or less reflects the MQTT protocol itself, where the QoS/Retain are a part of the PUBLISH/SUBSCRIBE packet: different clients subscribe differently to the same topic.

### Internal model

Internally we extend the signature of the `emqx_access_control:authorize/3` callback.
From
```erlang
-type pubsub() :: publish | subscribe.
...
-spec authorize(emqx_types:clientinfo(), emqx_types:pubsub(), emqx_types:topic()) ->
    allow | deny.
```

To
```erlang
-type pubsub() ::
    #{action_type := subscribe, qos := qos()}
    | #{action_type := publish, qos := qos(), retain := boolean()}.
...
-spec authorize(emqx_types:clientinfo(), emqx_types:pubsub(), emqx_types:topic()) ->
    allow | deny.
```

Thus, we make `emqx_types:pubsub()` extensible if needed.

We update all the entry points (including gateways) to pass the new extended action to the `authorize/3` function.

### External model

That is, how the user will configure the authorization rules.

We now explicitly differentiate between three types of rules:
* `raw` rules: rules obtained as plain data structures from a database or API, having the format like:
```erlang
#{
    <<"permission">> => <<"allow">>,
    <<"topic">> => <<"a/b/c">>,
    <<"action">> => <<"subscribe">>,
    <<"retain">> => <<"true">>,
    <<"qos">> => <<"1,2">>
}
%% or
#{
    <<"permission">> => <<"allow">>,
    <<"topic">> => [<<"a/b/c">>, <<"d">>],
    <<"action">> => <<"subscribe">>,
    <<"retain">> => <<"true">>,
    <<"qos">> => [1, 2]
}
...
```
We have a module `emqx_authz_rule_raw` to convert between `raw` and `dsl` rules.
* `dsl` rules: rules described with Erlang data structures. They are used by `mnesia` and `file` backends.
We use `emqx_authz_rule` to convert from `dsl` to `compiled` rules.
* `compiled` rules: slightly converter `dsl` rules, used for final matching.

We refactor the backends to utilize the common parser and validator of `raw` rules.

## Test refactoring

Previous test implementation was not very flexible and was not able to test the new model.
The problem was that the test both "inserted" the test data into a database and then also "queried" the database to check the results. However, many bugs may come from how the data is inserted into the database in the wild, because this is the part that is more likely to be customized by the user.

So we refactor the tests to be more like authn tests, where for each test we have its own storage schema and inserted data.
Example:
```erlang
...
#{
    name => qos_retain_in_query_result_as_integer,
    setup => [
        "CREATE TABLE acl(username VARCHAR(255), topic VARCHAR(255), "
        "permission VARCHAR(255), action VARCHAR(255),"
        "qos_i VARCHAR(255), retain_i VARCHAR(255))",

        "INSERT INTO acl(username, topic, permission, action, qos_i, retain_i)"
        " VALUES('username', 't1', 'allow', 'publish', 1, 1)"
    ],
    query =>
        "SELECT permission, action, topic, qos_i as qos, retain_i as retain"
        " FROM acl WHERE username = ${username}",
    client_info => #{
        username => <<"username">>
    },
    checks => [
        {allow, ?PUBLISH_ACTION(1, true), <<"t1">>},
        {deny, ?PUBLISH_ACTION(1, false), <<"t1">>},
        {deny, ?PUBLISH_ACTION(0, true), <<"t1">>}
    ]
},
...
```
Here we explicitly check that `retain` flag stored as an integer is handled correctly.

### <samp>🤖 Generated by Copilot at d49e101</samp>

This pull request enhances the authorization logic and features of the EMQ X broker by adding support for QoS and retain properties in the publish action, and by refactoring and improving the code quality and compatibility of the authorization modules and test suites. It also introduces a new placeholder for the retain flag, a new module for parsing and formatting raw rules, and some new macros and types for the action conditions. The pull request affects several files in the `apps/emqx_authz` and `apps/emqx` directories.

## PR Checklist

- [x] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up: [EMQX-10404](https://emqx.atlassian.net/browse/EMQX-10404) 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,14 +130,14 @@ $(foreach app,$(APPS),$(eval $(call gen-app-prop-target,$(app))))
 ct-suite: $(REBAR) merge-config clean-test-cluster-config
 ifneq ($(TESTCASE),)
 ifneq ($(GROUP),)
-	$(REBAR) ct -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE) --group $(GROUP)
+	$(REBAR) ct -c -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE) --group $(GROUP)
 else
-	$(REBAR) ct -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE)
+	$(REBAR) ct -c -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE)
 endif
 else ifneq ($(GROUP),)
-	$(REBAR) ct -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --group $(GROUP)
+	$(REBAR) ct -c -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --group $(GROUP)
 else
-	$(REBAR) ct -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)
+	$(REBAR) ct -c -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)
 endif
 
 .PHONY: cover
--- a/Makefile
+++ b/Makefile
@@ -130,14 +130,14 @@ $(foreach app,$(APPS),$(eval $(call gen-app-prop-target,$(app))))
 ct-suite: $(REBAR) merge-config clean-test-cluster-config
 ifneq ($(TESTCASE),)
 ifneq ($(GROUP),)
-	$(REBAR) ct -c -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE) --group $(GROUP)
+	$(REBAR) ct -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE) --group $(GROUP)
 else
-	$(REBAR) ct -c -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE)
+	$(REBAR) ct -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE)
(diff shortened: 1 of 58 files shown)


Makefile
```diff
@@ -130,14 +130,14 @@ $(foreach app,$(APPS),$(eval $(call gen-app-prop-target,$(app))))
...
 else
-	$(REBAR) ct -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE)
+	$(REBAR) ct -c -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --case $(TESTCASE)
 endif
 else ifneq ($(GROUP),)
-	$(REBAR) ct -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --group $(GROUP)
+	$(REBAR) ct -c -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)  --group $(GROUP)
 else
-	$(REBAR) ct -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)
+	$(REBAR) ct -c -v --readable=$(CT_READABLE) --name $(CT_NODE_NAME) --suite $(SUITE)
```
thalesmg, line 140: Nit: I wouldn't add `-c` by default here, as enabling cover compile slows down compilation/test execution _a lot_.

I use `ct-suite` as a way to run tests more quickly even if there's a single test module in an app exactly because of that.  :sweat_smile:

thalesmg (dismissed):
I still haven't gone through the tests, but looks good!  :beers:

id (approved):

PR #11134, opened 2023-06-24
fix: mask the value of the authorization header whatever its case

lafirest:
Fixes [EMQX-10629](https://emqx.atlassian.net/browse/EMQX-10269) 

### <samp>🤖 Generated by Copilot at d485f8f</samp>

Enhance `is_sensitive_key` function to filter out authorization header. Add test case for `emqx_utils.erl`.

## PR Checklist

### DIFF
--- a/apps/emqx_utils/src/emqx_utils.erl
+++ b/apps/emqx_utils/src/emqx_utils.erl
@@ -643,10 +643,7 @@ is_sensitive_key(<<"jwt">>) -> true;
 is_sensitive_key(authorization) -> true;
 is_sensitive_key("authorization") -> true;
 is_sensitive_key(<<"authorization">>) -> true;
-%% the authorization header is not case-sensitive
-is_sensitive_key("a" ++ _ = Key) -> is_authorization(Key);
-is_sensitive_key(<<"a", _/binary>> = Key) -> is_authorization(erlang:binary_to_list(Key));
-is_sensitive_key(_) -> false.
+is_sensitive_key(Key) -> is_authorization(Key).
 
 redact(Term) ->
     do_redact(Term, fun is_sensitive_key/1).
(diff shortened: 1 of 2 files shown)


apps/emqx_utils/src/emqx_utils.erl
```diff
@@ -777,6 +783,23 @@ redact2_test_() ->
     Keys = [secret, passcode],
     [{case_name(atom, Key), fun() -> Case(Key, Checker) end} || Key <- Keys].
 
+redact_is_authorization() ->
```
savonarola, line 786: seems the test is not run:

```suggestion
redact_is_authorization_test_() ->
```

but if enabled,it will probably fail on `"Authorization"` case 🤔

paulozulato: checking like that would solve this issue:
```
is_sensitive_key(jwt) -> true;
is_sensitive_key("jwt") -> true;
is_sensitive_key(<<"jwt">>) -> true;
is_sensitive_key(authorization) -> true;
is_sensitive_key(Key) when is_binary(Key) orelse is_list(Key) -> is_authorization(Key);
is_sensitive_key(_) -> false;
```

because any `"Authorization"` case would be checked by `is_authorization`, which converts to lowercase to check (the authorization string/atom case could be moved to is_authorization as well).

paulozulato (approved):

apps/emqx_utils/src/emqx_utils.erl
```diff
@@ -707,6 +707,19 @@ do_is_redacted(K, <<?REDACT_VAL>>, Fun) ->
 do_is_redacted(_K, _V, _Fun) ->
     false.
 
+%% This is ugly, however, the authorization is case-insensitive,
+%% the best way is to check chars one by one and quickly exit when any position is not equal,
+%% but in Erlang, this may not perform well, so here only check the first one
+is_authorization([Cap | _] = Key) when Cap == $a; Cap == $A ->
+    is_authorization2(Key);
+is_authorization(<<Cap, _/binary>> = Key) when Cap == $a; Cap == $A ->
```
savonarola, line 715: Nit
```suggestion
is_authorization(<<Cap, _/binary>> = Key) when Cap =:= $a; Cap =:= $A ->
```

savonarola (approved):

apps/emqx_utils/src/emqx_utils.erl
savonarola: I think the idea is to make a fast check before lowering each binary/string key.

PR #11148, opened 2023-06-27
fix(emqx_conf): don't sync the MFA to a leaved node

lafirest:
Fixes [EMQX-10113](https://emqx.atlassian.net/browse/EMQX-10113)

### <samp>🤖 Generated by Copilot at 7de4ef6</samp>

Added a feature to clean up cluster configuration commit records when a node leaves the cluster. Invoked this feature in the `leave` CLI command.

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/emqx_cluster_rpc.erl
+++ b/apps/emqx_conf/src/emqx_cluster_rpc.erl
@@ -282,7 +282,7 @@ on_mria_stop(_) ->
 
 %% @private
 init([Node, RetryMs]) ->
-    mria:register_callback(stop, fun ?MODULE:on_mria_stop/1),
+    register_mria_stop_cb(fun ?MODULE:on_mria_stop/1),
     {ok, _} = mnesia:subscribe({table, ?CLUSTER_MFA, simple}),
     State = #{node => Node, retry_interval => RetryMs, is_leaving => false},
     %% The init transaction ID is set in emqx_conf_app after
@@ -642,4 +642,23 @@ do_wait_for_emqx_ready2(N) ->
         false ->
             timer:sleep(100),
             do_wait_for_emqx_ready2(N - 1)
+    end.
+
+register_mria_stop_cb(Callback) ->
+    case mria_config:callback(stop) of
+        undefined ->
+            mria:register_callback(stop, Callback);
+        {ok, Previous} ->
+            mria:register_callback(
+                stop,
+                fun(Arg) ->
+                    Callback(Arg),
+                    case erlang:fun_info(Previous, arity) of
+                        {arity, 0} ->
+                            Previous();
+                        {arity, 1} ->
+                            Previous(Arg)
+                    end
+                end
+            )
     end.
--- a/apps/emqx_conf/src/emqx_cluster_rpc.erl
+++ b/apps/emqx_conf/src/emqx_cluster_rpc.erl
@@ -29,7 +29,7 @@
     status/0,
     skip_failed_commit/1,
     fast_forward_to_commit/2,
-    on_leave/0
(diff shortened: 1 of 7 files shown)


apps/emqx_conf/src/emqx_cluster_rpc.erl
```diff
@@ -353,7 +365,9 @@ catch_up(#{node := Node, retry_interval := RetryMs} = State, SkipResult) ->
         {aborted, Reason} ->
             ?SLOG(error, #{msg => "read_next_mfa_transaction_failed", error => Reason}),
             RetryMs
-    end.
+    end;
+catch_up(#{is_leaving := true}, _SkipResult) ->
```
paulozulato, line 369: shouldn't this function clause return a timeout as well?

apps/emqx_conf/src/emqx_cluster_rpc.erl
```diff
@@ -306,6 +315,9 @@ handle_call(skip_failed_commit, _From, State = #{node := Node}) ->
 handle_call({fast_forward_to_commit, ToTnxId}, _From, State) ->
     NodeId = do_fast_forward_to_commit(ToTnxId, State),
     {reply, NodeId, State, catch_up(State)};
+handle_call(on_leave, _From, State) ->
+    {atomic, ok} = transaction(fun ?MODULE:on_leave_clean/0, []),
```
ieQu1, line 319: Nit: maybe make it an anonymous fun to avoid unnecessary exports?

lafirest: I was told to be a good citizen for the hot upgrade, we should try to avoid using anonymous functions, although I know in V5 we don't support the hot upgrade now 
bildo

apps/emqx_management/src/emqx_mgmt_cli.erl
```diff
@@ -120,6 +120,7 @@ cluster(["join", SNode]) ->
             emqx_ctl:print("Failed to join the cluster: ~0p~n", [Error])
     end;
 cluster(["leave"]) ->
+    emqx_cluster_rpc:on_leave(),
     case ekka:leave() of
         ok ->
```
zhongwencool, line 125: What happen if ekka:leave failed but we already delete cluster_commit table ?

zhongwencool: we should also add delete cluster_commit when `force-leave Node` cli on line 132.
```
cluster(["force-leave", SNode]) ->
```

zhongwencool (changes requested):

apps/emqx_management/src/emqx_mgmt_cli.erl
lafirest: > What happen if ekka:leave failed but we already delete cluster_commit table ?

I have saw the ekka:leave code, it only fails if the node don'ts in the cluster, so I think here is safe.

ieQu1 (approved):

apps/emqx_conf/src/emqx_cluster_rpc.erl
```diff
@@ -624,3 +644,13 @@ do_wait_for_emqx_ready2(N) ->
...
             do_wait_for_emqx_ready2(N - 1)
     end.
+
+register_mria_stop_cb(Callback) ->
+    {ok, Previous} = mria_config:callback(stop),
+    mria:register_callback(
+        stop,
+        fun(Arg) ->
+            Callback(Arg),
+            Previous()
```
ieQu1, line 654: Please check arity of the previous callback: in the future it can change to /1

zhongwencool (approved):

PR #11163, opened 2023-06-29
fix(mongodb): hide `topology.pool_size` and fix it at 1

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10408

From an old conversation with @kjellwinblad:

> There are 3 pool_sizes
> - The buffer workers pool size, just exposed here: https://github.com/emqx/emqx/pull/9742
> - The topology.pool_size, which controls the pool size for the poolboy_pool in Kjell's
  diagram (on mongodb's side).
> - The pool_size from emqx_connector_mongo:mongo_fields that controls the ecpool pool
  size (on EMQX's side).
> So we actually want to set topology.pool_size = 1 and hide it from users.

### <samp>🤖 Generated by Copilot at 40aecf4</samp>

Hide `pool_size` option and set it to 1 for MongoDB plugin. This avoids confusion and improves compatibility with the MongoDB driver.

## PR Checklist

- [na] Added tests for the changes
- [na] Changed lines covered in coverage report

### DIFF
--- a/apps/emqx_mongodb/src/emqx_mongodb.erl
+++ b/apps/emqx_mongodb/src/emqx_mongodb.erl
@@ -97,7 +97,14 @@ fields(sharded) ->
     ] ++ mongo_fields();
 fields(topology) ->
     [
-        {pool_size, fun emqx_connector_schema_lib:pool_size/1},
+        {pool_size,
+            hoconsc:mk(
+                pos_integer(),
+                #{
+                    deprecated => {since, "5.1.1"},
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )},
         {max_overflow, fun max_overflow/1},
         {overflow_ttl, duration("overflow_ttl")},
         {overflow_check_period, duration("overflow_check_period")},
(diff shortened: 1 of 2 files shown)


thalesmg:
Will require changes in the frontend.  Attempt PR: https://github.com/emqx/emqx-dashboard5/pull/1568

apps/emqx_mongodb/src/emqx_mongodb.erl
```diff
@@ -97,7 +97,15 @@ fields(sharded) ->
...
 fields(topology) ->
     [
-        {pool_size, fun emqx_connector_schema_lib:pool_size/1},
+        {pool_size,
+            hoconsc:mk(
+                pos_integer(),
+                #{
+                    deprecated => {since, "5.1.1"},
+                    importance => ?IMPORTANCE_HIDDEN,
+                    converter => fun(_, _) -> 1 end
```
keynslug, line 106: Nit. IMO we don't have to go to such great lengths, default = 1 should be enough? Given the option is both deprecated and hidden. If we want to disregard it completely, isn't it better to set to 1 somewhere here?

https://github.com/emqx/emqx/blob/38bd8a81852caeda71ec58e874785c9736c49ef0/apps/emqx_mongodb/src/emqx_mongodb.erl#L370-L371

Also, does this converter work fine with `make_serializable`? Probably yes, but generally such converter seems "risky".

thalesmg: We also want to override any previously user-set values.

Hmmm I'll have to look into the `make_serializable` thing and where it's used.  But I assume it works, as we already do [something similar](https://github.com/emqx/emqx/blob/1ea92d400925fa5732868ec270071d680e7420ea/apps/emqx_bridge_mongodb/src/emqx_bridge_mongodb.erl#L54) in MongoDB bridge.

keynslug: 👍🏼 Got it. I have a feeling we need to work out some approach to how to deal with such deprecations.

Overall, for me it sounds more straightforward to just have `init_topology_options(R, [{pool_size, 1} | Acc]);` and tell a user though a log message that their configuration option is not being respected and why.

thalesmg: I've moved the override to the `on_start` callback and added a log to warn about the override.

keynslug (approved):

paulozulato (approved):

ieQu1 (approved):

PR #11164, opened 2023-06-29
fix(placeholder): add back support for nested keys inside binary encoded json

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10459

Without the need to call `json_decode` in the rule engine, we support placeholders with nested paths inside encoded JSON objects.

### <samp>🤖 Generated by Copilot at 9f929ca</samp>

This pull request adds support for nested variables in the payload templates for the emqx bridge to mysql. It modifies the `emqx_placeholder` module and its test suite, and also improves the test suite for the `emqx_bridge_mysql` module. The changes enable more flexible and dynamic data mapping between mqtt messages and mysql tables.

## PR Checklist

### DIFF
--- a/apps/emqx_utils/src/emqx_placeholder.erl
+++ b/apps/emqx_utils/src/emqx_placeholder.erl
@@ -257,7 +257,14 @@ quote_mysql(Str) ->
 
 lookup_var(Var, Value) when Var == ?PH_VAR_THIS orelse Var == [] ->
     Value;
-lookup_var([Prop | Rest], Data) ->
+lookup_var([Prop | Rest], Data0) ->
+    Data =
+        case emqx_utils_json:safe_decode(Data0, [return_maps]) of
+            {ok, Data1} ->
+                Data1;
+            {error, _} ->
+                Data0
+        end,
     case lookup(Prop, Data) of
         {ok, Value} ->
             lookup_var(Rest, Value);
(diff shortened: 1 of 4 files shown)


paulozulato (approved):

apps/emqx_utils/src/emqx_placeholder.erl
```diff
@@ -257,28 +257,8 @@ quote_mysql(Str) ->
...
-                error:{badkey, _} ->
-                    {error, undefined};
-                error:badarg ->
-                    {error, undefined}
-            end;
-        Value ->
-            {ok, Value}
-    end.
+lookup_var(Var, Data) ->
+    emqx_rule_maps:nested_get(Var, Data).
```
keynslug, line 261: Now `emqx_utils` depend on `emqx_rule_engine`. I specifically tried to avoid it before, thus this simple implementation.

I'd propose to just introduce another clause here, like in `emqx_rule_maps`.
https://github.com/emqx/emqx/blob/a757afcb7eca30c41fa049f70da4d2e4fc636612/apps/emqx_rule_engine/src/emqx_rule_maps.erl#L88-L89

<sup>(though I think that will muddle the interface slightly, and will have to hide some error conditions, it's still probably much simpler than having to introduce JSON decoding in all of this module's clients)</sup>

ieQu1 (approved):

PR #11175, opened 2023-06-30
fix(mysql_bridge): make nxdomain a 400 API error

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10460

image

### <samp>🤖 Generated by Copilot at 06b7d1b</samp>

This pull request improves the error handling and the SSL support of the emqx_bridge and emqx_mysql plugins. It returns a more accurate error code and message when starting a bridge with an invalid domain name, and it updates the `mysql` dependency to fix a connection bug with MySQL 8.0.26 or later.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_api.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_api.erl
@@ -985,9 +985,13 @@ call_operation(NodeOrAll, OperFunc, Args = [_Nodes, BridgeType, BridgeName]) ->
         {error, timeout} ->
             ?SERVICE_UNAVAILABLE(<<"Request timeout">>);
         {error, {start_pool_failed, Name, Reason}} ->
-            ?SERVICE_UNAVAILABLE(
-                bin(io_lib:format("Failed to start ~p pool for reason ~p", [Name, Reason]))
-            );
+            Msg = bin(io_lib:format("Failed to start ~p pool for reason ~p", [Name, Reason])),
+            case Reason of
+                nxdomain ->
+                    ?BAD_REQUEST(Msg);
+                _ ->
+                    ?SERVICE_UNAVAILABLE(Msg)
+            end;
         {error, not_found} ->
             BridgeId = emqx_bridge_resource:bridge_id(BridgeType, BridgeName),
             ?SLOG(warning, #{
--- a/apps/emqx_mysql/rebar.config
+++ b/apps/emqx_mysql/rebar.config
@@ -3,7 +3,7 @@
 {erl_opts, [debug_info]}.
 {deps, [
         %% NOTE: mind ecpool version when updating eredis_cluster version
-        {mysql, {git, "https://github.com/emqx/mysql-otp", {tag, "1.7.2"}}},
+        {mysql, {git, "https://github.com/emqx/mysql-otp", {tag, "1.7.3"}}},
         {emqx_connector, {path, "../../apps/emqx_connector"}},
         {emqx_resource, {path, "../../apps/emqx_resource"}}
 ]}.
--- a/changes/ee/fix-11175.en.md
+++ b/changes/ee/fix-11175.en.md
@@ -1,0 +1,1 @@
+Now when using a nonexistent hostname for connecting to MySQL will result in a 400 error rather than 503 in the HTTP API.


qzhuyan:
I don’t think returning 400 is the correct behavior. Users should be free to configure any server address even if it does not exist because the server could come up with new reg DNS name later.

thalesmg:
> I don’t think returning 400 is the correct behavior. Users should be free to configure any server address even if it does not exist because the server could come up with new reg DNS name later.

True, that's a possibility.  It's also possible that they made typo while inserting the URL, so that'll never resolve until they correct the hostname.

In both cases, I'd argue that the service unavailability comes from the user side: in the former case, they didn't provision the DNS entries yet; in the latter, the hostname is wrong.  So, for those two scenarios, a 4XX error seems more appropriate than a 503, which would indicate that EMQX itself is having problems.  Does this make sense?

qzhuyan:
I agree return 5xx is wrong. my point is as long as it is valid input, it should return 200. the pool should keep restarting until success.

thalesmg:
It will keep retrying, indeed.  

But this API is called, for example, when the user clicks the "Reconnect" button in the dashboard.  If we return 200 when there's a problem trying to start the bridge, I think it'll be misleading, since the "start" operation did fail.  We had some tickets about this in the past (e.g.: https://emqx.atlassian.net/browse/EMQX-10317).  So, if we do have some error information when attempting that operation, we should inform the user so it may help resolve the issue quicker.

sstrigler:
Did a similar thing for the webhooks bridge in the past. It's confusing to the user if they see an "ok" and then at the same time get an alarm that the bridge is not ok.

From a pure HTTP perspective I'd agree, a 200 makes more sense than a 400 but this is about UX and was requested to be changed like that.

Alternative (I think it had been discussed): return 200 but with the current status in the body that would show that it fails to connect, then the dashboard would have to be changed to display that status rather than the HTTP status code.

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -985,9 +985,13 @@ call_operation(NodeOrAll, OperFunc, Args = [_Nodes, BridgeType, BridgeName]) ->
         {error, timeout} ->
             ?SERVICE_UNAVAILABLE(<<"Request timeout">>);
         {error, {start_pool_failed, Name, Reason}} ->
-            ?SERVICE_UNAVAILABLE(
-                bin(io_lib:format("Failed to start ~p pool for reason ~p", [Name, Reason]))
-            );
+            Msg = bin(io_lib:format("Failed to start ~p pool for reason ~p", [Name, Reason])),
+            case Reason of
+                nxdomain ->
```
keynslug, line 990: Nit. This sounds like a somewhat limited (and probably not 100% correct) way to distinguish between "this service doesn't really exist there" and "this service doesn't work". I doubt if it's even possible to answer it correctly. For example, you'd have to also consider `econnrefused` / `econnreset` and a large swath of other posix errors, somehow make sure that this `nxdomain` isn't transient, make sure that if you established a connection, then make sure the other side is indeed mysql or something that follows protocol correctly, etc.

thalesmg: Indeed, I don't think it's possible to discern between those two cases.

The intention here was not to be comprehensive about all possible errors, but to give a starting point where we could distinguish between "there's a problem in EMQX" (i.e.: `ecpool` is crashing for unexpected reasons) and "there's a problem starting to the resource".

With that, then, we could extend that error handling for more cases as they arise, since, as you pointed out, there's a large swath of possible errors.  :sweat_smile:

thalesmg:
I've asked the product team (@Benniu ) for input on this discussion, and he believes we should return 400 for consistency with other bridges (and from a UI perspective too, I believe).

apps/emqx_bridge/src/emqx_bridge_api.erl
keynslug: Got it. I'd then let exceptions (i.e. _crashing for unexpected reasons_) propagate to the piece of code that converts them into `500`s, and convert everything else (i.e. expected errors) here into `400`s / `422`s. Currently, it's really odd that we're trying to tell the API client that _our_ API is unavailable when we cannot reach bridge target. 🫠

PR #11209, opened 2023-07-06
docs: update CONTRIBUTING.md to include information about change log

id:


qzhuyan (approved):

paulozulato (approved):
nit: maybe it should mention that these files should be written in English.
Also, we could normalize the enum notation (it's using `{,}` for ce/ee and `(|)` for feat/perf/fix.

paulozulato (approved):

thalesmg (approved):

PR #11211, opened 2023-07-06
consistently return 404 if resource not found

sstrigler:
Fixes [EMQX-9963](https://emqx.atlassian.net/browse/EMQX-9963)

## Summary
copilot:summary

## PR Checklist

- [-] ~If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up~


apps/emqx_gateway/src/emqx_gateway_api_clients.erl
```diff
@@ -234,8 +238,17 @@ subscriptions(delete, #{
     }
 }) ->
     with_gateway(Name0, fun(GwName, _) ->
-        _ = emqx_gateway_http:client_unsubscribe(GwName, ClientId, Topic),
-        {204}
+        case lookup_topic(GwName, ClientId, Topic) of
+            {ok, _} ->
+                case emqx_gateway_http:client_unsubscribe(GwName, ClientId, Topic) of
+                    {error, not_found} ->
+                        return_http_error(404, "Client not found");
```
keynslug, line 245: Just a thought: maybe simplify this to `Subscription not found`, without trying to find out if it's client or topic that's not found?

apps/emqx_gateway/src/emqx_gateway_api_clients.erl
```diff
@@ -218,8 +222,8 @@ subscriptions(post, #{
                 of
                     {error, not_found} ->
                         return_http_error(404, "client process not found");
-                    {error, Reason} ->
-                        return_http_error(400, Reason);
+                    {error, {bad_rpc, _} = Reason} ->
```
thalesmg, line 225: Q: should we keep a catch-all `{error, Reason}` clause?

apps/emqx_gateway/src/emqx_gateway_api_clients.erl
```diff
@@ -234,8 +238,17 @@ subscriptions(delete, #{
...
-        {204}
+        case lookup_topic(GwName, ClientId, Topic) of
+            {ok, _} ->
+                case emqx_gateway_http:client_unsubscribe(GwName, ClientId, Topic) of
+                    {error, not_found} ->
+                        return_http_error(404, "Client not found");
+                    ok ->
+                        {204}
+                end;
+            {error, not_found} ->
```
thalesmg, line 249: From the clauses, seems like `lookup_topic` might return other errors.  Maybe we need to add `{error, Reason}` here as well?

sstrigler: The other ones I don’t want to handle, it’s a `500` situation then.

apps/emqx_gateway/src/emqx_gateway_api_clients.erl
sstrigler: Wasn’t sure about this since it’s implementation dependent and there’s no strict contract. Some might return `{error, not_found}` if there’s no such subscription, others might not. Haven’t read through all the code tbh

apps/emqx_gateway/src/emqx_gateway_api_clients.erl
sstrigler: Tbh, I’d rather not handle a `500` situation at all. It’s kind of a contradiction. And also I think any `error` except `not_found` is unexpected.

sstrigler: Turns out ... it's needed :D

id:
@thalesmg @keynslug if you are happy with the PR, please go ahead and merge it.

PR #11223, opened 2023-07-07
fix: influxdb float serialization error

JimMoen:
Fixes <issue-or-jira-number>

### <samp>🤖 Generated by Copilot at 4a7ba95</samp>

Fix a bug in `emqx_bridge_influxdb_connector` that caused serialization errors when writing integers as floats to the InfluxDB bridge. Add a changelog file `fix-11223.en.md` to document the bug fix.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_influxdb/src/emqx_bridge_influxdb_connector.erl
+++ b/apps/emqx_bridge_influxdb/src/emqx_bridge_influxdb_connector.erl
@@ -638,8 +638,10 @@ value_type([UInt, <<"u">>]) when
     is_integer(UInt)
 ->
     {uint, UInt};
-value_type([Float]) when is_float(Float) ->
-    Float;
+%% write `1`, `1.0`, `-1.0` all as float
+%% see also: https://docs.influxdata.com/influxdb/v2.7/reference/syntax/line-protocol/#float
+value_type([Number]) when is_number(Number) ->
+    Number;
 value_type([<<"t">>]) ->
     't';
 value_type([<<"T">>]) ->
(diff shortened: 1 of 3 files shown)


apps/emqx_bridge_influxdb/src/emqx_bridge_influxdb_connector.erl
```diff
@@ -638,8 +638,10 @@ value_type([UInt, <<"u">>]) when
     is_integer(UInt)
 ->
     {uint, UInt};
-value_type([Float]) when is_float(Float) ->
-    Float;
+%% write `1`, `1.0`, `-1.0` all as float
+%% see also: https://docs.influxdata.com/influxdb/v2.7/reference/syntax/line-protocol/#float
+value_type([Number]) when is_number(Number) ->
+    Number;
```
savonarola, line 644: shouldn't we convert to float then?

JimMoen: In fact, if you write an integer without a decimal point as a field value, influxdb will think it is a float value, unless `i` or `u` is added after it when serializing the line protocol to let influxdb understand that this is an int or uint value.
We should not change its original literal value while writing.
Of course, the current indirect method is to multiply the value that needs to be written for the float type by 1.0 in the rule engine.

See this [line-protocol-float-example](https://docs.influxdata.com/influxdb/v2.7/reference/syntax/line-protocol/#float-field-value-examples)

HJianBo: I see.  SO, according to the InfluxDB protocol, the *Integer* literal value will actually be treated as a `Float` value.

HJianBo (approved):
LGTM

savonarola (approved):

PR #11236, opened 2023-07-10
perf(http): improve the speed of clients querying in default parameters

HJianBo:
In https://github.com//emqx/emqx/pull/9749 we fixed the problem of incorrect total calculation. In this PR we added an option to improve the query speed when the query condition is empty

Fixes https://emqx.atlassian.net/browse/EMQX-10178

Notes: Just a refactoring, the lines should be covered in the old tests

### <samp>🤖 Generated by Copilot at 5206e5f</samp>

Improved the performance and flexibility of clients query API by adding an option to use fast counting and custom query options. Changed `emqx_mgmt_api_clients`, `emqx_mgmt_api`, and related modules to support the new option.

## PR Checklist

- ~Added tests for the changes~

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api.erl
@@ -68,7 +68,7 @@
     %% multiple schemas of data, i.e: Built-in Authorization
     %%
     %% Default: false
-    fast_total_counting := boolean()
+    fast_total_counting => boolean()
 }.
 
 -type query_return() :: #{meta := map(), data := [term()]}.
--- a/apps/emqx_management/src/emqx_mgmt_api.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api.erl
@@ -29,7 +29,9 @@
 %% first_next query APIs
 -export([
     node_query/6,
+    node_query/7,
     cluster_query/5,
+    cluster_query/6,
     b2i/1
 ]).
 
@@ -56,6 +58,18 @@
 -type format_result_fun() ::
     fun((node(), term()) -> term())
     | fun((term()) -> term()).
+
+-type query_options() :: #{
+    %% Whether to use `ets:info/2` to get the total number of rows when the query conditions are
+    %% empty. It can significantly improves the speed of the query when the table stored large
+    %% amounts of data.
+    %%
+    %% However, it may cause the total number of rows to be inaccurate when the table stored in
+    %% multiple schemas of data, i.e: Built-in Authorization
+    %%
+    %% Default: false
+    fast_total_counting := boolean()
+}.
 
 -type query_return() :: #{meta := map(), data := [term()]}.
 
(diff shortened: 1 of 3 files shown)


apps/emqx_management/src/emqx_mgmt_api.erl
```diff
@@ -355,6 +394,8 @@ apply_total_query(QueryState = #{table := Tab}) ->
             Fun(Tab)
     end.
 
+counting_total_fun(_QueryState = #{qs := {[], []}, options := #{fast_total_counting := true}}) ->
+    fun(Tab) -> ets:info(Tab, size) end;
```
keynslug, line 398: Do we really need to introduce another option here? I'm a bit wary of the possibility that it may easily be conflicting with query's matchspec, if accidentally misused.

I'd propose that we allow `match_spec` to be `undefined` instead. Then, when we know that some query is equivalent to full table scan, we may add another clause, for example in `emqx_mgmt_api_clients`.
```
-spec qs2ms(list()) -> ets:match_spec() | undefined.
qs2ms([]) ->
    undefined;
qs2ms(Qs) ->
    {MtchHead, Conds} = qs2ms(Qs, 2, {#{}, []}),
    [{{'$1', MtchHead, '_'}, Conds, ['$_']}].
```

And then handle it in `emqx_mgmt_api`, i.e.:
```
            case maps:get(continuation, QueryState, undefined) of
                undefined when Ms == undefined ->
                    ets:select(Tab, {'_', [], ['$_']}, Limit);
                undefined ->
                    ets:select(Tab, Ms, Limit);
```

We would still need to go over the code and mark the full table scans with `match_spec => undefined` but at least this couldn't be inconsistent. What do you think?

HJianBo: Thanks for your review.  I got your concern but the implementation may introduce another inconsistency possibility🤔  : **"The caller must specify to return the entire record structure when constructing the MatchSpec"**. i.e:
```
qs2ms([]) ->
    undefined;
qs2ms(Qs) ->
     %% Only returns the first element for the record
     %% BUT it difference with '$_'  of the `Ms == undefined` case clause
      [{{'$1', MtchHead, '_'}, Conds, ['$1']}]
```
But my concern may be a bit redundant, due to all callers now returning the entire record for now🥲 .

keynslug: You're right, indeed it isn't suitable for such scenarios. As I see it, we'd have a few choices then:
* bring that same option (which introduces same inconsistency, though we could hide behind some DSL to make it a bit harder to misuse),
* disallow that (which weakens flexibility),
* perform syntactical analysis of the matchspec (which sounds way too complex).

But anyway, that was just a suggestion, it's fine to go either way.

HJianBo: Understood. Let's keep the current implementation for now. It is simple enough. and if we encounter any new issues in the future, we can fix them on a case-by-case.😉

lafirest (approved):

PR #11258, opened 2023-07-12
refactor workflows

id:
Fixes [EMQX-10334](https://emqx.atlassian.net/browse/EMQX-10334) among other things.

- Now `pull_request` and `push` events trigger only single respective "entrypoint" workflow which gives us much better control over what we want to trigger, when and how.
- PR Entrypoint first runs 'Sanity checks' workflow before scheduling anything else. 'Sanity checks' include most, if not all, fast checks that were spread over other workflows, including pre-commit hooks. This helps to reduce runner usage.
- New feature: 'Sanity checks' includes workflow linter. One can also run sanity checks manually before push via `scripts/pr-sanity-checks.sh`.
- Single place to process tag name and decide what artifacts to produce.
- Significantly reduce amount of identical artifacts, e.g. building the same docker image/package in different workflows. Compiling emqx for tests and building docker images are triggered simultaneously after passing sanity checks, and then trigger corresponding test runs.
- Reduce amount of places where we configure emqx-builder version, otp version, etc.
- Now all runs for a pull request are consolidated on one page (as you can see here). All artifacts are also available on one page.
- Moved windows package build from slim_packages to daily scheduled build.
- Moved test docker image build from slim_packages to its separate workflow since usually it completes faster. This allows to unblock dependent workflows faster.
- Minor improvements in several workflows.
- `docker-latest-*` tag is now obsolete, docker image is now tagged `latest` automatically if the tag which triggered the build is the latest tag in commit history, and strictly corresponds to pattern `^refs/tags/(v|e)[5-9]+\.[0-9]+\.[0-9]+$` or `^refs/tags/(v|e)[5-9]+\.[0-9]+\.[0-9]+\.[0-9]$`.
- Use self-hosted runners as much as possible.
- Introduce [shelltest](https://github.com/simonmichael/shelltestrunner)

Reference runs on my fork:
- push to master: https://github.com/id/emqx/actions/runs/5784965544
- push to release-51: https://github.com/id/emqx/actions/runs/5784978485
- push to ci/test: https://github.com/id/emqx/actions/runs/5784993123
- `git tag v5.0.99; git tag e5.0.99; git push origin --atomic v5.0.99 e5.0.99`
	- https://github.com/id/emqx/actions/runs/5785807703
	- https://github.com/id/emqx/actions/runs/5785807699


scripts/parse-git-ref.sh
```diff
@@ -0,0 +1,47 @@
...
+    LATEST=false
+elif [[ $1 =~ ^refs/tags/.+ ]]; then
+    echo "Unrecognized tag: $1"
+    exit 1
+elif [[ $1 =~ ^refs/heads/master$ ]]; then
+    PROFILE=emqx
+    EDITION=Opensource
+    RELEASE=false
+    LATEST=false
+elif [[ $1 =~ ^refs/heads/release-[5-9]+$ ]]; then
```
thalesmg, line 35:
```suggestion
elif [[ $1 =~ ^refs/heads/release-[5-9][0-9]+$ ]]; then
```

Otherwise, it might not match `release-51`.  Actually, it seems to be matching it by luck?

Tried a quick test:

```sh
$ [[ release-51 =~ ^release-[5-9]+ ]] && echo matches
matches
$ [[ release-51 =~ "^release-[5-9]+" ]] && echo matches
# no output
```

.github/workflows/spellcheck.yaml
```diff
@@ -0,0 +1,29 @@
+name: Build slim packages
```
thalesmg, line 1:
```suggestion
name: Spellcheck
```

.github/workflows/run_helm_tests.yaml
```diff
@@ -0,0 +1,151 @@
+name: Helm tests
+
+concurrency:
+  group: helm-${{ github.event_name }}-${{ github.ref }}
```
thalesmg, line 4: Q: should we use `github.ref` or `github.sha` as the last segment?  I saw another workflow that used `sha`.  :thinking: 

Perhaps using `ref` is preferred, since it'll abort ongoing runs for the same branch.

.github/workflows/static_checks.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - uses: actions/cache@v3
+        with:
+          path: "emqx_dialyzer_${{ matrix.otp }}_plt"
+          key: rebar3-dialyzer-plt-${{ matrix.profile }}-${{ matrix.otp }}-${{ hashFiles('rebar.*', 'apps/*/rebar.*', 'lib-ee/*/rebar.*') }}
+          restore-keys: |
+            rebar3-dialyzer-plt-${{ matrix.profile }}-${{ matrix.otp }}-
+      - name: run static checks
+        env:
+          PROFILE: ${{ matrix.profile }}
+        run: make static_checks
```
thalesmg, line 49: I noticed that `scripts/check-i18n-style.escript` apparently never exits with an error exit code, so it seems like that check is being ignored in practice.  :thinking: 

(not introduced by this PR)

.github/actions/pr-sanity-checks/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Run shellcheck
+      shell: bash
+      run: |
+        DEBIAN_FRONTEND=noninteractive apt-get update -qy && apt-get install -qy shellcheck
+        ./scripts/shellcheck.sh
+      # https://github.com/rhysd/actionlint/blob/main/docs/usage.md#use-actionlint-on-github-actions
+    - name: Check workflow files
+      shell: bash
+      run: |
+        bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
```
zmstone, line 27: maybe use the permalink: https://github.com/rhysd/actionlint/blob/6f31c56453b9c256bfb4940a3d1cf8eb596d26dd/scripts/download-actionlint.bash

zmstone: also pin the binary version:
* download the bash script
* execute `bash download-actionlink.sh 1.6.9`

zmstone (commented):
Maybe a separate task: we'll need to support the 4th number in the release version.
e.g. `v5.1.1.1` in case `v5.1.1` is broken.

Also, we'll need to support the possibility to rebuild an old tag (e.g. by adding the 4th version number).

id:
> Maybe a separate task: we'll need to support the 4th number in the release version. e.g. `v5.1.1.1` in case `v5.1.1` is broken.
> 
> Also, we'll need to support the possibility to rebuild an old tag (e.g. by adding the 4th version number).

yes, I'd rather do this in a separate task

zmstone:
> > Maybe a separate task: we'll need to support the 4th number in the release version. e.g. `v5.1.1.1` in case `v5.1.1` is broken.
> > Also, we'll need to support the possibility to rebuild an old tag (e.g. by adding the 4th version number).
> 
> yes, I'd rather do this in a separate task

Maybe we can make the 'LATEST=true' right from the beginning?
Ideally it should check if the reference (tag) being created is indeed the latest.

id:
@zmstone something like this?

scripts/parse-git-ref.sh
```diff
@@ -0,0 +1,72 @@
...
+    LATEST=false
+elif [[ $1 =~ ^refs/tags/.+ ]]; then
+    echo "Unrecognized tag: $1"
+    exit 1
+elif [[ $1 =~ ^refs/heads/master$ ]]; then
+    PROFILE=emqx
+    EDITION=Opensource
+    RELEASE=false
+    LATEST=false
+elif [[ $1 =~ ^refs/heads/release-[5-9][0-9]+$ ]]; then
```
thalesmg, line 55: I think that what triggered the regexp correct behavior was the usage of double quotes around the expression

thalesmg (approved):
:hot_pepper:

scripts/parse-git-ref.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# $1 is fully qualified git ref name, e.g. refs/tags/v5.1.0 or refs/heads/master
+
+is_latest() {
+    ref_name=$(basename "$1")
+    latest_ref_name=$(git describe --tags "$(git rev-list --tags --max-count=1)")
```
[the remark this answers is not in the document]
id: true, I'll think about a way to solve this

.github/workflows/_pr_entrypoint.yaml
```diff
@@ -0,0 +1,242 @@
...
+        run: |
+          DEBIAN_FRONTEND=noninteractive apt-get update -qy && apt-get install -qy shelltestrunner
+          scripts/shelltest/run_tests.sh
+      - name: Check workflow files
+        env:
+          ACTIONLINT_VSN: 1.6.25
+        run: |
+          wget https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VSN}/actionlint_${ACTIONLINT_VSN}_linux_amd64.tar.gz
+          tar zxf actionlint_${ACTIONLINT_VSN}_linux_amd64.tar.gz actionlint
+          # TODO: enable shellcheck when all the current issues are fixed
```
thalesmg, line 62: Q: we already have shellcheck a couple steps above.  Is this something else?

thalesmg (approved):

.github/workflows/_pr_entrypoint.yaml
id: this one would run shellcheck on the content of `run` steps in workflows.

zmstone (approved):
Great work!

PR #11260, opened 2023-07-12
fix(rule_maps): avoid losing data when using `emqx_rule_maps:nested_put`

thalesmg:
# targeting `release-51`

Fixes https://emqx.atlassian.net/browse/EMQX-10541

The cause was that a binary JSON payload was losing its keys when being handled by rule engine, leading to a message like `#{measurement => <<"temp">>, <<"value">> => undefined, ...}`

### <samp>🤖 Generated by Copilot at bffef38</samp>

This pull request enhances the test utilities and functionality for the HTTP and IoTDB bridges. It improves the `emqx_bridge_testlib` module to allow custom SQL statements for rules and actions. It adds a new test case to `emqx_bridge_iotdb_impl_SUITE` to verify the device ID extraction from the rule engine message payload. It also modifies the `emqx_rule_maps` module and its test suite to support updating binary-encoded JSON objects in nested maps.

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_maps.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_maps.erl
@@ -129,15 +129,6 @@ general_find({index, _}, List, _OrgData, Handler) when not is_list(List) ->
 
 do_put({key, Key}, Val, Map, _OrgData) when is_map(Map) ->
     maps:put(Key, Val, Map);
-do_put({key, Key}, Val, Data, _OrgData) when is_binary(Data) ->
-    case emqx_utils_json:safe_decode(Data, [return_maps]) of
-        {ok, Map = #{}} ->
-            %% Avoid losing other keys when the data is an encoded map...
-            Map#{Key => Val};
-        _ ->
-            %% Fallback to the general case otherwise.
-            #{Key => Val}
-    end;
 do_put({key, Key}, Val, Data, _OrgData) when not is_map(Data) ->
     #{Key => Val};
 do_put({index, {const, Index}}, Val, List, _OrgData) ->
--- a/apps/emqx_rule_engine/src/emqx_rule_maps.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_maps.erl
@@ -129,6 +129,15 @@ general_find({index, _}, List, _OrgData, Handler) when not is_list(List) ->
 
 do_put({key, Key}, Val, Map, _OrgData) when is_map(Map) ->
     maps:put(Key, Val, Map);
+do_put({key, Key}, Val, Data, _OrgData) when is_binary(Data) ->
+    case emqx_utils_json:safe_decode(Data, [return_maps]) of
+        {ok, Map = #{}} ->
+            %% Avoid losing other keys when the data is an encoded map...
+            Map#{Key => Val};
+        _ ->
+            %% Fallback to the general case otherwise.
+            #{Key => Val}
+    end;
 do_put({key, Key}, Val, Data, _OrgData) when not is_map(Data) ->
     #{Key => Val};
 do_put({index, {const, Index}}, Val, List, _OrgData) ->
--- a/apps/emqx_rule_engine/test/emqx_rule_engine_SUITE.erl
+++ b/apps/emqx_rule_engine/test/emqx_rule_engine_SUITE.erl
@@ -21,6 +21,7 @@
 
 -include_lib("eunit/include/eunit.hrl").
 -include_lib("common_test/include/ct.hrl").
+-include_lib("snabbkaffe/include/snabbkaffe.hrl").
 
 -include_lib("emqx/include/emqx.hrl").
 
@@ -582,6 +583,122 @@ t_ensure_action_removed(_) ->
 %%------------------------------------------------------------------------------
 %% Test cases for rule runtime
 %%------------------------------------------------------------------------------
+
+t_json_payload_decoding(_Config) ->
+    {ok, C} = emqtt:start_link(),
+    on_exit(fun() -> emqtt:stop(C) end),
+    {ok, _} = emqtt:connect(C),
+
+    Cases =
+        [
+            #{
+                select_fields =>
+                    <<"payload.measurement, payload.data_type, payload.value, payload.device_id">>,
+                payload => emqx_utils_json:encode(#{
+                    measurement => <<"temp">>,
+                    data_type => <<"FLOAT">>,
+                    value => <<"32.12">>,
+                    device_id => <<"devid">>
+                }),
+                expected => #{
+                    payload => #{
+                        <<"measurement">> => <<"temp">>,
+                        <<"data_type">> => <<"FLOAT">>,
+                        <<"value">> => <<"32.12">>,
+                        <<"device_id">> => <<"devid">>
+                    }
+                }
+            },
+            %% "last write wins" examples
+            #{
+                select_fields => <<"payload as p, payload.f as p.answer">>,
+                payload => emqx_utils_json:encode(#{f => 42, keep => <<"that?">>}),
+                expected => #{
+                    <<"p">> => #{
+                        <<"answer">> => 42
+                    }
+                }
+            },
+            #{
+                select_fields => <<"payload as p, payload.f as p.jsonlike.f">>,
+                payload => emqx_utils_json:encode(#{
+                    jsonlike => emqx_utils_json:encode(#{a => 0}),
+                    f => <<"huh">>
+                }),
+                %% behavior from 4.4: jsonlike gets wiped without preserving old "keys"
+                %% here we overwrite it since we don't explicitly decode it
+                expected => #{
+                    <<"p">> => #{
+                        <<"jsonlike">> => #{<<"f">> => <<"huh">>}
+                    }
+                }
+            },
+            #{
+                select_fields =>
+                    <<"payload as p, 42 as p, payload.measurement as p.measurement, 51 as p">>,
+                payload => emqx_utils_json:encode(#{
+                    measurement => <<"temp">>,
+                    data_type => <<"FLOAT">>,
+                    value => <<"32.12">>,
+                    device_id => <<"devid">>
+                }),
+                expected => #{
+                    <<"p">> => 51
+                }
+            },
+            %% if selected field is already structured, new values are inserted into it
+            #{
+                select_fields =>
+                    <<"json_decode(payload) as p, payload.a as p.z">>,
+                payload => emqx_utils_json:encode(#{
+                    a => 1,
+                    b => <<"2">>
+                }),
+                expected => #{
+                    <<"p">> => #{
+                        <<"a">> => 1,
+                        <<"b">> => <<"2">>,
+                        <<"z">> => 1
+                    }
+                }
+            }
+        ],
+    ActionFn = <<(atom_to_binary(?MODULE))/binary, ":action_response">>,
+    Topic = <<"some/topic">>,
+
+    ok = snabbkaffe:start_trace(),
+    on_exit(fun() -> snabbkaffe:stop() end),
+    on_exit(fun() -> delete_rule(?TMP_RULEID) end),
+    lists:foreach(
+        fun(#{select_fields := Fs, payload := P, expected := E} = Case) ->
+            ct:pal("testing case ~p", [Case]),
+            SQL = <<"select ", Fs/binary, " from \"", Topic/binary, "\"">>,
+            delete_rule(?TMP_RULEID),
+            {ok, _Rule} = emqx_rule_engine:create_rule(
+                #{
+                    sql => SQL,
+                    id => ?TMP_RULEID,
+                    actions => [#{function => ActionFn}]
+                }
+            ),
+            {_, {ok, Event}} =
+                ?wait_async_action(
+                    emqtt:publish(C, Topic, P, 0),
+                    #{?snk_kind := action_response},
+                    5_000
+                ),
+            ?assertMatch(
+                #{selected := E},
+                Event,
+                #{payload => P, fields => Fs, expected => E}
+            ),
+            ok
+        end,
+        Cases
+    ),
+    snabbkaffe:stop(),
+
+    ok.
 
 t_events(_Config) ->
     {ok, Client} = emqtt:start_link(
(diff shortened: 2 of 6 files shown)


apps/emqx_rule_engine/src/emqx_rule_maps.erl
```diff
@@ -129,6 +129,15 @@ general_find({index, _}, List, _OrgData, Handler) when not is_list(List) ->
 
 do_put({key, Key}, Val, Map, _OrgData) when is_map(Map) ->
     maps:put(Key, Val, Map);
+do_put({key, Key}, Val, Data, _OrgData) when is_binary(Data) ->
+    case emqx_utils_json:safe_decode(Data, [return_maps]) of
```
keynslug, line 133: Just a thought. There seems to be more and more places where we pay for such flexibility with a few wasted CPU cycles here (since this clause is likely to be hit). Given this is part of a hot path, it might be costly. I'd honestly try to transfer this responsibility out of this module to those modules that know for sure they're dealing with JSON (as a separate effort obviously).

keynslug, line 133: Q: Could there be an inadvertent violation of user's intent? I.e. user wants to have a field `"f":"{}"` but instead will get `"f":{}`, without an ability to enforce the former behavior.

thalesmg: Indeed, perhaps we could attempt the decoding once when the payload enters the rule engine to avoid this nested decoding, which can lead to the issue you commented below of inadvertently decoding what should be a string.

keynslug: So I couldn't resist doing some experiments on top of PR's current head (bffef386c18f7697055d15ccf8d2146f5cb64ff9). This patch doesn't seem to break few critical tests (`emqx_bridge_iotdb_impl_SUITE` and `emqx_rule_engine_SUITE` suites), just to be clear I'm not confident it won't break anything else.

```
diff --git a/apps/emqx_rule_engine/src/emqx_rule_maps.erl b/apps/emqx_rule_engine/src/emqx_rule_maps.erl
index b4dd8fc82..baf83ff6f 100644
--- a/apps/emqx_rule_engine/src/emqx_rule_maps.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_maps.erl
@@ -129,15 +129,6 @@ general_find({index, _}, List, _OrgData, Handler) when not is_list(List) ->
 
 do_put({key, Key}, Val, Map, _OrgData) when is_map(Map) ->
     maps:put(Key, Val, Map);
-do_put({key, Key}, Val, Data, _OrgData) when is_binary(Data) ->
-    case emqx_utils_json:safe_decode(Data, [return_maps]) of
-        {ok, Map = #{}} ->
-            %% Avoid losing other keys when the data is an encoded map...
-            Map#{Key => Val};
-        _ ->
-            %% Fallback to the general case otherwise.
-            #{Key => Val}
-    end;
 do_put({key, Key}, Val, Data, _OrgData) when not is_map(Data) ->
     #{Key => Val};
 do_put({index, {const, Index}}, Val, List, _OrgData) ->
diff --git a/apps/emqx_rule_engine/src/emqx_rule_runtime.erl b/apps/emqx_rule_engine/src/emqx_rule_runtime.erl
index f83aa2920..de1e92a3f 100644
--- a/apps/emqx_rule_engine/src/emqx_rule_runtime.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_runtime.erl
@@ -23,7 +23,6 @@
 -export([
     apply_rule/3,
     apply_rules/3,
-    clear_rule_payload/0,
     inc_action_metrics/2
 ]).
 
@@ -196,18 +195,18 @@ select_and_transform([], _Columns, Action) ->
 select_and_transform(['*' | More], Columns, Action) ->
     select_and_transform(More, Columns, maps:merge(Action, Columns));
 select_and_transform([{as, Field, Alias} | More], Columns, Action) ->
-    Val = eval(Field, Columns),
+    Val = eval(Field, [Action, Columns]),
     select_and_transform(
         More,
-        nested_put(Alias, Val, Columns),
+        Columns,
         nested_put(Alias, Val, Action)
     );
 select_and_transform([Field | More], Columns, Action) ->
-    Val = eval(Field, Columns),
+    Val = eval(Field, [Action, Columns]),
     Key = alias(Field, Columns),
     select_and_transform(
         More,
-        nested_put(Key, Val, Columns),
+        Columns,
         nested_put(Key, Val, Action)
     ).
 
@@ -217,25 +216,25 @@ select_and_collect(Fields, Columns) ->
     select_and_collect(Fields, Columns, {#{}, {'item', []}}).
 
 select_and_collect([{as, Field, {_, A} = Alias}], Columns, {Action, _}) ->
-    Val = eval(Field, Columns),
+    Val = eval(Field, [Action, Columns]),
     {nested_put(Alias, Val, Action), {A, ensure_list(Val)}};
 select_and_collect([{as, Field, Alias} | More], Columns, {Action, LastKV}) ->
-    Val = eval(Field, Columns),
+    Val = eval(Field, [Action, Columns]),
     select_and_collect(
         More,
         nested_put(Alias, Val, Columns),
         {nested_put(Alias, Val, Action), LastKV}
     );
 select_and_collect([Field], Columns, {Action, _}) ->
-    Val = eval(Field, Columns),
+    Val = eval(Field, [Action, Columns]),
     Key = alias(Field, Columns),
     {nested_put(Key, Val, Action), {'item', ensure_list(Val)}};
 select_and_collect([Field | More], Columns, {Action, LastKV}) ->
-    Val = eval(Field, Columns),
+    Val = eval(Field, [Action, Columns]),
     Key = alias(Field, Columns),
     select_and_collect(
         More,
-        nested_put(Key, Val, Columns),
+        Columns,
         {nested_put(Key, Val, Action), LastKV}
     ).
 
@@ -368,6 +367,16 @@ do_handle_action(RuleId, #{mod := Mod, func := Func, args := Args}, Selected, En
     inc_action_metrics(RuleId, Result),
     Result.
 
+eval({Op, _} = Exp, Context) when is_list(Context) andalso (Op == path orelse Op == var) ->
+    case Context of
+        [Columns] ->
+            eval(Exp, Columns);
+        [Columns | Rest] ->
+            case eval(Exp, Columns) of
+                undefined -> eval(Exp, Rest);
+                Val -> Val
+            end
+    end;
 eval({path, [{key, <<"payload">>} | Path]}, #{payload := Payload}) ->
     nested_get({path, Path}, may_decode_payload(Payload));
 eval({path, [{key, <<"payload">>} | Path]}, #{<<"payload">> := Payload}) ->
```

^ The idea here is to avoid rewriting select clause sources (`Columns`), opting for a bit more complex eval context (i.e. try to find value in `Action` accumulator first, then try in `Columns`) which seems to be consistent with rule_engine's evaluation model.

Yet I think the proper way to approach this is to expose some type annotations on the SQL parser level. Basically, if the parser knows that `payload.something` is accessed anywhere, it should be able to tell that _this SQL expects `payload` to be structured data_. With this information, runtime could attempt JSON decoding of `payload` before even starting evaluation (which would also render `safe_decode_and_cache/1` hack obsolete). Admittedly, it sounds way too involved for a bug fix. 🫠

Would like to hear what do you think about all of this. 🙈

thalesmg: Awesome, thanks for the patch!  :beers: 

I'm not deeply familiar with the rule engine internals, but it does sound better to avoid rewriting the parsed select clauses while building the result context, indeed!
~~It's more elegant, but I wonder if removing the new `do_put` clause won't make us decode the payload multiple times, then, since it won't be substituted by its decoded form.  :thinking:~~  Actually, the `safe_decode_and_cache` hack should prevent that.  :crossed_fingers: 

And indeed having extra info from the parser and using that to optimize evaluation sounds like a much smarter idea.  :bulb: 

I'll apply your patch and see if CI agrees with it.  Since this part seems to have some holes in coverage, I just hope it doesn't fall into one of those.  :see_no_evil: 🫠

thalesmg: It seems to work indeed, except for the test cases I've added for the new `do_put` clauses.  Now we "lose" data again if `nested_put` is called on a binary.

However, that module's purpose seems to be solely to support rule engine, and, since we've dealt with that case during evaluation, I think we don't need to offer support for ("correctly") inserting data inside a binary json.  Does this make sense?

keynslug: > I think we don't need to offer support for ("correctly") inserting data inside a binary json. Does this make sense?

Yes, it does make sense, at least in the context of the bug.

There are, however, a couple of edge cases that come to mind (which, honestly, should have came earlier 😅). For example, what should be the result of evaluation of this select (imagine `payload` is `{"f":42,"keep":"that?"}`):
```
SELECT payload as p, payload.f as p.answer FROM ...
```

It could be either `#{p => #{answer => 42}}` or `#{p => #{answer => 42, f => 42, keep => "that?"}}`. I can only speculate, because the evaluation model isn't described anywhere (supposedly), but the second outcome sounds like a more expected outcome to me (since SQL implies that `payload` is JSON).

Yet, this one is harder to answer I think. Imagine `payload` is `{"jsonlike":"{\"a\":0000}", "f":"huh"}`.
```
SELECT payload as p, payload.f as p.jsonlike.f FROM ...
```

Sorry for muddying the waters. 🙃

thalesmg: Hmmm good point.  In those cases, I _would expect_ that we avoid losing data, even if there are strange conversions.

So, it would be safer to keep that clause in `do_put`, looking from this angle.    :thinking:

keynslug: As for the second edge case, I'm not sure if we should. Imagine that `jsonlike` is not in fact JSON in payload, but it may _accidentally_ parse as JSON. User can probably do the following anyway, if he knows for sure.
```
SELECT payload as p, json_decode(payload.jsonlike) as p.jsonlike, payload.f as p.jsonlike.f FROM ...
```

But anyway, this is just a speculation. You're right that we may prefer one interpretation over another, depending on our goals. I'm just favoring _prefer to avoid unexpected outcomes_ goal here.

thalesmg: Well, losing keys in a map might be _unexpect_ from the user's point of view. :joy_cat: 

But indeed it's hard to say what the intentions are in all cases and cover them, specially when there is no precise specification of what should be the behavior.

IMO, I'd rather have things explicitly defined.  That is, even this use case of natively decoding the payload to JSON without calling `json_decode` I would forbid: the user should explicitly decode if they are sure it's a JSON.

Yet, I think we'll have to keep supporting this, as it's one of those "we always supported this, so we'll have to keep supporting" things...  :speak_no_evil:

keynslug (approved):
👍🏼

apps/emqx_rule_engine/test/emqx_rule_engine_SUITE.erl
```diff
@@ -583,6 +584,94 @@ t_ensure_action_removed(_) ->
...
+            #{
+                select_fields => <<"payload as p, payload.f as p.jsonlike.f">>,
+                payload => emqx_utils_json:encode(#{
+                    jsonlike => <<"{\"a\":0000}">>,
+                    f => <<"huh">>
+                }),
+                %% behavior from 4.4: jsonlike gets wiped without preserving old "keys"
+                expected => #{
+                    <<"p">> => #{
+                        <<"jsonlike">> => #{<<"f">> => <<"huh">>},
```
keynslug, line 632: Hm, strange, I thought that bringing that clause back to `emqx_rule_maps` should make this testcase emit `<<"jsonlike">> => #{<<"a">> => 0, <<"f">> => <<"huh">>},`, now I'm wondering why it doesn't. 🤔

Ah, jiffy thinks it's invalid JSON. Maybe use `jsonlike => <<"{\"a\":0}">>` in the payload to better illustrate what's happening here?

apps/emqx_rule_engine/src/emqx_rule_maps.erl
keynslug: > I think we'll have to keep supporting this, as it's one of those "we always supported this, so we'll have to keep supporting" things

One more random thought. I think that both those edge cases never were something that we always supported. Unless you're talking about 4.4, then I think it's fine to break such obscure guarantees in a major release, so that we won't need to support them throughout 5.x series. 🌝

keynslug: Spent some more time to think about what rubs wrong to me about supporting second edge case: it's the implicit decoding of arbitrarily deeply encoded JSON. So, with this SQL:
```
SELECT payload as p, payload.e as p.a.b.c.d FROM ...
```

And this payload:
```
{"e":42, "a":"{\"b\":\"{\\\"c\\\":\\\"{\\\\\\\"z\\\\\\\":{}}\\\"}\"}"}
```

The result becomes `{"p":{"a":{"b":{"c":{"d":42, "z":{}}}}}`, and I personally think it's kinda absurd to guarantee that behavior. 😅 But I'd like to hear what you think about that case.

apps/emqx_rule_engine/test/emqx_rule_engine_SUITE.erl
thalesmg: Hm true, I assumed that 0000 was a valid number in JSON :see_no_evil: :speak_no_evil:

apps/emqx_rule_engine/src/emqx_rule_maps.erl
thalesmg: > Unless you're talking about 4.4, then I think it's fine to break such obscure guarantees in a major release, so that we won't need to support them throughout 5.x series. 

Indeed, I don't know what guarantees we ever offered about those obscure uses, or if they worked like this for happenstance.  :see_no_evil: 
I agree that 5.x should be fine introducing such breaking changes.  On the one hand, it's just that sometime product concerns trump over those things.  On the other hand, I feel like this was never used like this.  Hard to tell the "average usage" by users.  :thinking:

apps/emqx_rule_engine/test/emqx_rule_engine_SUITE.erl
thalesmg: Actually, with that change, it diverges from the old e4.4 behavior: e4.4 drops the `a`, but the new version keeps it.

savonarola (approved):

PR #11269, opened 2023-07-14
fix: hide node.process_limit completely

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-10248
In previous versions, we used node_converter to maintain `process_limit = max_ports * 2`, 
all functionality worked properly.
However, `emqx_config:get_raw/1` and `emqx_config:get/1` would both expose `process_limit`, causing all APIs using them to return this value.
Now using `translation/1` instead of `converter` to completely hide it.

Before(show process_limit):
```
./bin/emqx_ctl conf show node
node {
  cookie = emqxsecretcookie
  data_dir = data
  dist_buffer_size = 8192
  global_gc_interval = 15m
  max_ports = 10000
  name = emqx@127.0.0.1
  role = core
  process_limit = 20000
}
```
After(Hide process_limit):
```
./bin/emqx_ctl conf show node
node {
  cookie = emqxsecretcookie
  data_dir = data
  dist_buffer_size = 8192
  global_gc_interval = 15m
  max_ports = 10000
  name = emqx@127.0.0.1
  role = core
}
```
Hide `node.process_limit` and also ensure `vm_args+P  = node.max_ports * 2` 

### <samp>🤖 Generated by Copilot at 131511b</samp>

Removed deprecated `process_limit` field and added `vm_args` schema for Erlang VM options. Updated tests to reflect the changes in `emqx_conf_schema.erl`.

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_schema.erl
+++ b/apps/emqx_conf/src/emqx_conf_schema.erl
@@ -86,8 +86,7 @@ roots() ->
                 sc(
                     ?R_REF("node"),
                     #{
-                        translate_to => ["emqx"],
-                        converter => fun node_converter/2
+                        translate_to => ["emqx"]
                     }
                 )},
             {"cluster",
@@ -446,9 +445,11 @@ fields("node") ->
             sc(
                 range(1024, 134217727),
                 #{
-                    mapping => "vm_args.+P",
+                    %% deprecated make sure it's disappeared in raw_conf user(HTTP API)
+                    %% but still in vm.args via translation/1
+                    %% ProcessLimit is always equal to MaxPort * 2 when translation/1.
+                    deprecated => true,
                     desc => ?DESC(process_limit),
-                    default => ?DEFAULT_MAX_PORTS * 2,
                     importance => ?IMPORTANCE_HIDDEN,
                     'readOnly' => true
                 }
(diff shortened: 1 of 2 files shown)


apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -1079,8 +1080,15 @@ translation("prometheus") ->
...
         {"vm_memory_collector_metrics", fun tr_vm_memory_collector/1},
         {"vm_msacc_collector_metrics", fun tr_vm_msacc_collector/1}
+    ];
+translation("vm_args") ->
+    [
+        {"+P", fun tr_vm_args_process_limit/1}
     ].
 
+tr_vm_args_process_limit(Conf) ->
+    2 * conf_get("node.max_ports", Conf, ?DEFAULT_MAX_PORTS).
```
thalesmg, line 1090: Maybe this should be _at least_ `2 * max_ports` rather than fixed at this number?

That is, we could specify a greater number of processes than `2 * max_ports` and have that?

zhongwencool: It doesn't seem like there are any scenarios that require such flexible settings.
Even if the user sets it to 2*maxports + 1, it will still become 4*maxports, so I don't think it's necessary to expose this parameter to users. 

Keep simple for user side.
http://erlang.org/documentation/doc-10.0/erts-10.0/doc/html/erl.html#+P

> NOTE: The actual maximum chosen may be much larger than the Number passed. Currently the runtime system often, but not always, chooses a value that is a power of 2. This might, however, be changed in the future. The actual value chosen can be checked by calling [erlang:system_info(process_limit)](http://erlang.org/documentation/doc-10.0/erts-10.0/doc/html/erlang.html#system_info_process_limit).

thalesmg (approved):

PR #11274, opened 2023-07-14
fix: get_config return target node's node conf

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-10552
Return the configuration information of the specified node, whereas previously it only returned the configuration information of the node that the dashboard requested. 
```
curl -X 'GET' \
  'http://10.42.3.4:18083/api/v5/configs?key=node&node=emqx%40127.0.0.2'\
  -H 'accept: text/plain' \
  -H 'Authorization: Basic ZGIyYTdmOWE2ZDRkYWJmMzpSdW05Qjh5RUl6d05QbHI4M05JeUxaOUF2REl3b2s2NDVYTGs4QVlOYTRaRkM='
node {
  cookie = dev-le-511-xxx-yyy-zzz
  data_dir = data
  dist_buffer_size = 8192
  global_gc_interval = 15m
  max_ports = 1048576
  name = emqx@127.0.0.1   <== It should be emqx@127.0.0.2
  role = core
}
```

### <samp>🤖 Generated by Copilot at 0051706</samp>

Refactored and upgraded the management API and its test cases to use the new `emqx_management_proto_v5` module. This module implements the latest RPC protocol for the management API and the broker nodes, improving the compatibility and functionality of the API with the EMQ X Broker features and configurations.

## PR Checklist

- ~Change log has been added to `changes/(ce|ee)/(feat|perf|fix)-<PR-id>.en.md` files~


apps/emqx_management/src/proto/emqx_management_proto_v5.erl
```diff
@@ -0,0 +1,96 @@
...
+    list_listeners/1,
+    subscribe/3,
+    unsubscribe/3,
+    unsubscribe_batch/3,
+
+    call_client/3,
+
+    get_full_config/1,
+    get_full_config_v2/1,
+    get_config_v2/2,
```
keynslug, line 37: Do we really need to introduce another layer of versioning here? Are there any clients of `get_full_config/1` left? If not, we should just drop it, this is a new proto version anyway.

keynslug (commented):
https://github.com/emqx/emqx/actions/runs/5552937066/jobs/10141236047?pr=11274

This also needs to be addressed (actually, every time there's new bpapi proto version).

apps/emqx_management/src/proto/emqx_management_proto_v5.erl
zhongwencool: yes, get_full_config/1 is using here(will deprecated when 5.2.0, but keep in 5.1.x):
https://github.com/zhongwencool/emqx/blob/fix-get-config-from-target-node/apps/emqx_management/src/emqx_mgmt_api_configs.erl#L375-L379
get_full_config_v2/1 is using here:
https://github.com/zhongwencool/emqx/blob/fix-get-config-from-target-node/apps/emqx_management/src/emqx_mgmt_api_configs.erl#L391-L398

keynslug: Ah, I see, thanks for clarifying that. I'd then propose 2 things:
* Avoid making another application's CLI module the target of RPC, I believe it's much better to keep the target in the same module, or at least in the same application, even if it will amount to copypasting some of the code. Middle ground would be moving the target of RPC to the `emqx_conf` module interface and marking it as such, for example with a comment.
* Give those functions a bit more descriptive names instead of `v1` / `v2`, so that it will be easier to understand what's the difference between them.

zhongwencool: done

apps/emqx_management/src/emqx_mgmt_api_configs.erl
```diff
@@ -363,20 +363,21 @@ find_suitable_accept(Headers, Perferences) when is_list(Perferences), length(Per
...
-            Found = lists:filter(fun(Accept) -> lists:member(Accept, Accepts) end, Perferences),
+            Found = lists:filter(fun(Accept) -> lists:member(Accept, Accepts) end, Preferences),
             case Found of
-                [] -> {error, no_suitalbe_accept};
+                [] -> {error, no_suitable_accept};
                 _ -> {ok, lists:nth(1, Found)}
             end
     end.
 
+%% Should deprecated json v1 since 5.2.0
```
HJianBo, line 375:
```suggestion
%% To return a JSON formatted configuration file, which is used to be compatible with the already
%% implemented `GET /configs` in the old versions 5.0 and 5.1.
%%
%% In e5.1.1, we support to return a hocon configuration file by `get_configs_v2/1`. It's more
%% useful for the user to read or reload the configuration file via HTTP API.
%%
%% The `get_configs_v1/1` should be deprecated since 5.2.0.
```

apps/emqx_management/test/emqx_mgmt_api_configs_SUITE.erl
```diff
@@ -283,7 +283,7 @@ t_configs_node({'init', Config}) ->
     ),
     Config;
 t_configs_node({'end', _}) ->
-    meck:unload([emqx, emqx_management_proto_v2]);
+    meck:unload([emqx, emqx_management_proto_v4]);
```
HJianBo, line 286: nit: should we add some test cases to ensure that the node parameter has taken effect?

PR #11215, opened 2023-07-06
feat(session): move most of persistence concerns behind session module

keynslug:
Fixes [EMQX-9593](https://emqx.atlassian.net/browse/EMQX-9593)

### <samp>🤖 Generated by Copilot at 436fb10</samp>

This pull request refactors the session persistence logic across several modules in the `emqx` application. It simplifies the code, removes unused or redundant components, and delegates the persistence and delivery operations to the `emqx_session` module. It also adds a license header to the `emqx_session.hrl` file.

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [ ] Added tests for the changes
- [ ] Changed lines covered in coverage report
- [ ] ~~Change log has been added to `changes/{ce,ee}/(feat|perf|fix)-<PR-id>.en.md` files~~
- [x] For internal contributor: there is a jira ticket to track this change
- [ ] ~~If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up~~
- [x] Schema changes are backward compatible

PR #11293, opened 2023-07-18
chore(ps): phase out existing session persistence mechanism

keynslug:
Part of [EMQX-9593](https://emqx.atlassian.net/browse/EMQX-9593).

Cut the existing session persistence stuff off the core modules.
* It's not really working currently in a cluster setup.
* It's not really working well.
* It would be much easier to integrate `emqx_durable_storage` w/o the need to support existing solution.

Essentially, a continuation of #11215 but without any attempt to preserve existing mechanism.

### <samp>🤖 Generated by Copilot at 3ba6d34</samp>

This pull request refactors the session management logic in EMQX, a MQTT broker, by introducing a new session store API and removing the dependency on the old persistent session store. It simplifies the code and improves the performance of the channel, session, and connection modules. It also updates the version number of the `emqx_eviction_agent` application and disables some tests that rely on the old session store. The affected files are `emqx_eviction_agent_channel.erl`, `emqx_channel.erl`, `emqx_cm.erl`, `emqx_session.erl`, `emqx_eviction_agent.app.src`, `emqx_eviction_agent_channel_SUITE.erl`, `emqx_broker.erl`, and `emqx_persistent_session_SUITE.erl`.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -277,65 +277,24 @@ open_session(true, ClientInfo = #{clientid := ClientId}, ConnInfo) ->
     Self = self(),
     CleanStart = fun(_) ->
         ok = discard_session(ClientId),
-        ok = emqx_persistent_session:discard_if_present(ClientId),
-        Session = create_session(ClientInfo, ConnInfo),
-        Session1 = emqx_persistent_session:persist(ClientInfo, ConnInfo, Session),
-        register_channel(ClientId, Self, ConnInfo),
-        {ok, #{session => Session1, present => false}}
+        ok = emqx_session:destroy(ClientId),
+        create_register_session(ClientInfo, ConnInfo, Self)
     end,
     emqx_cm_locker:trans(ClientId, CleanStart);
 open_session(false, ClientInfo = #{clientid := ClientId}, ConnInfo) ->
     Self = self(),
     ResumeStart = fun(_) ->
-        CreateSess =
-            fun() ->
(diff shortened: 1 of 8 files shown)


apps/emqx/src/emqx_cm.erl
```diff
@@ -277,65 +277,24 @@ open_session(true, ClientInfo = #{clientid := ClientId}, ConnInfo) ->
...
-            fun() ->
-                Session = create_session(ClientInfo, ConnInfo),
-                Session1 = emqx_persistent_session:persist(
-                    ClientInfo, ConnInfo, Session
-                ),
-                register_channel(ClientId, Self, ConnInfo),
-                {ok, #{session => Session1, present => false}}
-            end,
         case takeover_session(ClientId) of
-            {persistent, Session} ->
```
thalesmg, line 300: I wonder if we should keep this clause just to remind ourselves of dealing with it later (basically we could just raise an error like [`unreachable!`](https://doc.rust-lang.org/std/macro.unreachable.html))

thalesmg (approved):

apps/emqx/src/emqx_cm.erl
keynslug: I wondered about that as well, but decided not to, because I have a hunch that this whole function will need to be reworked anyway. Besides, if someone is lucky enough to respond `{persistent, Session}` here, we'll get `case_clause` error anyway, which is in some sense close to `unreachable!`'s meaning.

PR #11294, opened 2023-07-18
Mria fixes

ieQu1:
Fixes EMQX-10556, EMQX-10573

### <samp>🤖 Generated by Copilot at 6b592c8</samp>

This pull request updates the clustering and replication logic of emqx to use mria instead of ekka. It also refactors and fixes some issues in the rebar.config and mix.exs files to improve code quality and dependency management. The files affected are `emqx_mgmt_cli.erl`, `rebar.config.erl`, `apps/emqx/rebar.config`, and `mix.exs`.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_cli.erl
+++ b/apps/emqx_management/src/emqx_mgmt_cli.erl
@@ -110,7 +110,7 @@ broker(_) ->
 %% @doc Cluster with other nodes
 
 cluster(["join", SNode]) ->
-    case ekka:join(ekka_node:parse_name(SNode)) of
+    case mria:join(ekka_node:parse_name(SNode)) of
         ok ->
             emqx_ctl:print("Join the cluster successfully.~n"),
             cluster(["status"]);
@@ -120,7 +120,7 @@ cluster(["join", SNode]) ->
             emqx_ctl:print("Failed to join the cluster: ~0p~n", [Error])
     end;
 cluster(["leave"]) ->
-    case ekka:leave() of
+    case mria:leave() of
         ok ->
             emqx_ctl:print("Leave the cluster successfully.~n"),
             cluster(["status"]);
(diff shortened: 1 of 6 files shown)


mix.exs
```diff
@@ -339,96 +339,125 @@ defmodule EMQXUmbrella.MixProject do
...
+            ]
+
+          _ ->
+            [:emqx_telemetry]
+        end
+
+    system_apps ++
+      for app <- db_apps do
+        {app, :load}
+      end ++
```
thalesmg, line 456:
```suggestion
      Enum.map(db_apps, & {&1, :load}) ++
```

mix.exs
```diff
@@ -339,96 +339,125 @@ defmodule EMQXUmbrella.MixProject do
...
+      [emqx_machine: :permanent] ++
+      for app <- business_apps do
+        {app, :load}
+      end
```
thalesmg, line 460:
```suggestion
      Enum.map(business_apps, & {&1, :load})
```

Although, this seems to change most apps from `:permanent` to `load` only. :thinking: 
Shouldn't those be permanent?

ieQu1: Yes, this is the intention. Init script should not start any apps that depend on mria, because it causes mria to start before ekka initializes the callbacks and before `mgmt_cli` registers the commands. 
The apps will be started by `emqx_machine` (so the sequence is the same as during join/leave)

apps/emqx_management/src/emqx_mgmt_cli.erl
```diff
@@ -925,3 +927,26 @@ with_log(Fun, Msg) ->
...
+cluster_info() ->
+    RunningNodes = safe_call_mria(running_nodes, [], []),
+    StoppedNodes = safe_call_mria(cluster_nodes, [stopped], []),
+    #{
+        running_nodes => RunningNodes,
+        stopped_nodes => StoppedNodes
+    }.
+
+%% CLI starts before mria, so we should handle errors gracefully:
+safe_call_mria(Fun, Args, OnFail) ->
```
lafirest, line 940: I am afraid this function will be abused, Ignoring the exception is well for `get status`, but may be dangerous for other operations

ieQu1: It's just a UI module, it cannot perform any meaningful error handling or recovery anyway. Could you give an example situation where catching errors in the CLI module would be dangerous?

mix.exs
lafirest: Yes, you are right.

lafirest (approved):

thalesmg (approved):

PR #11306, opened 2023-07-19
fix(rule_metrics): notify rule metrics of late replies and expired requests

thalesmg:
# targeting `release-51`

Fixes https://emqx.atlassian.net/browse/EMQX-10600

### <samp>🤖 Generated by Copilot at f3c97d5</samp>

This pull request improves the handling of dropped replies by the HTTP bridge and the resource buffer worker. It adds a `reply_context` map to the `reply_fun` type and the `ReplyTo` tuple to pass information about the reply status. It also adds a new test case for the bridge expiration behavior and refactors the bridge URL configuration.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -824,7 +824,7 @@ reply_dropped(_ReplyTo = {Fn, Args, #{reply_dropped := true}}, Result) when
     is_function(Fn), is_list(Args)
 ->
     %% We want to avoid bumping metrics inside the buffer worker, since it's costly.
-    spawn(fun() -> erlang:apply(Fn, Args ++ [Result]) end),
+    emqx_pool:async_submit(Fn, Args ++ [Result]),
     ok;
 reply_dropped(_ReplyTo, _Result) ->
     ok.
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -366,6 +366,7 @@ resume_from_blocked(Data) ->
                     true -> #{dropped_expired => length(Batch)};
                     false -> #{}
                 end,
+            batch_reply_dropped(Batch, {error, request_expired}),
             NData = aggregate_counters(Data, Counters),
             ?tp(buffer_worker_retry_expired, #{expired => Batch}),
             resume_from_blocked(NData);
@@ -378,6 +379,7 @@ resume_from_blocked(Data) ->
         {batch, Ref, NotExpired, Expired} ->
             NumExpired = length(Expired),
             ok = update_inflight_item(InflightTID, Ref, NotExpired, NumExpired),
+            batch_reply_dropped(Expired, {error, request_expired}),
             NData = aggregate_counters(Data, #{dropped_expired => NumExpired}),
             ?tp(buffer_worker_retry_expired, #{expired => Expired}),
             %% We retry msgs in inflight window sync, as if we send them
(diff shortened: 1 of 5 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -809,6 +816,28 @@ reply_caller_defer_metrics(Id, ?REPLY(ReplyTo, HasBeenSent, Result), QueryOpts)
...
 
+%% This is basically used only by rule actions.  To avoid rule action metrics from
+%% becoming inconsistent when we drop messages, we need a way to signal rule engine that
+%% this action has reached a conclusion.
+-spec reply_dropped(reply_fun(), {error, late_reply | request_expired}) -> ok.
+reply_dropped(_ReplyTo = {Fn, Args, #{reply_dropped := true}}, Result) when
+    is_function(Fn), is_list(Args)
+->
+    %% We want to avoid bumping metrics inside the buffer worker, since it's costly.
+    spawn(fun() -> erlang:apply(Fn, Args ++ [Result]) end),
```
keynslug, line 827: TBH, this also sounds costly: it's per each dropped message, and their rate tend to rise when resources are already exhausted (i.e. CPU). But I do not have any good suggestions, besides perhaps using `emqx_pool:async_sumbit/2`.

keynslug (approved):
👍🏼  Good find, as always.

Still, can't help but wonder if we really need to spend that much effort on ensuring that action metrics are identical to the respective bridge metrics. Action metrics become somewhat pointless in the end.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
thalesmg: Indeed, but I wanted to avoid stressing these workers at least... :sweat_smile: 

Great suggestion, didn't know/remember this function.  :bulb: 

Maybe we could do something similar to the metrics aggregation inside them, and maybe send a message to a specific metrics worker that then evaluates it like `emqx_pool`, or just use `emqx_pool` directly, if it is not already too burdened.  For now, I'll replace this with async_submit, we could later improve this further I guess.

thalesmg:
> 👍🏼 Good find, as always.
> 
> Still, can't help but wonder if we really need to spend that much effort on ensuring that action metrics are identical to the respective bridge metrics. Action metrics become somewhat pointless in the end.

Indeed, but metrics are a recurring source of tickets from QA...  I guess users could get confused as well.  :sweat:

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -809,6 +816,28 @@ reply_caller_defer_metrics(Id, ?REPLY(ReplyTo, HasBeenSent, Result), QueryOpts)
...
 
+%% This is basically used only by rule actions.  To avoid rule action metrics from
+%% becoming inconsistent when we drop messages, we need a way to signal rule engine that
+%% this action has reached a conclusion.
+-spec reply_dropped(reply_fun(), {error, late_reply | request_expired}) -> ok.
+reply_dropped(_ReplyTo = {Fn, Args, #{reply_dropped := true}}, Result) when
+    is_function(Fn), is_list(Args)
+->
+    %% We want to avoid bumping metrics inside the buffer worker, since it's costly.
+    emqx_pool:async_submit({Fn, Args}),
```
keynslug, line 827: Nittiest of nits.
```suggestion
    emqx_pool:async_submit(Fn, Args),
```

keynslug (dismissed):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
thalesmg: :heavy_check_mark:

paulozulato (approved):

PR #11282, opened 2023-07-17
feat(index): add topic index facility 

JimMoen:
Fixes [EMQX-10006](https://emqx.atlassian.net/browse/EMQX-10006)

### <samp>🤖 Generated by Copilot at e20b6ae</samp>

This pull request refactors the rule engine module and API to use persistent term instead of ets table for storing and querying rules, and to use a metadata map and a macro for the rule record definition. This improves the performance, readability, and maintainability of the rule engine code. It also fixes some issues and enhances the test coverage of the rule engine API. The changes are documented in the change log file `changes/ce/perf-11282.en.md`.

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [x] Added tests for the changes
- [x] Changed lines covered in coverage report
- [x] Change log has been added to `changes/(ce|ee)/(feat|perf|fix)-<PR-id>.en.md` files
- [x] For internal contributor: there is a jira ticket to track this change
- [x] If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up
- [x] ~Schema changes are backward compatible~

## Checklist for CI (.github/workflows) changes

- [x] ~If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)~
- [x] ~Change log has been added to `changes/` dir for user-facing artifacts update~

PR #11313, opened 2023-07-20
perf(ruleeng): employ `emqx_topic_index` to speed up topic matching

keynslug:
Made an attempt to speed up the rule engine hot-path in spare time. The meat is in `emqx_topic_index` which utilizes ordered set semantics to match topics to topic filters. The `emqx_topic_index_SUITE` is basically just a `emqx_trie_SUITE` retrofitted to the new API, to verify they both give the same results with same testing data.

### <samp>🤖 Generated by Copilot at 3ba3fb4</samp>

This pull request introduces a new module `emqx_topic_index` that provides a fast and accurate topic index for matching topics to topic filters. It also refactors the `emqx_rule_engine` module to use the topic index for rule lookup and updates the topic index when rules are added or removed. Additionally, it defines a macro for the topic index table name and adds a test suite for the new module.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_topic_index.erl
+++ b/apps/emqx/src/emqx_topic_index.erl
@@ -1,0 +1,156 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+%% @doc Topic index for matching topics to topic filters.
+%%
+%% Works on top of ETS ordered_set table. Keys are parsed topic filters
+%% with record ID appended to the end, wrapped in a tuple to disambiguate from
+%% topic filter words. Existing table may be used if existing keys will not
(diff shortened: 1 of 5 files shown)


keynslug:
> Couldn't we convert the filter to a matchspec and make select?

Just to clarify: we need to tell if a topic matches some subset of topic filters stored in a ETS table, not vice versa. So we would need to convert topic `a/b/c` to something like `[["a", "b", "c"], ["a", "b", "c", "#"], ["a", "b", "+"], ["a", "b", "+", "#"], ...]` which explodes quite quickly. Maybe I'm missing something though.

savonarola:
Ah, I see; we are searching filter by topic, not vice versa.

apps/emqx/src/emqx_topic_index.erl
```diff
@@ -0,0 +1,156 @@
...
+
+-export([get_id/1]).
+-export([get_topic/1]).
+-export([get_record/2]).
+
+-type key(ID) :: [binary() | '+' | '#' | {ID}].
+-type match(ID) :: key(ID).
+
+new() ->
+    ets:new(?MODULE, [public, ordered_set, {write_concurrency, true}]).
```
JimMoen, line 48: `new/0` is not used anywhere except in test cases.

apps/emqx/src/emqx_topic_index.erl
```diff
@@ -0,0 +1,156 @@
...
+    stop.
+
+match_init(Topic) ->
+    case emqx_topic:words(Topic) of
+        [W = <<"$", _/bytes>> | Rest] ->
+            % NOTE
+            % This will effectively skip attempts to match special topics to `#` or `+/...`.
+            {Rest, [W]};
+        Words ->
+            {Words, []}
```
JimMoen, line 136:
```suggestion
        [W | RestWords] ->
            {RestWords, [W]}
```

JimMoen (changes requested):

JimMoen:
And Plz change base to `release-51`. We'll merge it into next release.

JimMoen:
Cherry picked and pushed on #11282

PR #11317, opened 2023-07-21
improve: use foldl/3 instead of foldr/3

caixiangyue:
foldl/3 is tail recursive and is usually preferred to foldr/3.
https://www.erlang.org/doc/man/lists.html#foldr-3


zhongwencool:
IMO: Use foldr mean we care about the list's order, and don't want to use lists:reverse.
PS:  Tail recursion is not always the optimal solution in all scenarios, especial in short list.

PR #10645, opened 2023-05-08
Data bridge target unavailable

paulozulato:
Fixes https://emqx.atlassian.net/browse/EMQX-9026

### <samp>🤖 Generated by Copilot at 6b23c31</samp>

This pull request enhances the error handling and reporting of the MySQL and PostgreSQL connectors when the target table is not created. It also updates the default SQL statement for the MySQL connector to match the new table schema. It affects the files `emqx_connector_mysql.erl`, `emqx_connector_pgsql.erl`, `emqx_resource_buffer_worker.erl`, and `emqx_ee_bridge_mysql.erl`.

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [x] Added tests for the changes
- [x] Changed lines covered in coverage report
- [x] Change log has been added to `changes/{ce,ee}/(feat|perf|fix)-<PR-id>.en.md` files
- [x] For internal contributor: there is a jira ticket to track this change
- [ ] ~If there should be document changes, a PR to emqx-docs.git is sent, or a jira ticket is created to follow up~
- [ ] ~Schema changes are backward compatible~

## Checklist for CI (.github/workflows) changes

- [ ] ~If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)~
- [ ] ~Change log has been added to `changes/` dir for user-facing artifacts update~

PR #11338, opened 2023-07-24
fix(postgres_bridge): fix table existence check and handle sync_required

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10629

During health checking, we check whether tables in the SQL statement exist.  Such check was done by asking the backend to parse the statement using a named prepared statements.  Concurrent health checks could then result in the error:

```erlang
{error,{error,error,<<"42P05">>,duplicate_prepared_statement,<<"prepared statement \"get_status\" already exists">>,[{file,<<"prepare.c">>},{line,<<"451">>},{routine,<<"StorePreparedStatement">>},{severity,<<"ERROR">>}]}}
```

This could lead to an inconsistent state in the driver process, which would crash later when a message from the backend (`READY_FOR_QUERY`, "idle"):

```
  2023-07-24T13:05:58.892043+00:00 [error] Generic server <0.2134.0> terminating. Reason: {'module could not be loaded',[{undefined,handle_message,[90,<<"I">>,...
```

Added calls to `epgsql:sync/1` for functions that could return `{error, sync_required}`.

Also, redundant calls to `parse2` were removed to reduce the number of requests.

### <samp>🤖 Generated by Copilot at 55e95ea</samp>

This pull request enhances the PostgreSQL bridge feature of EMQ X by improving the error handling and robustness of the `emqx_connector_pgsql` module and adding a new test case to the `emqx_bridge_pgsql_SUITE` module. The changes aim to prevent connection failures and bad parse errors when using the bridge to sync MQTT messages with PostgreSQL tables.

## PR Checklist

### DIFF
--- a/apps/emqx_connector/src/emqx_connector_pgsql.erl
+++ b/apps/emqx_connector/src/emqx_connector_pgsql.erl
@@ -62,6 +62,11 @@
         prepare_statement := epgsql:statement()
     }.
 
+%% FIXME: add `{error, sync_required}' to `epgsql:execute_batch'
+%% We want to be able to call sync if any message from the backend leaves the driver in an
+%% inconsistent state needing sync.
+-dialyzer({nowarn_function, [execute_batch/3]}).
+
 %%=====================================================================
 
 roots() ->
@@ -252,6 +257,8 @@ on_sql_query(InstId, PoolName, Type, NameOrSQL, Data) ->
                 reason => Reason
             }),
             case Reason of
+                sync_required ->
+                    {error, {recoverable_error, Reason}};
                 ecpool_empty ->
                     {error, {recoverable_error, Reason}};
                 {error, error, _, undefined_table, _, _} ->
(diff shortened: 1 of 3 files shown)


apps/emqx_connector/src/emqx_connector_pgsql.erl
```diff
@@ -344,6 +336,30 @@ do_check_prepares(State = #{pool_name := PoolName, prepare_sql := {error, Prepar
...
+                    validate_table_existence(Rest, SQL)
+            end;
+        _ ->
+            validate_table_existence(Rest, SQL)
+    catch
+        exit:{noproc, _} ->
+            validate_table_existence(Rest, SQL)
+    end;
+validate_table_existence([], _SQL) ->
+    %% All workers either replied an unexpected error; we will retry
```
paulozulato, line 359:
```suggestion
    %% All workers replied an unexpected error; we will retry
```

paulozulato: if all workers replied error, shouldn't we already move the bridge to `disconnected`?

paulozulato (approved):

apps/emqx_connector/src/emqx_connector_pgsql.erl
thalesmg: We are just checking for table existence.  It'll transition to disconnected only if we are certain that the table does not exist, as was the behavior before.  They may have just timed out, so we defer the check.

savonarola:
Maybe we are a bit overcomplicating things becase we are trying to unify two different scenarios in the health check:
1. form validation which occurs when a user interacts with the dashboard.
2. internal health checking of a resource when it tries to start or resume.

My idea is that:
* The first one may be arbitrarily complicated but done in a different resource connection, not related to the battle used ones
* The second one is when a resource is making an attempt to (re)start; they should be simple and robust. Its feels a bit awkward that we check table existence in _every_ connection of the pool for an _attempt_ to start.

thalesmg:
@savonarola  I think the goal here is to detect errors like this as early as possible and make the bridge drop (not even enqueuing) messages to signal that the operator needs to act. (refs: [1](https://emqx.atlassian.net/browse/EMQX-9026) and [2](https://github.com/emqx/emqx/pull/10645))

Checking for the table existence on each health check might be overzealousness.  It _might_ happen while the bridge is already connected, but we could decide in the future to relax this check and only perform it during initial connection, indeed.

PR #11365, opened 2023-07-28
fix(http_bridge): don't attempt to convert headers to atoms

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10653

### <samp>🤖 Generated by Copilot at bfa5f5f</samp>

This pull request adds support for customizing the atom conversion of webhook bridge config keys, especially the headers key, to avoid atom exhaustion. It also updates the versions of the emqx, emqx_bridge, emqx_bridge_http, and emqx_utils applications, and adds new test cases for the new features. The main files affected are `emqx_bridge_resource.erl`, `emqx_utils_maps.erl`, `emqx_bridge_http_schema.erl`, and their corresponding test modules.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_api.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_api.erl
@@ -541,30 +541,23 @@ schema("/bridges_probe") ->
     case emqx_dashboard_swagger:filter_check_request_and_translate_body(Request, RequestMeta) of
         {ok, #{body := #{<<"type">> := ConnType} = Params}} ->
             Params1 = maybe_deobfuscate_bridge_probe(Params),
-            try emqx_bridge_resource:create_dry_run(ConnType, maps:remove(<<"type">>, Params1)) of
+            case emqx_bridge_resource:create_dry_run(ConnType, maps:remove(<<"type">>, Params1)) of
                 ok ->
                     ?NO_CONTENT;
-                {error, #{kind := validation_error} = Reason} ->
+                {error, #{kind := validation_error} = Reason0} ->
+                    Reason = redact(Reason0),
                     ?BAD_REQUEST('TEST_FAILED', map_to_json(Reason));
                 {error, Reason0} when not is_tuple(Reason0); element(1, Reason0) =/= 'exit' ->
-                    Reason =
+                    Reason1 =
                         case Reason0 of
                             {unhealthy_target, Message} -> Message;
                             _ -> Reason0
                         end,
+                    Reason = redact(Reason1),
                     ?BAD_REQUEST('TEST_FAILED', Reason)
-            catch
-                %% We need to catch hocon validation errors here as well because,
-                %% currently, when defining the API union member selector, we can only use
-                %% references to fields, and they don't share whole-bridge validators if
-                %% they exist.  Such validators will only be triggered by
-                %% `create_dry_run'...
-                throw:{_Schema, [#{kind := _} = Reason0 | _]} ->
-                    Reason = redact(Reason0),
-                    ?BAD_REQUEST('TEST_FAILED', map_to_json(Reason))
             end;
         BadRequest ->
-            BadRequest
+            redact(BadRequest)
     end.
 
 maybe_deobfuscate_bridge_probe(#{<<"type">> := BridgeType, <<"name">> := BridgeName} = Params) ->
@@ -999,7 +992,9 @@ call_operation(NodeOrAll, OperFunc, Args = [_Nodes, BridgeType, BridgeName]) ->
         {error, timeout} ->
             ?BAD_REQUEST(<<"Request timeout">>);
         {error, {start_pool_failed, Name, Reason}} ->
-            Msg = bin(io_lib:format("Failed to start ~p pool for reason ~p", [Name, Reason])),
+            Msg = bin(
+                io_lib:format("Failed to start ~p pool for reason ~p", [Name, redact(Reason)])
+            ),
             ?BAD_REQUEST(Msg);
         {error, not_found} ->
             BridgeId = emqx_bridge_resource:bridge_id(BridgeType, BridgeName),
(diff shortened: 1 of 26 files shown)


apps/emqx_bridge/src/emqx_bridge_resource.erl
```diff
@@ -261,7 +267,12 @@ recreate(Type, Name, Conf, Opts) ->
 create_dry_run(Type, Conf0) ->
     TmpName = iolist_to_binary([?TEST_ID_PREFIX, emqx_utils:gen_id(8)]),
     TmpPath = emqx_utils:safe_filename(TmpName),
-    Conf = emqx_utils_maps:safe_atom_key_map(Conf0),
+    %% Some config keys might be maps that we don't want to covert to atoms.  This depends
+    %% on the schema of each bridge type.
+    Opts = #{
+        excluded_paths => config_keys_to_skip_atom_conversion(Type)
+    },
+    Conf = emqx_utils_maps:safe_atom_key_map(Conf0, Opts),
```
savonarola, line 275: Do I understand correctly that we have a raw config here converted to some kind of a "translated" config?

Is that convertion to atom keys a shortcut made instead of converting with schema 🤔 
Could we just make a full translation instead?

thalesmg: It should have already been validated by the API calls.

This was already like this for quite some time.  :thinking:

thalesmg: > Could we just make a full translation instead?

Sorry, what do you mean by a full translation?

savonarola: I mean, that in normal flow
* raw config comes through the API
* it is passed to `emqx_conf:update`
* callbacks receive translated (in terms of hocon) config, i.e. with atom keys, converted values, etc.
* callbacks start a new bridge

Here, as far as I understand, the dry-run checker receives a raw config, but does not send it to `emqx_conf:update` because it just wants to check resource connectivity. So it runs `emqx_utils_maps:safe_atom_key_map` as a shortcut for translation, just to make keys as atoms to make it possible to pass as the options to the resource.

My concern is that extending this shortcut usage, we start to build and support some sort of "parallel semi-transltion" mechanism which approximates `hocon:translate`. 😅 

Maybe we could do something like `Conf = hocon_tconf:check_plain(SomeSchema, RawConf, #{atom_key => true}, [Root])` instead.

thalesmg: That's a much better solution, indeed.  I've updated the branch to use that approach.  Thank you!  :beers:

apps/emqx_bridge_http/test/emqx_bridge_http_SUITE.erl
```diff
@@ -613,6 +636,38 @@ t_rule_action_expired(Config) ->
...
+            ?assertMatch(
+                {ok, {{_, 204, _}, _Headers, _Body}},
+                probe_bridge_api(BridgeConfig)
+            ),
+            emqx_bridge_testlib:create_bridge_api(
+                ?BRIDGE_TYPE,
+                ?BRIDGE_NAME,
+                BridgeConfig
+            ),
+            ok
```
paulozulato, line 665: should we check if the "some non existent atom" key was sent on header?

paulozulato (dismissed):

apps/emqx_bridge/src/emqx_bridge_resource.erl
thalesmg: Unfortunately, this seems to have consequences in some bridges that have validators and pattern matches that expect atoms or binaries in some places. :see_no_evil: 
I'll have to hunt and fix those.

thalesmg: Hopefully, those should be fixed now.  They were mostly places where `map()` values were expected and some validators that matched on raw binary keys.  Added a few more tests on those.  :crossed_fingers:

savonarola (dismissed):

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -553,6 +553,15 @@ schema("/bridges_probe") ->
                             _ -> Reason0
                         end,
                     ?BAD_REQUEST('TEST_FAILED', Reason)
+            catch
+                %% We need to catch hocon validation errors here as well because,
+                %% currently, when defining the API union member selector, we can only use
+                %% references to fields, and they don't share whole-bridge validators if
+                %% they exist.  Such validators will only be triggered by
+                %% `create_dry_run'...
+                throw:{_Schema, [#{kind := validation_error} = Reason0]} ->
```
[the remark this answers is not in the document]
thalesmg: I'm mirroring what we currently do in the non-throwing [clause above](https://github.com/emqx/emqx/blob/20443267348a753517606e896018687b38339e00/apps/emqx_bridge/src/emqx_bridge_api.erl#L547-L548).  I wonder which others might be.  :thinking:

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -541,7 +541,7 @@ schema("/bridges_probe") ->
     case emqx_dashboard_swagger:filter_check_request_and_translate_body(Request, RequestMeta) of
         {ok, #{body := #{<<"type">> := ConnType} = Params}} ->
             Params1 = maybe_deobfuscate_bridge_probe(Params),
-            case emqx_bridge_resource:create_dry_run(ConnType, maps:remove(<<"type">>, Params1)) of
+            try emqx_bridge_resource:create_dry_run(ConnType, maps:remove(<<"type">>, Params1)) of
                 ok ->
                     ?NO_CONTENT;
                 {error, #{kind := validation_error} = Reason} ->
```
zmstone, line 547: how does the code end up here before this PR ? I cannot seem to find a type check called anywhere.

thalesmg: I assume it was this code inside `create_dry_run`:

```erlang
            try
                ParseConf = parse_confs(bin(Type), TmpName, ConfNew),
                emqx_resource:create_dry_run_local(bridge_to_resource_type(Type), ParseConf)
            catch
                %% validation errors
                throw:Reason ->
                    {error, Reason}
```

thalesmg: But indeed there seems to be no type checking there either.  :thinking:

thalesmg:
Spellcheck CI failure unrelated to changes herein.

See #11397

zmstone (approved):

PR #11367, opened 2023-07-28
feat(gcp-iot): port GCP IoT Core compatibility layer from e4.4

paulozulato:
Fixes https://emqx.atlassian.net/browse/EMQX-10341

### <samp>🤖 Generated by Copilot at 159fcee</samp>

This pull request adds the gcp_device authentication mechanism for emqx_authn, which allows users to authenticate MQTT clients using Google Cloud Platform IoT Core device credentials. It also creates a new emqx_gcp_device application, which provides the API and logic for managing and importing GCP devices and their keys and configs. It updates the versions of the emqx_authn and emqx applications, and adds the necessary tests, documentation, and configuration files for the new features.

## PR Checklist


apps/emqx_authn/src/emqx_authn.erl
```diff
@@ -30,7 +30,16 @@
 
 -include("emqx_authn.hrl").
 
+-dialyzer({nowarn_function, [providers/1]}).
+
 providers() ->
+    providers(emqx_release:edition()).
```
thalesmg, line 36: It seems that there's a conflict on how to setup EE authenticators: https://github.com/emqx/emqx/pull/11386/files#diff-bc603baaf71ddf57d13fa3467b03a9e50b330e01b1322f3dba295df28093653aR44

paulozulato: Ack. I'll follow @lafirest's approach

apps/emqx_gcp_device/src/emqx_gcp_device.erl
```diff
@@ -0,0 +1,256 @@
...
+
+-spec get_device(deviceid()) -> {ok, formatted_device()} | not_found.
+get_device(DeviceId) ->
+    try ets:lookup(?TAB, DeviceId) of
+        [] ->
+            not_found;
+        [Device] ->
+            {ok, format_device(Device)}
+    catch
+        error:badarg ->
```
lafirest, line 89: I think in a normal situation, this clause should never be reached.

apps/emqx_gcp_device/src/emqx_gcp_device.erl
```diff
@@ -0,0 +1,256 @@
...
+        key_type => KeyType,
+        key => Key,
+        expires_at => ExpiresAt
+    }.
+
+-spec put_config(deviceid(), encoded_config()) -> ok.
+put_config(DeviceId, EncodedConfig) ->
+    Config = base64:decode(EncodedConfig),
+    Topic = config_topic(DeviceId),
+    Message = emqx_message:make(undefined, 1, Topic, Config, #{retain => true}, #{}),
```
lafirest, line 213: Q: is the `retain` must be true?

lafirest: And why not use the deviceId as the clientid (from) for this message

apps/emqx_gcp_device/src/emqx_gcp_device_api.erl
```diff
@@ -0,0 +1,488 @@
...
+    [
+        {key,
+            hoconsc:mk(
+                binary(),
+                #{
+                    desc => ?DESC(key),
+                    required => true,
+                    example => <<
+                        "-----BEGIN CERTIFICATE-----\n"
+                        "MIIBEjCBuAIJAPKVZoroXatKMAoGCCqGSM49BAMCMBExDzANBgNVBAMMBnVudXNl\n"
```
lafirest, line 304: I am incredulous that is this example suitable, maybe we could let it empty, otherwise some vulnerability scanners may report private key leakage.

apps/emqx_gcp_device/src/emqx_gcp_device_authn.erl
```diff
@@ -0,0 +1,233 @@
...
+    Jwk = jose_jwk:from_pem(KeyRaw),
+    case jose_jws:verify(Jwk, Jwt) of
+        {true, _, _} ->
+            true;
+        {false, _, _} ->
+            false
+    end.
+
+gcp_deviceid_from_clientid(#{clientid := <<"projects/", RestClientId/binary>>}) ->
+    case binary:split(RestClientId, <<"/">>, [global]) of
```
lafirest, line 202: nit: could use the `emqx_topic:tokens/1`

apps/emqx_gcp_device/src/emqx_gcp_device.erl
paulozulato: I'll remove the catch

apps/emqx_gcp_device/src/emqx_gcp_device_api.erl
paulozulato: This is available for GCP device CT ([c2_ec_cert](https://github.com/emqx/emqx/pull/11367/files#diff-1b5c04fcb1af507e8f002b11fc819a57838a54ec161dd72c798dd314ff19fe6f)), so I thought it would not be a problem to expose it here. I'll remove it anyway.

apps/emqx_gcp_device/src/emqx_gcp_device_authn.erl
paulozulato: It could be used, but, as `RestClientId` isn't a topic, I think it would be weird to use `emqx_topic:tokens` here. Also, if in the future we add a new separator for topic, it might have some non expected impact here (ok, it probably will not happen, but just to mention it 😅)

savonarola (dismissed):

apps/emqx_gcp_device/src/emqx_gcp_device.erl
```diff
@@ -0,0 +1,253 @@
...
+
+-spec create_table() -> ok.
+create_table() ->
+    ok = mria:create_table(?TAB, [
+        {rlog_shard, ?AUTH_SHARD},
+        {type, ordered_set},
+        {storage, disc_copies},
+        {record_name, emqx_gcp_device},
+        {attributes, record_info(fields, emqx_gcp_device)},
+        {storage_properties, [{ets, [{read_concurrency, true}]}]}
```
thalesmg, line 124: Q: do we need to also tune dets?

```suggestion
        {storage_properties, [{ets, [{read_concurrency, true}]}, {dets, [{auto_save, 10_000}]}]}
```

apps/emqx_gcp_device/src/emqx_gcp_device.erl
```diff
@@ -0,0 +1,253 @@
...
+-define(TAB, ?MODULE).
+
+%%--------------------------------------------------------------------
+%% API
+%%--------------------------------------------------------------------
+
+-spec put_device(formatted_device()) -> ok.
+put_device(FormattedDevice) ->
+    try
+        transaction(fun() -> put_device_no_transaction(FormattedDevice) end)
```
thalesmg, line 70: Q: if the operations are dirty, why do we need to wrap them in a transaction?

apps/emqx_gcp_device/src/emqx_gcp_device.erl
```diff
@@ -0,0 +1,253 @@
...
+    case ets:lookup(?TAB, DeviceId) of
+        [] ->
+            not_found;
+        [Device] ->
+            {ok, format_device(Device)}
+    end.
+
+-spec remove_device(deviceid()) -> ok.
+remove_device(DeviceId) ->
+    ok = transaction(fun() -> mria:dirty_delete({?TAB, DeviceId}) end),
```
thalesmg, line 92:
```suggestion
    ok = mria:dirty_delete({?TAB, DeviceId}),
```
?

apps/emqx_gcp_device/src/emqx_gcp_device.erl
```diff
@@ -0,0 +1,253 @@
...
+            actual_keys(Device)
+    catch
+        error:badarg ->
+            not_found
+    end.
+
+-spec import_devices([formatted_device()]) ->
+    {NumImported :: non_neg_integer(), NumError :: non_neg_integer()}.
+import_devices(FormattedDevices) when is_list(FormattedDevices) ->
+    transaction(fun() -> lists:foldl(fun import_device/2, {0, 0}, FormattedDevices) end).
```
thalesmg, line 110:
```suggestion
    lists:foldl(fun import_device/2, {0, 0}, FormattedDevices).
```
?

apps/emqx_gcp_device/src/emqx_gcp_device.erl
thalesmg: By using the same clientid, [tracing](https://github.com/emqx/emqx/blob/20443267348a753517606e896018687b38339e00/apps/emqx/src/emqx_broker.erl#L217) may pick up those messages.

apps/emqx_gcp_device/src/emqx_gcp_device_authn.erl
```diff
@@ -0,0 +1,233 @@
...
+        {mechanism, emqx_authn_schema:mechanism('gcp_device')}
+    ] ++ emqx_authn_schema:common_fields().
+
+refs(#{<<"mechanism">> := <<"gcp_device">>}) ->
+    [hoconsc:ref(?MODULE, gcp_device)];
+refs(_) ->
+    throw(#{
+        field_name => mechanism,
+        expected => "gcp_device"
+    }).
```
thalesmg, line 106: Q: will there be other mechanisms for GCP authn?

Since this module is specific to GCP and there's a single option for it, it seems that using an union selector might be unnecessary.

apps/emqx_gcp_device/src/emqx_gcp_device_authn.erl
```diff
@@ -0,0 +1,233 @@
...
+            ?tp(authn_gcp_device_check, #{
+                result => ignore, reason => "key not found", client => ClientInfo
+            }),
+            ?TRACE_AUTHN_PROVIDER(debug, "auth_ignored", #{
+                reason => "key not found",
+                client => ClientInfo
+            }),
+            ignore;
+        Keys ->
+            case any_key_matches(Keys, ClientInfo) of
```
thalesmg, line 165: Q: is it possible that `[] = Keys`?  If so, wouldn't it fall under case 3.1 of the comment above (result should be `ignore`)?

apps/emqx_gcp_device/src/emqx_gcp_device_api.erl
```diff
@@ -0,0 +1,479 @@
...
+                            <<"Bad Request">>
+                        )
+                    }
+            },
+        delete => #{
+            description => ?DESC(gcp_device_delete),
+            tags => ?TAGS,
+            parameters => [
+                {deviceid,
+                    hoconsc:mk(binary(), #{
```
thalesmg, line 168: Nit: since the same path param is used multiple times, it could be moved to a fn definition and shared.

apps/emqx_gcp_device/src/emqx_gcp_device.erl
savonarola: 🤔 10s looks like a reasonable timeout for autosave, but why do other tables do not configure it? Like `emqx_authn_mnesia`.

apps/emqx_gcp_device/src/emqx_gcp_device.erl
savonarola: For batch operations, this seem to improve performance significantly:  https://github.com/emqx/emqx/pull/11389/files

apps/emqx_gcp_device/src/emqx_gcp_device.erl
```diff
@@ -0,0 +1,253 @@
...
+    {atomic, ok} = mria:clear_table(?TAB),
+    ok.
+
+%%--------------------------------------------------------------------
+%% Internal functions
+%%--------------------------------------------------------------------
+
+-spec transaction(function()) -> term().
+transaction(Fun) ->
+    {atomic, Res} = mria:transaction(?AUTH_SHARD, Fun),
```
savonarola, line 139: We should use `mria:dirty_(a)sync/2` instead, probably.

apps/emqx_gcp_device/src/emqx_gcp_device_authn.erl
savonarola: `Keys = []` is considered to be `3.2.1` case

apps/emqx_gcp_device/src/emqx_gcp_device.erl
thalesmg: > 🤔 10s looks like a reasonable timeout for autosave, but why do other tables do not configure it? Like `emqx_authn_mnesia`.

Good question.

It seems most `disc_only` tables rely on the default 3 minutes autosave time, and it's not clear if that's intentional.   :thinking:

apps/emqx_gcp_device/src/emqx_gcp_device.erl
thalesmg: Since there's no need for a transaction, maybe it could reduce the number of operations by using `mria:async_dirty` (or `mria:sync_dirty`) as the linked PR does?

https://github.com/emqx/emqx/pull/11389/files#diff-dfb046d7c8660f7ef165880663126a5975583aa1c6e7ff57245d3ee77b1abb63R302

thalesmg: Now I see that "`transaction`" has been changed to call one of those functions.  :joy_cat: 

As a minor nit, I'd try to rename it to something else as it's not a transaction, I think.  Maybe `perform_dirty` or similar. :thinking:

apps/emqx_gcp_device/src/emqx_gcp_device.erl
thalesmg: Yes, now I see it has been changed to call `{,a}sync_dirty` here.  :smile_cat:

apps/emqx_gcp_device/src/emqx_gcp_device_api.erl
```diff
@@ -0,0 +1,459 @@
...
+    validate(Fields, Device, Validated#{key => Key});
+validate([expires_at | Fields], #{<<"expires_at">> := Expire} = Device, Validated) when
+    is_integer(Expire)
+->
+    validate(Fields, Device, Validated#{expires_at => Expire});
+validate([expires_at | _Fields], #{<<"expires_at">> := _}, _Validated) ->
+    {error, <<"invalid expires_at">>};
+validate([expires_at | Fields], Device, Validated) ->
+    validate(Fields, Device, Validated#{expires_at => 0});
+validate([Field | Fields], Device, Validated) when Field =:= deviceid; Field =:= key ->
```
lafirest, line 411: `Field =:= key` has matched in the above clause (L401)

apps/emqx_gcp_device/src/emqx_gcp_device_api.erl
```diff
@@ -0,0 +1,459 @@
...
+            {error, <<"invalid field: ", FieldBin/binary>>}
+    end;
+validate([keys | Fields], #{<<"keys">> := Keys} = Device, Validated) when is_list(Keys) ->
+    try validate_keys(Keys) of
+        {ok, ValidatedKeys} ->
+            validate(Fields, Device, Validated#{keys => ValidatedKeys});
+        {error, _} = Error ->
+            Error
+    catch
+        error:badarg ->
```
lafirest, line 438: Q:
Where will throw this error, I can't find the location of it in either `validate_key` or `validate`

apps/emqx_gcp_device/src/emqx_gcp_device.erl
```diff
@@ -0,0 +1,268 @@
...
+
+%%--------------------------------------------------------------------
+%% Internal API
+%%--------------------------------------------------------------------
+
+-spec create_table() -> ok.
+create_table() ->
+    ok = mria:create_table(?TAB, [
+        {rlog_shard, ?AUTH_SHARD},
+        {type, ordered_set},
```
lafirest, line 130: Can the type be  `set`?

paulozulato: Ack.

apps/emqx_gcp_device/src/emqx_gcp_device_api.erl
paulozulato: But at L401 it also checks for <<"key">> being present. If this field is not present, it'll match this case on L411.

apps/emqx_gcp_device/src/emqx_gcp_device_api.erl
paulozulato: I think that's not necessary. I'll remove this catch.

apps/emqx_gcp_device/src/emqx_gcp_device.erl
thalesmg: From past discussions, we try to use `ordered_set` since it has the same properties of `set` without too much of a performance penalty.

paulozulato: OK. So should we keep `ordered_set` here, right?
Also, just to know: what was the reason to try to use `ordered_set`? 🫣

lafirest (approved):

apps/emqx_gcp_device/src/emqx_gcp_device_sup.erl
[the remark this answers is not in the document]
paulozulato: it kept it here because app start needs to return a PID and I need the app start to be executed to create the gcp device table when application starts, otherwise one would not be allowed to import gcp devices before configuring the authenticator (if tables were created on gcp authenticator start instead of app start).

thalesmg (approved):

PR #11382, opened 2023-08-01
Auth refactor stage 1

savonarola:
Fixes [EMQX-9262](https://emqx.atlassian.net/browse/EMQX-9262)

This PR provides better isolation of `emqx_authn` and `emqx` apps.

In v4, `emqx_auth_*` apps are typical plugins which integrate into the core app through hooks.

In v5, things are more complicated: a part of the authentication logic resides in the core `emqx` app.
This is the `emqx_authentication` module and `emqx_authentication_config` handler. 

`emqx_authentication` manages authentication chains and provides additional injection points for registering authentication "providers".

Presumably, the reason for `emqx_authentication` residing inside `emqx` application is the presence of per-listener authentication. Listener configs belong to the core app, and so it should know how to update listener's authentication chains.

This also leads to ad-hoc hotwired code, like updating authn providers' certs directly in `emqx_listeners`. Also, such a code organization is a bit self-contradictory: 
* we **inject** providers from `emqx_authn` to `emqx` (`emqx_authn` -> `emqx` dependency);
* but we use **concrete providers** when fetching their schemas (`emqx` -> `emqx_authn` dependency). 

In the PR, we try to eliminate such an inconsistency, making the whole authentication _plugged_ into `emqx` more like it is done in v4.
This also gives a path to
* getting rid of provider injection since `emqx_authn` is now a concrete top-level plugin-like app and is free to depend on all the concrete providers and their schemas.
* free refactoring of `emqx_authn` to individual apps and its merging with `emqx_authz`.

To resolve the entanglement we make two changes:
* (the little one): we refactor the schema injecting trick a bit. Technically, the mechanism is the same, but now the whole injection is opaque to `emqx` and is controlled (minimally) by `emqx_conf`.
* (the major one): we introduce optional callbacks to `emqx_config_handler`, allowing us to handle config changes that are caused by some other changes made somewhere upper in the config tree.

This makes authn manipulations in listeners unnecessary because now we obtain them for free.

Also, we
* add some tests to test these pre-post actions;
* update authn tests to use new suite setup/teardown helpers.

### <samp>🤖 Generated by Copilot at 71a57e1</samp>

Refactor the authentication module to use hooks and schema injection. Move the authentication module code and header files from the `emqx` application to a new application `emqx_authn`. Update the access control module, the listeners module, the schema module, and the test modules to use the new hook API and schema injection API for authentication. Bump the version numbers of the `emqx` and `emqx_authn` applications.

## PR Checklist

- [na] Added tests for the changes
- [na] Change log has been added to `changes/(ce|ee)/(feat|perf|fix)-<PR-id>.en.md` files


thalesmg (commented):
Didn't finish reviewing yet :book:

apps/emqx/src/emqx_schema_hooks.erl
```diff
@@ -0,0 +1,118 @@
...
+        AllInjections,
+        ModuleInjections
+    ).
+
+inject_fields([]) ->
+    ok;
+inject_fields([{PointName, Fields} | Rest]) ->
+    case any_injections(PointName) of
+        true ->
+            inject_fields(Rest);
```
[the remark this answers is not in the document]
savonarola: Injection is idempotent and does not have "update semantics" (as it was before).
We also assume that there is only one module managing schema (like `emqx_conf_schema`), so even if written concurrently, it will be the same value.

apps/emqx_authn/rebar.config
```diff
@@ -34,4 +34,6 @@
 {cover_opts, [verbose]}.
 {cover_export_enabled, true}.
 
+{erl_first_files, ["src/emqx_authentication.erl"]}.
```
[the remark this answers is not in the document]
savonarola: This module defines a behaviour for providers which are in the same app. We have compile error otherwise.

This will be unnecessary once providers go away from this app.

zmstone:
Great summary in the PR description.

Just adding to this:
> Presumably, the reason for emqx_authentication residing inside emqx application is the presence of per-listener authentication. Listener configs belong to the core app, and so it should know how to update listener's authentication chains.

As I understood it, it was mainly because in the early stage of the design, we were aiming to provide another set of config/callback management for authn. For example, with the `emqx` core app, someone can implement their own config and callbacks to form a chain.
It should have been the same for `authz` but they were developed by different people, and they probably had different understandings to the abstraction.

apps/emqx/src/emqx_config_handler.erl
```diff
@@ -377,57 +399,242 @@ get_sub_config(ConfKey, Conf) when is_map(Conf) ->
...
+propagate_pre_config_updates_to_subconf_key(
+    Key,
+    #{
+        handlers := Handlers,
+        old_raw_conf := OldRawConf,
+        update_req := UpdateReq,
+        conf_key_path := ConfKeyPath
+    } = Ctx
+) ->
+    AtomKey = atom(Key),
```
savonarola, line 493: We should try to avoid this

apps/emqx/src/emqx_config_handler.erl
```diff
@@ -53,11 +53,17 @@
...
-    post_config_update/5
+    propagated_pre_config_update/3,
+    post_config_update/5,
+    propagated_post_config_update/5
 ]).
 
 -callback pre_config_update([atom()], emqx_config:update_request(), emqx_config:raw_config()) ->
-    {ok, emqx_config:update_request()} | {error, term()}.
+    ok | {ok, emqx_config:update_request()} | {error, term()}.
+-callback propagated_pre_config_update(
```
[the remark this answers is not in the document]
zmstone: or maybe this is more relevant: `{propagated_path, [listeners, '?', '?', authentiation]}`

apps/emqx/src/emqx_schema.erl
```diff
@@ -216,15 +215,14 @@ roots(high) ->
...
-        {?EMQX_AUTHENTICATION_CONFIG_ROOT_NAME, authentication(global)},
         %% NOTE: authorization schema here is only to keep emqx app pure
         %% the full schema for EMQX node is injected in emqx_conf_schema.
         {?EMQX_AUTHORIZATION_CONFIG_ROOT_NAME,
             sc(
                 ref(?EMQX_AUTHORIZATION_CONFIG_ROOT_NAME),
                 #{importance => ?IMPORTANCE_HIDDEN}
             )}
-    ];
+    ] ++ emqx_schema_hooks:injection_point('roots.high');
```
zmstone, line 225: please keep this at the old position in the list.

apps/emqx_conf/rebar.config
```diff
@@ -1,7 +1,10 @@
 %% -*- mode: erlang -*-
 
 {erl_opts, [debug_info]}.
-{deps, [{emqx, {path, "../emqx"}}]}.
+{deps, [
+    {emqx, {path, "../emqx"}},
+    {emqx_authn, {path, "../emqx_authn"}}
```
savonarola, line 6: Do we really need this?

apps/emqx/src/emqx_config_handler.erl
```diff
@@ -377,57 +399,242 @@ get_sub_config(ConfKey, Conf) when is_map(Conf) ->
...
-call_pre_config_update(#{?MOD := HandlerName}, OldRawConf, UpdateReq, ConfKeyPath) ->
-    case erlang:function_exported(HandlerName, pre_config_update, 3) of
+call_pre_config_update(Ctx) ->
+    case call_proper_pre_config_update(Ctx) of
+        {ok, NewUpdateReq0} ->
+            case
+                propagate_pre_config_updates_to_subconf(Ctx#{
+                    update_req => NewUpdateReq0
+                })
+            of
```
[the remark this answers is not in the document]
savonarola: I will update in a follow up.

apps/emqx/src/emqx_schema.erl
savonarola: ✅

apps/emqx/src/emqx_config_handler.erl
savonarola: ✅

apps/emqx/src/emqx_access_control.erl
```diff
@@ -29,6 +30,14 @@
 -compile(nowarn_export_all).
 -endif.
 
+-define(TRACE_RESULT(Label, Tag, Result, Reason), begin
```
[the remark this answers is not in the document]
savonarola: ✅

apps/emqx_conf/rebar.config
savonarola: Yes, because of `emqx_conf_cli` ✅

zmstone (approved):

Issue #11310, opened 2023-07-20
Core node restart caused 11 replica nodes to restart, resulting in 8 million connections being disconnected


Estheriii:
### What happened?

A cluster with 3 core nodes and 18 replica nodes has 13 million connections. Manually restarting a core node will cause 11 replica nodes to restart and 8 million connections to be disconnected

### What did you expect to happen?

One-third of the connections are disconnected, affecting up to 6 connections

### How can we reproduce it (as minimally and precisely as possible)?

start a cluster with operator-controller in k8s cluster
with 3 core and 18 replica
add 13 million connections
than delete a core node 

### EMQX version

<details>
5.0.21
```console
$ ./bin/emqx_ctl broker
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Log files

<details>

</details>
[log.zip](https://github.com/emqx/emqx/files/12101389/log.zip)

id:
@Estheriii I do not think it's a bug. If you do manual restart of a core node without [node evacuation](https://docs.emqx.com/en/enterprise/v5.1/deploy/cluster/rebalancing.html#node-evacuation-and-cluster-load-rebalancing), it is expected that the connections to that node will be dropped.

I would recommend to discuss this issue on the [emqx-operator forum](https://github.com/emqx/emqx-operator/discussions/).
This issue may be related: https://github.com/emqx/emqx-operator/discussions/628

Estheriii:
> @Estheriii I do not think it's a bug. If you do manual restart of a core node without [node evacuation](https://docs.emqx.com/en/enterprise/v5.1/deploy/cluster/rebalancing.html#node-evacuation-and-cluster-load-rebalancing), it is expected that the connections to that node will be dropped.
> 
> I would recommend to discuss this issue on the [emqx-operator forum](https://github.com/emqx/emqx-operator/discussions/). This issue may be related: [emqx/emqx-operator#628](https://github.com/emqx/emqx-operator/discussions/628)

I am simulating a node crash situation where a large number of connections are disconnected due to a node restart, indicating that such a cluster is not highly available

id:
Were clients not able to reconnect after you restarted the core node, or did you loose some data?

id:
> indicating that such a cluster is not highly available

[...]

PR #11388, opened 2023-08-03
fix: increase emqx_router_sup restart intensity

SergeTupchiy:
The goal is to tolerate occasional crashes that can happen under relatively normal conditions and don't seem critical to shutdown the whole app (emqx). For example, mria write/delete call delegated from a replicant to a core node may fail, if the core node is being stopped / restarted / not ready.

Fixes EMQX-10703, #11310

### <samp>🤖 Generated by Copilot at 2769736</samp>

The pull request updates the version number of the `emqx` application and enhances the fault tolerance of the `emqx_router` process by changing its supervisor flags.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_router_sup.erl
+++ b/apps/emqx/src/emqx_router_sup.erl
@@ -41,4 +41,9 @@ init([]) ->
         hash,
         {emqx_router, start_link, []}
     ]),
-    {ok, {{one_for_all, 0, 1}, [Helper, RouterPool]}}.
+    SupFlags = #{
+        strategy => one_for_one,
+        intensity => 10,
+        period => 100
+    },
+    {ok, {SupFlags, [Helper, RouterPool]}}.
--- a/apps/emqx/src/emqx.app.src
+++ b/apps/emqx/src/emqx.app.src
@@ -2,7 +2,7 @@
 {application, emqx, [
     {id, "emqx"},
     {description, "EMQX Core"},
-    {vsn, "5.1.4"},
+    {vsn, "5.1.5"},
     {modules, []},
     {registered, []},
     {applications, [
--- a/apps/emqx/src/emqx_router_helper.erl
+++ b/apps/emqx/src/emqx_router_helper.erl
@@ -146,13 +146,18 @@ handle_info({mnesia_table_event, Event}, State) ->
     ?SLOG(debug, #{msg => "unexpected_mnesia_table_event", event => Event}),
     {noreply, State};
 handle_info({nodedown, Node}, State = #{nodes := Nodes}) ->
-    global:trans(
-        {?LOCK, self()},
-        fun() ->
-            mria:transaction(?ROUTE_SHARD, fun ?MODULE:cleanup_routes/1, [Node])
-        end
-    ),
-    ok = mria:dirty_delete(?ROUTING_NODE, Node),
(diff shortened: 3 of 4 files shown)


apps/emqx/src/emqx_router_sup.erl
```diff
@@ -41,4 +41,9 @@ init([]) ->
         hash,
         {emqx_router, start_link, []}
     ]),
-    {ok, {{one_for_all, 0, 1}, [Helper, RouterPool]}}.
+    SupFlags = #{
+        strategy => one_for_one,
+        intensity => 10,
+        period => 100
+    },
+    {ok, {SupFlags, [Helper, RouterPool]}}.
```
SergeTupchiy, line 49: Note: we can also set `RouterPool` sup as a significant child, so that its own worker restart intensity is not multiplied with this (`emqx_router_sup`) intensity.. :thinking:

SergeTupchiy: Some context about the issue:
the below code in emqx_roter_helper is prone to errors, especially when called on a replicant:
https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_router_helper.erl#L155
The reason is that rpc to core node can fail in more or less normal circumstances (e.g. a core node restart). 
Instead of catching the errors inside emqx_router_helper, it seems reasonable to increase intensity to tolerate occasional crashes (probably in some other places as well) . 
Even if the helper fails to delete a record in that case (#L155), another node is expected to do the same, as nodedown message is sent on all the nodes.

zhongwencool: Since we know some situations can cause the code at line 155 to crash the process, why don't we fix it instead of logging an **incomprehensible Erlang crash error**?
This will cause great panic for our customers, leading to unnecessary trouble.
Perhaps we can catch the error on delete and, if there is an error, directly stop normally to let it restart.
IMO: we should not report avoidable crash logs to customers.

zhongwencool: If we keep this crash log, it is foreseeable that in the near future, many users will come to inquire about related issues, and we can only tell them that this has no impact. The crash log is useless for us.

SergeTupchiy: Thanks @zhongwencool,
Fixed. Actually it crashed with a `badmatch` error, so catching the error is not absolutely required :
```
2023-08-02T06:53:08.917759+00:00 [error] Generic server emqx_router_helper terminating. Reason: {{badmatch,{'EXIT',{aborted,{no_exists,emqx_routing_node}}}},
```
However, I covered and logged both exceptions and error tuples, as there is rather no strict return type guarantee. 
For example, Mnesia dirty operation may exit literally or return `{'EXIT', Reason}` tuple. 
On a replicant side, exit exceptions are not expected as they are normally caught and wrapped by rpc lib, but I'm not sure it won't change in future...

zmstone (approved):

apps/emqx/src/emqx_router_sup.erl
SergeTupchiy: After a discussion with @zmstone, we agreed to do router_helper cleanup only on core nodes, as doing it on a replicant is anyway redundant.
Updated the PR.

zhongwencool (approved):

PR #11401, opened 2023-08-07
fix: rule SQL mongo_date function should return a string in test mode

kjellwinblad:
The rule SQL mongo_date function should return a string with the format ISODate(*), where * is an ISO date string when running the rule in test mode.

This PR ports the functionality from https://github.com/emqx/emqx-enterprise/pull/1835 so that the when the user test SQL statements that call mongo_date the result will be formatet in the same way as in EMQX 4.4.

Fixes:
https://emqx.atlassian.net/browse/EMQX-10727

### <samp>🤖 Generated by Copilot at 3ed031d</samp>

Added support for ISODate format in mongo_date function for rule engine tests. Updated emqx_rule_engine version and added a test case to verify the new feature. Modified `emqx_rule_funcs.erl`, `emqx_rule_sqltester.erl`, `emqx_rule_engine.app.src`, and `emqx_rule_engine_api_rule_test_SUITE.erl`.

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_funcs.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_funcs.erl
@@ -1193,7 +1193,7 @@ mongo_date(MillisecondsTimestamp) ->
 mongo_date(Timestamp, Unit) ->
     InsertedTimeUnit = time_unit(Unit),
     ScaledEpoch = erlang:convert_time_unit(Timestamp, InsertedTimeUnit, millisecond),
-    convert_timestamp(ScaledEpoch).
+    mongo_date(ScaledEpoch).
 
 maybe_isodate_format(ErlTimestamp) ->
     case emqx_rule_sqltester:is_test_runtime_env() of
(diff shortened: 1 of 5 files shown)


apps/emqx_rule_engine/src/emqx_rule_funcs.erl
```diff
@@ -1185,16 +1185,31 @@ function_literal(Fun, Args) ->
...
+    maybe_isodate_format(erlang:timestamp()).
 
 mongo_date(MillisecondsTimestamp) ->
-    convert_timestamp(MillisecondsTimestamp).
+    maybe_isodate_format(convert_timestamp(MillisecondsTimestamp)).
 
 mongo_date(Timestamp, Unit) ->
     InsertedTimeUnit = time_unit(Unit),
     ScaledEpoch = erlang:convert_time_unit(Timestamp, InsertedTimeUnit, millisecond),
     convert_timestamp(ScaledEpoch).
```
thalesmg, line 1196: Q: doesn't this one need to be adapted as well?

thalesmg (approved):

apps/emqx_rule_engine/src/emqx_rule_funcs.erl
terry-xiaoyu: We do need to convert this when displaying it in sql-test.

kjellwinblad: Thanks, will fix of course!

thalesmg (approved):

PR #11424, opened 2023-08-10
fix(datetime): make sure the epoch is not larger than the maximum supported value

lafirest:
A user sets a large timestamp by API to the banned feature, the value of this timestamp is out of the supported range and will crash the list API of the banned feature, and meanwhile, there is no `clean_all` API in this feature, thus the user can't delete the bad date. 

There will be a series PRs to unify the `emqx_datetime` and `emqx_calender`

Fixes <issue-or-jira-number>

### <samp>🤖 Generated by Copilot at 526c61c</samp>

Export and implement `epoch_to_rfc3339` functions in `emqx_calendar` module and improve `emqx_datetime` module to handle maximum epoch timestamp. These changes are for better compatibility and validation of date and time formats.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_datetime.erl
+++ b/apps/emqx/src/emqx_datetime.erl
@@ -38,6 +38,12 @@
 -typerefl_from_string({epoch_second/0, ?MODULE, to_epoch_second}).
 -typerefl_from_string({epoch_millisecond/0, ?MODULE, to_epoch_millisecond}).
 
+%% the maximum value is the SECONDS_FROM_0_TO_10000 in the calendar.erl,
+%% here minus SECONDS_PER_DAY to tolerate timezone time offset,
+%% so the maximum date can reach 9999-12-31 which is ample.
+-define(MAXIMUM_EPOCH, 253402214400).
+-define(MAXIMUM_EPOCH_MILLI, 253402214400_000).
+
 to_epoch_second(DateTime) ->
     to_epoch(DateTime, second).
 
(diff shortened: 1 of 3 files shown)


apps/emqx/src/emqx_calendar.erl
```diff
@@ -32,6 +32,11 @@
     offset_second/1
 ]).
 
+-export([
+    epoch_to_rfc3339/1,
+    epoch_to_rfc3339/2
+]).
```
thalesmg, line 38: These new functions were introduced, but they are not used anywhere?

apps/emqx/src/emqx_datetime.erl
```diff
@@ -47,7 +52,7 @@ to_epoch_millisecond(DateTime) ->
 to_epoch(DateTime, Unit) ->
     try
         case string:to_integer(DateTime) of
-            {Epoch, []} when Epoch >= 0 -> {ok, Epoch};
+            {Epoch, []} when Epoch >= 0, Epoch =< ?MAXIMUM_EPOCH -> {ok, Epoch};
```
zmstone, line 55: have to take Unit into consideration ?

zmstone: ah sorry, it's datetime convert to epoch.

lafirest: the is a Unix timestamp, so the unit is `second` and it also is the default unit for `calendar:system_time_to_rfc3999/1`, even if misused with other units when calling `calendar:system_time_to_rfc3999/1`,  the value only becomes smaller and safer

zmstone: Sorry again.  Help me understand this.

If this is a string format of epoch on 9999-12-31, it would be `"253402214400"` which is valid.
If it's at millisecond level, it would be `"253402214400000"` which is invalid ?

zmstone: It seems `epoch_millisecond()` type is used in quite a few hocon schema fields.

zmstone: so there should be a `-define(MAXIMUM_EPOCH_MILLI, 253402214400_000).` and check the unit here.

lafirest: Yes, you are right, the unit is important because both the second and millisecond call this function

zmstone (approved):

PR #11445, opened 2023-08-15
fix: remove os_mon application in Windows release

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-10738

## Summary
copilot:summary

## PR Checklist


apps/emqx/src/emqx.app.src
```diff
@@ -14,7 +14,6 @@
         esockd,
         cowboy,
         sasl,
-        os_mon,
```
lafirest, line 17: Remove this may cause this application absent from the release package, I am not sure.

zhongwencool: I't ok, `os_mon` has already added in the root source: `reboot_lists.term` file.
It will also add to the release.

if we put `os_mon` in emqx.app.src, it will start as permanent, It will crash emqx application(then the whole vm is crash)  when os_mon's crash.
So I start `os_mon` application as temporarily in `emqx_os_mon` process's init.

lafirest (approved):

apps/emqx/src/emqx.app.src
thalesmg: Alternatively, it could be moved to `included_applications`, so it's loaded but not started by the application controller.

apps/emqx/src/emqx_os_mon.erl
```diff
@@ -83,12 +82,17 @@ current_sysmem_percent() ->
 %%--------------------------------------------------------------------
 
 init([]) ->
+    %% start os_mon temporarily
+    {ok, _} = application:ensure_all_started(os_mon),
```
thalesmg, line 86: Q: since this process is started under `emqx_sys_sup` and that in turn is under `emqx` supervision tree, could the application controller enter a dead lock trying to start another application here while itself is starting `emqx`?
Perhaps it would be safer to attempt to start `os_mon` in `handle_continue`?

thalesmg (approved):

apps/emqx/src/emqx_os_mon.erl
zhongwencool: We don't deal with included_applications in emqx_machine.  :(
only use applications and reboot_list.term to get deps digraph, 
So I don't want to add this to `include_applications` to mislead it.

PR #11466, opened 2023-08-17
fix: hocon converter not working when make_serializable is true

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-10790

Fixed a crash that occurred when setting the ssl_options.ciphers configuration option to an empty string (""). The SSL connection handler was not properly checking for empty cipher lists before attempting to use them, which resulted in exceptions being thrown. The code has been updated to validate that cipher lists are non-empty before use, avoiding these exceptions and resulting crashes. This ensures the empty string value for `ssl_options.ciphers` now works as intended.

image

### <samp>🤖 Generated by Copilot at ef431e3</samp>

This pull request improves the conversion and validation functions for some configuration fields in the `emqx_bridge_gcp_pubsub` and `emqx_prometheus` applications. It also adds a new test case for the `emqx_listeners_update_SUITE` module to check the handling of empty ssl options.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_metrics_worker.erl
+++ b/apps/emqx/src/emqx_metrics_worker.erl
@@ -495,7 +495,7 @@ terminate(_Reason, #state{metric_ids = MIDs}) ->
 
 stop(Name) ->
     try
-        gen_server:stop(Name)
+        gen_server:stop(Name, normal, 10000)
     catch
         exit:noproc ->
             ok;
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -2314,18 +2314,7 @@ ciphers_schema(Default) ->
         hoconsc:array(string()),
         #{
             default => default_ciphers(Default),
-            converter => fun
-                (undefined) ->
-                    [];
-                (<<>>) ->
-                    [];
-                ("") ->
-                    [];
-                (Ciphers) when is_binary(Ciphers) ->
-                    binary:split(Ciphers, <<",">>, [global]);
-                (Ciphers) when is_list(Ciphers) ->
-                    Ciphers
-            end,
+            converter => fun converter_ciphers/2,
             validator =>
                 case Default =:= quic of
                     %% quic has openssl statically linked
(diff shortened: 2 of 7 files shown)


apps/emqx/src/emqx_metrics_worker.erl
```diff
@@ -495,7 +495,7 @@ terminate(_Reason, #state{metric_ids = MIDs}) ->
 
 stop(Name) ->
     try
-        gen_server:stop(Name)
+        gen_server:stop(Name, normal, 10000)
```
zhongwencool, line 498: Is only use by SUITE, we need a timeout instead of infinity waiting

apps/emqx/src/emqx_schema.erl
```diff
@@ -2316,18 +2316,7 @@ ciphers_schema(Default) ->
...
-                    [];
-                (<<>>) ->
-                    [];
-                ("") ->
-                    [];
-                (Ciphers) when is_binary(Ciphers) ->
-                    binary:split(Ciphers, <<",">>, [global]);
-                (Ciphers) when is_list(Ciphers) ->
-                    Ciphers
-            end,
```
zhongwencool, line 2330: I replace all fun/1 with fun/2 in converter, because of hocon don't run converter fun when make_serializable is true( check_plain).
https://github.com/emqx/hocon/blob/master/src/hocon_tconf.erl#L1247-L1261
```
eval_arity1_converter(F, V, Opts) ->
    case is_make_serializable(Opts) of
        true ->
            %% the old version convert is only one way
            %% we can only hope V is serializable
            V;
        false ->
            F(V)
    end.
```

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub.erl
```diff
@@ -398,7 +398,9 @@ service_account_json_validator(Map) ->
             {error, #{missing_keys => MissingKeys}}
     end.
 
-service_account_json_converter(Map) when is_map(Map) ->
+service_account_json_converter(Val, #{make_serializable := true}) ->
+    Val;
+service_account_json_converter(Map, _Opts) when is_map(Map) ->
     ExpectedKeys = [
```
zhongwencool, line 404: @thalesmg Could you please take a look at this ? 
While the behavior of the code after changing it remains the same as before, 
we want to make sure our expected understanding is correct.

I replace all fun/1 with fun/2 in converter, because of hocon don't run converter fun when make_serializable is true( check_plain).
https://github.com/emqx/hocon/blob/master/src/hocon_tconf.erl#L1247-L1261

HJianBo (approved):

PR #11396, opened 2023-08-04
perf(ruleeng): employ `emqx_topic_index` to speed up topic matching

keynslug:
Fixes [EMQX-10712](https://emqx.atlassian.net/browse/EMQX-10712).

Follows up #11332, #11282, #11313 .

### <samp>🤖 Generated by Copilot at 674847b</samp>

This pull request introduces a new module `emqx_topic_index` that provides a fast and efficient way to match topics to topic filters for the rule engine. It also refactors the rule creation, update, and deletion logic in the `emqx_rule_engine` module to use the topic index and a common validation function. Additionally, it adds a test suite for the topic index module and a macro for the topic index table name.

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [x] Added tests for the changes
- [ ] ~~Added property-based tests for code which performs user input validation~~
- [x] Changed lines covered in coverage report
- [x] Change log has been added to `changes/(ce|ee)/(feat|perf|fix)-<PR-id>.en.md` files
- [x] For internal contributor: there is a jira ticket to track this change
- [ ] ~~Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket~~
- [x] Schema changes are backward compatible

PR #11495, opened 2023-08-22
0817 improve topic-index search

zmstone:
This PR is an attempt to further optimize topic-index search performance on https://github.com/emqx/emqx/pull/11396
Main changes are:
1. Optimized the trie-search algorithm to avoid repeated prefix matches
3. Made an abstraction of the trie-search algorithm. 
4. Implemented a `gb_trees` based index, which might be more suitable for rule-engine than `ets` based.

### <samp>🤖 Generated by Copilot at 70bbed4</samp>

This pull request refactors the topic index modules and tests to use a common trie-search logic and adds a new topic index implementation using gb_trees and persistent_term. The purpose is to improve code reuse, consistency, readability, and performance of topic matching. The affected files are `emqx_topic_index.erl`, `emqx_topic_index_SUITE.erl`, `emqx_topic_gbt.erl`, `emqx_trie_search.erl`, and `emqx_cluster_rpc.erl`.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_trie_search.erl
+++ b/apps/emqx/src/emqx_trie_search.erl
@@ -210,31 +210,35 @@ search_new(Words0, NewBase, NextF, Acc) ->
 
 %% Search to the bigger end of ordered collection of topics and topic-filters.
 search_up(Words, {Filter, _} = Cursor, NextF, Acc) ->
-    case compare(Filter, Words, false) of
+    case compare(Filter, Words, []) of
         match_full ->
             search_new(Words, Cursor, NextF, match_add(Cursor, Acc));
         match_prefix ->
             search_new(Words, Cursor, NextF, Acc);
         lower ->
             {Cursor, Acc};
-        [SeekWord | FilterTail] ->
+        UpFromPrefix ->
             % NOTE
             % This is a seek instruction.
             % If we visualize the `Filter` as `FilterHead ++ [_] ++ FilterTail`, we need to
             % seek to `FilterHead ++ [SeekWord]`. It carries the `FilterTail` because it's
             % much cheaper to return it from `compare/3` than anything more usable.
-            NewBase = base(seek(SeekWord, Filter, FilterTail)),
-            search_new(Words, NewBase, NextF, Acc)
+            search_new(Words, base(UpFromPrefix), NextF, Acc)
     end.
 
-seek(SeekWord, [_ | FilterTail], FilterTail) ->
-    [SeekWord];
-seek(SeekWord, [FilterWord | Rest], FilterTail) ->
-    [FilterWord | seek(SeekWord, Rest, FilterTail)].
-
 compare(NotFilter, _, _) when is_binary(NotFilter) ->
+    % All non-wildcards (topics) are sorted higher than wildcards
     lower;
-compare([], [], _) ->
+compare(['#'], _Words, _RPrefix) ->
+    % NOTE
+    %  Topic: a/b/c/d
+    % Filter: a/+/+/d/#
+    % We matched the topic to a topic filter with wildcard (possibly with pluses).
+    % We include it in the result set, and now need to try next entry in the table.
+    % Closest possible next entries that we must not miss:
+    % * a/+/+/d/# (same topic but a different ID)
+    match_full;
+compare([], [], _RPrefix) ->
     % NOTE
     %  Topic: a/b/c/d
     % Filter: a/+/+/d
@@ -244,7 +248,7 @@ compare([], [], _) ->
     % * a/+/+/d (same topic but a different ID)
     % * a/+/+/d/# (also a match)
     match_full;
-compare([], _Words, _) ->
+compare([], _Words, _RPrefix) ->
     % NOTE
     %  Topic: a/b/c/d
     % Filter: a/+/c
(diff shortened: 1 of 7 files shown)


apps/emqx/src/emqx_trie_search.erl
```diff
@@ -0,0 +1,344 @@
...
+%% The recursive entrypoint of the trie-search algorithm.
+%% Always start from the initial prefix and words.
+search_new(#ctx{prefix0 = Prefix, words0 = Words0} = C, NewBase, Acc0) ->
+    #acc{target = {Filter, _}} = Acc = move_up(C, Acc0, NewBase),
+    case Prefix of
+        [] ->
+            %% This is not a '$' topic, start from '+'
+            search_plus(C, Words0, Filter, [], Acc);
+        [DollarWord] ->
+            %% Start from the '$' word
```
thalesmg, line 245: Q: why are dollar-prefixed words treated specially compared to ordinary words?

thalesmg: It seems to be to avoid matching `+` and `#` according to another comment, but it _seems_ like it would be possible for this to still reach `search_plus` (not sure of an example offhand). :thinking:

zmstone: right, `search_plus` is effectively `search_up` in this case, will fix it. (and fix comment too).

zmstone: sorry, it won't reach `search_plus` on the first word.

e.g. `$SYS/a`
It would immediately compare `$SYS` to the first word in the filter.
however:
1. `compare` function only matches the *next* `#` (special case to ensure `a` matches `a/#`).
2. `$SYS` does not equal to '+'
So the compare result can never be `shorter` if the first word of the filter is `#` or `+`

zmstone: actually the code may never reach here because the initial base already skipped over the +/.. and # filters

apps/emqx/src/emqx_trie_search.erl
```diff
@@ -0,0 +1,312 @@
...
+search_new(#ctx{words0 = Words} = C, NewBase, Acc) ->
+    case move_up(C, NewBase) of
+        ?END ->
+            {?END, Acc};
+        {Filter, _} = T ->
+            search_plus(C, Words, Filter, [], T, Acc)
+    end.
+
+%% Try to use '+' as the next word in the prefix.
+search_plus(C, [W, X | Words], [W, X | Filter], RPrefix, T, Acc) ->
```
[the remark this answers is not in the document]
thalesmg: Also: if we do encounter the situation `[W, X | Words], [W, X | Filter]`, couldn't we fast-forward both `W` and `X` in one go?

apps/emqx/src/emqx_trie_search.erl
```diff
@@ -0,0 +1,312 @@
...
+word(<<"+">>) -> '+';
+word(<<"#">>) -> '#';
+word(Bin) -> Bin.
+
+%% match non-wildcard topics
+match_topics(#ctx{nextf = NextF} = C, Topic, {Topic, _} = Key, Acc) ->
+    %% found a topic match
+    match_topics(C, Topic, NextF(Key), add(C, Acc, Key));
+match_topics(#ctx{nextf = NextF} = C, Topic, {F, _}, Acc) when F < Topic ->
+    %% the last key is a filter, try jump to the topic
```
thalesmg, line 308: Q: since we're searching for _topic filters_, shouldn't we also add `F` to the result set as a valid match?

apps/emqx/src/emqx_trie_search.erl
```diff
@@ -0,0 +1,312 @@
...
+    search_up(C, W, Words, Filter, RPrefix, T, Acc).
+
+%% Search to the bigger end of ordered collection of topics and topic-filters.
+search_up(C, Word, Words, Filter, RPrefix, T, Acc) ->
+    case compare(Word, Filter, Words) of
+        {match, full} ->
+            search_new(C, T, add(C, Acc, T));
+        {match, prefix} ->
+            search_new(C, T, Acc);
+        lower ->
```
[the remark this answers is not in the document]
zmstone: this part is now re-rewritten

apps/emqx/src/emqx_trie_search.erl
```diff
@@ -0,0 +1,312 @@
...
+
+%% @doc Match given topic against the index and return _all_ matches.
+%% If `unique` option is given, return only unique matches by record ID.
+-spec matches(emqx_types:topic(), nextf(), opts()) -> [key(_)].
+matches(Topic, NextF, Opts) ->
+    search(Topic, NextF, Opts).
+
+%% @doc Entrypoint of the search for a given topic.
+search(Topic, NextF, Opts) ->
+    Words = words(Topic),
```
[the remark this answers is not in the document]
zmstone: The topic validation should have been done at higher level. e.g. when parsing MQTT packet.
Doing it again here is a waste. 
Also, in case this code is executed by a single process or a pool of processes, it's less scalable to put the validations at this low level.

apps/emqx/src/emqx_trie_search.erl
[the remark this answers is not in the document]
zmstone: good point.
now this part is re-written though.

apps/emqx/src/emqx_trie_search.erl
zmstone: Here, we have done searching for filters (wildcards), but start searching for non-wildcads.
If the last is a match, it should have already been added to Acc.

apps/emqx/src/emqx_trie_search.erl
```diff
@@ -0,0 +1,312 @@
...
+        higher ->
+            NewBase = base(lists:reverse([Word | RPrefix])),
+            search_new(C, NewBase, Acc);
+        shorter ->
+            search_plus(C, Words, tl(Filter), [Word | RPrefix], T, Acc)
+    end.
+
+%% Compare prefix word then the next words in suffix against the search-target
+%% topic or topic-filter.
+compare(_, NotFilter, _) when is_binary(NotFilter) ->
```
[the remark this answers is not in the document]
zmstone: Tried to document the new compare/3 function.

apps/emqx/src/emqx_trie_search.erl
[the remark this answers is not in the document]
zmstone: Sorry, since we are already traversing the words after split, it's not that expensive to add an assertion. Will add it.

zmstone:
moved to: https://github.com/emqx/emqx/pull/11517

PR #11517, opened 2023-08-24
feat(topicidx): iterate on trie search implementation

keynslug:
This is an attempt to iterate on #11495's impressive results, slightly improving performance and (arguably) simplicity. The gist of the search algorithm is essentially the same.

Synthetic benchmarks show that:
* Matching performance significantly better than `release-52` baseline on every workload.
* Matching performance slightly better than 62423b0b12deac9641152a5ff01236660fbf3746, between 5 and 15% depending on workload.
* GC pressure noticeably better than 62423b0b12deac9641152a5ff01236660fbf3746, spends ~25% less heap in [`fixed_15k` workload](https://gist.github.com/keynslug/c0393ab5189417df46dd48f745c90600#file-emqx_topicidx_bench-erl).

### <samp>🤖 Generated by Copilot at cf45e80</samp>

Refactor topic index module to use gb_trees and persistent_term, add generic trie-search algorithm, fix typo, and add dependency.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_trie_search.erl
+++ b/apps/emqx/src/emqx_trie_search.erl
@@ -137,7 +137,7 @@ get_topic({Filter, _ID}) when is_list(Filter) ->
 get_topic({Topic, _ID}) ->
     Topic.
 
--compile({inline, [base/1, move_up/2, match_add/2, compare/3]}).
+-compile({inline, [base/1, move_up/2, match_add/2, compare/4]}).
 
 %% Make the base-key which can be used to locate the desired search target.
 base(Prefix) ->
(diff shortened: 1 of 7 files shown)


apps/emqx/src/emqx_trie_search.erl
```diff
@@ -0,0 +1,349 @@
...
+    % * a/+/c/# (which is a match)
+    % * a/+/c/+ (also a match)
+    %
+    % TODO
+    % We might probably instead seek to a/+/c/# right away.
+    match_prefix;
+compare(['#'], _Words, _, _) ->
+    % NOTE
+    %  Topic: a/b/c/d
+    % Filter: a/+/+/d/#
```
zmstone, line 261: nit: maybe also mention `Filter: a/#`

apps/emqx/src/emqx_trie_search.erl
```diff
@@ -0,0 +1,349 @@
...
+    %  Topic: a/b/c/d
+    % Filter: a/+/c
+    % We found out that a topic filter is a prefix of the topic (possibly with pluses).
+    % We discard it, and now need to try next entry in the table.
+    % Closest possible next entries that we must not miss:
+    % * a/+/c/# (which is a match)
+    % * a/+/c/+ (also a match)
+    %
+    % TODO
+    % We might probably instead seek to a/+/c/# right away.
```
zmstone, line 256: the immediate next (if any) should be `a/+/c/#`, so there is no need to append another word which builds a new list.

keynslug:
Should now be even better performance-wise with all @zmstone's great insights.

apps/emqx/src/emqx_trie_search.erl
keynslug: There may also be a `a/+/c` but with a different ID. New list sounds expensive indeed, it seems possible to seek to `{[a, +, c], []}` instead (since any list compares greater than any tuple).

apps/emqx/src/emqx_trie_search.erl
keynslug: Did in fdd2a419639393901a082c2883504b95ae37f43a.

zmstone (approved):

apps/emqx/src/emqx_trie_search.erl
```diff
@@ -0,0 +1,346 @@
...
+        lower ->
+            % NOTE
+            %  Topic: a/b/c/d
+            % Filter: a/+/+/e/1 or a/b/+/d/1
+            % The topic is lower than a topic filter. But we're at the `+` position,
+            % so we emit a backtrack point to seek to:
+            % Seek: {2, c}
+            % We skip over part of search space, and seek to the next possible match:
+            % Next: a/+/c
+            {Pos, HW};
```
thalesmg, line 275: Q: wouldn't the next possible match in this case be the current backtrack result (i.e.: `a/+/c`) padded to the right with `+`'s (`a/+/c/+` in this example)?

rebar.config.erl
```diff
@@ -190,7 +190,8 @@ test_deps() ->
         {meck, "0.9.2"},
         {proper, "1.4.0"},
         {er_coap_client, {git, "https://github.com/emqx/er_coap_client", {tag, "v1.0.5"}}},
-        {erl_csv, "0.2.0"}
+        {erl_csv, "0.2.0"},
+        {eministat, "0.10.1"}
```
thalesmg, line 194: This doesn't seem to be used?

thalesmg: I just noticed it's used in the gist linked in the description.

thalesmg (approved):
:hot_pepper:

lafirest (approved):

apps/emqx/src/emqx_trie_search.erl
keynslug: We would miss `a/+/c/#` but in general it's great idea, we could indeed shave off few `NextF`s here or there by seeking _past_ some prefix. I'll try to experiment with that in a followup PR.

PR #11520, opened 2023-08-25
fix: packets_connack_sent is not incremented if the ack_flag is non-zero

zhongwencool:
Fixes: https://emqx.atlassian.net/browse/EMQX-10840

### <samp>🤖 Generated by Copilot at 28fd8d3</samp>

This pull request refactors the connection management logic from the `emqx_cm` module to the `emqx_channel` module, and updates the metrics and tests for the new format of the `CONNACK` packet. It also removes some unused functions and improves the performance of some channel operations by inlining functions.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -35,8 +35,6 @@
     insert_channel_info/3
 ]).
 
--export([connection_closed/1]).
-
 -export([
     get_chan_info/1,
     get_chan_info/2,
@@ -191,14 +189,6 @@ do_unregister_channel({_ClientId, ChanPid} = Chan) ->
     ets:delete_object(?CHAN_TAB, Chan),
     ok = emqx_hooks:run('channel.unregistered', [ChanPid]),
     true.
-
--spec connection_closed(emqx_types:clientid()) -> true.
-connection_closed(ClientId) ->
-    connection_closed(ClientId, self()).
-
--spec connection_closed(emqx_types:clientid(), chan_pid()) -> true.
-connection_closed(ClientId, ChanPid) ->
-    ets:delete_object(?CHAN_CONN_TAB, {ClientId, ChanPid}).
 
 %% @doc Get info of a channel.
 -spec get_chan_info(emqx_types:clientid()) -> maybe(emqx_types:infos()).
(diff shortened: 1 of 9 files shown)


apps/emqx/src/emqx_cm.erl
```diff
@@ -192,14 +190,6 @@ do_unregister_channel({_ClientId, ChanPid} = Chan) ->
     ok = emqx_hooks:run('channel.unregistered', [ChanPid]),
     true.
 
--spec connection_closed(emqx_types:clientid()) -> true.
-connection_closed(ClientId) ->
-    connection_closed(ClientId, self()).
-
--spec connection_closed(emqx_types:clientid(), chan_pid()) -> true.
-connection_closed(ClientId, ChanPid) ->
-    ets:delete_object(?CHAN_CONN_TAB, {ClientId, ChanPid}).
```
zhongwencool, line 201: - `?CHAN_CONN_TAB` table store `{{ClientId, ChanPid, Mod}`, so ets:delete_object do anything here.
- we have already deleted {clientid, ChanPid} in emqx_cm process after ChanPid shutdown here:
https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_cm.erl#L686-L692
So we don't need this connection_closed/1 function now.

apps/emqx/src/emqx_metrics.erl
```diff
@@ -493,7 +493,7 @@ inc_sent(Packet) ->
     inc('packets.sent'),
     do_inc_sent(Packet).
 
-do_inc_sent(?CONNACK_PACKET(ReasonCode)) ->
```
zhongwencool, line 496:
```erlang
-define(CONNACK_PACKET(ReasonCode), #mqtt_packet{
    header = #mqtt_packet_header{type = ?CONNACK},
    variable = #mqtt_packet_connack{
        ack_flags = 0,
        reason_code = ReasonCode
    }
}).
```
`?CONNACK_PACKET/1` lock the ack_flags to 0.

apps/emqx/src/emqx_cm.erl
zmstone: I remember at least the function was left unchanged previously by @savonarola 
but I cannot recall the discussion about why delete_object is still here.

savonarola: I added a clarification in 4.4
https://github.com/emqx/emqx/blob/main-v4.4/src/emqx_cm.erl#L173

>     %% We can't clean CHAN_CONN_TAB because records for dead connections
>     %% are required for `get_chann_conn_mod/1` function, and `get_chann_conn_mod/1`
>     %% is used for takeover.

I left the fun empty to avoid attempts to "fix" the things while we have the current mechanism of keeping offline sessions.

PR #11522, opened 2023-08-25
fix: bad error message when rule engine schema name is too long

kjellwinblad:
Fixes https://emqx.atlassian.net/browse/EMQX-10778

### <samp>🤖 Generated by Copilot at c65db82</samp>

Improved schema name validation and bumped up the application version. The changes prevent potential atom exhaustion and reflect the code updates in `emqx_schema_registry`.

## PR Checklist

- [  Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket

### DIFF
--- a/apps/emqx_schema_registry/src/emqx_schema_registry.erl
+++ b/apps/emqx_schema_registry/src/emqx_schema_registry.erl
@@ -64,7 +64,7 @@ get_serde(SchemaName) ->
 get_schema(SchemaName) ->
     case
         emqx_config:get(
-            [?CONF_KEY_ROOT, schemas, binary_to_atom(SchemaName)], undefined
+            [?CONF_KEY_ROOT, schemas, schema_name_bin_to_atom(SchemaName)], undefined
         )
     of
         undefined ->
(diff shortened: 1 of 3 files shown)


kjellwinblad:
I don't think a test case is necessary for this as it does not fix a critical bug. It just slightly improve the error message in an unusual situation when the user types an extremely long schema name. I have tested it manually.

apps/emqx_schema_registry/src/emqx_schema_registry.erl
```diff
@@ -333,6 +333,20 @@ async_delete_serdes(Names) ->
...
+    erlang:throw(
+        iolist_to_binary(
+            io_lib:format(
+                "Name is is too long."
+                " Please provide a shorter name (<= 255 bytes)."
+                " The name that is too long: \"~s\"",
+                [Bin]
+            )
+        )
+    );
```
zmstone, line 346: should be a part of the config schema?

kjellwinblad: This does not work when the name is a key. At least, this seems like the same problem as we had here https://github.com/emqx/emqx/pull/10911 and then we decided to make a simple fix until HOCON has support for this. Please correct me if I'm wrong.

thalesmg: Maybe we could check for the name length closer to the API boundary, like validating the input in the API handlers before attempting to do any operations.

kjellwinblad: @thalesmg `schema_registry:get_schema/1` is called from several API endpoints in `emqx_schema_registry_http_api`. We could of course do the check in all these places instead but I don't see that this makes things better. I would be happy to be convinced why this would be better. My thinking was to do the check as close as possible to where the error happened to make sure that all call sites where covered.

thalesmg: From what I can see, we would need to change only 2 places in `emqx_schema_registry_http_api`: the check in `'/schema_registry'(post, _)` and `param_path_schema_name`.  

The later is the type of the `:name` binding, and adding a validator there would add the check to all `/schema_registry/:name` handlers.  

Then, calls to the internal `get_schema` could assume that input was validated by the upper level.

thalesmg (approved):

PR #11533, opened 2023-08-28
feat(ds): close iterators when handling `UNSUBSCRIBE` packets

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-9742

### <samp>🤖 Generated by Copilot at aaa9a07</samp>

This pull request adds and improves the functionality of deleting subscriptions from the persistent session data store, which is part of the emqx durable storage layer. It modifies the `emqx_ds.erl`, `emqx_persistent_session_ds.erl`, `emqx_session.erl`, and `emqx_persistent_session_ds_proto_v1.erl` modules, and updates the `emqx_ds_SUITE.erl` integration test suite and the `Makefile`. The changes aim to handle the unsubscription scenario and ensure the consistency and cleanliness of the session data.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -146,16 +146,11 @@ del_subscription(IteratorID, TopicFilterBin, DSSessionID) ->
         begin
             TopicFilter = emqx_topic:words(TopicFilterBin),
             Ctx = #{iterator_id => IteratorID},
-            case IteratorID of
-                undefined ->
-                    ok;
-                _ ->
-                    ?tp_span(
-                        persistent_session_ds_close_iterators,
-                        Ctx,
-                        ok = ensure_iterator_closed_on_all_shards(IteratorID)
-                    )
-            end,
+            ?tp_span(
+                persistent_session_ds_close_iterators,
+                Ctx,
+                ok = ensure_iterator_closed_on_all_shards(IteratorID)
+            ),
             ?tp_span(
                 persistent_session_ds_iterator_delete,
                 Ctx,
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -185,8 +185,7 @@ ensure_all_iterators_closed(DSSessionID) ->
 
 %% RPC target.
 -spec do_ensure_all_iterators_closed(emqx_ds:session_id()) -> ok.
-do_ensure_all_iterators_closed(DSSessionID0) ->
-    DSSessionID = bin(DSSessionID0),
+do_ensure_all_iterators_closed(DSSessionID) ->
     ok = emqx_ds_storage_layer:discard_iterator_prefix(?DS_SHARD, DSSessionID),
     ok.
 
@@ -202,6 +201,3 @@ deserialize_message(Bin) ->
 
 is_store_enabled() ->
     emqx_config:get([persistent_session_store, ds]).
-
-bin(B) when is_binary(B) -> B;
-bin(A) when is_atom(A) -> atom_to_binary(A, utf8).
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -23,7 +23,8 @@
 -export([
     persist_message/1,
     open_session/1,
-    add_subscription/2
+    add_subscription/2,
+    del_subscription/3
 ]).
 
 -export([
@@ -32,7 +33,15 @@
 ]).
 
 %% RPC
--export([do_open_iterator/3]).
+-export([
+    ensure_iterator_closed_on_all_shards/1,
+    ensure_all_iterators_closed/1
+]).
+-export([
+    do_open_iterator/3,
+    do_ensure_iterator_closed/1,
+    do_ensure_all_iterators_closed/1
+]).
 
 %% FIXME
 -define(DS_SHARD, <<"local">>).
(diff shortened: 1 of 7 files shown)


apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -201,14 +202,35 @@ session_add_iterator(DSSessionId, TopicFilter) ->
...
+    IteratorRefId = {DSSessionId, TopicFilter},
+    {atomic, Res} =
+        mria:transaction(?DS_SHARD, fun() ->
+            case mnesia:read(?ITERATOR_REF_TAB, IteratorRefId, write) of
+                [] ->
+                    {ok, _WasSubscribed = false};
+                [#iterator_ref{}] ->
+                    ok = mnesia:delete(?ITERATOR_REF_TAB, IteratorRefId, write),
+                    {ok, _WasSubscribed = true}
+            end
```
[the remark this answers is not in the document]
ieQu1: I think @keynslug need to weigh on this, since write amplification is a real issue to be considered. Ideally we want to support clients that subscribe to large number of topics (in thousands), so it's a real issue.
For now, I think we can overlook this part and move on.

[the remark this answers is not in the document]
thalesmg: > TBH I don't see why it's important to respond with WasSubscribed flag,

I was trying to adhere the [previously defined typespec and doc](https://github.com/emqx/emqx/blob/8b3a9072f970fed8e2a79fba65ac5093d7637013/apps/emqx_durable_storage/src/emqx_ds.erl#L204-L208) for this function.  But, at this moment, it doesn't seem to be useful, indeed.

I'll then add a field to our `#session{}` record to keep track of the iterators, then.  That way, we can avoid the lookup before performing the multicall.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -130,6 +139,62 @@ do_open_iterator(TopicFilter, StartMS, IteratorID) ->
...
+    Nodes = emqx:running_nodes(),
+    Results = emqx_persistent_session_ds_proto_v1:close_all_iterators(Nodes, DSSessionID),
+    %% TODO: handle errors
+    true = lists:all(fun(Res) -> Res =:= {ok, ok} end, Results),
+    ok.
+
+%% RPC target.
+-spec do_ensure_all_iterators_closed(emqx_ds:session_id()) -> ok.
+do_ensure_all_iterators_closed(DSSessionID0) ->
+    DSSessionID = bin(DSSessionID0),
```
keynslug, line 194: Q: Is that indeed could be an atom? I'd expect only binaries as session IDs on that level.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -130,6 +139,62 @@ do_open_iterator(TopicFilter, StartMS, IteratorID) ->
...
 
+-spec del_subscription(emqx_ds:iterator_id() | undefined, emqx_types:topic(), emqx_ds:session_id()) ->
+    ok | {skipped, disabled}.
+del_subscription(IteratorID, TopicFilterBin, DSSessionID) ->
+    ?WHEN_ENABLED(
+        begin
+            TopicFilter = emqx_topic:words(TopicFilterBin),
+            Ctx = #{iterator_id => IteratorID},
+            case IteratorID of
+                undefined ->
```
keynslug, line 150: Nit. Looks unusual to handle `undefined` on that level. I'd move that to `emqx_session` level.

apps/emqx/src/emqx_session.erl
```diff
@@ -328,23 +335,32 @@ add_persistent_subscription(TopicFilterBin, ClientId, Session) ->
...
+            {ok, Session1#session{subscriptions = maps:remove(TopicFilter, Subs)}};
         error ->
             {error, ?RC_NO_SUBSCRIPTION_EXISTED}
     end.
 
+-spec remove_persistent_subscription(session(), emqx_types:topic(), emqx_types:clientid()) ->
+    session().
+remove_persistent_subscription(Session, TopicFilterBin, ClientId) ->
+    Iterators = Session#session.iterators,
+    IteratorId = maps:get(TopicFilterBin, Iterators, undefined),
```
keynslug, line 360: Q: Why not just `maps:get(TopicFilterBin, Iterators)`? Getting `undefined` here looks kinda impossible.

keynslug (commented):
👍🏼

apps/emqx/src/emqx_persistent_session_ds.erl
thalesmg: This was needed to [appease dialyzer](https://github.com/emqx/emqx/blob/14cf2d870aa845af49908768c60d6ceda595f608/apps/emqx/src/emqx_types.erl#L184)... :sweat:

apps/emqx/src/emqx_persistent_session_ds.erl
thalesmg: This was mostly to have idempotency for calls to this function: if this happens to be called with a session with this particular iterator id already removed (so, nothing to close), we still want to be sure to remove the iterator reference (that depends only the topic filter) from `?ITERATOR_REF_TAB`.

apps/emqx/src/emqx_session.erl
thalesmg: Indeed, I was thinking about idempotency as [here](https://github.com/emqx/emqx/pull/11533#discussion_r1315121515), in a scenario where the iterators are closed but the mnesia entry is not yet removed.

But, on second thought, even in such scenario I think that the ID would still be in the `#session{}` record at this point.

1. Midway while unsubscribing, we close/discard the iterators on all nodes, but the node goes down before `emqx_ds:session_del_iterator` is called.
2. The node is restarted, and the client reconnects.  It'll recreate the same iterators (but starting from `start_time` again) with the same ID.
3. If it tries to unsubscribe again, the iterator IDs will be in the session record, albeit they were reset.

I'll remove this default and the above clause.

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: I suspected that. 🫠 What do you think about defining `emqx_ds:session_id()` as just `binary()` so we don't have to write unreachable code here?

keynslug (approved):

qzhuyan (approved):

PR #11541, opened 2023-08-29
Support asynchronous mode of receiving responses to FT operations

savonarola:
Fixes [EEC-907](https://emqx.atlassian.net/browse/EEC-907)
EIP: https://github.com/emqx/eip/pull/80

* We add an option to send FT commands to `$file-async/...` topics. Then, success RC is returned on all kicked-off async operations (currently, only fin).
* We duplicate _final_ RC statuses to `$file-response/{ClientId}` topic in the form
```json
{
  "vsn": "0.1",
  "topic": "$file/[COMMAND]",
  "packet_id": 1,
  "reason_code": 0,
  "reason_description": "success"
}
```
* We also support [MQTT v5 request-response pattern](https://www.emqx.com/en/blog/mqtt5-request-response), that is, overriding topic for statuses with `Response-Topic` prop and sending `Correlation-Data` prop back.

### <samp>🤖 Generated by Copilot at 14bfa13</samp>

This pull request implements and tests a new feature for file transfer in the emqx_ft module, using a response topic and the emqx_ft_async_reply module to handle asynchronous replies from the storage workers. It also refactors the emqx_channel and emqx_cm modules to use macros and hooks, and fixes some syntax and naming issues. It adds, modifies, or deletes several files in the apps/emqx_ft and apps/emqx directories.

## PR Checklist

- [x] Created a follow-up jira ticket for docs [EMQX-10888](https://emqx.atlassian.net/browse/EMQX-10888), [EMQX-10889](https://emqx.atlassian.net/browse/EMQX-10889)

### DIFF
--- a/apps/emqx_ft/src/emqx_ft.erl
+++ b/apps/emqx_ft/src/emqx_ft.erl
@@ -90,6 +90,10 @@
 
 -define(FT_EVENT(EVENT), {?MODULE, EVENT}).
 
+-define(ACK_AND_PUBLISH(Result), {true, Result}).
+-define(ACK(Result), {false, Result}).
+-define(DELAY_ACK, delay).
+
 %%--------------------------------------------------------------------
 %% API for app
 %%--------------------------------------------------------------------
@@ -116,46 +120,34 @@ unhook() ->
 %% API
 %%--------------------------------------------------------------------
 
-decode_filemeta(Payload) when is_binary(Payload) ->
-    case emqx_utils_json:safe_decode(Payload, [return_maps]) of
-        {ok, Map} ->
-            decode_filemeta(Map);
-        {error, Error} ->
-            {error, {invalid_filemeta_json, Error}}
-    end;
-decode_filemeta(Map) when is_map(Map) ->
-    Schema = emqx_ft_schema:schema(filemeta),
-    try
-        Meta = hocon_tconf:check_plain(Schema, Map, #{atom_key => true, required => false}),
-        {ok, Meta}
-    catch
-        throw:{_Schema, Errors} ->
-            {error, {invalid_filemeta, Errors}}
-    end.
+decode_filemeta(Payload) ->
+    emqx_ft_schema:decode(filemeta, Payload).
 
 encode_filemeta(Meta = #{}) ->
-    Schema = emqx_ft_schema:schema(filemeta),
-    hocon_tconf:make_serializable(Schema, emqx_utils_maps:binary_key_map(Meta), #{}).
+    emqx_ft_schema:encode(filemeta, Meta).
+
+encode_response(Response) ->
+    emqx_ft_schema:encode(command_response, Response).
 
 %%--------------------------------------------------------------------
 %% Hooks
 %%--------------------------------------------------------------------
 
-on_message_publish(
-    Msg = #message{
-        id = _Id,
-        topic = <<"$file/", _/binary>>
-    }
-) ->
+on_message_publish(Msg = #message{topic = <<"$file-async/", _/binary>>}) ->
+    Headers = Msg#message.headers,
+    {stop, Msg#message{headers = Headers#{allow_publish => false}}};
+on_message_publish(Msg = #message{topic = <<"$file/", _/binary>>}) ->
     Headers = Msg#message.headers,
     {stop, Msg#message{headers = Headers#{allow_publish => false}}};
 on_message_publish(Msg) ->
     {ok, Msg}.
 
-on_message_puback(PacketId, #message{topic = Topic} = Msg, _PubRes, _RC) ->
+on_message_puback(PacketId, #message{from = From, topic = Topic} = Msg, _PubRes, _RC) ->
     case Topic of
-        <<"$file/", FileCommand/binary>> ->
-            {stop, on_file_command(PacketId, Msg, FileCommand)};
+        <<"$file/", _/binary>> ->
+            {stop, on_file_command(sync, From, PacketId, Msg, Topic)};
+        <<"$file-async/", _/binary>> ->
+            {stop, on_file_command(async, From, PacketId, Msg, Topic)};
         _ ->
             ignore
     end.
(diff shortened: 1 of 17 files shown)


apps/emqx_ft/include/emqx_ft.hrl
```diff
@@ -0,0 +1,22 @@
...
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+-ifndef(EMQX_FT_HRL).
+-define(EMQX_FT_HRL, true).
+
+-define(FT_RESPONSE_TOPIC, <<"$file/response">>).
```
keynslug, line 20: Few thoughts.
1. Wouldn't be more predictable / scalable for a client to subscribe to `$file/response/{{fileId}}`, so they won't be overwhelmed with feedback related to **all** transfers? Also should be more future-proof from the security standpoint.
2. Seems unfortunate that it collides with command topics space, might lead to some "messiness" (e.g. you can subscribe to `$file/response` but not to `$file/{{fileId}}`, you can still use `"response"` as fileId but feedback will look somewhat confusing). Maybe `$fileack/...` or something like that would be better?
3. Nit. It's not exactly a _responses_ that are going here (since they're entirely optional), maybe it's _status_ or _feedback_?

apps/emqx_ft/src/emqx_ft_schema.erl
```diff
@@ -282,6 +291,15 @@ schema(filemeta) ->
             {segments_ttl, hoconsc:mk(pos_integer())},
             {user_data, hoconsc:mk(json_value())}
         ]
+    };
+schema(command_response) ->
+    #{
+        roots => [
+            {vsn, hoconsc:mk(string(), #{default => <<"0.1.0">>})},
```
keynslug, line 298: Nit. I think it doesn't make much sense to use complex versioning here, e.g. `1.0` seems good enough.

apps/emqx_ft/src/emqx_ft_schema.erl
```diff
@@ -282,6 +291,15 @@ schema(filemeta) ->
...
+            {topic, hoconsc:mk(string())},
+            {packet_id, hoconsc:mk(pos_integer())},
+            {result_code, hoconsc:mk(non_neg_integer())}
```
keynslug, line 301: Given that we have response schema, it now seems natural to provide richer, file-transfer-specific feedback, and not just MQTT Reason Code. The latter reduces a lot of different situations to single _unspecified error_.

Admittedly, it needs some design work, so maybe this is followup PR's material.

keynslug (commented):
🔥

apps/emqx_ft/include/emqx_ft.hrl
savonarola: The first variant I implemented was actually `$file/{{fileId}}/response`, but this scheme doesn't work for malformed commands.
We decided to make the default response topic `$file/{{clientId}}/response` during the last discussion.
`{{fileId}}` is not included so that the client should not resubscribe on each file transfer.

apps/emqx_ft/src/emqx_ft_schema.erl
savonarola: ✅

apps/emqx_ft/src/emqx_ft.erl
```diff
@@ -116,65 +119,64 @@ unhook() ->
...
 %% Hooks
 %%--------------------------------------------------------------------
 
-on_message_publish(
-    Msg = #message{
-        id = _Id,
-        topic = <<"$file/", _/binary>>
-    }
-) ->
+on_message_publish(Msg = #message{topic = ?FT_RESPONSE_TOPIC}) ->
```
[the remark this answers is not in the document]
savonarola: Response topics may be overridden by a client, so it is problematic to forbid publishing consistently.

apps/emqx_ft/src/emqx_ft_schema.erl
savonarola: Yes, I planned to implement this after the PoC, done now.

apps/emqx_ft/include/emqx_ft.hrl
keynslug: > We decided to make the default response topic `$file/{{clientId}}/response` during the last discussion.
> `{{fileId}}` is not included so that the client should not resubscribe on each file transfer.

Good point. Although I'm still wondering: given the client can subscribe to a wildcard anytime, wouldn't it be strictly more flexible to publish to `$file/{{clientId}}/{{fileId}}/response` then?

apps/emqx_ft/src/emqx_ft.erl
```diff
@@ -89,6 +89,8 @@
 }.
 
 -define(FT_EVENT(EVENT), {?MODULE, EVENT}).
+-define(SEND_TO_TOPIC(Result), {true, Result}).
+-define(DONT_SEND_TO_TOPIC(Result), {false, Result}).
```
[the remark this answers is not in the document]
savonarola: 👍  + `undefined` as `DELAY_ACK`

apps/emqx_ft/test/emqx_ft_SUITE.erl
```diff
@@ -96,13 +105,18 @@ init_per_testcase(Case, Config) ->
     end.
 end_per_testcase(_Case, Config) ->
     _ = [ok = emqtt:stop(C) || {client, C} <- Config],
+    _ = emqx_ft_test_helpers:cleanup_ft_root(Config),
```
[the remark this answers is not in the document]
savonarola: >This is needed because of ClientID clashes, right?
In some tests, we have predefined ClientIDs

However, we may just make fr root folders unique.

id (approved):

PR #11574, opened 2023-09-07
feat(ds): add filter for message persistence

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10520

## Summary
copilot:summary

## PR Checklist

- [na] Change log has been added to `changes/(ce|ee)/(feat|perf|fix)-<PR-id>.en.md` files


ieQu1 (approved):

ieQu1 (approved):

apps/emqx/src/emqx_persistent_session_ds_router.erl
```diff
@@ -0,0 +1,230 @@
...
+%%--------------------------------------------------------------------
+%% Table Initialization
+%%--------------------------------------------------------------------
+
+init_tables() ->
+    mria_config:set_dirty_shard(?PS_ROUTER_SHARD, true),
+    ok = mria:create_table(?PS_ROUTER_TAB, [
+        {type, bag},
+        {rlog_shard, ?PS_ROUTER_SHARD},
+        {storage, ram_copies},
```
keynslug, line 55: Q: Shouldn't we preserve them on disk as well?

apps/emqx/src/emqx_persistent_session_ds_router.erl
```diff
@@ -0,0 +1,230 @@
...
+        false ->
+            mria_route_tab_delete(#ps_route{topic = Topic, dest = Dest})
+    end.
+
+%% @doc Take a real topic (not filter) as input, return the matching topics and topic
+%% filters associated with route destination.
+-spec match_routes(emqx_types:topic()) -> [emqx_types:route()].
+match_routes(Topic) when is_binary(Topic) ->
+    lookup_route_tab(Topic) ++
+        [match_to_route(M) || M <- match_filters(Topic)].
```
keynslug, line 109: There's potential for an optimization: currently, we don't need the list of matching routes, only the fact that there's at least one, and there's `emqx_topic_index:match/2` for that.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -98,6 +101,7 @@ open_session(ClientID) ->
 add_subscription(TopicFilterBin, DSSessionID) ->
     ?WHEN_ENABLED(
         begin
+            ok = emqx_persistent_session_ds_router:do_add_route(TopicFilterBin, DSSessionID),
```
keynslug, line 104: Q: This executes out of transaction context, right? If so, it seems there's possibility for an inconsistency: route is present in the persistent session router, but not in the session state. Admittedly, I might be missing something here.

Same question about `del_subscripton/2`.

keynslug (commented):
Sorry, I'm kinda late to the party. Good job! Got just a couple of questions.

apps/emqx/src/emqx_persistent_session_ds.erl
thalesmg: Indeed, subscriptions are (serially) processed by the connection process without any locking/transaction context.

I think I see what you mean.  Do you mean that the inconsistency might arise from the route is inserted into the PS routing table, and then the node crashes?  

Good point, perhaps it would be safer to first ensure the iterator is added to the session before adding the route to make it idempotent when restarting the subscription.

apps/emqx/src/emqx_persistent_session_ds_router.erl
thalesmg: True, I took what we have for the normal router, but it probably makes more sense to keep this in disk.  Perhaps we should set it to `rocksdb_copies` (or fallback to `disc_copies` for Windows).

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: > perhaps it would be safer to first ensure the iterator is added to the session before adding the route to make it idempotent when restarting the subscription.

Indeed, it should be better. But there's also a possibility that no client will ever try to restart the subscription again.

thalesmg:
Thanks for the review, @keynslug !  :beers: 

I've made a follow up PR to address those issues: https://github.com/emqx/emqx/pull/11598

apps/emqx/src/emqx_persistent_session_ds.erl
thalesmg: Yes, there's that possibility.

By writing to the iterator reference table first, inside a transaction, I was thinking that it could cover that scenario when the client connects again.  My idea was that the client, upon reconnecting, would have its session data loaded, and when encountering that topic associated with its DS session ID, it would also call `add_subscription` to repeat the process. 

Also, we're currently using dirty operations for writing the routes, so there's no use in calling it inside `emqx_ds:session_add_iterator`'s transaction, for example.

Do you see a way to make this more consistent?  :thinking:

thalesmg: >it would also call add_subscription to repeat the process.

Actually, I was thinking of `emqx_session:subscribe`.  On second thought, I wonder if this is redundant/overkill...  :thinking:

keynslug: Good points.

Under the assumption that the client will eventually attempt to repeat any failed operation until successful completion, we indeed should get away with simple sequence of writes that's only idempotent (but not transactional). But it's very strong assumption that probably won't hold in practice.

> Do you see a way to make this more consistent? 🤔

Only 2 things come to mind so far.
1. Wrap all subscription-related writes in a transaction.
2. Turn session router into an asynchronous in-memory index of the sessions subscriptions table.

On the other hand, the risk of having inconsistent state between sessions and session router tables doesn't compromise correctness (not 100% sure, please tell if you agree): i.e. we only risk persisting messages that no client will ever read. So, I think it might be fine for now, but we'll probably need to think up a way how to reconcile this inconsistency eventually.

> On second thought, I wonder if this is redundant/overkill...

In general, for a transactionless idempotent multi-record write we supposedly should designate a kind of a "write root", a record that's written the last, and the existence of which can later tell that the whole write sequence was successful, so we wouldn't need to repeat (and a way to tell if a record is part of incomplete write, in order to GC those records eventually).

With subscriptions, "write root" is probably session's subscription record in mnesia. So write sequence supposedly should be _iterators_ -> _session route_ -> _session subscription record_. Thus we will know for sure that, if this _session subscription record_ exists, everything written before it also exists.

These are general thoughts, and this scheme is likely not universally applicable. Yet, should be fine as a kind of coarse mental model. What do you think?

keynslug: > perhaps it would be safer to first ensure the iterator is added to the session before adding the route to make it idempotent when restarting the subscription.

Sorry for the confusion. AFAICT it's more correct the way it currently is in this PR.

I.e. without transactions (or some other mechanism) we cannot hope for consistency anyway, so we should write the route record before subscription record. Otherwise, the client will not see messages written after `StartMS` but before it reestablishes the connection (and optionally resubscribes).

PR #11592, opened 2023-09-12
feat(session): introduce session implementation concept

keynslug:
Fixes [EMQX-10827](https://emqx.atlassian.net/browse/EMQX-10827).

There are 2 session implementations:
1. `emqx_session_mem` which is basically slightly reworked `emqx_session`.
2. `emqx_persistent_session_ds` (WIP) which uses `emqx_durable_storage` for session / message persistence.

Few pieces of `emqx_session` logic that were separated from it for some reason is now hidden behind `emqx_session_mem`, notably takeovers and replays. Few other pieces that should have been part of `emqx_session` but were part of `emqx_channel` are still there (due to refactoring complexity), they are highlighted with comments.

There are a lot of TODOs.
* [x] `emqx_eviction_agent_channel`
* [x] Rolling upgrade safety.
* [x] Working persistent session tests.
* [x] Dialyzer complaints.

### <samp>🤖 Generated by Copilot at 4719be1</samp>

Refactor session and message persistence modules. Split `emqx_session.hrl` into two files for different storage backends. Simplify timer and queue management in `emqx_channel.erl` and `emqx_mqueue.erl`. Remove unnecessary dependencies and logic from `emqx_cm.erl` and `emqx_sup.erl`. Rename `emqx_persistent_session` to `emqx_persistent_message` and implement a new message persistence layer. Create `emqx_session_events.erl` to handle session events.

## PR Checklist


apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -160,4 +160,18 @@ assert_messages_order([Msg | Ls1], [{publish, #{payload := No}} | Ls2]) ->
...
+    % NOTE
+    % Introduce randomness so that natural order is not the same as arrival order.
+    iolist_to_binary(
+        io_lib:format("~4.16.0B [~B] [~s]", [
+            rand:uniform(16#10000) - 1,
+            I,
+            calendar:system_time_to_rfc3339(erlang:system_time(millisecond), [
+                {unit, millisecond},
+                {offset, "Z"}
+            ])
```
thalesmg, line 175: Nano nit:

```suggestion
            emqx_utils_calendar:now_to_rfc3339(millisecond)
```

apps/emqx/include/emqx_session.hrl
```diff
@@ -17,39 +17,7 @@
...
-    %% have not been completely acknowledged
-    awaiting_rel :: map(),
-    %% Maximum number of awaiting QoS2 messages allowed
-    max_awaiting_rel :: non_neg_integer() | infinity,
-    %% Awaiting PUBREL Timeout (Unit: millisecond)
-    await_rel_timeout :: timeout(),
-    %% Created at
-    created_at :: pos_integer()
-}).
+-define(IS_SESSION_IMPL_MEM(S), (is_tuple(S) andalso element(1, S) =:= session)).
```
ieQu1, line 20: Q: can we use `is_record(S, session)` here and below?

apps/emqx/src/emqx_channel.erl
```diff
@@ -1220,14 +1221,18 @@ handle_info(
...
         #channel{
-            conn_state = ConnState
+            conn_state = ConnState,
+            clientinfo = ClientInfo,
+            session = Session
         }
 ) when
     ConnState =:= connected orelse ConnState =:= reauthenticating
 ->
+    {Intent, Session1} = emqx_session:disconnect(ClientInfo, Session),
```
keynslug, line 1231: Discussion point. @thalesmg @ieQu1 

Here we ask the session what it wants the channel to do upon disconnect (`Intent`), because we don't really want the channel backed by new persistent session implementation to live after the client has gone. This requires one more function in the session API surface, and I naturally wonder if there's a way to simplify this.

For reference, [this is](https://github.com/emqx/emqx/blob/b8184966ab40e847c2b0b5e2bf6399aa48fff0fc/apps/emqx/src/emqx_session_mem.erl#L713-L715) how `emqx_session_mem` implements it, and [this is](https://github.com/emqx/emqx/blob/b8184966ab40e847c2b0b5e2bf6399aa48fff0fc/apps/emqx/src/emqx_persistent_session_ds.erl#L283-L284) how `emqx_persistent_session_ds` does.

apps/emqx/src/emqx_cm.erl
```diff
@@ -262,96 +262,64 @@ set_chan_stats(ClientId, ChanPid, Stats) ->
...
+    emqx_cm_locker:trans(ClientId, fun(_) ->
+        case emqx_session:open(ClientInfo, ConnInfo) of
+            {true, Session, ReplayContext} ->
+                ok = register_channel(ClientId, Self, ConnInfo),
+                {ok, #{session => Session, present => true, replay => ReplayContext}};
+            false ->
                 create_register_session(ClientInfo, ConnInfo, Self)
         end
-    end,
-    emqx_cm_locker:trans(ClientId, ResumeStart).
```
keynslug, line 292: Discussion point. @thalesmg @ieQu1 

Half of this stuff is now part of `takeover_channel_session/2` function, which [is used by `emqx_session_mem` itself](https://github.com/emqx/emqx/blob/b8184966ab40e847c2b0b5e2bf6399aa48fff0fc/apps/emqx/src/emqx_session_mem.erl#L207-L220) (since it's essentially part of in-memory session logic). That's why there's now a functor for "stuff that should happen after takover begin but before takeover end".

What was named `Pendings` before is now hidden behind opaque `ReplayContext` term, named that way because it [gets passed back to `emqx_session:replay/3` later](https://github.com/emqx/emqx/blob/b8184966ab40e847c2b0b5e2bf6399aa48fff0fc/apps/emqx/src/emqx_channel.erl#L2028).

This needs a closer look to ensure that logic is essentially the same as before. See also next comment.

apps/emqx/src/emqx_channel.erl
```diff
@@ -2051,30 +2022,25 @@ maybe_resume_session(#channel{resuming = false}) ->
...
     clientinfo = ClientInfo
 }) ->
-    {ok, Publishes, Session1} = emqx_session:replay(ClientInfo, Session),
-    case emqx_session:deliver(ClientInfo, Pendings, Session1) of
-        {ok, Session2} ->
-            {ok, Publishes, Session2};
-        {ok, More, Session2} ->
-            {ok, lists:append(Publishes, More), Session2}
-    end.
+    emqx_session:replay(ClientInfo, ReplayContext, Session).
```
keynslug, line 2028: Discussion point. @thalesmg @ieQu1 

This is `ReplayContext` at play here. All this logic, including some [side-effects](https://github.com/emqx/emqx/pull/11592/commits/b8184966ab40e847c2b0b5e2bf6399aa48fff0fc#diff-27c0844d0e2116cd96569c34369a4b485d68529f0d21cb075c19f8bc5a59f477L595) from `process_connect/2` was [moved to `emqx_session_mem:replay/3`](https://github.com/emqx/emqx/blob/b8184966ab40e847c2b0b5e2bf6399aa48fff0fc/apps/emqx/src/emqx_session_mem.erl#L677-L690). Again, this needs a careful look that logic is kept (mostly) intact.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_session.erl
```diff
@@ -78,67 +76,65 @@ stats(#{session := Session}) ->
     emqx_session:stats(Session).
 
 puback(ClientInfo, MsgId, Session) ->
-    with_sess(?FUNCTION_NAME, [ClientInfo, MsgId], Session).
+    with_sess(puback, [ClientInfo, MsgId], Session).
```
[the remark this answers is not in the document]
keynslug: Leftovers from the final iteration. I'll revert those changes.

apps/emqx/include/emqx_session.hrl
keynslug: Quick experiment shows that it would require bringing both records in the scope where these marcos are used.

apps/emqx/include/emqx_session.hrl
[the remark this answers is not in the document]
keynslug: I probably need to state that I plan to revert that one, because it complicates RPC compatibility a lot.

> Do I understand correctly, that this change makes mqueue more of a pure data structure and removes MQTT specifics from it?

Yes. There's also this mess that is the result of misplaced responsibility (IMO) that I wanted to get rid of:
https://github.com/emqx/emqx/blob/f2346b2e9a7341fdefbb9da091d3769d8fbf3a13/apps/emqx/src/emqx_session.erl#L604-L606

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_session.erl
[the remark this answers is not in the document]
thalesmg: IMO, the new version is more grep-friendly when searching for usage sites of this function.

apps/emqx/src/emqx_cm.erl
thalesmg: I still have to look more closely; as an initial observation we seem to have a chain of calls `emqx_cm` -> `emqx_session_mem` -> `emqx_cm` -> `emqx_session_mem` (via the continuation), and that only happens in the memory implementation.  Since that function is now only used by `emqx_session_mem`, perhaps the function could be entirely moved to `emqx_session_mem`.

apps/emqx/src/emqx_channel.erl
keynslug: > That is, there's no need to keep the channel process around for takeover when we have persistent sessions, so we terminate it?

Right. Also parts of `emqx_channel` logic that kick in between channel disconnect and expiration / takeover are specific to `emqx_session_mem`. So this is also a kind of safety measure, to ensure that idling channel with ds persistent sessions processes will not crash when they receive some message for some reason.

<sup>Thanks for the question! It seems that I missed the fact that a channel with in-memory session can still try to takeover a persistent session (which will fail miserably). I'll need to take that into account, however it sounds like a stuff for a followup PRs, especially given that the semantics of that case are unclear.</sup>

apps/emqx/src/emqx_cm.erl
keynslug: Good point, I think it indeed should ideally be there. I believe I thought about that but was stopped by the fact that the logic of `takeover_channel_session/2` heavily relies on `emqx_cm_proto_v2`, hence must stay in `emqx_cm`. However, we may try to end up in a middle ground: export `takeover_begin` / `takeover_end` from `emqx_cm`, and use them in `emqx_cm`. What do you think?

thalesmg: I thought that the reason was to avoid calling `emqx_cm_provo_v2` outside `emqx_cm`.  But, IMO, making the flow harder to follow outweighs calling `emqx_cm_proto_v2` in `emqx_session_mem`.  

If the naming of the proto module feels weird, we could introduce `emqx_session_mem_proto_v1` that calls the same old `emqx_cm:takeover_{session,finish}` functions, as they will exist on older versions.

> However, we may try to end up in a middle ground: export takeover_begin / takeover_end from emqx_cm, and use them in emqx_cm. What do you think?

IIUC, `takeover_session` (begin) and `takeover_finish` are already exported since they are RPC targets, so they could be called from `emqx_session_mem`?

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -47,95 +68,261 @@
...
+    #sessionds{
+        id = ClientID,
+        subscriptions = #{},
+        conf = Conf
+    }.
+
+-spec open(clientinfo(), conninfo()) ->
+    {true, session()} | false.
+open(#{clientid := ClientID}, _ConnInfo) ->
+    open_session(ClientID).
```
thalesmg, line 101: Upon closer inspection, I think the PS implementation will need to do something similar to `emqx_cm:pick_channel` (perhaps calling `kick_session/2` for the living client and `discard_session/2` for the potentially stale ones would be more precise) to kick any living client during takeover.

apps/emqx/src/emqx_channel.erl
thalesmg: The reference about the side-effects unfortunately was lost after some rebase, but I assume it's  `emqx_utils:drain_deliver()` and possibly `maybe_{,n}ack` in `session_mem`?

keynslug: Which one? One is still here, where `{takeover, 'end'}` is being handled. Another wasn't marked with such comment before for some reason:
https://github.com/emqx/emqx/blob/e089fda260574cac279faae6efaee798f87703be/apps/emqx/src/emqx_channel.erl#L603-L605

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: Very good point.

apps/emqx/src/emqx_channel.erl
thalesmg: I meant this link from the first message: https://github.com/emqx/emqx/pull/11592/commits/b8184966ab40e847c2b0b5e2bf6399aa48fff0fc#diff-27c0844d0e2116cd96569c34369a4b485d68529f0d21cb075c19f8bc5a59f477L595

keynslug: Ah, sorry. 😅 Yes, that's the `emqx_utils:drain_deliver()` call, highlighted in my previous comment.

apps/emqx/src/emqx_session.erl
```diff
@@ -831,115 +458,100 @@ run_terminate_hooks(ClientInfo, takenover, Session) ->
...
+should_discard(MsgDeliver) ->
+    is_banned_msg(MsgDeliver).
 
-batch_n(Inflight) ->
-    case emqx_inflight:max_size(Inflight) of
-        0 -> ?DEFAULT_BATCH_N;
-        Sz -> Sz - emqx_inflight:size(Inflight)
-    end.
+is_banned_msg(#message{from = ClientId}) ->
+    [] =:= emqx_banned:look_up({clientid, ClientId}).
```
thalesmg, line 533: Nit: I know this function was like this before.  But the logic and the name seem reversed?

```suggestion
is_not_banned_msg(#message{from = ClientId}) ->
    [] =:= emqx_banned:look_up({clientid, ClientId}).
```

apps/emqx/src/emqx_session.erl
```diff
@@ -831,115 +458,100 @@ run_terminate_hooks(ClientInfo, takenover, Session) ->
...
+    ok.
 
--compile({inline, [sort_fun/2, batch_n/1, with_ts/1, age/2]}).
+%%--------------------------------------------------------------------
 
-sort_fun({_, A}, {_, B}) ->
-    A#inflight_data.timestamp =< B#inflight_data.timestamp.
+-spec should_discard(message() | emqx_types:deliver()) -> boolean().
+should_discard(MsgDeliver) ->
+    is_banned_msg(MsgDeliver).
```
thalesmg, line 530: Assuming the logic of `is_banned_msg` is correct, and since this is the only fn used in `should_discard`, it seems like this ought to be:

```suggestion
-spec should_keep(message() | emqx_types:deliver()) -> boolean().
should_keep(MsgDeliver) ->
    is_not_banned_msg(MsgDeliver).
```

thalesmg (commented):
Awesome work!  :beers:

apps/emqx/test/emqx_persistent_session_SUITE.erl
```diff
@@ -444,6 +372,7 @@ t_assigned_clientid_persistent_session(Config) ->
     {ok, Client2} = emqtt:start_link([
         {clientid, AssignedClientId},
         {proto_ver, v5},
+        {properties, #{'Session-Expiry-Interval' => 30}},
```
keynslug, line 375: Discussion point. @thalesmg @ieQu1 
Moderately stupid consequence of how the session implementation [is now chosen](https://github.com/emqx/emqx/blob/3e679df1b6b1bf7e8d8769a7c13734ad57481a24/apps/emqx/src/emqx_session.erl#L557-L565) is that this line is now required for this testcase to pass under `emqx_persistent_session_ds` implementation. Oddly enough, the spec does not explicitly say what the Client wants when it says CleanStart = 0 **and** SessionExpiryInterval = 0 (which is the default), yet after one more read-through I have an impression that the spec wants us to continue with existing session, the `emqx_persistent_session_ds` in this case, meaning that the testcase should pass w/o this added line. Though it's still not clear what SessionExpiryInterval = 0 means then. I wonder how would you interpret the spec?

apps/emqx_eviction_agent/test/emqx_eviction_agent_channel_SUITE.erl
```diff
@@ -199,40 +189,6 @@ t_get_connected_client_count(_Config) ->
         emqx_cm:get_connected_client_count()
     ).
 
-t_persistence(_Config) ->
```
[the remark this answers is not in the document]
keynslug: I was trying to keep it, but it has been failing: the reason is `emqx_eviction_agent_channel:call(Pid, kick)` call that makes the session state perish. Moreover, I believe this testcase is too specific to the former session persistence implementation to justify keeping it.

apps/emqx/src/emqx_channel.erl
```diff
@@ -944,10 +937,12 @@ handle_deliver(
         clientinfo = ClientInfo
     }
 ) ->
+    % NOTE
+    % This is essentially part of `emqx_session_mem` logic, thus call it directly.
```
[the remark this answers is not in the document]
keynslug: It actually relates to the whole clause, which ideally should have been part of `emqx_session_mem` in some way or another.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1075,6 +1070,10 @@ return_connack(AckPacket, Channel) ->
             },
             {Packets, NChannel2} = do_deliver(Publishes, NChannel1),
             Outgoing = [?REPLY_OUTGOING(Packets) || length(Packets) > 0],
+            % NOTE
+            % Session timers are not restored here, so there's a tiny chance that
+            % the session becomes stuck, when it already has no place to track new
+            % messages.
```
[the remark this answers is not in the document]
keynslug: No, the opposite actually, it's the consequence of it. I just wanted to highlight it, and leave the behavior unchanged. Do you think it would be better to just restore the timers and drop the comment?

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -695,6 +687,16 @@ replay(ClientInfo, Session) ->
...
 
+-spec dedup(clientinfo(), [emqx_types:message()], [emqx_types:deliver()], session()) ->
+    [emqx_types:message()].
+dedup(ClientInfo, Pendings, DeliversLocal, Session) ->
+    PendingsLocal1 = emqx_session:enrich_delivers(ClientInfo, DeliversLocal, Session),
+    PendingsLocal2 = lists:filter(
+        fun(Msg) -> not lists:keymember(Msg#message.id, #message.id, Pendings) end,
+        PendingsLocal1
+    ),
+    append(Pendings, PendingsLocal2).
```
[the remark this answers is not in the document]
keynslug: Yes. I'm pretty sure it wasn't safe (i.e. incorrect), yet surprisingly that slightly modified takeover testsuite hasn't been able to catch that. I probably need to look into `emqx_takeover_SUITE` more carefully.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -220,11 +247,20 @@ session_get_iterator_id(DSSessionId, TopicFilter) ->
...
-        mria:transaction(?DS_SHARD, fun() ->
-            mnesia:delete(?ITERATOR_REF_TAB, IteratorRefId, write)
-        end),
-    ok.
+    transaction(fun() ->
+        mnesia:delete(?ITERATOR_REF_TAB, IteratorRefId, write)
+    end).
+
+session_read_iterators(DSSessionId) ->
+    % NOTE: somewhat convoluted way to trick dialyzer
```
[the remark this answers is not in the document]
keynslug: Interesting idea! However, worth noting the result is not the same, i.e. after that change dialyzer won't be able to spot mistakes in places where record is constructed anymore, only in places where we explicitly say that a function accepts / returns them. Still, might be better than this abomination.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1303,66 +1308,65 @@ handle_timeout(
...
-            NChannel = Channel#channel{session = NSession},
-            {ok, clean_timer(await_timer, NChannel)};
-        {ok, Timeout, NSession} ->
-            NChannel = Channel#channel{session = NSession},
-            {ok, reset_timer(await_timer, Timeout, NChannel)}
+    case emqx_session:handle_timeout(ClientInfo, Name, Session) of
+        {ok, [], NSession} ->
+            {ok, Channel#channel{session = NSession}};
+        {ok, Replies, NSession} ->
+            handle_out(publish, Replies, Channel#channel{session = NSession})
```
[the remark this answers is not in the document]
keynslug: No, I think we don't. The point of `{emqx_session, _}` timers is that the session implementation is exclusively responsible for them.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -2083,42 +2063,56 @@ handle_timeout(
...
         {ok, Publishes, Timeout, NSession} ->
             NChannel = Channel#channel{session = NSession},
             %% XXX: These replay messages should awaiting register acked?
-            handle_out(publish, Publishes, reset_timer(retry_timer, Timeout, NChannel))
+            handle_out(publish, Publishes, reset_timer(Name, Timeout, NChannel))
     end;
 handle_timeout(
     _TRef,
-    expire_awaiting_rel,
+    {emqx_session, _Name},
```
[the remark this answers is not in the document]
keynslug: Nowhere yet. 🫠 The idea is that we will use `emqx_session:ensure_timer/3` sooner than later somewhere in the persistent session implementation code, so there's some stuff to make sure this will work.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -2083,42 +2063,56 @@ handle_timeout(
...
-        {ok, NSession} ->
-            {ok, clean_timer(await_timer, Channel#channel{session = NSession})};
-        {ok, Timeout, NSession} ->
-            {ok, reset_timer(await_timer, Timeout, Channel#channel{session = NSession})}
+    case emqx_mqttsn_session:handle_timeout(ClientInfo, Name, Session) of
+        {ok, [], NSession} ->
+            {ok, Channel#channel{session = NSession}};
+        {ok, Publishes, NSession} ->
+            %% XXX: These replay messages should awaiting register acked?
+            handle_out(publish, Publishes, Channel#channel{session = NSession})
```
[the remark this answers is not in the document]
keynslug: Again, the idea here is we don't need to do this on the channel level, because these timers are the responsibility of the session itself.

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -0,0 +1,788 @@
...
+    case Dropped of
+        undefined ->
+            Session#session{mqueue = NewQ};
+        _Msg ->
+            Reason =
+                case emqx_mqueue:info(store_qos0, Q) of
+                    false when QOS =:= ?QOS_0 -> qos0_msg;
+                    _ -> queue_full
+                end,
+            _ = emqx_session_events:handle_event(ClientInfo, {dropped, Dropped, Reason}),
```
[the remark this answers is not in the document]
keynslug: Thanks for the idea! Made slightly different changes in 05347848bf9e84d36f1968b91b66a2ee51da1e37.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -47,146 +69,332 @@
...
+    SubOpts = maps:get(props, Iterator),
+    ok = del_subscription(TopicFilter, Iterator, ID),
+    {ok, Session#{iterators := maps:remove(TopicFilter, Iters)}, SubOpts};
+unsubscribe(
+    _TopicFilter,
+    _Session = #{}
+) ->
+    {error, ?RC_NO_SUBSCRIPTION_EXISTED}.
+
+-spec get_subscription(emqx_types:topic(), session()) ->
```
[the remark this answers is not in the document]
keynslug: This is actually a bug, it breaks `emqx_session` contract. Attempted to fix in a242e3098.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -47,146 +69,332 @@
...
+subscribe(
+    TopicFilter,
+    SubOpts,
+    Session = #{id := ID, iterators := Iters}
+) ->
+    % TODO: max_subscriptions
+    Iterator = add_subscription(TopicFilter, SubOpts, ID),
+    {ok, Session#{iterators := Iters#{TopicFilter => Iterator}}}.
+
+-spec unsubscribe(topic(), session()) ->
```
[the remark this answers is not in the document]
keynslug: Attempted to fix in a242e30987b3a9c82ac3669035fd2fd7bd4969ef

apps/emqx/src/emqx_channel.erl
```diff
@@ -1075,6 +1070,10 @@ return_connack(AckPacket, Channel) ->
             },
             {Packets, NChannel2} = do_deliver(Publishes, NChannel1),
             Outgoing = [?REPLY_OUTGOING(Packets) || length(Packets) > 0],
+            % NOTE
```
ieQu1, line 1073: Nit (code style): it's better to use `%%` for beginning-of-line comments.

apps/emqx/src/emqx_channel.erl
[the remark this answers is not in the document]
thalesmg: No, I think we should leave that to another PR.  I thought this was an intermediate commit leftover, sorry.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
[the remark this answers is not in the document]
thalesmg: Sure.  But it seems like the `emqx_session` / `emqx_session_mem` modules are not resetting/clearing the timers during `publish`, while this `mqttsn_channel` module previously did so?

apps/emqx/test/emqx_session_mem_SUITE.erl
```diff
@@ -0,0 +1,613 @@
...
+    ok = meck:expect(emqx_hooks, run, fun
+        ('message.dropped', [Msg, _By, ReasonName]) ->
+            self() ! {'message.dropped', ReasonName, Msg},
+            ok;
+        (_Hook, _Arg) ->
+            ok
+    end),
+
+    Session = session(#{max_awaiting_rel => 2, awaiting_rel => #{PacketId1 = 1 => ts(millisecond)}}),
+    begin
```
ieQu1: Unrelated to the PR, but `begin`/`end` block does absolutely nothing here. In the future it can be replaced with comment: `Ensure that publishing a new message with the packet ID that matches ID of a packet awaiting release generates an exception.`

apps/emqx/src/emqx_channel.erl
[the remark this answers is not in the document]
keynslug: The changeset is stable (barring few good points and nits in the comments). Common session timers are handled in the previous clause. This is a plumbing for _custom_ timers to be used inside `emqx_persistent_session_ds` later. Admittedly, this is a kind of leftover from the previous iteration, do you think it would be less confusing if we move these into a followup PR?

apps/emqx/src/emqx_channel.erl
```diff
@@ -944,10 +937,12 @@ handle_deliver(
         clientinfo = ClientInfo
     }
 ) ->
+    % NOTE
```
ieQu1, line 940: Nit, coding style: `%` -> `%%` to please the default emacs autoindentation settings

apps/emqx/test/emqx_proper_types.erl
```diff
@@ -134,8 +134,8 @@ sessioninfo() ->
     ?LET(
         Session,
         #session{
-            clientid = clientid(),
```
ieQu1, line 137: Forgotten change?

apps/emqx/src/emqx_channel.erl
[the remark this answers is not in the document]
keynslug: Thanks for the input! Backed them out as part of e3b5d75.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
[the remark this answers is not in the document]
keynslug: Backed them out as part of e3b5d75, this stuff will go in the followup PR.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -64,10 +64,11 @@
     id := emqx_ds:session_id(),
     created_at := _Millisecond :: non_neg_integer(),
     expires_at := _Millisecond :: non_neg_integer() | never,
-    iterators := map(),
     props := map()
 }.
 
+-type iterators() :: #{topic() => iterator()}.
```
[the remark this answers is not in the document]
keynslug: Strictly speaking, we can not rely on types defined in the `emqx` app since it's not a dependency. Refined the type in 2a1e0ad48. Was unsure about renaming, I have a feeling that on this level `topic()` reflects the nature of what it is better than `words()`, which is essentially a name for specific representation.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -77,16 +77,14 @@
...
-
 -type message_store_opts() :: #{}.
 
 -type message_stats() :: #{}.
 
 -type message_id() :: binary().
 
 %% Parsed topic:
--type topic() :: list(binary()).
+-type topic() :: list(binary() | '+' | '#' | '').
```
ieQu1, line 87: Hmm, I don't think this is correct: by the spec the topic cannot include `+` or `#` levels. This change essentially turned this type into a `topic_filter()` type, equivalent to `-type words() :: list(word()).` from `emqx_types`

apps/emqx_durable_storage/src/emqx_ds.erl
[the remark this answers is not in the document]
thalesmg: Fair enough, we can't reference the `emqx` app indeed.

Though `list(binary())` might be slightly imprecise, since we end up using `emqx_topic:words/1` to create those values, which introduce the atoms cited above.  Perhaps we could reflect that in this module's type?

```erlang
-type topic() :: list(binary() | '' | '+' | '#').
```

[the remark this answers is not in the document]
thalesmg: Just saw this change in another commit.  :smile_cat:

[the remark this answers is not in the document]
keynslug: Absolutely! This one is now called `topic_filter()` after 35e91ac.

thalesmg (approved):
:hot_pepper: :hot_pepper: :hot_pepper:

apps/emqx/test/emqx_persistent_session_SUITE.erl
keynslug: Sounds reasonable, thanks!

Still have some doubts. The [MQTT-3.1.2-5] and this piece from the spec

> If it is set to 0, or is absent, the Session ends when the Network Connection is closed. 

...Kinda leave the impression that (if read literally) we should take the existing session, but update its expiry interval to 0. So that it will perish once the _new_ client connection goes away.

<sup>On a sidenote: perhaps it should be fine to decide which interpretation to assume in a followup PR. 🤔 </sup>

thalesmg: I agree, if the original connection had `SessionExpiryInterval > 0`, those items suggest the behavior you described.  

If the _original connection_ had `SessionExpiryInterval = 0`, then I think even with `CleanStart = 0` we should destroy the session rather than letting it take over.

keynslug: > I agree, if the original connection had `SessionExpiryInterval > 0`, those items suggest the behavior you described.

👍🏼 

This is exactly what's happening in the testcase. Which then means we need to look in persistent session storage first for any CONNECT with `CleanStart = 0` (we're kinda lucky there's only one persistent session storage currently 🌝).

I'll try to tackle that in the current PR, hope it won't require too much effort.

> If the _original connection_ had `SessionExpiryInterval = 0`, then I think even with `CleanStart = 0` we should destroy the session rather than letting it take over.

Agree. It should have already been destroyed in fact, by the time the next connection is established.

keynslug: Made an attempt to revamp this in 05b304ebe.

apps/emqx/src/emqx_session.erl
```diff
@@ -831,115 +469,109 @@ run_terminate_hooks(ClientInfo, takenover, Session) ->
...
+-spec choose_impl_candidates(conninfo()) -> [module()].
+choose_impl_candidates(#{expiry_interval := EI}) ->
+    choose_impl_candidates(EI, emqx_persistent_message:is_store_enabled()).
 
-set_field(Name, Value, Session) ->
-    Pos = emqx_utils:index_of(Name, record_info(fields, session)),
-    setelement(Pos + 1, Session, Value).
+choose_impl_candidates(_, false) ->
+    [emqx_session_mem];
+choose_impl_candidates(0, true) ->
```
thalesmg, line 564: Nit:

```suggestion
choose_impl_candidates(_, _IsPSStoreEnabled = false) ->
    [emqx_session_mem];
choose_impl_candidates(0, _IsPSStoreEnabled = true) ->
```

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -144,28 +145,36 @@ message_stats() ->
...
             [] ->
-                Session = export_record(session_create(SessionId, Props)),
-                {true, Session, #{}}
+                false
         end
     end).
 
+-spec session_ensure(session_id(), _Props :: map()) ->
+    {ok, session(), iterators()}.
+session_ensure(SessionId, Props) ->
```
thalesmg, line 171: Nit: the function name gives me the impression that it would have the previous `session_open` semantics: if a matching session already exists, it's returned; otherwise, a fresh one is returned; rather than always returning a fresh one.  :thinking: 

Perhaps:

```suggestion
session_ensure_new(SessionId, Props) ->
```

thalesmg (approved):

apps/emqx_durable_storage/src/emqx_ds.erl
keynslug: ✅ a2ddd9d5f.

apps/emqx/src/emqx_session.erl
keynslug: ✅ a2ddd9d5f.

keynslug: ✅ a2ddd9d5f.

thalesmg (approved):

PR #11607, opened 2023-09-14
feat: check cluster.static.seeds's integrity

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-11206

When nodes are removed from a static cluster through the CLI, an updated message will suggest that users also update their configuration files.

1. Start a cluster with nodes A, B and C.
```
cluster {
  name = emqxcl
  discovery_strategy = static
  static { seeds = [A, B, C]
}
```
2. After two weeks of running, finding only two nodes is sufficient.
3. Remove node C through `./emqx ctl cluster leave A` and `./emqx stop` on node C.
4. It is suggested that the user remove C from Node A/B's seeds, otherwise A'B will attempt to connect to C.

### <samp>🤖 Generated by Copilot at 1d9afd2</samp>

This pull request enhances the `emqx_mgmt_cli` module with a new function to check the cluster configuration integrity and a corresponding test case. The function `check_cluster_config_integrity/2` detects and reports any nodes that are configured as seeds but not part of the cluster. The test case `t_cluster_info/1` in `emqx_mgmt_cli_SUITE.erl` covers this scenario and asserts the expected output.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_cli.erl
+++ b/apps/emqx_management/src/emqx_mgmt_cli.erl
@@ -46,13 +46,17 @@
     data/1
 ]).
 
+-ifdef(TEST).
+-export([cluster_info/0]).
+-endif.
+
 -spec load() -> ok.
 load() ->
     Cmds = [Fun || {Fun, _} <- ?MODULE:module_info(exports), is_cmd(Fun)],
     lists:foreach(fun(Cmd) -> emqx_ctl:register_command(Cmd, {?MODULE, Cmd}, []) end, Cmds).
 
 is_cmd(Fun) ->
-    not lists:member(Fun, [init, load, module_info]).
+    not lists:member(Fun, [init, load, module_info, cluster_info]).
 
 %%--------------------------------------------------------------------
 %% @doc Node status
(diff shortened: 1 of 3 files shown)


apps/emqx_management/src/emqx_mgmt_cli.erl
```diff
@@ -939,11 +943,54 @@ with_log(Fun, Msg) ->
...
+                msg => "node_configured_in_seeds_but_missing_in_cluster",
+                orphan_nodes_in_seeds => NodesMissInCluster
+            },
+            ?SLOG(warning, Msg),
+            Msg;
+        {_, [_ | _]} ->
+            %% User should add orphan nodes into seeds manually or stop the orphan nodes.
+            %% so we call those orphan_nodes as `orphan_nodes_not_in_seeds`.
+            Msg = Msg0#{
+                msg => "node_in_cluster_but_not_configured_in_seeds",
```
ieQu1, line 985: I think this warning message is not necessary. One could configure one node as the seed and use it alone to bootstrap a cluster with however many nodes. It's not necessary to list all of the nodes as seeds: once the node connects to the cluster, the information about the rest of the nodes will be exchanged through other means.

ieQu1:
I think manual `emqx_ctl cluster join` leave is generally incompatible with the automatic cluster discovery. Maybe the simplest solution is to print a warning when these commands are executed and discovery strategy is not set to `manual`.

PR #11613, opened 2023-09-14
feat(ds): introduce keyspace concept

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10579

This introduces the concept of "keyspaces" to our durable storage (DS) implementation, and also refactors some places where "shard" and "keyspace" would be mixed up.

We might want to tune the storage options differently for distinct sets of topics, the keyspaces.  The keyspace is composed by one or more shards.

- Keyspaces are identified simply by binary strings.
- DS configuration is scoped by keyspaces instead of shards.
- Starting a new DS shard requires definining to which keyspace the shard belongs.

### <samp>🤖 Generated by Copilot at 772690d</samp>

This pull request adds support for multiple keyspaces in the durable storage module, which allows different settings and options for different types of data in Cassandra. It refactors the configuration, message storage, storage layer, and test modules to use and pass the keyspace name as an extra argument. It also updates the persistent session module to use a macro for the default keyspace name.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -46,7 +46,7 @@
 
 %% FIXME
 -define(DS_SHARD, <<"local">>).
--define(DEFAULT_KEYSPACE, <<"#">>).
+-define(DEFAULT_KEYSPACE, default).
 
 -define(WHEN_ENABLED(DO),
     case is_store_enabled() of
@@ -60,10 +60,16 @@
 init() ->
     ?WHEN_ENABLED(begin
         ok = emqx_ds:ensure_shard(
-            ?DS_SHARD,
             ?DEFAULT_KEYSPACE,
+            ?DS_SHARD,
             #{
-                dir => filename:join([emqx:data_dir(), ds, messages, ?DS_SHARD])
+                dir => filename:join([
+                    emqx:data_dir(),
+                    ds,
+                    messages,
+                    atom_to_binary(?DEFAULT_KEYSPACE),
+                    ?DS_SHARD
+                ])
             }
         ),
         ok = emqx_persistent_session_ds_router:init_tables(),
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -45,8 +45,9 @@
 ]).
 
 %% FIXME
--define(DS_SHARD, <<"local">>).
+-define(DS_SHARD_ID, <<"local">>).
 -define(DEFAULT_KEYSPACE, default).
+-define(DS_SHARD, {?DEFAULT_KEYSPACE, ?DS_SHARD_ID}).
 
 -define(WHEN_ENABLED(DO),
     case is_store_enabled() of
@@ -60,15 +61,14 @@
 init() ->
     ?WHEN_ENABLED(begin
         ok = emqx_ds:ensure_shard(
-            ?DEFAULT_KEYSPACE,
             ?DS_SHARD,
             #{
                 dir => filename:join([
                     emqx:data_dir(),
                     ds,
                     messages,
-                    atom_to_binary(?DEFAULT_KEYSPACE),
-                    ?DS_SHARD
+                    ?DEFAULT_KEYSPACE,
+                    ?DS_SHARD_ID
                 ])
             }
         ),
@@ -97,7 +97,9 @@ store_message(Msg) ->
     ID = emqx_message:id(Msg),
     Timestamp = emqx_guid:timestamp(ID),
     Topic = emqx_topic:words(emqx_message:topic(Msg)),
-    emqx_ds_storage_layer:store(?DS_SHARD, ID, Timestamp, Topic, serialize_message(Msg)).
+    emqx_ds_storage_layer:store(
+        ?DS_SHARD, ID, Timestamp, Topic, serialize_message(Msg)
+    ).
 
 has_subscribers(#message{topic = Topic}) ->
     emqx_persistent_session_ds_router:has_any_route(Topic).
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -46,6 +46,7 @@
 
 %% FIXME
 -define(DS_SHARD, <<"local">>).
+-define(DEFAULT_KEYSPACE, <<"#">>).
 
 -define(WHEN_ENABLED(DO),
     case is_store_enabled() of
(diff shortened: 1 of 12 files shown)


apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -333,13 +342,13 @@ create_gen(GenId, Since, {Module, Options}, S = #s{db = DBHandle, cf_generations
     },
     {ok, Gen, S#s{cf_generations = NewCFs ++ CFs}}.
 
--spec open_db(emqx_ds:shard(), options()) -> {ok, state()} | {error, _TODO}.
-open_db(Shard, Options) ->
+-spec open_db(emqx_ds:shard(), emqx_ds:keyspace(), options()) -> {ok, state()} | {error, _TODO}.
+open_db(Shard, Keyspace, Options) ->
     DBDir = unicode:characters_to_list(maps:get(dir, Options, Shard)),
```
keynslug, line 347: Nit. I have a feeling that keyspace should be part of default `DBDir` now.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -135,9 +143,10 @@
 %% API funcions
 %%================================================================================
 
--spec start_link(emqx_ds:shard(), emqx_ds_storage_layer:options()) -> {ok, pid()}.
-start_link(Shard, Options) ->
-    gen_server:start_link(?REF(Shard), ?MODULE, {Shard, Options}, []).
+-spec start_link(emqx_ds:shard(), emqx_ds:keyspace(), emqx_ds_storage_layer:options()) ->
+    {ok, pid()}.
+start_link(Shard, Keyspace, Options) ->
+    gen_server:start_link(?REF(Shard), ?MODULE, {Shard, Keyspace, Options}, []).
```
keynslug, line 149: Q: Shouldn't `Keyspace` also be part of process name here?

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -46,6 +46,7 @@
 
 %% FIXME
 -define(DS_SHARD, <<"local">>).
+-define(DEFAULT_KEYSPACE, <<"#">>).
```
keynslug, line 49: Nit. I get the idea, yet at the moment I feel that we'll have some atom names for any keyspaces in the future config anyway. Maybe just?
```suggestion
-define(DEFAULT_KEYSPACE, default).
```

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -96,10 +98,10 @@
 %% API funcions
 %%================================================================================
 
--spec ensure_shard(shard(), emqx_ds_storage_layer:options()) ->
+-spec ensure_shard(shard(), keyspace(), emqx_ds_storage_layer:options()) ->
```
ieQu1, line 101: Nit: Keyspace, Shard is a more natural order of the arguments.

apps/emqx_durable_storage/src/emqx_ds_message_storage_bitmask.erl
```diff
@@ -236,16 +237,18 @@ create_new(DBHandle, GenId, Options) ->
 %% Reopen the database
 -spec open(
     emqx_ds:shard(),
```
ieQu1, line 239: I suggest using `Keyspace, Shard` ordering of the arguments here as well

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
thalesmg: Since each shard belongs to a single keyspace, I was thinking that it might not be needed.

But if we assume that shards from different keyspaces might use the same name, then that's a good idea.

apps/emqx/src/emqx_persistent_session_ds.erl
thalesmg: Sure.  It wasn't clear to me what we would use for keyspace identifiers, so I assumed they would be topic filters.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
thalesmg: Now I'm thinking if keyspace also shouldn't be an argument to `storage_layer:store/5`, if shard ids may collide.  :thinking:

apps/emqx_durable_storage/test/props/prop_replay_message_storage.erl
```diff
@@ -255,7 +256,7 @@ iterate_shim(Shim, Iteration) ->
 open_db(Filepath, Options) ->
     {ok, Handle} = rocksdb:open(Filepath, [{create_if_missing, true}]),
     {Schema, CFRefs} = emqx_ds_message_storage_bitmask:create_new(Handle, ?GEN_ID, Options),
-    DB = emqx_ds_message_storage_bitmask:open(?ZONE, Handle, ?GEN_ID, CFRefs, Schema),
+    DB = emqx_ds_message_storage_bitmask:open(?SHARD, ?KEYSPACE, Handle, ?GEN_ID, CFRefs, Schema),
```
ieQu1, line 259: Parameters should be swapped here too.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
keynslug: Might be worth exploring making it part of the _shard_ type instead? I.e. `-type shard() :: {_Keyspace, _ShardID}.`.

ieQu1 (approved):

apps/emqx_durable_storage/test/props/prop_replay_message_storage.erl
```diff
@@ -10,7 +10,9 @@
 -define(WORK_DIR, ["_build", "test"]).
 -define(RUN_ID, {?MODULE, testrun_id}).
 
--define(ZONE, ?MODULE).
+-define(KEYSPACE, ?MODULE).
+-define(SHARD_ID, <<(atom_to_binary(?KEYSPACE))/binary, "_shard">>).
```
keynslug, line 14: Nit. Maybe just `<<"shard">>` is enough? 🤔

keynslug (approved):

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -98,8 +102,8 @@
 
 -spec ensure_shard(shard(), emqx_ds_storage_layer:options()) ->
     ok | {error, _Reason}.
-ensure_shard(Shard, Options) ->
-    case emqx_ds_storage_layer_sup:start_shard(Shard, Options) of
+ensure_shard({Keyspace, ShardId}, Options) ->
+    case emqx_ds_storage_layer_sup:start_shard({Keyspace, ShardId}, Options) of
```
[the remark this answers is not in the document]
keynslug: Just got a feeling that there are a tad too many places where this is happening, in the `emqx_ds_storage_layer_sup` module for example. 🫠 That's just a nit, feel free to ignore it.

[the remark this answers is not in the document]
thalesmg: I wanted to assert this new shape in some places.  I'll change it back to reduce the diff size.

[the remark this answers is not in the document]
keynslug: Agree that assertions have to happen somewhere. My point that the code would be just slightly simpler if they happen once per control flow branch.

keynslug (approved):
🔥🔥🔥

PR #11617, opened 2023-09-15
fix: keep builtin_console actions not complain args

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-10992
Don't need changelog, v5.2.0 is ok.

There was previously a bug allowing invalid fields:

The API for creation/update did not validate the args field, and rules were being created with default args.

The configuration file also did not validate the args field.

Now after being fixed, validation has been enabled:

This causes any rules/configurations saved earlier with invalid args to no longer work.

We fixed the validate here:
https://github.com/emqx/emqx/commit/78c5a779d7c721663470609acc80c52efd3c7886#diff-5799d1ec01ad3359596ab22f7781d2b34e8f7909194bb0785f7d1dbe677a0d88L202-L208

compatibility with `actions. args`
```
rule_engine {
  ignore_sys_message = true
  jq_function_default_timeout = 10s
  rules {
    rule_u54w {
      actions = [
        {
          args {}
          function = console
        }
      ]
      description = ""
      enable = true
      metadata {created_at = 1694748634418}
      name = ""
      sql = "SELECT\n  *\nFROM\n  \"t/#\" \n  WHERE payload.message.header.function='pmMsgReq'"
    }
  }
}
```
And 
image

### <samp>🤖 Generated by Copilot at d72348c</samp>

Fix a validation bug for the `builtin_action_console` rule action and update the `emqx_rule_engine` version. The bug caused the console action to fail validation when saving the rule config due to an invalid `args` field. The fix removes the `args` field from the console action config and adds it to the schema as a deprecated and ignored field. The version update reflects the bug fix and other changes in the application.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -647,7 +647,7 @@ set_keepalive(put, #{bindings := #{clientid := ClientID}, body := Body}) ->
         error ->
             {400, 'BAD_REQUEST', "Interval Not Found"};
         {ok, Interval} ->
-            case emqx_mgmt:set_keepalive(emqx_mgmt_util:urldecode(ClientID), Interval) of
+            case emqx_mgmt:set_keepalive(ClientID, Interval) of
                 ok -> lookup(#{clientid => ClientID});
                 {error, not_found} -> {404, ?CLIENTID_NOT_FOUND};
                 {error, Reason} -> {400, #{code => 'PARAMS_ERROR', message => Reason}}
--- a/apps/emqx_rule_engine/src/emqx_rule_engine_schema.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_engine_schema.erl
@@ -114,6 +114,7 @@ fields("builtin_action_console") ->
         {args,
             ?HOCON(map(), #{
                 deprecated => true,
+                importance => ?IMPORTANCE_HIDDEN,
                 desc => "The arguments of the built-in 'console' action",
                 default => #{}
             })}
--- a/apps/emqx_rule_engine/src/emqx_rule_engine_schema.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_engine_schema.erl
@@ -98,10 +98,25 @@ fields("builtin_action_republish") ->
     ];
 fields("builtin_action_console") ->
     [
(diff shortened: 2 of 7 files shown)


kjellwinblad (dismissed):
Looks ok to me but would be great with a test case to ensure that we don't break this in the future!

zmstone (dismissed):

zmstone (dismissed):

apps/emqx_rule_engine/src/emqx_rule_engine_schema.erl
```diff
@@ -98,10 +98,25 @@ fields("builtin_action_republish") ->
...
+
+        %% The "args" field was not validated by the HOCON schema before 5.2.0,
+        %% which allowed rules to be created with invalid "args" data.
+        %% In 5.2.1 the validation was added,
+        %% so existing rules saved with invalid "args" would now fail validation
+        %% To maintain backward compatibility for existing rule data that may contain invalid "args",
+        %% the field needs to be included in the schema even though it is not a valid field.
+        {args,
+            ?HOCON(map(), #{
+                deprecated => true,
```
thalesmg, line 116:
```suggestion
                deprecated => true,
                importance => ?IMPORTANCE_HIDDEN,
```
?

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -647,7 +647,7 @@ set_keepalive(put, #{bindings := #{clientid := ClientID}, body := Body}) ->
         error ->
             {400, 'BAD_REQUEST', "Interval Not Found"};
         {ok, Interval} ->
-            case emqx_mgmt:set_keepalive(emqx_mgmt_util:urldecode(ClientID), Interval) of
```
thalesmg, line 650: Q: is this change intended for this PR?  Seems unrelated.  :thinking:

thalesmg (approved):
Nit: maybe this could have a test case?

apps/emqx_management/src/emqx_mgmt_api_clients.erl
zhongwencool: Yes, It's not related, I just fix this incidentally, because `keepalive` api is hidden for swagger.

JimMoen (approved):

lafirest (approved):

terry-xiaoyu (approved):

paulozulato (approved):

PR #11683, opened 2023-09-25
test: refactor emqx_rpc unit tests

zmstone:
replace  `prop_emqx_rcp` with simple eunit test cases.
the old tests generate random node names, 
as a result, 3/4 of the proper+eunit test run is flooded with gen_rpc error/crash logs.


apps/emqx/src/emqx_rpc.erl
thalesmg, line 1: It seems that testcases for `cast` and `multicall` (gen_rpc) are still needed?  Only `erpc:multicall` is being tested now.

thalesmg (approved):
should we test `cast` as `prop_emqx_rpc` did before?

zmstone:
> should we test `cast` as `prop_emqx_rpc` did before?

will add one in another pr

PR #11703, opened 2023-09-28
fix(ldap): use the search result as bind target

lafirest:
## Summary

Avoid using `base_dn` template in LDAP bind query.
Prior to this change, `base_dn` is a template which can be rendered then used in LDAP search argument, and also used to render the DN in bind query.
Such implementation forces the DN to contain values which are template-able in EMQX. e.g. `uid=${username}`
However, often there are requirements to search user in LDAP using other information such as `email` which is not a part of DN.
After this change, the filter condition will need to be specific enough to find only one user, then use the DN from the query result for binding.

## PR Checklist

### DIFF
--- a/apps/emqx_ldap/src/emqx_ldap_bind_worker.erl
+++ b/apps/emqx_ldap/src/emqx_ldap_bind_worker.erl
@@ -58,14 +58,12 @@ on_stop(InstId, _State) ->
 
 on_query(
     InstId,
-    {bind, Data},
+    {bind, DN, Data},
     #{
-        base_tokens := DNTks,
         bind_password := PWTks,
         bind_pool_name := PoolName
     } = State
 ) ->
-    DN = emqx_placeholder:proc_tmpl(DNTks, Data),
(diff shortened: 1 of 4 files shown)


apps/emqx_ldap/src/emqx_ldap_bind_worker.erl
```diff
@@ -58,14 +58,12 @@ on_stop(InstId, _State) ->
...
     InstId,
-    {bind, Data},
+    {bind, DN, Data},
     #{
-        base_tokens := DNTks,
         bind_password := PWTks,
         bind_pool_name := PoolName
     } = State
 ) ->
-    DN = emqx_placeholder:proc_tmpl(DNTks, Data),
```
zmstone, line 68: I think we can keep this. 
If DN is not a template, rendering it should be just a no-op

zmstone: should probably deprecate `base_dn` field, and make it optional?

PR #11705, opened 2023-09-28
partial feat: implement basics of kinesis consumer coordinator

thalesmg:
Part of https://emqx.atlassian.net/browse/EMQX-10668

This adds just the basic election and shard distribution mechanisms for Kinesis consumer
coordinators.  A lot is still missing before the bridge is working:

- Nudge (or check periodically) followers and the leader itself to start/stop shard workers.
- Implement the workers themselves with LSN persistence.
- Config schema.
- More tests for shard distribution logic.

## Summary
copilot:summary

## PR Checklist


apps/emqx_bridge_kinesis/src/consumer/emqx_bridge_kinesis_consumer_coordinator.erl
```diff
@@ -0,0 +1,546 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_bridge_kinesis_consumer_coordinator).
```
zmstone, line 4: need @doc to describe the coordinator protocol

thalesmg: :heavy_check_mark: https://github.com/emqx/emqx/pull/11705/commits/548b3896470007d9eaa456e005af8a28b195b6b6

zmstone:
So we need to find a place to store LSN, mnesia is probably the only option. 
But I am not entirely sure if it's necessary to use Mnesia to distribute assignments.
Mnesia data is async-replicated to replica nodes while the membership changes happen in different data channel.

Some discussion points:
* Elect leader only among the core nodes, so they should have the least chance of looking at stale state.
* Based on the number of nodes in the cluster, leader decide which shard goes to which node and `cast` the assignments to all members (self-included)
* Leader periodically do the casts.
* Up on receiving a new assignment, member revoke old assignments and start new assignments with least interruption.

zmstone:
I think it would benefit other application if there is a generic leadership among the core nodes.
Maybe provided by `ekka` or `mria`, the leader process should have a incremental 'generation', or 'epoch' ID which can be used to fence off stale leaders.
The leader should have some primitive APIs
1. Is this node the leader
2. What's the current generation.
3. Accepts local processes to observe leadership/membership changes.

The Kinesis coordinator process can directly get the leader ship and membership information by observing the generic process.

thalesmg:
> But I am not entirely sure if it's necessary to use Mnesia to distribute assignments. Mnesia data is async-replicated to replica nodes while the membership changes happen in different data channel.

I thought of storing the assignments in Mnesia because each node could be restarted and resume work with the last committed config, regardless of the leader being online or not.  Otherwise, the leader must be online for nodes to be restarted, and also they must contact the leader to ask the assignments.

> 
> Some discussion points:
> 
>     * Elect leader only among the core nodes, so they should have the least chance of looking at stale state.

Since we don't use `mria:ro_transaction`, replicant leaders will see the latest data, as the transactions will happen in a core node?

>     * Up on receiving a new assignment, member revoke old assignments and start new assignments with least interruption.

If a shard is transferred between 2 nodes by removing the assignment from node A, assigning it to node B, and then casting notifications to A and B, there's larger potential for message duplication, right?  i.e.: node A might still process some messages while node B also consumes such messages after starting its worker.  

Granted, this is not handled in this PR either.  Under normal operations, the leader could coordinate a more graceful transition to avoid duplication.  If nodes crash, they would read their assignments from Mnesia and shouldn't duplicate messages either, I think, save the case where the last consumed LSN was not committed.

thalesmg:
> I think it would benefit other application if there is a generic leadership among the core nodes. Maybe provided by `ekka` or `mria`, the leader process should have a incremental 'generation', or 'epoch' ID which can be used to fence off stale leaders. The leader should have some primitive APIs

This sounds like Raft? 

>     1. Is this node the leader

`mria_membership:leader()` has this API, but it always returns the oldest core node (regardless of whether it's online).  We'd probably need some other API.

>     3. Accepts local processes to observe leadership/membership changes.

Here we use `mria_membership:monitor` API and common Erlang monitors for monitoring membership/leadership changes, respectively.  Won't these suffice?

thalesmg:
>  But I am not entirely sure if it's necessary to use Mnesia to distribute assignments.

Another potential issue of not persisting the assignments: if the leader is restarted, then care must be taken to have the assignments be consistent, otherwise there might be a lot of work of shuffling shards around if they are less deterministic.

thalesmg:
Superseded by #13157 .

PR #11722, opened 2023-10-06
fix(kafka_producer): send messages to wolff producer to buffer even when connector is in `connecting` state (sync query mode)

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-11085

Messages would not be sent to wolff if the connection was down, so they were effectively lost.

### <samp>🤖 Generated by Copilot at ea170df</samp>

Added a new module `emqx_bridge_kafka_impl_producer` to implement the Kafka bridge producer logic using the `emqx_resource` behaviour. Improved the query handling and resource information in the `emqx_resource_buffer_worker` module. Bumped the version of the `emqx_resource` application and renamed a variable for clarity.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -1048,7 +1048,9 @@ call_query(QM, Id, Index, Ref, Query, QueryOpts) ->
             ?RESOURCE_ERROR(not_found, "resource not found")
     end.
 
-do_call_query(QM, Id, Index, Ref, Query, #{is_buffer_supported := true} = QueryOpts, Resource) ->
+do_call_query(QM, Id, Index, Ref, Query, QueryOpts, #{query_mode := ResQM} = Resource) when
+    ResQM =:= simple_async; ResQM =:= simple_sync
+->
     %% The connector supports buffer, send even in disconnected state
     #{mod := Mod, state := ResSt, callback_mode := CBM} = Resource,
     CallMode = call_mode(QM, CBM),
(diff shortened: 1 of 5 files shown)


qzhuyan (approved):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1048,7 +1048,9 @@ call_query(QM, Id, Index, Ref, Query, QueryOpts) ->
             ?RESOURCE_ERROR(not_found, "resource not found")
     end.
 
-do_call_query(QM, Id, Index, Ref, Query, #{is_buffer_supported := true} = QueryOpts, Resource) ->
+do_call_query(QM, Id, Index, Ref, Query, QueryOpts, #{query_mode := ResQM} = Resource) when
```
zmstone, line 1051: do the authn authz resources end up calling this path?

thalesmg: Yes, they end up calling this function.

thalesmg: Here I introduce a new query mode type: #11724 
Perhaps it would be more appropriate to restrict this behavior to that more specific type.

thalesmg: But I think we would also need to introduce a new query mode for the async version of a buffered connector, to have the correct semantics as well.  :thinking:

thalesmg: https://github.com/emqx/emqx/pull/11724/commits/5467e32b9c9b1126b819d79c61787b1d3e2dccf6

PR #11803, opened 2023-10-23
[NOMERGE] ci(helm): test gen_rpc transports

qzhuyan:
Just for testing.


qzhuyan:
gen_rpc `tls1.3` tests failed on master, so the helm test is proved to be working.
but no idea how tls1.2 works...

PR #11809, opened 2023-10-24
feat(mqttbridge): support file-sourced secrets as passwords

keynslug:
Part of [EMQX-10808](https://emqx.atlassian.net/browse/EMQX-10808).

### <samp>🤖 Generated by Copilot at 7b109e7</samp>

This pull request introduces a new feature to handle secret values, such as passwords, in the configuration of EMQ X. It adds a new `secret()` type to the HOCON schema and a new `emqx_secret` module to wrap and unwrap secret values using closures. It also updates the MQTT bridge connector and its test suite to use the new feature, and adds Swagger API documentation for it.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_schema_secret.erl
+++ b/apps/emqx/src/emqx_schema_secret.erl
@@ -1,0 +1,107 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+%% @doc HOCON schema that defines _secret_ concept.
+-module(emqx_schema_secret).
+
+-include_lib("emqx/include/logger.hrl").
+-include_lib("typerefl/include/types.hrl").
+
+-export([mk/1]).
+
+%% HOCON Schema API
+-export([convert_secret/2]).
+
+%% Target of `emqx_secret:wrap/3`
+-export([load/1]).
+
+%% @doc Secret value.
+-type t() :: binary().
+
+%% @doc Source of the secret value.
+%% * "file://...": file path to a file containing secret value.
+%% * other binaries: secret value itself.
+-type source() :: iodata().
+
+-type secret() :: binary() | function().
+-reflect_type([secret/0]).
+
+-define(SCHEMA, #{
+    required => false,
+    format => <<"password">>,
+    sensitive => true,
+    converter => fun ?MODULE:convert_secret/2
+}).
+
+-dialyzer({nowarn_function, source/1}).
+
+%%
+
+-spec mk(#{atom() => _}) -> hocon_schema:field_schema().
+mk(Overrides = #{}) ->
+    hoconsc:mk(secret(), maps:merge(?SCHEMA, Overrides)).
+
+convert_secret(undefined, #{}) ->
+    undefined;
+convert_secret(Secret, #{make_serializable := true}) ->
+    unicode:characters_to_binary(source(Secret));
+convert_secret(Secret, #{}) when is_function(Secret, 0) ->
+    Secret;
+convert_secret(Secret, #{}) when is_integer(Secret) ->
+    wrap(integer_to_binary(Secret));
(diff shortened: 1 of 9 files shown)


apps/emqx/src/emqx_secret.erl
```diff
@@ -29,13 +29,38 @@
...
     unwrap(Term());
 unwrap(Term) ->
     Term.
+
+%% @doc Inspect the term wrapped in a secret closure.
+-spec term(t(_)) -> _Term.
+term(Wrap) when is_function(Wrap, 0) ->
+    case erlang:fun_info(Wrap, module) of
+        {module, ?MODULE} ->
+            {env, Env} = erlang:fun_info(Wrap, env),
```
zmstone, line 62: this is a great idea.
we can keep the conventional arity-0 wrapper but also make possible to inspect other values.

If we are to make `file://` conventional for EMQX secret files, maybe no need to introduce another module, just implement everything here in this module.

for example:
```
wrap(<<"file://", Path/binary>>) ->
    Source = ?PASSWORD_FILE(Path) ,
    fun() ->
            Path1 = element(2, Source),
            read_secret_file(Path1)
    end;
wrap(Secret) ->
    fun() -> Secret end.

unwrap(Term) when is_function(Term, 0) ->
    %% Handle potentially nested funs
    unwrap(Term());
unwrap(Term) ->
    Term.

source(Wrap) ->
    case erlang:fun_info(Wrap, env) of
        {env, [F]} when is_function(F) ->
            source(F);
        {env, [?PASSWORD_FILE(Path)]} ->
            <<"file://", Path/binary>>;
        {env, [Secret]} ->
            Secret
    end.
```

apps/emqx/src/emqx_schema_secret.erl
```diff
@@ -0,0 +1,107 @@
...
+            throw(invalid_string)
+    catch
+        error:_ ->
+            throw(invalid_type)
+    end.
+
+-spec wrap(source()) -> emqx_secret:t(t()).
+wrap(Source) ->
+    try
+        _Secret = load(Source),
```
zmstone, line 79: we can probably leave the exception to runtime.
i.e. allow the file to be missing when emqx boots up or when new config is added.

integrations usually has a retry loop, so sysadmin can crate the file after seeing error messages.

apps/emqx/src/emqx_secret.erl
keynslug: > maybe no need to introduce another module, just implement everything here in this module.

The reason for a very generic `wrap/3` and a separate module is that I wanted to ensure that `emqx_secret` module won't see any changes for sufficiently long time, due to hot-upgrade concerns.

apps/emqx/src/emqx_schema_secret.erl
keynslug: Good point. On the other hand, should this be configurable? I.e. are there potential situations when we want to crash during configuration loading?

apps/emqx/src/emqx_secret.erl
```diff
@@ -29,13 +29,38 @@
...
+-spec unwrap(t(T)) -> T.
 unwrap(Term) when is_function(Term, 0) ->
     %% Handle potentially nested funs
     unwrap(Term());
 unwrap(Term) ->
     Term.
+
+%% @doc Inspect the term wrapped in a secret closure.
+-spec term(t(_)) -> _Term.
+term(Wrap) when is_function(Wrap, 0) ->
```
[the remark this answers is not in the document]
keynslug: This somewhat hacky approach is motivated by the fact that we want to keep `emqx_secret` following the convention (followed in turn by quite a few libraries) that says that wrapped secrets are 0-arity functions.

apps/emqx/src/emqx_secret.erl
```diff
@@ -19,23 +19,52 @@
...
         Term
     end.
 
+%% @doc Wrap a loader function call over a term in a secret closure.
+%% This is slightly more flexible form of `wrap/1` with the same basic purpose.
+-spec wrap(atom(), _Term) -> t(_).
+wrap(Function, Term) ->
+    fun() ->
+        apply(?LOADER, Function, [Term])
+    end.
```
zmstone, line 50: maybe:
```suggestion
-spec wrap(atom(), _Term) -> t(_).
wrap_dynamic(Term) ->
    fun() ->
        apply(?LOADER, load, [Term])
    end.
```

zmstone: I mean if `Term` is already a contract between the caller and `emqx_secret_loader:Function`, maybe just keep it to one place.

keynslug: Fixed in d278486.

id:
May be test it in higher level tests as well?
- https://github.com/emqx/emqx/blob/master/scripts/ui-tests/docker-compose.yaml
- https://github.com/emqx/emqx/blob/master/.github/workflows/run_helm_tests.yaml#L100

zmstone (approved):

keynslug:
> May be test it in higher level tests as well?

Basically no objections, but this PR currently affects only handling of secrets in the MQTT bridge app, so it's kinda unclear how to test this feature there.

PR #11813, opened 2023-10-24
Fix/william/genrpc ssl port align

qzhuyan:
Fixes <issue-or-jira-number>

### <samp>🤖 Generated by Copilot at 6b9e8bc</samp>

This pull request introduces a new gen_rpc implementation for emqx nodes that supports SSL/TLS and two modes of port discovery. It also adds a new configuration option `ssl_client_port` and updates the helm charts to expose the new ports.

## PR Checklist

### DIFF
--- a/.github/workflows/run_helm_tests.yaml
+++ b/.github/workflows/run_helm_tests.yaml
@@ -76,8 +76,6 @@ jobs:
           EMQX_RPC__CACERTFILE: /opt/emqx/etc/certs/cacert.pem
           EMQX_RPC__CIPHERS: TLS_AES_256_GCM_SHA384,TLS_AES_128_GCM_SHA256
           EMQX_RPC__TLS_VERSIONS: "[tlsv1.3]"
-          EMQX_RPC__SSL_SERVER_PRT: 5370
-          EMQX_RPC__PORT_DISCOVERY: manual
         EOL
     - name: Prepare emqxConfig.EMQX_RPC using ssl1.2
       working-directory: source
@@ -91,7 +89,6 @@ jobs:
           EMQX_RPC__CACERTFILE: /opt/emqx/etc/certs/cacert.pem
           EMQX_RPC__CIPHERS: TLS_AES_256_GCM_SHA384,TLS_AES_128_GCM_SHA256
           EMQX_RPC__TLS_VERSIONS: "[tlsv1.2]"
-          EMQX_RPC__SSL_SERVER_PRT: 5370
           EMQX_RPC__PORT_DISCOVERY: manual
         EOL
     - name: run emqx on chart (k8s)
--- a/.github/workflows/run_helm_tests.yaml
+++ b/.github/workflows/run_helm_tests.yaml
@@ -76,6 +76,8 @@ jobs:
           EMQX_RPC__CACERTFILE: /opt/emqx/etc/certs/cacert.pem
           EMQX_RPC__CIPHERS: TLS_AES_256_GCM_SHA384,TLS_AES_128_GCM_SHA256
           EMQX_RPC__TLS_VERSIONS: "[tlsv1.3]"
+          EMQX_RPC__SSL_SERVER_PRT: 5370
+          EMQX_RPC__PORT_DISCOVERY: manual
         EOL
     - name: Prepare emqxConfig.EMQX_RPC using ssl1.2
       working-directory: source
@@ -89,6 +91,8 @@ jobs:
           EMQX_RPC__CACERTFILE: /opt/emqx/etc/certs/cacert.pem
           EMQX_RPC__CIPHERS: TLS_AES_256_GCM_SHA384,TLS_AES_128_GCM_SHA256
           EMQX_RPC__TLS_VERSIONS: "[tlsv1.2]"
+          EMQX_RPC__SSL_SERVER_PRT: 5370
+          EMQX_RPC__PORT_DISCOVERY: manual
         EOL
     - name: run emqx on chart (k8s)
       if: matrix.discovery == 'k8s'
--- a/apps/emqx_conf/src/emqx_conf_schema.erl
+++ b/apps/emqx_conf/src/emqx_conf_schema.erl
@@ -1151,6 +1151,7 @@ translation("emqx") ->
 translation("gen_rpc") ->
     [
         {"default_client_driver", fun tr_gen_rpc_default_client_driver/1},
+        {"ssl_client_port", fun tr_gen_rpc_ssl_client_port/1},
         {"ssl_client_options", fun tr_gen_rpc_ssl_options/1},
         {"ssl_server_options", fun tr_gen_rpc_ssl_options/1},
         {"socket_ip", fun(Conf) ->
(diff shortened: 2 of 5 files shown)


deploy/charts/emqx-enterprise/templates/StatefulSet.yaml
```diff
@@ -121,6 +121,10 @@ spec:
           {{- end }}
           - name: ekka
             containerPort: 4370
+          - name: genrpc-manual
+            containerPort: 5369
+          - name: genrpc2-auto
+            containerPort: 5370
```
zmstone, line 127: there should be no need for this. ?

deploy/charts/emqx/templates/StatefulSet.yaml
```diff
@@ -121,6 +121,10 @@ spec:
           {{- end }}
           - name: ekka
             containerPort: 4370
+          - name: genrpc-manual
+            containerPort: 5369
+          - name: genrpc2-auto
+            containerPort: 5370
```
zmstone, line 127: same here

deploy/charts/emqx-enterprise/templates/StatefulSet.yaml
qzhuyan: we need it for port port_discovery = stateless in gen_rpc, yes this is confusion.

.github/workflows/run_helm_tests.yaml
```diff
@@ -89,6 +89,7 @@ jobs:
           EMQX_RPC__CACERTFILE: /opt/emqx/etc/certs/cacert.pem
           EMQX_RPC__CIPHERS: TLS_AES_256_GCM_SHA384,TLS_AES_128_GCM_SHA256
           EMQX_RPC__TLS_VERSIONS: "[tlsv1.2]"
+          EMQX_RPC__PORT_DISCOVERY: manual
```
zmstone, line 92: is this for redundance ? because it's set to manual in docker entrypoint script.
https://github.com/emqx/emqx/blob/master/deploy/docker/docker-entrypoint.sh#L52

deploy/charts/emqx-enterprise/templates/StatefulSet.yaml
qzhuyan: https://github.com/emqx/gen_rpc/blob/46dd22100391f3d87daca989cde7e2ecc52a17b5/README.md?plain=1#L117

.github/workflows/run_helm_tests.yaml
qzhuyan: Oh I thought it was auto-discovery as it is the default in the schema. I set it to manual want to test both. shall we remove it from docker-entrypoint? I don't see good reason to have it there.

qzhuyan: or we should change the schema default to manual

zmstone: The override in docker entrypoint was added long ago, we keep it for backward compatibility reasons. (so that v4 and v5 has the same port mapping convention).

We cannot change the default in schema because for RPM/DEB/TGZ installation, it's offset based port mapping. --- also kept for backward compatibility reasons.

deploy/charts/emqx-enterprise/templates/StatefulSet.yaml
zmstone: with the override set in docker entrypoint, i don't think we need this port.

apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -1151,6 +1151,7 @@ translation("emqx") ->
 translation("gen_rpc") ->
     [
         {"default_client_driver", fun tr_gen_rpc_default_client_driver/1},
+        {"ssl_client_port", fun tr_gen_rpc_ssl_client_port/1},
```
zmstone, line 1154: this is probably the only change needed.

or fix this instead: gen_rpc has a default value for ssl clients to connect to 5370
https://github.com/emqx/gen_rpc/blob/46dd22100391f3d87daca989cde7e2ecc52a17b5/src/gen_rpc.app.src#L35

but we have set the same server port on 5369 for both tcp and ssl:
https://github.com/emqx/emqx/blob/a4ab744d4ed47526f889718163830eeb9c77359e/apps/emqx_conf/src/emqx_conf_schema.erl#L798-L813

zmstone: for completeness, we should actually add `tcp_client_port` too

apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -1226,6 +1227,9 @@ collector_enabled(disabled, _) -> [].
 tr_gen_rpc_default_client_driver(Conf) ->
     conf_get("rpc.protocol", Conf).
 
+tr_gen_rpc_ssl_client_port(Conf) ->
+    conf_get("rpc.ssl_server_port", Conf).
```
zmstone, line 1231: this is wrong, because in case one configure to use `manual`, and change server port for node2, then this config on node1 should be node2's server port, but not server port of itself.

apps/emqx_conf/src/emqx_conf_schema.erl
zmstone: actually, to support `manual` strategy properly, it should not be a simple translation and user server port. 
should rather make the tcp and ssl ports configurable for the clients.

zmstone: if we force all nodes in the cluster to use the same server port, then maybe just a translation is enough.

apps/emqx_conf/src/emqx_conf_schema.erl
qzhuyan: This is actually what this PR tries to solve, without this mapping, gen_rpc use the default port 5370. In regular manual mode, the server is listening on 5369.

https://github.com/emqx/gen_rpc/blob/46dd22100391f3d87daca989cde7e2ecc52a17b5/src/gen_rpc.app.src#L35

My theory is for intra-cluster communication each node 's gen_rpc client should connect to the same port that the server is listening on.  The port specified here is a *base* port, whatever the port discovery method in use , they should get aligned. 

In case of manual mode and all nodes have rpc.ssl_server_port set to `X` emqx1 listens on `X` emqx2 listens on `X` and the client of emqx2 connects to emqx1 should connect to`X`. 

In case of stateless mode and rpc.ssl_server_port set to `X` emqx1 listens on `X+1` emqx2 listens on `X+2` and the client of emqx2 connects to emqx1 should connect to`X+1`.  the client of emqx1 connected to emqx2 should connect to `X+2`


> this is wrong, because in case one configure to use manual, and change server port for node2, then this config on node1 should be node2's server port, but not server port of itself.

do you mean node1 rpc.ssl_server_port set to `X` and node2 rpc.ssl_server_port set to `Y` where X=/=Y?
this to me is a config error.

zmstone: right, it makes no sense to configure different ports when `manual`.
basically means no way to support more than one server port when there are more than 2 nodes.

PR #11825, opened 2023-10-27
build: refactor Makefile -ct and -prop target generation

zmstone:
Most of the cases, the time spent executing `find-apps.sh` and `find-suites.sh` are wasted.
The only benefit is maybe `make` command target auto-completion.
 
After this PR, we no longer pre-generate all the `-ct` and `-prop` targets, rather generate only the one that is in the specified build target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,8 @@
+ifeq ($(DEBUG),1)
+DEBUG_INFO = $(info $1)
+else
+DEBUG_INFO = @:
+endif
 REBAR = $(CURDIR)/rebar3
 BUILD = $(CURDIR)/build
 SCRIPTS = $(CURDIR)/scripts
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -101,31 +101,48 @@ static_checks:
...
+# use a different name to store the user-set SUITES environment variable
+# because SUITES is going to mutate
+CT_SUITES := $(SUITES)
 
-.PHONY: $(APPS:%=%-ct)
+## example:
+## env SUITES=apps/appname/test/test_SUITE.erl CASES=t_foo make apps/appname-ct
 define gen-app-ct-target
 $1-ct: $(REBAR) merge-config clean-test-cluster-config
+	@export SUITES=$(CT_SUITES)
```
id, line 116: I think this is not necessary. If `SUITES` is set by the user, it will be available for `find-suites.sh` as is.

id (approved):

id (approved):

PR #11827, opened 2023-10-27
refactor: OTP_VSN and ELIXIR_VSN are no longer needed in Makefile

zmstone:
These two variables are set when we had elixir and otp version in package file names.
Now they are stale.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,17 +23,6 @@ endif
 export EMQX_DASHBOARD_VERSION ?= v1.5.0
 export EMQX_EE_DASHBOARD_VERSION ?= e1.3.0
 
-# `:=` should be used here, otherwise the `$(shell ...)` will be executed every time when the variable is used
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,6 +34,10 @@ CT_COVER_EXPORT_PREFIX ?= $(PROFILE)
 
 export REBAR_GIT_CLONE_OPTIONS += --depth=1
 
+.PHONY: nothing
```
qzhuyan, line 37: what is this? make nothing?

id (approved):

Makefile
zmstone: for testing purpose.

thalesmg: :scissors: ?

zmstone: I can do `time make nothing` to see how much speedup it is .

thalesmg (approved):

PR #11842, opened 2023-10-30
EMQX-11272 fix kafka bridge v1 import

SergeTupchiy:
Fixes EMQX-11272

### <samp>🤖 Generated by Copilot at 620b3d9</samp>

This pull request adds a feature to upgrade backup data to the latest schema, fixes a bug in the `emqx_connector` module, and introduces a new `kafka` connector type. These changes improve the data management, stability, and flexibility of the EMQ X broker.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_v2.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_v2.erl
@@ -16,7 +16,7 @@
 -module(emqx_bridge_v2).
 
 -behaviour(emqx_config_handler).
-% -behaviour(emqx_config_backup).
+-behaviour(emqx_config_backup).
 
 -include_lib("emqx/include/emqx.hrl").
 -include_lib("emqx/include/logger.hrl").
@@ -75,6 +75,11 @@
 -export([
     post_config_update/5,
     pre_config_update/3
+]).
+
+%% Data backup
+-export([
+    import_config/1
 ]).
 
 %% Compatibility API
@@ -730,6 +735,14 @@ bridge_v2_type_to_connector_type(kafka_producer) ->
     kafka_producer;
 bridge_v2_type_to_connector_type(azure_event_hub) ->
     azure_event_hub.
+
+%%====================================================================
+%% Data backup API
+%%====================================================================
+
+import_config(RawConf) ->
+    %% bridges v2 structure
+    emqx_bridge:import_config(RawConf, <<"bridges_v2">>, ?ROOT_KEY, config_key_path()).
 
 %%====================================================================
 %% Config Update Handler API
--- a/apps/emqx_connector/src/emqx_connector.erl
+++ b/apps/emqx_connector/src/emqx_connector.erl
@@ -134,7 +134,7 @@ post_config_update([?ROOT_KEY], _Req, NewConf, OldConf, _AppEnv) ->
             Result = perform_connector_changes([
                 #{action => fun emqx_connector_resource:remove/4, data => Removed},
                 #{
-                    action => fun emqx_connector_resource:create/3,
+                    action => fun emqx_connector_resource:create/4,
                     data => Added,
                     on_exception_fn => fun emqx_connector_resource:remove/4
                 },
--- a/apps/emqx_connector/src/schema/emqx_connector_schema.erl
+++ b/apps/emqx_connector/src/schema/emqx_connector_schema.erl
@@ -56,7 +56,7 @@ enterprise_fields_connectors() -> [].
 
 -endif.
 
-connector_type_to_bridge_types(kafka_producer) -> [kafka_producer];
+connector_type_to_bridge_types(kafka_producer) -> [kafka, kafka_producer];
 connector_type_to_bridge_types(azure_event_hub) -> [azure_event_hub].
 
 actions_config_name() -> <<"bridges_v2">>.
--- a/apps/emqx_management/src/emqx_mgmt_data_backup.erl
+++ b/apps/emqx_management/src/emqx_mgmt_data_backup.erl
@@ -462,11 +462,12 @@ import_cluster_hocon(BackupDir, Opts) ->
     case filelib:is_regular(HoconFileName) of
         true ->
             {ok, RawConf} = hocon:files([HoconFileName]),
-            {ok, _} = validate_cluster_hocon(RawConf),
+            RawConf1 = upgrade_raw_conf(emqx_conf:schema_module(), RawConf),
+            {ok, _} = validate_cluster_hocon(RawConf1),
             maybe_print("Importing cluster configuration...~n", [], Opts),
             %% At this point, when all validations have been passed, we want to log errors (if any)
             %% but proceed with the next items, instead of aborting the whole import operation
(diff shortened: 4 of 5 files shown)


apps/emqx_connector/src/schema/emqx_connector_schema.erl
```diff
@@ -56,7 +56,7 @@ enterprise_fields_connectors() -> [].
 
 -endif.
 
-connector_type_to_bridge_types(kafka_producer) -> [kafka_producer];
+connector_type_to_bridge_types(kafka_producer) -> [kafka_producer, kafka];
```
SergeTupchiy, line 59: @kjellwinblad, it worked fine (I've imported a  config with `bridges.kafka` config, but I haven't verified if listing aliases here follows the initial concept

kjellwinblad: kafka_producer should not be needed in this list since what we are interested in here is Bridge V1 bridges and there is no Bridge V1 bridge type called kafka_producer. Can you remove kafka_producer from this list and write a comment that explains that this function is a mapping from connector types to the corresponding Bridge V1 types?

apps/emqx_connector/src/emqx_connector.erl
```diff
@@ -134,7 +134,7 @@ post_config_update([?ROOT_KEY], _Req, NewConf, OldConf, _AppEnv) ->
             Result = perform_connector_changes([
                 #{action => fun emqx_connector_resource:remove/4, data => Removed},
                 #{
-                    action => fun emqx_connector_resource:create/3,
+                    action => fun emqx_connector_resource:create/4,
```
kjellwinblad, line 137: Good catch!

kjellwinblad: Since there does not seem to be any test case that catches this maybe we should add this but I think this can be done in another PR. Can you add a jira task for adding such a test case?

kjellwinblad (changes requested):

apps/emqx_connector/src/schema/emqx_connector_schema.erl
SergeTupchiy: Fixed

apps/emqx_connector/src/emqx_connector.erl
SergeTupchiy: Added a ticket: https://emqx.atlassian.net/browse/EMQX-11281

apps/emqx_management/src/emqx_mgmt_data_backup.erl
```diff
@@ -476,6 +477,15 @@ import_cluster_hocon(BackupDir, Opts) ->
             #{}
     end.
 
+upgrade_raw_conf(SchemaMod, RawConf) ->
+    case erlang:function_exported(SchemaMod, upgrade_raw_conf, 1) of
```
thalesmg, line 481: To ensure this is really loaded and the check for exported functions work properly (happened in the past with other similar calls :sweat_smile: ):

```suggestion
    _ = SchemaMod:module_info(),
    case erlang:function_exported(SchemaMod, upgrade_raw_conf, 1) of
```

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -731,6 +736,14 @@ bridge_v2_type_to_connector_type(kafka_producer) ->
 bridge_v2_type_to_connector_type(azure_event_hub) ->
     azure_event_hub.
 
+%%====================================================================
+%% Data backup API
+%%====================================================================
+
+import_config(RawConf) ->
+    %% bridges v2 structure
+    emqx_bridge:import_config(RawConf, <<"bridges_v2">>, ?ROOT_KEY, config_key_path()).
```
SergeTupchiy, line 745: we need these callbacks to properly detect and pretty-print changes, as each part of the config can have its own structure. That's why can't simply call `emqx_conf:update/3` in `emqx_mgmt_data_backup` module...

apps/emqx_management/src/emqx_mgmt_data_backup.erl
SergeTupchiy: Fixed :slightly_smiling_face:

kjellwinblad (approved):
LGTM

apps/emqx_connector/src/schema/emqx_connector_schema.erl
SergeTupchiy: @kjellwinblad, apparently it broke `emqx_bridge:create/3`, see: https://github.com/emqx/emqx/actions/runs/6695364058/job/18191346775?pr=11842
I've rolled back to `[kafka_producer, kafka]`.

apps/emqx_connector/src/schema/emqx_connector_schema.erl
```diff
@@ -56,7 +56,7 @@ enterprise_fields_connectors() -> [].
 
 -endif.
 
-connector_type_to_bridge_types(kafka_producer) -> [kafka_producer];
+connector_type_to_bridge_types(kafka_producer) -> [kafka, kafka_producer];
```
thalesmg, line 59: Q: since this function is only used for converting v1 bridges to v2 bridges apparently, maybe only `kafka` should be in this list?

thalesmg (approved):

apps/emqx_connector/src/schema/emqx_connector_schema.erl
thalesmg: I still need to understand better how it's used, though.  It sounds strange that both are needed in this list.

SergeTupchiy: @thalesmg , please take a look at this comment: https://github.com/emqx/emqx/pull/11842#discussion_r1376595952

thalesmg: Yes, I saw it.  Still, strange that both need to exist.  Again, I don't understand the compatibility layer very well yet.  :see_no_evil:

SergeTupchiy: Me neither.. 🙈

kjellwinblad: Yes, this is strange. We should look into why kafka_producer is needed here (seems strange). Will add an item in the TODO canvas and merge this.

kjellwinblad: @SergeTupchiy I think the problem is that you removed `kafka` which should be there since it is the old type we upgrade from. What should not be needed is `kafka_producer`. I will merge this since it should not cause any problems to have both `kafka_producer` and `kafka` in the list and we can fix it in another PR.

SergeTupchiy: I didn't remove kafka from here, first I added kafka, so it became `[kafka, kafka_producer]`, then I changed to `[kafka]` and it broke emqx_bridge:create.

PR #11862, opened 2023-11-01
 fix(bridge_v2 API): optional cascading delete operation

kjellwinblad:
This PR makes the delete HTTP API operation for Bridge V2 behave in the same way as in the Bridge V1 API.

Fixes:
https://emqx.atlassian.net/browse/EMQX-11293

### <samp>🤖 Generated by Copilot at 1e935e9</samp>

This pull request adds a new feature and a test case to the bridge HTTP API. It allows deleting a bridge and its dependent rule actions in one request by passing a query string parameter `also_delete_dep_actions`. It also fixes some issues with the swagger spec and the error handling for the delete operation.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_v2_api.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_v2_api.erl
@@ -70,7 +70,11 @@
 namespace() -> "bridge_v2".
 
 api_spec() ->
-    emqx_dashboard_swagger:spec(?MODULE, #{check_schema => true}).
+    %% TODO
+    %% The check_schema option needs to be set to false so we get the
+    %% query_string to the delete operation. We can change this once
+    %% we have fixed the schmea for the delete operation.
+    emqx_dashboard_swagger:spec(?MODULE, #{check_schema => false}).
 
 paths() ->
     [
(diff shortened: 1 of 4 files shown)


apps/emqx_bridge/src/emqx_bridge_v2_api.erl
```diff
@@ -70,7 +70,11 @@
 namespace() -> "bridge_v2".
 
 api_spec() ->
-    emqx_dashboard_swagger:spec(?MODULE, #{check_schema => true}).
+    %% TODO
+    %% The check_schema option needs to be set to false so we get the
+    %% query_string to the delete operation. We can change this once
+    %% we have fixed the schmea for the delete operation.
+    emqx_dashboard_swagger:spec(?MODULE, #{check_schema => false}).
```
thalesmg, line 77: I think we shouldn't disable the schema check here.  Having unexpected terms passed down to the handlers is confusing, and not being able to rely on the type checking/coercion makes things harder.

I think we just have to add the expected parameters to the `parameters` key in the API spec schema.

thalesmg (approved):

thalesmg (approved):

zmstone (approved):

PR #11868, opened 2023-11-02
fix: willmsg not published in takeover

### LINKED ISSUE #10551: Will messages not sent during client session takeover, violating MQTT 3.1.1 specification
### What happened?

When an existing MQTT session is taken over by a new connection reusing the same client ID, the will message for the prior connection is not being published. This was tested using MQTT 3.1.1 clients and EMQX 5.0.19.

The will message from the first connection must be published when it is closed as per the MQTT 3.1.1 spec. [search ref MQTT-3.1.2-8 in the [spec](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html)]

Using a client of your choice, open two different connections using the same client ID and will messages. As each client connects, the other will be kicked off, but the will messages will not be published.

For example, with `mosquitto_sub` as the client, I ran these two commands in two different consoles, and never saw either will message published, even as they repeatedly kicked each other:
```
mosquitto_sub -L "mqtt://un:pw@testbroker.com/sandbox" -i debug_test1 -v --will-payload SESSION1_WILL --will-topic sandbox/willtest
mosquitto_sub -L "mqtt://un:pw@testbroker.com/sandbox" -i debug_test1 -v --will-payload SESSION2_WILL --will-topic sandbox/willtest
```

This used to behave differently in EMQX 3.x, and used to publish the will messages in this session takeover.

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
sysdescr  : EMQX
version   : 5.0.19
datetime  : 2023-04-27T13:46:46.540526478+00:00
uptime    : 38 seconds
```

</details>

### OS version

<details>

```console
$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
$ uname -a
Linux c48122d7b7e6 5.19.0-40-generic #41~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar 31 16:00:14 UTC 2 x86_64 GNU/Linux
```
</details>

### Log files

<details>

[node_dump_20230427_142450.tar.gz](https://github.com/emqx/emqx/files/11344792/node_dump_20230427_142450.tar.gz)

</details>

qzhuyan:
Fixes #10551

### <samp>🤖 Generated by Copilot at b497f84</samp>

This pull request fixes a bug with will messages not being published after session takeover, adds a feature to gracefully close the transport layer when shutting down a connection, refactors and simplifies the code for handling will messages and channel termination, and updates the test suite to cover different scenarios for session takeover. It also adds a change log file `changes/ce/fix-11868.en.md` to document the bug fix.

## PR Checklist


apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -78,30 +93,144 @@ t_takeover(_) ->
...
+            [{fun publish_msg/2, [Msg]} || Msg <- Client2Msgs] ++
+            [
+                {
+                    fun(CTX) ->
+                        timer:sleep(1000),
+                        CTX
+                    end,
+                    []
+                }
+            ],
```
keynslug, line 201: Nit. Maybe just do `?drainMailbox(1000)` further down instead?

apps/emqx/src/emqx_channel.erl
```diff
@@ -1137,16 +1137,14 @@ handle_call(
     kick,
     Channel = #channel{
         conn_state = ConnState,
-        will_msg = WillMsg,
-        clientinfo = ClientInfo,
         conninfo = #{proto_ver := ProtoVer}
     }
 ) ->
-    (WillMsg =/= undefined) andalso publish_will_msg(ClientInfo, WillMsg),
+    Channel0 = maybe_publish_will_msg(kick, Channel),
```
keynslug, line 1143: Q: Probably out-of-scope of this PR, but still. If the Will message has a delay, in case of MQTTv5 we'll arm the timer in `maybe_publish_will_msg/2` but then just shut down the connection process right away, right?

keynslug, line 1143: Q: Also kinda confused why this case needs separate handling at the time of `handle_call/2`? The first impression is the same would be achieved by just removing this code and separate clause with `terminate({shutdown, kicked}, Channel)` from `terminate/2`.

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -169,3 +306,30 @@ payload(I) ->
...
+    ),
+    Filtered = [
+        Msg
+     || #{payload := P} = Msg <- List,
+        P =/= Payload
+    ],
+    {IsWill, Filtered}.
+
+%% @doc assert emqtt *client* process exits as expected.
+assert_client_exit(Pid, v5) ->
```
keynslug, line 331: Tiny nit. Looks like it should be either `assert_client_takenover(Pid, v5)` or `assert_client_exit(Pid, takenover, v5)`, otherwise sounds too broad.

keynslug (commented):
💯

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -169,3 +306,30 @@ payload(I) ->
...
+            (_) -> false
+        end,
+        List
+    ),
+    Filtered = [
+        Msg
+     || #{payload := P} = Msg <- List,
+        P =/= Payload
+    ],
+    {IsWill, Filtered}.
```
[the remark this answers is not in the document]
qzhuyan: I want to ensure the Payload is the willmsg payload. comparing length does not feel safe.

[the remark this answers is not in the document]
keynslug: Why? `Filtered` will have strictly less elements than `List` only if `List` has at least one `Msg` with `P =:= Payload`.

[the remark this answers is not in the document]
qzhuyan: oh, I miss read it, thanks for clarifying.

apps/emqx/src/emqx_channel.erl
qzhuyan: For MQTTv5,
If the session is not timeout, it won't shutdown connection process right away.

If session timeout first, then connection process shutdown will lead to willmsg publish with reason ({shutdown, expired})
if willmsg timeout first when the connection process is still alive, connection process will publish the willmsg and set willmsg to null.

Such cases are covered in paho python test.

qzhuyan: I am not 100% sure if kick also means delete session here. 
I will double check in the spec.

keynslug: I probably need to clarify that the question is specifically about `handle_call(kick, ...)`, which is probably not part of the paho test suite.

So imagine that the Will Message has 5 seconds delay. What will happen here AFAICT:
1. Call to `maybe_publish_will_msg/2` will arm will message timer to fire 5 seconds after.
2. Call to `shutdown(kicked, ok, ...)` will make `handle_call/2` to return with `{shutdown, kicked, ok, ...}`.
3. Outer `emqx_connection:handle_call/3` will translate that to `{stop, {shutdown, kicked}, ...}`.
4. This will effectively translate to `emqx_channel:terminate({shutdown, kicked}, ...)` call.
5. This clause in `terminate/2` does nothing with will messages, so the process will just exit without sending the Will Message.

Session Expiry Interval doesn't seem to affect anything in this codepath.

qzhuyan: right.

 here it does not even check if the connection is allowed. (passing REAUTH ??) and kick means closing connection and deleting session (session ends).  according to the spec:

```
The Server SHOULD publish Will Messages promptly after the Network Connection is closed and the Will Delay Interval has passed, or when the Session ends, whichever occurs first. 
```

shall I just ignore 'promptly' and publish it in `terminate/2` after send 'DISCONNECT' to client?  hum... by sending 'DISCONNECT' the conn process here may stuck for a while.

I will see if I could handle it in maybe_publish_will_msg by checking the conn_state and the reason....

apps/emqx/src/emqx_channel.erl
```diff
@@ -1430,17 +1428,9 @@ terminate(normal, Channel) ->
...
-    Reason =:= takenover
-->
-    run_terminate_hook(Reason, Channel);
-terminate(Reason, Channel = #channel{clientinfo = ClientInfo, will_msg = WillMsg}) ->
-    %% since will_msg is set to undefined as soon as it is published,
-    %% if will_msg still exists when the session is terminated, it
-    %% must be published immediately.
-    WillMsg =/= undefined andalso publish_will_msg(ClientInfo, WillMsg),
-    run_terminate_hook(Reason, Channel).
+terminate(Reason, Channel) ->
```
qzhuyan, line 1431: I do think we need a type spec for the `Reason`.

apps/emqx/src/emqx_channel.erl
```diff
@@ -2111,16 +2101,43 @@ ensure_disconnected(
...
-            ok = publish_will_msg(ClientInfo, WillMsg),
-            Channel#channel{will_msg = undefined};
-        I ->
-            ensure_timer(will_message, timer:seconds(I), Channel)
+maybe_publish_will_msg(
+    _Reason,
+    Channel = #channel{conninfo = #{proto_ver := ?MQTT_PROTO_V3}, will_msg = WillMsg}
+) ->
+    %% Unconditionally publish will message for MQTT 3.1.1
+    ok = publish_will_msg(Channel#channel.clientinfo, WillMsg),
```
qzhuyan, line 2113: since we moved all the logic into this fun, we should also check the conn_state

qzhuyan:
I did some updates. but need to dig more about the possible reasons of shutdown and connection states where willmsg publishing is allowed.

apps/emqx/src/emqx_channel.erl
qzhuyan: I think "Session ends" here means EMQX connection process terminates due to `{shutdown, expired}`( session expired) OR `{shutdown, kicked}` (session deleted).

But I don't know if there are more 'reasons' casues  "Session ends".

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -224,6 +224,68 @@ t_takeover_willmsg_clean_session(Config) ->
...
+    {IsWill1, ReceivedNoWill0} = filter_payload(Received, <<"willpayload_delay10">>),
+    {IsWill2, ReceivedNoWill} = filter_payload(ReceivedNoWill0, <<"willpayload_2">>),
+    assert_messages_missed(AllMsgs, ReceivedNoWill),
+    assert_messages_order(AllMsgs, ReceivedNoWill),
+    %% THEN: payload <<"willpayload_delay10">> should be published without delay.
+    ?assert(IsWill1),
+    ?assertNot(IsWill2),
+    emqtt:stop(CPid2),
+    emqtt:stop(CPidSub),
+    ?assert(not is_process_alive(CPid1)),
```
keynslug, line 286: Tiny nit. The `assert_client_exit/3` call several lines before returns only when `CPid1` process exits, so this line seems kinda unnecessary.

apps/emqx/test/emqx_shared_sub_SUITE.erl
```diff
@@ -936,6 +936,7 @@ t_session_kicked({init, Config}) when is_list(Config) ->
 t_session_kicked({'end', Config}) when is_list(Config) ->
     emqx_config:put_zone_conf(default, [mqtt, max_inflight], 0);
 t_session_kicked(Config) when is_list(Config) ->
+    emqx_logger:set_log_level(debug),
```
keynslug, line 939: Nit. Leftovers?

apps/emqx/src/emqx_channel.erl
```diff
@@ -2099,9 +2118,19 @@ ensure_disconnected(
 %%--------------------------------------------------------------------
 %% Maybe Publish will msg
 -spec maybe_publish_will_msg(Reason, channel()) -> channel() when
-    Reason :: kick | sock_closed | {shutdown, atom()}.
+    Reason :: takenover | kicked | discarded | expired | sock_closed | {shutdown, atom()}.
```
keynslug, line 2121: Q: Why there's no `not_authorized` that's being handled in one of the following clauses?

apps/emqx/src/emqx_channel.erl
```diff
@@ -2112,30 +2141,51 @@ maybe_publish_will_msg(
...
+        %% Unsure...
+        Reason =:= {shutdown, internal_error} orelse
+        Reason =:= kicked
 ->
-    %% Must publish now without delay and cancel the will message timer.
+    %% For the cases that session MUST be gone.
+    %% a. expired (session expired)
+    %% c. discarded (Session ends because another process starts new session with the same clientid)
+    %% b. kicked. (kicked by operation)
+    %% d. internal_error (maybe not recoverable)
```
keynslug, line 2154: Agree. It seems a bit unsafe to assume that in the event of hitting internal error publishing should still work.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1428,19 +1436,17 @@ terminate(_, #channel{conn_state = idle} = _Channel) ->
...
 ->
-    run_terminate_hook(Reason, Channel);
-terminate(Reason, Channel = #channel{clientinfo = ClientInfo, will_msg = WillMsg}) ->
-    %% since will_msg is set to undefined as soon as it is published,
-    %% if will_msg still exists when the session is terminated, it
-    %% must be published immediately.
-    WillMsg =/= undefined andalso publish_will_msg(ClientInfo, WillMsg),
-    run_terminate_hook(Reason, Channel).
+    Channel1 = maybe_publish_will_msg(Reason, Channel),
+    run_terminate_hook(Reason, Channel1);
```
keynslug, line 1446: Nit. This clause's logic is identical to the next catch-all clause, the only difference is taking the inner atom from `shutdown` tuple. I'd argue this makes the shutdown logic harder to follow.

In general, I suspect that one thing that could help to make things simpler to follow is separating `maybe_publish_will_msg` into a set of 2 functions: one is for _scheduling_ willmsg publishing (e.g. essentially called only when `sock_closed`), another one is for deciding what to do on channel _termination_ (when there's no point to schedule anything). This should also eliminate the need to do something explicitly on `handle_call(kick, ...)`, i.e. [here](https://github.com/emqx/emqx/blob/6f779dc663e0aab9108aa0a85fe250bff333997e/apps/emqx/src/emqx_channel.erl#L1153).

keynslug (commented):
👍🏼

apps/emqx/src/emqx_channel.erl
qzhuyan: > Nit. This clause's logic is identical to the next catch-all clause, the only difference is taking the inner atom from shutdown tuple. I'd argue this makes the shutdown logic harder to follow.

I thought the same and I did remove it until I found `run_terminate_hook` relies on the Reason (kicked) instead of `{shutdown, kicked}`, the hooks are exposed to the users so I think it better to keep it.

qzhuyan: > In general, I suspect that one thing that could help to make things simpler to follow is separating maybe_publish_will_msg into a set of 2 functions: one is for scheduling willmsg publishing (e.g. essentially called only when sock_closed), another one is for deciding what to do on channel termination (when there's no point to schedule anything). This should also eliminate the need to do something explicitly on handle_call(kick, ...), i.e. [here](https://github.com/emqx/emqx/blob/6f779dc663e0aab9108aa0a85fe250bff333997e/apps/emqx/src/emqx_channel.erl#L1153).

I found scheduling willmsg publishing will not always work when the process terminates before the timer get fired. 
Simple fix would be to just extend the process lifetime but that is costly and it cannot be done in `terminate` without changing the whole call stack.

keynslug: Ah, good catch. However, it also means that terminate hook now will be called with `expired`, instead of `{shutdown, expired}` as before. Although I'm not sure if former behavior is important to preserve here.

keynslug: > I found scheduling willmsg publishing will not always work when the process terminates before the timer get fired.

I mean, that seems the reason why the logic is hard to follow: currently `maybe_publish_will_msg` is called both from sock_closed context (where scheduling timers makes sense) and terminate context (where scheduling timers doesn't make sense). Yet some of its codepaths end up with timers being scheduled anyway, which is confusing: it's very hard to tell in which context they are scheduled. Thus it seems worthwhile to try to separate this into `maybe_schedule_will_msg/1` and `maybe_publish_will_msg_on_terminate/2`, at least on the first look.

apps/emqx/src/emqx_channel.erl
qzhuyan: I just list them, may miss something, the entire call path, call stack is hard to follow.

apps/emqx/src/emqx_channel.erl
qzhuyan: Now I removed it, look unnecessary now.

apps/emqx/test/emqx_takeover_SUITE.erl
qzhuyan: I will leave it it is test code, assertions just helps.

apps/emqx/src/emqx_channel.erl
```diff
@@ -2111,15 +2121,146 @@ ensure_disconnected(
...
+    Channel = #channel{
+        clientinfo = ClientInfo,
+        conninfo = #{clientid := ClientId},
+        will_msg = WillMsg
+    }
+) when
+    Reason =:= expired orelse
+        Reason =:= discarded orelse
+        Reason =:= kicked orelse
+        Reason =:= ?chan_terminating orelse
```
qzhuyan, line 2192: this keeps the old behavior.

apps/emqx/src/emqx_channel.erl
qzhuyan: I found this in spec:
```
In the case of a Server shutdown or failure, the Server MAY defer publication of Will Messages until a subsequent restart. If this happens, there might be a delay between the time the Server experienced failure and when the Will Message is published.
```
So I think it is ok to publish/not publish/delay publish will when server error.

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -78,37 +113,750 @@ t_takeover(_) ->
...
+    ?assertNotEqual([], ReceivedNoWill),
+    emqtt:stop(CPidSub),
+    ?assert(not is_process_alive(CPid1)),
+    ?assert(not is_process_alive(CPid2)),
+    ok.
+
+t_takeover_session_then_abnormal_disconnect(Config) ->
+    ?config(mqtt_vsn, Config) =:= v5 orelse ct:fail("MQTTv5 Only"),
+    process_flag(trap_exit, true),
+    emqx_logger:set_log_level(debug),
```
qzhuyan, line 599: remove

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -78,37 +113,750 @@ t_takeover(_) ->
...
+    ?assertEqual([], ReceivedNoWill1),
+    emqtt:stop(CPidSub),
+    ?assert(not is_process_alive(CPid1)),
+    ?assert(not is_process_alive(CPid2)),
+    ok.
+
+t_takeover_session_then_abnormal_disconnect_2(Config) ->
+    ?config(mqtt_vsn, Config) =:= v5 orelse ct:fail("MQTTv5 Only"),
+    process_flag(trap_exit, true),
+    emqx_logger:set_log_level(debug),
```
qzhuyan, line 665: remove

apps/emqx/src/emqx_channel.erl
```diff
@@ -2111,15 +2121,149 @@ ensure_disconnected(
...
+        conninfo = #{clientid := ClientId}
+    }
+) ->
+    %% Handles other Unknown Reasons.
+    case will_delay_interval(WillMsg) of
+        0 ->
+            ?tp(debug, willmsg_other_publish, #{clientid => ClientId, reason => Reason}),
+            _ = publish_will_msg(ClientInfo, WillMsg),
+            remove_willmsg(Channel);
+        I when I > 0 ->
```
qzhuyan, line 2265: add comments why we don't need to take care of the session expiry here.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1398,19 +1411,16 @@ terminate(_, #channel{conn_state = idle} = _Channel) ->
     ok;
 terminate(normal, Channel) ->
     run_terminate_hook(normal, Channel);
-terminate({shutdown, kicked}, Channel) ->
-    run_terminate_hook(kicked, Channel);
 terminate({shutdown, Reason}, Channel) when
-    Reason =:= discarded;
-    Reason =:= takenover
+    Reason =:= expired orelse
```
tigercl, line 1415: Reason =:= expired orelse
Reason =:= takenover orelse
Reason =:= kicked orelse
Reason =:= discarded

apps/emqx/src/emqx_channel.erl
```diff
@@ -2188,15 +2198,152 @@ ensure_disconnected(
...
+maybe_publish_will_msg(
+    _Reason,
+    Channel = #channel{
+        conn_state = ConnState,
+        conninfo = #{clientid := ClientId}
+    }
+) when
+    ConnState =:= idle orelse
+        ConnState =:= connecting orelse
+        ConnState =:= reauthenticating
```
tigercl, line 2248:
```
ConnState =:= idle orelse
ConnState =:= connecting orelse
ConnState =:= reauthenticating
```

apps/emqx/src/emqx_channel.erl
```diff
@@ -2188,15 +2198,152 @@ ensure_disconnected(
...
+->
+    %% Wrong state to publish
+    ?tp(debug, willmsg_wrong_state, #{clientid => ClientId}),
```
tigercl, line 2251: I think EMQX should also send the will message if re-authentication fails causing the connection to be closed

apps/emqx/src/emqx_channel.erl
qzhuyan: they are same as my changes?

code formating is done by `make fmt`, not me.

apps/emqx/src/emqx_channel.erl
```diff
@@ -2227,15 +2237,152 @@ session_disconnect(_ClientInfo, _ConnInfo, undefined) ->
 
 %%--------------------------------------------------------------------
 %% Maybe Publish will msg
-
-maybe_publish_will_msg(Channel = #channel{will_msg = undefined}) ->
+%% @doc May publish will message [MQTT-3.1.2-8]
```
zmstone, line 2240:
```suggestion
%% @doc Maybe publish will message [MQTT-3.1.2-8]
```

apps/emqx/src/emqx_channel.erl
```diff
@@ -2227,15 +2237,152 @@ session_disconnect(_ClientInfo, _ConnInfo, undefined) ->
...
     Channel;
-maybe_publish_will_msg(Channel = #channel{clientinfo = ClientInfo, will_msg = WillMsg}) ->
+maybe_publish_will_msg(_Reason, Channel = #channel{will_msg = undefined}) ->
+    %% No will message to publish
+    Channel;
+maybe_publish_will_msg(
+    _Reason,
+    Channel = #channel{
+        conn_state = ConnState,
+        conninfo = #{clientid := ClientId}
```
zmstone, line 2282: Is it guaranteed to have a properly initialized conninfo?
the process may shutdown before gets assigned with a clientid.

e.g. CONNECT packet might be delayed processing due to being rate limited, and the process is to be shutdown for whatever reason.

qzhuyan: I think L2275 covers that?

apps/emqx/src/emqx_channel.erl
zmstone: I don't agree.
EMQX should never publish a will for unauthenticated or unauthorised clients.

qzhuyan: @zmstone I think we meant the reauth after conencted (authenticated and authorised) because session state won't be updated with willmsg if auth during connect is failed. 

according to the spec [MQTT-3.1.2-8] line 617-618, if the client connection isn't closed with Reason 0 (NOERROR), server should try to publish willmsg.

apps/emqx/src/emqx_channel.erl
```diff
@@ -2227,15 +2237,152 @@ session_disconnect(_ClientInfo, _ConnInfo, undefined) ->
...
 %%--------------------------------------------------------------------
 %% Maybe Publish will msg
-
-maybe_publish_will_msg(Channel = #channel{will_msg = undefined}) ->
+%% @doc May publish will message [MQTT-3.1.2-8]
+%%   When willmsg presents the decision whether or when to publish the Will Message are effected by
+%%   the followings:
+%%   - connecion state
+%%   - If it is MQTT normal disconnection (RC: 0)
+%%   - If it is MQTT normal disconnection (RC: 4)
```
qzhuyan, line 2245: also Rc != 0

apps/emqx/src/emqx_channel.erl
```diff
@@ -2227,15 +2237,152 @@ session_disconnect(_ClientInfo, _ConnInfo, undefined) ->
...
+        | expired
+        %% Connection is terminating because of socket close/error
+        | sock_closed
+        %% Connection is terminating with Reasons
+        | {shutdown, atom()}
+        %% Connection is terminating, delay willmsg publish is impossible.
+        | ?chan_terminating
+        %% Connection is terminating because of normal MQTT disconnection, implies delete the session.
+        | normal.
+maybe_publish_will_msg(normal, Channel) ->
```
qzhuyan, line 2272: Spec says that if it is a normal disconnect(rc=0) from client, willmsg MUST not be published. MQTT-3.14.4-3

If server send disconnect(rc=0), willmsg should still be sent. (unsure in which case server should use RC=0)

qzhuyan: removed

apps/emqx/src/emqx_channel.erl
```diff
@@ -1134,16 +1146,14 @@ handle_call(
     kick,
     Channel = #channel{
         conn_state = ConnState,
-        will_msg = WillMsg,
-        clientinfo = ClientInfo,
         conninfo = #{proto_ver := ProtoVer}
     }
 ) ->
-    (WillMsg =/= undefined) andalso publish_will_msg(ClientInfo, WillMsg),
+    Channel0 = maybe_publish_will_msg(kicked, Channel),
```
qzhuyan, line 1152: I think it can be deferred to terminate/2

apps/emqx/src/emqx_channel.erl
```diff
@@ -1157,7 +1167,8 @@ handle_call(
             shutdown(kicked, ok, Channel1)
     end;
 handle_call(discard, Channel) ->
-    disconnect_and_shutdown(discarded, ok, Channel);
+    Channel0 = maybe_publish_will_msg(discarded, Channel),
```
qzhuyan, line 1170: I think it can be deferred to terminate/2

qzhuyan: Nope, we have states to check

apps/emqx/src/emqx_channel.erl
qzhuyan: anyway, they are  init from here
https://github.com/emqx/emqx/blob/bb6a95bae809cbb13a13ce17307b018fc1729696/apps/emqx/src/emqx_channel.erl#L246

called from:
https://github.com/emqx/emqx/blob/bb6a95bae809cbb13a13ce17307b018fc1729696/apps/emqx/src/emqx_connection.erl#L349

zmstone:
kick should not trigger will message due to security reason.

zmstone (approved):

Issue #10694, opened 2023-05-14
Most of the Rest API under under `/api/v5​/gateways​/{name}​/listeners​/{id}​/authentication/`  have some problems


xiaobaxiang:
### What happened?

I want add some exproto gateways authentication user info.
I test the `/api/v5​/gateways​/{name}​/listeners​/{id}​/authentication` get the response code 204.
`/api/v5​/gateways​/{name}​/listeners​/{id}​/authentication​/users` ,`/api/v5​/gateways​/{name}​/listeners​/{id}​/authentication​/users` get response code 404.the response:
```json
{
  "code": "RESOURCE_NOT_FOUND",
  "message": "Resource not found. path: gateway.exproto.listeners.tcp.default.authentication"
}
```

### What did you expect to happen?

I can add some  exproto gateways authentication user info.

### How can we reproduce it (as minimally and precisely as possible)?

Test on the swagger document. `/api-docs/index.html`

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
# paste output here
```
sysdescr  : EMQX
version   : 5.0.23
datetime  : 2023-05-14T17:12:03.537834935+08:00
uptime    : 8 days, 22 hours, 24 minutes, 25 seconds
</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
NAME="Alibaba Cloud Linux"
VERSION="3 (Soaring Falcon)"
ID="alinux"
ID_LIKE="rhel fedora centos anolis"
VERSION_ID="3"
PLATFORM_ID="platform:al8"
PRETTY_NAME="Alibaba Cloud Linux 3 (Soaring Falcon)"
ANSI_COLOR="0;31"
HOME_URL="https://www.aliyun.com/"

$ uname -a
# paste output here
Linux iZ8vbg1vhlai3gpx5v8cg6Z 5.10.134-13.1.al8.x86_64 #1 SMP Mon Feb 6 14:54:50 CST 2023 x86_64 x86_64 x86_64 GNU/Linux

# On Windows:
C:\> wmic os get Caption, Version, BuildNumber, OSArchitecture
# paste output here
```

</details>

### Log files

<details>

</details>

xiaobaxiang:
yes,I do.
call `api/v5/gateways/exproto` returned the following:
```json
{
  "created_at": "2023-05-15T18:25:35.300+08:00",
  "name": "exproto",
  "started_at": "2023-05-15T18:25:35.311+08:00",
[...]

PR #11881, opened 2023-11-04
fix(gateway): api

cisiqo:
Fixes [https://github.com/emqx/emqx/issues/10694](https://github.com/emqx/emqx/issues/10694)

Also i think this type of api can be discarded because the `authentication` has been moved out of the `listeners `.

### <samp>🤖 Generated by Copilot at bf2ad73</samp>

Fix HTTP gateway crash when authentication chain is not configured. Add try-catch block to `emqx_gateway_http.erl` to use default authentication chain if config path is not found.

## PR Checklist

### DIFF
--- a/apps/emqx_gateway/src/emqx_gateway_http.erl
+++ b/apps/emqx_gateway/src/emqx_gateway_http.erl
@@ -251,10 +251,14 @@ authn(GwName, ListenerId) ->
     {_, Type, Name} = emqx_gateway_utils:parse_listener_id(ListenerId),
     Path = [gateway, GwName, listeners, Type, Name, ?AUTHN],
     ChainName = emqx_gateway_utils:listener_chain(GwName, Type, Name),
-    wrap_chain_name(
-        ChainName,
-        emqx_utils_maps:jsonable_map(emqx:get_raw_config(Path))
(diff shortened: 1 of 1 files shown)


apps/emqx_gateway/src/emqx_gateway_http.erl
```diff
@@ -251,10 +251,14 @@ authn(GwName, ListenerId) ->
...
-        ChainName,
-        emqx_utils_maps:jsonable_map(emqx:get_raw_config(Path))
-    ).
+    try
+        wrap_chain_name(
+            ChainName,
+            emqx_utils_maps:jsonable_map(emqx:get_raw_config(Path))
+        )
+    catch
+        _:{config_not_found, _} -> authn(GwName)
```
HJianBo, line 260: We can't just return a Gateway-level authenticator as a Listener-level authenticator. That's not correct.

HJianBo (changes requested):

apps/emqx_gateway/src/emqx_gateway_http.erl
cisiqo: OK

PR #11882, opened 2023-11-05
emqx_bridge_v2 module improvements

kjellwinblad:
This PR improves the readability of the  emqx_bridge_v2 module:
* Adds specs to the most important public facing functions
* Changes names to public compatibility layer functions so they are all prefixed with bridge_v1
* removes the Kafka specific knowledge from emqx_bridge_v2 and makes it possible to add new Bridge V2 bridges without modifying the emqx_bridge application.

### <samp>🤖 Generated by Copilot at de726ed</samp>

This pull request refactors and improves the compatibility layer API for bridge v1 and v2 configurations. It adds specs, types, and validation functions for bridge v2, and renames some functions for clarity and consistency. It also updates some specs and tests to use the `resource_status()` type.

## PR Checklist


apps/emqx_bridge/src/proto/emqx_bridge_proto_v5.erl
```diff
@@ -49,7 +49,7 @@ introduced_in() ->
 list_bridges_on_nodes(Nodes) ->
     erpc:multicall(Nodes, emqx_bridge, list, [], ?TIMEOUT).
 
--type key() :: atom() | binary() | [byte()].
```
thalesmg, line 52: I think we'll need to keep this typespec untouched to please the BPAPI checks.

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
```diff
@@ -708,3 +719,17 @@ kafka_ext_header_value_validator(Value) ->
...
             }
     end.
+
+-include_lib("emqx_bridge/include/emqx_bridge_v2_register.hrl").
+
+register_bridge_v2() ->
+    emqx_bridge_v2_register_bridge_type(#{
+        %% Should be provided by all bridges. Even if the bridge_v2_type_name is
+        %% the same as the bridge_v1_type_named.
+        'bridge_v1_type_name' => kafka,
```
sstrigler, line 729: What if we have new bridges that never had a v1?

kjellwinblad: Then they will still have to have a v1 type name for the compatibility layer. It can be the same as the v2 type name...

kjellwinblad: We could of coarse have a nicer interface that have this assumption and don't force the user to provide both when they are the same...

apps/emqx_bridge/include/emqx_bridge_v2_register.hrl
```diff
@@ -0,0 +1,153 @@
...
+    'schema_struct_field' := atom() | binary()
+}) -> ok.
+emqx_bridge_v2_register_bridge_type(BridgeTypeInfo) ->
+    try
+        %% We must prevent overwriting so we take a lock when writing to persistent_term
+        global:set_lock(
+            {
+                internal_emqx_bridge_v2_persistent_term_info_key(),
+                internal_emqx_bridge_v2_persistent_term_info_key()
+            },
```
thalesmg, line 24: Just thinking out loud: I wonder if we could avoid locking if each module has its own exclusive key (e.g.: `{common_info_key(), bridge_type()}`), and then coalesced to a single value when starting the `emqx_bridge` up.

apps/emqx_bridge/include/emqx_bridge_v2_register.hrl
```diff
@@ -0,0 +1,153 @@
...
+    end,
+    ok.
+
+internal_register_bridge_type_with_lock(BridgeTypeInfo) ->
+    InfoMap0 = persistent_term:get(internal_emqx_bridge_v2_persistent_term_info_key()),
+    %% The Bridge V1 type is also a bridge V2 type due to backwards compatibility
+    InfoMap1 = emqx_utils_maps:deep_force_put(
+        [
+            bridge_v2_type_names,
+            maps:get(bridge_v1_type_name, BridgeTypeInfo)
```
thalesmg, line 60: This could be absent for bridges created after this refactoring effort.

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -86,25 +108,101 @@
...
+        %% We must prevent overwriting so we take a lock when writing to persistent_term
+        global:set_lock(
+            {
+                internal_emqx_bridge_v2_persistent_term_info_key(),
+                internal_emqx_bridge_v2_persistent_term_info_key()
+            },
+            [node()],
+            infinity
+        ),
+        internal_maybe_create_initial_bridge_v2_info_map()
```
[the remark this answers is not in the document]
kjellwinblad: This is needed in case there are no bridges V2 that register themselves.

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -794,12 +906,20 @@ connector_type(Type) ->
...
+    case maps:get(Type, BridgeV2TypeToConnectorTypeMap, undefined) of
+        undefined -> bridge_v2_type_to_connector_type_old(Type);
+        ConnectorType -> ConnectorType
+    end.
+
+% bridge_v2_type_to_connector_type_old(kafka) ->
+%     %% backward compatible
+%     kafka_producer;
+% bridge_v2_type_to_connector_type_old(kafka_producer) ->
+%     kafka_producer;
```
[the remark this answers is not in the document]
kjellwinblad: Intentionally left it here while trying to get the CI task to go through to make it easy to see what the function did originally. Will remove of coarse.

apps/emqx_bridge/include/emqx_bridge_v2_register.hrl
kjellwinblad: The name can be the same, but yes it makes sense to let users register a bridge without providing a v1_type_name and which would mean that the v1 and v2 type names are the same.

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -86,25 +108,101 @@
...
+}.
+
+-type bridge_v2_type() :: binary() | atom() | [byte()].
+-type bridge_v2_name() :: binary() | atom() | [byte()].
+
+%%====================================================================
+
+%%====================================================================
+
+-include_lib("emqx_bridge/include/emqx_bridge_v2_register.hrl").
```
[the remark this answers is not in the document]
kjellwinblad: Yes, it is included so we can call: `internal_maybe_create_initial_bridge_v2_info_map()` which we need to do in case no bridges are registering themselves.

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
kjellwinblad: > New bridges after this refactoring might never have had a v1 equivalent. For example, Confluent (which is a shim over Kafka like AEH).

See comments above. Even if they where created after the introduction of v2 they would still need a v1 type for the compatibility layer (even if they are the same). The alternative is to hide certain bridges from the compatibility layer but I don't think that will give us much benefits.

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -26,6 +35,15 @@
 
 -define(ROOT_KEY, bridges_v2).
 
+-on_load(on_load/0).
+
+%% Getting registered bridge schemas:
+
+-export([
+    bridge_v2_type_to_schame_stuct_field/1,
```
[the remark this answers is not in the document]
kjellwinblad: Fixed, thanks!

apps/emqx_bridge/include/emqx_bridge_v2_register.hrl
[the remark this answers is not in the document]
kjellwinblad: Fixed, thanks!

apps/emqx_bridge/include/emqx_bridge_v2_register.hrl
kjellwinblad: We could do that but I think that could be a future improvement in that case. Unless the single node locking in `global` is extremely expensive (which I don't see any reason for it to be) the performance benefits if any should be very minimal. Also, the the locked critical section is so short so I think the the wait on lock overhead will be minimal even when we have refactored all 20 bridges (I'm not even sure module loading is happening in parallel).

apps/emqx_bridge/include/emqx_bridge_v2_register.hrl
```diff
@@ -0,0 +1,169 @@
...
+%% might not be loaded when the bridge module is loaded.
+-spec emqx_bridge_v2_register_bridge_type(#{
+    %% Should be provided by all bridges. Even if the bridge_v2_type_name is
+    %% the same as the bridge_v1_type_named.
+    'bridge_v1_type_name' := atom(),
+    'bridge_v2_type_name' := atom(),
+    'connector_type' := atom(),
+    'schema_module' := atom(),
+    'schema_struct_field' := atom() | binary()
+}) -> ok.
```
zmstone, line 32: how about just a module, and everything else should be returned by the modules APIs
```suggestion
-spec emqx_bridge_v2_register_module(module()) -> ok.
```

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
thalesmg: I think new bridges shouldn't have a v1 equivalent, since we're moving towards actions and bridges v1 is now a legacy API.  :thinking:

apps/emqx_bridge/include/emqx_bridge_v2_register.hrl
thalesmg: I mean that new bridges wouldn't have a V1 counterpart (or shouldn't, I think).

apps/emqx_bridge/include/emqx_bridge_v2_register.hrl
kjellwinblad: Good suggestion. Fixed! Rebased due to conflicts. The three last commits contains the real changes.

apps/emqx_bridge/src/emqx_bridge_v2.erl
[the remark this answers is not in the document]
kjellwinblad: Fixed

apps/emqx_bridge/include/emqx_bridge_v2_register.hrl
kjellwinblad: Now, the V1 name has been made optional.

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
kjellwinblad: The V1 name is now optional.

apps/emqx_bridge/src/emqx_action_info.erl
```diff
@@ -0,0 +1,197 @@
+%%
```
thalesmg, line 1:
```suggestion
%%--------------------------------------------------------------------
%% Copyright (c) 2023 EMQ Technologies Co., Ltd. All Rights Reserved.
%%
```

apps/emqx_bridge_azure_event_hub/src/emqx_bridge_azure_event_hub_action_info.erl
thalesmg, line 1: Missing copyright header

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
```diff
@@ -526,7 +526,18 @@ fields(consumer_kafka_opts) ->
 fields(resource_opts) ->
     SupportedFields = [health_check_interval],
     CreationOpts = emqx_resource_schema:create_opts(_Overrides = []),
-    lists:filter(fun({Field, _}) -> lists:member(Field, SupportedFields) end, CreationOpts).
+    lists:filter(fun({Field, _}) -> lists:member(Field, SupportedFields) end, CreationOpts);
+fields(bridge_v2_field) ->
```
thalesmg, line 530: Nit (since this is a new thing):

```suggestion
fields(action_field) ->
```

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
```diff
@@ -526,7 +526,18 @@ fields(consumer_kafka_opts) ->
...
+    {kafka_producer,
+        mk(
+            hoconsc:map(name, ref(emqx_bridge_kafka, kafka_producer_action)),
+            #{
+                desc => <<"Kafka Producer Bridge V2 Config">>,
```
thalesmg, line 535:
```suggestion
                desc => <<"Kafka Producer Action Config">>,
```

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
```diff
@@ -526,7 +526,18 @@ fields(consumer_kafka_opts) ->
...
+                required => false
+            }
+        )};
+fields(action) ->
+    fields(bridge_v2_field).
```
thalesmg, line 540:
```suggestion
    fields(action_field).
```

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_action_info.erl
thalesmg, line 1: Missing copyright header

apps/emqx_conf/src/emqx_conf.erl
```diff
@@ -151,6 +151,9 @@ reset(Node, KeyPath, Opts) ->
 %% @doc Called from build script.
 %% TODO: move to a external escript after all refactoring is done
 dump_schema(Dir, SchemaModule) ->
+    %% TODO: Load all apps instead of only emqx_dashboard
+    %% as this will help schemas that searches for apps with
+    %% relevant schema definitions
```
thalesmg, line 156: Q: would do the injecting in [`emqx_conf_schema:roots/0`](https://github.com/emqx/emqx/blob/5aeb1feada4afedc15cacf21fb68c2495a4aedec/apps/emqx_conf/src/emqx_conf_schema.erl#L96) like auth does be a possible solution?

apps/emqx_bridge/src/emqx_action_info.erl
```diff
@@ -0,0 +1,197 @@
...
+
+action_info_modules() ->
+    ActionInfoModules = [
+        action_info_modules(App)
+     || {App, _, _} <- application:loaded_applications()
+    ],
+    lists:usort(lists:flatten(ActionInfoModules) ++ hard_coded_action_info_modules()).
+
+action_info_modules(App) ->
+    case application:get_env(App, emqx, action_info_module) of
```
thalesmg, line 149: Taking Kafka app.src for example:

```erlang
{env, [{emqx_action_info_module, emqx_bridge_kafka_action_info}]},
```

Shouldn't this be:

```suggestion
    case application:get_env(App, emqx_action_info_module) of
```

? (also, `application:get_env/3` doesn't return `{ok, _}`)

apps/emqx_bridge/src/schema/emqx_bridge_v2_schema.erl
```diff
@@ -85,7 +86,21 @@ post_request() ->
...
 api_schema(Method) ->
     EE = ?MODULE:enterprise_api_schemas(Method),
-    hoconsc:union(bridge_api_union(EE)).
+    APISchemas = registered_api_schemas(Method),
+    hoconsc:union(bridge_api_union(EE ++ APISchemas)).
+
+registered_api_schemas(Method) ->
+    %% We *must* do this to ensure the module is really loaded, especially when we use
+    %% `call_hocon' from `nodetool' to generate initial configurations.
+    _ = emqx_bridge_v2:module_info(),
```
thalesmg, line 95: Nit: since the checks for exported functions are done inside `emqx_action_info`, I think this is not necessary here:

```suggestion
```

apps/emqx_bridge/src/schema/emqx_bridge_v2_schema.erl
```diff
@@ -133,7 +148,17 @@ roots() ->
     end.
 
 fields(actions) ->
-    [] ++ enterprise_fields_actions().
+    %% We *must* do this to ensure the module is really loaded, especially when we use
+    %% `call_hocon' from `nodetool' to generate initial configurations.
+    _ = emqx_bridge_v2:module_info(),
```
thalesmg, line 153: Nit:

```suggestion
```

apps/emqx_bridge/src/schema/emqx_bridge_v2_schema.erl
```diff
@@ -85,7 +86,21 @@ post_request() ->
...
+
+registered_api_schemas(Method) ->
+    %% We *must* do this to ensure the module is really loaded, especially when we use
+    %% `call_hocon' from `nodetool' to generate initial configurations.
+    _ = emqx_bridge_v2:module_info(),
+    RegistredSchmeas = emqx_action_info:registered_schema_modules(),
+    [
+        api_ref(SchemaModule, atom_to_binary(BridgeV2Type), Method ++ "_bridge_v2")
+     || {BridgeV2Type, SchemaModule} <- RegistredSchmeas
+    ].
```
thalesmg, line 100:
```suggestion
    RegisteredSchemas = emqx_action_info:registered_schema_modules(),
    [
        api_ref(SchemaModule, atom_to_binary(BridgeV2Type), Method ++ "_bridge_v2")
     || {BridgeV2Type, SchemaModule} <- RegisteredSchemas
    ].
```

apps/emqx_bridge/src/emqx_action_info.erl
```diff
@@ -0,0 +1,197 @@
...
+    [].
+
+hard_coded_action_info_modules() ->
+    hard_coded_action_info_modules_common() ++ hard_coded_action_info_modules_ee().
+
+%% ====================================================================
+%% API
+%% ====================================================================
+
+action_type_to_connector_type(Type) when not is_atom(Type) ->
```
thalesmg, line 63: Q: there are some places that already assume the convention of having `connector_name = action_name` (such as some config checks/callbacks).  If this callback seems to allow different names, wouldn't that lead to future problems?

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -88,25 +92,54 @@
...
     bridge_v2_type_to_connector_type/1,
-    is_bridge_v2_type/1,
-    lookup_and_transform_to_bridge_v1/2,
-    list_and_transform_to_bridge_v1/0,
+    is_bridge_v2_type/1
+]).
+
+%% Compatibility Layer API
+%% All public functions for the compatibility layer should be prefixed with
+%% birdge_v1_
```
thalesmg, line 104:
```suggestion
%% bridge_v1_
```

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -1006,7 +1046,7 @@ unpack_bridge_conf(Type, PackedConf, TopLevelConf) ->
 %%
 %% * The corresponding bridge v2 should exist
 %% * The connector for the bridge v2 should have exactly one channel
-is_valid_bridge_v1(BridgeV1Type, BridgeName) ->
+bridge_v1_is_valid(BridgeV1Type, BridgeName) ->
```
thalesmg, line 1049: I feel like this should return `false` for entirely new bridges created after this refactoring (which should not be supported by V1 API)?

kjellwinblad: I think this is a design decision. I think it is fine new bridges are also supported in the legacy API. That would be more consistent in my opinion since internally there is no difference between refactored bridges and bridges created from scratch as action/connector. I think we can adjust this function according to the decision once a decision has been made.

apps/emqx_bridge/src/emqx_action_info.erl
kjellwinblad: I think we should fix those places that relay on this assumption that is true today but might not be true in the future. I think most places already do the conversion?

apps/emqx_bridge/src/emqx_action_info.erl
kjellwinblad: Good catch!

apps/emqx_bridge/src/emqx_action_info.erl
thalesmg: Well, we had agreed on that convention, so I'm not sure _those_ places should be changed.  Unless we retracted that convention?

apps/emqx_conf/src/emqx_conf.erl
kjellwinblad: I might misunderstand how that functionality works but it seems to me that it hard codes the schema module names in the emqx_conf app which I think would be good to avoid.

kjellwinblad: For actions we currently (with this PR) hard code the info module names implicitly via `emqx_action_info:hard_coded_action_info_modules/0` which I think is a little bit better since it is in a lower level in the dependency hierarchy.

apps/emqx_bridge/src/emqx_action_info.erl
kjellwinblad: Ok, sorry. I know that we decided to use this convention for the existing bridges but not that we decided that this would be true for all future bridges as well. We can very easily enforce this convention by removing the conversion function and change the places that is calling it but I think this can be done in a different PR?

thalesmg: Agreed, this doesn't need to be done here.

apps/emqx_bridge_azure_event_hub/src/emqx_bridge_azure_event_hub_action_info.erl
thalesmg, line 1: Missing copyright header (adding a new comment because the old one is marked as outdated, for some reason)

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_action_info.erl
thalesmg, line 1: Missing copyright header (adding a new comment because the old one is marked as outdated, for some reason)

thalesmg (dismissed):
I think there are just a couple copyright headers missing, otherwise LGTM

apps/emqx_bridge_azure_event_hub/src/emqx_bridge_azure_event_hub_action_info.erl
kjellwinblad: Sorry, my bad! Fixed!

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_action_info.erl
kjellwinblad: Sorry, my bad! Fixed!

apps/emqx_bridge_azure_event_hub/src/emqx_bridge_azure_event_hub_action_info.erl
kjellwinblad: Sorry, my bad! Fixed!

thalesmg (dismissed):

thalesmg (approved):

id:
--> https://github.com/emqx/emqx/pull/11945

PR #11886, opened 2023-11-06
fix(plugins): fix backward compatibility

savonarola:
Fixes [EMQX-11324](https://emqx.atlassian.net/browse/EMQX-11324)

### <samp>🤖 Generated by Copilot at 8438d73</samp>

This pull request improves the support and testing of legacy plugin templates that use deprecated hookpoints. It refactors the `emqx_hookpoints.erl` module to use a map and log warnings, and adds a new test case in `emqx_plugins_SUITE.erl` for the legacy templates.

## PR Checklist

- [na] Changed lines covered in coverage report

### DIFF
--- a/apps/emqx/src/emqx_hookpoints.erl
+++ b/apps/emqx/src/emqx_hookpoints.erl
@@ -16,6 +16,8 @@
 
 -module(emqx_hookpoints).
 
+-include("logger.hrl").
+
 -type callback_result() :: stop | any().
 -type fold_callback_result(Acc) :: {stop, Acc} | {ok, Acc} | stop | any().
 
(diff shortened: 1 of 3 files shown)


thalesmg (dismissed):

apps/emqx/src/emqx_hookpoints.erl
```diff
@@ -206,27 +212,42 @@ when
...
 verify_hookpoint(HookPoint) when is_binary(HookPoint) -> ok;
 verify_hookpoint(HookPoint) ->
-    case maps:is_key(HookPoint, registered_hookpoints()) of
-        true ->
-            ok;
-        false ->
-            error({invalid_hookpoint, HookPoint})
+    case maps:find(HookPoint, registered_hookpoints()) of
+        {ok, valid} -> ok;
+        {ok, deprecated} -> ?SLOG(warning, #{msg => deprecated_hookpoint, hookpoint => HookPoint});
```
zmstone, line 243: it means the hook is added, but the callback is never going to be evaluated?
should we maybe try to fix it to the correct endpoint ?

savonarola: Making them aliases of the correct endpoints may break the existing behaviour 🤔 Previously, they were never actually called, but we will start calling them.

thalesmg (approved):

PR #11921, opened 2023-11-09
EMQX-10534 integrate otel log handler

SergeTupchiy:
Closes:  https://emqx.atlassian.net/browse/EMQX-10534

Documentation follow-up ticket: https://emqx.atlassian.net/browse/EMQX-11385
Tests follow-up ticket: https://emqx.atlassian.net/browse/EMQX-11432

### <samp>🤖 Generated by Copilot at 0cfb424</samp>

This pull request adds support for external log handlers using the OpenTelemetry protocol in EMQ X. It updates the dependencies, the configuration schema, and the application version in `mix.exs` and `emqx_opentelemetry.app.src`. It adds a new behaviour `emqx_external_log_handler` and a new module `emqx_otel_log_handler` that implements it. It modifies `emqx_config_logger` to enable the external log handlers. It also adds a new schema `otel_log_handler` and `otel_exporter` to define the configuration options for the OpenTelemetry log handler and exporter.

## PR Checklist


apps/emqx/src/bhvrs/emqx_external_log_handler.erl
```diff
@@ -0,0 +1,28 @@
...
+-callback tr_handlers(Conf :: map()) ->
+    [{handler, logger:handler_id(), module(), logger:handler_config()}].
+
+-export([tr_handlers/1]).
+
+tr_handlers(Conf) ->
+    lists:flatmap(fun(M) -> M:tr_handlers(Conf) end, tr_modules()).
+
+tr_modules() ->
+    [emqx_otel_log_handler].
```
SergeTupchiy, line 28: This probably should be better decoupled from the main `emqx` app, but I tried to keep it simple for now.

apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -996,6 +996,27 @@ fields("log") ->
                     aliases => [file_handlers],
                     importance => ?IMPORTANCE_HIGH
                 }
+            )},
+        {"external",
```
SergeTupchiy, line 1000: There is only ome `opentelemetry` type for now, introducing `external` type appears worthwhile for potential future extensions.  However, it's also possible to simply add `log.opentelemetry` - should be discussed.

apps/emqx/src/bhvrs/emqx_external_log_handler.erl
zmstone: can we install the handler at runtime? i.e. in emqx_opentelemetry application

apps/emqx_opentelemetry/src/emqx_otel_schema.erl
```diff
@@ -29,18 +29,18 @@ roots() -> ["opentelemetry"].
 
 fields("opentelemetry") ->
     [
-        {exporter,
+        {"exporter",
```
zmstone, line 32: keep using atoms for filed names

thalesmg: Q: why is it better to use atoms for these?

apps/emqx/src/bhvrs/emqx_external_log_handler.erl
SergeTupchiy: @zmstone, yes, but it would mean that before` emqx_opentelemetry` app is started, logs won't be sent to the external otel collector/backend. Unless we start emqx_opentelemetry earlier than any other app.

SergeTupchiy: BTW, current `otel_log_handler` implementation anyway doesn't allow to start a handler at early stage (when kernel logger starts). I think we have two options for now:
- change otel_log_handler to start under kernel's `logger_sup`, start accepting log event writes to the table but postpone 'full init' before kernel is fully started. 
   Advantages: the handler can work similarly to native log handlers with no log event losses.
   Drawbacks:  - have to rely on some kernel internals (logger_sup),
                        - more complicated log_handler design is required
 - keep otel_log_handler as is and accept the fact that it can't be initialized at early boot stage, meaning some logs will always be missing (probably up to the point when emqx_conf app initiates).

SergeTupchiy: Note: my previous comment is not absolutely correct. There was actually a deadlock trying to start ssl app (opentelemetry_exporter dependency) in adding_handler/1 logger cb. I'm fixing that now in opentelemetry lib, so starting a handler under kernel logger_sup shouldn't be required.

zmstone: I don't think it's so critical to have the boot logs sent to otel ?

SergeTupchiy: Moved everything to `emqx_opentelemetry` app

apps/emqx_opentelemetry/src/emqx_otel_schema.erl
SergeTupchiy: Changed back to atoms

SergeTupchiy:
NOTE: 
Thanks to `upgrade_raw_conf/1`, the new feature is backward compatible with cluster.hocon/emqx.conf that define metrics config under `opentelmetry` root key.
However,  it requires Dashboard UI update to conform with the new schema.

rel/i18n/emqx_otel_schema.hocon
```diff
@@ -1,15 +1,44 @@
 emqx_otel_schema {
 
 opentelemetry.desc: "Open Telemetry Toolkit configuration"
+opentelemetry.label: "Open Telemetry"
+
+otel_logs.desc:
+"""Open Telemetry Logs configuration. If enabled, EMQX Installs a log handler that formats events according to Open Telemetry log data model and
```
ieQu1, line 7:
```suggestion
"""Open Telemetry Logs configuration. If enabled, EMQX installs a log handler that formats events according to Open Telemetry log data model and
```

rel/i18n/emqx_otel_schema.hocon
```diff
@@ -1,15 +1,44 @@
...
+exports them to the configured Open Telemetry collector or back-end."""
```
ieQu1, line 8:
```suggestion
exports them to the configured Open Telemetry collector or backend."""
```

rel/i18n/emqx_otel_schema.hocon
```diff
@@ -1,15 +1,44 @@
...
+otel_metrics.label: "Open Telemetry Metrics"
+
+enable.desc: "Enable or disable Open Telemetry signal."
+enable.label: "Enable."
 
 exporter.desc: "Open Telemetry Exporter"
+exporter.label: "Exporter"
+
+max_queue_size.desc:
+"""The maximum queue size. After the size is reached Open Telemetry signals are dropped."""
```
ieQu1, line 21: Q: What is the strategy for dropping the signals? Does it drop the new signals until the queue is drained, or discards the queued data as well?

rel/i18n/emqx_otel_schema.hocon
```diff
@@ -1,15 +1,44 @@
...
+max_queue_size.label: "Max Queue Size"
+
+exporting_timeout.desc: "The time Open Telemetry signal export can run before it is cancelled."
+exporting_timeout.label: "Exporting Timeout"
+
+scheduled_delay.desc: "The delay interval between two consecutive exports of Open Telemetry signal."
```
zmstone, line 27:
```suggestion
scheduled_delay.desc: "The delay interval between two consecutive exports of Open Telemetry signals."
```

apps/emqx_opentelemetry/src/emqx_otel_schema.erl
```diff
@@ -42,41 +74,127 @@ fields("opentelemetry") ->
...
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )},
+        {ssl_options,
+            ?HOCON(
+                ?R_REF(emqx_schema, "ssl_client_opts"),
+                #{
+                    default => #{<<"enable">> => false},
+                    desc => ?DESC(exporter_ssl),
+                    importance => ?IMPORTANCE_HIGH
```
zmstone, line 180:
```suggestion
                    importance => ?IMPORTANCE_LOW
```

apps/emqx_opentelemetry/src/emqx_otel_schema.erl
```diff
@@ -42,41 +74,127 @@ fields("opentelemetry") ->
...
+                    default => grpc,
+                    desc => ?DESC(exporter_protocol),
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )},
+        {ssl_options,
+            ?HOCON(
+                ?R_REF(emqx_schema, "ssl_client_opts"),
+                #{
+                    default => #{<<"enable">> => false},
```
zmstone, line 178: it seems ssl enable/disable is derived from the endpoint url scheme.
maybe create a new type here and call emqx_schema:common_ssl_opts_schema (export it) for the fields.

i.e. no need for `enable` field.

rel/i18n/emqx_otel_schema.hocon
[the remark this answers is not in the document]
SergeTupchiy: Yes, exactly, the idea is to convert all logs to OTLP protocol and send them to the configured sink, e.g. opentelemetry collector or back-end. 
An example showing a bridge error logs received by opentelemetry-collector:
Screenshot from 2023-11-22 11-10-17

rel/i18n/emqx_otel_schema.hocon
SergeTupchiy: It discards new signals. No need to discard the queued data, since olp mechanism generally ensures that the 'queue' stops growing once the limit is reached. The 'queue' is actually not a real queue data structure, but an ETS table. 
We keep using 'max_queue_size' because it's a term used in OpenTelemetry specification and our ETS -based implementation doesn't violate the spec. 
Each insert to ETS table is done after decrementing and getting the value of the atomic counter. Once counter is decremented to 0 (it starts with max_queue_size), the new events are drooped and the processor (gen_statem process) is notified to start a next exporting cycle. 
The actual implementation is slightly more complex since there are two ETS tables and they are reversed periodically, so that when the processor starts exporting, another table becomes available for inserts.
https://github.com/emqx/opentelemetry-erlang/blob/main/apps/opentelemetry/src/otel_batch_olp.erl#L178

rel/i18n/emqx_otel_schema.hocon
SergeTupchiy: Thanks, fixed

rel/i18n/emqx_otel_schema.hocon
SergeTupchiy: Fixed

rel/i18n/emqx_otel_schema.hocon
SergeTupchiy: Thanks, fixed

apps/emqx_opentelemetry/src/emqx_otel_schema.erl
SergeTupchiy: Fixed

apps/emqx_opentelemetry/src/emqx_otel_schema.erl
SergeTupchiy: Fixed slightly differently, following similar cases in other schemas, e.g.: https://github.com/emqx/emqx/blob/master/apps/emqx_ldap/src/emqx_ldap.erl#L100

rel/i18n/emqx_otel_schema.hocon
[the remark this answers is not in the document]
ieQu1: Cool, thanks for the clarification!

rel/i18n/emqx_otel_schema.hocon
ieQu1: Thanks. Perhaps it's worth mentioning in the documentation?

zmstone (approved):

PR #11937, opened 2023-11-13
Emqx 11318 crash in get bridges v 2 if a broken bridge is configured

sstrigler:
Fixes [EMQX-11318](https://emqx.atlassian.net/browse/EMQX-11318)

### <samp>🤖 Generated by Copilot at 631f54e</samp>

This pull request fixes a bug and improves the error handling of the bridge v2 API. It changes the `emqx_bridge_v2_api.erl` and `emqx_bridge_v2.erl` files to handle invalid or uninitialized bridges more gracefully. It also adds a test case in `emqx_bridge_v2_api_SUITE.erl` to verify the API behavior.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_v2.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_v2.erl
@@ -362,39 +362,22 @@ uninstall_bridge_v2(
 uninstall_bridge_v2(
     BridgeV2Type,
     BridgeName,
-    Config
-) ->
-    uninstall_bridge_v2_helper(
-        BridgeV2Type,
-        BridgeName,
-        combine_connector_and_bridge_v2_config(
-            BridgeV2Type,
-            BridgeName,
-            Config
-        )
-    ).
-
-uninstall_bridge_v2_helper(
-    _BridgeV2Type,
-    _BridgeName,
-    {error, Reason} = Error
-) ->
-    ?SLOG(error, Reason),
-    Error;
-uninstall_bridge_v2_helper(
-    BridgeV2Type,
-    BridgeName,
     #{connector := ConnectorName} = Config
 ) ->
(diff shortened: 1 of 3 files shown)


apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -390,11 +368,16 @@ uninstall_bridge_v2_helper(
...
     ok = emqx_resource:clear_metrics(BridgeV2Id),
-    %% Deinstall from connector
-    ConnectorId = emqx_connector_resource:resource_id(
-        connector_type(BridgeV2Type), ConnectorName
-    ),
-    emqx_resource_manager:remove_channel(ConnectorId, BridgeV2Id).
+    case combine_connector_and_bridge_v2_config(BridgeV2Type, BridgeName, Config) of
+        {error, _} ->
+            ok;
+        _CombinedConfig ->
```
thalesmg, line 374: Tiny nit (since we don't care about the combined config, this may sound more clear when reading this part):

```suggestion
    case validate_referenced_connectors(BridgeV2Type, ConnectorName, BridgeName) of
        {error, _} ->
            ok;
        ok ->
```

thalesmg (approved):

apps/emqx_bridge/src/emqx_bridge_v2.erl
thalesmg: I know...  It's just that there's this validate function whose purpose is to check the referenced connector.  Since we don't care about the combined config, it seems easier to understand the purpose of a `validate_referenced_connectors` call here rather than `combine_...` which we discard the result of.  :wink:

sstrigler: Will fix at follow-up PR

PR #11939, opened 2023-11-13
fix(connector): validate connector name before converting ssl certs

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-11336

See also: https://github.com/emqx/emqx/pull/11540

### <samp>🤖 Generated by Copilot at d108dd2</samp>

This pull request adds a new feature to validate the bridge and connector names using a regular expression and fixes a bug in the V1 HTTP API for creating bridges with SSL options. It updates the schema, the `emqx_bridge` and `emqx_connector` modules, and the corresponding test suites. It also adds new test cases for different bridge and connector types and configuration paths.

## PR Checklist

### DIFF
--- a/apps/emqx_connector/src/emqx_connector.erl
+++ b/apps/emqx_connector/src/emqx_connector.erl
@@ -108,18 +108,28 @@ config_key_path() ->
 pre_config_update([?ROOT_KEY], RawConf, RawConf) ->
     {ok, RawConf};
 pre_config_update([?ROOT_KEY], NewConf, _RawConf) ->
-    {ok, convert_certs(NewConf)};
+    case multi_validate_connector_names(NewConf) of
+        ok ->
+            {ok, convert_certs(NewConf)};
+        Error ->
+            Error
+    end;
 pre_config_update(_, {_Oper, _, _}, undefined) ->
     {error, connector_not_found};
 pre_config_update(_, {Oper, _Type, _Name}, OldConfig) ->
     %% to save the 'enable' to the config files
     {ok, OldConfig#{<<"enable">> => operation_to_enable(Oper)}};
 pre_config_update(Path, Conf, _OldConfig) when is_map(Conf) ->
-    case emqx_connector_ssl:convert_certs(filename:join(Path), Conf) of
-        {error, Reason} ->
-            {error, Reason};
-        {ok, ConfNew} ->
-            {ok, ConfNew}
+    case validate_connector_name_in_config(Path) of
+        ok ->
+            case emqx_connector_ssl:convert_certs(filename:join(Path), Conf) of
+                {error, Reason} ->
+                    {error, Reason};
+                {ok, ConfNew} ->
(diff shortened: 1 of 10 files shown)


apps/emqx_connector/src/emqx_connector.erl
```diff
@@ -458,3 +460,55 @@ ensure_no_channels(Configs) ->
         {error, Reason, _State} ->
             {error, Reason}
     end.
+
+to_bin(A) when is_atom(A) -> atom_to_binary(A, utf8);
+to_bin(B) when is_binary(B) -> B.
+
+validate_connector_name(ConnectorName0) ->
+    ConnectorName = to_bin(ConnectorName0),
+    case re:run(ConnectorName, ?MAP_KEY_RE, [{capture, none}]) of
```
zmstone, line 469: there is `emqx_resource:validate_name/1`

thalesmg: :heavy_check_mark:

apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_producer_SUITE.erl
```diff
@@ -317,3 +317,28 @@ t_same_name_azure_kafka_bridges(AehConfig) ->
         end
     ),
     ok.
+
+t_create_with_bad_name(Config0) ->
```
sstrigler, line 321: Not so sure about this, why test it here too? It's been reported for AEH, but the problem has nothing to do with it. Future people might start to wonder why this is tested here. Do we want to test that with every bridge?

Certainly we want to test whether the compat layer (i.e. using a v2 bridge works through the old `/bridges` API endpoint), but isn't that done elsewhere too?

Anyway, maybe leave a comment at least, and explain why this is here.

apps/emqx_connector/src/emqx_connector.erl
```diff
@@ -458,3 +468,51 @@ ensure_no_channels(Configs) ->
...
+    catch
+        throw:Error ->
+            {error, Error}
+    end.
+
+validate_connector_name_in_config(Path) ->
+    case Path of
+        [?ROOT_KEY, _ConnectorType, ConnectorName] ->
+            validate_connector_name(ConnectorName);
+        _ ->
```
sstrigler, line 488: Is this something to expect? When? Otherwise, wouldn't it be better to just crash?

thalesmg: True, I don't think this should be reached.  I'll removed it.

thalesmg: Actually, no, it may happen if someone does `emqx_conf:update([connectors, type], ...)`.

kjellwinblad (approved):

ieQu1 (approved):

PR #11968, opened 2023-11-17
Bridge V1 config upgrade downgrade fixup callbacks

kjellwinblad:
Adds:

* convenience function for creating action schemas
* callbacks to the emqx_action_info module for doing fixes (such as changing a field name) after the automatic split of a Bridge V1 config or the merge of connector and action configs for the compatibility layer.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 797563f</samp>

This pull request improves the compatibility and flexibility of the bridge v1 and v2 formats by adding custom conversion logic for different action types. It also refactors some schema modules to simplify the code and fix some typos. The main files affected are `emqx_action_info.erl`, `emqx_bridge_api.erl`, `emqx_bridge_v2_schema.erl`, `emqx_bridge_v2.erl`, and `emqx_connector_schema.erl`.

## PR Checklist


apps/emqx_bridge/src/emqx_action_info.erl
```diff
@@ -25,15 +25,25 @@
...
 -callback bridge_v1_type_name() -> atom().
 -callback action_type_name() -> atom().
 -callback connector_type_name() -> atom().
 -callback schema_module() -> atom().
-
--optional_callbacks([bridge_v1_type_name/0]).
+%% Define this if the automatic config downgrade is not enough for the bridge.
+-callback action_to_bridge_v1_fixup(map()) -> term().
+%% Define this if the automatic config upgrade is not enough for the bridge.
+-callback bridge_v1_to_action_fixup(map()) -> term().
```
thalesmg, line 40: Nit

```suggestion
-callback action_to_bridge_v1_fixup(map()) -> map().
%% Define this if the automatic config upgrade is not enough for the bridge.
-callback bridge_v1_to_action_fixup(map()) -> map().
```

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_action_info.erl
```diff
@@ -20,3 +22,19 @@ action_type_name() -> kafka_producer.
...
 
 schema_module() -> emqx_bridge_kafka.
+
+action_to_bridge_v1_fixup(Config) ->
+    rename(<<"parameters">>, <<"kafka">>, Config).
+
+rename(OldKey, NewKey, Map) ->
+    case maps:find(OldKey, Map) of
+        {ok, Value} ->
+            maps:remove(OldKey, maps:put(NewKey, Value, Map));
```
thalesmg, line 32: Nit:

```suggestion
            maps:put(NewKey, Value, maps:remove(OldKey, Map));
```

Because renaming a key to itself would remove it from the result otherwise.

apps/emqx_bridge_azure_event_hub/src/emqx_bridge_azure_event_hub_action_info.erl
```diff
@@ -20,3 +22,19 @@ action_type_name() -> azure_event_hub_producer.
...
 
 schema_module() -> emqx_bridge_azure_event_hub.
+
+action_to_bridge_v1_fixup(Config) ->
+    rename(<<"parameters">>, <<"kafka">>, Config).
+
+rename(OldKey, NewKey, Map) ->
+    case maps:find(OldKey, Map) of
+        {ok, Value} ->
+            maps:remove(OldKey, maps:put(NewKey, Value, Map));
```
thalesmg, line 32: Nit:

```suggestion
            maps:put(NewKey, Value, maps:remove(OldKey, Map));
```

Because renaming a key to itself would remove it from the result otherwise.

Maybe this could be added to `emqx_utils_maps`.

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_action_info.erl
```diff
@@ -20,3 +22,19 @@ action_type_name() -> kafka_producer.
...
+        error ->
+            Map
+    end.
+
+bridge_v1_to_action_fixup(Config) ->
+    KafkaField = emqx_utils_maps:deep_get([<<"parameters">>, <<"kafka">>], Config, #{}),
+    Config1 = emqx_utils_maps:deep_remove([<<"parameters">>, <<"kafka">>], Config),
```
thalesmg, line 39: I think there's no `kafka` under parameters.  `kafka` was renamed to `parameters` in e5.3.1

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_action_info.erl
```diff
@@ -20,3 +22,19 @@ action_type_name() -> kafka_producer.
 connector_type_name() -> kafka_producer.
 
 schema_module() -> emqx_bridge_kafka.
+
+action_to_bridge_v1_fixup(Config) ->
```
thalesmg, line 26: Nit: since it's currently the same, this callback and its inverse could delegate to Kafka's module to avoid repetition.

apps/emqx_bridge_azure_event_hub/src/emqx_bridge_azure_event_hub_action_info.erl
```diff
@@ -20,3 +22,19 @@ action_type_name() -> azure_event_hub_producer.
...
+        error ->
+            Map
+    end.
+
+bridge_v1_to_action_fixup(Config) ->
+    KafkaField = emqx_utils_maps:deep_get([<<"parameters">>, <<"kafka">>], Config, #{}),
+    Config1 = emqx_utils_maps:deep_remove([<<"parameters">>, <<"kafka">>], Config),
```
thalesmg, line 39: I think there's no `kafka` under parameters.  `kafka` was renamed to `parameters` in e5.3.1

apps/emqx_connector/src/schema/emqx_connector_schema.erl
```diff
@@ -203,11 +203,21 @@ transform_old_style_bridges_to_connector_and_actions_of_type(
...
+            %% Take fields that should be in the top level of the action map
+            TopLevelFields = [<<"resource_opts">>, <<"enable">>, <<"connector">>],
+            TopLevelActionFields = maps:with(TopLevelFields, ActionMap),
+            ParametersActionFields = maps:without(TopLevelFields, ActionMap),
+            %% Action map should be wrapped under parameters key
+            WrappedParameters = #{<<"parameters">> => ParametersActionFields},
+            FinalActionMap = maps:merge(TopLevelActionFields, WrappedParameters),
+            FixedActionMap = emqx_action_info:bridge_v1_to_action_fixup(
+                BridgeType, FinalActionMap
+            ),
```
thalesmg, line 215: I think I see why the fixup code in Kafka/AEH expects a `kafka` key under `parameters`, which shouldn't happen according to the schema.  Which is a bit confusing.

Couldn't we call the `bridge_v1_to_action_fixup` first, and do the automatic heuristic **only** if the callback is undefined?  That way, all the massaging needed is handled by the node and is closely tied to the v1/action schemas rather than an intermediate step like the wrapping here introduces.

apps/emqx_bridge/src/schema/emqx_bridge_v2_schema.erl
```diff
@@ -150,6 +154,24 @@ examples(Method) ->
...
+
+make_action_schema(ActionParametersRef) ->
+    [
+        {enable, mk(boolean(), #{desc => ?DESC("config_enable"), default => true})},
+        {connector,
+            mk(binary(), #{
+                desc => ?DESC(emqx_connector_schema, "connector_field"), required => true
+            })},
+        {description, emqx_schema:description_schema()},
+        {local_topic, mk(binary(), #{required => false, desc => ?DESC(mqtt_topic)})},
```
thalesmg, line 169: This is not common to consumer actions.

```suggestion
```

thalesmg (commented):
I think this should target `master`, as it'll sync faster, and some people have already started to port actions based off it.  :sweat_smile:

thalesmg (approved):

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_action_info.erl
kjellwinblad: This was supposed to do the renaming but perhaps it is done in another way now....

kjellwinblad: The Kafka field is moved into parameters by the automatic upgrade code.

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_action_info.erl
kjellwinblad: Yes, we could avoid repetition of the this code in the azure bridge. This is what you meant right?

apps/emqx_bridge/src/emqx_action_info.erl
```diff
@@ -130,23 +130,46 @@ action_to_bridge_v1_fixup(ActionOrBridgeType, Config) ->
...
     end.
 
-bridge_v1_to_action_fixup(ActionOrBridgeType, Config) ->
+bridge_v1_to_action_fixup(ActionOrBridgeType, Config0) ->
     Module = get_action_info_module(ActionOrBridgeType),
     case erlang:function_exported(Module, bridge_v1_to_action_fixup, 1) of
         true ->
-            Module:bridge_v1_to_action_fixup(Config);
+            Config1 = Module:bridge_v1_to_action_fixup(Config0),
+            common_bridge_v1_to_action_adapter(Config1);
```
kjellwinblad, line 138: I think the `common_bridge_v1_to_action_adapter` call should be done before calling the `bridge_v1_to_action_fixup` callback. That way one can always fix things that the "automatic" translation does wrong. What is the reason for doing this after?

thalesmg: The reason was that, IMO, it's easier to reason about the transformation for each bridge if one considers only the old and new schemas, rather than something intermediate done by this "automatic" translation, that could change at some point.

PR #11984, opened 2023-11-20
Integrate OpenTelmetry tracing 

SergeTupchiy:
Fixes: EMQX-10535 

Documentation follow-up ticket: https://emqx.atlassian.net/browse/EMQX-11385
~Tests follow-up ticket: https://emqx.atlassian.net/browse/EMQX-11432~ (done in this PR)

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 9ec360f</samp>

This pull request introduces the OpenTelemetry integration feature, refactors the channel and connection modules, improves the routing table schema and performance, fixes some bugs and dialyzer warnings, and updates some tests and dependencies.

## PR Checklist


apps/emqx_opentelemetry/rebar.config
```diff
@@ -1,8 +1,16 @@
 %% -*- mode: erlang -*-
 
-{deps, [
-    {emqx, {path, "../emqx"}}
-]}.
+{deps,
+    [{emqx, {path, "../emqx"}}
+    %% trace
+    , {opentelemetry_api, {git_subdir, "https://github.com/emqx/opentelemetry-erlang", {tag, "v1.4.4-emqx"}, "apps/opentelemetry_api"}}
```
SergeTupchiy, line 6: Moved the deps here to fix Elixir build that couldn't find included hrl files: https://github.com/emqx/emqx/actions/runs/6932019327/job/18855083220?pr=11984#step:3:16138

apps/emqx/src/emqx_external_trace.erl
```diff
@@ -0,0 +1,109 @@
...
+    trace_process_publish/3,
+    start_trace_send/2,
+    end_trace_send/1,
+    event/1,
+    event/2
+]).
+
+-define(PROVIDER, {?MODULE, trace_provider}).
+
+-define(with_provider(IfRegisitered, IfNotRegisired),
```
thalesmg, line 42: Nit:

```suggestion
-define(with_provider(IfRegistered, IfNotRegistered),
```

apps/emqx_opentelemetry/src/emqx_otel_trace.erl
```diff
@@ -0,0 +1,275 @@
...
+-spec toggle_registered(boolean()) -> ok | {error, term()}.
+toggle_registered(true = _Enable) ->
+    emqx_external_trace:register_provider(?MODULE);
+toggle_registered(false = _Enable) ->
+    _ = emqx_external_trace:unregister_provider(?MODULE),
+    ok.
+
+-spec ensure_traces(map()) -> ok | {error, term()}.
+ensure_traces(#{enable := true} = Conf) ->
+    start(Conf);
```
thalesmg, line 59: This seems to be called from the [`application:start/2`](https://github.com/emqx/emqx/blob/3eda1f16f1f758c883b1ec0913d056a490cbe9f2/apps/emqx_opentelemetry/src/emqx_otel_app.erl#L27) callback, but inside `start/1` here we may do [some calls](https://github.com/emqx/emqx/blob/3eda1f16f1f758c883b1ec0913d056a490cbe9f2/apps/emqx_opentelemetry/src/emqx_otel_trace.erl#L80-L81) to `application`, which may deadlock the application startup.

apps/emqx/src/emqx_external_trace.erl
SergeTupchiy: :see_no_evil:  thanks a lot, fixed

apps/emqx_opentelemetry/src/emqx_otel_trace.erl
SergeTupchiy: This seemed safe overall, as application controller spawns another process to start an app :thinking:
 https://github.com/erlang/otp/blob/OTP-25.3.2.2/lib/kernel/src/application_controller.erl#L688
https://github.com/erlang/otp/blob/OTP-25.3.2.2/lib/kernel/src/application_controller.erl#L1385
Perhaps need to re-think it...

thalesmg: As we talked about, maybe it's safe in this use case, indeed.  I'm curious about others' opinions.  :smile_cat: 
I can't recall the conditions in the past in which I bumped into this kind of deadlock, might be during the `stop` callback, which is sync.  :thinking:

SergeTupchiy: @thalesmg, following our discussion, I've updated otel lib and got rid of using `application:ensure_all_started/1` completely: https://github.com/emqx/emqx/pull/11984/commits/16407d6beb9a49aea6df185746eba321e61cc9c8
This must be a better and safer design anyway. 
Thanks for raising this concern! :raised_hands:

apps/emqx/src/emqx_channel.erl
```diff
@@ -398,8 +398,10 @@ handle_in(?PACKET(_), Channel = #channel{conn_state = ConnState}) when
     handle_out(disconnect, ?RC_PROTOCOL_ERROR, Channel);
 handle_in(Packet = ?PUBLISH_PACKET(_QoS), Channel) ->
     case emqx_packet:check(Packet) of
-        ok -> process_publish(Packet, Channel);
-        {error, ReasonCode} -> handle_out(disconnect, ReasonCode, Channel)
+        ok ->
+            emqx_external_trace:trace_process_publish(Packet, Channel, fun process_publish/2);
```
zmstone, line 402: should this work?
```suggestion
            emqx_external_trace:trace_process_publish(Packet, fun(PacketWithTrace) -> process_publish(PacketWithTrace) end);
```

zmstone: ok, need to extrace client id later.
channel_attributes does not seem to be an expensive operation, so we can extract it and pass it down as arg

PR #11990, opened 2023-11-20
Implement a prototype of durable message sharding

ieQu1:
Fixes EMQX-11007

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_message.erl
+++ b/apps/emqx/src/emqx_persistent_message.erl
@@ -57,7 +57,9 @@ storage_backend() ->
 storage_backend(#{builtin := #{enable := true}}) ->
     #{
         backend => builtin,
-        storage => {emqx_ds_storage_bitfield_lts, #{}}
+        storage => {emqx_ds_storage_bitfield_lts, #{}},
+        n_shards => 16,
+        replication_factor => 3
     }.
 
 %%--------------------------------------------------------------------
--- a/apps/emqx/src/emqx_persistent_message.erl
+++ b/apps/emqx/src/emqx_persistent_message.erl
@@ -54,12 +54,14 @@ is_persistence_enabled() ->
 storage_backend() ->
     storage_backend(emqx_config:get([session_persistence, storage])).
 
-storage_backend(#{builtin := #{enable := true}}) ->
+storage_backend(#{
+    builtin := #{enable := true, n_shards := NShards, replication_factor := ReplicationFactor}
+}) ->
     #{
         backend => builtin,
         storage => {emqx_ds_storage_bitfield_lts, #{}},
-        n_shards => 16,
-        replication_factor => 3
+        n_shards => NShards,
+        replication_factor => ReplicationFactor
     }.
 
 %%--------------------------------------------------------------------
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -442,7 +442,9 @@ del_subscription(TopicFilter, DSSessionId) ->
 create_tables() ->
     ok = emqx_ds:open_db(?PERSISTENT_MESSAGE_DB, #{
         backend => builtin,
-        storage => {emqx_ds_storage_bitfield_lts, #{}}
+        storage => {emqx_ds_storage_bitfield_lts, #{}},
+        n_shards => 255,
+        replication_factor => 3
     }),
     ok = mria:create_table(
         ?SESSION_TAB,
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -443,7 +443,7 @@ create_tables() ->
     ok = emqx_ds:open_db(?PERSISTENT_MESSAGE_DB, #{
         backend => builtin,
         storage => {emqx_ds_storage_bitfield_lts, #{}},
-        n_shards => 255,
+        n_shards => 16,
         replication_factor => 3
     }),
     ok = mria:create_table(
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -440,12 +440,6 @@ del_subscription(TopicFilter, DSSessionId) ->
 %%--------------------------------------------------------------------
 
 create_tables() ->
-    ok = emqx_ds:open_db(?PERSISTENT_MESSAGE_DB, #{
-        backend => builtin,
-        storage => {emqx_ds_storage_bitfield_lts, #{}},
-        n_shards => 16,
-        replication_factor => 3
-    }),
     ok = mria:create_table(
         ?SESSION_TAB,
         [
--- a/apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
@@ -53,6 +53,7 @@
 %% tags:
 -define(STREAM, 1).
 -define(IT, 2).
+-define(BATCH, 3).
 
 %% keys:
 -define(tag, 1).
@@ -91,12 +92,12 @@
 
 -type message_id() :: emqx_ds_storage_layer:message_id().
 
--record(batch, {
-    messages :: [emqx_types:message()],
-    misc = #{} :: map()
-}).
+-define(batch_messages, 2).
 
--type batch() :: #batch{}.
+-type batch() :: #{
+    ?tag := ?BATCH,
+    ?batch_messages := [emqx_types:message()]
+}.
 
 %%================================================================================
 %% API functions
@@ -130,7 +131,7 @@ drop_db(DB) ->
 store_batch(DB, Messages, Opts) ->
     Shard = shard_of_messages(DB, Messages),
     Node = node_of_shard(DB, Shard),
-    Batch = #batch{messages = Messages},
+    Batch = #{?tag => ?BATCH, ?batch_messages => Messages},
     emqx_ds_proto_v1:store_batch(Node, DB, Shard, Batch, Opts).
 
 -spec get_streams(emqx_ds:db(), emqx_ds:topic_filter(), emqx_ds:time()) ->
@@ -214,7 +215,7 @@ do_drop_db_v1(DB) ->
     emqx_ds:message_store_opts()
 ) ->
     emqx_ds:store_batch_result().
-do_store_batch_v1(DB, Shard, #batch{messages = Messages}, Options) ->
+do_store_batch_v1(DB, Shard, #{?tag := ?BATCH, ?batch_messages := Messages}, Options) ->
     emqx_ds_storage_layer:store_batch({DB, Shard}, Messages, Options).
 
 -spec do_get_streams_v1(
--- a/apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
@@ -39,7 +39,7 @@
     do_next_v1/4
 ]).
 
--export_type([shard_id/0, builtin_db_opts/0, stream/0, iterator/0, message_id/0]).
+-export_type([shard_id/0, builtin_db_opts/0, stream/0, iterator/0, message_id/0, batch/0]).
 
 -include_lib("emqx_utils/include/emqx_message.hrl").
 
@@ -91,6 +91,13 @@
 
 -type message_id() :: emqx_ds_storage_layer:message_id().
 
+-record(batch, {
+    messages :: [emqx_types:message()],
+    misc = #{} :: map()
+}).
+
+-type batch() :: #batch{}.
+
 %%================================================================================
 %% API functions
 %%================================================================================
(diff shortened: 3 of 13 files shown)


apps/emqx/src/emqx_persistent_message.erl
```diff
@@ -40,9 +40,14 @@
 
 init() ->
     ?WHEN_ENABLED(begin
+        NShards = emqx_config:get([persistent_session_store, n_shards]),
```
ieQu1, line 43: This will have to be rewritten to accommodate the new schema.

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
```diff
@@ -82,42 +91,46 @@
 
 -type message_id() :: emqx_ds_storage_layer:message_id().
 
+-record(batch, {
+    messages :: [emqx_types:message()],
+    misc = #{} :: map()
+}).
+
+-type batch() :: #batch{}.
```
thalesmg, line 99:
```suggestion
-type batch() :: 
    #{
        ?tag := ?BATCH,
        ?messages := [emqx_types:message()]
    }.
```

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -0,0 +1,338 @@
...
+            {ok, FD} = file:open(Filename, [write]),
+            io:format(FD, "~p.", [Site]),
+            file:close(FD)
+    end,
+    {atomic, ok} = mria:transaction(?SHARD, fun ?MODULE:claim_site/2, [Site, node()]),
+    persistent_term:put(?emqx_ds_builtin_site, Site),
+    ok.
+
+-spec this_site() -> site().
+this_site() ->
```
thalesmg, line 304: Nit: move this fn to the "API functions" section, as it's part of the public API?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -0,0 +1,338 @@
...
+
+-spec create_shards(emqx_ds:db(), pos_integer(), pos_integer()) -> ok.
+create_shards(DB, NShards, ReplicationFactor) ->
+    Shards = [integer_to_binary(I) || I <- lists:seq(0, NShards - 1)],
+    Sites = sites(),
+    lists:foreach(
+        fun(Shard) ->
+            Hashes0 = [{hash(Shard, Site), Site} || Site <- Sites],
+            Hashes = lists:sort(Hashes0),
+            {_, Sites} = lists:unzip(Hashes),
```
thalesmg, line 315: `Sites` is a bound variable at this point, but I think the purpose here is not to assert anything.

```suggestion
    Sites0 = sites(),
    lists:foreach(
        fun(Shard) ->
            Hashes0 = [{hash(Shard, Site), Site} || Site <- Sites0],
            Hashes = lists:sort(Hashes0),
            {_, Sites} = lists:unzip(Hashes),
```

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -0,0 +1,338 @@
...
+
+-spec drop_db_trans(emqx_ds:db()) -> ok.
+drop_db_trans(DB) ->
+    mnesia:delete({?META_TAB, DB}),
+    [mnesia:delete({?SHARD_TAB, Shard}) || Shard <- shards(DB)],
+    ok.
+
+-spec claim_site(site(), node()) -> ok.
+claim_site(Site, Node) ->
+    mnesia:write(#?NODE_TAB{site = Site, node = Node}).
```
thalesmg, line 236: Q: despite being rare, shouldn't we still check if the `Site` being claimed hasn't already been taken by another node?  We're already in a transaction here.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -0,0 +1,338 @@
...
+-define(META_TAB, emqx_ds_builtin_metadata_tab).
+%% Mapping from Site to the actual Erlang node:
+-define(NODE_TAB, emqx_ds_builtin_node_tab).
+%% Shard metadata:
+-define(SHARD_TAB, emqx_ds_builtin_shard_tab).
+
+-record(?META_TAB, {
+    db :: emqx_ds:db(),
+    db_props :: emqx_ds_replication_layer:builtin_db_opts()
+}).
```
thalesmg, line 74:
```suggestion
    misc = #{} :: map(),
}).
```
?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -0,0 +1,338 @@
...
+
+-record(?NODE_TAB, {
+    site :: site(),
+    node :: node(),
+    misc = #{} :: map()
+}).
+
+-record(?SHARD_TAB, {
+    shard :: {emqx_ds:db(), emqx_ds_replication_layer:shard_id()},
+    %% Sites that the
```
thalesmg, line 84: Nit: missing rest of explanation

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -0,0 +1,338 @@
...
+%% Internal functions
+%%================================================================================
+
+ensure_tables() ->
+    Majority = true,
+    ok = mria:create_table(?META_TAB, [
+        {rlog_shard, ?SHARD},
+        {majority, Majority},
+        {type, ordered_set},
+        {storage, rocksdb_copies},
```
thalesmg, line 265: Nit: these tables might the [this workaround](https://github.com/emqx/emqx/blob/6c9a8461f7cc1771835dd2bc7ad0def10a802649/apps/emqx/src/emqx_persistent_session_ds.erl#L497) for windows.

ieQu1: Our entire builtin backend depends on rocksdb, so it won't work on windows anyway.

thalesmg: `emqx_ds_sup` will start, start this process, and it'll try to create the tables, which will fail on windows and prevent the node from booting even if this feature is disabled, if I understood correctly.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
ieQu1: I think we can just extend `db_props`.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
ieQu1: Good catch!

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
ieQu1: It may happen in theory when the node name changes. I think we should just overwrite the old record.

thalesmg: If it's the same node that generated that site ID, yes; but a different node might clobber an already running node site id, in principle.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
ieQu1: Hmm, good point. But the proper solution is to abort the startup of the application early when on Windows, and don't create the tables.

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
```diff
@@ -82,42 +91,47 @@
 
 -type message_id() :: emqx_ds_storage_layer:message_id().
 
+-define(TAG, 0).
+-define(BATCH_MESSAGES, 1).
+
+-type batch() :: #{
+    ?TAG := batch,
+    ?BATCH_MESSAGES := [emqx_types:message()]
```
thalesmg, line 99: Nit, to keep style consistent:

```suggestion
-define(batch_messages, 1).

-type batch() :: #{
    ?tag := batch,
    ?batch_messages := [emqx_types:message()]
```

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
ieQu1: Site ID contains 64 bits of entropy, I assume that it should be random enough.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
ieQu1: Done. Supervisor won't have any children unless rocksdb backend is available.

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
```diff
@@ -82,42 +91,47 @@
 
 -type message_id() :: emqx_ds_storage_layer:message_id().
 
+-define(TAG, 0).
```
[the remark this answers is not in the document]
ieQu1: Thanks, fixed.

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
ieQu1: Fixed, thanks.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -440,10 +440,6 @@ del_subscription(TopicFilter, DSSessionId) ->
 %%--------------------------------------------------------------------
 
 create_tables() ->
-    ok = emqx_ds:open_db(?PERSISTENT_MESSAGE_DB, #{
```
ieQu1, line 443: This database is created in `emqx_persistent_message:init()`, no need to do it twice.

thalesmg (approved):

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
ieQu1: Also Windows support is dropped :)

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -0,0 +1,340 @@
...
+%% DS database metadata:
+-define(META_TAB, emqx_ds_builtin_metadata_tab).
+%% Mapping from Site to the actual Erlang node:
+-define(NODE_TAB, emqx_ds_builtin_node_tab).
+%% Shard metadata:
+-define(SHARD_TAB, emqx_ds_builtin_shard_tab).
+
+-record(?META_TAB, {
+    db :: emqx_ds:db(),
+    db_props :: emqx_ds_replication_layer:builtin_db_opts()
```
keynslug, line 73: Q: Just want to clarify some confusion.

I was under the impression that one of the purposes of generations is to allow changing [storage schema](https://github.com/emqx/emqx/blob/088de9476cd267f2d611ed33cbe1dfb628d8c96d/apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl#L63-L69), yet here the [options](https://github.com/emqx/emqx/blob/088de9476cd267f2d611ed33cbe1dfb628d8c96d/apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl#L55-L60) (from which the schema is derived) are fixed as part of DB record. Is it specifically only for the initial storage generation?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -0,0 +1,340 @@
...
+        {type, ordered_set},
+        {storage, rocksdb_copies},
+        {record_name, ?NODE_TAB},
+        {attributes, record_info(fields, ?NODE_TAB)}
+    ]),
+    ok = mria:create_table(?SHARD_TAB, [
+        {rlog_shard, ?SHARD},
+        {majority, Majority},
+        {type, ordered_set},
+        {storage, ram_copies},
```
keynslug, line 287: Q: Why do we use `ram_tables` for this state? Won't we lose it in the event of complete cluster shutdown (as opposed to `?META_TAB` contents)?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -0,0 +1,340 @@
...
+-spec sites() -> [site()].
+sites() ->
+    eval_qlc(qlc:q([Site || #?NODE_TAB{site = Site} <- mnesia:table(?NODE_TAB)])).
+
+-spec shard_leader(emqx_ds:db(), emqx_ds_replication_layer:shard_id()) ->
+    {ok, node()} | {error, no_leader_for_shard}.
+shard_leader(DB, Shard) ->
+    case mnesia:dirty_read(?SHARD_TAB, {DB, Shard}) of
+        [#?SHARD_TAB{leader = Leader}] ->
+            {ok, Leader};
```
keynslug, line 166: Q: Doesn't this return `{ok, undefined}` to the caller where there's no leader yet?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
ieQu1: It's a bug.

PR #11993, opened 2023-11-21
feat(sessds): introduce plugguble backends

savonarola:
Fixes [EMQX-11391](https://emqx.atlassian.net/browse/EMQX-11391)

## PR Checklist

- [x] Added tests for the changes (happy path is modified)


apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -215,3 +220,13 @@ next(DB, Iter, BatchSize) ->
...
 %% Internal functions
 %%================================================================================
+
+ds_module(#{backend := builtin}) ->
+    emqx_ds_replication_layer;
+ds_module(#{options := #{ds_module := Module}}) ->
+    Module.
+
+ds_config(#{backend := builtin} = Config) when not is_map_key(storage, Config) ->
+    Config#{storage => {emqx_ds_storage_bitfield_lts, #{}}};
```
ieQu1, line 230: `emqx_ds` module shouldn't make any assumptions about the configuration for the backend. It shouldn't inject the default configuration or modify it.

savonarola: But it already does, setting replication impl for built-in storage.

Should we inject that too?

savonarola: Morover, only it is only replication impl who knows how to make use of `storage` key.

ieQu1: No, storage key is a property of `emqx_ds_replication_layer` (which can be seen as a backend that _happens_ to reside in the same OTP application). It should be theoretically possible to pull `emqx_ds_replication_layer`, `emqx_ds_storage_layer`, ... away from `emqx_ds` application. 
`emqx_ds` module itself doesn't use or care about `storage`.

PR #11994, opened 2023-11-21
ci: remove windows from CI

zmstone:
Starting from v5.4.0, we will stop release windows packages.
Other windows related files are still around, such as emqx.cmd
will delete them in follow-up PRs.


ieQu1 (approved):

id (approved):

thalesmg (approved):

michaelcplusplus:
We have been successfully using EMQX Broker on Windows systems for our customers for several years. The system load has so far been moderate and we have not experienced any problems with the Windows installations.

Why is support for Windows being discontinued? Is there a way to continue using the Broker in versions above 5.4.0 on Windows?
Best regards
Michael

PR #12017, opened 2023-11-23
EMQX-11319 data backup http api

SergeTupchiy:
Fixes <issue-or-jira-number>

## PR Checklist


apps/emqx_management/src/emqx_mgmt_api_data_backup.erl
```diff
@@ -0,0 +1,361 @@
...
+    end.
+
+data_import(post, #{body := #{<<"filename">> := FileName} = Body}) ->
+    case safe_parse_node(Body) of
+        {error, Msg} ->
+            {400, #{code => 'BAD_REQUEST', message => Msg}};
+        FileNode ->
+            CoreNode = core_node(FileNode),
+            response(
+                emqx_mgmt_data_backup_proto_v1:import_file(CoreNode, FileNode, FileName, infinity)
```
SergeTupchiy, line 211: Only core node is allowed to import a backup, because it implies using `mnesia:restore/2`. Calling it on a replicant node would mean that Mnesia data is only restored locally.

apps/emqx_management/src/emqx_mgmt_api_data_backup.erl
```diff
@@ -0,0 +1,361 @@
...
+response({ok, Res}) ->
+    {200, Res};
+response(ok) ->
+    {204};
+response({error, Reason}) ->
+    {400, #{code => ?BAD_REQUEST, message => emqx_mgmt_data_backup:format_error(Reason)}}.
+
+list_backup_files(Page, Limit) ->
+    Start = Page * Limit - Limit + 1,
+    lists:sublist(list_backup_files(), Start, Limit).
```
SergeTupchiy, line 296: Naive paging, but I think the number of stored backup files is not expected to be very large...
Additionally, this simply lists `*.tar.gz` files in `<emqx data dir>/backup`. Adding some Mnesia DB layer for backups is possible but it would be hard to keep it in sync with FS state, especially taking into account the fact that users can directly manage `backup` dir without the dashboard..

ieQu1 (approved):

PR #12029, opened 2023-11-26
feat(sessds): provide QoS2 message replay support

keynslug:
Fixes [EMQX-11049](https://emqx.atlassian.net/browse/EMQX-11049).

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_message_ds_replayer.erl
+++ b/apps/emqx/src/emqx_persistent_message_ds_replayer.erl
@@ -19,7 +19,13 @@
 -module(emqx_persistent_message_ds_replayer).
 
 %% API:
--export([new/0, open/1, next_packet_id/1, replay/1, commit_offset/3, poll/3, n_inflight/1]).
+-export([new/0, open/1, next_packet_id/1, n_inflight/1]).
+
+-export([poll/4, replay/2, commit_offset/4, commit_marker/4]).
+
+-export([seqno_to_packet_id/1, packet_id_to_seqno/2]).
+
+-export([committed_until/2]).
 
 %% internal exports:
 -export([]).
@@ -27,13 +33,19 @@
 -export_type([inflight/0, seqno/0]).
 
 -include_lib("emqx/include/logger.hrl").
--include_lib("emqx_utils/include/emqx_message.hrl").
 -include("emqx_persistent_session_ds.hrl").
 
 -ifdef(TEST).
 -include_lib("proper/include/proper.hrl").
 -include_lib("eunit/include/eunit.hrl").
 -endif.
+
+-define(EPOCH_SIZE, 16#10000).
+
+-define(ACK, 0).
+-define(COMP, 1).
+
+-define(TRACK_FLAG(WHICH), (1 bsl WHICH)).
 
 %%================================================================================
 %% Type declarations
@@ -42,14 +54,19 @@
 %% Note: sequence numbers are monotonic; they don't wrap around:
 -type seqno() :: non_neg_integer().
 
+-type track() :: ack | comp.
+-type marker() :: rec.
+
 -record(inflight, {
     next_seqno = 1 :: seqno(),
-    acked_until = 1 :: seqno(),
+    commits = #{ack => 1, comp => 1, rec => 1} :: #{track() | marker() => seqno()},
     %% Ranges are sorted in ascending order of their sequence numbers.
     offset_ranges = [] :: [ds_pubrange()]
 }).
 
 -opaque inflight() :: #inflight{}.
+
+-type reply_fun() :: fun((seqno(), emqx_types:message()) -> emqx_session:reply()).
 
 %%================================================================================
 %% API funcions
@@ -61,10 +78,12 @@ new() ->
 
 -spec open(emqx_persistent_session_ds:id()) -> inflight().
 open(SessionId) ->
-    Ranges = ro_transaction(fun() -> get_ranges(SessionId) end),
-    {AckedUntil, NextSeqno} = compute_inflight_range(Ranges),
+    {Ranges, RecUntil} = ro_transaction(
+        fun() -> {get_ranges(SessionId), get_marker(SessionId, rec)} end
+    ),
+    {Commits, NextSeqno} = compute_inflight_range(Ranges),
     #inflight{
-        acked_until = AckedUntil,
+        commits = Commits#{rec => RecUntil},
         next_seqno = NextSeqno,
         offset_ranges = Ranges
     }.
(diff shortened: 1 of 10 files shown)


apps/emqx/src/emqx_persistent_session_ds.hrl
```diff
@@ -64,10 +68,27 @@
     %%   message in the range.
     iterator :: emqx_ds:iterator(),
     %% Reserved for future use.
-    misc = #{} :: map()
+    misc = #{} :: #{
+        %% What commit tracks this range is part of.
+        %% This is rarely stored: we only need to persist it when the range
+        %% contains QoS 2 messages.
+        ?T_tracks => non_neg_integer(),
```
keynslug, line 75: Q: Are we already past the point where we can add new record fields here? This would slightly help to improve code readability.

apps/emqx/src/emqx_persistent_message_ds_replayer.erl
```diff
@@ -75,59 +94,81 @@ next_packet_id(Inflight0 = #inflight{next_seqno = LastSeqno}) ->
     {seqno_to_packet_id(LastSeqno), Inflight}.
 
 -spec n_inflight(inflight()) -> non_neg_integer().
-n_inflight(#inflight{next_seqno = NextSeqno, acked_until = AckedUntil}) ->
-    range_size(AckedUntil, NextSeqno).
+n_inflight(#inflight{offset_ranges = Ranges}) ->
+    %% TODO
```
ieQu1, line 98: Just a vague idea: can we have two ephemeral (non-persistent) counters for this? One is for QoS1 messages, and another is for QoS2.
- Once the process reconnects, we reconstruct the value of this counters by scanning the ranges.
- Otherwise, we just update it with each message. PUBACK for QoS1 message decrements QoS1 counter; PUBCOMP for QoS2 message decrements QoS2 counter.
- n_inflight = counter_qos1 + counter_qos2.

I see the problem that it requires storing some properties (like QoS) for each outgoing message. But if we hold onto the batch for longer (we probable should, since it's required to process hooks for PUBACK), we could get this information from the batch itself.

apps/emqx/src/emqx_persistent_session_ds.hrl
ieQu1: I don't think we enabled table creation on the master just yet. Should be fine to modify the schema?

keynslug: Agree, I also think it should. Will make the necessary changes.

apps/emqx/src/emqx_persistent_message_ds_replayer.erl
```diff
@@ -227,84 +302,178 @@ find_checkpoints(Ranges) ->
...
-    RangeKept ++ discard_acked_ranges(SessionId, AckedUntil, Checkpoints, Rest);
-discard_acked_ranges(_SessionId, _AckedUntil, _Checkpoints, Ranges) ->
-    %% The rest of ranges (if any) still have unacked messages.
-    Ranges.
+    TComp =
+        case Until > CompUntil of
+            true -> ?TRACK_FLAG(?COMP) band Tracks;
+            false -> 0
+        end,
+    TAck bor TComp.
```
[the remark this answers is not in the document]
keynslug: Yeah, I can see why. Although this is in part deliberate: there are already a few places (typespecs / default values) that state there are only 2 of them right now (and that doesn't seem to change anytime soon), yet IMO making this function generic would make the code more complex and harder to understand.

apps/emqx/src/emqx_persistent_message_ds_replayer.erl
```diff
@@ -138,9 +182,23 @@ poll(SessionId, Inflight0, WindowSize) when WindowSize > 0, WindowSize < 16#7fff
         true ->
             %% TODO: Wrap this in `mria:async_dirty/2`?
             Streams = shuffle(get_streams(SessionId)),
-            fetch(SessionId, Inflight0, Streams, FreeSpace, [])
+            fetch(ReplyFun, SessionId, Inflight0, Streams, FreeSpace, [])
     end.
 
+-spec committed_until(track() | marker(), inflight()) -> seqno().
```
thalesmg, line 188: Nit: maybe add a comment saying that the returned seqno is _exclusive_ (i.e., itself is not committed).  Or _inclusive_, if I got it wrong.  :see_no_evil:

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -394,7 +421,24 @@ handle_timeout(_ClientInfo, get_streams, Session) ->
...
     {ok, replies(), session()}.
 replay(_ClientInfo, [], Session = #{inflight := Inflight0}) ->
-    {Replies, Inflight} = emqx_persistent_message_ds_replayer:replay(Inflight0),
+    AckedUntil = emqx_persistent_message_ds_replayer:committed_until(ack, Inflight0),
+    RecUntil = emqx_persistent_message_ds_replayer:committed_until(rec, Inflight0),
+    CompUntil = emqx_persistent_message_ds_replayer:committed_until(comp, Inflight0),
+    ReplyFun = fun
+        (_Seqno, #message{qos = ?QOS_0}) ->
+            {false, []};
+        (Seqno, #message{qos = ?QOS_1}) when Seqno < AckedUntil ->
```
thalesmg, line 430: (if the results of `committed_until` are inclusive) shouldn't this be `=<`?

keynslug: They are indeed exclusive. I.e. "committed until 42" means the last committed seqno is 41.

apps/emqx/src/emqx_persistent_message_ds_replayer.erl
keynslug: 34c9c02

apps/emqx/src/emqx_persistent_message_ds_replayer.erl
```diff
@@ -75,59 +97,81 @@ next_packet_id(Inflight0 = #inflight{next_seqno = LastSeqno}) ->
...
             {false, Inflight0}
     end.
 
--spec poll(emqx_persistent_session_ds:id(), inflight(), pos_integer()) ->
+-spec poll(reply_fun(), emqx_persistent_session_ds:id(), inflight(), pos_integer()) ->
     {emqx_session:replies(), inflight()}.
-poll(SessionId, Inflight0, WindowSize) when WindowSize > 0, WindowSize < 16#7fff ->
-    #inflight{next_seqno = NextSeqNo0, acked_until = AckedSeqno} =
-        Inflight0,
+poll(ReplyFun, SessionId, Inflight0, WindowSize) when WindowSize > 0, WindowSize < ?EPOCH_SIZE ->
```
ieQu1, line 172:
```suggestion
poll(ReplyFun, SessionId, Inflight0, WindowSize) when WindowSize > 0, WindowSize =< ?EPOCH_SIZE ->
```

?

keynslug: > Followed by the Two Byte Integer representing the Receive Maximum value.

The `?EPOCH_SIZE` is defined as `16#10000` (same constant that's used for the purpose of skipping seqnos), so it currently seems correct?

ieQu1: Right, sorry.

apps/emqx/src/emqx_persistent_session_ds.hrl
```diff
@@ -68,6 +76,17 @@
 }).
 -type ds_pubrange() :: #ds_pubrange{}.
 
+-record(ds_marker, {
```
ieQu1, line 79: It would be nice to have a comment explaining what "marker" is. Since `_MarkerName` is a free variable, it's not obvious from the definition that currently it currently has a single value of `rec` and that it tracks the last session sequence acked by the client.

Maybe even consider renaming the record to "ds_committed_offset", and "marker name" to "commit type". But it's entirely optional.

ieQu1 (approved):

PR #12053, opened 2023-11-29
1129 prepare for OTP 26

zmstone:
Prepare for OTP 26 in the CI.

* Change to use `peer` for cluster tests (more clean up is required in follow-up PRs)
* Change to match `+0.0` instead of `0.0` to mute the warning message (some dependencies still have this warning, needs to fix later)
* Changed triple-double-quotes to single-double-quotes to mute the compile warnings
* Flaky test fixes

### DIFF
--- a/apps/emqx/test/emqx_shared_sub_SUITE.erl
+++ b/apps/emqx/test/emqx_shared_sub_SUITE.erl
@@ -63,6 +63,7 @@ init_per_suite(Config) ->
         end,
     emqx_common_test_helpers:boot_modules(all),
     emqx_common_test_helpers:start_apps([]),
+    emqx_logger:set_log_level(debug),
     [{dist_pid, DistPid} | Config].
 
 end_per_suite(Config) ->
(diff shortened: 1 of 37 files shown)


apps/emqx/test/emqx_shared_sub_SUITE.erl
```diff
@@ -1253,34 +1254,25 @@ recv_msgs(Count, Msgs) ->
     end.
 
 start_slave(Name, Port) ->
```
id, line 1256: rename to `peer`?

id (approved):

apps/emqx/test/emqx_shared_sub_SUITE.erl
```diff
@@ -63,6 +63,7 @@ init_per_suite(Config) ->
         end,
     emqx_common_test_helpers:boot_modules(all),
     emqx_common_test_helpers:start_apps([]),
+    emqx_logger:set_log_level(debug),
```
savonarola, line 66: Do we need to keep this level?

savonarola (approved):

apps/emqx/test/emqx_shared_sub_SUITE.erl
zmstone: no. Will remove it in the next pr.

PR #12058, opened 2023-11-29
sync release-53 to master

id:


apps/emqx_connector/src/schema/emqx_connector_schema.erl
```diff
@@ -357,7 +391,17 @@ roots() ->
...
+            mk(
+                hoconsc:map(name, ref(emqx_bridge_http_schema, "config_connector")),
+                #{
+                    alias => [webhook],
+                    desc => <<"HTTP Connector Config">>,
+                    required => false
+                }
+            )}
+    ] ++ enterprise_fields_connectors();
 fields("node_status") ->
```
id, line 405: should we use string in both clauses, or atom in both clauses?

thalesmg: Some places that have clauses like `"get_connector"`, `"get_bridge_v2"` kinda rely on strings to match on parts of the field name.  This one in particular apparently is indifferent: it could use either.

Rory-Z (dismissed):

thalesmg (approved):

thalesmg (approved):

PR #11958, opened 2023-11-16
feat(sessds): preserve acks / replays in session state

keynslug:
Fixes [EMQX-9745](https://emqx.atlassian.net/browse/EMQX-9745).

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at f30b04f</samp>

This pull request enhances the persistent message and session features of emqx, by introducing a new API and data structure for replaying inflight messages, refactoring the session data structure and functions to use topic filters and subscriptions, and improving the performance and correctness of offset and range management. It also cleans up some unused or obsolete code in `emqx_persistent_session_ds.hrl`.

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [x] Added tests for the changes
- [ ] ~~Added property-based tests for code which performs user input validation~~
- [ ] Changed lines covered in coverage report
- [ ] Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] For internal contributor: there is a jira ticket to track this change
- [ ] ~~Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket~~
- [x] Schema changes are backward compatible

PR #12099, opened 2023-12-05
fix(sessds): make streams ephemeral

keynslug:
Fixes [EMQX-11537](https://emqx.atlassian.net/browse/EMQX-11537).

Release version: v5.4

In other words, do not store them in mnesia. Keep only references to them as part of pubranges, a reference now is a hash of a stream. As long as we use streams themselves as their identities this must be safe.

Pros:
* Now that a set of streams always reflects a set of current subscriptions, unsubscribe just works.
* Less objects to manage in mnesia.

## PR Checklist


ieQu1:
How does it play with the concepts of fully replayed streams? We need to keep them to make stream scheduler work

keynslug:
> We need to keep them to make stream scheduler work

Good point. There are already pubrange checkpoints which can probably hold `eos` in place of the last iterator, which supposedly would give the same effect. Anyway, this is worth a discussion.

ieQu1 (commented):
I am afraid this will become a premature pessimization. In fact, I think we should move more data _away_ from ds_pubrange. I introduced this table as a quick hack to save the mapping from packet sequence numbers to the iterators, in case we need to replay messages. 

- I don't want `ds_pubrange` to become a God object, upon which the entire session logic relies. 
- `ds_pubrange` objects are the most frequently written ones (https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_persistent_message_ds_replayer.erl#L276 - it's in the hot loop!; they can multiply uncontrollably if the client reads data in small chunks. 
- I think this object is the prime candidate for **removal** rather than extension

The number of stream objects, on the other hand, is well known since `emqx_ds` tries to minimize the number of streams.

ieQu1 (changes requested):
I am afraid this will become a premature pessimization. In fact, I think we should move more data _away_ from ds_pubrange. I introduced this table as a quick hack to save the mapping from packet sequence numbers to the iterators, in case we need to replay messages. 

- I don't want `ds_pubrange` to become a God object, upon which the entire session logic relies. 
- `ds_pubrange` objects are the most frequently written ones (https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_persistent_message_ds_replayer.erl#L276 - it's in the hot loop!; they can multiply uncontrollably if the client reads data in small chunks. 
- I think this object is the prime candidate for **removal** rather than extension

The number of stream objects, on the other hand, is well known since `emqx_ds` tries to minimize the number of streams.

keynslug:
Very arguable points. But I'll just scrap this PR in favor of a more conservative approach. A couple of things for the context / history.

> I introduced this table as a quick hack

This table was introduced as part of #11958.

> they can multiply uncontrollably if the client reads data in small chunks

They definitely will, if: 1) streams are "live" and constantly but infrequently receive new writes, 2) client rarely acks publishes, or acks them in large batches.

PR #12125, opened 2023-12-07
chore(ds): return DS key from `next` and add `update_iterator` callback

thalesmg:
Part of https://emqx.atlassian.net/browse/EMQX-10943

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds.erl
@@ -28,7 +28,7 @@
 -export([store_batch/2, store_batch/3]).
 
 %% Message replay API:
--export([get_streams/3, make_iterator/4, next/3]).
+-export([get_streams/3, make_iterator/4, update_iterator/3, next/3]).
 
 %% Misc. API:
 -export([]).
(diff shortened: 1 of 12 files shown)


apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -212,6 +218,11 @@ get_streams(DB, TopicFilter, StartTime) ->
 make_iterator(DB, Stream, TopicFilter, StartTime) ->
     ?module(DB):make_iterator(DB, Stream, TopicFilter, StartTime).
 
+-spec make_iterator_from_key(db(), stream(), topic_filter(), time(), message_key()) ->
```
thalesmg, line 221: I think I'll actually change this to:

```suggestion
-spec make_iterator_from_key(db(), iterator(), message_key()) ->
```

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -125,6 +128,9 @@
 -callback make_iterator(db(), ds_specific_stream(), topic_filter(), time()) ->
     make_iterator_result(ds_specific_iterator()).
 
+-callback make_iterator_from_key(db(), ds_specific_iterator(), message_key()) ->
```
ieQu1, line 131: Q: why does this function takes iterator instead of stream as the second argument?

thalesmg: Because the use case here is for the cache to construct a new iterator based on the last seen key.  Otherwise, we would need all the info to create an iterator from scratch: https://github.com/emqx/emqx/pull/12125#discussion_r1419260458

And when using `next`, the arguments do not contain the original arguments, but the iterators internally do.

ieQu1: I see, it makes sense. But a more appropriate name for the function I guess is `update_iterator`?

ieQu1 (approved):

PR #12127, opened 2023-12-07
fix(sessds): drop streams related to deleted subscriptions

keynslug:
Fixes [EMQX-11537](https://emqx.atlassian.net/browse/EMQX-11537).

## Summary

Also stop conflating same `emqx_ds:stream()` that is part of more than one subscription into a single `#ds_stream{}`. This could have been the source of subtle bugs, for instance when start times of those subscriptions differ significantly.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -88,7 +88,6 @@
 %% an atom, in theory (?).
 -type id() :: binary().
 -type topic_filter() :: emqx_types:topic().
--type topic_filter_words() :: emqx_ds:topic_filter().
 -type subscription_id() :: {id(), topic_filter()}.
 -type subscription() :: #{
     start_time := emqx_ds:time(),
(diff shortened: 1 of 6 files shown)


apps/emqx/src/emqx_persistent_session_ds.hrl
```diff
@@ -41,6 +45,8 @@
     ref :: _StreamRef,
     stream :: emqx_ds:stream(),
     rank :: emqx_ds:stream_rank(),
+    topic_filter :: emqx_types:topic(),
+    start_time :: emqx_ds:time(),
```
keynslug, line 49: Again, this needs special attention. I followed the same reasoning as [here](https://github.com/emqx/emqx/pull/12029/#discussion_r1405695081), but please shout loudly if it's not acceptable anymore.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -845,62 +846,73 @@ do_ensure_all_iterators_closed(_DSSessionID) ->
...
         ?PERSISTENT_MESSAGE_DB,
         Stream,
-        TopicFilter,
+        emqx_topic:words(TopicFilter),
         StartTime
     ),
     DSStream = #ds_stream{
         session = DSSessionId,
-        ref = StreamRef,
+        ref = emqx_persistent_sequence:next({?SESSION_STREAM_SEQ, DSSessionId}),
```
savonarola, line 903: Curious: what are pros and cons against just keeping the counter in session?

keynslug: Pros: slightly more isolated and reusable, a bit less complexity of maintaining a counter (updates don't need to be propagated anywhere), no need to do read/write cycle of the whole session. Cons: one more table brings maintenance burden. To be honest, that was just a quickest way to solve this, given that the data model is in flux we might even drop it quite soon.

ieQu1:
Fixed here: https://github.com/emqx/emqx/pull/12251

PR #12151, opened 2023-12-12
dont allow action types as v 1 bridge types

sstrigler:
Fixes [EMQX-11582](https://emqx.atlassian.net/browse/EMQX-11582)

Release version: v/e5.4.0

## PR Checklist

[EMQX-11582]: https://emqx.atlassian.net/browse/EMQX-11582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_bridge/src/emqx_action_info.erl
+++ b/apps/emqx_bridge/src/emqx_action_info.erl
@@ -266,6 +266,7 @@ get_info_map(Module) ->
                     {_BridgeV1TypeFun, _BridgeV1Types} = BridgeV1TypeTuple ->
                         BridgeV1TypeTuple;
                     [BridgeV1Type | _] = BridgeV1Types0 ->
+                        %% FIXME currently needed for kafka and gcp_pubsub
                         {BridgeV1Type, BridgeV1Types0};
                     BridgeV1Type ->
                         {BridgeV1Type, [BridgeV1Type]}
--- a/apps/emqx_bridge/src/emqx_action_info.erl
+++ b/apps/emqx_bridge/src/emqx_action_info.erl
@@ -37,6 +37,7 @@
 
 -callback bridge_v1_type_name() ->
     atom()
+    | [atom()]
     | {
         fun(({ActionConfig :: map(), ConnectorConfig :: map()}) -> Type :: atom()),
         TypeList :: [atom()]
@@ -65,9 +66,12 @@
 ]).
 
 %% ====================================================================
-%% Hadcoded list of info modules for actions
-%% TODO: Remove this list once we have made sure that all relevants
-%% apps are loaded before this module is called.
+%% FIXME: Hard-coded list of info modules for actions. Remove this list once we
+%% have made sure that all relevant apps are loaded before this module is
+%% called.
+%% Alternative idea: every-time an `emqx_action_info` implementation is loaded
+%% (-on_load(...)) calls emqx_action_info:add_action_info_module which updates
+%% the cache.
 %% ====================================================================
 
 -if(?EMQX_RELEASE_EDITION == ee).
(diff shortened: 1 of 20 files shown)


apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -1399,72 +1409,49 @@ bridge_v1_remove(
...
-    ConnectorType = connector_type(BridgeV2Type),
-    case emqx_connector_resource:get_channels(ConnectorType, ConnectorName) of
-        {ok, []} ->
-            false;
-        _ ->
-            true
-    end.
+connector_has_channels(ConnectorType, ConnectorName) ->
+    {ok, Channels} = emqx_connector_resource:get_channels(ConnectorType, ConnectorName),
+    length(Channels) =/= 0.
```
thalesmg, line 1454:
```suggestion
    case emqx_connector_resource:get_channels(ConnectorType, ConnectorName) of
        {ok, [_ | _]} -> true;
        _ -> false
    end.
```
?

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -851,6 +852,15 @@ connector_type(Type) ->
 bridge_v2_type_to_connector_type(Type) ->
     emqx_action_info:action_type_to_connector_type(Type).
 
+connector_id(BridgeType, BridgeName) ->
+    case lookup_conf(BridgeType, BridgeName) of
+        #{connector := ConnectorName} ->
+            ConnectorType = bin(connector_type(BridgeType)),
+            <<"connector:", (bin(ConnectorType))/binary, ":", (bin(ConnectorName))/binary>>;
```
thalesmg, line 859: `emqx_connector_resource:resource_id`?

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -1399,72 +1409,49 @@ bridge_v1_remove(
...
-    RemoveConnector = lists:member(connector, RemoveDeps),
-    case emqx_bridge_lib:maybe_withdraw_rule_action(BridgeType, BridgeName, RemoveDeps) of
+bridge_v1_check_deps_and_remove(BridgeV1Type, BridgeName, RemoveDeps) ->
+    BridgeV2Type = ?MODULE:bridge_v1_type_to_bridge_v2_type(BridgeV1Type),
+    case emqx_bridge_lib:maybe_withdraw_rule_action(BridgeV2Type, BridgeName, RemoveDeps) of
         ok ->
-            case remove(BridgeType, BridgeName) of
-                ok when RemoveConnector ->
-                    maybe_delete_channels(BridgeType, BridgeName, ConnectorName);
+            #{connector := ConnectorName} = lookup_conf(BridgeV2Type, BridgeName),
```
thalesmg, line 1420: This will change the behavior to crash if the action doesn't exist for whatever reason.  Can we ensure that's the case?  
It seems that this function is used elsewhere in the compatibility layer, maybe we could do this in a less explosive manner for now.

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -461,9 +463,9 @@ schema("/bridges_probe") ->
         }
     }.
 
-'/bridges'(post, #{body := #{<<"type">> := BridgeType0, <<"name">> := BridgeName} = Conf0}) ->
-    BridgeType = upgrade_type(BridgeType0),
-    case emqx_bridge:lookup(BridgeType, BridgeName) of
+'/bridges'(post, #{body := #{<<"type">> := BridgeType, <<"name">> := BridgeName} = Conf0}) ->
+    ActionType = upgrade_type(BridgeType),
+    case emqx_bridge:lookup(ActionType, BridgeName) of
```
thalesmg, line 468: Just a note for posterity: we should change `lookup` to operate only on v1 types.

apps/emqx_bridge/src/emqx_action_info.erl
```diff
@@ -259,9 +247,10 @@ action_info_modules(App) ->
 initial_info_map() ->
     #{
         action_type_names => #{},
-        bridge_v1_type_to_action_type => #{},
-        action_type_to_bridge_v1_type => #{},
-        action_type_to_connector_type => #{},
+        % [FIXME] remove 'mqtt/http' once implemented
+        bridge_v1_type_to_action_type => #{mqtt => mqtt, http => http},
+        action_type_to_bridge_v1_type => #{mqtt => mqtt, http => http},
+        action_type_to_connector_type => #{mqtt => mqtt, http => http},
```
thalesmg, line 253: Strange that http is needed to be added here, since it's already migrated from `webhook` to `http` :thinking: .

apps/emqx_bridge_mongodb/test/emqx_bridge_mongodb_SUITE.erl
```diff
@@ -150,7 +151,9 @@ init_per_testcase(_Testcase, Config) ->
 end_per_testcase(_Testcase, Config) ->
     clear_db(Config),
     delete_bridge(Config),
-    [] = emqx_connector:list(),
+    [emqx_bridge_v2:remove(T, N) || #{type := T, name := N} <- emqx_bridge_v2:list()],
+    [emqx_connector:remove(T, N) || #{type := T, name := N} <- emqx_connector:list()],
```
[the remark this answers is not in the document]
sstrigler: nice

apps/emqx_bridge/src/emqx_bridge_v2.erl
sstrigler: That would hide an unexpected error

thalesmg: The idea for the suggestion is: if the channel doesn't exist, it doesn't have channels.

sstrigler: you mean the connector... hm... I will change even though personally I'm not a big fan of such fuzziness.

thalesmg: Oops, yes, the connector.  :speak_no_evil: 

I also like strictness.  But I think that, in this case, it's semantically correct to return false if the connector doesn't exist.

thalesmg: An action pointing to a non-existent connector, sure, shouldn't be a valid state, but I'm just focusing on this specific function.

apps/emqx_bridge/src/emqx_action_info.erl
thalesmg: I think this is a side-effect of renaming the http bridge in V1 schema: https://github.com/emqx/emqx/blob/79e7327042d2a6cf9010681088a811fb57f32512/apps/emqx_bridge/src/schema/emqx_bridge_schema.erl#L171-L175

I believe this renaming shouldn't have been done, and that we should treat v1 schema as immutable as possible now that we're migrating to actions/connectors/sources.

PR #12157, opened 2023-12-12
feat(ds): add an API for making new generations

lafirest:
Fixes [EMQX-11223](https://emqx.atlassian.net/browse/EMQX-11223)

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds.erl
@@ -22,7 +22,7 @@
 -module(emqx_ds).
 
 %% Management API:
--export([open_db/2, drop_db/1]).
+-export([open_db/2, update_db_config/2, add_generation/1, drop_db/1]).
 
 %% Message storage API:
 -export([store_batch/2, store_batch/3]).
@@ -95,7 +95,7 @@
 
 %% Timestamp
 %% Earliest possible timestamp is 0.
-%% TODO granularity?  Currently, we should always use micro second, as that's the unit we
+%% TODO granularity?  Currently, we should always use milliseconds, as that's the unit we
 %% use in emqx_guid.  Otherwise, the iterators won't match the message timestamps.
 -type time() :: non_neg_integer().
 
@@ -123,6 +123,10 @@
 %%================================================================================
 
 -callback open_db(db(), create_db_opts()) -> ok | {error, _}.
+
+-callback update_db_config(db(), create_db_opts()) -> ok | {error, _}.
+
+-callback add_generation(db()) -> ok | {error, _}.
 
 -callback drop_db(db()) -> ok | {error, _}.
 
@@ -153,6 +157,24 @@ open_db(DB, Opts = #{backend := Backend}) when Backend =:= builtin orelse Backen
         end,
     persistent_term:put(?persistent_term(DB), Module),
     ?module(DB):open_db(DB, Opts).
+
+-spec update_db_config(db(), create_db_opts()) -> ok.
+update_db_config(DB, Opts) ->
+    case persistent_term:get(?persistent_term(DB), undefined) of
+        undefined ->
+            ok;
+        Module ->
+            Module:update_db_config(DB, Opts)
+    end.
+
+-spec add_generation(db()) -> ok.
+add_generation(DB) ->
+    case persistent_term:get(?persistent_term(DB), undefined) of
+        undefined ->
+            ok;
+        Module ->
+            Module:add_generation(DB)
+    end.
 
 %% @doc TODO: currently if one or a few shards are down, they won't be
 
--- a/apps/emqx_durable_storage/src/emqx_ds.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds.erl
@@ -22,7 +22,7 @@
 -module(emqx_ds).
 
 %% Management API:
--export([open_db/2, update_db_config/2, add_generation/1, drop_db/1]).
+-export([open_db/2, add_generation/2, add_generation/1, drop_db/1]).
 
 %% Message storage API:
 -export([store_batch/2, store_batch/3]).
@@ -124,9 +124,9 @@
 
 -callback open_db(db(), create_db_opts()) -> ok | {error, _}.
 
--callback update_db_config(db(), create_db_opts()) -> ok | {error, _}.
-
 -callback add_generation(db()) -> ok | {error, _}.
+
+-callback add_generation(db(), create_db_opts()) -> ok | {error, _}.
 
 -callback drop_db(db()) -> ok | {error, _}.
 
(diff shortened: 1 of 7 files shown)


apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -116,6 +116,8 @@
 
 -callback open_db(db(), create_db_opts()) -> ok | {error, _}.
 
+-callback update_db_config(db(), create_db_opts()) -> ok | {error, _}.
```
savonarola, line 119: The "update" naming hides any other applications of generation usage, like retention; do we intend this?

@lafirest  @ieQu1 what do you think?

ieQu1: Sorry, I'm not sure I understood the question. This callback signature looks fine to me.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -197,6 +207,7 @@ make_iterator(
 next(Shard, Iter = #{?tag := ?IT, ?generation := GenId, ?enc := GenIter0}, BatchSize) ->
     #{module := Mod, data := GenData} = generation_get(Shard, GenId),
     Current = generation_current(Shard),
+
```
ieQu1, line 210: Accidental change?

apps/emqx_durable_storage/src/emqx_ds.erl
savonarola: In other words, do we plan to create new generations _only_ when we want to update the config?

ieQu1: I see, you're right. We may indeed need to create new generations without updating the config.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -116,6 +116,10 @@
...
+
+-callback add_generation(db()) -> ok | {error, _}.
```
savonarola, line 121: Maybe we leave just `add_generation` with opts? Or `add_generation` with and without opts.

Otherwise the api may be a bit confusing:
* if we want to add gen changing opts, then we should use `update_db_config`
* if we want to add gen keeping opts, then we should use `add_generation`

lafirest: Yes,  them can be combined by `add_generation/0`, `add_generation/1`,  @ieQu1 does such modification meet our goals?

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -154,6 +158,24 @@ open_db(DB, Opts = #{backend := Backend}) when Backend =:= builtin orelse Backen
...
     ?module(DB):open_db(DB, Opts).
 
+-spec add_generation(db()) -> ok.
+add_generation(DB) ->
+    case persistent_term:get(?persistent_term(DB), undefined) of
+        undefined ->
+            ok;
+        Module ->
+            Module:add_generation(DB)
+    end.
```
thalesmg, line 168: Tiny nit (applies to the other arity as well): since we expect the DB to exist when calling this (differently from `drop`), we might be more assertive. 

```suggestion
    ?module(DB):add_generation(DB).
```

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -369,3 +459,30 @@ eval_qlc(Q) ->
...
+    filter_shards(DB, const(true)).
+
+-spec filter_shards(emqx_ds:db(), fun((_) -> boolean())) ->
+    [emqx_ds_replication_layer:shard_id()].
+filter_shards(DB, Predicte) ->
+    filter_shards(DB, Predicte, fun(#?SHARD_TAB{shard = {_, ShardId}}) ->
+        ShardId
+    end).
+
+filter_shards(DB, Predicte, Maper) ->
```
thalesmg, line 473:
```suggestion
filter_shards(DB, Predicate, Mapper) ->
```

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -346,7 +436,7 @@ create_shards(DB, NShards, ReplicationFactor) ->
             Hashes0 = [{hash(Shard, Site), Site} || Site <- AllSites],
             Hashes = lists:sort(Hashes0),
             {_, Sites} = lists:unzip(Hashes),
-            [First | _] = ReplicaSet = lists:sublist(Sites, 1, ReplicationFactor),
+            [First | ReplicaSet] = lists:sublist(Sites, 1, ReplicationFactor),
```
thalesmg, line 439: Q: shouldn't the whole replicaset also include the first node, since it also contains the data?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -21,20 +21,26 @@
 %% implementation details from this module.
 -module(emqx_ds_replication_layer_meta).
 
+-compile(inline).
```
thalesmg, line 24: Q: is this an intended change?  Maybe it was forgotten here?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
lafirest: I think these two place have a little conflict https://github.com/emqx/emqx/blob/363c56c0857b076d07db678e6a78deaed8a2e9b3/apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl#L162, if the `ReplicaSet` included all sites, here should not check the `InSync` set?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
lafirest: Yes, I don't want to write too much table traversal, but also do not like introducing costs of the anonymous function in the general traversal

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -145,22 +154,34 @@ start_link() ->
...
+    end).
+
+-spec my_owned_shards(emqx_ds:db()) -> [emqx_ds_replication_layer:shard_id()].
+my_owned_shards(DB) ->
+    Self = node(),
+    filter_shards(DB, fun(#?SHARD_TAB{leader = Leader}) ->
+        Self =:= Leader
+    end).
+
+-spec leader_nodes(emqx_ds:db()) -> [node()].
```
ieQu1, line 173: Should we allow just one leader?

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -154,6 +158,24 @@ open_db(DB, Opts = #{backend := Backend}) when Backend =:= builtin orelse Backen
...
+
+-spec add_generation(db(), create_db_opts()) -> ok.
+add_generation(DB, Opts) ->
+    case persistent_term:get(?persistent_term(DB), undefined) of
+        undefined ->
+            ok;
```
ieQu1, line 174: I think it's better to return an error or throw an exception if the database doesn't exist.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -154,6 +158,24 @@ open_db(DB, Opts = #{backend := Backend}) when Backend =:= builtin orelse Backen
     persistent_term:put(?persistent_term(DB), Module),
     ?module(DB):open_db(DB, Opts).
 
+-spec add_generation(db()) -> ok.
+add_generation(DB) ->
+    case persistent_term:get(?persistent_term(DB), undefined) of
+        undefined ->
+            ok;
```
ieQu1, line 165: I think it's better to return an error or throw an exception if the database doesn't exist.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
ieQu1: Indeed, thanks.

ieQu1 (approved):

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
thalesmg: My understanding was that replica set means the set of all nodes that contain a replica of the data, and in-sync was the subset of those that are caught up with replication.

HJianBo (approved):

PR #12166, opened 2023-12-14
feat: refactor MQTT bridge to source, action, and connector

kjellwinblad:
This commit:

* refactors the MQTT V1 bridge into connector, source and action
* Extends the compatibility layer so it works for sources
* Fixes the MQTT bridge test suite so that all test cases passes

We still need to add a HTTP API handling sources. Also, we still need to add HTTP API example schemes and examples for the MQTT connector/action/source.

We should also make sure that we handle the corner cases of the MQTT V1 bridge automatic upgrade downgrade in a sufficiently good way:

* An error is currently thrown when converting an MQTT V1 bridge without egress or ingress config.
* If there is a source and action with the same name we will currently throw an error in the compatibility layer.
* We will also throw an error when converting an MQTT V1 bridge with both ingress and egress.

The above is probably the right thing to do  but we have to make sure that we return a reasonable error to the user when this happens.

(partly)
Fixes:
https://emqx.atlassian.net/browse/EMQX-11489

Fixes <issue-or-jira-number>

Release version: v/e5.?

## PR Checklist


apps/emqx_bridge/src/emqx_action_info.erl
```diff
@@ -49,19 +52,26 @@
...
--callback bridge_v1_config_to_connector_config(BridgeV1Config :: map()) -> map().
+-callback bridge_v1_config_to_connector_config(BridgeV1Config :: map()) ->
+    map() | {ConnectorTypeName :: atom(), map()}.
 %% Define this if the automatic config upgrade is not enough for the bridge.
 %% If you want to make use of the automatic config upgrade, you can call
 %% emqx_action_info:transform_bridge_v1_config_to_action_config/4 in your
 %% implementation and do some adjustments on the result.
 -callback bridge_v1_config_to_action_config(BridgeV1Config :: map(), ConnectorName :: binary()) ->
-    map().
+    map() | {source | action, ActionTypeName :: atom(), map()} | 'none'.
```
thalesmg, line 62: Note to self: revisit this "none".

kjellwinblad (commented):
Just had a quick look but what I saw looks good to me.

(I guess the sources API still needs to be added)

thalesmg:
> (I guess the sources API still needs to be added)

yes, I was still doing some refactoring in preparation to add the sources API

thalesmg:
Still missing tests for the whole `/sources` API, but some simple manual tests indicate that it seems to work, and old tests using the V1 API are passing with minor changes at most.  So I think this can be reviewed.

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -48,6 +48,11 @@
 -export([lookup_from_local_node/2]).
 -export([get_metrics_from_local_node/2]).
 
+%% only for testting/mocking
```
qzhuyan, line 51: nit `testing`

qzhuyan (dismissed):

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_pubsub_schema.erl
```diff
@@ -0,0 +1,143 @@
...
+        hoconsc:mk(
+            hoconsc:ref(?MODULE, ingress_parameters),
+            #{
+                required => true,
+                desc => ?DESC("source_parameters")
+            }
+        )
+    );
+fields(ingress_parameters) ->
+    Fields0 = emqx_bridge_mqtt_connector_schema:fields("ingress"),
```
thalesmg, line 83: - [ ] currently, this contains 2 nested keys: `local` and `remote`.  We probably want to flatten this structure.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_pubsub_schema.erl
```diff
@@ -0,0 +1,143 @@
...
+        hoconsc:mk(
+            hoconsc:ref(?MODULE, action_parameters),
+            #{
+                required => true,
+                desc => ?DESC("action_parameters")
+            }
+        )
+    );
+fields(action_parameters) ->
+    Fields0 = emqx_bridge_mqtt_connector_schema:fields("egress"),
```
thalesmg, line 59: - [ ] currently, this contains 2 nested keys: `local` and `remote`.  We probably want to flatten this structure.

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -1136,7 +1353,52 @@ bridge_v1_lookup_and_transform_helper(
     end.
 
 lookup_conf(Type, Name) ->
-    case emqx:get_config([?ROOT_KEY, Type, Name], not_found) of
+    lookup_conf(?ROOT_KEY_ACTIONS, Type, Name).
+
+lookup_conf_if_one_of_sources_actions(Type, Name) ->
```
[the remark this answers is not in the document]
kjellwinblad: Fixed

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -1082,6 +1286,19 @@ bridge_v1_lookup_and_transform(ActionType, Name) ->
             Error
     end.
 
+lookup_actions_or_sources(ActionType, Name) ->
```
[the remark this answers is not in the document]
kjellwinblad: Fixed

apps/emqx_bridge/src/emqx_bridge_api.erl
```diff
@@ -547,9 +552,17 @@ schema("/bridges_probe") ->
         Id,
         case emqx_bridge_v2:is_bridge_v2_type(BridgeType) of
             true ->
-                BridgeV2Type = emqx_bridge_v2:bridge_v2_type_to_connector_type(BridgeType),
-                ok = emqx_bridge_v2:reset_metrics(BridgeV2Type, BridgeName),
-                ?NO_CONTENT;
+                try
+                    ConfRootKey = emqx_bridge_v2:get_conf_root_key_if_only_one(
```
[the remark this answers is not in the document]
kjellwinblad: Fixed

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -1575,7 +1871,7 @@ actions_convert_from_connectors(Conf) ->
         Conf
     ).
```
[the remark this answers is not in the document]
kjellwinblad: After spending some time looking into `emqx_action_info:action_convert_from_connector` I'm still not sure what it is doing. However, I think we can improve this naming or add a comment in later PR since improving the readability of this is unrelated to the main purpose of this PR.

apps/emqx_bridge/src/emqx_action_info.erl
thalesmg: Maybe `undefined` is more common?

But we can rename it in a follow up, as no place seems to be using it yet.

qzhuyan (approved):

PR #12171, opened 2023-12-14
feat(ds): add cache

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10943

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -81,9 +81,6 @@
 
 -export([print_session/1, seqno_diff/4]).
 
-%% `emqx_config_handler' API
--export([post_config_update/5, propagated_post_config_update/5]).
-
 -ifdef(TEST).
 -export([
     session_open/2,
@@ -901,55 +898,6 @@ expiry_interval(ConnInfo) ->
 
 bump_interval() ->
     emqx_config:get([session_persistence, last_alive_update_interval]).
-
-%%--------------------------------------------------------------------------------
-%% `emqx_config_handler' API
-%%--------------------------------------------------------------------------------
-
--define(ROOT_KEY, session_persistence).
--define(CACHE_PREFETCH_PATH, [?ROOT_KEY, cache_prefetch_topic_filters]).
--define(CACHE_GC_INTERVAL_PATH, [?ROOT_KEY, cache_gc_interval]).
-
-post_config_update(
-    ?CACHE_PREFETCH_PATH, _Req, NewConf, OldConf, _AppEnv
-) ->
-    handle_update_cache_prefetch_topic_filters(NewConf, OldConf),
-    ok;
-post_config_update(
-    ?CACHE_GC_INTERVAL_PATH, _Req, NewConf, _OldConf, _AppEnv
-) ->
-    handle_update_cache_gc_interval(NewConf),
-    ok;
-post_config_update(_Path, _Req, _NewConf, _OldConf, _AppEnv) ->
-    ok.
-
-propagated_post_config_update(
-    ?CACHE_PREFETCH_PATH, _Req, NewConf, OldConf, _AppEnvs
-) ->
-    handle_update_cache_prefetch_topic_filters(NewConf, OldConf),
-    ok;
-propagated_post_config_update(
-    ?CACHE_GC_INTERVAL_PATH, _Req, NewConf, _OldConf, _AppEnvs
-) ->
-    handle_update_cache_gc_interval(NewConf),
-    ok;
-propagated_post_config_update(_Path, _UpdateReq, _NewConf, _OldConf, _AppEnvs) ->
-    ok.
-
-handle_update_cache_prefetch_topic_filters(NewTopics, OldTopics) ->
-    Added = lists:usort(NewTopics -- OldTopics),
-    Removed = lists:usort(OldTopics -- NewTopics),
-    emqx_ds_cache:remove_cached_topic_filters(?PERSISTENT_MESSAGE_DB, Removed),
-    emqx_ds_cache:add_cached_topic_filters(?PERSISTENT_MESSAGE_DB, Added),
-    ok.
-
-handle_update_cache_gc_interval(NewGCInterval) ->
-    emqx_ds_cache:set_gc_interval(?PERSISTENT_MESSAGE_DB, NewGCInterval),
-    ok.
-
--undef(CACHE_GC_INTERVAL_PATH).
--undef(CACHE_PREFETCH_PATH).
--undef(ROOT_KEY).
 
 %%--------------------------------------------------------------------
 %% SeqNo tracking
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -81,6 +81,9 @@
 
 -export([print_session/1, seqno_diff/4]).
 
+%% `emqx_config_handler' API
+-export([post_config_update/5, propagated_post_config_update/5]).
+
 -ifdef(TEST).
 -export([
     session_open/2,
@@ -713,7 +716,7 @@ enqueue_batch(IsReplay, BatchSize, Srs0, Session = #{inflight := Inflight0}, Cli
             true -> ItBegin0;
             false -> ItEnd0
         end,
-    case emqx_ds:next(?PERSISTENT_MESSAGE_DB, ItBegin, BatchSize) of
+    case emqx_ds:next(?PERSISTENT_MESSAGE_DB, ItBegin, BatchSize, #{use_cache => true}) of
         {ok, ItEnd, Messages} ->
             {Inflight, LastSeqnoQos1, LastSeqnoQos2} = process_batch(
                 IsReplay, Session, ClientInfo, FirstSeqnoQos1, FirstSeqnoQos2, Messages, Inflight0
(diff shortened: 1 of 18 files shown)


apps/emqx_durable_storage/src/emqx_ds_cache.erl
```diff
@@ -0,0 +1,513 @@
...
+%%--------------------------------------------------------------------------------
+%% `gen_server' API
+%%--------------------------------------------------------------------------------
+
+-spec init(start_opts()) -> {ok, state()}.
+init(Opts) ->
+    process_flag(trap_exit, true),
+    DB = maps:get(db, Opts),
+    Tid = create_table(),
+    persistent_term:put(?cache_ptkey(DB), Tid),
```
savonarola, line 162: Why don't we use just a named table?

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -230,12 +242,36 @@ update_iterator(DB, OldIter, DSKey) ->
 
 -spec next(db(), iterator(), pos_integer()) -> next_result().
 next(DB, Iter, BatchSize) ->
-    ?module(DB):next(DB, Iter, BatchSize).
+    Opts = #{use_cache => false},
+    next(DB, Iter, BatchSize, Opts).
+
+-spec next(db(), iterator(), pos_integer(), next_opts()) -> next_result().
+next(DB, Iter, BatchSize, Opts) ->
+    ?module(DB):next(DB, Iter, BatchSize, Opts).
```
savonarola, line 250: Seems that the caching is done on the replication layer, i.e. inside the concrete `emqx_ds` implementation. However, the caching options are raised to the `emqx_ds` interface. 

So, on the one hand, the implementations should be aware of this option; on the other hand, they cannot use the caching bc. it is replication backend-specific.

So maybe we should
* Either lift caching to the `emqx_ds` level, so that other backends could get it for free;
* Or remove caching from the `emqx_ds` interface, so that each backend opaquely implements caching (if needed)
?

apps/emqx_durable_storage/src/emqx_ds_cache.erl
```diff
@@ -0,0 +1,513 @@
...
+    ets:fun2ms(fun(#entry{inserted_at = InsertedAt}) when
+        InsertedAt + Threshold =< Now
+    ->
+        true
+    end).
+
+now_ms() ->
+    erlang:system_time(millisecond).
+
+start_timer(Event, Timeout) ->
```
savonarola, line 501: nit: `emqx_utils:start_timer`

apps/emqx_durable_storage/src/emqx_ds_cache.erl
```diff
@@ -0,0 +1,513 @@
...
+) ->
+    {Seqno, [entry()]}
+when
+    Seqno :: pos_integer().
+to_entries(Now, Seqno, Stream, Batch) ->
+    to_entries(Now, Seqno, Stream, Batch, []).
+
+-spec to_entries(
+    integer(),
+    Seqno,
```
savonarola, line 367: nit: mb add types for Now and Seqno

```
-spec to_entries(
    timestamp(),
    seqno(),
    emqx_ds:stream(),
    [{emqx_ds:message_key(), emqx_types:message()}],
    [entry()]
) ->
    {seqno(), [entry()]}
```

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -1029,6 +1032,57 @@ receive_maximum(ConnInfo) ->
...
+) ->
+    handle_update_cache_gc_interval(NewConf),
+    ok;
+propagated_post_config_update(_Path, _UpdateReq, _NewConf, _OldConf, _AppEnvs) ->
+    ok.
+
+handle_update_cache_prefetch_topic_filters(NewTopics, OldTopics) ->
+    Added = lists:usort(NewTopics -- OldTopics),
+    Removed = lists:usort(OldTopics -- NewTopics),
+    emqx_ds_cache:remove_cached_topic_filters(?PERSISTENT_MESSAGE_DB, Removed),
```
savonarola, line 1072: Seems that the config is for persistent sessions in general;
However it configures `emqx_ds_cache` which belongs to a concrete backend (replication layer)

savonarola: Maybe we could somehow isolate `emqx_ds` universal code from concrete "main" implementation 😓 Because otherwise it is unclear that some shortcuts from universal code to the replication backend break `emqx_ds` interface 🤔

apps/emqx_durable_storage/src/emqx_ds_cache.erl
```diff
@@ -0,0 +1,513 @@
...
+start_link(Opts) ->
+    #{db := DB} = Opts,
+    gen_server:start_link(?REF(DB), ?MODULE, Opts, []).
+
+-spec stop(emqx_ds:db()) -> ok.
+stop(DB) ->
+    gen_server:stop(?REF(DB)).
+
+-spec child_spec(emqx_ds:db(), start_opts()) -> supervisor:child_spec().
+child_spec(DB, Opts) ->
```
savonarola, line 111: Unused?

apps/emqx_durable_storage/src/emqx_ds.erl
savonarola: Or is this the intended methos of integration? 🤔
Should the other backends, like replication layer, query the "global" cache themselves, when use_cache is true? And query the db otherwise?

apps/emqx_durable_storage/src/emqx_ds_cache.erl
thalesmg: Yes, each DB will have its own cache.

apps/emqx_durable_storage/src/emqx_ds_cache.erl
thalesmg: that one wraps the message with `{timeout, TRef, _}`

apps/emqx/src/emqx_persistent_session_ds.erl
thalesmg: I think that the PS DS cache may be the same for all backends, since the DB is the same.  There's no need to differentiate between the backends in this case.  

For example, FDB may just do the same wrapping as replication_layer does.  I did it in replication layer because the backend needs to implement `last_seen_key` in order for the cache to work properly.

apps/emqx_durable_storage/src/emqx_ds.erl
thalesmg: > Either lift caching to the emqx_ds level, so that other backends could get it for free;

The cache relies on `last_seen_key` being implemented to work properly, so I think it would be best left for each backend to do similar wrapping :sweat_smile: 

WDYT?

apps/emqx_durable_storage/src/emqx_ds_cache.erl
thalesmg: :heavy_check_mark: :scissors:

apps/emqx_durable_storage/src/emqx_ds_cache.erl
thalesmg: :heavy_check_mark:

apps/emqx/src/emqx_persistent_session_ds.erl
savonarola: >in replication layer because the backend needs to implement last_seen_key in order for the cache to work properly.

🤔  But we have `last_seen_key` in the `emqx_ds` behaviour... so we can be sure that it is implemented from the `emqx_ds` layer

thalesmg: it's an optional callback: https://github.com/emqx/emqx/pull/12171/files#diff-c7f537324b94a1e8f072e809aeb185a9adda2c89d8aaea78438c3aa4e5c5f4c1R151

thalesmg: some DBs/backends might not be able to implement it; for example, a backend for retained messages might not implement it, as iterations will probably be done by scanning all topics matching a filter without any strict notion of time/order.

apps/emqx_durable_storage/src/emqx_ds_cache.erl
```diff
@@ -0,0 +1,510 @@
...
+            {ok, LastKey, Batch};
+        [] when Key =:= ?EOS_KEY(Stream) ->
+            ?tp(ds_cache_eos_found, #{}),
+            {ok, end_of_stream};
+        [] ->
+            %% Possible if entry was removed just before we started iterating, or if only
+            %% the last seen key of the iterator is contained in the cache.  If the latter
+            %% case, then we should return a cache "hit" with empty results; otherwise,
+            %% clients that are up to date with their iterators will start spamming RPCs
+            %% again.
```
thalesmg, line 322: TODO: add some tests to check this.

thalesmg: Should rethink and revert this: if a topic filter is later removed, any client whose last seen key is in the cache might get forever stuck...

thalesmg:
Note: we may need to add back the callback to extract the topic filter from an iterator...

Otherwise, if we cache `foo/#` but a client subscribes to `foo/1`, it may receive messages from unwanted topics...

apps/emqx_durable_storage/src/emqx_ds_cache.erl
thalesmg: One idea is to return `false` if the stream is not currently being tracked, otherwise return an empty batch to prevent the client from doing the RPC.

apps/emqx_durable_storage/test/emqx_ds_cache_SUITE.erl
```diff
@@ -0,0 +1,397 @@
...
+                    2_000
+                ),
+            [{_Rank, Stream}] = emqx_ds:get_streams(DB, TopicFilter, NowMS),
+            {ok, Iter0} = emqx_ds:make_iterator(DB, Stream, TopicFilter, NowMS),
+            {ok, Iter1, Batch1} = emqx_ds:next(DB, Iter0, 1, #{use_cache => false}),
+            {ok, _Iter2, Batch2} = emqx_ds:next(DB, Iter1, 2, #{use_cache => true}),
+            %% We don't report a cache miss in this case, as this case is equivalent to
+            %% when the client is up to date with the stream and it's the cache that needs
+            %% to fetch more.
+            %% TODO: what to do if the client started from an old key and it got GC'ed?
```
thalesmg, line 218: actually, it'll then notice that last seen key is missing from the cache and fetch again; should test this.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -1029,6 +1032,57 @@ receive_maximum(ConnInfo) ->
...
+
+post_config_update(
+    ?CACHE_PREFETCH_PATH, _Req, NewConf, OldConf, _AppEnv
+) ->
+    handle_update_cache_prefetch_topic_filters(NewConf, OldConf),
+    ok;
+post_config_update(
+    ?CACHE_GC_INTERVAL_PATH, _Req, NewConf, _OldConf, _AppEnv
+) ->
+    handle_update_cache_gc_interval(NewConf),
```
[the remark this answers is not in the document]
thalesmg: Removed

apps/emqx/src/emqx_persistent_session_ds_sup.erl
```diff
@@ -55,7 +55,8 @@ do_init(_Opts) ->
     },
     CoreChildren = [
         worker(session_gc_worker, emqx_persistent_session_ds_gc_worker, []),
-        worker(message_gc_worker, emqx_persistent_message_ds_gc_worker, [])
+        worker(message_gc_worker, emqx_persistent_message_ds_gc_worker, []),
+        worker(persistent_message_cache, emqx_ds_cache, [emqx_persistent_message:cache_opts()])
```
thalesmg, line 59: Should we spawn this under `emqx_ds_builtin_db_sup`?

ieQu1: Yes, I think it's a good idea. Also, I don't see why don't replicant have the cache?

thalesmg:
#12453

Issue #12187, opened 2023-12-17
allow jwt acl to passthrough


zmstone:
### What would you like to be added or enhanced?

Have done a bit more playing around, haven't really gotten anywhere.  It does definitely seem like any topics not in the jwt acl are just rejected instantly without ever touching the authZ chain.  Unless there's some setting to control that I haven't spotted, that's a real shame

### Why is this needed?

to make jwt acl claims work with other authz sources.
or let the general `no_match` config to decide the allow or deny action.

PR #12189, opened 2023-12-17
feat: support regular raw acl rule in jwt acl claim

zmstone:
Fixes https://github.com/emqx/emqx/issues/12187 and [EMQX-11372](https://emqx.atlassian.net/browse/EMQX-11372)
Release version: `v/e5.5.0`

## PR Checklist

### DIFF
--- a/apps/emqx_auth_jwt/src/emqx_authn_jwt.erl
+++ b/apps/emqx_auth_jwt/src/emqx_authn_jwt.erl
@@ -219,14 +219,24 @@ verify(undefined, _, _, _) ->
 verify(JWT, JWKs, VerifyClaims, AclClaimName) ->
     case do_verify(JWT, JWKs, VerifyClaims) of
         {ok, Extra} ->
-            {ok, acl(Extra, AclClaimName)};
+            try
+                {ok, acl(Extra, AclClaimName)}
+            catch
+                throw:{bad_acl_rule, Reason} ->
+                    %% it's a invalid token, so ok to log
+                    ?TRACE_AUTHN_PROVIDER("bad_acl_rule", Reason#{jwt => JWT}),
+                    {error, bad_username_or_password}
+            end;
         {error, {missing_claim, Claim}} ->
+            %% it's a invalid token, so it's ok to log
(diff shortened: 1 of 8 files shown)


apps/emqx_auth_jwt/src/emqx_authn_jwt.erl
```diff
@@ -219,14 +219,24 @@ verify(undefined, _, _, _) ->
 verify(JWT, JWKs, VerifyClaims, AclClaimName) ->
     case do_verify(JWT, JWKs, VerifyClaims) of
         {ok, Extra} ->
-            {ok, acl(Extra, AclClaimName)};
+            try
+                {ok, acl(Extra, AclClaimName)}
+            catch
+                throw:{bad_acl_rule, Reason} ->
+                    %% it's a invalid token, so ok to log
+                    ?TRACE_AUTHN_PROVIDER("bad_acl_rule", Reason#{jwt => JWT}),
```
savonarola, line 227: Previously, the validation was lazy, that is, invalid rules were handled on authz stage. This is kept for v1 rules;
For v2 we now break connection, don't we consider this to be a breaking change?

apps/emqx_auth_jwt/test/emqx_authz_jwt_SUITE.erl
```diff
@@ -99,7 +99,47 @@ t_topic_rules(_Config) ->
         },
         <<"username">> => <<"username">>
     },
-    JWT = generate_jws(Payload),
+    test_topic_rules(JWT).
+
+t_topic_rules_v2(_Config) ->
```
savonarola, line 104: nit: maybe add qos/retain to v2 examples

savonarola (approved):

apps/emqx_auth_jwt/src/emqx_authn_jwt.erl
zmstone: A invalid token will cause any pub/sub action to fail, maybe it's better to close connection early.
It's a new format, so maybe it's not entirely a breaking change since all the v1 clients will not notice any change.

PR #12196, opened 2023-12-18
EMQX-10891 route cleanup optimization

SergeTupchiy:
Fixes EMQX-10891 (partially for now)

Release version: v/e5.?

## PR Checklist


apps/emqx/src/emqx_router.erl
```diff
@@ -361,17 +363,12 @@ has_route_tab_entry(Topic, Dest) ->
...
-    mria:transaction(?ROUTE_SHARD, fun() ->
-        [
-            mnesia:delete_object(?ROUTE_TAB, Route, write)
-         || Pat <- Patterns,
-            Route <- mnesia:match_object(?ROUTE_TAB, Pat, write)
-        ]
-    end).
+    lists:foreach(
+        fun(Pattern) ->
+            {atomic, ok} = mria:match_delete(?ROUTE_TAB, #route{_ = '_', dest = Pattern})
```
SergeTupchiy, line 368: Note: it's two separate transactions now (`mnesia:match_delete/2` is a transaction that can't be nested).

apps/emqx/src/emqx_router.erl
```diff
@@ -440,38 +437,20 @@ has_route_v2(Topic, Dest) ->
...
-    ),
-    ok = ets:foldl(
-        fun(#route{dest = Dest} = Route, ok) ->
-            case get_dest_node(Dest) of
-                Node ->
-                    mria:dirty_delete_object(?ROUTE_TAB, Route);
-                _ ->
-                    ok
-            end
+    lists:foreach(
```
SergeTupchiy, line 440: Note: both v1 and v2 cleanups are backward compatible at Mnesia level: older nodes that don't have Mnesia patch will commit the transaction correctly (thanks to the fact that Mnesia internally was ready for match_delete implementation). 
Older Mria releases would fail on both core and replicant nodes (even though Mria ptotcol version is bumped and new replicant nodes will only connect to new cores), However, the failure will lead to shard restart and re-bootstrapping which  will eventually bring all nodes to the consistent state.

apps/emqx/src/emqx_router.erl
```diff
@@ -435,8 +451,25 @@ has_route_v2(Topic, Dest) ->
     end.
 
 cleanup_routes_v2(Node) ->
-    % NOTE
-    % No point in transaction here because all the operations on filters table are dirty.
+    ?with_fallback(
+        lists:foreach(
```
SergeTupchiy, line 455: Note: `lists:foreach/2` is probably too silent.. :thinking:  Though we eventually ignore return value of `cleanup_routes/1` in `emqx_router_helper`, it may have sense to return `ok`/`error` instead of `ok`.

SergeTupchiy:
There are two more potential smaller optimizations:
- currently `emqx_router_helper` attempts to cleanup routes twice (because it receives two nodedown messages, at least when a replicant is down), more details about it in another PR thread: https://github.com/emqx/emqx/pull/12237#discussion_r1445024824
- every core node attempts to cleanup routes (strictly one after another, thanks to the global lock, but each node tries to acquire the lock indefinitely, since `retries` defaults to `infinity`): https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_router_helper.erl#L154
   Each call to `mnesia:match_delete/2` is a transaction with full table lock and it will cause full table scan on every node each time, so probably it makes sense to minimize its usage. 
   It's possible to set the number of `global` lock retries, e. .g. if we set it to 0, only one node that is the first to win the lock will do the cleanup. 
  This is probably less fault tolerant if the lock winner somehow fails to cleanup routes, but we can either retry cleanup on fail, or set retries slightly higher than 0, e.g. to  `number of core nodes / 2` (however, in this case each can still win the lock and repeat cleanup if faster than sleep time between retries).

ieQu1 (dismissed):

ieQu1 (dismissed):

thalesmg:
>  It's possible to set the number of global lock retries, e. .g. if we set it to 0, only one node that is the first to win the lock will do the cleanup.

It may lead to _no nodes_ running the cleanup, if they happen to start the transaction concurrently.

For the DS GC worker, [we set retries to 1](https://github.com/emqx/emqx/blob/cc00dd80ee5f5e0aeb24788c70f15c4b43851717/apps/emqx/src/emqx_persistent_session_ds_gc_worker.erl#L87-L93) to account for that race condition, but I'm not sure it'll always make one node win, in particular if the cluster is large enough.  :thinking:

Rory-Z (approved):

SergeTupchiy:
> > It's possible to set the number of global lock retries, e. .g. if we set it to 0, only one node that is the first to win the lock will do the cleanup.
> 
> It may lead to _no nodes_ running the cleanup, if they happen to start the transaction concurrently.
> 
> For the DS GC worker, [we set retries to 1](https://github.com/emqx/emqx/blob/cc00dd80ee5f5e0aeb24788c70f15c4b43851717/apps/emqx/src/emqx_persistent_session_ds_gc_worker.erl#L87-L93) to account for that race condition, but I'm not sure it'll always make one node win, in particular if the cluster is large enough. 🤔
> 
> We'd probably need leader election with consensus to address this more properly, I think.

Thaks @thalesmg  :raised_hands: Indeed, 0 retries is not an option. :see_no_evil: 1 retry definitely doesn't mean there is always only one winner, but it can probably have an affect. The first sleep retry time used by global is random value up to 250 ms and this cleanup transaction can take longer, at least in a highly loaded cluster.
Anyway, it's probably not absolutely worth changing it. :thinking:

SergeTupchiy:
I'm going to merge and try to address the discussed concerns in a separate PR. 
The fact that nodes eagerly cleanup routes of the down node also increases the risk of some late cleanup of new routes occurring after the down node is back and adds these new routes faster than some of the other node does a subsequent cleanup..

ieQu1 (approved):

PR #12210, opened 2023-12-20
feat(schema): provide type-level documentation snippets

keynslug:
For stuff like `duration()`, `bytesize()` and `secret()`, etc.

The latter will look like this in schema dumps:
```
  ...
  {
    "typedoc" : "A string holding some sensitive information, such as a password. When secret starts with <code>file://</code>, the rest of the string is interpreted as a path to a file containing the secret itself: whole content of the file except any trailing whitespace characters is considered a secret value.",
    "type" : "Secret",
    "text" : "password",
    "hash" : "V-authorization-S-authorization-sources-S-ldap-password",
    "doc" : "The password associated with the bridge, used for authentication with the external database."
  },
  ...
```

Fixes [EMQX-11579](https://emqx.atlassian.net/browse/EMQX-11579).

## PR Checklist


scripts/schema-dump-reformat.escript
```diff
@@ -18,7 +18,8 @@ main(_) ->
     halt(1).
 
 reformat(Json) ->
-    emqx_conf:reformat_schema_dump(fix(Json)).
```
zmstone, line 21: this script can be deleted.

apps/emqx_conf/src/emqx_conf_schema_types.erl
```diff
@@ -165,37 +167,37 @@ readable("duration()") ->
...
         dashboard => #{type => duration},
-        docgen => #{type => "String", example => <<"12m">>}
+        docgen => #{type => "Duration", example => <<"12m">>, desc => ?DESC(duration)}
     };
 readable("timeout_duration_s()") ->
     #{
         swagger => #{type => string, example => <<"1h">>},
         dashboard => #{type => duration},
-        docgen => #{type => "String", example => <<"1h">>}
+        docgen => #{type => "Duration(s)", example => <<"1h">>, desc => ?DESC(duration)}
```
zmstone, line 194: the `_s` or `_ms` types are only different internally.
Should be the same user-facing type. i.e. users should not need to know if this is converted to seconds or to milliseconds internally.

rel/i18n/emqx_conf_schema_types.hocon
```diff
@@ -0,0 +1,12 @@
+emqx_conf_schema_types {
+
+    duration.desc:
+    """A string that represents a time duration, for example: <code>10s</code>, <code>2.5m</code>, <code>1h30m</code>, <code>1W2D</code>, or <code>2345ms</code>, which is the smallest unit. When precision is specified, finer portions of the duration may be ignored: writing <code>1200ms</code> for <code>Duration(s)</code> is equivalent to writing <code>1s</code>. It doesn't matter if units are in upper or lower case."""
```
zmstone, line 4:
```suggestion
    """A string that represents a time duration, for example: <code>10s</code>, <code>2.5m</code>, <code>1h30m</code>, <code>1W2D</code>, or <code>2345ms</code>, which is the smallest unit. When precision is specified, finer portions of the duration may be ignored: writing <code>1200ms</code> for <code>Duration(s)</code> is equivalent to writing <code>1s</code>. The unit part is case-insensitive."""
```

rel/i18n/emqx_conf_schema_types.hocon
```diff
@@ -0,0 +1,12 @@
...
+
+    bytesize.desc:
+    """A string that represents a number of bytes, for example: <code>10B</code>, <code>640kb</code>, <code>4MB</code>, <code>1GB</code>. Units are interpreted as powers of 1024, and the unit part is case-insensitive."""
+
+    secret.desc:
+    """A string holding some sensitive information, such as a password. When secret starts with <code>file://</code>, the rest of the string is interpreted as a path to a file containing the secret itself: whole content of the file except any trailing whitespace characters is considered a secret value."""
```
zmstone, line 10:
```suggestion
    """A string holding some sensitive information, such as a password. When secret starts with <code>file://</code>, the rest of the string is interpreted as a path to a file containing the secret itself: whole content of the file except any trailing whitespace characters is considered a secret value. Note: when clustered, all EMQX nodes should have the same file present before using such config values."""
```

apps/emqx_conf/src/emqx_conf.erl
```diff
@@ -314,18 +316,17 @@ gen_flat_doc(RootNames, #{full_name := FullName, fields := Fields} = S) ->
...
-    [format_field(Field) | format_fields(Fields)].
+format_fields(Fields, DescResolver) ->
+    [format_field(F, DescResolver) || F <- Fields].
 
-format_field(#{name := Name, aliases := Aliases, type := Type} = F) ->
+format_field(#{name := Name, aliases := Aliases, type := Type} = F, DescResolver) ->
     L = [
         {text, Name},
         {type, format_type(Type)},
+        {typedoc, format_type_desc(Type, DescResolver)},
```
zmstone, line 329: For this version, maybe we can append `dos` after `typedoc`  (insert a `<br/>` in between)

apps/emqx_conf/src/emqx_conf_schema_types.erl
keynslug: There's a small issue with that: users are allowed to write, for instance, `900ms` for the `duration_s()` value, but the result may be very unexpected for them if they don't know the precision. OTOH no similar issue with `duration_ms()`, it's just `Duration` in docs.

apps/emqx_conf/src/emqx_conf.erl
zmstone: i mean, don't introduce a new field.

keynslug: I was under impression that such manipulations should be under the docsite team responsibility.

apps/emqx_conf/src/emqx_conf_schema_types.erl
zmstone: good point

zmstone (approved):

PR #12219, opened 2023-12-22
fix: FT deobfuscate secret key

zhongwencool:
Fixes <issue-or-jira-number>

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_ft/test/emqx_ft_api_SUITE.erl
+++ b/apps/emqx_ft/test/emqx_ft_api_SUITE.erl
@@ -22,6 +22,8 @@
 -include_lib("common_test/include/ct.hrl").
 -include_lib("stdlib/include/assert.hrl").
 
+-define(SECRET_ACCESS_KEY, <<"fake_secret_access_key">>).
+
 -import(emqx_dashboard_api_test_helpers, [host/0, uri/1]).
 
 all() -> emqx_common_test_helpers:all(?MODULE).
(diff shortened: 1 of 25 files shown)


apps/emqx_ft/test/emqx_ft_api_SUITE.erl
```diff
@@ -405,23 +411,47 @@ t_configure(Config) ->
...
+                            }
                         }
                     }
                 }
-            },
+            ),
             Config
         )
     ),
+    %% put secret as ******, check the secret is unchanged
```
keynslug, line 434: This strikes me as an odd design choice: what if user's secret is indeed `******`? It will just be silently discarded. Obviously, it's not possible to have an AWS secret that looks like `******` but it's sort of a general observation.

Design-wise, what if we respond according to slightly different schema for this field, so a response won't be ready for a blind reuse? I.e. `..., "secret_access_key": {"secret":"******"}, ...`. Though I don't know the use case well enough, so take my words with a grain of salt.

keynslug: Right. But after this change it won't be possible to change secret to `******` if I'm reading this right?

thalesmg: Maybe if the API obfuscation were to return `null` / `undefined`, then it would be a clearer signal that the intention is truly to set the password to `******`.  :thinking:

PR #12227, opened 2023-12-22
fix route race condition possible in core/replicant clusters

SergeTupchiy:
The problem was that emqx_router:has_route/2 check may observe a stale route, which deletion is not yet replicated from the core node to the local replicant node.

For example:
1. The only one subscriber per a given topic A unsubscribes from a replicant node.
2. The route to A is deleted by the emqx_broker.
3. Mria makes RPC to a core node, it succeeds and returns.
4. The client resubscribes or another client subscribes to the same topic A.
5. emqx_broker tries to add a route again: emqx_router:do_add_route(Topic).
6. emqx_router checks if the route is present.
7. The stale route is present because deletion not replicated yet, so no route is being added.
8. Route deletion is replicated locally but it's too late: we already have a local subscriber without a route.

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_broker.erl
+++ b/apps/emqx/src/emqx_broker.erl
@@ -160,15 +160,20 @@ do_subscribe(Topic, SubPid, SubOpts) ->
 do_subscribe2(Topic, SubPid, SubOpts) when is_binary(Topic) ->
     %% FIXME: subscribe shard bug
     %% https://emqx.atlassian.net/browse/EMQX-10214
-    case emqx_broker_helper:get_sub_shard(SubPid, Topic) of
+    {Shard, Seq} = emqx_broker_helper:get_sub_shard_and_seq(SubPid, Topic),
+    IsFirst = Seq =:= 1,
+    case Shard of
         0 ->
             true = ets:insert(?SUBSCRIBER, {Topic, SubPid}),
             true = ets:insert(?SUBOPTION, {{Topic, SubPid}, SubOpts}),
-            call(pick(Topic), {subscribe, Topic});
+            case IsFirst of
+                true -> call(pick(Topic), {subscribe, Topic});
+                false -> ok
+            end;
         I ->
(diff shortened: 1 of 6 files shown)


apps/emqx/src/emqx_broker.erl
```diff
@@ -160,15 +160,20 @@ do_subscribe(Topic, SubPid, SubOpts) ->
...
-    case emqx_broker_helper:get_sub_shard(SubPid, Topic) of
+    {Shard, Seq} = emqx_broker_helper:get_sub_shard_and_seq(SubPid, Topic),
+    IsFirst = Seq =:= 1,
+    case Shard of
         0 ->
             true = ets:insert(?SUBSCRIBER, {Topic, SubPid}),
             true = ets:insert(?SUBOPTION, {{Topic, SubPid}, SubOpts}),
-            call(pick(Topic), {subscribe, Topic});
+            case IsFirst of
+                true -> call(pick(Topic), {subscribe, Topic});
```
SergeTupchiy, line 170: This is probably too optimistic: if the first subscriber failed to register the route, the next one won't even try it.. :face_with_head_bandage: 
Need a more robust solution.

SergeTupchiy: Maybe should keep `emqx_router:has_route/2` check but skip it for the first subscriber, so that the first one always adds a route (even if the previous stale route is still present). 
Or just serialize writes to `?SUBSCRIBER` tab by means of `emqx_broker` pool...

SergeTupchiy: Another possible solution is to implement new Mria API that will synchronously update local replicant state and avoid core->replicant replication for the node that initiated a write op.

keynslug: > Or just serialize writes to `?SUBSCRIBER` tab by means of emqx_broker pool...

I'm also starting to think about going into this direction.

SergeTupchiy:
Closed in favor of https://github.com/emqx/emqx/pull/12243

PR #12228, opened 2023-12-23
feat(helm): only keep last digits for ipv6 ports

clementnuss:
permits specifying ports such as :::18083, a somewhat strange but needed
syntax when trying to make emqx listen on IPv6 addresses
The solution here is to split by colon `:` and to only keep the last
entry, which will corespond to the actual number for the port.

can be tested by using following values.yaml config:
```yaml
## EMQX configuration item, see the documentation (https://hub.docker.com/r/emqx/emqx)
emqxConfig:
  EMQX_CLUSTER__DISCOVERY_STRATEGY: "dns"
  EMQX_DASHBOARD__DEFAULT_USERNAME: "admin"
  EMQX_DASHBOARD__DEFAULT_PASSWORD: "public"
  EMQX_LISTENERS__TCP__DEFAULT__BIND: ":::1234"
  EMQX_LISTENERS__TCP__INTERNAL__BIND: ":::3456"
  EMQX_DASHBOARD__LISTENERS__HTTP__BIND: ":::31415"
  EMQX_DASHBOARD__LISTENERS__HTTPS__BIND: ":::65432

```

### DIFF
--- a/deploy/charts/emqx/templates/StatefulSet.yaml
+++ b/deploy/charts/emqx/templates/StatefulSet.yaml
@@ -102,22 +102,22 @@ spec:
           {{- end }}
           ports:
           - name: mqtt
-            containerPort: {{ .Values.emqxConfig.EMQX_LISTENERS__TCP__DEFAULT__BIND | default 1883 }}
+            containerPort: {{ splitList ":" ( .Values.emqxConfig.EMQX_LISTENERS__TCP__DEFAULT__BIND | default "1883" ) | last }}
           - name: mqttssl
-            containerPort: {{ .Values.emqxConfig.EMQX_LISTENERS__SSL__DEFAULT__BIND | default 8883 }}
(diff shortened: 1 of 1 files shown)


Rory-Z:
I think the value of the `EMQX_LISTENERS__TCP__DEFAULT__BIND ` need a string, not list, so maybe it's will not work.
@id Am I right ?

clementnuss:
you are right that it is a string, but what `splitList` actually does is: split the input string, return the splitted parts as a list.

```
{{ splitList ":" ( .Values.emqxConfig.EMQX_LISTENERS__TCP__DEFAULT__BIND | default "1883" ) | last }}
``` 

so I split by column `:` and take the last entry, which will the last port number.
In the config for emqx however, the port is specified correctly with the (strange) triple-column notation, and IPv6 addresses work.

image

Rory-Z (approved):

Rory-Z:
Great, LGTM

clementnuss:
just did a rebase from master 🙃

PR #12237, opened 2023-12-27
feat(routing): add route sync process pool

keynslug:
Fixes [EMQX-11483](https://emqx.atlassian.net/browse/EMQX-11483).

## Summary

Dedicated to synchronizing local state updates with the global view of the routing state. This should help with:
* Free `emqx_broker` pool from doing global routing updates.
* Minimize roundtrips over high-latency links to core nodes.
* ~~Make it much harder to reach inconsistent routing state~~ Split into #12243.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_broker.erl
+++ b/apps/emqx/src/emqx_broker.erl
@@ -604,21 +604,35 @@ do_dispatch({shard, I}, Topic, Msg) ->
 %%
 
 maybe_add_route(_Existed = false, Topic, ReplyTo) ->
-    add_route(emqx_config:get([broker, routing, batch_sync, enable]), Topic, ReplyTo);
+    sync_route(add, Topic, #{reply => ReplyTo});
 maybe_add_route(_Existed = true, _Topic, _ReplyTo) ->
     ok.
 
-add_route(_BatchSync = true, Topic, ReplyTo) ->
-    emqx_router_syncer:push(add, Topic, node(), #{reply => ReplyTo});
-add_route(_BatchSync = false, Topic, _ReplyTo) ->
-    emqx_router:do_add_route(Topic, node()).
-
 maybe_delete_route(_Exists = false, Topic) ->
-    delete_route(emqx_config:get([broker, routing, batch_sync, enable]), Topic);
+    sync_route(delete, Topic, #{});
 maybe_delete_route(_Exists = true, _Topic) ->
     ok.
 
-delete_route(_BatchSync = true, Topic) ->
-    emqx_router_syncer:push(delete, Topic, node(), #{});
-delete_route(_BatchSync = false, Topic) ->
+sync_route(Action, Topic, ReplyTo) ->
+    EnabledOn = emqx_config:get([broker, routing, batch_sync, enable_on]),
+    case EnabledOn of
+        all ->
+            push_sync_route(Action, Topic, ReplyTo);
+        none ->
+            regular_sync_route(Action, Topic);
+        Role ->
+            case mria_config:whoami() of
+                Role ->
+                    push_sync_route(Action, Topic, ReplyTo);
+                _Disabled ->
+                    regular_sync_route(Action, Topic)
+            end
+    end.
+
+push_sync_route(Action, Topic, Opts) ->
+    emqx_router_syncer:push(Action, Topic, node(), Opts).
+
+regular_sync_route(add, Topic) ->
+    emqx_router:do_add_route(Topic, node());
+regular_sync_route(delete, Topic) ->
     emqx_router:do_delete_route(Topic, node()).
--- a/apps/emqx/src/emqx_broker.erl
+++ b/apps/emqx/src/emqx_broker.erl
@@ -621,10 +621,10 @@ sync_route(Action, Topic, ReplyTo) ->
         none ->
             regular_sync_route(Action, Topic);
         Role ->
-            case mria_config:whoami() of
-                Role ->
+            case Role =:= mria_config:whoami() of
+                true ->
                     push_sync_route(Action, Topic, ReplyTo);
-                _Disabled ->
+                false ->
                     regular_sync_route(Action, Topic)
             end
     end.
--- a/apps/emqx/src/emqx_broker.erl
+++ b/apps/emqx/src/emqx_broker.erl
@@ -167,7 +167,13 @@ do_subscribe(Topic, SubPid, SubOpts) when is_binary(Topic) ->
     %% `unsubscribe` codepath. So we have to pick a worker according to the topic,
     %% but not shard. If there are topics with high number of shards, then the
     %% load across the pool will be unbalanced.
-    call(pick(Topic), {subscribe, Topic, SubPid, I});
+    Sync = call(pick(Topic), {subscribe, Topic, SubPid, I}),
+    case Sync of
+        ok ->
+            ok;
+        Ref when is_reference(Ref) ->
+            emqx_router_syncer:wait(Ref)
+    end;
 do_subscribe(Topic = #share{group = Group, topic = RealTopic}, SubPid, SubOpts) when
     is_binary(RealTopic)
 ->
@@ -491,8 +497,8 @@ safe_update_stats(Tab, Stat, MaxStat) ->
 call(Broker, Req) ->
     gen_server:call(Broker, Req, infinity).
 
-cast(Broker, Msg) ->
-    gen_server:cast(Broker, Msg).
+cast(Broker, Req) ->
+    gen_server:cast(Broker, Req).
 
 %% Pick a broker
 pick(Topic) ->
@@ -506,18 +512,18 @@ init([Pool, Id]) ->
     true = gproc_pool:connect_worker(Pool, {Pool, Id}),
     {ok, #{pool => Pool, id => Id}}.
 
-handle_call({subscribe, Topic, SubPid, 0}, _From, State) ->
+handle_call({subscribe, Topic, SubPid, 0}, {From, _Tag}, State) ->
     Existed = ets:member(?SUBSCRIBER, Topic),
     true = ets:insert(?SUBSCRIBER, {Topic, SubPid}),
-    Result = maybe_add_route(Existed, Topic),
+    Result = maybe_add_route(Existed, Topic, From),
     {reply, Result, State};
-handle_call({subscribe, Topic, SubPid, I}, _From, State) ->
+handle_call({subscribe, Topic, SubPid, I}, {From, _Tag}, State) ->
     Existed = ets:member(?SUBSCRIBER, Topic),
     true = ets:insert(?SUBSCRIBER, [
         {Topic, {shard, I}},
         {{shard, Topic, I}, SubPid}
     ]),
-    Result = maybe_add_route(Existed, Topic),
+    Result = maybe_add_route(Existed, Topic, From),
     {reply, Result, State};
 handle_call(Req, _From, State) ->
     ?SLOG(error, #{msg => "unexpected_call", call => Req}),
@@ -597,12 +603,12 @@ do_dispatch({shard, I}, Topic, Msg) ->
 
 %%
 
-maybe_add_route(_Existed = false, Topic) ->
-    emqx_router:do_add_route(Topic);
-maybe_add_route(_Existed = true, _Topic) ->
+maybe_add_route(_Existed = false, Topic, ReplyTo) ->
+    emqx_router_syncer:push(add, Topic, node(), #{reply => ReplyTo});
+maybe_add_route(_Existed = true, _Topic, _ReplyTo) ->
     ok.
 
 maybe_delete_route(_Exists = false, Topic) ->
-    emqx_router:do_delete_route(Topic);
+    emqx_router_syncer:push(delete, Topic, node(), #{});
 maybe_delete_route(_Exists = true, _Topic) ->
     ok.
--- a/apps/emqx/src/emqx_broker.erl
+++ b/apps/emqx/src/emqx_broker.erl
@@ -604,11 +604,21 @@ do_dispatch({shard, I}, Topic, Msg) ->
 %%
 
 maybe_add_route(_Existed = false, Topic, ReplyTo) ->
-    emqx_router_syncer:push(add, Topic, node(), #{reply => ReplyTo});
+    add_route(emqx_config:get([broker, routing, batch_sync, enable]), Topic, ReplyTo);
 maybe_add_route(_Existed = true, _Topic, _ReplyTo) ->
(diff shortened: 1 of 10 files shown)


apps/emqx/src/emqx_broker_sup.erl
```diff
@@ -59,4 +66,4 @@ init([]) ->
         modules => [emqx_broker_helper]
     },
 
-    {ok, {{one_for_all, 0, 1}, [BrokerPool, SharedSub, Helper]}}.
+    {ok, {{one_for_all, 0, 1}, [BrokerPool, SyncerPool, SharedSub, Helper]}}.
```
SergeTupchiy, line 69: As `BrokerPool` depends on `SyncerPool`, would it make sense to start `SyncerPool` before 'BrokerPool` ?  
It should not create any problems actually, since listeners are started only after all theses pools, so this is more like extra safety suggestion..

keynslug: Oh right, indeed.

apps/emqx/src/emqx_router.erl
```diff
@@ -213,12 +222,70 @@ do_delete_route(Topic) when is_binary(Topic) ->
...
+mria_delete_route(v1, Topic, Dest, Ctx) ->
+    mria_delete_route_v1(Topic, Dest, Ctx).
+
+-spec do_batch(Batch) -> Errors when
+    %% Operation :: {add, ...} | {delete, ...}.
+    Batch :: #{Route => _Operation :: tuple()},
+    Errors :: #{Route => _Error},
+    Route :: {emqx_types:topic(), dest()}.
+do_batch(Batch) ->
+    Nodes = batch_get_dest_nodes(Batch),
```
SergeTupchiy, line 238: Side question: I wonder if we can avoid calling `emqx_router_helper:monitor/1` on every `add_route`. 
It seems to be redundant, as emqx_router_helper already receives `{membership,{node,down, Node}}` because it calls `ekka:monitor(membership)`. Alternatively, router_helper can also add local node to `emqx_routing_node` tab once in `init/1` cb.
Performance-wise, it would probably make little sense but the code can be slightly simplified.

SergeTupchiy: Yes, it's expected to receive it from `mria_membership` process: https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_router_helper.erl#L104

Mria than notifies all monitor callers if node is down:
1. `mria_node_monitor` tracks all connected nodes: https://github.com/emqx/mria/blob/main/src/mria_node_monitor.erl#L78
2. `mria_node_monitor` notifies `mria_membership` when some connected node (either core or replicant) is down: 
https://github.com/emqx/mria/blob/main/src/mria_node_monitor.erl#L133
2. `mria membership` notifies subscribed processes (those that called `ekka/mria:monitor`): https://github.com/emqx/mria/blob/main/src/mria_membership.erl#L509

It looks like emqx_router_helper:monitor/1 is used as a safety measure if underlying ekka/mria node monitor doesn't work (or probably it behaved differently in the past). 
I think we can either stop using `?ROUTING_NODE` and rely solely on ekka/mria in emqx_router_helper, or just write local node to `?ROUTING_NODE` tab once in `emqx_router_helper:init/1`. 
Probably it's worth a separate task/PR, though. :thinking:

SergeTupchiy: @zmstone, @ieQu1 may I ask you to have a look at this thread?

keynslug: Thanks for clarifying! My impression was that this thing with `?ROUTING_NODE` table was introduced as a kind of optimization to avoid full table scans for nodes that do not serve any clients whatsoever. And now it looks like this no longer works as intended for quite some time.

SergeTupchiy: Yes, it probably was the intention indeed. However, it actually tries to cleanup routes of the down node twice (though during the second call there is nothing more to clean and it much faster):
An example of node down (testing it now in scope of down node routes cleanup optimization):
```
call emqx_router_helper:handle_info({nodedown,'emqx1@127.0.0.1'},#{nodes => ['emqx1@127.0.0.1']}) 
Routes cleaned up in: 6977ms
(<0.2635.0>) call emqx_router_helper:handle_info({membership,{node,down,'emqx1@127.0.0.1'}},#{nodes => []}) 
(<0.2635.0>) call emqx_router_helper:handle_info({nodedown,'emqx1@127.0.0.1'},#{nodes => []}) ({gen_server,
Routes cleaned up in: 411ms
```

apps/emqx/src/emqx_schema.erl
```diff
@@ -1399,6 +1399,22 @@ fields("broker_routing") ->
                     'readOnly' => true,
                     desc => ?DESC(broker_routing_storage_schema)
                 }
+            )},
+        {"batch_sync",
```
zmstone, line 1403: maybe `batch_sync_from_replicant` and set default to `true`.

apps/emqx/src/emqx_router_syncer.erl
```diff
@@ -0,0 +1,328 @@
...
+
+handle_call(_Call, _From, State) ->
+    {reply, ignored, State}.
+
+handle_cast(_Msg, State) ->
+    {noreply, State}.
+
+handle_info(Push = ?PUSH(_, _), State) ->
+    %% NOTE: Wait a bit to collect potentially overlapping operations.
+    ok = timer:sleep(?MIN_SYNC_INTERVAL),
```
zmstone, line 125: use https://www.erlang.org/doc/man/erlang#yield-0 in prod ?

apps/emqx/src/emqx_router_syncer.erl
```diff
@@ -0,0 +1,328 @@
...
+                    replyctx_send(maps:get(Route, Errors, ok), Ctx)
+            end
+        end,
+        Batch
+    ).
+
+replyctx_send(_Result, []) ->
+    noreply;
+replyctx_send(Result, {MRef, Pid}) ->
+    Pid ! {MRef, Result}.
```
zmstone, line 212: Nit: prefer ignoring the result in the earliest place, but not delegate to callers.
```suggestion
replyctx_send(_Result, []) ->
    ok;
replyctx_send(Result, {MRef, Pid}) ->
    _ = erlang:send(Pid, {MRef, Result}),
    ok.
```

apps/emqx/src/emqx_router_syncer.erl
```diff
@@ -0,0 +1,328 @@
...
+        n_add => maps:size(maps:filter(fun(_, ?ROUTEOP(A)) -> A == add end, Batch)),
+        n_delete => maps:size(maps:filter(fun(_, ?ROUTEOP(A)) -> A == delete end, Batch)),
+        prio_highest => maps:fold(fun(_, ?ROUTEOP(_, P), M) -> min(P, M) end, none, Batch),
+        prio_lowest => maps:fold(fun(_, ?ROUTEOP(_, P), M) -> max(P, M) end, 0, Batch)
+    }),
+    %% TODO: retry if error?
+    Errors = run_batch(Batch),
+    ok = send_replies(Errors, Batch),
+    NState = State#{stash := Stash3},
+    %% TODO: postpone if only ?PRIO_BG operations left?
```
zmstone, line 150: probably negligible, and hard to test.
```suggestion
```

apps/emqx/src/emqx_router_syncer.erl
```diff
@@ -0,0 +1,328 @@
...
+
+run_batch(Batch) ->
+    emqx_router:do_batch(Batch).
+
+%%
+
+stash_new() ->
+    #{}.
+
+stash_empty(Stash) ->
```
zmstone, line 224:
```suggestion
is_stash_empty(Stash) ->
```

apps/emqx/src/emqx_router_syncer.erl
```diff
@@ -0,0 +1,328 @@
...
+            case merge_route_op(RouteOp, ?ROUTEOP(Action, Prio, Ctx)) of
+                undefined ->
+                    maps:remove(Route, Stash);
+                RouteOpMerged ->
+                    Stash#{Route := RouteOpMerged}
+            end
+    end.
+
+merge_route_op(?ROUTEOP(Action, _Prio1, Ctx1), DestOp = ?ROUTEOP(Action)) ->
+    %% NOTE: This should not happen anyway.
```
zmstone, line 257: Q: Why this doesn't happen ?

apps/emqx/src/emqx_router_syncer.erl
```diff
@@ -0,0 +1,328 @@
...
+    _ = replyctx_send(ignored, Ctx1),
+    DestOp;
+merge_route_op(?ROUTEOP(_Action1, _Prio1, Ctx1), ?ROUTEOP(_Action2, _Prio2, Ctx2)) ->
+    %% NOTE: Operations cancel each other.
```
zmstone, line 261: this assumes the base is always consitent.
to be more idempotent, maybe only cancel the old op and keep the new op?

apps/emqx/src/emqx_router.erl
```diff
@@ -213,12 +222,70 @@ do_delete_route(Topic) when is_binary(Topic) ->
...
+mria_batch(v2, Batch) ->
+    mria_batch_v2(Batch);
+mria_batch(v1, Batch) ->
+    mria_batch_v1(Batch).
+
+mria_batch_v2(Batch) ->
+    mria:async_dirty(?ROUTE_SHARD, fun mria_batch_run/2, [v2, Batch]).
+
+mria_batch_v1(Batch) ->
+    {atomic, Res} = mria:transaction(?ROUTE_SHARD, fun mria_batch_run/2, [v1, Batch]),
```
zmstone, line 251: use remote calls, 
and this won't work during rolling upgrade ?

apps/emqx/src/emqx_router_syncer.erl
```diff
@@ -0,0 +1,328 @@
...
+
+-spec push(action(), emqx_types:topic(), emqx_router:dest(), Opts) ->
+    ok | _WaitRef :: reference()
+when
+    Opts :: #{reply => pid()}.
+push(Action, Topic, Dest, Opts) ->
+    Worker = gproc_pool:pick_worker(?POOL, Topic),
+    Prio = designate_prio(Action, Opts),
+    Context = mk_push_context(Opts),
+    Worker ! ?PUSH(Prio, {Action, Topic, Dest, Context}),
```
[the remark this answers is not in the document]
keynslug: Hm, I don't understand why it should have. Unmatched return?

apps/emqx/src/emqx_router.erl
zmstone: agree, no need to monitor twice.

apps/emqx/src/emqx_router_syncer.erl
keynslug: Assumption here is that if we trust `emqx_broker` to correctly sync updates over the same topic, then it should only push "add, delete, add, delete, ..." but not "add, add, delete, delete, ...". If it doesn't occasionally, this `ignored` currently will case a badmatch in the channel process.

apps/emqx/src/emqx_router_syncer.erl
keynslug: Idea here is to try to do nothing if we have flapping clients with moderately unique subscription patterns. It's indeed somewhat too specific, but should be relatively easy to test.

apps/emqx/src/emqx_router_syncer.erl
```diff
@@ -0,0 +1,328 @@
...
+        {MRef, _} ->
+            MRef;
+        [] ->
+            ok
+    end.
+
+-spec wait(_WaitRef :: reference()) ->
+    ok | {error, _Reason}.
+wait(MRef) ->
+    %% FIXME: timeouts
```
zmstone, line 92: We cannot do timeouts because the client will reconnect and add more load otherwise.
What we can do is the check is to inject some check callback functions here and run a check after some timeout,
keep waiting if check returns ok, otherwise give up waiting and return error.

The only interrupting event should be socket close or error events which are already delivered to the mailbox.
However doing receive here will make it hard to handle such events by emqx_channel etc,
so maybe a good option is to check if the socket is closed by calling `inet:getopts/1`.

zmstone: the same applies to emqx_broker:handle_call, there is no need to handle the call if SUBACK is not possible to be sent.

apps/emqx/src/emqx_router_syncer.erl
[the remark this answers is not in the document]
zmstone: yes.

apps/emqx/src/emqx_schema.erl
zmstone: cannot default to `true` because of rolling upgrade.

apps/emqx/src/emqx_router_syncer.erl
keynslug: Good idea.

> this assumes the base is always consitent.

Worth noting that it's not really "always", e.g. add + delete in _different_ batches won't be able to cancel each other.

apps/emqx/src/emqx_router.erl
SergeTupchiy: thanks @zmstone , I'll address it in a separate PR, agreed with @keynslug .

apps/emqx/src/emqx_router_syncer.erl
keynslug: Since it carries a big red warning callout in the doc, I changed the interval to `0` instead for the time being. Strangely enough, I was always pretty sure that it's equivalent to scheduling out the current process but cannot find any evidence in the documentation confirming that.

apps/emqx/src/emqx_broker.erl
```diff
@@ -597,12 +603,36 @@ do_dispatch({shard, I}, Topic, Msg) ->
...
+sync_route(Action, Topic, ReplyTo) ->
+    EnabledOn = emqx_config:get([broker, routing, batch_sync, enable_on]),
+    case EnabledOn of
+        all ->
+            push_sync_route(Action, Topic, ReplyTo);
+        none ->
+            regular_sync_route(Action, Topic);
+        Role ->
+            case mria_config:whoami() of
+                Role ->
```
zmstone, line 625: nit:
```suggestion
            case Role =:= mria_config:whoami() of
                true ->
```

zmstone (dismissed):

thalesmg (approved):

savonarola (approved):

PR #12243, opened 2023-12-28
fix(router): rely on local sync state updates to propagate routes

keynslug:
## Summary

Instead of (potentially lagging) global table state itself. Local state should be the only source of truth anyway. Moreover,
synchronize parts of local state update logic through the broker pool to ensure consistency.

The problem was that `emqx_router:has_route/2` check may observe a stale route, which deletion is not yet replicated from the core node to the local replicant node.

For example:
1. The only one subscriber per a given topic A unsubscribes from a replicant node.
2. The route to A is deleted by the `emqx_broker`.
3. Mria makes RPC to a core node, it succeeds and returns.
4. The client resubscribes or another client subscribes to the same topic A.
5. emqx_broker tries to add a route again: `emqx_router:do_add_route(Topic)`.
6. `emqx_router` checks if the route is present.
7. The stale route is present because deletion not replicated yet, so no route is being added.
8. Route deletion is replicated locally but it's too late: we already have a local subscriber without a route.

Thanks to @SergeTupchiy for spotting this and coming up with a testcase.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_broker.erl
+++ b/apps/emqx/src/emqx_broker.erl
@@ -157,21 +157,19 @@ with_subid(SubId, SubOpts) ->
 do_subscribe(Topic, SubPid, SubOpts) when is_binary(Topic) ->
     %% FIXME: subscribe shard bug
     %% https://emqx.atlassian.net/browse/EMQX-10214
-    case emqx_broker_helper:get_sub_shard(SubPid, Topic) of
-        0 ->
-            true = ets:insert(?SUBSCRIBER, {Topic, SubPid}),
-            true = ets:insert(?SUBOPTION, {{Topic, SubPid}, SubOpts}),
-            call(pick(Topic), {subscribe, Topic});
-        I ->
-            true = ets:insert(?SUBSCRIBER, {{shard, Topic, I}, SubPid}),
-            true = ets:insert(?SUBOPTION, {{Topic, SubPid}, maps:put(shard, I, SubOpts)}),
-            call(pick({Topic, I}), {subscribe, Topic, I})
-    end;
+    I = emqx_broker_helper:get_sub_shard(SubPid, Topic),
+    true = ets:insert(?SUBOPTION, {{Topic, SubPid}, with_shard_idx(I, SubOpts)}),
+    call(pick(Topic), {subscribe, Topic, SubPid, I});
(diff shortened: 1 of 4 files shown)


apps/emqx/src/emqx_broker.erl
```diff
@@ -204,15 +199,12 @@ do_unsubscribe2(Topic, SubPid, SubOpts) when
...
-        I ->
-            true = ets:delete_object(?SUBSCRIBER, {{shard, Topic, I}, SubPid}),
-            cast(pick({Topic, I}), {unsubscribed, Topic, I})
-    end;
+    I = maps:get(shard, SubOpts, 0),
+    case I of
+        0 -> emqx_exclusive_subscription:unsubscribe(Topic, SubOpts);
+        _ -> ok
+    end,
+    cast(pick(Topic), {unsubscribed, Topic, SubPid, I});
```
SergeTupchiy, line 207: Just thinking aloud: a side effect of `cast(pick(Topic)` instead of `cast(pick({Topic, I})` is less even workload distribution over broker pool workers. 
Say, topic A has a very large number of subscribers. If all/most of A subscribers go down/unsubscribe, they will be processed by the same broker process, probably making it less responsive for a while.
Consistency-wise it's good and simple, though.. :thinking:

keynslug: Very good point. To be honest I'm struggling to come up with a better alternative that does not involve significant rework of route syncing logic. Maybe I'm missing something? From the correctness/consistency standpoint, my assumptions are:
1. Order of changes of each route is important, and should be preserved in the whole system: we achieve this by synchronizing and sending changes of the same topic from the same process.<sup>Though admittedly, mnesia might compromise it somewhere down the line.</sup>
1. If we are relying on local state to minimize global routing state changes, it's important that some operations on ETS tables on the same topic should not be interleaved (in this case, `ets:member` + `ets:insert` vs `ets:delete`s).

SergeTupchiy: I'm also thinking about alternatives and have not found any better solution either yet..

zmstone: alternative is to one route for each shard.

zmstone: > alternative is to one route for each shard.
not claiming it's a better solution though.

zmstone (approved):

PR #12251, opened 2024-01-03
Refactor and optimize persistent session

ieQu1:
Fixes https://emqx.atlassian.net/browse/EMQX-11556

Release version: v/e5.?

## Summary

Refactor code to make more invariants correct by construction.

- Hide Mnesia tables behind a CRUD module. Introduce dirty session state to avoid frequent mria transactions 
- Factor out inflight buffer to a different module
- Use separate tracks of PacketIds for QoS1 and QoS2 messages
- Avoid storing pubranges, allow only one pubrange per stream
- Better logical separation of concerns

### Fan-out to wildcard subscribers (message size = 0)
```bashs
emqttb --pushgw --loiter 10min \
               @pub --metadata -N 1000 --qos 1 --topic foo/%n --pubinterval 100ms --size 0 \
               @sub --qos 1 --parse-metadata --topic foo/# -N 10 --expiry 30 \
               @a -a pub/pubinterval -V 10000000 --Ti 0.1 --Kp 0.0009 --pvar [scenarios,sub,{},metrics,e2e_latency] --setpoint 1000000 --min 1000 --max 100000000 \
               @g --inflight 1000 &
```

https://snapshots.raintank.io/dashboard/snapshot/ccmEzWR08hLfU9zdxQL9iEqG4gD6eqKC

Storage inefficiency factor:
```bash
tar cf  data.tar data
tar czf data.tar.xz data
echo "$(stat --format %s data.tar) / $(stat --format %s data.tar.xz)" | bc -l

#+RESULTS:
: 7.470356703038198
```
### Fan-out to wildcard subscribers (message size = 1k)

https://snapshots.raintank.io/dashboard/snapshot/fg0qoTi8v3vzvzoTCcw7uGGFdJnI1dGa

## PR Checklist


apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -0,0 +1,579 @@
...
+%% Persistent map.
+%%
+%% Pmap accumulates the updates in a term stored in the heap of a
+%% process, so they can be committed all at once in a single
+%% transaction.
+%%
+%% It should be possible to make frequent changes to the pmap without
+%% stressing Mria.
+%%
+%% It's implemented as three maps: `clean', `dirty' and `tombstones'.
```
ieQu1, line 71: Tombstones were removed.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -0,0 +1,279 @@
...
+        S0
+    ).
+
+%%================================================================================
+%% Internal functions
+%%================================================================================
+
+ensure_iterator(TopicFilter, StartTime, SubId, {{RankX, RankY}, Stream}, S) ->
+    %% TODO: hash collisions
+    Key = {SubId, erlang:phash2(Stream)},
```
ieQu1, line 141: This is the same problem as Thales was facing... I need to come up with an idea how to store the streams without worrying about collisions.

apps/emqx/src/emqx_persistent_session_ds.hrl
```diff
@@ -25,75 +25,44 @@
...
+    rank_y :: emqx_ds:rank_y(),
+    %% Iterator at the beginning and end of the last batch:
+    it_begin :: emqx_ds:iterator() | undefined,
+    it_end :: emqx_ds:iterator() | end_of_stream,
+    %% Key that points at the beginning of the batch:
+    batch_size = 0 :: non_neg_integer(),
+    %% Session sequence number at the time when the batch was fetched:
+    first_seqno_qos1 = 0 :: emqx_persistent_session_ds:seqno(),
+    first_seqno_qos2 = 0 :: emqx_persistent_session_ds:seqno(),
+    %% Number of messages collected in the last batch:
```
ieQu1, line 63: Misleading comment

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -0,0 +1,582 @@
...
+
+kv_pmap_restore(Table, SessionId) ->
+    MS = [{#kv{k = {SessionId, '_'}, _ = '_'}, [], ['$_']}],
+    Objs = mnesia:select(Table, MS, read),
+    [{K, encoder(decode, Table, V)} || #kv{k = {_, K}, v = V} <- Objs].
+
+kv_pmap_delete(Table, SessionId) ->
+    MS = [{#kv{k = {SessionId, '$1'}, _ = '_'}, [], ['$1']}],
+    Keys = mnesia:select(Table, MS, read),
+    [mnesia:delete(Table, {SessionId, K}, write) || K <- Keys],
```
thalesmg, line 539: Could we use `mnesia:match_delete` from our OTP fork?

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -0,0 +1,279 @@
...
+            end
+        end,
+        [],
+        S
+    ),
+    lists:sort(fun compare_streams/2, Streams).
+
+-spec find_new_streams(emqx_persistent_session_ds_state:t()) ->
+    [{emqx_persistent_session_ds_state:stream_key(), emqx_persistent_session_ds:stream_state()}].
+find_new_streams(S) ->
```
thalesmg, line 64: Q: it seems that the only difference between this function and `find_replay_streams` is that this one shuffles the results?

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
ieQu1: RocksDB doesn't like maps in the keys.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -0,0 +1,279 @@
...
+    %% some way to get the replays un-stuck at the cost of potentially
+    %% losing messages during replay (or just kill the stuck channel
+    %% after timeout?)
+    Comm1 = emqx_persistent_session_ds_state:get_seqno(?committed(?QOS_1), S),
+    Comm2 = emqx_persistent_session_ds_state:get_seqno(?committed(?QOS_2), S),
+    shuffle(
+        emqx_persistent_session_ds_state:fold_streams(
+            fun(Key, Stream, Acc) ->
+                case is_fully_acked(Comm1, Comm2, Stream) of
+                    true ->
```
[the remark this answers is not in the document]
ieQu1: This is actually the trick that increases the speed of replay: we limit the number of "pubranges" to only one per stream. This way we avoid expensive garbage collection, since we can simply overwrite the state of the stream replay.

However, it means we cannot fetch messages from the stream that has unacknowledged messages: they will be lost during replay. This is why this function returns only the streams that have been fully acked.

apps/emqx/src/emqx_persistent_session_ds_state.erl
ieQu1: Sure, but I'd prefer to save it for a later PR.

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -0,0 +1,582 @@
...
+        cache :: #{K => V},
+        dirty :: #{K => dirty | del}
+    }.
+
+%% Session metadata:
+-define(created_at, created_at).
+-define(last_alive_at, last_alive_at).
+-define(conninfo, conninfo).
+%% Unique integer used to create unique identities
+-define(last_id, last_id).
```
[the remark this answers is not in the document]
ieQu1: Translation of a map from/to a compact representation will be done here: https://github.com/emqx/emqx/pull/12251/files#diff-4f926b90b373b839679213fbdb3c30c784f5575bc3d4b721afce21466dded77eR394 (I haven't implemented any encoder callbacks yet).

I found that integer-keyed maps are rather unpleasant for debugging, I'd like to avoid them when possible.

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -0,0 +1,582 @@
...
+            Res
+    end.
+
+ro_transaction(Fun) ->
+    {atomic, Res} = mria:ro_transaction(?DS_MRIA_SHARD, Fun),
+    Res.
+
+-compile({inline, check_sequence/1}).
+
+-ifdef(CHECK_SEQNO).
```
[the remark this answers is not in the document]
ieQu1: It's a flag that can be enabled manually for local debugging.

apps/emqx_exhook/test/emqx_exhook_SUITE.erl
```diff
@@ -1,5 +1,5 @@
 %%--------------------------------------------------------------------
-%% Copyright (c) 2020-2023 EMQ Technologies Co., Ltd. All Rights Reserved.
+%% Copyright (c) 2020-2024 EMQ Technologies Co., Ltd. All Rights Reserved.
```
ieQu1, line 2: Auto-update copyright hook... 
This change is accidental, but I guess it won't hurt.

apps/emqx/test/emqx_persistent_messages_SUITE.erl
```diff
@@ -307,11 +283,6 @@ t_qos0_only_many_streams(_Config) ->
             receive_messages(3)
         ),
 
-        ?assertMatch(
```
ieQu1, line 310: No more pubranges...

apps/emqx/test/emqx_persistent_messages_SUITE.erl
```diff
@@ -205,31 +205,7 @@ t_session_subscription_iterators(Config) ->
                 messages => [Message1, Message2, Message3, Message4]
             }
         end,
-        fun(Trace) ->
-            ct:pal("trace:\n  ~p", [Trace]),
-            case ?of_kind(ds_session_subscription_added, Trace) of
```
ieQu1, line 210: This postcondition was verifying a property that is now assured by the `persistent_session_ds_state` module. Removed it to simplify the tests.

apps/emqx/src/emqx_schema.erl
```diff
@@ -1777,7 +1777,7 @@ fields("session_persistence") ->
             sc(
                 pos_integer(),
                 #{
-                    default => 1000,
+                    default => 100,
```
ieQu1, line 1780: Since the introduction of the mqueue buffer for persistent sessions, it's no longer necessary to use huge batches.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -845,124 +657,201 @@ do_ensure_all_iterators_closed(_DSSessionID) ->
...
+                    SeqNoQos2 = SeqNoQos20;
+                ?QOS_1 ->
+                    SeqNoQos1 = inc_seqno(?QOS_1, SeqNoQos10),
+                    SeqNoQos2 = SeqNoQos20;
+                ?QOS_2 ->
+                    SeqNoQos1 = SeqNoQos10,
+                    SeqNoQos2 = inc_seqno(?QOS_2, SeqNoQos20)
+            end,
+            {
+                case Msg#message.qos of
```
savonarola, line 770: nit: just
```suggestion
                case Qos of
```
?

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -1029,75 +888,230 @@ receive_maximum(ConnInfo) ->
...
+bump_interval() ->
+    emqx_config:get([session_persistence, last_alive_update_interval]).
+
+%%--------------------------------------------------------------------
+%% SeqNo tracking
+%% --------------------------------------------------------------------
+
+-spec commit_seqno(puback | pubrec | pubcomp, emqx_types:packet_id(), session()) ->
+    {ok, emqx_types:message(), session()} | {error, _}.
+commit_seqno(Track, PacketId, Session = #{id := SessionId, s := S}) ->
```
savonarola, line 900: Nit: Seems that `put_seqno` does not actually _commit_ (_to the database_) 🤔

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -403,214 +433,117 @@ deliver(_ClientInfo, _Delivers, Session) ->
...
+%% @doc Force syncing of the transient state to persistent storage
+sync(ClientId) ->
+    case emqx_cm:lookup_channels(ClientId) of
+        [Pid] ->
+            Ref = monitor(process, Pid),
+            Pid ! {emqx_session, #req_sync{from = self(), ref = Ref}},
+            receive
+                {'DOWN', Ref, process, _Pid, Reason} ->
+                    {error, Reason};
+                Ref ->
```
savonarola, line 542: demonitor?

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -0,0 +1,584 @@
...
+
+kv_pmap_persist(Tab, SessionId, Key, Val0) ->
+    %% Write data to mnesia:
+    Val = encoder(encode, Tab, Val0),
+    mnesia:write(Tab, #kv{k = {SessionId, Key}, v = Val}, write).
+
+kv_pmap_restore(Table, SessionId) ->
+    MS = [{#kv{k = {SessionId, '_'}, _ = '_'}, [], ['$_']}],
+    Objs = mnesia:select(Table, MS, read),
+    [{K, encoder(decode, Table, V)} || #kv{k = {_, K}, v = V} <- Objs].
```
savonarola, line 536: Nit: we may select `{K, V}` by the first `MS`

ieQu1: Done.

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: We definitely don't want to create a transaction on every message ack, so not committing the state is intentional. Renamed to `update_seqno`.

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: Thanks, fixed.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -845,124 +658,201 @@ do_ensure_all_iterators_closed(_DSSessionID) ->
...
+            #{props := SubOpts} = emqx_topic_gbt:get_record(SubMatch, Subs),
+            emqx_session:enrich_message(ClientInfo, Msg0, SubOpts, UpgradeQoS)
+        end
+    ],
+    {Inflight, LastSeqNoQos1, LastSeqNoQos2} = lists:foldl(
+        fun(Msg = #message{qos = Qos}, {Acc, SeqNoQos10, SeqNoQos20}) ->
+            case Qos of
+                ?QOS_0 ->
+                    SeqNoQos1 = SeqNoQos10,
+                    SeqNoQos2 = SeqNoQos20;
```
keynslug, line 762: Q: Admittedly, it's unclear to me why there's now 2 independent seqno tracks. Nice thing is `SeqNoRec - SeqNoComp` now gives something meaningful, but this property doesn't seem to be used (almost) anywhere. Or does it, do we plan to?

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -1029,75 +889,230 @@ receive_maximum(ConnInfo) ->
...
+    emqx_config:get([session_persistence, last_alive_update_interval]).
+
+%%--------------------------------------------------------------------
+%% SeqNo tracking
+%% --------------------------------------------------------------------
+
+-spec update_seqno(puback | pubrec | pubcomp, emqx_types:packet_id(), session()) ->
+    {ok, emqx_types:message(), session()} | {error, _}.
+update_seqno(Track, PacketId, Session = #{id := SessionId, s := S}) ->
+    SeqNo = packet_id_to_seqno(PacketId, S),
```
keynslug, line 902: Q: Should we reject things like `PUBREC(PacketID = 42)` (i.e. packet id from different track)?

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -1029,75 +889,230 @@ receive_maximum(ConnInfo) ->
...
+            %% We skip sequence numbers that lead to PacketId = 0 to
+            %% simplify math. Note: it leads to occasional gaps in the
+            %% sequence numbers.
+            NextSeqno + 1;
+        _ ->
+            NextSeqno
+    end.
+
+%% Note: we use the least significant bit to store the QoS. Even
+%% packet IDs are QoS1, odd packet IDs are QoS2.
```
keynslug, line 974: Nit. Outdated comment?

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -0,0 +1,590 @@
...
+kv_pmap_delete(Table, SessionId, Key) ->
+    %% Note: this match spec uses a fixed primary key, so it doesn't
+    %% require a table scan, and the transaction doesn't grab the
+    %% whole table lock:
+    mnesia:delete(Table, {SessionId, Key}, write).
+
+%%
+
+transaction(Fun) ->
+    mria:async_dirty(?DS_MRIA_SHARD, Fun).
```
keynslug, line 553: This looks risky. For consistency to be preserved the order of operations in `Fun` should be correct. Right now it almost certainly isn't (e.g. because not-yet-persisted state is kept in a map, streams' iterators are updated in unspecified order, etc.), so the consistency looks easy to break.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -403,214 +433,118 @@ deliver(_ClientInfo, _Delivers, Session) ->
...
-    {ok, [], emqx_session:ensure_timer(?TIMER_GET_STREAMS, Interval, Session)};
-handle_timeout(_ClientInfo, ?TIMER_BUMP_LAST_ALIVE_AT, Session0) ->
+    Session = emqx_session:ensure_timer(
+        ?TIMER_GET_STREAMS,
+        Interval,
+        Session0#{s => S}
+    ),
+    {ok, [], Session};
+handle_timeout(_ClientInfo, ?TIMER_BUMP_LAST_ALIVE_AT, Session0 = #{s := S0}) ->
+    S = emqx_persistent_session_ds_state:commit(bump_last_alive(S0)),
```
keynslug, line 458: I'd expects users will want this configurable, especially when QoS2 flows are involved.

apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
```diff
@@ -221,6 +218,7 @@ t_session_subscription_idempotency(Config) ->
     ClientId = <<"myclientid">>,
     ?check_trace(
         begin
+            #{timetrap => 20_000},
```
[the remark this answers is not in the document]
ieQu1: Thanks, stupid mistake.

apps/emqx/src/emqx_topic_gbt.erl
```diff
@@ -39,11 +39,11 @@
 -type match(ID) :: key(ID).
 
 -opaque t(ID, Value) :: gb_trees:tree(key(ID), Value).
--opaque t() :: t(_ID, _Value).
+-type t() :: t(_ID, _Value).
```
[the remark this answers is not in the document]
ieQu1: When `t/0` type is declared as opaque, dialyzer stops recognizing the fact that it's an alias to `t/2` and creates a bunch of bogus warnings when these two types are used together.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -0,0 +1,279 @@
...
+    Key = {SubId, erlang:phash2(Stream)},
+    case emqx_persistent_session_ds_state:get_stream(Key, S) of
+        undefined ->
+            ?SLOG(debug, #{
+                msg => new_stream, key => Key, stream => Stream
+            }),
+            {ok, Iterator} = emqx_ds:make_iterator(
+                ?PERSISTENT_MESSAGE_DB, Stream, TopicFilter, StartTime
+            ),
+            NewStreamState = #ifs{
```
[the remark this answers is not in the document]
ieQu1: `ifs` (I forgot to mention in the comment that it stands for in-flight state) is quite stupid name indeed, but it's short (bytes count when we store stuff in DB).

apps/emqx/src/emqx_session.erl
```diff
@@ -409,12 +409,8 @@ enrich_delivers(ClientInfo, Delivers, Session) ->
...
     [];
 enrich_delivers(ClientInfo, [D | Rest], UpgradeQoS, Session) ->
-    case enrich_deliver(ClientInfo, D, UpgradeQoS, Session) of
-        [] ->
-            enrich_delivers(ClientInfo, Rest, UpgradeQoS, Session);
-        Msg ->
-            [Msg | enrich_delivers(ClientInfo, Rest, UpgradeQoS, Session)]
-    end.
+    enrich_deliver(ClientInfo, D, UpgradeQoS, Session) ++
+        enrich_delivers(ClientInfo, Rest, UpgradeQoS, Session).
```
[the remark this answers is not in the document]
ieQu1: `++` is a [BIF](https://github.com/erlang/otp/blob/master/erts/emulator/beam/bif.tab#L328) though, and we save on a pattern-matching. Not sure what is more expensive in terms of micro-optimizations, always returning a list is nicer from types perspective.

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: Not sure if I understand the remark. `bump_interval()` function inside `bump_last_alive()` reads the configuration. `?TIMER_BUMP_LAST_ALIVE_AT` is the id of the timer event.

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: With separate tracks we don't have to remember the QoS of the sent packets in `emqx_inflight` to check the correctness of PUBRELs. 
Otherwise we'd have to keep track of the QoS for each sent packet ID somewhere, that's extra memory and extra CPU.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -403,214 +433,118 @@ deliver(_ClientInfo, _Delivers, Session) ->
...
+        fun({_StreamKey, Stream}, SessionAcc) ->
+            replay_batch(Stream, SessionAcc, ClientInfo)
+        end,
+        Session0,
+        Streams
+    ),
+    %% Note: we filled the buffer with the historical messages, and
+    %% from now on we'll rely on the normal inflight/flow control
+    %% mechanisms to replay them:
+    {ok, [], pull_now(Session)}.
```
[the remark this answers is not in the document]
ieQu1: It shouldn't make a lot of difference, but it makes code easier to follow, since there's only one place where we drain messages.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -845,124 +658,201 @@ do_ensure_all_iterators_closed(_DSSessionID) ->
...
-            end
-        end,
-        ExistingStreams,
-        TopicStreams
-    ).
+fetch_new_messages([], Session, _ClientInfo) ->
+    Session;
+fetch_new_messages([I | Streams], Session0 = #{inflight := Inflight}, ClientInfo) ->
+    BatchSize = emqx_config:get([session_persistence, max_batch_size]),
+    case emqx_persistent_session_ds_inflight:n_buffered(all, Inflight) >= BatchSize of
```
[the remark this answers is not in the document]
ieQu1: It sets up a limit on the number of buffered messages, that is `< BatchSize * 2`. I didn't introduce a new config parameter of `max_buffer_size` for simplicity: in the first approximation, the session doesn't gain anything in terms of throughput by storing more than two full batches in the process heap.

apps/emqx/src/emqx_persistent_session_ds_state.erl
ieQu1: Agree that this part is a bit sketchy. However, session must NOT rely on `mnesia_locker`. If we need a locking mechanism to prevent multiple versions of session from clobbering each others' persistent data, it should be explicit.

keynslug: It's not about concurrent updates, it's more about atomicity: `async_dirty` is not atomic in presence of failures I believe. It could have been much less of a problem if the operations were issued in the consistency-preserving order, but they are not.

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: Sorry, I was thinking about the configurability of the circumstances under which the session state is persisted.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -1029,75 +888,230 @@ receive_maximum(ConnInfo) ->
...
+    SeqNo = packet_id_to_seqno(PacketId, S),
+    case Track of
+        puback ->
+            MinTrack = ?committed(?QOS_1),
+            MaxTrack = ?next(?QOS_1);
+        pubrec ->
+            MinTrack = ?rec,
+            MaxTrack = ?next(?QOS_2);
+        pubcomp ->
+            MinTrack = ?committed(?QOS_2),
```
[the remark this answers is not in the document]
keynslug: Apologies, that question / remark should have been linked to the next line, that is `MaxTrack = ?next(?QOS_2)`.

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: I specifically tried to avoid adding new config parameters in this PR, since I find constant adding and deprecation of config parameters quite annoying.
 
If there is real-life need to fine tune these parameters separately, we can always add a new timer/new config.

apps/emqx/src/emqx_persistent_session_ds_state.erl
ieQu1: If we intend to move the sessions to DS, we cannot rely on the atomicity of updates either. What would be the consistency preserving order here?

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -845,124 +658,201 @@ do_ensure_all_iterators_closed(_DSSessionID) ->
...
-                TopicFilterWords = emqx_topic:words(TopicFilter),
-                renew_topic_streams(SessionId, TopicFilterWords, StartTime, Streams)
-            end,
-            ExistingStreams,
-            Subscriptions
-        )
-    end),
-    ok.
+fetch_new_messages(Session = #{s := S}, ClientInfo) ->
+    Streams = emqx_persistent_session_ds_stream_scheduler:find_new_streams(S),
```
keynslug, line 665: While thinking about consistency soundness of those changes, it just occurred to me that this is unsafe because `renew_streams/1` is nondeterministic. If the session fails to be persisted then the client will very likely see different order of messages during reconnect and replay.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -275,39 +267,86 @@ print_session(ClientId) ->
...
+    Session = #{id := ID, s := S0}
 ) ->
-    case subs_lookup(TopicFilter, Subs) of
-        Subscription = #{} ->
-            NSubscription = update_subscription(TopicFilter, Subscription, SubOpts, ID),
-            NSubs = subs_insert(TopicFilter, NSubscription, Subs),
-            {ok, Session#{subscriptions := NSubs}};
+    case subs_lookup(TopicFilter, S0) of
         undefined ->
-            % TODO: max_subscriptions
```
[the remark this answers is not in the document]
ieQu1: Indeed

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: There are three evils:
1) Transactions (with a singleton locker process handling everything in the node)
2) Storing everything in one huge record, and receiving comments about write amplification :wink: 
3) Async dirty with flimsy guarantees.

I chose 3 mostly because mnesia here is as temporary as it gets.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -622,204 +555,83 @@ storage() ->
...
-    Subscriptions = session_read_subscriptions(DSSessionId, write),
-    lists:foreach(
-        fun(#ds_sub{id = DSSubId} = DSSub) ->
-            TopicFilter = subscription_id_to_topic_filter(DSSubId),
-            ok = emqx_persistent_session_ds_router:do_delete_route(TopicFilter, DSSessionId),
-            ok = session_del_subscription(DSSub)
+session_ensure_new(Id, ConnInfo, Conf) ->
+    Now = now_ms(),
+    S0 = emqx_persistent_session_ds_state:create_new(Id),
+    S1 = emqx_persistent_session_ds_state:set_conninfo(ConnInfo, S0),
```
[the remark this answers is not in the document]
ieQu1: Makes sense, fixed.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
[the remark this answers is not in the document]
ieQu1: Fixed. Renamed `ifs` to `srs` (stream replay state), and added a comment.

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: Yeah. 🫠 Kinda feel the pain.

> because mnesia here is as temporary as it gets.

Then why not choose 1 instead? Could at least free ourselves from thinking about those edge cases. Now it feels like an unsafe optimization that looks sort of odd as part of a refactoring effort.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -0,0 +1,279 @@
...
+    CommQos1 = emqx_persistent_session_ds_state:get_seqno(?committed(?QOS_1), S0),
+    CommQos2 = emqx_persistent_session_ds_state:get_seqno(?committed(?QOS_2), S0),
+    %% 1. For each subscription, find the X ranks that were fully replayed:
+    Groups = emqx_persistent_session_ds_state:fold_streams(
+        fun({SubId, _Stream}, StreamState = #ifs{rank_x = RankX, rank_y = RankY}, Acc) ->
+            Key = {SubId, RankX},
+            case
+                {maps:get(Key, Acc, undefined), is_fully_replayed(CommQos1, CommQos2, StreamState)}
+            of
+                {undefined, true} ->
```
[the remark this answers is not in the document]
ieQu1: Thanks, applied.

apps/emqx/test/emqx_persistent_session_SUITE.erl
```diff
@@ -640,7 +644,7 @@ t_publish_many_while_client_is_gone_qos1(Config) ->
         #mqtt_msg{topic = <<"loc/1/2/42">>, payload = <<"M4">>, qos = 1},
         #mqtt_msg{topic = <<"t/42/foo">>, payload = <<"M5">>, qos = 1},
         #mqtt_msg{topic = <<"loc/3/4/5">>, payload = <<"M6">>, qos = 1},
-        #mqtt_msg{topic = <<"msg/feed/me">>, payload = <<"M7">>, qos = 1}
+        #mqtt_msg{topic = <<"msg/feed/me2">>, payload = <<"M7">>, qos = 1}
```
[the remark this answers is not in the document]
ieQu1: Rolled back.

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: Heh, both you and I probably missed the thread. This one is about nondeterministic replays.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
thalesmg: Right.  The binary representation of the stream term then?

apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
```diff
@@ -51,12 +51,12 @@ init_per_testcase(TestCase, Config) when
 init_per_testcase(t_session_gc = TestCase, Config) ->
     Opts = #{
         n => 3,
-        roles => [core, core, replicant],
```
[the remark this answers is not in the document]
ieQu1: Flakiness of the common test environment.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -845,124 +665,201 @@ do_ensure_all_iterators_closed(_DSSessionID) ->
...
+        {ok, end_of_stream} ->
+            %% No new messages; just update the end iterator:
+            {Srs0#srs{it_begin = ItBegin, it_end = end_of_stream, batch_size = 0}, Inflight0};
+        {error, _} when not IsReplay ->
+            ?SLOG(info, #{msg => "failed_to_fetch_batch", iterator => ItBegin}),
+            {Srs0, Inflight0}
+    end.
+
+%% key_of_iter(#{3 := #{3 := #{5 := K}}}) ->
+%%     K.
```
thalesmg, line 741: :scissors: ?

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -0,0 +1,584 @@
...
+        table :: atom(),
+        cache :: #{K => V},
+        dirty :: #{K => dirty | del}
+    }.
+
+-type metadata() ::
+    #{
+        ?created_at => emqx_persistent_session_ds:timestamp(),
+        ?last_alive_at => emqx_persistent_session_ds:timestamp(),
+        ?expiry_interval => emqx_types:conninfo(),
```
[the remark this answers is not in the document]
ieQu1: Right, I removed conninfo from the state after Andrew's remark, but forgot to update the type here.

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -0,0 +1,584 @@
...
+%%         true ->
+%%             Fun();
+%%         false ->
+%%             {atomic, Res} = mria:transaction(?DS_MRIA_SHARD, Fun),
+%%             Res
+%%     end.
+
+%% ro_transaction(Fun) ->
+%%     {atomic, Res} = mria:ro_transaction(?DS_MRIA_SHARD, Fun),
+%%     Res.
```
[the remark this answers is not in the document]
ieQu1: Good to keep this comment here as an eyesore/reminder that state update is not atomic, and we need to do something about it.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -0,0 +1,277 @@
...
+remove_fully_replayed_streams(S0) ->
+    CommQos1 = emqx_persistent_session_ds_state:get_seqno(?committed(?QOS_1), S0),
+    CommQos2 = emqx_persistent_session_ds_state:get_seqno(?committed(?QOS_2), S0),
+    %% 1. For each subscription, find the X ranks that were fully replayed:
+    Groups = emqx_persistent_session_ds_state:fold_streams(
+        fun({SubId, _Stream}, StreamState = #srs{rank_x = RankX, rank_y = RankY}, Acc) ->
+            Key = {SubId, RankX},
+            case is_fully_replayed(CommQos1, CommQos2, StreamState) of
+                true when is_map_key(Key, Acc) ->
+                    Acc#{Key => {true, RankY}};
```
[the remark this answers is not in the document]
ieQu1: Yes, it relies on the fact that `find_replay_streams` never returns different Ys for the same X.

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: > But I think LTS may lead to independent streams containing the same topic (the paths before and after "learning")?

This must never happen within a given generation: `get_streams` for the LTS must always return the _original_ index for the topic.
New topics will end up in the wildcard-augmented stream, but the old ones are stored in their own streams.

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -0,0 +1,584 @@
...
+    transaction(fun() ->
+        delete(SessionId),
+        #{
+            id => SessionId,
+            metadata => #{},
+            subscriptions => emqx_topic_gbt:new(),
+            streams => pmap_open(?stream_tab, SessionId),
+            seqnos => pmap_open(?seqno_tab, SessionId),
+            ranks => pmap_open(?rank_tab, SessionId),
+            ?set_dirty
```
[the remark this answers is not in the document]
ieQu1: We don't want to skip the first `commit` when the new session is created.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -0,0 +1,277 @@
...
+%% Type declarations
+%%================================================================================
+
+%%================================================================================
+%% API functions
+%%================================================================================
+
+-spec find_replay_streams(emqx_persistent_session_ds_state:t()) ->
+    [{emqx_persistent_session_ds_state:stream_key(), emqx_persistent_session_ds:stream_state()}].
+find_replay_streams(S) ->
```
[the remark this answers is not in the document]
ieQu1: Great idea, fixed.

apps/emqx/test/emqx_persistent_session_ds_state_tests.erl
```diff
@@ -0,0 +1,370 @@
...
+precondition(_, _) ->
+    true.
+
+postcondition(S, {call, ?MODULE, iterate_sessions, [_]}, Result) ->
+    {Sessions, _} = lists:unzip(Result),
+    %% No lingering sessions:
+    ?assertMatch([], Sessions -- maps:keys(S)),
+    %% All committed sessions are visited by the iterator:
+    CommittedSessions = lists:sort([K || {K, #s{committed = true}} <- maps:to_list(S)]),
+    ?assertMatch([], CommittedSessions -- Sessions),
```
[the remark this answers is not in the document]
ieQu1: I'm not sure if it's useful property.

apps/emqx/test/emqx_persistent_session_ds_state_tests.erl
```diff
@@ -0,0 +1,370 @@
...
+        SessionId,
+        Idx,
+        fun(Map) -> maps:remove(Key, Map) end,
+        S
+    );
+next_state(S, _V, {call, ?MODULE, commit, [SessionId]}) ->
+    update(
+        SessionId,
+        #s.committed,
+        fun(_) -> true end,
```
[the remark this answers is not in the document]
ieQu1: Here, `committed` means "has been committed at least once since the time of creation", rather than "not dirty".

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: This could be useful for debugging later.

apps/emqx/test/emqx_persistent_session_SUITE.erl
```diff
@@ -895,26 +913,27 @@ t_publish_many_while_client_is_gone(Config) ->
...
     {ok, _} = emqtt:ConnFun(Client3),
 
-    %% Only the last unacked QoS 1 message should be retransmitted.
+    %% Check that the messages are retransmitted with DUP=1:
     Msgs3 = receive_messages(NPubs, _Timeout = 2000),
     ct:pal("Msgs3 = ~p", [Msgs3]),
     ?assertMatch(
-        [#{topic := <<"t/100/foo">>, payload := <<"M11">>, qos := 1, dup := true}],
-        Msgs3
+        [<<"M10">>, <<"M11">>, <<"M12">>],
```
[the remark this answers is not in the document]
ieQu1: Fixed.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -0,0 +1,312 @@
...
+%%
+%% 2. For each RankX, it never advances RankY until _all_ streams with
+%% the same RankX are replayed.
+%%
+%% 3. Once all streams with the given rank are replayed, it advances
+%% the RankY to the smallest known RankY that is greater than replayed
+%% RankY.
+%%
+%% 4. If the RankX has never been replayed, it keeps the streams with
+%% the smallest RankX.
```
thalesmg, line 119:
```suggestion
%% 2. For each RankX, it never advances RankY until _all_ streams with
%% the same RankY are replayed.
%%
%% 3. Once all streams with the given rank are replayed, it advances
%% the RankY to the smallest known RankY that is greater than replayed
%% RankY.
%%
%% 4. If the RankX has never been replayed, it keeps the streams with
%% the smallest RankY.
```
?

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -1029,75 +896,229 @@ receive_maximum(ConnInfo) ->
...
+    Min = max(0, NextSeqNo - WindowSize),
+    Max = max(0, NextSeqNo - 1),
+    range(Min, Max).
+
+%% Generate a sequence number:
+next_seqno_gen() ->
+    ?LET(
+        {Epoch, Offset},
+        {non_neg_integer(), range(0, ?EPOCH_SIZE)},
+        Epoch bsl 15 + Offset
```
thalesmg, line 1040: Nit:

```suggestion
        Epoch bsl ?EPOCH_BITS + Offset
```

thalesmg (approved):

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -403,214 +434,123 @@ deliver(_ClientInfo, _Delivers, Session) ->
...
-    %% FIXME: This is a temporary workaround to avoid crashes when starting on Windows
-    case mria:rocksdb_backend_available() of
-        true ->
-            rocksdb_copies;
-        _ ->
-            disc_copies
+    emqx_persistent_session_ds_state:create_tables().
+
+%% @doc Force syncing of the transient state to persistent storage
+sync(ClientId) ->
```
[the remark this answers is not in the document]
ieQu1: Right, could be useful for the tests that have to inspect the state of the session.

apps/emqx/src/emqx_persistent_session_ds_stream_scheduler.erl
ieQu1: %% 2. For each RankX, it never advances RankY until _all_ streams with
%% the same RankX are replayed.

This is actually correct: streams with the different RankX can have different RankY. Other comment already implies that this function always returns the streams with the same RankY for the given X.

apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
```diff
@@ -164,59 +165,27 @@ is_persistent_connect_opts(#{properties := #{'Session-Expiry-Interval' := EI}})
...
-        lists:member(K, [ds_session_gc, ds_session_gc_lock_taken]),
-        N =/= node()
-    ]),
-    ?assertEqual(lists:usort(CoreNodes), GCNodes).
-
 %%------------------------------------------------------------------------------
 %% Testcases
 %%------------------------------------------------------------------------------
 
-t_non_persistent_session_subscription(_Config) ->
```
ieQu1, line 193: I removed this testcase because it tests in-memory session. Not sure if it's relevant for this testsuite. Also, testcases that verify absence of certain trace events are unreliable on their own: if a trace event is renamed or removed, the test will always pass.

apps/emqx/src/emqx_persistent_session_ds.erl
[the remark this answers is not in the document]
ieQu1: I made the logic more strict, so this shouldn't be relevent anymore.

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: > From the [EIP](https://github.com/emqx/eip/blob/162edbbbfcc6dce1eed5f164d516cb9e7214330e/active/0023-rocksdb-message-persistence.md?plain=1#L24), we don't provide guarantees about ordering between different topics.

Just wanted to clarify. The problem lies in how existing sessions are replayed in the event of a failure, when recent session state updates were lost. The client will reconnect and see that the session was present, but then receive some of the messages it has already seen but with different PacketIDs. This essentially violates [MQTT spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901238), and makes it harder for clients to process messages idempotently.

PR #12290, opened 2024-01-10
fix(retain): add `hasnext` into the meta data for the `GET /retain/messages`

lafirest:
[EMQX-11709](https://emqx.atlassian.net/browse/EMQX-11709)

Release version: `v/e5.5.0`

## PR Checklist

### DIFF
--- a/apps/emqx_retainer/src/emqx_retainer_api.erl
+++ b/apps/emqx_retainer/src/emqx_retainer_api.erl
@@ -181,15 +181,27 @@ lookup_retained(get, #{query_string := Qs}) ->
     Page = maps:get(<<"page">>, Qs, 1),
     Limit = maps:get(<<"limit">>, Qs, emqx_mgmt:default_row_limit()),
     Topic = maps:get(<<"topic">>, Qs, undefined),
-    {ok, Msgs} = emqx_retainer_mnesia:page_read(undefined, Topic, Page, Limit),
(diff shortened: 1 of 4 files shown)


apps/emqx_retainer/src/emqx_retainer_api.erl
```diff
@@ -184,7 +184,17 @@ lookup_retained(get, #{query_string := Qs}) ->
...
-        meta => #{page => Page, limit => Limit, count => emqx_retainer_mnesia:size(?TAB_MESSAGE)}
+        meta => #{
+            page => Page,
+            limit => Limit,
+            count =>
+                case Topic of
+                    undefined ->
+                        emqx_retainer_mnesia:size(?TAB_MESSAGE);
+                    _ ->
+                        0
```
zmstone, line 195: `length(Msgs)` ?

zmstone, line 195: I wonder if it makes more sense to simply not return count, or as string value `"unknown"`

zmstone:
missing changelog.

apps/emqx_retainer/src/emqx_retainer_api.erl
lafirest: We should not break with the consistent of the data schema, and this mechanism also has be applied in other APIs

zmstone (approved):

zmstone (approved):

PR #12294, opened 2024-01-10
fix(emqx_router_helper): make sure late routes cleanup cannot remove new routes of the restarted node

SergeTupchiy:
Release version: v/e5.?

## PR Checklist


apps/emqx/test/emqx_routing_SUITE.erl
```diff
@@ -307,8 +303,92 @@ t_slow_rlog_routing_consistency(Config) ->
...
+                _Timeout = infinity
+            ),
+
+            emqx_cth_cluster:restart(Replicant, ReplSpec),
+            ok = rpc:call(Replicant, emqx_router, do_add_route, [Topic1]),
+
+            %% wait for Mria route replication,
+            %% cannot poll has_route, because late cleanup may wipe it away
+            %% if the test is about to fail due to a race
+            timer:sleep(1000),
```
keynslug, line 345: Q: Why is this sleep needed?

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -101,6 +101,12 @@ monitor(Node) when is_atom(Node) ->
 
 init([]) ->
     process_flag(trap_exit, true),
+    %% Make sure no other node can do late routes cleanup when this node is up.
+    true = global:set_lock({?LOCK(node()), self()}),
```
keynslug, line 91: Good idea! If we do `global:set_lock({?LOCK(node()), self()}, [node()])` here, I wonder will it be as safe as it is now? Otherwise there's a risk of blocking the node startup if some replicant somewhere is unreachable. Or maybe I'm missing something. 🤔

apps/emqx/test/emqx_routing_SUITE.erl
SergeTupchiy: To wait for core -> replicant  new route replication, the comment above explains it (though probably it is not very clear..). 
The test case assumes that cleanups may ocur after node is up again, so the cleanups can remove new routes. 
The expected order of events:
1. restart replicant
2. add new route on the replicant
3. wait a little to give some time to a core node to replicat this new route back to the replicant
4. make sure all cleanups have been done (via snabkaffe)
5. check that every node has the new route but doesn't have the old one.

apps/emqx/test/emqx_routing_SUITE.erl
```diff
@@ -307,8 +303,92 @@ t_slow_rlog_routing_consistency(Config) ->
...
+    Config;
+t_cleanup_routes_race('end', Config) ->
+    ok = snabbkaffe:stop(),
+    [Core1, Core2, _Repl] = ?config(cluster, Config),
+    lists:foreach(
+        fun unmock_router_helper/1,
+        [Core1, Core2]
+    ).
+
+t_cleanup_routes_race(Config) ->
```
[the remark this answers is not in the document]
SergeTupchiy: Yeah, perhaps I should move it to `emqx_router_helper_SUITE`... `emqx_routing_suite` uses a 'real' cluster, while emqx_router_helper_SUITE` doesn't, so my initial intention was to keep cluster-based routing test cases in the same suite, but it probably makes little sense...

apps/emqx/src/emqx_router_helper.erl
SergeTupchiy: Honestly, that's was suggested by @ieQu1, so not my idea. :slightly_smiling_face: 
You mean that this node tries to acquire the lock on all currently live nodes (`[node()|nodes()]`) but some of these nodes are unreachable? If so, `global` must definitely handle it by ignoring unreachable nodes:
```
set_lock_on_nodes(Id, Nodes) ->
    case local_lock_check(Id, Nodes) of 
        true ->
            Msg = {set_lock, Id},
            {Replies, _} = 
                gen_server:multi_call(Nodes, global_name_server, Msg),
```
Here it ignores BadNodes, so if a node is unreachable, the lock will be acquired only on live nodes.

keynslug: Yes, but it does so with `infinity` timeout (according to `gen_server:multicall/3`). Granted it won't realistically become stuck, but I must admit it's not really clear how long the multicall itself will take in the worst case (i.e. when some node gone down but this node haven't noticed that yet).

> If we do `global:set_lock({?LOCK(node()), self()}, [node()])` here

Anyway, disregard that, it's unsafe, haven't thought it through enough. 🫠

keynslug: However, I don't honestly see an alternative that is simple enough, so maybe just keep that in mind / leave a comment here? Admittedly, this situation shouldn't really happen often.

SergeTupchiy: I think the worst case (if a node is really unresponsive) depends on the `ticktime`, which, acc. to the docs is 
```
MaxT = NetTickTime + NetTickTime / NetTickIntensity
``` 
(https://www.erlang.org/doc/man/kernel_app#net_ticktime)
So, even though gen_server multicall uses infinity timeout, it's expected to receive 'DOWN' msg if the node is not responding..

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -192,5 +112,37 @@ code_change(_OldVsn, State, _Extra) ->
...
+                %% since replicants participate in locking (see init/1).
+                [node() | nodes()],
+                %% It's important to do a limited number of retries (not infinity).
+                %% Otherwise, if a down node comes back in a short time
+                %% and wins the lock in init/1,
+                %% emqx_router_helepr on another node can desparetly retry it indefinitely.
+                2
+            );
+        replicant when Node =:= node() ->
+            %% Mnesia is down locally, but emqx_router_helper is still running and holds the lock,
```
SergeTupchiy, line 142: This whole concept `effectively` treats any crashes too seriously...
if `mria_membership` crashes on a replicant:
1. emqx_router_helper on this replicant node will delete the lock.
2. another node will receive `mria, down` event, acquire a lock and cleanup the routes
However, mria_membership proc can be restarted and the replicant will continue its operation, as though nothing happen...

Probably it's better to do routes cleanups only when the node is really down but not as soon as Mria (or even Mnesia is down)? :thinking:

apps/emqx/test/emqx_routing_SUITE.erl
```diff
@@ -307,8 +303,92 @@ t_slow_rlog_routing_consistency(Config) ->
...
+            timer:sleep(1000),
+            ?assertMatch({ok, _}, snabbkaffe:receive_events(SubRef)),
+
+            lists:foreach(
+                fun(N) ->
+                    ?assertNot(rpc:call(N, emqx_router, has_route, [Topic, Replicant])),
+                    ?assert(rpc:call(N, emqx_router, has_route, [Topic1, Replicant]))
+                end,
+                Nodes
+            )
```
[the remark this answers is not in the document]
SergeTupchiy: Done, thanks.

apps/emqx/src/emqx_router_helper.erl
SergeTupchiy: Added a comment

apps/emqx/test/emqx_routing_SUITE.erl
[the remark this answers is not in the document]
SergeTupchiy: Moved to `emqx_router_helper_SUITE`

apps/emqx/src/emqx_router_helper.erl
SergeTupchiy: Changed `{mria, down}` events to be sent only when `mria_membership` terminates with `shutdown` reason, so that if it crashes and restarts due to some errors, it doesn't cause any cleanups. 
If Mria is shutdown on a remote node, it makes sense to cleanup routes even if that node keeps being connected at Erlang distr level..

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -129,5 +123,45 @@ code_change(_OldVsn, State, _Extra) ->
...
+                %% as A's lock is still present on node C (C and A still can communicate).
+                %% If retries were set to infinity, there will be a risk for emqx_router_helper
+                %% to be blocked forever on node B, trying to acquire the lock and do cleanup.
+                2
+            );
+        replicant when Node =:= node() ->
+            %% Mnesia (or Mria) is down locally,
+            %% but emqx_router_helper is still running and holds the lock,
+            %% remove it so that other healthy nodes can do cleanup.
+            global:del_lock({?LOCK(Node), self()});
```
keynslug, line 163: Q: Hm, is it really something that can happen during normal operation? What would be the circumstances that trigger it? I'm under impression that mria / mnesia app is started in permanent mode in production releases, so it going down should cause whole node to terminate anyway.

keynslug (commented):
🎉

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -192,5 +122,61 @@ code_change(_OldVsn, State, _Extra) ->
...
+maybe_drain_nodedown_msg(What, DownNode) when DownNode =/= node(), What =/= node ->
+    receive
+        {membership, {node, down, DownNode}} -> ok
+    after ?DRAIN_TIMEOUT -> ok
+    end;
+maybe_drain_nodedown_msg(_, _) ->
+    ok.
+
+locking_nodes() ->
+    locking_nodes(mria_rlog:role()).
```
zmstone, line 137: this and `mria_config:whoami()`, which is the correct API ?

SergeTupchiy: I think both are equally usable. `whoami/0` will tel if the rlog is disabled and will return `mnesia`. 
`role/0` always return `core` or `replicant`.

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -192,5 +122,61 @@ code_change(_OldVsn, State, _Extra) ->
...
+
+locking_nodes(core) ->
+    mria_mnesia:running_nodes();
+locking_nodes(replicant) ->
+    %% No reliable source to get the list of core nodes, unfortunately.
```
zmstone, line 142: maybe not critical to this PR. but it would be good to have mria brodcast core nodes to all replicants ?

SergeTupchiy: For sure, it has this mechanism, but I find it is not very suitable, especially when called upon app start in `init/1`, please see my comment: https://github.com/emqx/emqx/pull/12294#discussion_r1452553704

SergeTupchiy: The replicant node inserts info about core nodes by doing async pings, so at the point when emqx_router_helper is started, it can probably get an empty core nodes list, unless enough time has passed..

apps/emqx/src/emqx_router_helper.erl
SergeTupchiy: @zmstone what do you think about ignoring mnesia/mria down events  in the router_helper and only do cleanups when nodedown is received?

apps/emqx/src/emqx_router_helper.erl
SergeTupchiy: @keynslug suggested a good alternative for now:
```
mria:async_dirty(?ROUTE_SHARD, fun mria_mnesia:running_nodes/0, []).
```

PR #12354, opened 2024-01-18
Apply brdige post config changes in parallel

SergeTupchiy:
Fixes EMQX-11751

Each bridge creation can take signficant time. 
For example, if `start_after_created = true` (which is default), and the bridge connector implementation doesn't set `disconnected` state but only uses `connected` or `connecting`,  the bridge creation can be blocked for up to `start_timeout` (defaults to 5s) if the data sink is not reachable/healthy. 

When the batch of bridges is being updated/created, there no need to wait for a bridge creation before proceeding to the next one.
We can create all of them in parallel and wait for results/errors. 

This can greatly improve config load / backup file import performance, when there are many bridges and some/all data sinks are not reachable/healthy. 

Release version: v/e5.5

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge.erl
+++ b/apps/emqx_bridge/src/emqx_bridge.erl
@@ -103,33 +103,37 @@
 
 load() ->
     Bridges = emqx:get_config([?ROOT_KEY], #{}),
-    lists:foreach(
+    emqx_utils:pforeach(
         fun({Type, NamedConf}) ->
-            lists:foreach(
+            emqx_utils:pforeach(
                 fun({Name, Conf}) ->
                     %% fetch opts for `emqx_resource_buffer_worker`
                     ResOpts = emqx_resource:fetch_creation_opts(Conf),
                     safe_load_bridge(Type, Name, Conf, ResOpts)
                 end,
-                maps:to_list(NamedConf)
+                maps:to_list(NamedConf),
+                infinity
             )
         end,
-        maps:to_list(Bridges)
+        maps:to_list(Bridges),
+        infinity
     ).
 
 unload() ->
     unload_hook(),
     Bridges = emqx:get_config([?ROOT_KEY], #{}),
-    lists:foreach(
+    emqx_utils:pforeach(
         fun({Type, NamedConf}) ->
-            lists:foreach(
+            emqx_utils:pforeach(
                 fun({Name, _Conf}) ->
                     _ = emqx_bridge_resource:stop(Type, Name)
                 end,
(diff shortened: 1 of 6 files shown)


SergeTupchiy:
NOTE: the behaviour is now slightly changed. Previously, if one bridge post_config operation failed, we didn't even try the next bridge. 
Now, we do everything in parallel, so one bridge failure doesn't stop us from trying to create/update/delete other bridges. 
I think it's only for the best, since bridges are independent.
All the errors are collected and eventually returned from `post_config_update` callbacks.

zhongwencool:
we should also do this work in `emqx_connector:post_config_update` ?
It is the one actually connecting to the external host in v2.

SergeTupchiy:
> we should also do this work in `emqx_connector:post_config_update` ? It is the one actually connecting to the external host in v2.

thanks, done

apps/emqx_bridge/src/emqx_bridge.erl
```diff
@@ -103,34 +103,40 @@
 
 load() ->
     Bridges = emqx:get_config([?ROOT_KEY], #{}),
-    lists:foreach(
+    _ = emqx_utils:pmap(
```
zmstone, line 106: maybe make a `emqx_utils:pforeach/2` so there is no need to do `_ = `

zmstone:
I could not quickly find where but I remember there were some ordering applied to deal with deletes of actions and connectors due to the dependency.
Please make sure the order is not broken when changed to parallel

changes/ce/perf-12354.en.md
```diff
@@ -0,0 +1,3 @@
+Apply post config bridge changes in parallel.
+This can greatly improve the performance when many bridges are being changed,
```
ieQu1, line 2: Nit:
```suggestion
This can greatly improve the performance when multiple bridges are being changed, 
```

SergeTupchiy: fixed

ieQu1 (approved):

SergeTupchiy:
> I could not quickly find where but I remember there were some ordering applied to deal with deletes of actions and connectors due to the dependency between these two root entries. Please make sure the order is not broken when changed to parallel

It only runs a single action (delete,create,update) items in parallel and waits for all items before returning from post_config_update.. :thinking: 

It doesn't introduce a risk of something like deleting both actions and connectors at the same time, this must be already serialized at emqx_config level.

Connector must not be deleted if it's used by at least one action, but I don't see how this can be broken by this change.. :thinking:

zmstone (approved):

PR #12361, opened 2024-01-21
feat(ds): implement raft-based replication

keynslug:
Fixes [EMQX-11756](https://emqx.atlassian.net/browse/EMQX-11756).

## Summary

Uses Raft through https://github.com/rabbitmq/ra to replicate writes across nodes: each shard is now a Raft consensus group. Replicas are guaranteed to not diverge, therefore we're free to pick any replica to perform reads.

* [x] Total order of messages in each shard
* [x] Session replay tests work
* [x] All tests are green

### Roadmap

This is mostly out-of-scope for this PR (maybe except for first two), but good to have written down.

1. Measure performance impact.

    Read performance shouldn't really suffer. Things to explore: issue `release_cursor` less often, make smarter assumptions about shard servers, do not waste cycles on constructing names / IDs.

2. Perform rest of shard-level operations through consensus.

    Namely: adding generations. At first look it seems important because of snapshotting. Initially we could make it work synchronously, where adding a generation will block message writes. Then think about doing asynchronously while preserving safety.

3. Solve snapshotting story.

    We could relax strictness and take the most recent snapshots, instead of taking them at the specified Raft index. Supposedly, this should be safe as long as writes are idempotent, and repeating write operations already present in a snapshot will not accidentally cause write duplication. At first glance, the only thing needed is making _add generation_ part of Raft logs.

4. Design good first start UX story.

    In other words: how operators are supposed to ensure that the cluster they want to deploy will start and fairly allocate shards between them. Currently node in the cluster does not know how large the cluster is supposed to be, and could steal too much shards / replicas.

5. Perform cluster-wide changes through consensus.

    Related to the above. Super important to unlock safe rebalancing. Quite tricky because of uncertainty about initial cluster membership.

6. Implement rebalancing.

7. Skip RocksDB WAL.

    This is a rather long-term goal. We already have a Raft log which serves basically the same purpose, so disabling RocksDB WAL and taking control over dumping memtables / releasing Raft log entries should help to avoid keeping two logs.

## PR Checklist

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds.erl
@@ -295,6 +295,7 @@ drop_db(DB) ->
         undefined ->
             ok;
         Module ->
+            _ = persistent_term:erase(?persistent_term(DB)),
             Module:drop_db(DB)
     end.
 
--- a/apps/emqx_durable_storage/src/emqx_ds.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds.erl
@@ -43,6 +43,7 @@
 
 %% Misc. API:
 -export([count/1]).
+-export([timestamp_us/0]).
 
 -export_type([
     create_db_opts/0,
@@ -147,9 +148,11 @@
 -type error(Reason) :: {error, recoverable | unrecoverable, Reason}.
 
 %% Timestamp
+%% Each message must have unique timestamp.
 %% Earliest possible timestamp is 0.
-%% TODO granularity?  Currently, we should always use milliseconds, as that's the unit we
-%% use in emqx_guid.  Otherwise, the iterators won't match the message timestamps.
+%% Granularity: microsecond.
+%% TODO: Currently, we should always use milliseconds, as that's the unit we
+%% use in emqx_guid. Otherwise, the iterators won't match the message timestamps.
 -type time() :: non_neg_integer().
 
 -type message_store_opts() ::
@@ -393,6 +396,10 @@ count(DB) ->
 %%================================================================================
 %% Internal exports
 %%================================================================================
+
+-spec timestamp_us() -> time().
+timestamp_us() ->
+    erlang:system_time(microsecond).
 
 %%================================================================================
 %% Internal functions
--- a/apps/emqx_durable_storage/src/emqx_ds.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds.erl
@@ -150,9 +150,6 @@
 %% Timestamp
 %% Each message must have unique timestamp.
 %% Earliest possible timestamp is 0.
-%% Granularity: microsecond.
-%% TODO: Currently, we should always use milliseconds, as that's the unit we
-%% use in emqx_guid. Otherwise, the iterators won't match the message timestamps.
 -type time() :: non_neg_integer().
 
 -type message_store_opts() ::
--- a/apps/emqx_durable_storage/src/emqx_ds_builtin_db_sup.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_builtin_db_sup.erl
@@ -21,18 +21,14 @@
 -behaviour(supervisor).
 
 %% API:
--export([start_db/2, start_shard/1, start_egress/1, stop_shard/1, ensure_shard/1]).
--export([status/1]).
+-export([start_db/2, start_shard/1, start_egress/1, stop_shard/1, ensure_shard/1, ensure_egress/1]).
+-export([which_shards/1]).
 
 %% behaviour callbacks:
 -export([init/1]).
--export([handle_call/3, handle_cast/2, handle_info/2, terminate/2]).
 
 %% internal exports:
 -export([start_link_sup/2]).
-
-%% FIXME
--export([lookup_shard_meta/2]).
 
 %%================================================================================
 %% Type declarations
@@ -41,14 +37,14 @@
 -define(via(REC), {via, gproc, {n, l, REC}}).
 
 -define(db_sup, ?MODULE).
--define(shard_sup, emqx_ds_builtin_db_shard_sup).
+-define(shards_sup, emqx_ds_builtin_db_shards_sup).
 -define(egress_sup, emqx_ds_builtin_db_egress_sup).
+-define(shard_sup, emqx_ds_builtin_db_shard_sup).
 
 -record(?db_sup, {db}).
--record(?shard_sup, {db}).
+-record(?shards_sup, {db}).
 -record(?egress_sup, {db}).
-
--define(shard_meta(DB, SHARD), {?MODULE, DB, SHARD}).
+-record(?shard_sup, {db, shard}).
 
 %%================================================================================
 %% API funcions
@@ -60,8 +56,8 @@ start_db(DB, Opts) ->
 
 -spec start_shard(emqx_ds_storage_layer:shard_id()) ->
     supervisor:startchild_ret().
-start_shard(Shard = {DB, _}) ->
-    supervisor:start_child(?via(#?shard_sup{db = DB}), shard_spec(DB, Shard)).
+start_shard({DB, Shard}) ->
+    supervisor:start_child(?via(#?shards_sup{db = DB}), shard_spec(DB, Shard)).
 
 -spec start_egress(emqx_ds_storage_layer:shard_id()) ->
     supervisor:startchild_ret().
@@ -70,28 +66,24 @@ start_egress({DB, Shard}) ->
 
 -spec stop_shard(emqx_ds_storage_layer:shard_id()) -> ok | {error, _}.
 stop_shard(Shard = {DB, _}) ->
-    Sup = ?via(#?shard_sup{db = DB}),
+    Sup = ?via(#?shards_sup{db = DB}),
     ok = supervisor:terminate_child(Sup, Shard),
     ok = supervisor:delete_child(Sup, Shard).
 
 -spec ensure_shard(emqx_ds_storage_layer:shard_id()) ->
     ok | {error, _Reason}.
 ensure_shard(Shard) ->
-    case start_shard(Shard) of
-        {ok, _Pid} ->
-            ok;
-        {error, {already_started, _Pid}} ->
-            ok;
-        {error, Reason} ->
-            {error, Reason}
-    end.
+    ensure_started(start_shard(Shard)).
 
-status(DB) ->
-    State = sys:get_state(?via({allocator, DB})),
-    maps:get(status, State).
+-spec ensure_egress(emqx_ds_storage_layer:shard_id()) ->
+    ok | {error, _Reason}.
+ensure_egress(Shard) ->
+    ensure_started(start_egress(Shard)).
 
-lookup_shard_meta(DB, Shard) ->
-    persistent_term:get(?shard_meta(DB, Shard)).
+-spec which_shards(emqx_ds:db()) ->
+    [_Child].
+which_shards(DB) ->
+    supervisor:which_children(?via(#?shards_sup{db = DB})).
 
 %%================================================================================
 %% behaviour callbacks
@@ -103,7 +95,7 @@ init({#?db_sup{db = DB}, DefaultOpts}) ->
     Opts = emqx_ds_replication_layer_meta:open_db(DB, DefaultOpts),
     ok = start_ra_system(DB, Opts),
     Children = [
-        sup_spec(#?shard_sup{db = DB}, []),
+        sup_spec(#?shards_sup{db = DB}, []),
         sup_spec(#?egress_sup{db = DB}, []),
         shard_allocator_spec(DB, Opts)
     ],
@@ -113,8 +105,8 @@ init({#?db_sup{db = DB}, DefaultOpts}) ->
         period => 1
     },
     {ok, {SupFlags, Children}};
-init({#?shard_sup{db = _DB}, _}) ->
-    %% Spec for the supervisor that manages the worker processes for
+init({#?shards_sup{db = _DB}, _}) ->
+    %% Spec for the supervisor that manages the supervisors for
     %% each local shard of the DB:
     SupFlags = #{
         strategy => one_for_one,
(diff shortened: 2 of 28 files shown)


apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
```diff
@@ -344,3 +339,167 @@ shard_of_messages(DB, [#message{from = From} | _]) ->
...
+    iolist_to_binary(io_lib:format("~s_~s", [DB, Shard])).
+
+ra_server_name(DB, Shard, Site) ->
+    DBBin = atom_to_binary(DB),
+    binary_to_atom(<<"ds_", DBBin/binary, Shard/binary, "_", Site/binary>>).
+
+%%
+
+init(#{db := DB, shard := Shard}) ->
+    _ = erlang:put(emqx_ds_db_shard, {DB, Shard}),
```
thalesmg, line 474: Q: is there a particular reason not to put this in the state map?

apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl
```diff
@@ -412,11 +410,10 @@ make_key(#s{keymappers = KeyMappers, trie = Trie}, #message{timestamp = Timestam
 ]) ->
     binary().
 make_key(KeyMapper, TopicIndex, Timestamp, Varying) ->
-    UniqueInteger = erlang:unique_integer([monotonic, positive]),
```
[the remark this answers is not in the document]
keynslug: Since we want total order of messages per stream, this PR just enforces total order per shard through allocating [unique pseudo-realtime timestamps](https://github.com/emqx/emqx/blob/c199c0b1613b7de18f760eb1e1cf4a1efc9e11d0/apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl#L491) to messages (admittedly, in a quite ugly way). Since we don't want timestamps to occasionally jump backwards during leader election it's important to have that as part of Raft log / state, and uniqueness (i.e. strict monotonicity) is trivial to add on top of that.

With that we don't need this extra uniqueness, which furthermore is not really guaranteed to be unique across shard's replica set.

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
```diff
@@ -344,3 +339,167 @@ shard_of_messages(DB, [#message{from = From} | _]) ->
...
+    ok = ra:start_server(_System = default, #{
+        id => LocalServer,
+        uid => UID,
+        cluster_name => ClusterName,
+        initial_members => Servers,
+        machine => {module, ?MODULE, #{db => DB, shard => Shard}},
+        log_init_args => #{uid => UID}
+    }),
+    case Servers of
+        [LocalServer | _] ->
```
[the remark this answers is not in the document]
keynslug: Yes. Apparently ra servers do not start elections automatically, and we need to poke them, thus this quick attempt to trigger it only once. Needs some testing to verify that it works well enough.

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
keynslug: My line of thinking is that `State` is essentially a product of applying Raft log entries, and is periodically snapshotted and dumped to disk. This tuple feels odd as part of state / snapshot, it's more like a runtime-only information needed to locate relevant Erlang processes. On the other hand I agree that it feels hacky to put it in the process dictionary.

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
```diff
@@ -344,3 +339,167 @@ shard_of_messages(DB, [#message{from = From} | _]) ->
...
+) ->
+    %% NOTE
+    %% Unique timestamp tracking real time closely.
+    %% With microsecond granularity it should be nearly impossible for it to run
+    %% too far ahead than the real time clock.
+    Timestamp = max(Latest + 1, TimestampLocal),
+    Messages = assign_timestamps(Timestamp, MessagesIn),
+    Result = emqx_ds_storage_layer:store_batch(erlang:get(emqx_ds_db_shard), Messages, Options),
+    %% NOTE: Last assigned timestamp.
+    NLatest = Timestamp + length(Messages) - 1,
```
savonarola, line 495: Nit: m.b. return from `NLatest` from `assign_timestamps`

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -94,9 +94,11 @@
 -type next_result() :: next_result(iterator()).
 
 %% Timestamp
+%% Each message must have unique timestamp.
```
ieQu1, line 97: Note: storage lts introduces a "unique integer" https://github.com/emqx/emqx/blob/5bc67cb288f6953e73654ce1784b22d98b2639b4/apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl#L446
to keep the original timestamp.

I think creation of this number can be moved into the replication layer, and made deterministic (smth. like raft batch seqno + message index in the batch). This way we'll avoid storing two unique IDs.

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
ieQu1: I wouldn't object to using process dictionary too much, as long as the terms are static. 
Process dictionary becomes rather sticky when it's used to store temporary data, because it doesn't get cleaned up on exception.

apps/emqx_durable_storage/src/emqx_ds.erl
keynslug: > this PR shifts the responsibility of making the timestamps unique to the application

To the replication layer actually.

> We shouldn't move the responsibility of making message timestamps unique to the API consumer

Agree. This PR doesn't move the responsibility though. The replication layer modifies message timestamps to ensure uniqueness. API consumer will definitely be surprised though, but it's a PoC-grade hack at this point.

> we can order the messages according to the leader's clock

They are already ordered according to the leader's clock (to be more precise, shard's consensus group tracks strictly monotonically increasing clock).

apps/emqx_utils/src/emqx_utils_stream.erl
```diff
@@ -91,6 +93,31 @@ map(F, S) ->
         end
     end.
 
+%% @doc Zip a list of streams into a stream producing lists of their respective values.
+%% The resulting stream is as long as the shortest of the input streams.
+-spec zip([stream(X)]) -> stream([X]).
```
thalesmg, line 98: Nit: with this name, I'd expect the typespec to be something like `zip([stream(_)]) -> stream(tuple())`, similar to `lists:zip`, `Enum.zip` and others.  But I see how this has the same "shape" as a stream of tuples.  :thinking: 

Also, this function seems to be unused in this PR?

keynslug: I should probably rename it. Agree that it brokes expectations of what _zip_s are, however I found that having it emit rows as lists is a bit more usable overall, especially given it's generic in the number of things being zipped (e.g. one can do `zipwith/2` through `erlang:apply/2`).

> Also, this function seems to be unused in this PR?

Yeah it was used in the previous iteration. Do you think it's better to drop it?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -308,20 +257,60 @@ terminate(_Reason, #s{}) ->
...
+            mnesia:abort({insufficient_sites_online, NSites, AllSites})
+    end,
+    case mnesia:match_object(?SHARD_TAB, ?SHARD_PAT({DB, '_'}), write) of
+        [] ->
+            ok;
+        Records ->
+            ShardsAllocated = [Shard || #?SHARD_TAB{shard = {_DB, Shard}} <- Records],
+            mnesia:abort({shards_already_allocated, ShardsAllocated})
+    end,
+    {Allocation, _} = lists:mapfoldl(
```
ieQu1, line 292: This seems to replace Rendezvous Hashing shard allocation with an algorithm that simply cycles through the available replicas (?).
Any reason to rollback to that?

keynslug: I described it a bit in the commit message. It was allocating shards too unfairly, and that was hurting the quality and reproducibility of performance testing a lot. Given usual realistic cluster and shard set sizes it will also help to allocate fairly in production clusters. I'd also argue that having a degree of predictability here also helps a bit in operation and maintenance.

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
```diff
@@ -500,3 +471,244 @@ do_get_delete_streams_v4(DB, Shard, TopicFilter, StartTime) ->
 
 list_nodes() ->
     mria:running_nodes().
+
+%% TODO
+%% Too large for normal operation, need better backpressure mechanism.
+-define(RA_TIMEOUT, 60 * 1000).
+
+-define(SAFERPC(SERVER, EXPR), make_safe_rpc(SERVER, fun() -> EXPR end)).
```
ieQu1, line 479: This is not the safest RPC, since it creates a closure object. Upgrading code will break it. It's safer to use something like `{fun my_fun/1, ArgsList}` or an mfa.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_shard.erl
```diff
@@ -0,0 +1,205 @@
...
+    terminate/2
+]).
+
+-define(PTERM(DB, SHARD, L), {?MODULE, DB, SHARD, L}).
+-define(MEMOIZE(DB, SHARD, EXPR),
+    case persistent_term:get(__X_Key = ?PTERM(DB, SHARD, ?LINE), undefined) of
+        undefined ->
+            ok = persistent_term:put(__X_Key, __X_Value = (EXPR)),
+            __X_Value;
+        __X_Value ->
```
ieQu1, line 47: Dangerous macro. If someone invokes it several times in the same function, they will get a badmatch. I would prefer it to be rewritten as a function that takes an argument possibly created with the help of a macro.

apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl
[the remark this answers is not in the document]
thalesmg: I'm thinking about if that could also be a problem, now from the opposite side.  :sweat_smile: 

When [moving session data from mnesia to DS](https://github.com/emqx/emqx/pull/12675), I took advantage of the fact that the message timestamps are respected by the storage layer to upsert keys.  Now that `emqx_ds_replication_layer:assign_timestamps` enforces unique timestamps, it'll no longer be possible to upsert a key if one wants to do that in a particular DB.  :thinking:

apps/emqx_durable_storage/src/emqx_ds_builtin_db_sup.erl
```diff
@@ -86,45 +92,78 @@ ensure_shard(Shard) ->
...
+    },
+    Opts = emqx_ds_replication_layer_meta:get_options(DB),
+    Children = [
+        shard_storage_spec(DB, Shard, Opts),
+        shard_replication_spec(DB, Shard, Opts)
+    ],
     {ok, {SupFlags, Children}}.
 
+start_ra_system(DB, #{replication_options := ReplicationOpts}) ->
+    DataDir = filename:join([emqx:data_dir(), DB, dsrepl]),
```
thalesmg, line 140:
```suggestion
    DataDir = filename:join([emqx_ds:base_dir(), DB, dsrepl]),
```

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -530,51 +532,64 @@ open_shard(ShardId, DB, CFRefs, ShardSchema) ->
...
+        {error, Reason} ->
+            {error, Reason}
+    end.
+
+-spec handle_update_config(server_state(), emqx_ds:time(), emqx_ds:create_db_opts()) ->
+    server_state().
+handle_update_config(S0 = #s{schema = Schema}, Since, Options) ->
+    Prototype = maps:get(storage, Options),
+    S = S0#s{schema = Schema#{prototype := Prototype}},
+    handle_add_generation(S, Since).
```
thalesmg, line 592: Now this doesn't always return the server state: it may return `{error, Reason}`.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -472,16 +482,12 @@ init({ShardId, Options}) ->
...
-    S = add_generation(S1, Since),
+handle_call(#call_update_config{since = Since, options = Options}, _From, S0) ->
+    S = #s{} = handle_update_config(S0, Since, Options),
     commit_metadata(S),
     {reply, ok, S};
-handle_call(add_generation, _From, S0) ->
-    Since = emqx_message:timestamp_now(),
-    S = add_generation(S0, Since),
+handle_call(#call_add_generation{since = Since}, _From, S0) ->
+    S = #s{} = handle_add_generation(S0, Since),
```
thalesmg, line 490: May need to check for `{error, _}` here.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -472,16 +482,12 @@ init({ShardId, Options}) ->
     commit_metadata(S),
     {ok, S}.
 
-handle_call({update_config, Options}, _From, #s{schema = Schema} = S0) ->
-    Prototype = maps:get(storage, Options),
-    S1 = S0#s{schema = Schema#{prototype := Prototype}},
-    Since = emqx_message:timestamp_now(),
-    S = add_generation(S1, Since),
+handle_call(#call_update_config{since = Since, options = Options}, _From, S0) ->
+    S = #s{} = handle_update_config(S0, Since, Options),
```
thalesmg, line 486: May need to check for `{error, _}` here.

apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl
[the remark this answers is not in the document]
keynslug: > Now that `emqx_ds_replication_layer:assign_timestamps` enforces unique timestamps, it'll no longer be possible

Conceptually, this needs to be enforced only for messages that needs to be replayed in a strict order. If we need to accommodate for another storage semantics, nothing should stop us from expanding API here, e.g. adding flags or properties to _store batch_ command.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_shard.erl
```diff
@@ -0,0 +1,207 @@
...
+    handle_cast/2,
+    terminate/2
+]).
+
+%%
+
+start_link(DB, Shard, Opts) ->
+    gen_server:start_link(?MODULE, {DB, Shard, Opts}, []).
+
+shard_servers(DB, Shard) ->
```
ieQu1, line 46: Nit (can be done in a follow-up): it would be nice to have type specs for all the exported functions.

ieQu1 (dismissed):

ieQu1 (approved):

ieQu1 (approved):

emqxqa:
[Test Execution]()

PR #12380, opened 2024-01-24
fix: EMQX_DATA_DIR not set

zhongwencool:
Fixes <issue-or-jira-number>
```
2024-01-24T11:22:58.000646+08:00 [warning] env: {EMQX_DATA_DIR}, msg: cannot_resolve_env_variable, original: ${EMQX_DATA_DIR}
```
introduce here: https://github.com/emqx/emqx/commit/57074015c6f1a1983193ffe382e7dff29386427b#diff-7cd7fcf74d91ac97a19ed681624f639922e0426dcbae0032d0070248d8199000

1. The `data_dir` currently comes from only one place, which is `node.data_dir`. 
If it's not configured, the EMQX won't start.
2. We don't currently need to introduce a new variable(ENV) to add more complexity.
3. `DATA_DIR` is sourced from node.data_dir and can be used.
```
%% emqx line 558
DATA_DIR="$(get_boot_config 'node.data_dir')"
```

PS: I don't want to rename `DATA_DIR` to `EMQX_DATA_DIR` either, because configuration items starting with `EMQX_ `, and we don't provide EMQX_DATA_DIR config.

Release version: v/e5.6.0

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -1896,7 +1896,7 @@ fields("session_storage_backend_builtin") ->
                 string(),
                 #{
                     desc => ?DESC(session_builtin_data_dir),
-                    default => <<"${EMQX_DATA_DIR}">>,
+                    default => <<"${DATA_DIR}">>,
                     importance => ?IMPORTANCE_LOW,
                     converter => fun(Path, Opts) ->
                         naive_env_interpolation(ensure_unicode_path(Path, Opts))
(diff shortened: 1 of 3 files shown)


zmstone:
I think we should use `EMQX_DATA_DIR` everywhere, not the other way around.

thalesmg:
> I think we should use `EMQX_DATA_DIR` everywhere, not the other way around.

Yes, I think that would make env var naming more consistent, and it won't affect `node.data_dir` as well, as it's a templated variable from `vm.args.cloud` defined at build time.

zhongwencool:
I would prefer not to confuse users by using `EMQX_DATA_DIR`. If the team believes it is not a significant concern, then I kindly request that you take responsibility for addressing it. 
I have decided to close this pull request. @thalesmg

Issue #12344, opened 2024-01-17
Excessive memory spike after a MQTT client publishes a large payload (e.g. ~100mb), causing broker crash


bernard-bear:
### What happened?

We are attempting to use the EMQX broker to send MQTT messages containing large payloads. To test this, we published a large message (~100mb) to the broker via a MQTT client (MQTT.js). However, upon publishing the message, memory usage of the node that the client is connected to increases significantly over a short duration, from about 200mb to 1.8 GB within 20 seconds (and presumably increases even more). Subsequently, the node crashes as it is out of memory. This is likely triggered by Google Kubernetes Engine.

See this video for the memory spike immediately after the large payload was published by the client:
[EMQX broker crash memory usage.webm](https://github.com/emqx/emqx/assets/138452214/23710ab4-c9d9-43fd-b785-058bcc142340)

We've tested on various payload sizes (e.g. 50mb, 60mb, 70mb, 80mb, 90mb), and observed similar memory spikes (only 90mb and 100mb triggers a crash, and this is due to the memory limit of 2GB). On average, it seems that the memory spike is about 20 times of the payload (i.e. if the payload is 50mb, memory usage increases by around 1000mb)

The behaviour described above happens even if the max packet size is set to a low value (e.g. 1mb), which means that the memory spike occurs even before the publish packet has been accepted by the broker.

### What did you expect to happen?

The broker should not have such excessive memory usage. The memory spike should not be 20 times but somewhere closer to 1 - 5 times.

### How can we reproduce it (as minimally and precisely as possible)?

With a MQTT client (e.g. MQTT.js), connect to the broker and publish a payload that is about 100 megabytes. The payload should be published to a topic that the client is authorised to publish to.

This code can be used to generate the payload:
```
const payload = JSON.stringify({ field: 'x'.repeat(1000 * 1000 * 100) })
```

### EMQX version

<details>
5.3.2
</details>

### OS version

<details>
1.26.6-gke.1700

Using EMQX operator to deploy
</details>

### Log files

<details>
EMQX logs from GKE:

```
[...]

PR #12392, opened 2024-01-25
perf: new ws listener option to disable UTF-8 validation

qzhuyan:
perf improvement for  #12344

Release version: v/e5.5

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -1104,6 +1104,14 @@ fields("ws_opts") ->
             sc(
                 ref("deflate_opts"),
                 #{}
+            )},
+        {"validate_utf8",
+            sc(
+                boolean(),
+                #{
+                    default => true,
+                    desc => ?DESC(fields_ws_opts_validate_utf8)
+                }
             )}
     ];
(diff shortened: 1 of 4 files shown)


apps/emqx/src/emqx_schema.erl
```diff
@@ -1103,6 +1103,14 @@ fields("ws_opts") ->
             sc(
                 ref("deflate_opts"),
                 #{}
+            )},
+        {"validate_utf8",
+            sc(
+                boolean(),
+                #{
+                    default => true,
```
zmstone, line 1111: default to `false`?

zmstone:
maybe change base to `master`.

apps/emqx/src/emqx_schema.erl
qzhuyan: the default in cowboy is `true`, and I want to keep legacy behaviour. set it to false only when needed.

qzhuyan:
note,
on a machine of 6 cores, apple m2. with reasonable socket buffer size, 
once I set it to `false`,  memory peak goes from 660MB to 470MB while testing with a 100M publish payload.

thalesmg (approved):

PR #12410, opened 2024-01-26
perf(emqx_broker): pick broker pool worker by topic/shard pair to dis…

SergeTupchiy:
Fixes EMQX-11812

https://github.com/emqx/emqx/pull/12243 solved a probable race condition but the performance dropped significantly for fan-out scenario (lots of subscribers per one topic).

This PR addresses the performance drop and apparently doesn't introduce any new (or previously discovered)  routes race conditions (may worth extra verification...)

Big thanks @zmstone  for brainstorming it.

Release version: v/e5.5

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_broker.erl
+++ b/apps/emqx/src/emqx_broker.erl
@@ -85,6 +85,16 @@
 %% Guards
 -define(IS_SUBID(Id), (is_binary(Id) orelse is_atom(Id))).
 
+-define(cast_or_eval(Pid, Msg, Expr),
+    case Pid =:= self() of
+        true ->
+            _ = Expr,
+            ok;
+        false ->
+            cast(Pid, Msg)
+    end
+).
+
 -spec start_link(atom(), pos_integer()) -> startlink_ret().
 start_link(Pool, Id) ->
     ok = create_tabs(),
@@ -159,15 +169,7 @@ do_subscribe(Topic, SubPid, SubOpts) when is_binary(Topic) ->
     %% https://emqx.atlassian.net/browse/EMQX-10214
     I = emqx_broker_helper:get_sub_shard(SubPid, Topic),
     true = ets:insert(?SUBOPTION, {{Topic, SubPid}, with_shard_idx(I, SubOpts)}),
-    %% NOTE
-    %% We are relying on the local state to minimize global routing state changes,
-    %% thus it's important that some operations on ETS tables on the same topic
-    %% should not be interleaved: `ets:member/2` + `ets:insert/2` that are part of
-    %% broker's `subscribe` codepath, and `ets:delete_object/2` that's part of
-    %% `unsubscribe` codepath. So we have to pick a worker according to the topic,
-    %% but not shard. If there are topics with high number of shards, then the
-    %% load across the pool will be unbalanced.
-    Sync = call(pick(Topic), {subscribe, Topic, SubPid, I}),
+    Sync = call(pick({Topic, I}), {subscribe, Topic, SubPid, I}),
     case Sync of
         ok ->
             ok;
@@ -218,15 +220,7 @@ do_unsubscribe2(Topic, SubPid, SubOpts) when
         0 -> emqx_exclusive_subscription:unsubscribe(Topic, SubOpts);
         _ -> ok
     end,
-    %% NOTE
-    %% We are relying on the local state to minimize global routing state changes,
-    %% thus it's important that some operations on ETS tables on the same topic
-    %% should not be interleaved: `ets:member/2` + `ets:insert/2` that are part of
-    %% broker's `subscribe` codepath, and `ets:delete_object/2` that's part of
-    %% `unsubscribe` codepath. So we have to pick a worker according to the topic,
-    %% but not shard. If there are topics with high number of shards, then the
-    %% load across the pool will be unbalanced.
-    cast(pick(Topic), {unsubscribed, Topic, SubPid, I});
+    cast(pick({Topic, I}), {unsubscribed, Topic, SubPid, I});
 do_unsubscribe2(#share{group = Group, topic = Topic}, SubPid, _SubOpts) when
     is_binary(Group), is_binary(Topic), is_pid(SubPid)
 ->
@@ -501,8 +495,8 @@ cast(Broker, Req) ->
     gen_server:cast(Broker, Req).
 
 %% Pick a broker
-pick(Topic) ->
-    gproc_pool:pick_worker(broker_pool, Topic).
+pick(TopicShard) ->
+    gproc_pool:pick_worker(broker_pool, TopicShard).
 
 %%--------------------------------------------------------------------
 %% gen_server callbacks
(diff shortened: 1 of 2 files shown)


apps/emqx/src/emqx_broker.erl
```diff
@@ -518,17 +502,27 @@ handle_call({subscribe, Topic, SubPid, 0}, {From, _Tag}, State) ->
...
 handle_call({subscribe, Topic, SubPid, I}, {From, _Tag}, State) ->
-    Existed = ets:member(?SUBSCRIBER, Topic),
-    true = ets:insert(?SUBSCRIBER, [
-        {Topic, {shard, I}},
-        {{shard, Topic, I}, SubPid}
-    ]),
+    %% This will attempt to add a route per each next shard.
+    %% The overhead must be negligible, but the consistency in general
+    %% / race conditions safety is expected to be stronger.
+    Existed = ets:member(?SUBSCRIBER, {shard, Topic, I}),
```
SergeTupchiy, line 508: Though I'm not really sure that it solves any issues at the moment, I like the idea that each pool worker is only responsible for its subset of shards and doesn't try to check `ets:member(?SUBSCRIBER, Topic)`, since another worker may modify that state in between. 
It completely eliminates any possible race conditions between several workers processing different shards of the same topic. 
The cost is that we add route per each shard (even though it's normally expected to be already added for any shard > 0). 
But this must be a really small overhead..

apps/emqx/src/emqx_broker.erl
```diff
@@ -518,17 +502,27 @@ handle_call({subscribe, Topic, SubPid, 0}, {From, _Tag}, State) ->
     Result = maybe_add_route(Existed, Topic, From),
     {reply, Result, State};
 handle_call({subscribe, Topic, SubPid, I}, {From, _Tag}, State) ->
-    Existed = ets:member(?SUBSCRIBER, Topic),
-    true = ets:insert(?SUBSCRIBER, [
-        {Topic, {shard, I}},
-        {{shard, Topic, I}, SubPid}
-    ]),
+    %% This will attempt to add a route per each next shard.
```
zmstone, line 505:
```suggestion
    %% This will attempt to add a route per each new shard.
```

apps/emqx/src/emqx_broker.erl
```diff
@@ -538,12 +532,14 @@ handle_cast({unsubscribed, Topic, SubPid, I}, State) ->
     true = ets:delete_object(?SUBSCRIBER, {{shard, Topic, I}, SubPid}),
     case ets:member(?SUBSCRIBER, {shard, Topic, I}) of
         false ->
-            ets:delete_object(?SUBSCRIBER, {Topic, {shard, I}});
+            ets:delete_object(?SUBSCRIBER, {Topic, {shard, I}}),
+            %% Do not attempt to delete any routes here,
+            %% let it be handled only by the same pool worker per topic (0 shard),
+            %% so that all route deletes are serialized.
+            cast(pick({Topic, 0}), {unsubscribed, Topic, shard, I});
```
zmstone, line 539: check if the picked pid is `self()` ?

apps/emqx/src/emqx_broker.erl
SergeTupchiy: Done.

apps/emqx/src/emqx_broker.erl
SergeTupchiy: Done, thanks.

zmstone (approved):

apps/emqx/src/emqx_broker.erl
SergeTupchiy: It's conflicting with https://github.com/emqx/emqx/blob/release-55/apps/emqx/src/emqx_router_syncer.erl#L320.
This can happen with this PR, and `ignored` reply will cause badmatch error.
For example:
Shard 0 is busy waiting for deleting a route (e.g. RPC takes much time). Meanwhile, more than max shard size new subscribers are coming.
So, some shard N subscribe call can be definitely processed while there is no route yet and about the same time when other shard 0 subscribe requests are queued for processing/being processed.

apps/emqx/src/emqx_broker.erl
```diff
@@ -518,32 +502,48 @@ handle_call({subscribe, Topic, SubPid, 0}, {From, _Tag}, State) ->
...
+            false ->
+                ets:delete_object(?SUBSCRIBER, {Topic, {shard, I}}),
+                %% Do not attempt to delete any routes here,
+                %% let it be handled only by the same pool worker per topic (0 shard),
+                %% so that all route deletes are serialized.
+                case pick({Topic, 0}) of
+                    Pid when Pid =:= self() ->
+                        maybe_delete_route(Topic);
+                    Pid1 ->
+                        cast(Pid1, {unsubscribed, Topic, shard, I})
```
zmstone, line 542:
```suggestion
                        cast(Pid1, {unsubscribed, Topic, shard, 0})
```

apps/emqx/src/emqx_broker.erl
```diff
@@ -518,32 +502,48 @@ handle_call({subscribe, Topic, SubPid, 0}, {From, _Tag}, State) ->
     Result = maybe_add_route(Existed, Topic, From),
     {reply, Result, State};
```
zmstone, line 503:
```suggestion
    ok = maybe_add_route(Existed, Topic, From),
    {reply, ok, State};
```

apps/emqx/src/emqx_broker.erl
```diff
@@ -518,32 +502,48 @@ handle_call({subscribe, Topic, SubPid, 0}, {From, _Tag}, State) ->
...
     true = ets:delete_object(?SUBSCRIBER, {{shard, Topic, I}, SubPid}),
-    case ets:member(?SUBSCRIBER, {shard, Topic, I}) of
-        false ->
-            ets:delete_object(?SUBSCRIBER, {Topic, {shard, I}});
-        true ->
-            true
-    end,
-    Exists = ets:member(?SUBSCRIBER, Topic),
-    _Result = maybe_delete_route(Exists, Topic),
+    _ =
```
zmstone, line 531:
```suggestion
```

apps/emqx/src/emqx_broker.erl
```diff
@@ -518,32 +502,48 @@ handle_call({subscribe, Topic, SubPid, 0}, {From, _Tag}, State) ->
...
+        case ets:member(?SUBSCRIBER, {shard, Topic, I}) of
+            false ->
+                ets:delete_object(?SUBSCRIBER, {Topic, {shard, I}}),
+                %% Do not attempt to delete any routes here,
+                %% let it be handled only by the same pool worker per topic (0 shard),
+                %% so that all route deletes are serialized.
+                case pick({Topic, 0}) of
+                    Pid when Pid =:= self() ->
+                        maybe_delete_route(Topic);
```
zmstone, line 540:
```suggestion
                        _ = maybe_delete_route(Topic),
                        ok;
```

apps/emqx/src/emqx_broker.erl
```diff
@@ -518,32 +502,48 @@ handle_call({subscribe, Topic, SubPid, 0}, {From, _Tag}, State) ->
...
+        end,
+    true = ets:insert(?SUBSCRIBER, Recs1),
     Result = maybe_add_route(Existed, Topic, From),
     {reply, Result, State};
 handle_call(Req, _From, State) ->
     ?SLOG(error, #{msg => "unexpected_call", call => Req}),
     {reply, ignored, State}.
 
+handle_cast({unsubscribed, Topic, shard, _I}, State) ->
+    _Result = maybe_delete_route(Topic),
```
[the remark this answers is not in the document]
SergeTupchiy: Note: `{t/#, 42}` can't send {delete, t/#}, only `{t/#, 0}` is allowed to do that. 

Let's imagine, `{t/#, 0}` broker sees that there are absolutely no subscribers (ets:member(?SUBSCRIBER, t/#) =:= true`) so it sends `{delete, t/#}` to route syncer. This indeed can be arrived in unspecified order with `{add, t/#}` from {t/#, 43}. 
But, if `{t/#, 43}` manages to send `{add, t/#}`, then there are must be some more subscribers for shard 0 that must be processed by `{t/#, 0}` broker.
 This is guareded by the atomic ETS sequence: if shard 43 is trying to add routes, than there are 42 shards behind. 
So, the final batch may look like:
```
{add, t/#},        % from {t/#, 43}
{delete, t/#}     % from {t,#, 0}
{add, t/#},        % from {t/#, 0} (can be several) 
```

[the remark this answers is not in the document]
SergeTupchiy: Well, it's probably still possible, though quite hard to imagine :sweat:
* shard `{t/#, 0}` deletes a route
* shard `{t/#, 43}` adds a route
These msgs  are re-ordered: `add, delete`. 
shard `{t/#, 0}` tries to add the route again (because if the sequence reached shard `{t/#, 43}`, than 0 shard must be full. 
However, none of the `{t/#, 0}` shard subscribers wins the race with `{t/#, 43}`, i,e, {t/#, 43} inserted a subscriber to ETS tab, and  `ets:member(?SUBSCRIBER, Topic)` returned `true` for all `{t/#, 0}` subscribers..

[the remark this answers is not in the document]
keynslug: > Note: `{t/#, 42}` can't send `{delete, t/#}`, only `{t/#, 0}` is allowed to do that.

Oops, sorry for the typo, I meant exactly that. 🫠

> shard `{t/#, 0}` tries to add the route again (because if the sequence reached shard `{t/#, 43}`, than 0 shard must be full.

Right. Thanks for the detailed explanation! I was under impression that counter updates are not atomic, and can stray away from the actual number of subscribers, assuming clients can be killed without notice. And it looks like it can overestimate that number AFAICT (though would be cool if you'll confirm or deny that).

But even with that in mind, I completely agree it's a very hard to reach scenario. Unsure if there are other scenarios with more processes / different reorderings with easier to reach inconsistencies.

> Perhaps we need to eliminate adding routes in shards other than 0 and cast a message to `pick({Topic, 0})` (similarly to how it worked before).

Wouldn't it defeat the purpose of this optimization? Or not too much?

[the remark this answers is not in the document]
SergeTupchiy: > Right. Thanks for the detailed explanation! I was under impression that counter updates are not atomic, and can stray away from the actual number of subscribers, assuming clients can be killed without notice. And it looks like it can overestimate that number AFAICT (though would be cool if you'll confirm or deny that).

You are right, clients can die fast and the sequence is reclaimed only eventually, but when clients die routes are not removed automatically. So, the sequence can indeed reflect a stale count of alive subscribers as well as doesn't immediately start allocating new subscribers to shard `{t/#, 0}`. 
But it doesn't seem dangerous for routes management :thinking:  If a number of subscribes die and new ones come, there is indeed no guarantee that at least one will go to shard `{t/#, 0}`, as sequence update is not immediate. 
However, when this happens once any new subscriber inserts itself into ?SUBSCRIBER and prevents the route from being removed. 

>Wouldn't it defeat the purpose of this optimization? Or not too much?

I think it shouldn't at all, we will cast one message per shard, something like:
```
handle_call({subscribe, Topic, SubPid, I}, {From, _Tag}, State) ->
    Existed = ets:member(?SUBSCRIBER, {shard, Topic, I}),
    Recs = [{{shard, Topic, I}, SubPid}],
    Recs1 =
        case Existed of
            false -> 
 +               cast(pick({Topic, 0}), {subscribe, Topic}),
                [{Topic, {shard, I}} | Recs];
            true -> 
                Recs
        end,
 -   true = ets:insert(?SUBSCRIBER, Recs1),
 -   Result = maybe_add_route(Existed, Topic, From),
    {reply, Result, State};
```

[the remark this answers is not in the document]
keynslug: > clients can die fast and the sequence is reclaimed only eventually

True. I was also thinking about the fact that `nextval` is called before inserting record into `?SUPOPTION` table, meaning that a client can bump the sequence and then die right away, so the sequence is never reclaimed. Extremely hard to achieve, although imaginable in a long-running system.

> I think it shouldn't at all, we will cast one message per shard, something like:

Nice! 💯 Thanks for the clarification, that should indeed exclude the possibility of any race conditions. This will probably need some polishing to preserve "call" semantics (i.e. reply when the route is propagated to the global table), but it seems easy to do.

apps/emqx/src/emqx_broker.erl
```diff
@@ -159,15 +159,7 @@ do_subscribe(Topic, SubPid, SubOpts) when is_binary(Topic) ->
     %% https://emqx.atlassian.net/browse/EMQX-10214
     I = emqx_broker_helper:get_sub_shard(SubPid, Topic),
```
keynslug, line 160: Another idea, seemingly close to what @zmstone proposed: bump this counter in a pool worker process (together with inserting `?SUBSCRIBER` / `?SUBOPTIONS`) to ensure consistency, and rely on its value to make decisions on routing changes.

Pros seem to be less lookups, and load is slightly less skewed towards 0-th shard (though this one is almost negligible). Cons are that (supposedly?) insert to `?SUBOPTION` has to be performed in the pool worker process, and potentially unfair client-to-shard allocation, since client process can only read this value to pick a shard but not bump it. Curious what do you think about it?

zmstone (approved):

apps/emqx/src/emqx_broker.erl
SergeTupchiy: Fixed

apps/emqx/src/emqx_broker.erl
SergeTupchiy: This is interesting, I also tend to to think that during abrupt load spikes it may lead to slightly worse sharding.. 
At the same time, it may be quite negligible.

keynslug: Disregard that: it's still possible to have operations reordered, because there would be different processes acting on the same routes.

keynslug (approved):
🎉

PR #12450, opened 2024-02-01
ci: use v3 artifacts actions in build packages

id:
We need to use v3 actions family in `build_packages` because v4 is on NodeJS 20, and NodeJS 20 is not supported on older platforms.
There was a messed up change when `actions/upload-artifacts` in `macos` job used v4 family, `linux` job v3 family, and `actions/download-artifacts` was v4 family. Since v4 and v3 are incompatible (i.e., v4 cannot download v3 artifcats, and vice versa), when we published latest release, it only got macos artifacts.

### DIFF
--- a/.github/workflows/build_packages.yaml
+++ b/.github/workflows/build_packages.yaml
@@ -95,13 +95,12 @@ jobs:
         apple_developer_identity: ${{ secrets.APPLE_DEVELOPER_IDENTITY }}
         apple_developer_id_bundle: ${{ secrets.APPLE_DEVELOPER_ID_BUNDLE }}
         apple_developer_id_bundle_password: ${{ secrets.APPLE_DEVELOPER_ID_BUNDLE_PASSWORD }}
-    - uses: actions/upload-artifact@c7d193f32edcb7bfad88892161225aeda64e9392 # v4.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/build_packages.yaml
```diff
@@ -246,5 +247,5 @@ jobs:
             echo "unknown profile $PROFILE"
             exit 1
         fi
-        aws s3 cp --recursive packages/$PROFILE s3://${{ secrets.AWS_S3_BUCKET }}/$s3dir/${{ github.ref_name }}
-        aws cloudfront create-invalidation --distribution-id ${{ secrets.AWS_CLOUDFRONT_ID }} --paths "/$s3dir/${{ github.ref_name }}/*"
+        aws s3 cp --recursive packages/$PROFILE s3://$AWS_S3_BUCKET/$s3dir/$REF_NAME
```
thalesmg, line 250:
```suggestion
        aws s3 cp --recursive "packages/$PROFILE" "s3://$AWS_S3_BUCKET/$s3dir/$REF_NAME"
```

thalesmg (approved):

.github/workflows/build_packages.yaml
id: I'll take care of this next time.

PR #12453, opened 2024-02-01
feat(ds): add cache

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-10943

Release version: v/e5.6

## PR Checklist

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds_builtin_db_sup.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_builtin_db_sup.erl
@@ -23,6 +23,13 @@
 %% API:
 -export([start_db/2, start_shard/1, start_egress/1, stop_shard/1, ensure_shard/1]).
 
+%% Cache Worker API:
+-export([
+    ensure_cache_worker_started/4,
+    ensure_cache_worker_stopped/2,
+    get_cache_workers/1
+]).
+
 %% behaviour callbacks:
 -export([init/1]).
 
@@ -38,10 +45,14 @@
 -define(db_sup, ?MODULE).
 -define(shard_sup, emqx_ds_builtin_db_shard_sup).
 -define(egress_sup, emqx_ds_builtin_db_egress_sup).
+-define(cache_sup, emqx_ds_cache_sup).
+-define(cache_worker_sup, emqx_ds_cache_worker_sup).
 
 -record(?db_sup, {db}).
 -record(?shard_sup, {db}).
 -record(?egress_sup, {db}).
+-record(?cache_sup, {db}).
+-record(?cache_worker_sup, {db}).
 
 %%================================================================================
 %% API funcions
@@ -79,6 +90,40 @@ ensure_shard(Shard) ->
             {error, Reason}
     end.
 
+-spec ensure_cache_worker_started(
+    emqx_ds:db(),
+    emqx_ds:ds_specific_stream(),
+    emqx_ds:topic_filter(),
+    emqx_ds:time()
+) ->
+    ok | {error, _Reason}.
+ensure_cache_worker_started(DB, Stream, TopicFilter, StartTime) ->
+    case start_cache_worker(DB, Stream, TopicFilter, StartTime) of
+        {ok, _Pid} ->
+            ok;
+        {error, {already_started, _Pid}} ->
+            ok;
+        {error, Reason} ->
+            {error, Reason}
+    end.
+
+-spec ensure_cache_worker_stopped(
+    emqx_ds:db(),
+    emqx_ds:ds_specific_stream()
+) ->
+    ok | {error, _Reason}.
+ensure_cache_worker_stopped(DB, Stream) ->
+    case stop_cache_worker(DB, Stream) of
+        ok ->
+            ok;
+        {error, not_found} ->
+            ok
+    end.
+
+%% For debugging/testing
+get_cache_workers(DB) ->
+    supervisor:which_children(?via(#?cache_worker_sup{db = DB})).
+
 %%================================================================================
 %% behaviour callbacks
 %%================================================================================
@@ -95,7 +140,15 @@ init({#?db_sup{db = DB}, DefaultOpts}) ->
         end,
         MyShards
     ),
-    Children = [sup_spec(#?shard_sup{db = DB}, []), sup_spec(#?egress_sup{db = DB}, [])],
+    Children =
+        [
+            sup_spec(#?shard_sup{db = DB}, []),
+            sup_spec(#?egress_sup{db = DB}, [])
+        ] ++
+            [
+                sup_spec(#?cache_sup{db = DB}, DefaultOpts)
+             || emqx_ds_cache_coordinator:is_cache_enabled()
+            ],
     SupFlags = #{
         strategy => one_for_all,
         intensity => 0,
@@ -121,6 +174,30 @@ init({#?egress_sup{db = DB}, _}) ->
     SupFlags = #{
         strategy => one_for_one,
         intensity => 0,
+        period => 1
+    },
+    {ok, {SupFlags, Children}};
+init({#?cache_sup{db = DB}, DefaultOpts}) ->
+    %% Spec for the supervisor that manages the cache coordinator process and the cache
+    %% worker supervisor, which in turn manages one worker process per cached stream:
+    Children = [
+        cache_coordinator_spec(DB, DefaultOpts),
+        sup_spec(#?cache_worker_sup{db = DB}, DefaultOpts)
+    ],
+    SupFlags = #{
+        strategy => rest_for_one,
+        intensity => 5,
+        period => 1
+    },
+    {ok, {SupFlags, Children}};
+init({#?cache_worker_sup{db = _DB}, _}) ->
+    %% Spec for the supervisor that manages the cache worker processes that are
+    %% dynamically started and stopped by the cache coordinator as it discovers new and
+    %% deleted streams, respectively:
+    Children = [],
+    SupFlags = #{
+        strategy => one_for_one,
+        intensity => 1,
         period => 1
     },
     {ok, {SupFlags, Children}}.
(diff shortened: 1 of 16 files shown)


apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -763,7 +763,7 @@ enqueue_batch(IsReplay, BatchSize, Srs0, Session = #{inflight := Inflight0}, Cli
             true -> ItBegin0;
             false -> ItEnd0
         end,
-    case emqx_ds:next(?PERSISTENT_MESSAGE_DB, ItBegin, BatchSize) of
+    case emqx_ds:next(?PERSISTENT_MESSAGE_DB, ItBegin, BatchSize, #{use_cache => true}) of
```
ieQu1, line 766: Why do we need a new option for the cache? Does it change the semantics of `next`?

apps/emqx_durable_storage/src/emqx_ds_builtin_db_sup.erl
```diff
@@ -150,3 +166,21 @@ egress_spec(DB, Shard) ->
...
+cache_coordinator_spec(DB, Opts) ->
+    #{
+        id => {coordinator, DB},
+        start => {emqx_ds_cache_coordinator, start_link, [DB, Opts]},
+        shutdown => 5_000,
+        restart => permanent,
+        type => worker
+    }.
+
+cache_worker_sup_spec(DB) ->
```
ieQu1, line 179: Is it possible to use `sup_spec` function here as well for the sake of consistency?

apps/emqx_durable_storage/src/emqx_ds_cache_coordinator.erl
```diff
@@ -0,0 +1,335 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2024 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
```
ieQu1, line 3: Please add Apache license header here

apps/emqx/src/emqx_persistent_session_ds.erl
thalesmg: Semantics, no.  But we need a way to call next without going through the cache.  One example is the cache worker: when it tries to update the cache, it should go through it.

apps/emqx_durable_storage/src/emqx_ds_cache_coordinator.erl
thalesmg: Sure, I'll add it.  

As a side note, a few DS modules also don't have this header (bitmask_keymapper is one example).

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: This could be solved by having an internal API in the replication layer that always goes directly to the RPC/RocksDB, and exposing the cached version to `emqx_ds` callback module (similar to how egress workers intercept calls).

Exposing this option directly in the API only makes sense when we want to give the API consumers an option to bypass cache, but I can't come up with a situation when it would be needed.

thalesmg: If the cache is to be used by other backends, then shouldn't it be controllable from `emqx_ds`, like egress is as well?

thalesmg: Egress also has its options exposed:

https://github.com/emqx/emqx/blob/ed8660cb681a03a07b6d4f5faa59dc35a9518314/apps/emqx_durable_storage/src/emqx_ds.erl#L251-L257

thalesmg: Inverted the default behavior to use the cache by default in the replication layer.

ieQu1: Options exposed by egress actually change the semantics of the call: for example, `async => true` means the message can be lost before being persisted.

apps/emqx_durable_storage/src/emqx_ds_cache_worker.erl
```diff
@@ -0,0 +1,226 @@
...
+-record(get_table, {}).
+-record(pull, {}).
+
+%%================================================================================
+%% API
+%%================================================================================
+
+-spec start_link(emqx_ds:db(), emqx_ds:stream(), emqx_ds:topic_filter(), emqx_ds:time()) ->
+    supervisor:startchild_ret().
+start_link(DB, Stream, TopicFilter, StartTime) ->
```
ieQu1, line 66: I think in the long term the cache workers shouldn't care about topic filters, and simply cache the entire contents of the stream. This will allow to share the contents of the cache between different clients that potentially subscribe to different topics, that happen to be mapped to the same stream.

apps/emqx/src/emqx_persistent_session_ds.erl
thalesmg: Ok.  Still, if I understand correctly, the cache could be used by other backends.  If so, the way to call `next` without attempting to use the case should live in `emqx_ds`, right?

apps/emqx_durable_storage/src/emqx_ds_cache_worker.erl
```diff
@@ -0,0 +1,226 @@
...
+    gen_server:start_link(
+        ?via(#?cache_worker{db = DB, stream = Stream}),
+        ?MODULE,
+        {DB, Stream, TopicFilter, StartTime},
```
ieQu1, line 70: Why do we need to supply topic filter and start time to the cache worker? It seems inconsistent with the process ID, that only allows to have one cache worker for the DB and the stream.
If different different topic filters happen to map to the same stream, does it mean they will compete for the process registration?

apps/emqx_durable_storage/src/emqx_ds_cache_worker.erl
thalesmg: The current API to create iterators require the topic filter to be provided, so I guess it'll require new APIs to iterate over the stream?

apps/emqx_durable_storage/src/emqx_ds_cache_worker.erl
thalesmg: The current API to create iterators require the topic filter to be provided, so I don't think there's currently a way around it?

thalesmg: Also, we've discussed that the first version of the cache could use statically configured topic filters, and later make it automatically track streams based on usage or other heuristics.  So the future version might not need to specify the topic filter and use `#`.

apps/emqx_durable_storage/src/emqx_ds_cache_worker.erl
thalesmg: Or do you suggest to track `#` for the stream?

apps/emqx_durable_storage/src/emqx_ds_cache_worker.erl
thalesmg: I've just found a fundamental problem with the cache right now: it is currently just relying on seqnos to detect gaps, without checking that the message topic actually matches the iterator topic filter...  It'll require some rethinking.  :see_no_evil:

apps/emqx_durable_storage/src/emqx_ds_cache.hrl
```diff
@@ -0,0 +1,36 @@
...
+
+-record(cache_entry, {
+    key :: ?CACHE_KEY(emqx_ds:message_key()) | ?EOS_KEY,
+    seqno :: seqno(),
+    inserted_at :: timestamp(),
+    message :: emqx_types:message() | ?EOS_MARKER
+}).
+-type cache_entry() :: #cache_entry{}.
+
+-type seqno() :: non_neg_integer().
```
ieQu1, line 31: Nit: It's better to avoid defining types in the headers, since they'll end up duplicated in each module using the hrl.

apps/emqx_durable_storage/src/emqx_ds_cache.hrl
```diff
@@ -0,0 +1,36 @@
...
+-define(EMQX_DS_CACHE_HRL, true).
+
+-define(CACHE_KEY(LASTKEY), {LASTKEY}).
+-define(EOS_KEY, []).
+-define(EOS_MARKER, []).
+
+-record(cache_entry, {
+    key :: ?CACHE_KEY(emqx_ds:message_key()) | ?EOS_KEY,
+    seqno :: seqno(),
+    inserted_at :: timestamp(),
```
ieQu1, line 26: Can we use the timestamp from the key? We don't have to be very precise in the cache eviction flow.

thalesmg: From the key inside the message, you mean?  If we assume only more or less recent keys are added to cache (i.e., we don't start caching from the distant past), I guess so.

If it's the `#cache_entry.key` you mean, I think there's no way to extract it from the `emqx_ds:message_key()` in the general case without yet another callback.  :sweat_smile:

thalesmg:
### Using non-wildcard subscriptions (`t/%n`)

Left: without cache
Right: with cache
image

### Using wildcard subscriptions (`t/#`)

Left: without cache
Right: with cache
image

thalesmg:
I ran the test with 1 k subscriber and 1 k publishers (non-wildcard) .  Network usage drops a bit, but I guess it's because the received message rate also drops substantially...  CPU usage is similar in both scenarios (very high, ~ 100 %).  With the cache, it stays at ~ 100 % even at "rest"...  :see_no_evil:

Curiously, loadgen (LG) CPU usage also seemed higher with the cache.  :thinking: 

Left: without cache.
Right: with cache.

3rd image is LG.

- https://snapshots.raintank.io/dashboard/snapshot/6fTbZxfHF9xMpVoZThZtGFHkcKPzaqaA
- https://snapshots.raintank.io/dashboard/snapshot/bSeCCMF3Dcii6nBEAGI9Ls7ioa1rCx02
- https://snapshots.raintank.io/dashboard/snapshot/3RBgFb4xPDFLUzCMpEJ5wkRQHEHpCv4H

emqttb

emqx

lg

PR #12475, opened 2024-02-03
Use a more compact data structure to represent streams

ieQu1:
Fixes <issue-or-jira-number>

Release version: v/e5.?

## Summary
Previously the builtin DS database represented streams as maps. This was problematic for several reasons:

- Maps, even with integer fields, are relatively heavy when serialized to binary form
- Order comparison of maps is broken, so they can't be reliably used as rocksDB keys, so we had to hash them, which could lead to hash collisions

This PR changes the internal representation of the streams to lists, and removes the hashing. 

## PR Checklist


apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
```diff
@@ -311,22 +313,43 @@ do_drop_db_v1(DB) ->
...
     emqx_ds_storage_layer:store_batch({DB, Shard}, Messages, Options).
 
+%% Remove me in EMQX 5.6
+-dialyzer({nowarn_function, do_get_streams_v1/4}).
 -spec do_get_streams_v1(
     emqx_ds:db(), emqx_ds_replication_layer:shard_id(), emqx_ds:topic_filter(), emqx_ds:time()
+) ->
+    [{integer(), emqx_ds_storage_layer:stream_v1()}].
+do_get_streams_v1(_DB, _Shard, _TopicFilter, _StartTime) ->
+    error(obsolete_api).
```
keynslug, line 323: Is the idea that we don't care about rolling upgrade stability for the time being?

keynslug (approved):

apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl
ieQu1: Yes. I added a disclaimer that new releases can currently break durable sessions in the documentation draft.

apps/emqx_durable_storage/src/proto/emqx_ds_proto_v4.erl
```diff
@@ -0,0 +1,147 @@
...
+-export([
+    drop_db/2,
+    store_batch/5,
+    get_streams/5,
+    make_iterator/6,
+    next/5,
+    update_iterator/5,
+    add_generation/2,
+
+    %% introduced in v3
```
thalesmg, line 31: Nit: maybe only indicate the changes in this version, i.e., get_streams and make_iterator v2?

thalesmg (approved):

apps/emqx_durable_storage/src/proto/emqx_ds_proto_v4.erl
ieQu1: We'll never end up having to use the old versions, I hope.

thalesmg: I meant to make it the diff between proto versions clearer at a glance.

ieQu1: D'oh, sorry I misunderstood your comment. I will fix it in a later PR.

PR #12500, opened 2024-02-09
feat(ds): list disconnected persistent sessions in clients API

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-11540

Note that not all information provided by disconnected in-memory sessions is available to
disconnected persistent sessions, nor does all of them make sense.

Release version: v/e5.7

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds_state.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds_state.erl
@@ -36,10 +36,20 @@
 -export([get_rank/2, put_rank/3, del_rank/2, fold_ranks/3]).
 -export([get_subscriptions/1, put_subscription/4, del_subscription/3]).
 
--export([make_session_iterator/0, session_iterator_next/2]).
+-export([
+    make_session_iterator/0,
+    session_iterator_next/2,
+    session_count/0
+]).
 
 -export_type([
-    t/0, metadata/0, subscriptions/0, seqno_type/0, stream_key/0, rank_key/0, session_iterator/0
+    t/0,
+    metadata/0,
+    subscriptions/0,
+    seqno_type/0,
+    stream_key/0,
+    rank_key/0,
+    session_iterator/0
 ]).
 
 -include("emqx_mqtt.hrl").
@@ -359,17 +369,18 @@ del_rank(Key, Rec) ->
 fold_ranks(Fun, Acc, Rec) ->
     gen_fold(ranks, Fun, Acc, Rec).
 
+-spec session_count() -> non_neg_integer().
+session_count() ->
+    %% N.B.: this is potentially costly.  Should not be called in hot paths.
+    %% `mnesia:table_info(_, size)' is always zero for rocksdb, so we need to traverse...
+    do_session_count(make_session_iterator(), 0).
+
 -spec make_session_iterator() -> session_iterator().
 make_session_iterator() ->
-    case mnesia:dirty_first(?session_tab) of
-        '$end_of_table' ->
-            '$end_of_table';
-        Key ->
-            Key
-    end.
+    mnesia:dirty_first(?session_tab).
 
 -spec session_iterator_next(session_iterator(), pos_integer()) ->
-    {[{emqx_persistent_session_ds:id(), metadata()}], session_iterator()}.
+    {[{emqx_persistent_session_ds:id(), metadata()}], session_iterator() | '$end_of_table'}.
 session_iterator_next(Cursor, 0) ->
     {[], Cursor};
 session_iterator_next('$end_of_table', _N) ->
@@ -563,6 +574,18 @@ ro_transaction(Fun) ->
 %% ro_transaction(Fun) ->
 %%     {atomic, Res} = mria:ro_transaction(?DS_MRIA_SHARD, Fun),
 %%     Res.
+
+%%
+
+do_session_count('$end_of_table', N) ->
+    N;
+do_session_count(Cursor, N) ->
+    case session_iterator_next(Cursor, 1) of
+        {[], _} ->
+            N;
+        {_, NextCursor} ->
+            do_session_count(NextCursor, N + 1)
+    end.
 
 -compile({inline, check_sequence/1}).
 
--- a/apps/emqx_management/src/emqx_mgmt.erl
+++ b/apps/emqx_management/src/emqx_mgmt.erl
@@ -66,6 +66,9 @@
     do_kickout_clients/1
 ]).
 
+%% Internal exports
+-export([lookup_running_client/2]).
+
 %% Internal functions
 -export([do_call_client/2]).
 
@@ -314,10 +317,16 @@ nodes_info_count(PropList) ->
 %%--------------------------------------------------------------------
 
 lookup_client({clientid, ClientId}, FormatFun) ->
-    lists:append([
-        lookup_client(Node, {clientid, ClientId}, FormatFun)
-     || Node <- emqx:running_nodes()
-    ]);
+    IsPersistenceEnabled = emqx_persistent_message:is_persistence_enabled(),
+    case lookup_running_client(ClientId, FormatFun) of
+        [] when IsPersistenceEnabled ->
+            case emqx_persistent_session_ds_state:print_session(ClientId) of
+                undefined -> [];
+                Session -> [maybe_format(FormatFun, {ClientId, Session})]
+            end;
+        Res ->
+            Res
+    end;
 lookup_client({username, Username}, FormatFun) ->
     lists:append([
         lookup_client(Node, {username, Username}, FormatFun)
@@ -632,6 +641,16 @@ create_banned(Banned) ->
 
 delete_banned(Who) ->
     emqx_banned:delete(Who).
+
+%%--------------------------------------------------------------------
+%% Internal exports
+%%--------------------------------------------------------------------
+
+lookup_running_client(ClientId, FormatFun) ->
+    lists:append([
+        lookup_client(Node, {clientid, ClientId}, FormatFun)
+     || Node <- emqx:running_nodes()
+    ]).
 
 %%--------------------------------------------------------------------
 %% Internal Functions.
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -698,26 +698,13 @@ list_clients(QString) ->
         case maps:get(<<"node">>, QString, undefined) of
             undefined ->
                 Options = #{fast_total_counting => true},
-                emqx_mgmt_api:cluster_query(
-                    ?CHAN_INFO_TAB,
-                    QString,
-                    ?CLIENT_QSCHEMA,
-                    fun ?MODULE:qs2ms/2,
-                    fun ?MODULE:format_channel_info/2,
-                    Options
-                );
+                list_clients_cluster_query(QString, Options);
             Node0 ->
                 case emqx_utils:safe_to_existing_atom(Node0) of
                     {ok, Node1} ->
-                        QStringWithoutNode = maps:without([<<"node">>], QString),
-                        emqx_mgmt_api:node_query(
-                            Node1,
-                            ?CHAN_INFO_TAB,
-                            QStringWithoutNode,
-                            ?CLIENT_QSCHEMA,
-                            fun ?MODULE:qs2ms/2,
-                            fun ?MODULE:format_channel_info/2
-                        );
+                        QStringWithoutNode = maps:remove(<<"node">>, QString),
+                        Options = #{},
+                        list_clients_node_query(Node1, QStringWithoutNode, Options);
                     {error, _} ->
                         {error, Node0, {badrpc, <<"invalid node">>}}
                 end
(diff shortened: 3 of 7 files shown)


apps/emqx_management/src/emqx_mgmt.erl
```diff
@@ -633,6 +642,16 @@ create_banned(Banned) ->
 delete_banned(Who) ->
     emqx_banned:delete(Who).
 
+%%--------------------------------------------------------------------
+%% Internal exports
+%%--------------------------------------------------------------------
+
+lookup_clientid_ets(ClientId, FormatFun) ->
```
ieQu1, line 649: Rename to `lookup_running_client` ?

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -30,6 +31,34 @@ init_per_suite(Config) ->
 end_per_suite(_) ->
     emqx_mgmt_api_test_util:end_suite().
 
+init_per_testcase(t_persistent_sessions = TestCase, Config) ->
+    AppSpecs = [
+        emqx_durable_storage,
```
keynslug, line 36: Nit. Should not be needed to specify that explicitly, since it's a direct dependency of `emqx`.

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -171,6 +200,257 @@ t_clients(_) ->
...
+            ?retry(
+                100,
+                20,
+                ?assertMatch(
+                    {ok, {{_, 200, _}, _, #{<<"data">> := []}}},
+                    list_request(APIPort)
+                )
+            ),
+
+            %% Scenario 2
```
keynslug, line 290: Nit. Given these are different test scenarios, it feels odd to squeeze them into a single testcase. Having 5 separate testcases should make possible future regressions more isolated and easier to analyze.

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -171,6 +200,257 @@ t_clients(_) ->
...
+            ?assertMatch(
+                {ok,
+                    {{_, 200, _}, _, #{
+                        <<"data">> := [_, _, _],
+                        <<"meta">> := #{
+                            <<"count">> := 4,
+                            <<"hasnext">> := true
+                        }
+                    }}},
+                P1
```
keynslug, line 426: Q: Why do we match `P1` here, and not `P3`? Similar question about `P2` below.

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -171,6 +200,257 @@ t_clients(_) ->
...
+            ),
+
+            %% Scenario 3
+            %% 1) Client connects and is listed as connected.
+            %% 2) Client connects to *another node* and takes over, listed only once.
+            ?tp(notice, "scenario 3", #{}),
+            ClientId3 = <<"c3">>,
+            _C5 = Connect(#{port => Port1, clientid => ClientId3}),
+            AssertSingleClient(N1, ClientId3, connected),
+            process_flag(trap_exit, true),
```
keynslug, line 318: Nit. I vaguely remember CT runner having problems with stuck and/or failing testcases that set `trap_exit` to `true`.

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -983,7 +1113,26 @@ format_channel_info(WhichNode, {_, ClientInfo0, ClientStats}) ->
             maps:without(RemoveList, ClientInfoMap),
             TimesKeys
         )
-    ).
+    );
+format_channel_info(undefined, {ClientId, PSInfo0 = #{}}) ->
```
keynslug, line 1117: Nit. Turning this into a separate `format_persistent_session_info(ClientId, PSInfo)` function could make things a bit clearer.

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -851,6 +838,145 @@ do_unsubscribe(ClientID, Topic) ->
...
+            end
+    end.
+
+init_persistent_session_iterator() ->
+    emqx_persistent_session_ds_state:make_session_iterator().
+
+no_persistent_sessions() ->
+    case emqx_persistent_message:is_persistence_enabled() of
+        true ->
+            emqx_persistent_session_ds_state:is_end_of_table(init_persistent_session_iterator());
```
keynslug, line 935: Q: Why use separate function here while there's a simple comparison with `$end_of_table` down below?

keynslug (commented):
👍🏼

apps/emqx_management/src/emqx_mgmt_api_clients.erl
thalesmg: Good question.  In my first attempt, it was just like that.  But dialyzer complained about such a comparison here (because the iterator is opaque), so I had to introduce this function...

Curiously enough, it's not complaining about the use case in the function guard, in this same file.  :thinking:

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
thalesmg: D'oh, bad copy and paste...  :see_no_evil:

apps/emqx_management/src/emqx_mgmt_api_clients.erl
thalesmg: Maybe we could make that typespec:

```erlang
-opaque session_iterator_inner() :: emqx_persistent_session_ds:id().
-type session_iterator() ::  session_iterator_inner() | '$end_of_table'.
```

keynslug: This spec feels a bit odd to me: hard to make sense of a union of opaque type and non-opaque type. I think it's better either to use `is_end_of_table/1` everywhere, to turn it into a regular type, or to change iteration function's contract into something like `next(iterator()) -> {_Items, {more, iterator()} | fin}`.

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -171,6 +215,290 @@ t_clients(_) ->
...
+            ?retry(
+                100,
+                20,
+                ?assertMatch(
+                    {ok, {{_, 200, _}, _, #{<<"data">> := []}}},
+                    list_request(APIPort)
+                )
+            ),
+
+            ok
```
ieQu1, line 291: Nit: not needed?

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -171,6 +215,290 @@ t_clients(_) ->
...
+                20,
+                ?assertMatch(
+                    {ok, {{_, 200, _}, _, #{<<"data">> := []}}},
+                    list_request(APIPort, "node=" ++ atom_to_list(N1))
+                )
+            ),
+            ok = emqtt:stop(C2),
+            ok = erpc:call(N1, emqx_persistent_session_ds, destroy_session, [ClientId]),
+
+            ok
```
ieQu1, line 330: Nit: not needed?

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -171,6 +215,290 @@ t_clients(_) ->
...
```
ieQu1, line 373: Same here?

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -171,6 +215,290 @@ t_clients(_) ->
...
+        end,
+        []
+    ),
+    ok.
+
+t_persistent_sessions5(Config) ->
+    [N1, N2] = ?config(nodes, Config),
+    APIPort = 18084,
```
ieQu1, line 381: Why is it using non-standard port?

thalesmg: Because the suite starts the management on the default port for the other tests.

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -442,3 +770,111 @@ time_string_to_epoch(DateTime, Unit) when is_binary(DateTime) ->
...
+            Error
+    end.
+
+maybe_json_decode(X) ->
+    case emqx_utils_json:safe_decode(X, [return_maps]) of
+        {ok, Decoded} -> Decoded;
+        {error, _} -> X
+    end.
+
+list_request(Port) ->
```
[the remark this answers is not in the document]
thalesmg: Could be, I think it's a matter of style/choice.

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
thalesmg: Those `ok`s are just for ease of adding new lines without adding commas to the diff.  e.g.: `?assert(...).` would become `?assert(...),` just to add a newline.

I think those `ok`s are innocuous, hence I prefer to leave them there.

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -18,18 +18,64 @@
...
+        t_persistent_sessions1,
+        t_persistent_sessions2,
+        t_persistent_sessions3,
+        t_persistent_sessions4,
+        t_persistent_sessions5
+    ].
 
 init_per_suite(Config) ->
+    emqx_common_test_helpers:clear_screen(),
+
```
keynslug, line 45: 🫠

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -359,6 +367,12 @@ del_rank(Key, Rec) ->
...
 
+-spec session_count() -> non_neg_integer().
+session_count() ->
+    %% N.B.: this is potentially costly.  Should not be called in hot paths.
+    %% `mnesia:table_info(_, size)' is always zero for rocksdb, so we need to traverse...
+    do_session_count(make_session_iterator(), 0).
+
 -spec make_session_iterator() -> session_iterator().
 make_session_iterator() ->
     case mnesia:dirty_first(?session_tab) of
```
keynslug, line 378: Q: What this case clause is for?

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -53,7 +60,8 @@
 
 -type subscriptions() :: emqx_topic_gbt:t(_SubId, emqx_persistent_session_ds:subscription()).
 
--opaque session_iterator() :: emqx_persistent_session_ds:id() | '$end_of_table'.
+-opaque session_iterator_inner() :: emqx_persistent_session_ds:id().
+-type session_iterator() :: session_iterator_inner() | '$end_of_table'.
```
keynslug, line 64: At first glance it honestly looks essentially equivalent to `-type session_iterator() :: emqx_persistent_session_ds:id() | '$end_of_table'.`.

keynslug (dismissed):

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
thalesmg: Oops, removed.

apps/emqx/src/emqx_persistent_session_ds_state.erl
thalesmg: Good question.  I didn't introduce it, but seems like we could "eta reduce" it.  :thinking:

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
ieQu1: Not sure if we have to optimize code for "minimal diffs". I would prefer to optimize for "readability at any given moment".

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
ieQu1: And why can't we reuse the default setup? 
Also, is it possible to call the Erlang interface directly, bypassing from/to JSON conversion?

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -53,7 +59,7 @@
 
 -type subscriptions() :: emqx_topic_gbt:t(_SubId, emqx_persistent_session_ds:subscription()).
 
--opaque session_iterator() :: emqx_persistent_session_ds:id() | '$end_of_table'.
+-type session_iterator() :: emqx_persistent_session_ds:id() | '$end_of_table'.
```
ieQu1, line 62: What triggered the change from opaque to transparent? Does the new code rely on the internal structure of the iterator? Ideally it should be avoided, because I am planning to change the storage from mnesia to DS in near future.

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
thalesmg: > And why can't we reuse the default setup?

Those tests are using an isolated, clustered setup.  To avoid having to stop the running applications from the suite, it's simpler to just set another port here.

> Also, is it possible to call the Erlang interface directly, bypassing conversion from/to JSON and the HTTP?

I believe it's best to test the user-facing API in this case, since it's the object of the test.  Specially during the bridge refactoring tasks, we observed that tests that don't use such APIs tend to hinder future changes to the code.

apps/emqx/src/emqx_persistent_session_ds_state.erl
thalesmg: What triggered this was the check for no persistent sessions, and also if the iteration has reached the end of the table.

ieQu1: P.S. I think the return type of `session_interator_next` should be adjusted instead like this:

```
-spec session_iterator_next(session_iterator(), pos_integer()) ->
    {[{emqx_persistent_session_ds:id(), metadata()}], session_iterator() | '$end_of_table'}.
```

thalesmg: Good idea, I'll adjust it.

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -564,6 +580,13 @@ ro_transaction(Fun) ->
 %%     {atomic, Res} = mria:ro_transaction(?DS_MRIA_SHARD, Fun),
 %%     Res.
 
+%%
+
+do_session_count('$end_of_table', N) ->
+    N;
+do_session_count(Cursor, N) ->
+    do_session_count(mnesia:dirty_next(?session_tab, Cursor), N + 1).
```
ieQu1, line 588: Can be implemented via `session_iterator_next`?

apps/emqx/src/emqx_persistent_session_ds_state.erl
ieQu1: This check is just comparing the first batch with `[]` and batch size = 1? I think this can be done in the msgmt code to keep the interface for the CRUD module clean.

ieQu1 (approved):

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -359,17 +369,18 @@ del_rank(Key, Rec) ->
 fold_ranks(Fun, Acc, Rec) ->
     gen_fold(ranks, Fun, Acc, Rec).
 
+-spec session_count() -> non_neg_integer().
```
ieQu1, line 372: Ideally, I would prefer to move this function to mgmt. This module was designed to encapsulate session states completely devoid of the business logic, rather than a convenient place for the helper functions.

If we expose a session count function from this module, to me it means that we commit ourselves to providing session count to business logic. But since the current implementation requires a full sweep of the table, perhaps it's best to avoid it for now.

thalesmg: Good point.  I'll fix this in a follow up.

Issue #12484, opened 2024-02-06
Nodes can't communicate  in static cluster


Rotario:
### What happened?

I'm running a 2 core node cluster on fly.io
my cluster settings are:
```
  EMQX_CLUSTER__PROTO_DIST = "inet6_tcp"
  EMQX_CLUSTER__DISCOVERY_STRATEGY = "static"
  EMQX_CLUSTER__STATIC__SEEDS = "[emqx@xxxxx.vm.emqx.internal,emqx@xxxxxx.vm.emqx.internal]"
```
They're running fine most of the time:
![image](https://github.com/emqx/emqx/assets/29485059/abaa88ff-b5a6-4898-8702-6aa5f75b511b)

But when testing I've found that sometimes they get stuck and can't communicate and spit out the following logs:
```
2024-02-06T00:35:56Z app[6e82e56f2e3038] lhr [info]2024-02-06T00:35:56.990082+00:00 [error] event=connect_to_remote_server, peer=emqx@2874de3c1675d8.vm.emqx.internal, port=5369, reason=nxdomain                  
2024-02-06T00:35:56Z app[6e82e56f2e3038] lhr [info]2024-02-06T00:35:56.990243+00:00 [error] crasher: initial call: gen_rpc_client:init/1, pid: <0.7306.0>, registered_name: [], exit: {{badrpc,nxdomain},[{gen_serv
er,init_it,6,[{file,"gen_server.erl"},{line,835}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,240}]}]}, ancestors: [gen_rpc_client_sup,gen_rpc_sup,<0.2123.0>], message_queue_len: 0, messages: [], l
inks: [<0.2129.0>], dictionary: [], trap_exit: true, status: running, heap_size: 2586, stack_size: 28, reductions: 2710; neighbours:   
```

This results in the fact that if I subscribe to one node, I don't receive messages sent to the other node.
I've found a restart fixes this, but I can't deploy EMQX into production if this occurs.
What could potentially be causing this please? Is it the DNS resolution of the domain names from the cloud provider (docs here: https://fly.io/docs/networking/private-networking/)? If so I'm waiting for https://github.com/emqx/emqx/pull/12467 to get merged so I can move to AAAA discovery. This might fix the issue.

### What did you expect to happen?

I expected two nodes to communicate correctly using my cloud provider's private IP address resolution mechanisms

### How can we reproduce it (as minimally and precisely as possible)?

[...]

PR #12541, opened 2024-02-20
Fail fast if node host is not IP address but cluster strategy is `dns` with `a` or `aaaa` record type

zmstone:
Fixes https://github.com/emqx/emqx/issues/12484

Release version: v/e5.6.0

## Summary

Added a schema validation to fail fast if the host part of node name is not IP but cluster strategy is `dns` with `a` or `aaaa` record type.

Example log:
```
2024-02-20T16:07:47.733469+01:00 [error] #{reason => integrity_validation_failure,result => #{domain => "local.host",explain => "Node name must be of name@IP format for DNS cluster discovery strategy with 'a' record type."},kind => validation_error,validation_name => check_node_name_and_discovery_strategy}
```

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_schema.erl
+++ b/apps/emqx_conf/src/emqx_conf_schema.erl
@@ -42,6 +42,8 @@
 %% internal exports for `emqx_enterprise_schema' only.
 -export([ensure_unicode_path/2, convert_rotation/2, log_handler_common_confs/2]).
 
+-define(DEFAULT_NODE_NAME, <<"emqx@127.0.0.1">>).
+
 %% Static apps which merge their configs into the merged emqx.conf
 %% The list can not be made a dynamic read at run-time as it is used
 %% by nodetool to generate app.<time>.config before EMQX is started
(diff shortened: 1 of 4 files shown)


ieQu1 (approved):

PMExtra:
How can I deploy a cluster by docker swarm in 5.6.0+?

The container IP address is flexible and cannot be got before container start.

Does it mean I cannot auto discovery nodes by DNS automatically anymore?

id:
@PMExtra 

- If you use A or AAAA dns record types for auto discovery, container IP address should be registered in DNS server at the time of EMQX startup. If you cannot register IP address when container starts up, it won't work anyway.
- if that is not possible, you should use SRV DNS record type

PMExtra:
@id 

Docker Swarm can use `endpoint_mode: dnsrr` to create round-robin A records.

The problem is that I cannot get the container's IP address before it starts. The IP is assigned by the Swarm network and changes with each rescheduling, so I cannot set the EMQX_NODE__NAME variable to include it in advance.

Maybe we could support a macro in `EMQX_NODE__NAME` that gets replaced with the container’s actual IP address when EMQX starts up.

For example:
`EMQX_NODE__NAME=emqx@{ip.default}` or `EMQX_NODE__NAME=emqx@{ip.eth1}`

This macro would then be replaced with the IP address of the default route interface or the specified network interface.

id:
@PMExtra may be don't set it at all? If not set, EMQX will discover its IP address automatically on startup https://github.com/emqx/emqx/blob/master/deploy/docker/docker-entrypoint.sh#L15

PR #12544, opened 2024-02-20
0220 delete non prod code in emqx_resource.erl

zmstone:
Delete unused code.

### DIFF
--- a/apps/emqx_resource/src/proto/emqx_resource_proto_v2.erl
+++ b/apps/emqx_resource/src/proto/emqx_resource_proto_v2.erl
@@ -1,0 +1,34 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2022-2024 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


qzhuyan (dismissed):

apps/emqx_resource/src/proto/emqx_resource_proto_v2.erl
```diff
@@ -0,0 +1,34 @@
...
+%%--------------------------------------------------------------------
+
+-module(emqx_resource_proto_v2).
+
+-behaviour(emqx_bpapi).
+
+-export([
+    introduced_in/0,
+    reset_metrics/1
+]).
```
thalesmg, line 24: Q: shouldn't this have the other non-deprecated functions as well (i.e., `recreate` and `remove`)?

thalesmg (approved):

apps/emqx_resource/src/proto/emqx_resource_proto_v2.erl
zmstone: I don't think so.
The older version nodes will call the `v1` module (which is not deleted yet) during rolling upgrade.

zmstone: ah, forgot to add comments.

thalesmg (approved):

apps/emqx_resource/src/proto/emqx_resource_proto_v2.erl
zmstone: Deleted the "deprecated" comments.
We need the v1 module to stick around to pin the `_local` flavor APIs for xref and dialyzer checks.

PR #12553, opened 2024-02-21
refactor: delete zone schema roots

zmstone:
it's never really used as real root configs


ieQu1 (approved):

zmstone:
It's actually needed to generate config API schema.

PR #12562, opened 2024-02-21
Refactor schema for the durable storage

ieQu1:
Fixes https://emqx.atlassian.net/browse/EMQX-11862

Release version: v/e5.?

## Summary
- Reduce configuration nesting 
- Allow to change storage layout
- Add dynamic dispatching of the DS configuration handler 

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_ds_schema.erl
+++ b/apps/emqx/src/emqx_ds_schema.erl
@@ -1,0 +1,245 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2024 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+%% @doc Schema for EMQX_DS databases.
+-module(emqx_ds_schema).
+
+%% API:
+-export([schema/0, translate_builtin/1]).
+
+%% Behavior callbacks:
+-export([fields/1, desc/1, namespace/0]).
+
+-include("emqx_schema.hrl").
+-include_lib("typerefl/include/types.hrl").
+-include_lib("hocon/include/hoconsc.hrl").
+-include_lib("hocon/include/hocon_types.hrl").
+
+%%================================================================================
+%% Type declarations
+%%================================================================================
+
+%%================================================================================
+%% API
+%%================================================================================
+
+translate_builtin(#{
+    backend := builtin,
+    n_shards := NShards,
+    replication_factor := ReplFactor,
+    layout := Layout
+}) ->
+    Storage =
+        case Layout of
+            #{
+                type := wildcard_optimized,
+                bits_per_topic_level := BitsPerTopicLevel,
+                epoch_bits := EpochBits,
+                topic_index_bytes := TIBytes
+            } ->
+                {emqx_ds_storage_bitfield_lts, #{
+                    bits_per_topic_level => BitsPerTopicLevel,
+                    topic_index_bytes => TIBytes,
+                    epoch_bits => EpochBits
+                }};
+            #{type := reference} ->
+                {emqx_ds_storage_reference, #{}}
+        end,
+    #{
+        backend => builtin,
+        n_shards => NShards,
+        replication_factor => ReplFactor,
+        storage => Storage
+    }.
+
+%%================================================================================
+%% Behavior callbacks
+%%================================================================================
+
+namespace() ->
+    durable_storage.
+
+schema() ->
+    [
+        {"messages",
+            ds_schema(#{
+                desc => ?DESC(messages),
+                importance => ?IMPORTANCE_HIDDEN,
+                default =>
+                    #{
+                        <<"backend">> => builtin
+                    }
+            })}
+    ].
+
+fields("builtin") ->
+    %% Schema for the builtin backend:
+    [
+        {"backend",
+            sc(
+                builtin,
+                #{
+                    importance => ?IMPORTANCE_MEDIUM,
+                    'readOnly' => true,
+                    default => builtin,
+                    desc => ?DESC(builtin)
+                }
+            )},
+        {"_config_handler",
+            sc(
+                {module(), atom()},
+                #{
+                    importance => ?IMPORTANCE_HIDDEN,
+                    'readOnly' => true,
+                    default => {?MODULE, translate_builtin}
+                }
+            )},
+        {"data_dir",
+            sc(
+                string(),
+                #{
+                    desc => ?DESC(builtin_data_dir),
+                    mapping => "emqx_durable_storage.db_data_dir",
+                    required => false,
+                    importance => ?IMPORTANCE_MEDIUM
+                }
+            )},
+        {"n_shards",
+            sc(
+                pos_integer(),
+                #{
+                    importance => ?IMPORTANCE_MEDIUM,
+                    desc => ?DESC(builtin_n_shards),
+                    default => 16
+                }
+            )},
+        {"replication_factor",
+            sc(
+                pos_integer(),
+                #{
+                    default => 3,
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )},
+        {"egress",
+            sc(
+                ref("builtin_egress"),
+                #{
+                    desc => ?DESC(builtin_egress),
+                    importance => ?IMPORTANCE_MEDIUM
+                }
+            )},
+        {"layout",
+            sc(
+                hoconsc:union([
+                    ref("layout_builtin_wildcard_optimized"), ref("layout_builtin_reference")
+                ]),
+                #{
+                    desc => ?DESC(builtin_layout),
+                    importance => ?IMPORTANCE_HIDDEN,
+                    default =>
+                        #{
+                            <<"type">> => wildcard_optimized
+                        }
+                }
+            )}
+    ];
+fields("builtin_egress") ->
+    [
+        {"max_items",
+            sc(
+                pos_integer(),
+                #{
+                    default => 1000,
+                    mapping => "emqx_durable_storage.egress_batch_size",
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )},
+        {"flush_interval",
+            sc(
+                emqx_schema:timeout_duration_ms(),
+                #{
+                    default => 100,
+                    mapping => "emqx_durable_storage.egress_flush_interval",
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )}
+    ];
+fields("layout_builtin_wildcard_optimized") ->
+    [
+        {"type",
+            sc(
+                wildcard_optimized,
+                #{
+                    desc => ?DESC(layout_wildcard_optimized),
+                    'readOnly' => true,
+                    default => wildcard_optimized
+                }
+            )},
+        {"bits_per_topic_level",
+            sc(
+                range(1, 64),
+                #{
+                    default => 64,
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )},
+        {"epoch_bits",
+            sc(
+                range(0, 64),
+                #{
+                    default => 10,
+                    importance => ?IMPORTANCE_MEDIUM,
+                    desc => ?DESC(wildcard_optimized_epoch_bits)
+                }
+            )},
+        {"topic_index_bytes",
+            sc(
+                pos_integer(),
+                #{
+                    default => 4,
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )}
+    ];
+fields("layout_builtin_reference") ->
+    [
+        {"type",
+            sc(
+                reference,
+                #{'readOnly' => true}
+            )}
+    ].
+
+desc(_) ->
+    undefined.
+
+%%================================================================================
+%% Internal functions
+%%================================================================================
+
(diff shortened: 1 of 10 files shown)


apps/emqx/src/emqx_persistent_message.erl
```diff
@@ -52,31 +52,20 @@ is_persistence_enabled() ->
...
-        n_shards => NShards,
-        replication_factor => ReplicationFactor
-    };
-storage_backend(#{
-    fdb := #{enable := true} = FDBConfig
-}) ->
-    FDBConfig#{backend => fdb}.
+storage_backend(Path) ->
+    ConfigTree = #{'_config_handler' := {Module, Function}} = emqx_config:get(Path),
+    io:format("Config tree: ~p~n", [ConfigTree]),
```
savonarola, line 65: Remove?

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -0,0 +1,260 @@
...
+
+%%================================================================================
+%% Type declarations
+%%================================================================================
+
+%%================================================================================
+%% API
+%%================================================================================
+
+translate_builtin(#{
```
savonarola, line 39: Why not use a regular converter? 🤔

apps/emqx/src/emqx_persistent_message.erl
ieQu1: Yes, shouldn't be in the final PR. Need it for debug now

apps/emqx/src/emqx_ds_schema.erl
ieQu1: The goal here is not to convert a value of a field to another value in the config, but to provide a way to dynamically convert the config when a new emqx_ds database is created. Not sure if there is a standard HOCON way of doing so.

apps/emqx/src/emqx_persistent_message.erl
ieQu1: Done

rel/i18n/emqx_ds_schema.hocon
```diff
@@ -0,0 +1,35 @@
...
+"""File system directory where the database is located."""
+
+builtin_n_shards.desc:
+"""The builtin durable storage partitions data into shards.
+This configuration parameter defines the number of shards.
+Please note that it takes effect only during the initialization of the durable storage database.
+Changing this configuration parameter after the database has been already created won't take any effect."""
+
+builtin_egress.desc:
+"""Configuration related to the buffering outcoming messages to the shards."""
```
thalesmg, line 19:
```suggestion
"""Configuration related to the buffering of outgoing messages to the shards."""
```

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -0,0 +1,245 @@
...
+    }.
+
+%%================================================================================
+%% Behavior callbacks
+%%================================================================================
+
+namespace() ->
+    durable_storage.
+
+schema() ->
```
thalesmg, line 75: Shouldn't this be `roots`?

thalesmg (approved):

apps/emqx/src/emqx_ds_schema.erl
ieQu1: Stone told me that it's not required to have roots in the schema modules that are grafted to the main `emqx_schema`.

rel/i18n/emqx_ds_schema.hocon
```diff
@@ -0,0 +1,35 @@
...
+"""File system directory where the database is located."""
+
+builtin_n_shards.desc:
+"""The builtin durable storage partitions data into shards.
+This configuration parameter defines the number of shards.
+Please note that it takes effect only during the initialization of the durable storage database.
+Changing this configuration parameter after the database has been already created won't take any effect."""
+
+builtin_egress.desc:
+"""Configuration related to the buffering of outcoming messages to the shards."""
```
thalesmg, line 19:
```suggestion
"""Configuration related to the buffering of outgoing messages to the shards."""
```
?

thalesmg (approved):

rel/i18n/emqx_ds_schema.hocon
```diff
@@ -0,0 +1,35 @@
...
+Depending on the type of workload and the topic structure, different types of strategies for storing the data can be employed to maximize efficency of the replay."""
+
+
+layout_wildcard_optimized.desc:
+"""_Wildcard-optimized_ layout is designed to maximize the throughput of the wildcard subscriptions covering large numbers of topics."""
+
+wildcard_optimized_epoch_bits.desc:
+"""Wildcard-optimized layout partitions messages recorded at different times into "epochs".
+Each epoch can be consumed by the subscribers as a batch.
+Generally, larger epochs lead to higher throughput of subscribers, however currently they may increase latency."""
```
zmstone, line 33:
```suggestion
Generally, larger epochs lead to higher throughput of subscribers, however may increase latency."""
```

rel/i18n/emqx_ds_schema.hocon
```diff
@@ -0,0 +1,35 @@
...
+"""Configuration related to the buffering of messages from the local node to the shard leader."""
+
+builtin_layout.desc:
+"""Storage layout is a method of arranging messages from various topics and clients on disc.
+
+Depending on the type of workload and the topic structure, different types of strategies for storing the data can be employed to maximize efficency of the replay."""
+
+
+layout_wildcard_optimized.desc:
+"""_Wildcard-optimized_ layout is designed to maximize the throughput of the wildcard subscriptions covering large numbers of topics."""
```
zmstone, line 28: maybe give an example here ?

rel/i18n/emqx_ds_schema.hocon
```diff
@@ -0,0 +1,35 @@
...
+Please note that it takes effect only during the initialization of the durable storage database.
+Changing this configuration parameter after the database has been already created won't take any effect."""
+
+builtin_egress.desc:
+"""Configuration related to the buffering of messages from the local node to the shard leader."""
+
+builtin_layout.desc:
+"""Storage layout is a method of arranging messages from various topics and clients on disc.
+
+Depending on the type of workload and the topic structure, different types of strategies for storing the data can be employed to maximize efficency of the replay."""
```
zmstone, line 24: "replay" is an unfamiliar term for MQTT users, maybe just say "read messages from durable storage", as it's applicable for reading session messages in general.

```suggestion
Depending on the type of workload and the topic structure, different types of strategies for storing the data can be employed to maximize efficiency of reading messages from the durable storage."""
```

rel/i18n/emqx_ds_schema.hocon
```diff
@@ -0,0 +1,35 @@
+emqx_ds_schema {
+
+messages.desc:
+"""Configuration related to the durable storage of MQTT messages."""
+
+builtin.desc:
+"""Builtin session storage backend utilizing embedded RocksDB key-value store."""
+
+builtin_data_dir.desc:
+"""File system directory where the database is located."""
```
zmstone, line 10: it has no default value in the schema. 
so it's necessary to describe from where the default value would be inherited if missing from the config.

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -0,0 +1,245 @@
...
+                    desc => ?DESC(builtin_n_shards),
+                    default => 16
+                }
+            )},
+        {"replication_factor",
+            sc(
+                pos_integer(),
+                #{
+                    default => 3,
+                    importance => ?IMPORTANCE_HIDDEN
```
zmstone, line 134: Q: hidden because it's not implemented yet ?

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -0,0 +1,245 @@
...
+                }
+            )},
+        {"egress",
+            sc(
+                ref("builtin_egress"),
+                #{
+                    desc => ?DESC(builtin_egress),
+                    importance => ?IMPORTANCE_MEDIUM
```
zmstone, line 142: this field should be hidden because all fields are hidden for "builtin_egress" type.

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -0,0 +1,245 @@
...
+            )},
+        {"replication_factor",
+            sc(
+                pos_integer(),
+                #{
+                    default => 3,
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )},
+        {"egress",
```
zmstone, line 137: not sure if 'egress' is the best name here, any alternatives?

apps/emqx/src/emqx_ds_schema.erl
ieQu1: I have changed the visibility of the fields in the last commit.

apps/emqx/src/emqx_ds_schema.erl
ieQu1: Yes, exactly.

apps/emqx/src/emqx_ds_schema.erl
ieQu1: I guess it can be just named "buffer", but there are just too many different buffers.

rel/i18n/emqx_ds_schema.hocon
ieQu1: Rewritten that part.

rel/i18n/emqx_ds_schema.hocon
ieQu1: Done.

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -0,0 +1,257 @@
...
+                #{
+                    importance => ?IMPORTANCE_MEDIUM,
+                    desc => ?DESC(builtin_local_write_buffer)
+                }
+            )},
+        {layout,
+            sc(
+                hoconsc:union([
+                    ref(layout_builtin_wildcard_optimized)
+                    % , ref(layout_builtin_reference)
```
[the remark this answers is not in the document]
ieQu1: Hidden behind TEST flag.

apps/emqx/src/emqx_persistent_message.erl
```diff
@@ -52,31 +52,17 @@ is_persistence_enabled() ->
...
-        storage => {emqx_ds_storage_bitfield_lts, #{}},
-        n_shards => NShards,
-        replication_factor => ReplicationFactor
-    };
-storage_backend(#{
-    fdb := #{enable := true} = FDBConfig
-}) ->
-    FDBConfig#{backend => fdb}.
+storage_backend(Path) ->
+    ConfigTree = #{'_config_handler' := {Module, Function}} = emqx_config:get(Path),
```
keynslug, line 64: IMO odd to have that as part of the schema: it can be changed, it can end up in the `cluster.hocon` even if it does not differ from the default. Isn't it easier to assume for now that we should just call `emqx_ds_schema:something(ConfigTree)` here?

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -0,0 +1,266 @@
...
+                    desc => ?DESC(builtin_local_write_buffer_max_items)
+                }
+            )},
+        {flush_interval,
+            sc(
+                emqx_schema:timeout_duration_ms(),
+                #{
+                    default => 100,
+                    mapping => "emqx_durable_storage.egress_flush_interval",
+                    importance => ?IMPORTANCE_MEDIUM,
```
keynslug, line 176: I'd propose to leave that one hidden for now, since current egress implementation will likely need to undergo a rework in the future in my impression.

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -0,0 +1,266 @@
...
+            )},
+        {epoch_bits,
+            sc(
+                range(0, 64),
+                #{
+                    default => 10,
+                    importance => ?IMPORTANCE_MEDIUM,
+                    desc => ?DESC(wildcard_optimized_epoch_bits)
+                }
+            )},
```
keynslug, line 208: I wonder if we could turn that option into a time interval instead. Work on the consistent replication has so far shown that one of the most practical ways to deal with ordering guarantees is to just use microsecond-presicion timestamps for message keys, but changing precision will suddenly make this option backward-incompatible. One significant downside is that we won't be able to respect specified value exactly and need to tell the user what's the actual epoch size somehow.

apps/emqx/src/emqx_persistent_message.erl
ieQu1: Agree, admittedly, it's not the nicest of the hacks. However, `emqx_ds_schema:something` would have to know how to handle configuration of backends injected on EE. 

True, we could add some sort of advanced backend registration mechanism in the DS domain, but it wouldn't solve the problem of configuration translation, which is part of the emqx domain. This could be redesigned in the future.

apps/emqx/src/emqx_ds_schema.erl
ieQu1: I think I mentioned in the other PR that using microsecond-precision timestamp for serialization is fine, but it must be implemented in a way that does NOT affect neither EMQX application, NOR the storage layer. Otherwise we leak abstractions from the replication layer.

EMQX application owns the `#message` record, and assumes that the timestamp is in milliseconds. DS must store and restore message as is. 

There is no reason why the storage layer must be desined otherwise. I guess proper way to address this is to pass unique message id (which could be microsecond) from replication layer to the storage layer, instead of generating it in the storage layer.

apps/emqx/src/emqx_ds_schema.erl
ieQu1: Done

apps/emqx/src/emqx_ds_schema.erl
ieQu1: ...Made it hidden for now.

keynslug: Thanks!

> EMQX application owns the #message record, and assumes that the timestamp is in milliseconds.

That's why I mentioned message key and not emqx-level message timestamp.

> NOR the storage layer.

Arguable. I do not honestly see why storage layer could not accommodate replication layer needs and handle timestamps with variable precision for the sake of overall simplicity and compactness.

zmstone (approved):

PR #12574, opened 2024-02-23
fix(s3): disallow partially defined credentials in schema

keynslug:
Fixes [EMQX-11888](https://emqx.atlassian.net/browse/EMQX-11888).

Release version: e5.6

## Summary

Running S3 client with only partially defined credentials does not make sense and usually causes crashes somewhere down the stack. Better to disallow this on the schema level.

## PR Checklist

- [ ] ~~Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files~~
- [ ] Schema changes are backward compatible **(not entirely, but this is again the point of the PR)**


apps/emqx_s3/src/emqx_s3_schema.erl
```diff
@@ -169,6 +170,27 @@ desc(s3_upload) ->
...
     "Options for the HTTP transport layer used by the S3 client".
 
+validations() ->
+    [{no_partially_defined_credentials, fun validate_no_partially_defined_credentials/1}].
+
+validate_no_partially_defined_credentials(Config) ->
+    AccessKeyId = hocon_maps:get("s3.access_key_id", Config),
+    SecretAccessKey = hocon_maps:get("s3.secret_access_key", Config),
+    case {AccessKeyId, SecretAccessKey} of
+        {_Defined, undefined} ->
```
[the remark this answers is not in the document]
keynslug: Oh right! 🫠 Thanks.

apps/emqx_s3/src/emqx_s3_schema.erl
```diff
@@ -169,6 +170,27 @@ desc(s3_upload) ->
 desc(transport_options) ->
     "Options for the HTTP transport layer used by the S3 client".
 
+validations() ->
```
zmstone, line 173: Since the values are not cross roots, so there is no maybe need to add a root level validator which is evaluated by all config checks even if it's not checking this root.
IIRC, we can implement roots/0 like this.

```
roots() ->
  [{s3, hoconsc:mk(hoconsc:ref(s3), #{validator => fun(Conf) -> ... end})].
```

keynslug: What do you mean by "cross roots"?

zmstone: I mean if one needs to verify `root1.key1 = bar` is valid given `root0.key0 = foo`, then it's a cross-root validation,
one would have to add a global validator.

e.g. the recently added `cluster.discoverty_strategy = dns` vs `node.name = emqx@FQDN` check.

PR #12589, opened 2024-02-26
Make retainer pluggable

savonarola:
Fixes [EMQX-11837](https://emqx.atlassian.net/browse/EMQX-11837)

1. removes any direct usages of retainer backends; code is refactored a bit
2. schema is extended to support alternative backend. 

### Problem

Changing schema is the most challenging part.

For a long time, for retainer backend options, we had the following config
```
backend {
   type = built_in_database
   storage_type = ram
   ...
}
```

with the intention to change the type and the set of options for an alternative backend:

```
backend {
   type = some_new_backend
   new_option = some_value
   ...
}
```

However, we discouraged this approach many times because it led to inconsistent configurations after the change. When the bootstrap config is again merged with the cluster config, we obtain something like

```
backend {
   type = some_new_backend
   storage_type = ram
   new_option = some_value
   ...
}

```

So 
1. One one hand, the used approach is actually impossible to extend;
2. On the other hand, we need to keep backward compatibility.

### Proposed solution

This PR introduces the following structure:

```
backend { # this always configures the builtin backend
   type = built_in_database # this is deprecated & hidden
   enable = false # default is true
   storage_type = ram
   ...
}
external_backends { # these are injected additional backends
   fdb_ds => {
      enable => true
   }
}
```

The term `external` somewhat reflects that the storage and its settings are in the other apps and config sections (the ds configs introduced by Dmitrii).

### Alternative approaches

Alternative approach was to
* Introduce not `external_backends` but just `backends`;
* Make a converter placing legacy `backend` section under `backends.builtin` path.

But things seemed to be too complex, and this led to strange behavior even in tests...

@zmstone @ieQu1, what do you think?

Release version: v/e5.?

### DIFF
--- a/apps/emqx_retainer/src/emqx_retainer.erl
+++ b/apps/emqx_retainer/src/emqx_retainer.erl
@@ -32,8 +32,7 @@
 
 -export([
     delete_message/2,
-    store_retained/2,
-    get_backend_module/0
+    store_retained/2
 ]).
 
 -export([
@@ -45,7 +44,15 @@
     page_read/3,
     post_config_update/5,
     stats_fun/0,
-    retained_count/0
+    retained_count/0,
+    backend_module/0,
+    backend_module/1,
+    enabled/0
+]).
+
+%% For testing only
+-export([
+    context/0
 ]).
 
 %% gen_server callbacks
@@ -54,8 +61,7 @@
     handle_call/3,
     handle_cast/2,
     handle_info/2,
-    terminate/2,
-    code_change/3
+    terminate/2
 ]).
 
 -export_type([
@@ -71,29 +77,35 @@
     context := undefined | context(),
     clear_timer := undefined | reference()
 }.
--type context() :: term().
+
+-type backend_state() :: term().
+
+-type context() :: #{
+    module := module(),
+    state := backend_state()
+}.
 
 -type topic() :: emqx_types:topic().
 -type message() :: emqx_types:message().
 -type cursor() :: undefined | term().
--type backend() :: emqx_retainer_mnesia | module().
+-type has_next() :: boolean().
 
 -define(DEF_MAX_PAYLOAD_SIZE, (1024 * 1024)).
 -define(DEF_EXPIRY_INTERVAL, 0).
 -define(MAX_PAYLOAD_SIZE_CONFIG_PATH, [retainer, max_payload_size]).
 
--callback create(map()) -> context().
--callback update(context(), map()) -> ok | need_recreate.
--callback close(context()) -> ok.
--callback delete_message(context(), topic()) -> ok.
--callback store_retained(context(), message()) -> ok.
--callback read_message(context(), topic()) -> {ok, list()}.
--callback page_read(context(), topic(), non_neg_integer(), non_neg_integer()) ->
-    {ok, HasNext :: boolean(), list()}.
--callback match_messages(context(), topic(), cursor()) -> {ok, list(), cursor()}.
--callback clear_expired(context()) -> ok.
--callback clean(context()) -> ok.
--callback size(context()) -> non_neg_integer().
+-callback create(hocon:config()) -> backend_state().
+-callback update(backend_state(), hocon:config()) -> ok | need_recreate.
+-callback close(backend_state()) -> ok.
+-callback delete_message(backend_state(), topic()) -> ok.
+-callback store_retained(backend_state(), message()) -> ok.
+-callback read_message(backend_state(), topic()) -> {ok, list(message())}.
+-callback page_read(backend_state(), emqx_maybe:t(topic()), non_neg_integer(), non_neg_integer()) ->
+    {ok, has_next(), list(message())}.
+-callback match_messages(backend_state(), topic(), cursor()) -> {ok, list(message()), cursor()}.
+-callback clear_expired(backend_state()) -> ok.
+-callback clean(backend_state()) -> ok.
+-callback size(backend_state()) -> non_neg_integer().
 
 %%------------------------------------------------------------------------------
 %% Hook API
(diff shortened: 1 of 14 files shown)


apps/emqx_retainer/test/emqx_retainer_backend_SUITE.erl
```diff
@@ -0,0 +1,104 @@
...
+
+-module(emqx_retainer_backend_SUITE).
+
+-compile(export_all).
+-compile(nowarn_export_all).
+
+-include_lib("eunit/include/eunit.hrl").
+-include_lib("common_test/include/ct.hrl").
+
+all() -> emqx_common_test_helpers:all(?MODULE).
```
thalesmg, line 25: Nit: maybe move this under the `Setups` heading.

apps/emqx_retainer/test/emqx_retainer_dummy.erl
```diff
@@ -0,0 +1,97 @@
...
+                    default => <<"emqx_retainer_dummy">>,
+                    importance => ?IMPORTANCE_HIDDEN
+                }
+            )},
+        {enable,
+            hoconsc:mk(
+                boolean(),
+                #{
+                    desc => <<"enable dummy backend">>,
+                    required => false,
```
thalesmg, line 93: Nano nit:

```suggestion
                    required => true,
```

(so that it would fail if `external_backends.dummy = {}`)

apps/emqx_retainer/test/emqx_retainer_SUITE.erl
```diff
@@ -789,6 +787,11 @@ t_compatibility_for_deliver_rate(_) ->
         Parser(DeliveryInf)
     ).
 
+t_update_config(_) ->
+    OldConf = emqx_config:get_raw([retainer]),
+    NewConf = emqx_utils_maps:deep_put([<<"backend">>, <<"storage_type">>], OldConf, <<"disk">>),
+    emqx_retainer:update_config(NewConf).
```
thalesmg, line 793: Q: shouldn't we assert a result here?

apps/emqx_retainer/src/emqx_retainer_schema.erl
```diff
@@ -214,3 +238,16 @@ retainer_converter(#{<<"deliver_rate">> := Delivery} = Conf, Opts) ->
...
+    BuiltInBackend = maps:get(<<"backend">>, Config, #{}),
+    ExternalBackends = maps:values(maps:get(<<"external_backends">>, Config, #{})),
+    Enabled = lists:filter(fun(#{<<"enable">> := E}) -> E end, [BuiltInBackend | ExternalBackends]),
+    case Enabled of
+        [#{}] ->
+            ok;
+        _Conflicts = [_ | _] ->
+            {error, multiple_enabled_backends};
+        _None = [] ->
+            {error, no_enabled_backend}
```
thalesmg, line 252: Q: it seems we allow disabling the retainer feature, won't this disallow such functionality?

apps/emqx_retainer/src/emqx_retainer_cli.erl
```diff
@@ -107,6 +125,23 @@ count() ->
...
+    end.
+
+if_mnesia_backend(Fun) ->
+    case emqx_retainer:backend_module() of
+        emqx_retainer_mnesia -> Fun();
+        _ -> ?PRINT_MSG("Command only applicable for builtin backend~n")
+    end.
+
+if_mnesia_enabled(Fun) ->
+    if_mnesia_backend(fun() ->
```
thalesmg, line 145:
```suggestion
    if_enabled(fun() ->
```

apps/emqx_retainer/src/emqx_retainer.erl
```diff
@@ -414,24 +458,46 @@ check_timer(Timer, undefined, _) ->
...
-create_resource(Context, #{type := built_in_database} = Cfg) ->
-    emqx_retainer_mnesia:create_resource(Cfg),
-    Context.
-
--spec close_resource(context()) -> ok | {error, term()}.
-close_resource(#{resource_id := ResourceId}) ->
-    emqx_resource:stop(ResourceId);
-close_resource(_) ->
-    ok.
+-spec enabled_backend_config(hocon:config()) -> emqx_maybe:t(hocon:config()).
```
thalesmg, line 461: Nit: this seems to actually be:

```suggestion
-spec enabled_backend_config(hocon:config()) -> hocon:config() | no_return().
```

apps/emqx_retainer/src/emqx_retainer.erl
```diff
@@ -414,24 +458,46 @@ check_timer(Timer, undefined, _) ->
...
+enabled_backend_config(#{backend := Backend, external_backends := ExternalBackends} = Config) ->
+    AllBackends = [Backend | maps:values(ExternalBackends)],
+    case lists:search(fun(#{enable := Enable}) -> Enable end, AllBackends) of
+        {value, EnabledBackend} -> EnabledBackend;
+        false -> error({no_enabled_backend, Config})
```
thalesmg, line 466: From the other functions, apparently this should be:

```suggestion
        false -> undefined
```

apps/emqx_retainer/src/emqx_retainer_schema.erl
savonarola: The retainer itself is disabled at higher level.

thalesmg: Maybe this validation should happen only if `retainer.enable = true`?

apps/emqx_retainer/test/emqx_retainer_backend_SUITE.erl
savonarola: ✅

apps/emqx_retainer/test/emqx_retainer_SUITE.erl
savonarola: ✅

thalesmg: Apparently this always returns ok?

thalesmg (approved):

savonarola:
Merging, since we do not introduce incompatible changes to the emqx community/enterprise. We may change the schema later.

PR #12540, opened 2024-02-20
feat: pulsar bridge v2

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-11465

Release version: v/e5.6.0

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [ ] Added tests for the changes
- [ ] Added property-based tests for code which performs user input validation
- [ ] Changed lines covered in coverage report
- [ ] Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [ ] For internal contributor: there is a jira ticket to track this change
- [ ] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket
- [ ] Schema changes are backward compatible

## Checklist for CI (.github/workflows) changes

- [ ] If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)
- [ ] Change log has been added to `changes/` dir for user-facing artifacts update

PR #12600, opened 2024-02-27
Fix bridge v2 and connector schema check

zhongwencool:
Fixes <issue-or-jira-number>
Previously, using the wrong schema never returned an error.

cherry pick from https://github.com/emqx/emqx/pull/12540, to make PR smaller is better for review.
this change cover by https://github.com/emqx/emqx/pull/12540 's test case.

Release version: v/e5.6.0

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_v2.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_v2.erl
@@ -1611,12 +1611,12 @@ split_and_validate_bridge_v1_config(BridgeV1Type, BridgeName, RawConf, PreviousR
         }
     },
     try
-        hocon_tconf:check_plain(
+        _ = hocon_tconf:check_plain(
             emqx_connector_schema,
             NewFakeConnectorConfig,
             #{atom_key => false, required => false}
         ),
-        hocon_tconf:check_plain(
+        _ = hocon_tconf:check_plain(
             emqx_bridge_v2_schema,
             NewFakeBridgeV2Config,
             #{atom_key => false, required => false}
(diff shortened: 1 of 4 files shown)


apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -1610,8 +1612,13 @@ split_and_validate_bridge_v1_config(BridgeV1Type, BridgeName, RawConf, PreviousR
     },
     try
         hocon_tconf:check_plain(
-            emqx_schema,
-            NewFakeGlobalConfig,
```
zhongwencool, line 1614: the emqx_schema is always return ok and validate nothing.

apps/emqx_resource/include/emqx_resource.hrl
```diff
@@ -154,7 +154,7 @@
 %% boolean
 -define(START_AFTER_CREATED, true).
 
--define(TEST_ID_PREFIX, "_probe_:").
```
zhongwencool, line 157: https://github.com/emqx/hocon/blob/master/src/hocon_tconf.erl#L81
map key need to be `"^[A-Za-z0-9]+[A-Za-z0-9-_]*$"` so we change the prefix to t_probe_

apps/emqx_resource/src/emqx_resource.erl
```diff
@@ -650,6 +651,10 @@ forget_allocated_resources(InstanceId) ->
     true = ets:delete(?RESOURCE_ALLOCATION_TAB, InstanceId),
     ok.
 
+deallocate_resource(InstanceId, Key) ->
```
zhongwencool, line 654: I use this function in pulsar PR. just add here is ok.

apps/emqx_resource/include/emqx_resource.hrl
zmstone: is `:` needed ?

zmstone:
bad rebase ?

apps/emqx_resource/include/emqx_resource.hrl
thalesmg: the places that concatenate this with the name don't add a `:`, but I assume it'll fail hocon's key format check

thalesmg (approved):

PR #12633, opened 2024-03-04
refactor: get-channel-status don't call get-status

zhongwencool:
Fixes <issue-or-jira-number>

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource.erl
+++ b/apps/emqx_resource/src/emqx_resource.erl
@@ -128,6 +128,7 @@
 ]).
 
 -export([apply_reply_fun/2]).
+-export([channel_status/3]).
 
 %% common validations
 -export([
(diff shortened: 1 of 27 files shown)


apps/emqx_resource/src/emqx_resource.erl
```diff
@@ -628,6 +629,19 @@ apply_reply_fun({F, A}, Result) when is_function(F) ->
 apply_reply_fun(From, Result) ->
     gen_server:reply(From, Result).
```
kjellwinblad, line 631: Do we really need this function? I think the `emqx_resource_manger` should be designed so that `on_get_channel_status` for a channel is only called when the channel is actually added. Is it not designed in this way already? If not, I think we can fix it at a higher level. The resource state contains map with all the IDs of added resources so it should be easy to fix.

I think a more descriptive name would make the code easier to understand if we do need to fix it in the `on_get_channel_status` implementations.

zhongwencool: One thing I can confirm is that there is no need to invoke on_get_status in on_get_channel_status.
However, it seems unnecessary to judge whether channel_id exists in on_get_channel_status. From the current code, it is certain to exist. I need to investigate further.
Yet, in the end, on_get_channel_status should have a default implementation, and most bridges don't need to implement this function.

PR #12639, opened 2024-03-04
test(emqx_broker_SUITE): fix flaky test case

kjellwinblad:
Fixes https://emqx.atlassian.net/browse/EMQX-11933

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -721,8 +721,8 @@ do_get_chann_conn_mod(ClientId, ChanPid) ->
     end.
 
 mark_channel_connected(ChanPid) ->
-    ?tp(emqx_cm_connected_client_count_inc, #{chan_pid => ChanPid}),
     ets:insert_new(?CHAN_LIVE_TAB, {ChanPid, true}),
+    ?tp(emqx_cm_connected_client_count_inc, #{chan_pid => ChanPid}),
     ok.
 
 mark_channel_disconnected(ChanPid) ->
--- a/apps/emqx/test/emqx_broker_SUITE.erl
+++ b/apps/emqx/test/emqx_broker_SUITE.erl
@@ -559,45 +559,70 @@ t_connected_client_count_transient_takeover(Config) when is_list(Config) ->
     %% we spawn several clients simultaneously to cause the race
     %% condition for the client id lock
     NumClients = 20,
+    ConnectSuccessCntr = counters:new(1, []),
+    ConnectFailCntr = counters:new(1, []),
+    ConnectFun =
+        fun() ->
+            process_flag(trap_exit, true),
+            try
+                {ok, ConnPid} =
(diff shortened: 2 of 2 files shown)


apps/emqx/test/emqx_broker_SUITE.erl
```diff
@@ -559,45 +559,70 @@ t_connected_client_count_transient_takeover(Config) when is_list(Config) ->
...
+    ConnectFun =
+        fun() ->
+            try
+                process_flag(trap_exit, true),
+                {ok, ConnPid} =
+                    emqtt:start_link([
+                        {clean_start, true},
+                        {clientid, ClientID}
+                        | Config
+                    ]),
```
keynslug, line 573: Does it make sense to move it out of try expression? So that it's indeed only a connect call left inside.

apps/emqx/src/emqx_cm.erl
```diff
@@ -723,6 +723,7 @@ do_get_chann_conn_mod(ClientId, ChanPid) ->
 mark_channel_connected(ChanPid) ->
     ?tp(emqx_cm_connected_client_count_inc, #{chan_pid => ChanPid}),
     ets:insert_new(?CHAN_LIVE_TAB, {ChanPid, true}),
+    ?tp(emqx_cm_connected_client_count_inc_done, #{chan_pid => ChanPid}),
```
keynslug, line 726: Maybe should be enough just to move the tracepoint, instead of introducing a new one? Feels a bit too much for such a trivial operation.

keynslug (commented):
👍🏼

thalesmg (approved):

apps/emqx/src/emqx_cm.erl
kjellwinblad: I think you are right. I saw that `mark_channel_disconnected` had `emqx_cm_connected_client_count_dec` and `emqx_cm_connected_client_count_dec_done` and I thought it made sense to make them symmetrical but there is probably no good reson for that so I will just move it instead of introducing a new one (it might even fix some other flakiness).

apps/emqx/test/emqx_broker_SUITE.erl
kjellwinblad: I think we want to catch if the `emqtt:start_link` fails as well and treat it as a connect failure but the `process_flag` call can be moved out from the try block.

apps/emqx/src/emqx_cm.erl
kjellwinblad: Fixed and force pushed

apps/emqx/test/emqx_broker_SUITE.erl
kjellwinblad: Moved the process_flag call up one level and force pushed

thalesmg (approved):

PR #12675, opened 2024-03-08
feat(ds): move session data from `mnesia` to `emqx_ds`

thalesmg:
Partially fixes https://emqx.atlassian.net/browse/EMQX-11841

Release version: v/e5.7

## Summary

This implements storing session data into DS itself, instead of relying on mnesia.

By default, things will stay as before, using mnesia to store session data.

To enable this new feature, one must compile the project setting the environment variable `STORE_STATE_IN_DS=1`.

## PR Checklist

### DIFF
--- a/apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
+++ b/apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
@@ -25,6 +25,7 @@ all() ->
     emqx_common_test_helpers:all(?MODULE).
 
 init_per_suite(Config) ->
+    emqx_common_test_helpers:clear_screen(),
     TCApps = emqx_cth_suite:start(
         app_specs(),
         #{work_dir => emqx_cth_suite:work_dir(Config)}
@@ -198,6 +199,22 @@ force_last_alive_at(ClientId, Time) ->
     S = emqx_persistent_session_ds_state:set_last_alive_at(Time, S0),
     _ = emqx_persistent_session_ds_state:commit(S),
     ok.
+
+%% Helper for determining if rpc faults should be injected.
+%% When storing session data in DS, failing here could lead to a session being considered
+%% new and not loaded, since we read non-atomically from the DB.  If getting the streams
+%% fail due to rpc problems, it leniently returns an empty list of streams, and that
+%% results in a "new" session.
+is_restoring_session() ->
+    {current_stacktrace, Stacktrace} = process_info(self(), current_stacktrace),
+    is_restoring_session(Stacktrace).
+
+is_restoring_session([]) ->
+    false;
+is_restoring_session([{emqx_persistent_session_ds, session_open, _, _} | _Rest]) ->
+    true;
+is_restoring_session([_ | Rest]) ->
+    is_restoring_session(Rest).
 
 %%------------------------------------------------------------------------------
 %% Testcases
@@ -649,7 +666,7 @@ t_session_replay_retry(_Config) ->
         ClientsPub
     ),
 
-    Pubs0 = emqx_common_test_helpers:wait_publishes(NClients, 5_000),
+    Pubs0 = emqx_common_test_helpers:wait_publishes(NClients, 6_000),
     NPubs = length(Pubs0),
     ?assertEqual(NClients, NPubs, ?drainMailbox(1_500)),
 
@@ -658,9 +675,18 @@ t_session_replay_retry(_Config) ->
     %% Make `emqx_ds` believe that roughly half of the shards are unavailable.
     ok = emqx_ds_test_helpers:mock_rpc_result(
         fun(_Node, emqx_ds_replication_layer, _Function, [_DB, Shard | _]) ->
-            case erlang:phash2(Shard) rem 2 of
-                0 -> unavailable;
-                1 -> passthrough
+            %% When storing session data in DS, failing here could lead to a session being
+            %% considered new and not loaded, since we read non-atomically from the DB.
+            %% If getting the streams fail due to rpc problems, it leniently returns an
+            %% empty list of streams, and that results in a "new" session.
+            case is_restoring_session() of
+                true ->
+                    passthrough;
+                false ->
+                    case erlang:phash2(Shard) rem 2 of
+                        0 -> unavailable;
+                        1 -> passthrough
+                    end
             end
         end
     ),
--- a/apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
+++ b/apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
@@ -14,6 +14,8 @@
 
 -import(emqx_common_test_helpers, [on_exit/1]).
 
+-define(SESSION_DB, emqx_persistent_session).
+
 %%------------------------------------------------------------------------------
 %% CT boilerplate
 %%------------------------------------------------------------------------------
@@ -199,22 +201,6 @@ force_last_alive_at(ClientId, Time) ->
     S = emqx_persistent_session_ds_state:set_last_alive_at(Time, S0),
     _ = emqx_persistent_session_ds_state:commit(S),
     ok.
-
-%% Helper for determining if rpc faults should be injected.
-%% When storing session data in DS, failing here could lead to a session being considered
-%% new and not loaded, since we read non-atomically from the DB.  If getting the streams
-%% fail due to rpc problems, it leniently returns an empty list of streams, and that
-%% results in a "new" session.
-is_restoring_session() ->
-    {current_stacktrace, Stacktrace} = process_info(self(), current_stacktrace),
-    is_restoring_session(Stacktrace).
-
-is_restoring_session([]) ->
-    false;
-is_restoring_session([{emqx_persistent_session_ds, session_open, _, _} | _Rest]) ->
-    true;
-is_restoring_session([_ | Rest]) ->
-    is_restoring_session(Rest).
 
 %%------------------------------------------------------------------------------
 %% Testcases
@@ -674,20 +660,18 @@ t_session_replay_retry(_Config) ->
 
     %% Make `emqx_ds` believe that roughly half of the shards are unavailable.
     ok = emqx_ds_test_helpers:mock_rpc_result(
-        fun(_Node, emqx_ds_replication_layer, _Function, [_DB, Shard | _]) ->
-            %% When storing session data in DS, failing here could lead to a session being
-            %% considered new and not loaded, since we read non-atomically from the DB.
-            %% If getting the streams fail due to rpc problems, it leniently returns an
-            %% empty list of streams, and that results in a "new" session.
-            case is_restoring_session() of
-                true ->
-                    passthrough;
-                false ->
-                    case erlang:phash2(Shard) rem 2 of
-                        0 -> unavailable;
-                        1 -> passthrough
-                    end
-            end
+        fun
+            (_Node, emqx_ds_replication_layer, _Function, [?SESSION_DB | _]) ->
+                %% When storing session data in DS, failing here could lead to a session
+                %% being considered new and not loaded, since we read non-atomically from
+                %% the DB.  If getting the streams fail due to rpc problems, it leniently
+                %% returns an empty list of streams, and that results in a "new" session.
+                passthrough;
+            (_Node, emqx_ds_replication_layer, _Function, [_DB, Shard | _]) ->
+                case erlang:phash2(Shard) rem 2 of
+                    0 -> unavailable;
+                    1 -> passthrough
+                end
         end
     ),
 
--- a/apps/emqx/src/emqx_ds_schema.erl
+++ b/apps/emqx/src/emqx_ds_schema.erl
@@ -80,7 +80,7 @@ namespace() ->
 schema() ->
     [
         {messages,
-            ds_schema(ref(builtin), #{
+            ds_schema(ref(builtin), 'durable_storage.messages', #{
                 default =>
                     #{
                         <<"backend">> => builtin
@@ -89,7 +89,7 @@ schema() ->
                 desc => ?DESC(messages)
             })},
         {sessions,
-            ds_schema(ref(builtin_session), #{
+            ds_schema(ref(builtin_session), 'durable_storage.sessions', #{
                 default =>
                     #{
                         <<"backend">> => builtin
@@ -290,11 +290,11 @@ desc(_) ->
 %% Internal functions
 %%================================================================================
 
-ds_schema(MainRef, Options) ->
+ds_schema(MainRef, InjectionPoint, Options) ->
     sc(
         hoconsc:union([
             MainRef
-            | emqx_schema_hooks:injection_point('durable_storage.backends', [])
+            | emqx_schema_hooks:union_option_injections(InjectionPoint, [])
         ]),
         Options
     ).
--- a/apps/emqx/src/emqx_ds_schema.erl
+++ b/apps/emqx/src/emqx_ds_schema.erl
@@ -80,13 +80,22 @@ namespace() ->
 schema() ->
     [
         {messages,
-            ds_schema(#{
+            ds_schema(ref(builtin), #{
                 default =>
                     #{
                         <<"backend">> => builtin
                     },
                 importance => ?IMPORTANCE_MEDIUM,
                 desc => ?DESC(messages)
+            })},
+        {sessions,
+            ds_schema(ref(builtin_session), #{
+                default =>
+                    #{
+                        <<"backend">> => builtin
+                    },
+                importance => ?IMPORTANCE_MEDIUM,
+                desc => ?DESC(sessions)
             })}
     ].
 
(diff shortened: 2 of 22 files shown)


apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
```diff
@@ -25,6 +25,7 @@ all() ->
     emqx_common_test_helpers:all(?MODULE).
 
 init_per_suite(Config) ->
+    emqx_common_test_helpers:clear_screen(),
```
ieQu1, line 28: Forgotten?

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -229,36 +273,50 @@ fields(layout_builtin_reference) ->
...
 %%================================================================================
 
-ds_schema(Options) ->
+ds_schema(MainRef, InjectionPoint, Options) ->
     sc(
         hoconsc:union([
-            ref(builtin)
-            | emqx_schema_hooks:injection_point('durable_storage.backends', [])
+            MainRef
+            | emqx_schema_hooks:union_option_injections(InjectionPoint, [])
```
thalesmg, line 295: This addresses this comment: https://github.com/emqx/emqx-platform/pull/135#discussion_r1514765105

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -478,94 +592,221 @@ pmap_del(
...
+->
+    %% This introduces a redundant extra topic level for the metadata domain, but it also
+    %% makes the topic structure uniform in relation to the other domains.
+    BinKey = key_encode(Domain, Key),
+    to_topic(Domain, SessionId, BinKey).
+
+to_topic(Domain, SessionId0, BinKey) when is_binary(BinKey) ->
+    SessionId = emqx_http_lib:uri_encode(SessionId0),
+    emqx_topic:join([
+        <<"session">>,
```
ieQu1, line 708: I don't see the reason to have a common prefix for the topics.

thalesmg: I think it's a good idea to have namespaces, in case we might want to add other data to this database, such as some kind of metadata not tied to a particular clientid.

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -478,94 +592,221 @@ pmap_del(
...
+        rank_key()
+    ) ->
+        binary();
+    (
+        ?seqno_domain,
+        emqx_persistent_session_ds:id(),
+        seqno_type()
+    ) ->
+        binary().
+domain_topic_encode(Domain, SessionId, Key) when
```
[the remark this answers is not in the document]
thalesmg: They all have the same implementation.  :thinking:

apps/emqx/src/emqx_schema_hooks.erl
```diff
@@ -56,12 +67,22 @@ injection_point(PointName) ->
 injection_point(PointName, Default) ->
     persistent_term:get(?HOOKPOINT_PT_KEY(PointName), Default).
 
+-spec union_option_injections(hookpoint()) -> [hocon_schema:field()].
```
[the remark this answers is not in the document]
thalesmg: This has more context: https://github.com/emqx/emqx-platform/pull/135#discussion_r1514765105

Basically, we're injection union members, not fields.

apps/emqx/test/emqx_persistent_session_SUITE.erl
```diff
@@ -139,6 +139,7 @@ end_per_group(_, _Config) ->
     ok.
 
 init_per_testcase(TestCase, Config) ->
+    emqx_common_test_helpers:clear_screen(),
```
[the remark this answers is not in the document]
thalesmg: I think it doesn't need to be all or nothing...  

If there is somewhere to place it that will affect things globally (I don't know where), it can be added there later, without affecting existing places (it does nothing the 2nd+ time it's called).  Even if this place doesn't exist, I don't think it's a reason to forbid it from being there, since it's a no-op if you don't use it.

apps/emqx/src/emqx_schema_hooks.erl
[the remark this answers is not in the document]
ieQu1: Is it just about the naming, fields vs. union members? I've tested the injection into the platform code, and it worked. 

If so, maybe the callback can be renamed to `inject_schema`, and have its return type generalized?

apps/emqx/test/emqx_persistent_session_SUITE.erl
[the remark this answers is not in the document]
ieQu1: I don't see why persistent session suite should behave differently in terms of UI from all the other test suites. It should be a job of the test framework to check the user preferences and apply them consistently.

I think this call can be moved to `emqx_common_test_helpers:init_per_testcase`.

apps/emqx/src/emqx_persistent_session_ds_state.erl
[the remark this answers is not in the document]
ieQu1: This implementation is two lines of code. And it's not the same, since at least some of the cases have unused parameters.

apps/emqx/test/emqx_persistent_session_SUITE.erl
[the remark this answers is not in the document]
thalesmg: Sorry, what do you mean.

As I mentioned before, it's a no-op when the `CLEAR_SCREEN=true` environment variable is not set...  So leaving it there won't make any difference at all, nothing jarring. 

And I can't see how that's related to UI.  :thinking:

apps/emqx/src/emqx_schema_hooks.erl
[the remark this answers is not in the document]
thalesmg: It'll add fields to `durable_storage.{messages,session}`, but we want to add constructors to the union type.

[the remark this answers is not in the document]
ieQu1: I see, thanks.

apps/emqx/src/emqx_persistent_session_ds_state.erl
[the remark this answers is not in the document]
thalesmg: Yes, those cases are handled by `key_encode` already.

apps/emqx/test/emqx_persistent_session_SUITE.erl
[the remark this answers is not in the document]
ieQu1: Can you explain to someone unfamiliar with the code, why does `emqx_persistent_session_SUITE` test suite cleans the display, while, for example, `emqx_flapping_SUITE` doesn't? Is it related to the business logic at hand?

apps/emqx/src/emqx_schema_hooks.erl
```diff
@@ -129,5 +165,68 @@ inject_fields(PointName, Fields) ->
...
+        ModuleInjections
+    ).
+
+any_union_option_injections(PointName) ->
+    persistent_term:get(?UNION_OPTION_PT_KEY(PointName), undefined) =/= undefined.
+
+call_if_implemented(Mod, Fun, Args, Default) ->
+    %% ensure modeule is loaded
+    _ = Mod:module_info(),
+    case erlang:function_exported(Mod, Fun, length(Args)) of
```
ieQu1, line 227: This can be rewritten as 
```
lists:member({Fun, length(Args)}, Mod:module_info(exports)).
```

apps/emqx/test/emqx_persistent_session_SUITE.erl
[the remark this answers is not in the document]
thalesmg: Both won't do such a thing, unless you have `CLEAR_SCREEN=true` set?  And I guess it's not a common environment variable to be set.  Do you think it may be commonly set by devs?

apps/emqx/src/emqx_persistent_session_ds_state.erl
```diff
@@ -478,94 +593,221 @@ pmap_del(
...
-        {attributes, record_info(fields, kv)}
+from_domain_msg(#message{topic = Topic, payload = Bin}) ->
+    #{
+        domain := Domain,
+        session_id := _SessionId,
+        key := _Key
+    } = Data = domain_topic_decode(Topic),
+    Data#{val => val_decode(Domain, Bin)}.
+
+-spec domain_topic_encode
```
ieQu1, line 663: I find that a 30-line signature to a 2-line function that just passes the parameters into another function significantly *reduces* readability of the code. This is an internal function within a self-contained module that handles a well-defined set of data.

apps/emqx/src/emqx_persistent_session_ds_state.erl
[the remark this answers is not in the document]
ieQu1: So there is no need for this function then?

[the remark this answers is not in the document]
thalesmg: In my opinion, it doesn't hurt, simply a matter of style.  But ok, I'll remove it.

apps/emqx/src/emqx_persistent_session_ds_state.erl
thalesmg: Removed.

thalesmg:
Maybe we could add a temporary CI job to run emqx's app suites+eunit+proper tests with this flag enabled.

apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
```diff
@@ -658,9 +675,18 @@ t_session_replay_retry(_Config) ->
...
-            case erlang:phash2(Shard) rem 2 of
-                0 -> unavailable;
-                1 -> passthrough
+            %% When storing session data in DS, failing here could lead to a session being
+            %% considered new and not loaded, since we read non-atomically from the DB.
+            %% If getting the streams fail due to rpc problems, it leniently returns an
+            %% empty list of streams, and that results in a "new" session.
+            case is_restoring_session() of
+                true ->
+                    passthrough;
```
keynslug, line 684: This sounds weird, though I have to admit I haven't yet finished reviewing.

I think that read failure should not be equivalent to having no session record. Is there any good reason to do that?

On the test code: wouldn't it be simpler to match on `DB` and simulate failure only for the message DB? Current test code looks both super-hacky and extemely fragile.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -163,7 +163,14 @@
...
         %% Default: `false'.
-        atomic => boolean()
+        atomic => boolean(),
+        %% Defines whether to automatically assign strictly monotonically increasing
+        %% internal timestamps when storing messages (which precludes upserting even if
+        %% crafting timestamps in the message), or allow repeated timestamps in the DB
+        %% (which allows upserting keys, depending on the properties of the layout
+        %% keymapper).
+        %% Default: `true'.
+        auto_assign_timestamps => boolean()
```
keynslug, line 173: Hm, I have to say I 100% thought that we agreed on the approach where this is part of batch entry rather than whole batch. Having it per-batch seemingly brings too much of unnecessary complexity.

I.e. something close to:
```
-spec store_batch(db(), [emqx_types:message() | {_TsID :: emqx_ds:time(), emqx_types:message()}], message_store_opts()) ->
    store_batch_result().
```

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
```diff
@@ -1535,3 +1546,30 @@ connect_client(Opts) ->
...
+%% before attempting to GC a session.
+stop_and_commit(Client) ->
+    {ok, {ok, _}} =
+        ?wait_async_action(
+            emqtt:stop(Client),
+            #{?snk_kind := persistent_session_ds_terminate}
+        ),
+    ok.
+
+delete_all_sessions(N) ->
```
keynslug, line 1561: I'd definitely vote for not reaching that deep into the implementation here. Should be enough to `emqtt:disconnect(C, ?RC_SUCCESS, #{'Session-Expiry-Interval' => 0})` each client? After all, we strive to respect the spec in this regard, thereby these sessions should disappear from API responses anyway.

apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
thalesmg: > I think that read failure should not be equivalent to having no session record. Is there any good reason to do that?

Just that I'm not sure how we should handle (recoverable) failures in this case.  :thinking:   
Sleep and retry?

> On the test code: wouldn't it be simpler to match on DB and simulate failure only for the message DB? Current test code looks both super-hacky and extemely fragile.

Sounds like a much better idea, I'll fix it.

apps/emqx_durable_storage/src/emqx_ds.erl
thalesmg: > Hm, I have to say I 100% thought that we agreed on the approach where this is part of batch entry rather than whole batch.

Yes, and I thought that it implies having this option per-batch, so that we know whether to auto assign or not.

With the proposed spec change, wouldn't that move the auto-assignment to the caller instead of replication layer?   Or are you suggesting that, if `TsID` is defined by the caller, it is not subject to the auto-assign?

I didn't understand that suggestion when we talked before.  I'll adapt to use this then.

apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
keynslug: > Sleep and retry?

Good question. This warrants some discussion but as a safe bet I'd propagate this failure to the client, and let them decide if they want to retry.

Also wondering what would happen now if locking is unavailable in a mria transaction, probably something effectively similar?

thalesmg: > Good question. This warrants some discussion but as a safe bet I'd propagate this failure to the client, and let them decide if they want to retry.

This comes from `get_streams` being [lenient with failures](https://github.com/emqx/emqx/blob/a9ce4d7645511be61094b001868380fb701da4b4/apps/emqx_durable_storage/src/emqx_ds_replication_layer.erl#L225-L231), so it could affect other portions of the code as well.

> Also wondering what would happen now if locking is unavailable in a mria transaction, probably something effectively similar?

Not sure I understand the question, but indeed not having transactions will lead to potential inconsistencies (at least in the builtin backend; FDB could avoid them, at least if deletions are not involved).

apps/emqx_durable_storage/src/emqx_ds.erl
thalesmg: Sounds like a cleaner approach indeed, thanks for the suggestion.  :smile_cat:

apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
keynslug: > This comes from `get_streams` being lenient with failures

Oh I see now. TBH this is (again) feels like a hack: we probably don't want to hide failures on that level.

> Not sure I understand the question

I mean, I'm trying to understand what would happen to the client which is trying to open an existing session when current storage (mria) is unavailable.

thalesmg: > I mean, I'm trying to understand what would happen to the client which is trying to open an existing session when current storage (mria) is unavailable.

Good question, not sure on the exact answer.  Since it uses dirty reads and writes, my guess is that it'll work with the local copy of the data (maybe hang in replicants while they can't RPC a core). :thinking:

apps/emqx_durable_storage/src/emqx_ds.erl
thalesmg: :heavy_check_mark:

apps/emqx/integration_test/emqx_persistent_session_ds_SUITE.erl
thalesmg: Changed the test code to match on the DB.

apps/emqx_management/test/emqx_mgmt_api_clients_SUITE.erl
thalesmg: :heavy_check_mark:

apps/emqx/src/emqx_schema_hooks.erl
```diff
@@ -129,5 +165,68 @@ inject_fields(PointName, Fields) ->
...
+        end,
+        AllInjections,
+        ModuleInjections
+    ).
+
+any_union_option_injections(PointName) ->
+    persistent_term:get(?UNION_OPTION_PT_KEY(PointName), undefined) =/= undefined.
+
+call_if_implemented(Mod, Fun, Args, Default) ->
+    %% ensure modeule is loaded
```
savonarola, line 225:
```suggestion
    %% ensure module is loaded
```

id:
@thalesmg replaced by #13539 and can be closed?

PR #12676, opened 2024-03-11
feat: Add sticky clientid shared subscription strategies

hholstil:
Add new sticky_clientid and sticky_clientid_leastpubs strategies for shared subscriptions.

Release version: v/e5.7.0

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_shared_sub.erl
+++ b/apps/emqx/src/emqx_shared_sub.erl
@@ -80,18 +80,22 @@
     | round_robin
     | round_robin_per_group
     | sticky
+    | sticky_clientid
+    | sticky_prioritize_leastload
     | local
     | hash_clientid
     | hash_topic.
 
 -define(SERVER, ?MODULE).
-
 -define(SHARED_SUB_QOS1_DISPATCH_TIMEOUT_SECONDS, 5).
 -define(IS_LOCAL_PID(Pid), (is_pid(Pid) andalso node(Pid) =:= node())).
 -define(ACK, shared_sub_ack).
 -define(NACK(Reason), {shared_sub_nack, Reason}).
 -define(NO_ACK, no_ack).
 -define(SUBSCRIBER_DOWN, noproc).
+
+%% new publisher rate filter time constant for sticky_prioritize_leastload strategy
+-define(LEASTLOAD_RATE_FILTER_TIME_CONSTANT, 1000.0).
 
 -type redispatch_to() :: ?REDISPATCH_TO(emqx_types:group(), emqx_types:topic()).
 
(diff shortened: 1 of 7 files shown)


apps/emqx/src/emqx_shared_sub.erl
```diff
@@ -72,20 +73,29 @@
...
+    | sticky_clientid
+    | sticky_clientid_leastpubs
     | local
     | hash_clientid
     | hash_topic.
 
 -define(SERVER, ?MODULE).
 -define(TAB, emqx_shared_subscription).
 -define(SHARED_SUBS_ROUND_ROBIN_COUNTER, emqx_shared_subscriber_round_robin_counter).
+-define(SHARED_SUBS_LEASTPUBS_COUNTER, emqx_shared_subscriber_least_pubs_counter).
```
zmstone, line 97: please help to refactor the code to use one ETS table for both.
e.g. rename `emqx_shared_subscriber_round_robin_counter` to `emqx_shared_subscriber_dispatch_counter`.

changes/ce/feat-12676.en.md
```diff
@@ -0,0 +1,9 @@
+Added new shared subscription strategies
+
+- `sticky_clientid`
+
+    Dispatch messages from a publisher (distinguished by their client ID) always to to the same subscriber as long as both the publisher and subscriber sessions exist. The initial subscriber is selected using the hash_clientid strategy. The difference to hash_clientid is that the subscriber associated to a certain publisher will not change even if new subscribers join the group.
```
zmstone, line 5: If I understand the intention of this correctly, you want to make it possible for publisher client to pair up a subscriber client and survive the publisher client reconnect ?

The challenges are
1. The publishing client reconnect will trigger session terminate, and I noticed that session terminate un-sticks the client.
2. When clustered, if a client reconnects, it may connect to another node in the cluster, however the stickiness state is local (in a ETS table).

zmstone: > The main goal is to make sure the subscriber does not change when new subscribers get added to the shared subscription group (or when subscribers are removed, unless it's the current subscriber)

the current sticky strategy should achieve this? Or is it the initial random pick you don’t like?

> I couldn't figure out a better way to make sure that the sticky session data gets cleaned up when it's no longer needed. If the client sets the session expiry interval to a non-zero value the sticky session should stay until that expires regardless of reconnection, right?

process dictionary should do, no need for ets

> Do you think it would make sense to share the sticky state between cluster nodes? What would be the best way to accomplish this?

no, that would be too expensive

zmstone: > The current hash_clientid strategy works well if the subscribers for the group stay constant, but not when they get added/removed (due to kubernetes autoscaling, for example). The hash_clientid strategy picks the Nth subscriber in the list for each message based on the hash but it's not necessarily the same subscriber if the list/count has changed.

I meant the current `sticky` strategy, but did NOT mean the current `hash_clientid` strategy.

> The initial pick is a separate thing, I think it would be nice to have an option for using the subscriber with the least amount of load (which is the goal with the sticky_clientid_leastpubs strategy).

To this, I agree. Maybe worth a separate PR.

> But regardless of which one is used, the sticky state for each client should be cleared when it's no longer needed, right?

Process dictionary goes away when the client process terminates, so there is no need for any cleanup.

zmstone: Hmm. there is perhaps some misunderstanding about how EMQX message dispatch works.
The `sticky` strategy ties to bind a PUBLISH client with a SUBSCRIBE client as long as they are both alive.
If another PUBLISH client sends a message, the `sticky` strategy allows a fresh initialization (random) of the bound.

Same for the process dictionary, the stickiness state is stored in the publisher's process dictionary, but not subscriber.

zmstone: Maybe add a config `initial_sticky_pick = random | hash_clientid | hash_topic | least_loaded`

apps/emqx/src/emqx_shared_sub.erl
```diff
@@ -404,9 +468,12 @@ init([]) ->
     {atomic, PMon} = mria:transaction(?SHARED_SUB_SHARD, fun ?MODULE:init_monitors/0),
     ok = emqx_utils_ets:new(?SHARED_SUBSCRIBER, [protected, bag]),
     ok = emqx_utils_ets:new(?ALIVE_SHARED_SUBSCRIBERS, [protected, set, {read_concurrency, true}]),
-    ok = emqx_utils_ets:new(?SHARED_SUBS_ROUND_ROBIN_COUNTER, [
+    ok = emqx_utils_ets:new(?SHARED_SUBS_DISPATCH_COUNTER, [
         public, set, {write_concurrency, true}
     ]),
+    emqx_hooks:put(
+        'client.disconnected', {?MODULE, on_client_disconnected, [self()]}, ?HP_LOWEST
```
zmstone, line 475: sorry for not thinking it though during previous reviews.
I am afraid this is not quite reliable because a client process can be killed (or crash), in which case, it has no chance to call the hooks.

We would need to monitor the publisher clients, however it would be too expensive to keep track of the monitoring references because there could potentially millions of publisher clients vs only a few subscribers.

Also, in case of clustering, nodes may have different view of which one is least loaded, so I am unsure if this is really going to be more balanced than random or hash based strategy.

zmstone: weighted random might work.
e.g. bump a counter for each pick, for `sticky_prioritize_leastload` strategy, pick with condition `Rand > 1 - (PickCount+1) / (TotalCount+N)`.
Maybe reset the counters every 10 seconds to reflect real time load.

changes/ce/feat-12676.en.md
zmstone: Thank you. I'll have a look soon.

zmstone:
> > Hi @hholstil Thank you for splitting this PR out. And sorry for the delayed review for this one. Now that the initial pick PR is merged, could you please rebase ?
> 
> Thanks for getting the other PR merged. I will rebase this, but unfortunately don't have the time to do it right now so it may take a while.
> 
> I'm not 100% happy with the weights/rate algorithm so I was still thinking about alternatives. Since the shared subscriber clients themselves know how much load they have, could they perhaps periodically publish this information to a (configurable) MQTT topic (by their client id) which the shared subscription handler in EMQX would then internally subscribe? This way the random weights could be updated by the latest status information published by the subscriber clients. I haven't researched how feasible this would be to implement so this is basically just brainstorming, but do you think it would make any sense?

Good idea.
But instead of publishing load data to a topic, maybe it's better to allow subscriber processes to compute their load and insert the load to a replicated table (e.g. `shared_sub_state` which should has key as the pid, and value can be the load/weight)
When picking subscriber per `least_load` strategy, it should sort the pids per their load.

hholstil:
This is outdated and it's no longer possible for me to contribute with the recent licensing changes.

PR #12708, opened 2024-03-14
fix: DynamoDB connector status check takes too long

kjellwinblad:
The DynamoDB connector status checks takes very long when the server is unavailable which makes the resource manager blocked for a long time. This causes calls to update the Bridge config with the Bridge V1 API fail due to a timeout when it calls the resource manager to remove the channel.

A better fix would be to change the resource manager so that the status check cannot block it for a long time. However, this is more complicated so it needs to be done in a later commit. A new ticket has been created for this task https://emqx.atlassian.net/browse/EMQX-12015 .

Fixes:
https://emqx.atlassian.net/browse/EMQX-11984

Fixes <issue-or-jira-number>

Release version: v/e5.?

## PR Checklist

No need for changelog since this fixes unreleased code.

### DIFF
--- a/apps/emqx_bridge_dynamo/test/emqx_bridge_dynamo_SUITE.erl
+++ b/apps/emqx_bridge_dynamo/test/emqx_bridge_dynamo_SUITE.erl
@@ -88,7 +88,9 @@ init_per_suite(Config) ->
 
 end_per_suite(_Config) ->
     emqx_mgmt_api_test_util:end_suite(),
-    ok = emqx_common_test_helpers:stop_apps([emqx_bridge, emqx_resource, emqx_conf, erlcloud]),
+    ok = emqx_common_test_helpers:stop_apps([
+        emqx_rule_engine, emqx_bridge, emqx_resource, emqx_conf, erlcloud
+    ]),
     ok.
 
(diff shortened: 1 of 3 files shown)


apps/emqx_bridge_dynamo/test/emqx_bridge_dynamo_SUITE.erl
```diff
@@ -273,6 +275,24 @@ create_bridge_http(Params) ->
         Error -> Error
     end.
 
+update_bridge_http(#{<<"type">> := Type, <<"name">> := Name} = Config) ->
```
thalesmg, line 278: Nit: there are `get` and `update` helper fns in `emqx_bridge_testlib`.

thalesmg (approved):

apps/emqx_bridge_dynamo/test/emqx_bridge_dynamo_SUITE.erl
kjellwinblad: Thanks! I will remember that next time! I don't think it will cause much harm to have a little bit of code duplication in a test suite so I will merge this now and maybe we can come back and fix this in a later PR.

PR #12739, opened 2024-03-19
feat(sessds): Add zone overrides for session durability settings

ieQu1:
Fixes https://emqx.atlassian.net/browse/EMQX-11928

Release version: v/e5.7

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_config_SUITE.erl
+++ b/apps/emqx/test/emqx_config_SUITE.erl
@@ -463,5 +463,17 @@ zone_global_defaults() ->
                 backoff_new_conn => true,
                 enable => false
             },
-        stats => #{enable => true}
+        stats => #{enable => true},
+        session_persistence =>
+            #{
+                enable => false,
+                batch_size => 100,
+                force_persistence => false,
+                idle_poll_interval => 100,
+                last_alive_update_interval => 5000,
+                message_retention_period => 86400000,
+                renew_streams_interval => 5000,
+                session_gc_batch_size => 100,
+                session_gc_interval => 600000
+            }
     }.
--- a/apps/emqx/src/emqx_persistent_message.erl
+++ b/apps/emqx/src/emqx_persistent_message.erl
@@ -19,9 +19,10 @@
 -behaviour(emqx_config_handler).
 
 -include("emqx.hrl").
+-include_lib("emqx/include/logger.hrl").
 
 -export([init/0]).
--export([is_persistence_enabled/0, force_ds/0]).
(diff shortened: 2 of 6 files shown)


apps/emqx/test/emqx_config_SUITE.erl
```diff
@@ -463,5 +463,17 @@ zone_global_defaults() ->
...
+            #{
+                enable => false,
+                batch_size => 100,
+                force_persistence => false,
+                idle_poll_interval => 100,
+                last_alive_update_interval => 5000,
+                message_retention_period => 86400000,
+                renew_streams_interval => 5000,
+                session_gc_batch_size => 100,
+                session_gc_interval => 600000
```
ieQu1, line 477: GC settings can't be overridden per zone. Shall we move them to a separate config tree?

keynslug: Probably. Perhaps keep them only in the global `session_persistence` schema but not in the zone-level one? Another alternative seems to be having one session storage per zone, which sounds way too involved for the task. 🤔

ieQu1: I don't think it's possible to remove fields of the original record from zone override.

keynslug: > I don't think it's possible to remove fields of the original record from zone override.

At first glance it sounds like a limitation of current implementation rather than something fundamentally impossible.

E.g. what if we have [here](https://github.com/emqx/emqx/blob/46877e979b638ae17ec79464e7af0100ef4e0b72/apps/emqx/src/emqx_zone_schema.erl#L69-L70) another clause that looks like this? So that we could refer to `session_persistence_overridable` record fields instead of the root itself.
```
fields({Name, BaseSchemaName}) ->
    [{N, no_default(Sc)} || {N, Sc} <- emqx_schema:fields(BaseSchemaName)];
```
(Granted `zones_without_default/0` and `global_zones_without_default/0` would also require few changes, but still)

ieQu1: I don't want to make changes to the configuration framework to handle a special case.

ieQu1: Conclusion: we already have certain parameters in the MQTT subtree that are present in the zone override, but don't take any effect (not overridable). Therefore, the current state of the PR is "good enough".

keynslug (approved):

thalesmg (approved):

emqxqa:
[Test Execution]()

PR #12780, opened 2024-03-25
feat(dsrepl): transfer storage snapshot during ra snapshot recovery

keynslug:
Fixes [EMQX-11979](https://emqx.atlassian.net/browse/EMQX-11979).
Fixes [EMQX-12089](https://emqx.atlassian.net/browse/EMQX-12089) (partially).

## Summary

Implement snapshot logic following `ra_snapshot` behaviour that pushes complete RocksDB checkpoint over ra snapshot transfer transport, when leader requests a follower to accept a snapshot and recover from it. This usually happens when the follower lags behind so much that there are no log segments covering the gap anymore.

Even though `ra` actually requests point-in-time snapshot, `emqx_ds_replication_snapshot` currently ignores this and transfers most recent RocksDB checkpoint instead. This should be safe as long as storage layer is idempotent: there are few fixes in this PR to ensure that.

Regular distribution channel is used by `ra` for snapshot transfers, and since they can be quite huge the channel's performance might be affected. If it proves to be an issue, `emqx_ds_storage_snapshot` should hopefully require minimal changes to be able to run in a separate process and utilize side channel for communication with the target server.

## PR Checklist


apps/emqx_durable_storage/src/emqx_ds_storage_snapshot.erl
```diff
@@ -0,0 +1,325 @@
...
+    term_to_binary(Header).
+
+make_packet(Header, Rest) ->
+    HeaderBytes = term_to_binary(Header),
+    <<HeaderBytes/binary, Rest/binary>>.
+
+parse_packet(Packet) ->
+    try binary_to_term(Packet, [safe, used]) of
+        {Header = ?PAT_HEADER(), Length} ->
+            Rest = binary:part(Packet, Length, byte_size(Packet) - Length),
```
thalesmg, line 318: Maybe:

```suggestion
            {_, Rest} = split_binary(Packet, Length),
```

apps/emqx_durable_storage/src/emqx_ds_storage_snapshot.erl
```diff
@@ -0,0 +1,325 @@
...
+}).
+-define(PAT_EOS(), #{
+    '$' := eos
+}).
+
+-define(PAT_HEADER(), #{'$' := _}).
+
+%%
+
+-record(reader, {
```
thalesmg, line 67: Q: is this record sent between nodes?  If so, shouldn't we use maps to reduce problems with rolling upgrades?

apps/emqx_durable_storage/src/emqx_ds_replication_snapshot.erl
```diff
@@ -0,0 +1,229 @@
...
+    accept_chunk/2,
+    complete_accept/2,
+
+    recover/1,
+    validate/1,
+    read_meta/1
+]).
+
+%% Read state.
+-record(rs, {
```
thalesmg, line 39: Q: can this record (and `#ws{}`) be sent over the wire to other nodes?

apps/emqx_durable_storage/src/emqx_ds_replication_snapshot.erl
```diff
@@ -0,0 +1,229 @@
...
+            error(Reason)
+    end.
+
+start_snapshot_writer(WS) ->
+    ShardId = shard_id(WS),
+    logger:info(#{
+        msg => "dsrepl_snapshot_write_started",
+        shard => ShardId
+    }),
+    _ = emqx_ds_builtin_db_sup:terminate_storage(ShardId),
```
thalesmg, line 171: Shouldn't we also terminate the egress for this shard?

Also: some operations don't rely on the storage layer process running (e.g.: `get_streams`, `next`).  Should those be disabled when the shard is down while restoring a snapshot?

apps/emqx_durable_storage/src/emqx_ds_replication_snapshot.erl
```diff
@@ -0,0 +1,229 @@
...
+            ?tp(dsrepl_snapshot_write_complete, #{writer => SnapWriter}),
+            _ = emqx_ds_storage_snapshot:release_writer(SnapWriter),
+            Result = complete_accept(WS#ws{writer = SnapWriter}),
+            ?tp(dsrepl_snapshot_accepted, #{shard => shard_id(WS)}),
+            Result;
+        {error, Reason} ->
+            ?tp(dsrepl_snapshot_write_error, #{reason => Reason, writer => SnapWriter0}),
+            _ = emqx_ds_storage_snapshot:abort_writer(SnapWriter0),
+            error(Reason)
+    end.
```
thalesmg, line 189: Seems like `storage_snapshot:write_chunk/2` might also return `{next, SnapWriter}`.

apps/emqx_durable_storage/src/emqx_ds_storage_snapshot.erl
keynslug: No, only packets are traveling over network (produced by `make_packet/1,2`).

apps/emqx_durable_storage/src/emqx_ds_replication_snapshot.erl
keynslug: Same, they should not.

apps/emqx_durable_storage/src/emqx_ds_replication_snapshot.erl
keynslug: > Shouldn't we also terminate the egress for this shard?

Not really I think. Those have more or less separate lifecycles. Shard as a whole is likely functioning: the shard's leader is on another node, and performing the snapshot transfer in the background. It does not block writes on the leader.

> Should those be disabled when the shard is down while restoring a snapshot?

Ideally, yes. But this is not the only reason why they should, RPCs could arrive during regular node's shutdown and startup. So it's more of a general concern, which I actually started working on separately. However, from safety perspective we likely want to perform reads on leaders as well (because of this safe epoch cutoff in `emqx_ds_storage_bitfield_lts`), so that might not be a big problem anyway eventually (because it's hard to be a leader w/o running storage).

apps/emqx_durable_storage/src/emqx_ds_replication_snapshot.erl
keynslug: This is the last chunk, so it's somewhat unexpected to have `{next, _Writer}` here. Opted for a good old `case_clause` here for now.

apps/emqx_durable_storage/src/emqx_ds_storage_snapshot.erl
```diff
@@ -0,0 +1,325 @@
...
+    case file:read(IoDev, Size) of
+        {ok, Chunk} ->
+            ChunkSize = byte_size(Chunk),
+            HasMore = ChunkSize div Size,
+            RFile1 = RFile0#rfile{pos = Pos + ChunkSize},
+            case ChunkSize < Size of
+                false ->
+                    Status = next,
+                    RFile = RFile1;
+                true ->
```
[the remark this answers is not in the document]
keynslug: Actually, `HasMore` is intentionally a `0 | 1` here. But I'll try to make it look nicer, thanks for the suggestion.

apps/emqx_durable_storage/src/emqx_ds_lts.erl
```diff
@@ -277,7 +279,7 @@ get_id_for_key(#trie{static_key_size = Size}, _State, _Token) ->
     %% If we want to impress computer science crowd, sorry, I mean to
     %% minimize storage requirements, we can even employ Huffman coding
     %% based on the frequency of messages.
-    <<Int:(Size * 8)>> = crypto:strong_rand_bytes(Size),
+    <<Int:(Size * 8), _/bytes>> = crypto:hash(sha256, term_to_binary([State | Token])),
```
ieQu1, line 282: I wonder if it affects performance in any way. IIRC this path is only taken for the new topics?

apps/emqx_durable_storage/test/emqx_ds_replication_SUITE.erl
```diff
@@ -0,0 +1,202 @@
...
+message(Topic, Payload, PublishedAt) ->
+    #message{
+        from = <<?MODULE_STRING>>,
+        topic = Topic,
+        payload = Payload,
+        timestamp = PublishedAt,
+        id = emqx_guid:gen()
+    }.
+
+consume(Node, DB, Shard, TopicFilter, StartTime) ->
```
ieQu1, line 138: I saw implementation of similar function in several places (e.g. `emqx_ds_SUITE`). Move it to a library function?

apps/emqx_durable_storage/src/emqx_ds_replication_snapshot.erl
ieQu1: > Shouldn't we also terminate the egress for this shard?

No. Egress is on the client side. Storage is on the server side.

apps/emqx_durable_storage/src/emqx_ds_replication_snapshot.erl
```diff
@@ -0,0 +1,229 @@
...
+%% Reading a snapshot.
+%% This is triggered by the leader when it finds out that a follower is
+%% behind so much that there are no log segments covering the gap anymore.
+%% This process, on the other hand, MUST involve reading the storage snapshot,
+%% (in addition to the log snapshot) to reconstruct the storage state on the
+%% target node.
+
+-spec begin_read(_SnapshotDir :: file:filename(), _Context :: #{}) ->
+    {ok, ra_snapshot:meta(), rs()} | {error, _Reason :: term()}.
+begin_read(Dir, _Context) ->
```
ieQu1, line 83: Please add a comment answering the following questions:

What entity has logical ownership of the snapshot?
What is its lifetime?
What happens in exceptional situations (e.g. remote followers gets shot down amidst snapshot transfer)? Will snapshot keep lingering?

keynslug: Good point. I expanded a bit on that in ee409ae (also listed a couple of shortcomings / TODOs).

> Is there an API for listing the existing snapshots and their owners?

Not right now, but will be great to have eventually.

ieQu1 (approved):

emqxqa:
[Test Execution]()

PR #12781, opened 2024-03-25
Add metrics for the builtin durable storage.

ieQu1:
Fixes https://emqx.atlassian.net/browse/EMQX-11009

Release version: v/e5.7

## PR Checklist


apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
```diff
@@ -151,80 +154,183 @@ terminate(_Reason, _S) ->
...
-        case N >= NMax of
-            true ->
-                _ = erlang:cancel_timer(S2#s.tref),
-                do_flush(S2);
-            false ->
-                S2
+        #{},
+        Messages
+    ),
+    maps:foreach(
```
ieQu1, line 305: `maps:fold`

apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl
```diff
@@ -510,6 +517,7 @@ traverse_interval(ITHandle, KeyMapper, Filter, Cutoff, Key, Val, It0, Acc0, N) -
                     Acc = [{Key, Msg} | Acc0],
                     traverse_interval(ITHandle, KeyMapper, Filter, Cutoff, It, Acc, N - 1);
                 false ->
+                    inc_counter(?DS_LTS_COLLISION_COUNTER),
```
thalesmg, line 520: Since we're using the same counters to track `delete` operations, we might want to add a `inc_counter(?DS_LTS_COLLISION_COUNTER)` to l. 615.

https://github.com/emqx/emqx/blob/17b3b25b07d1b028b10020232325e00c1e98cf86/apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl#L602-L615

apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
```diff
@@ -151,80 +154,194 @@ terminate(_Reason, _S) ->
...
+    T0 = erlang:monotonic_time(microsecond),
+    Result = emqx_ds_replication_layer:ra_store_batch(DB, Shard, Messages),
+    T1 = erlang:monotonic_time(microsecond),
+    emqx_ds_builtin_metrics:observe_egress_flush_time(Metrics, T1 - T0),
+    case Result of
         ok ->
+            emqx_ds_builtin_metrics:inc_egress_batches(Metrics),
+            emqx_ds_builtin_metrics:inc_egress_messages(Metrics, S#s.n),
+            emqx_ds_builtin_metrics:inc_egress_bytes(Metrics, S#s.n_bytes),
+            lists:foreach(fun(From) -> gen_server:reply(From, ok) end, Replies),
```
thalesmg, line 224: This duplicates l. 229.

apps/emqx/src/emqx_persistent_message.erl
```diff
@@ -114,6 +114,7 @@ needs_persistence(Msg) ->
 
 -spec store_message(emqx_types:message()) -> emqx_ds:store_batch_result().
 store_message(Msg) ->
+    emqx_metrics:inc('messages.persisted'),
```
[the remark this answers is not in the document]
ieQu1: The goal of this counter is to compute the ratio of messages that are _supposed_ to be persisted vs. ones routed normally. Result doesn't matter.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
```diff
@@ -151,80 +154,194 @@ terminate(_Reason, _S) ->
...
+cancel_timer(#s{tref = undefined}) ->
+    ok;
+cancel_timer(#s{tref = TRef}) ->
+    _ = erlang:cancel_timer(TRef),
+    ok.
+
+%% @doc Return approximate size of the MQTT message (it doesn't take
+%% all things into account, for example headers and extras)
+payload_size(#message{payload = P, topic = T}) ->
+    size(P) + size(T).
```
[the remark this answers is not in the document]
ieQu1: To avoid re-running performance tests I chose a "trivially cheap" implementation. 
This can be improved later.

apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl
ieQu1: I was thinking about that, but the main purpose of these counters is to tune the LTS layout parameters. From this perspective, delete and next are the same.

apps/emqx_utils/src/emqx_metrics_worker.erl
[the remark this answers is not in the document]
ieQu1: I want to minimize unrelated changes that go beyond DS and directly affected modules.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
```diff
@@ -151,80 +154,193 @@ terminate(_Reason, _S) ->
...
+do_flush(S0 = #s{n = 0}) ->
+    S0;
 do_flush(
-    S = #s{batch = Messages, pending_replies = Replies, db = DB, shard = Shard}
+    S = #s{queue = Q, pending_replies = Replies, db = DB, shard = Shard, metrics_id = Metrics}
 ) ->
-    case emqx_ds_replication_layer:ra_store_batch(DB, Shard, lists:reverse(Messages)) of
+    Messages = queue:to_list(Q),
+    T0 = erlang:monotonic_time(microsecond),
+    Result = emqx_ds_replication_layer:ra_store_batch(DB, Shard, Messages),
```
keynslug, line 216: Might be a concern for a separate PR, but a discussion here for visibility won't hurt.

Currently `ra_store_batch/3` doesn't actually follow the `emqx_ds:error(_)` contract. Partially overlooked, partially because some conditions do not actually fit "retry or report error" logic, or at least not when it comes to clients.

If we leave out non-important details, performing a state change through `ra` can basically either succeed, fail (for example, when no server are reachable), or _time out_. Timeout is essentially "no knowledge if it succeeded, failed, or will succeed / fail eventually". It could happen for example when leader is alive, persisted the log entry but currently has no quorum to commit. And since we don't know the outcome, and have no idempotency on that level, we (arguably) can neither retry nor report an error to the client.

Admittedly, we can
1. Report such conditions as errors, and let `emqx_broker` decide what to do with them. I'd argue a dedicated error class might be good to have then.
1. Make the logic on that level idempotent, however it sounds quite comprehensive.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
```diff
@@ -151,80 +154,193 @@ terminate(_Reason, _S) ->
...
-            true = erlang:garbage_collect(),
+            erlang:garbage_collect(),
+            S#s{
+                n = 0,
+                n_bytes = 0,
+                queue = queue:new(),
+                pending_replies = []
+            };
+        {error, recoverable, Reason} ->
+            %% Retry sending the batch:
```
keynslug, line 237: Q: Just to clarify the direction. Last time we discussed this time, my understanding of the gist of discussion was: as this egress implementation's goal is not to provide direct feedback to publishers (which we want to be unaffected by persistence suddenly kicking in) the strategy here should be "load shedding", e.g. throwing away parts of the queue when it starts to grow too much, avoiding retries, and even eschewing backpressure. Do you want to adopt a different strategy here?

apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
```diff
@@ -151,80 +154,193 @@ terminate(_Reason, _S) ->
...
-    %% or until the buffer is filled.
-    S2 =
-        case Sync of
-            true ->
-                S1#s{pending_replies = [From | Replies]};
-            false ->
-                gen_server:reply(From, ok),
-                S1
+            S;
+        Err = {error, unrecoverable, _} ->
```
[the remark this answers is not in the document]
ieQu1: I will add a debug log here. Since this is a hot loop, there's a risk of flooding the logs with an enormous number of messages that will hide the actual root cause.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
ieQu1: This is a valid point. I will add a comment to document this discussion, and extend the API to accommodate this usecase. 
BUT the previous implementation didn't achieve this goal either. It would pause processing (blocking the writing process), _then_ it would drop the messages. Which looks like the worst of both worlds.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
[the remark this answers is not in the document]
ieQu1: Done. Added logger domain to the process for finer grained logging filtering.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
ieQu1: Replaced call with a cast for async batches. The previous solution with half-async call was weird.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_egress.erl
ieQu1: Good point. I limited the number of retries to 0 for now.

keynslug (dismissed):

id (approved):

emqxqa:
[Test Execution]()

PR #12798, opened 2024-03-27
feat(client mgmt api): add cursor-based list API

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12028

Release version: v/e5.7

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -38,6 +38,7 @@
 
 -export([
     clients/2,
+    list_clients_v2/2,
     kickout_clients/2,
     client/2,
     subscriptions/2,
@@ -62,6 +63,10 @@
 
 %% for batch operation
 -export([do_subscribe/3]).
+
+-ifdef(TEST).
+-export([parse_cursor/2, serialize_cursor/1]).
+-endif.
 
 -define(TAGS, [<<"Clients">>]).
 
@@ -95,6 +100,14 @@
     message => <<"Client connection has been shutdown">>
 }).
 
+%% tags
+-define(CURSOR_VSN1, 1).
+-define(CURSOR_TYPE_ETS, 1).
+-define(CURSOR_TYPE_DS, 2).
+%% field keys
+-define(CURSOR_ETS_NODE_IDX, 1).
+-define(CURSOR_ETS_CONT, 2).
+
 namespace() -> undefined.
 
 api_spec() ->
@@ -103,6 +116,7 @@ api_spec() ->
 paths() ->
     [
         "/clients",
+        "/clients_v2",
         "/clients/kickout/bulk",
         "/clients/:clientid",
         "/clients/:clientid/authorization/cache",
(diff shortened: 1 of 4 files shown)


apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -890,6 +945,218 @@ list_clients(QString) ->
...
+        false ->
+            done
+    end.
+
+next_ets_cursor(Nodes, NodeIdx, Cont) ->
+    case NodeIdx > length(Nodes) of
+        true ->
+            initial_ds_cursor();
+        false ->
+            Node = lists:nth(NodeIdx, Nodes),
```
savonarola, line 1073: Do nodes preserve count across requests?

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -890,6 +945,218 @@ list_clients(QString) ->
...
+                iterator => init_persistent_session_iterator()
+            };
+        false ->
+            done
+    end.
+
+next_ets_cursor(Nodes, NodeIdx, Cont) ->
+    case NodeIdx > length(Nodes) of
+        true ->
+            initial_ds_cursor();
```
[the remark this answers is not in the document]
thalesmg: I can see that.  :sweat_smile: 

This function is intended to give the next cursor, _given an ETS cursor as input_.  Not sure how to better name it to reflect this.  :thinking:

apps/emqx_management/src/emqx_mgmt_api_clients.erl
thalesmg: No, if the node list changes between requests, the total returned list of clients will change.  But this is the case with the current API as well, IIUC: if the client is asking for page 1, page 2, ..., and then the node list changes before it asks for, say, page 3, it'll also get strange results.

One initial idea I had was to encode the current node and the remaining nodes in the cursor.  But that introduces the problem of having to deal with the change in the cluster in a different manner:  what to do if one of the next nodes is off?  Or a new node is introduced but not in the `next` list?  Since the old API seems to suffer from this in a similar fashion, I left this idea out.  Which also makes the cursor slimmer when serializing it in the query params.  :sweat_smile:

savonarola (approved):

apps/emqx_management/src/emqx_mgmt_api_clients.erl
savonarola: Maybe protect from the case when `NodeId > length(Nodes)`? Just end the iteration, instead of crash?

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -890,6 +945,218 @@ list_clients(QString) ->
...
+next_ds_cursor('$end_of_table') ->
+    done;
+next_ds_cursor(Iter) ->
+    #{
+        type => ?CURSOR_TYPE_DS,
+        iterator => Iter
+    }.
+
+parse_cursor(CursorBin, Nodes) ->
+    try base64:decode(CursorBin, #{mode => urlsafe, padding => false}) of
```
HJianBo, line 1091: it's only supported on OTP 26, but still using OTP 25 in docker images. Maybe we should consider the following funcs?
```
jose_base64url:decode/1,
```
or  `emqx_base62.erl`

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -890,6 +945,218 @@ list_clients(QString) ->
...
+    },
+    do_list_clients_v2(Nodes, Cursor, QString0, Acc).
+
+do_list_clients_v2(_Nodes, Cursor = done, _QString, Acc) ->
+    format_results(Acc, Cursor);
+do_list_clients_v2(Nodes, Cursor = #{type := ?CURSOR_TYPE_ETS, node := Node}, QString0, Acc0) ->
+    {Rows, NewCursor} = do_ets_select(Nodes, QString0, Cursor),
+    Acc1 = maps:update_with(rows, fun(Rs) -> [{Node, Rows} | Rs] end, Acc0),
+    Acc = #{limit := Limit, n := N} = maps:update_with(n, fun(N) -> N + length(Rows) end, Acc1),
+    case N >= Limit of
```
zmstone, line 978: we should try to compare with `>= Limit + 1` or `> Limit` but do not return the last item in the list (and the last item's key should be the next cursor).
this should eliminate the false positive for `hasnext`

thalesmg: I think just the last seen clientid is not enough for a cursor, since we have to i) traverse different nodes which have different first and last clientids; ii) after all ETS tables are traversed, we traverse DS DB, which is not ordered by clientid.  That's why I used the current cursor implementation.

apps/emqx_management/src/emqx_mgmt_api_clients.erl
thalesmg: Surprisingly, `jose_base64url:{encode,decode}` seem to fail roundtrip serialization-deserialization...  🫠

I'll change it to use `emqx_base62`.  Thanks!

thalesmg: https://github.com/emqx/emqx/pull/12848

apps/emqx_management/src/emqx_mgmt_api_clients.erl
thalesmg: On second thought, I think I see now what you meant.  Using just the last clientid is still not enough as a cursor, but we could use that and restart `ets:select` using it instead of holding on to the continuation.  :thinking:

emqxqa:
[Test Execution]()

PR #12800, opened 2024-03-28
sync release-56

id:


savonarola (approved):

apps/emqx/src/emqx_inflight.erl
```diff
@@ -36,8 +36,7 @@
     max_size/1,
     is_full/1,
     is_empty/1,
-    window/1,
-    query/2
+    window/1
```
ieQu1, line 39: @SergeTupchiy is this ok?

ieQu1 (approved):

apps/emqx/src/emqx_inflight.erl
SergeTupchiy: Yes, I moved query to emqx_session_mem as it is now has to convert Inflight tree to list and apply content-based logic (sort by message timestamp), it makes little sense to keep that code in `emqx_inflight`.

thalesmg (approved):

emqxqa:
[Test Execution]()

PR #12811, opened 2024-03-28
ci: restore build tgz in build packages

id:


.github/workflows/build_packages.yaml
```diff
@@ -151,7 +151,23 @@ jobs:
...
+        BUILDER_SYSTEM: force_docker
+      run: |
+        ./scripts/buildx.sh \
+          --profile $PROFILE \
+          --arch $ARCH \
+          --builder $BUILDER \
+          --elixir $IS_ELIXIR \
+          --pkgtype tgz
+    - name: build pkg
+      if: matrix.with_elixir == 'no'
```
thalesmg, line 170: This will exclude the single build that contains Elixir from having a `.deb` package.  Is this intentional?

id: Yes, I replicated previous behavior https://github.com/emqx/emqx/blob/release-55/.github/workflows/build_packages.yaml#L177

```
if [ "${IS_ELIXIR:-}" == 'yes' ]; then
  make "${PROFILE}-elixir-tgz"
else
  make "${PROFILE}-tgz"
  make "${PROFILE}-pkg"
fi
```

Frankly, I'm not sure what was the motivation to not build system packages for emqx with elixir.

thalesmg (approved):

PR #12812, opened 2024-03-28
feat(resource manager): perform non-blocking health checks

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12015

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_manager.erl
+++ b/apps/emqx_resource/src/emqx_resource_manager.erl
@@ -60,6 +60,9 @@
 % Behaviour
 -export([init/1, callback_mode/0, handle_event/4, terminate/3]).
 
+%% Internal exports.
+-export([worker_resource_health_check/1]).
+
 % State record
 -record(data, {
     id,
@@ -73,7 +76,15 @@
     state,
     error,
     pid,
-    added_channels,
+    added_channels = #{},
+    %% Reference to process performing resource health check.
+    hc_workers = #{resource => #{}, channel => #{}} :: #{
+        resource | channel := #{{pid(), reference()} => true}
+    },
+    %% Callers waiting on health check
+    hc_pending_callers = #{resource => [], channel => []} :: #{
+        resource | channel := [gen_server:from()]
+    },
     extra
 }).
 -type data() :: #data{}.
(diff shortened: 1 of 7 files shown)


apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -153,13 +164,13 @@ create(ResId, Group, ResourceType, Config, Opts) ->
     case SpawnBufferWorkers andalso lists:member(QueryMode, [sync, async]) of
         true ->
             %% start resource workers as the query type requires them
-            ok = emqx_resource_buffer_worker_sup:start_workers(ResId, Opts),
-            case maps:get(start_after_created, Opts, ?START_AFTER_CREATED) of
-                true ->
-                    wait_for_ready(ResId, maps:get(start_timeout, Opts, ?START_TIMEOUT));
```
thalesmg, line 159: I think this wait was inside the `true` clause by pure accident: I can't see a reason why we should wait for the resource to be ready only when spawning buffer workers and if the query mode is `{,a}sync`.  :scream_cat: :see_no_evil:

thalesmg:
This fix is enough to fix the [original test case](https://github.com/emqx/emqx/pull/12708) when the Dynamo health check timeout is increased to 30 s, so perhaps we don't need to refactor the health check for the channels just yet.  We may leave it for a future follow up PR if the need arises.

apps/emqx_resource/src/emqx_resource_manager.erl
kjellwinblad: I also don't see any reason so it is probably correct to fix this.

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -865,56 +994,6 @@ get_channel_status_channel_added(#data{id = ResId, mod = Mod, state = State}, Ch
     RawStatus = emqx_resource:call_channel_health_check(ResId, ChannelId, Mod, State),
```
kjellwinblad, line 994: Can the manual channel health check block the resource manager? Can this cause any problems?

kjellwinblad (approved):
Looks good to me

apps/emqx_resource/src/emqx_resource_manager.erl
thalesmg: https://github.com/emqx/emqx/pull/12830

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -835,26 +855,135 @@ handle_not_connected_and_not_connecting_remove_channel(From, ChannelId, Data) ->
...
+    {keep_state, Data}.
+
+-spec spawn_health_check_worker(data()) -> {pid(), reference()}.
+spawn_health_check_worker(#data{} = Data) ->
+    spawn_monitor(?MODULE, worker_resource_health_check, [Data]).
+
+%% separated so it can be spec'ed and placate dialyzer tantrums...
+-spec worker_resource_health_check(data()) -> no_return().
+worker_resource_health_check(Data) ->
+    HCRes = emqx_resource:call_health_check(Data#data.id, Data#data.mod, Data#data.state),
```
HJianBo, line 903: Q: Is it necessary to use spawn_link here so that the health_checker process doesn't leak if emqx_resource_manager crash.

thalesmg: https://github.com/emqx/emqx/pull/12847

zmstone: terminate/2 is maybe not evaluated if the process crashes.

emqxqa:
[Test Execution]()

PR #12825, opened 2024-04-02
feat(dsrepl): add APIs to manage DB replication sites

keynslug:
Fixes [EMQX-12110](https://emqx.atlassian.net/browse/EMQX-12110).

## Summary

Basic operation is `assign_db_sites/2` that could be employed in both lazy and transactional fashion, where the latter could be optimized for minimal set of membership changes.

## PR Checklist

- [ ] ~~Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files~~

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
@@ -49,6 +49,7 @@
     assign_db_sites/2,
     replica_set_transitions/2,
     update_replica_set/3,
+    db_sites/1,
     replica_set/2,
     target_set/2
 ]).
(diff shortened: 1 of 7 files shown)


apps/emqx_durable_storage/src/emqx_ds_replication_layer_meta.erl
```diff
@@ -86,15 +101,20 @@
 
 -record(?SHARD_TAB, {
     shard :: {emqx_ds:db(), emqx_ds_replication_layer:shard_id()},
+    %% Sites that currently contain the data:
+    replica_set :: [site()],
     %% Sites that should contain the data when the cluster is in the
     %% stable state (no nodes are being added or removed from it):
-    replica_set :: [site()],
+    target_set :: [site()] | undefined,
```
thalesmg, line 108: Just checking: since the feature is not GA yet, we don't need to provide a table migration for this new field even if someone is experimenting with it, right?

keynslug: Good question. I assume we don't.

ieQu1 (approved):

emqxqa:
[Test Execution]()

PR #12827, opened 2024-04-03
Add rule ID tracing support

kjellwinblad:
This PR:

* Adds ruleid as a new tracing type
* Adds test cases for the ruleid tracing type
* Adds the rule_ids key to the logger process meta data before calling the connector module so that traces in the connector instance will be associated with the appropriate rule id
* Adds rule id tracing for the HTTP connector (templets rendered successfully trace)

Fixes https://emqx.atlassian.net/browse/EMQX-12025 (partially)

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_trace/emqx_trace.erl
+++ b/apps/emqx/src/emqx_trace/emqx_trace.erl
@@ -66,6 +66,9 @@
 -export_type([ip_address/0]).
 -type ip_address() :: string().
 
+-export_type([ruleid/0]).
+-type ruleid() :: binary().
+
 publish(#message{topic = <<"$SYS/", _/binary>>}) ->
     ignore;
 publish(#message{from = From, topic = Topic, payload = Payload}) when
@@ -517,6 +520,9 @@ to_trace(#{type := ip_address, ip_address := Filter} = Trace, Rec) ->
         Error ->
             Error
     end;
+to_trace(#{type := ruleid, ruleid := Filter} = Trace, Rec) ->
+    Trace0 = maps:without([type, ruleid], Trace),
+    to_trace(Trace0, Rec#?TRACE{type = ruleid, filter = Filter});
 to_trace(#{type := Type}, _Rec) ->
     {error, io_lib:format("required ~s field", [Type])};
 to_trace(#{payload_encode := PayloadEncode} = Trace, Rec) ->
--- a/apps/emqx/src/emqx_trace/emqx_trace.erl
+++ b/apps/emqx/src/emqx_trace/emqx_trace.erl
@@ -28,7 +28,8 @@
     subscribe/3,
     unsubscribe/2,
     log/3,
-    log/4
+    log/4,
+    rendered_action_template/2
 ]).
 
 -export([
@@ -85,6 +86,25 @@ unsubscribe(<<"$SYS/", _/binary>>, _SubOpts) ->
     ignore;
 unsubscribe(Topic, SubOpts) ->
     ?TRACE("UNSUBSCRIBE", "unsubscribe", #{topic => Topic, sub_opts => SubOpts}).
+
+rendered_action_template(ActionID, RenderResult) ->
+    Msg = io_lib:format("action_template_rendered(~s)", [ActionID]),
+    TraceResult = ?TRACE("QUERY_RENDER", Msg, RenderResult),
+    case logger:get_process_metadata() of
+        #{stop_action_after_render := true} ->
+            %% We throw an unrecoverable error to stop action before the
+            %% resource is called/modified
+            StopMsg = io_lib:format(
+                "action_stopped_after_render(~s): "
+                "Action stopped after template render due to test setting.",
+                [ActionID]
+            ),
+            MsgBin = iolist_to_binary(StopMsg),
+            error({unrecoverable_error, MsgBin});
+        _ ->
+            ok
+    end,
+    TraceResult.
 
 log(List, Msg, Meta) ->
     log(debug, List, Msg, Meta).
--- a/apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
+++ b/apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
@@ -661,13 +661,22 @@ process_request_and_action(Request, ActionState, Msg) ->
     ),
     BodyTemplate = maps:get(body, ActionState),
     Body = render_request_body(BodyTemplate, RenderTmplFunc, Msg),
-    #{
+    RenderResult = #{
         method => Method,
         path => Path,
         body => Body,
         headers => Headers,
         request_timeout => maps:get(request_timeout, ActionState)
-    }.
+    },
+    ?TRACE(
+        "QUERY_RENDER",
+        "http_connector_successfully_rendered_request",
+        #{
+            request => Request,
+            render_result => RenderResult
+        }
+    ),
+    RenderResult.
 
 merge_proplist(Proplist1, Proplist2) ->
     lists:foldl(
--- a/apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
+++ b/apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
@@ -371,6 +371,29 @@ on_query(
         }
     ),
     NRequest = formalize_request(Method, BasePath, Request),
+    case NRequest of
+        {Path, Headers} ->
+            emqx_trace:rendered_action_template(
+                InstId,
+                #{
+                    path => Path,
+                    method => Method,
+                    headers => Headers,
+                    timeout => Timeout
+                }
+            );
+        {Path, Headers, Body} ->
+            emqx_trace:rendered_action_template(
+                InstId,
+                #{
+                    path => Path,
+                    method => Method,
+                    headers => Headers,
+                    timeout => Timeout,
+                    body => Body
+                }
+            )
+    end,
     Worker = resolve_pool_worker(State, KeyOrNum),
     Result0 = ehttpc:request(
         Worker,
@@ -480,6 +503,29 @@ on_query_async(
         }
     ),
     NRequest = formalize_request(Method, BasePath, Request),
+    case NRequest of
+        {Path, Headers} ->
+            emqx_trace:rendered_action_template(
+                InstId,
+                #{
+                    path => Path,
+                    method => Method,
+                    headers => Headers,
+                    timeout => Timeout
+                }
+            );
+        {Path, Headers, Body} ->
+            emqx_trace:rendered_action_template(
+                InstId,
+                #{
+                    path => Path,
+                    method => Method,
+                    headers => Headers,
+                    timeout => Timeout,
+                    body => Body
+                }
+            )
+    end,
     MaxAttempts = maps:get(max_attempts, State, 3),
     Context = #{
         attempt => 1,
@@ -661,22 +707,13 @@ process_request_and_action(Request, ActionState, Msg) ->
     ),
     BodyTemplate = maps:get(body, ActionState),
     Body = render_request_body(BodyTemplate, RenderTmplFunc, Msg),
-    RenderResult = #{
+    #{
         method => Method,
         path => Path,
         body => Body,
         headers => Headers,
         request_timeout => maps:get(request_timeout, ActionState)
-    },
-    ?TRACE(
-        "QUERY_RENDER",
-        "http_connector_successfully_rendered_request",
-        #{
-            request => Request,
-            render_result => RenderResult
-        }
-    ),
-    RenderResult.
+    }.
 
 merge_proplist(Proplist1, Proplist2) ->
     lists:foldl(
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -64,8 +64,10 @@
 
 -define(COLLECT_REQ_LIMIT, 1000).
 -define(SEND_REQ(FROM, REQUEST), {'$send_req', FROM, REQUEST}).
--define(QUERY(FROM, REQUEST, SENT, EXPIRE_AT), {query, FROM, REQUEST, SENT, EXPIRE_AT}).
--define(SIMPLE_QUERY(FROM, REQUEST), ?QUERY(FROM, REQUEST, false, infinity)).
+-define(QUERY(FROM, REQUEST, SENT, EXPIRE_AT, TRACE_CTX),
+    {query, FROM, REQUEST, SENT, EXPIRE_AT, TRACE_CTX}
+).
+-define(SIMPLE_QUERY(FROM, REQUEST, TRACE_CTX), ?QUERY(FROM, REQUEST, false, infinity, TRACE_CTX)).
 -define(REPLY(FROM, SENT, RESULT), {reply, FROM, SENT, RESULT}).
 -define(INFLIGHT_ITEM(Ref, BatchOrQuery, IsRetriable, AsyncWorkerMRef),
     {Ref, BatchOrQuery, IsRetriable, AsyncWorkerMRef}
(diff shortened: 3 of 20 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -64,8 +64,10 @@
 
 -define(COLLECT_REQ_LIMIT, 1000).
 -define(SEND_REQ(FROM, REQUEST), {'$send_req', FROM, REQUEST}).
--define(QUERY(FROM, REQUEST, SENT, EXPIRE_AT), {query, FROM, REQUEST, SENT, EXPIRE_AT}).
--define(SIMPLE_QUERY(FROM, REQUEST), ?QUERY(FROM, REQUEST, false, infinity)).
+-define(QUERY(FROM, REQUEST, SENT, EXPIRE_AT, RULE_ID),
+    {query, FROM, REQUEST, SENT, EXPIRE_AT, RULE_ID}
```
zmstone, line 68: It should be a more generic trace context instead of RULE_ID.
```suggestion
    {query, FROM, REQUEST, SENT, EXPIRE_AT, TRACE_CONETXT}
```
`TRACE_CONTEXT` is either `undefined` (when no matching trace), or a map with the matching IDs (e.g. `clientid` and `ruleid`).

zmstone (changes requested):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
kjellwinblad: There is never more than one rule_id per `?QUERY` but I think its worth making it more easy to extend anyway. For example, we need to forward information about if the action should be stopped after the template is rendered and I think this info can be passed in the `TRACE_CONTEXT` as well.

kjellwinblad: > i.e. one should be able to trace rules with clientid fitler.

Ok, I see, so I will add client_id to the trace message's meta data.

apps/emqx_rule_engine/src/emqx_rule_runtime.erl
```diff
@@ -77,48 +85,80 @@ apply_rule(Rule = #{id := RuleID}, Columns, Envs) ->
...
+                    stacktrace => StkTrace
+                },
+                warning
+            ),
             {error, {Error, StkTrace}}
+    after
+        reset_process_trace_metadata(Columns)
     end.
 
+set_process_trace_metadata(RuleID, #{clientid := ClientID}) ->
```
zmstone, line 142: we can try to avoid using process dictionary for this?
we can maybe try to check if current rule is being traced or not directly by calling emqx_trace module
if not being traced, log nothing, if it's being traced emqx_trace directly to log it.

kjellwinblad: > we can try to avoid using process dictionary for this?

We can but I don't think that makes the code better. This would mean that we need to pass the `ClientID` to calls that do tracing. I checked other `?TRACE(...)` calls and it seems common to pass tracing meta data by setting it implicitly using `logger:update_process_metadata(...)`. Maybe it is good to not clutter the code more than necessary by adding parameters that are just needed for tracing?

> we can maybe try to check if current rule is being traced or not directly by calling emqx_trace module
if not being traced, log nothing, if it's being traced emqx_trace directly to log it.

If we don't need to do this optimization (e.g., for performance reason), I think it is better to leave what will actually be logged to the logger/tracing framework (logger). All added trace log messages have level debug and if my understanding is correct they will thrown away unless there is a matching trace, or the user has set the global log level to debug. Also remember that since we have now added clientid to the metadata we might want a trace message to show up in a trace even though the rule is not traced.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
kjellwinblad: @zmstone  As you suggested, I have made this field into TRACE_CONTEXT and added clientid so that that rules can be traced using the clientid filter.

kjellwinblad:
Documentation Jira ticket: https://emqx.atlassian.net/browse/EDOC-269

apps/emqx_rule_engine/src/emqx_rule_runtime.erl
```diff
@@ -411,13 +460,47 @@ do_handle_action(
...
     %% the function can also throw 'out_of_service'
     Args = maps:get(args, Action, []),
     Result = Mod:Func(Selected, Envs, Args),
-    inc_action_metrics(RuleId, Result),
+    TraceCtx = do_handle_action_get_trace_context(Action),
+    inc_action_metrics(TraceCtx, Result),
+    trace_action(Action, "call_action_function_result", #{result => Result}, debug),
     Result.
 
+do_handle_action_get_trace_context(Action) ->
```
zmstone, line 473: we should try NOT to build trace context (return `undefined`) if there is currently no ongoing trace (even better if there is no matching trace).

kjellwinblad: Thanks for the suggestion. Can you explain why you think this optimization is needed? Performance reasons?

rel/i18n/emqx_rule_api_schema.hocon
```diff
@@ -66,6 +66,18 @@ test_context.desc:
 test_context.label:
 """Event Conetxt"""
 
+test_rule_environment.desc:
+"""The environment that will be passed to the rule when it is applied. A default environment will be used if no environment is given."""
```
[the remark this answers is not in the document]
kjellwinblad: After studying the code closer I have figured out that the the only time the environment is used for anything is when a custom action function is used and then it is merged together with the input context/columns before it is passed to the function as a parameter. So I think the best is to remove the environment parameter from the API. If we figure out that it is needed for anything later we can always add the parameter.

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -83,6 +87,25 @@ unsubscribe(<<"$SYS/", _/binary>>, _SubOpts) ->
 unsubscribe(Topic, SubOpts) ->
     ?TRACE("UNSUBSCRIBE", "unsubscribe", #{topic => Topic, sub_opts => SubOpts}).
 
+rendered_action_template(ActionID, RenderResult) ->
+    Msg = io_lib:format("action_template_rendered(~s)", [ActionID]),
```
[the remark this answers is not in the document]
kjellwinblad: Fixed

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -83,6 +87,25 @@ unsubscribe(<<"$SYS/", _/binary>>, _SubOpts) ->
...
+    TraceResult = ?TRACE("QUERY_RENDER", Msg, RenderResult),
+    case logger:get_process_metadata() of
+        #{stop_action_after_render := true} ->
+            %% We throw an unrecoverable error to stop action before the
+            %% resource is called/modified
+            StopMsg = io_lib:format(
```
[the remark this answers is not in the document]
kjellwinblad: Fixed

apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
```diff
@@ -371,6 +371,29 @@ on_query(
...
     ),
     NRequest = formalize_request(Method, BasePath, Request),
+    case NRequest of
+        {Path, Headers} ->
+            emqx_trace:rendered_action_template(
+                InstId,
+                #{
+                    path => Path,
+                    method => Method,
+                    headers => Headers,
```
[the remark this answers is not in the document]
kjellwinblad: Fixed

apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
```diff
@@ -371,6 +371,29 @@ on_query(
...
+                    timeout => Timeout
+                }
+            );
+        {Path, Headers, Body} ->
+            emqx_trace:rendered_action_template(
+                InstId,
+                #{
+                    path => Path,
+                    method => Method,
+                    headers => Headers,
```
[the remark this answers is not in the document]
kjellwinblad: Fixed

apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
```diff
@@ -480,6 +503,29 @@ on_query_async(
...
```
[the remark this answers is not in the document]
kjellwinblad: Fixed

apps/emqx_rule_engine/test/emqx_rule_engine_SUITE.erl
```diff
@@ -154,13 +158,13 @@ init_per_suite(Config) ->
     emqx_rule_funcs_demo:module_info(),
     application:load(emqx_conf),
     ok = emqx_common_test_helpers:start_apps(
-        [emqx_conf, emqx_rule_engine, emqx_auth, emqx_bridge],
+        [emqx, emqx_conf, emqx_rule_engine, emqx_auth, emqx_bridge],
```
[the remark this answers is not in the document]
kjellwinblad: Good point, I think I added it as I thought it was needed to start the trace server but it is the emqx_modules application that starts the trace server.

apps/emqx_rule_engine/src/emqx_rule_runtime.erl
kjellwinblad: Fixed

kjellwinblad:
Thanks @zmstone for good comments. I have addressed all of them in the latest commit.

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -83,6 +87,27 @@ unsubscribe(<<"$SYS/", _/binary>>, _SubOpts) ->
...
+            %% We throw an unrecoverable error to stop action before the
+            %% resource is called/modified
+            StopMsg = lists:flatten(
+                io_lib:format(
+                    "action_stopped_after_render(~ts): "
+                    "Action stopped after template render due to test setting.",
+                    [ActionID]
+                )
+            ),
+            MsgBin = iolist_to_binary(StopMsg),
```
zmstone, line 104: nit: for "~ts", it is not guaranteed to be iolist. must call unicode:characters_to_binary instead.
however, given that ActionID is restricted string, it's ok.

last time someone had to find all the ~s in the code base, change to ~ts, if we take care of this in general, should save some time when this has to be done again.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1093,11 +1104,80 @@ call_query(QM, Id, Index, Ref, Query, QueryOpts) ->
...
+    ClientIDs = lists:foldl(fun collect_client_id/2, #{}, Requests),
+    StopAfterRender = lists:foldl(fun collect_stop_after_render/2, no_info, Requests),
+    StopAfterRenderVal =
+        case StopAfterRender of
+            only_true ->
+                logger:update_process_metadata(#{stop_action_after_render => false}),
+                true;
+            only_false ->
+                false;
+            mixed ->
```
zmstone, line 1130: can modify do_collect_requests to avoid collecting mixed requests in one batch

kjellwinblad: > can modify do_collect_requests to avoid collecting mixed requests in one batch

Sure, this is possible but is it something that we want to do? This might introduce complex logic and some overhead for something that will be very rare in practice (at least if we document that this problem can arise when setting stop_action_after_render).

If we want it, does it need to be implemented before we merge this PR or can it be introduced later?

apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
```diff
@@ -480,6 +503,29 @@ on_query_async(
...
     ),
     NRequest = formalize_request(Method, BasePath, Request),
+    case NRequest of
+        {Path, Headers} ->
+            emqx_trace:rendered_action_template(
+                InstId,
+                #{
+                    path => Path,
+                    method => Method,
+                    headers => Headers,
```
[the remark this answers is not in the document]
kjellwinblad: My, bad sorry.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
zmstone: I’m sure it will be less complex than this collect logic. 
the batch collection is pretty straightforward.

zmstone: Better to fix it before merge it

kjellwinblad: Ok, I will look into it.

kjellwinblad: Thanks for the discussion. Fixed here [cf56050](https://github.com/emqx/emqx/pull/12827/commits/cf56050759d96806b1e1cb39a564cdf7f3acbb44)

kjellwinblad:
@zmstone  This commit tries to fix the problems with the trace entries that you brought up during the demo: https://github.com/emqx/emqx/pull/12827/commits/6cf29ba68883f425b8101b12b22fd213231bbd98

(Formatting the rule_ids map as a list in the trace will be fixed later in the JSON trace formatter PR)

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -619,6 +623,73 @@ flush(Data0) ->
             end
     end.
 
+stop_before_mixed_stop_after_render(
```
zmstone, line 626: maybe this is more readable:
```
stop_before_mixed_stop_after_render(?QUERY(_, _, _, _, TraceCtx), State) ->
    do_stop_before_mixed_stop_after_render(TraceCtx, State).

do_stop_before_mixed_stop_after_render(TraceCtx, initial_state) ->
    do_stop_before_mixed_stop_after_render(TraceCtx, get_stop_flag(TraceCtx));
do_stop_before_mixed_stop_after_render(TraceCtx, StopFlag) ->
    case get_stop_flag(TraceCtx) =:= StopFlag  of
        true ->
            StopFlag;
        false ->
            true
    end.

get_stop_flag(#{stop_action_after_render := true}) -> stop_action_after_render;
get_stop_flag(_) -> no_stop_action_after_render.
```

kjellwinblad: Thanks, I like your suggestion. I think it needs some small fixes and I will change the name to stop_batching.

kjellwinblad: Refactored thanks to your suggestion: [b16b9d8](https://github.com/emqx/emqx/pull/12827/commits/b16b9d8fcc3a4a3ca36f0c30f341c88995d8470c). Thanks, I think it is much more readable now.

zmstone (approved):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1093,11 +1125,53 @@ call_query(QM, Id, Index, Ref, Query, QueryOpts) ->
         {ok, _Group, #{status := ?status_connecting, error := unhealthy_target}} ->
             {error, {unrecoverable_error, unhealthy_target}};
         {ok, _Group, Resource} ->
-            do_call_query(QM, Id, Index, Ref, Query, QueryOpts, Resource);
+            set_rule_id_trace_meta_data(Query),
+            QueryResult = do_call_query(QM, Id, Index, Ref, Query, QueryOpts, Resource),
+            %% do_call_query does not throw an exception as the call to the
+            %% resource is wrapped in a try catch expression so we will always
+            %% unset the trace meta data
+            unset_rule_id_trace_meta_data(),
```
zmstone, line 1133: do this in a try after.

emqxqa:
[Test Execution]()

PR #12828, opened 2024-04-03
feat(dsrepl): move snapshot state to manager process

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12143

Release version: v/e5.7

## PR Checklist

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds_snapshot_manager.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_snapshot_manager.erl
@@ -1,0 +1,361 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2024 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+-module(emqx_ds_snapshot_manager).
+
+-behaviour(gen_server).
+
+%% API
+-export([
+    start_link/2,
+
+    begin_read/2,
+    read_chunk/3,
+
+    begin_accept/2,
+    accept_chunk/2,
+    complete_accept/2
+]).
+
+%% `gen_server' API
+-export([
+    init/1,
+
+    handle_call/3,
+    handle_cast/2,
+    handle_info/2
+]).
+
+-include_lib("snabbkaffe/include/trace.hrl").
+
+%%------------------------------------------------------------------------------
+%% Type declarations
+%%------------------------------------------------------------------------------
+
+-define(via(SHARD_ID), {via, gproc, {n, l, {?MODULE, SHARD_ID}}}).
+
+-type shard_id() :: {emqx_ds:db(), emqx_ds_replication_layer:shard_id()}.
+
+-type state() :: #{
+    rs := rs() | undefined,
+    ws := ws() | undefined
+}.
+
+-type ra_state() :: emqx_ds_replication_layer:ra_state().
+
+%% Read state.
+-record(rs, {
+    phase :: machine_state | storage_snapshot,
+    started_at :: _Time :: integer(),
+    state :: ra_state() | undefined,
+    reader :: emqx_ds_storage_snapshot:reader() | undefined
+}).
+
+%% Write state.
+-record(ws, {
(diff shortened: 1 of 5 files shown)


apps/emqx_durable_storage/src/emqx_ds_snapshot_manager.erl
```diff
@@ -0,0 +1,428 @@
...
+    gen_server:start_link(?via({DB, ShardId}), ?MODULE, #{db => DB, shard_id => ShardId}, []).
+
+-spec begin_read(_SnapshotDir :: file:filename(), _Context :: #{}) ->
+    {ok, ra_snapshot:meta(), ext_rs()} | {error, _Reason :: term()}.
+begin_read(Dir, _Context) ->
+    RS0 = #rs{
+        phase = machine_state,
+        started_at = erlang:monotonic_time(millisecond)
+    },
+    case ra_log_snapshot:recover(Dir) of
```
keynslug, line 109: Now there's replication-level concerns trickled on that level for some reason. Should be much clearer if anything related to `ra_snapshot` behaviour implementation stays in the respective module.

apps/emqx_durable_storage/src/emqx_ds_snapshot_manager.erl
```diff
@@ -0,0 +1,428 @@
...
+cleanup_reader(State) ->
+    State.
+
+%% Writer (follower state)
+
+-spec handle_accept_chunk(binary(), state()) ->
+    {ok | {error, _Reason :: term()}, state()}.
+handle_accept_chunk(Chunk, #{ws := WS0 = #ws{phase = reader_caller_proc}} = State0) ->
+    ReaderPid = binary_to_term(Chunk),
+    MRef = monitor(process, ReaderPid),
```
keynslug, line 339: I'm not really against such approach, but:
1. We must have the guarantee that this DOWN signal won't arrive _before_ last chunk.
2. This has questionable forward-compatibility (probably need to wrap it in some kind of well-defined packet?).

apps/emqx_durable_storage/src/emqx_ds_snapshot_manager.erl
```diff
@@ -0,0 +1,428 @@
...
+%%--------------------------------------------------------------------
+-module(emqx_ds_snapshot_manager).
+
+-behaviour(gen_server).
+
+%% API
+-export([
+    start_link/2,
+
+    begin_read/2,
```
ieQu1, line 24: Snapshot manager should just do garbage collection of RocksdB snapshots.

apps/emqx_durable_storage/src/emqx_ds_snapshot_manager.erl
```diff
@@ -0,0 +1,428 @@
...
+        db => DB,
+        shard_id => ShardId,
+        domain => [ds, storage_layer, snapshot]
+    }),
+    clear_all_checkpoints(DB, ShardId),
+    {ok, init_state()}.
+
+terminate(_Reason, State0) ->
+    State1 = cleanup_writer(State0),
+    _ = cleanup_reader(State1),
```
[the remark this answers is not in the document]
thalesmg: True, didn't account for the possibility of more than one reader for a shard.  :see_no_evil: 

But we should have only one writer per shard, since there's no concurrency there.

apps/emqx_durable_storage/src/emqx_ds_snapshot_manager.erl
thalesmg: > We must have the guarantee that this DOWN signal won't arrive before last chunk.

Great point.

> This has questionable forward-compatibility (probably need to wrap it in some kind of well-defined packet?).

I'm not seeing what makes this chunk more special than the others :see_no_evil: .  You mean serializing the PID is unsafe somehow?

apps/emqx_durable_storage/src/emqx_ds_snapshot_manager.erl
```diff
@@ -0,0 +1,428 @@
...
+        dir = Dir,
+        meta = Meta
+    },
+    ExtWS = #{phase => machine_state, ws => WS},
+    {ok, ExtWS}.
+
+-spec accept_chunk(binary(), ext_ws()) ->
+    {ok, ext_ws()} | {error, _Reason :: term()}.
+accept_chunk(Chunk, ExtWS = #{shard_id := ShardId}) ->
+    case gen_server:call(?via(ShardId), #accept_chunk{chunk = Chunk}, infinity) of
```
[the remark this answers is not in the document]
thalesmg: There was no enforcing at this point.  I just moved most of the existing logic and ownership over the files to this module from the original one.  Would it be possible that `ra` would trigger more than one snapshot transfer on the same follower shard?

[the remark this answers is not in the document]
keynslug: > There was no enforcing at this point.

Yeah, I know, it's also a shortcoming of the original implementation.

> Would it be possible for `ra` to trigger more than one snapshot transfer on the same follower shard?

If you mean "initiate a snapshot writer", then I believe this is possible, this could (probably need to check 🫠) happen when leader changes while transfer is in progress.

PR #12861, opened 2024-04-11
test: delete flaky test for now

zmstone:
It's identical test as influxdb, but flaky for greptimedb
probably not worth the time investigating because the test mostly covers non-driver related logic: the buffer worker's return value assertions.


ieQu1 (approved):

emqxqa:
[Test Execution]()

PR #12867, opened 2024-04-11
refactor: do not generate i18n msgid for dashboard schema json

zmstone:
Targeting release: 5.7.0

Originally the special message of format`$msgid:namespace.id.tag` was intended for dashboard to render the UI with a generic i18n lookup with the provided ID. However, frontend team have their own plan.
This means labels and descriptions are maintained in two different copies, however there is no plan to unify.

After this change, I18n is only used to generate docs:
1. runtime swagger spec at the path `/api-docs`, which is also generated during build time and exported as https://docs.emqx.com/en/enterprise/v5.6/admin/api-docs.html
2. build-time config manual doc generation, which is exported as https://docs.emqx.com/en/enterprise/v5.6.0/hocon/
Both result in the resolved i18n text, but not msgid

No release note because this only affects dashboard which does not use the removed schema field anyway.
Nobody else is using the schemas.

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
@@ -81,7 +81,7 @@
     ])
 ).
 
--define(SPECIAL_LANG_MSGID, <<"$msgid">>).
+-define(NO_I18N, undefined).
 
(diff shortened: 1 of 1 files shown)


ieQu1 (approved):

emqxqa:
[Test Execution]()

PR #12750, opened 2024-03-20
0316 introduce client_attrs

zmstone:
Fixes https://emqx.atlassian.net/browse/EMQX-11882

Release version: v/e5.7.0

## Summary

Added `client_attrs` field to `clientinfo` object.
- `client_attrs` can be initialized with a regexp based extraction to set one `client_attrs.NAME` field.
- `client_attrs` can also be initialized from CONNECT packet's User-Property
- If there is a `client_attrs` field found in authentication results, it is merged to clientinfo `client_attrs` object
  - So far only HTTP auth and JWT auth supports extending `client_attrs`
- The placeholder `${client_attrs.NAME}` is allowed in authentication and authorization templates

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [x] Added tests for the changes
- [x] Added property-based tests for code which performs user input validation
- [ ] Changed lines covered in coverage report
- [x] Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [ ] For internal contributor: there is a jira ticket to track this change
- [ ] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket
- [ ] Schema changes are backward compatible

## Checklist for CI (.github/workflows) changes

- [ ] If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)
- [ ] Change log has been added to `changes/` dir for user-facing artifacts update

PR #12872, opened 2024-04-12
0412 use variform for client_attrs

zmstone:
Fixes [EMQX-11882](https://emqx.atlassian.net/browse/EMQX-11882)
This PR rewrites https://github.com/emqx/emqx/pull/12750 which is not released yet.

Release version: v/e5.7.0

## Summary

- Allow multiple `client_attrs` initial extractions
- Support using variform for `client_attrs` initial extraction
- Provide regexp based extraction as a variform function (so the config schema introduced in #12750 is rewritten)

## PR Checklist

[EMQX-11882]: https://emqx.atlassian.net/browse/EMQX-11882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_utils/src/emqx_variform_bif.erl
+++ b/apps/emqx_utils/src/emqx_variform_bif.erl
@@ -14,13 +14,11 @@
 %% limitations under the License.
 %%--------------------------------------------------------------------
 
-%% Predefined functions string templating
--module(emqx_variform_str).
+%% Predefined functions for variform expressions.
+-module(emqx_variform_bif).
 
 %% String Funcs
 -export([
-    coalesce/1,
-    coalesce/2,
     lower/1,
     ltrim/1,
     ltrim/2,
@@ -47,14 +45,21 @@
     replace/4,
     regex_match/2,
     regex_replace/3,
+    regex_extract/2,
     ascii/1,
     find/2,
     find/3,
     join_to_string/1,
     join_to_string/2,
     unescape/1,
-    nth/2
+    any_to_str/1
 ]).
+
+%% Array functions
+-export([nth/2]).
+
+%% Control functions
+-export([coalesce/1, coalesce/2]).
 
 -define(IS_EMPTY(X), (X =:= <<>> orelse X =:= "" orelse X =:= undefined)).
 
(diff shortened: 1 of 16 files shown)


terry-xiaoyu (approved):

apps/emqx_utils/src/emqx_variform_bif.erl
```diff
@@ -47,15 +45,37 @@
...
     unescape/1,
-    nth/2
+    any_to_str/1
 ]).
 
+%% Array functions
+-export([nth/2]).
+
+%% Control functions
+-export([coalesce/1, coalesce/2]).
```
savonarola, line 62: Wondering: don't we want to make `coalesce` lazy?

Also, a question: how is a customer intended to emulate `IFF(.., .., ...)` (functional if)? Via `coalesce`?

savonarola (approved):

apps/emqx_utils/src/emqx_variform_bif.erl
zmstone: > Wondering: don't we want to make coalesce lazy?
Good idea, maybe for improvement. The `eval` loop can be enhanced.

> emulate IFF(cond, a, b)

It's easy: `coalesce([coalesce([a, '']), b])`

zmstone: can add a `IFF` function

zmstone: To make it more complete, we can:

- `coalesce` should be a reserved word in the expression token scanner
- will need to add `iif` function (reserved word too) for if this else that
- to support `iif`, we'll need more compare functions, such as `gt`, `gte`, `eq`, `lt`, `lte`.
- then there is perhaps to be match functions too

Then if we add variable bindings, it'll become even more complete.

Finally allow user defined functions to be a programing language.

savonarola: The idea of the use case that occurred to me while reviewing was checking on emptiness in the `iif` condition.

emqxqa:
[Test Execution]()

PR #12874, opened 2024-04-12
API for durable subscriptions

ieQu1:
Fixes https://emqx.atlassian.net/browse/EMQX-12030 https://emqx.atlassian.net/browse/EMQX-12032 https://emqx.atlassian.net/browse/EMQX-12200

Release version: v/e5.?

## Summary

- Track PacketIDs of unacked QoS2 PUBLISH messages received from the client
- Add REST API for listing the durable subscriptions
- Fix inconsistent behavior in regard to overlapping subscriptions
- Separate subscription object from the immutable subscription state object
- Improve REST API response for offline durable sessions

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -75,7 +75,8 @@
 
 %% Managment APIs:
 -export([
-    list_client_subscriptions/1
+    list_client_subscriptions/1,
+    get_client_subscription/2
 ]).
 
 %% session table operations
@@ -735,6 +736,11 @@ list_client_subscriptions(ClientId) ->
         false ->
             {error, not_found}
     end.
+
+-spec get_client_subscription(emqx_types:clientid(), emqx_types:topic()) ->
+    subscription() | undefined.
+get_client_subscription(ClientId, Topic) ->
+    emqx_persistent_session_ds_subs:cold_get_subscription(ClientId, Topic).
 
 %%--------------------------------------------------------------------
 %% Session tables operations
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -367,10 +367,10 @@ subscribe(
 subscribe(
     TopicFilter,
     SubOpts,
-    Session = #{id := ID, s := S0, props := #{upgrade_qos := UpgradeQoS}}
+    Session = #{id := ID}
 ) ->
     {UpdateRouter, S1} = emqx_persistent_session_ds_subs:on_subscribe(
-        TopicFilter, UpgradeQoS, SubOpts, S0
+        TopicFilter, SubOpts, Session
     ),
     case UpdateRouter of
         true ->
@@ -379,9 +379,8 @@ subscribe(
             ok
     end,
     S = emqx_persistent_session_ds_state:commit(S1),
-    ?tp(persistent_session_ds_subscription_added, #{
-        topic_filter => TopicFilter, is_new => UpdateRouter
-    }),
+    UpdateRouter andalso
+        ?tp(persistent_session_ds_subscription_added, #{topic_filter => TopicFilter, session => ID}),
     {ok, Session#{s => S}}.
 
 -spec unsubscribe(topic_filter(), session()) ->
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -184,7 +184,9 @@
     seqno_q2_dup,
     seqno_q2_rec,
     seqno_q2_next,
-    n_streams
+    n_streams,
+    awaiting_rel_cnt,
+    awaiting_rel_max
 ]).
 
 %%
@@ -206,7 +208,8 @@ open(#{clientid := ClientID} = ClientInfo, ConnInfo, MaybeWillMsg, Conf) ->
     ok = emqx_cm:takeover_kick(ClientID),
     case session_open(ClientID, ClientInfo, ConnInfo, MaybeWillMsg) of
         Session0 = #{} ->
-            Session = Session0#{props => Conf},
+            Session1 = Session0#{props => Conf},
+            Session = do_expire(ClientInfo, Session1),
             {true, ensure_timers(Session), []};
         false ->
             false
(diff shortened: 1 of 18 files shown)


apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -1006,15 +1074,13 @@ enqueue_transient(ClientInfo, Msg0, Session = #{s := S, props := #{upgrade_qos :
...
-    Subs = emqx_persistent_session_ds_state:get_subscriptions(S),
-    Msgs = [
-        Msg
-     || SubMatch <- emqx_topic_gbt:matches(Msg0#message.topic, Subs, []),
-        Msg <- begin
-            #{props := SubOpts} = emqx_topic_gbt:get_record(SubMatch, Subs),
-            emqx_session:enrich_message(ClientInfo, Msg0, SubOpts, UpgradeQoS)
-        end
-    ],
+    case emqx_persistent_session_ds_state:get_subscription(Msg0#message.topic, S) of
```
ieQu1, line 1077: Matching topic of the message against subscriptions literally is wrong, but that's what mem session is doing, apparently.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -984,27 +988,29 @@ enqueue_batch(IsReplay, BatchSize, Srs0, Session = #{inflight := Inflight0}, Cli
...
+    {_DsMsgKey, Msg0} = KV,
     Comm1 = emqx_persistent_session_ds_state:get_seqno(?committed(?QOS_1), S),
     Comm2 = emqx_persistent_session_ds_state:get_seqno(?committed(?QOS_2), S),
     Dup1 = emqx_persistent_session_ds_state:get_seqno(?dup(?QOS_1), S),
     Dup2 = emqx_persistent_session_ds_state:get_seqno(?dup(?QOS_2), S),
     Rec = emqx_persistent_session_ds_state:get_seqno(?rec, S),
-    Subs = emqx_persistent_session_ds_state:get_subscriptions(S),
-    Msgs = [
-        Msg
-     || SubMatch <- emqx_topic_gbt:matches(Topic, Subs, []),
```
keynslug, line 931: Q: What happens now if 2 different subscriptions map to the same stream? Looking at stream scheduler code it supposedly is already taken care of, but it's still not entirely clear to me.

keynslug (dismissed):
Impressive work!

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: Streams for each subscriptions are independent. There will be a copy of the stream for each subscription.

apps/emqx/src/emqx_persistent_session_ds.hrl
```diff
@@ -65,7 +65,9 @@
     last_seqno_qos2 = 0 :: emqx_persistent_session_ds:seqno(),
     %% This stream belongs to an unsubscribed topic-filter, and is
     %% marked for deletion:
-    unsubscribed = false :: boolean()
+    unsubscribed = false :: boolean(),
+    %% Reference to the subscription state:
+    sub_state_id :: emqx_persistent_session_ds_subs:subscription_state_id()
```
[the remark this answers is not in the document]
ieQu1: Do you mean why not store the key-value map of states in the subscription record itself?
Separation helps with the following concerns:
- Unsubscribing logic is much simpler: now it's possible to simply delete or update the subscription record, and it won't affect the replay since the SRS records reference subscription states instead of subscriptions.
- Replay (hot path) logic is also simpler: it only needs one lookup to get the subopts.

apps/emqx/src/emqx_persistent_session_ds_subs.erl
```diff
@@ -73,26 +73,37 @@
...
+                        id => SubId,
+                        current_state => SStateId,
+                        start_time => now_ms()
+                    },
+                    S = emqx_persistent_session_ds_state:put_subscription(
+                        TopicFilter, Subscription, S3
+                    ),
+                    {true, S};
+                false ->
+                    {false, S0}
```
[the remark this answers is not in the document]
ieQu1: Good catch, will fix.

apps/emqx/src/emqx_persistent_session_ds_subs.erl
```diff
@@ -73,26 +73,37 @@
...
     emqx_persistent_session_ds:topic_filter(),
-    boolean(),
     emqx_types:subopts(),
-    emqx_persistent_session_ds_state:t()
+    emqx_persistent_session_ds:session()
 ) ->
     {_UpdateRouter :: boolean(), emqx_persistent_session_ds_state:t()}.
-on_subscribe(TopicFilter, UpgradeQoS, SubOpts, S0) ->
+on_subscribe(TopicFilter, SubOpts, #{s := S0, props := Props}) ->
+    #{upgrade_qos := UpgradeQoS, max_subscriptions := MaxSubscriptions} = Props,
```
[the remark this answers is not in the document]
ieQu1: It will make it impossible to change this configuration, there are valid usecases for doing so.

apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
```diff
@@ -210,8 +226,184 @@ do_subscriptions_query(QString) ->
...
+%% @doc Drop elements from the stream until encountered N elements
+%% matching the predicate function.
+-spec consume_n_matching(
+    fun((T) -> Q),
+    fun((Q) -> boolean()),
+    non_neg_integer(),
+    emqx_utils_stream:stream(T)
+) -> {[Q], emqx_utils_stream:stream(T) | empty}.
+consume_n_matching(Map, Pred, N, S) ->
+    consume_n_matching(Map, Pred, N, S, []).
```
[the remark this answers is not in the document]
ieQu1: I am not sure if stream-based APIs are the right way to go:
- The backend has to "rewind" the stream to the page on every call
- Since the steam API is stateless, but the underlying data is mutable, this leaves holes and/or duplicate issues when the matching elements are added between the calls.

I see this solution as temporary. In the future I hope we can adopt cursor-based REST API on the frontend and iterator-based API on the backend.

apps/emqx/src/emqx_persistent_session_ds.hrl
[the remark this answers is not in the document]
keynslug: > Do you mean why not store the key-value map of states in the subscription record itself?

No, I was thinking about SRS record: what's the goal of introducing subscription state table and referencing it as opposed to just storing Subopts + UpgradeQoS there?

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: I think it's not the same actually. In-mem sessions receive `deliver()`s and use the [_routing_ topic to look up the subscription](https://github.com/emqx/emqx/blob/0f426e6e77273bc18640698defaa0edc856b1cdf/apps/emqx/src/emqx_session.erl#L428) and not the _message_ topic. Those are different, _routing_ topic comes from the routing table and is basically topic filter of some subscription.

apps/emqx/src/emqx_persistent_session_ds.hrl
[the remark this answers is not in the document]
ieQu1: SRS record should be as lean as possible. While it's possible to pack the existing subscription state efficiently now, relying on this is not future proof.

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: Ok, I think this will work exactly the same then.

ieQu1: This is part of the shared channel logic, isn't it?

keynslug: Oh right, I see now, sorry for the confusion. That actually means that those messages have already been passed through `enrich_message/4` before hitting this function, so this second "enrichment" is not really needed?

ieQu1: Thanks, fixed.

keynslug (approved):

id (approved):

emqxqa:
[Test Execution]()

PR #12934, opened 2024-04-25
feat(s3-bridge): implement aggregated upload action

keynslug:
Fixes [EMQX-12204](https://emqx.atlassian.net/browse/EMQX-12204).

Release version: e5.7

## Summary

Action that performs time-based aggregation of Rule SQL results and uploads them to S3 API compatible storage services in a specified container (i.e. CSV).

TODOs.
* [ ] Graceful handling of rule changes.
* [x] Graceful handling of termination signals for delivery processes.
* [ ] Basic compression support.
* [x] Annotating uploads with appropriate Content-Type.

## PR Checklist

[EMQX-12204]: https://emqx.atlassian.net/browse/EMQX-12204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_bridge_s3/src/emqx_bridge_s3_aggreg_csv.erl
+++ b/apps/emqx_bridge_s3/src/emqx_bridge_s3_aggreg_csv.erl
@@ -16,14 +16,22 @@
 
 -record(csv, {
     columns :: [binary()] | undefined,
-    order :: [binary()],
-    separator :: char() | iodata(),
-    delimiter :: char() | iodata(),
+    column_order :: [binary()],
+    %% A string or character that separates each field in a record from the next.
+    %% Default: ","
+    field_separator :: char() | iodata(),
+    %% A string or character that delimits boundaries of a record.
+    %% Default: "\n"
+    record_delimiter :: char() | iodata(),
     quoting_mp :: _ReMP
 }).
 
 -type container() :: #csv{}.
--type options() :: #{column_order => [column()]}.
+
+-type options() :: #{
+    %% Which columns have to be ordered first in the resulting CSV?
+    column_order => [column()]
+}.
 
 -type record() :: emqx_bridge_s3_aggregator:record().
 -type column() :: binary().
@@ -34,9 +42,9 @@
 new(Opts) ->
     {ok, MP} = re:compile("[\\[\\],\\r\\n\"]", [unicode]),
     #csv{
-        order = maps:get(column_order, Opts, []),
-        separator = $,,
-        delimiter = $\n,
+        column_order = maps:get(column_order, Opts, []),
+        field_separator = $,,
+        record_delimiter = $\n,
         quoting_mp = MP
     }.
 
@@ -56,28 +64,28 @@ close(#csv{}) ->
 
 %%
 
-mk_columns(Record, #csv{order = ColumnOrder}) ->
+mk_columns(Record, #csv{column_order = ColumnOrder}) ->
     Columns = [emqx_utils_conv:bin(C) || C <- lists:sort(maps:keys(Record))],
     Unoredered = Columns -- ColumnOrder,
     ColumnOrder ++ Unoredered.
 
 -spec emit_header([column()], container()) -> iodata().
-emit_header([C], #csv{delimiter = Delim}) ->
+emit_header([C], #csv{record_delimiter = Delim}) ->
     [C, Delim];
-emit_header([C | Rest], CSV = #csv{separator = Sep}) ->
+emit_header([C | Rest], CSV = #csv{field_separator = Sep}) ->
     [C, Sep | emit_header(Rest, CSV)];
-emit_header([], #csv{delimiter = Delim}) ->
+emit_header([], #csv{record_delimiter = Delim}) ->
     [Delim].
 
 -spec emit_row(record(), container()) -> iodata().
 emit_row(Record, CSV = #csv{columns = Columns}) ->
     emit_row(Record, Columns, CSV).
 
-emit_row(Record, [C], CSV = #csv{delimiter = Delim}) ->
+emit_row(Record, [C], CSV = #csv{record_delimiter = Delim}) ->
     [emit_cell(C, Record, CSV), Delim];
-emit_row(Record, [C | Rest], CSV = #csv{separator = Sep}) ->
+emit_row(Record, [C | Rest], CSV = #csv{field_separator = Sep}) ->
     [emit_cell(C, Record, CSV), Sep | emit_row(Record, Rest, CSV)];
-emit_row(#{}, [], #csv{delimiter = Delim}) ->
+emit_row(#{}, [], #csv{record_delimiter = Delim}) ->
     [Delim].
 
 emit_cell(Column, Record, CSV) ->
--- a/apps/emqx_bridge_s3/src/emqx_bridge_s3_aggreg_csv.erl
+++ b/apps/emqx_bridge_s3/src/emqx_bridge_s3_aggreg_csv.erl
@@ -57,7 +57,7 @@ close(#csv{}) ->
 %%
 
 mk_columns(Record, #csv{order = ColumnOrder}) ->
-    Columns = lists:sort(maps:keys(Record)),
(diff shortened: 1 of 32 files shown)


apps/emqx_bridge_s3/src/emqx_bridge_s3_aggreg_csv.erl
```diff
@@ -0,0 +1,97 @@
...
+    new/1,
+    fill/2,
+    close/1
+]).
+
+-export_type([container/0]).
+
+-record(csv, {
+    columns :: [binary()] | undefined,
+    order :: [binary()],
```
savonarola, line 19:
```suggestion
    column_order :: [binary()],
```

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggreg_csv.erl
```diff
@@ -0,0 +1,97 @@
...
+    separator :: char() | iodata(),
+    delimiter :: char() | iodata(),
```
savonarola, line 21: Nit: m.b. record_separator/column_separator and line_separator?

keynslug: Good point. Turns out there's tiny difference in semantics between _separator_ and _delimiter_, so I just put appropriate prefixes for clarity.

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
```diff
@@ -0,0 +1,486 @@
...
+        _ ->
+            write_records(Name, Buffer, Records)
+    end.
+
+write_records(Name, Buffer = #buffer{fd = Writer}, Records) ->
+    case emqx_bridge_s3_aggreg_buffer:write(Records, Writer) of
+        ok ->
+            ?tp(s3_aggreg_records_written, #{action => Name, records => Records}),
+            ok;
+        {error, Reason} when Reason == terminated orelse Reason == closed ->
```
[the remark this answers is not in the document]
keynslug: Not entirely sure but likely not, good point.

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
```diff
@@ -0,0 +1,486 @@
...
+-define(SRVREF(NAME), {via, gproc, {n, l, {?MODULE, NAME}}}).
+
+%%
+
+start_link(Name, Opts) ->
+    gen_server:start_link(?SRVREF(Name), ?MODULE, mk_state(Name, Opts), []).
+
+push_records(Name, Timestamp, Records = [_ | _]) ->
+    %% FIXME: Error feedback.
+    case pick_buffer(Name, Timestamp) of
```
savonarola, line 54: Could be selected buffer just invalidated after pick?
Would that result e.g. in rotating buffer twice?

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
```diff
@@ -0,0 +1,486 @@
...
+    ets:update_counter(Tab, Counter, {2, Size}).
+
+del_counter({Tab, Counter}) ->
+    ets:delete(Tab, Counter).
+
+%%
+
+create_tab(Name) ->
+    Tab = ets:new(?MODULE, [public, set, {write_concurrency, auto}]),
+    ok = persistent_term:put({?MODULE, Name}, Tab),
```
savonarola, line 474: Why do we need to register tabs through `persistent_term`?

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
keynslug: Yes it can. The idea is that `file:write/2` then returns `{error, terminated}`, which will translate to `rotate_buffer/2`, and if the buffer in question has been already rotated by the time `handle_rotate_buffer/2` aggregator process will just respond with the current buffer (since it has different FD), which will most likely be a buffer for the next time window in this case.

Though I may be missing something important, please write down a scenario that you think could lead to weird outcomes.

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
```diff
@@ -0,0 +1,486 @@
...
+    end;
+push_records(_Name, _Timestamp, []) ->
+    ok.
+
+tick(Name, Timestamp) ->
+    case pick_buffer(Name, Timestamp) of
+        #buffer{} ->
+            ok;
+        _Outdated ->
+            send_close_buffer(Name, Timestamp)
```
savonarola, line 69: Why do we need to send buffer on tick (which is used only for status check)?

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
keynslug: To avoid doing `atom_to_list` / `list_to_atom` cycles in a hot loop (unless I'm missing something obvious).

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
keynslug: The idea was to avoid dealing with current time and timers in aggregator code, to make it more easily testable and reusable. Truth be told it's not exactly clear if these goals are met, given that this gave rise to API warts like that: without timers, someone has to poke the aggregator to kick off any outdated buffer deliveries. 🫠

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
```diff
@@ -0,0 +1,486 @@
...
+init(St0 = #st{name = Name}) ->
+    _ = erlang:process_flag(trap_exit, true),
+    St1 = St0#st{tab = create_tab(Name)},
+    St = recover(St1),
+    _ = announce_current_buffer(St),
+    {ok, St}.
+
+handle_call({next_buffer, Timestamp}, _From, St0) ->
+    St = #st{buffer = Buffer} = handle_next_buffer(Timestamp, St0),
+    {reply, Buffer, St, 0};
```
savonarola, line 167: nit: mb `handle_continue`?

keynslug: This is basically a measure to prioritize handling calls over starting queued deliveries. My assumption is that this way we can handle all queued calls first, and only then proceed to deal with `#st.queued`. I now wonder if it's correct assumption, and if `handle_continue` would have the same effect?

savonarola (approved):

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
savonarola: I see

`handle_continue` seems to follow immediately, before the rest of the queue

apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl
savonarola: As far as I understood, on failure, `write_records` tries to obtain another buffer through `rotate_buffer` which unconditionally makes rotation...
I was thinking about a paranoid scenario when, due to concurrency, several writers invalidate each other's buffers and leave many closed empty buffers 😅

keynslug: Yeah, that indeed could happen! There's [this clause](https://github.com/emqx/emqx/blob/3afee4ed8b2ac8f557a2b7eeeae4e4a4456bee69/apps/emqx_bridge_s3/src/emqx_bridge_s3_aggregator.erl#L228) that (hopefully) should basically deduplicate those _rotate_ requests.

yanzhiemq:
[TestExecution](https://emqx.atlassian.net/browse/QA-2158)

PR #12948, opened 2024-04-29
fix(http connector): deobfuscate sensitive headers

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12213

Release version: v/e5.7

## PR Checklist

### DIFF
--- a/apps/emqx_utils/src/emqx_utils_redact.erl
+++ b/apps/emqx_utils/src/emqx_utils_redact.erl
@@ -152,19 +152,24 @@ redact_v(_V) ->
     ?REDACT_VAL.
 
 deobfuscate(NewConf, OldConf) ->
+    deobfuscate(NewConf, OldConf, fun(_) -> false end).
+
+deobfuscate(NewConf, OldConf, IsSensitiveFun) ->
     maps:fold(
         fun(K, V, Acc) ->
             case maps:find(K, OldConf) of
                 error ->
-                    case is_redacted(K, V) of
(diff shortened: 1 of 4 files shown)


zmstone (approved):

apps/emqx_utils/src/emqx_utils_redact.erl
```diff
@@ -152,19 +152,24 @@ redact_v(_V) ->
...
         fun(K, V, Acc) ->
             case maps:find(K, OldConf) of
                 error ->
-                    case is_redacted(K, V) of
+                    case is_redacted(K, V, IsSensitiveFun) of
                         %% don't put redacted value into new config
                         true -> Acc;
                         false -> Acc#{K => V}
                     end;
+                {ok, OldV} when is_map(V), is_map(OldV), ?IS_KEY_HEADERS(K) ->
```
zmstone, line 167: `redact` also handles when value is a tuple list, but i guess it's never the case in values needed to be deobfuscatted ?

thalesmg: Good point.  It would be more symmetrical to treat those as well.  But, from what I can see, deobfuscation happens only at HTTP API boundaries (when receiving user input) and deals with raw configurations, so I think tuples are unexpected types in those situations.

yanzhiemq:
[TestExecution](https://emqx.atlassian.net/browse/QA-2153)

PR #12955, opened 2024-04-30
port: TLS partial chain

qzhuyan:
Fixes <issue-or-jira-number>

Release version: v/e5.8.0

## PR Checklist


apps/emqx/src/emqx_const_v2.erl
```diff
@@ -0,0 +1,127 @@
...
+        true ->
+            do_verify_ext_key_usage(CertExtL, T);
+        false ->
+            false
+    end.
+
+%% @doc Helper tls cert extension
+-spec ext_key_opts
+    (string()) -> [OidString :: string() | public_key:oid()];
+    (undefined) -> undefined.
```
thalesmg, line 105: Nit: this function doesn't actually support `undefined` as an input in its current form: `string:tokens/2` will raise if `Str = undefined`.

apps/emqx/src/emqx_const_v2.erl
```diff
@@ -0,0 +1,127 @@
...
+                ?'id-kp-serverAuth';
+            ("codeSigning") ->
+                ?'id-kp-codeSigning';
+            ("emailProtection") ->
+                ?'id-kp-emailProtection';
+            ("timeStamping") ->
+                ?'id-kp-timeStamping';
+            ("ocspSigning") ->
+                ?'id-kp-OCSPSigning';
+            ([$O, $I, $D, $: | OidStr]) ->
```
thalesmg, line 122: Micro nit:

```suggestion
            ("OID:" ++ OidStr) ->
```

apps/emqx/src/emqx_const_v2.erl
```diff
@@ -0,0 +1,127 @@
...
+            true ->
+                {trusted_ca, TrustedOne};
+            false ->
+                {trusted_ca, TrustedTwo}
+        end
+    end.
+
+make_tls_verify_fun(verify_cert_extKeyUsage, KeyUsages) ->
+    AllowedKeyUsages = ext_key_opts(KeyUsages),
+    {fun verify_fun_peer_extKeyUsage/3, AllowedKeyUsages}.
```
thalesmg, line 53: Q: since this returns `{function(), [term()]}`, do we need to keep it in `emqx_const_v2`?  If we were to return a fully qualified fn like `{fun some_other_module:.../3, AllowedKeyUsages}`, wouldn't that eliminate the need to keep it here, immutable?

apps/emqx/src/emqx_const_v2.erl
```diff
@@ -0,0 +1,127 @@
...
+    %% OTP verify_peer default
+    {unknown, UserState};
+verify_fun_peer_extKeyUsage(_, valid, UserState) ->
+    %% OTP verify_peer default
+    {valid, UserState};
+verify_fun_peer_extKeyUsage(
+    #'OTPCertificate'{tbsCertificate = #'OTPTBSCertificate'{extensions = ExtL}},
+    %% valid peer cert
+    valid_peer,
+    AllowedKeyUsages
```
thalesmg, line 72: Nit: this sounds more like `RequiredKeyUsages` (all must be present in the extension list) than "allowed" (only those are allowed, but suggests that the certificate extension list _could_ be empty).

qzhuyan: yes `Required` is better term.

apps/emqx/src/emqx_const_v2.erl
qzhuyan: @zmstone correct me if I am wrong.
I remember we have `const` module to avoid `badfun` in upgrade and RPC. const module are there and never get realoded/upgraded.

Here it returns a SSL callback `fun` that stick to the version of a module which is risky get unloaded during upgrade, we put it here to ensure the `fun` survives after upgrade.

thalesmg: Indeed, but I thought that using a fully qualified function (external call) like `Mod:Fun/N` would always resolve to the current version of the code, so it wouldn't become a `badfun` like a closure/lambda.

thalesmg (approved):
~~I still think `make_tls_verify_fun` doesn't need to live in `emqx_const_v2`, but maybe @zmstone could double check if my claim makes sense.~~

Nevermind: we want to ensure this function call is not rename nor (re)moved, anyway.

qzhuyan:
We could have a followup if we decide to move it.

apps/emqx/src/emqx_schema.erl
```diff
@@ -2109,6 +2109,22 @@ common_ssl_opts_schema(Defaults, Type) ->
...
+                }
+            )},
+        {"verify_peer_ext_key_usage",
+            sc(
+                string(),
+                #{
+                    required => false,
+                    desc => ?DESC(common_ssl_opts_verify_peer_ext_key_usage)
+                }
+            )},
```
zmstone, line 2127: should be added only to listener schema, but not clients.

changes/ce/feat-11721.en.md
```diff
@@ -0,0 +1,5 @@
+Port two TLS handshake validation features from emqx 4.4
```
zmstone, line 1: Enhance TLS listener to support more flexible TLS verifications.

changes/ce/feat-11721.en.md
```diff
@@ -0,0 +1,5 @@
...
+
+- partial_chain support
+- Certificate KeyUsage Validation
```
zmstone, line 4: explain in more details what these features are, and add examples too.

rel/i18n/emqx_schema.hocon
```diff
@@ -678,6 +678,18 @@ common_ssl_opts_schema_verify.desc:
 common_ssl_opts_schema_verify.label:
 """Verify peer"""
 
+common_ssl_opts_schema_partial_chain.desc:
+"""Enable or disable peer verification with partial_chain"""
```
zmstone, line 682: Document what each enum value means
- true
- false
- two_cacerts_from_cacertfile
- cacert_from_cacertfile means

rel/i18n/emqx_schema.hocon
```diff
@@ -678,6 +678,18 @@ common_ssl_opts_schema_verify.desc:
...
+
+common_ssl_opts_schema_partial_chain.label:
+"""Partial chain"""
+
+common_ssl_opts_verify_peer_ext_key_usage.desc:
+"""Verify Extended Key Usage in Peer's certificate"""
```
zmstone, line 688: explain the value scheme.
comma separated keys, and the special format of `OID:` etc.

apps/emqx/src/emqx_schema.erl
qzhuyan: they ({verify_fun, [custom_verify()](https://www.erlang.org/doc/man/ssl#type-custom_verify)}  and {partial_chain, [root_fun()](https://www.erlang.org/doc/man/ssl#type-root_fun)} ) are common in TLS doc, so I think they should be common for both client and server? Their defaults has no impact on the defaults of client.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2498)

PR #12957, opened 2024-04-30
adjust build scripts

id:
- minimize discrepancies with the private fork
- add macos 14


.github/workflows/check_deps_integrity.yaml
```diff
@@ -14,32 +14,30 @@ jobs:
...
       - run: ./scripts/check-deps-integrity.escript
       - name: Setup mix
-        env:
-          MIX_ENV: emqx-enterprise
-          PROFILE: emqx-enterprise
         run: |
           mix local.hex --force
           mix local.rebar --force
           mix deps.get
       - run: ./scripts/check-elixir-deps-discrepancies.exs
```
thalesmg, line 34: Just a nano suggestion that might help debug future issues, since we're touching here:

```suggestion
      - name: print mix dependency tree
        run: mix deps.tree
      - run: ./scripts/check-elixir-deps-discrepancies.exs
```

mix.exs
```diff
@@ -113,8 +113,9 @@ defmodule EMQXUmbrella.MixProject do
     set_emqx_app_system_env(apps, profile_info, version)
   end
 
-  defp umbrella_apps(profile_info) do
-    enterprise_apps = enterprise_umbrella_apps()
+  defp umbrella_apps(profile_info = %{release_type: release_type}) do
+    enterprise_apps = enterprise_umbrella_apps(release_type)
+    excluded_apps = excluded_apps(release_type)
```
thalesmg, line 118: Nit: I think this is not needed here, since we enumerate the umbrella apps under `apps/*`, and the excluded apps are external ones such as `quicer`, `observer`, etc. .

thalesmg (approved):

mix.exs
id: Yes, correct. I added this to reduce number of different lines in mix.exs between this and what we have in the private fork.

thalesmg (approved):

rebar.config.erl
```diff
@@ -543,10 +553,11 @@ dialyzer(Config) ->
         end,
 
     AppNames = app_names(),
+    ExcludedApps = excluded_apps(standard),
 
     KnownApps = [Name || Name <- AppsToAnalyse, lists:member(Name, AppNames)],
 
-    AppsToExclude = AppNames -- KnownApps,
+    AppsToExclude = ExcludedApps -- KnownApps ++ AppNames,
```
zmstone, line 560: This is effectively the same as: `AppsToExclude = ExcludedApps -- (KnownApps ++ AppNames)`
`KnownApps` is a subset of `AppNames`, I don't understand why the ++ here.

id: did you mean `AppsToExclude = ExcludedApps -- AppNames`?

.github/workflows/_push-entrypoint.yaml
```diff
@@ -8,7 +8,6 @@ on:
   push:
     tags:
       - 'v*'
-      - 'e*'
```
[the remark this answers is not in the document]
id: My plan was to do opposite change in another repo. 
May be even simpler approach would be to block corresponding tags on repo settings level.

rebar.config.erl
zmstone: yes.

rebar.config.erl
```diff
@@ -543,10 +553,8 @@ dialyzer(Config) ->
         end,
 
     AppNames = app_names(),
-
-    KnownApps = [Name || Name <- AppsToAnalyse, lists:member(Name, AppNames)],
-
-    AppsToExclude = AppNames -- KnownApps,
+    ExcludedApps = excluded_apps(standard),
+    AppsToExclude = ExcludedApps -- AppNames,
```
id, line 557: This does not look right. What exactly we are trying to do here?

thalesmg (approved):

zmstone (approved):

emqxqa:
[Test Execution]()

PR #12958, opened 2024-04-30
fix: rule trace formatting, republish and console stop after rendering

kjellwinblad:
* Better rule trace formatting for many trace entries
* The republish and console actions have got working stop after rendering functionality
* (bugfix): make sure we return something JSON compatible when calling the rule apply API
* (bugfix): trace should be working immediately after the API call to create it returns

Fixes:
https://emqx.atlassian.net/browse/EMQX-12275
https://emqx.atlassian.net/browse/EMQX-12276

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_logger_jsonfmt.erl
+++ b/apps/emqx/src/emqx_logger_jsonfmt.erl
@@ -229,6 +229,13 @@ best_effort_json_obj(Map, Config) ->
             do_format_msg("~p", [Map], Config)
     end.
 
+json_value(true, _Config) ->
+    true;
+json_value(false, _Config) ->
+    false;
+json_value(V, Config) ->
+    json(V, Config).
+
 json(A, _) when is_atom(A) -> atom_to_binary(A, utf8);
 json(I, _) when is_integer(I) -> I;
 json(F, _) when is_float(F) -> F;
@@ -317,7 +324,7 @@ json_kv(K0, V, Config) ->
     K = json_key(K0),
     case is_map(V) of
         true -> {K, best_effort_json_obj(V, Config)};
-        false -> {K, json(V, Config)}
+        false -> {K, json_value(V, Config)}
     end.
 
 json_key(A) when is_atom(A) -> json_key(atom_to_binary(A, utf8));
--- a/apps/emqx/src/emqx_logger_jsonfmt.erl
+++ b/apps/emqx/src/emqx_logger_jsonfmt.erl
@@ -229,14 +229,7 @@ best_effort_json_obj(Map, Config) ->
             do_format_msg("~p", [Map], Config)
     end.
 
-json_value(true, _Config) ->
-    true;
-json_value(false, _Config) ->
-    false;
-json_value(V, Config) ->
-    json(V, Config).
-
-json(A, _) when is_atom(A) -> atom_to_binary(A, utf8);
+json(A, _) when is_atom(A) -> A;
 json(I, _) when is_integer(I) -> I;
 json(F, _) when is_float(F) -> F;
 json(P, C) when is_pid(P) -> json(pid_to_list(P), C);
@@ -324,7 +317,7 @@ json_kv(K0, V, Config) ->
     K = json_key(K0),
     case is_map(V) of
         true -> {K, best_effort_json_obj(V, Config)};
-        false -> {K, json_value(V, Config)}
+        false -> {K, json(V, Config)}
     end.
 
 json_key(A) when is_atom(A) -> json_key(atom_to_binary(A, utf8));
--- a/apps/emqx/src/emqx_trace/emqx_trace.erl
+++ b/apps/emqx/src/emqx_trace/emqx_trace.erl
@@ -405,7 +405,14 @@ code_change(_, State, _Extra) ->
     {ok, State}.
 
 insert_new_trace(Trace) ->
-    transaction(fun emqx_trace_dl:insert_new_trace/1, [Trace]).
+    case transaction(fun emqx_trace_dl:insert_new_trace/1, [Trace]) of
+        {error, _} = Error ->
+            Error;
+        Res ->
+            %% We call this to ensure the trace is active when we return
+            check(),
+            Res
+    end.
 
 update_trace(Traces) ->
     Now = now_second(),
--- a/apps/emqx/src/emqx_trace/emqx_trace.erl
+++ b/apps/emqx/src/emqx_trace/emqx_trace.erl
@@ -31,7 +31,8 @@
     log/4,
     rendered_action_template/2,
     make_rendered_action_template_trace_context/1,
-    rendered_action_template_with_ctx/2
+    rendered_action_template_with_ctx/2,
+    is_rule_trace_active/0
 ]).
 
 -export([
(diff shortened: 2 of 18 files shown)


apps/emqx/src/emqx_logger_jsonfmt.erl
```diff
@@ -229,6 +229,13 @@ best_effort_json_obj(Map, Config) ->
...
     end.
 
+json_value(true, _Config) ->
+    true;
+json_value(false, _Config) ->
+    false;
+json_value(V, Config) ->
+    json(V, Config).
+
 json(A, _) when is_atom(A) -> atom_to_binary(A, utf8);
```
zmstone, line 239:
```suggestion
json(A, _) when is_atom(A) -> A;
```

apps/emqx/src/emqx_trace/emqx_trace_json_formatter.erl
```diff
@@ -57,12 +57,46 @@ prepare_key_value(K, {Formatter, V}, PEncode) when is_function(Formatter, 1) ->
...
+    is_integer(I2),
+    is_integer(I3),
+    is_integer(I4)
 ->
-    %% This is unlikely anything else then info about a HTTP request so we make
-    %% it more structured
-    prepare_key_value(K, #{status => Status, headers => Headers, body => Body}, PEncode);
+    %% We assume this is an IP address
+    {host,
+        unicode:characters_to_binary([
```
zmstone, line 71: inet:ntoa/1

apps/emqx/src/emqx_trace/emqx_trace_json_formatter.erl
```diff
@@ -57,12 +57,46 @@ prepare_key_value(K, {Formatter, V}, PEncode) when is_function(Formatter, 1) ->
...
+            <<".">>,
+            integer_to_binary(I3),
+            <<".">>,
+            integer_to_binary(I4)
+        ])};
+prepare_key_value(K, {ok, StatusCode, Headers}, PEncode) when
+    is_integer(StatusCode), StatusCode >= 200, StatusCode < 300, is_list(Headers)
+->
+    prepare_key_value(K, {ok, StatusCode, Headers, <<"">>}, PEncode);
+prepare_key_value(K, {ok, StatusCode, Headers, Body}, PEncode) when
```
zmstone, line 84: this seems to be the wrong module to format http results?

kjellwinblad: In a way yes. However there are a few connectors that return the result in the same format so my idea was that it would make the code less cluttered if these results got formatted implicitly. It might be still be best to be more explicit? I'm okay with both so I can make these into lacy format functions in the connectors if you think this is better?

kjellwinblad: Just realized that we also need to handle the format functions in the other log formanters as well!

apps/emqx/src/emqx_logger_jsonfmt.erl
kjellwinblad: This works because all other atoms than true and false are converted to strings by the JSON library, right?

thalesmg: Indeed, `jiffy` seems to encode atoms as JSON strings:

```elixir
iex(emqx@127.0.0.1)1> :jiffy.encode(%{a: :a})
"{\"a\":\"a\"}"
```

apps/emqx/src/emqx_trace/emqx_trace_json_formatter.erl
kjellwinblad: I'm having a discussion with @thalesmg . Returning the formatter in the result from the connector is a quite ugly hack that could lead to problems. His suggestion is to add a connector callback for trace formatting the result. The formatter module can infer which callback to use from the Action ID. I think this sounds like a reasonable suggestion so I will try to implement this tomorrow.

apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
```diff
@@ -413,6 +414,24 @@ on_query(
...
+    %% add a function to lazily format the trace entry.
+    case emqx_trace:is_rule_trace_active() of
+        true ->
+            {ok, {fun trace_format_result/1, Res}};
+        false ->
+            Res
+    end.
+
+trace_format_result({ok, Status, Headers, Body}) ->
+    #{status => Status, headers => Headers, body => Body};
```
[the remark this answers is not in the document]
kjellwinblad: Not sure. We redact the headers that we send to the server. These are the headers that we get in response from the server. Does it make sense to redact these as well?

apps/emqx/src/emqx_logger_jsonfmt.erl
kjellwinblad: Fixed: [e0d7a87](https://github.com/emqx/emqx/pull/12958/commits/e0d7a876ec4c92b800ae5bd9ff12e9cd1bc2e9a9)

apps/emqx/src/emqx_trace/emqx_trace_json_formatter.erl
kjellwinblad: Fixed: [e0d7a87](https://github.com/emqx/emqx/pull/12958/commits/e0d7a876ec4c92b800ae5bd9ff12e9cd1bc2e9a9)

apps/emqx/src/emqx_trace/emqx_trace_json_formatter.erl
kjellwinblad: Implemented a solution in line with @thalesmg's suggestion here: [e0d7a87](https://github.com/emqx/emqx/pull/12958/commits/e0d7a876ec4c92b800ae5bd9ff12e9cd1bc2e9a9)

zmstone (approved):

apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
```diff
@@ -503,33 +507,68 @@ on_query_async(
...
+            https ->
+                <<"https://">>
+        end,
+    unicode:characters_to_binary([
+        SchemeStr,
+        Host,
+        <<":">>,
+        erlang:integer_to_binary(Port),
+        Path
+    ]).
```
thalesmg, line 570: Tiny nit: there's also `uri_string:recompose/1`.

```elixir
iex(emqx@127.0.0.1)7> :uri_string.recompose(%{host: "a.b.c", path: "/some/path", scheme: "https", port: 8080})
"https://a.b.c:8080/some/path"
```

thalesmg (approved):

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -108,23 +119,25 @@ rendered_action_template(<<"action:", _/binary>> = ActionID, RenderResult) ->
...
             %% resource is called/modified
-            StopMsg = lists:flatten(
+            ActionIDStr =
+                case ActionID of
+                    Bin when is_binary(Bin) ->
+                        Bin;
+                    Term ->
+                        ActionIDFormatted = io_lib:format("~tw", [Term]),
+                        unicode:characters_to_binary(ActionIDFormatted)
+                end,
```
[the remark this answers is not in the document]
kjellwinblad: Good suggestion. I will keep this in mind for later. I don't think we should use this here since the emqx app should not depend on the emq_rule_engine app.

apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
kjellwinblad: Thanks for the suggestion. Will try to fix this in a different PR.

emqxqa:
[Test Execution]()

PR #12968, opened 2024-05-03
ci: send data to coveralls from emqx/emqx only

id:


thalesmg (commented):
~~Maybe we could skip the whole `make_cover` job if the repo doesn't match?~~

thalesmg (approved):

thalesmg (approved):

PR #12971, opened 2024-05-03
fix(auth,http): improve URI handling

savonarola:
Fixes [EMQX-12257](https://emqx.atlassian.net/browse/EMQX-12257)

Release version: v/e5.8.0

This PR addresses three entangled problems.

---

The first one is directly described in the task above. We have templated URIs in many parts of the code, and URI templates cannot be parsed with the standard URI parser (`uri_string`). So, multiple ad-hoc simplified parses are created, which may be error-prone. 

**We solve this by introducing a single loose parser** `emqx_utils_uri`.

---

The second one comes from the fact that the path that forms a part of the URI and the path that forms the "request-target" in the corresponding HTTP request are different concepts. The former may not start with a slash, while the latter MUST be absolute.

For example, in the valid URI `http://example.com?q=x`, the path is empty. However, the HTTP request `GET ?q=x HTTP/1.1` is invalid. It should be sent in the form `GET /?q=x HTTP/1.1`. See also: https://datatracker.ietf.org/doc/html/rfc7230#section-5.3

Browsers and high-level HTTP libraries automatically normalize the path, e.g.:
* https://github.com/sneako/finch/blob/71f469c46d3d8ea9993d986789cef8d14bcc819c/lib/finch/request.ex#L124
* https://github.com/benoitc/hackney/blob/7ae3831c1a56896d05d862de2c44cc751ae7790f/src/hackney_request.erl#L32

But `ehttpc` is a wrapper around quite a low-level HTTP client, `gun`, which does not make any URI parsing at all. And over `ehttpc` there are quite a lot of layers, so it was not quite clear which part of the code should be responsible for normalization. 

**We enforce normalization on** `emqx_bridge_http_connector` **level**.

---

Third, `emqx_bridge_http_connector` API is redundant. It accepts `base_url`, which is prepended to all requests, but:
* the `path` field of `base_url` is never used. Everywhere `base_url` is an explicitly cut-out scheme-host-port part.
* even if it were going to be used, the `path`-related part is not functional, because the correct validation and correct merging of paths is not implemented.

**So we do not accept** `base_url` **anymore and accept an** `request_base() :: #{scheme := term(), host := term(), port := term()}` **instead.**

## PR Checklist

- [x] Added property-based tests for code which performs user input validation (property-like ones)

[EMQX-12257]: https://emqx.atlassian.net/browse/EMQX-12257?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_auth/src/emqx_auth_utils.erl
+++ b/apps/emqx_auth/src/emqx_auth_utils.erl
@@ -1,0 +1,78 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2024 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%
+%% Licensed under the Apache License, Version 2.0 (the "License");
+%% you may not use this file except in compliance with the License.
+%% You may obtain a copy of the License at
+%%
+%%     http://www.apache.org/licenses/LICENSE-2.0
+%%
+%% Unless required by applicable law or agreed to in writing, software
+%% distributed under the License is distributed on an "AS IS" BASIS,
+%% WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%% See the License for the specific language governing permissions and
+%% limitations under the License.
+%%--------------------------------------------------------------------
+
+-module(emqx_auth_utils).
+
+%% TODO
+%% Move more identical authn and authz helpers here
+
+-export([parse_url/1]).
+
+%%--------------------------------------------------------------------
+%% API
+%%--------------------------------------------------------------------
+
+-spec parse_url(binary()) ->
+    {_Base :: emqx_utils_uri:request_base(), _Path :: binary(), _Query :: binary()}.
+parse_url(Url) ->
+    Parsed = emqx_utils_uri:parse(Url),
+    case Parsed of
+        #{scheme := undefined} ->
+            throw({invalid_url, {no_scheme, Url}});
+        #{authority := undefined} ->
+            throw({invalid_url, {no_host, Url}});
+        #{authority := #{userinfo := Userinfo}} when Userinfo =/= undefined ->
+            throw({invalid_url, {userinfo_not_supported, Url}});
+        #{fragment := Fragment} when Fragment =/= undefined ->
+            throw({invalid_url, {fragments_not_supported, Url}});
+        _ ->
+            case emqx_utils_uri:request_base(Parsed) of
+                {ok, Base} ->
+                    {Base, emqx_utils_uri:path(Parsed),
+                        emqx_maybe:define(emqx_utils_uri:query(Parsed), <<>>)};
+                {error, Reason} ->
+                    throw({invalid_url, {invalid_base, Reason, Url}})
+            end
+    end.
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+
+templates_test_() ->
+    [
+        ?_assertEqual(
+            {
+                #{port => 80, scheme => http, host => "example.com"},
+                <<"">>,
+                <<"client=${clientid}">>
+            },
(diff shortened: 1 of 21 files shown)


apps/emqx_auth/src/emqx_auth_utils.erl
```diff
@@ -0,0 +1,78 @@
...
+%% API
+%%--------------------------------------------------------------------
+
+-spec parse_url(binary()) ->
+    {_Base :: emqx_utils_uri:request_base(), _Path :: binary(), _Query :: binary()}.
+parse_url(Url) ->
+    Parsed = emqx_utils_uri:parse(Url),
+    case Parsed of
+        #{scheme := undefined} ->
+            throw({invalid_url, {no_schema, Url}});
```
thalesmg, line 34:
```suggestion
            throw({invalid_url, {no_scheme, Url}});
```

thalesmg (approved):

apps/emqx_auth/src/emqx_auth_utils.erl
```diff
@@ -0,0 +1,78 @@
...
+    Parsed = emqx_utils_uri:parse(Url),
+    case Parsed of
+        #{scheme := undefined} ->
+            throw({invalid_url, {no_scheme, Url}});
+        #{authority := undefined} ->
+            throw({invalid_url, {no_host, Url}});
+        #{authority := #{userinfo := Userinfo}} when Userinfo =/= undefined ->
+            throw({invalid_url, {userinfo_not_supported, Url}});
+        #{fragment := Fragment} when Fragment =/= undefined ->
+            throw({invalid_url, {fragments_not_supported, Url}});
```
zmstone, line 40: why this is considered invalid ?

savonarola: fragments specify locations in a fetched resource, but we do not use them anyhow

apps/emqx_utils/src/emqx_utils_uri.erl
```diff
@@ -0,0 +1,392 @@
...
+%% > well-formed URI reference into its components.
+%%
+%% > ^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?
+%%
+%% We skip capturing some unused parts of the regex.
+
+-define(URI_REGEX,
+    ("^(?:(?<scheme>[^:/?#]+):)?(?<authority>//[^/?#]*)?"
+    "(?<path>[^?#]*)(?<query>\\?[^#]*)?(?<fragment>#.*)?")
+).
```
zmstone, line 215: nit: move it to the top.

apps/emqx_utils/src/emqx_utils_uri.erl
```diff
@@ -0,0 +1,392 @@
...
+
+uri_regexp() ->
+    {ok, RE} = re:compile(?URI_REGEX, [caseless]),
```
zmstone, line 218: maybe worth caching it in persistent term in a on_load callback.
simple profiling shows re:compile is 20x as persistent_term:get

savonarola: ✅ 
Currently, we do not parse URIs anywhere in hot loops only on resource/auth creation.
But let the code be a bit more foolproof and efficient.

apps/emqx_utils/src/emqx_utils_uri.erl
savonarola: ✅

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2506)

PR #12973, opened 2024-05-03
refactor: extract blob aggregator logic to separate application

thalesmg:
This will allow other bridges to share the same aggregation logic as S3 aggregated action.

Aiming for e5.7 as that will relieve us of backward compatibility issues that would be introduced in the buffer file header by the renaming of the buffer module.

Part of https://emqx.atlassian.net/browse/EMQX-12280

Release version: e5.7

## PR Checklist


apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -0,0 +1,183 @@
...
+-type state() :: #delivery{}.
+
+-type init_opts() :: #{
+    callback_module := module(),
+    any() => term()
+}.
+
+-type impl_specific_upload_state() :: term().
+
+-callback init_upload_state(buffer(), map()) -> impl_specific_upload_state().
```
keynslug, line 45: Nit. I'd argue that it's already apparent that it's implementation specific.
```suggestion
-callback init_upload_state(buffer(), map()) -> upload_state().
```

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -0,0 +1,183 @@
...
+    system_terminate/4,
+    format_status/2
+]).
+
+-record(delivery, {
+    name :: _Name,
+    callback_module :: module(),
+    container :: emqx_connector_aggreg_csv:container(),
+    reader :: emqx_connector_aggreg_buffer:reader(),
+    upload :: impl_specific_upload_state(),
```
keynslug, line 32: Q: If we are to generalize this, maybe _upload_ is too narrow name for that? Admittedly, I can't propose anything better. `transfer`?

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -0,0 +1,183 @@
...
+process_append_records(
+    Records,
+    Delivery = #delivery{
+        callback_module = Mod,
+        container = Container0,
+        upload = Upload0
+    }
+) ->
+    {Writes, Container} = emqx_connector_aggreg_csv:fill(Records, Container0),
+    {ok, Upload} = Mod:process_append(Writes, Upload0),
```
keynslug, line 138: Just a sidenote essentially.
The append / write API split feels a bit too specific here, given it's sort of a quirk of `emqx_s3_upload` API. On the other hand, it allows for a bit more flexible approach to retries (i.e. append does not usually fail, while write does).

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -0,0 +1,183 @@
...
+}.
+
+-type impl_specific_upload_state() :: term().
+
+-callback init_upload_state(buffer(), map()) -> impl_specific_upload_state().
+
+-callback process_append(iodata(), impl_specific_upload_state()) ->
+    {ok, impl_specific_upload_state()}.
+
+-callback process_write(impl_specific_upload_state()) -> impl_specific_upload_state().
```
keynslug, line 50: From the generalization standpoint, feels a bit weird: I'd expect that `append/2` does not fail (because it basically implies no IO) but `write/1` does.

apps/emqx_bridge_s3/src/emqx_bridge_s3_connector.erl
```diff
@@ -376,3 +391,83 @@ render_content(Template, Data) ->
...
+-spec lookup(emqx_template:accessor(), {_Name, buffer()}) ->
+    {ok, integer() | string()} | {error, undefined}.
+lookup([<<"action">>], {Name, _Buffer}) ->
+    {ok, mk_fs_safe_string(Name)};
+lookup(Accessor, {_Name, Buffer = #buffer{}}) ->
+    lookup_buffer_var(Accessor, Buffer);
+lookup(_Accessor, _Context) ->
+    {error, undefined}.
+
+lookup_buffer_var([<<"datetime">>, Format], #buffer{since = Since}) ->
```
keynslug, line 452: Nit. I wonder if it's possible to avoid relying on record definition here. It's essentially an implementation detail, feels awkward to pull it to that level. Admittedly, this is something that can be refactored later.

keynslug (commented):
🔥 Looks great. Mostly nits.

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
thalesmg: `transfer` sounds good to me.  :smile_cat:

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -0,0 +1,188 @@
...
+    ok.
+
+%%
+
+handle_msg({system, From, Msg}, Delivery, Parent, Debug) ->
+    sys:handle_system_msg(Msg, From, Parent, ?MODULE, Debug, Delivery);
+handle_msg({'EXIT', Parent, Reason}, Delivery, Parent, Debug) ->
+    system_terminate(Reason, Parent, Debug, Delivery);
+handle_msg(_Msg, Delivery, Parent, Debug) ->
+    loop(Parent, Debug, Delivery).
```
zmstone, line 173: to allow hot-beam reload without sys calls.
```suggestion
    ?MODULE:loop(Parent, Debug, Delivery).
```

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -0,0 +1,188 @@
...
+
+-spec system_continue(pid(), [sys:debug_option()], state()) -> no_return().
+system_continue(Parent, Debug, Delivery) ->
+    loop(Delivery, Parent, Debug).
```
zmstone, line 177:
```suggestion
    ?MODULE:loop(Delivery, Parent, Debug).
```

apps/emqx_connector_aggregator/BSL.txt
```diff
@@ -0,0 +1,94 @@
+Business Source License 1.1
+
+Licensor:             Hangzhou EMQ Technologies Co., Ltd.
+Licensed Work:        EMQX Enterprise Edition
+                      The Licensed Work is (c) 2023
```
savonarola, line 5:
```suggestion
                      The Licensed Work is (c) 2024
```

apps/emqx_connector_aggregator/BSL.txt
```diff
@@ -0,0 +1,94 @@
...
+                      Hangzhou EMQ Technologies Co., Ltd.
+Additional Use Grant: Students and educators are granted right to copy,
+                      modify, and create derivative work for research
+                      or education.
+Change Date:          2028-01-26
```
savonarola, line 10: nit: +4 years from `now()`

zmstone (commented):
a bit lat to review this feature.
I wonder why the aggregator is designed to store erlang terms but not the formatted data.
e.g. for csv, why not append formatted rows to the buffer file.

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.app.src
```diff
@@ -0,0 +1,14 @@
+{application, emqx_connector_aggregator, [
+    {description, "EMQX Enterprise Connector Data Aggregator"},
+    {vsn, "0.1.0"},
+    {registered, []},
+    {applications, [
+        kernel,
+        stdlib,
+        emqx_resource
```
savonarola, line 8: Curious: why do we need to depend on `emqx_resource`?

keynslug:
> I wonder why the aggregator is designed to store erlang terms but not the formatted data.

Primary reason: because initial requirements had other, non-stream-friendly containers for uploads (i.e. Parquet). Secondary reason: make writes cheaper.

Another positive side-effect is that it's a bit simpler approach when changes to bridge config and rules are taken into account.

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -0,0 +1,188 @@
...
+-export([start_link/3]).
+
+%% Internal exports
+-export([
+    init/4,
+    loop/3
+]).
+
+%% Sys
+-export([
```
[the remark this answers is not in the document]
keynslug: I initially thought that this way it might look simpler in the end, when compared to `gen_statem` with its overhead for basically an "interruptible task". But apparently it does not, so that's probably a good idea to adopt them.

zmstone:
> Parquet

got it, thanks.

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -0,0 +1,188 @@
...
+    format_status/2
+]).
+
+-record(delivery, {
+    name :: _Name,
+    callback_module :: module(),
+    container :: emqx_connector_aggreg_csv:container(),
+    reader :: emqx_connector_aggreg_buffer:reader(),
+    transfer :: transfer_state(),
+    empty :: boolean()
```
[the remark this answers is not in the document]
thalesmg: These were just moved from the previous locations without significant modifications.  Maybe we could do that improvement in a follow up?

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.app.src
thalesmg: Good point, I think this app doesn't actually use it.  I'll remove it.

savonarola (approved):

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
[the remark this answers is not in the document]
zmstone: ok

zmstone (approved):

emqxqa:
[Test Execution]()

PR #12989, opened 2024-05-08
fix(plugin): add a backup for the plugin config file

JimMoen:
Fixes <issue-or-jira-number>

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_plugins/src/emqx_plugins.erl
+++ b/apps/emqx_plugins/src/emqx_plugins.erl
@@ -94,6 +94,8 @@
 -define(RAW_BIN, binary).
 -define(JSON_MAP, json_map).
 
+-define(MAX_KEEP_BACKUP_CONFIGS, 10).
+
 %% "my_plugin-0.1.0"
 -type name_vsn() :: binary() | string().
 %% the parse result of the JSON info file
@@ -287,7 +289,7 @@ get_config(NameVsn, #{format := ?CONFIG_FORMAT_MAP}, Default) ->
 %% the avro Json Map and plugin config ALWAYS be valid before calling this function.
 put_config(NameVsn, AvroJsonMap, _DecodedPluginConfig) ->
     AvroJsonBin = emqx_utils_json:encode(AvroJsonMap),
-    ok = write_avro_bin(NameVsn, AvroJsonBin),
+    ok = backup_and_write_avro_bin(NameVsn, AvroJsonBin),
     ok = persistent_term:put(?PLUGIN_PERSIS_CONFIG_KEY(NameVsn), AvroJsonMap),
     ok.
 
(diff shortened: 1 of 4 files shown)


apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -1057,8 +1059,69 @@ maybe_create_config_dir(NameVsn) ->
...
+    %% e.g. read-only file system
+    Path = avro_config_file(NameVsn),
+    _ = filelib:ensure_dir(Path),
+    TmpFile = Path ++ ".tmp",
+    case file:write_file(TmpFile, AvroBin) of
+        ok ->
+            backup_and_replace(Path, TmpFile);
+        {error, Reason} ->
+            ?SLOG(error, #{
+                msg => "failed_to_save_conf_file",
```
thalesmg, line 1075:
```suggestion
                msg => "failed_to_save_plugin_conf_file",
```

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -1057,8 +1059,69 @@ maybe_create_config_dir(NameVsn) ->
             {error, {mkdir_failed, ConfigDir, Reason}}
     end.
 
-write_avro_bin(NameVsn, AvroBin) ->
-    ok = file:write_file(avro_config_file(NameVsn), AvroBin).
+%% @private Backup the current config to a file with a timestamp suffix and
+%% then save the new config to the config file.
+backup_and_write_avro_bin(NameVsn, AvroBin) ->
+    %% this may fail, but we don't care
+    %% e.g. read-only file system
```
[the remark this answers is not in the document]
JimMoen: `ensure_dir` relates to `Path`

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -1082,30 +1142,38 @@ read_file_fun(Path, ErrMsg, #{read_mode := ?JSON_MAP}) ->
     end.
 
 %% Directorys
+-spec plugin_dir(name_vsn()) -> string().
 plugin_dir(NameVsn) ->
-    filename:join([install_dir(), NameVsn]).
+    wrap_list_path(filename:join([install_dir(), NameVsn])).
 
+-spec plugin_config_dir(name_vsn()) -> string().
 plugin_config_dir(NameVsn) ->
```
zmstone, line 1150: Not related to this PR, but something we might have missed before for plugins in general:
we have to think about how to support upgrades if name-vsn is in the path to plugin data directory.

We should probably put plugins data in EMQX's data directory, and not include `vsn` part in the path,
so that when a plugin upgrades from for example `0.1.0` to `0.1.1`, it can still load old config.

In the plugin example, we can probably put a version number in the config schema, so to encourage developers to think about compatibility.
e.g. for configs, there is a `version` field, so the plugin code can perform upgrade-on-read.

zmstone (approved):

apps/emqx_plugins/src/emqx_plugins.erl
JimMoen: But this cannot handle the situation where the schema is changed when the plug-in version is upgraded.
In other words, schema, configuration, and plugin version should be strongly related.
But in order to preserve the old configuration during the plugin upgrade process. We can store it in the data directory.
nit: Upgrading means deleting the old version of the plugin and installing the new version.

zmstone: the plugin should export a upgrade_config callback which will be called by plugin framework after loading the config, but before checking the config against the schema.

emqxqa:
[Test Execution]()

PR #12994, opened 2024-05-08
fix(sessds): use milder log level for regular error conditions

keynslug:
Release version: v5.7

## Summary

Use milder log level for regular error conditions, especially when such events are emitted in (potentially) tight loops, i.e. pulling messages from iterators pointing to remote shards.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -622,7 +622,7 @@ replay_streams(Session0 = #{replay := [{StreamKey, Srs0} | Rest]}, ClientInfo) -
             replay_streams(Session#{replay := Rest}, ClientInfo);
         {error, recoverable, Reason} ->
             RetryTimeout = ?TIMEOUT_RETRY_REPLAY,
-            ?SLOG(warning, #{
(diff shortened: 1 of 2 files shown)


thalesmg (approved):
Just a thought: should we use [log throttling](https://github.com/emqx/emqx/blob/6e5d04e9faba6be22444cc3553e4ff3b359e9623/apps/emqx_conf/src/emqx_conf_schema.erl#L81-L88) for those messages that are in tight loops?

keynslug:
> Just a thought: should we use [log throttling](https://github.com/emqx/emqx/blob/6e5d04e9faba6be22444cc3553e4ff3b359e9623/apps/emqx_conf/src/emqx_conf_schema.erl#L81-L88) for those messages that are in tight loops?

I think it's actually better not to throttle debug level events, since once throttled they become less useful for debugging. On the other hand, by shoving those kinds of events to the debug logs we risk hiding from the user some conditions where the cluster is in half-broken state. So that's a good question. I basically adopted the same basic approach the egress is currently using, though it's probably worth revisiting later.

emqxqa:
[Test Execution]()

PR #13003, opened 2024-05-09
fix(aggregator): namespace aggregator ids with action type

thalesmg:
Release version: e5.7

## Summary

Otherwise, actions of different types but same names will clash when starting the aggregator supervision tree.

## PR Checklist

### DIFF
--- a/apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
+++ b/apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
@@ -25,13 +25,15 @@
 ]).
 
 -record(delivery, {
-    name :: _Name,
+    id :: id(),
     callback_module :: module(),
     container :: emqx_connector_aggreg_csv:container(),
     reader :: emqx_connector_aggreg_buffer:reader(),
     transfer :: transfer_state(),
     empty :: boolean()
 }).
+
+-type id() :: term().
 
 -type state() :: #delivery{}.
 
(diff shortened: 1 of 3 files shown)


apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -25,14 +25,16 @@
 ]).
 
 -record(delivery, {
-    name :: _Name,
+    id :: aggreg_id(),
```
keynslug, line 28: Nit. This reads like _delivery id_ rather than _which aggregator this delivery was spawned by_.

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
```diff
@@ -25,14 +25,16 @@
...
     callback_module :: module(),
     container :: emqx_connector_aggreg_csv:container(),
     reader :: emqx_connector_aggreg_buffer:reader(),
     transfer :: transfer_state(),
     empty :: boolean()
 }).
 
+-type aggreg_id() :: {_Type, _Name}.
```
keynslug, line 36: Also I feel that on that level it's not really correct to narrow this type that much, it's currently leaking from `emqx_bridge_s3_connector`, which is not part of this library. On this level it's the same as `emqx_connector_aggregator`'s name, which has basically no requirements on how it should look like.

keynslug (approved):

apps/emqx_connector_aggregator/src/emqx_connector_aggreg_delivery.erl
keynslug: Sorry for not being clear enough: I meant that `#delivery.id` sounds like it's something that should uniquely identify a delivery. But it's actually not: there can be multiple deliveries with the same `#delivery.id`.

But it's just a nit after all, not that important.

thalesmg: Aha, sorry, I missed the point.  :sweat_smile: 

I'll fix the field and binding names in a follow up then.  Thanks!

PR #13017, opened 2024-05-10
fix(subs mgmt api): attempt to return mixed clients ds/non-ds in the same page

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12294

Release version: v/e5.7

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
@@ -443,13 +443,13 @@ maybe_fetch_from_second_query(Params) ->
     } = Params,
     NumRows1 = length(Data1),
     {Data, HN} =
-        case NumRows1 >= Limit of
+        case (NumRows1 >= Limit) orelse HN1 of
             true ->
                 {Data1, HN1 orelse C2 > 0};
             false ->
                 #{data := Data2, meta := #{hasnext := HN2}} =
                     Q2(QString0#{<<"limit">> := Limit - NumRows1}),
-                {Data1 ++ Data2, HN1 or HN2}
+                {Data1 ++ Data2, HN2}
         end,
     #{
         data => Data,
--- a/apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
@@ -310,7 +310,12 @@ consume_n_matching(Map, Pred, N, S) ->
 consume_n_matching(_Map, _Pred, _N, [], Acc) ->
     {lists:reverse(Acc), []};
 consume_n_matching(_Map, _Pred, 0, S, Acc) ->
-    {lists:reverse(Acc), S};
+    case emqx_utils_stream:next(S) of
+        [] ->
+            {lists:reverse(Acc), []};
+        _ ->
+            {lists:reverse(Acc), S}
+    end;
 consume_n_matching(Map, Pred, N, S0, Acc) ->
     case emqx_utils_stream:next(S0) of
         [] ->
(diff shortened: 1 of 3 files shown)


apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
```diff
@@ -396,10 +401,21 @@ merge_queries(QString0, Q1, Q2) ->
             Q2Page = ceil(C1 / Limit),
             case Page =< Q2Page of
                 true ->
-                    #{data := Data, meta := #{hasnext := HN}} = Q1(QString0),
+                    #{data := Data1, meta := #{hasnext := HN1}} = Q1(QString0),
+                    NumRows1 = length(Data1),
+                    case NumRows1 >= Limit of
+                        true ->
+                            Data = Data1,
```
zmstone, line 408: this kind of variable bind is very confusing, it looks like assertion

suggestion: `{Data, HasNext} = case ... of`
or even better, write a func to handle the case part.

thalesmg: :heavy_check_mark:

apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
```diff
@@ -310,7 +310,12 @@ consume_n_matching(Map, Pred, N, S) ->
 consume_n_matching(_Map, _Pred, _N, [], Acc) ->
     {lists:reverse(Acc), []};
 consume_n_matching(_Map, _Pred, 0, S, Acc) ->
-    {lists:reverse(Acc), S};
+    case emqx_utils_stream:next(S) of
+        [] ->
+            {lists:reverse(Acc), []};
```
savonarola, line 315:
```suggestion
            {lists:reverse(Acc), emqx_utils_stream:empty()};
```
?

apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
```diff
@@ -421,6 +431,31 @@ merge_queries(QString0, Q1, Q2) ->
...
+    } = Params,
+    NumRows1 = length(Data1),
+    {Data, HN} =
+        case NumRows1 >= Limit of
+            true ->
+                {Data1, HN1 orelse C2 > 0};
+            false ->
+                #{data := Data2, meta := #{hasnext := HN2}} =
+                    Q2(QString0#{<<"limit">> := Limit - NumRows1}),
+                {Data1 ++ Data2, HN1 or HN2}
```
savonarola, line 452: Curious: if we assume that HN1 may be true, why do we use the second query?

thalesmg: In this arm of the `case`, the number of returned rows from `Q1` is fewer than `Limit`, so I think we might assume `HN1 = false` in most cases.  But I think you have a good point: this situation is not necessarily wrong, I'll check for this condition.  Thanks!  :beers:

apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
thalesmg: Actually, that's part of the fix: there is this check to determine if `hasnext = true`: https://github.com/thalesmg/emqx/blob/0dee2d67b1645b2ae2bcb46d2817c56b03b89f6f/apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl#L258

apps/emqx_management/src/emqx_mgmt_api_subscriptions.erl
thalesmg: :heavy_check_mark:

zmstone (approved):

emqxqa:
[TestExecution]()

PR #13020, opened 2024-05-11
Fix qlc leak 

savonarola:
**DO NOT MERGE**

Adhoc 5.6.0 fix for https://emqx.atlassian.net/browse/EMQX-12319


savonarola:
In master, merged as https://github.com/emqx/emqx/pull/13025

PR #13025, opened 2024-05-13
Get rid of qlc in retainer

savonarola:
Fixes [EMQX-12367](https://emqx.atlassian.net/browse/EMQX-12367)

Release version: v5.8.0/e5.8.0

## PR Checklist

[EMQX-12367]: https://emqx.atlassian.net/browse/EMQX-12367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_retainer/src/emqx_retainer_mnesia.erl
+++ b/apps/emqx_retainer/src/emqx_retainer_mnesia.erl
@@ -176,14 +176,14 @@ clear_expired(_) ->
 clear_expired() ->
     NowMs = erlang:system_time(millisecond),
     S0 = ets_stream(?TAB_MESSAGE),
-    S1 = stream_filter(
+    S1 = emqx_utils_stream:filter(
         fun(#retained_message{expiry_time = ExpiryTime}) ->
             ExpiryTime =/= 0 andalso ExpiryTime < NowMs
         end,
         S0
     ),
     DirtyWriteIndices = dirty_indices(write),
-    stream_foreach(
+    emqx_utils_stream:foreach(
         fun(RetainedMsg) ->
             delete_message_with_indices(RetainedMsg, DirtyWriteIndices)
         end,
@@ -198,7 +198,7 @@ delete_message(_State, Topic) ->
         true ->
             S = search_stream(Tokens, 0),
             DirtyWriteIndices = dirty_indices(write),
-            stream_foreach(
+            emqx_utils_stream:foreach(
                 fun(RetainedMsg) -> delete_message_with_indices(RetainedMsg, DirtyWriteIndices) end,
                 S
             )
@@ -207,7 +207,7 @@ delete_message(_State, Topic) ->
 read_message(_State, Topic) ->
     {ok, read_messages(Topic)}.
 
-match_messages(State, Topic, undefined) ->
+match_messages(_State, Topic, undefined) ->
     Tokens = topic_to_tokens(Topic),
     Now = erlang:system_time(millisecond),
     S = msg_stream(search_stream(Tokens, Now)),
@@ -358,7 +358,7 @@ search_stream(Index, FilterTokens, Now) ->
         fun(#retained_index{key = Key}) -> emqx_retainer_index:restore_topic(Key) end,
         IndexRecordStream
     ),
-    MatchingTopicStream = stream_filter(
+    MatchingTopicStream = emqx_utils_stream:filter(
         fun(TopicTokens) -> match(IsExactMs, TopicTokens, FilterTokens) end,
         TopicStream
     ),
@@ -366,7 +366,7 @@ search_stream(Index, FilterTokens, Now) ->
         fun(TopicTokens) -> ets:lookup(?TAB_MESSAGE, TopicTokens) end,
         MatchingTopicStream
     ),
-    FoundAndValidLookedUpRetainMsgStream = stream_filter(
+    FoundAndValidLookedUpRetainMsgStream = emqx_utils_stream:filter(
         fun
             ([#retained_message{expiry_time = ExpiryTime}]) ->
                 ExpiryTime =:= 0 orelse ExpiryTime >= Now;
@@ -708,33 +708,7 @@ are_indices_updated(Indices) ->
 ets_stream(Tab) ->
     emqx_utils_stream:ets(
         fun
-            (undefined) -> ets:match_object(Tab, '$1', 1);
+            (undefined) -> ets:match_object(Tab, '_', 1);
             (Cont) -> ets:match_object(Cont)
         end
     ).
-
-stream_foreach(F, S) ->
-    case emqx_utils_stream:next(S) of
-        [X | Rest] ->
-            F(X),
-            stream_foreach(F, Rest);
-        [] ->
-            ok
-    end.
-
-%% TODO: move to emqx_utils_stream
-stream_filter(F, S) ->
-    FilterNext = fun FilterNext(St) ->
-        case emqx_utils_stream:next(St) of
-            [X | Rest] ->
-                case F(X) of
-                    true ->
-                        [X | stream_filter(F, Rest)];
-                    false ->
-                        FilterNext(Rest)
-                end;
-            [] ->
-                []
-        end
-    end,
-    fun() -> FilterNext(S) end.
--- a/apps/emqx_retainer/src/emqx_retainer_mnesia.erl
+++ b/apps/emqx_retainer/src/emqx_retainer_mnesia.erl
@@ -23,7 +23,6 @@
 -include_lib("emqx/include/logger.hrl").
 -include_lib("snabbkaffe/include/snabbkaffe.hrl").
 -include_lib("stdlib/include/ms_transform.hrl").
--include_lib("stdlib/include/qlc.hrl").
 
 %% emqx_retainer callbacks
 -export([
@@ -64,7 +63,6 @@
 
 -define(META_KEY, index_meta).
 
--define(CLEAR_BATCH_SIZE, 1000).
 -define(REINDEX_BATCH_SIZE, 1000).
 -define(REINDEX_DISPATCH_WAIT, 30000).
 -define(REINDEX_RPC_RETRY_INTERVAL, 1000).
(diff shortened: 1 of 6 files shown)


apps/emqx_utils/src/emqx_utils_stream.erl
```diff
@@ -22,6 +22,8 @@
     list/1,
     mqueue/1,
     map/2,
+    filter/2,
+    foreach/2,
```
keynslug, line 26: Tiny nit. Move to `%% Evaluating` section?

apps/emqx_utils/src/emqx_utils_stream.erl
```diff
@@ -93,6 +95,33 @@ map(F, S) ->
...
+                    true ->
+                        [X | filter(F, Rest)];
+                    false ->
+                        FilterNext(Rest)
+                end;
+            [] ->
+                []
+        end
+    end,
+    fun() -> FilterNext(S) end.
```
keynslug, line 113: Q: Worth to simplify to?
```suggestion
    fun FilterNext(F, S) ->
        case next(S) of
            [X | Rest] ->
                case F(X) of
                    true ->
                        [X | filter(F, Rest)];
                    false ->
                        FilterNext(F, Rest)
                end;
            [] ->
                []
        end
    end.
```

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -333,58 +327,63 @@ do_store_retained_index(Key, ExpiryTime) ->
...
+    ),
+    TopicStream = emqx_utils_stream:map(
+        fun(#retained_index{key = Key}) -> emqx_retainer_index:restore_topic(Key) end,
+        IndexRecordStream
+    ),
+    MatchingTopicStream = emqx_utils_stream:filter(
+        fun(TopicTokens) -> match(IsExactMs, TopicTokens, FilterTokens) end,
+        TopicStream
+    ),
+    LookedUpRetainMsgStream = emqx_utils_stream:map(
```
keynslug, line 365: Q: Could something like a (currently nonexistent) `flatmap/2` help here a bit with readability?

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -177,91 +175,87 @@ clear_expired(_) ->
...
             _ ->
                 Tokens = topic_to_tokens(Topic),
-                msg_table(search_table(Tokens, Now))
+                msg_stream(search_stream(Tokens, Now))
         end,
-    OrderedQH = qlc:sort(QH, {order, fun compare_message/2}),
-    Cursor = qlc:cursor(OrderedQH),
+    %% This is very inefficient, but we are limited with inherited API
+    S1 = emqx_utils_stream:list(
+        lists:sort(
```
keynslug, line 243: Ouch. This will always consume process heap, when compared to `qlc:sort/2`, which can offload sorts to disks AFAIK, right?

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -706,3 +704,11 @@ are_indices_updated(Indices) ->
         _ ->
             false
     end.
+
+ets_stream(Tab) ->
+    emqx_utils_stream:ets(
+        fun
+            (undefined) -> ets:match_object(Tab, '_', 1);
```
keynslug, line 711: I suspect it would be a good idea to see how this rewrite affects performance.

I have a feeling that this approach (iterating one record at a time) will hit it the hardest, however it doesn't seem too hard to adapt it to larger pages though, on the `emqx_utils_stream` side.

keynslug (approved):

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
savonarola: Yes, that seems worth implementing

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
savonarola: `qlc`'s possible benefits were actually accidental 😓
Pre-`qlc` implementation is also supposed in-memory sorting.
To provide any guarantees for this method, we should either refuse sorting or add a time-based index.

apps/emqx_utils/src/emqx_utils_stream.erl
savonarola: 🤔 this gives a 2-arity function while stream is a 0-arity one

keynslug: Oh right. Disregard please.

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -177,91 +175,87 @@ clear_expired(_) ->
...
-        more ->
-            case qlc_next_answers(Cursor, Limit) of
-                {closed, Rows} ->
-                    {ok, false, Rows};
-                {more, Rows} ->
-                    qlc:delete_cursor(Cursor),
-                    {ok, true, Rows}
-            end
+    case emqx_utils_stream:consume(NSkip, S1) of
+        {_, S2} ->
```
keynslug, line 250: Another idea: could not-yet-implemented `emqx_utils_stream:skip/2` help here by making things a bit nicer and not wasting extra memory?

apps/emqx_utils/src/emqx_utils_stream.erl
savonarola: ✅

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
savonarola: ✅

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
savonarola: ✅

thalesmg (dismissed):

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -333,58 +325,56 @@ do_store_retained_index(Key, ExpiryTime) ->
...
+        fun(TopicTokens) -> match(IsExactMs, TopicTokens, FilterTokens) end,
+        TopicStream
+    ),
+    RetainMsgStream = emqx_utils_stream:chainmap(
+        fun(TopicTokens) -> emqx_utils_stream:list(ets:lookup(?TAB_MESSAGE, TopicTokens)) end,
+        MatchingTopicStream
+    ),
+    ValidRetainMsgStream = emqx_utils_stream:filter(
+        fun(#retained_message{expiry_time = ExpiryTime}) ->
+            ExpiryTime =:= 0 orelse ExpiryTime >= Now
```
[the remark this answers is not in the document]
savonarola: ✅

apps/emqx_utils/src/emqx_utils_stream.erl
```diff
@@ -102,6 +104,64 @@ map(F, S) ->
...
+drop(N, S) ->
+    DropNext = fun DropNext(M, St) ->
+        case next(St) of
+            [_X | Rest] when M > 0 ->
+                DropNext(M - 1, Rest);
+            Next ->
+                Next
+        end
+    end,
+    fun() -> DropNext(N, S) end.
```
keynslug, line 145: Q: Wouldn't it make more sense to turn this function into an eager evaluator, similar to `consume/2`? It shouldn't even change the spec.

keynslug (dismissed):
🔥

apps/emqx_utils/src/emqx_utils_stream.erl
savonarola: Depending on whether we put it into the Evaluating or the Combinator section, both variants seem to make sense.

When constructing a stream, it may be unexpected or undesirable to have something evaluated (e.g., when we construct a csv-reading-parsing stream and want to skip the header, we probably do not expect eager reading of a file)

On the other hand, in the actual use case, we just scroll the stream to the desired page, i.e. consume it.

IMO lazy variant is more generic because it wraps eager dropping in just a single call, so the overhead of 
consume(drop(...)) over eager drop is tiny 🤔

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -177,91 +177,83 @@ clear_expired(_) ->
...
-    QH = msg_table(search_table(Tokens, Now)),
+    S = msg_stream(search_stream(Tokens, Now)),
     case batch_read_number() of
         all_remaining ->
-            {ok, qlc:eval(QH), undefined};
+            {ok, emqx_utils_stream:consume(S), undefined};
         BatchNum when is_integer(BatchNum) ->
-            Cursor = qlc:cursor(QH),
-            match_messages(State, Topic, {Cursor, BatchNum})
+            match_messages(undefined, Topic, {S, BatchNum})
```
zmstone, line 220: If State is to be kept for future use.

```suggestion
            match_messages(State, Topic, {S, BatchNum})
```

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -177,91 +177,83 @@ clear_expired(_) ->
...
-        {closed, Rows} ->
-            {ok, Rows, undefined};
-        {more, Rows} ->
-            {ok, Rows, {Cursor, BatchNum}}
+match_messages(_State, _Topic, {S0, BatchNum}) ->
+    case emqx_utils_stream:consume(BatchNum, S0) of
+        {Rows, S1} ->
+            {ok, Rows, {S1, BatchNum}};
+        Rows when is_list(Rows) ->
+            {ok, Rows, undefined}
```
zmstone, line 227: when does this clause happen ?

zmstone: nvm, it's a part of the stream api contract.

zmstone (dismissed):

thalesmg (dismissed):

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2519)

PR #13043, opened 2024-05-14
fix(kafka trace): do not include query type in trace entry

kjellwinblad:
For Kafka that has its internal buffering the on_query_async callback is used both for sync and async queries (for sync queries the caller waits for a response directly after calling on_query_async). Therefore, before this commit, the trace included incorrect query type information. This has been fixed by removing the query type information from the trace. This should be okay since the query type is not essential information and can be derived from the configuration.

Fixes:
https://emqx.atlassian.net/browse/EMQX-12363

Release version: v/e5.?

## PR Checklist


zmstone (approved):

emqxqa:
[TestExecution]()

PR #13055, opened 2024-05-16
test: bump up emqx_builder to add zsh to erlang container

zhongwencool:
Fixes <issue-or-jira-number>

No need to test, this just adds zsh to the development erlang docker image for convenient development and debugging.

Release version: v/e5.8.0

## PR Checklist


scripts/ct/run.sh
```diff
@@ -316,7 +319,7 @@ set +e
 if [ "$STOP" = 'yes' ]; then
     $DC down --remove-orphans
 elif [ "$ATTACH" = 'yes' ]; then
-    docker exec -it "$ERLANG_CONTAINER" bash
+    docker exec -it -u root "$ERLANG_CONTAINER" zsh
```
zhongwencool, line 322: Emqx-builder installs zsh using root. Add .zshrc in /root/.zshrc.Some docker-compose.yaml start erlang by the emqx user, which makes zsh unavailable.So if we start with attach, we use root directly.

Rory-Z: maybe can add `/home/emqx/.zshrc` ?
but why do we need zsh ?

thalesmg: I think it's best to not run the shell as root, specially when developing, because it'll create lots of root-owned files in the host file system that then require root to modify/delete.

@Rory-Z 's idea sounds good, or perhaps there's a way to run `zsh` with the normal user?

zhongwencool: https://github.com/emqx/emqx/blob/master/.ci/docker-compose-file/docker-compose.yaml#L24
We already mount /emqx to emqx dev home. 
Q1: How to  add `.zshrc` and `.oh-my-zsh` to this dir.
Q2: Sometime we use root to run, if we add .zshrc to emqx user not root, how to enable zsh when use root user ?
https://github.com/emqx/emqx/blob/master/.ci/docker-compose-file/docker-compose.yaml#L31

zhongwencool: >  but why do we need zsh ?
this builder image is only used for development and running CI tests. It will not be released to users. So adding zsh（oh-my-zsh） is to improve the development experience.

Rory-Z: If just improve the development experience, maybe we can add `.zshrc` by ourself,  because just install zsh is not enought, must have a good zshrc to have a good experience.

Other: There may be some differences between zsh and bash in subtle places. Please force our scripts to use bash.

PR #13063, opened 2024-05-17
fix(plugin): serde not found on new-joined nodes

JimMoen:
Fixes [EMQX-12349](https://emqx.atlassian.net/browse/EMQX-12349) [EMQX-12350](https://emqx.atlassian.net/browse/EMQX-12350)

Release version: v/e5.7.0

## PR Checklist

- ~[ ] Added tests for the changes~
- ~[ ] Added property-based tests for code which performs user input validation~
- ~[ ] Changed lines covered in coverage report~
- ~[ ] Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files~


apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -1148,7 +1181,7 @@ plugin_dir(NameVsn) ->
 
 -spec plugin_config_dir(name_vsn()) -> string().
 plugin_config_dir(NameVsn) ->
-    wrap_list_path(filename:join([plugin_dir(NameVsn), "data", "configs"])).
+    wrap_list_path(filename:join([emqx:data_dir(), "plugins", NameVsn])).
```
zhongwencool, line 1184: Good idea to save to `data/plugins` dir.
Please ensure all config is deleted when uninstall/delete plugins.

apps/emqx_plugins/src/emqx_plugins_app.erl
```diff
@@ -27,8 +27,8 @@
 
 start(_Type, _Args) ->
     %% load all pre-configured
-    ok = emqx_plugins:ensure_started(),
     {ok, Sup} = emqx_plugins_sup:start_link(),
+    ok = emqx_plugins:ensure_started(),
```
zhongwencool, line 31: Maybe try catch and log this on start.
we can't let plugin start failed lead emqx start failed.
```
Erlang/OTP 26 [erts-14.2.1] [emqx] [64-bit] [smp:8:8] [ds:8:8:8] [async-threads:4] [jit:ns]

Listener tcp:default on 0.0.0.0:1883 started.
Listener ssl:default on 0.0.0.0:8883 started.
Listener ws:default on 0.0.0.0:8083 started.
Listener wss:default on 0.0.0.0:8084 started.
Listener tcp:default on 0.0.0.0:1883 stopped.
Listener ssl:default on 0.0.0.0:8883 stopped.
Listener ws:default on 0.0.0.0:8083 stopped.
Listener wss:default on 0.0.0.0:8084 stopped.
Kernel pid terminated (application_controller) ("{application_start_failure,emqx_machine,{{shutdown,{failed_to_start_child,emqx_machine_boot,{'EXIT',{{failed_to_start_app,emqx_plugins,{emqx_plugins,{bad_return,{{emqx_plugins_app,start,[normal,[]]},{'EXIT',{#{function => fun emqx_plugins:do_ensure_started/1,errors => [{\"emqx_xxx-1.2.2\",#{reason => {bad_return,{{emqx_xxx_app,start,[normal,[]]},{'EXIT',{{badkey,<<\"auth_url\">>},[{erlang,map_get,[<<\"auth_url\">>,#{}],[{error_info,#{module => erl_erts_errors}}]},{emqx_xxx_app,config,0,[{file,\"/Users/zw/emqx/emqx_xxx/src/emqx_xxx_app.erl\"},{line,54}]},{emqx_xxx_app,start,2,[{file,\"/Users/zw/emqx/emqx_xxx/src/emqx_xxx_app.erl\"},{line,25}]},{application_master,start_it_old,4,[{file,\"application_master.erl\"},{line,293}]}]}}}},error_msg => \"failed_to_start_plugin_app\",app => emqx_xxx,err_app => emqx_xxx}}]},[{emqx_plugins,for_plugins,1,[{file,\"emqx_plugins.erl\"},{line,1016}]},{emqx_plugins_app,start,2,[{f
```

zhongwencool (changes requested):

apps/emqx_plugins/src/emqx_plugins.erl
JimMoen: maybe we can not 🤔
Consider the situation of plugin upgrade:
The new version of the plugin still needs to use the configuration of the old version.

* Upgrade: The current implementation is that you must first delete the old version of the plug-in and then install the new version.

zhongwencool: 1. Currently, I don't see anywhere deleting this `plugins/NameVersion/*`. I think it should be deleted when the user deletes the package.
2. Don't see any code do copy the old configuration to the new version when upgrading. And the new version of the configuration may not be compatible with the old version, just let the users handle it. **If the use already uninstall the package, it's mean all code and config should be deleted**, If we want to handle the upgrade, we should add a new action call **upgrade** in the future.

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -1013,8 +1013,11 @@ configured() ->
 
 for_plugins(ActionFun) ->
     case lists:flatmap(fun(I) -> for_plugin(I, ActionFun) end, configured()) of
-        [] -> ok;
-        Errors -> erlang:error(#{function => ActionFun, errors => Errors})
+        [] ->
+            ok;
+        Errors ->
+            ?SLOG(error, #{function => ActionFun, errors => Errors}),
```
zhongwencool, line 1039: Maybe we should use `critical` instead of `error`.
The plugin not running properly is a very serious mistake.
Plan is to add a alarm type in `emqx_alarm` in the future.

zhongwencool: Also: we cant change this for_plugins/1 function, 
If we enable plugins from API failed ,we should crash and return the error to api.
We only catch the error when emqx start.

apps/emqx_plugins/src/emqx_plugins.erl
JimMoen: The config file should not be deleted due to plugin-upgrade. Actual behavior is removal-installed plugin replacement process.
If the `plugins/Name/*` (note: **NO Version** as suffix) files be removed after plugin uninstalled.
The new version of plugin could not find the previous `config.hocon` file.

HJianBo (approved):
LGTM

zhongwencool (approved):

emqxqa:
[TestExecution]()

PR #13069, opened 2024-05-17
feat: implement azure blob storage action

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12280

Release version: e5.8

## PR Checklist


qzhuyan (approved):

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_connector.erl
```diff
@@ -0,0 +1,763 @@
...
+    } = TransferState0,
+    case BufferSize > MaxBlockSize of
+        true ->
+            {IOData, NewBuffer} = take_chunk(Buffer, MaxBlockSize),
+            ?tp(azure_blob_storage_will_write_chunk, #{}),
+            do_process_write(IOData, NewBuffer, TransferState0);
+        false ->
+            NewBuffer = [],
+            do_process_write(Buffer, NewBuffer, TransferState0)
+    end.
```
keynslug, line 415: Apparently having this append / write split backfires slightly. 🫠

Honestly both of those things feel suboptimal:
1. Call Azure API on every append, which is usually on the scale of 1-10s of KiBs assuming moderate event rate.
2. Cut >4GiB large buffer somewhere close to its end, which will likely cause 2-3x of that in heap memory consumption. This most likely will never happen, but still.

I wonder what if:
1. Introduce `min_block_size` config parameter, with something like 10 MiB as the default.
2. During append: append as usual, but put the buffer under some `next_block` map entry if its size is larger that `min_block_size` (i.e. minimize latency impact / overhead / cost (?) of too frequent API requests).
3. However, if it's to become larger than `max_block_size` does not append but instead put existing buffer under `next_block` key (i.e. avoid cutting iolists).
4. During write, write `next_block` if it's defined.

On more thing: currently, having a single append larger than 4 GiB is impossible I think, because of [this](https://github.com/emqx/emqx/blob/d61f4078e2b1ecd9493d8469de8057bf24307b31/apps/emqx_connector_aggregator/src/emqx_connector_aggreg_buffer.erl#L50). Admittedly, it's sort of an implicitly defined limit, but I'd argue still could be relied upon to get rid of complexity of "cut off exactly this much of this `iolist()`".

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_connector.erl
```diff
@@ -0,0 +1,763 @@
...
+lookup_buffer_var([<<"datetime">>, Format], #{since := Since}) ->
+    {ok, format_timestamp(Since, Format)};
+lookup_buffer_var([<<"datetime_until">>, Format], #{until := Until}) ->
+    {ok, format_timestamp(Until, Format)};
+lookup_buffer_var([<<"sequence">>], #{seq := Seq}) ->
+    {ok, Seq};
+lookup_buffer_var([<<"node">>], #{}) ->
+    {ok, mk_fs_safe_string(atom_to_binary(erlang:node()))};
+lookup_buffer_var(_Binding, _Context) ->
+    {error, undefined}.
```
keynslug, line 476: Might be a good time to introduce something like a small `emqx_connector_aggreg_buffer_ctx` module, that implements `emqx_template` behaviour for `#buffer{}`s, and put this stuff (except for `<<"node">>`) there? Though we'd have to keep `mk_fs_safe_string/1` on this level. Not the prettiest solution, but could help minimize code duplication and the need for that weird `buffer_map()` type.

keynslug (commented):
🚀

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_action_schema.erl
```diff
@@ -0,0 +1,295 @@
...
+        {max_records,
+            hoconsc:mk(
+                pos_integer(),
+                #{
+                    required => false,
+                    default => 1_000_000,
+                    desc => ?DESC("aggregation_max_records")
+                }
+            )},
+        {max_block_size,
```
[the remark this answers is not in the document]
thalesmg: Indeed.  Moved to action parameters.  :heavy_check_mark:

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_connector.erl
thalesmg: :heavy_check_mark:

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_action_schema.erl
```diff
@@ -0,0 +1,310 @@
...
+        {max_records,
+            hoconsc:mk(
+                pos_integer(),
+                #{
+                    required => false,
+                    default => 1_000_000,
+                    desc => ?DESC("aggregation_max_records")
+                }
+            )},
+        {min_block_size,
```
keynslug, line 148: I feel this is also a bit out of place here in `aggregation` options. 🌝

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_action_schema.erl
```diff
@@ -0,0 +1,310 @@
...
+                    default => <<"10mb">>,
+                    importance => ?IMPORTANCE_HIDDEN,
+                    required => true,
+                    validator => fun block_size_validator/1
+                }
+            )}
+    ];
+fields(common_action_parameters) ->
+    [
+        {max_block_size,
```
keynslug, line 161: Q: Is there anything this parameter may affect for direct uploads? The idea is to prematurely drop the event on the floor if it's for some reason >4GiB in size?

keynslug (commented):
Great! A couple of nits.

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_action_schema.erl
thalesmg: Well, this one really only affects the aggregation part of the `aggregated` mode, differently from `max_block_size` which indeed meant Azure's API limits.  :thinking:

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_action_schema.erl
thalesmg: Good point, currently we don't check it, but this limit indeed is Azure's maximum upload per block.
(More precisely, for a single "Put Block Blob" call, it's possible to send 5000 MiB, and each "Put Blob" is 4000 MiB)

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_action_schema.erl
keynslug: I get your point, though I'd still think about the delivery process as being separate from the aggregation process.

thalesmg: I see.  I've moved it one level up here: https://github.com/emqx/emqx/pull/13069/commits/34d8432db0761003a386eb6c80da8e765fca0507

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_action_schema.erl
thalesmg: Added a check and test here: https://github.com/emqx/emqx/pull/13069/commits/a92eb3c60f89241460ae50dadfd0d65f85b8c1fa

apps/emqx_connector_aggregator/src/emqx_connector_aggregator_schema.erl
```diff
@@ -0,0 +1,89 @@
...
+            mk(
+                csv,
+                #{
+                    required => true,
+                    desc => ?DESC("container_csv")
+                }
+            )},
+        {column_order,
+            mk(
+                hoconsc:array(string()),
```
zmstone, line 68:
```suggestion
                hoconsc:array(binary()),
```

apps/emqx_connector_aggregator/src/emqx_connector_aggregator_schema.erl
```diff
@@ -0,0 +1,89 @@
...
+%%------------------------------------------------------------------------------
+
+%%------------------------------------------------------------------------------
+%% API
+%%------------------------------------------------------------------------------
+
+container() ->
+    {container,
+        hoconsc:mk(
+            %% TODO: Support selectors once there are more than one container.
```
zmstone, line 36: remove?

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_action_schema.erl
```diff
@@ -0,0 +1,310 @@
...
+        | fields(common_action_parameters)
+    ];
+fields(aggreg_parameters) ->
+    [
+        {mode, mk(aggregated, #{required => true, desc => ?DESC("aggregated_mode")})},
+        {aggregation, mk(ref(aggregation), #{required => true, desc => ?DESC("aggregation")})},
+        %% Container in the Azure Blob Storage domain, not aggregation.
+        {container,
+            mk(
+                string(),
```
zmstone, line 110: ?
```suggestion
                binary(),
```

apps/emqx_connector_aggregator/src/emqx_connector_aggregator_schema.erl
thalesmg: This is like this before, so I'm not sure we should change this.

apps/emqx_connector_aggregator/src/emqx_connector_aggregator_schema.erl
thalesmg: This comment was just transported from its previous location.

apps/emqx_bridge_azure_blob_storage/src/emqx_bridge_azure_blob_storage_action_schema.erl
thalesmg: This must be a `string()` for `erlazure`.

apps/emqx_connector_aggregator/src/emqx_connector_aggregator_schema.erl
thalesmg: i.e. also the product of a refactoring

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3355)

PR #13072, opened 2024-05-19
Idempotent drop_generation

ieQu1:
Fixes [EMQX-12402](https://emqx.atlassian.net/browse/EMQX-12402)

Release version: v/e5.7

## Summary

Various small, but important fixes.

- Add an option to execute read operations on the leader.
- `drop_generation` operation can be replayed multiple times by the replication layer, but it's not idempotent. This PR adds a workaround that avoids a crash when `drop_generation` doesn't succeed. In the future, however, we want to make `drop_generation` idempotent in a nicer way.
- Wrap storage layer events in a small structure containing the generation ID, to make sure events are handled by the same layout CBM & context that produced them. 
- Fix crash when storage event arrives to the dropped generation (there was a bug in `storage_layer:generation_at` function, which has been removed for good).
- Implement `format_status` callback for several workers to avoid dumping contents of the entire message buffer to the console.
- Move the responsibility of `end_of_stream` detection to the layout CBM. Previously storage layer used a heuristic: old generations that return an empty batch won't produce more data. This is, obviously, incorrect: for example, bitfield-LTS layout MAY return empty batch while waiting for safe cutoff time. 
- `reference` layout has been enabled in prod build. It could be useful for integration testing. 
- Fix incorrect epoch calculation in `bitfield_lts:handle_event` callback that lead to missed safe cutoff time updates, and effectively, subscribers being unable to fetch messages until a fresh batch was published.

## PR Checklist

[EMQX-12402]: https://emqx.atlassian.net/browse/EMQX-12402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_ds_schema.erl
+++ b/apps/emqx/src/emqx_ds_schema.erl
@@ -246,7 +246,8 @@ fields(layout_builtin_reference) ->
                 reference,
                 #{
                     'readOnly' => true,
-                    importance => ?IMPORTANCE_LOW
+                    importance => ?IMPORTANCE_LOW,
+                    desc => ?DESC(layout_builtin_reference_type)
                 }
             )}
     ].
@@ -257,6 +258,8 @@ desc(builtin_local_write_buffer) ->
     ?DESC(builtin_local_write_buffer);
 desc(layout_builtin_wildcard_optimized) ->
     ?DESC(layout_builtin_wildcard_optimized);
+desc(layout_builtin_reference) ->
+    ?DESC(layout_builtin_reference);
 desc(_) ->
     undefined.
 
--- a/apps/emqx/src/emqx_ds_schema.erl
+++ b/apps/emqx/src/emqx_ds_schema.erl
@@ -246,7 +246,7 @@ fields(layout_builtin_reference) ->
                 reference,
                 #{
                     'readOnly' => true,
-                    importance => ?IMPORTANCE_HIDDEN
+                    importance => ?IMPORTANCE_LOW
                 }
             )}
     ].
@@ -273,17 +273,12 @@ ds_schema(Options) ->
         Options
     ).
 
--ifndef(TEST).
-builtin_layouts() ->
-    [ref(layout_builtin_wildcard_optimized)].
--else.
 builtin_layouts() ->
     %% Reference layout stores everything in one stream, so it's not
     %% suitable for production use. However, it's very simple and
     %% produces a very predictabale replay order, which can be useful
     %% for testing and debugging:
     [ref(layout_builtin_wildcard_optimized), ref(layout_builtin_reference)].
--endif.
 
 sc(Type, Meta) -> hoconsc:mk(Type, Meta).
 
--- a/apps/emqx_durable_storage/src/emqx_ds_replication_layer_shard.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_replication_layer_shard.erl
@@ -115,11 +115,11 @@ get_servers_local_preferred(DB, Shard) ->
         Servers when is_list(Servers) ->
             ok
     end,
-    case lists:keyfind(node(), 2, Servers) of
+    case lists:keytake(node(), 2, Servers) of
         false ->
             Servers;
-        Local when is_tuple(Local) ->
-            [Local | lists:delete(Local, Servers)]
+        {value, Local, Rest} ->
+            [Local | Rest]
     end.
 
 lookup_leader(DB, Shard) ->
--- a/apps/emqx_durable_storage/src/emqx_ds_replication_layer_shard.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_replication_layer_shard.erl
@@ -28,8 +28,7 @@
 
 %% Dynamic server location API
 -export([
-    servers/3,
-    server/3
+    servers/3
 ]).
 
 %% Membership
@@ -83,15 +82,14 @@ server_name(DB, Shard, Site) ->
 
 %%
 
--spec servers(emqx_ds:db(), emqx_ds_replication_layer:shard_id(), Order) -> [server(), ...] when
-    Order :: leader_preferred | undefined.
-servers(DB, Shard, _Order = leader_preferred) ->
+-spec servers(emqx_ds:db(), emqx_ds_replication_layer:shard_id(), Order) -> [server()] when
+    Order :: leader_preferred | local_preferred | undefined.
+servers(DB, Shard, leader_preferred) ->
     get_servers_leader_preferred(DB, Shard);
+servers(DB, Shard, local_preferred) ->
+    get_servers_local_preferred(DB, Shard);
 servers(DB, Shard, _Order = undefined) ->
     get_shard_servers(DB, Shard).
-
-server(DB, Shard, _Which = local_preferred) ->
-    get_server_local_preferred(DB, Shard).
 
 get_servers_leader_preferred(DB, Shard) ->
     %% NOTE: Contact last known leader first, then rest of shard servers.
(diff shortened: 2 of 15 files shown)


apps/emqx_durable_storage/test/emqx_ds_storage_SUITE.erl
```diff
@@ -27,25 +27,6 @@ opts() ->
 
 %%
 
-t_idempotent_store_batch(_Config) ->
```
ieQu1, line 30: Well, this doesn't hold, since writing batches to the old generations was a bug.
@keynslug Is this property important for e.g. replication layer?

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -586,6 +589,24 @@ init({ShardId, Options}) ->
     commit_metadata(S),
     {ok, S}.
 
+format_status(#s{shard_id = ShardId, db = DB, cf_refs = CFRefs, schema = Schema, shard = Shard}) ->
```
keynslug, line 592: Hm, I thought that the argument of `format_status/1` callback is not the state itself but something that contains it among other things.
https://www.erlang.org/doc/man/gen_server#type-format_status

apps/emqx_durable_storage/src/emqx_ds_replication_layer_shard.erl
```diff
@@ -84,15 +83,14 @@ server_name(DB, Shard, Site) ->
 %%
 
 -spec servers(emqx_ds:db(), emqx_ds_replication_layer:shard_id(), Order) -> [server(), ...] when
-    Order :: leader_preferred | undefined.
-servers(DB, Shard, _Order = leader_preferred) ->
+    Order :: leader_preferred | local_preferred | undefined.
+servers(DB, Shard, leader_preferred) ->
```
keynslug, line 87: Just to be sure we're aware of that: this works only for shards that have a replica on the local node, because `ra_leaderboard` does not track leaders otherwise (i.e. when all of the servers are supposed to live on remote nodes).

ieQu1: Does it track members at least?

keynslug: Leaderboard tracks members + leaders for all shards that have local replicas. Otherwise `get_shard_servers/2` is called that simply lists currently allocated shard servers, which obviously lacks leadership information.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
ieQu1: Thanks, fixed.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_shard.erl
ieQu1: I think we need a more generic mechanism for shooting requests to the leader. It'll help to split storing batches into prepare/commit stages as well. I'll leave it as is for now, if you don't object.

apps/emqx_durable_storage/test/emqx_ds_storage_SUITE.erl
ieQu1: Well, `commit_batch` is still idempotent, since `prepare` adds generation id into the cooked batch. I think splitting the operation should ultimately help with this. There are two obstacles:

1. ra doesn't provide an API to call aux server with a timeout (can be worked around)
2. there's no API get leader from a node that is not part of the shard cluster.

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -246,7 +246,7 @@ fields(layout_builtin_reference) ->
                 reference,
                 #{
                     'readOnly' => true,
-                    importance => ?IMPORTANCE_HIDDEN
+                    importance => ?IMPORTANCE_LOW
```
ieQu1, line 249: Having reference layout in the release build is useful. However hocon complains that a non-hidden record cannot have all of its fields hidden.

apps/emqx_durable_storage/src/proto/emqx_ds_proto_v4.erl
```diff
@@ -179,8 +179,7 @@ make_delete_iterator(Node, DB, Shard, Stream, TopicFilter, StartTime) ->
     | {ok, end_of_stream}
     | {error, _}.
 delete_next(Node, DB, Shard, Iter, Selector, BatchSize) ->
-    emqx_rpc:call(
-        Shard,
+    erpc:call(
```
ieQu1, line 182: Using janky gen_rpc is only justified when forwarding large binaries, however this is not the case here.

apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl
```diff
@@ -546,13 +561,15 @@ handle_event(_ShardId, State = #s{gvars = Gvars}, Time, tick) ->
             LastWrittenTs = 0
     end,
     case Latch of
-        false when ?EPOCH(State, Time) > ?EPOCH(State, LastWrittenTs) ->
+        false when ?EPOCH(State, Time) > ?EPOCH(State, LastWrittenTs) + 1 ->
```
thalesmg, line 564: Q: why the + 1 is needed here?

thalesmg (dismissed):

apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl
ieQu1: It delays event emission by one epoch to add a little safety margin.

ieQu1: P.S. Added a comment.

apps/emqx_durable_storage/src/emqx_ds_replication_layer_shard.erl
```diff
@@ -104,17 +102,24 @@ get_servers_leader_preferred(DB, Shard) ->
...
-            pick_random(get_online_servers(DB, Shard))
+            Servers = get_online_servers(DB, Shard);
+        Servers when is_list(Servers) ->
+            ok
+    end,
+    case lists:keyfind(node(), 2, Servers) of
+        false ->
+            Servers;
+        Local when is_tuple(Local) ->
+            [Local | lists:delete(Local, Servers)]
```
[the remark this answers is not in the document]
ieQu1: Thanks, fixed.

apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl
```diff
@@ -546,13 +572,17 @@ handle_event(_ShardId, State = #s{gvars = Gvars}, Time, tick) ->
...
+            %% Note: + 1 above delays the event by one generation to
```
keynslug, line 576: Nit. Either I misunderstand the idea, or "generation" sounds slightly confusing in this context? Shouldn't it be "epoch"?

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -297,13 +301,14 @@ store_batch(Shard, Messages, Options) ->
...
 ) -> {ok, cooked_batch()} | ignore | emqx_ds:error(_).
-prepare_batch(Shard, Messages = [{Time, _Msg} | _], Options) ->
+prepare_batch(Shard, Messages = [_ | _], Options) ->
     %% NOTE
     %% We assume that batches do not span generations. Callers should enforce this.
     ?tp(emqx_ds_storage_layer_prepare_batch, #{
         shard => Shard, messages => Messages, options => Options
     }),
-    {GenId, #{module := Mod, data := GenData}} = generation_at(Shard, Time),
+    GenId = generation_current(Shard),
```
keynslug, line 310: I suspect this could introduce message duplication under _snapshot transfer + log replay_ scenarios. I.e. imagine a following sequence of entries in the logs:
1. `{T1 = 42, Msg1}`
2. `{T2 = 43, Msg2}`
3. `{T3 = 44, add_generation}`
4. `{T4 = 45, Msg3}`

If they are reapplied on top of the storage state that _already had them applied_, couldn't `Msg1` and `Msg2` end up both in previous and in the current generation?

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -444,15 +449,12 @@ update_iterator(
...
     case generation_get(Shard, GenId) of
         #{module := Mod, data := GenData} ->
-            Current = generation_current(Shard),
-            case Mod:next(Shard, GenData, GenIter0, BatchSize, Now) of
-                {ok, _GenIter, []} when GenId < Current ->
-                    %% This is a past generation. Storage layer won't write
-                    %% any more messages here. The iterator reached the end:
-                    %% the stream has been fully replayed.
-                    {ok, end_of_stream};
+            IsCurrent = GenId =:= generation_current(Shard),
```
keynslug, line 452: I wonder if this loss of write idempotency could be alleviated by picking the "current" generation according to `Now` here, instead of the most recent storage state? I have a feeling this also relaxes the requirement to read from leaders, but am not 100% sure.

The risk here I guess is having `Now` pointing to a dropped generation (from the _storage state_ perspective), but picking the earliest alive in that case should work the same I guess.

apps/emqx_durable_storage/src/emqx_ds_storage_bitfield_lts.erl
ieQu1: Right, it was a typo.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
ieQu1: Yes, it will happen. But running prepare in the main state machine code should be removed. Prepare should run on the leader (in aux callback), and add a cooked batch to the raft log. Committing cooked batch to the storage is idempotent.

thalesmg (approved):

emqxqa:
[TestExecution]()

PR #13077, opened 2024-05-20
fix: action config update would sometimes not be reflected in connector

kjellwinblad:
Before this commit the following happened sometimes:

1. action status is connected
2. action config is updated to something that should change the status to disconnected
3. action status is still connected and the old config is being used by the connector even though the config has been correctly updated.

The reason for this bug is that the post_config_hook runs before the global EMQX config is updated. The post config hook is adding the new config to the connector. Since the new config causes the action to get status disconnected, the adding of the action to the connector is retried when the health check runs but this time the config will be loaded from the global config which could cause the old config to be loaded (this happens when the global config has not had time to get updated).

The above problem has been fixed in this commit by only reading action configs from the global config when the connector starts/restarts and instead store the latest configs for the actions in the connector.

Fixes:
https://emqx.atlassian.net/browse/EMQX-12376

Release version: v/e5.7

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_manager.erl
+++ b/apps/emqx_resource/src/emqx_resource_manager.erl
@@ -1038,7 +1038,8 @@ continue_resource_health_check_not_connected(NewStatus, Data0) ->
 
 handle_manual_channel_health_check(From, #data{state = undefined}, _ChannelId) ->
     {keep_state_and_data, [
-        {reply, From, channel_status({error, resource_disconnected}, undefined)}
+        {reply, From,
+            maps:remove(config, channel_status({error, resource_disconnected}, undefined))}
     ]};
 handle_manual_channel_health_check(
     From,
@@ -1072,13 +1073,15 @@ handle_manual_channel_health_check(
     is_map_key(ChannelId, Channels)
 ->
     %% No ongoing health check: reply with current status.
-    {keep_state_and_data, [{reply, From, maps:get(ChannelId, Channels)}]};
+    {keep_state_and_data, [{reply, From, maps:remove(config, maps:get(ChannelId, Channels))}]};
 handle_manual_channel_health_check(
     From,
     _Data,
     _ChannelId
 ) ->
-    {keep_state_and_data, [{reply, From, channel_status({error, channel_not_found}, undefined)}]}.
+    {keep_state_and_data, [
+        {reply, From, maps:remove(config, channel_status({error, channel_not_found}, undefined))}
+    ]}.
 
 -spec channels_health_check(resource_status(), data()) -> data().
 channels_health_check(?status_connected = _ConnectorStatus, Data0) ->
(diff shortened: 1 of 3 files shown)


apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -1479,62 +1512,61 @@ channel_status() ->
...
+        error => Error,
+        config => ChannelConfig
     };
-channel_status({error, Reason}) ->
+channel_status({error, Reason}, ChannelConfig) ->
     #{
         status => ?status_disconnected,
-        error => Reason
+        error => Reason,
+        config => ChannelConfig
```
thalesmg, line 1569: I believe we return this channel status in the APIs.  If so, we might want to redact this config in the API handlers, as it might contain sensitive data.  It might be already handled there, but worth confirming it.

thalesmg (dismissed):

apps/emqx_resource/src/emqx_resource_manager.erl
kjellwinblad: Good point. I will check.

kjellwinblad: Configs are now removed from get_channel_status calls and the lookup calls: [0f10b0b](https://github.com/emqx/emqx/pull/13077/commits/0f10b0b6c33bdf3e8113d2952f23b5550ad9a60e)

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -1426,6 +1490,14 @@ maybe_reply(Actions, From, Reply) ->
...
 -spec data_record_to_external_map(data()) -> resource_data().
 data_record_to_external_map(Data) ->
+    AddedChannelsList = maps:to_list(Data#data.added_channels),
+    AddedChannelsListWithoutConfigs =
+        [
+            {ChanID, maps:remove(config, Status)}
+         || {ChanID, Status} <- AddedChannelsList
+        ],
+    AddedChannelsWithoutConfigs =
+        maps:from_list(AddedChannelsListWithoutConfigs),
```
thalesmg, line 1500: Nit:

```suggestion
    AddedChannelsWithoutConfigs =
        maps:map(
            fun(_ChanID, Status) -> maps:remove(config, Status) end,
            Data#data.added_channels
        ),
```

thalesmg (dismissed):

apps/emqx_resource/src/emqx_resource_manager.erl
kjellwinblad: Fixed [88c96e2](https://github.com/emqx/emqx/pull/13077/commits/88c96e26de41199f4fb66f69be2960a04511098f)

Also added another fix to not leak configs in alarm.

thalesmg (approved):

PR #13078, opened 2024-05-21
feat: return 415 when UNSUPPORTED_MEDIA_TYPE

zhongwencool:
Fixes [12401](https://emqx.atlassian.net/browse/EMQX-12401)

Release version: v/e5.8.0

## PR Checklist

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
@@ -30,7 +30,7 @@
 -export([base_path/0]).
 -export([relative_uri/1, get_relative_uri/1]).
 -export([compose_filters/2]).
--export([is_content_type_json/2, validate_content_type/3]).
+-export([validate_content_type_json/2, validate_content_type/3]).
 
 -export([
     filter_check_request/2,
@@ -153,18 +153,22 @@ spec(Module, Options) ->
         ),
     {ApiSpec, components(lists:usort(AllRefs), Options)}.
 
-is_content_type_json(Params, Meta) ->
+validate_content_type_json(Params, Meta) ->
     validate_content_type(Params, Meta, <<"application/json">>).
 
 %% tip: Skip content-type check if body is empty.
-validate_content_type(#{body := Body} = Params, #{method := Method}, Expect) when
+validate_content_type(
+    #{body := Body, headers := Headers} = Params,
+    #{method := Method},
+    Expect
+) when
     (Method =:= put orelse
         Method =:= post orelse
         method =:= patch) andalso
         Body =/= #{}
 ->
     ExpectSize = byte_size(Expect),
-    case find_content_type(Params) of
+    case maps:get(<<"content-type">>, Headers, undefined) of
         <<Expect:ExpectSize/binary, _/binary>> ->
             {ok, Params};
         _ ->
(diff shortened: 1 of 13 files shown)


apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
```diff
@@ -152,6 +153,30 @@ spec(Module, Options) ->
...
+            {ok, Params};
+        _ ->
+            {415, 'UNSUPPORTED_MEDIA_TYPE', <<"content-type:", Expect/binary, " Required">>}
+    end;
+validate_content_type(Params, _Meta, _Expect) ->
+    {ok, Params}.
+
+find_content_type(#{headers := #{<<"content-type">> := Type}}) -> Type;
+find_content_type(#{headers := #{<<"Content-Type">> := Type}}) -> Type;
+find_content_type(_Headers) -> error.
```
ieQu1, line 178: HTTP RFC declares that header names are content-insensitive, this check should take it into account.

Perhaps, it can handle happy cases as implemented below, and in this clause run a more comprehensive check for pathological cases like `CoNtEnT-TyPe`.

> HTTP headers let the client and the server pass additional information with an HTTP request or response. An HTTP header consists of its case-insensitive name followed by a colon (:), then by its value. [Whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) before the value is ignored.

apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
```diff
@@ -30,6 +30,7 @@
 -export([base_path/0]).
 -export([relative_uri/1, get_relative_uri/1]).
 -export([compose_filters/2]).
+-export([is_content_type_json/2, validate_content_type/3]).
```
zmstone, line 33:
```suggestion
-export([validate_content_type_json/2, validate_content_type/3]).
```

apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
zmstone: I think cowboy normalizes header names to lower-case:
https://github.com/emqx/cowboy/blob/957dc318057b1adf1c57c021cd097d2820a13880/src/cowboy_http.erl#L605

ieQu1: Makes sense. Then second clause is not needed.

apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
```diff
@@ -152,6 +153,30 @@ spec(Module, Options) ->
...
+    Expect
+) when
+    (Method =:= put orelse
+        Method =:= post orelse
+        method =:= patch) andalso
+        Body =/= #{}
+->
+    ExpectSize = byte_size(Expect),
+    case maps:get(<<"content-type">>, Headers, undefined) of
+        <<Expect:ExpectSize/binary, _/binary>> ->
```
zmstone, line 172: Q: why not a exact match?
```
case maps:get(<<"content-type">>, Headers, undefined) of
    Expect ->
```

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2529)

PR #13100, opened 2024-05-23
fix: add action rendered trace for s3 in aggregated mode

kjellwinblad:
Fixes:
https://emqx.atlassian.net/browse/EMQX-12429

Manually tested

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_s3/src/emqx_bridge_s3_connector.erl
+++ b/apps/emqx_bridge_s3/src/emqx_bridge_s3_connector.erl
@@ -366,10 +366,7 @@ run_aggregated_upload(InstId, ChannelID, Records, #{aggreg_id := AggregId}) ->
     Timestamp = erlang:system_time(second),
     emqx_trace:rendered_action_template(ChannelID, #{
         mode => aggregated,
-        records => #emqx_trace_format_func_data{
-            function = fun render_records/1,
-            data = Records
-        }
+        records => Records
     }),
     case emqx_connector_aggregator:push_records(AggregId, Timestamp, Records) of
         ok ->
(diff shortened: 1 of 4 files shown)


apps/emqx_bridge_s3/src/emqx_bridge_s3_connector.erl
```diff
@@ -372,6 +379,13 @@ run_aggregated_upload(InstId, Records, #{aggreg_id := AggregId}) ->
             {error, {unrecoverable_error, Reason}}
     end.
 
+render_records(Records) ->
+    try
+        [unicode:characters_to_binary(R) || R <- Records]
```
keynslug, line 384: Q: `Records` are maps 99.9% of time, productions of a Rule SQL expression, to be more specific. I wonder why do we attempt to stringify them like that first?

kjellwinblad: You are right. This was done for non-aggregate mode so I assumed we should do the same for aggregate mode as well but I see now that this is incorrect. In non-aggregate mode we render a template. I will fix. Btw, I accidentally pushed a commit for another issue to this PR. Do you think you can review that one as well so we can merge both with the same PR?

kjellwinblad: Removed the formatting here: [89b47e8](https://github.com/emqx/emqx/pull/13100/commits/89b47e8ffccbeff5b9149f3fc2ab9866d4e611fa)

keynslug (approved):
Seems reasonable! Not too confident about fb7688a due to lack of context.

emqxqa:
[TestExecution]()

PR #13130, opened 2024-05-27
fix(redis trace): add separators in redis batch action trace

kjellwinblad:
Logger will transform data that looks like IO data into a string which made redis batch traces look like the spaces had been removed from the strings. To prevent this, we render the batched commands into binary string separated by spaces and semicolon. The components of a single command is separated by spaces and the commands in a batch are separated by semicolons.

Fixes:
https://emqx.atlassian.net/browse/EMQX-12428

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_redis/src/emqx_bridge_redis_connector.erl
+++ b/apps/emqx_bridge_redis/src/emqx_bridge_redis_connector.erl
@@ -6,6 +6,7 @@
 -include_lib("emqx/include/logger.hrl").
 -include_lib("emqx_resource/include/emqx_resource.hrl").
 -include_lib("snabbkaffe/include/snabbkaffe.hrl").
+-include_lib("emqx/include/emqx_trace.hrl").
 
 -behaviour(emqx_resource).
 
@@ -143,7 +144,13 @@ on_batch_query(
             [{ChannelID, _} | _] = BatchData,
             emqx_trace:rendered_action_template(
                 ChannelID,
-                #{commands => Cmds, batch => ture}
+                #{
+                    commands => #emqx_trace_format_func_data{
+                        function = fun trace_format_commands/1,
+                        data = Cmds
+                    },
+                    batch => true
+                }
             ),
             Result = query(InstId, {cmds, Cmds}, RedisConnSt),
             ?tp(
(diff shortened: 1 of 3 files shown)


qzhuyan (approved):

thalesmg (dismissed):

apps/emqx_bridge_redis/src/emqx_bridge_redis_connector.erl
```diff
@@ -162,6 +169,11 @@ on_batch_query(
             Error
     end.
 
+trace_format_commands(Commands0) ->
+    Commands1 = [[unicode:characters_to_list(S) || S <- C] || C <- Commands0],
+    Commands2 = [lists:join(" ", C) || C <- Commands1],
+    unicode:characters_to_binary(lists:join("; ", Commands2)).
```
zmstone, line 175: `unicode:characthers_to_binary` works with `iodata()` type, so there is no need to convert them into a string list then join with space then convert back to binary.

kjellwinblad: Sorry, yes, you are right. The conversion on line 173 was done as I first used string:join which only worked with list strings but then I figured out that I could just as well use lists:join but forgot to remove the conversion on the previous line. While looking at another issue I found that it is actually best_effort_json that cause the problem we are trying to solve here. best_effort_json has heuristics that tries to convert things that looks like IO data into into strings. But this heuristics in at least two places (this connector and the oracle connector). I think perhaps a better way to solve it is to wrap the list in something that forces best_effort_json to treat it as a list but then we have to handle this wrapping in all formatters...

zmstone: if you mean this macro?
`-define(IS_STRING(String), (is_list(String) orelse is_binary(String))).`

yes, the macro can be optimized to check if the list style string also meets this condition: `String =/=[] andalso is_integer(hd(String))`

However, I actually mean this:
```
    Commands1 = [lists:join(" ", C) || C <- Commands0],
    unicode:characters_to_binary(lists:join("; ", Commands1)).
```

savonarola: Should we use unicode tranformations at all to dump commands? 🤔 
Redis is not unicode aware, it operates with byte sequences as commands and operands.
Doesn't converting to unicode bring umbiguity of what actually was sent to redis 🤔

zmstone: redis is not unicode aware, then it's EMQX's responsibility to make sure unicode code-points are turned into byte streams (utf8), so I think it does make sense to format logs as utf8 chars

our config is read as utf8, and the templates are rendered utf8

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2304)

PR #13133, opened 2024-05-27
chore: port diverged modules back to oss

zmstone:
A few modules diverged more than necessary.
This is an attempt to port them back to upstream.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2299)

PR #13140, opened 2024-05-28
fix: trace formatter should not crash when given a non-binary payload

kjellwinblad:
Fixes:
https://emqx.atlassian.net/browse/EMQX-12474

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_trace/emqx_trace_formatter.erl
+++ b/apps/emqx/src/emqx_trace/emqx_trace_formatter.erl
@@ -89,8 +89,17 @@ weight({packet, _}) -> {0, packet};
 weight({payload, _}) -> {2, payload};
 weight({K, _}) -> {1, K}.
 
-format_packet(undefined, _) -> "";
-format_packet(Packet, Encode) -> emqx_packet:format(Packet, Encode).
+format_packet(undefined, _) ->
+    "";
+format_packet(Packet, Encode) ->
+    try
+        emqx_packet:format(Packet, Encode)
+    catch
+        _:_ ->
+            %% We don't want to crash if there is a field named packet with
+            %% some other type of value
+            Packet
+    end.
 
 format_payload(undefined, _) ->
     "";
@@ -100,7 +109,11 @@ format_payload(Payload, text) when ?MAX_PAYLOAD_FORMAT_LIMIT(Payload) ->
     unicode:characters_to_list(Payload);
 format_payload(Payload, hex) when ?MAX_PAYLOAD_FORMAT_LIMIT(Payload) -> binary:encode_hex(Payload);
 format_payload(<<Part:?TRUNCATED_PAYLOAD_SIZE/binary, _/binary>> = Payload, Type) ->
-    emqx_packet:format_truncated_payload(Part, byte_size(Payload), Type).
+    emqx_packet:format_truncated_payload(Part, byte_size(Payload), Type);
+format_payload(Payload, _) ->
+    %% We don't want to crash if there is a field named payload with some other
+    %% type of value
+    Payload.
 
 to_iolist(Atom) when is_atom(Atom) -> atom_to_list(Atom);
 to_iolist(Int) when is_integer(Int) -> integer_to_list(Int);
--- a/changes/ce/fix-13140.en.md
+++ b/changes/ce/fix-13140.en.md
@@ -1,0 +1,1 @@
+The issue causing text traces for the republish action to crash and not display correctly has been resolved.


apps/emqx/src/emqx_trace/emqx_trace_formatter.erl
```diff
@@ -89,8 +89,17 @@ weight({packet, _}) -> {0, packet};
...
-format_packet(undefined, _) -> "";
-format_packet(Packet, Encode) -> emqx_packet:format(Packet, Encode).
+format_packet(undefined, _) ->
+    "";
+format_packet(Packet, Encode) ->
+    try
+        emqx_packet:format(Packet, Encode)
+    catch
+        _:_ ->
+            %% We don't want to crash if there is a field named packet with
```
zmstone, line 99: could you provide an example here?

apps/emqx/src/emqx_trace/emqx_trace_formatter.erl
```diff
@@ -100,7 +109,11 @@ format_payload(Payload, text) when ?MAX_PAYLOAD_FORMAT_LIMIT(Payload) ->
     unicode:characters_to_list(Payload);
 format_payload(Payload, hex) when ?MAX_PAYLOAD_FORMAT_LIMIT(Payload) -> binary:encode_hex(Payload);
 format_payload(<<Part:?TRUNCATED_PAYLOAD_SIZE/binary, _/binary>> = Payload, Type) ->
-    emqx_packet:format_truncated_payload(Part, byte_size(Payload), Type).
+    emqx_packet:format_truncated_payload(Part, byte_size(Payload), Type);
+format_payload(Payload, _) ->
```
zmstone, line 113: if payload is a json-map, maybe it makes sense to format?

apps/emqx/src/emqx_trace/emqx_trace_formatter.erl
kjellwinblad: There is a trace that includes the output of a rule. As the the rule can output almost any map it can contain a field named packet. Not 100% sure if the key can be an atom in that case. Perhaps the optimal solution is to not look for the packet field in all levels of the map as we may only want to format it when it is in the top level. What do you think? The payload field is perhaps different since the user can change how it is formatted and will perhaps expect it to be formatted this way wherever it may be.

apps/emqx/src/emqx_trace/emqx_trace_formatter.erl
kjellwinblad: Not 100% sure about this but I think it makes sense to keep it as a map in this case because I think this is what will be most useful for users.

apps/emqx/src/emqx_trace/emqx_trace_formatter.erl
zmstone: this is ok.
I missed that this is done recursively.
maybe good to explain as comment.

apps/emqx/src/emqx_trace/emqx_trace_formatter.erl
zmstone: right, since it's returning Payload as is, it should be included in the output.

apps/emqx/src/emqx_trace/emqx_trace_formatter.erl
```diff
@@ -100,7 +109,11 @@ format_payload(Payload, text) when ?MAX_PAYLOAD_FORMAT_LIMIT(Payload) ->
...
+    %% We don't want to crash if there is a field named payload with some other
+    %% type of value
+    Payload.
```
zmstone, line 116: If JSON map is the only possible format, maybe we should format it into JSON and truncate.
Otherwise it can be quite bloated when tracing a large SQL output.

kjellwinblad: One will probably use some kind JSON viewer when viewing the JSON traces (for example, the one provided by the dashboard rule tester). This will allow the user tor expand to collapse fields as needed which I think remove the bloat problem that you mention above.

I think the tracing feature is mostly intended to inspect the flow triggered by a few messages or rule applications and not to inspect massive amount of data which makes the bloat problem less relevant. As user using the rule tester in the dashboard I would like to see the whole rule output in a structured way and getting a truncated JSON string (due to some hard coded limit) could probably be quite annoying. So I don't think we should encode JSON maps as text and then truncate them.

zmstone: fair enough.

zmstone (approved):

thalesmg (approved):
:heavy_check_mark: 

Would it make sense to add a smoke test for this case?

kjellwinblad:
@thalesmg 
> Would it make sense to add a smoke test for this case?

It would not hurt. I added a task for this so I can pick it up later:

https://emqx.atlassian.net/browse/EMQX-12483

@zmstone 
> maybe good to explain as comment.

Also does not hurt. Will try to add this at the same time as I do the task above.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2302)

PR #13163, opened 2024-05-30
Inject shared subcription handling into durable session

savonarola:
Fixes [EMQX-12305](https://emqx.atlassian.net/browse/EMQX-12305).

[EMQX-12305]: https://emqx.atlassian.net/browse/EMQX-12305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/emqx/include/emqx_durable_session_metadata.hrl
```diff
@@ -31,5 +31,7 @@
 -define(clientinfo, clientinfo).
 -define(protocol, protocol).
 -define(offline_info, offline_info).
+-define(owner_session, owner_session).
```
ieQu1, line 34: That's not session metadata.

apps/emqx/include/emqx_persistent_session_internals.hrl
```diff
@@ -68,7 +68,29 @@
     %% marked for deletion:
     unsubscribed = false :: boolean(),
     %% Reference to the subscription state:
-    sub_state_id :: emqx_persistent_session_ds_subs:subscription_state_id()
+    sub_state_id :: emqx_persistent_session_ds_subs:subscription_state_id(),
+    stream_owner :: emqx_persistent_session_ds:stream_owner()
```
ieQu1, line 72: 4> size(term_to_binary(owner_shared_sub)).
19
5> size(term_to_binary(owner_session)).
16
6> size(term_to_binary(0)).
3
7> size(term_to_binary(1)).
3

apps/emqx/include/emqx_persistent_session_internals.hrl
ieQu1, line 1: Why rename the file? It's already in `emqx_persistent_session_ds` directory, this should imply that it's unrelated to memory session. Maybe rename it to `internals.hrl` instead.

apps/emqx/include/emqx_persistent_session_internals.hrl
ieQu1: ...We store this things durably and transfer over RPC. Let's not do premature pessimization.

apps/emqx/include/emqx_persistent_session_internals.hrl
```diff
@@ -68,7 +68,29 @@
     %% marked for deletion:
     unsubscribed = false :: boolean(),
     %% Reference to the subscription state:
-    sub_state_id :: emqx_persistent_session_ds_subs:subscription_state_id()
+    sub_state_id :: emqx_persistent_session_ds_subs:subscription_state_id(),
```
ieQu1, line 71: We cannot change this record. It already could be stored somewhere.

Can we just store different types of SRS in different containers and track ownership in the code?

apps/emqx/include/emqx_persistent_session_internals.hrl
savonarola: It is moved to include, because the new app will need to work with `#srs{}`

apps/emqx/include/emqx_persistent_session_internals.hrl
savonarola: We probably can, but we cannot treat records polymorphically. All code working with srs should be rewritten to use some polymorphic accessors 🤔

savonarola: We may try to store ownership separately in the session state.

savonarola: Probably we may completely get rid of this

apps/emqx/include/emqx_persistent_session_internals.hrl
ieQu1: Hmm, I smell a logical problem here. We missed something part during the internal discussions. Shared subscription shouldn't care about `#srs`, because this record contains seqnos. Seqnos are *private to the session* (they are used for PacketID generation), and as such should not leave it. 
Session and shared subscription worker should exchange a different type of data.

apps/emqx/include/emqx_persistent_session_internals.hrl
ieQu1: In lieu of this remark https://github.com/emqx/emqx/pull/13163#discussion_r1621285076 this should not be needed indeed.

apps/emqx/include/emqx_persistent_session_internals.hrl
savonarola: >Session and shared subscription worker should exchange a different type of data.
Yes, the group leader obviously should not receive this. 

But the agent part (that lives inside the session) still needs them to do something like this: https://github.com/emqx/emqx/pull/13163/files#diff-6091e87982530a697e6e3e1e7ca69c55c8410cc7892aefab6fd66f0b39c8e292R293

ieQu1: Can this part of agent be moved to emqx app? If it needs to inspect internal session data structures, it is logically part of the session.  

Internal data structures were made hidden for a reason.

savonarola: In other words, inside `emqx_persistent_session_ds_shared_subs` there should be some kind of stream scheduler, but with another behaviour (like, not respecting ranks).

ieQu1: Can the server always overwrite ranks to `{0, 0}` when lending streams to the session?

savonarola: We may but we needn't because session does not know about the ranks, it is the scheduler who knows.

ieQu1: ...Perhaps it will allow to reuse more of the scheduler logic in the session?

savonarola: I simplified the PR, it still makes sense because it defined boundary between session and shared subscription agent (`emqx_persistent_session_ds_shared_subs`).

If we do not want to share `#srs{}` with other apps, then we will need another boundary inside `emqx_persistent_session_ds_shared_subs`, because there is a complex state machine interacting with the server, it should be the part of DSL app.

ieQu1: > If we do not want to share #srs{} with other apps

Absolutely not.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -601,6 +642,45 @@ handle_timeout(_ClientInfo, Timeout, Session) ->
     ?SLOG(warning, #{msg => "unknown_ds_timeout", timeout => Timeout}),
     {ok, [], Session}.
 
+%%--------------------------------------------------------------------
+%% Geneic messages
```
ieQu1, line 646:
```suggestion
%% Generic messages
```

ieQu1 (dismissed):

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3363)

PR #13180, opened 2024-06-04
0604 optimize tcp ssl connection data send performance

zmstone:
Fixes [EMQX-12515](https://emqx.atlassian.net/browse/EMQX-12515)

Release version: v/e5.7.1

### Perf1: do not call inet getstat before each and every send

In a stress test environment, when alarm is enabled, there were
a lot of long_schedule warnings with the stacktrace pointing to
congestion alarm based inet:getstat function.
When alarm is disabled, the one single client is able to
hold 40,000 QoS 1 messages per second (inflight=32) throughput.
When alarm is enabled, its mqueue overflows.

### Perf2: eliminate the loop-back fake async inet_reply message

Now that EMQX is on OTP 26 by default, the fake inet_reply
message when the send call returns ok only adds overhead.

For TCP (gen_tcp module), OTP 26 no longer supports direct port_command
so we had to swith to gen_tcp:send, then fake an async reply:
see https://github.com/emqx/esockd/pull/181

For TLS (ssl module), it has never supported async send anyway.

This commit switches to call Transport:send/2 directly,
this should result in one less loop-back message for each message.

For OTP 25 on which EMQX can still be compiled, one will have to
suffer the performance penalty for gen_tcp:send function to
do a non-optimized 'receive' for the send result.

## PR Checklist

[EMQX-12515]: https://emqx.atlassian.net/browse/EMQX-12515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/ce/feat-13180.en.md
+++ b/changes/ce/feat-13180.en.md
@@ -1,0 +1,1 @@
+Improve client message handling performance when running on OTP 26.
--- a/apps/emqx/src/emqx_connection.erl
+++ b/apps/emqx/src/emqx_connection.erl
@@ -914,7 +914,6 @@ send(IoData, #state{transport = Transport, socket = Socket, channel = Channel}) 
     Oct = iolist_size(IoData),
     ok = emqx_metrics:inc('bytes.sent', Oct),
     inc_counter(outgoing_bytes, Oct),
-    emqx_congestion:maybe_alarm_conn_congestion(Socket, Transport, Channel),
     case Transport:async_send(Socket, IoData, []) of
         ok ->
             ok;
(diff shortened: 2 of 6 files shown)


terry-xiaoyu (approved):

thalesmg (approved):

zhongwencool (approved):

thalesmg (approved):

changes/ce/feat-13180.en.md
```diff
@@ -0,0 +1 @@
+Improve client message handling performance when running on OTP 26.
```
qzhuyan, line 1: So we removed the function of conn congestion alarm?

zmstone: it's still called, but only triggered by the stats emit timer.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2593)

PR #13182, opened 2024-06-04
feat(oracle connector): add possible to configure and use SSL

kjellwinblad:
This is not tested since I have not managed to set up SSL in the oracle docker XE image. It should be tested at least manually. Perhpas QA could help with that?

Fixes:
https://emqx.atlassian.net/browse/EMQX-12288

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_oracle/src/emqx_oracle.erl
+++ b/apps/emqx_oracle/src/emqx_oracle.erl
@@ -112,7 +112,7 @@ on_start(
         case emqx_utils_maps:deep_get([ssl, enable], Config, false) of
             true ->
                 SSLOptsMap = maps:get(ssl, Config),
-                SSLOpts = maps:to_list(SSLOptsMap),
+                SSLOpts = emqx_tls_lib:to_client_opts(SSLOptsMap),
                 [{ssl, SSLOpts} | Options0];
             false ->
                 Options0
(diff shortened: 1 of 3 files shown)


apps/emqx_oracle/src/emqx_oracle.erl
```diff
@@ -108,12 +108,21 @@ on_start(
         {timeout, ?OPT_TIMEOUT},
         {app_name, "EMQX Data To Oracle Database Action"}
     ],
+    Options1 =
+        case emqx_utils_maps:deep_get([ssl, enable], Config, false) of
+            true ->
+                SSLOptsMap = maps:get(ssl, Config),
+                SSLOpts = maps:to_list(SSLOptsMap),
+                [{ssl, SSLOpts} | Options0];
```
thalesmg, line 116: Nit: could `emqx_tls_lib:to_client_opts` be of use here?

kjellwinblad:
The driver does not seem to work without a patch of OTP. There is patches for OTP 25 but not for OTP 26: https://github.com/erlangbureau/jamdb_oracle/issues/98#issuecomment-2150320459

kjellwinblad:
I have tried the following patch

https://github.com/erlangbureau/jamdb_oracle/blob/master/test/ssl-10.8-otp-25.patch

On OTP versions:

OTP-25.2.3
OTP-25.3.2.12

These don't work. `ssl:connect` get stuck after the patch has been applied. The following command gets stuck:

```
ssl:connect("oracle-test-instance.cerkngbiw6l9.eu-west-1.rds.amazonaws.com", 2484, [{verify, verify_none}]).
```

id:
We decided to reject this one.

PR #13197, opened 2024-06-06
fix(bridge-s3): pass SSL options through `convert_certs/2`

keynslug:
Fixes [EMQX-12522](https://emqx.atlassian.net/browse/EMQX-12522).

Release version: e5.7.

## Summary

Regular mechanism has not been working because S3 Bridge schema structured differently.

## PR Checklist

[EMQX-12522]: https://emqx.atlassian.net/browse/EMQX-12522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_connector/src/emqx_connector.erl
+++ b/apps/emqx_connector/src/emqx_connector.erl
@@ -126,17 +126,26 @@ pre_config_update([?ROOT_KEY, _Type, _Name], Oper, OldConfig) when
 ->
     %% to save the 'enable' to the config files
     {ok, OldConfig#{<<"enable">> => operation_to_enable(Oper)}};
-pre_config_update([?ROOT_KEY, _Type, Name] = Path, Conf = #{}, _OldConfig) ->
+pre_config_update([?ROOT_KEY, _Type, Name] = Path, Conf = #{}, ConfOld) ->
     case validate_connector_name(Name) of
         ok ->
             case emqx_connector_ssl:convert_certs(filename:join(Path), Conf) of
-                {error, Reason} ->
-                    {error, Reason};
                 {ok, ConfNew} ->
-                    {ok, ConfNew}
+                    connector_pre_config_update(Path, ConfNew, ConfOld);
+                {error, Reason} ->
+                    {error, Reason}
             end;
         Error ->
             Error
+    end.
+
+connector_pre_config_update([?ROOT_KEY, Type, Name] = Path, ConfNew, ConfOld) ->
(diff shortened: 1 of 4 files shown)


apps/emqx_connector/src/emqx_connector.erl
```diff
@@ -126,19 +126,28 @@ pre_config_update([?ROOT_KEY, _Type, _Name], Oper, OldConfig) when
...
+                {error, Reason} ->
+                    {error, Reason}
             end;
         Error ->
             Error
     end.
 
+connector_pre_config_update([?ROOT_KEY, Type, Name] = Path, ConfNew, ConfOld) ->
+    Mod = emqx_connector_info:config_transform_module(Type),
+    case Mod =/= undefined andalso erlang:function_exported(Mod, pre_config_update, 4) of
```
thalesmg, line 144: I wonder if we could use the `propagated_pre_config_update/3` callback instead of this.  Not entirely sure how to use it, might need to register `emqx_bridge_s3` as a handler for this path.  :thinking:

keynslug: Interesting idea.

> might need to register `emqx_bridge_s3` as a handler for this path.

Yeah I guess so. This is actually what feels off a bit: need to register config handlers, and thus need to have the knowledge about config roots and config path structure on this level as a result. I'll try to tackle and see how it looks anyway.

keynslug: Another problem is that having both `[connectors, '?', '?']` and `[connectors, s3, '?']` config handlers is impossible currently, which is actually reasonable (because defining ordering and semantics of multiple `pre_config_update` / `post_config_update`s is tricky and would make this mechanism much harder to reason about), while setting config handler on `[connectors, s3, '?', transport_options]` feels super weird.

keynslug: It could probably help to tear the handler behaviour apart, because `pre` and `post` are _very_ different things IMO. Or defer this "messing with certfiles" to something like a converter that's aware both of the config tree and the HOCON schema, and make sure that callers are aware that these converters may have side-effects.

But the amount of work both of those would ensue feels a bit too much for the scope of this PR to be honest. 🫠

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2446)

PR #13209, opened 2024-06-07
chore: remove redundant health check call

thalesmg:
Since `emqx_resource_manager:channel_health_check` already does a base resource health check, the call was repeated.

Fixes [EMQX-12521](https://emqx.atlassian.net/browse/EMQX-12521)

Release version: v/e5.7.1

## PR Checklist

[EMQX-12521]: https://emqx.atlassian.net/browse/EMQX-12521?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_manager.erl
+++ b/apps/emqx_resource/src/emqx_resource_manager.erl
@@ -381,7 +381,6 @@ channel_health_check(ResId, ChannelId) ->
 add_channel(ResId, ChannelId, Config) ->
     Result = safe_call(ResId, {add_channel, ChannelId, Config}, ?T_OPERATION),
     %% Wait for health_check to finish
-    _ = health_check(ResId),
     _ = channel_health_check(ResId, ChannelId),
     Result.
 


zmstone:
according to thedescription in jira. Hc is done 3 times. This eliminates one, but there is another one left

thalesmg:
> according to thedescription in jira. Hc is done 3 times. This eliminates one, but there is another one left

I think this change can be done safely, regardless of the ticket, so can be merged as is.

Also, as mentioned in the ticket, maybe it's not so bad to do 2 in a row if it's only for probing the config?  I'm not sure we can remove either of the remaining ones, due to different code paths reaching there.  @kjellwinblad , do you have any insights about that?

zmstone (approved):

zmstone:
> > according to thedescription in jira. Hc is done 3 times. This eliminates one, but there is another one left
> 
> I think this change can be done safely, regardless of the ticket, so can be merged as is.
> 
> Also, as mentioned in the ticket, maybe it's not so bad to do 2 in a row if it's only for probing the config? I'm not sure we can remove either of the remaining ones, due to different code paths reaching there. @kjellwinblad , do you have any insights about that?

I agree to merge this one as it is.
However I think the third HC can be easily avoided too:
`emqx_resource_manager:add_channel/3` can be extended with a `/4` variation which does not immediately return `ok`, but for HC result, and call `/4` for `dryrun` resource. --- this is essentially what `emqx_bridge_v2:create_dry_run_helper` does anyway: add (which triggers an HC), then HC (and wait for result)

thalesmg:
> I agree to merge this one as it is. However I think the third HC can be easily avoided too: `emqx_resource_manager:add_channel/3` can be extended with a `/4` variation which does not immediately return `ok`, but wait for HC result, and call `/4` for `dryrun` resource. --- this is essentially what `emqx_bridge_v2:create_dry_run_helper` does anyway: add (which triggers an HC), then HC (and wait for result)

:heavy_check_mark: 
https://github.com/emqx/emqx/pull/13220

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2595)

PR #13218, opened 2024-06-10
feat(dsrepl): enable WAL-less batch writes

keynslug:
Fixes [EMQX-12468](https://emqx.atlassian.net/browse/EMQX-12468).

Release version: v5.8

Depends on https://github.com/emqx/erlang-rocksdb/pull/33.

## Summary

Use WAL-less write mode in `emqx_ds_storage_layer` backends, which should be safe because `ra` keeps the Raft log anyway, assuming `ra` has good correctness / durability guarantees in regards to its own WAL and log segment writes.

Safety is achieved through following mechanisms.
1. The implementation of `ra_machine` from time to time asks Ra to release _all_ log entries up to the current Ra index.
2. According to internal logic, Ra often satisfies that request by taking a snapshot at the current Ra index.
3. Taking the snapshot involves flushing _all yet unflushed_ column families to disk at the storage layer.
4. Once this is done, and snapshot is successfully taken, Ra marks the logs releasable and collects them eventually.

Open questions:
1. Do we want to be able to toggle WAL-less mode as a whole or on the storage layer level, through a set of options?
2. Do we need to introduce a safety margin, by requesting release not up to the current but to some past Raft index? So that there's some extra logs to replay during snapshot recovery. This will involve a bit more messy state tracking to make it work correctly.
3. How to decide when it's time to request releasing logs? Ideally this should happen as often as is optimal for RocksDB, e.g. by default RocksDB flushes memtables once it hits 64 MiB. For that we need to have some sort of batch accounting, but this sounds suboptimal, both in the context of Ra server process (because this will block it) and in the context of egress process (because these numbers will become part of the Raft log).
4. Correspondingly, should automatic flushing be disabled?

## PR Checklist

[EMQX-12468]: https://emqx.atlassian.net/browse/EMQX-12468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
@@ -317,9 +317,6 @@ drop_shard(Shard) ->
 ) ->
     emqx_ds:store_batch_result().
 store_batch(Shard, Messages, Options) ->
-    ?tp(emqx_ds_storage_layer_store_batch, #{
-        shard => Shard, messages => Messages, options => Options
-    }),
     case prepare_batch(Shard, Messages, #{}) of
         {ok, CookedBatch} ->
             commit_batch(Shard, CookedBatch, Options);
@@ -334,9 +331,12 @@ store_batch(Shard, Messages, Options) ->
     [{emqx_ds:time(), emqx_types:message()}],
     batch_prepare_opts()
 ) -> {ok, cooked_batch()} | ignore | emqx_ds:error(_).
-prepare_batch(Shard, Messages = [{Time, _} | _], _Options) ->
+prepare_batch(Shard, Messages = [{Time, _} | _], Options) ->
     %% NOTE
     %% We assume that batches do not span generations. Callers should enforce this.
+    ?tp(emqx_ds_storage_layer_prepare_batch, #{
+        shard => Shard, messages => Messages, options => Options
+    }),
     %% FIXME: always store messages in the current generation
     case generation_at(Shard, Time) of
         {GenId, #{module := Mod, data := GenData}} ->
@@ -344,9 +344,6 @@ prepare_batch(Shard, Messages = [{Time, _} | _], _Options) ->
             Result =
                 case Mod:prepare_batch(Shard, GenData, Messages) of
                     {ok, CookedBatch} ->
-                        ?tp(emqx_ds_storage_layer_batch_cooked, #{
-                            shard => Shard, gen => GenId, batch => CookedBatch
-                        }),
                         {ok, #{?tag => ?COOKED_BATCH, ?generation => GenId, ?enc => CookedBatch}};
                     Error = {error, _, _} ->
                         Error
--- a/apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
@@ -71,8 +71,7 @@
     options/0,
     prototype/0,
     cooked_batch/0,
-    batch_store_opts/0,
-    db_write_opts/0
+    batch_store_opts/0
 ]).
 
 -include_lib("snabbkaffe/include/snabbkaffe.hrl").
@@ -223,9 +222,6 @@
 %% Generation callbacks
 %%================================================================================
 
-%% See: `rocksdb:write_options()'.
--type db_write_opts() :: [_Option].
-
 %% Create the new schema given generation id and the options.
 %% Create rocksdb column families.
 -callback create(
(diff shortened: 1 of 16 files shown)


apps/emqx_durable_storage/src/emqx_ds_storage_reference.erl
```diff
@@ -115,7 +115,7 @@ commit_batch(_ShardId, #s{db = DB, cf = CF}, Messages) ->
         end,
         Messages
     ),
-    Res = rocksdb:write_batch(DB, Batch, _WriteOptions = []),
```
ieQu1, line 118: This should be passed as a parameter to the storage layer. Replication-less edition will use rocksdb WAL.

apps/emqx_utils/src/emqx_utils_stream.erl
```diff
@@ -190,7 +190,7 @@ transpose_tail(S, Tail) ->
 %% @doc Make a stream by concatenating multiple streams.
 -spec chain([stream(X)]) -> stream(X).
 chain(L) ->
-    lists:foldl(fun chain/2, empty(), L).
+    lists:foldr(fun chain/2, empty(), L).
```
ieQu1, line 193: What is the reason of this change?

keynslug: This preserves the order of `L`, as evidenced by some of the `emqx_ds` tests. Without this change, once `add_generation` events are added into a test event stream, [this call](https://github.com/emqx/emqx/blob/a37376e3530aa10b141ffb9b0e95c0149452379d/apps/emqx_durable_storage/test/emqx_ds_test_helpers.erl#L287-L288) results in a stream which goes from newer generations to older.

ieQu1: Thanks!

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -305,21 +330,21 @@ store_batch(Shard, Messages, Options) ->
...
-    emqx_ds:message_store_opts()
+    batch_prepare_opts()
 ) -> {ok, cooked_batch()} | ignore | emqx_ds:error(_).
-prepare_batch(Shard, Messages = [{Time, _} | _], Options) ->
+prepare_batch(Shard, Messages = [{Time, _} | _], _Options) ->
     %% NOTE
     %% We assume that batches do not span generations. Callers should enforce this.
-    ?tp(emqx_ds_storage_layer_prepare_batch, #{
-        shard => Shard, messages => Messages, options => Options
-    }),
```
ieQu1, line 315: ?

keynslug: This was almost exactly the same tracepoint as `emqx_ds_storage_layer_store_batch`. I transformed it into `emqx_ds_storage_layer_batch_cooked` with some extra content.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -305,21 +330,21 @@ store_batch(Shard, Messages, Options) ->
...
     %% FIXME: always store messages in the current generation
     case generation_at(Shard, Time) of
         {GenId, #{module := Mod, data := GenData}} ->
             T0 = erlang:monotonic_time(microsecond),
             Result =
-                case Mod:prepare_batch(Shard, GenData, Messages, Options) of
+                case Mod:prepare_batch(Shard, GenData, Messages) of
```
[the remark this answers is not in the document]
keynslug: There's currently no options that can reasonably affect "cooking" but can affect writing itself (unless obviously we want to bake the write options into the cooked batch). So `Options` "moved" into the `commit` callback.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
ieQu1: I see, thanks. Now `store_batch` is merely a convenience wrapper that combines two separate primitives together, so I think it's better to remove `emqx_ds_storage_layer_store_batch` tracepoint and keep this one.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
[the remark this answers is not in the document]
ieQu1: I think it's worth to reserve the possibility of passing options to both callbacks, unless there is a logical or consistency-related reason not to allow `prepare_batch` to access option.

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3405)

PR #13238, opened 2024-06-12
fix(http authz): handle unknown content types in responses

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12530

Release version: v/e5.7.1

## PR Checklist


keynslug (approved):
The bug report's stacktrace points to https://github.com/emqx/emqx/blob/6fe8a09e97d8bfc0b6aff640072f188ca167e0de/apps/emqx_auth_http/src/emqx_authz_http.erl#L248-L249, which looks like another issue: we can't really issue authz request with an arbitrary content type.

I guess the schema is too permissive in the `content-type` header values, the validator would help but also make the schema backward-incompatible.

thalesmg:
> The bug report's stacktrace points to
> 
> https://github.com/emqx/emqx/blob/6fe8a09e97d8bfc0b6aff640072f188ca167e0de/apps/emqx_auth_http/src/emqx_authz_http.erl#L248-L249
> , which looks like another issue: we can't really issue authz request with an arbitrary content type.
> 
> I guess the schema is too permissive in the `content-type` header values, the validator would help but also make the schema backward-incompatible.

True.  The ticket is a bit confusing because it shows a configuration where the _request content-type_ is `application/json`, but stacktrace suggests another issue.  I found that a _response content-type_ also had a similar crash, so I at least started with this one until a more precise ticket is made.  :see_no_evil:

PR #13242, opened 2024-06-13
feat:  make the dashboard restart quicker

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-11334
Related PR: https://github.com/emqx/minirest/pull/124

When the dashboard generates a dispatch, it is very slow, taking around 5-10 seconds or more. This results in a delay of 5-10 seconds each time the dashboard is restarted during a cluster join.
With this change, the previous dispatch will be used during the dashboard restart (the dispatch is the same if the emqx version remains unchanged). This will allow the dashboard back to normal without waiting for a new dispatch to be generated.

PS: The dispatch always be regenerated after dashboard restart, The dispatch cache also be updated as expected.

Release version: v/e5.8.0

## PR Checklist

### DIFF
--- a/apps/emqx/include/logger.hrl
+++ b/apps/emqx/include/logger.hrl
@@ -119,13 +119,7 @@ end).
 -endif.
 
 %% print to 'user' group leader
--define(ULOG(Fmt, Args),
-    io:format(user, "~ts " ++ Fmt, [emqx_utils_calendar:now_to_rfc3339(millisecond) | Args])
-).
--define(ELOG(Fmt, Args),
-    io:format(standard_error, "~ts " ++ Fmt, [
-        emqx_utils_calendar:now_to_rfc3339(millisecond) | Args
-    ])
-).
+-define(ULOG(Fmt, Args), io:format(user, Fmt, Args)).
+-define(ELOG(Fmt, Args), io:format(standard_error, Fmt, Args)).
 
 -endif.
--- a/apps/emqx/include/logger.hrl
+++ b/apps/emqx/include/logger.hrl
@@ -119,7 +119,13 @@ end).
 -endif.
 
 %% print to 'user' group leader
--define(ULOG(Fmt, Args), io:format(user, Fmt, Args)).
--define(ELOG(Fmt, Args), io:format(standard_error, Fmt, Args)).
+-define(ULOG(Fmt, Args),
+    io:format(user, "~ts " ++ Fmt, [emqx_utils_calendar:now_to_rfc3339(millisecond) | Args])
+).
+-define(ELOG(Fmt, Args),
+    io:format(standard_error, "~ts " ++ Fmt, [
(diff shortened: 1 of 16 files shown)


zmstone:
Can we compile dispatch to a static module for each version? So it’s going to be static even for the initial boot

zhongwencool:
> Can we compile dispatch to a static module for each version? So it’s going to be static even for the initial boot

Yes, we can use a static way. However, the dynamic way is more flexible for patching minor versions.

zmstone:
> > Can we compile dispatch to a static module for each version? So it’s going to be static even for the initial boot
> 
> Yes, we can use a static way. However, the dynamic way is more flexible for patching minor versions.

How about this: we use persistent_term to store the dynamically compiled dispatches, if exists, we use it, otherwise use the static module.

For upgrade/patching, if youmean patching as running `emqx relup ...` command, then we can add a step to recompile the dispatches after upgraded.
If you mean patching as loading some newer version beams, then the problem still exists, one will need to recompile the dispatches after loading the new beams.

zhongwencool:
> > > Can we compile dispatch to a static module for each version? So it’s going to be static even for the initial boot
> > 
> > 
> > Yes, we can use a static way. However, the dynamic way is more flexible for patching minor versions.
> 
> How about this: we use persistent_term to store the dynamically compiled dispatches, if exists, we use it, otherwise use the static module.
> 
> For upgrade/patching, if youmean patching as running `emqx relup ...` command, then we can add a step to recompile the dispatches after upgraded. If you mean patching as loading some newer version beams, then the problem still exists, one will need to recompile the dispatches after loading the new beams.

Done.
We cache all API rules in the 'priv' directory of emqx_dashboard.
Starting listeners with this cache ensures immediate functionality. Additionally, we regenerate it in emqx_dashboard_listener to ensure that patch changes are applied.

zmstone (approved):

apps/emqx/include/logger.hrl
```diff
@@ -119,7 +119,13 @@ end).
 -endif.
 
 %% print to 'user' group leader
--define(ULOG(Fmt, Args), io:format(user, Fmt, Args)).
--define(ELOG(Fmt, Args), io:format(standard_error, Fmt, Args)).
+-define(ULOG(Fmt, Args),
```
zmstone, line 122: This is probably a change which deserves a proper change log, and release in 5.8

changes/ce/feat-13242.en.md
```diff
@@ -0,0 +1 @@
+Significantly increased the startup speed of the EMQX management dashboard.
```
zmstone, line 1:
```suggestion
Significantly increased the startup speed of EMQX dashboard listener.
```

HJianBo:
I found this PR was merged into the release-57 branch. SO actually it will be released in 5.7.1

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2636)

PR #13276, opened 2024-06-17
fix(dsstore): persist inherited LTS trie

keynslug:
Release version: v/e5.7.1

## Summary

Before this commit, inherited trie was actually only kept in memory cache.

Also simplify storage backend behaviour around inheriting previous generation's legacy.

## PR Checklist

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
@@ -69,7 +69,6 @@
     shard_id/0,
     options/0,
     prototype/0,
-    post_creation_context/0,
     cooked_batch/0
 ]).
 
@@ -194,25 +193,19 @@
 
 -type options() :: map().
 
--type post_creation_context() ::
-    #{
-        shard_id := emqx_ds_storage_layer:shard_id(),
-        db := rocksdb:db_handle(),
-        new_gen_id := emqx_ds_storage_layer:gen_id(),
-        old_gen_id := emqx_ds_storage_layer:gen_id(),
-        new_cf_refs := cf_refs(),
-        old_cf_refs := cf_refs(),
-        new_gen_runtime_data := _NewData,
-        old_gen_runtime_data := _OldData
-    }.
-
 %%================================================================================
 %% Generation callbacks
 %%================================================================================
 
 %% Create the new schema given generation id and the options.
 %% Create rocksdb column families.
--callback create(shard_id(), rocksdb:db_handle(), gen_id(), Options :: map()) ->
+-callback create(
+    shard_id(),
+    rocksdb:db_handle(),
+    gen_id(),
+    Options :: map(),
+    PrevRuntimeData :: term()
+) ->
     {_Schema, cf_refs()}.
 
 %% Open the existing schema
@@ -258,9 +251,7 @@
 
 -callback handle_event(shard_id(), _Data, emqx_ds:time(), CustomEvent | tick) -> [CustomEvent].
 
--callback post_creation_actions(post_creation_context()) -> _Data.
-
--optional_callbacks([post_creation_actions/1, handle_event/4]).
+-optional_callbacks([handle_event/4]).
 
 %%================================================================================
 %% API for the replication layer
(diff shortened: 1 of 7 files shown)


apps/emqx_durable_storage/test/emqx_ds_storage_bitfield_lts_SUITE.erl
```diff
@@ -23,6 +23,8 @@
 -include_lib("snabbkaffe/include/snabbkaffe.hrl").
 -include_lib("stdlib/include/assert.hrl").
 
+-import(emqx_utils_conv, [bin/1]).
```
ieQu1, line 26: Better to avoid imports.

keynslug: Dropped in 10e78ea194e4b8f5592b198c9fff79723905985e.

apps/emqx_durable_storage/test/emqx_ds_storage_bitfield_lts_SUITE.erl
```diff
@@ -177,20 +177,33 @@ t_new_generation_inherit_trie(_Config) ->
...
-            Batch = [
-                begin
-                    Topic = emqx_topic:join(["wildcard", integer_to_binary(I), "suffix", Suffix]),
-                    {TS, make_message(TS, Topic, integer_to_binary(TS))}
-                end
+            Batch1 = [
+                {TS, make_message(TS, make_topic(wildcard, I, suffix, Suffix), bin(TS))}
              || I <- lists:seq(1, 200),
-                TS <- Timestamps,
+                TS <- lists:seq(0, 1_000, 10),
```
ieQu1, line 183: It's pretty confusing that `TS` is used for both timestamp and as a topic level. I think it's better to use a constant timestamp for all messages (I believe it's irrelevant in this test), and rename `TS` to `J`.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -194,25 +193,19 @@
...
 %% Create rocksdb column families.
--callback create(shard_id(), rocksdb:db_handle(), gen_id(), Options :: map()) ->
+-callback create(
+    shard_id(),
+    rocksdb:db_handle(),
+    gen_id(),
+    Options :: map(),
+    PrevRuntimeData :: term()
+) ->
     {_Schema, cf_refs()}.
```
ieQu1, line 209:
```suggestion
    Schema | undefined
) ->
    {Schema, cf_refs()}.
```

apps/emqx_durable_storage/test/emqx_ds_storage_bitfield_lts_SUITE.erl
```diff
@@ -475,6 +485,12 @@ make_message(PublishedAt, Topic, Payload) when is_binary(Topic) ->
         payload = Payload
     }.
 
+make_topic(T1, T2, T3, T4) ->
+    make_topic([T1, T2, T3, T4]).
```
ieQu1, line 489: This function looks very specific to a particular testcase. Please remove it and use `make_topic/1` with list argument.

apps/emqx_durable_storage/test/emqx_ds_storage_bitfield_lts_SUITE.erl
keynslug: It's actually not used as a topic level, but it indeed should be irrelevant.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
keynslug: The last argument is actually `RuntimeData` (i.e. the result of `open/5`), not `Schema`.

ieQu1 (approved):

apps/emqx_durable_storage/test/emqx_ds_storage_bitfield_lts_SUITE.erl
```diff
@@ -211,10 +224,7 @@ t_replay(_Config) ->
     ok = emqx_ds_storage_layer:store_batch(?SHARD, Batch1, []),
     %% Create wildcard topics `wildcard/+/suffix/foo' and `wildcard/+/suffix/bar':
     Batch2 = [
-        begin
-            Topic = emqx_topic:join(["wildcard", integer_to_list(I), "suffix", Suffix]),
-            {TS, make_message(TS, Topic, integer_to_binary(TS))}
-        end
+        {TS, make_message(TS, make_topic([wildcard, I, suffix, Suffix]), bin(TS))}
```
ieQu1, line 227: Same here, TS is probably not needed.

keynslug: Had a different impression actually: sounds important for a replay test to have messages spread both in time and in "topic space" (here, `TS` is not part of a topic).

Anyway, it'd probably be better to tackle the logic of unrelated testcase in a separate PR.

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2645)

Issue #11685, opened 2023-09-26
The keepalive_timeout and actual timeout of EMQX do not match.


wengdunliang:
### What happened?

I immediately disconnected the client after connecting to EMQX. Set the Keep-Alive Multiplier to 1 and the client keep-alive time is 120. The EMQX actually disconnected after 240 seconds.
**I set the Keep-Alive Multiplier to 1 only because if it is set to 1.5, the actual timeout is 360 seconds.**

### What did you expect to happen?

The EMQX disconnected after 120 seconds.

### Anything else we need to know?

I directly turned off the power of the embedded device to disconnect MQTT, so this may not be reproducible on the PC.

### EMQX version

v5.2.1

### Log files

2023-09-26T10:25:09.971438+08:00 [MQTT] da16x_B4BE@192.168.1.161:38554 msg: mqtt_packet_received, packet: CONNECT(Q0, R0, D0, ClientId=da16x_B4BE, ProtoName=MQIsdp, ProtoVsn=3, CleanStart=true, KeepAlive=120, Username=devices, Password=******, Will(Q2, R0, Topic=smc_will_dev Payload={"sn":"008DB260"}))
2023-09-26T10:25:09.972369+08:00 [SOCKET] da16x_B4BE@192.168.1.161:31105 msg: emqx_connection_terminated, reason: {shutdown,discarded}
2023-09-26T10:25:09.973186+08:00 [MQTT] da16x_B4BE@192.168.1.161:38554 msg: mqtt_packet_sent, packet: CONNACK(Q0, R0, D0, AckFlags=0, ReasonCode=0)
2023-09-26T10:25:09.977985+08:00 [MQTT] da16x_B4BE@192.168.1.161:38554 msg: mqtt_packet_received, packet: SUBSCRIBE(Q1, R0, D0, PacketId=1 TopicFilters=[smc_response_D0B065A8(#{nl => 0,qos => 2,rap => 0,rh => 0})])
2023-09-26T10:25:09.978404+08:00 [SUBSCRIBE] da16x_B4BE@192.168.1.161:38554 msg: subscribe, sub_id: da16x_B4BE, sub_opts: [nl: 0, qos: 2, rap: 0, rh: 0, sub_props: []], topic: smc_response_D0B065A8
[...]

PR #13315, opened 2024-06-21
feat: improve keepalive_multiplier and keepalive_check_interval

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-11151

https://github.com/emqx/emqx/issues/11685

Add a new keepalive config: mqtt.keepalive_check_interval(default is 30s)
The time interval to check for mqtt packets.
range is `[1s, Interval(in connect packet) div 2 s]`.

So the time to determine if a timeout is needed is:

- min:   interval * multiplier 
- max:  interval * multiplier + 1 * check_interval

eg:
interval = 60s, multiplier = 1.5, check_interval = 30s
should be keepalive_timeout between `90s(60*1.5) and 120s(60*1.5+30)`

If the user expects to reduce this section. You can reduce the factor:
interval = 60s, multiplier = 1.5, check_interval = 10s
should be keepalive_timeout between `90s(60 * 1.5) and 100s(60 * 1.5+10)`

Release version: v/e5.7.2

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -544,8 +544,10 @@ handle_in(
         {error, ReasonCode} ->
             handle_out(disconnect, ReasonCode, Channel)
     end;
-handle_in(?PACKET(?PINGREQ), Channel) ->
-    {ok, ?PACKET(?PINGRESP), Channel};
+handle_in(?PACKET(?PINGREQ), Channel = #channel{keepalive = Keepalive}) ->
+    {ok, NKeepalive} = emqx_keepalive:check(Keepalive),
+    NChannel = Channel#channel{keepalive = NKeepalive},
+    {ok, ?PACKET(?PINGRESP), reset_timer(keepalive, NChannel)};
 handle_in(
     ?DISCONNECT_PACKET(ReasonCode, Properties),
     Channel = #channel{conninfo = ConnInfo}
@@ -1229,11 +1231,12 @@ handle_call(
     {keepalive, Interval},
     Channel = #channel{
         keepalive = KeepAlive,
-        conninfo = ConnInfo
+        conninfo = ConnInfo,
+        clientinfo = #{zone := Zone}
     }
 ) ->
     ClientId = info(clientid, Channel),
-    NKeepalive = emqx_keepalive:update(timer:seconds(Interval), KeepAlive),
+    NKeepalive = emqx_keepalive:update(Zone, Interval, KeepAlive),
     NConnInfo = maps:put(keepalive, Interval, ConnInfo),
     NChannel = Channel#channel{keepalive = NKeepalive, conninfo = NConnInfo},
     SockInfo = maps:get(sockinfo, emqx_cm:get_chan_info(ClientId), #{}),
@@ -1333,22 +1336,22 @@ die_if_test_compiled() ->
     | {shutdown, Reason :: term(), channel()}.
 handle_timeout(
     _TRef,
-    {keepalive, _StatVal},
+    keepalive,
     Channel = #channel{keepalive = undefined}
 ) ->
     {ok, Channel};
 handle_timeout(
     _TRef,
-    {keepalive, _StatVal},
+    keepalive,
     Channel = #channel{conn_state = disconnected}
 ) ->
     {ok, Channel};
 handle_timeout(
     _TRef,
-    {keepalive, StatVal},
+    keepalive,
     Channel = #channel{keepalive = Keepalive}
 ) ->
-    case emqx_keepalive:check(StatVal, Keepalive) of
+    case emqx_keepalive:check(Keepalive) of
         {ok, NKeepalive} ->
             NChannel = Channel#channel{keepalive = NKeepalive},
             {ok, reset_timer(keepalive, NChannel)};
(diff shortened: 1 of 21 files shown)


apps/emqx/src/emqx_keepalive.erl
```diff
@@ -31,7 +32,8 @@
 
 -record(keepalive, {
     interval :: pos_integer(),
-    statval :: non_neg_integer()
+    statval :: non_neg_integer(),
+    repeat = 0 :: non_neg_integer()
```
zmstone, line 36:
```suggestion
    %% The number of idle intervals allowed before disconnecting the client.
    idle_intervals = 0 :: non_neg_integer()
```

apps/emqx/src/emqx_keepalive.erl
```diff
@@ -31,7 +32,8 @@
 
 -record(keepalive, {
     interval :: pos_integer(),
-    statval :: non_neg_integer()
+    statval :: non_neg_integer(),
```
zmstone, line 35:
```suggestion
    %% the received packets since last keepalive check
    statval :: non_neg_integer(),
```

apps/emqx/src/emqx_channel.erl
```diff
@@ -2322,7 +2324,7 @@ ensure_keepalive_timer(disabled, Channel) ->
 ensure_keepalive_timer(Interval, Channel = #channel{clientinfo = #{zone := Zone}}) ->
     Multiplier = get_mqtt_conf(Zone, keepalive_multiplier),
     RecvCnt = emqx_pd:get_counter(recv_pkt),
-    Keepalive = emqx_keepalive:init(RecvCnt, round(timer:seconds(Interval) * Multiplier)),
+    Keepalive = emqx_keepalive:init(RecvCnt, round(timer:seconds(Interval) * Multiplier / 2)),
```
zmstone, line 2327: Move this calculation to emqx_keepalive.erl.

Also:
I believe the reason for MQTT specification to define `1.5` as multiplier is to give 0.5 of the interval shift for the checks.
To better align the timer expire even with the intervals, we should check more often, a good default is `Interval div 2`, and allow client to be idle for 3 checks. 

If we hard-code the number of idle checks to 2, the error (误差）becomes bigger when the interval is larger.

In order to allow users to fine-tune the parameters to lower the error (误差）, we should make it possible for users to adjust both the interval, and the number of idle checks.

```
TimerInterval = timer:seconds(Interval) div 2,
MaxIdleInterval = floor(2 * Multiplier + 0.000001),
```

As an example, for default multiplier `1.5` and interval=`12s`, the check is done every `6s` and disconnected after `18s` but before `24s`.

Another example: if one configures `multiplier=9` and `interval=2s`, then the check is done every `1s`, the client is STILL disconnected after `18s` but before `19s`.

I remember discussing this with someone quite a long time ago, but in the end it was implemented wrong for some reason I cannot recall.

zmstone (changes requested):

apps/emqx/src/emqx_channel.erl
zhongwencool: There's a previous discussion here ? https://github.com/emqx/emqx/issues/11685#issuecomment-1764108957

But, IMO: Make multiplier high and interval too small, although it can be used this way, it is obviously not good(This is not really what the user wants). interval = 1s means that ping requests are sent every second. This is not a recommended design for either the client or the server.

It's a good idea to make this `2` a configurable item.

rel/i18n/emqx_schema.hocon
```diff
@@ -855,6 +855,21 @@ The default value 1.5 is following the MQTT 5.0 specification. This multiplier i
 mqtt_keepalive_multiplier.label:
 """Keep Alive Multiplier"""
 
+mqtt_keepalive_check_factor.desc:
```
zmstone, line 858: As discussed in slack, to make it easier to configure:
```suggestion
mqtt_keepalive_check_interval.desc:
```

and clamp the config value in the range of `[1000, (KaIntervalSeconds div 2) * 1000]`

rel/i18n/emqx_schema.hocon
```diff
@@ -855,6 +855,15 @@ The default value 1.5 is following the MQTT 5.0 specification. This multiplier i
 mqtt_keepalive_multiplier.label:
 """Keep Alive Multiplier"""
 
+mqtt_keepalive_check_interval.desc:
+"""The frequency of checking for incoming MQTT packets determines how often the server will check for new MQTT packets.
+If a certain amount of time passes without any packets being sent from the client,this time will be added up.
```
zmstone, line 860:
```suggestion
If a certain amount of time passes without any packets being sent from the client, this time will be added up.
```

rel/i18n/emqx_schema.hocon
```diff
@@ -855,6 +855,15 @@ The default value 1.5 is following the MQTT 5.0 specification. This multiplier i
...
+Once the accumulated time exceeds the keepalive interval * the keepalive multiplier, the connection will be terminated.
```
zmstone, line 861:
```suggestion
Once the accumulated time exceeds `keepalive-interval * keepalive-multiplier`, the connection will be terminated.
```

rel/i18n/emqx_schema.hocon
```diff
@@ -855,6 +855,15 @@ The default value 1.5 is following the MQTT 5.0 specification. This multiplier i
...
+The default is set to 30 seconds, with a minimum value of 1 second and a maximum value of Interval/2."""
```
zmstone, line 862:
```suggestion
The default is set to 30 seconds, with a minimum value of 1 second and a maximum value of `keepalive-interval / 2`."""
```

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2720)

PR #13331, opened 2024-06-25
fix(api_keys): improved the error message when bootstrapping api_key

lafirest:
Fixes EMQX-12210

Release version: v/e5.7.2

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_auth.erl
+++ b/apps/emqx_management/src/emqx_mgmt_auth.erl
@@ -357,6 +357,11 @@ init_bootstrap_file(File) ->
             init_bootstrap_file(File, Dev, MP);
         {error, Reason0} ->
             Reason = emqx_utils:explain_posix(Reason0),
+            FmtReason = emqx_utils:format(
+                "load API bootstrap file failed, file:~ts, reason:~ts",
+                [File, Reason]
+            ),
+
             ?SLOG(
                 error,
                 #{
@@ -365,7 +370,8 @@ init_bootstrap_file(File) ->
                     reason => Reason
                 }
             ),
-            {error, Reason}
+
+            {error, FmtReason}
     end.
 
 init_bootstrap_file(File, Dev, MP) ->


apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -365,6 +369,7 @@ init_bootstrap_file(File) ->
                     reason => Reason
```
zhongwencool, line 369: msg has been printed in the log. There should be no need to repeat it.
only return the format-reason but print explain-reason in log.

apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -356,7 +356,11 @@ init_bootstrap_file(File) ->
             {ok, MP} = re:compile(<<"(\.+):(\.+)(?::(\.+))?$">>, [ungreedy]),
             init_bootstrap_file(File, Dev, MP);
         {error, Reason0} ->
-            Reason = emqx_utils:explain_posix(Reason0),
+            Reason = io:format(
+                "load API bootstrap file failed, file:~ts, reason:~ts",
+                [File, emqx_utils:explain_posix(Reason0)]
+            ),
```
zhongwencool, line 362:
```suggestion
           Reason = emqx_utils:explain_posix(Reason0),
            FmtReason = emqx_utils:format((
                "load API-Key's bootstrap file failed, file:~ts, reason:~ts",
                [File, Reason]
            ),
```

zhongwencool (changes requested):

apps/emqx_management/src/emqx_mgmt_auth.erl
lafirest: No, most of the time this function is called at startup and will cause EMQX to crash, thus failing to print messages in the log file.

zhongwencool (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-2888)

PR #13370, opened 2024-06-28
New durable storage layout with explicit index for LTS wildcards

ieQu1:
Fixes https://emqx.atlassian.net/browse/EMQX-12536

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx_ds_builtin_local/src/emqx_ds_builtin_local.erl
+++ b/apps/emqx_ds_builtin_local/src/emqx_ds_builtin_local.erl
@@ -46,6 +46,12 @@
     shard_of_message/4
 ]).
 
+%% Internal exports:
+-export([
+    do_next/3,
+    do_delete_next/4
+]).
+
 -export_type([db_opts/0, shard/0, iterator/0, delete_iterator/0]).
 
 -include_lib("emqx_utils/include/emqx_message.hrl").
@@ -295,19 +301,8 @@ update_iterator(DB, Iter0 = #{?tag := ?IT, ?shard := Shard, ?enc := StorageIter0
     end.
 
 -spec next(emqx_ds:db(), iterator(), pos_integer()) -> emqx_ds:next_result(iterator()).
-next(DB, Iter0 = #{?tag := ?IT, ?shard := Shard, ?enc := StorageIter0}, N) ->
-    ShardId = {DB, Shard},
-    T0 = erlang:monotonic_time(microsecond),
-    Result = emqx_ds_storage_layer:next(ShardId, StorageIter0, N, current_timestamp(ShardId)),
-    T1 = erlang:monotonic_time(microsecond),
-    emqx_ds_builtin_metrics:observe_next_time(DB, T1 - T0),
-    case Result of
-        {ok, StorageIter, Batch} ->
-            Iter = Iter0#{?enc := StorageIter},
-            {ok, Iter, Batch};
-        Other ->
-            Other
-    end.
+next(DB, Iter, N) ->
+    with_worker(do_next, [DB, Iter, N]).
 
 -spec get_delete_streams(emqx_ds:db(), emqx_ds:topic_filter(), emqx_ds:time()) ->
     [emqx_ds:ds_specific_delete_stream()].
@@ -347,7 +342,36 @@ make_delete_iterator(DB, ?delete_stream(Shard, InnerStream), TopicFilter, StartT
 
 -spec delete_next(emqx_ds:db(), delete_iterator(), emqx_ds:delete_selector(), pos_integer()) ->
     emqx_ds:delete_next_result(emqx_ds:delete_iterator()).
-delete_next(DB, Iter = #{?tag := ?DELETE_IT, ?shard := Shard, ?enc := StorageIter0}, Selector, N) ->
+delete_next(DB, Iter, Selector, N) ->
+    with_worker(do_delete_next, [DB, Iter, Selector, N]).
+
+%%================================================================================
+%% Internal exports
+%%================================================================================
+
+current_timestamp(ShardId) ->
+    emqx_ds_builtin_local_meta:current_timestamp(ShardId).
+
+-spec do_next(emqx_ds:db(), iterator(), pos_integer()) -> emqx_ds:next_result(iterator()).
+do_next(DB, Iter0 = #{?tag := ?IT, ?shard := Shard, ?enc := StorageIter0}, N) ->
+    ShardId = {DB, Shard},
+    T0 = erlang:monotonic_time(microsecond),
+    Result = emqx_ds_storage_layer:next(ShardId, StorageIter0, N, current_timestamp(ShardId)),
+    T1 = erlang:monotonic_time(microsecond),
+    emqx_ds_builtin_metrics:observe_next_time(DB, T1 - T0),
+    case Result of
+        {ok, StorageIter, Batch} ->
+            Iter = Iter0#{?enc := StorageIter},
+            {ok, Iter, Batch};
+        Other ->
+            Other
+    end.
+
+-spec do_delete_next(emqx_ds:db(), delete_iterator(), emqx_ds:delete_selector(), pos_integer()) ->
+    emqx_ds:delete_next_result(emqx_ds:delete_iterator()).
+do_delete_next(
+    DB, Iter = #{?tag := ?DELETE_IT, ?shard := Shard, ?enc := StorageIter0}, Selector, N
+) ->
     ShardId = {DB, Shard},
     case
         emqx_ds_storage_layer:delete_next(
@@ -363,13 +387,6 @@ delete_next(DB, Iter = #{?tag := ?DELETE_IT, ?shard := Shard, ?enc := StorageIte
     end.
 
 %%================================================================================
-%% Internal exports
-%%================================================================================
-
-current_timestamp(ShardId) ->
-    emqx_ds_builtin_local_meta:current_timestamp(ShardId).
-
-%%================================================================================
 %% Internal functions
 %%================================================================================
 
@@ -380,3 +397,20 @@ timeus_to_timestamp(undefined) ->
     undefined;
 timeus_to_timestamp(TimestampUs) ->
     TimestampUs div 1000.
+
+with_worker(F, A) ->
+    Parent = self(),
+    Ref = make_ref(),
+    {_Pid, MRef} = spawn_opt(
+        fun() ->
+            Parent ! {Ref, apply(?MODULE, F, A)}
+        end,
+        [monitor, {min_heap_size, 10000}]
+    ),
+    receive
+        {Ref, Result} ->
+            erlang:demonitor(MRef, [flush]),
+            Result;
+        {'DOWN', MRef, _, _, _, Info} ->
+            {error, unrecoverable, Info}
+    end.
--- a/apps/emqx_ds_builtin_raft/test/emqx_ds_replication_SUITE.erl
+++ b/apps/emqx_ds_builtin_raft/test/emqx_ds_replication_SUITE.erl
@@ -29,15 +29,12 @@
     emqx_ds_test_helpers:on(NODES, fun() -> BODY end)
 ).
 
-opts() ->
-    opts(#{}).
-
-opts(Overrides) ->
+opts(Config, Overrides) ->
+    Layout = ?config(layout, Config),
     maps:merge(
         #{
             backend => builtin_raft,
-            %% storage => {emqx_ds_storage_reference, #{}},
-            storage => {emqx_ds_storage_bitfield_lts, #{epoch_bits => 10}},
+            storage => Layout,
             n_shards => 16,
             n_sites => 1,
             replication_factor => 3,
@@ -58,7 +55,7 @@ appspec(emqx_durable_storage) ->
 
 t_metadata(init, Config) ->
     Apps = emqx_cth_suite:start([emqx_ds_builtin_raft], #{
-        work_dir => emqx_cth_suite:work_dir(?FUNCTION_NAME, Config)
+        work_dir => ?config(work_dir, Config)
     }),
     [{apps, Apps} | Config];
 t_metadata('end', Config) ->
@@ -108,7 +105,7 @@ t_replication_transfers_snapshots(init, Config) ->
             {t_replication_transfers_snapshots2, #{apps => Apps}},
             {t_replication_transfers_snapshots3, #{apps => Apps}}
         ],
-        #{work_dir => emqx_cth_suite:work_dir(?FUNCTION_NAME, Config)}
+        #{work_dir => ?config(work_dir, Config)}
     ),
     Nodes = emqx_cth_cluster:start(NodeSpecs),
     [{nodes, Nodes}, {specs, NodeSpecs} | Config];
@@ -125,9 +122,10 @@ t_replication_transfers_snapshots(Config) ->
     Nodes = [Node, NodeOffline | _] = ?config(nodes, Config),
     _Specs = [_, SpecOffline | _] = ?config(specs, Config),
     ?check_trace(
+        #{timetrap => 30_000},
         begin
             %% Initialize DB on all nodes and wait for it to be online.
-            Opts = opts(#{n_shards => 1, n_sites => 3}),
+            Opts = opts(Config, #{n_shards => 1, n_sites => 3}),
             ?assertEqual(
                 [{ok, ok} || _ <- Nodes],
                 erpc:multicall(Nodes, emqx_ds, open_db, [?DB, Opts])
@@ -139,8 +137,11 @@ t_replication_transfers_snapshots(Config) ->
             ),
 
             %% Stop the DB on the "offline" node.
-            ok = emqx_cth_cluster:stop_node(NodeOffline),
-            _ = ?block_until(#{?snk_kind := ds_ra_state_enter, state := leader}, 500, 0),
+            ?wait_async_action(
+                ok = emqx_cth_cluster:stop_node(NodeOffline),
+                #{?snk_kind := ds_ra_state_enter, state := leader},
+                5_000
+            ),
 
             %% Fill the storage with messages and few additional generations.
             emqx_ds_test_helpers:apply_stream(?DB, Nodes -- [NodeOffline], Stream),
@@ -153,9 +154,10 @@ t_replication_transfers_snapshots(Config) ->
                     ?snk_meta := #{node := NodeOffline}
                 })
             ),
-            ?assertEqual(
-                ok,
-                erpc:call(NodeOffline, emqx_ds, open_db, [?DB, opts()])
+
+            ok = ?ON(
+                NodeOffline,
+                emqx_ds:open_db(?DB, opts(Config, #{}))
             ),
 
             %% Trigger storage operation and wait the replica to be restored.
@@ -183,7 +185,7 @@ t_rebalance(init, Config) ->
             {t_rebalance3, #{apps => Apps}},
             {t_rebalance4, #{apps => Apps}}
         ],
-        #{work_dir => emqx_cth_suite:work_dir(?FUNCTION_NAME, Config)}
+        #{work_dir => ?config(work_dir, Config)}
     ),
     [{nodes, Nodes} | Config];
 t_rebalance('end', Config) ->
@@ -206,7 +208,7 @@ t_rebalance(Config) ->
         begin
             Sites = [S1, S2 | _] = [ds_repl_meta(N, this_site) || N <- Nodes],
             %% 1. Initialize DB on the first node.
-            Opts = opts(#{n_shards => 16, n_sites => 1, replication_factor => 3}),
+            Opts = opts(Config, #{n_shards => 16, n_sites => 1, replication_factor => 3}),
             [
                 ?assertEqual(ok, ?ON(Node, emqx_ds:open_db(?DB, Opts)))
              || Node <- Nodes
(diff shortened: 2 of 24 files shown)


apps/emqx_durable_storage/asn.1/DurableMessage.asn
```diff
@@ -0,0 +1,90 @@
...
+  -- Wrapper for any data that doesn't comply with the strict schema:
+  Misc ::= CHOICE {
+    flag MiscFlag,
+    header MiscProperty,
+    property MiscProperty,
+    extra MiscProperty
+  }
+
+  -- Both key and value are interpreted as binaries:
+  UserProperty ::= SEQUENCE {
```
ieQu1, line 32: Note: currently it's unused. I am not sure if we really want to store it for each message. Might remove it.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -0,0 +1,739 @@
...
+
+finalize_loop(It0, Op, Acc) ->
+    case Op of
+        next -> NextTS = It0#it.ts + 1;
+        {seek, NextTS} -> ok
+    end,
+    It = It0#it{ts = NextTS},
+    {ok, It, lists:reverse(Acc)}.
+
+next_step(
```
savonarola, line 508: To clarify for myself:

If we have `a/b/c` topic filter for `+ -> + -> c` lts, we will have 3 indexes to lookup:
```
[
   {static_key(c), 0, hash("a")},
   {static_key(c), 1, hash("b")},
   {static_key(c)} // data stream
]
```

If we have numerous messages in `a/y/c` and `x/b/c` and rare `a/b/c` messages we will anyway iterate  `x/b/c -> ts1` -> `a/y/c -> ts1 > ts2` -> `x/b/c -> ts3 > ts2` -> `a/y/c -> ts4 > ts3` till we get to the first `a/b/c` ?

ieQu1: Yes. It is theoretically possible to iterate level indices in any order, but this doesn't exploit this possibility. If need arises, it can be implemented.

apps/emqx_durable_storage/src/emqx_ds_lts.erl
```diff
@@ -450,12 +583,51 @@ follow_path(#trie{} = T, State, Path) ->
...
+tokens_of_path([]) ->
+    [].
+
+%% Wrapper for type checking only:
+-compile({inline, ets_insert_new/2}).
+-spec ets_insert_new(ets:tid(), trans()) -> boolean().
+ets_insert_new(Tid, Trans) ->
+    ets:insert_new(Tid, Trans).
+
+compress_topic(_StaticKey, [], [], Acc) ->
```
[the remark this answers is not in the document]
ieQu1: I think it's a leftover.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
savonarola: Could this result to a significant degradation vs keymapper (which can handle concrete streams falling into wilcard lts path perfectly, i.e. without any excess reads)?

savonarola: Just as a brainstorm,

we may add a "full" index for the varying part

```
[
   {static_key(c), all, hash(compressed_topic)},
   {static_key(c), 0, hash(0th segment)},
   {static_key(c), 1, hash(1st segment)},
   {static_key(c)} // data stream
]
```

to handle concrete topics, which may be a frequent case 🤔

ieQu1: I guess it could work.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -0,0 +1,739 @@
...
+    %%     ts => It0#it.ts, tf => It0#it.compressed_tf, bs => BatchSize, tmax => TMax, op => Op
+    %% }),
+    #ctx{s = S, tmax = TMax, iters = Iterators} = Ctx,
+    #it{static_index = StaticIdx, compressed_tf = CompressedTF} = It0,
+    case next_step(S, StaticIdx, CompressedTF, Iterators, undefined, Op) of
+        none ->
+            %% ?tp(notice, skipstream_loop_result, #{r => none}),
+            inc_counter(?DS_SKIPSTREAM_LTS_EOS),
+            finalize_loop(It0, Op, Acc);
+        {seek, TS} when TS > TMax ->
```
keynslug, line 481: Q: I wonder why it's still important to have this stopping criterion here?

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -0,0 +1,739 @@
...
+                })
+        end,
+    Ctx = #ctx{
+        shard = Shard,
+        s = S,
+        iters = Iterators,
+        topic_structure = TopicStructure,
+        filter = words(CompressedTF),
+        tmax = TMax
+    },
```
keynslug, line 465: While this is much better from readability perspective, in my experience it keeping this on the stack (i.e. as arguments) helps significantly in hot loops.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -0,0 +1,739 @@
...
+    do_init_iterators(S, Static, words(CompressedTF), 1).
+
+do_init_iterators(S, Static, ['+' | TopicFilter], WildcardLevel) ->
+    %% Ignore wildcard levels in the topic filter:
+    do_init_iterators(S, Static, TopicFilter, WildcardLevel + 1);
+do_init_iterators(S, Static, [Constraint | TopicFilter], WildcardLevel) ->
+    %% Create iterator for the index stream:
+    #s{hash_bytes = HashBytes, db = DB, data_cf = DataCF} = S,
+    Hash = hash(HashBytes, Constraint),
+    {ok, ItHandle} = rocksdb:iterator(DB, DataCF, get_key_range(Static, WildcardLevel, Hash)),
```
keynslug, line 420: Q: Why aren't we using `It#it.ts` (and possibly `TMax`) in the `get_key_range` computation?

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -0,0 +1,739 @@
...
+get_delete_streams(_Shard, #s{trie = Trie}, TopicFilter, _StartTime) ->
+    get_streams(Trie, TopicFilter).
+
+make_iterator(_Shard, #s{trie = Trie}, #stream{static_index = StaticIdx}, TopicFilter, StartTime) ->
+    {ok, TopicStructure} = emqx_ds_lts:reverse_lookup(Trie, StaticIdx),
+    CompressedTF = emqx_ds_lts:compress_topic(StaticIdx, TopicStructure, TopicFilter),
+    {ok, #it{
+        static_index = StaticIdx,
+        ts = StartTime,
+        compressed_tf = emqx_topic:join(CompressedTF)
```
keynslug, line 250: Nit. It kind of seems like `compressed_tf` is used in "tokenized" form much more often in this module.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -0,0 +1,739 @@
...
+            false
+    end.
+
+-spec mk_key(emqx_ds_lts:static_key(), wildcard_idx(), wildcard_hash(), ts()) -> binary().
+mk_key(StaticIdx, 0, <<>>, Timestamp) ->
+    %% Data stream is identified by wildcard level = 0
+    <<StaticIdx/binary, 0:?wcb, Timestamp:?tsb>>;
+mk_key(StaticIdx, N, Hash, Timestamp) when N > 0 ->
+    %% Index stream:
+    <<StaticIdx/binary, N:?wcb, Hash/binary, Timestamp:?tsb>>.
```
keynslug, line 621: Speculation. If these wildcard indices are allocated other way around (i.e. `16#ffff` for the message stream, `16#fffe` for the first index, etc.), then a single iterator supposedly could be used to iterate through in the `next_loop/6` (obviously, at the cost of ability to visit indices in any order). It's unclear how much this would affect performance, but I guess it might be worth measuring.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -0,0 +1,739 @@
...
+
+hash(HashBytes, '') ->
+    hash(HashBytes, <<>>);
+hash(HashBytes, TopicLevel) ->
+    {Hash, _} = split_binary(erlang:md5(TopicLevel), HashBytes),
```
keynslug, line 626: Tiny nit. While `<<Hash:HashBytes/bytes, _/binary>>` looks uglier, it doesn't require allocating a tuple on the heap.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -0,0 +1,739 @@
...
+                                    %% Hash collision. Advance to the
+                                    %% next timestamp:
+                                    inc_counter(?DS_SKIPSTREAM_LTS_HASH_COLLISION),
+                                    {seek, NextTS + 1}
+                            end;
+                        _ ->
+                            %% This is index stream. Keep going:
+                            next_step(S, StaticIdx, CompressedTF, Iterators, NextTS, {seek, NextTS})
+                    end;
+                NextTS when NextTS > ExpectedTS, N > 0 ->
```
keynslug, line 557: Nit. The `NextTS when NextTS > ExpectedTS, N == 0` clause is notably missing. I guess we don't really expect that because it would mean our storage is corrupted and atomicity of some batch write was lost, but I think having a comment would help.

keynslug (commented):
A cool idea!

apps/emqx_ds_builtin_local/src/emqx_ds_builtin_local.erl
```diff
@@ -380,3 +397,20 @@ timeus_to_timestamp(undefined) ->
...
+
+with_worker(F, A) ->
+    Parent = self(),
+    Ref = make_ref(),
+    {_Pid, MRef} = spawn_opt(
+        fun() ->
+            Parent ! {Ref, apply(?MODULE, F, A)}
+        end,
+        [monitor, {min_heap_size, 10000}]
+    ),
```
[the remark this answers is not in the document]
ieQu1: I don't think this needs a lot of testing because Raft backend always runs table scan in a worker process, even for the calls that end up at the local node. (There's a gen_rpc optimization that bypasses network layer for such calls, but it still spawns a temporary process for various reasons). Any code that uses an g?e(n_)?rpc implicitly does the same thing. 
And my tests show that inhibiting GC induced by running table scan in the same process leads to better throughput.

apps/emqx_ds_builtin_raft/test/emqx_ds_replication_SUITE.erl
```diff
@@ -984,12 +997,36 @@ sample(N, List) ->
...
+init_per_testcase(TCName, Config0) ->
+    Config1 = [{work_dir, emqx_cth_suite:work_dir(TCName, Config0)} | Config0],
+    emqx_common_test_helpers:init_per_testcase(?MODULE, TCName, Config1).
+
 end_per_testcase(TCName, Config) ->
     ok = snabbkaffe:stop(),
-    emqx_common_test_helpers:end_per_testcase(?MODULE, TCName, Config).
+    Result = emqx_common_test_helpers:end_per_testcase(?MODULE, TCName, Config),
+    catch emqx_ds:drop_db(TCName),
+    emqx_cth_suite:clean_work_dir(?config(work_dir, Config)),
```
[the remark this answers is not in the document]
ieQu1: Running replication suite _once_ generates 10s of GB of garbage, especially when it's unstable and retrying something. Commenting this line for debugging is easy enough.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: Future replication layers can use it to communicate high watermark of fully replicated data down the line.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: This is result of final refactoring. Original version passed everything through function arguments as you described. I didn't find any measurable difference in the throughput or flamegraph before and after introducing context record.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: Erlang will employ TLV encoding to encode list of binaries, which has overhead. For example:
  
```
5> size(term_to_binary([<<"a">>, <<"b">>])).
19
6> size(term_to_binary(<<"a/b">>)).
9
```

Given that iterator objects are stored and transfered over network, it's _a_ factor. There's no way to predict what cost will be larger:

- of parsing the topic
- of reading and writing larger iterator objects from disk and sending them over network.

So I chose to join them for easier debugging, I guess.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: Interesting idea. Not sure about all the corner cases of using a single iterator, but at this stage it's very easy to change the order of indexes from incrementing to decrementing.

ieQu1: P.S. Not sure how this will work. Wildcard index is in the prefix of the key, timestamp is the suffix. In order to jump to the next index level, the iterator will have to jump over all indexes and back.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: Why is it needed? AFAIK, specifying iterator range helps to avoid potentially expensive SST lookup when iterator overshoots the expected range.

Here:

1) All iterators always move forward, so it can't get behind `it.ts`
2) Key with T>TMax is likely freshly written

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: This will break `a/+` lookups.

ieQu1: > Could this result to a significant degradation vs keymapper (which can handle concrete streams falling into а wilcard lts path perfectly, i.e. without any excess reads)?

1. Perfectly defined topic filters without wildcards, by definition, have to scan less messages.
2. Old layout with keymapper *cannot* do a perfect scan either. Keymapper simply creates an implicit skiplist that allows to jump to the next epoch. You still to have go over all epochs. Longer epochs can optimize the number of such jumps, but lead to larger e2e latency.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: Replication layer must not advance NextTS past the point that was replicated.

ieQu1: Added a comment.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -0,0 +1,739 @@
...
+-type s() :: #s{}.
+
+-record(stream, {
+    static_index :: emqx_ds_lts:static_key()
+}).
+
+-record(it, {
+    static_index :: emqx_ds_lts:static_key(),
+    ts :: ts(),
+    compressed_tf :: binary()
```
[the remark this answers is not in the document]
ieQu1: Added a comment.

apps/emqx_ds_builtin_raft/test/emqx_ds_replication_SUITE.erl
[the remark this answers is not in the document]
keynslug: > Commenting this line for debugging is easy enough.

Not that easy when tests fail on CI.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
keynslug: I see.

> GC

My point is that this approach usually contributes to higher GC pressure (i.e. garbage needs to be collected much more often).

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
keynslug: I think it can also help to narrow down a set of SSTs that need to be even considered for a scan, so a tighter range can in theory lead to less IO.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
keynslug: > the iterator will have to be moved backwards

Assumption is to start from the index with the smallest `N`, i.e. first from `16#fffc`, then `16#fffd`, etc.

> It also contradicts your other comment about specifying a tighter range for the iterators.

Yes, I know. They are independent comments.

apps/emqx_ds_builtin_local/src/emqx_ds_builtin_local.erl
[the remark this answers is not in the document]
keynslug: > And my tests show that inhibiting GC induced by running table scan in the same process leads to more stable throughput.

Even if, say, there are only 2 new messages in a stream?

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: This tuple is never updated (it contains _immutable_ loop context), why would it create any garbage?

keynslug: Ah right, good point, I missed that. Disregard the comment.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
keynslug: I see now, right. Jumping backwards is inevitable in this scheme.

> What does this achieve?

The same I guess: having a single iterator. After all, this is a speculation about if this can be beneficial to performance to manage a single iterator instead of many (e.g. for small batch sizes).

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
keynslug: > This will break `a/+` lookups.

True. Well, not really _breaks_ but doesn't help, i.e. forces to scan through the message stream (in that case I guess having `hash(Varying)` in the key would be a must). But on the other hand, helps to do less index scans when iterators have more specific topic filters.

ieQu1: Sorry, do you think that this layout will have *unacceptable* performance for a particularly common traffic pattern? Please describe the traffic pattern, and I will test it.

Otherwise I don't see the reason to brainstorm a *totally different* solution B in the comments for solution A.

keynslug: > Otherwise I don't see the reason to brainstorm a totally different solution B in the comments for solution A.

Sure, disregard my intervention then.

apps/emqx_ds_builtin_local/src/emqx_ds_builtin_local.erl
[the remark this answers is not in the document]
ieQu1: Like I said, the "optimization" of running scan in the same process is not going to stick, whether it's more or not.

1) It's specific to builtin local backend (which is not our top priority). If something can't be reused in Raft (or other HA backend), we don't implement it in local.
2) It can't be used with async fetch (must have for *session* performance), we don't want to block PUBACKS and mess up MQTT channel flow control

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: Suppose we find out that creating rocksdb iterators is very expensive. With the current architecture we have ~N_clients * N_wildcard_filters iterators. If N_clients = 1000000, and N_wildcard_filters=3, then with this optimization we can go from 3000000 iterators to 1000000.

Or we can focus our efforts on different things, like RAM cache and stream events, and reduce the number of RocksDB iterators from 3000000 to 30000, because sessions will stop hammering storage.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
ieQu1: Also this discussion seem to focus on the fact that topic index lookups can miss. Yes, but notice that this layout *uses* the timestamp of mismatched topic level indexes to advance iterator.

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3424)

PR #13391, opened 2024-07-02
perf(sessds): rotate through streams with iterators when fetching

keynslug:
Fixes [EMQX-12611](https://emqx.atlassian.net/browse/EMQX-12611).

## PR Checklist

[EMQX-12611]: https://emqx.atlassian.net/browse/EMQX-12611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
@@ -66,7 +66,7 @@
     n_awaiting_rel/1
 ]).
 
--export([iterate/1]).
+-export([iter_next/1]).
 
 -export([make_session_iterator/0, session_iterator_next/2]).
 
@@ -484,6 +484,9 @@ fold_streams(Fun, Acc, Rec) ->
 -spec iter_streams(_StartAfter :: stream_key() | beginning, t()) ->
     iter(stream_key(), emqx_persistent_session_ds:stream_state()).
 iter_streams(After, Rec) ->
+    %% NOTE
+    %% No special handling for `beginning', as it always compares less
+    %% than any `stream_key()'.
     gen_iter_after(?streams, After, Rec).
 
 -spec n_streams(t()) -> non_neg_integer().
@@ -544,8 +547,8 @@ n_awaiting_rel(Rec) ->
 
 %%
 
--spec iterate(iter(K, V)) -> {K, V, iter(K, V)} | none.
-iterate(It0) ->
+-spec iter_next(iter(K, V)) -> {K, V, iter(K, V)} | none.
+iter_next(It0) ->
     gen_iter_next(It0).
 
 %%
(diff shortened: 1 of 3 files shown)


apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
```diff
@@ -534,6 +544,12 @@ n_awaiting_rel(Rec) ->
 
 %%
 
+-spec iterate(iter(K, V)) -> {K, V, iter(K, V)} | none.
+iterate(It0) ->
```
thalesmg, line 548: Tiny nit: maybe:

```suggestion
next(It0) ->
```

?

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
```diff
@@ -670,13 +694,110 @@ pmap_commit(
...
+            gbt_fold_iter(Fun, Fun(K, V, Acc), It);
+        _ ->
+            Acc
+    end.
+
+gbt_size(Cache) ->
+    gb_trees:size(Cache).
+
+gbt_iter_after(After, Cache) ->
+    It0 = gb_trees:iterator_from(After, Cache),
```
thalesmg, line 790: Q: does `beginning` happen to work due to the Erlang term order and due to the particular keys we use here?  Maybe we should have an explicit clause for `beginning`?

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -87,23 +111,41 @@ find_new_streams(S) ->
...
+            case Filter(Key, Stream) of
+                true ->
+                    {Key, Stream, ItStream};
+                false ->
+                    next_stream(ItStream)
+            end;
+        none ->
+            %% Restart the iteration from the beginning:
+            ItStream = ItStream0#iter{it = ItCont},
+            next_stream(ItStream)
```
thalesmg, line 142: Shouldn't we decrement `limit` before recursing here?

ieQu1 (approved):

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
ieQu1: Yes, I guess this name is better for the sake of naming consistency. We use `next` everywhere else.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
keynslug: Yes, the term order helps. I will leave a comment here to describe that we rely on it here.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
keynslug: Good point. We haven't gone through one more element, thus no decrement. The situation you're describing can happen in general, but should not here: `it_cont` returning `none` right away means that `limit` is `0`. But it still sounds like a good idea to defend from that honestly.

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3412)

PR #13398, opened 2024-07-03
fix: don't destory when authz'source unchanged

zhongwencool:
Fixes https://emqx.atlassian.net/browse/EMQX-12630

https://github.com/emqx/emqx/discussions/13385#discussioncomment-9941273

Release version: v/e5.7.2

## PR Checklist

### DIFF
--- a/apps/emqx_auth/src/emqx_authz/emqx_authz.erl
+++ b/apps/emqx_auth/src/emqx_authz/emqx_authz.erl
@@ -317,18 +317,32 @@ do_post_config_update(?ROOT_KEY, _Conf, NewConf) ->
 
 overwrite_entire_sources(Sources) ->
     PrevSources = lookup(),
-    NewSourcesTypes = lists:map(fun type/1, Sources),
-    EnsureDelete = fun(S) ->
-        TypeName = type(S),
-        Opts =
-            case lists:member(TypeName, NewSourcesTypes) of
-                true -> #{clear_metric => false};
-                false -> #{clear_metric => true}
-            end,
-        ensure_deleted(S, Opts)
-    end,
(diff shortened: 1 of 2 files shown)


apps/emqx_auth/src/emqx_authz/emqx_authz.erl
```diff
@@ -317,18 +317,32 @@ do_post_config_update(?ROOT_KEY, _Conf, NewConf) ->
...
-    lists:foreach(EnsureDelete, PrevSources),
-    create_sources(Sources).
+    #{
+        removed := Removed,
+        added := Added,
+        identical := Identical,
+        changed := Changed
+    } = emqx_utils:diff_lists(Sources, PrevSources, fun type/1),
+    lists:foreach(
+        fun(S) -> ensure_deleted(S, #{clear_metric => true}) end,
```
zmstone, line 327: I find it unreasonable to delete the data even if we just want to disable then enable it back.

zmstone:
we should perhaps never deleted data even if built-in authz is removed (same for authn).
For data purge, can expose an API/CLI to do so.

zmstone:
> we should perhaps never deleted data even if built-in authz is removed (same for authn). For data purge, can expose an API/CLI to do so.

Discussed with Guowei, in 5.7.2, we can just fix it to survive enable/disable.
The new API/CLI can be added in a minor, 5.8 or 5.9

apps/emqx_auth/src/emqx_authz/emqx_authz.erl
zhongwencool: this is only for removed source. if we just disable / enable source, it will in changed sources. not removed.

zmstone (approved):

PR #13451, opened 2024-07-11
fix: do not convert a empty file name to a empty list

lafirest:
Fixes <issue-or-jira-number>

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -3355,6 +3355,8 @@ default_listener(SSLListener) ->
 %% otherwise always return string.
 naive_env_interpolation(undefined) ->
     undefined;
+naive_env_interpolation(<<>>) ->
+    <<>>;
 naive_env_interpolation(Bin) when is_binary(Bin) ->
     naive_env_interpolation(unicode:characters_to_list(Bin, utf8));
 naive_env_interpolation("$" ++ Maybe = Original) ->


id (approved):

apps/emqx/src/emqx_schema.erl
```diff
@@ -3355,6 +3355,8 @@ default_listener(SSLListener) ->
 %% otherwise always return string.
 naive_env_interpolation(undefined) ->
     undefined;
+naive_env_interpolation(<<>>) ->
+    <<>>;
```
zmstone, line 3359: undefined ?

zhongwencool:
The naive_env_interpolation function is returning a binary result, which seems a bit unusual or unexpected.
My suggestion:
1. make this field not required and no default value
```
fields("banned") ->
    [
        {bootstrap_file,
            sc(
                binary(),
                #{ required => false,
                    desc => ?DESC("banned_bootstrap_file")
                }
            )}
    ].
```
then use with:
```
case emqx_schema:naive_env_interpolation(emqx:get_config([banned, bootstrap_file], undefined)) of
undefine -> ignore;
File -> ...
end.
```

apps/emqx/src/emqx_schema.erl
lafirest: I think use empty string to indicate there is no bootstrap file is reasonable, and we also need to handle the empty string clause

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3049)

PR #13455, opened 2024-07-11
sync new mix build work to master

thalesmg:
## Summary

No changes that should affect main rebar3 build or old mix build.

## Todo

- [x] ASN.1 compiler
- [x] Run `merge-config` before release
- [ ] Check application envs in app.src and copy that.
- [ ] Copy app vsns to mix.exs files.
- [ ] Investigate if there's a way to make `ct:run_test` flush IO to console as well, not only to HTML reports.
- [x] Fix `mix ct --groups` to accept lists of group paths
- [x] Allow scoping eunit to a list of modules/tests
- [ ] Add coverdata support to `ct`, `eunit`, `proper`
- [x] Dialyzer
- [ ] Dialyzer pretty messages (should we vender dialyxir?)
- [ ] Xref
- [ ] Go through build warnings and fix them
- [x] Move `compile.grpc.ex` and maybe other compilers to separate helper app to avoid shipping their beam files.
- [ ] Check that plugin templates and emqtt keep working after we remove all rebar.config files.

## PR Checklist


apps/emqx_mix_utils/lib/mix/tasks/compile.asn1.ex
```diff
@@ -0,0 +1,82 @@
...
+    add_to_path_and_cache(:asn1)
+
+    Mix.Project.get!()
+    config = Mix.Project.config()
+    app_root = File.cwd!()
+
+    asn1_srcs = config[:asn1_srcs] || []
+    asn1_opts = config[:asn1_opts] || %{}
+    out_dir = Map.fetch!(asn1_opts, :out_dir)
+    opts = [:per, :noobj, outdir: to_charlist(out_dir)]
```
ieQu1, line 24: Compiler options are the only important part, yet they're hidden in the middle of some function. They should be at least moved to a visible place.
Also, it hardcodes PER encoding for all asn.1 files, which is too limiting. 

Can we just call `erlc` via shell? Why do we need a wrapper that essentially does the it, but in 80 lines?

apps/emqx_mix_utils/lib/mix/tasks/compile.asn1.ex
```diff
@@ -0,0 +1,82 @@
...
+  defp compile(src_path, opts, app_root, manifest_modified_time) do
+    src_path =
+      app_root
+      |> Path.join(src_path)
+      |> Path.expand()
+    if stale?(src_path, manifest_modified_time) do
+      Mix.shell().info("compiling asn1 file: #{src_path}")
+      :ok = :asn1ct.compile(to_charlist(src_path), opts)
+    else
+      Mix.shell().info("file up to date, not compiling: #{src_path}")
```
ieQu1, line 43:
```suggestion
      Mix.shell().info("file is up to date, not compiling: #{src_path}")
```

ieQu1: P.S. I am surprised that mix requires writing handling of `is_stale` logic manually. Isn't it bread and butter of build systems?

apps/emqx_mix_utils/lib/mix/tasks/compile.asn1.ex
thalesmg: Since the current solution [already uses hard-coded flags](https://github.com/emqx/emqx/blob/ffa69df6f8fc01317b986de223744fdb4cf07343/apps/emqx_durable_storage/rebar.config#L7) and command to compile, I don't think there's much difference nor need to make it generic or different.  We just use this to compile only EMQX project.  As new uses cases emerge, we can simply update the command here.

Anyways, I've moved the flags to being an option exposed in `mix.exs`, just like what's now with `rebar.config`.

apps/emqx_mix_utils/lib/mix/tasks/compile.asn1.ex
thalesmg: Thanks, fixed.

apps/emqx_mix_utils/lib/mix/tasks/compile.asn1.ex
ieQu1: > Since the current solution [already uses hard-coded flags](https://github.com/emqx/emqx/blob/ffa69df6f8fc01317b986de223744fdb4cf07343/apps/emqx_durable_storage/rebar.config#L7) and command to compile

There's a big difference: "current solution" is simply a hardcoded hook _in the application that defines a consumes the schema_. It's not a _plugin_, it's local and not generic. 

This file is named `apps/emqx_mix_utils/lib/mix/tasks/compile.asn1.ex`, i.e. it's a generic util for compiling ASN.1. It's not part of durable_storage, so if anyone else uses it and needs e.g. BER, they change the flag and it changes all encoders and in the worst case old data becomes unreadable. 

So I ask to move this compile hook back to durable_storage. We don't need to create and maintain a generic ASN.1 mix plugin.

thalesmg: The compile options have been moved to `emqx_durable_storage/mix.exs`, so they are more explicit.  

Also, again, it's as generic as our build pipeline needs it to be, since this `emqx_mix_utils` app is only used for mix utilities and not released anywhere.  Currently, it's sufficient to build our project.  If we need to expand it at a later time, we can.

ieQu1 (approved):

ieQu1:
Approved, but man, that's a lot of logic AND boilerplate in a build system. I'm not sold on Mix/Elixir :trollface:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3425)

PR #13464, opened 2024-07-12
chore: attempt to reduce race condition supervisor noproc shutdown error logs

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12442

e.g.:
```
2024-05-23T08:52:39.811845+00:00 [error] Supervisor: {local,emqx_resource_manager_sup}. Context: shutdown_error. Reason: noproc. Offender: id=<<99, 101, 110, 115, 111, 114, 101, 100>>,pid=<0.7752.1030>.
```

It could be just a race condition, as it seems to be the case for resource manager: i) a call is made to the process to stop it; ii) the call times out; iii) the after clause ends up calling supervisor:terminate_child; iv) while the supervisor is finding the child to terminate, the process actually finishes terminating, and the supervisor receives a noproc reason back.

Release version: v/e5.8

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_manager.erl
+++ b/apps/emqx_resource/src/emqx_resource_manager.erl
@@ -59,7 +59,7 @@
 ]).
 
 % Server
--export([start_link/5]).
+-export([start_link/5, where/1]).
 
 % Behaviour
 -export([init/1, callback_mode/0, handle_event/4, terminate/3]).
@@ -146,6 +146,9 @@
 %%------------------------------------------------------------------------------
 %% API
 %%------------------------------------------------------------------------------
+
+where(ResId) ->
+    gproc:where(?NAME(ResId)).
 
 %% @doc Called from emqx_resource when starting a resource instance.
 %%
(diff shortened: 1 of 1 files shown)


thalesmg:
<details>
<summary>
Script to test filter:
</summary>

```elixir
cfg0 = :logger.get_primary_config()
cfg = %{cfg0 | filters: []}
:logger.set_primary_config(cfg)

spec = fn name ->
  %{
    id: name,
    start: {B, :start_link, []},
    restart: :transient,
    shutdown: 1_000
  }
end
sup = :emqx_resource_manager_sup

## Without filter: logs line.  Ex:
## 2024-07-18T17:58:10.940498+00:00 [error] Supervisor: {local,emqx_resource_manager_sup}. Context: shutdown_error. Reason: noproc. Offender: id=<<112,49>>,pid=<0.33402.0>.

{:ok, p1} = Supervisor.start_child(sup, spec.("p1"))
Supervisor.terminate_child(sup, "p1")
Process.sleep(100)
Supervisor.delete_child(sup, "p1")

filter = fn
  %{msg: {:report, %{label: {:supervisor, :shutdown_error}, report: report}}}, on_match ->
    if :noproc == Keyword.get(report, :reason) do
      on_match
    else
      :ignore
    end

  _, _ ->
    :ignore
end
cfg = %{cfg0 | filters: [{:drop_noproc_shutdown_errors, {filter, :stop}}]}
:logger.set_primary_config(cfg)

## With the filter, shouldn't log.
{:ok, p1} = Supervisor.start_child(sup, spec.("p1"))
Supervisor.terminate_child(sup, "p1")
Process.sleep(100)
Supervisor.delete_child(sup, "p1")
```
</details>

apps/emqx/src/config/emqx_config_logger.erl
```diff
@@ -276,13 +277,28 @@ do_formatter(text, CharsLimit, SingleLine, TimeOffSet, Depth, TsFormat, WithMfa,
...
+        progress -> Filters
     end.
 
 filter_audit(_, _) -> stop.
 
+is_supervisor_noproc_shutdown_error(
+    #{msg := {report, #{label := {supervisor, shutdown_error}, report := Report}}}, OnMatch
+) when is_list(Report) ->
+    case proplists:get_value(reason, Report) of
+        noproc ->
```
keynslug, line 294: My 2 cents. Those don't feel like events that are okay to ignore, it hints that something weird is going on in the system that certainly needs attention.

thalesmg: It could be just a race condition, as it seems to be the case for resource manager: i) a call is made to the process to [stop it](https://github.com/emqx/emqx/blob/464e2027421c10009a0a2fff219bd0303e466dce/apps/emqx_resource/src/emqx_resource_manager.erl#L278-L296); ii) the call times out; iii) the `after` clause ends up calling `supervisor:terminate_child`; iv) while the supervisor is finding the child to terminate, the process actually finishes terminating, and the supervisor receives a [`noproc` reason](https://github.com/emqx/otp/blob/e6c6b1a4fffffe77fe36da1e38390b856cfa3e6e/lib/stdlib/src/supervisor.erl#L913) back.

Yet, this is just one particular case: I don't think there's a way to distinguish between a harmless race and a bug, indeed.  Maybe the user should just ignore such log noise.  :thinking:

thalesmg: Also, I believe the user in question is using a version that doesn't contain the `force_kill` patch, so any timeouts while stopping a resource might lead to such noise.

I've added a monitor to the process and now wait for the exit signal in force_kill.  This way, we shouldn't get the race condition of the process dying while the supervisor selects the child pid.

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -465,7 +468,12 @@ force_kill(ResId) ->
         undefined ->
             ok;
         Pid when is_pid(Pid) ->
+            Ref = monitor(process, Pid),
```
zmstone, line 471: we need this wait around the `remove` and `stop` calls, not just `force_kill`.
 because `force_kill` is only called in the `{error, timeout}` clause

apps/emqx_resource/src/emqx_resource_manager_sup.erl
```diff
@@ -37,9 +39,40 @@ ensure_child(ResId, Group, ResourceType, Config, Opts) ->
...
+
+ensure_dead(ResId) ->
+    maybe
+        Pid = emqx_resource_manager:where(ResId),
+        true ?= is_pid(Pid),
+        true ?= is_process_alive(Pid),
+        delete_child(ResId)
+    else
+        false ->
+            ok
```
zmstone, line 74: this is not necessary.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3444)

PR #13474, opened 2024-07-15
feat(ds): allow isolated batches with preconditions

keynslug:
Part of [EMQX-12309](https://emqx.atlassian.net/browse/EMQX-12309).

Release version: v5.8

## Summary

There are essentially 2 separate DS API changes, along with corresponding basic-and-somewhat-broken implementation on the `emqx_ds_backend_local` level.
1. _Timestamping_ (naming suggestions are welcome), a part of DB config that tells how the backend should allocate message key timestamps for messages in write batches.
2. _Preconditions_ (same here) and batch isolation, facilities to express more complex write operations, including "compare-and-swap" atomic writes.

While timestamping support in the local backend looks more or less correct, the implementation of batch isolation is currently half-broken. The idea is that isolated batches skip the write buffer. Having a concept of isolated batches in the buffer implementation will inevitably make it much more complex while having little sense (besides probably having better ordering guarantees). But this in turn breaks timestamping and preconditions, because there's no serialization point anymore. This will be fixed once we agree on the shape of the API changes.

## PR Checklist

[EMQX-12309]: https://emqx.atlassian.net/browse/EMQX-12309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_durable_storage/include/emqx_ds.hrl
+++ b/apps/emqx_durable_storage/include/emqx_ds.hrl
@@ -21,4 +21,23 @@
     preconditions = [] :: [emqx_ds:precondition()]
 }).
 
+-record(message_matcher, {
+    %% Fields identifying the message:
+    %% Client identifier
+    from :: binary(),
+    %% Topic that the message is published to
+    topic :: emqx_types:topic(),
+    %% Timestamp (Unit: millisecond)
+    timestamp :: integer(),
+
+    %% Fields the message is matched against:
+    %% Message Payload
+    payload,
+    %% Message headers
+    headers = #{} :: emqx_types:headers(),
+    %% Extra filters
+    %% Reserved for the forward compatibility purposes.
+    filters = #{}
+}).
+
 -endif.
--- a/apps/emqx_durable_storage/include/emqx_ds.hrl
+++ b/apps/emqx_durable_storage/include/emqx_ds.hrl
@@ -16,4 +16,9 @@
 -ifndef(EMQX_DS_HRL).
 -define(EMQX_DS_HRL, true).
 
+-record(dsbatch, {
+    operations :: [emqx_ds:operation()],
+    preconditions = [] :: [emqx_ds:precondition()]
+}).
+
 -endif.
--- a/apps/emqx_durable_storage/src/emqx_ds.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds.erl
@@ -169,6 +169,13 @@
 -type generic_db_opts() ::
     #{
         backend := atom(),
+        %% Force strictly monotonic message timestamps.
+        %% Default: `true'.
+        %% Messages are assigned unique, strictly monotonically increasing timestamps.
+        %% Those timestamps form a total order per each serialization key.
+        %% If `false' then message timestamps are respected; timestamp, topic and
+        %% serialization key uniquely identify a message.
+        force_monotonic_timestamps => boolean(),
         serialize_by => clientid | topic,
         _ => _
     }.
--- a/apps/emqx_durable_storage/src/emqx_ds.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds.erl
@@ -99,8 +99,8 @@
 
 -type message() :: emqx_types:message().
 
-%% Message pattern.
--type message(Pattern) :: emqx_types:message(Pattern).
+%% Message matcher.
+-type message_matcher(Payload) :: #message_matcher{payload :: Payload}.
 
 %% A batch of storage operations.
 -type batch() :: [operation()] | #dsbatch{}.
@@ -110,7 +110,7 @@
     message()
     %% Delete a message.
     %% Does nothing if the message does not exist.
-    | {delete, message(_)}.
+    | {delete, message_matcher('_')}.
 
 %% Precondition.
 %% Fails whole batch if the storage already has the matching message (`if_exists'),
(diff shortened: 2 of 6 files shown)


apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -163,12 +184,20 @@
         %% shard (if applicable to the backend), as the batch will be split accordingly
         %% even if this flag is `true'.
         %% Default: `false'.
-        atomic => boolean()
+        atomic => boolean() | isolate
```
ieQu1, line 187: What is semantic of `atomic=>isolate` and how is it different from `atomic => true`?

keynslug: In essence,
* `atomic => true` means that either whole set of operations is applied or none at all,
* `atomic => isolate` means the same + that set of operations applied in isolation from any other set, basically to be sure that any failing precondition will affect only this set of operations.

Admittedly, backends can decide on their own that some level of isolation is needed if there are preconditions in a batch, so this explicitness is not necessary.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -94,6 +93,28 @@
...
+    %% Delete a message.
+    %% Does nothing if the message does not exist.
+    | {delete, message(_)}.
+
+%% Precondition.
+%% Fails whole batch if the message in question does not match (`while'), or
+%% does match (`unless'). Useful to construct batches with "compare-and-set"
+%% semantics.
+-type precondition() ::
+    {while | unless, message(iodata() | '_')}.
```
savonarola, line 116: Nit: the semantics of `while` is not very easy to grasp 😅

savonarola (dismissed):

apps/emqx_durable_storage/src/emqx_ds.erl
ieQu1: I think `atomic => true` should do what `atomic => isolate` does.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -94,6 +97,32 @@
...
+
+%% A batch of storage operations.
+-type batch() :: [operation()] | #dsbatch{}.
+
+-type operation() ::
+    %% Store a message.
+    message()
+    %% Delete a message.
+    %% Does nothing if the message does not exist.
+    | {delete, message(_)}.
```
ieQu1, line 113: Since we decided to split out preconditions, why not do the same to deletes?

apps/emqx_utils/include/emqx_message.hrl
```diff
@@ -33,7 +33,7 @@
     %% Topic that the message is published to
     topic :: emqx_types:topic(),
     %% Message Payload
-    payload :: emqx_types:payload(),
+    payload,
```
ieQu1, line 36: Is it possible to avoid this change? I'd prefer to relax the type in `emqx_ds` than to risk introducing a missed typing issue anywhere in emqx code.

Proper solution could look like this: remove all types from the record, and define `emqx_types:message()` as `#message{id :: .., payload :: .., ...}`.

Then define another type smth. like `message_match_pattern` that allows '_'.

apps/emqx_durable_storage/src/emqx_ds.erl
keynslug: ...And supposedly should be part of storage-layer-level batches.

apps/emqx_utils/include/emqx_message.hrl
ieQu1: 1) At least it would be nice to leave this type commented out with an explanation. 
2) But maybe it's worth to add another record just for precondition matching, say, `#message_matcher{}` with the same fields.

apps/emqx_durable_storage/src/emqx_ds.erl
ieQu1: Ok.

ieQu1: Another question: why is it `{delete, #message{}}` and not `{delete, Key}`? Deletion by key is trivial for DS, deletion by message requires non-trivial matching.

ieQu1: Nvm, answered my own question.

apps/emqx_durable_storage/include/emqx_ds.hrl
```diff
@@ -16,4 +16,28 @@
...
+    %% Topic that the message is published to
+    topic :: emqx_types:topic(),
+    %% Timestamp (Unit: millisecond)
+    timestamp :: integer(),
+
+    %% Fields the message is matched against:
+    %% Message Payload
+    payload,
+    %% Message headers
+    headers = #{} :: emqx_types:headers(),
```
ieQu1, line 37: Nit: if we don't support matching of certain fields (like headers), maybe they should not be here to avoid making a false promise to the API consumer.

apps/emqx_durable_storage/include/emqx_ds.hrl
```diff
@@ -16,4 +16,28 @@
...
 
+-record(dsbatch, {
+    operations :: [emqx_ds:operation()],
+    preconditions = [] :: [emqx_ds:precondition()]
+}).
+
+-record(message_matcher, {
+    %% Fields identifying the message:
+    %% Client identifier
+    from :: binary(),
```
ieQu1, line 27: Nit:
```suggestion
    from :: binary() | '_',
```

apps/emqx_durable_storage/include/emqx_ds.hrl
keynslug: True. But this one is something that could be demanded in the near future and trivial to implement (e.g. with _subset of_ semantics).

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3436)

PR #13487, opened 2024-07-17
feat(cluster link): refactor http api, add status and metrics

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12627

Release version: e5.8

## PR Checklist

### DIFF
--- a/apps/emqx_cluster_link/src/emqx_cluster_link.erl
+++ b/apps/emqx_cluster_link/src/emqx_cluster_link.erl
@@ -26,28 +26,10 @@
     on_message_publish/1
 ]).
 
-%% metrics API
--export([
-    maybe_create_metrics/1,
-    drop_metrics/1,
-
-    get_metrics/1,
-    routes_inc/2
-]).
-
 -include("emqx_cluster_link.hrl").
 -include_lib("emqx/include/emqx.hrl").
 -include_lib("emqx/include/emqx_hooks.hrl").
 -include_lib("emqx/include/logger.hrl").
-
-%%--------------------------------------------------------------------
-%% Type definitions
-%%--------------------------------------------------------------------
-
--define(METRICS, [
-    ?route_metric
-]).
--define(RATE_METRICS, []).
 
 %%--------------------------------------------------------------------
 %% emqx_external_broker API
@@ -149,32 +131,6 @@ put_hook() ->
 
 delete_hook() ->
     emqx_hooks:del('message.publish', {?MODULE, on_message_publish, []}).
-
-%%--------------------------------------------------------------------
-%% metrics API
-%%--------------------------------------------------------------------
-
-get_metrics(ClusterName) ->
-    Nodes = emqx:running_nodes(),
-    Timeout = 15_000,
-    Results = emqx_metrics_proto_v2:get_metrics(Nodes, ?METRIC_NAME, ClusterName, Timeout),
-    sequence_multicall_results(Nodes, Results).
-
-maybe_create_metrics(ClusterName) ->
-    case emqx_metrics_worker:has_metrics(?METRIC_NAME, ClusterName) of
-        true ->
-            ok = emqx_metrics_worker:reset_metrics(?METRIC_NAME, ClusterName);
-        false ->
-            ok = emqx_metrics_worker:create_metrics(
-                ?METRIC_NAME, ClusterName, ?METRICS, ?RATE_METRICS
-            )
-    end.
-
-drop_metrics(ClusterName) ->
-    ok = emqx_metrics_worker:clear_metrics(?METRIC_NAME, ClusterName).
-
-routes_inc(ClusterName, Val) ->
-    catch emqx_metrics_worker:inc(?METRIC_NAME, ClusterName, ?route_metric, Val).
 
 %%--------------------------------------------------------------------
 %% Internal functions
@@ -297,16 +253,3 @@ maybe_filter_incomming_msg(#message{topic = T} = Msg, ClusterName) ->
         true -> with_sender_name(Msg, ClusterName);
         false -> []
     end.
-
--spec sequence_multicall_results([node()], emqx_rpc:erpc_multicall(term())) ->
-    {ok, [{node(), term()}]} | {error, [term()]}.
-sequence_multicall_results(Nodes, Results) ->
-    case lists:partition(fun is_ok/1, lists:zip(Nodes, Results)) of
-        {OkResults, []} ->
-            {ok, [{Node, Res} || {Node, {ok, Res}} <- OkResults]};
-        {_OkResults, BadResults} ->
-            {error, BadResults}
-    end.
-
-is_ok({_Node, {ok, _}}) -> true;
-is_ok(_) -> false.
--- a/apps/emqx_cluster_link/src/emqx_cluster_link.erl
+++ b/apps/emqx_cluster_link/src/emqx_cluster_link.erl
@@ -26,10 +26,28 @@
     on_message_publish/1
 ]).
 
+%% metrics API
+-export([
+    maybe_create_metrics/1,
+    drop_metrics/1,
+
+    get_metrics/1,
+    routes_inc/2
+]).
+
 -include("emqx_cluster_link.hrl").
 -include_lib("emqx/include/emqx.hrl").
 -include_lib("emqx/include/emqx_hooks.hrl").
 -include_lib("emqx/include/logger.hrl").
+
+%%--------------------------------------------------------------------
+%% Type definitions
+%%--------------------------------------------------------------------
+
+-define(METRICS, [
+    ?route_metric
+]).
+-define(RATE_METRICS, []).
 
 %%--------------------------------------------------------------------
 %% emqx_external_broker API
@@ -131,6 +149,32 @@ put_hook() ->
 
 delete_hook() ->
     emqx_hooks:del('message.publish', {?MODULE, on_message_publish, []}).
+
+%%--------------------------------------------------------------------
+%% metrics API
+%%--------------------------------------------------------------------
+
+get_metrics(ClusterName) ->
+    Nodes = emqx:running_nodes(),
+    Timeout = 15_000,
+    Results = emqx_metrics_proto_v2:get_metrics(Nodes, ?METRIC_NAME, ClusterName, Timeout),
+    sequence_multicall_results(Nodes, Results).
+
+maybe_create_metrics(ClusterName) ->
+    case emqx_metrics_worker:has_metrics(?METRIC_NAME, ClusterName) of
+        true ->
+            ok = emqx_metrics_worker:reset_metrics(?METRIC_NAME, ClusterName);
+        false ->
+            ok = emqx_metrics_worker:create_metrics(
+                ?METRIC_NAME, ClusterName, ?METRICS, ?RATE_METRICS
+            )
+    end.
+
+drop_metrics(ClusterName) ->
+    ok = emqx_metrics_worker:clear_metrics(?METRIC_NAME, ClusterName).
+
+routes_inc(ClusterName, Val) ->
+    catch emqx_metrics_worker:inc(?METRIC_NAME, ClusterName, ?route_metric, Val).
 
 %%--------------------------------------------------------------------
 %% Internal functions
@@ -253,3 +297,16 @@ maybe_filter_incomming_msg(#message{topic = T} = Msg, ClusterName) ->
         true -> with_sender_name(Msg, ClusterName);
         false -> []
     end.
+
+-spec sequence_multicall_results([node()], emqx_rpc:erpc_multicall(term())) ->
+    {ok, [{node(), term()}]} | {error, [term()]}.
+sequence_multicall_results(Nodes, Results) ->
+    case lists:partition(fun is_ok/1, lists:zip(Nodes, Results)) of
+        {OkResults, []} ->
+            {ok, [{Node, Res} || {Node, {ok, Res}} <- OkResults]};
+        {_OkResults, BadResults} ->
+            {error, BadResults}
+    end.
+
+is_ok({_Node, {ok, _}}) -> true;
+is_ok(_) -> false.
--- a/apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
+++ b/apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
@@ -22,8 +22,8 @@
 
 -export([
     '/cluster/links'/2,
-    '/cluster/links/:name'/2,
-    '/cluster/links/:name/metrics'/2
+    '/cluster/links/link/:name'/2,
+    '/cluster/links/link/:name/metrics'/2
 ]).
 
 -define(CONF_PATH, [cluster, links]).
@@ -39,8 +39,8 @@ api_spec() ->
 paths() ->
     [
         "/cluster/links",
-        "/cluster/links/:name",
-        "/cluster/links/:name/metrics"
+        "/cluster/links/link/:name",
+        "/cluster/links/link/:name/metrics"
     ].
 
 schema("/cluster/links") ->
@@ -69,9 +69,9 @@ schema("/cluster/links") ->
                     }
             }
     };
-schema("/cluster/links/:name") ->
+schema("/cluster/links/link/:name") ->
     #{
-        'operationId' => '/cluster/links/:name',
+        'operationId' => '/cluster/links/link/:name',
         get =>
             #{
                 description => "Get a cluster link configuration",
@@ -117,9 +117,9 @@ schema("/cluster/links/:name") ->
                     }
             }
     };
-schema("/cluster/links/:name/metrics") ->
+schema("/cluster/links/link/:name/metrics") ->
     #{
-        'operationId' => '/cluster/links/:name/metrics',
+        'operationId' => '/cluster/links/link/:name/metrics',
         get =>
             #{
                 description => "Get a cluster link metrics",
@@ -173,11 +173,11 @@ fields(node_metrics) ->
         fun() -> handle_create(Name, Body) end
     ).
 
-'/cluster/links/:name'(get, #{bindings := #{name := Name}}) ->
+'/cluster/links/link/:name'(get, #{bindings := #{name := Name}}) ->
     with_link(Name, fun(Link) -> handle_lookup(Name, Link) end, not_found());
-'/cluster/links/:name'(put, #{bindings := #{name := Name}, body := Params0}) ->
+'/cluster/links/link/:name'(put, #{bindings := #{name := Name}, body := Params0}) ->
     with_link(Name, fun() -> handle_update(Name, Params0) end, not_found());
-'/cluster/links/:name'(delete, #{bindings := #{name := Name}}) ->
+'/cluster/links/link/:name'(delete, #{bindings := #{name := Name}}) ->
     with_link(
         Name,
         fun() ->
@@ -192,7 +192,7 @@ fields(node_metrics) ->
         not_found()
     ).
 
-'/cluster/links/:name/metrics'(get, #{bindings := #{name := Name}}) ->
+'/cluster/links/link/:name/metrics'(get, #{bindings := #{name := Name}}) ->
     with_link(Name, fun() -> handle_metrics(Name) end, not_found()).
 
 %%--------------------------------------------------------------------
--- a/apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
+++ b/apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
@@ -7,6 +7,7 @@
 
 -include_lib("hocon/include/hoconsc.hrl").
 -include_lib("emqx/include/http_api.hrl").
+-include_lib("emqx_utils/include/emqx_utils_api.hrl").
 
 -export([
     api_spec/0,
@@ -14,7 +15,10 @@
     schema/1
 ]).
 
--export([config/2]).
+-export([
+    '/cluster/links'/2,
+    '/cluster/links/:name'/2
+]).
 
 -define(CONF_PATH, [cluster, links]).
 -define(TAGS, [<<"Cluster">>]).
@@ -24,12 +28,13 @@ api_spec() ->
 
 paths() ->
     [
-        "/cluster/links"
+        "/cluster/links",
+        "/cluster/links/:name"
     ].
 
 schema("/cluster/links") ->
     #{
-        'operationId' => config,
+        'operationId' => '/cluster/links',
         get =>
             #{
                 description => "Get cluster links configuration",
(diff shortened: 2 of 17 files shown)


thalesmg:
Todo: maybe we should also check the health of the actor and combine it with that of the forwarder.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
```diff
@@ -7,86 +7,454 @@
...
+            AggregatedMetrics = aggregate_metrics(NodeMetrics),
+            Response = #{metrics => AggregatedMetrics, node_metrics => NodeMetrics},
+            ?OK(Response)
+    end.
+
+aggregate_metrics(NodeMetrics) ->
+    ErrorLogger = fun(_) -> ok end,
+    lists:foldl(
+        fun(#{metrics := Metrics}, Acc) ->
+            emqx_utils_maps:best_effort_recursive_sum(Metrics, Acc, ErrorLogger)
```
keynslug, line 254: I'd say we'll need to explain what those metrics mean.

Currently, the `routes` counter is tracked in _extrouter_, and thus means "number of routes pushed to me by `ClusterName`, aggregated by actor". Because different actors may push identical routes, this number will most of the time be larger than "number of routes `ClusterName` is interested in". I guess the last one is harder to measure because of the initial design goal of having leaderless route replication. To be clear it's still possible to measure through a (potentially expensive) full-scan of the extrouter table.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
```diff
@@ -7,86 +7,454 @@
...
+        end,
+        #{},
+        NodeMetrics
+    ).
+
+format_metrics(Node, Metrics) ->
+    Routes = emqx_utils_maps:deep_get([counters, ?route_metric], Metrics, 0),
+    #{
+        node => Node,
+        metrics => #{
```
keynslug, line 264: I think we could as well expose resource metrics of the message forwarding resource?

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
```diff
@@ -7,86 +7,454 @@
...
+            ?route_metric => Routes
+        }
+    }.
+
+add_status(Name, Link) ->
```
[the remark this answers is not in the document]
thalesmg: No, it indeed [fetches and adds the status](https://github.com/thalesmg/emqx/blob/d9832252d8f1cafe30796fe555a4da088421359a/apps/emqx_cluster_link/src/emqx_cluster_link_api.erl#L270-L272) to a given link map.

apps/emqx_cluster_link/src/emqx_cluster_link_config.erl
```diff
@@ -28,11 +30,15 @@
 
 -export([
     %% General
+    create/1,
+    delete/1,
+    update_one_link/1,
```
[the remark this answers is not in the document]
thalesmg: Well, it's already under the `cluster_link` namespace...  But I'll add the suffix then.

apps/emqx_cluster_link/src/emqx_cluster_link_config.erl
```diff
@@ -320,3 +414,11 @@ do_convert_certs(LinkName, SSLOpts) ->
             ),
             throw({bad_ssl_config, Reason})
     end.
+
+safe_take(Name, Transformations) ->
+    case lists:splitwith(fun(#{<<"name">> := N}) -> N =/= Name end, Transformations) of
+        {_Front, []} ->
+            not_found;
+        {Front, [Found | Rear]} ->
+            {ok, {Found, Front, Rear}}
```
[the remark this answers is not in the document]
thalesmg: The idea was to be more idiomatic with the `{ok, _} | error` idiom...  Though, I'd need to use `error` to be more adherent, indeed.  :wink:

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
```diff
@@ -114,3 +482,41 @@ links_config_example() ->
...
+    case emqx_cluster_link_config:link_raw(Name) of
+        undefined ->
+            NotFoundFn();
+        Link0 = #{} ->
+            Link = fill_defaults_single(Link0),
+            {arity, Arity} = erlang:fun_info(FoundFn, arity),
+            case Arity of
+                1 -> FoundFn(Link);
+                0 -> FoundFn()
+            end
```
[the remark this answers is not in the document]
thalesmg: :checkered_flag:

apps/emqx_cluster_link/src/emqx_cluster_link_config.erl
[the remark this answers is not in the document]
thalesmg: `update_one_link` was already there before the PR.

:checkered_flag:

apps/emqx_cluster_link/src/emqx_cluster_link_config.erl
```diff
@@ -133,6 +194,37 @@ remove_handler() ->
...
+        NewRawConf = convert_certs(maybe_increment_ps_actor_incr(NewRawConf0, OldRawConf)),
+        {ok, NewRawConf}
+    else
+        _ ->
+            {error, already_exists}
+    end;
+pre_config_update(?LINKS_PATH, {update, LinkRawConf}, OldRawConf) ->
+    #{<<"name">> := Name} = LinkRawConf,
+    maybe
+        {ok, {_Found, Front, Rear}} = safe_take(Name, OldRawConf),
```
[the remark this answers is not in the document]
thalesmg: :see_no_evil:

apps/emqx_cluster_link/src/emqx_cluster_link.erl
```diff
@@ -26,11 +26,29 @@
...
 
+%% metrics API
+-export([
+    maybe_create_metrics/1,
+    drop_metrics/1,
+
+    get_metrics/1,
+    routes_inc/2
+]).
+
```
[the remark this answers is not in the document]
thalesmg: :checkered_flag:

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
```diff
@@ -7,86 +7,454 @@
...
+    Links.
+
+get_all_link_status_cluster() ->
+    case emqx_cluster_link_mqtt:get_all_resources_cluster() of
+        {error, BadResults} ->
+            ?SLOG(warning, #{
+                msg => "cluster_link_api_all_status_bad_erpc_results",
+                results => BadResults
+            }),
+            [];
```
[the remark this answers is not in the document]
thalesmg: :checkered_flag:

apps/emqx_cluster_link/src/emqx_cluster_link.erl
```diff
@@ -253,3 +297,16 @@ maybe_filter_incomming_msg(#message{topic = T} = Msg, ClusterName) ->
...
+sequence_multicall_results(Nodes, Results) ->
+    case lists:partition(fun is_ok/1, lists:zip(Nodes, Results)) of
+        {OkResults, []} ->
+            {ok, [{Node, Res} || {Node, {ok, Res}} <- OkResults]};
+        {_OkResults, BadResults} ->
+            {error, BadResults}
+    end.
+
+is_ok({_Node, {ok, _}}) -> true;
+is_ok(_) -> false.
```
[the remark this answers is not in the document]
thalesmg: :checkered_flag:

apps/emqx_cluster_link/src/emqx_cluster_link_config.erl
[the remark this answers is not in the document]
thalesmg: :check

apps/emqx_cluster_link/src/emqx_cluster_link_config.erl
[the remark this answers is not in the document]
thalesmg: :checkered_flag:

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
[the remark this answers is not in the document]
keynslug: Oh sorry, I get the idea now. Disregard the comment.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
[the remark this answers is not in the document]
keynslug: RIght, I see now.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
thalesmg: > Because different actors may push identical routes, this number will most of the time be larger than "number of routes ClusterName is interested in

I thought duplications were handled in this code: https://github.com/emqx/emqx/blob/311419f6212b2a69233514d6366b70e273c8f09e/apps/emqx_cluster_link/src/emqx_cluster_link_extrouter.erl#L281-L299

Could there still be duplicates even with the check above, then?

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
```diff
@@ -296,9 +311,9 @@ get_all_link_status_cluster() ->
                 msg => "cluster_link_api_all_status_bad_erpc_results",
                 results => BadResults
             }),
-            [];
+            ?INTERNAL_ERROR(<<"Failed to fetch cluster link status, please check the logs.">>);
```
keynslug, line 314: I'll take my suggestion back, sorry for the confusion. 🫠 I haven't pieced together the fact that it's just the resource status that gets different value if any of the RPCs have failed. Please feel free to revert that change.

My general question is basically: why log a warning if we can carry the error information with API responses, where it could be more visible and potentially more helpful? In this case, in the `node_status` field. What do you think about that?

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
keynslug: Oh wait. I misread that change initially, I think it _might be_ incorrect actually. Hope to be wrong, please help to verify.

```
    case MCounter band Marker of
        0 when OpName =:= add ->
            Res = mria:dirty_update_counter(?EXTROUTE_TAB, Entry, Marker),
            _ = emqx_cluster_link:routes_inc(ClusterName, 1),
```

^ This will bump counter each time an actor pushes a route not yet present in the extrouter table.
Two different actor will have different lanes, which means different `Marker` values, so them pushing the same route would cause 2 counter bumps.

```
            case mria:dirty_update_counter(?EXTROUTE_TAB, Entry, -Marker) of
                0 ->
                    Record = #extroute{entry = Entry, mcounter = 0},
                    ok = mria:dirty_delete_object(?EXTROUTE_TAB, Record),
                    _ = emqx_cluster_link:routes_inc(ClusterName, -1),
```

^ This deletion is actually could happen only when there are no more occupied lanes, which means that every actor had this route deleted. Moreover, this deletion might not happen at all, if there was concurrent `dirty_update_counter(+Marker)` in the meantime.

So those counter bumps are not actually symmetrical.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
thalesmg: Good idea, I'll move the errors back to the API response.

thalesmg: Remembered another reason: some terms might encode to garbled json values...

thalesmg: Yes, that's what I used in the latest commits to overcome that.  But I think only recently it has started to be used, I don't spot it being used often in some APIs.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
thalesmg: :checkered_flag:

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
keynslug: I see, I was trying to say that it sounds to me like this is a problem to solve regardless of where error details go: in API responses or in the logs.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
thalesmg: I see, thanks for the explanation.  

Though, if there's only one actor per cluster link, it should still be safe to bump the counters, right?  Since the counter is scoped per cluster link name and `Marker`s are unique per actor?

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
thalesmg: I tried to push those reasons to the status and metrics API responses in the latest commits.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
keynslug: Number of actors per link is generally more than one, there are usually 1 or 2 actors per node. The idea is that each node replicates only its own route updates.

thalesmg: Oh, true, there could be 2 actors (one for DS, the other for non-DS routes).  :see_no_evil: 

It sounds like the safest option then is to have a periodic full table scan that updates those metrics as gauges, then.

keynslug: Yeah. 🫠 I remember we had discussion with SergeTupchiy about the feasibility of having an extrouter table per cluster link, that would have probably helped.

I guess if we can tolerate small errors under high concurrency (in the "10 actors push the same route at the exact same moment to different target cluster nodes" sense), we can still rely on the result of `dirty_update_counter/3` and bump the counter if this was the first update, i.e. mirroring the current counter decrease logic in https://github.com/emqx/emqx/pull/13487/files/d9832252d8f1cafe30796fe555a4da088421359a#diff-2fe6367be77c076ca17ca1254e75b4a65a0de905a1e6af8badd52a347fe17455R296. I suspect that under really really high concurrency we risk bumping the counter several times on `add` operations and not decreasing it at all on `del` operations, but this assumption probably needs a double-check.

thalesmg: One table per link would make it much simpler, indeed.  :smiley_cat: 

In my last commit, I added a bookkeeper process similar to what we have for persistent sessions.  WDYT?

apps/emqx_cluster_link/src/emqx_cluster_link_bookkeeper.erl
```diff
@@ -0,0 +1,84 @@
...
+    ClusterNames = cluster_names(),
+    tally_routes(ClusterNames),
+    ensure_timer(#tally_routes{}, emqx_cluster_link_config:tally_routes_interval()),
+    ok.
+
+tally_routes([ClusterName | ClusterNames]) ->
+    Tab = emqx_cluster_link_extrouter:extroute_tab(),
+    Pat = emqx_cluster_link_extrouter:cluster_routes_ms(ClusterName),
+    NumRoutes = ets:select_count(Tab, Pat),
+    emqx_cluster_link_metrics:routes_set(ClusterName, NumRoutes),
```
keynslug, line 81: Just basically to kick off a discussion: having so "full-scan-intensive" job just to have lagging but accurate metrics feels a bit wasteful to me. I wonder if we can have a hybrid approach, where (as before) a counter tracks per-node route additions and deletions (accepting the risk of slight drift under high concurrency) + this bookkeeper that tracks cluster-wide gauges, but it runs much more rarely and resets the per-node counters on each run. Then `NRoutes(ClusterName) = Gauge(ClusterName) + Sum(Counter(Node, ClusterName) || N <- Nodes)`. Bonus point is almost no lagging, downside is lack of 100% accuracy.

I'm not sure though how it fits into the current metrics models, please share any thoughts. Maybe I'm wrong about this perceived wastefulness?

keynslug (commented):
Solid work. 🔥

apps/emqx_cluster_link/src/emqx_cluster_link_bookkeeper.erl
keynslug: Ah wait, it won't work as I've imagined, counter resets will happen at different times on different nodes, which will deform the aggregation results.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
```diff
@@ -270,12 +279,27 @@ aggregate_metrics(NodeMetrics) ->
...
+            'dropped' => Get([counters, 'dropped'], ResourceMetrics),
+            'retried' => Get([counters, 'retried'], ResourceMetrics),
+            'received' => Get([counters, 'received'], ResourceMetrics),
+
+            'queuing' => Get([gauges, 'queuing'], ResourceMetrics),
+            'inflight' => Get([gauges, 'inflight'], ResourceMetrics),
+
+            'rate' => Get([rate, 'matched', current], ResourceMetrics),
+            'rate_last5m' => Get([rate, 'matched', last5m], ResourceMetrics),
+            'rate_max' => Get([rate, 'matched', max], ResourceMetrics)
```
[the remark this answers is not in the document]
thalesmg: I thought of that initially, but I thought of following the current convention (from other metric APIs) of flattening stuff.
We can namespace it and, if frontend or QA/product complains later, we can just flatten it back.

apps/emqx_cluster_link/src/emqx_cluster_link_schema.erl
```diff
@@ -30,7 +30,18 @@ namespace() -> "cluster".
...
+    #{
+        cluster => [
+            {links, links_schema(#{})},
+            {tally_routes_interval,
+                hoconsc:mk(
+                    emqx_schema:timeout_duration(), #{
+                        default => <<"15s">>,
+                        importance => ?IMPORTANCE_HIDDEN
+                    }
+                )}
```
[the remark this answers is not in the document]
thalesmg: That way will be harder to have a higher frequency for tests.  :thinking:

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
```diff
@@ -237,10 +237,19 @@ handle_metrics(Name) ->
...
+                ({Node, {ok, RouterMetrics}, ResError}, {OkAccIn, ErrAccIn}) ->
+                    OkAcc = [format_metrics(Node, RouterMetrics, _ResourceMetrics = #{}) | OkAccIn],
+                    {OkAcc, [{Node, #{resource => ResError}} | ErrAccIn]};
+                ({Node, RouterError, {ok, ResourceMetrics}}, {OkAccIn, ErrAccIn}) ->
+                    OkAcc = [format_metrics(Node, _RouterMetrics = #{}, ResourceMetrics) | OkAccIn],
+                    {OkAcc, [{Node, #{router => RouterError}} | ErrAccIn]};
+                ({Node, RouterError, ResourceError}, {OkAccIn, ErrAccIn}) ->
+                    {OkAccIn, [
+                        {Node, #{router => RouterError, resource => ResourceError}} | ErrAccIn
+                    ]}
```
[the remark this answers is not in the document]
thalesmg: Not sure what you mean by "2 separate list comprehensions :melting_face: ", but I'll refactor this.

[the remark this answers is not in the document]
keynslug: Right, my bad, it's a bit trickier than that. Thanks!

apps/emqx_cluster_link/src/emqx_cluster_link_schema.erl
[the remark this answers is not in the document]
thalesmg: Fair enough, let's go with that.

apps/emqx_cluster_link/src/emqx_cluster_link_api.erl
```diff
@@ -281,7 +281,7 @@ aggregate_metrics(NodeMetrics) ->
 
 format_metrics(Node, RouterMetrics, ResourceMetrics) ->
     Get = fun(Path, Map) -> emqx_utils_maps:deep_get(Path, Map, 0) end,
-    Routes = Get([counters, ?route_metric], RouterMetrics),
+    Routes = Get([gauges, ?route_metric], RouterMetrics),
```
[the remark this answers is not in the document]
thalesmg: Ah, true, the routes are cluster-global, not per node.  :see_no_evil: 
Thanks for spotting that!

apps/emqx_cluster_link/src/emqx_cluster_link_bookkeeper.erl
thalesmg: Maybe that could work.  Each node would keep track of its own counters and have those fast but potentially inaccurate counters that would be reset by the bookkeeper.  The difference though is that we should have `NRoutes(ClusterName) = Gauge(ClusterName) + Counter(ClusterName)` done on each node, and then we can take the maximum value when aggregating.  Again, there's room for lagging metrics or a bit of inaccuracy, but sounds reasonable.

The problem I can think of now is what you mentioned before about having potentially more than 1 actor inserting the same route...  Not sure how we could easily deduplicate that.  :thinking:

keynslug: > The problem I can think of now is what you mentioned before about having potentially more than 1 actor inserting the same route... Not sure how we could easily deduplicate that. 🤔

Yeah I meant that we should essentially mirror the (former) delete logic, i.e.:
```
    case MCounter band Marker of
        0 when OpName =:= add ->
            case mria:dirty_update_counter(?EXTROUTE_TAB, Entry, Marker) of
                Marker -> %% (likely) we're first, bump metric.
                _More -> %% not the first actor to push that route.
            end;
        Marker when OpName =:= add ->
            %% Already added.
            MCounter;
        Marker when OpName =:= delete ->
            case mria:dirty_update_counter(?EXTROUTE_TAB, Entry, -Marker) of
                0 ->
                    %% (likely) the last actor to delete that route, decrease the metric
                    Record = #extroute{entry = Entry, mcounter = 0},
                    ok = mria:dirty_delete_object(?EXTROUTE_TAB, Record),
                    0;
                C ->
                    %% there are still actors holding onto this route
                    C
            end;
        0 when OpName =:= delete ->
            %% Already deleted.
            MCounter
    end.
```

keynslug: > The difference though is that we should have `NRoutes(ClusterName) = Gauge(ClusterName) + Counter(ClusterName)` done on each node, and then we can take the maximum value when aggregating.

Won't it be a bit off because counters are still per-node, so even the maximum will not reflect changes made by _all_ actors (which can be connected to different nodes)? I.e. won't something like this will be closer to the truth?
```
[G1 | Gauges] = [Gauge(ClusterName @ N) || N <- Nodes],
[M1 | Metrics] = [Counter(ClusterName @ N) || N <- Nodes],
G1 + M1 + sum([G - G1 + M || {G, M} <- zip(Gauges, Metrics)])
```

It's obvious this starts to look complex and hard to reason about, so might be a subject for a followup. 🫠

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3449)

PR #13495, opened 2024-07-19
chore: hide enable flags from schema and config examples

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12730

Release version: v/e5.8

```sh
ͳ jq 'path(.. | select(.text? == "enable")) as $path | getpath($path | . += ["hash"])' < _build/docgen/emqx-enterprise/schema-v2-en.json
```

## PR Checklist

### DIFF
--- a/rel/config/examples/gateway.exproto.conf.example
+++ b/rel/config/examples/gateway.exproto.conf.example
@@ -16,7 +16,6 @@ gateway.exproto {
     ## Configurations for request to ConnectionHandler service
     handler {
         address = "http://127.0.0.1:9001"
-        ssl_options {enable = false}
     }
 
     listeners.tcp.default {
--- a/rel/config/examples/gateway.exproto.conf.example
+++ b/rel/config/examples/gateway.exproto.conf.example
@@ -16,6 +16,7 @@ gateway.exproto {
     ## Configurations for request to ConnectionHandler service
     handler {
         address = "http://127.0.0.1:9001"
+        ssl_options {enable = false}
     }
 
     listeners.tcp.default {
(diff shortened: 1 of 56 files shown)


zmstone:
Some of the enable flags having default value `false` will be problematic if it's not documented and not provided in the examples either. 
To avoid this, we should unify all the enable flags to have default value `true`, for the ones with `default => false` today, we should change the parent fields' schema to union with `disabled` and set its default value to `disabled`.

rel/config/examples/gateway.exproto.conf.example
```diff
@@ -16,7 +16,6 @@ gateway.exproto {
     ## Configurations for request to ConnectionHandler service
     handler {
         address = "http://127.0.0.1:9001"
-        ssl_options {enable = false}
```
zmstone, line 19: we'll need to keep this for now.

rel/config/examples/plugins.conf.example
```diff
@@ -9,10 +9,8 @@ plugins {
             ##     Format: {name}-{version}
             ##     Note: name and version should be what it is in the plugin application
             name_vsn = "my_acl-0.1.0",
-
-            enable = true ## enable this plugin
         },
-        {name_vsn = "my_rule-0.1.1", enable = false}
```
zmstone, line 15: we'll need to keep this as an example to show how to disable a plugin.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3446)

PR #13518, opened 2024-07-24
feat(kafka producer): allow dynamic topics (mkIII)

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12656

Release version: e5.7.2

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
+++ b/apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
@@ -295,6 +295,7 @@ fields("config_producer") ->
 fields("config_consumer") ->
     fields(kafka_consumer);
 fields(kafka_producer) ->
+    %% Schema used by bridges V1.
     connector_config_fields() ++ producer_opts(v1);
 fields(kafka_producer_action) ->
     [
@@ -306,6 +307,10 @@ fields(kafka_producer_action) ->
         {tags, emqx_schema:tags_schema()},
         {description, emqx_schema:description_schema()}
     ] ++ producer_opts(action);
+fields(pre_configured_topic) ->
+    [
+        {topic, mk(string(), #{required => true, desc => ?DESC(kafka_topic)})}
+    ];
 fields(kafka_consumer) ->
     connector_config_fields() ++ fields(consumer_opts);
 fields(ssl_client_opts) ->
@@ -364,9 +369,41 @@ fields(socket_opts) ->
                 validator => fun emqx_schema:validate_tcp_keepalive/1
             })}
     ];
+fields(v1_producer_kafka_opts) ->
+    OldSchemaFields =
+        [
+            topic,
+            message,
+            max_batch_bytes,
+            compression,
+            partition_strategy,
+            required_acks,
+            kafka_headers,
+            kafka_ext_headers,
+            kafka_header_value_encode_mode,
+            partition_count_refresh_interval,
+            partitions_limit,
+            max_inflight,
+            buffer,
+            query_mode,
+            sync_query_timeout
+        ],
+    Fields = fields(producer_kafka_opts),
+    lists:filter(
+        fun({K, _V}) -> lists:member(K, OldSchemaFields) end,
+        Fields
+    );
 fields(producer_kafka_opts) ->
     [
         {topic, mk(string(), #{required => true, desc => ?DESC(kafka_topic)})},
+        {pre_configured_topics,
+            mk(
+                hoconsc:array(ref(pre_configured_topic)),
+                #{
+                    default => [],
+                    desc => ?DESC("producer_pre_configured_topics")
+                }
+            )},
         {message, mk(ref(kafka_message), #{required => false, desc => ?DESC(kafka_message)})},
         {max_batch_bytes,
             mk(emqx_schema:bytesize(), #{default => <<"896KB">>, desc => ?DESC(max_batch_bytes)})},
@@ -675,15 +712,15 @@ resource_opts() ->
 %% However we need to keep it backward compatible for generated schema json (version 0.1.0)
 %% since schema is data for the 'schemas' API.
 parameters_field(ActionOrBridgeV1) ->
-    {Name, Alias} =
+    {Name, Alias, Ref} =
         case ActionOrBridgeV1 of
             v1 ->
-                {kafka, parameters};
+                {kafka, parameters, v1_producer_kafka_opts};
             action ->
-                {parameters, kafka}
+                {parameters, kafka, producer_kafka_opts}
         end,
     {Name,
-        mk(ref(producer_kafka_opts), #{
+        mk(ref(Ref), #{
             required => true,
             aliases => [Alias],
             desc => ?DESC(producer_kafka_opts),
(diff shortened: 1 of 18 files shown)


apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
```diff
@@ -306,6 +307,10 @@ fields(kafka_producer_action) ->
         {tags, emqx_schema:tags_schema()},
         {description, emqx_schema:description_schema()}
     ] ++ producer_opts(action);
+fields(pre_configured_topic) ->
```
zmstone, line 310: there should be no need to pre-configure topics now?
given wolff now supports adding topics in the send2 call.

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
```diff
@@ -364,9 +369,41 @@ fields(socket_opts) ->
...
+            sync_query_timeout
+        ],
+    Fields = fields(producer_kafka_opts),
+    lists:filter(
+        fun({K, _V}) -> lists:member(K, OldSchemaFields) end,
+        Fields
+    );
 fields(producer_kafka_opts) ->
     [
         {topic, mk(string(), #{required => true, desc => ?DESC(kafka_topic)})},
```
zmstone, line 398:
```suggestion
        {topic, mk(emqx_schema:template(), #{required => true, desc => ?DESC(kafka_topic)})},
```

changes/ee/feat-13452.en.md
```diff
@@ -0,0 +1 @@
+Added to possibility to configure a list of predefined Kafka topics to Kafka producer actions, and also to use templates to define the destination Kafka topic.
```
zmstone, line 1: Please verify if the impact of unknown topic is correct below.
```suggestion
Kafka producer's `topic` config now supports template.
The topics must be created already in Kafka. If a message is rendered towards a non-existing topic in Kafka (given Kafka disabled topic auto-creation), the message will fail with an unrecoverable error. This can cause other messages in the same action to delay.
```

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
```diff
@@ -549,22 +604,29 @@ on_get_channel_status(
...
 
-check_topic_and_leader_connections(ClientId, KafkaTopic, MaxPartitions) ->
+check_topic_and_leader_connections(ActionResId, ClientId, MKafkaTopic, MaxPartitions) ->
     case wolff_client_sup:find_client(ClientId) of
         {ok, Pid} ->
-            ok = check_topic_status(ClientId, Pid, KafkaTopic),
-            ok = check_if_healthy_leaders(ClientId, Pid, KafkaTopic, MaxPartitions);
+            maybe
+                true ?= is_binary(MKafkaTopic),
+                ok = check_topic_status(ClientId, Pid, MKafkaTopic),
```
zmstone, line 612: how does it behave when it's a template ?

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
```diff
@@ -773,20 +837,19 @@ handle_telemetry_event(_EventId, _Metrics, _MetaData, _HandlerConfig) ->
 %% Note: don't use the instance/manager ID, as that changes everytime
 %% the bridge is recreated, and will lead to multiplication of
 %% metrics.
--spec telemetry_handler_id(resource_id()) -> binary().
-telemetry_handler_id(ResourceID) ->
-    <<"emqx-bridge-kafka-producer-", ResourceID/binary>>.
+-spec telemetry_handler_id(action_resource_id()) -> binary().
+telemetry_handler_id(ActionResId) ->
+    <<"emqx-bridge-kafka-producer-", ActionResId/binary>>.
```
zmstone, line 842: nit: 
```suggestion
    ActionResId.
```

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
```diff
@@ -591,8 +653,10 @@ error_summary(Map, [Error]) ->
 error_summary(Map, [Error | More]) ->
     Map#{first_error => Error, total_errors => length(More) + 1}.
 
-check_if_healthy_leaders(ClientId, ClientPid, KafkaTopic, MaxPartitions) when is_pid(ClientPid) ->
-    case wolff_client:get_leader_connections(ClientPid, KafkaTopic, MaxPartitions) of
+check_if_healthy_leaders(ActionResId, ClientId, ClientPid, KafkaTopic, MaxPartitions) when
+    is_pid(ClientPid)
+->
+    case wolff_client:get_leader_connections(ClientPid, ActionResId, KafkaTopic, MaxPartitions) of
```
zmstone, line 659: how does it behave if it's a template ?

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
thalesmg: Due to the line above this one being inside a `maybe` block, this line won't execute and we'll just return `ok` if the client is found.

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
thalesmg: Since this is only called in `check_topic_and_leader_connections/4`, and we have the `maybe` clause guarding this call, it won't be called if the topic is a template.

changes/ee/feat-13452.en.md
thalesmg: Yes, it'll result in an unrecoverable error.  We check for that in this test line: https://github.com/emqx/emqx/pull/13518/files#diff-a270708602dcaf19e272a18dd76d32b8378d5951b49fad623ea9d975889a01e1R1036

The only small thing I'd say is that I _think_ the "this can cause other messages in the same action to delay" won't apply to Kafka, since it doesn't use builtin buffer workers that may block.

```suggestion
Kafka producer action's `topic` config now supports templates.

The topics must be already created in Kafka. If a message is rendered towards a non-existing topic in Kafka (given Kafka disabled topic auto-creation), the message will fail with an unrecoverable error.  Also, if a message does not contain enough information to render to the configured template (e.g.: the template is `t-${t}` and the message context does not define `t`), this message will also fail with an unrecoverable error. 

This same feature is also available for Azure Event Hubs and Confluent Platform producer integrations.
```

zmstone: > Yes, it'll result in an unrecoverable error. We check for that in this test line: https://github.com/emqx/emqx/pull/13518/files#diff-a270708602dcaf19e272a18dd76d32b8378d5951b49fad623ea9d975889a01e1R1036
> 
> The only small thing I'd say is that I _think_ the "this can cause other messages in the same action to delay" won't apply to Kafka, since it doesn't use builtin buffer workers that may block.

but unrecoverable error would result in resource state transition which needs time to recover from health checks ?

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3233)

PR #13539, opened 2024-07-29
feat(session): move session data to DS

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-11841

Release version: v/e5.8

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -145,9 +145,6 @@
     (is_binary(CLIENTID) orelse (is_atom(CLIENTID) andalso CLIENTID =/= undefined))
 ).
 
-%% When dealing with race conditions during new session creation.
--define(MAX_RETRIES, 5).
-
 %% linting overrides
 -elvis([
     {elvis_style, invalid_dynamic_call, #{ignore => [emqx_cm]}},
@@ -308,31 +305,10 @@ open_session(_CleanStart = false, ClientInfo = #{clientid := ClientId}, ConnInfo
         end
     end).
 
-%% Used only when create a *new* session, not resuming a new one.
-create_register_session(ClientInfo, ConnInfo, MaybeWillMsg, ChanPid) ->
-    Attempts = 0,
-    create_register_session(ClientInfo, ConnInfo, MaybeWillMsg, ChanPid, Attempts).
-
-create_register_session(ClientInfo = #{clientid := ClientId}, ConnInfo, MaybeWillMsg, ChanPid, N) ->
-    case emqx_session:create(ClientInfo, ConnInfo, MaybeWillMsg) of
-        {error, recoverable, session_already_exists} when N < ?MAX_RETRIES ->
-            %% Race condition: session data could still have been replicating and now
-            %% became visible as we attempted to delete it in `open_session'; try again.
-            ok = emqx_session:destroy(ClientInfo, ConnInfo),
-            timer:sleep(100),
-            create_register_session(ClientInfo, ConnInfo, MaybeWillMsg, ChanPid, N + 1);
-        {error, recoverable, session_already_exists} ->
-            {error, failed_to_create_new_session};
-        {error, recoverable, Reason} when N < ?MAX_RETRIES ->
-            ?SLOG(info, #{msg => "cm_failed_to_create_new_session", reason => Reason, retry => true}),
-            timer:sleep(100),
-            create_register_session(ClientInfo, ConnInfo, MaybeWillMsg, ChanPid, N + 1);
-        {error, unrecoverable, Reason} ->
-            {error, Reason};
-        {ok, Session} ->
-            ok = register_channel(ClientId, ChanPid, ConnInfo),
-            {ok, #{session => Session, present => false}}
-    end.
+create_register_session(ClientInfo = #{clientid := ClientId}, ConnInfo, MaybeWillMsg, ChanPid) ->
+    Session = emqx_session:create(ClientInfo, ConnInfo, MaybeWillMsg),
+    ok = register_channel(ClientId, ChanPid, ConnInfo),
+    {ok, #{session => Session, present => false}}.
 
 %% @doc Try to takeover a session from existing channel.
 -spec takeover_session_begin(emqx_types:clientid()) ->
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -145,6 +145,9 @@
     (is_binary(CLIENTID) orelse (is_atom(CLIENTID) andalso CLIENTID =/= undefined))
 ).
 
+%% When dealing with race conditions during new session creation.
+-define(MAX_RETRIES, 5).
+
 %% linting overrides
 -elvis([
     {elvis_style, invalid_dynamic_call, #{ignore => [emqx_cm]}},
@@ -305,10 +308,31 @@ open_session(_CleanStart = false, ClientInfo = #{clientid := ClientId}, ConnInfo
         end
     end).
 
-create_register_session(ClientInfo = #{clientid := ClientId}, ConnInfo, MaybeWillMsg, ChanPid) ->
-    Session = emqx_session:create(ClientInfo, ConnInfo, MaybeWillMsg),
-    ok = register_channel(ClientId, ChanPid, ConnInfo),
-    {ok, #{session => Session, present => false}}.
+%% Used only when create a *new* session, not resuming a new one.
+create_register_session(ClientInfo, ConnInfo, MaybeWillMsg, ChanPid) ->
+    Attempts = 0,
+    create_register_session(ClientInfo, ConnInfo, MaybeWillMsg, ChanPid, Attempts).
+
+create_register_session(ClientInfo = #{clientid := ClientId}, ConnInfo, MaybeWillMsg, ChanPid, N) ->
+    case emqx_session:create(ClientInfo, ConnInfo, MaybeWillMsg) of
+        {error, recoverable, session_already_exists} when N < ?MAX_RETRIES ->
+            %% Race condition: session data could still have been replicating and now
+            %% became visible as we attempted to delete it in `open_session'; try again.
+            ok = emqx_session:destroy(ClientInfo, ConnInfo),
+            timer:sleep(100),
+            create_register_session(ClientInfo, ConnInfo, MaybeWillMsg, ChanPid, N + 1);
+        {error, recoverable, session_already_exists} ->
+            {error, failed_to_create_new_session};
+        {error, recoverable, Reason} when N < ?MAX_RETRIES ->
+            ?SLOG(info, #{msg => "cm_failed_to_create_new_session", reason => Reason, retry => true}),
+            timer:sleep(100),
+            create_register_session(ClientInfo, ConnInfo, MaybeWillMsg, ChanPid, N + 1);
+        {error, unrecoverable, Reason} ->
+            {error, Reason};
+        {ok, Session} ->
+            ok = register_channel(ClientId, ChanPid, ConnInfo),
+            {ok, #{session => Session, present => false}}
+    end.
 
 %% @doc Try to takeover a session from existing channel.
 -spec takeover_session_begin(emqx_types:clientid()) ->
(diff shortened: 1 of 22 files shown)


apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -88,6 +88,14 @@ namespace() ->
 
 schema() ->
     [
+        {sessions,
+            ds_schema(#{
+                default =>
+                    #{
+                        <<"backend">> => ?DEFAULT_BACKEND
```
ieQu1, line 95: I don't think we should make this backend configurable at all. For each type of release we should hardcode the backend offering the highest degree of consistency:
- OSS: local
- EE: raft
- ...

apps/emqx/src/emqx_persistent_message.erl
```diff
@@ -74,6 +74,16 @@ storage_backend(Path) ->
     ConfigTree = #{'_config_handler' := {Module, Function}} = emqx_config:get(Path),
     apply(Module, Function, [ConfigTree]).
 
+-ifdef(STORE_STATE_IN_DS).
+initialize_session_ds_state() ->
+    ok = emqx_persistent_session_ds_state:open_db(storage_backend([durable_storage, sessions])).
+%% ELSE ifdef(STORE_STATE_IN_DS).
```
ieQu1, line 80: It's usually a nice idea to add comments to the long macros, but our automatic formatter makes them so ugly it's not really worth it.

apps/emqx/src/emqx_persistent_session_ds/session_internals.hrl
```diff
@@ -19,6 +19,8 @@
 -include("emqx_persistent_message.hrl").
 -include("emqx_durable_session_metadata.hrl").
 
+-define(PERSISTENT_SESSION_DB, emqx_persistent_session).
```
ieQu1, line 22: DURABLE_SESSION_META or DURABLE_SESSION_STATE?

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
```diff
@@ -570,17 +904,67 @@ session_iterator_next(Cursor0, N) ->
...
+key_encode(?awaiting_rel_domain, PacketId) ->
+    integer_to_binary(PacketId).
+
+-spec key_decode(domain(), binary()) -> term().
+key_decode(?metadata_domain, Bin) ->
+    Bin;
+key_decode(?subscription_domain, Bin) ->
+    topic_level_to_term(Bin);
+key_decode(?subscription_state_domain, Bin) ->
+    topic_level_to_term(Bin);
```
ieQu1, line 945: Subscription state id is an integer, transforming it to binary then to base64 is quite inefficient.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
```diff
@@ -570,17 +904,67 @@ session_iterator_next(Cursor0, N) ->
...
 %% All mnesia reads and writes are passed through this function.
 %% Backward compatiblity issues can be handled here.
+-ifdef(STORE_STATE_IN_DS).
+val_encode(_Domain, Term) ->
+    term_to_binary(Term).
+
+val_decode(_Domain, Bin) ->
+    binary_to_term(Bin).
+
+-spec key_encode(domain(), term()) -> binary().
```
ieQu1, line 923: Note: key in the DB doesn't necessarily need to be the same as key in pmap. Is it possible to introduce integer handles (created via `get_unique_id`) for objects in PMAP, that will be invisible to the session, but used for (de)serialization?
From my experience, term_to_binary/binary_to_term tends to make unwelcome appearances in the flamegraph all too often.

thalesmg: I think we may need to be able to map a key to a topic key deterministically, at least for some cases.

For example, `cold_get_subscription_state` and `cold_get_subscription` directly access specific topics in a session's subscriptions:

https://github.com/thalesmg/emqx/blob/9b7a747c430a2d8c27ad32032e76c3a744664752/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl#L645-L649

So, if we were to generate unique integers for each subscription or subscription state in the PMAP, I think we would then need to scan all subscriptions / subscription states to find the one matching the argument.

Maybe it's a fine tradeoff, since these functions which are apparently only called from the management APIs?

ieQu1: Performance of cold* operations is not very important.

apps/emqx/src/emqx_persistent_session_ds/session_internals.hrl
thalesmg: Renamed to the latter.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
thalesmg: Actually, I think we can't generate unique ids for each time we encounter a key.  We need the mapping from pmap key to topic key to be deterministic since we want to overwrite any existing messages with the same key when committing.

ieQu1: This mapping can be stored in the pmap structure, perhaps?

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
```diff
@@ -570,17 +904,67 @@ session_iterator_next(Cursor0, N) ->
...
+    topic_level_to_term(Bin);
+key_decode(?rank_domain, Bin) ->
+    topic_level_to_term(Bin);
+key_decode(?seqno_domain, Bin) ->
+    binary_to_integer(Bin);
+key_decode(?awaiting_rel_domain, Bin) ->
+    binary_to_integer(Bin).
+
+term_to_topic_level(Term) ->
+    base64:encode(term_to_binary(Term), #{mode => urlsafe, padding => false}).
```
ieQu1, line 956: Do we need base64? Since these "topics" probably won't be exposed via MQTT protocol, we don't care if topic level is valid UTF8.

thalesmg: It was a way to ensure that the topic level doesn't contain invalid characters like `/`, `+`, `#`, whatever the input is.  We can use another encoding that satisfies that, sure.

apps/emqx/test/emqx_persistent_session_ds_state_tests.erl
```diff
@@ -316,12 +417,57 @@ get_state(SessionId) ->
...
+%% END ifdef(STORE_STATE_IN_DS).
+-endif.
 
+-ifdef(STORE_STATE_IN_DS).
+clean() ->
+    ets:delete(?tab),
+    emqx_ds:drop_db(?DB),
+    application:stop(emqx_ds_backends),
+    application:stop(emqx_ds_builtin_local),
+    mria:stop(),
```
qzhuyan, line 460: why do we need to stop `mria` here just for clean test env or it is a must?

thalesmg: It's to clean up the test env, indeed.

Note that this eunit module already did that prior to the changes herein, so this is not a new addition.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
thalesmg: It can.  But we will still need to be stored alongside the metadata, as the mapping must be reversible.  Otherwise, we won't be able to overwrite existing internal keys after closing and then reopening sessions.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
ieQu1: Fair enough.

ieQu1: On a second thought, DS API treats topics as lists of binaries, and wildcards in the topic filters as atoms `+` and `#`. There should not be any `binary()` -> `[word()]` parsing.
The only exception is `skipstream_lts` layout's topic compaction... It does not have escaping for `/`'s in the topic levels currently. But that can be fixed.

thalesmg: > On a second thought, DS API treats topics as lists of binaries, and wildcards in the topic filters as atoms + and #. There should not be any binary() -> [word()] parsing.

I think I'm a bit lost, sorry.  Where would that binary() -> [word()] parsing be involved in?

The highlighted `term_to_topic_level` function was used for encoding arbitrary terms (such as `{[word()], nil()}`, for example) to a single topic level.  We need to encode to binary and then escape bad characters (like `#`) because we need to project terms to a binary topic level.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -186,7 +187,13 @@
...
         %% `store_batch'.
         %% Default: `true'.
-        sync => boolean()
+        sync => boolean(),
+        %% Whether the whole batch given to `store_batch' should be inserted atomically as
+        %% a unit.  Note: the whole batch must be crafted so that it belongs to a single
+        %% shard (if applicable to the backend), as the batch will be split accordingly
+        %% even if this flag is `true'.
+        %% Default: `false'.
+        atomic => boolean()
```
thalesmg, line 196: This flag apparently is still being "checked" by [DS buffer](https://github.com/emqx/emqx/blob/e42021d31482fe25bc6b3a043dcc1ae7f52b3476/apps/emqx_durable_storage/src/emqx_ds_buffer.erl#L84), but it doesn't do anything.  :thinking:

apps/emqx/src/emqx_cm.erl
```diff
@@ -304,10 +304,18 @@ open_session(_CleanStart = false, ClientInfo = #{clientid := ClientId}, ConnInfo
...
 create_register_session(ClientInfo = #{clientid := ClientId}, ConnInfo, MaybeWillMsg, ChanPid) ->
-    Session = emqx_session:create(ClientInfo, ConnInfo, MaybeWillMsg),
-    ok = register_channel(ClientId, ChanPid, ConnInfo),
-    {ok, #{session => Session, present => false}}.
+    case emqx_session:create(ClientInfo, ConnInfo, MaybeWillMsg) of
+        {error, recoverable, session_already_exists} ->
+            %% Race condition: session data could still have been replicating and now
+            %% became visible as we attempted to delete it in `open_session'; try again.
+            ok = emqx_session:destroy(ClientInfo, ConnInfo),
+            create_register_session(ClientInfo, ConnInfo, MaybeWillMsg, ChanPid);
```
thalesmg, line 314: TODO: sleep and retry a finite amount of times.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
```diff
@@ -654,19 +1169,77 @@ pmap_open(Table, SessionId) ->
...
+        dirty = Dirty#{IntK => dirty}
+    }.
+
+get_or_gen_internal_key(K, KeyMapping, _Domain, _Cache) when
+    is_map_key(K, KeyMapping)
+->
+    IntK = maps:get(K, KeyMapping),
+    {IntK, KeyMapping};
+get_or_gen_internal_key(K, KeyMapping0, Domain, Cache) ->
+    IntK = erlang:unique_integer(),
```
thalesmg, line 1209: Must generate keys in a different manner for streams to preserve their order relation.

qzhuyan (dismissed):
as agreed, we merge it first as it is controlled by build flag. I will review it later.

qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3623)

PR #13589, opened 2024-08-07
fix(rule engine api): check that user provided id is valid

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12838

Release version: v/e5.8

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
@@ -20,6 +20,7 @@
 -include_lib("emqx/include/logger.hrl").
 -include_lib("hocon/include/hoconsc.hrl").
 -include_lib("typerefl/include/types.hrl").
+-include_lib("emqx_utils/include/emqx_utils_api.hrl").
 
 -behaviour(minirest_api).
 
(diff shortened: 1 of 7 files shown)


apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
```diff
@@ -405,9 +406,11 @@ param_path_id() ->
                                 },
                                 #{tag => ?TAG}
                             ),
-                            {400, #{code => 'BAD_REQUEST', message => ?ERR_BADARGS(Reason)}}
+                            ?BAD_REQUEST(?ERR_BADARGS(Reason))
                     end
-            end
+            end;
+        _BadId ->
+            ?BAD_REQUEST(<<"rule id must be a string">>)
```
id, line 413: To be 100% clear, the issue is that a user can create a rule where id is literal string "null". Does this patch handle it?

thalesmg: Hmm I managed to reproduce the issue by passing the JSON null value, not a string `"null"`.  :thinking: 

I'll add an assertion for the `"null"` case, but I _think_ it shouldn't be a problem, so long it's a string.

thalesmg: Turns out it is a bug in hocon: https://github.com/emqx/hocon/pull/299

id (approved):

PR #13600, opened 2024-08-12
ci: fix RUN_FROM for the docker image

id:
`RUN_FROM` was using builder image and the final docker image size was > 1GB.

The image has been rebuilt and updated, this is to ensure it stays this way in case we need to release another one 5.7.x.

Release version: v/e5.7.*


zmstone (approved):

qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3554)

PR #13609, opened 2024-08-14
update ci

id:


.github/workflows/build_and_push_docker_images.yaml
```diff
@@ -20,13 +20,13 @@ on:
...
+        required: false
       DOCKER_HUB_TOKEN:
-        required: true
+        required: false
       AWS_ACCESS_KEY_ID:
-        required: true
+        required: false
       AWS_SECRET_ACCESS_KEY:
-        required: true
+        required: false
```
thalesmg, line 29: Q: wouldn't it be better to fail earlier if those are missing?

id: For sure. But I've got a weird "Secret `DOCKER_HUB_USER` is required, but not provided while calling ..." error from github without this change. Couldn't find the root cause.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3848)

PR #13623, opened 2024-08-15
ci: use build container in codeql workflow

id:
CodeQL analysis for cpp does not work when we build in container, but run analyze on host.


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3851)

PR #13630, opened 2024-08-15
Sync shared subs from master

savonarola:


apps/emqx_ds_shared_sub/src/emqx_ds_shared_sub_proto.erl
```diff
@@ -57,6 +51,20 @@
     agent_metadata/0
 ]).
 
+-define(log_agent_msg(ToLeader, Msg),
+    ?tp(debug, shared_sub_proto_msg, #{
```
thalesmg, line 55: Logs that were at `warning` level now are `debug`.  Is it intentional?

thalesmg (approved):

apps/emqx_ds_shared_sub/src/emqx_ds_shared_sub_proto.erl
savonarola: Yes

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3565)

PR #13650, opened 2024-08-20
Feat: Suppport datalayers integration

HJianBo:
Fixes https://emqx.atlassian.net/browse/EMQX-12861
Release version: v/e5.8.0

## PR Checklist

- ~Added property-based tests for code which performs user input validation~

- ~If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)~


.ci/docker-compose-file/docker-compose-datalayers-tcp.yaml
```diff
@@ -0,0 +1,13 @@
+version: '3.9'
+
+services:
+  datalayers_server_tcp:
+    container_name: datalayers_tcp
+    image: datalayers/datalayers:${DATALAYERS_TAG}
+    expose:
+      - "8361"
+    environment:
+      DATALAYERS_SERVER__TIMEZONE: "Etc/GMT"
```
zmstone, line 10: Q: why this zone ?

apps/emqx_bridge_datalayers/README.md
```diff
@@ -0,0 +1,43 @@
+# EMQX Datalayers Bridge
+
+[Datalayers](https://docs.datalayers.cn/datalayers/latest/) is a multimodal,
```
zmstone, line 3: this link is 404

id: may be just https://datalayers.io/
EN doc site is empty

HJianBo: Their Product team told me they only have the Chinese site for now  🤒

.ci/docker-compose-file/docker-compose-datalayers-tcp.yaml
HJianBo: It is actually to guarantee they have the same time string (i.e: 2024-08-20T22:09:03.136418Z) between the API query result with `calendar:system_time_to_rfc3339` in tests

https://github.com/HJianBo/emqx/blob/1a572cb93033c1b2047de99f0fc573251bfa359e/apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_SUITE.erl#L566

.ci/docker-compose-file/.env
```diff
@@ -12,6 +12,7 @@ OPENTS_TAG=9aa7f88
 KINESIS_TAG=2.1
 HSTREAMDB_TAG=v0.19.3
 HSTREAMDB_ZK_TAG=3.8.1
+DATALAYERS_TAG=nightly
```
thalesmg, line 15: I think it would be best to pin a fixed version here.  Otherwise, it might break CI and take some time to realize it's due to an upstream image update.

apps/emqx_bridge/src/schema/emqx_bridge_enterprise.erl
thalesmg, line 1: This module shouldn't have any changes, I believe.  This is for v1 bridges, and deprecated.  Since this a new bridge, we don't need to support v1 API for it.

apps/emqx_bridge_datalayers/mix.exs
```diff
@@ -0,0 +1,32 @@
...
+    ]
+  end
+
+  def application do
+    [extra_applications: UMP.extra_applications()]
+  end
+
+  def deps() do
+    [
+      {:influxdb, github: "emqx/influxdb-client-erl", tag: "1.1.13"},
```
thalesmg, line 26: Since this is (now) a shared dependency with `emqx_bridge_influxdb`, it would be best to move it to a `common_dep/1` clause in the root `mix.exs`, so we can centralized shared deps there.

.ci/docker-compose-file/.env
HJianBo: Of course, I'll send another PR to update once the Datalayers have a new stable tag.

apps/emqx_bridge_datalayers/mix.exs
HJianBo: Thanks for your comments. Since I don't understand the mix compilation process, I'm unsure if my new commit understands what you mean.

Please help review it again.

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_connector.erl
```diff
@@ -0,0 +1,894 @@
...
+    ConnConfig0 = maps:without([write_syntax, precision], BridgeV1Config),
+    ConnConfig1 =
+        case emqx_utils_maps:indent(parameters, IndentKeys, ConnConfig0) of
+            #{parameters := #{database := _}} = Conf -> Conf
+        end,
+    emqx_utils_maps:update_if_present(
+        resource_opts,
+        fun emqx_connector_schema:project_to_connector_resource_opts/1,
+        ConnConfig1
+    ).
```
thalesmg, line 236:
```suggestion
```

thalesmg (commented):
General question: the connector implementation here looks very similar to the existing influxdb connector.  

Would it be possible to make a thin wrapper around the existing implementation or, even better, to simply use the existing module, for better code reuse?  

An example of a connector + action that does this is [Confluent](https://github.com/emqx/emqx/blob/9a31238a69507f8a5a3c92fe90773ed27c2a6cac/apps/emqx_bridge_confluent/src/emqx_bridge_confluent_producer_connector_info.erl#L30-L31), which simply defers the `emqx_resource` implementation to the existing Kafka implementation.

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers.erl
```diff
@@ -0,0 +1,405 @@
...
+%% Examples
+-export([
+    bridge_v2_examples/1,
+    conn_bridge_examples/1,
+    connector_examples/1
+]).
+
+-type write_syntax() :: list().
+-reflect_type([write_syntax/0]).
+-typerefl_from_string({write_syntax/0, ?MODULE, to_datalayers_lines}).
```
thalesmg, line 35: Is the implementation the exact same as the one in `emqx_bridge_influxdb:write_syntax()`?  If so, we could just use that to centralize the implementation.

scripts/ct/run.sh
```diff
@@ -247,18 +247,22 @@ for dep in ${CT_DEPS}; do
...
+	        ;;
+	    azurite)
+	        FILES+=( '.ci/docker-compose-file/docker-compose-azurite.yaml' )
+	        ;;
+	    couchbase)
+	        FILES+=( '.ci/docker-compose-file/docker-compose-couchbase.yaml' )
+	        ;;
+	    kdc)
+	        FILES+=( '.ci/docker-compose-file/docker-compose-kdc.yaml' )
+	        ;;
```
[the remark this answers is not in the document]
HJianBo: Seems a little better to keep it aligned with the previous lines?

HJianBo:
> Would it be possible to make a thin wrapper around the existing implementation

Yes. Stone also told me that. and I have created [EMQX-12936](https://emqx.atlassian.net/browse/EMQX-12936) to trace that. I'm tending to do it in the next release (may e5.8.1), since the e5.8.0 is about to close🤒.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3661)

PR #13656, opened 2024-08-20
feat(message transformation dry run): add more input fields

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12932

Release version: e5.8

## PR Checklist

### DIFF
--- a/apps/emqx_message_transformation/src/emqx_message_transformation_http_api.erl
+++ b/apps/emqx_message_transformation/src/emqx_message_transformation_http_api.erl
@@ -39,7 +39,8 @@
 -define(METRIC_NAME, message_transformation).
 
 -type user_property() :: #{binary() => binary()}.
--reflect_type([user_property/0]).
+-type publish_properties() :: #{binary() => binary() | integer()}.
+-reflect_type([user_property/0, publish_properties/0]).
 
 %%-------------------------------------------------------------------------------------------------
 %% `minirest' and `minirest_trails' API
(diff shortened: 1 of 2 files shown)


apps/emqx_message_transformation/src/emqx_message_transformation_http_api.erl
```diff
@@ -305,15 +306,29 @@ fields(dryrun_input_message) ->
...
             mk(
                 typerefl:alias("map(binary(), binary())", user_property()),
                 #{default => #{}}
-            )}
+            )},
+        {username, mk(binary(), #{required => false})}
+    ];
+fields(dryrun_input_flags) ->
+    [
+        {dup, mk(boolean(), #{default => false})},
```
zmstone, line 330: I don't think any message should be explicit set dup flag to 'true'.
dup is only set when a message is re-transmitted after session resume.

thalesmg: This is for dry run testing a transformation, so it would make sense to allow simulating a duplicate message?

zmstone: the dup flag might be set by the publish-client, but it's cleared before calling the message.publish hook
https://github.com/emqx/emqx/blob/acb8c994fa4825af988f36fa6929853e0204a21d/apps/emqx/src/emqx_broker.erl#L246

the dup flag is only set before re-sending a message to subscriber-client
https://github.com/emqx/emqx/blob/master/apps/emqx/src/emqx_session_mem.erl#L661

thalesmg: I see, thanks for the clarification.  Removed this.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3631)

PR #13688, opened 2024-08-23
test(trace): fix flaky test

zmstone:
409 only happens when two requests were sent in the same second.


qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3669)

PR #13689, opened 2024-08-23
test: ensure prevent_overlapping_partitions from ERL_FLAGS

zmstone:
The last fix did not work because the flag was read at init time when `global` starts.

This is an attempt to fix test flakiness 
```
'global' at node 'test@127.0.0.1' requested disconnect from node 't_crash_stop_recover2@127.0.0.1' in order to prevent overlapping partitions
```


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3670)

PR #13724, opened 2024-08-29
chore(connector aggregator): trigger delivery when max records is pushed

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12927

Release version: e5.8.1

## PR Checklist

### DIFF
--- a/apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
+++ b/apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
@@ -85,28 +85,34 @@ buffer_to_map(#buffer{} = Buffer) ->
 %%
 
 write_records_limited(Name, Buffer = #buffer{max_records = undefined}, Records) ->
-    write_records(Name, Buffer, Records);
+    write_records(Name, Buffer, Records, _NumWritten = undefined);
 write_records_limited(Name, Buffer = #buffer{max_records = MaxRecords}, Records) ->
     NR = length(Records),
     case inc_num_records(Buffer, NR) of
         NR ->
             %% NOTE: Allow unconditionally if it's the first write.
-            write_records(Name, Buffer, Records);
+            write_records(Name, Buffer, Records, NR);
         NWritten when NWritten > MaxRecords ->
             NextBuffer = rotate_buffer(Name, Buffer),
             write_records_limited(Name, NextBuffer, Records);
-        _ ->
-            write_records(Name, Buffer, Records)
+        NWritten ->
+            write_records(Name, Buffer, Records, NWritten)
     end.
 
-write_records(Name, Buffer = #buffer{fd = Writer}, Records) ->
+write_records(Name, Buffer = #buffer{fd = Writer, max_records = MaxRecords}, Records, NumWritten) ->
     case emqx_connector_aggreg_buffer:write(Records, Writer) of
         ok ->
             ?tp(connector_aggreg_records_written, #{action => Name, records => Records}),
+            case is_number(NumWritten) andalso NumWritten >= MaxRecords of
+                true ->
+                    rotate_buffer_async(Name, Buffer);
+                false ->
+                    ok
+            end,
             ok;
         {error, terminated} ->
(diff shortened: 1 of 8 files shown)


apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -99,10 +99,12 @@ write_records_limited(Name, Buffer = #buffer{max_records = MaxRecords}, Records)
             write_records(Name, Buffer, Records)
     end.
 
-write_records(Name, Buffer = #buffer{fd = Writer}, Records) ->
+write_records(Name, Buffer = #buffer{fd = Writer, max_records = MaxRecords}, Records) ->
     case emqx_connector_aggreg_buffer:write(Records, Writer) of
         ok ->
             ?tp(connector_aggreg_records_written, #{action => Name, records => Records}),
+            get_num_records(Buffer) >= MaxRecords andalso
```
zmstone, line 106: why not line 95?

thalesmg: data has not been written at that point.  there, the buffer is rotated prior to actually writing, we want to write and then trigger the rotation if max has been reached.

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -99,10 +99,12 @@ write_records_limited(Name, Buffer = #buffer{max_records = MaxRecords}, Records)
             write_records(Name, Buffer, Records)
```
keynslug, line 99: I guess you can just pass down `NWritten` from here, to skip going into ETS for basically the same number.

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -99,10 +99,12 @@ write_records_limited(Name, Buffer = #buffer{max_records = MaxRecords}, Records)
...
     end.
 
-write_records(Name, Buffer = #buffer{fd = Writer}, Records) ->
+write_records(Name, Buffer = #buffer{fd = Writer, max_records = MaxRecords}, Records) ->
     case emqx_connector_aggreg_buffer:write(Records, Writer) of
         ok ->
             ?tp(connector_aggreg_records_written, #{action => Name, records => Records}),
+            get_num_records(Buffer) >= MaxRecords andalso
+                (_ = rotate_buffer(Name, Buffer)),
```
keynslug, line 107: Nit. This blocks the caller, not for long though. Still, could be a motivation for a separate async variant.

keynslug (dismissed):

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -85,28 +85,34 @@ buffer_to_map(#buffer{} = Buffer) ->
...
+                true ->
+                    rotate_buffer_async(Name, Buffer);
+                false ->
+                    ok
+            end,
             ok;
         {error, terminated} ->
             BufferNext = rotate_buffer(Name, Buffer),
-            write_records(Name, BufferNext, Records);
+            write_records(Name, BufferNext, Records, NumWritten);
```
keynslug, line 115: Ouch. Now that I'm looking at this change, shouldn't that have been `write_records_limited(Name, BufferNext, Records)` all this time? I.e. even before your changes. Otherwise, we could miss tracking number of records for that first, could we?

keynslug (commented):
🔥

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
thalesmg: Indeed, good catch, thanks.  :see_no_evil:
We should "restart" the write attempt if we've just rotated the buffer.  Pushed a fix.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3897)

PR #13742, opened 2024-09-02
fix: topic matching when deliver retainer msg

JimMoen:
Fixes [EMQX-12499](https://emqx.atlassian.net/browse/EMQX-12499)

Release version: v/e5.8.1

## PR Checklist

- ~[ ] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket~

[EMQX-12499]: https://emqx.atlassian.net/browse/EMQX-12499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
+++ b/apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
@@ -374,7 +374,7 @@ check_clientid_banned({Msg, _Topic} = Input, _) ->
         true ->
             ?tp(
                 debug,
-                ignore_retained_message_deliver,
+                ignore_retained_message_due_to_banned,
                 #{
                     reason => ?msg_pub_from_is_banned,
                     clientid => Msg#message.from
@@ -388,7 +388,7 @@ check_clientid_banned({Msg, _Topic} = Input, _) ->
         false ->
             ?tp(
                 debug,
-                ignore_retained_message_deliver,
+                ignore_retained_message_due_to_topic_not_match,
                 #{
                     reason => ?msg_topic_not_match,
                     msg_topic => Msg#message.topic,
--- a/apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
+++ b/apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
@@ -334,24 +334,61 @@ deliver_in_batches(Msgs, BatchSize, Pid, Topic, Limiter0) ->
             Drop
     end.
 
+-define(DELIVER_ALLOWED, true).
+-define(DELIVER_NOT_ALLOWED, false).
+
 deliver_to_client([Msg | T], Pid, Topic) ->
+    Pipe = emqx_utils:pipeline(
+        [
+            fun check_clientid_banned/2,
+            fun 'check_prefixed_$_with_wildcard'/2
+        ],
+        {Msg, Topic},
+        ?DELIVER_NOT_ALLOWED
+    ),
     _ =
-        case emqx_banned:check_clientid(Msg#message.from) of
-            false ->
+        case Pipe of
+            {ok, _, ?DELIVER_ALLOWED} ->
                 Pid ! {deliver, Topic, Msg};
-            true ->
-                ?tp(
-                    notice,
-                    ignore_retained_message_deliver,
-                    #{
-                        reason => "client is banned",
-                        clientid => Msg#message.from
-                    }
-                )
+            {error, _, _} ->
+                ok
         end,
     deliver_to_client(T, Pid, Topic);
 deliver_to_client([], _, _) ->
     ok.
+
+check_clientid_banned({Msg, _Topic} = Input, _) ->
+    case emqx_banned:check_clientid(Msg#message.from) of
+        false ->
+            {ok, Input, ?DELIVER_ALLOWED};
+        true ->
+            ?tp(
+                notice,
+                ignore_retained_message_deliver,
+                #{
+                    reason => client_is_banned,
+                    clientid => Msg#message.from
+                }
+            ),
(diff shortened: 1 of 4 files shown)


apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -334,25 +334,62 @@ deliver_in_batches(Msgs, BatchSize, Pid, Topic, Limiter0) ->
...
+            ),
+            {error, client_is_banned, ?DELIVER_NOT_ALLOWED}
+    end.
+
+'check_prefixed_$_with_wildcard'({Msg, Topic} = Input, _) ->
+    case emqx_topic:match(Msg#message.topic, Topic) of
+        false ->
+            ?tp(
+                notice,
+                ignore_retained_message_deliver,
```
keynslug, line 381: Nit. Doesn't seem like a good idea to me to log `notice` event in this case to be honest. After all, it's the same as unmatched topic filter, which is not even a log event right now, rightly so.

JimMoen: `Msg` here are absolutely matched to a given `Topic`
(Use topic index to retrieve from retainer table)

The topic matching is actually to ensure that messages starting with `$` are not sent to subscriptions starting with `+` or wildcard sub `#`

> The Server MUST NOT match Topic Filters starting with a wildcard character (# or +) with Topic Names beginning with a $ character [MQTT-4.7.2-1]

So in general, the retain messages to be delivered are filtered here, and the log level should be `notice` (I just temporarily used the same notice as the client banned check) or `info`.
They should all use the `info` level to me.

JimMoen: Nit:
Message filtering should be performed before attempting to deliver any messages

```diff
 deliver(Messages, Pid, Topic, Limiter) ->
     case erlang:is_process_alive(Pid) of
         false ->
             no_receiver;
         _ ->
             BatchSize = emqx_conf:get([retainer, flow_control, batch_deliver_number], undefined),
+            NMessages = filter_delivery(Messages, Topic),
             case BatchSize of
                 0 ->
-                    deliver_to_client(Messages, Pid, Topic),
+                    deliver_to_client(NMessages, Pid, Topic),
                     {ok, Limiter};
                 _ ->
-                    deliver_in_batches(Messages, BatchSize, Pid, Topic, Limiter)
+                    deliver_in_batches(NMessages, BatchSize, Pid, Topic, Limiter)
             end
     end.
```

keynslug: >  They should all use the info level to me.

We're just following the spec, IMO there's no point in logging anything other than `debug` event, or perhaps just a tracepoint.

JimMoen: right, debug is enough when following spec 😎

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -334,25 +335,71 @@ deliver_in_batches(Msgs, BatchSize, Pid, Topic, Limiter0) ->
...
-    deliver_to_client(T, Pid, Topic);
+deliver_to_client([Msg | Rest], Pid, Topic) ->
+    Pid ! {deliver, Topic, Msg},
+    deliver_to_client(Rest, Pid, Topic);
 deliver_to_client([], _, _) ->
     ok.
 
+-define(DELIVER_ALLOWED, true).
+-define(DELIVER_NOT_ALLOWED, false).
+-define(msg_pub_from_is_banned, msg_pub_from_is_banned).
```
zmstone, line 346:
```suggestion
-define(msg_pub_from_is_banned, publisher_client_banned).
```

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -334,25 +335,71 @@ deliver_in_batches(Msgs, BatchSize, Pid, Topic, Limiter0) ->
...
+                }
+            ),
+            {error, ?msg_pub_from_is_banned, ?DELIVER_NOT_ALLOWED}
+    end.
+
+'check_prefixed_$_with_wildcard'({Msg, Topic} = Input, _) ->
+    case emqx_topic:match(Msg#message.topic, Topic) of
+        false ->
+            ?tp(
+                debug,
```
zmstone, line 390: no need for logging, not even debug, because such exclusions are not logged during regular routing.

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -334,25 +335,71 @@ deliver_in_batches(Msgs, BatchSize, Pid, Topic, Limiter0) ->
...
 
+-define(DELIVER_ALLOWED, true).
+-define(DELIVER_NOT_ALLOWED, false).
+-define(publisher_client_banned, publisher_client_banned).
+-define(msg_topic_not_match, msg_topic_not_match).
+
+filter_delivery(Messages, Topic) ->
+    FilterFun =
+        fun(Msg) ->
+            Pipe = emqx_utils:pipeline(
```
zmstone, line 352: use `maybe` will be more readable.
so there is no need to follow the odd interface of `emqx_utils:pipeline`

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3923)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-3924)

PR #13800, opened 2024-09-13
ci: always run compile job in push entrypoint

id:
other jobs which depend on it only run on releases


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4069)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4070)

PR #13817, opened 2024-09-17
refactor: add is_exist API for resource existence check

zmstone:
Only a internal refactoring, no interface changes.

This is to prepare for another refactoring which can help to reduce the number of ets copies unless absolutely necessary.


qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4246)

PR #13819, opened 2024-09-17
feat(sessds): use selective realtime routing for QoS 0 messages

keynslug:
Fixes [EMQX-13091](https://emqx.atlassian.net/browse/EMQX-13091).

Release version: v5.8.x

## Summary

Use selective realtime routing for QoS 0 messages, so that they will not hit DS most of the time, thus freeing up resources for QoS 1/2 messages. This inevitably breaks existing subscriber message ordering guarantees.

Currently, selective routing is always enabled, with one notable exception: unless there's `upgrade_qos` enabled somewhere (even if there are no active sessions with `upgrade_qos = true` subscriptions).

### Open questions

1. Do we need to feature-flag this optimization?
2. (As of d9e6e7af36ecded7bfde6a2a536ca5c19a57115d) Rolling upgrade is subtly broken right now: older nodes will not know about `noqos0` persistent routes, and will potentially drop messages. Do we need to take care of that?

### TODOs

* Good understanding of how this should interact with DS Shared Subs stuff.
* Regressions.
* Feature-specific testcases.

## PR Checklist


apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_router.erl
```diff
@@ -17,38 +17,51 @@
...
 
 -type route() :: #ps_route{}.
 -type dest() :: emqx_persistent_session_ds:id() | #share_dest{}.
 
+%% Subscription scope.
+%% Scopes enable limited form of _selective routing_.
+%%  * `root` scope routes "work" for every message.
+%%  * `noqos0` scope routes work only for QoS 1/2 messages.
+%% Mirrors `emqx_broker:subscope()`.
+-type subscope() :: root | noqos0.
```
ieQu1, line 60: `root` -> `all` or `any`

apps/emqx/src/emqx_broker.erl
```diff
@@ -72,12 +78,15 @@
     code_change/3
 ]).
 
--import(emqx_utils_ets, [lookup_value/2, lookup_value/3]).
+%% Tests-only
+-export([
```
ieQu1, line 82: -ifdef(TEST)?

keynslug: Then we would have to guard both the export attribute and every function in the implementation chain because there would be warnings when compiling under non-test profile, and this feels a bit too ugly. 🫠

apps/emqx/src/emqx_broker_helper.erl
```diff
@@ -84,25 +88,47 @@ lookup_subid(SubPid) when is_pid(SubPid) ->
...
--spec get_sub_shard(pid(), emqx_types:topic()) -> non_neg_integer().
-get_sub_shard(SubPid, Topic) ->
-    case create_seq(Topic) of
-        Seq when Seq =< ?SHARD -> 0;
-        _ -> erlang:phash2(SubPid, shards_num()) + 1
+-spec get_sub_shard(pid(), emqx_types:topic(), emqx_broker:subscope()) -> non_neg_integer().
+get_sub_shard(SubPid, Topic, Scope) ->
+    NShards = shards_num(),
+    Offset = shards_scope_offset(Scope) * NShards,
+    case create_seq(Topic, Scope) of
```
ieQu1, line 95: Why include scope into the shard?

apps/emqx/src/emqx_broker.erl
ieQu1: `nowarn_unused_function`?

apps/emqx/src/emqx_broker.erl
```diff
@@ -94,6 +103,17 @@
     end
 ).
 
+%% Subscription scope.
+%% When incoming message matches the subscription topic filter, scope can additionally
+%% narrow down message delivery.
+%% * Scope `any` means a subscription with this scope is interested in all messages.
+%% * Scope `qos0` means a subscription needs _only_ QoS 0 messages.
+%%   This is currently employed by durable sessions to route QoS 0 messages through
+%%   the realtime channel to avoid storing them in the durable storage.
```
ieQu1, line 112:
```suggestion
%%   the direct channel to avoid storing them in the durable storage.
```

Nit: I prefer term "direct", because term "realtime" is, stricly speaking, wrong: there's nothing realtime about forwarding gen_rpc:ordered_cast's via a TCP connection.

apps/emqx/src/emqx_broker.erl
```diff
@@ -94,6 +103,17 @@
...
+%% Currently, this affects only the dispatching phase, but not the routing phase.
+%% Messages are routed between nodes regardless of which subscription scopes nodes have.
+-type subscope() :: any | qos0.
```
ieQu1, line 115: Suggestion, maybe expose something like this instead:

`-type channel() :: direct | direct_0.`

...to reduce the number of terms we have to juggle around when reasoning about different ways to forward messages between the clients from the high-level perspective.

Instead of speaking in terms of "durable" vs. "realtime" (or "direct") with scope any or scope qos0, we can just say we have three "logical channels":

- direct
- durable
- direct_0

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -474,16 +475,20 @@ unsubscribe(
...
-        {ok, S1, #{id := SubId, subopts := SubOpts}} ->
+    case emqx_persistent_session_ds_subs:on_unsubscribe(TopicFilter, Session0) of
+        {ok, S1, #{mode := {persistent, _}, id := SubId, subopts := SubOpts}} ->
             {S2, SchedS} = emqx_persistent_session_ds_stream_scheduler:on_unsubscribe(
                 SubId, S1, SchedS0
             ),
             S = emqx_persistent_session_ds_state:commit(S2),
             Session = Session0#{s := S, stream_scheduler_s := SchedS},
             {ok, Session, SubOpts};
+        {ok, S1, #{mode := realtime, subopts := SubOpts}} ->
```
ieQu1, line 488: It's not really `realtime`...

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_router.erl
```diff
@@ -17,38 +17,48 @@
...
 -include("emqx_ps_ds_int.hrl").
 
 -export([init_tables/0]).
 
 %% Route APIs
 -export([
-    do_add_route/2,
-    do_delete_route/2,
+    add_route/2,
+    add_route/3,
```
ieQu1, line 28: I would question if exposing `add_route/3` is safe. Clients on different nodes (or in different zones) can choose to set it differently, leading to the following situation:  

- first client can be in a zone with upgrade_qos=true
- second one can be in a zone with upgrade_qos=false

Second subscriber doesn't expect QoS0 messages in the durable storage for the topic, while the first one can subscribe to the same topic with QoS1, which will lead to message duplication for the first client. I don't see message deduplication in the session code, correct me if I missed it.

But regardless, having QoS0 messages in both channels negates the advantage of this bypass, as the client still have to read and drop a bunch of messages.

Therefore, as I suggested here https://emqxio.slack.com/archives/C04H4RHKKEY/p1725653029483629 , upgrade_qos for durable sessions should be a _cluster-global_ setting. And as such, `add_route/2` function should decide the "scope" depending on this config and it should ignore client's desire.

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1, line 1: I think there should be some code for dropping QoS0 messages from durable storage batches when direct channel is used.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_router.erl
ieQu1: Another question: how does persistent_session router handle change of this setting?

apps/emqx/src/emqx_broker.erl
```diff
@@ -464,12 +504,19 @@ subscriber_down(SubPid) ->
 %% Management APIs
 %%--------------------------------------------------------------------
 
+-spec subscriptions() ->
+    [{{emqx_types:topic() | emqx_types:share(), pid()}, emqx_types:subopts()}].
+subscriptions() ->
+    MatchHead = #subscription{sub = '$1', opts = '$2', _ = '_'},
+    Contruct = {{'$1', '$2'}},
```
savonarola, line 511:
```suggestion
    Construct = {{'$1', '$2'}},
```
🤔

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_router.erl
keynslug: There's this [measure](https://github.com/emqx/emqx/pull/13819/files#diff-eb9115c096b866c464ff4ad37b615f6f62f9f4dd18980b974679ae27ee1a7400R338-R346). Currently, if any zone's or global MQTT config has `upgrade_qos = true`*, this logic of selective routing is outright disabled. I opted for such an approach, because 1) to have less discrepancies between session implementations, and 2) this option is used very rarely across our customers.

> And as such, `add_route/2` function should decide the "scope" depending on this config and it should ignore client's desire.

I don't think it's a good idea to have router concern itself with MQTT settings, it's session's domain after all. That's why it's currently exposed as an API detail.

---

* Granted, it doesn't currently handle situations where there's no `upgrade_qos = true` configured anywhere but some existing session has `upgrade_qos = true` in the subscription state, but I haven't found a sane way to handle this.

keynslug: > Another question: how does persistent_session router handle change of this setting?

Routes with different scopes are basically separate routes, there's no updates in that sense.

apps/emqx_utils/src/emqx_utils_ets.erl
```diff
@@ -23,13 +23,18 @@
...
+    lookup_value/4
 ]).
 
 -export([keyfoldl/3]).
 
 -export([delete/1]).
 
+-export_type([pat/0]).
+
+-type pat() :: '_' | '$1' | '$2'.
```
savonarola, line 36: we also may make it take an argument to clarify intentions
```
-type matchable(T) :: T | '_' | '$1' | '$2'.
```

apps/emqx/src/emqx_broker.erl
keynslug: I've read the proposal but I think that _channel_ is:
* Already has an established meaning in the `emqx` app, there's `emqx_channel`, `emqx_cm`, etc.
* Something from the communication domain, but in reality communication is happening through the same mechanisms as before.
* Does not map very well to the routing context, upon which most of this functionality is built.

Because of these I decided to go with something closer to the routing context and build on top of them.

On the other hand, agree with `direct` vs `realtime`, will do quick refactoring.

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: If we don't take into account rolling upgrades for a moment, currently having both QoS0 messages in DS and direct channels / scopes is not possible, unless I'm missing something. On the other hand, in the event of a rolling upgrade it's currently possible, I will adjust this.

apps/emqx_utils/src/emqx_utils_ets.erl
keynslug: I actually tried that first, but it looked more heavyweight, so I backed down.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_router.erl
```diff
@@ -83,117 +93,104 @@ init_tables() ->
             ]}
         ]}
     ]),
+    %% NOTE
+    %% Holds non-wildcard routes with any scopes other than the `any` scope.
+    ok = mria:create_table(?PS_ROUTER_EXT_TAB, [
```
savonarola, line 98: Do we need new tables for backward compatibility?

keynslug: Yeah, sort of. I haven't found a good way to structure the existing table so that rolling upgrades are safe (suggestions are welcome).

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_subs.erl
```diff
@@ -85,91 +104,126 @@
...
+        %% This is existing subscription:
+        Sub = #{} ->
+            Mode = get_subscription_mode(Sub),
+            case subscription_mode(TopicFilter, SubOpts, Props) of
+                Mode ->
+                    %% ...And the mode hasn't changed, update it:
+                    S = update_subscription(TopicFilter, SubOpts, Sub, Props, S0);
+                _Different ->
+                    %% ...But the mode is different, recreate it:
+                    S1 = delete_subscription(SessionId, TopicFilter, Sub, S0),
```
savonarola, line 125: With `SubId` changed, how scheduler's `update_stream_subscription_state_ids` will work? Seems it won't be able to update stream's sub states 🤔 Will they become orphaned?

savonarola (commented):
To verify, without `upgrade_qos`, we lose the ability to read qos0 from the past?

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_subs.erl
```diff
@@ -85,91 +104,126 @@
 ) ->
     {ok, emqx_persistent_session_ds_state:t()} | {error, ?RC_QUOTA_EXCEEDED}.
 on_subscribe(TopicFilter, SubOpts, #{id := SessionId, s := S0, props := Props}) ->
-    #{upgrade_qos := UpgradeQoS, max_subscriptions := MaxSubscriptions} = Props,
+    #{max_subscriptions := MaxSubscriptions} = Props,
```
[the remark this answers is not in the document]
keynslug: Shared subscriptions still essentially use [`add_route/2`](https://github.com/emqx/emqx/blob/1324330a2984f77de7091dc44536daa69d5b0cda/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_shared_subs.erl#L183-L185) that puts the route into `any` scope, which means any message will end up in DS regardless of QoS.

This currently has a weird consequence that making a shared subscription could lead to message duplication for unrelated regular subscriptions, a side effect similar to [the one described by Dmitrii](https://github.com/emqx/emqx/pull/13819#discussion_r1777511758).

Of course this needs to be addressed, maybe as part of another PR that enables similar optimization specifically for shared subscriptions.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_subs.erl
keynslug: ~~I believe they will be taken care of by `emqx_persistent_session_ds_subs:gc/1` eventually.~~ Good catch, scheduler needs to be informed about deleted subscriptions.

apps/emqx/src/emqx_persistent_session_ds.erl
keynslug: Good point anyway, need to think more about that. The fact that the user can toggle `upgrade_qos` on and off in the runtime makes things even more complicated.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_router.erl
ieQu1: > There's this [measure](https://github.com/emqx/emqx/pull/13819/files#diff-eb9115c096b866c464ff4ad37b615f6f62f9f4dd18980b974679ae27ee1a7400R338-R346).

1. Can a client call this API with wrong parameters due to bug, race condition or being restored from a very old state?
2. If not, how likely is such bug (or race condition) in the future?
3. If not, what if the session code is modified by someone else?
4. If such situation happen, does _one_ misbehaving client affect _all other_ clients subscribed to the same topic?

apps/emqx/src/emqx_persistent_session_ds.erl
ieQu1: MQTT spec prohibits repeated deliver of QoS0 messages, we should ensure complience to the spec in all situations.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_subs.erl
savonarola: If we create a new subscription, it will obtain new streams (starting from a new point?). This will probably lead (unlike `update_subscription` case) to the loss of continuity of message flow.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_router.erl
ieQu1: P.S. QoS0 handling for shared subscriptions provides a legit case for the /3 API, but shared QoS0 subscription is very finicky question problem in itself. It requires study of MQTT spec.

ieQu1:
> To verify, without upgrade_qos, we lose the ability to read qos0 from the past?

Yes. See https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901230

> QoS 1 and QoS 2 messages pending transmission to the Client and OPTIONALLY QoS 0 messages pending transmission to the Client.

With this change we treat OPTIONALLY as no.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_router.erl
keynslug: > If any answer is yes then this is not a very robust API.

Then there's little to no robust APIs. I'd also like to point that:
1. Currently, `upgrade_qos` is part of the session state.
2. Currently, `upgrade_qos` changes take effect at specific points in time, and that's also part of the session logic.

At the moment I don't see how all this can become part of the router. However, I'll try to fit everything better together so that this API becomes simpler.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_subs.erl
keynslug: Yes. This is currently considered fine, since we expect `upgrade_qos` is rarely used and thus even rarer toggled in runtime. Please share if you think this needs to be reevaluated.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_router.erl
ieQu1: > Currently, upgrade_qos is part of the session state.

Previously, router wasn't QoS-selective. Now it is.

PR #13825, opened 2024-09-18
ci: do not run emqx app tests on release

id:
also switch to using self hosted runners in forks


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4252)

PR #13838, opened 2024-09-19
perf: optimize resource status/state/error cache

zmstone:
Fixes: https://emqx.atlassian.net/browse/EMQX-11308
Release version: v/e5.8.1

## Summary

Prior to this fix, resource manager state data is cached in ets table (via grpoc).
It's quite a bloated map to copy for each and every message to send,
it's especially expensive when there are multiple channels added to one connector.

This PR breaks up the cache into 3 parts.

- Connector `status`, `error` and `config` are cached in ets table `emqx_resource_cache`,
  since `status` and `error` tend to change often, it makes less sense to cache in persistent-term. 
  Also,`status` and `error` are grouped into the same record field of the cached tuple, so it can be looked-up with `ets:lookup_element` to avoid copying the full record (`config` can be large).
- Connector's channels (actions) are cached in the same ets table, but using a per-channel key.
  This allows us to call `ets:lookup_element` to directly get channel status/error instead of copying all the channels in a map then lookup inside the map.
- The relatively stable parts, such as callback module, query mode, and callback state are cached in persistent-term.
  This can avoid copying the terms when processing the queries.

## Test results
```
emqtt_bench pub -s -I 100 -L 0 -h localhost -p 1883 -t p/1 -c 1000 -s 100
```
Same test scenario, comparing to 5.8.0
- Total RAM 500MB -> 300MB.
- Connection process RAM 310KB -> 106KB.

## PR Checklist

- [~] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket

### DIFF
--- a/apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
+++ b/apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
@@ -532,11 +532,11 @@ do_send_msg(sync, KafkaTopic, KafkaMessage, Producers, SyncTimeout) ->
             {error, timeout}
     end;
 do_send_msg(async, KafkaTopic, KafkaMessage, Producers, AsyncReplyFn) ->
-    %% * Must be a batch because wolff:cast2 are batch APIs
+    %% * Must be a batch because wolff send and cast are batch APIs
     %% * Must be a single element batch because wolff books calls, but not batch sizes
     %%   for counters and gauges.
     Batch = [KafkaMessage],
-    {_Partition, Pid} = wolff:cast2(
+    {_Partition, Pid} = wolff:send2(
         Producers, KafkaTopic, Batch, {fun ?MODULE:on_kafka_ack/3, [AsyncReplyFn]}
     ),
     %% This Pid is returned, but not monitored by caller
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -1302,36 +1302,31 @@ extract_connector_id(Id) when is_binary(Id) ->
 %% Check if channel is installed in the connector state.
 %% There is no need to query the conncector if the channel is not
 %% installed as the query will fail anyway.
-pre_query_channel_check(Id, {Id, _} = _Request, ChanSt, QueryOpts) ->
+pre_query_channel_check(Id, {Id, _} = _Request, ChanSt, IsSimpleQuery) ->
     case emqx_resource_manager:channel_status_is_channel_added(ChanSt) of
         true ->
             ok;
         false ->
-            error_if_channel_is_not_installed(Id, QueryOpts)
+            error_if_channel_is_not_installed(Id, IsSimpleQuery)
     end;
-pre_query_channel_check(Id, {Id, _} = _Request, ?NO_CHANNEL, QueryOpts) ->
-    %% a per-channel requst, but channel was not added or deleted (due to reace)
-    error_if_channel_is_not_installed(Id, QueryOpts);
-pre_query_channel_check(_Id, _Request, _ChanSt, _QueryOpts) ->
-    %% Not a per-channel requst
+pre_query_channel_check(_Id, _Request, _ChanSt, _IsSimpleQuery) ->
+    %% Not a per-channel request
     ok.
 
-error_if_channel_is_not_installed(Id, QueryOpts) ->
+error_if_channel_is_not_installed(Id, IsSimpleQuery) ->
     %% Fail with a recoverable error if the channel is not installed and there are buffer
     %% workers involved so that the operation can be retried.  Otherwise, this is
     %% unrecoverable.  It is emqx_resource_manager's responsibility to ensure that the
     %% channel installation is retried.
-    IsSimpleQuery = is_simple_query(QueryOpts),
-    case IsSimpleQuery of
-        true ->
-            {error,
-                {unrecoverable_error,
-                    iolist_to_binary(io_lib:format("channel: \"~s\" not operational", [Id]))}};
-        false ->
-            {error,
-                {recoverable_error,
-                    iolist_to_binary(io_lib:format("channel: \"~s\" not operational", [Id]))}}
-    end.
+    Msg = iolist_to_binary(io_lib:format("channel: \"~s\" not operational", [Id])),
+    ErrorType =
+        case IsSimpleQuery of
+            true ->
+                unrecoverable_error;
+            false ->
+                recoverable_error
+        end,
+    {error, {ErrorType, Msg}}.
 
 is_always_send(#{query_mode := M}, _) when ?IS_BYPASS(M) ->
     %% The query overrides the query mode of the resource, send even in disconnected state
@@ -1393,7 +1388,7 @@ apply_query_fun(
         ?APPLY_RESOURCE(
             call_query,
             begin
-                case pre_query_channel_check(Id, Request, ChanSt, QueryOpts) of
+                case pre_query_channel_check(Id, Request, ChanSt, is_simple_query(QueryOpts)) of
                     ok ->
                         Mod:on_query(extract_connector_id(Id), Request, ResSt);
                     Error ->
@@ -1437,7 +1432,7 @@ apply_query_fun(
             AsyncWorkerMRef = undefined,
             InflightItem = ?INFLIGHT_ITEM(Ref, Query, IsRetriable, AsyncWorkerMRef),
             ok = inflight_append(InflightTID, InflightItem),
-            case pre_query_channel_check(Id, Request, ChanSt, QueryOpts) of
+            case pre_query_channel_check(Id, Request, ChanSt, IsSimpleQuery) of
                 ok ->
                     case
                         Mod:on_query_async(
@@ -1479,7 +1474,9 @@ apply_query_fun(
         ?APPLY_RESOURCE(
             call_batch_query,
             begin
-                case pre_query_channel_check(Id, FirstRequest, ChanSt, QueryOpts) of
+                case
+                    pre_query_channel_check(Id, FirstRequest, ChanSt, is_simple_query(QueryOpts))
+                of
                     ok ->
                         Mod:on_batch_query(extract_connector_id(Id), Requests, ResSt);
                     Error ->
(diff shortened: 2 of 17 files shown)


HJianBo (dismissed):

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1290,29 +1297,23 @@ extract_connector_id(Id) when is_binary(Id) ->
...
+pre_query_channel_check(Id, {Id, _} = _Request, ChanSt, QueryOpts) ->
+    case emqx_resource_manager:channel_status_is_channel_added(ChanSt) of
         true ->
             ok;
         false ->
             error_if_channel_is_not_installed(Id, QueryOpts)
     end;
-pre_query_channel_check({Id, _} = _Request, _Channels, QueryOpts) ->
+pre_query_channel_check(Id, {Id, _} = _Request, ?NO_CHANNEL, QueryOpts) ->
+    %% a per-channel requst, but channel was not added or deleted (due to reace)
```
thalesmg, line 1313:
```suggestion
    %% a per-channel request, but channel was not added or deleted (due to race)
```

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1290,29 +1297,23 @@ extract_connector_id(Id) when is_binary(Id) ->
...
     error_if_channel_is_not_installed(Id, QueryOpts);
-pre_query_channel_check(_Request, _Channels, _QueryOpts) ->
+pre_query_channel_check(_Id, _Request, _ChanSt, _QueryOpts) ->
+    %% Not a per-channel requst
```
thalesmg, line 1316:
```suggestion
    %% Not a per-channel request
```

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1321,44 +1322,33 @@ error_if_channel_is_not_installed(Id, QueryOpts) ->
     %% unrecoverable.  It is emqx_resource_manager's responsibility to ensure that the
     %% channel installation is retried.
     IsSimpleQuery = is_simple_query(QueryOpts),
-    case is_channel_id(Id) of
-        true when IsSimpleQuery ->
+    case IsSimpleQuery of
```
thalesmg, line 1325: It seems like this case clause can be reduced to just the error.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1290,29 +1297,23 @@ extract_connector_id(Id) when is_binary(Id) ->
...
-    case emqx_resource_manager:channel_status_is_channel_added(ChannelStatus) of
+pre_query_channel_check(Id, {Id, _} = _Request, ChanSt, QueryOpts) ->
+    case emqx_resource_manager:channel_status_is_channel_added(ChanSt) of
         true ->
             ok;
         false ->
             error_if_channel_is_not_installed(Id, QueryOpts)
     end;
-pre_query_channel_check({Id, _} = _Request, _Channels, QueryOpts) ->
+pre_query_channel_check(Id, {Id, _} = _Request, ?NO_CHANNEL, QueryOpts) ->
```
thalesmg, line 1312: This clause should be before the previous, since it's more specific on `ChanSt`

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1205,27 +1205,34 @@ handle_async_worker_down(Data0, Pid) ->
 -spec call_query(force_sync | async_if_possible, _, _, _, _, _) -> _.
 call_query(QM, Id, Index, Ref, Query, QueryOpts) ->
     ?tp(call_query_enter, #{id => Id, query => Query, query_mode => QM}),
-    case emqx_resource_manager:lookup_cached(extract_connector_id(Id)) of
+    case emqx_resource_cache:get_runtime(Id) of
```
thalesmg, line 1208: Since `emqx_resource_cache:get_runtime/1` apparently expects only channel IDs, won't this prevent calling the connector with special queries, like calling postgres with `{query, "select * from ..."}`?

apps/emqx_resource/src/emqx_resource_cache.erl
```diff
@@ -0,0 +1,331 @@
...
+    id :: binary(),
+    group :: binary(),
+    manager_pid :: pid(),
+    st_err :: st_err(),
+    config :: term(),
+    cb :: term(),
+    extra = []
+}).
+
+-type chan_key() :: {ConnectorId :: binary(), ChannelID :: binary()}.
```
thalesmg, line 43: Nit:

```suggestion
-type chan_key() :: {connector_res_id(), action_res_id() | source_res_id()}.
```

(maybe we could redefine the existing `channel_id` to `action_res_id() | source_res_id()`)

apps/emqx_resource/src/emqx_resource_cache.erl
```diff
@@ -0,0 +1,331 @@
...
+    lists:foldl(
+        fun(
+            #channel{
+                id = {ConnectorId, ChannelId},
+                status = Status,
+                error = Error
+            },
+            Acc
+        ) ->
+            Key = iolist_to_binary([ChannelId, ":", ConnectorId]),
```
thalesmg, line 323: Q: do we need to append the connector id here, since we're already searching for channels from a specific connector?
Maybe we could only construct it like this in `make_resource_data`, and `erase_channel` could work with tuples/iolists.

apps/emqx_resource/src/emqx_resource_cache.erl
```diff
@@ -0,0 +1,331 @@
...
+            not_found;
+        #{mod := Mod} ->
+            {ok, Mod}
+    end.
+
+get_cb(ID) ->
+    case get_cb_pt(ID) of
+        undefined ->
+            %% maybe it's a dryrun connector
+            ets:lookup_element(?CACHE, ID, #connector.cb, undefined);
```
thalesmg, line 150: Nit:

```suggestion
            ets:lookup_element(?CACHE, ID, #connector.cb, not_found);
```

apps/emqx_resource/src/emqx_resource_cache.erl
```diff
@@ -0,0 +1,331 @@
...
+    ?NO_CHANNEL;
+get_channel_status(ChanKey) ->
+    ets:lookup_element(?CACHE, ChanKey, #channel.status, ?NO_CHANNEL).
+
+get_cb_pt(ID) ->
+    try
+        persistent_term:get(?CB_PT_KEY(ID))
+    catch
+        error:badarg ->
+            undefined
```
thalesmg, line 225: Nit:

```suggestion
            not_found
```

apps/emqx_resource/src/emqx_resource_cache_cleaner.erl
```diff
@@ -0,0 +1,78 @@
...
+                emqx_pmon:erase(Pid, Pmon);
+            error ->
+                Pmon
+        end,
+    {noreply, State#{pmon => NewPmon}};
+handle_info(_Info, State) ->
+    {noreply, State}.
+
+terminate(_Reason, _State) ->
+    ok.
```
thalesmg, line 65: Since there's no cleanup, we don't need `trap_exit`?

apps/emqx_resource/src/emqx_resource_cache_cleaner.erl
```diff
@@ -0,0 +1,78 @@
...
+
+code_change(_OldVsn, State, _Extra) ->
+    {ok, State}.
```
thalesmg, line 68: :scissors: ?

apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
```diff
@@ -532,11 +532,11 @@ do_send_msg(sync, KafkaTopic, KafkaMessage, Producers, SyncTimeout) ->
             {error, timeout}
     end;
 do_send_msg(async, KafkaTopic, KafkaMessage, Producers, AsyncReplyFn) ->
-    %% * Must be a batch because wolff:cast2 are batch APIs
+    %% * Must be a batch because wolff send and cast are batch APIs
     %% * Must be a single element batch because wolff books calls, but not batch sizes
     %%   for counters and gauges.
     Batch = [KafkaMessage],
-    {_Partition, Pid} = wolff:cast2(
+    {_Partition, Pid} = wolff:send2(
```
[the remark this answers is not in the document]
zmstone: it's still async, blocked until the message is queued.
the sync API is `send_sync`

apps/emqx_resource/src/emqx_resource_cache.erl
zmstone: this is intended to be `undefined` to align with the type spec and other clauses

apps/emqx_resource/src/emqx_resource_cache.erl
zmstone: maybe, but `erase_old_channels` takes the external format chain IDs.

apps/emqx_resource/src/emqx_resource_cache_cleaner.erl
zmstone: I tend to always add trap_exist to avoid surprises in the future.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
zmstone: this clause is not needed.

apps/emqx_resource/src/emqx_resource_cache.erl
thalesmg: that field seems to be just `term()` :thinking: 

it's just a minor nit, since everywhere else in this module we return `not_found` on such occasions

apps/emqx_resource/src/emqx_resource_cache.erl
thalesmg: exactly, erase_old_channels could handle tuples/iolists.

thalesmg (approved):

apps/emqx_resource/src/emqx_resource_cache.erl
zmstone: sorry, i meant default value.
but i will push a new commit to use `no_state` instead.

apps/emqx_resource/src/emqx_resource_cache.erl
zmstone: I mean this is externalized 
https://github.com/emqx/emqx/pull/13838/files#diff-67a1a968ee134d34263236a53df607c87f2688b085300b25dee20b423b9f2e5dR104

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4267)

PR #14004, opened 2024-10-16
fix(cluster-link): correctly replicate overlapping route intersections

keynslug:
Fixes [EMQX-13229](https://emqx.atlassian.net/browse/EMQX-13229)

Release version: e5.8.2

## Summary

Cluster Linking involves route replication and allows to configure a _set of topic filters_ linked cluster is interested in. This set of topic filters essentially represents a subset of topic space, and it's currently infeasible to always have exact and minimal representation of such set: 1) a topic filter in this set can be a "subset" of another topic filter, or 2) a pair of topic filters can be overlapping.

For the first case this PR introduces config validation that rejects such redundant topic filters.

The second case is worse though: if route intersects with 2 topic filters that are overlapping, we may have to replicate this route 2 times, in cases where these intersections are themselves overlapping. Before these changes, the route replication assumed that it's enough to replicate just a single non-empty intersection. Usually the first one would be replicated: e.g. with `t/+` subscription and `[t/1, t/#]` link filters, only `t/1` would be the only replicated route.

This PR also refactors `emqx_retainer` slightly, because the responsibility of filtering out special topics became split between the backend and the dispatcher after `emqx_topic:match/2` refinements. This responsibility is now exclusively in the backend, where it's also cheaper to do.

## PR Checklist

[EMQX-13229]: https://emqx.atlassian.net/browse/EMQX-13229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/emqx/src/emqx_topic.erl
```diff
@@ -150,6 +168,29 @@ intersect_join(W, Words) -> [W | Words].
 wildcard_intersection(W, W) -> W;
 wildcard_intersection(_, _) -> '+'.
 
+%% @doc Finds out if topic / topic filter T1 is a subset of topic / topic filter T2.
+-spec subset(topic() | words(), topic() | words()) -> boolean().
```
zmstone, line 172: is_subset ?

zmstone (approved):

savonarola:
Just curious: what is the summary of the original logic problem? I.e. from the first view it's unclear why overlapping routes prevented messages from being published (as the related issue states).

keynslug:
> Just curious: what is the summary of the original logic problem?

Before these changes, the route replication assumed that it's enough to replicate just a single non-empty intersection. Usually the first one would be replicated: e.g. with `t/+` subscription and `[t/1, t/#]` link filters, only `t/1` would be the only replicated route.

Added this into the PR description.

apps/emqx_retainer/test/emqx_retainer_SUITE.erl
savonarola, line 1: NB: This test is also used by fdb retainer

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -381,12 +381,24 @@ search_stream(Tokens, Now) ->
...
     Ms = make_message_match_spec(Tokens, Now),
-    emqx_utils_stream:ets(
+    MsgStream = emqx_utils_stream:ets(
         fun
             (undefined) -> ets:select(?TAB_MESSAGE, Ms, 1);
             (Cont) -> ets:select(Cont)
         end
-    );
+    ),
+    case Tokens of
```
savonarola, line 390: Maybe a corresponding update is needed for the fdb retainer

savonarola (approved):

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
keynslug: Will take care of that once the next sync PR is merged.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4775)

PR #14008, opened 2024-10-16
fix(connector aggregator): use `{continue, _}` instead of 0 timeout

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13302

Release version: e5.8.2

## Problem

1) First allocated buffer gets 3 records, its delivery is started and completes just fine.
2) As the 3rd record from the first batch is pushed, it triggers the allocation of a
   second buffer.
3) After delivery is completed, 3 more records are pushed into this 2nd buffer.
4) As the 3rd record from the second batch is pushed, it triggers the allocation of a
   3rd buffer.  This time, however, no delivery process is started.
5) Seconds later, probably due to a ~tick~ timeout, delivery of this 3rd, empty buffer is
   started and completed, indeed reporting to be empty.  Delivery of the 2nd buffer is
   never started.

Albeit the causing sequence of events causing the race condition is not yet exactly determined,
this is an attempt to avoid the possibility of other events being processed before an
enqueued delivery is started.

Examples:

- https://github.com/emqx/emqx/actions/runs/11365721751/job/31614944950#step:5:510
  - [logs-emqx-enterprise-apps_emqx_bridge_snowflake-sg1_1.zip](https://github.com/user-attachments/files/17411909/logs-emqx-enterprise-apps_emqx_bridge_snowflake-sg1_1.zip)

- https://github.com/emqx/emqx/actions/runs/11367014664?pr=14008
  - [logs-emqx-enterprise-apps_emqx_bridge_snowflake-sg1_1_b.zip](https://github.com/user-attachments/files/17411914/logs-emqx-enterprise-apps_emqx_bridge_snowflake-sg1_1_b.zip)

### Hypothesis

1) 0 timeout is aborted if any request or message arrives concurrently.
2) Last pushed record triggers `cast {rotate_buffer, FD}`, which places buffer in
   `#st.queued` and allocates a fresh empty buffer and sets it as current.
3) Concurrently, some request or message arrives, cancelling the 0-timeout that would
   enqueue the delivery of the full buffer.
4) After `tick` timeouts, it triggers `cast {close_buffer, Timestamp}`, which processes
   only the current empty buffer.

## PR Checklist

### DIFF
--- a/apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
+++ b/apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
@@ -231,19 +231,12 @@ terminate(_Reason, #st{name = Name}) ->
 %%
 
 handle_next_buffer(Timestamp, St = #st{buffer = #buffer{until = Until}}) when Timestamp < Until ->
-    ?tp(connector_aggreg_handle_next_buffer_too_soon, #{
-        timestamp => Timestamp, buffer => St#st.buffer
-    }),
     St;
 handle_next_buffer(Timestamp, St0 = #st{buffer = Buffer = #buffer{since = PrevSince}}) ->
-    ?tp(connector_aggreg_handle_next_buffer, #{timestamp => Timestamp, buffer => Buffer}),
     BufferClosed = close_buffer(Buffer),
     St = enqueue_closed_buffer(BufferClosed, St0),
     handle_next_buffer(Timestamp, PrevSince, St);
 handle_next_buffer(Timestamp, St = #st{buffer = undefined}) ->
-    ?tp(connector_aggreg_handle_next_buffer_undefined, #{
-        timestamp => Timestamp, buffer => St#st.buffer
-    }),
     handle_next_buffer(Timestamp, Timestamp, St).
 
 handle_next_buffer(Timestamp, PrevSince, St0) ->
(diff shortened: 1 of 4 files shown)


apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -186,10 +191,10 @@ init(St0 = #st{name = Name}) ->
 
 handle_call({next_buffer, Timestamp}, _From, St0) ->
     St = #st{buffer = Buffer} = handle_next_buffer(Timestamp, St0),
-    {reply, Buffer, St, 0};
+    {reply, Buffer, St, {continue, enqueue_delivery}};
```
keynslug, line 194: If I remember the original idea correctly, this 0 timeout is here to serve 2 purposes. The assumption is that if there's a `{next_buffer, ...}` call, high chance that there are actually _a lot_ of concurrent `{next_buffer, ...}` calls (same with `{rotate_buffer, ...}`, etc.).
1. If there are, make sure that all of them are processed and replied to first, only then start delivery (which is moderately expensive).
2. At the same time, do not overwhelm the current process with a lot of `enqueue_delivery` requests.

However you're absolutely right that this mechanism is way too brittle.

That said, what do you think about doing `gen_server:cast/2` to `self()` in the first clause of `enqueue_closed_buffer/2`? This IIUC should mostly preserve the current mechanics.

keynslug (commented):
Good find, as usual! Should keep reminding myself that putting important logic behind gen_server timeouts is rarely a good idea. 🫠

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
thalesmg: Sure, I think that achieves the same goal.  I've pushed a patch for that.

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -227,12 +234,19 @@ terminate(_Reason, #st{name = Name}) ->
 %%
 
 handle_next_buffer(Timestamp, St = #st{buffer = #buffer{until = Until}}) when Timestamp < Until ->
+    ?tp(connector_aggreg_handle_next_buffer_too_soon, #{
+        timestamp => Timestamp, buffer => St#st.buffer
+    }),
```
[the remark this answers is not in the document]
thalesmg: These are not used in any tests.  I peppered them to try and find the sequence of events that leads to the original bug.  But apparently just by doing that it became less frequent.  :see_no_evil: 

---

Well, once the first `next_buffer` rotates the buffer, the other ones are "too soon", in a way. :speak_no_evil: 

But I'll remove the debugging TPs, anyway.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4589)

PR #14023, opened 2024-10-18
fix(dashboard monitor): don't sum data for local aggregation

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13262

Release version: v/e5.8.2

## Summary

When compacting or downsampling data for local node only, since data points are repeated snapshots until a change is recorded, we need to keep the latest values instead of summing them up.

#### Without patch

https://github.com/user-attachments/assets/3e3152ed-0575-46e4-ba97-abb0b2ccec13

#### With patch

https://github.com/user-attachments/assets/f899b21d-d36b-4d0a-bbd0-86657f3af249

## PR Checklist

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_monitor.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_monitor.erl
@@ -214,7 +214,7 @@ all_data() ->
 inplace_downsample() ->
     All = all_data(),
     Now = erlang:system_time(millisecond),
-    Compacted = compact(Now, All, []),
+    Compacted = compact(Now, All),
     {Deletes, Writes} = compare(All, Compacted, [], []),
     {atomic, ok} = mria:transaction(
         mria:local_content_shard(),
@@ -255,18 +255,23 @@ compare([{T0, _} | _] = All, [{T1, Data1} | Compacted], Deletes, Writes) when T0
     compare(All, Compacted, Deletes, [{T1, Data1} | Writes]).
 
 %% compact the data points to a smaller set of buckets
-compact(_Now, [], Acc) ->
+%% Pre-condition: data fed to this function must be sorted chronologically.
+compact(Now, Data) ->
+    Gauges = gauges(),
+    compact(Now, Data, Gauges, []).
+
+compact(_Now, [], _Gauges, Acc) ->
     lists:reverse(Acc);
-compact(Now, [{Time, Data} | Rest], Acc) ->
+compact(Now, [{Time, Data} | Rest], Gauges, Acc) ->
     Interval = sample_interval(Now - Time),
     Bucket = round_down(Time, Interval),
-    NewAcc = merge_to_bucket(Bucket, Data, Acc),
-    compact(Now, Rest, NewAcc).
+    NewAcc = merge_to_bucket(Bucket, Data, Gauges, Acc),
+    compact(Now, Rest, Gauges, NewAcc).
 
-merge_to_bucket(Bucket, Data, [{Bucket, Data0} | Acc]) ->
-    NewData = merge_sampler_maps(Data, Data0),
+merge_to_bucket(Bucket, Data, Gauges, [{Bucket, Data0} | Acc]) ->
+    NewData = merge_local_sampler_maps(Data0, Data, Gauges),
     [{Bucket, NewData} | Acc];
-merge_to_bucket(Bucket, Data, Acc) ->
+merge_to_bucket(Bucket, Data, _Gauges, Acc) ->
     [{Bucket, Data} | Acc].
 
 %% for testing
@@ -277,6 +282,7 @@ randomize(Count, Data) when is_map(Data) ->
 randomize(Count, Data, Age) when is_map(Data) andalso is_integer(Age) ->
     Now = erlang:system_time(millisecond) - 1,
     StartTs = Now - Age,
+    Gauges = gauges(),
     lists:foreach(
         fun(_) ->
             Ts = round_down(StartTs + rand:uniform(Age), timer:seconds(10)),
@@ -285,7 +291,7 @@ randomize(Count, Data, Age) when is_map(Data) andalso is_integer(Age) ->
                 [] ->
                     store(Record);
                 [#emqx_monit{data = D} = R] ->
-                    store(R#emqx_monit{data = merge_sampler_maps(D, Data)})
+                    store(R#emqx_monit{data = merge_local_sampler_maps(Data, D, Gauges)})
             end
         end,
         lists:seq(1, Count)
@@ -370,6 +376,7 @@ merge_sampler_maps(M1, M2) when is_map(M1) andalso is_map(M2) ->
     Fun = fun(Key, Acc) -> merge_values(Key, M1, Acc) end,
     lists:foldl(Fun, M2, ?SAMPLER_LIST).
 
+%% `M1' is assumed to be newer data compared to anything `M2' has seen.
 merge_local_sampler_maps(M1, M2, Gauges) when is_map(M1) andalso is_map(M2) ->
     Fun = fun(Key, Acc) -> merge_local_values(Key, M1, Acc, Gauges) end,
     lists:foldl(Fun, M2, ?SAMPLER_LIST).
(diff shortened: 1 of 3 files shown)


apps/emqx_dashboard/test/emqx_dashboard_monitor_SUITE.erl
```diff
@@ -198,8 +195,6 @@ t_inplace_downsample(_Config) ->
     ok = gen_server:call(emqx_dashboard_monitor, dummy, infinity),
     All1 = emqx_dashboard_monitor:all_data(),
     All = drop_dummy_data_points(All1),
-    AllSent = lists:map(fun({_, #{sent := S}}) -> S end, All),
-    ?assertEqual(Total, lists:sum(AllSent)),
```
thalesmg, line 202: We shouldn't assert anything about the sum, because these are not individual data points, but "snapshots" of the counters at given points in time.

apps/emqx_dashboard/src/emqx_dashboard_monitor.erl
```diff
@@ -263,9 +263,9 @@ compact(Now, [{Time, Data} | Rest], Acc) ->
     NewAcc = merge_to_bucket(Bucket, Data, Acc),
     compact(Now, Rest, NewAcc).
 
-merge_to_bucket(Bucket, Data, [{Bucket, Data0} | Acc]) ->
-    NewData = merge_sampler_maps(Data, Data0),
-    [{Bucket, NewData} | Acc];
+%% Assumes input data is sorted chronologically.
```
thalesmg, line 266: Table type is `set`, but it seems to be ordered when it reaches here...  Does it maintain insertion order?  Or is it just luck that `ets:select` returns thinks in the correct order?  :thinking:

apps/emqx_dashboard/test/emqx_dashboard_monitor_SUITE.erl
```diff
@@ -660,19 +655,142 @@ t_persistent_session_stats(Config) ->
 
     ok.
 
+%% Checks that we get consistent data when changing the requested time window for
+%% `/monitor'.
+t_smoke_test_monitor_multiple_windows(Config) ->
```
thalesmg, line 660: This tests seems to capture well the issue without the patch: if you check the connection count on each `?retry`, it continually increases and reaches ~ 28 connections before failing.

apps/emqx_dashboard/src/emqx_dashboard_monitor.erl
zmstone: it's sorted by lists:keysort

zmstone:
> When compacting or downsampling data for local node only, since data points are repeated snapshots until a change is recorded, we need to keep the latest values instead of summing them up.

Maybe we just need a local merge function for gauges like this (for counters, it's delta stored in db):
https://github.com/emqx/emqx/blob/2bf8ee62368fc8c4f25a23bc382a04b72b6ee793/apps/emqx_dashboard/src/emqx_dashboard_monitor.erl#L374-L381

changes/ce/fix-14023.en.md
```diff
@@ -0,0 +1 @@
+Fixed an issue with the `GET /monitor` HTTP API where returned data could return values greater than reality depending on the requested time window.
```
zmstone, line 1: list the gauges:
```
      disconnected_durable_sessions,
      subscriptions_durable,
      subscriptions,
      topics,
      connections,
      live_connections
```

zmstone, line 1: it only happens to data points older than 1h ?
```suggestion
Fixed an issue with the `GET /monitor` HTTP API where returned data could return values greater than reality for sampled metrics older than 1 hour.
```

thalesmg: No, it happens to fresh points as well, as shown in the video.

apps/emqx_dashboard/src/emqx_dashboard_monitor.erl
thalesmg: Not in all places: https://github.com/emqx/emqx/blob/e378169e5227f29b3b838cbceb0ef3eb131ee5da/apps/emqx_dashboard/src/emqx_dashboard_monitor.erl#L322-L325

I assume this place would also need this sort.

apps/emqx_dashboard/src/emqx_dashboard_monitor.erl
```diff
@@ -380,6 +385,12 @@ merge_values(disconnected_durable_sessions, M1, M2) ->
 merge_values(Key, M1, M2) ->
     sum_values(Key, M1, M2).
 
+merge_local_values(Key, M1, M2, Gauges) when is_map_key(Key, Gauges) ->
+    %% First argument is assumed to be from a newer timestamp, so we keep the latest.
+    M2#{Key => maps:get(Key, M1, maps:get(Key, M2, 0))};
```
zmstone, line 390: call `max_values` ?

zmstone: and max_values needs a fix.
it always return M2 if key is not found in both M1 and M2.

changes/ce/fix-14023.en.md
zmstone: maybe we should make it clear in change log:
the data points older than 1h is permanently altered, the data points within 1h is only displayed wrong.

zmstone:
this could be rather confusing users, when made the changes, I only tested with counters, forgot about the gauges.
please help to add this to known issues of 5.8.1 https://github.com/emqx/emqx-docs/blob/release-5.8/en_US/changes/known-issues-5.8.md

apps/emqx_dashboard/src/emqx_dashboard_monitor.erl
thalesmg: it's not the same: we're taking the latest value, not the maximum

apps/emqx_dashboard/src/emqx_dashboard_monitor.erl
```diff
@@ -319,10 +319,10 @@ do_sample(Node, Time) when is_integer(Time) ->
 
 do_sample_local(Time) ->
     MS = ets:fun2ms(fun(#emqx_monit{time = T} = A) when T >= Time -> A end),
-    FromDB = ets:select(?TAB, MS),
+    FromDB = lists:keysort(#emqx_monit.time, ets:select(?TAB, MS)),
```
zmstone, line 322: sorry, i was maybe wrong. 
there is actually no need to sort after all ? it turns into a map anyway.

thalesmg: d'oh, you're right.   
sorted in the wrong spot :see_no_evil:

changes/ce/fix-14023.en.md
thalesmg: :checkered_flag:

thalesmg: :check

apps/emqx_dashboard/src/emqx_dashboard_monitor.erl
thalesmg: I already added a sort in `ts_list/1`.

thalesmg:
> this could be rather confusing users, when made the changes, I only tested with counters, forgot about the gauges. please help to add this to known issues of 5.8.1 https://github.com/emqx/emqx-docs/blob/release-5.8/en_US/changes/known-issues-5.8.md

:checkered_flag: 

https://github.com/emqx/emqx-docs/pull/2700

apps/emqx_dashboard/test/emqx_dashboard_monitor_SUITE.erl
```diff
@@ -249,23 +244,52 @@ t_downsample_1h(_Config) ->
...
 test_downsample(MaxAge, DataPoints) ->
+    ok = pause_monitor_process(),
     Now = erlang:system_time(millisecond) - 1,
     Interval = emqx_dashboard_monitor:sample_interval(MaxAge),
     StartTs = round_down(Now - MaxAge, Interval),
-    ok = emqx_dashboard_monitor:clean(0),
+    clean_data(),
     %% insert the start mark for deterministic test boundary
-    ok = write(StartTs, sent_1()),
-    ok = insert_data_points(DataPoints - 1, StartTs, Now),
```
thalesmg, line 264: This was a source of flakiness: `Now` could actually change the `sample_interval` together with the computed `StartTs` from the expected `Interval`, making "strange" timestamps appear.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4776)

PR #14053, opened 2024-10-22
fix(ds-lts): respect special topic matching rules in LTS-based layouts

keynslug:
Release version: v5.8.2

## Summary

There are a couple of advantages doing it on such a low level:
1. It's possible to avoid scanning whole streams.
2. Topic is already matched against (compressed) topic filter on the layout level anyway, so it would be wasteful to add one more step of matching and filtering on higher levels.

## PR Checklist

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds_lts.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_lts.erl
@@ -76,13 +76,14 @@
 %% Fixed size binary or integer, depending on the options:
 -type static_key() :: non_neg_integer() | binary().
 
-%% Trie root:
+%% Trie roots:
 -define(PREFIX, prefix).
+-define(PREFIX_SPECIAL, special).
 %% Special prefix root for reverse lookups:
 -define(rlookup, rlookup).
 -define(rlookup(STATIC), {?rlookup, STATIC}).
 
--type state() :: static_key() | ?PREFIX.
+-type state() :: static_key() | ?PREFIX | ?PREFIX_SPECIAL.
 
 -type varying() :: [level() | ?PLUS].
 
@@ -229,6 +230,11 @@ trie_copy_learned_paths(OldTrie, NewTrie) ->
 
 %% @doc Lookup the topic key. Create a new one, if not found.
 -spec topic_key(trie(), threshold_fun(), [level()]) -> msg_storage_key().
+topic_key(Trie, ThresholdFun, [<<"$", _/bytes>> | _] = Tokens) ->
+    %% [MQTT-4.7.2-1]
+    %% Put any topic starting with `$' into a separate root so they won't match
+    %% with `#' / `+/...' subscriptions.
+    do_topic_key(Trie, ThresholdFun, 0, ?PREFIX_SPECIAL, Tokens, [], []);
 topic_key(Trie, ThresholdFun, Tokens) ->
     do_topic_key(Trie, ThresholdFun, 0, ?PREFIX, Tokens, [], []).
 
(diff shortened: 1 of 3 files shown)


apps/emqx_durable_storage/test/emqx_ds_storage_layout_SUITE.erl
```diff
@@ -292,6 +299,28 @@ t_replay(Config) ->
     ?assert(check(?SHARD, <<"#">>, 0, Messages)),
     ok.
 
+t_replay_sys(_Config) ->
```
ieQu1, line 302: What property does this testcase try to verify? Why do I need to care about this property? Please add a comment answering these questions.

ieQu1:
My take on this:
1) Either SYS messages should never end up in the storage
2) Or LTS trie should have a different root prefix for them

apps/emqx_durable_storage/src/emqx_ds_lts.erl
```diff
@@ -229,6 +230,11 @@ trie_copy_learned_paths(OldTrie, NewTrie) ->
 
 %% @doc Lookup the topic key. Create a new one, if not found.
 -spec topic_key(trie(), threshold_fun(), [level()]) -> msg_storage_key().
+topic_key(Trie, ThresholdFun, [<<"$", _/bytes>> | _] = Tokens) ->
+    %% [MQTT-4.7.2-1]
+    %% Put any topic starting with `$' into a separate root so they won't match
+    %% with `#' / `+/...' subscriptions.
+    do_topic_key(Trie, ThresholdFun, 0, ?PREFIX_SPECIAL, Tokens, [], []);
```
ieQu1, line 237:
```suggestion
topic_key(Trie, ThresholdFun, [<<"$", Root/bytes>> | _] = Tokens) ->
    %% [MQTT-4.7.2-1]
    %% Put any topic starting with `$' into a separate root so they won't match
    %% with `#' / `+/...' subscriptions.
    do_topic_key(Trie, ThresholdFun, 0, ?PREFIX_SPECIAL(Root), Tokens, [], []);
```

?

ieQu1: ...Otherwise topic with different root will be treated the same; it's a bug.

ieQu1 (changes requested):
Current implementation will not distinguish between `$SYS` and `$SOMETHING_ELSE`. Maybe we don't have `$SOMETHING_ELSE` now, but we may in the future.

apps/emqx_durable_storage/src/emqx_ds_lts.erl
keynslug: Separate roots idea sounds fine. But, just to clarify:

> Otherwise topic with different root will be treated the same; it's a bug.

I don't see why. Currently, we don't throw away the first token here. Are you thinking about situations where this `special` root starts to have `+` edge?

One thing with separate root for each `$token` idea is that thresholds and `threshold_fun` will have subtly different meaning, which complicates understanding a bit.

keynslug: So I haven't managed to catch the bug you've described with 9e67f1f1e054166a91361690db006f2009c3622d, but changed to _multiple separate roots_ approach anyway in 404b198702da513c299e79b9d92b5d54be4697c7.

ieQu1: > One thing with separate root for each $token idea is that thresholds and threshold_fun will have subtly different meaning, which complicates understanding a bit.

Using thresholds (optional hints for **optimizing** the storage) to implement features that affect **consistency** is not the right way.

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6653)

PR #14090, opened 2024-10-28
fix: don't crash retainer api when retainer is disabled

zhongwencool:
fix: https://askemq.com/t/topic/9875

Release version: v/e5.8.2

image

## PR Checklist

### DIFF
--- a/apps/emqx_retainer/src/emqx_retainer.erl
+++ b/apps/emqx_retainer/src/emqx_retainer.erl
@@ -326,28 +326,43 @@ dispatch(Context, Topic) ->
 
 -spec delete_message(context(), topic()) -> ok.
 delete_message(Context, Topic) ->
-    Mod = backend_module(Context),
-    BackendState = backend_state(Context),
-    Mod:delete_message(BackendState, Topic).
+    with_backend_module(
+        Context,
+        fun(Mod, BackendState) ->
+            Mod:delete_message(BackendState, Topic)
+        end,
+        ok
+    ).
 
 -spec read_message(context(), topic()) -> {ok, list(message())}.
 read_message(Context, Topic) ->
-    Mod = backend_module(Context),
-    BackendState = backend_state(Context),
-    Mod:read_message(BackendState, Topic).
+    with_backend_module(
+        Context,
+        fun(Mod, BackendState) -> Mod:read_message(BackendState, Topic) end,
+        {ok, []}
+    ).
 
 -spec page_read(context(), emqx_maybe:t(topic()), deadline(), non_neg_integer(), non_neg_integer()) ->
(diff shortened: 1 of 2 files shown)


HJianBo (approved):

apps/emqx_retainer/src/emqx_retainer.erl
```diff
@@ -326,28 +326,43 @@ dispatch(Context, Topic) ->
...
 delete_message(Context, Topic) ->
-    Mod = backend_module(Context),
-    BackendState = backend_state(Context),
-    Mod:delete_message(BackendState, Topic).
+    with_backend_module(
+        Context,
+        fun(Mod, BackendState) ->
+            Mod:delete_message(BackendState, Topic)
+        end,
+        ok
```
keynslug, line 334: I don't think that it's a good idea to silently return `ok` if nothing actually happens when the user expects something to happen (e.g. destructive or otherwise mutating operation).

I would say that it's the API that should be aware of the fact that retainer can be disabled, and tell the user about that explicitly.

zhongwencool: The setting of this API is that even if it is disabled, the corresponding message needs to be queried. So I think if it is disabled, it should also support deletion.

There is only one case where it is undefined.
That is, when enable=false is configured, there is no context in init when it starts. In other cases, it is necessary to support CURD message even when enable=false is enabled.

keynslug: > it is necessary to support CURD message even when enable=false .

In this case it's necessary for the API to be explicit and respond to the user with a client error saying that retainer is disabled.

zhongwencool: I believe the design is intended to ensure that all existing messages can be operated on, even if the 'enable' status is set to false.

So that use can delete the message from dashboard.

keynslug: Thanks. I see your point clearer. But the merged solution does not provides the user with that IIUC. The user clicks on "delete" button, API says 200 OK even if nothing has actually happened, and the message is still there. I'd expect this will be reported as a bug.

Correct me if I'm wrong, but to work the way you imagine we need to:
1. Always initialize backend context.
2. If the user disables retainer, just kill hooks, but leave the process with context and the infrastructure (e.g. GC) as is.

zhongwencool: > Always initialize backend context. 
Initializing with default build-in-db even if  user don't set backend. it that ok ?

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4868)

PR #14106, opened 2024-10-29
fix(kafka consumer): ensure no repeated topics in the same connector

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13389

Currently, brod treats a consumer process to a specific topic as a singleton (per client id / connector), meaning that the first subscriber to a given topic will define the consumer options for all other consumers, and those options persist even after the original consumer group is terminated.  We enforce that, if the user wants to consume multiple times from the same topic, then they must create a different connector.

Release version: e5.8.2

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_consumer.erl
+++ b/apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_consumer.erl
@@ -251,7 +251,7 @@ on_add_channel(ConnectorResId, ConnectorState0, SourceResId, SourceConfig) ->
         kafka_client_id := ClientID,
         installed_sources := InstalledSources0
     } = ConnectorState0,
-    case start_consumer(SourceConfig, ConnectorResId, SourceResId, ClientID) of
+    case start_consumer(SourceConfig, ConnectorResId, SourceResId, ClientID, ConnectorState0) of
         {ok, SourceState} ->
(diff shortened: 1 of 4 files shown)


apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_consumer.erl
```diff
@@ -494,6 +496,30 @@ start_consumer(Config, ConnectorResId, SourceResId, ClientID) ->
...
+            %% all new topics
+            ok;
+        NewTopics ->
+            ExistingTopics0 = KafkaTopics -- NewTopics,
+            ExistingTopics = lists:join(<<", ">>, ExistingTopics0),
+            Message = iolist_to_binary([
+                <<"Topics ">>,
+                ExistingTopics,
+                <<" already exist in other sources associated with this connector.">>,
+                <<" If you want to repeat topics, create new connector and source(s).">>
```
zmstone, line 518: this is when creating a new source, right? maybe this:

To consume from the same topic in different sources, please add a new connector.

thalesmg: It might happen when updating an existing source, too.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-4877)

PR #14113, opened 2024-10-30
fix(machine boot): serialize start/stop apps

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-11702

Release version: v/e5.9.0

## Summary

The original issue was that a join request (`ekka:join/1`) was issued while the node was
still starting all its applications.  When joining a cluster, mria calls its `stop`
callback before stopping itself.  However, previous to this patch, that callback is empty
when the node has not yet started.  Hence, mria stopped itself while application were
being started.  When one of these attempted to create its tables, it crashed the whole
node:

```
{"Kernel pid terminated",application_controller,"{application_start_failure,emqx_machine,{{shutdown,{failed_to_start_child,emqx_machine_boot,{'EXIT',{{failed_to_start_app,emqx_gcp_device,{emqx_gcp_device,{bad_return,{{emqx_gcp_device_app,start,[normal,[]]},{'EXIT',{noproc,{gen_server,call,[mria_schema,{create_table,emqx_gcp_device,[...
```

To fix this, here we introduce a process to serialize starting/stopping all applications,
and also register the stop mria callback sooner.  This process starts all applications in
an uninterruptible fashion, meaning that the startup sequence must complete before it's
stopped by the join request.

## PR Checklist

- ~~[ ] Added tests for the changes~~ tested manually with patched `mria_schema:create_table` rigged with a sleep time; could not reproduce issue after patch.


apps/emqx_machine/src/emqx_machine_sup.erl
```diff
@@ -29,10 +29,18 @@ start_link() ->
 
 init([]) ->
     Terminator = child_worker(emqx_machine_terminator, [], transient),
+    ReplicantHealthProbe = child_worker(emqx_machine_replicant_health_probe, [], transient),
+    %% Must start before `post_boot'.
+    Booter = child_worker(emqx_machine_app_booter, [], permanent),
```
zmstone, line 34: we can maybe turn `emqx_machine_boot` into a gen_statem or gen_server.
and use the `init/1` callback to boot-up instead of calling `post_boot` then exit.

apps/emqx_machine/src/emqx_machine_app_booter.erl
```diff
@@ -0,0 +1,129 @@
...
+    ?SLOG(notice, #{msg => "(re)starting_emqx_apps"}),
+    lists:foreach(fun start_one_app/1, emqx_machine_boot:sorted_reboot_apps()),
+    ?tp(emqx_machine_boot_apps_started, #{}).
+
+handle_stop_apps() ->
+    ?SLOG(notice, #{msg => "stopping_emqx_apps"}),
+    _ = emqx_alarm_handler:unload(),
+    ok = emqx_conf_app:unset_config_loaded(),
+    ok = emqx_plugins:ensure_stopped(),
+    lists:foreach(fun stop_one_app/1, lists:reverse(emqx_machine_boot:sorted_reboot_apps())).
```
zmstone, line 105: maybe this block should continue be called by the original caller processes, this gen-call can return `ok` just to make sure boot-up is completed.

`emqx_machine_boot:is_boot_complete/1` which blocks and wait for the process to return.

the original caller of this block is not just ekka, but emqx_machine_terminator too.

apps/emqx_machine/src/emqx_machine_sup.erl
thalesmg: I think there's still a need for `post_boot`:  the first time we boot, we print the current EMQX version, and start autocluster; that happens only once during the node's uptime, even if it joins other clusters multiple times.

zmstone: but the process is only `init`ialized once, right?
i mean do not make it temporary, start it as a gen_xxx and keep it running, so there is no need to introduce a new module and new process.

thalesmg: I think keeping it in a separate module helps to keep things clean and separated, since there's a bunch of stuff already in machine_boot.

I'll move the post_init stuff to the booter process then.

apps/emqx_machine/src/emqx_machine_app_booter.erl
thalesmg: :checkered_flag:

id:
@thalesmg @zmstone release version in Jira is set to 5.8.3, but here it says 5.9.0 and targets master. We need to either update Jira, or change target branch here.

thalesmg:
> @thalesmg @zmstone release version in Jira is set to 5.8.3, but here it says 5.9.0 and targets master. We need to either update Jira, or change target branch here.

@id the release version was not defined in Jira when this was opened, so I assumed `master` would be appropriate.

thalesmg:
rebased onto r583

zmstone (approved):

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5060)

Issue #13215, opened 2024-06-10
But I found that it was sent successfully, and I also subscribed to this topic. Should return 200 and messageid


jasonccs:
### What happened?

When I publish delayed messages by adjusting API/v5/publish
{
"Payload encoding": "plain",
"Topic": "$delayed/10/go",
"QoS": 0,
"Clientid": "string",
"Payload": "hello emqx API",
"Properties":{
"Payload format indicator": 0,
"Message_expiry_interval": 20,
"Response_topic": "go",
//"Correlation_data": "string",
"User_properties":{
"Foo": "bar"
},
"Content_type": "application/json"
},
"Retain": false
}
The response is{
"Message": "no_matching_subscribers",
"Reason code": 16
}But I found that it was sent successfully, and I also subscribed to this topic. Should return 200 and messageid

### What did you expect to happen?

{
    "message": "no_matching_subscribers",
    "reason_code": 16
}

### EMQX version
5.7
<details>

```console
$ ./bin/emqx_ctl broker
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

zmstone:
@jasonccs thank you for the report.
We'll follow up with a fix, but hard to commit which release.

jasonccs:
这是来自QQ邮箱的假期自动回复邮件！
    您好，我最近正在休假中，无法亲自回复您的邮件。我将在假期结束后，第一时间给您回复。

PR #14182, opened 2024-11-07
fix(publish api): return 200 for delayed messages even when there are no subscribers

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-12780

Fixes https://github.com/emqx/emqx/issues/13215

Release version: v/e5.9.0

## PR Checklist


keynslug (approved):

apps/emqx_management/src/emqx_mgmt_api_publish.erl
```diff
@@ -272,6 +272,8 @@ is_ok_deliver({_NodeOrShare, _MatchedTopic, {error, _}}) -> false.
 %% No preceding payload format indicator to compare against.
 %% Content-Type check should be done at HTTP layer but not here.
 %% 153                Payload format invalid                  400
+publish_result_to_http_reply(#message{topic = <<"$delayed/", _/binary>>} = Message, []) ->
```
keynslug, line 275: Thinking a bit more about this, more future-proof I guess would be for the `emqx_delayed` hook to signal to the broker that "look, this message is taken care of, answer with `RC_SUCCESS` to the client", and the result of `publish/1` would reflect that.

thalesmg: Makes sense.  However, at this point, the only thing that comes back from the delayed message hook is an empty destination list, so currently we can't return, for example, the message with special headers signaling the reason code to the hook without changing this hook's API.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6669)

PR #14192, opened 2024-11-08
feat(jwt_auth): make will msg be sent in case of disconnetion on auth expire

savonarola:
Fixes [EMQX-12708](https://emqx.atlassian.net/browse/EMQX-12708)

Release version: v/e5.8.3

## Summary

The problem description is quite simple:
* JWT authenticator provides both **authentication** and **authorization rules** that we put into the `clientinfo`.
* They both have the same expiration deadline (the exp of JWT).
* We schedule a disconnect timer for the connection after the successful JWT auth, so that client is kicked when its auth expires.
* We kick the client, and its will message cannot pass the authz checks because they have just expired (together with the auth).

A straightforward solution is to add some thresholds to authz expiration deadline to give the will message a chance to pass authorization. However, since the disconnection is triggered by a timer, the event may be arbitrarily late if the connection is under load, leading to unstable behavior: most of the time, the messages are not lost, but sometimes they are.

The funny part is that the JWT auth does not "recognize" the message the sending of which it induced itself 😅 because the information is lost several times:
* The connection does not know which authenticator set the authz rules and the expiration time.
* When the expiration timer triggers disconnect, the code that sends the will message does not know the reason for sending.
* Finally, the called authorization app does not have enough context — it does not know at all that it was called for a will message and for which reason the will message is sent.

In the proposed solution, we try to prevent expiration data loss so that the authorizer at least receives back the exact data that it set:
* We make the connection expiration timer more specific — it becomes the "auth expiration" timer set from the `auth_expire_at` value returned from authenticators;
* then, handling the timer, we introduce a specific `maybe_publish_will_msg` reason: `auth_expired`;
* so in `maybe_publish_will_msg` with `auth_expired` argument, we know that we are sending the will message maybe a bit later than the desired moment `auth_expire_at`. 
* finally, we inject this moment as the time into the authorization context (`clientinfo`). This allows the authorizer to work without a side effect (fetching time) and compare the expiration moment with itself.

[EMQX-12708]: https://emqx.atlassian.net/browse/EMQX-12708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -1486,13 +1486,16 @@ handle_timeout(
     end;
 handle_timeout(
     _TRef,
-    connection_expire,
+    connection_auth_expire,
     #channel{conn_state = ConnState} = Channel0
 ) ->
-    Channel1 = clean_timer(connection_expire, Channel0),
+    Channel1 = clean_timer(connection_auth_expire, Channel0),
     case ConnState of
-        disconnected -> {ok, Channel1};
-        _ -> handle_out(disconnect, ?RC_NOT_AUTHORIZED, Channel1)
+        disconnected ->
+            {ok, Channel1};
+        _ ->
+            Channel2 = maybe_publish_will_msg(auth_expired, Channel1),
+            handle_out(disconnect, ?RC_NOT_AUTHORIZED, Channel2)
     end;
 handle_timeout(TRef, Msg, Channel) ->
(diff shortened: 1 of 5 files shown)


zmstone (approved):

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6671)

PR #14227, opened 2024-11-14
Consistently update progress for durable shared subscriptions

savonarola:
Fixes
[EMQX-13472](https://emqx.atlassian.net/browse/EMQX-13472)
[EMQX-13473](https://emqx.atlassian.net/browse/EMQX-13473)

Release version: e5.8.3

# Summary

The major changes are:
* The stream renewal for shared subs does not require external `renew_streams` call anymore, it is done in `on_info` callback just after the stream lease events are received from the leader. This is an internal change of the shared subs.
* The `on_replay` callback (that reports progress) is called from the two places of the DS session:
  * on seqno release when the iterator span was completely consumed;
  * on stream enqueue (needed only for qos0).
 This change required little changes to the scheduler's interface: we need to catch moments when a stream is fully acked.
* Internal shared subs module refactoring: We eliminate "scheduled actions" machinery by just making each subscription to the same topic a standalone consumer from the leader's perspective. So, in case of quick unsubscribe/subscribe, they (opaquely to each other) may coexist in the session, the old ones handling stream revocation and the new one waiting for stream leases.

[EMQX-13472]: https://emqx.atlassian.net/browse/EMQX-13472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMQX-13473]: https://emqx.atlassian.net/browse/EMQX-13473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/priv/bpapi.versions
+++ b/apps/emqx/priv/bpapi.versions
@@ -28,6 +28,7 @@
 {emqx_ds_beamsplitter,1}.
 {emqx_ds_new_streams,1}.
 {emqx_ds_shared_sub,1}.
+{emqx_ds_shared_sub,2}.
 {emqx_eviction_agent,1}.
 {emqx_eviction_agent,2}.
 {emqx_eviction_agent,3}.
--- a/apps/emqx/src/emqx.app.src
+++ b/apps/emqx/src/emqx.app.src
@@ -2,7 +2,7 @@
 {application, emqx, [
     {id, "emqx"},
     {description, "EMQX Core"},
-    {vsn, "5.4.2"},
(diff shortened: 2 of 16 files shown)


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5091)

PR #14229, opened 2024-11-15
fix(dsrepl): ensure preconditions are idemotent under log replays

keynslug:
Fixes [EMQX-12848](https://emqx.atlassian.net/browse/EMQX-12848).

Release version: e5.8.3

## Summary

Implements "trivial" approach mentioned in the ticket.

When log replay is involved, preconditions should work as if they observe the storage state at the specific point corresponding to the Raft index, without the effects of any subsequent log entries that may actually have been already applied to the storage state before.

Because it's impossible with how we currently manage storage state and snapshots, emulate that through keeping track of the latest Raft index in the storage itself, and skip any (preconditions-involving) batches that seem to be already applied by reading the latest Raft index record on apply.

## PR Checklist


apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -662,6 +685,39 @@ delete_next(
             {ok, end_of_stream}
     end.
 
+-spec store_global(shard_id(), _KVs :: #{binary() => binary()}, batch_store_opts()) ->
```
ieQu1, line 688: Please add a comment describing purpose and semantics of this function

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5108)

PR #14248, opened 2024-11-19
feat(router): require node to cleanup routes on restart

keynslug:
Fixes [EMQX-11852](https://emqx.atlassian.net/browse/EMQX-11852).

Release version: v5.8.4

## Summary

Tolerate transient connectivity issues. It's now assumed that receiving a `{node, down, ...}` event does not mean that the node in question is dead, but that it has experienced connectivity loss.

Now, the router helper:
* Cleans any leftover routes from previous incarnations upon restart.
* Consequently, expects cluster nodes to clean up themselves upon restart.
* Cleans routes of a cluster node that decided to leave upon receiving such event.
* Cleans routes of cluster nodes that are down for more than 1 hour.

Notable consequences of these changes are:
* Global routing table will hold offline nodes' entries for a while, instead of cleaning them up as soon as node is considered offline.
* Node startup sequence now includes an additional step of cleaning up route leftovers from its previous incarnation.
    Currently, this boils down to what essentially a full-scan, so the time it takes the node to start now depends on the size of the routing table. However, when running under emqx/otp this full-scan should use optimized implementation. The same is true when Mria autoheal is involved.

## PR Checklist

[EMQX-11852]: https://emqx.atlassian.net/browse/EMQX-11852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_router_helper.erl
+++ b/apps/emqx/src/emqx_router_helper.erl
@@ -31,6 +31,7 @@
 -export([
     start_link/0,
     monitor/1,
+    is_routable/1,
     purge/0,
     purge_force/0
 ]).
@@ -49,6 +50,8 @@
 ]).
 
 -record(routing_node, {name, const = unused}).
+
+-define(TAB_STATUS, ?MODULE).
 
 -define(LOCK(RESOURCE), {?MODULE, RESOURCE}).
 
@@ -95,6 +98,19 @@ monitor(Node) when is_atom(Node) ->
         false -> mria:dirty_write(?ROUTING_NODE, #routing_node{name = Node})
     end.
 
+%% @doc Is given node considered routable?
+%% I.e. should the broker attempt to forward messages there, even if there are
+%% routes to this node in the routing table?
+-spec is_routable(node()) -> boolean().
+is_routable(Node) when Node == node() ->
+    true;
+is_routable(Node) ->
+    try
+        lookup_node_reachable(Node)
+    catch
+        error:badarg -> true
+    end.
+
 %% @doc Schedule dead node purges.
 -spec purge() -> scheduled.
 purge() ->
@@ -114,6 +130,8 @@ purge_force() ->
 
 init([]) ->
     process_flag(trap_exit, true),
+    %% Initialize a table to cache node status.
+    Tab = ets:new(?TAB_STATUS, [protected, named_table, set, {read_concurrency, true}]),
     %% Monitor nodes lifecycle events.
     ok = ekka:monitor(membership),
     %% Cleanup any routes left by old incarnations of this node (if any).
@@ -126,6 +144,7 @@ init([]) ->
     State = #{
         down => #{},
         left => #{},
+        tab_node_status => Tab,
         timer_reconcile => TRef
     },
     {ok, State, hibernate}.
@@ -173,10 +192,12 @@ handle_membership_event(_Event, State) ->
     State.
 
 record_node_down(Node, State = #{down := Down}) ->
+    _ = ets:insert(?TAB_STATUS, {Node, down, _Reachable = false}),
     Record = #{since => erlang:monotonic_time(millisecond)},
     State#{down := Down#{Node => Record}}.
 
 record_node_left(Node, State = #{left := Left}) ->
+    _ = ets:insert(?TAB_STATUS, {Node, left, _Reachable = false}),
     Record = #{since => erlang:monotonic_time(millisecond)},
     State#{left := Left#{Node => Record}}.
 
@@ -184,10 +205,14 @@ record_node_alive(Node, State) ->
     forget_node(Node, State).
 
 forget_node(Node, State = #{down := Down, left := Left}) ->
+    _ = ets:delete(?TAB_STATUS, Node),
     State#{
         down := maps:remove(Node, Down),
         left := maps:remove(Node, Left)
     }.
+
+lookup_node_reachable(Node) ->
+    ets:lookup_element(?TAB_STATUS, Node, 3, _Default = true).
 
 handle_reconcile(State) ->
     TRef = schedule_task(reconcile, ?RECONCILE_INTERVAL),
@@ -196,8 +221,10 @@ handle_reconcile(State) ->
     schedule_purges(TS, reconcile(NState)).
 
 reconcile(State) ->
-    %% Fallback: avoid purging live nodes, if missed lifecycle events for whatever reason.
-    lists:foldl(fun record_node_alive/2, State, mria:running_nodes()).
+    %% NOTE
+    %% This is a fallback mechanism. Avoid purging / ignoring live nodes, if missed
+    %% lifecycle events for whatever reason.
+    lists:foldl(fun record_node_alive/2, State, mria:running_nodes() ++ nodes()).
 
 select_outdated(NodesInfo, Since0) ->
     maps:keys(maps:filter(fun(_, #{since := Since}) -> Since < Since0 end, NodesInfo)).
--- a/apps/emqx/src/emqx_router_helper.erl
+++ b/apps/emqx/src/emqx_router_helper.erl
@@ -30,7 +30,9 @@
 %% API
 -export([
     start_link/0,
-    monitor/1
+    monitor/1,
+    purge/0,
+    purge_force/0
 ]).
 
 %% Internal export
@@ -46,16 +48,19 @@
     code_change/3
 ]).
 
-%% Internal exports (RPC)
--export([
-    cleanup_routes/1
-]).
-
 -record(routing_node, {name, const = unused}).
 
--define(LOCK, {?MODULE, cleanup_routes}).
+-define(LOCK(RESOURCE), {?MODULE, RESOURCE}).
 
--dialyzer({nowarn_function, [cleanup_routes/1]}).
+%% How often to reconcile nodes state? (ms)
+-define(RECONCILE_INTERVAL, 10 * 60_000).
+%% How soon should a dead node be purged? (ms)
+-define(PURGE_DEAD_TIMEOUT, 60 * 60_000).
+%% How soon should a left node be purged? (ms)
+%% This is a fallback, left node is expected to be purged right away.
+-define(PURGE_LEFT_TIMEOUT, 15_000).
+
+% -define(PURGE_LEFT_FALLBACK_TIMEOUT, {15_000, '±', 2_500}).
 
 %%--------------------------------------------------------------------
 %% Mnesia bootstrap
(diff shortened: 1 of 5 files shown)


apps/emqx/src/emqx_router_helper.erl
```diff
@@ -188,6 +190,169 @@ code_change(_OldVsn, State, _Extra) ->
...
+    %% Include stopped nodes as well.
+    mria_membership:nodelist().
+
+pick_responsible(Task) ->
+    %% Pick a responsible core node.
+    %% We expect the same node to be picked as responsible across the cluster (unless
+    %% the cluster is highly turbulent).
+    Nodes = lists:sort(mria_membership:running_core_nodelist()),
+    N = length(Nodes),
+    N > 0 andalso lists:nth(1 + erlang:phash2(Task, N), Nodes).
```
savonarola, line 352: nit: the return value calculation and its matching was a bit difficult to understand

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -86,39 +102,65 @@ start_link() ->
...
-        [],
-        mnesia:dirty_all_keys(?ROUTING_NODE)
-    ),
+    _ = purge_this_node(),
+    %% Setup periodic stats reporting.
     ok = emqx_stats:update_interval(route_stats, fun ?MODULE:stats_fun/0),
-    {ok, #{nodes => Nodes}, hibernate}.
+    TRef = schedule_task(reconcile, ?RECONCILE_INTERVAL),
+    State = #{
+        down => #{},
```
savonarola, line 154: Curious: why do we additional collections besides `?TAB_STATUS`?

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -188,6 +190,169 @@ code_change(_OldVsn, State, _Extra) ->
...
+choose_timeout(Baseline) ->
+    Baseline.
+
+%%
+
+am_core() ->
+    mria_config:whoami() =/= replicant.
+
+cores() ->
+    %% Include stopped nodes as well.
```
savonarola, line 343: Curious: why do we selects stopped nodes? To make the lock intentionally fail?

savonarola:
>Cleans any leftover routes from previous incarnations upon restart.
>Expects cluster nodes to clean up themselves upon restart.

How do these cases differ?

apps/emqx/src/emqx_router_helper.erl
keynslug: Yes. And to fail whichever transactions happening on a minority of cluster nodes if there is a partition.

keynslug: ...However I just realized that `global:trans/4` is extremely lax about unreachable nodes, so it makes little sense actually. In this case, my hope is `mria` / `mnesia` will reject writes originating on a minority.

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -188,6 +230,228 @@ code_change(_OldVsn, State, _Extra) ->
...
+    State;
+handle_membership_event({node, up, Node}, State) ->
+    _ = record_node_alive(Node),
+    State;
+handle_membership_event({mnesia, up, Node}, State = #{last_membership := Membership}) ->
+    State#{last_membership := lists:usort([Node | Membership])};
+handle_membership_event(_Event, State) ->
+    State.
+
+record_node_down(Node) ->
```
zmstone, line 248: write a warning message ?

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -188,6 +230,228 @@ code_change(_OldVsn, State, _Extra) ->
...
+record_node_left(Node) ->
+    NS = #ns{
+        node = Node,
+        status = left,
+        since = erlang:monotonic_time(millisecond)
+    },
+    ets:insert(?TAB_STATUS, NS).
+
+record_node_alive(Node) ->
+    forget_node(Node).
```
zmstone, line 274: write a warning message ?

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -14,6 +14,33 @@
...
+%%    The assumption is that the node has crashed (worst-case), so the
+%%    previous incarnation's routes are still present upon restart.
+%% 2. Managing portions of global routing table belonging to dead / "left"
+%%    cluster members, i.e. members that are not supposed to come back
+%%    online again.
+%%
+%% Only core nodes are responsible for the latter task. Moreover, helper
+%% adopts the following operational model:
+%% 1. Core nodes are supposed to be explicitly evicted (or "left") from
+%%    the cluster. Even if a core node is marked down for several hours,
```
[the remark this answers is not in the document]
keynslug: I've tried to illustrate with something much larger than 15 minutes, which only affects replicants.

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -46,16 +77,29 @@
...
--export([
-    cleanup_routes/1
-]).
-
 -record(routing_node, {name, const = unused}).
 
--define(LOCK, {?MODULE, cleanup_routes}).
+-define(TAB_STATUS, ?MODULE).
+
+-define(LOCK(RESOURCE), {?MODULE, RESOURCE}).
```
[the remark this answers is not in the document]
keynslug: Yes, now lock is per node-to-be-purged, it is intentional change. I'm not sure what is the action point: do you mean we need to keep the old lock construct?

apps/emqx/src/emqx_router_helper.erl
keynslug: There's at least https://github.com/emqx/mria/blob/603bea2481917041e502952e834ccc53ddc876f0/src/mria_membership.erl#L340, though I wouldn't be surprised if there are more.

Why warning though? Nodes coming up and down are regular lifecycle events. Nevertheless, if we want to increase visibility, it's probably better to do it on `mria` side?

apps/emqx/src/emqx_router_helper.erl
keynslug: In a similar vein, there's https://github.com/emqx/mria/blob/603bea2481917041e502952e834ccc53ddc876f0/src/mria_membership.erl#L324 already.

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -86,39 +130,77 @@ start_link() ->
...
+    true;
+is_routable(Node) ->
+    try
+        lookup_node_reachable(Node)
+    catch
+        error:badarg -> true
+    end.
+
+%% @doc Schedule dead node purges.
+-spec purge() -> scheduled.
```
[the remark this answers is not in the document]
keynslug: Yes, was used in tests before. Idea is that it still might be useful in a `remote_console` environment. Fine dropping it if you think this is an overkill.

apps/emqx/src/emqx_router_helper.erl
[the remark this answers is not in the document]
zmstone: just a question, do you see any risk of cleaning up two or more nodes concurrently.

apps/emqx/src/emqx_router_helper.erl
zmstone: Sorry, maybe not here, but at the place where one node decides to clean-up a peer node.
This is for better traceability in case some routes missing, we can look for this log to see when the routes were cleared by which node.

keynslug: There's https://github.com/emqx/emqx/blob/4c1135eb4bd69dad849a21ce45412dc715ae20e2/apps/emqx/src/emqx_router_helper.erl#L362. There's no event _before_ this is done, might add it as well I guess.

Still unsure about warning level, because the current expectation is that a node is not supposed to completely clean its own routes during orderly shutdown. I was thinking about changing it to "node should clean its routes during shutdown", and then this event indeed can be turned into a warning (because it would then imply node crash / permanent connectivity loss), and thus become more visible. What do you think?

apps/emqx/src/emqx_router_helper.erl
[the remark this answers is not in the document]
keynslug: No, I don't. That's not a very good judgement though, I guess if you have some concerns, then it means we need more tests.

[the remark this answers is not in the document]
zmstone: can't think of anything either. added the comment just to trigger more discussion.

apps/emqx/src/emqx_router_helper.erl
zmstone: Although not documented, a proper shutdown is to evict the node first
in this sense, any abrupt shutdown is considered non-graceful and worth some warning.
Maybe write a `warning` if there is ANY route to this node, and `debug` if ignore

keynslug: >  Maybe write a `warning` if there is ANY route to this node, and `debug` if ignore

It is like that right now.

> Although not documented, a proper shutdown is to evict the node first

With the current expectation it would still produce a warning (because evicted / "left" node don't clean own routes during shutdown).

But in general I can imagine the one who operates the cluster might just want to `systemctl restart emqx` for whatever reason, do you think it still needs to be considered non-graceful shutdown?

zmstone: > It is similar right now, info and debug if node had no routes.

Great!

> Do you think it still needs to be considered non-graceful shutdown?

I don't really want to argue which way of shutdown is or is not graceful.
What I want is some logs to help investigation if something odd happens, and it seems raising it to warning is the simplest way.
The downside is someone may ask why this is a warning, what should I (as a sys-admin) do, we can maybe add more context in the log like: "if this peer node is shutdown by admin, ignore this warning message" ?

keynslug: > raising it to warning is the simplest way.

True. Okay, will raise the event to a warning.

In general though, can't help but argue that it's always the simplest way. So if we always follow the same principle we'll end up with logs sprinkled with warnings that are not actually warnings.

keynslug: Done in 7b5e59f05d2430162261f8057346f908f700e250.

apps/emqx/src/emqx_router_helper.erl
```diff
@@ -86,39 +130,77 @@ start_link() ->
...
+purge() ->
```
zmstone, line 153: schedule_purge

zmstone (dismissed):

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5492)

PR #14256, opened 2024-11-20
build: make sure elastic bridge is ee only

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5173)

PR #14267, opened 2024-11-22
fix: do not redact secret value when it is file path

zmstone:
Release version: v/e5.8.4

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_schema_secret.erl
+++ b/apps/emqx/src/emqx_schema_secret.erl
@@ -71,8 +71,8 @@ convert_secret(Secret, #{}) ->
     end.
 
 -spec wrap(source()) -> emqx_secret:t(t()).
-wrap(<<"file://", Filename/binary>>) ->
-    emqx_secret:wrap_load({file, Filename});
+wrap(<<"file://", _Filename/binary>> = Secret) ->
+    emqx_secret:wrap_load({file, Secret});
 wrap(Secret) ->
     emqx_secret:wrap(Secret).
 
--- a/apps/emqx/src/emqx_secret_loader.erl
+++ b/apps/emqx/src/emqx_secret_loader.erl
@@ -22,9 +22,13 @@
 
 -export_type([source/0]).
 
--type source() :: {file, file:filename_all()}.
+-type source() :: {file, string() | binary()}.
 
 -spec load(source()) -> binary() | no_return().
+load({file, <<"file://", Path/binary>>}) ->
+    file(Path);
+load({file, "file://" ++ Path}) ->
+    file(Path);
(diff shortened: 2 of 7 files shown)


apps/emqx/src/emqx_schema_secret.erl
```diff
@@ -71,8 +71,8 @@ convert_secret(Secret, #{}) ->
     end.
 
 -spec wrap(source()) -> emqx_secret:t(t()).
-wrap(<<"file://", Filename/binary>>) ->
-    emqx_secret:wrap_load({file, Filename});
+wrap(<<"file://", _Filename/binary>> = Secret) ->
+    emqx_secret:wrap_load({file, Secret});
```
keynslug, line 75: Q: Why this change is needed?

apps/emqx/src/emqx_secret_loader.erl
```diff
@@ -22,9 +22,13 @@
 
 -export_type([source/0]).
 
--type source() :: {file, file:filename_all()}.
+-type source() :: {file, string() | binary()}.
 
 -spec load(source()) -> binary() | no_return().
+load({file, <<"file://", Path/binary>>}) ->
+    file(Path);
+load({file, "file://" ++ Path}) ->
```
keynslug, line 30: Nit. RIght now it's not possible to wrap a `string()`, only `binary()`.

apps/emqx/src/emqx_schema_secret.erl
zmstone: this is to keep the original prefix so there is no need to concatenate "file://" after calling `emqx_secret:term`.

keynslug (approved):

apps/emqx/src/emqx_schema_secret.erl
keynslug: Just to clarify: the idea was to keep this `file://...` convention on the schema level only, so that lower-level code would not have to know about it.

apps/emqx/src/emqx_secret_loader.erl
```diff
@@ -22,9 +22,13 @@
...
+    file(Path);
 load({file, Filename}) ->
     file(Filename).
```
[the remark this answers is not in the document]
zmstone: removed.

apps/emqx/src/emqx_secret_loader.erl
zmstone: right. but I'd keep this clause here just in case.

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5333)

PR #14278, opened 2024-11-25
refactor: add extra field to connection and ws_connection states

zmstone:
Release version: v/e5.8.4

## Summary

This is to make future hot-upgrade easier so there is no need to wake up potentially millions of processes to evaluate state upgrade functions.


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5337)

PR #14303, opened 2024-11-29
fix(scram_auth): fix scram http auth and enable tests

savonarola:
Fixes [EMQX-13598](https://emqx.atlassian.net/browse/EMQX-13598)

Release version: v/e5.9.0

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [x] Added tests for the changes (enabled existing tests)
- [na] Added property-based tests for code which performs user input validation
- [ ] Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] For internal contributor: there is a jira ticket to track this change
- [x] Schema changes are backward compatible

PR #14315, opened 2024-12-02
fix(scram_auth): move changelog to the right place

savonarola:
Follow-up for https://github.com/emqx/emqx/pull/14303


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6689)

PR #14317, opened 2024-12-02
fix(mgmt): drop incorrect but inactive clause in paging accumulation

keynslug:
Release version: v5.8.4 (?)

## Summary

This will make paging accumulation work in more general cases (as shown by the proptest). Before this commit, feeding node-query results containing more than `Limit` number of rows produced incorrect results. However, there is no evidence this incorrect codepath was ever active.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api.erl
@@ -54,7 +54,11 @@
 ]).
 
 -ifdef(TEST).
+-include_lib("proper/include/proper.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
 -export([paginate_test_format/1]).
+
 -endif.
 
 -export_type([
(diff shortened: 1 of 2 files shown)


apps/emqx_management/src/emqx_mgmt_api.erl
```diff
@@ -707,20 +709,19 @@ format_query_result(
...
-                end,
-                [],
-                RowsAcc
-            )
-        )
+        data => format_query_data(FmtFun, RowsAcc, Opts)
     }.
 
+format_query_data(FmtFun, RowsAcc, Opts) ->
+    %% NOTE: `RowsAcc` is reversed in the node-order, `lists:foldl/3` is correct here.
```
zmstone, line 716: make sense to `lists:append(lists:reverse(...))` ?

zmstone (approved):

zmstone (commented):
add a changelog

apps/emqx_management/src/emqx_mgmt_api.erl
keynslug: The `FmtFun` needs to run per each row, but the list is `[{Node, [Row]}]`, so I think the current implementation is needed (and also likely the cheapest one).

keynslug:
> add a changelog

Done. Worth mentioning it's not a user-facing issue, as far as I can tell, this is more of a preventive fix.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5341)

PR #13666, opened 2024-08-22
fix: the emqx ctl command not working after relup

terry-xiaoyu:
Fixes https://emqx.atlassian.net/browse/EMQX-12935

Release version: v/e5.8.0

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [ ] Added tests for the changes
- [ ] Added property-based tests for code which performs user input validation
- [ ] Changed lines covered in coverage report
- [ ] Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [ ] For internal contributor: there is a jira ticket to track this change
- [ ] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket
- [ ] Schema changes are backward compatible

## Checklist for CI (.github/workflows) changes

- [ ] If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)
- [ ] Change log has been added to `changes/` dir for user-facing artifacts update

PR #14323, opened 2024-12-03
fix: emqx help command

JimMoen:
Fixes [EMQX-13611](https://emqx.atlassian.net/browse/EMQX-13611)

Release version: v/e5.8.3

## PR Checklist

[EMQX-13611]: https://emqx.atlassian.net/browse/EMQX-13611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/bin/emqx
+++ b/bin/emqx
@@ -216,6 +216,19 @@ BASE_RUNNER_ROOT_DIR="${BASE_RUNNER_ROOT_DIR:-$RUNNER_ROOT_DIR}"
 # shellcheck disable=SC1090,SC1091
 . "$RUNNER_ROOT_DIR"/releases/emqx_vars
 
+RELUP_DIR="relup"
+RELUP_PATH="$RUNNER_ROOT_DIR/$RELUP_DIR"
+
+if [ -f "$RELUP_PATH/version" ]; then
+    TARGET_VSN=$(cat "$RELUP_PATH/version")
+    export BASE_RUNNER_ROOT_DIR
+    ## only print for boot commands to avoid messing the CLI outputs
+    if [ "$IS_BOOT_COMMAND" = 'yes' ]; then
+        loginfo "Loading emqx from hot-upgrade dir: $RELUP_PATH"
+    fi
+    exec "$RELUP_PATH/$TARGET_VSN"/bin/emqx "$@"
+fi
+
 COMMAND="${1:-}"
 if [ -z "$COMMAND" ]; then
     usage 'help'
@@ -250,19 +263,6 @@ case "${COMMAND}" in
         IS_BOOT_COMMAND='no'
         ;;
 esac
-
-RELUP_DIR="relup"
-RELUP_PATH="$RUNNER_ROOT_DIR/$RELUP_DIR"
-
-if [ -f "$RELUP_PATH/version" ]; then
-    TARGET_VSN=$(cat "$RELUP_PATH/version")
-    export BASE_RUNNER_ROOT_DIR
-    ## only print for boot commands to avoid messing the CLI outputs
-    if [ "$IS_BOOT_COMMAND" = 'yes' ]; then
-        loginfo "Loading emqx from hot-upgrade dir: $RELUP_PATH"
-    fi
-    exec "$RELUP_PATH/$TARGET_VSN"/bin/emqx "$@"
-fi
 
 # defined in emqx_vars
 export RUNNER_ROOT_DIR
--- a/bin/emqx
+++ b/bin/emqx
@@ -12,6 +12,7 @@ if [ "$DEBUG" -eq 2 ]; then
   set -x
   export PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 fi
+GREP='grep --color=never'
 
 logerr() {
     if [ "${TERM:-dumb}" = dumb ]; then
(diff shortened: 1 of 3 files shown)


bin/emqx
```diff
@@ -71,25 +76,28 @@ else
...
+export BINDIR="$ERTS_DIR/bin"
+export EMU="beam"
+export PROGNAME="erl"
+export ERTS_LIB_DIR="$RUNNER_ROOT_DIR/lib"
 
+RUNNER_SCRIPT="$RUNNER_BIN_DIR/$REL_NAME"
+CODE_LOADING_MODE="${CODE_LOADING_MODE:-embedded}"
+REL_DIR="$RUNNER_ROOT_DIR/releases/$REL_VSN"
+
+COMMAND="${1:-}"
```
zmstone, line 100: we should avoid doing so before switching user and `exec` (line 122 left, and 128 right)

zmstone: for help command, there is no need to switch user.

bin/emqx
```diff
@@ -71,25 +76,28 @@ else
...
+export SCHEMA_MOD
+export IS_ENTERPRISE
 
-if [ "${2:-}" = 'help' ]; then
-    ## 'ctl' command has its own usage info
-    if [ "$COMMAND" != 'ctl' ]; then
-        usage "$COMMAND"
-        exit 0
-    fi
-fi
```
zmstone, line 91: I guess it's enough to just move this part to after usage function is defeined ?

bin/emqx
JimMoen: - `usage()` requires $REL_NAME in `emqx_vars`
- `emqx_vars` requires `BASE_RUNNER_ROOT_DIR`
- we should call newer `usage()` or other commands after relup. which means the `exec` should put before it.

JimMoen:
The bug was introduced by pr #13666
request review about the relup part by @terry-xiaoyu

bin/emqx
```diff
@@ -50,110 +51,6 @@ die() {
...
-export ERTS_LIB_DIR="$RUNNER_ROOT_DIR/lib"
-DYNLIBS_DIR="$RUNNER_ROOT_DIR/dynlibs"
-
-assert_node_alive() {
-    if ! relx_nodetool "ping" > /dev/null; then
-        exit 1
-    fi
-}
-
 usage() {
```
zmstone, line 54: we'll need to initialize variables used in `usage` function before it.
e.g. `REL_NAME`

zmstone: `REL_NAME` seems to be the only var used in usage. 
maybe just replace it with `emqx` hard-coded.

JimMoen: Prefer to keep REL_NAME, which is used by many other functions.
In addition, I have tested that as long as I can ensure that the `REL_NAME` variable has been initialized before calling the `usage()` function for the first time, it will be fine.
first call at line 234 right. https://github.com/JimMoen/emqx/blob/fix-help-command/bin/emqx#L234

zmstone: ok, agree.

zmstone: meaning we move below chunk to before `usage` function, and that's all we need ?

```
# We need to find real directory with emqx files on all platforms
# even when bin/emqx is symlinked on several levels
# - readlink -f works perfectly, but `-f` flag has completely different meaning in BSD version,
#   so we can't use it universally.
# - `stat -f%R` on MacOS does exactly what `readlink -f` does on Linux, but we can't use it
#   as a universal solution either because GNU stat has different syntax and this argument is invalid.
#   Also, version of stat which supports this syntax is only available since MacOS 12
if [ "$(uname -s)" == 'Darwin' ]; then
    product_version="$(sw_vers -productVersion | cut -d '.' -f 1)"
    if [ "$product_version" -ge 12 ]; then
        # if homebrew coreutils package is installed, GNU version of stat can take precedence,
        # so we use absolute path to ensure we are calling MacOS default
        RUNNER_ROOT_DIR="$(cd "$(dirname "$(/usr/bin/stat -f%R "$0" || echo "$0")")"/..; pwd -P)"
    else
        # try our best to resolve link on MacOS <= 11
        RUNNER_ROOT_DIR="$(cd "$(dirname "$(readlink "$0" || echo "$0")")"/..; pwd -P)"
    fi
else
    RUNNER_ROOT_DIR="$(cd "$(dirname "$(realpath "$0" || echo "$0")")"/..; pwd -P)"
fi

BASE_RUNNER_ROOT_DIR="${BASE_RUNNER_ROOT_DIR:-$RUNNER_ROOT_DIR}"
# shellcheck disable=SC1090,SC1091
. "$RUNNER_ROOT_DIR"/releases/emqx_vars
```

JimMoen: yes, and **before `usage`** -> **after `usage`** btw.
need a review by @terry-xiaoyu for relup chunk
```
RELUP_DIR="relup"
RELUP_PATH="$RUNNER_ROOT_DIR/$RELUP_DIR"

if [ -f "$RELUP_PATH/version" ]; then
    TARGET_VSN=$(cat "$RELUP_PATH/version")
    export BASE_RUNNER_ROOT_DIR
    ## only print for boot commands to avoid messing the CLI outputs
    if [ "$IS_BOOT_COMMAND" = 'yes' ]; then
        loginfo "Loading emqx from hot-upgrade dir: $RELUP_PATH"
    fi
    exec "$RELUP_PATH/$TARGET_VSN"/bin/emqx "$@"
fi
```

PR #14327, opened 2024-12-03
chore: bump coveralls-erl

thalesmg:
This adds retries for when coveralls returns 520, as it sometimes does.


id (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5344)

PR #14362, opened 2024-12-06
refactor(resource): forbid changing resource state from `on_get_status` return, upgrade `mysql-otp` and `ecpool`

thalesmg:
Release version: v/e5.8.4

Fixes https://emqx.atlassian.net/browse/EMQX-13514
Fixes https://emqx.atlassian.net/browse/EMQX-13639

## Summary

Since we do health checks asynchronously, this opens up too much space for race conditions where state is clobbered when a connector/resource health check is returned while the state has been changed by other operations.

Consider the following sequence (which happens for MySQL and Postgres connectors, for example):

1. Resource state is S0, and has no channels added yet.
2. A resource health check (HC) starts using `S0` to run.
3. Concurrently, a new channel (action) is added, changing `S0` to `S1` containing the new channel.
4. Resource HC returns with a modified state `S2` based on `S0` which does not contain the new channel.  This state replaces `S1`, and the channel is forever lost.

## PR Checklist

  - Issue was tested manually.  Added a recovery automated test case, but it didn't manage capture the original issue. 

### DIFF
--- a/apps/emqx_bridge_kinesis/src/emqx_bridge_kinesis_impl_producer.erl
+++ b/apps/emqx_bridge_kinesis/src/emqx_bridge_kinesis_impl_producer.erl
@@ -94,11 +94,15 @@ on_stop(InstanceId, _State) ->
     ?status_connected
     | ?status_disconnected
     | {?status_disconnected, state(), {unhealthy_target, string()}}.
-on_get_status(_InstanceId, #{pool_name := Pool} = State) ->
+on_get_status(_InstanceId, #{pool_name := _Pool} = State) ->
+    do_get_status(State, []).
+
+-spec do_get_status(state(), nil() | [_Stream]) -> _.
+do_get_status(#{pool_name := Pool}, StreamArgs) ->
     case
         emqx_resource_pool:health_check_workers(
             Pool,
-            {emqx_bridge_kinesis_connector_client, connection_status, []},
+            {emqx_bridge_kinesis_connector_client, connection_status, StreamArgs},
             ?HEALTH_CHECK_TIMEOUT,
             #{return_values => true}
         )
@@ -111,7 +115,7 @@ on_get_status(_InstanceId, #{pool_name := Pool} = State) ->
                 false ->
                     Unhealthy = lists:any(fun(S) -> S =:= {error, unhealthy_target} end, Values),
                     case Unhealthy of
-                        true -> {?status_disconnected, State, {unhealthy_target, ?TOPIC_MESSAGE}};
+                        true -> {?status_disconnected, {unhealthy_target, ?TOPIC_MESSAGE}};
                         false -> ?status_disconnected
                     end
             end;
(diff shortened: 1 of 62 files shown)


apps/emqx_bridge_mongodb/src/emqx_bridge_mongodb_connector.erl
```diff
@@ -61,13 +61,8 @@ on_get_channel_status(InstanceId, _ChannelId, State) ->
 on_get_channels(InstanceId) ->
     emqx_bridge_v2:get_channels_for_connector(InstanceId).
 
-on_get_status(InstanceId, ConnectorState = #{connector_state := DriverState0}) ->
-    case emqx_mongodb:on_get_status(InstanceId, DriverState0) of
-        {Status, DriverState, Reason} ->
-            {Status, ConnectorState#{connector_state := DriverState}, Reason};
```
zmstone, line 67: What's the original purpose of this state mutation here ?

thalesmg: I can't answer for the original author, but looking at the code, we can see it's accommodating for the possibility of the state changing for whatever reason, given the previous signature for the `emqx_resource:on_get_status` callback.  Which is changed here in this PR.

zmstone: ok, so it means this state is never mutated for mongodb anyway. which is good.
resolving this comment.

apps/emqx_bridge_kinesis/src/emqx_bridge_kinesis_impl_producer.erl
```diff
@@ -94,11 +94,15 @@ on_stop(InstanceId, _State) ->
...
+on_get_status(_InstanceId, #{pool_name := _Pool} = State) ->
+    do_get_status(State, []).
+
+-spec do_get_status(state(), nil() | [_Stream]) -> _.
+do_get_status(#{pool_name := Pool}, StreamArgs) ->
     case
         emqx_resource_pool:health_check_workers(
             Pool,
-            {emqx_bridge_kinesis_connector_client, connection_status, []},
+            {emqx_bridge_kinesis_connector_client, connection_status, StreamArgs},
```
zmstone, line 105: Q: so `[]` means check all channels?

zmstone (approved):

apps/emqx_bridge_kinesis/src/emqx_bridge_kinesis_impl_producer.erl
thalesmg: This was done only to please Elvis.  :sweat_smile: 

Essentially, the HC logic was the same, except that the resource HC (connector) doesn't have a specific stream to check, so it just tries to list streams: https://github.com/emqx/emqx/blob/ddbc493a3128cb10fd9f0aa8533febc5ae320093/apps/emqx_bridge_kinesis/src/emqx_bridge_kinesis_connector_client.erl#L134-L142

thalesmg: ... but it doesn't actually test each one.  And when one stream is specified, it's checked differently:

https://github.com/emqx/emqx/blob/ddbc493a3128cb10fd9f0aa8533febc5ae320093/apps/emqx_bridge_kinesis/src/emqx_bridge_kinesis_connector_client.erl#L131-L132

https://github.com/emqx/emqx/blob/ddbc493a3128cb10fd9f0aa8533febc5ae320093/apps/emqx_bridge_kinesis/src/emqx_bridge_kinesis_connector_client.erl#L166-L167

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5363)

PR #14448, opened 2024-12-19
ci: trigger update-emqx-docs workflow directly on tag push

id:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5505)

PR #14453, opened 2024-12-19
ci(helm): fix flaky test

thalesmg:
Sometimes, `kubectl port-forward` terminates itself if the service is not available or the only endpoint becomes unresponsive, or there're other instabilities.

Potential example: https://github.com/emqx/emqx/actions/runs/12418695854/job/34673287762#step:14:1

### DIFF
--- a/.github/workflows/run_helm_tests.yaml
+++ b/.github/workflows/run_helm_tests.yaml
@@ -138,7 +138,7 @@ jobs:
         done
     - name: Setup 18083 port forwarding
       run: |
-        nohup kubectl port-forward service/${EMQX_NAME} 18083:18083 > /dev/null &
+        nohup bash -c "while true; do kubectl port-forward service/${EMQX_NAME} 18083:18083 ; done" &
     - name: Get auth token
       run: |
         curl --head -X GET --retry 10 --retry-connrefused --retry-delay 6 http://localhost:18083/status


.github/workflows/run_helm_tests.yaml
```diff
@@ -138,7 +138,7 @@ jobs:
         done
     - name: Setup 18083 port forwarding
       run: |
-        nohup kubectl port-forward service/${EMQX_NAME} 18083:18083 > /dev/null &
+        nohup bash -c "while true; do kubectl port-forward service/${EMQX_NAME} 18083:18083 ; done" &
```
zmstone, line 141: so this is not a busy loop ?

thalesmg: only if it were to crash constantly;  under normal operations, it starts and stays up until you kill it or the service has no endpoints (it seems)

zmstone: make sense to add a `sleep 1;` in case it exits immediately for some seconds?

thalesmg: :checkered_flag:

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5798)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5799)

Issue #10215, opened 2023-03-23
Interesting QoS Handling


ZodiacLyu:
## Description
I notice that a client can upgrade or downgrade the QoS of a specific message on EMQ X Broker, **regardless of its own subscription level or QoS level of the original message**.
Take the **QoS Downgrade** as example, the Client subscribes to the topic setting QoS=2 and the Device publishes a QoS 2 message as well. The Client certainly receives the QoS 2 message. According to the specification, the Client should reply a PUBREC message, however, if it replies a PUBACK message that used to finish a QoS 1 publish, the QoS 2 publish will be ended in 1 RTT.
I use the session mechanism to verify my point of view: if the publish is uncompleted, the client will receive the repeated PUBLISH message after a re-connection setting cleansession=0.
## PoC Code
Here is the PoC code, while the authentication and the authorization of a client and a device are needed.
```python
import threading
import time

from scapy.contrib.mqtt import *
from scapy.compat import raw
from scapy.sendrecv import sniff
import socket

iface = ""  # iface communicating with the broker
brokerIP = ""
brokerPort = 1883  # The default listen port of MQTT is 1883

clientID = "client"
clientUsername = "client"
clientPassword = "client"

deviceID = "device"
deviceUsername = "device"
devicePassword = "device"

topic = "test/status"
message = "This is a message"

class sniffer(threading.Thread):
    def __init__(self, port):
        threading.Thread.__init__(self)
        self.port = port

    def run(self):
        lfilter = lambda pkt: TCP in pkt and pkt.dport == self.port and pkt.sport == brokerPort and pkt.getlayer(
            "TCP").flags & 8 == 8
        self.pkt = sniff(
            iface=iface,
            lfilter=lfilter,
            timeout=3
        )

if __name__ == '__main__':
    clientCONN = MQTT() / MQTTConnect(protoname="MQTT", protolevel=4, clientId=clientID, usernameflag=True, passwordflag=True, username=clientUsername, password=clientPassword, cleansess=0)
    clientSUB = MQTT(QOS=1) / MQTTSubscribe(topics=MQTTTopicQOS(topic=topic, QOS=2), msgid=1)

    deviceCONN = MQTT() / MQTTConnect(protoname="MQTT", protolevel=4, clientId=deviceID, usernameflag=True, passwordflag=True, username=deviceUsername, password=devicePassword)
    devicePUB = MQTT(QOS=2) / MQTTPublish(topic=topic, value=message, msgid=5)

[...]

PR #14498, opened 2025-01-07
Introduce subscription API for durable storage

ieQu1:
Fixes [EMQX-13655](https://emqx.atlassian.net/browse/EMQX-13655), [EMQX-13411](https://emqx.atlassian.net/browse/EMQX-13411)

Release version: v/e5.9

## Summary

1. Introduce subscribe API for `emqx_ds`
2. Make durable session fully event-driven, no more periodic timers
3. Introduce a fuzzer suite for the durable session
4. Fix some legacy issues found by the fuzzer

## PR Checklist

[EMQX-13655]: https://emqx.atlassian.net/browse/EMQX-13655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMQX-13411]: https://emqx.atlassian.net/browse/EMQX-13411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
@@ -690,7 +690,7 @@ init_for_subscription(
     SchedS0 = #s{new_stream_subs = NewStreamSubs, sub_metadata = SubMetadata}
 ) ->
     #{id := SubId, start_time := StartTime} = Subscription,
-    TopicFilter = emqx_topic:words(TopicFilterBin),
+    TopicFilter = emqx_ds:topic_words(TopicFilterBin),
     %% Start watching the streams immediately:
     NewStreamsWatch = watch_streams(TopicFilter),
     %% Create the initial record for subscription:
@@ -781,7 +781,7 @@ renew_streams_for_x(S0, SubId, RankX, SchedS0 = #s{sub_metadata = SubMeta0}) ->
                 undefined ->
                     {[], S0, SchedS0};
                 Subscription ->
-                    TopicFilter = emqx_topic:words(TopicFilterBin),
+                    TopicFilter = emqx_ds:topic_words(TopicFilterBin),
                     #sub_metadata{pending_streams = Cache} = SubS0,
                     Pending0 = maps:get(RankX, Cache, []),
                     {NewSRSIds, S, Pending} = do_renew_streams_for_x(
@@ -1147,7 +1147,7 @@ watch_streams(TopicFilter) ->
 
 unwatch_streams(TopicFilterBin, Ref, NewStreamSubs) ->
     ?tp(debug, ?sessds_sched_unwatch_streams, #{
-        topic_filter => emqx_topic:words(TopicFilterBin), ref => Ref
+        topic_filter => emqx_ds:topic_words(TopicFilterBin), ref => Ref
     }),
     emqx_ds_new_streams:unwatch(?PERSISTENT_MESSAGE_DB, Ref),
     maps:remove(Ref, NewStreamSubs).


apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_inflight.erl
```diff
@@ -0,0 +1,357 @@
+%%--------------------------------------------------------------------
```
ieQu1, line 1: Note: this file was renamed from `_ds_buffer` to inflight, but Git treats this as change. There are not many actual changes in this file.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1639,14 +1639,18 @@ handle_info({disconnect, ReasonCode, ReasonName, Props}, Channel) ->
     handle_out(disconnect, {ReasonCode, ReasonName, Props}, Channel);
 handle_info({puback, PacketId, PubRes, RC}, Channel) ->
     do_finish_publish(PacketId, PubRes, RC, Channel);
-handle_info({'DOWN', Ref, process, Pid, Reason}, Channel) ->
-    case emqx_hooks:run_fold('client.monitored_process_down', [Ref, Pid, Reason], []) of
-        [] -> {ok, Channel};
-        Msgs -> {ok, Msgs, Channel}
-    end;
-handle_info(Info, Channel = #channel{session = Session0, clientinfo = ClientInfo}) ->
+handle_info(Info, Channel0 = #channel{session = Session0, clientinfo = ClientInfo}) ->
```
ieQu1, line 1642: When subscription dies on the DS side, the session gets a DOWN message. Therefore now we should pass DOWN messages to the session.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -569,251 +994,95 @@ to_BQ1(Key, SRS, S = #s{bq1 = BQ1}) ->
...
+                it_begin = Iterator,
+                it_end = Iterator,
+                sub_state_id = CurrentSubState
+            },
+            {
+                [Key],
+                emqx_persistent_session_ds_state:put_stream(Key, NewStreamState, S)
+            };
+        {error, Class = unrecoverable, Reason} ->
+            %% TODO: Handling of recoverable errors is non-trivial,
```
ieQu1, line 1072: Fix todo

apps/emqx/test/emqx_persistent_session_ds_fuzzer.erl
```diff
@@ -0,0 +1,738 @@
...
+     }.
+
+%% @doc Proper generator for `emqtt:connect' parameters:
+connect_(S) ->
+    ?LET(
+        ReceiveMaximum,
+        range(1, 32),
+        begin
+            DynamicOpts = #{
+                properies => #{'Receive-Maximum' => ReceiveMaximum}
```
thalesmg, line 180:
```suggestion
                properties => #{'Receive-Maximum' => ReceiveMaximum}
```

apps/emqx/test/emqx_persistent_session_ds_fuzzer.erl
```diff
@@ -0,0 +1,738 @@
...
+%% @doc Proper generator that creates a message in one of the topics.
+message(MsgId, #{subs := Subs}) ->
+    %% Create bias towards topics that the session is subscribed to:
+    TopicFreq = [{5, maps:keys(Subs)}, {1, ?topics}],
+    Topics = [{Freq, T} || {Freq, Topics} <- TopicFreq, T <- Topics],
+    ?LET(
+        {Topic, From, QoS},
+        {frequency(Topics), oneof(?publishers), qos()},
+        #message{
+            id = <<>>,
```
thalesmg, line 196:
```suggestion
            id = MsgId,
```
?

apps/emqx/test/emqx_persistent_session_ds_fuzzer.erl
```diff
@@ -0,0 +1,738 @@
...
+            %% Ack:
+            case QoS of
+                ?QOS_0 ->
+                    ok;
+                ?QOS_1 ->
+                    ?tp(info, ?sessds_test_out_puback, #{packet_id => PID}),
+                    emqtt:puback(client_pid(), PID);
+                ?QOS_2 ->
+                    ?tp(info, ?sessds_test_out_pubrec, #{packet_id => PID}),
+                    emqtt:pubrec(client_pid(), PID)
```
thalesmg, line 327: Maybe occasionally replying with the wrong ack (i.e., puback for qos =/= 1, pubrec/pubcomp for qos =/= 2) might help simulate [rogue/buggy clients](https://github.com/emqx/emqx/issues/10215)?

ieQu1: Yes, it should be possible to extend the fuzzer, but I'd prefer to do it in a later PR.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -537,6 +580,12 @@ count(DB) ->
 timestamp_us() ->
     erlang:system_time(microsecond).
 
+%% @doc Version of `emqx_topic:words()' that doesn't transform empty
+%% topic into atom ''
+-spec topic_words(binary()) -> topic().
+topic_words(<<>>) -> [];
+topic_words(Bin) -> emqx_topic:words(Bin).
```
thalesmg, line 587: Q: is the intention here really to prevent only `<<>>` from becoming `['']`?  Non-empty topics with empty levels will also violate the `emqx_ds:topic()` typespec by returning `''` atoms instead of always being `list(binary())`.

Ex:

```elixir
iex(emqx@127.0.0.1)11> :emqx_topic.words("/")
[:"", :""]
```

apps/emqx_durable_storage/src/emqx_ds_beamformer_rt.erl
```diff
@@ -0,0 +1,360 @@
...
+    Topic = emqx_topic:tokens(Msg#message.topic),
+    emqx_ds_beamformer_waitq:matching_keys(Stream, Topic, Queue).
+
+queue_drop(Queue, #sub_state{stream = Stream, topic_filter = TF, req_id = ID, client = Client}) ->
+    emqx_ds_beamformer_waitq:delete(Stream, TF, {node(Client), ID}, Queue).
+
+%% queue_update(Queue) ->
+%%     fun(_NextKey, Req = #sub_state{stream = Stream, topic_filter = TF, req_id = ReqId}) ->
+%%         emqx_ds_beamformer_waitq:insert(Stream, TF, ReqId, Req, Queue)
+%%     end.
```
thalesmg, line 333: :scissors: ?

apps/emqx_durable_storage/src/emqx_ds_beamformer_catchup.erl
```diff
@@ -0,0 +1,429 @@
...
+    process_batch(S, Stream, TopicFilter, Rest, Candidates, Beams).
+
+%% It's always worth trying to fulfill the oldest requests first,
+%% because they have a better chance of producing a batch that
+%% overlaps with other pending requests.
+%%
+%% This function implements a heuristic that tries to find such poll
+%% request. It simply compares the keys (and nothing else) within a
+%% small sample of pending polls, and picks request with the smallest
+%% key as the starting point.
```
thalesmg, line 356: Q: this suggests that the oldest request has the smallest key, but since `keypos = 1` for the queue, but this field apparently is a stream according to `?queue_elem`;  wouldn't the oldest request have the smallest `StartKey` (i.e., `StartKey` should be the first element in the `?queue_elem` tuple)?

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -413,6 +431,68 @@ update_iterator(_Shard, _Data, OldIter, DSKey) ->
...
+                {ok, #it{last_key = NextTS}, [_]} when NextTS =< FastForwardTo ->
+                    {error, unrecoverable, has_data};
+                {ok, It, _} ->
+                    {ok, It};
+                Err ->
+                    Err
+            end
+    end.
+
+-record(match_ctx, {compressed_topic, stream_key}).
```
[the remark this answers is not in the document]
ieQu1: While I agree that having such section implies all records should be defined there, this particular one is highly contextual for these specific functions.

apps/emqx_durable_storage/src/emqx_ds.erl
ieQu1: Right... That's a copy of an old function from skipstream_lts, though. I wouldn't touch it just yet. Instead clarify the behavior everywhere.

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -173,108 +244,160 @@
...
+%% Global builder:
+-record(beam_builder, {
+    cbm :: module(),
+    sub_tab :: ets:tid(),
+    lagging :: boolean(),
+    shard_id :: _Shard,
+    queue_drop :: fun(),
+    queue_update :: fun(),
+    global_n_msgs = 0 :: non_neg_integer(),
+    per_node = #{} :: #{node => #beam_builder_node{}}
```
thalesmg, line 286: Tiny nit: 
```suggestion
    per_node = #{} :: #{node() => #beam_builder_node{}}
```

apps/emqx_durable_storage/src/emqx_ds_beamformer_rt.erl
```diff
@@ -0,0 +1,360 @@
...
+            exit(retry);
+        Other ->
+            Pack =
+                case Other of
+                    {ok, end_of_stream} ->
+                        end_of_stream;
+                    {error, unrecoverable, _} ->
+                        Other
+                end,
+            Ids = emqx_ds_beamformer_waitq:matching_keys(Stream, ['#'], Queue),
```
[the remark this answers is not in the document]
ieQu1: Can you please clarify this comment?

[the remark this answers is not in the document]
ieQu1: Topics with `queue`, `share` and `sys` prefixes should not end up in the storage. This line just means give all subscribers for the stream.

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -295,590 +418,952 @@ poll(Node, ReturnAddr, Shard, Iterator, Opts = #{timeout := Timeout}) ->
...
+                    false ->
+                        %% This subscription is not keeping up
+                        %% with the acks. Freeze the
+                        %% subscription by dropping it from
+                        %% the worker's active queue:
+                        Flags = Flags0 bor ?DISPATCH_FLAG_STUCK,
+                        disown_stuck_subscription(DBShard, SubTab, SubS),
+                        QueueDrop(SubS0)
+                end,
+                DispatchMask = emqx_ds_dispatch_mask:enc_finalize(BatchSize + 1, MaskEncoder),
```
thalesmg, line 1144: Q: why the `+ 1` here, since (apparently) there's no message being appended to the batch?  Maybe add a comment explaining it?

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -295,590 +418,952 @@ poll(Node, ReturnAddr, Shard, Iterator, Opts = #{timeout := Timeout}) ->
...
+) ->
+    ets:delete(fc_tab(DBShard), SubId),
+    ets:delete(SubTab, SubId),
+    case ets:take(owner_tab(DBShard), SubId) of
+        [#owner_tab{owner = Owner, mref = MRef}] ->
+            %% Remove client monitoring:
+            demonitor(MRef, [flush]),
+            ets:delete(MTab, MRef),
+            %% Remove stuck subscription (if stuck):
+            ets:delete(disowned_sub_tab(DBShard), SubId),
```
thalesmg, line 1231: Since we get `stuck_sub_tab = SubTab` from the gen_statem data, it seems that such table is already [what's in the gvar](https://github.com/emqx/emqx/blob/fe6e9966ec532166549afefd03d895fd5729cceb/apps/emqx_durable_storage/src/emqx_ds_beamformer.erl#L726-L728).  If really so, we have already deleted the subscription in the 2nd line of this function.

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -295,590 +418,952 @@ poll(Node, ReturnAddr, Shard, Iterator, Opts = #{timeout := Timeout}) ->
...
+%% @doc Send fake monitor messages to all subscribers registered in
+%% the table, nudging them to resubscribe.
+on_worker_down(InheritedSubTab) ->
+    MS = {#sub_state{client = '$1', req_id = '$2', _ = '_'}, [], [{{'$1', '$2'}}]},
+    notify_down_loop(ets:select(InheritedSubTab, [MS], 100)).
+
+notify_down_loop('$end_of_table') ->
+    ok;
+notify_down_loop({Matches, Cont}) ->
+    _ = [Pid ! {'DOWN', Ref, process, self(), worker_crash} || {Pid, Ref} <- Matches],
```
thalesmg, line 1287: Nit: if we're faking a `DOWN` signal from the worker that has just died (and beamformer received in the real `DOWN` signal), wouldn't it be more "realistic" to use the worker pid here instead of beamformer's?

ieQu1: The intention is to hide the presence of workers from the client.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -1524,68 +1518,113 @@ maybe_set_offline_info(S, Id) ->
...
-                    _ ->
-                        {[], SchedS0}
-                end,
-            S1 = emqx_persistent_session_ds_state:put_seqno(SeqNoKey, SeqNo, S0),
+        {ok, Inflight} when Track =:= pubrec ->
+            S = put_seqno(SeqNoKey, SeqNo, S0),
+            Session = Session0#{inflight := Inflight, s := S},
+            {ok, Msg, schedule_delivery(Session)};
+        {ok, Inflight1} ->
+            {UblockedStreams, S1, SchedS} =
```
thalesmg, line 1583: Minor nit:

```suggestion
            {UnblockedStreams, S1, SchedS} =
```

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -295,590 +412,954 @@ poll(Node, ReturnAddr, Shard, Iterator, Opts = #{timeout := Timeout}) ->
...
+%% Handle wakeup subscription call:
+handle_event(
+    {call, From},
+    #wakeup_sub_req{id = SubId},
+    State,
+    D0 = #d{pending = Pending, stuck_sub_tab = Tab}
+) ->
+    case ets:take(Tab, SubId) of
+        [SubS = #sub_state{flowcontrol = {_MaxUnacked, ARef}}] ->
+            atomics:put(ARef, ?fc_idx_stuck, 0),
```
thalesmg, line 841: It seems this could race with the `atomics:put(_, stuck, 1)` from `keep_and_seqno` which is called externally via `beams_conclude/3`?

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -295,590 +412,954 @@ poll(Node, ReturnAddr, Shard, Iterator, Opts = #{timeout := Timeout}) ->
...
+send_out_final_term_to_node(DBShard, SubTab, Term, Node, Reqs) ->
+    Mask = emqx_ds_dispatch_mask:encode([true]),
+    Destinations = lists:map(
+        fun(
+            #sub_state{
+                client = Client, req_id = SubId, it = It, flowcontrol = FC
+            }
+        ) ->
+            {_MaxUnacked, ARef} = FC,
+            SeqNo = atomics:add_get(ARef, ?fc_idx_seqno, 1),
```
thalesmg, line 1161: Q: could this also race with `atomics:add_get(ARef, ?fc_idx_seqno, Nmsgs)` from `keep_and_seqno`?

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -295,590 +418,952 @@ poll(Node, ReturnAddr, Shard, Iterator, Opts = #{timeout := Timeout}) ->
...
+    {keep_state, D, {reply, From, Ret}};
+%% Handle disown request:
+handle_event(
+    cast,
+    #disown_stuck_req{sub_state = SubState},
+    _State,
+    #d{dbshard = DBShard}
+) ->
+    #sub_state{req_id = SubRef} = SubState,
+    case ets:update_element(owner_tab(DBShard), SubRef, {#owner_tab.owner, undefined}) of
```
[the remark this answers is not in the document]
ieQu1: What would be the scenario for such race condition?

This event is sent by the worker after it disowns the subscription.

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
ieQu1: What is the precise scenario for the race condition? 

This event is _only_ triggered by a client that _knows_ that its subscription is stuck (i.e. ?fc_idx_stuck` = 1).

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
ieQu1: This function is called from the same process that calls beams_conclude -> keep_and_seqno.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_shared_subs_agent.erl
```diff
@@ -107,34 +109,50 @@ open(Topics, Opts) ->
     ?shared_subs_agent:open(Topics, Opts).
 
 -spec pre_subscribe(t(), share_topic_filter(), emqx_types:subopts()) -> ok | {error, term()}.
+pre_subscribe(undefined, _, _) ->
```
ieQu1, line 112: This is not great.

Mini-rant: by silencing dialyzer warnings you just kick the problem down the line and let someone else to deal with it. Consider an example:

```
module A:

-dialyzer({nowarn_function, foo/1}).
foo(_) ->
  %% Noreturn, gets silenced
  1 = foo.
```

```
module B:

bar(X) ->
  case X of
    true ->
      A:foo(X), %% Since dialyzer figured out that A:foo(..) -> no_return(), it concludes that the following code is unreachable:
      C:baz(X);
    false ->
      C:baz(X)
  end.
```

```
module C:

baz(false) ->
  ...;
baz(true) ->
  %% Dialyzer will simply report that this clause is unreachable.
  ...
```

As you can see, a silenced success typing error in one module can pop up however many function calls apart, depending on how unlucky you are.

apps/emqx_durable_storage/include/emqx_ds.hrl
```diff
@@ -42,12 +42,23 @@
 
 -record(poll_reply, {
     ref :: reference(),
-    userdata,
-    payload :: emqx_ds:next_result() | poll_timeout
+    payload :: emqx_ds:next_result() | poll_timeout,
```
ieQu1, line 45: poll_timeout is unused

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -569,251 +1000,95 @@ to_BQ1(Key, SRS, S = #s{bq1 = BQ1}) ->
...
-) ->
-    ?tp(sessds_stream_state_trans, #{
-        key => Key,
-        to => u
-    }),
-    S#s{
-        ready = del_ready(Key, R),
-        pending = maps:remove(Key, P)
-        %% We do not remove the stream from the blocked queues.
-        %% It will be removed on seqno release.
```
savonarola, line 602: Seems we have lost this behaviour 🤔 

Previous logic wat the following:

* When a client unsubscribes, it may have unacked publishes received.
* We expect that the client will still ack everything from the inflight buffer.
* When it finally acks the last inflight message from the SRS span, we call on_seqno_release, detect "unblock" and may report the consistent progress (call `emqx_persistent_session_ds_shared_subs:on_streams_replay`), even for an unsubscribed stream.
* Having received consistent progress for an unsubscribed stream, the shared subs agent releases the stream.

Now, when a session unsubscribed, it won't receive any  and fully replayed streams from `emqx_persistent_session_ds_stream_scheduler:on_seqno_release`, and won't be able to report a consistent progress.

apps/emqx/src/emqx_persistent_session_ds.erl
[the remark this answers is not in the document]
ieQu1: I'd prefer to avoid making scheduler a god module. It already handles too much in my opinion.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -252,212 +305,253 @@ find_replay_streams(S) ->
...
+    #poll_reply{ref = Ref, size = Size, seqno = SeqNo} = Reply,
+    case SchedS#s.ds_subs of
+        #{Ref := #ds_sub{seqno = Atomic, stream_key = StreamKey}} ->
+            case atomics:add_get(Atomic, ?ix_seqno, Size) of
+                SeqNo ->
+                    %% SeqNos match:
+                    {true, StreamKey, SchedS};
+                Other ->
+                    %% SeqNos don't match, and we've just corrupted
+                    %% the counter by `add_get'. Resubscribe now:
```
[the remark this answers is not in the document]
ieQu1: Seqnos are frequently changing mutable variables. My profiling shows that storing them in an immutable map creates pressure on GC.

apps/emqx/src/emqx_persistent_session_ds.erl
```diff
@@ -224,19 +200,32 @@
...
+    %% previous messages.
+    inflight := emqx_persistent_session_ds_inflight:t(),
     stream_scheduler_s := emqx_persistent_session_ds_stream_scheduler:t(),
     %% In-progress replay:
-    %% List of stream replay states to be added to the inflight buffer.
+    %% List of stream replay states to be added to the inflight.
     replay := [{_StreamKey, stream_state()}, ...] | undefined,
-    new_stream_subs := emqx_persistent_session_ds_subs:new_stream_subs(),
+    %% Flags:
+    ?TIMER_DRAIN_INFLIGHT := boolean(),
```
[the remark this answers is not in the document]
ieQu1: The message it sends emulates a timeout message due to channel implementation details.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_subs.erl
```diff
@@ -74,27 +85,17 @@
         _ => _
     },
     %% Optional field that is added when subscription state becomes
-    %% outdated:
-    superseded_by => subscription_state_id()
+    %% outdated (note: do not use the value, as GC may delete
+    %% subscription states referenced by `superseded_by'):
+    superseded_by => subscription_state_id(),
+    %% Optional field used by shared subscriptions:
+    share_topic_filter => #share{}
```
savonarola, line 92: As far as I see, it is not used now actually. I'll make a task for myself to consistently remove this field after merge.

apps/emqx/test/emqx_persistent_session_ds_fuzzer.erl
```diff
@@ -0,0 +1,741 @@
...
+        }
+    ).
+
+publish_(S = #{message_seqno := SeqNo}) ->
+    ?LET(
+        BatchSize,
+        range(1, 10),
+        ?LET(
+            Msgs,
+            [message(I, S) || I <- lists:seq(SeqNo, SeqNo + BatchSize)],
```
savonarola, line 220: Nit: some clarification would be helpful; for the same thing, we have `MsgId` and `SeqNo` names and `emqx_ds:time()` type.

apps/emqx_durable_storage/src/emqx_dsx.erl_
savonarola, line 1: Nit: rename back or remove?

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -173,108 +252,201 @@
...
 
--define(fulfill_loop, fulfill_loop).
--define(housekeeping_loop, housekeeping_loop).
+%% Per-node builder:
+-record(beam_builder_node, {
+    global_n_msgs :: non_neg_integer() | undefined,
+    n_msgs = 0 :: non_neg_integer(),
+    %% List of messages and keys, reversed:
+    pack = [] :: [{emqx_ds:message_key(), emqx_types:message()}],
+    subs = #{} :: #{reference() => #beam_builder_sub{}}
```
savonarola, line 281: Just curious: is there any difference (e.g., in data size) expected with transposed packing (i.e., mapping messages to the bitmap-encoded set of subscribers)?

ieQu1: I am not sure I understand the question.

apps/emqx_durable_storage/src/emqx_dsx.erl_
ieQu1: I intend to fix this module in a later PR.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
ieQu1: Fixed. Now the scheduler returns unblocked unsubscribed streams.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -252,212 +305,253 @@ find_replay_streams(S) ->
...
         _ ->
-            %% Send poll request:
-            PollOpts = PollOpts0#{reply_to => Ref},
-            {ok, Ref} = emqx_ds:poll(?PERSISTENT_MESSAGE_DB, Iterators, PollOpts),
+            ?tp(warning, ?sessds_unexpected_stream_notification, #{ref => Ref}),
+            {[], S0, SchedS0}
+    end.
+
+%% @doc Verify sequence number of DS subscription reply
+-spec verify_reply(#poll_reply{}, emqx_persistent_session_ds_state:t(), t()) ->
```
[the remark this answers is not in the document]
ieQu1: Fixed.

apps/emqx/src/emqx_connection.erl
```diff
@@ -935,43 +935,41 @@ handle_outgoing(Packets, State = #state{channel = _Channel}) ->
...
-                ?LOG(error, #{
-                    input_packet => Packet,
-                    exception => Reason,
-                    stacktrace => Stacktrace
-                }),
-                erlang:error(?FRAME_SERIALIZE_ERROR)
-        end
+    send(serialize_and_inc_stats(State, Packet), State).
+
+serialize_and_inc_stats(#state{serialize = Serialize}, Packet) ->
```
ieQu1, line 942: I'm not sure why this function was carried in the first place, but this simplified version leads to a marginal improvement in throughput.

apps/emqx/src/emqx_persistent_session_ds.erl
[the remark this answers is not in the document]
ieQu1: P.S. My idea for separation of session into modules is the following:

- Portions of functionality that are logically self-contained can be extracted into modules.
- Main module orchestrates them all.
- There's at most one level of indirection, i.e. session -> component, rather than session -> component -> component -> ...

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -511,329 +618,695 @@ is_fully_acked(Srs, S) ->
...
+) ->
+    ret().
+init_for_subscription(
+    TopicFilterBin,
+    Subscription,
+    S0,
+    SchedS0 = #s{new_stream_subs = NewStreamSubs, sub_metadata = SubMetadata}
+) ->
+    #{id := SubId, start_time := StartTime} = Subscription,
+    TopicFilter = emqx_topic:words(TopicFilterBin),
```
keynslug, line 693: Q: Shouldn't this use `emqx_ds:topic_words/1`?

keynslug (commented):
Impressive work.

apps/emqx_durable_storage/src/emqx_ds_storage_reference.erl
```diff
@@ -253,41 +259,43 @@ scan_stream(Shard, S, _Stream, TopicFilter, LastSeenKey, BatchSize, TMax, IsCurr
...
-                Acc#{Topic => 1};
-            ({delete, _Msg}, Acc) ->
-                Acc
-        end,
-        #{},
-        Messages
-    ),
-    [{#stream{}, T} || T <- maps:keys(Topics)].
+batch_events(_Shard, _, _Messages) ->
+    %% FIXME:
```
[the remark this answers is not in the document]
ieQu1: I think reference backend should be phased out: it was created mostly because bitfield_lts was *weird*. Skipstream LTS doesn't reorder messages, so there's no need for a simple "test" layout.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
```diff
@@ -39,6 +39,9 @@
     unpack_iterator/3,
     scan_stream/8,
     message_matcher/3,
+    fast_forward/5,
+    message_match_context/5,
+    iterator_match_context/3,
```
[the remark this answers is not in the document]
ieQu1: No.

1. Bitfield LTS is unlikely to be used.
2. Reference layout is test-only, I've switched all tests to skipstream so there's no need for it anymore.

We should keep these layouts in read-only mode for a while.

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -295,590 +467,969 @@ poll(Node, ReturnAddr, Shard, Iterator, Opts = #{timeout := Timeout}) ->
...
+            },
+            %% Store another reference to the flow control object in a
+            %% different table readable by the client during suback:
+            ets:insert(fc_tab(Shard), #fctab{
+                sub_ref = SubRef, max_unacked = MaxUnacked, aref = FlowControl
+            }),
+            %% Create an entry in the owner table:
+            ets:insert(owner_tab(Shard), #owner_tab{sub_ref = SubRef, mref = MRef}),
+            %% Schedule dispaching to the catchup worker:
+            D = D0#d{pending = [SubState | Pending]},
```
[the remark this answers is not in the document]
ieQu1: Right. Something was lost during refactoring.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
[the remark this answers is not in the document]
ieQu1: I'll follow-up by disabling these layouts in the schema.

apps/emqx_durable_storage/src/emqx_ds_beamformer.erl
```diff
@@ -295,590 +467,969 @@ poll(Node, ReturnAddr, Shard, Iterator, Opts = #{timeout := Timeout}) ->
...
+            case emqx_ds_beamformer_catchup:enqueue(Worker, Reqs) of
+                ok ->
+                    Unmatched;
+                Error ->
+                    ?tp(
+                        error,
+                        emqx_ds_beamformer_delegate_error,
+                        #{shard => DBShard, worker => Worker, error => Error}
+                    ),
+                    Reqs ++ Unmatched
```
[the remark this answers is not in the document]
ieQu1: I want to fix it later, handling intermittent errors (like timeouts) is not trivial 'cause they're likely to lead to the double ownership problem. 

Sync call with infinite timeout is not the final solution here. I'll add a comment.

apps/emqx_durable_storage/src/emqx_ds_storage_skipstream_lts.erl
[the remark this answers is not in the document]
keynslug: > Bitfield LTS is unlikely to be used.

It's currently employed as a storage layout for shared subscriptions, and I think is still usable in similar use cases (giving smaller storage footprint). Granted these use cases currently don't need subscriptions.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1647,14 +1647,18 @@ handle_info({disconnect, ReasonCode, ReasonName, Props}, Channel) ->
...
     Session = emqx_session:handle_info(Info, Session0, ClientInfo),
-    {ok, Channel#channel{session = Session}}.
+    Channel = Channel0#channel{session = Session},
+    case Info of
+        {'DOWN', Ref, process, Pid, Reason} ->
```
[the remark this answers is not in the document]
keynslug: Right. What I meant is: if the session took ownership and handled an event, ideally it shouldn't probably be passed into the hooks for further processing.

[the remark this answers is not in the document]
ieQu1: There's no way to tell whether if session did anything to the event.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6759)

PR #14505, opened 2025-01-08
ci: attempt to reduce flakiness

thalesmg:
The exact reason is not yet fully understood, but adding this flag may help reduce the
need to repeat this workflow.


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6705)

PR #14508, opened 2025-01-08
Perf: emqx_cm and emqx_broker_helper 

### LINKED ISSUE #14482: One of the nodes in EMQX has expired sessions that are not being cleared, which leads to the node experiencing an OOM (Out of Memory) condition
### What happened?

Environment:

- EMQX Version: 5.7.2
- Operating System Version: CentOS 7
- Cluster Nodes: 4

The client uses temporary sessions with a Session Expiry Interval of 1 minutes and a QoS level of 1.
There are approximately millions of long connections, with a message publishing and receiving frequency of less than 100 messages per second. Each message is smaller than 1KB.

After the service starts, within a few hours, one of the nodes occasionally reaches 98% memory usage and then crashes.
image

image
image

Using Crashdump Viewer to analyze the dump file, the result is that one process is occupying 36GB of memory.
image
image

When a node crashes, the total number of connections in the cluster exceeds 20 million (with only about 4 million active connections).
image
image
image

The session expiration time is set to 1 minutes, and logically, the ratio of total connections to active connections shouldn't be this high.
Upon querying the "client" information in the dashboard, it was found that sessions on one node could not be cleared, leading to an accumulation of sessions on that node. The final result is:

- Node 1 has 160,000 disconnected sessions.
- Node 2 also has 160,000.
- Node 3 has 1.6 million disconnected sessions and keeps increasing, ultimately leading to OOM and the node crashing.
- Node 4 also has 160,000.

Therefore, there must be an issue with the 'clearing expired sessions' task on Node 3. 
Please provide assistance on how to locate the specific problem.

Memory status of Node 3 is as follows:

image

1. After a session expires (disconnected for more than 1 minute), it should be cleared.
2. The memory of the machine should not keep increasing, eventually leading to an OOM.

_No response_

_No response_

### EMQX version

5.7.2

### OS version

NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"

### Log files

_No response_

zmstone:
Fixes #14482

Release version: v/e5.8.5

## Summary

Prior to this change, `emqx_utils:drain_down(?BATCH_SIZE)` might repeatedly scan process mailbox to drain the `DOWN` messages in `emqx_broker_helper` and `emqx_cm`.
This PR changes the receive loop to always collect the head of the first `BATCH_SIZE` messages in the mailbox including both register requests and `DOWN` messages.
Also, `BATCH_SIZE` is change from `100,000` to `1000` for smaller term collection.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_broker_helper.erl
+++ b/apps/emqx/src/emqx_broker_helper.erl
@@ -167,11 +167,11 @@ init([]) ->
     {ok, #{}}.
 
 handle_call(Req, _From, State) ->
-    ?SLOG(error, #{msg => "emqx_borker_helper_unexpected_call", call => Req}),
+    ?SLOG(error, #{msg => "emqx_broker_helper_unexpected_call", call => Req}),
     {reply, ignored, State}.
 
 handle_cast(Msg, State) ->
-    ?SLOG(error, #{msg => "emqx_borker_helper_unexpected_cast", cast => Msg}),
+    ?SLOG(error, #{msg => "emqx_broker_helper_unexpected_cast", cast => Msg}),
     {noreply, State}.
 
 handle_info({register_sub, SubPid, SubId}, State) ->
@@ -181,7 +181,7 @@ handle_info({'DOWN', _MRef, process, SubPid, _Reason}, State) ->
     ok = collect_and_handle([], [SubPid]),
     {noreply, State};
 handle_info(Info, State) ->
-    ?SLOG(error, #{msg => "emqx_borker_helper_unexpected_info", info => Info}),
+    ?SLOG(error, #{msg => "emqx_broker_helper_unexpected_info", info => Info}),
     {noreply, State}.
 
 terminate(_Reason, _State) ->
--- a/apps/emqx/src/emqx_broker_helper.erl
+++ b/apps/emqx/src/emqx_broker_helper.erl
@@ -48,6 +48,9 @@
     code_change/3
 ]).
 
+%% Internal APIs
+-export([clean_down/1]).
+
 -ifdef(TEST).
 -compile(export_all).
 -compile(nowarn_export_all).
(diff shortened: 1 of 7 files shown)


ieQu1 (approved):

thalesmg (dismissed):

apps/emqx/src/emqx_broker_helper.erl
```diff
@@ -194,6 +195,45 @@ code_change(_OldVsn, State, _Extra) ->
...
+    after 0 ->
+        {Reg, Down}
+    end.
+
+handle_register(Reg) ->
+    lists:foreach(fun do_handle_register/1, Reg).
+
+do_handle_register({SubId, SubPid}) ->
+    true = (SubId =:= undefined) orelse ets:insert(?SUBID, {SubId, SubPid}),
+    _ = erlang:monitor(process, SubPid),
```
keynslug, line 228: Q: Can we simplify further, and just rely on a single monitor set up by `emqx_cm`? Say, make it submit a composite job to the generic pool.

apps/emqx/src/emqx_broker_helper.erl
```diff
@@ -146,6 +150,8 @@ table_size(Tab) when is_atom(Tab) -> ets:info(Tab, size).
 %%--------------------------------------------------------------------
 
 init([]) ->
+    process_flag(priority, high),
```
keynslug, line 153: Nit. I'd propose to put that behind a configuration flag, feels a bit risky to me to distort the scheduling like that. Or maybe introduce a pool of helpers?

zmstone: We can maybe remove this flag for helper process, only set it for emqx_cm.
from the observations, it is mostly always on top when sort by reductions.

apps/emqx/src/emqx_broker_helper.erl
zmstone: that'd be the next PR.

zmstone:
fixed test cases with this commit: `
test: ensure emqx async pool worker is started for test` rebased as the last commit.

apps/emqx/src/emqx_broker_helper.erl
keynslug: Right. Also, I'm thinking all the other improvements (off-heap, no pmon, no selective receive) might be already enough to severely decrease the average reductions and memory management cost per disconnect?

IIUC `high` priority should not change a lot when CPU is not completely utilized, but when it is, cleaning up disconnects will take disproportionate priority over all other tasks, and the less there are CPU cores the more pronounced the effect will be. I guess we could consider this is better than accumulating huge message queue, but it doesn't actually guarantee that it stays bounded. For that I guess we'd need to introduce some form of backpressure during `register_sub`?

keynslug (approved):

shengbinxu:
Hi, @zmstone , thank you for your hard work.
Is there an estimated release time for v5.8.5? We want to use this version in the production environment as soon as possible.

zmstone:
> Hi, @zmstone , thank you for your hard work. Is there an estimated release time for v5.8.5? We want to use this version in the production environment as soon as possible.

Hi @shengbinxu 
The planed release date is Feb.6

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5843)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5844)

shengbinxu:
> > Hi, @zmstone , thank you for your hard work. Is there an estimated release time for v5.8.5? We want to use this version in the production environment as soon as possible.
> 
> Hi @shengbinxu The planed release date is Feb.6

@zmstone hi,zmstone, Happy New Year, has the release date changed?

zmstone:
Hi @shengbinxu 
Happy new year to you too.

Sorry that I remembered it wrong.
Feb.6 was our code-freeze day for 5.8.5. 
The release date is actually planed on Feb.20.

apps/emqx/src/emqx_broker_helper.erl
zmstone: https://github.com/emqx/emqx/pull/16413

PR #12920, opened 2024-04-23
fix: STOMP heartbeat

cisiqo:
Fixes https://github.com/emqx/emqx/issues/12911

Fix server not send heartbeat to client and client send heartbeat to server, the server keep alive.

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [ ] Added tests for the changes
- [ ] Added property-based tests for code which performs user input validation
- [ ] Changed lines covered in coverage report
- [ ] Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [ ] For internal contributor: there is a jira ticket to track this change
- [ ] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket
- [ ] Schema changes are backward compatible

## Checklist for CI (.github/workflows) changes

- [ ] If changed package build workflow, pass [this action](https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)
- [ ] Change log has been added to `changes/` dir for user-facing artifacts update

Issue #12911, opened 2024-04-22
STOMP GW does not send heartbeat to client


shinedd:
### What happened?

I am using obuspa as a stomp client to test with emqx stomp gw, and i get a error message:
![image](https://github.com/emqx/emqx/assets/106661626/49e3d4d6-d769-425f-8b09-f5d87b3c58d1)

### What did you expect to happen?

Could emqx stomp gw support sending heartbeat to stomp client

### EMQX version

sysdescr  : EMQX
version   : 5.5.1
datetime  : 2024-04-22T19:25:59.961416581+08:00
uptime    : 10 days, 9 hours, 35 minutes, 26 seconds

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

HJianBo:
Hi, @shinedd I'm not sure what's kinds of `Heartbeat` you mentioned, because the standard stomp v1.2 protocol spec doesn't have heartbeat definition https://stomp.github.io/stomp-specification-1.2.html#Client_Frames

Could you give me some code examples to reproduce that?

shinedd:
I'm new for stomp, i just find this problem during testing with tr369 client:
https://github.com/BroadbandForum/obuspa
And when i change cx > 0 or cy > 5000
![image](https://github.com/emqx/emqx/assets/106661626/189f1342-c7b4-44f8-b1f2-673da472cc84)
I will get this error log

shinedd:
When i set cy to 5000, i found the heartbeat interval is about 15s in captured package
![image](https://github.com/emqx/emqx/assets/106661626/dbd3702b-f1b5-4e0d-83e4-494511d26fb3)

cisiqo:
@shinedd yes, my branch had fix the bug, you can use to test.

cisiqo:
Oh, no. Please pay attention to this https://github.com/emqx/emqx/pull/12920#issuecomment -2167477207

shinedd:
I have tested with this branch https://github.com/cisiqo/emqx/tree/patch-6
The test results are the same

cisiqo:
Try it again, i have fix it.

shinedd:
![image](https://github.com/emqx/emqx/assets/106661626/b15ac89c-6f71-4af3-8af4-e6a6af647cb0)
it seems work, tks

PR #14528, opened 2025-01-11
fix(stomp): Received client heartbeat reset the count

cisiqo:
Fixes [https://github.com/emqx/emqx/pull/12920](https://github.com/emqx/emqx/pull/12920)
Fixes [https://github.com/emqx/emqx/issues/12911](https://github.com/emqx/emqx/issues/12911)

Stomp client sending heartbeat to server, the connection always a living.

Release version: v5.8.4

## PR Checklist


apps/emqx_gateway_stomp/src/emqx_stomp_channel.erl
```diff
@@ -683,8 +683,7 @@ handle_in(
     ?PACKET(?CMD_HEARTBEAT),
     Channel = #channel{heartbeat = Heartbeat}
 ) ->
-    NewVal = emqx_pd:get_counter(recv_pkt),
-    NewHeartbeat = emqx_stomp_heartbeat:reset(incoming, NewVal, Heartbeat),
+    NewHeartbeat = emqx_stomp_heartbeat:reset(incoming, 0, Heartbeat),
```
zmstone, line 686: Not sure if I fully understand the code, but I believe the fix is to set `repeat` to `0`, but not to ground the `statval` (received packets).
https://github.com/emqx/emqx/blob/cf6e56aadfdb212406286db611524d8e01ff173f/apps/emqx_gateway_stomp/src/emqx_stomp_heartbeat.erl#L96-L101

Consider this case:
client sends one heartbeat, and never send anything again.
Since `statval` is set to `0`, the heartbeat timer check function `emqx_stomp_heartbeat:check/3` will never return `{error, timeout}`

zmstone (commented):
Hi @cisiqo 
Thank your for the PR.
I have added a comment.

apps/emqx_gateway_stomp/src/emqx_stomp_channel.erl
cisiqo: I have tested this case and it is perfectly fine. You can also test it. 
When the client times out, `NewVal` and `OldVal` are equal.

In the master branch, you can also test this case:
Client timing continuous sending heartbeat.

zmstone: You are right. Setting `statval` to 0 can indeed avoid timeout.

But the right fix is to set repeat=0 which is designed as the 'backoff' purpose, initialize at 0 and check fail when >=1 is to allow the next heart to arrive between 1 ~ 2 intervals (statistically 1.5)

cisiqo: As you said, I modified and tested it and it does fix the problem.
I see that you have a full understanding of the issue and you should fix it better than I did.

zmstone:
Thank you @cisiqo for testing it.

PR #14540, opened 2025-01-13
Introduce histograms for measuring latency

savonarola:
Fixes [EMQX-1460](https://emqx.atlassian.net/browse/EMQX-1460)

Release version: 6.0.0

## PR Checklist

- [x] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket: https://emqx.atlassian.net/browse/EMQX-13834

[EMQX-1460]: https://emqx.atlassian.net/browse/EMQX-1460?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/emqx_auth/src/emqx_authn/emqx_authn_schema.erl
```diff
@@ -213,6 +218,20 @@ enable(importance) -> ?IMPORTANCE_NO_DOC;
...
+            ?HOCON(
+                ?R_REF(emqx_auth_cache_schema, config),
+                #{
+                    desc => ?DESC(authentication_cache),
+                    importance => ?IMPORTANCE_LOW,
+                    default => emqx_auth_cache_schema:default_config()
+                }
+            )},
+        {"total_latency_metric_buckets",
+            emqx_schema:histogram_buckets_sc(#{desc => ?DESC(authn_latency_buckets)})}
```
keynslug, line 233: Just to start a discussion. My (rather uneducated) opinion is that having such configuration options feels like micromanagement. Do you envision we'll eventually have similar option for each latency metric? Or will it be per-subsystem or similarly coarse granularity? To better understand: I assume that it's first and foremost for Prometheus consumers, but I wonder if it's something that the Dashboard wants to control, consume and render as well?

apps/emqx_utils/src/emqx_metrics_worker.erl
```diff
@@ -72,16 +76,35 @@
 -define(SAMPLING, 1).
 -endif.
 
+-define(DEFAULT_HIST_BUCKETS, [10, 50, 250, 1_000, 5_000, 25_000, 100_000, 500_000, 2_500_000]).
```
keynslug, line 79: Q: Why are these different from `emqx_schema` defaults?

apps/emqx/src/emqx_access_control.erl
```diff
@@ -266,10 +279,18 @@ format_retain_flag(true) ->
...
+run_hooks(Name, Args, Acc) when Name == 'client.authenticate'; Name == 'client.authorize' ->
     ok = emqx_metrics:inc(Name),
-    emqx_hooks:run_fold(Name, Args, Acc).
+    {Time, Value} = timer:tc(
+        fun() -> emqx_hooks:run_fold(Name, Args, Acc) end
+    ),
+    catch emqx_metrics_worker:observe_hist(
+        ?ACCESS_CONTROL_METRICS_WORKER,
+        Name,
+        total_latency,
```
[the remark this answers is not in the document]
zmstone: The decision is to instrument the entry-point for now.
For backend-specific latency (or plugin callback eval latency) can be added later (if needed).

apps/emqx_auth/src/emqx_authn/emqx_authn_schema.erl
savonarola: >if it's something that the Dashboard wants to control
As far as I understand from discussions, no, we do not plan to expose these histograms to the Dashboard. 

Additionally, as far as I understand, it is also quite challenging to implement smooth and correct aggregation of histograms across the cluster (and that would be reimplementing what Prometheus already does well 😅)

>feels like micromanagement
I somewhat agree 🫠 

Another option is to make a more universal endpoint for bucket configuration somewhere in the Prometheus app:
* It is already coupled with emqx's internals and knows all the metrics workers and metrics' names, so we will not need the `update_buckets` interface.
* As Prometheus is probably the only consumer of histograms, this endpoint may be naturally treated as _configuring the required view_ for Prometheus.

apps/emqx/src/emqx_schema.erl
```diff
@@ -3418,6 +3421,54 @@ parse_port(Port) ->
             {error, "bad_port_number"}
     end.
 
+histogram_buckets_sc(Meta0) ->
+    DefaultMeta = #{
+        default => <<"5, 10, 20, 50, 100, 200, 500, 1000, 2000, 5000, 10000, 20000">>,
```
zmstone, line 3426: suggest defaults: `10ms,100ms,1s,5s,30s,+Inf`

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
```diff
@@ -115,9 +115,9 @@ paths() ->
         "/authentication/:id/users",
         "/authentication/:id/users/:user_id",
         "/authentication/order",
-        "/authentication_cache",
-        "/authentication_cache/status",
-        "/authentication_cache/reset"
+        "/authentication/settings",
+        "/authentication/settings/node_cache/status",
```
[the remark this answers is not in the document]
savonarola: >Also node_cache sounds overly specific

For authz, we already had the `cache` thing, which stands for the in-process cache. So we needed something different for the node-wise cache (`node_cache`).

So for authn, I made the same name for the same thing, trying not to confuse the user.

apps/emqx/src/emqx_schema.erl
zmstone: I guess +Inf is added automatically, so skip it.

apps/emqx_auth/src/emqx_authn/emqx_authn_schema.erl
zmstone: I was in the previous discussions,
so i can add some: it's hard to provide a sane-default for everyone, e.g. some deployments might have a slow LDAP which takes 10s for each query, one may choose to create buckets like `5ms,1s,10s,20s` where 5ms is to cover the cache hits.

zmstone: I might have misunderstood your point.
I was commenting about why we need to make the buckets configurable, maybe your point is if we need to make buckets configurable for each different metrics, today we have `total_latency_metric_buckets` should we need to add `postgres_latency_metric_buckets` in the future ?
Maybe not.
So I'm for just `latecny_metric_buckets`.

zmstone: > I wonder if it's something that the Dashboard wants to control, consume and render as well

No, no intention to expose it for Dashboard use.

zmstone (approved):

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6729)

PR #14584, opened 2025-01-20
Authenticator App (TOTP 2FA) for Enterprise edition dashboard

zmstone:
Fixes [EMQX-13509](https://emqx.atlassian.net/browse/EMQX-13509)

Release version: v/e5.9.0

## Summary

See emqx_dashboard/README.md

## PR Checklist

- [x] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a [follow-up jira ticket](https://emqx.atlassian.net/browse/EMQX-13837)

[EMQX-13509]: https://emqx.atlassian.net/browse/EMQX-13509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/emqx_dashboard/test/emqx_dashboard_mfa_SUITE.erl
```diff
@@ -0,0 +1,342 @@
...
+init_per_suite(Config) ->
+    ?EE_ONLY(
+        begin
+            Apps = emqx_cth_suite:start(
+                [
+                    emqx,
+                    emqx_conf,
+                    emqx_management,
+                    emqx_mgmt_api_test_util:emqx_dashboard(
+                        "dashboard.listeners.http { enable = true, bind = 18083 }"
```
thalesmg, line 50: Nit: the `/0` version of this function already have this as the default config.

apps/emqx_dashboard/test/emqx_dashboard_mfa_SUITE.erl
```diff
@@ -0,0 +1,342 @@
...
+                    )
+                ],
+                #{work_dir => emqx_cth_suite:work_dir(Config)}
+            ),
+            {ok, _} = emqx_common_test_http:create_default_app(),
```
thalesmg, line 55: Nit: `emqx_mgmt_api_test_util:emqx_dashboard` already takes care of this.

apps/emqx_dashboard/test/emqx_dashboard_mfa_SUITE.erl
```diff
@@ -0,0 +1,342 @@
...
+    %% now EMQX is expecting viewer1 to provide a token for the first time
+    %% but login with a bad password should continue to fail
+    {ok, 401, BadPass1} = login(BadPwd),
+    ok = assert_return_code("BAD_USERNAME_OR_PWD", BadPass1),
+    ok = ExpectTotpMissingFn(true),
+    ok = ExpectBadTotpFn(BadTotp),
+    %% secret is still presented after a failed login with bad token
+    ok = ExpectTotpMissingFn(true),
+    %% try to login with good TOTP but bad password
+    %% this should mark the TOTP setup complete,
```
thalesmg, line 172: Should there be side-effects if the provided password is wrong?  :thinking:

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -238,7 +291,13 @@ login(post, #{body := Params}) ->
                 })};
         {error, R} ->
             ?SLOG(info, #{msg => "dashboard_login_failed", username => Username, reason => R}),
-            {401, ?BAD_USERNAME_OR_PWD, <<"Auth failed">>}
+            %% check if this error was created by emqx_dashboard_mfa
+            case emqx_dashboard_mfa:is_mfa_error(R) of
+                true ->
+                    {403, ?BAD_MFA_TOKEN, R};
```
thalesmg, line 297: Nit: bad MFA errors seem to be either missing token or bad token.  It would seem like the response status code for both would be 401 (Unauthorized), since authentication didn't complete, as opposed to 403 (Forbidden; authenticated but denied access to resource/action)?

rel/i18n/emqx_dashboard_api.hocon
```diff
@@ -90,4 +90,23 @@ backend.desc:
...
+"""Delete MFA state for a dashboard user.
+After the state is deleted, the user will have MFA explicitly disabled, to re-enabled MFA, call the `POST` method."""
+
+mfa_token.desc:
+"""Multifactor authenticaion token."""
+
+mfa_status.desc:
+"""MFA status.
+- `none`: no MFA enabled
+- `disabled`: MFA is disabled for the user
```
thalesmg, line 109: Q: is there a reason to differentiate between the two?

Perhaps the intention of `disabled` is to mean "MFA state/secret is already set up, but we don't require using it for this user"?

rel/i18n/emqx_dashboard_api.hocon
```diff
@@ -90,4 +90,23 @@ backend.desc:
...
+- Otherwis: The MFA mechanism in use, e.g. `totp`"""
```
thalesmg, line 110:
```suggestion
- Otherwise: The MFA mechanism in use, e.g. `totp`"""
```

apps/emqx_dashboard/README.md
```diff
@@ -15,3 +15,48 @@ Note, at this point EMQX Dashboard itself is an independent frontend project and
...
+
+Detailed steps are:
+
+- The 1st `/login`, the body only includes `username` and `password`, without providing an MFA token.
+  - If the user does not require MFA, then the process proceeds to password hash comparison.
+  - If the user has set up MFA, there are two scenarios:
+    - If an MFA token has never been submitted before (indicating that the Authenticator App needs to be initialized):
+      - The backend verifies the password. If the password is incorrect, it returns a `401` error to indicate bad username or password. If the password is correct, the process continues.
+      - A `403` error is returned, and the body like `{"code": "BAD_MFA_TOKEN", "message": {"error": "missing_mfa_token", "mechanism": "totp", "cluster": "EMQX Cluster Name", "secret": "ABCDJBSWY3DPEHPK3PXP"}}`.
+      - The Dashbaord uses this secret to generate a QR code for Authenticator App.
```
thalesmg, line 48:
```suggestion
      - The Dashboard uses this secret to generate a QR code for Authenticator App.
```

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -378,44 +451,153 @@ all_users() ->
...
+    case get_mfa_enabled_state(Username) of
+        {ok, State} ->
+            %% MFA is configured, and a token is provided
+            case emqx_dashboard_mfa:verify(State, MfaToken) of
+                ok ->
+                    %% Token is ok, no state change
+                    ok;
+                {ok, NewState} ->
+                    %% Token is ok, state changed, should update db
+                    {ok, ok} = set_mfa_state(Username, NewState),
```
thalesmg, line 556: Since this is a side effect that sets the MFA state, I think we should validate the password before doing this (currently it's the other way around).

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -378,44 +451,153 @@ all_users() ->
...
+            %% No MFA enabled, but provided with a token
+            %% Can happen if MFA is disabled for this user but the
+            %% dashboad is still providing a token for some reason,
+            %% proceed to password check
+            ok
+    end.
+
+%% Initialize MFA state if there is a default MFA settings configured.
+maybe_init_mfa_state(Username) ->
+    try emqx:get_config([dashboard, default_mfa]) of
```
thalesmg, line 572: Nit: maybe use `emqx:get_config/2` with a default of `none`?

apps/emqx_dashboard/test/emqx_dashboard_mfa_SUITE.erl
zmstone: TOTP secret is only provided when password was once correct.
The second login attempt providing a valid TOTP code indicates that the authenticator app is set up.
Otherwise the user will get presented with the QR code again.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
zmstone: changed to use 401

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
zmstone: TOTP secret is only provided when password was once correct.
The second login attempt providing a valid TOTP code indicates that the authenticator app is set up.
Otherwise the user will get presented with the QR code again.

thalesmg: Right, but it would add an extra layer of security if we require the password to be correct for any state altering side-effects to take place.  :thinking:

apps/emqx_dashboard_sso/test/emqx_dashboard_sso_cli_SUITE.erl
```diff
@@ -22,7 +22,7 @@
     )
 ).
 
-all() -> [t_add, t_passwd, t_del].
+all() -> [t_add, t_passwd, t_del, t_mfa].
```
thalesmg, line 25: Nit:

```suggestion
all() -> emqx_common_test_helpers:all(?MODULE).
```

?

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
thalesmg: I mean, it's multi-factor authn because there are multiple proofs given when authentication, e.g., password + TOTP.

thalesmg (dismissed):

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
zmstone: ok. fixed.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -378,44 +477,146 @@ all_users() ->
...
+check(Username, Password, MfaToken) ->
     case lookup_user(Username) of
         [#?ADMIN{pwdhash = PwdHash} = User] ->
-            case verify_hash(Password, PwdHash) of
-                ok -> {ok, User};
-                error -> {error, <<"password_error">>}
-            end;
+            IsPwdOk = (ok =:= verify_hash(Password, PwdHash)),
+            MfaVerifyResult = verify_mfa_token(Username, MfaToken, IsPwdOk),
+            check_mfa_pwd(MfaVerifyResult, IsPwdOk, User);
```
thalesmg, line 516: I think that we shouldn't even do anything further once we establish that the password is incorrect.

i.e.:

```suggestion
            maybe
                true ?= (ok =:= verify_hash(Password, PwdHash)) || {error, <<"password_error"},
                MfaVerifyResult = verify_mfa_token(Username, MfaToken),
                check_mfa_pwd(MfaVerifyResult, User)
            end;
```

thalesmg (approved):

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
zmstone: For TOTP, we can (and should) return token error over password error.
Otherwise a brute-force attacker can send any token to guess the password like without MFA.

For other mechanisms which costs to send the token (e.g. email / SMS, if we will need to implement), password needs to be checked before token.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6722)

PR #14599, opened 2025-01-22
async reply session takeover

qzhuyan:
New connection process has blocking call to wait for the old process exit.

with this change, new process will get async reply that can speed up the session takeover because
graceful_shutdown_transport is yet another blocking call to the transport may with timeouts.

Fixes <issue-or-jira-number>

Release version: v/e5.?

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_connection.erl
+++ b/apps/emqx/src/emqx_connection.erl
@@ -171,6 +171,9 @@
 
 -define(LOG(Level, Data), ?SLOG(Level, (Data)#{tag => "MQTT"})).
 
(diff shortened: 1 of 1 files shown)


ieQu1:
This may or may not corrupt the state of durable session during takeover. I'd like to test this locally.

qzhuyan:
> This may or may not corrupt the state of durable session during takeover. I'd like to test this locally.

Thank you for the feedback, that is what I am worried about.  the tricky thing is it is not easy to test. 
I have another safer approach, more complicated,  I could move on to that.

PR #14617, opened 2025-01-27
test: fix flaky test

thalesmg:


apps/emqx/test/emqx_persistent_session_SUITE.erl
```diff
@@ -774,17 +774,20 @@ t_publish_many_while_client_is_gone_qos1(Config) ->
 
     ?assert(NMsgs2 < NPubs, {NMsgs2, '<', NPubs}),
     ?assert(NMsgs2 > NPubs2, {NMsgs2, '>', NPubs2}),
-    ?assert(NMsgs2 >= NPubs - NAcked, Msgs2),
-    NSame = NMsgs2 - NPubs2,
+    %% Once reconnected, we should receive all the published messages above (`Msgs1 ++
+    %% Msgs2'), *except* those from `Msgs1' that we acked.
```
thalesmg, line 778: That is: we must not get the same messages which we had already acked, but we may get fewer than all the published and not-yet-acked messages.

apps/emqx/test/emqx_persistent_session_SUITE.erl
```diff
@@ -774,17 +774,20 @@ t_publish_many_while_client_is_gone_qos1(Config) ->
...
+    ?assert(NMsgs2 =< NPubs - NAcked, #{msgs2 => Msgs2, n_pubs => NPubs, n_acked => NAcked}),
```
thalesmg, line 779: That is: we must not get the same messages which we had already acked, but we may get fewer than all the published and not-yet-acked messages.

keynslug: It's a bit unclear why we may receive _less_ than _NPubs_ – _NAcked_ here. Waited not enough to receive this much? I wonder if this changes if the current timeout of 2 seconds is increased.

On a second thought, probably need to experiment with that.

thalesmg: Maybe.  Still, isn't the condition incorrect?  Shouldn't the number of received messages be _less or equal_ to the _maximum_ expected amount, instead of _greater or equal_ to it?

keynslug: Strictly speaking, it should be _equal_ but not less: we should expect to receive every yet unacked message, eventually. I think it was _greater_ because of how earlier implementations diverged from the specification, and might have resent already acked messages (marked as DUPes).

thalesmg: It should be _eventually equal_, indeed, but I'd still expect _less or equal_ for any amount of time, in particular if the receive interval happens to be too short.  :thinking:

keynslug: This is true, theoretically speaking. Practically speaking however, if we don't get all the messages in, let's say, 10 seconds, there's very likely a bug, which would have been obscured by the "less than or equal" assertion otherwise.

thalesmg: I've changed the assertion to `==` and increased the receive timeout, so we don't obscure potential bugs, by having either fewer or more messages than expected.

keynslug (approved):
❤️

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6735)

PR #14627, opened 2025-01-30
feat(rule engine): add `$events/alarm_{,de}activated`

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13636

Release version: v/e5.8.5

## Summary

With this new event, it's possible for rules to be triggered by alarm toggling without the
need to set `rule_engine.ignore_sys_message = false` and using the existing
`$SYS/broker/+/alarms/+` topic filter.  Using this new event, the frontend may implement
an UX for tying webhooks to alarms.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_hookpoints.erl
+++ b/apps/emqx/src/emqx_hookpoints.erl
@@ -79,6 +79,10 @@
     'session.takeovered'
 ]).
 
+-type rule_engine_conf() :: #{event_topic := emqx_types:topic()}.
+-type transformation_context() :: #{name := binary()}.
+-type validation_context() :: #{name := binary()}.
+
 %%-----------------------------------------------------------------------------
 %% Callbacks
 %%-----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


apps/emqx/src/emqx_hookpoints.erl
```diff
@@ -89,6 +108,12 @@
 %%
 %% By default, callbacks are executed in the channel process context.
 
+-callback 'alarm.activated'(alarm_activated_context(), rule_engine_conf()) ->
```
savonarola, line 111: That's not quite clear. It seems that alarms and hooks are in the `emqx` app and out of the context of the rule engine 🤔

thalesmg: The rule engine just wants to receive alarm info, that's why it's a new hook.

savonarola: I mean, emqx app doesn't even know that any rule engine exist.

savonarola: The callback signatures just document the arguments in absence of the additional hook arguments, so we probably may just remove the `rule_engine_conf()` arg from the specs.

thalesmg: Ah, I see now what you mean.

I've removed mentions to that config, even though that particular callback impl injects it in the hook.

savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5895)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5895)

PR #14632, opened 2025-02-03
ci: fix schema registry image that is gone

thalesmg:
```
Error response from daemon: repository public.ecr.aws/ews-network/confluentinc/cp-schema-registry not found: name unknown: The repository with name 'confluentinc/cp-schema-registry' does not exist in the registry with id 'ews-network'
```


id (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-5896)

PR #14671, opened 2025-02-10
fix(mqtt action): make `tcp_closed` and `closed` recoverable errors

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13880

Release version: v/e5.8.5

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
+++ b/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
@@ -407,6 +407,10 @@ handle_send_result({ok, Reply}) ->
 handle_send_result({error, Reason}) ->
     {error, classify_error(Reason)}.
 
+classify_reply(Reply = #{reason_code := ?RC_PACKET_IDENTIFIER_IN_USE}) ->
+    %% If `emqtt' client restarted, it may re-use packet ids that the remote broker still
+    %% has memory of.  We should retry.
+    {recoverable_error, Reply};
 classify_reply(Reply = #{reason_code := _}) ->
     {unrecoverable_error, Reply}.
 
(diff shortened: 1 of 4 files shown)


apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -407,6 +407,10 @@ handle_send_result({ok, Reply}) ->
 handle_send_result({error, Reason}) ->
     {error, classify_error(Reason)}.
 
+classify_reply(Reply = #{reason_code := ?RC_PACKET_IDENTIFIER_IN_USE}) ->
+    %% If `emqtt' client restarted, it may re-use packet ids that the remote broker still
+    %% has memory of.  We should retry.
+    {recoverable_error, Reply};
```
keynslug, line 413: Hmm, this sounds like a protocol violation. Is it `emqtt` failing proper session recovery under `CleanStart = false`? Probably here, where it relies on `clean_start` flag instead of `session_present`:
```
connected(info, immediate_retry, State = #state{clean_start = false}) ->
```

Tangentially, why does publishing side need `CleanStart = true`? When buffering and retries are involved on connector side, shouldn't this be enough?

thalesmg: I think the problem is allowing `clean_start = false` when the emqtt process can be restarted at any point and lose all state such as packet ids.  In this case, I think we ought to retry so we don't silently drop messages in practice.  This came from discussions with @zmstone .

> Tangentially, why does publishing side need CleanStart = true? When buffering and retries are involved on connector side, shouldn't this be enough?

`clean_start = true` is the default, which works much better with the restarting and retrying nature of the buffering layer.  `clean_start = false` + retries is exactly what _could_ lead to a state where the connector is reusing packet ids expected by the remote broker.  If we always start afresh and retry, there's no issue.

keynslug: > `clean_start = false` + retries is exactly what could lead to a state where the connector is reusing packet ids expected by the remote broker.

Right, this matches my understanding, I wonder why it's even possible to configure that (on publishing side).

changes/ce/fix-14671.en.md
```diff
@@ -0,0 +1 @@
+Fixed an issue with MQTT Action where messages could fail to be sent and retried in the event of a rare race condition linked to MQTT Connector having its connection closed.
```
zmstone, line 1:
```suggestion
Fixed an issue with MQTT Action where messages could fail to be sent and retried in the event of a race condition linked to MQTT Connector having its connection closed.
```

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6140)

PR #14694, opened 2025-02-13
fix(actions/sources api): use correct `enable` field to denote action/source state

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13913

Release version: v/e5.8.5

## Summary

Previous value was being taken from `resource_data`, which actually corresponds to the connector state (enabled/disabled).


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6131)

PR #14697, opened 2025-02-13
fix(actions/sources): don't mix action/source dependencies when deleting

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13889

Release version: v/e5.8.5

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_v2.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_v2.erl
@@ -371,7 +371,7 @@ check_deps_and_remove(ConfRootKey, BridgeType, BridgeName, AlsoDeleteActions) ->
             false -> []
         end,
     case
-        emqx_bridge_lib:maybe_withdraw_rule_action(
+        maybe_withdraw_rule_action(
             ConfRootKey,
             BridgeType,
             BridgeName,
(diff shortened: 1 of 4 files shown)


apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -388,6 +388,40 @@ check_deps_and_remove(ConfRootKey, BridgeType, BridgeName, AlsoDeleteActions) ->
 %% Helpers for CRUD API
 %%--------------------------------------------------------------------
 
+maybe_withdraw_rule_action(ConfRootKey, BridgeType, BridgeName, DeleteDeps) ->
```
keynslug, line 391: Nit. Does "withdraw" simply means "remove" in this context? If so, maybe just use "remove" then?

keynslug (approved):

apps/emqx_bridge/src/emqx_bridge_v2.erl
thalesmg: I assume so.  I just "forked" this from the previous `emqx_bridge_lib:maybe_...`, we could rename both later.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6133)

PR #14700, opened 2025-02-14
fix(alarm): run hooks in a temporary process

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13917

Release version: v/e5.8.5

### DIFF
--- a/apps/emqx/src/emqx_alarm.erl
+++ b/apps/emqx/src/emqx_alarm.erl
@@ -60,7 +60,8 @@
 
 %% Internal exports
 -export([
-    do_start_worker/0
+    do_start_worker/0,
+    worker_loop/0
 ]).
 
 -record(activated_alarm, {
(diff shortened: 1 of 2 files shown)


qzhuyan:
will the activation and deactivation of alerts run out of order?

thalesmg:
> will the activation and deactivation of alerts run out of order?

great observation, I've switched to using a companion process

apps/emqx/src/emqx_alarm.erl
```diff
@@ -538,8 +539,15 @@ start_worker() ->
     proc_lib:start_link(?MODULE, do_start_worker, []).
 
 do_start_worker() ->
+    %% Drop check after OTP 26 is dropped.
+    case erlang:function_exported(proc_lib, set_label, 1) of
+        true ->
+            proc_lib:set_label({?MODULE, worker});
```
zmstone, line 545: use binary label.
<<"alarm_event_worker">>

thalesmg: :checkered_flag:

zmstone (approved):

PR #14716, opened 2025-02-17
fix(actions & sources): add and remove channels asynchronously when updating configs

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13893

Fixes https://emqx.atlassian.net/browse/EMQX-13896

Fixes https://emqx.atlassian.net/browse/EMQX-13807

Release version: v/e5.8.5

## PR Checklist


apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -1706,6 +1804,19 @@ handle_retry_add_channel(Data0, ChannelId) ->
             keep_state_and_data
     end.
 
+handle_retry_remove_channel(Data0, ChannelId) ->
+    ?tp(retry_remove_channel, #{channel_id => ChannelId}),
+    From = undefined,
+    maybe
+        {ok, _} ?= maps:find(ChannelId, Data0#data.added_channels),
```
savonarola, line 1811: Nit:
m.b. for more transparent consistency and introspection, we could keep all the pending events and timers in the `added_channels` map, together with channel status.

thalesmg:
After discussions, this'll be moved to 5.9.0.

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -1052,6 +1046,18 @@ id(BridgeType, BridgeName, ConnectorName) ->
...
 
+get_resource_ids(ConfRootKey, Type, Name) ->
+    try
+        ChannelResId = id_with_root_name(ConfRootKey, Type, Name),
+        ConnResId = extract_connector_id_from_bridge_v2_id(ChannelResId),
+        {ok, {ConnResId, ChannelResId}}
+    catch
+        throw:Reason ->
+            {error, Reason};
+        error:{error, Reason} ->
```
zmstone, line 1057: `extract_connector_id_from_bridge_v2_id` should `throw` ?

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -1052,6 +1046,18 @@ id(BridgeType, BridgeName, ConnectorName) ->
 source_id(BridgeType, BridgeName, ConnectorName) ->
     id_with_root_name(?ROOT_KEY_SOURCES, BridgeType, BridgeName, ConnectorName).
 
+get_resource_ids(ConfRootKey, Type, Name) ->
+    try
+        ChannelResId = id_with_root_name(ConfRootKey, Type, Name),
```
zmstone, line 1051: cannot seem to find a `throw` down this call

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -2010,6 +2004,8 @@ extract_connector_id_from_bridge_v2_id(Id) ->
     case binary:split(Id, <<":">>, [global]) of
         [<<"action">>, _Type, _Name, <<"connector">>, ConnectorType, ConnecorName] ->
             <<"connector:", ConnectorType/binary, ":", ConnecorName/binary>>;
+        [<<"source">>, _Type, _Name, <<"connector">>, ConnectorType, ConnecorName] ->
```
zmstone, line 2007: Q: why this is missing before?

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -38,7 +38,9 @@
     channel_health_check/2,
     add_channel/3,
```
zmstone, line 39: Q: are the sync APIs still needed ?

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -765,18 +813,36 @@ handle_event(enter, _OldState, ?state_stopped = State, Data0) ->
 %% The following events can be handled in any other state
 %%--------------------------
 handle_event(
-    {call, From}, #add_channel{channel_id = ChannelId, config = Config}, State, Data
+    {call, From}, #add_channel{channel_id = ChannelId, config = Config}, State, Data0
+) ->
+    {Actions, Data} = handle_not_connected_add_channel(From, ChannelId, Config, State, Data0),
```
zmstone, line 818: nit: inconsistent naming?
m.b. handle_add_channel_not_connected

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -2100,3 +2223,76 @@ abort_health_checks_for_channel(Data0, ChannelId) ->
...
+    Ops2 = lists:map(fun(#{op := Op, n := N}) -> {N, Op} end, maps:values(Ops1)),
+    Ops = lists:keysort(1, Ops2),
+    lists:map(fun({_N, Op}) -> Op end, Ops).
+
+operation_channel_id(#add_channel{channel_id = ChannelId, config = _Config}) ->
+    ChannelId;
+operation_channel_id(#remove_channel{channel_id = ChannelId}) ->
+    ChannelId.
+
+collect_channel_operations(0, Acc) ->
```
zmstone, line 2288: nit: collect and compress can be done in one go?
use `lists:keyfind` to find the channel ID to preserve the order.
so there is no need for indexing map to list, sorting etc.

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -2100,3 +2223,76 @@ abort_health_checks_for_channel(Data0, ChannelId) ->
...
+            (#remove_channel{channel_id = ChannelId}, {AccActions, AccData}) ->
+                {Actions, Data} = handle_remove_channel(From, ChannelId, AccData),
+                {Actions ++ AccActions, Data}
+        end,
+        {[], Data0},
+        Ops
+    ).
+
+collect_and_compress_channel_operations(Acc) ->
+    MaxRequests = 500,
```
zmstone, line 2261: It's unlikely to collect 500 messages in mailbox, but if it happens, isn't it better to compact them all as soon as possible?

apps/emqx_bridge/src/emqx_bridge_v2.erl
thalesmg: https://github.com/emqx/emqx/blob/23f18a3daccf9975664c4a105d56e0332361db79/apps/emqx_bridge/src/emqx_bridge_v2.erl#L1060-L1067

apps/emqx_bridge/src/emqx_bridge_v2.erl
thalesmg: https://github.com/emqx/emqx/blob/23f18a3daccf9975664c4a105d56e0332361db79/apps/emqx_bridge/src/emqx_bridge_v2.erl#L1060-L1067

apps/emqx_resource/src/emqx_resource_manager.erl
thalesmg: there are some tests calling it (most shouldn't :melting_face: ), but dry run helper also calls it, and it's appropriate for dry run, I think

apps/emqx_resource/src/emqx_resource_manager.erl
thalesmg: not sure I understand, but that requires traversing the accumulator multiple times searching for the channel id, right?

I think we won't have so many messages that this is a performance issue.

apps/emqx_resource/src/emqx_resource_manager.erl
thalesmg: yes, as mentioned in the previous comment, I don't think we'll have that many.  it's just an upper limit.

>  isn't it better to compact them all as soon as possible?

isn't that what this function does?  not sure I understand.

apps/emqx_resource/src/emqx_resource_manager.erl
thalesmg: it's possible to introspect the timers with `sys:get_status/1`, which might be better, since we also may have postponed events which will not show up with `sys:get_state/1`, if that's the intention

apps/emqx_resource/src/emqx_resource_manager.erl
thalesmg: :checkered_flag: (in the newer PR)

apps/emqx_bridge/test/emqx_bridge_testlib.erl
```diff
@@ -126,6 +126,11 @@ create_bridge(Config, Overrides) ->
     ct:pal("creating bridge with config: ~p", [BridgeConfig]),
     emqx_bridge:create(BridgeType, BridgeName, BridgeConfig).
 
+%% For running RPCs in peers.
+infer_dashboard_host_port() ->
```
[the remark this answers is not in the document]
thalesmg: :checkered_flag:  (in the newer PR)

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -2100,3 +2223,76 @@ abort_health_checks_for_channel(Data0, ChannelId) ->
...
+                    New = #{op => Op, n => N},
+                    {OpAcc#{ChannelId => New}, N + 1}
+            end
+        end,
+        {#{}, 0},
+        Ops0
+    ),
+    Ops2 = lists:map(fun(#{op := Op, n := N}) -> {N, Op} end, maps:values(Ops1)),
+    Ops = lists:keysort(1, Ops2),
+    lists:map(fun({_N, Op}) -> Op end, Ops).
```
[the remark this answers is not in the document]
thalesmg: :checkered_flag:  (in the newer PR)

apps/emqx_resource/src/emqx_resource_manager.erl
zmstone: I mean if there are more than 500, the remaining ones will be handled in the next chunk, possibly takes even more time.

thalesmg: I see.  I think it's important to have an upper limit like this, though, or else we might be stuck collecting requests forever if timing and frequency of requests is just right.

PR #14724, opened 2025-02-18
refactor: delete unused callback from authz source behaviour

zmstone:
Release: v/e5.8.5

In anther task, I noticed that LDAP authz schema had a `desc` field provided by a callback, this made i18n complicated.
Then I realize that this callback is never called anywhere else, so I decided to remove it.


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6195)

PR #14726, opened 2025-02-19
docs: return bytesize in max_heap validation error message

thalesmg:
Fixes (again) https://emqx.atlassian.net/browse/EMQX-13901

Release version: v/e5.8.5


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6198)

PR #14765, opened 2025-02-24
fix(sqlserver): try to infer connected instance name

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13954

Release version: e5.9.0

## Summary

Previously, we could not infer when the user furnished an explicit port for SQL Server, and always added the default port if not explicitly defined.

For Named Instances, we need to explicitly define a port to connect to when connecting with the ODBC driver.  And the driver happily connects to whatever instance is running on that port, completely ignoring the given Instance Name, if any.

Here, we demand that the port is explicitly defined when an instance name is given, and we also attempt to infer differences between desired and connected instance names during health checks.

## PR Checklist

- [x] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket  -- https://github.com/emqx/emqx-docs/pull/2841
- [na] ~~Schema changes are backward compatible~~  if anyone uses named instances without port, they'll fail the validation.

### DIFF
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -55,7 +55,7 @@
 -type cipher() :: map().
 -type port_number() :: 1..65535.
 -type server_parse_option() :: #{
-    default_port => port_number(),
+    default_port => integer(),
     no_port => boolean(),
     supported_schemes => [string()],
     default_scheme => string()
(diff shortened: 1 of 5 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6776)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6776)

PR #14773, opened 2025-02-25
Refactor limiters

savonarola:
Fixes [EMQX-13124](https://emqx.atlassian.net/browse/EMQX-13124)

For a brief description and organization of the code, see [apps/emqx/src/emqx_limiter/README.md](https://github.com/lafirest/emqx/blob/refactor/limiter/apps/emqx/src/emqx_limiter/README.md)

The major updates are:
* The limiters are now simple token buckets;
* The semantics of rate limiting options (`messages_rate= "1000/1m"`) is now straightforward, meaning the bucket will be refilled 1000 tokens each 1m;
* We implemented burst rate, that is, temporarily granting larger capacity. Internally, that is modeled just as an increased overall capacity, but we intentionally do not specify this. 
* We implemented a hot config upgrade for the rate limits.

Release version: v/e5.9.0

## PR Checklist

- [ ] Schema changes are backward compatible **NO**, see the changelog.


thalesmg (commented):
:hot_pepper: 

(still reviewing)

apps/emqx/src/emqx_channel.erl
```diff
@@ -629,6 +628,7 @@ process_publish(Packet = ?PUBLISH_PACKET(QoS, Topic, PacketId), Channel) ->
                 },
                 #{topic => Topic, tag => "AUTHZ"}
```
zmstone, line 629: the "AUTHZ" tag here is no longer accurate?

apps/emqx/src/emqx_limiter/src/emqx_limiter_client.erl
```diff
@@ -0,0 +1,90 @@
...
+            {Result, Limiter#{state := NewState}};
+        Result when is_boolean(Result) ->
+            {Result, Limiter}
+    catch
+        error:Reason ->
+            ?SLOG_THROTTLE(
+                error,
+                #{
+                    msg => failed_to_consume_from_limiter,
+                    reason => Reason,
```
zmstone, line 65: Q: how to recover from this error ?

changes/ce/feat-14773.en.md
```diff
@@ -0,0 +1,12 @@
+Improved rate-limiting functionality (`bytes_rate`, `messages_rate`, `max_conn_rate` configured for zones or listeners).
+* The rate-limiting algorithm is simplified to work more predictably. It does not try to impose backpressure on the client(s). Instead, it drops the messages. For QoS1/QoS2 messagest the appropriate error code is also returned.
```
thalesmg, line 2:
```suggestion
* The rate-limiting algorithm is simplified to work more predictably. It does not try to impose backpressure on the client(s). Instead, it drops the messages. For QoS1/QoS2 messages the appropriate reason code is also returned.
```

changes/ce/feat-14773.en.md
```diff
@@ -0,0 +1,12 @@
...
+* Rate limiting now allows reconfiguration at runtime.
+* Rate limiting configuration now allows to specify window size and limit per each window explicitly:
+`messages_rate = "300/5m"` means 300 messages per 5 minutes (with any possible fluctuation within this period),
+while `messages_rate = "10/10s"` means 10 messages per 10 seconds. Although the configured RPS is the same, the latter variant is much more strict about fluctuations in the actual rate.
+* Burst rate is provided via the `messages_burst`, `bytes_burst`, `max_conn_burst` zone and listener options. The burst value `messages_burst = 1000/h` means a client can send 1000 additional messages each hour without applying rate limiting.
+
+## Breaking changes
```
thalesmg, line 9: Maybe it's better to create a separate `breaking-14773.en.md` file so that it may go into the appropriate changelog section.

apps/emqx/test/emqx_limiter_SUITE.erl
```diff
@@ -0,0 +1,203 @@
...
+run_connector() ->
+    {ok, C} = emqtt:start_link([{host, "127.0.0.1"}, {port, 1883}]),
+    case emqtt:connect(C) of
+        {ok, _} ->
+            {ok, _} = emqtt:publish(C, <<"test">>, <<"a">>, 1),
+            ok = emqtt:stop(C);
+        {error, _Reason} ->
+            ok
+    end,
+    ct:sleep(10),
```
thalesmg, line 190: Nit:

```suggestion
            {ok, _} = emqtt:publish(C, <<"test">>, <<"a">>, 1);
        {error, _Reason} ->
            ok
    end,
    ok = emqtt:stop(C);
    ct:sleep(10),
```

apps/emqx/test/emqx_channel_SUITE.erl
```diff
@@ -1126,33 +1167,10 @@ mock_cm_open_session() ->
...
-bucket_cfg() ->
-    #{rate => 10, initial => 0, burst => 0}.
-
-add_bucket() ->
-    emqx_limiter_server:add_bucket(?MODULE, message_routing, bucket_cfg()).
-
-del_bucket() ->
-    emqx_limiter_server:del_bucket(?MODULE, message_routing).
+    emqx_limiter_container:create_by_names(
+        [bytes, messages], #{}, emqx_limiter:internal_allocator()
```
thalesmg, line 1172: Maybe this function is not used?  There doesn't seem to be a `emqx_limiter:internal_allocator/0` function.  :thinking:

apps/emqx/src/emqx_limiter/src/emqx_limiter_client.erl
savonarola: >how to recover from this error ?
Create limiters back, e.g. by recreating the listener,

Personally, I'd prefer to just propagate error here to avoid inconsistent state.

The workaround comes from https://github.com/emqx/emqx/blob/master/apps/emqx/test/emqx_listeners_SUITE.erl#L59
where there is a scenario where we may "fail" to stop a listener and stop it again.

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_bucket_registry.erl
```diff
@@ -0,0 +1,47 @@
...
+%%  API
+%%--------------------------------------------------------------------
+
+-spec find_bucket(limiter_id()) -> bucket_ref() | undefined.
+find_bucket(LimiterId) ->
+    persistent_term:get(LimiterId, undefined).
+
+-spec insert_bucket(limiter_id(), bucket_ref()) -> ok.
+insert_bucket(LimiterId, Bucket) ->
+    persistent_term:put(LimiterId, Bucket).
```
zmstone, line 42: If our ambition is to support 1M limiters (for multi-tenancy, one limiter for each tenant), would it make more sense to use ETS instead ?

apps/emqx_retainer/test/emqx_retainer_SUITE.erl
```diff
@@ -2,7 +2,7 @@
 %% Copyright (c) 2020-2025 EMQ Technologies Co., Ltd. All Rights Reserved.
 %%
 %% Licensed under the Apache License, Version 2.0 (the "License");
-%% you may not use this file except in compliance with the License.
+%% you may not use this ifle except in compliance with the License.
```
thalesmg, line 5:
```suggestion
%% you may not use this file except in compliance with the License.
```

thalesmg (commented):
:1st_place_medal:

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_bucket_registry.erl
savonarola: Currently, there is an indirection: the client contains `limiter_id`; by `limiter_id`, it gets bucket ref from PT, which contains counter ref and index. This is done to cope with probable allocator crashes. On restart, it will recreate and reregister the bucket, and the client will just use the new bucket on the next call.

We may instead:
* In the limiter client, cache bucket ref directly (counter & index);
* In the allocator, at the start, try to find and reuse the old bucket.

However, with 1M limiters, we may encounter the following problems:

**Settings update**. Currently, limiter settings are also cached in the PT. If a limiter becomes disabled
* the PT value is updated
* and the connected limiter clients (channels) see that immediately and do not try to consume the bucket.

With 1M limiters, maintaining 1M PTs with settings also would be very costly; probably we would want to use a global setting instead: limiting enabled for all tenants/disabled for all tenants.

**Bucket refill**. 
With 1M tenants we somehow would have to refill 1M buckets somehow.  It would be unpractical to do that in the current way (counters and timers). Instead, we may try to make something like self-refilling limiter, like the `exclusive` one, but based on atomics and self-refilling with compare-and-set by one of the clients.

apps/emqx/src/emqx_limiter/src/emqx_limiter.erl
```diff
@@ -0,0 +1,378 @@
...
+            ok = Module:delete_group(Group),
+            ok = emqx_limiter_registry:unregister_group(Group)
+    end.
+
+%%--------------------------------------------------------------------
+%% Zone config update
+%%--------------------------------------------------------------------
+
+add_handler() ->
+    ok = emqx_config_handler:add_handler([mqtt, limiter], ?MODULE),
```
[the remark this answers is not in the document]
savonarola: There is no legal way to configure zones in runtime except the default one 😃

apps/emqx/src/emqx_limiter/src/emqx_limiter_exclusive.erl
```diff
@@ -0,0 +1,165 @@
...
+->
+    {false, State0};
+try_consume_regular(
+    #{last_time := LastTime, tokens := Tokens0} = State0,
+    #{capacity := Capacity, burst_capacity := BurstCapacity, interval := Interval},
+    Amount,
+    Now
+) ->
+    Inc = Capacity * (Now - LastTime) / Interval,
+    Tokens = erlang:min(Capacity + BurstCapacity, Tokens0 + Inc),
```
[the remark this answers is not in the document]
savonarola: Then we would wipe the burst tokens, if there are any.

For simplicity, we just silently allow to reach the burst capacity if the client consumes slowly than the regular rate. 
This allows to provide our commitment:
* a client may consume with regular rate;
* each burst interval it may consume burst tokens. 

We do not try to punish a user for understanding that they basically have a little larger capacity and rate and may just use burst to consume steadily with slightly increased rate.

After feedback from the cloud team, we may tune this.

apps/emqx/test/emqx_limiter_SUITE.erl
savonarola: If connect fails, doesn't then `C` exit itsef?

apps/emqx_gateway/src/bhvrs/emqx_gateway_conn.erl
```diff
@@ -309,7 +309,6 @@ init_state(WrappedSock, Peername, Options, FrameMod, ChannMod) ->
     },
     ActiveN = emqx_gateway_utils:active_n(Options),
     %% FIXME: TODO
```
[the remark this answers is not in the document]
savonarola: Still to be implemented, I think 😅

apps/emqx/src/emqx_limiter/src/emqx_limiter_client.erl
```diff
@@ -0,0 +1,90 @@
...
+    catch
+        error:Reason ->
+            ?SLOG_THROTTLE(
+                error,
+                #{
+                    msg => failed_to_put_back_to_limiter,
+                    reason => Reason,
+                    module => Module
+                },
+                #{tag => "AUTHZ"}
```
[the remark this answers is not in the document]
savonarola: ✅

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_bucket_registry.erl
```diff
@@ -0,0 +1,47 @@
...
+
+-spec delete_bucket(limiter_id()) -> ok.
+delete_bucket(LimiterId) ->
+    _ = persistent_term:erase(LimiterId),
```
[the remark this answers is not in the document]
savonarola: 👍 ✅

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_allocator.erl
```diff
@@ -0,0 +1,337 @@
...
+update(Group) ->
+    gen_server:call(?VIA_GPROC(Group), #update{}).
+
+%%--------------------------------------------------------------------
+%%% gen_server callbacks
+%%--------------------------------------------------------------------
+
+-spec init(list()) -> {ok, state()}.
+init([Group, LimiterNames]) ->
+    State0 = #{
```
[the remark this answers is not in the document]
savonarola: ✅

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_allocator.erl
```diff
@@ -0,0 +1,337 @@
...
+        #{burst_capacity := 0} ->
+            do_ensure_burst_timers(State, Names);
+        #{burst_capacity := BurstCapacity, burst_interval := Interval} when BurstCapacity > 0 ->
+            BurstTimers =
+                case BurstTimers0 of
+                    #{Interval := IntervalNames0} ->
+                        IntervalNames = [Name | IntervalNames0],
+                        BurstTimers0#{Interval => IntervalNames};
+                    _ ->
+                        _ = erlang:send_after(Interval, self(), {tick_burst_alloc_event, Interval}),
```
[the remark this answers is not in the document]
savonarola: ✅

apps/emqx/src/emqx_channel.erl
savonarola: ✅

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_shared.erl
```diff
@@ -0,0 +1,161 @@
...
+-spec child_spec(emqx_limiter:group(), [emqx_limiter:name()]) ->
+    supervisor:child_spec().
+child_spec(Group, Names) ->
+    #{
+        id => Group,
+        start => {emqx_limiter_allocator, start_link, [Group, Names]},
+        restart => permanent,
+        shutdown => 5000,
+        type => worker,
+        modules => [?MODULE]
```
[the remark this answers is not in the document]
savonarola: ✅

apps/emqx/src/emqx_limiter/src/emqx_limiter_schema.erl
```diff
@@ -16,277 +16,126 @@
...
-                ?HOCON(?R_REF(node_opts), #{
-                    desc => deprecated_desc(Type),
+            {alloc_interval,
+                ?HOCON(emqx_schema:duration_ms(), #{
+                    desc => ?DESC(alloc_interval),
+                    default => <<"100ms">>,
                     importance => ?IMPORTANCE_HIDDEN,
-                    required => {false, recursively},
-                    aliases => alias_of_type(Type)
+                    deprecated => {since, "5.9.0"}
```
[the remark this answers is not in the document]
savonarola: ✅

apps/emqx/src/emqx_limiter/src/emqx_limiter_schema.erl
```diff
@@ -16,277 +16,126 @@
...
-     || {Name, Example} <-
-            lists:zip(short_paths(), [<<"1000/s">>, <<"1000/s">>, <<"100MB/s">>])
+        {Rate,
+            ?HOCON(rate_type(), #{
+                desc => ?DESC(Rate),
+                required => false
+            })},
+        {Burst,
+            ?HOCON(rate_type(), #{
+                desc => ?DESC(burst),
```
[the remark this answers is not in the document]
savonarola: ✅

changes/ce/feat-14773.en.md
savonarola: ✅

apps/emqx/src/emqx_limiter/src/emqx_limiter_client.erl
savonarola: So, the recovery is to stop a corresponding listener and start it back.

zmstone: Maybe add the possible cause and recover to the log.
e.g. `cause => "limiter not initialized", hint => "Try restart the listener"`

also, should log the limiter identifier too ?

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_bucket_registry.erl
zmstone: > compare-and-set by one of the clients.

You mean one of the connected clients has to take a exclusive lock as an owner of the bucket?

zmstone: Q: If say, the total number of tenants is less than 1000, should the current implementation work?

apps/emqx/src/emqx_limiter/src/emqx_limiter.erl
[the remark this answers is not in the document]
thalesmg: Not sure I understand.  

I remember I saw a TODO in one of the modules that suggested that we might need a handler such as this to trigger updates when updating a zone config (trying to find again which module it was :see_no_evil: ).

apps/emqx/test/emqx_limiter_SUITE.erl
thalesmg: true

apps/emqx/src/emqx_limiter/src/emqx_limiter_registry.erl
```diff
@@ -0,0 +1,208 @@
...
+handle_cast(Req, State) ->
+    ?SLOG(error, #{msg => "unexpected_cast", cast => Req}),
+    {noreply, State}.
+
+handle_info(Req, State) ->
+    ?SLOG(error, #{msg => "unexpected_info", info => Req}),
+    {noreply, State}.
+
+terminate(_Reason, #{group_names := GroupNames} = _State) ->
+    ?SLOG(warning, #{msg => "terminating", group_names => sets:to_list(GroupNames)}),
```
thalesmg, line 171: Maybe make this a debug level log, and also add a bit more information to `msg` to distinguish from whence it comes from?

thalesmg (approved):
:shipit:

apps/emqx/src/emqx_limiter/src/emqx_limiter.erl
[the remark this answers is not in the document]
savonarola: I added config listeners e.g., here https://github.com/emqx/emqx/blob/c8bfa6e6b02ba1b55ca93c40fd6fa62f88faed11/apps/emqx_retainer/src/emqx_retainer.erl#L298

But this was actually an overkill because there are no API/CLI endpoints to:
* add a new zone;
* change settings of a zone other than `default`.

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_allocator.erl
```diff
@@ -0,0 +1,346 @@
...
+                        Timers0#{Interval => [Name]}
+                end,
+            do_ensure_alloc_timers(State#{timers => Timers}, Names)
+    end.
+
+handle_alloc_timer(#{timers := Timers0} = State0, Interval) ->
+    {Names, Timers} = maps:take(Interval, Timers0),
+    State1 = State0#{timers => Timers},
+    State2 = alloc_tokens(regular, State1, Names),
+    ensure_alloc_timers(State2, Names).
```
keynslug, line 302: Nit. Shouldn't this be other way around, ideally? Imagining there could be a slight drift if there is a significant number of buckets to "alloc".

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_allocator.erl
```diff
@@ -0,0 +1,346 @@
...
+    BucketRef = #{counter => Counter, index => Index},
+    ok = emqx_limiter_bucket_registry:insert_bucket(LimiterId, BucketRef),
+    #{
+        counter => Counter,
+        index => Index,
+        correction => 0,
+        last_alloc_time => now_ms_monotonic()
+    }.
+
+alloc_tokens(
```
keynslug, line 164: Just a thought. I wonder how would an implementation based on pair of atomics per bucket (i.e. one for tokens, another for "refill epoch", for "self-refilling") compare against this one with timers, e.g.: scalability under high contention and/or massive number of buckets (my guess is the current one should be better up until some large number).

keynslug (approved):
🔥

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_bucket_registry.erl
savonarola: >You mean one of the connected clients has to take a exclusive lock as an owner of the bucket?
I don't think we need a lock, we may have something like this:
```erlang
try_consume(#{
   last_update  := LastUpdate0,
   last_update_atomic_ref := LastUpdateARef,
   last_update_atomic_ix := LastUpdateAIx,
   tokens_counter_ref := TokensCRef,
   tokens_counter_ix := TokensCIx,
   ...
}, Amount) ->
#{capacity := Cap, interval := Interval} = get_settings(...),
...
%% Try to consume existing tokens
%% If there are no tokens, then:
...
Now = now_ns_monotonic(),
case LastUpdate0 + Interval > Now of
  true -> ... %% failure
  false ->
     LastUpdate1 = atomics:get(LastUpdateARef, LastUpdateAIx),
     case LastUpdate1 + Interval > Now of
        true ->
           %% failure, someone recently refilled the bucket, only update last_update
        false ->
           case compare_exchange(LastUpdateARef, LastUpdateAIx, LastUpdate1, Now) of
              ok ->
                  %% we updated the LastUpdate atomic, so we refill the bucket
                  counter:add(...),
                  %% cache last_update, try use refilled tokens
              LastUpdate2 ->
                  %% conflict, someone else updated, cache last_update, try use refilled tokens
           end
     end
end

           
      
     




```

savonarola: > If say, the total number of tenants is less than 1000, should the current implementation work?

That depends on the configured rates, but I wouldn't try to use the alloc limiters for 1000 tenants either.

apps/emqx/src/emqx_limiter/src/emqx_limiter.erl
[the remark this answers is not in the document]
thalesmg: I see now, thanks.  Maybe it would be "future-proof" to add the handlers now (so we don't get weird behavior when such APIs are introduced), but we may as well defer to when we actually have the problem.

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_allocator.erl
savonarola: ✅ 
As far as I can see, the existing algorithm tries to take drift into account.

zmstone (dismissed):
Great work!
Maybe a follow-up PR to implement the self-filling algorithm ?

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6790)

PR #14774, opened 2025-02-25
Fix plugin cluster sync

JimMoen:
Fixes [EMQX-10080](https://emqx.atlassian.net/browse/EMQX-10080)

Release version: v/e5.9.0

## PR Checklist

- ~[ ] Added tests for the changes~ 
  - another pr for cluster plugin sync tests
- ~[ ] Added property-based tests for code which performs user input validation~
- ~[ ] Changed lines covered in coverage report~
- ~[ ] Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files~
- ~[ ] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket~

- ~[ ] If changed package build workflow, pass [this action]~(https://github.com/emqx/emqx/actions/workflows/build_packages.yaml) (manual trigger)

[EMQX-10080]: https://emqx.atlassian.net/browse/EMQX-10080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
@@ -37,7 +37,8 @@
     update_plugin/2,
     plugin_config/2,
     plugin_schema/2,
-    update_boot_order/2
+    update_boot_order/2,
+    sync_plugin/2
 ]).
 
 -export([
@@ -52,7 +53,9 @@
     ensure_action/2,
     ensure_action/3,
     do_update_plugin_config/3,
-    do_update_plugin_config_v4/2
+    do_update_plugin_config_v4/2,
+    ensure_existed/1,
+    sync_plugin_cluster/2
 ]).
 
 -define(NAME_RE, "^[A-Za-z]+\\w*\\-[\\w-.]*$").
@@ -75,7 +78,8 @@ paths() ->
         "/plugins/:name/:action",
         "/plugins/:name/config",
         "/plugins/:name/schema",
-        "/plugins/:name/move"
+        "/plugins/:name/move",
+        "/plugins/cluster_sync"
     ].
 
 schema("/plugins") ->
@@ -251,6 +255,26 @@ schema("/plugins/:name/move") ->
                 400 => emqx_dashboard_swagger:error_codes(['MOVE_FAILED'], <<"Move failed">>)
             }
         }
+    };
+schema("/plugins/cluster_sync") ->
+    #{
+        'operationId' => sync_plugin,
+        post => #{
+            summary => <<"Sync specific version of plugin to cluster">>,
+            description =>
+                "Forces a plugin to be synced to the cluster and removes other versions of the plugin.",
+            tags => ?TAGS,
+            'requestBody' => sync_request_body(),
+            responses => #{
+                204 => <<"Sync plugin successfully">>,
+                400 => emqx_dashboard_swagger:error_codes(
+                    ['BAD_PLUGIN_INFO'], <<"Bad Plugin Name Vsn">>
+                ),
+                404 => emqx_dashboard_swagger:error_codes(
+                    ['NOT_FOUND'], <<"Plugin Not Found">>
+                )
+            }
+        }
     }.
 
 fields(plugin) ->
@@ -363,6 +387,14 @@ fields(position) ->
                 }
             )}
     ];
+fields(sync_request) ->
+    [
+        {name,
+            hoconsc:mk(string(), #{
+                example => "emqx_plugin_demo-5.1-rc.2",
+                required => true
+            })}
+    ];
 fields(running_status) ->
     [
         {node, hoconsc:mk(string(), #{example => "emqx@127.0.0.1"})},
@@ -394,6 +426,17 @@ move_request_body() ->
             move_to_after => #{
                 summary => <<"move plugin after other plugins">>,
                 value => #{position => <<"after:emqx_plugin_demo-5.1-rc.2">>}
+            }
+        }
+    ).
+
+sync_request_body() ->
+    emqx_dashboard_swagger:schema_with_examples(
+        hoconsc:ref(?MODULE, sync_request),
+        #{
+            sync_from_node => #{
+                summary => <<"sync specific version of plugin to cluster">>,
+                value => #{name => <<"emqx_plugin_demo-5.1-rc.2">>}
             }
         }
     ).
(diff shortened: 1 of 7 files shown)


JimMoen:
from @zmstone
> Q1: What happens if nodes have different plugins installed?
Q2: And what happens if nodes have conflicts? e.g. same plugin but different version.
Q3: Who is to decide which one should be copied from and to ?

---

**A1:**
It happens before a node (we call it `NewJoined` here) joined a cluster (we call it `Cluster` here).
After `NewJoined` have joined the `Cluster`. 
On the Plugin page of the Dashboard, you can see the plugins installed on all nodes (note that the plugins are not synchronized between nodes at this time). When you try to enable a plugin, the plugin and its configuration (if it have one) will be synchronized between nodes, and then an attempt will be made to enable the plugin.

**A2:**
The situation is almost the same as A1. However, in order to keep the actual applied plugins the same between nodes, an API is provided in this PR to force synchronization between nodes.
Cluster Plugin sync cURL request:

<details> 

```curl
curl -X 'POST' \
  'http://<host>:<port>/api/v5/plugins/cluster_sync' \
  -H 'accept: */*' \
  -H 'Authorization: Basic xxx' \
  -d '{
  "name": "simple_emqx_plugin-1.0.0", // Which plugin NameVsn will be synced to cluster
  "node": "emqx1@emqx1-cluster.emqx.io" // And which node the plugin from.
}'
```

</details>

**A3:**
We leave this decision to the user.

---

Known issue: Since the Name in the plugin's `NameVsn` and the erlang app in the release have the same name, and the upgrade of the plugin package may not be accompanied by the upgrade of the erlang app, we are temporarily unable to distinguish which plugin package the plugin beam loaded on different nodes comes from.
But what is certain is that after calling the plugin synchronization API, restarting the plugin can ensure consistency between nodes.

zmstone (dismissed):

apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -245,6 +249,25 @@ schema("/plugins/:name/move") ->
                 400 => emqx_dashboard_swagger:error_codes(['MOVE_FAILED'], <<"Move failed">>)
             }
         }
+    };
+schema("/plugins/cluster_sync") ->
+    #{
+        'operationId' => sync_plugin,
+        post => #{
+            summary => <<"Sync specific version of plugin from given node">>,
```
savonarola, line 257: Maybe "from current node"? Seems there is no node-like parameter specified.

apps/emqx_management/src/proto/emqx_mgmt_api_plugins_proto_v4.erl
```diff
@@ -0,0 +1,83 @@
...
+    rpc:multicall(
+        Nodes,
+        emqx_mgmt_api_plugins,
+        do_update_plugin_config,
+        [NameVsn, AvroJsonMap, MaybeAvroValue],
+        10000
+    ).
+
+-spec ensure_existed(node(), binary() | string()) -> emqx_rpc:call_result().
+ensure_existed(Node, NameVsn) ->
```
savonarola, line 75: Is this function used anywhere?

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -244,9 +248,26 @@ ensure_disabled(NameVsn) ->
 %% reside in all the plugin install dirs.
 -spec purge(name_vsn()) -> ok.
 purge(NameVsn) ->
+    ?SLOG(debug, #{msg => "purge_plugin", name_vsn => NameVsn}),
     _ = maybe_purge_plugin_config(NameVsn),
     purge_plugin(NameVsn).
 
+-spec maybe_purge_others(name_vsn()) -> ok.
+maybe_purge_others(NameVsn) ->
```
savonarola, line 256: Nit: m.b. purge other versions?

savonarola:
Maybe a test needed?

apps/emqx_management/src/emqx_mgmt_api_plugins.erl
JimMoen: nice catch. it's my bad.
The previous design required a specified node name, but this parameter was later removed.

apps/emqx_management/src/proto/emqx_mgmt_api_plugins_proto_v4.erl
JimMoen: Added it for the same reason, but forgot to delete it.

apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -363,6 +387,14 @@ fields(position) ->
                 }
             )}
     ];
+fields(sync_request) ->
+    [
+        {name,
```
savonarola, line 392: nit: we may reference the existing name field here, or use its validator.

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -42,6 +42,10 @@
     ensure_installed/0,
     ensure_installed/1,
     ensure_installed/2,
+
+    get_plugin_tar_from_node/2,
```
savonarola, line 46:
```suggestion
    get_package_from_node/2,
```

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -42,6 +42,10 @@
...
+    get_plugin_config_from_node/2,
```
savonarola, line 47: Nit 
```suggestion
    get_config_from_node/2,
```

apps/emqx_management/src/proto/emqx_mgmt_api_plugins_proto_v4.erl
```diff
@@ -1,7 +1,17 @@
+%% Copyright (c) 2022-2025 EMQ Technologies Co., Ltd. All Rights Reserved.
```
savonarola, line 1: AFAIK, we do not add Apache License header anymore because of license change.

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -253,6 +259,25 @@ write_package(NameVsn, Bin) ->
...
+purge_other_versions(NameVsn) ->
+    {AppName, _AppVsn} = emqx_plugins_utils:parse_name_vsn(NameVsn),
+    lists:foreach(
+        fun
+            (Dir) when Dir =:= NameVsn ->
+                ok;
+            (Path) ->
+                case emqx_plugins_fs:is_installed(Path) of
+                    true ->
+                        ok = delete_cached_config(Path),
```
savonarola, line 272: `delete_cached_config` receives `NameVsn`, not path

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -253,6 +259,25 @@ write_package(NameVsn, Bin) ->
 is_package_present(NameVsn) ->
     emqx_plugins_fs:is_tar_present(NameVsn).
 
+-spec purge_other_versions(name_vsn()) -> ok.
+purge_other_versions(NameVsn) ->
```
savonarola, line 263: Maybe we should 
* call `list_plugins` here
* find plugin(s) with the same AppName
* call ensure_stopped, ensure_uninstalled

because the plugin that is being removed may actually be running

apps/emqx_plugins/test/emqx_plugins_SUITE.erl
```diff
@@ -882,6 +882,173 @@ group_t_cluster_leave(Config) ->
...
+            vsn => "5.1.0",
+            tag => "5.1.0",
+            shdir => OldInstallDir
+        }),
+
+    #{package := NewPackage, release_name := NewPluginName} =
+        get_demo_plugin_package(#{
+            release_name => ?EMQX_PLUGIN_TEMPLATE_RELEASE_NAME,
+            git_url => ?EMQX_PLUGIN_TEMPLATE_URL,
+            vsn => "5.9.0-beta.1",
```
savonarola, line 903:
```suggestion
            vsn => "5.9.0-beta.3",
```

JimMoen: it's ok to keep beta.1 here.
cluster sync and purge should work with any two different versions.

savonarola (approved):

PR #14799, opened 2025-02-28
feat(sessmem): run takeover end phase after returning connack

keynslug:
Fixes [EMQX-13354](https://emqx.atlassian.net/browse/EMQX-13354).

Release version: e5.9

## Summary

See individual commits.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -332,7 +332,7 @@ takeover_session_begin(_ClientId, undefined) ->
 -spec takeover_session_end(takeover_state()) ->
     {ok, _ReplayContext} | {error, _Reason}.
 takeover_session_end({ConnMod, ChanPid}) ->
-    case wrap_rpc(emqx_cm_proto_v2:takeover_finish(ConnMod, ChanPid)) of
+    case emqx_cm_proto_v3:takeover_finish(ConnMod, ChanPid) of
         {ok, Pendings} ->
             {ok, Pendings};
         {error, _} = Error ->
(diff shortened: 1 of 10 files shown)


apps/emqx/src/emqx_cm.erl
```diff
@@ -358,7 +352,7 @@ pick_channel(ClientId) ->
             ?SLOG(warning, #{msg => "more_than_one_channel_found", chan_pids => ChanPids}),
             lists:foreach(
                 fun(StalePid) ->
-                    catch discard_session(ClientId, StalePid)
+                    discard_session(ClientId, StalePid)
```
qzhuyan, line 355: Just some findings, while I am investigating some issue in the same area, I found one issue:

The computation of StalePids seems wrong in cluster setup or even on single node as well, we could not assume 'later' channel is the latest one as the inserts are now all dirty, meaning we might discard the wrong channel and take over the wrong session.

keynslug: Curious how you define _wrong_ here? Having more than one channel for a given ClientID is an anomaly, which in my opinion suggests that there's no good definition of _wrong_ or _right_ channel: e.g. imagine there was a brain-split, and 2 lingering sessions now hold disjoint sets of subscriptions / mqueues / etc.

So we either have to bring strict serializability to the table, or tolerate multiple entries and rely on some heuristic. Do you suggest using "pick the most recent one" heuristic, regardless of the potential clock skew?

zmstone (commented):
there is a session-present flag in CONNACK, https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901078
if takeover failed, the client may lose all its subscriptions ?

zmstone:
Maybe a less radical change is to fail take-over more aggressively.

- Change wait time from 15s to 5s
- Force kill old sessions once there are more than 1 old session found (typically means the client is stuck on doing something after connected, authentication, authorization, data integration)

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6822)

PR #14802, opened 2025-02-28
chore(plugins): require explicit CLI authorization to install packages

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13979

Release version: v/e5.8.6

## Summary

After discussions, the solution to the potential issue was to require explicit authorization to install a plugin (at specific version) via the CLI before HTTP API accepts the package.

## PR Checklist

- [x] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket  --  https://emqx.atlassian.net/browse/EMQX-13988

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
@@ -458,7 +458,9 @@ do_install_package(NameVsn, FileName, Bin) ->
             Msg = iolist_to_binary([
                 <<"Package is not allowed installation;">>,
                 <<" first allow it to be installed by running:">>,
-                <<" `emqx ctl plugins allow NAME-VSN`">>
+                <<" `emqx ctl plugins allow ">>,
+                NameVsn,
+                <<"`">>
             ]),
             {403, #{code => 'FORBIDDEN', message => Msg}}
     end.
--- a/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
@@ -120,7 +120,7 @@ schema("/plugins/install") ->
             responses => #{
                 204 => <<"Install plugin successfully">>,
                 400 => emqx_dashboard_swagger:error_codes(
-                    ['UNEXPECTED_ERROR', 'ALREADY_INSTALLED', 'BAD_PLUGIN_INFO']
+                    ['UNEXPECTED_ERROR', 'ALREADY_INSTALLED', 'BAD_PLUGIN_INFO', 'FORBIDDEN']
                 )
             }
         }
@@ -418,7 +418,7 @@ upload_install(post, #{body := #{<<"plugin">> := Plugin}}) when is_map(Plugin) -
                     AppDir = filename:join(emqx_plugins:install_dir(), AppName),
                     case filelib:wildcard(AppDir ++ ?VSN_WILDCARD) of
                         [] ->
-                            do_install_package(FileName, Bin);
+                            do_install_package(NameVsn, FileName, Bin);
                         OtherVsn ->
                             {400, #{
                                 code => 'ALREADY_INSTALLED',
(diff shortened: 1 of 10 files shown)


apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -154,6 +158,28 @@ app_dir(AppName, Apps) ->
             {error, not_found}
     end.
 
+%% Note: this is only used for the HTTP API.
+%% We could use `application:set_env', but the typespec for it makes dialyzer sad when it
+%% seems a non-atom key...
+-spec allow_installation(binary() | string()) -> ok.
+allow_installation(NameVsn0) ->
+    NameVsn = bin(NameVsn0),
+    _ = persistent_term:put({?MODULE, NameVsn}, true),
```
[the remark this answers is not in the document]
thalesmg: As I tried to convey in the comment, dialyzer complains about using non-atom keys in app env parameter keys.

> {allowed_installations, #{Name => true}}.

I don't quite understand what's this, but if you mean to use `application:set_env(emqx_plugins, allow_installations, #{...})`, then that could introduce some race conditions with reading and writing back the value, which the persistent term approach doesn't have.  :thinking:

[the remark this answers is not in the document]
zmstone: This is only triggered from manual step, race is less of a concern. Worst case is to run the command again to allow if it accidentally get lost by setting allow for another plugin name-vsn at the same time.

My point is to avoid using pt for non-static, non-performance critical code paths.

changes/ce/fix-14802.en.md
```diff
@@ -0,0 +1,3 @@
+Added new `emqx ctl plugins allow NAME-VSN` CLI command for plugins.
+
+Now, it is required to run thist command in the CLI before attempting to install a package via the HTTP API, to harden security.
```
zmstone, line 3:
```suggestion
Now, it is required to run this command in the CLI before attempting to install a package via the HTTP API, to harden security.
```

apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -450,6 +450,21 @@ upload_install(post, #{}) ->
...
 
+do_install_package(NameVsn, FileName, Bin) ->
+    case emqx_plugins:is_allowed_installation(NameVsn) of
+        true ->
+            do_install_package(FileName, Bin);
+        false ->
+            Msg = iolist_to_binary([
+                <<"Package is not allowed installation;">>,
+                <<" first allow it to be installed by running:">>,
+                <<" `emqx ctl plugins allow ">>,
```
zmstone, line 461: I notified frontend team to add some special handing for 403 returned from this API.
They will maybe render this message in a pop-up, maybe ask  what's the best quote for this message. 
Unsure ` is the best choice

thalesmg: without any change, it already displays a readable error message:

image

apps/emqx_management/src/emqx_mgmt_cli.erl
```diff
@@ -329,6 +331,8 @@ plugins(_) ->
             {"plugins <command> [Name-Vsn]", "e.g. 'start emqx_plugin_template-5.0-rc.1'"},
             {"plugins list", "List all installed plugins"},
             {"plugins describe  Name-Vsn", "Describe an installed plugins"},
+            {"plugins allow     Name-Vsn",
```
zmstone, line 334: thinking about it again, should there perhaps be a `disallow` command ?

thalesmg: I imagined it should

thalesmg: :checkered_flag:

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6363)

PR #14840, opened 2025-03-07
feat: multi-tenancy limiters

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13820

Release version: e5.9.0

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -408,6 +408,13 @@ handle_in(
                     {?RC_CONTINUE_AUTHENTICATION, NProperties},
                     NChannel#channel{conn_state = reauthenticating}
                 );
+            {error, NReasonCode, NChannel} ->
+                case ConnState of
+                    connecting ->
+                        handle_out(connack, NReasonCode, NChannel);
+                    _ ->
+                        handle_out(disconnect, NReasonCode, NChannel)
+                end;
             {error, NReasonCode} ->
                 case ConnState of
                     connecting ->
@@ -536,10 +543,11 @@ process_connect(?CONNECT_PACKET(ConnPkt) = Packet, Channel) ->
                 fun run_conn_hooks/2,
                 fun check_connect/2,
                 fun enrich_client/2,
-                %% set_log_meta should happen after enrich_client
-                %% because client ID assign and override
+                %% `set_log_meta' should happen after `enrich_client'
+                %% because client ID assign and override.
+                %% Even though authentication may also inject new attributes, we call it
+                %% here so that next steps have more logger context, and call it again after auth.
                 fun set_log_meta/2,
-                fun adjust_limiter/2,
                 fun check_banned/2,
                 fun count_flapping_event/2
             ],
@@ -560,8 +568,8 @@ process_connect(?CONNECT_PACKET(ConnPkt) = Packet, Channel) ->
                     post_process_connect(Properties, NChannel3);
                 {continue, Properties, NChannel2} ->
                     handle_out(auth, {?RC_CONTINUE_AUTHENTICATION, Properties}, NChannel2);
-                {error, ReasonCode} ->
-                    handle_out(connack, ReasonCode, NChannel1)
+                {error, ReasonCode, NChannel2} ->
+                    handle_out(connack, ReasonCode, NChannel2)
             end;
         {error, ReasonCode, NChannel} ->
             handle_out(connack, ReasonCode, NChannel)
@@ -2056,6 +2064,11 @@ fix_mountpoint(ClientInfo = #{mountpoint := MountPoint}) ->
     MountPoint1 = emqx_mountpoint:replvar(MountPoint, ClientInfo),
     ClientInfo#{mountpoint := MountPoint1}.
 
+fix_mountpoint(_PipelineOutput, #channel{clientinfo = ClientInfo0} = Channel0) ->
+    ClientInfo = fix_mountpoint(ClientInfo0),
+    Channel = Channel0#channel{clientinfo = ClientInfo},
+    {ok, Channel}.
+
 %%--------------------------------------------------------------------
 %% Set log metadata
 
@@ -2157,7 +2170,7 @@ authenticate(Packet, Channel) ->
                 {ok, _, _} -> ?EXT_TRACE_SET_STATUS_OK();
                 %% TODO: Enhanced AUTH
                 {continue, _, _} -> ok;
-                {error, _} -> ?EXT_TRACE_SET_STATUS_ERROR()
+                {error, _, _} -> ?EXT_TRACE_SET_STATUS_ERROR()
             end,
             Res
         end,
@@ -2185,14 +2198,14 @@ process_authenticate(
             auth_cache => AuthCache
         },
     Credential = maybe_add_cert(Credential0, Channel),
-    do_authenticate(Credential, Channel);
+    authentication_pipeline(Credential, Channel);
 process_authenticate(
     ?CONNECT_PACKET(#mqtt_packet_connect{password = Password}),
     #channel{clientinfo = ClientInfo} = Channel
 ) ->
     %% Auth with CONNECT packet for MQTT v3
     Credential = maybe_add_cert(ClientInfo#{password => Password}, Channel),
-    do_authenticate(Credential, Channel);
+    authentication_pipeline(Credential, Channel);
 process_authenticate(
     ?AUTH_PACKET(_, #{'Authentication-Method' := AuthMethod} = Properties),
     #channel{
@@ -2205,7 +2218,7 @@ process_authenticate(
     case emqx_mqtt_props:get('Authentication-Method', ConnProps, undefined) of
         AuthMethod ->
             AuthData = emqx_mqtt_props:get('Authentication-Data', Properties, undefined),
-            do_authenticate(
+            authentication_pipeline(
                 ClientInfo#{
                     auth_method => AuthMethod,
                     auth_data => AuthData,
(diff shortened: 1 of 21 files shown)


apps/emqx_mt/src/emqx_mt_api.erl
```diff
@@ -193,9 +323,117 @@ client_count(get, #{bindings := #{ns := Ns}}) ->
         {error, not_found} -> ?NOT_FOUND("Namespace not found")
     end.
 
+tenant_limiter(get, #{bindings := #{ns := Ns}}) ->
+    %% TODO: check NS exists and was explicitly created.
```
thalesmg, line 327: This and similar occurrences are probably best handled in follow ups, when we implement more of the config?

apps/emqx_mt/src/emqx_mt_state.erl
```diff
@@ -26,21 +26,34 @@
...
+%% Type declarations
+%%------------------------------------------------------------------------------
+
 -define(EMQX_MT_SHARD, emqx_mt_shard).
 
 %% mria tables
 -define(RECORD_TAB, emqx_mt_record).
 -define(COUNTER_TAB, emqx_mt_counter).
 -define(NS_TAB, emqx_mt_ns).
+-define(LIMITER_TAB, emqx_mt_limiter).
```
zmstone, line 49: there can be other configs in the future, maybe give it a more generic name ?
e.g. `emqx_mt_config`, and limiter being one of the configs.

apps/emqx_mt/src/emqx_mt_api.erl
```diff
@@ -157,12 +281,18 @@ last_clientid_in_query() ->
...
-mk(Type, Props) -> hoconsc:mk(Type, Props).
-
-array(Type) -> hoconsc:array(Type).
+fields(limiter_in) ->
+    [
+        {bytes, mk(ref(limiter_options), #{})},
+        {messages, mk(ref(limiter_options), #{})}
+    ];
+fields(limiter_out) ->
+    fields(limiter_in);
```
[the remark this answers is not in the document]
thalesmg: they share the same schema now, so we can reuse it as an implementation detail.

giving separate names makes the distinction between them clearer, IMO

apps/emqx_mt/src/emqx_mt_api.erl
```diff
@@ -213,3 +451,34 @@ example_client_list() ->
...
+                    <<"burst">> => <<"200MB/1m">>
+                },
+                <<"messages">> => #{
+                    <<"rate">> => <<"3000/1s">>,
+                    <<"burst">> => <<"40/1m">>
+                }
+            }
+    }.
+
+limiter_out(LimiterConfigs) ->
```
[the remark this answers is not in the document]
thalesmg: > the `_in` and `_out` naming clashes the rate limiter direction, maybe `_req` or `_rsp`?

sorry, what do you mean with "clashes with rate limiter direction"?  we don't seem to have anything like that in the limiter configs or modules that I can see.  :thinking:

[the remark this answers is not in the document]
thalesmg: if the concern is about future types of limiters, maybe their adapter functions could be `{in,out}bound_limiter_{in,out}`, so there would be no clash

apps/emqx_mt/src/emqx_mt_state.erl
thalesmg: :checkered_flag:

apps/emqx_mt/src/emqx_mt_api.erl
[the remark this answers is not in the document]
zmstone: when i read `limiter_out`, i thought it's about the out-going message rate limiter. this is what i meant by 'clashing'.

apps/emqx_conf/src/emqx_conf_schema_inject.erl
```diff
@@ -29,6 +29,7 @@ schemas(Edition) ->
         authz() ++
         shared_subs(Edition) ++
         bridges(Edition) ++
+        channel_customizations() ++
```
zmstone, line 32: this injection has little to do with schema, maybe use a new injection mechanism.
e.g. `limiter_override_cb` as persistent term key and call it directly from emqx_mt app

apps/emqx/src/emqx_channel.erl
```diff
@@ -2088,6 +2105,23 @@ is_clientid_namespaced(ClientId, Tns) when is_binary(Tns) andalso Tns =/= <<>> -
...
     false.
 
+%%--------------------------------------------------------------------
+%% Adjust limiter
+
+adjust_limiter(_ConnPkt, #channel{clientinfo = ClientInfo} = Channel0) ->
+    #{zone := Zone, listener := ListenerId} = ClientInfo,
+    maybe
+        Tns = get_tenant_namespace(ClientInfo),
+        true ?= Tns /= undefined,
```
zmstone, line 2115: this assumes the limiter adjust only happens for multi-tenancy.
i think it's more logical that here in emqx_channel, it just evaluates the callback without checks.
i.e. it's the `emqx_mt` callback's job to decide if adjust should happen or not based on `Tns =/= undefined`.

apps/emqx_mt/src/emqx_mt_limiter.erl
```diff
@@ -0,0 +1,191 @@
...
+    emqx_limiter:update_group(client_group(Ns), LimiterConfigs).
+
+delete_client_limiter_group(Ns) ->
+    emqx_limiter:delete_group(client_group(Ns)).
+
+%%------------------------------------------------------------------------------
+%% `emqx_schema_hooks' API
+%%------------------------------------------------------------------------------
+
+injected_values() ->
```
zmstone, line 110: this is no longer needed ?

apps/emqx/src/emqx_limiter/src/emqx_limiter_schema.erl
```diff
@@ -125,6 +128,35 @@ to_burst(Str) ->
...
+
+duration_ms_to_str(Interval) ->
+    AllowedTimeUnits = [<<"d">>, <<"h">>, <<"m">>, <<"s">>, <<"ms">>],
+    emqx_schema:duration_ms_to_str(Interval, AllowedTimeUnits).
+
+byte_capacity_to_str(0) ->
+    <<"0">>;
+byte_capacity_to_str(C0) when C0 rem ?GIGABYTE == 0 ->
+    C1 = C0 div ?GIGABYTE,
+    <<(integer_to_binary(C1))/binary, "gb">>;
```
zmstone, line 150:
```suggestion
    <<(integer_to_binary(C1))/binary, "GB">>;
```

apps/emqx/src/emqx_limiter/src/emqx_limiter_schema.erl
```diff
@@ -125,6 +128,35 @@ to_burst(Str) ->
...
+byte_capacity_to_str(C0) when C0 rem ?MEGABYTE == 0 ->
+    C1 = C0 div ?MEGABYTE,
+    <<(integer_to_binary(C1))/binary, "mb">>;
+byte_capacity_to_str(C0) when C0 rem ?KILOBYTE == 0 ->
+    C1 = C0 div ?KILOBYTE,
+    <<(integer_to_binary(C1))/binary, "kb">>;
```
zmstone, line 156:
```suggestion
    <<(integer_to_binary(C1))/binary, "KB">>;
```

zmstone (approved):

zmstone (dismissed):

zmstone (approved):

apps/emqx/src/emqx_channel.erl
```diff
@@ -2088,6 +2111,45 @@ is_clientid_namespaced(ClientId, Tns) when is_binary(Tns) andalso Tns =/= <<>> -
...
+-undef(PT_LIMITER_ADJUST_KEY).
+
+adjust_limiter(_ConnPkt, #channel{clientinfo = ClientInfo} = Channel0) ->
+    maybe
+        Fn = get_limiter_adjustment_fn(),
+        true ?= is_function(Fn, 1),
+        #{zone := Zone, listener := ListenerId} = ClientInfo,
+        Tns = get_tenant_namespace(ClientInfo),
+        Context = #{zone => Zone, listener_id => ListenerId, tns => Tns},
+        case Fn(Context) of
```
savonarola, line 2141: Nit
```
{ok, NewLimiter} ?= Fn(Context)
```

apps/emqx/src/emqx_channel.erl
```diff
@@ -2088,6 +2111,45 @@ is_clientid_namespaced(ClientId, Tns) when is_binary(Tns) andalso Tns =/= <<>> -
...
+    ok.
+-undef(PT_LIMITER_ADJUST_KEY).
+
+adjust_limiter(_ConnPkt, #channel{clientinfo = ClientInfo} = Channel0) ->
+    maybe
+        Fn = get_limiter_adjustment_fn(),
+        true ?= is_function(Fn, 1),
+        #{zone := Zone, listener := ListenerId} = ClientInfo,
+        Tns = get_tenant_namespace(ClientInfo),
+        Context = #{zone => Zone, listener_id => ListenerId, tns => Tns},
```
savonarola, line 2140: Nit

The `adjust` hook naming is quite generic, but the context content somewhat discovers that it is a tenant-specific limiter.
Also, we have two places for limiter creation.

Maybe
* call here a (new) regular fold hook (`emqx_hooks` based)
* pass the default (zone-listener) limiter as the default acc to it?

Thus we also won't need an additional hooking mechanism.

apps/emqx_mt/src/emqx_mt_limiter.erl
```diff
@@ -0,0 +1,185 @@
...
+create_limiter(Zone, ListenerId, Ns, Name) ->
+    TenantLimiters = create_tenant_limiters(Zone, Ns, Name),
+    ClientLimiters = create_client_limiters(ListenerId, Ns, Name),
+    Clients = TenantLimiters ++ ClientLimiters,
+    emqx_limiter_composite:new(Clients).
+
+create_tenant_limiters(Zone, Ns, Name) ->
+    ZoneLimiterId = {zone_group(Zone), Name},
+    ZoneLimiterClient = emqx_limiter:connect(ZoneLimiterId),
+    case emqx_mt_config:get_tenant_limiter_config(Ns) of
```
savonarola, line 166: Curious

If the explicit ns configs appear later, do we agree to allow existing clients to remain unlimited by the new ns limits?

thalesmg: Yes, I discussed this with @zmstone and reached this conclusion:

> we can keep it simple: only new ones will use the new config.
worst case, one can query all clients using the ns and kick them to force reconnect.

apps/emqx/src/emqx_channel.erl
thalesmg: > The adjust hook naming is quite generic, but the context content somewhat discovers that it is a tenant-specific limiter.

The _current context_ only needs this, so it doesn't imply it's _only_ for tenant stuff.  More stuff may be added to the context in the future, as the need arises.  It's useful to use a map for this, since they are extensible.

The problem I see with creating the limiter only at this point is that we'll have to create an "inconsistent" `#channel{}` record with `quota = undefined`, until it reaches this point of the processing pipeline.  I wanted to have _something_ there until we have enough information to potentially change it.  :thinking:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6817)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6817)

PR #14861, opened 2025-03-12
test(crl cache): attempt to fix flaky test

thalesmg:
```
%%% emqx_crl_cache_SUITE ==> t_revoked: FAILED
%%% emqx_crl_cache_SUITE ==> {{case_clause,{error,connack_timeout}},
 [{emqx_crl_cache_SUITE,t_revoked,1,
                        [{file,"/__w/emqx/emqx/apps/emqx/test/emqx_crl_cache_SUITE.erl"},
                         {line,899}]},
  {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1794}]},
  {test_server,run_test_case_eval1,6,[{file,"test_server.erl"},{line,1303}]},
  {test_server,run_test_case_eval,9,[{file,"test_server.erl"},{line,1235}]}]}
```

### DIFF
--- a/apps/emqx/src/emqx_ssl_crl_cache.erl
+++ b/apps/emqx/src/emqx_ssl_crl_cache.erl
@@ -65,6 +65,8 @@
 -include_lib("ssl/src/ssl_internal.hrl").
 -include_lib("public_key/include/public_key.hrl").
 
+-include_lib("snabbkaffe/include/trace.hrl").
+
 -include("logger.hrl").
 
 -behaviour(ssl_crl_cache_api).
@@ -177,7 +179,9 @@ do_insert(URI, CRLs) ->
     case uri_string:normalize(URI, [return_map]) of
         #{scheme := "http", path := _} ->
             Key = cache_key(URI),
-            ssl_manager:insert_crls(Key, CRLs);
+            Res = ssl_manager:insert_crls(Key, CRLs),
+            ?tp("emqx_ssl_crl_cache_inserted", #{key => Key}),
+            Res;
         _ ->
             {error, {only_http_distribution_points_supported, URI}}
     end.
--- a/apps/emqx/test/emqx_crl_cache_SUITE.erl
+++ b/apps/emqx/test/emqx_crl_cache_SUITE.erl
@@ -348,7 +348,11 @@ start_emqx_with_crl_cache(#{is_cached := IsCached} = Opts, TC, Config) ->
     case IsCached of
         true ->
             %% wait the cache to be filled
-            emqx_crl_cache:refresh(?DEFAULT_URL),
+            {_, {ok, _}} =
+                ?wait_async_action(
+                    emqx_crl_cache:refresh(?DEFAULT_URL),
+                    #{?snk_kind := "emqx_ssl_crl_cache_inserted"}
+                ),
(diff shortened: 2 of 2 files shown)


keynslug (approved):

thalesmg:
> In these occasions TLS socket appears to be blocked however, by a CRL server request perhaps?

I don't think it _should_, because in this test we [first ensure](https://github.com/emqx/emqx/blob/a32125f6d610881f7dc03fda0a0c24618cc6435b/apps/emqx/test/emqx_crl_cache_SUITE.erl#L348-L352) the CRL is cached.

It could be the case if the `client-revoked.cert.pem` distribution point didn't match what we cached.  Inspecting the certificate, it seems to match.  :thinking:

thalesmg:
... though, on second thought, we are possibly _not waiting enough_, so you possibly have a point. :melting_face: 

We wait for the HTTP GET to be made, but not the actual write to the ETS table...  :melting_face: 

Now that in 5.x we have a fork of `ssl_crl_cache`, we can add a tracepoint there to be more accurate.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6504)

thalesmg:
Still happens :melting_face: 

The server does detect the certificate is revoked, as seen in the test logs:

https://github.com/emqx/emqx/actions/runs/13906098844/job/38909929379?pr=14867#step:5:356

```
 =NOTICE REPORT==== 17-Mar-2025::17:48:50.116672 ===
TLS server: In state wait_cert at ssl_handshake.erl:2200 generated SERVER ALERT: Fatal - Certificate Revoked

=INFO REPORT==== 17-Mar-2025::17:48:50.117267 ===
    supervisor: {esockd_connection_sup,<0.238227.0>}
    errorContext: ssl_error
    reason: {tls_alert,
                {certificate_revoked,
                    "TLS server: In state wait_cert at ssl_handshake.erl:2200 generated SERVER ALERT: Fatal - Certificate Revoked\n"}}
    offender: [{pid,<0.238227.0>},
               {name,connection},
               {mfargs,
                   {emqx_connection,start_link,
                       [#{zone => default,
                          listener => {ssl,default},
                          enable_authn => true}]}}]
```

... but, for some reason, this error seems to sometimes fail to be replied to the `emqtt:connect/1` caller.  :thinking:

thalesmg:
Thinking about this snippet:

https://github.com/emqx/emqtt/blob/58b1140fd24ec90ac0c2403afce791a8da702b96/src/emqtt.erl#L1531-L1540

If we have a [TLS socket](https://github.com/emqx/emqtt/blob/58b1140fd24ec90ac0c2403afce791a8da702b96/src/emqtt_internal.hrl#L17), then the error that the `emqtt` receives above will never match the internal `#ssl_socket{}` record, right?  So the error will be lost, and we reach the `connack_timeout` error seen in CI. :thinking:

thalesmg:
After adding some debug logs to the test, I found out why the process hangs:  https://github.com/emqx/emqtt/pull/279

keynslug:
> After adding some debug logs to the test, I found out why the process hangs: [emqx/emqtt#279](https://github.com/emqx/emqtt/pull/279)

Good find!

PR #14867, opened 2025-03-13
fix(resource manager): correctly handle channel updates

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14016

Release version: v/e5.9.0

## Summary

Now that we perform channel changes (add/remove) asynchronously during config updates,
there’s a chance that we may not reset the metrics when updating an action/source.

An example caught in CI:

https://github.com/emqx/emqx/actions/runs/13839069149/job/38722213036?pr=14860#step:4:680

```
%%% emqx_bridge_v2_kafka_producer_SUITE ==> t_overflow_rule_metrics: FAILED
%%% emqx_bridge_v2_kafka_producer_SUITE ==> 
Failure/Error: ?assertMatch(# { counters := # { dropped := 1 , 'dropped.queue_full' := 1 , 'dropped.expired' := 0 , 'dropped.other' := 0 , matched := 1 , success := 0 , failed := 0 , late_reply := 0 , retried := 0 } }, emqx_bridge_v2 : get_metrics ( Type , ActionName ))
  expected: = # { counters := # { dropped := 1 , 'dropped.queue_full' := 1 , 'dropped.expired' := 0 , 'dropped.other' := 0 , matched := 1 , success := 0 , failed := 0 , late_reply := 0 , retried := 0 } }
       got: #{counters =>
                  #{dropped => 2,success => 0,matched => 2,failed => 0,
                    received => 0,retried => 0,'retried.success' => 0,
                    'retried.failed' => 0,late_reply => 0,
                    'dropped.expired' => 0,'dropped.queue_full' => 2,
                    'dropped.resource_not_found' => 0,
                    'dropped.resource_stopped' => 0,'dropped.other' => 0},
              rate => #{matched => #{max => 2.0,current => 2.0,last5m => 2.0}},
              gauges => #{queuing => 0,queuing_bytes => 0},
              slides => #{},hists => #{}}
      line: 1687
```

Possible timeline:

1) `emqx_resource_manager:remove_channel_async` is called.  We don’t drop the metrics at
   this point; that’s done by the resource manager when it actually removes the channel.
2) In an update, `emqx_bridge_v2:install_bridge_v2` is called just after that, and it
   immediately creates the metrics outside the resource manager request.
3) This “create metrics” will preserve the old counters, and if the removal request is
   compressed together with the add request internally by the resource manager, it may not
   drop the metrics (which would also lead to problems…)

The causes were actually bad handling of updated channel configurations:

* When a connector resource goes `disconnected` and back to `connected`, it completely
  wipes the resource state, while we considered the channel to be installed in resource
  manager, which led to inconsistencies.
* If: a channel was already added and a new `#add_channel{}` request arrived later (or if
  a remove/add pair was present but remove failed), it wouldn't update the channel.

## PR Checklist

- [na] ~~Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files~~ this async config change change is unreleased.

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_manager.erl
+++ b/apps/emqx_resource/src/emqx_resource_manager.erl
@@ -180,6 +180,7 @@
 -record(start_channel_health_check, {channel_id :: channel_id()}).
 -record(retry_add_channel, {channel_id :: channel_id()}).
 -record(retry_remove_channel, {channel_id :: channel_id()}).
+-record(retry_update_channel, {channel_id :: channel_id()}).
 -record(get_channel_configs, {}).
 
 -type generic_timeout(Id, Content) :: {{timeout, Id}, timeout(), Content}.
@@ -192,8 +193,12 @@
 -type retry_remove_channel_action() :: generic_timeout(
     #retry_remove_channel{}, #retry_remove_channel{}
 ).
+-type retry_update_channel_action() :: generic_timeout(
+    #retry_update_channel{}, channel_config()
+).
 -type retry_add_channel_event() :: #retry_add_channel{}.
 -type retry_remove_channel_event() :: #retry_remove_channel{}.
+-type retry_update_channel_event() :: #retry_update_channel{}.
 
 %%------------------------------------------------------------------------------
 %% API
@@ -759,11 +764,6 @@ handle_event(internal, start_resource, ?state_connecting, Data) ->
     start_resource(Data, undefined);
 handle_event(state_timeout, health_check, ?state_connecting, Data) ->
     start_resource_health_check(Data);
-handle_event(
-    cast, #remove_channel{channel_id = _ChannelId} = Op, ?state_connecting = State, Data0
-) ->
-    {Actions, Data} = collect_and_handle_channel_operations_not_connected(Op, State, Data0),
-    {keep_state, Data, Actions};
 %%--------------------------
 %% State: CONNECTED
 %% The connected state is entered after a successful on_start/2 of the callback mod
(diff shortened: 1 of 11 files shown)


apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -991,35 +975,96 @@ start_resource(Data, From) ->
...
+        {#{}, #{}},
         ChannelIdConfigTuples
     ),
-    Data#data{added_channels = NewChannels}.
+    remove_channels_from_previous_incarnation(maps:keys(ChannelsToRemove), PreviousIncarnationData),
+    %% Immediately add attempt to add channels to resource state to avoid inconsistent
+    %% cached resource state while concurrently serving requests.
+    maps:fold(
+        fun(ChannelId, Config, {AccActions, AccData}) ->
+            {Actions, Data} = handle_add_channel(_From = undefined, AccData, ChannelId, Config),
```
thalesmg, line 1023: Should be add_channels_in_list

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -991,35 +975,96 @@ start_resource(Data, From) ->
                 },
                 #{tag => tag(Group, Type)}
             ),
-            _ = maybe_alarm(?status_disconnected, IsDryRun, ResId, Err, Data#data.error),
+            _ = maybe_alarm(?status_disconnected, IsDryRun, ResId, Err, Data0#data.error),
             %% Add channels and raise alarms
-            ensure_channel_metrics_exist(Data),
-            {Actions0, NewData1} = channels_health_check(?status_disconnected, add_channels(Data)),
+            ensure_channel_metrics_exist(Data0),
+            {Actions0, Data1} = add_channels_to_fresh_state(Data0, Data0),
```
thalesmg, line 981: Shouldn't add if disconnected

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6828)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6828)

PR #14885, opened 2025-03-17
250315 refactor resource id for authn authz

zmstone:
Release version: 5.9.0

## Summary

This is pure refactoring to improve logging data (in order to help AI to parse the log fields).
Previously the `resource_id` in the logs does not quite align with the authenticator ID (which is of format `MECHANISM:BACKEND`).

This PR changes the internal format of the resource ID to `authn:TYPE:NUMBER` such as `authn:postgres:1` and the `owner_id` to `MECHANISM:BACKEND` such as `password_based:postgres`.

Same for authz connectors.
- `authz:TYPE:NUMBER` for internal resource ID such as `auth:mysql:2`
- `TYPE` for `owner_id`, such as `mysql`.

### DIFF
--- a/apps/emqx_auth_mysql/src/emqx_authn_mysql.erl
+++ b/apps/emqx_auth_mysql/src/emqx_authn_mysql.erl
@@ -47,7 +47,11 @@ create(Config0) ->
 
 update(Config0, #{resource_id := ResourceId} = _State) ->
     {Config, NState} = parse_config(Config0),
-    case emqx_authn_utils:update_resource(emqx_mysql, Config, ResourceId) of
+    case
+        emqx_authn_utils:update_resource(
+            emqx_mysql, Config, ResourceId, ?AUTHN_MECHANISM_BIN, ?AUTHN_BACKEND_BIN
+        )
+    of
         {error, Reason} ->
             error({load_config_error, Reason});
         {ok, _} ->
--- a/apps/emqx_auth_mysql/src/emqx_authn_mysql.erl
+++ b/apps/emqx_auth_mysql/src/emqx_authn_mysql.erl
@@ -38,7 +38,7 @@ create(_AuthenticatorID, Config) ->
     create(Config).
 
 create(Config0) ->
-    ResourceId = emqx_authn_utils:make_resource_id(<<"mysql">>),
+    ResourceId = emqx_authn_utils:make_resource_id(?AUTHN_BACKEND_BIN),
     {Config, State} = parse_config(Config0),
     {ok, _Data} = emqx_authn_utils:create_resource(
         ResourceId, emqx_mysql, Config, ?AUTHN_MECHANISM_BIN, ?AUTHN_BACKEND_BIN
--- a/apps/emqx_auth_mysql/src/emqx_authn_mysql.erl
+++ b/apps/emqx_auth_mysql/src/emqx_authn_mysql.erl
@@ -16,8 +16,6 @@
 
 -module(emqx_authn_mysql).
 
--include_lib("emqx_auth/include/emqx_authn.hrl").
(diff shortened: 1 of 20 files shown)


apps/emqx_auth_postgresql/src/emqx_authn_postgresql.erl
```diff
@@ -42,12 +42,14 @@ create(_AuthenticatorID, Config) ->
     create(Config).
 
 create(Config0) ->
-    ResourceId = emqx_authn_utils:make_resource_id(?MODULE),
+    ResourceId = emqx_authn_utils:make_resource_id(<<"postgres">>),
```
JimMoen, line 45: maybe `<<"postgresql">>` like `?AUTHN_BACKEND_BIN`?

apps/emqx_auth_mysql/src/emqx_authn_mysql.erl
```diff
@@ -37,9 +38,11 @@ create(_AuthenticatorID, Config) ->
     create(Config).
 
 create(Config0) ->
-    ResourceId = emqx_authn_utils:make_resource_id(?MODULE),
+    ResourceId = emqx_authn_utils:make_resource_id(<<"mysql">>),
     {Config, State} = parse_config(Config0),
-    {ok, _Data} = emqx_authn_utils:create_resource(ResourceId, emqx_mysql, Config),
+    {ok, _Data} = emqx_authn_utils:create_resource(
+        ResourceId, emqx_mysql, Config, ?AUTHN_MECHANISM_BIN, ?AUTHN_BACKEND_BIN
+    ),
```
JimMoen, line 45: nit: Notice that some function calls use hard coded atom such as `'mysql'`, and some use hard coded binary `<<"mysql">>`.
Using macros uniformly may help improve readability.

funcs:
call `emqx_authn_utils:make_resource_id/1` used binary
call `emqx_authz_utils:make_resource_id/1` used atom
but they all return `binary()`

call `emqx_authn_utils:create_resource/5` used binary as `owner_id`
call `emqx_authz_utils:create_resource/5` used atom as `owner_id`
And record field `owner_id` is new added, Unified with atom/binary might be better?

zmstone: changed to use macros everywhere (except for jwt).
it's ok to be `binary()` or `atom()` since it's only intended for logging.

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -164,6 +164,27 @@
     ST =:= ?status_connecting; ST =:= ?status_connected; ST =:= ?status_disconnected
 ).
 
+%% If owner_id exists, log owner ID, otherwise log resource_id
+-define(LOG(LEVEL, FIELDS, DATA),
```
savonarola, line 168: Maybe wrap the body into a function or use a function? Otherwise, we will get "variable unsafe" if we use this macro twice in a scope.

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -164,6 +164,27 @@
...
+                #{
+                    tag => tag(DATA#data.group, DATA#data.type)
+                }
+            );
+        _ ->
+            ?SLOG(LEVEL, maps:merge(FIELDS, #{resource_id => binary_to_list(DATA#data.id)}), #{
+                tag => tag(DATA#data.group, DATA#data.type)
+            })
+    end
+).
```
thalesmg, line 186: Q: couldn't we just do something like `logger:set_process_metadata(#{...})` and this extra info would get logged (at least by the JSON formatter)?

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6833)

PR #14893, opened 2025-03-18
test: add debug logs for investigating flakiness

thalesmg:
Even after https://github.com/emqx/emqx/pull/14861, the problem still happens.

The server does recognize the certificate as revoked, but the `emqtt` process sometimes hangs and doesn't seem to receive the SSL error / SSL socket closed event.


thalesmg:
https://github.com/emqx/emqx/pull/14894

PR #14897, opened 2025-03-19
250319 sqlserver reconnect if health check failed

zmstone:
Release version: 5.8.6

## Summary

This is a followup fix for https://github.com/emqx/emqx/pull/14880

Force odbc reconnect if `SELECT 1` query got unexpected result.
Cannot simulate such error with real server, so the test is done by injecting error with a mock.

### DIFF
--- a/apps/emqx_bridge_sqlserver/src/emqx_bridge_sqlserver_connector.erl
+++ b/apps/emqx_bridge_sqlserver/src/emqx_bridge_sqlserver_connector.erl
@@ -217,6 +217,7 @@ on_start(
         {driver, Driver},
         {database, Database},
         {pool_size, PoolSize},
+        {auto_reconnect, 2},
         {on_disconnect, {?MODULE, disconnect, []}}
     ],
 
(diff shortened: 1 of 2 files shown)


savonarola (approved):

apps/emqx_bridge_sqlserver/src/emqx_bridge_sqlserver_connector.erl
```diff
@@ -374,9 +376,9 @@ do_get_status(Conn) ->
     case execute(Conn, <<"SELECT 1">>) of
         {selected, [[]], [{1}]} ->
             ok;
-        {error, Reason} ->
-            {error, Reason};
         Other ->
+            %% force restart if SELECT 1 returns error
+            disconnect(Conn),
```
thalesmg, line 381: Not needed if we use `auto_reconnect` and return `?status_disconnected` on errors.

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6549)

apps/emqx_bridge_sqlserver/src/emqx_bridge_sqlserver_connector.erl
thalesmg: not sure I understand

zmstone: without this disconnect here, the added test does not pass [here](https://github.com/emqx/emqx/pull/14897/files#diff-9460ba192c8e462aab2e4c59631f37c2fdbabe170e47af19e9447c986cc7cec7R321)

if returning `disconnected` status from health-check would trigger the entire pool to restart,
it does not explain why the odbc connection was not restarted (the on_stop did not happen?)
I thought it was because this connector does not call emqx_resource:allocate_resource, but after a closer look that does not seem to be related.

PR #14905, opened 2025-03-20
feat(mt api): add "kick all clients in namespace" endpoint

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14034

Release version: 5.9.0

## PR Checklist

- [na] ~~Change log has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files~~ unreleased feature


apps/emqx_mt/src/emqx_mt_client_kicker.erl
```diff
@@ -0,0 +1,165 @@
...
+
+%%------------------------------------------------------------------------------
+%% Type declarations
+%%------------------------------------------------------------------------------
+
+-define(ID(NS), {?MODULE, NS}).
+-define(SUP, emqx_mt_sup).
+
+-define(ns, ns).
+-define(last_clientid, last_clientid).
```
zmstone, line 39: maybe prev_clientid

apps/emqx_mt/src/emqx_mt_client_kicker.erl
```diff
@@ -0,0 +1,165 @@
...
+
+-define(kick, kick).
+
+-define(KICK_INTERVAL, 250).
```
zmstone, line 43: maybe kick a batch in a tight loop for 1000 clients then pause

apps/emqx_mt/src/emqx_mt_client_kicker.erl
```diff
@@ -0,0 +1,165 @@
...
+]).
+
+-include("emqx_mt.hrl").
+
+%%------------------------------------------------------------------------------
+%% Type declarations
+%%------------------------------------------------------------------------------
+
+-define(ID(NS), {?MODULE, NS}).
+-define(SUP, emqx_mt_sup).
```
zmstone, line 36: the root sup is a one_for_all. 
it's better to start the kicker workers under their own supervisor (simple_one_for_one)

thalesmg: this is a `temporary` process.  it will never be restarted, hence should not affect its siblings, right?

zmstone: temporary process's termination causes siblings to terminate since the supervisor is one_for_all

thalesmg: > one_for_all - If one child process terminates **_and is to be restarted_**, all other child processes are terminated and then all child processes are restarted.

Since this process is not to be restarted, it should be fine?

thalesmg: This simple test case demonstrates that it's indeed the case that temporary children (which are not significant) don't matter at all for restart intensity or for `one_for_all`:

```elixir
defmodule MySup do
  use Supervisor

  def start_link() do
    Supervisor.start_link(__MODULE__, nil, [])
  end

  def init(_) do
    children = [%{
      id: :existing,
      start: {Agent, :start_link, [fn ->
        IO.inspect("I'm starting!!!")
        1
      end]}}]
    Supervisor.init(children, strategy: :one_for_all, max_restarts: 1, max_seconds: 1)
  end
end

defmodule Temporary do
  use GenServer

  def child_spec(id) do
    %{
      id: id,
      start: {__MODULE__, :start_link, []},
      restart: :temporary
    }
  end

  def start_link() do
    GenServer.start_link(__MODULE__, nil, [])
  end

  def init(_) do
    {:ok, %{}, {:continue, :die}}
  end

  def handle_continue(_continue, _state) do
    raise "boom"
  end
end

{:ok, sup} = MySup.start_link()
## Should see "I'm starting!!!" from the child.

## Now we start our temporary child which will crash.
## You will not see the initialization message from the other child.
child_spec = Temporary.child_spec({:crasher, 1})
{:ok, _} = Supervisor.start_child(sup, child_spec)
```

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6835)

PR #14913, opened 2025-03-21
bump emqx enterprise dashboard version to e1.8.6

id:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6621)

PR #14936, opened 2025-03-27
fix(router): purge orphaned routing nodes during reconcile

keynslug:
Release version: 5.9.0

Fixes https://emqx.atlassian.net/browse/EMQX-14055.

## Summary

_Orphaned_ node here means a node that is marked as "routing node" in the respective table but cluster membership does not know about it.

Behavior is slightly changed, most notably:
* Nodes which are leaving / have left the cluster will be purged after 10 second delay, after cluster membership changes are completed.
* Nodes which have been "force-left" right after going down will be purged after 10 second delay, after cluster membership changes are completed; in other cases in ~1 minute during the next reconcile (same as before).

This also means that flapping cluster nodes may delay purges for considerable amount of time.

## PR Checklist

- [ ] Changed lines covered by tests
- [x] Change log has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files

### DIFF
--- a/apps/emqx/rebar.config
+++ b/apps/emqx/rebar.config
@@ -29,7 +29,7 @@
     {gproc, {git, "https://github.com/emqx/gproc", {tag, "0.9.0.1"}}},
     {cowboy, {git, "https://github.com/emqx/cowboy", {tag, "2.9.2"}}},
     {esockd, {git, "https://github.com/emqx/esockd", {tag, "5.13.1"}}},
-    {ekka, {git, "https://github.com/emqx/ekka", {tag, "0.21.1"}}},
+    {ekka, {git, "https://github.com/emqx/ekka", {tag, "0.21.2"}}},
     {gen_rpc, {git, "https://github.com/emqx/gen_rpc", {tag, "3.4.1"}}},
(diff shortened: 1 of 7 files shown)


zmstone (commented):
maybe extend the node leave/force-leave command/request to force purge routes.
or a separate cli for the rescue

zmstone (commented):
will you add a new ctl command to support manually clean routes ?

Makefile
```diff
@@ -28,7 +28,7 @@ CT_NODE_NAME ?= 'test@127.0.0.1'
 CT_READABLE ?= true
 CT_COVER_EXPORT_PREFIX ?= $(PROFILE)
 
-export REBAR_GIT_CLONE_OPTIONS += --depth=1
+export REBAR_GIT_CLONE_OPTIONS += --depth=3
```
zmstone, line 31: why this change ?

keynslug: Oops, accidentally put local only changes into a commit. Thanks for pointing out.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6930)

PR #14965, opened 2025-04-01
chore(new mix build): fix circular dependency

thalesmg:
Note: this only affects the new mix build


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6929)

PR #14966, opened 2025-04-01
feat(dashboard api): allow deleting default admin when others exist

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14053

Release version: 5.9.0

## Summary

This essentially allows one to rename the default admin user after a cluster is already started by creating a new admin and deleting the default one.

If there is at least one admin user other than the default one, we may delete the latter.

Also, since the `emqx_dashboard` application attempts to create the default user every time it starts, we use the same logic to avoid creating it.  

## PR Checklist

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard.erl
@@ -28,7 +28,7 @@
 %% Authorization
 -export([authorize/1]).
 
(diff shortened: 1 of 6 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6928)

PR #14975, opened 2025-04-02
fix(listener): use `esockd:reset_options/2` for online updates

keynslug:
Fixes [EMQX-14050](https://emqx.atlassian.net/browse/EMQX-14050).

Release version: 5.9

## Summary

* Employ https://github.com/emqx/esockd/pull/202 to perform online listener options update that were failing before, e.g.: restricting allowed TLS versions to `tlsv1.3` without changing any defaults.
* Remove workarounds for CRL options as no longer necessary.
* Massage options supplied to `emqx_tls_lib` more carefully, avoiding redefining defaults.
    * Move knowledge of `user_lookup_fun`, `crl_check` and `tls_auth_ext` options and their defaults there as well.

## PR Checklist

[EMQX-14050]: https://emqx.atlassian.net/browse/EMQX-14050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/emqx/src/emqx_tls_lib.erl
```diff
@@ -591,71 +592,106 @@ do_drop_invalid_certs([KeyPath | KeyPaths], SSL) ->
...
+    Ciphers = integral_ciphers(Versions, conf_get_opt(ciphers, Opts)),
+    DefaultUserLookupFun =
+        case Versions of
+            ['tlsv1.3'] -> undefined;
+            [_ | _] -> {fun emqx_tls_psk:lookup/3, undefined}
+        end,
+    TLSServerOpts = [
+        {versions, Versions},
+        {ciphers, Ciphers}
+        | emqx_utils:flattermap(
```
savonarola, line 610: Should we crash if an unknown option is found?
To avoid a situation where an option is added to the hocon schema, but this converter was not updated?

apps/emqx/test/emqx_listeners_SUITE.erl
```diff
@@ -382,6 +382,73 @@ t_ssl_update_opts(Config) ->
...
+                }
+            }}
+        ),
+
+        C3 = emqtt_connect_ssl(Host, Port, [
+            {certfile, filename:join(PrivDir, "client.pem")},
+            {keyfile, filename:join(PrivDir, "client.key")}
+            | ClientSSLOpts
+        ]),
+
```
savonarola, line 441: Should we check that connecting with previous options is not possible?

savonarola (approved):

apps/emqx/test/emqx_listeners_SUITE.erl
keynslug: Similar scenario is verified in `t_ssl_update_opts/1`, so I guess we can skip verifying it here?

apps/emqx/src/emqx_tls_lib.erl
keynslug: Good question. First thing to note is that `ocsp` option is already unknown in some sense (but we're aware of it). I'd argue that introduction of new schema fields should be paired with relevant feature tests. But, if it's for some reason not, and the field in question is optional, this "crash if unknown" logic would introduce a potential risk crashing listeners in runtime later, on someone's production system.

qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6939)

PR #14976, opened 2025-04-02
250402 precondition for authn

zmstone:
Fixes [EMQX-13883](https://emqx.atlassian.net/browse/EMQX-13883)

Release version: 5.9.0

## Summary

This will be useful when need to apply different auth mechanism based on client information.

As an example, if one wants to use HTTP auth for one listener, and MySQL auth for another, 
before 5.9, both authenticator will be called for clients from any of the listeners.
now it's possible to skip-over with a precondition like: `str_eq(listener, 'tcp:default')` -- which means invoke this authenticator only when listener is `tcp:default`.

Current auth chain on the left side, the new chain after this PR on the right side.

image

## PR Checklist

[EMQX-13883]: https://emqx.atlassian.net/browse/EMQX-13883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_auth/src/emqx_authn/emqx_authn_chains.erl
+++ b/apps/emqx_auth/src/emqx_authn/emqx_authn_chains.erl
@@ -35,6 +35,7 @@
     id :: binary(),
     provider :: module(),
     enable :: boolean(),
+    precondition :: undefined | emqx_variform:compiled(),
     state :: map()
 }).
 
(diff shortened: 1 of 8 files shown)


zmstone:
> Do we still need now individual authn chains for listeners?

Probably not, but we'll keep it there for now.

apps/emqx_auth/src/emqx_authn/emqx_authn_chains.erl
```diff
@@ -747,7 +748,39 @@ maybe_add_stacktrace('throw', Data, _Stacktrace) ->
...
+    Credential = add_cert_info(Credential0),
+    case emqx_variform:render(Precondition, Credential) of
+        {ok, <<"true">>} ->
+            true;
+        Other ->
+            Other
+    end.
+
+%% Add cert info to credential if it exists to aid the authentication placeholders.
+add_cert_info(#{cn := CN, dn := DN} = Credential) ->
```
savonarola, line 763: Nit: 
Now we have some duplication with `emqx_auth_template:rename_client_info_vars/1`

apps/emqx_auth/src/emqx_authn/emqx_authn_chains.erl
```diff
@@ -802,24 +835,55 @@ hook() ->
...
-                        provider = Provider,
-                        enable = Enable,
-                        state = State
-                    },
-                    {ok, Authenticator};
-                {error, Reason} ->
-                    {error, Reason}
-            end
+            do_create_authenticator2(AuthenticatorID, Config, Provider)
+    end.
```
savonarola, line 845: Nit:
maybe a good case for maybe, smth like:
```erlang
maybe
  Type = authn_type(Config),
  {ok, Provider} ?= find_provider(Type, Providers),
  {ok, Precondition} ?= compile_precondition(Config),
  {ok, State} ?= Provider:create(AuthenticatorID, Config),
  #{enable := Enable} = Config,
  {ok, #authenticator{...}}
end
```

savonarola (approved):

apps/emqx_auth/src/emqx_authn/emqx_authn_chains.erl
zmstone: A second thought, maybe won't fix this nit.
Current style incites better code coverage.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6938)

PR #14803, opened 2025-03-01
New license type

zmstone:

PR #14985, opened 2025-04-05
250404 refactor license take2

zmstone:
Release version: 5.9.0

## Summary

Added a RPC call `emqx_cluster:can_i_join/1` to if peer node allows the requesting node to join cluster.
Previously (5.9, but not released yet), the check is done locally, as a result it was not possible to join cluster without adding a license first --- this was considered a bad user experience.
Now the new node can join cluster without a license added, then license will be fetched from peer node after joined.

## PR Checklist

- [~] For internal contributor: there is a jira ticket to track this change (this is a follow-up fix for https://github.com/emqx/emqx/pull/14803

### DIFF
--- a/apps/emqx/src/proto/emqx_cluster_proto_v1.erl
+++ b/apps/emqx/src/proto/emqx_cluster_proto_v1.erl
@@ -16,6 +16,6 @@
 introduced_in() ->
     "5.9.0".
 
--spec can_i_join(node(), node()) -> ok | {error, string()} | {badrpc, term()}.
+-spec can_i_join(node(), node()) -> ok | {error, string()}.
 can_i_join(SelfNode, PeerNode) ->
     erpc:call(PeerNode, emqx_cluster, can_i_join, [SelfNode]).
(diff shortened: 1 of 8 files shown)


apps/emqx/src/proto/emqx_cluster_proto_v1.erl
```diff
@@ -0,0 +1,21 @@
...
+    introduced_in/0,
+    can_i_join/2
+]).
+
+-include("bpapi.hrl").
+
+introduced_in() ->
+    "5.9.0".
+
+-spec can_i_join(node(), node()) -> ok | {error, string()} | {badrpc, term()}.
```
thalesmg, line 19: Nit: I think `erpc:call` does not return `{badrpc, _}` values.

```suggestion
-spec can_i_join(node(), node()) -> ok | {error, string()}.
```

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-6987)

PR #15013, opened 2025-04-09
feat(rule engine api): add action details field

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13790

Release version: 5.9.0

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_api_schema.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_api_schema.erl
@@ -105,8 +105,21 @@ fields("rule_info") ->
                     desc => ?DESC("ri_last_modified_at"),
                     example => "2021-12-24T15:00:44.153+08:00"
                 }
+            )},
+        {"action_details",
+            sc(
+                hoconsc:array(ref("action_details")),
+                #{
+                    desc => ?DESC("ri_action_details")
+                }
             )}
     ] ++ fields("rule_creation");
+fields("action_details") ->
+    [
+        {"type", sc(binary(), #{desc => ?DESC("ri_action_details_type")})},
+        {"name", sc(binary(), #{desc => ?DESC("ri_action_details_name")})},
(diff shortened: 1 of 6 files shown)


apps/emqx_rule_engine/src/emqx_rule_engine_api.erl
```diff
@@ -609,6 +638,34 @@ do_format_action(#{mod := Mod, func := Func}) ->
...
+    #{action_summary_index := SummaryIndex} = Context,
+    Actions0 =
+        lists:filtermap(
+            fun
+                ({bridge, Type, Name}) ->
+                    {true, {bin(Type), bin(Name)}};
+                ({bridge_v2, Type, Name}) ->
+                    {true, {bin(Type), bin(Name)}};
+                (_) ->
+                    false
```
zmstone, line 651: no need to return republish actions ?

apps/emqx_rule_engine/src/emqx_rule_api_schema.erl
```diff
@@ -105,8 +105,21 @@ fields("rule_info") ->
...
+                hoconsc:array(ref("action_details")),
+                #{
+                    desc => ?DESC("ri_action_details")
+                }
             )}
     ] ++ fields("rule_creation");
+fields("action_details") ->
+    [
+        {"type", sc(binary(), #{desc => ?DESC("ri_action_details_type")})},
+        {"name", sc(binary(), #{desc => ?DESC("ri_action_details_name")})},
```
zmstone, line 120: make sense to return action ID instead?  or is split type and name useful to the frontend ?

ysfscream: Right now, actions don’t seem to have a dedicated ID — the combination of `type + name` is a unique identifier. So I think it’s fine as it is.

Kinplemelon: type and name is okay for us, `{type}:{name}` as action ID is more of an unspecified convention

PR #15014, opened 2025-04-09
feat(dashboard): lock login after several unsuccessful attempts

savonarola:
Fixes [EMQX-14093](https://emqx.atlassian.net/browse/EMQX-14093)

Release version: 5.9

## PR Checklist

[EMQX-14093]: https://emqx.atlassian.net/browse/EMQX-14093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_admin.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_admin.erl
@@ -424,8 +424,11 @@ change_password(Username, OldPasswd, NewPasswd, MfaToken) ->
 %% @doc Change password from CLI.
 change_password_trusted(Username, Password) when is_binary(Username), is_binary(Password) ->
     case legal_password(Password) of
-        ok -> change_password_hash(Username, hash(Password));
-        Error -> Error
+        ok ->
+            ok = emqx_dashboard_login_lock:reset(Username),
+            change_password_hash(Username, hash(Password));
+        Error ->
+            Error
     end.
 
 change_password_hash(Username, PasswordHash) ->
@@ -620,6 +623,7 @@ sign_token(Username, Password) ->
 sign_token(Username, Password, MfaToken) ->
     ExpiredTime = emqx:get_config([dashboard, password_expired_time], 0),
     maybe
+        ok ?= emqx_dashboard_login_lock:verify(Username),
         {ok, User} ?= check(Username, Password, MfaToken),
         {ok, Result} ?= verify_password_expiration(ExpiredTime, User),
         {ok, Role, Token} ?= emqx_dashboard_token:sign(User, Password),
--- a/apps/emqx_dashboard/src/emqx_dashboard_admin.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_admin.erl
@@ -34,6 +34,9 @@
     clear_mfa_state/1,
     set_mfa_state/2,
     get_mfa_state/1,
+    clear_login_lock/1,
+    set_login_lock/2,
+    get_login_lock/1,
     force_add_user/4,
     remove_user/1,
     update_user/3,
@@ -147,13 +150,13 @@ get_mfa_enabled_state(Username) ->
     end.
 
 get_mfa_state(Username) ->
-    case ets:lookup(?ADMIN, Username) of
-        [#?ADMIN{extra = #{mfa_state := S}}] ->
+    case get_extra(Username) of
+        {ok, #{mfa_state := S}} ->
             {ok, S};
-        [_] ->
+        {ok, _} ->
             {error, no_mfa_state};
-        [] ->
-            {error, username_not_found}
+        {error, _} = Error ->
+            Error
     end.
 
 %% @doc Delete MFA state from user record.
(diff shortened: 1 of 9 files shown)


apps/emqx_dashboard/src/emqx_dashboard_app.erl
```diff
@@ -31,7 +31,8 @@ start(_StartType, _StartArgs) ->
     Tables = lists:append([
         emqx_dashboard_admin:create_tables(),
         emqx_dashboard_token:create_tables(),
-        emqx_dashboard_monitor:create_tables()
+        emqx_dashboard_monitor:create_tables(),
+        emqx_dashboard_login_lock:create_tables()
```
zmstone, line 35: It's an overkill to create a table for this small feature.
Possible to make use of the `extra` field in the admin table? like the way TOTP is supported

apps/emqx_dashboard/src/emqx_dashboard_login_lock.erl
```diff
@@ -0,0 +1,280 @@
...
+%% API
+%%--------------------------------------------------------------------
+
+-spec register_unsuccessful_login(username()) -> ok.
+register_unsuccessful_login(Username) ->
+    NowUs = now_us(),
+    Key = {Username, NowUs},
+    Attempt = #login_attempts{key = Key, extra = #{}},
+    max_attempts_reached(Username, NowUs) andalso lock(Username, NowUs),
+    ok = mria:dirty_write(?TAB_ATTEMPTS, Attempt).
```
ieQu1, line 77: Does this write happen even if the username is locked?

If so, this feature becomes an easily explaitable DOS vector: since unsuccessful login attempts are stored in a `disc_copies` table (a.k.a. ets with some disc log persistence), every attempt will create a record in RAM. Fire them fast enough, and you crash the server.

apps/emqx_dashboard/src/emqx_dashboard_login_lock.erl
```diff
@@ -0,0 +1,280 @@
...
+
+%%--------------------------------------------------------------------
+%% Internal API
+%%--------------------------------------------------------------------
+
+create_tables() ->
+    ok = mria:create_table(?TAB_ATTEMPTS, [
+        {type, ordered_set},
+        {rlog_shard, ?DASHBOARD_SHARD},
+        {storage, disc_copies},
```
ieQu1, line 118: Why disc_copies and not, say, rocksdb_copies? Do we need to cache contents of the entire table in RAM?

apps/emqx_dashboard/src/emqx_dashboard_login_lock.erl
savonarola: Does this write happen even if the username is locked?

No, the record is added only on failed password check, and in case of a present lock, the password check is not performed.

apps/emqx_dashboard/src/emqx_dashboard_login_lock.erl
savonarola: The number of records is assumed to be the same order as the number of users, i.e. quite small.

ieQu1:
I question whether we need this functionality built into the broker, instead of writing an instruction how to use fail2ban with EMQX.

apps/emqx_dashboard/src/emqx_dashboard_login_lock.erl
ieQu1: What about non-existent usernames?

apps/emqx_dashboard/src/emqx_dashboard_login_lock.erl
ieQu1: Do we need this data constantly in RAM? It looks like this table is read only after an unsucessful login attempt to calculate the number of entries in a certain interval. 

This event shouldn't happen at high frequency, so `mnesia:select` should do.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -424,8 +449,11 @@ change_password(Username, OldPasswd, NewPasswd, MfaToken) ->
...
 change_password_trusted(Username, Password) when is_binary(Username), is_binary(Password) ->
     case legal_password(Password) of
-        ok -> change_password_hash(Username, hash(Password));
-        Error -> Error
+        ok ->
+            ok = emqx_dashboard_login_lock:reset(Username),
+            change_password_hash(Username, hash(Password));
+        Error ->
+            Error
     end.
```
JimMoen, line 457: Is it only possible to unlock the by changing the password through cli?

Considered: 
- the wrong password was entered by mistake due to ignoring the case lock.
- the wrong keyboard layout was configured, but the user did not find it when entering password, resulting in the wrong characters being entered (for example, qwerty -> azerty is configured).

At this time, usery may not want to change the password, just need to unlock it manually.
like: `pam_tally –user USER –reset`

savonarola: >Is it only possible to unlock by changing the password through cli?

That's the original UX requirement https://emqx.atlassian.net/browse/EEC-1160

>the wrong keyboard layout was configured

As far as I understand, the scenarios of human errors are mostly covered with a sufficient number of attempts (5 by default).

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7035)

PR #15015, opened 2025-04-10
ci: delete advanced mqtt features tests using jmeter

zmstone:
the tests are rather opaque and hard to troubleshoot if they fail.
for the auth tests, i value their coverage with different database versions.


qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7025)

PR #15036, opened 2025-04-14
fix(resource): trigger fallback actions when action is an "unhealthy target"

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14109

Release version: 5.9.0

## Summary

Previously, when the Action was marked as an "unhealthy target", the request would not even reach the buffering layer, where the fallback action triggering logic lives.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -2549,8 +2549,8 @@ result_context(Queries) ->
     #{?queries => Queries}.
 
 -doc """
-This an internal export to be used only in `emqx_resourcee` application in places where
-the request is fropped/fails before it even reaches the buffering layer (here).  An
+This an internal export to be used only in `emqx_resource` application in places where
+the request is dropped/fails before it even reaches the buffering layer (here).  An
 example of such situation is when the resource is deemed an "unhealthy target".
 """.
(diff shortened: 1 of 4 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -2545,6 +2548,27 @@ request_context(QueryOpts) ->
 result_context(Queries) ->
     #{?queries => Queries}.
 
+-doc """
+This an internal export to be used only in `emqx_resourcee` application in places where
```
savonarola, line 2552:
```suggestion
This an internal export to be used only in `emqx_resource` application in places where
```

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -2545,6 +2548,27 @@ request_context(QueryOpts) ->
...
+the request is fropped/fails before it even reaches the buffering layer (here).  An
```
savonarola, line 2553:
```suggestion
the request is dropped/fails before it even reaches the buffering layer (here).  An
```

savonarola (dismissed):

savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7057)

PR #15037, opened 2025-04-14
feat(limiter): handle dynamic zone update for limiters

savonarola:
Fixes [EMQX-14105](https://emqx.atlassian.net/browse/EMQX-14105)

Release version: 5.9

Although this goes as a fix, it seems we never supported dynamic zone update for rate limiting. 
This was just revealed by more strict listener creation.

## PR Checklist

[EMQX-14105]: https://emqx.atlassian.net/browse/EMQX-14105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_limiter/src/emqx_limiter.erl
+++ b/apps/emqx/src/emqx_limiter/src/emqx_limiter.erl
@@ -12,16 +12,6 @@
     init/0
 ]).
 
-%% Zone limiter management
--export([
-    create_zone_limiters/0,
-    update_zone_limiters/0,
-    delete_zone_limiters/0,
-    create_zone_limiters/1,
-    update_zone_limiters/1,
-    delete_zone_limiters/1
-]).
-
 %% Listener limiter management
 -export([
     create_listener_limiters/2,
@@ -45,10 +35,7 @@
 
 %% Config Listener
 -export([
-    add_handler/0,
-    remove_handler/0,
-    post_config_update/5,
-    propagated_post_config_update/5
+    post_zone_config_update/2
 ]).
 
 %%  Config helpers
@@ -62,7 +49,7 @@
 
 -export_type([zone/0, group/0, name/0, id/0, options/0]).
 
--type zone() :: group().
+-type zone() :: atom().
 -type group() :: term().
 -type name() :: atom().
 -type id() :: {group(), name()}.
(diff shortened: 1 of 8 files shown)


apps/emqx/src/emqx_limiter/src/emqx_limiter.erl
```diff
@@ -211,20 +171,34 @@ delete_group(Group) ->
...
+        end,
+        Added
+    ),
+    ok = maps:foreach(
+        fun(Zone, {_OldZoneConfig, ZoneConfig}) ->
+            update_limiters_for_zone(Zone, limiter_config_for_zone(ZoneConfig))
+        end,
+        Changed
+    ),
+    %% Never delete the default zone
```
thalesmg, line 197: Q: if the `default` zone _is indeed_ deleted (for whatever reason), shouldn't we also delete its limiters to avoid inconsistencies or leaking resources?

apps/emqx/test/emqx_limiter_SUITE.erl
```diff
@@ -140,16 +140,96 @@ t_metrics_quota_exceeded(_Config) ->
...
+    NewZoneConfig = OldZoneConfig#{<<"myzone">> => #{}},
+    {ok, _} = emqx:update_config([zones], NewZoneConfig),
+
+    %% Update the global defaults to have limits
+    emqx:update_config([mqtt, limiter], #{<<"messages_rate">> => <<"1/500ms">>}),
+
+    %% Check that the zone's limiter has applied the new limits
+    {_, LimiterOptions} = emqx_limiter_registry:find_group({zone, myzone}),
+    ?assertMatch(
+        #{capacity := 1},
```
thalesmg, line 220: Mini nit: should we assert that `capacity := infinity` before setting the global config?

thalesmg (dismissed):

apps/emqx/test/emqx_limiter_SUITE.erl
savonarola: ✅

apps/emqx/src/emqx_limiter/src/emqx_limiter.erl
savonarola: We do not have any logic that forbids the removal of used zones, so I left it to avoid the most common probable errors.
We shouldn't hit resource leakage, we preserve a few small values of constant size.

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7055)

PR #15039, opened 2025-04-14
more API features and fixes for fallback actions

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14116

Release version: 5.9.0

## PR Checklist


apps/emqx/src/emqx_config.erl
```diff
@@ -954,3 +956,43 @@ put_config_post_change_actions(_Key, _NewValue) ->
 
 config_files() ->
     application:get_env(emqx, config_files, []).
+
+remove_computed_fields(Config) ->
```
zmstone, line 960: maybe this should be a flag for the `Opts` arg for `hocon_tconf:map_translate/3`, so it will will not try to return computed field?.

apps/emqx_bridge/src/schema/emqx_bridge_v2_schema.erl
```diff
@@ -653,6 +665,50 @@ fallback_actions_republish_compute(Args0, _HoconOpts) ->
...
+                                (_, Acc2) ->
+                                    Acc2
+                            end,
+                            Acc1,
+                            FBAs
+                        );
+                    (_Name, _Conf, Acc1) ->
+                        Acc1
+                end,
+                Acc0,
```
zmstone, line 701: maybe break into smaller functions

apps/emqx/src/emqx_config.erl
thalesmg: it would have to be for `hocon_tconf:check_plain` in this case, but could be an alternative.

apps/emqx_bridge/src/schema/emqx_bridge_v2_schema.erl
thalesmg: :checkered_flag:

apps/emqx/src/emqx_config.erl
thalesmg: https://github.com/emqx/hocon/pull/306

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7075)

PR #15043, opened 2025-04-14
feat(dsraft): report basic DB, shard and shard transition metrics

keynslug:
Fixes [EMQX-13909](https://emqx.atlassian.net/browse/EMQX-13909).

Release version: 5.9

## PR Checklist


apps/emqx_prometheus/src/emqx_prometheus_ds_builtin_raft.erl
```diff
@@ -0,0 +1,116 @@
...
+-spec collect_mf(_Registry, Callback) -> ok when
+    _Registry :: prometheus_registry:registry(),
+    Callback :: prometheus_collector:collect_mf_callback().
+collect_mf(?PROMETHEUS_DSRAFT_REGISTRY, Callback) ->
+    Metrics = emqx_prometheus_cluster:raw_data(?MODULE, ?GET_PROM_DATA_MODE()),
+    collect_metric_families(Callback, metrics_meta(), Metrics);
+collect_mf(_, _) ->
+    ok.
+
+%% @private
```
savonarola, line 71: Nit: not clear what `private` refers to

apps/emqx_prometheus/src/emqx_prometheus_ds_builtin_raft.erl
```diff
@@ -0,0 +1,116 @@
...
+collect(<<"json">>) ->
+    #{
+        %% TODO
```
savonarola, line 74: Nit: Maybe explicitly state that we are not going to implement this and return empty?

savonarola (approved):

ieQu1:
Why is there an entirely separate API endpoint for Raft backend metrics, while the rest of DS metrics are available in the main EMQX prometheus endpoint? This is inconsistent from any perspective. All DS metrics should be in one place.

keynslug:
> Why is there an entirely separate API endpoint for Raft backend metrics, while the rest of DS metrics are available in the main EMQX prometheus endpoint? This is very inconsistent. All DS metrics should be in one place.

Because main endpoint is extremely overloaded, and huge pain to work with. If consistency is the goal here, I'd rather prefer to have separate endpoint for all DS metrics instead.

ieQu1 (changes requested):
I don't think it's acceptable design as it is. Either move all DS metrics to the new endpoint, or add new metrics to the old place.

Please just think about how confusing it will be for the users who want to monitor DS performance.

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7405)

PR #15052, opened 2025-04-16
ci: fix artifact upload in scheduled build packages

id:
Need to add branch from matrix to artifact name to avoid name clash.


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7403)

PR #15064, opened 2025-04-17
Dev/william/LSR

qzhuyan:
Fixes [EMQX-14233](https://emqx.atlassian.net/browse/EMQX-14233)

Release version: NOT 5.10.0

New Experimental cluster Linear Session Registry (LSR). ==> [EIP](https://github.com/emqx/eip/pull/93)

Added support for handling large-scale client reconnections. 
This feature prioritizes accepting the most recent client connection attempt over rejecting it in favor of older connections.

Enable the feature by configuring:

```
broker.enable_linear_session_registry = true
broker.enable_session_registry = false
```

Note: The feature is disabled by default and has the following limitations:  

- EMQX cluster nodes require precise time synchronization (time difference < 5ms) for accurate operation.  

- Channel registry history is not supported.

Untested Scenarios:

- Limited to MQTT connections; other transport or gateway protocols are untested.  

- Tested only with in-memory sessions, not durable storage.  

- Node evacuation with older EMQX versions has not been tested.

## PR Checklist

[EMQX-14233]: https://emqx.atlassian.net/browse/EMQX-14233?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -222,8 +222,8 @@ register_channel_local(ClientId, ChanPid, #{conn_mod := ConnMod, transport_start
     ok;
 register_channel_local(ClientId, ChanPid, ConnInfo) ->
     %% For backward compatibility.
-    %% when transport_started_at is absent, that means it is a call from older EMQX version
-    %% where the connected time must be known, but defaults to 0 for safety.
+    %% when `transport_started_at` is absent, that means it is a call from older EMQX version
+    %% where the MQTT connected time (`connected_at`) must be known, but defaults to 0 for safety.
     TS = maps:get(connected_at, ConnInfo, 0),
     register_channel_local(ClientId, ChanPid, ConnInfo#{transport_started_at => TS}).
 
@@ -331,11 +331,12 @@ set_chan_stats(ClientId, ChanPid, Stats) when ?IS_CLIENTID(ClientId) ->
 global_chan_cnt() ->
     case emqx_lsr:mode() of
         enabled ->
-            emqx_lsr:count_local_d();
+            emqx_lsr:count_dirty();
         disabled ->
-            emqx_cm_registry:count_local_d();
+            emqx_cm_registry:count_dirty();
         migration_enabled ->
-            emqx_cm_registry:count_local_d() + emqx_lsr:count_local_d()
+            %% It is inaccurate anyway...
+            emqx_cm_registry:count_dirty() + emqx_lsr:count_dirty()
     end.
 
 %% @doc Open a session and also register it in global channel/session registry
--- a/apps/emqx/src/emqx_lsr.erl
+++ b/apps/emqx/src/emqx_lsr.erl
@@ -21,7 +21,7 @@
     max_channel_d/1,
     register_channel/4,
     unregister_channel/2,
-    count_local_d/0,
+    count_dirty/0,
     do_cleanup_channels/1
 ]).
 
@@ -121,8 +121,8 @@ ch_pid(undefined) ->
 ch_pid(#lsr_channel{pid = Pid}) ->
     Pid.
 
--spec count_local_d() -> non_neg_integer().
-count_local_d() ->
+-spec count_dirty() -> non_neg_integer().
+count_dirty() ->
     mnesia:table_info(?LSR_TAB, size).
 
 %%--------------------------------------------------------------------
--- a/apps/emqx/src/emqx_cm_registry.erl
+++ b/apps/emqx/src/emqx_cm_registry.erl
@@ -37,7 +37,7 @@
     do_cleanup_channels/1
 ]).
 
--export([count_local_d/0]).
+-export([count_dirty/0]).
 
 -ifdef(TEST).
 %% For testing only
@@ -63,7 +63,8 @@ start_link() ->
 %%--------------------------------------------------------------------
 %% APIs
 %%--------------------------------------------------------------------
-count_local_d() ->
+-spec count_dirty() -> non_neg_integer().
+count_dirty() ->
     mnesia:table_info(?CHAN_REG_TAB, size).
 
 %% @doc Is the global registry enabled?


changes/ee/feat-15064.en.md
```diff
@@ -0,0 +1,27 @@
+New Experimental cluster Linear Session Registry (LSR)
+
+Added support for handling large-scale client reconnections. 
+This feature prioritizes accepting the most recent client connection attempt over rejecting it in favor of older connections.
+
+Enable the feature by configuring:
+
+```
+broker.enable_linear_channel_registry = true
```
qzhuyan, line 9: session registry

rel/i18n/emqx_schema.hocon
```diff
@@ -1507,6 +1507,9 @@ Advantages of Disabling<br/>
 - Reduced Memory Usage: Turning off the session registry can lower the overall memory footprint of the system.
 - Improved Performance: Without the overhead of maintaining a global registry, the node can process client connections faster."""
 
+broker_enable_linear_session_registry.desc:
+"""Enable Linearizable Session Registry for correctness."""
```
ieQu1, line 1511: Please elaborate what this feature does. Perhaps rehash the contents of changes/ee/feat-15064.en.md ?

apps/emqx/src/emqx_types.erl
```diff
@@ -163,6 +163,7 @@
     receive_maximum => non_neg_integer(),
     expiry_interval => non_neg_integer(),
     atom() => term()
+    % transport_started_at => integer(),
```
ieQu1, line 166: Forgotten code?

apps/emqx/src/emqx_types.erl
```diff
@@ -180,9 +181,10 @@
     anonymous => boolean(),
     cn => binary(),
     dn => binary(),
+    %% Following fields are never used in RPC calls.
```
ieQu1, line 184: Nit: "The following fields"

apps/emqx/include/emqx_lsr.hrl
```diff
@@ -0,0 +1,20 @@
...
+-define(lsr_err_channel_outdated, lsr_err_channel_outdated).
+-define(lsr_err_restart_takeover, lsr_err_restart_takeover).
+
+-record(lsr_channel, {
+    id :: lsr_session_id(),
+    pid :: pid() | '$1',
+    vsn :: integer() | undefined | '$2'
+}).
+
+-type lsr_session_id() :: emqx_types:clientid().
```
ieQu1, line 17: Nit: avoid type declaration in headers. It's better to find some module to house the type and export it from there.

apps/emqx/include/emqx_lsr.hrl
```diff
@@ -0,0 +1,20 @@
...
+
+-define(lsr_err_max_retries, lsr_err_max_retries).
+-define(lsr_err_channel_outdated, lsr_err_channel_outdated).
+-define(lsr_err_restart_takeover, lsr_err_restart_takeover).
+
+-record(lsr_channel, {
+    id :: lsr_session_id(),
+    pid :: pid() | '$1',
+    vsn :: integer() | undefined | '$2'
+}).
```
ieQu1, line 15: Nit: when a record is used in match, usually it's better to avoid in-place field type declarations. This code can be changed as following:

emqx_lsr.hrl:
```
-record(lsr_channel, {id, pid, vsn}).
```

emqx_cm.erl

```
-type lsr_channel() :: #lsr_channel{id :: lsr_session_id(), pid :: pid(), ...}.
```
```

File: apps/emqx_eviction_agent/src/proto/emqx_eviction_agent_proto_v4.erl [hunk 7]
```diff
@@ -0,0 +1,41 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+-module(emqx_eviction_agent_proto_v4).
```
ieQu1, line 5: Add 
```
deprecated_since() -> 
   "5.10.0".
```
to v3?

apps/emqx/test/emqx_lsr_SUITE.erl
```diff
@@ -0,0 +1,860 @@
...
+            {Res, []} = rpc:multicall(Nodes, emqx_cm, lookup_channels, [ClientId]),
+            [_] = lists:usort(lists:flatten(Res)),
+            ok
+        end
+    ).
+
+%% t_robustness(init, _Config) ->
+%%     ok.
+%% t_robustness(_) ->
+%%     todo.
```
ieQu1, line 712: Forgotten code?

qzhuyan: it is a reminder code for todo

apps/emqx/src/emqx_types.erl
qzhuyan: commented out for bpapi spec backward compatibility.

I will add the note

apps/emqx/src/emqx_types.erl
qzhuyan: commented out for bpapi spec backward compatibility.
I will add the note

changes/ee/feat-15064.en.md
```diff
@@ -0,0 +1,27 @@
+New Experimental cluster Linear Session Registry (LSR)
+
+Added support for handling large-scale client reconnections. 
+This feature prioritizes accepting the most recent client connection attempt over rejecting it in favor of older connections.
+
+Enable the feature by configuring:
+
+```
+broker.enable_linear_session_registry = true
```
zmstone, line 9: why isn't it enabled by default?

changes/ee/feat-15064.en.md
```diff
@@ -0,0 +1,27 @@
...
+broker.enable_session_registry = false
```
zmstone, line 10: this flag was to turn off global session registration completely.
i.e. no cross-node session take-over at all.
I think we should keep this flag as it is intended:
`true`: either LSR or the legacy registration
`false`: no registration at all, no LSR, no legacy registration.

changes/ee/feat-15064.en.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+Note: The feature is disabled by default and has the following limitations:  
+
+- EMQX cluster nodes require precise time synchronization (time difference < 5ms) for accurate operation.  
+
+- Channel registry history is not supported.
+
+Untested Scenarios:
```
zmstone, line 19: Q: are they TODOs ? if yes, maybe turn this section to "future work"

changes/ee/feat-15064.en.md
qzhuyan: Yes. it is indeed to turn off global session registration completely becase LSR is a replacement of `global session registration`,  node could read from LSR to complte cross node session takeover/discard.

EIP described the need of parallel enabling of both LSR and legacy `global` session registry to possibley handle rolling upgrades. (not addressed in this PR yet)

apps/emqx/src/emqx_cm.erl
```diff
@@ -99,6 +104,9 @@
     do_get_chann_conn_mod/2
 ]).
 
+%% for mgmt
+-export([global_chan_cnt/0]).
```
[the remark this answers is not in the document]
qzhuyan: it is channel actually, as multiple channels for same session could exist when these processes are terminating.

apps/emqx/include/emqx_lsr.hrl
ieQu1: Yes, also it creates copies of the same type in multiple modules, which may be confusing in dialyzer errors.

changes/ee/feat-15064.en.md
zmstone: maybe better to write "future work" and "limitations" instead of "untested"

apps/emqx/src/emqx_cm.erl
```diff
@@ -163,25 +171,46 @@ insert_channel_info(ClientId, Info, Stats) when ?IS_CLIENTID(ClientId) ->
...
+    case emqx_lsr:register_channel(ClientInfo, ChanPid, ConnInfo) of
+        ok ->
+            register_channel_local(ClientId, ChanPid, ConnInfo);
+        {error, _} = Err ->
+            Err
+    end;
+register_channel(ClientId, ChanPid, ConnInfo) when
     is_pid(ChanPid) andalso ?IS_CLIENTID(ClientId)
 ->
+    %% Note that: It maybe be called in a locked transaction.
```
ieQu1, line 191: Nit:
```suggestion
    %% Note that: It may be called in a locked transaction.
```

apps/emqx/src/emqx_cm.erl
```diff
@@ -163,25 +171,46 @@ insert_channel_info(ClientId, Info, Stats) when ?IS_CLIENTID(ClientId) ->
...
     ok = cast({registered, ChanPid}),
     true = ets:insert(?CHAN_TAB, Chan),
-    ok = emqx_cm_registry:register_channel(Chan),
     ok = mark_channel_connected(ChanPid),
-    ok.
+    ok;
+register_channel_local(ClientId, ChanPid, ConnInfo) ->
+    %% For backward compatibility.
+    %% when transport_started_at is absent, that means it is a call from older EMQX version
+    %% where the connected time must be known, but defaults to 0 for safty.
```
ieQu1, line 211:
```suggestion
    %% where the connected time must be known, but defaults to 0 for safety.
```

apps/emqx/src/emqx_lsr.erl
```diff
@@ -0,0 +1,285 @@
...
+unregister_channel({ClientId, ChanPid}, Vsn) ->
+    Ch = #lsr_channel{
+        id = ClientId,
+        pid = ChanPid,
+        vsn = Vsn
+    },
+    mria:dirty_delete_object(?LSR_TAB, Ch),
+    ok.
+
+-spec lookup_channels_d(emqx_types:clientid()) -> [lsr_channel()].
```
ieQu1, line 79: Nit: dirty_lookup_channels

apps/emqx/src/emqx_cm.erl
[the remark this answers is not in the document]
qzhuyan: > if it's not deleted after the transport is closed, then it's session.

I believe the `session` here is MQTT session ? then a client should have at most one session in the EMQX cluster,  this is part of the issue this PR try to fix.

A client could have many channels (closing, pending close, peer closed, opening, running, stuck) but there should have at most one session running when the client is connected or disconnected (session expire >0).

apps/emqx/include/emqx_lsr.hrl
qzhuyan: good point!  but I prefer to have types in place for readability. I will add a note for these values used for match spec.

changes/ee/feat-15064.en.md
qzhuyan: My bad I didn't write the doc for this experimental feature, here I updated it. 
https://github.com/emqx/emqx/pull/15064/commits/a79f17171461fea318144e92f514a2bda886f6fa

LSR provides global registry but it doesn't want to be controlled by `enable_session_registry` becasue it is used by `emqx_cm_registry` which is the old 'session registry' and has other bugs that I don't want to touch in this PR.

having two options to control both has the flexibility to operator like for upgrade or feature enabling.

apps/emqx/src/emqx_cm.erl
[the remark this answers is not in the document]
zmstone: If you insist this API should be named "channel", it's fine as it is only used internally.
However, it's is called by API handler which says "count sessions", so I think it's important to explain:

In EMQX's layered process state design:
- "connection" is to make transport agnostic (TCP, TLS, WS, WSS, QUIC) 
- "channel" wraps around "connection", with a goal to make this layer of the state agnostic to "connection" layer
- "session" is is the highest level which is exposed to users (management API callers).

When observed from external (dashboard or management API), one can see
- "connection": as it maps to a TCP link (MQTT client)
- "session": as it maps to a MQTT session 
 
"channel" however, is completely an internal abstraction layer.

apps/emqx/src/emqx_cm.erl
```diff
@@ -164,24 +172,60 @@ insert_channel_info(ClientId, Info, Stats) when ?IS_CLIENTID(ClientId) ->
...
     ok = cast({registered, ChanPid}),
     true = ets:insert(?CHAN_TAB, Chan),
-    ok = emqx_cm_registry:register_channel(Chan),
     ok = mark_channel_connected(ChanPid),
-    ok.
+    ok;
+register_channel_local(ClientId, ChanPid, ConnInfo) ->
+    %% For backward compatibility.
+    %% when transport_started_at is absent, that means it is a call from older EMQX version
+    %% where the connected time must be known, but defaults to 0 for safety.
```
zmstone, line 226:
```suggestion
    %% where the connected time must be unknown, but defaults to 0 for safety.
```

qzhuyan: connected time is known.

apps/emqx/src/emqx_cm.erl
```diff
@@ -265,7 +327,19 @@ set_chan_stats(ClientId, ChanPid, Stats) when ?IS_CLIENTID(ClientId) ->
...
-%% @doc Open a session.
+-spec global_chan_cnt() -> non_neg_integer().
+global_chan_cnt() ->
+    case emqx_lsr:mode() of
+        enabled ->
+            emqx_lsr:count_local_d();
+        disabled ->
+            emqx_cm_registry:count_local_d();
+        migration_enabled ->
+            emqx_cm_registry:count_local_d() + emqx_lsr:count_local_d()
```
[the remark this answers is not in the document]
qzhuyan: Yes count twice, it is written in the schema description.  We could only do dedup but I don't think it worth it, image you have a table of 5M records. 

We couldn't take max of two becasue there are cases that clients are in the emqx_cm_registry but not in LSR as old nodes may still serve traffic for new clients.  

It is inaccurate anyway, if you insist I could change it to max.

rel/i18n/emqx_schema.hocon
```diff
@@ -1507,6 +1507,26 @@ Advantages of Disabling<br/>
 - Reduced Memory Usage: Turning off the session registry can lower the overall memory footprint of the system.
 - Improved Performance: Without the overhead of maintaining a global registry, the node can process client connections faster."""
 
+broker_linear_session_registry.desc:
+"""Enable the Linearizable Session Registry (LSR) for improved correctness.
+Default: Disabled
+
+In real-world scenarios, a client may establish multiple network connections to the same or different nodes within a cluster due to repeated reconnection attempts.
+The LSR feature ensures that the most recent client connection is prioritized and accepted, rather than being rejected in favor of older ones.
```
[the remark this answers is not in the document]
qzhuyan: In cluser setup, ekka locker only protects the action (takeover, discard..) not the registration data, because updating 'global registration' are done async dirty, there is no order of the writes(new channel will be written after old channel) so takeover the older channel instead of the new could happen.

see 
https://github.com/emqx/emqx/blob/cd84ae83fcbb696b46779dde44a18dd94fabade6/apps/emqx/src/emqx_cm.erl#L334-L340

apps/emqx/src/emqx_cm.erl
```diff
@@ -265,7 +327,19 @@ set_chan_stats(ClientId, ChanPid, Stats) when ?IS_CLIENTID(ClientId) ->
...
     end.
 
-%% @doc Open a session.
+-spec global_chan_cnt() -> non_neg_integer().
+global_chan_cnt() ->
+    case emqx_lsr:mode() of
+        enabled ->
+            emqx_lsr:count_local_d();
+        disabled ->
+            emqx_cm_registry:count_local_d();
```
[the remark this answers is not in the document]
qzhuyan: I changed the names to `count_dirty`

PR #15083, opened 2025-04-22
fix(schema registry): convert SSL/TLS certificates for external http serde

thalesmg:
Inspired by https://emqx.atlassian.net/browse/EMQX-14159, but doesn't fix it (seems to be a frontend issue).

Release version: 5.9.0

## Summary

The certificates were save directly to config, without being converted to files, which should be fine.  Those certificate contents were already being returned by the HTTP API, but frontend does not display them.

## PR Checklist


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7151)

PR #15118, opened 2025-04-25
feat(emqx): add option to tune max allowed QoS per client subscription

keynslug:
Fixes [EMQX-9746](https://emqx.atlassian.net/browse/EMQX-9746).

Release version: 5.9.x

## Summary

Useful to limit QoS level requested by the client in a `SUBSCRIBE` packet, on a per-subscription basis.

Currently, limited to a small number of predicates evaluated over topic present in `SUBSCRIBE` request.

## PR Checklist


rel/i18n/emqx_schema.hocon
```diff
@@ -124,6 +124,22 @@ mqtt_max_qos_allowed.desc:
 mqtt_max_qos_allowed.label:
 """Max QoS"""
 
+mqtt_subscription_max_qos_rules.desc:
+"""Maximum Subscription QoS allowance rules. Useful to limit QoS level requested by the client in a `SUBSCRIBE` packet. Take precedence over `max_qos_allowed` option, which then acts as the default when no rule applies."""
```
ieQu1, line 128:
```suggestion
"""Maximum Subscription QoS allowance rules. Useful to limit QoS level requested by the client in a `SUBSCRIBE` packet. Takes precedence over `max_qos_allowed` option, which then acts as the default when no rule applies."""
```

ieQu1 (approved):

rel/i18n/emqx_schema.hocon
```diff
@@ -124,6 +124,22 @@ mqtt_max_qos_allowed.desc:
...
+
+mqtt_subscription_max_qos_rules.label:
+"""Max Subscription QoS Rules"""
+
+topic_qos_rule_match.desc:
+"""Expression evaluated over the given topic to decide if the rule applies.
+Follows `<topic-predicate>(topic, '<topic-argument>')` syntax where `<topic-argument>` is a string containing literal topic filter, and `<topic-predicate>` is one of:
+* `topic_equal`,
+* `topic_intersects`,
+* `topic_subset_of`."""
```
zmstone, line 138: we'll need some more description (with examples) for each predicate

apps/emqx/src/emqx_schema.erl
```diff
@@ -1882,6 +1902,25 @@ compile_variform(Expression, _Opts) ->
             throw(#{expression => Expression, reason => Reason})
     end.
 
+compile_topic_rule_match(Expression, Opts) ->
+    %% NOTE
+    %% Allow severely constrained variform expression as syntax for now, for
+    %% simplicity and predictability.
+    case compile_variform(Expression, Opts) of
+        Compiled = #{form := {call, TopicFn, [{var, "topic"}, {str, ArgStr}]}} when
```
zmstone, line 1910: if we are only compiling the expression, but not actually rendering the result, there is perhaps no need to force the usage of var `topic`, after all, there is only one possible input which is the topic used in SUBSCRIBE request.

keynslug: The goal here is to keep it forward-compatible with variform (or some future superset of it) while having limited subset as a first iteration, and having implicit arguments might compromise it in my opinion. To be clear, this is an UX concern: I think it's better to (eventually) have a single expression language in such contexts rather than several slightly different ones.

Moreover, this struct may potentially be used in other contexts besides SUBSCRIBE requests (most of contexts that come to mind admittedly also have just one well-defined `topic`, but still).

rel/i18n/emqx_schema.hocon
keynslug: Expanded schema doc in a0db7099528b0e877be1a9ab1836963c62bb3024, please tell if you think it's not the best place for that.

rel/i18n/emqx_schema.hocon
```diff
@@ -124,6 +124,35 @@ mqtt_max_qos_allowed.desc:
 mqtt_max_qos_allowed.label:
 """Max QoS"""
 
+mqtt_subscription_max_qos_rules.desc:
+"""Maximum Subscription QoS allowance rules. Useful to limit QoS level requested by the client in a `SUBSCRIBE` packet. Take precedence over `max_qos_allowed` option, which then acts as the default when no rule applies."""
```
ieQu1, line 128: It would be nice to have an example configuration.

rel/i18n/emqx_schema.hocon
```diff
@@ -124,6 +124,35 @@ mqtt_max_qos_allowed.desc:
...
+
+* `topic_equal`
+
+    Given topic or topic filter is simply equal to the argument.
+
+* `topic_subset_of`
+
+    Tests if given topic filter effectively covers a subset of topics that are covered by the topic filter in the argument.
+
+    For example, if `topic` is `t/conf/+` then `topic_subset_of(topic, 't/+/#')` is _true_ but `topic_subset_of(topic, 't/conf/+/+')` is _false_.
```
ieQu1, line 145: > `topic_subset_of(topic, 't/conf/+/+')`

What is `topic` in this expression? Is it a constant token that user should add into the configuration?

ieQu1: If I understand correctly, the full form of rule is 
```
{is_match: topic_subset_of(foo/#), qos: 0}
```

Why not
```
{topic: subset_of(foo/#), qos: 0}
```

apps/emqx/src/emqx_schema.erl
ieQu1: > I think it's better to (eventually) have a single expression language in such contexts rather than several slightly different ones.

What is variform? This is the only description of a programming language with this name I found on the internet: https://docs.emqx.com/en/emqx/latest/configuration/configuration.html#variform-expressions 

There's no BNF, no denotational semantics, just a couple of examples. It's not well known even among EMQX developers :wink: I suggest we postpone deep integration of this single expression language until it's well documented.

rel/i18n/emqx_schema.hocon
keynslug: > Why not

So that in the future it would be easier to extend to, for example:
```
{is_match: "topic_subset_of(topic, 'foo/#') and starts_with(clientid, 'bar')", qos: 0}
```

keynslug: Simplified to `{is_match: "topic.is_subset_of('foo/#')", qos: 0}` in 455059b201b4a03753ecd12aa5892aaa0f0c97a7.

rel/i18n/emqx_schema.hocon
keynslug: Added in 455059b201b4a03753ecd12aa5892aaa0f0c97a7. Please tell if examples need refinement.

apps/emqx/src/emqx_schema.erl
ieQu1: > The primary goal is forward compatibility: if we need to introduce more bindings in the evaluation context (e.g. listener or clientid) in the future, and more predicates, then having implicit arguments is going to bite.

What will happen **currently** if I try to use these additional predicates? Crash? Rule will silenty fail to match?

rel/i18n/emqx_schema.hocon
ieQu1: Oh, fantastic. Maybe we should extend scope of the PR to include . completion a la IntelliSense and AI copilot support? QA will have a field day with it.

apps/emqx/src/emqx_schema.erl
keynslug: Currently there's no additional predicates, so it's an error that will cause configuration to be rejected as invalid.

rel/i18n/emqx_schema.hocon
keynslug: Hah, good idea!

> Since we went from

On a serious note though, we didn't. There's 3 possible predicates over a topic, and that's it.

zmstone (dismissed):

rel/i18n/emqx_schema.hocon
ieQu1: I suggest we follow YAGNI (You Ain't Gonna Need It) rule and stick to the original requirements from the product. If we ever need to extend matching to different parameters, in addition to `{topic: ..., qos: 0}` we can add `{clientid: ...}`. Or go full DSL if you wish, but not until there's a demand for this.

ieQu1 (changes requested):
This is not according to the requirements. Requirements didn't include any DSL integrations.

id:
@keynslug change target branch to `release-510`

keynslug:
Updated according to the recent discussions, see 45d3762fca41c30a854844ff6eed79952a8d172b.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7643)

PR #15169, opened 2025-05-07
ci: delete schema-dump from push workflow

zmstone:
schema-dump is now only required in pr work flow (spellcheck)


id:
No, it's also used in https://github.com/emqx/emqx-docs/blob/release-5.9/.github/workflows/update-api-and-cfg-manual.yaml#L51.

We should it fix it though. Schema dump generation should be moved to update-emqx-docs.yaml.

PR #15174, opened 2025-05-07
feat(schema registry): support protobuf source bundles

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14089

Release version: 5.10.0

## PR Checklist

### DIFF
--- a/apps/emqx_schema_registry/src/emqx_schema_registry.erl
+++ b/apps/emqx_schema_registry/src/emqx_schema_registry.erl
@@ -53,11 +53,28 @@
 
 -define(BAD_SCHEMA_NAME, <<"bad_schema_name">>).
 
+-type schema_config_req() :: schema() | protobuf_bundle_req().
+
 -type schema() :: #{
     type := serde_type(),
     source := binary(),
     description => binary()
 }.
+
+%% #{
+%%     <<"type">> => <<"protobuf">>,
+%%     <<"source">> => #{
+%%         <<"type">> => <<"bundle">>,
+%%         <<"files">> => [
+%%             #{
+%%                 <<"root">> => true,
+%%                 <<"path">> => <<"some.proto">>,
+%%                 <<"contents">> => <<"message Bah {...">>
+%%             }
+%%         ]
+%%     }
+%% }
+-type protobuf_bundle_req() :: map().
 
 %%-------------------------------------------------------------------------------------------------
 %% API
@@ -125,7 +142,7 @@ get_schema_raw_with_defaults(Name) ->
             {error, not_found}
     end.
 
--spec add_schema(schema_name(), schema()) -> ok | {error, term()}.
+-spec add_schema(schema_name(), schema_config_req()) -> ok | {error, term()}.
 add_schema(Name, Schema) ->
     RawSchema = emqx_utils_maps:binary_key_map(Schema),
     Res = emqx_conf:update(
--- a/apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
+++ b/apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
@@ -442,7 +442,7 @@ delete_protobuf_bundle_data_dir(Name) ->
 copy_protobuf_bundle_to_data_dir(Name, RootFile, OtherFiles) ->
     DataDir = protobuf_bundle_data_dir(Name),
     ok = filelib:ensure_path(DataDir),
-    %% Root file is validate earlier in the pipeline.
+    %% Root file is validated earlier in the pipeline.
     #{<<"path">> := RootPath0} = RootFile,
     RootPath = filename:join([DataDir, RootPath0]),
     ok = filelib:ensure_dir(RootPath),
--- a/apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
+++ b/apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
@@ -115,10 +115,12 @@ delete_external_registry(Name) ->
 %% `emqx_config_handler' API
 %%------------------------------------------------------------------------------
 
-pre_config_update(?SCHEMA_CONF_PATH(Name), NewRawConf, _OldConf) ->
+pre_config_update(?SCHEMA_CONF_PATH(Name), RawConf0, _OldConf) ->
     maybe
         ok ?= validate_name(Name),
-        convert_certs(Name, NewRawConf)
+        {ok, RawConf1} ?= convert_certs(Name, RawConf0),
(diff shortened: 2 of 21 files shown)


changes/ee/feat-15174.en.md
```diff
@@ -0,0 +1 @@
+Added support to upload Protobuf source file bundles for Schema Registry.
```
zmstone, line 1: maybe give a cURL example

apps/emqx_schema_registry/src/emqx_schema_registry.erl
```diff
@@ -53,12 +53,29 @@
 
 -define(BAD_SCHEMA_NAME, <<"bad_schema_name">>).
 
+-type schema_config_req() :: schema() | protobuf_bundle_req().
+
 -type schema() :: #{
     type := serde_type(),
     source := binary(),
```
zmstone, line 60: why not expand this type to `source := binary() | map()`?

apps/emqx_schema_registry/src/emqx_schema_registry_schema.erl
```diff
@@ -111,7 +111,15 @@ fields(avro) ->
 fields(protobuf) ->
     [
         {type, mk(?protobuf, #{required => true, desc => ?DESC("schema_type_protobuf")})}
-        | common_fields(binary())
+        | common_fields(hoconsc:union([binary(), ref(protobuf_bundle_source)]))
+    ];
+fields(protobuf_bundle_source) ->
+    [
+        {type, mk(bundle, #{required => true, desc => ?DESC("protobuf_source_bundle_type")})},
+        {root_proto_path,
```
zmstone, line 119: I like it.
I think we should aim to convert other schema sources to files too, so to reference the name in cluster.hocon.
this makes below things slightly more complicated though:

1. initial copy of the directories https://github.com/emqx/emqx/blob/08562c47a7c71d391526baba6d5ea8f38c1bac9d/apps/emqx_conf/src/emqx_conf_app.erl#L72
2. data backup(export) and import

apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
```diff
@@ -314,3 +316,152 @@ convert_certs(Name, #{<<"parameters">> := #{<<"ssl">> := SSL} = Params0} = RawCo
...
+        ok ?= validate_protobuf_imports({path, RootPath}),
+        NewSource = #{
+            <<"type">> => <<"bundle">>,
+            <<"root_proto_path">> => RootPath
+        },
+        RawConf = RawConf0#{<<"source">> := NewSource},
+        {ok, RawConf}
+    else
+        {error, {invalid_or_missing_imports, _} = Reason} ->
+            delete_protobuf_bundle_data_dir(Name),
```
zmstone, line 355: should be in a try after clause ?
i.e. purge this dir if failed.

apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
```diff
@@ -314,3 +316,152 @@ convert_certs(Name, #{<<"parameters">> := #{<<"ssl">> := SSL} = Params0} = RawCo
...
+
+delete_protobuf_bundle_data_dir(Name) ->
+    DataDir = protobuf_bundle_data_dir(Name),
+    _ = file:del_dir_r(DataDir),
+    ok.
+
+copy_protobuf_bundle_to_data_dir(Name, RootFile, OtherFiles) ->
+    DataDir = protobuf_bundle_data_dir(Name),
+    ok = filelib:ensure_path(DataDir),
+    %% Root file is validate earlier in the pipeline.
```
zmstone, line 445:
```suggestion
    %% Root file is validated earlier in the pipeline.
```

apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
```diff
@@ -314,3 +316,152 @@ convert_certs(Name, #{<<"parameters">> := #{<<"ssl">> := SSL} = Params0} = RawCo
...
+            {error, <<"must have exactly one root file">>};
+        [_, _ | _] ->
+            {error, <<"must have exactly one root file">>};
+        [RootFile] ->
+            {ok, RootFile, OtherFiles}
+    end.
+
+protobuf_bundle_data_dir(Name) ->
+    DataDir = emqx:data_dir(),
+    filename:join([DataDir, "schemas", Name]).
```
zmstone, line 381: maybe add a space?
```suggestion
    filename:join([DataDir, "schemas", "proto", Name]).
```

apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
thalesmg: should only purge if failed, not always, hence this clause; `after` would always purge

changes/ee/feat-15174.en.md
thalesmg: :checkered_flag:

apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
thalesmg: should only delete new files...  if updating an existing schema, should not delete existing files if new schema fails to import.

apps/emqx_schema_registry/src/emqx_schema_registry_schema.erl
zmstone: yes, new node will sort of "rebase"

apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
zmstone: sorry, you are right, we should not delete in the 'after' clause.

I guess what I wanted to say is purge in case of failure

> should only delete new files

to make it simple, i guess we can either

- do not allow update, force delete first.
- use a temporary dir for the new schemas, delete the old and then rename (also lowers the chance of half-updated directory in case crashed in the middle)

thalesmg: > use a temporary dir for the new schemas, delete the old and then rename (also lowers the chance of half-updated directory in case crashed in the middle)

I was thinking of this approach;  there's still a small chance of being left in a broken state, which I was trying to avoid, but I guess we can't.  :crying_cat_face: 

> to make it simple: do not allow update, force delete first.

I think we can't dismiss this case here: when importing config backups, we need to overwrite what's there.  :thinking:

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7432)

PR #15175, opened 2025-05-07
feat(rule engine): support matching event topics with wildcards

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-13920

Release version: 5.10.0

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_events.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_events.erl
@@ -127,13 +127,8 @@ unload() ->
         event_names()
     ).
 
-unload(Topic) ->
-    lists:foreach(
-        fun(HookPoint) ->
-            emqx_hooks:del(HookPoint, {?MODULE, hook_fun_name(HookPoint)})
-        end,
-        match_event_names(Topic)
-    ).
+unload(EventName) ->
+    emqx_hooks:del(EventName, {?MODULE, hook_fun_name(EventName)}).
 
 %%--------------------------------------------------------------------
 %% Callbacks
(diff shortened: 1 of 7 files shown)


apps/emqx_rule_engine/src/emqx_rule_events.erl
```diff
@@ -121,8 +128,12 @@ unload() ->
     ).
 
 unload(Topic) ->
```
zmstone, line 130: unload should be expanded before calling.
unload_hooks_for_rule/1 should expand topic to matching events,
then for each event, check if there are other rules matching this event, and unhook this specific event if no other rule is selecting it.

thalesmg: not sure I understand this comment.  :thinking: 

> unload_hooks_for_rule/1 should expand topic to matching events,
then for each event, check if there are other rules matching this event, and unhook this specific event if no other rule is selecting it.

isn't it doing this already?

zmstone: the caller of unload checks if the topic (e.g. `$event/#`) has any overlapping with other rules, if yes, skip unloading ALL event hooks.
It should instead: enumerate each matching event, check if this is the last rule, unhook if yes.

thalesmg: I see now, good catch.  :see_no_evil:

zmstone (dismissed):

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7417)

PR #15190, opened 2025-05-11
Support hard-coded QoS in message transform

zmstone:
Release version: 5.9.1 or 5.10.0

## Summary

Previously, the Variform expression only supports boolean literals as expression root (hardcoding).
Now it supports numbers (integer, float) and strings too.

## PR Checklist


ieQu1:
I'm a bit concerned that more and more places start using "variform", while this language remains pretty much undocumented. An insufficiently documented DSL will cause great confusion among the users, and, even worse, each developer will have their own idea about the language, so everyone will end up extending it in incompatible ways. 
  
Below I'll pretend to be a user who is not familiar with this DSL and who can't or doesn't want to dig through the Erlang code.

Here's what I see in the docs:
https://docs.emqx.com/en/emqx/latest/configuration/configuration.html#variform-expressions

> Variform is a lightweight, expressive language designed for string manipulation and runtime evaluation. It is not a full-fledged programming language but a specialized tool that can be embedded within configurations for EMQX to perform string operations dynamically.

...

Ok, it's time to learn about this new language. 

> NULL value:
> In Variform expressions, a value-binding reference or sub-expression evaluation may result in an undefined value, which is represented as an empty string ('').
> It is important to note that if a JSON-decoded field is null, it is treated as undefined value (hence ''), but not string value "null".

Wait... NULLs, undefined values, JSON-encoded fields? Is this language based on JSON?.. I don't know anything about the syntax or semantics of variform yet, but the docs are already talking about some corner cases. Let's keep reading.

> To illustrate:
> function_call(clientid, another_function_call(username))

`clientid`, `username`? Are these variables? What is the scope of variform expressions? Are `clientid` and `username` always in the scope? 

> This expression combines or manipulates clientid and username to generate a new string value.

Well, `function_call` and `another_function_call` sound like generic placeholder names for functions. But the above paragraph says that they generate a *string* value. Do *all* variform expressions generate string values?..

Let's keep reading:

> Below are the functions that can be used in the expressions:
> String functions:
> [String Operation Functions](https://docs.emqx.com/en/emqx/latest/data-integration/rule-sql-builtin-functions.html#string-operation-functions)

The link leads to "Built-in SQL Functions" page. Is variform a subset EMQX SQL language?..

...

Control flow:

> iif(Condition, ThenExpression, ElseExpression): returns ThenExpression if Condition yields true or a non-empty string value

So there's type coercion, and `if` operator is C-style: https://ideone.com/n1j6g7 (which is confusing).
BTW, I already see places where the semantics of conditionals is totally different: https://github.com/emqx/emqx/discussions/15161 There the expression isn't considered valid when it's a non-empty string value, it's considered valid when it evaluates to string "true" (?). If tooltip is correct, that's the third way to represent true.

I suggest before we commit to adding variform everywhere, we should

1) Properly document it, for both users and developers
2) Clean up the semantics. Avoid type coercion (it should be impossible to write `iif("false", ...`))
3) Consider the performance. variform interpreter is written in Erlang... If some variform-based configuration ends up in the hot path (e.g. during processing of PUBLISH'es), then it may seriously hurt the performance, with little ways to troubleshoot it. (We can't ask all users who complain about the performance to share their entire configs to make sure there's no large variform expression in the hot path...). 
4) Optionally, implement some sort of type checking

zmstone:
@ieQu1  I agree the document should be improved. created [jira](https://emqx.atlassian.net/browse/EMQX-14240) to follow up.
I will later send 2 PRs, one to https://github.com/emqx/emqx-dev-docs for internal references, another to https://github.com/emqx/emqx-docs for users' references.

We can discuss semantics and performance in the PR to https://github.com/emqx/emqx-dev-docs.

changes/ee/fix-15190.en.md
```diff
@@ -0,0 +1 @@
+Support setting hard-coded QoS and topic in message transromation.
```
thalesmg, line 1:
```suggestion
Support setting hard-coded QoS and topic in message transformation.
```

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7428)

PR #15206, opened 2025-05-13
test(ft): attempt to fix flaky test case

thalesmg:
See https://github.com/emqx/emqx/pull/14994#discussion_r2033167363

### DIFF
--- a/apps/emqx_ft/test/emqx_ft_storage_fs_gc_SUITE.erl
+++ b/apps/emqx_ft/test/emqx_ft_storage_fs_gc_SUITE.erl
@@ -11,6 +11,7 @@
 -include_lib("stdlib/include/assert.hrl").
 -include_lib("snabbkaffe/include/snabbkaffe.hrl").
 -include_lib("common_test/include/ct.hrl").
+-include_lib("emqx/include/asserts.hrl").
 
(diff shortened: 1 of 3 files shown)


apps/emqx_ft/test/emqx_ft_storage_fs_gc_SUITE.erl
```diff
@@ -183,74 +283,71 @@ t_gc_complete_transfers(_Config) ->
...
-            ?snk_kind := garbage_collection,
-            stats := #gcstats{
-                files = Files,
-                directories = 4,
-                space = Space
-            }
-        } when Files == (?NSEGS(S1, SS1)) andalso Space > S1,
-        infinity,
-        0
-    ),
```
keynslug, line 228: The PR looks like an overkill to be honest. 🫠 I'd definitely suggest instead to just simplify test assertions: i.e. collect all the `garbage_collection` traces for 6 (?) seconds and accumulate stats in the check phase of `?check_trace(...)`. After all, what we really care about is that every incomplete transfer is collected eventually.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7438)

PR #15207, opened 2025-05-13
ci: fix azurite container health check

thalesmg:
Apparently, microsoft decided to change some things in the already tagged image.... 🫠

- `curl` no longer exists in the image.
- `GET /` now seems to return a 400 error, which would make `curl -f` fail anyway.


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7440)

PR #15208, opened 2025-05-13
Transactional key-value API for DS

ieQu1:
Fixes [EMQX-14049](https://emqx.atlassian.net/browse/EMQX-14049)

Release version: 5.10.0

## Summary

The best time to drop mnesia was 4 years ago when we started mria project, the next best time is now. This PR introduces a key-value transactional API for the durable storage with optimistic concurrency control.

## PR Checklist

[EMQX-14049]: https://emqx.atlassian.net/browse/EMQX-14049?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
+++ b/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
@@ -993,6 +993,7 @@ t_13_smoke_kv_tx(Config) ->
                 ok,
                 emqx_ds_open_db(DB, Opts)
             ),
+            timer:sleep(100),
             %% 1. Start a write-only transaction to create some data:
             {ok, Tx1} = emqx_ds:new_kv_tx(DB, #{
                 generation => 1,
@@ -1072,6 +1073,7 @@ t_14_kv_wildcard_deletes(Config) ->
                 ok,
                 emqx_ds_open_db(DB, Opts)
             ),
+            timer:sleep(100),
             %% 1. Insert test data:
             {ok, Tx1} = emqx_ds:new_kv_tx(DB, TXOpts),
             Msgs0 =
@@ -1134,6 +1136,7 @@ t_15_kv_write_serial(Config) ->
                 ok,
                 emqx_ds_open_db(DB, Opts)
             ),
+            timer:sleep(100),
             %% 1. Insert test data:
             {ok, Tx1} = emqx_ds:new_kv_tx(DB, TXOpts),
             Ops = #{
@@ -1173,6 +1176,7 @@ t_16_kv_preconditions(Config) ->
                 ok,
                 emqx_ds_open_db(DB, Opts)
             ),
+            timer:sleep(100),
             %% 1. Insert test data:
             {ok, Tx1} = emqx_ds:new_kv_tx(DB, TXOpts),
             Ops1 = #{
--- a/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
+++ b/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
@@ -1456,24 +1456,9 @@ all() ->
     [{group, Backend} || Backend <- backends()].
 
 groups() ->
-    AllTCs = emqx_common_test_helpers:all(?MODULE),
+    TCs = emqx_common_test_helpers:all(?MODULE),
     lists:map(
         fun(Backend) ->
-            TCs =
-                case Backend of
-                    emqx_ds_builtin_local ->
-                        AllTCs;
-                    _ ->
-                        AllTCs --
-                            [
-                                t_13_smoke_kv_tx,
-                                t_14_kv_wildcard_deletes,
-                                t_15_kv_write_serial,
-                                t_16_kv_preconditions,
-                                t_17_tx_wrapper,
-                                t_18_async_trans
-                            ]
-                end,
             {Backend, TCs}
         end,
         backends()
--- a/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
+++ b/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
@@ -993,7 +993,6 @@ t_13_smoke_kv_tx(Config) ->
                 ok,
                 emqx_ds_open_db(DB, Opts)
             ),
-            timer:sleep(100),
             %% 1. Start a write-only transaction to create some data:
             {ok, Tx1} = emqx_ds:new_kv_tx(DB, #{
                 generation => 1,
@@ -1073,7 +1072,6 @@ t_14_kv_wildcard_deletes(Config) ->
                 ok,
                 emqx_ds_open_db(DB, Opts)
             ),
-            timer:sleep(100),
             %% 1. Insert test data:
             {ok, Tx1} = emqx_ds:new_kv_tx(DB, TXOpts),
             Msgs0 =
@@ -1136,7 +1134,6 @@ t_15_kv_write_serial(Config) ->
                 ok,
                 emqx_ds_open_db(DB, Opts)
             ),
-            timer:sleep(100),
             %% 1. Insert test data:
             {ok, Tx1} = emqx_ds:new_kv_tx(DB, TXOpts),
             Ops = #{
@@ -1176,7 +1173,6 @@ t_16_kv_preconditions(Config) ->
                 ok,
                 emqx_ds_open_db(DB, Opts)
             ),
-            timer:sleep(100),
             %% 1. Insert test data:
             {ok, Tx1} = emqx_ds:new_kv_tx(DB, TXOpts),
             Ops1 = #{
--- a/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
+++ b/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
@@ -1057,7 +1057,7 @@ t_13_smoke_kv_tx(Config) ->
 %% API
 t_14_kv_wildcard_deletes(Config) ->
     DB = ?FUNCTION_NAME,
-    TXOpts = #{shard => {auto, <<"me">>}, timeout => infinity},
+    TXOpts = #{shard => {auto, <<"me">>}, timeout => infinity, generation => 1},
     ?check_trace(
         begin
             %% Open the database
@@ -1121,7 +1121,7 @@ t_14_kv_wildcard_deletes(Config) ->
 %% transaction serial.
 t_15_kv_write_serial(Config) ->
     DB = ?FUNCTION_NAME,
-    TXOpts = #{shard => {auto, <<"me">>}, timeout => infinity},
+    TXOpts = #{shard => {auto, <<"me">>}, timeout => infinity, generation => 1},
     Topic = [<<"foo">>],
     ?check_trace(
         begin
@@ -1159,7 +1159,7 @@ t_15_kv_write_serial(Config) ->
 
 t_16_kv_preconditions(Config) ->
     DB = ?FUNCTION_NAME,
-    TXOpts = #{shard => {auto, <<"me">>}, timeout => infinity},
+    TXOpts = #{shard => {auto, <<"me">>}, timeout => infinity, generation => 1},
     Topic1 = [<<"foo">>],
     Topic2 = [<<"bar">>],
     ?check_trace(
@@ -1337,6 +1337,7 @@ t_18_async_trans(Config) ->
                 ok,
                 emqx_ds_open_db(DB, Opts)
             ),
+            timer:sleep(100),
             %% 1. Successful async write transaction:
             {async, Ref1, hello} =
                 emqx_ds:trans(
--- a/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
+++ b/apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
@@ -978,6 +978,431 @@ t_sub_catchup_unrecoverable(Config) ->
         []
     ).
 
+%% Verify functionality of the low-level transaction API.
+t_13_smoke_kv_tx(Config) ->
+    DB = ?FUNCTION_NAME,
+    Owner = <<"test_clientid">>,
+    ?check_trace(
+        begin
+            %% Open the database
+            Opts = maps:merge(opts(Config), #{
+                store_kv => true,
+                storage => {emqx_ds_storage_kv_skipstream_lts, #{}}
+            }),
+            ?assertMatch(
+                ok,
+                emqx_ds_open_db(DB, Opts)
+            ),
+            %% 1. Start a write-only transaction to create some data:
+            {ok, Tx1} = emqx_ds:new_kv_tx(DB, #{
+                generation => 1,
+                shard => emqx_ds:shard_of(DB, Owner),
+                timeout => infinity
+            }),
+            Ops1 = #{
+                ?ds_tx_write => [
+                    {[<<"foo">>], <<"payload0">>},
+                    {[<<"t">>, <<"1">>], <<"payload1">>},
+                    {[<<"t">>, <<"2">>], <<"payload2">>}
+                ]
+            },
+            %% Commit:
+            ?assertMatch(
+                {ok, _Serial},
+                do_commit_tx(DB, Tx1, Ops1)
+            ),
+
+            %% Verify side effects of the transaction:
+            ?assertEqual(
+                [{[<<"t">>, <<"1">>], <<"payload1">>}],
+                emqx_ds:dirty_read(DB, [<<"t">>, <<"1">>])
+            ),
+            ?assertEqual(
+                [{[<<"t">>, <<"2">>], <<"payload2">>}],
+                emqx_ds:dirty_read(DB, [<<"t">>, <<"2">>])
+            ),
+            %%   All topics together:
+            ?assertEqual(
+                [
+                    {[<<"foo">>], <<"payload0">>},
+                    {[<<"t">>, <<"1">>], <<"payload1">>},
+                    {[<<"t">>, <<"2">>], <<"payload2">>}
+                ],
+                lists:sort(emqx_ds:dirty_read(DB, ['#']))
+            ),
+
+            %% 2. Create a transaction that deletes one of the topics:
+            {ok, Tx2} = emqx_ds:new_kv_tx(DB, #{
+                generation => 1,
+                shard => {auto, Owner},
+                timeout => infinity
+            }),
+            Ops2 = #{
+                ?ds_tx_delete_topic => [[<<"t">>, <<"2">>]]
+            },
+            ?assertMatch({ok, _Serial}, do_commit_tx(DB, Tx2, Ops2)),
+            %% Verify that data in t/2 is gone:
+            ?assertMatch(
+                [
+                    {[<<"foo">>], <<"payload0">>},
+                    {[<<"t">>, <<"1">>], <<"payload1">>}
+                ],
+                lists:sort(emqx_ds:dirty_read(DB, ['#']))
+            )
+        end,
+        []
+    ).
+
+%% Verify correctness of wildcard topic deletions in the transactional
+%% API
+t_14_kv_wildcard_deletes(Config) ->
+    DB = ?FUNCTION_NAME,
+    TXOpts = #{shard => {auto, <<"me">>}, timeout => infinity},
+    ?check_trace(
+        begin
+            %% Open the database
+            Opts = maps:merge(opts(Config), #{
+                store_kv => true,
+                storage => {emqx_ds_storage_kv_skipstream_lts, #{}}
+            }),
+            ?assertMatch(
+                ok,
+                emqx_ds_open_db(DB, Opts)
+            ),
+            %% 1. Insert test data:
+            {ok, Tx1} = emqx_ds:new_kv_tx(DB, TXOpts),
+            Msgs0 =
+                [{[<<"foo">>], <<"payload">>}] ++
+                    [
+                        {[<<"t">>, <<I:16>>, <<J:16>>], <<I:16, J:16>>}
+                     || I <- lists:seq(1, 10),
+                        J <- lists:seq(1, 10)
+                    ],
+            Ops1 = #{?ds_tx_write => Msgs0},
+            ?assertMatch({ok, _}, do_commit_tx(DB, Tx1, Ops1)),
+            %% Verify that all data has been inserted:
+            ?assertEqual(
+                100,
+                emqx_ds:fold_topic(
+                    fun(_, _, _, _, Acc) -> Acc + 1 end,
+                    0,
+                    [<<"t">>, '+', '+'],
+                    #{db => DB}
+                )
+            ),
+            %% 2. Issue a new transaction that deletes t/10/+:
+            {ok, Tx2} = emqx_ds:new_kv_tx(DB, TXOpts),
+            Ops2 = #{
+                ?ds_tx_delete_topic => [[<<"t">>, <<10:16>>, '+']]
+            },
+            ?assertMatch({ok, _}, do_commit_tx(DB, Tx2, Ops2)),
+            Msgs1 = lists:filter(
+                fun
+                    ({[<<"t">>, <<10:16>>, _], _}) ->
+                        false;
+                    (_) ->
+                        true
+                end,
+                Msgs0
+            ),
+            %% Verify side effects, t/100/+ is gone:
+            ?assertEqual(
+                lists:sort(Msgs1),
+                lists:sort(emqx_ds:dirty_read(DB, ['#']))
+            )
+        end,
+        []
+    ).
+
+%% Verify that `?ds_tx_serial' is properly substituted with the
+%% transaction serial.
+t_15_kv_write_serial(Config) ->
+    DB = ?FUNCTION_NAME,
+    TXOpts = #{shard => {auto, <<"me">>}, timeout => infinity},
+    Topic = [<<"foo">>],
+    ?check_trace(
+        begin
+            %% Open the database
+            Opts = maps:merge(opts(Config), #{
+                store_kv => true,
+                storage => {emqx_ds_storage_kv_skipstream_lts, #{}}
+            }),
+            ?assertMatch(
+                ok,
+                emqx_ds_open_db(DB, Opts)
+            ),
+            %% 1. Insert test data:
+            {ok, Tx1} = emqx_ds:new_kv_tx(DB, TXOpts),
+            Ops = #{
+                ?ds_tx_write => [{Topic, ?ds_tx_serial}]
+            },
+            {ok, Serial1} = do_commit_tx(DB, Tx1, Ops),
+            %% Verify that foo is set to Serial:
+            ?assertEqual(
+                [{Topic, Serial1}],
+                emqx_ds:dirty_read(DB, Topic)
+            ),
+            %% Repeat the procedure and make sure serial increases
+            {ok, Tx2} = emqx_ds:new_kv_tx(DB, TXOpts),
+            {ok, Serial2} = do_commit_tx(DB, Tx2, Ops),
+            ?assertEqual(
+                [{Topic, Serial2}],
+                emqx_ds:dirty_read(DB, Topic)
+            ),
+            ?assert(Serial2 > Serial1, [Serial2, Serial1])
+        end,
+        []
+    ).
+
+t_16_kv_preconditions(Config) ->
+    DB = ?FUNCTION_NAME,
+    TXOpts = #{shard => {auto, <<"me">>}, timeout => infinity},
+    Topic1 = [<<"foo">>],
+    Topic2 = [<<"bar">>],
+    ?check_trace(
+        begin
+            %% Open the database
+            Opts = maps:merge(opts(Config), #{
+                store_kv => true,
+                storage => {emqx_ds_storage_kv_skipstream_lts, #{}}
+            }),
+            ?assertMatch(
+                ok,
+                emqx_ds_open_db(DB, Opts)
+            ),
+            %% 1. Insert test data:
+            {ok, Tx1} = emqx_ds:new_kv_tx(DB, TXOpts),
+            Ops1 = #{
+                ?ds_tx_write => [{Topic1, <<"payload0">>}]
+            },
+            ?assertMatch({ok, _}, do_commit_tx(DB, Tx1, Ops1)),
+
+            %% 2. Preconditions, successful.
+            {ok, Tx2} = emqx_ds:new_kv_tx(DB, TXOpts),
+            Ops2 = #{
+                ?ds_tx_expected => [
+                    {Topic1, '_'},
+                    {Topic1, <<"payload0">>}
+                ],
+                ?ds_tx_unexpected => [Topic2],
+                ?ds_tx_write => [{Topic1, <<"payload1">>}]
+            },
+            ?assertMatch(
+                {ok, _},
+                do_commit_tx(DB, Tx2, Ops2)
+            ),
+            ?assertEqual(
+                [{Topic1, <<"payload1">>}],
+                emqx_ds:dirty_read(DB, ['#'])
+            ),
+
+            %% 3. Precondiction fail, unexpected message:
+            {ok, Tx3} = emqx_ds:new_kv_tx(DB, TXOpts),
+            Ops3 = #{
+                ?ds_tx_unexpected => [Topic1],
+                ?ds_tx_write => [{Topic1, <<"fail">>}]
+            },
+            ?assertMatch(
+                ?err_unrec({precondition_failed, _}),
+                do_commit_tx(DB, Tx3, Ops3)
+            ),
+
+            %% 4 Preconditions, fail, expected message not found:
+            {ok, Tx4} = emqx_ds:new_kv_tx(DB, TXOpts),
+            Ops4 = #{
+                ?ds_tx_expected => [{Topic2, '_'}],
+                ?ds_tx_write => [{Topic2, <<"fail">>}]
+            },
+            ?assertMatch(
+                ?err_unrec({precondition_failed, _}),
+                do_commit_tx(DB, Tx4, Ops4)
+            ),
+
+            %% 5 Preconditions, fail, value is different:
+            {ok, Tx5} = emqx_ds:new_kv_tx(DB, TXOpts),
+            Ops5 = #{
+                ?ds_tx_expected => [{Topic1, <<"payload0">>}],
+                ?ds_tx_write => [{Topic1, <<"fail">>}]
+            },
+            ?assertMatch(
+                ?err_unrec(
+                    {precondition_failed, [
+                        #{expected := <<"payload0">>, got := <<"payload1">>, topic := Topic1}
+                    ]}
+                ),
+                do_commit_tx(DB, Tx5, Ops5)
+            ),
+
+            %% Verify that side effects of failed transactions weren't
+            %% applied:
+            ?assertEqual(
+                [{Topic1, <<"payload1">>}],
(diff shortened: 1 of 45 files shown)


apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -119,6 +157,10 @@
 
 -type deletion() :: {delete, message_matcher('_')}.
 
+-type kv_pair() :: {topic(), binary()}.
+
+-type kv_matcher() :: {message_key(), binary() | '_'}.
```
ieQu1, line 162: Is this used?

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -305,12 +353,97 @@
...
+
+-type transaction_opts() :: #{
+    db := db(),
+    %% Specifies shard of transaction. If set to `{auto, Term}' then
+    %% the shard is derived automatically by calling
+    %% `shard_of(DB, Term)'.
+    shard := shard() | {auto, _},
+
+    %% If not specified, the last generation will be used:
+    generation => generation(),
```
ieQu1, line 393: => to :=

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -585,6 +745,362 @@ count(DB) ->
...
+%%
+%% == Options ==
+%%
+%% - `db': name of the database
+%%
+%% - `shard': Specify the shard directly. If set to `{auto, Term}',
+%% then the shard is derived by calling `shard_of(DB, Term)'.
+%%
+%% - `generation': Specify generation for the transaction. If omitted,
+%% current generation is assumed.
```
ieQu1, line 834: no longer true

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -585,6 +745,362 @@ count(DB) ->
...
+%%
+%% - `sync': If set to `false', this function will return immediately
+%% without waiting for commit. Commit outcome will be sent as a
+%% message. `true' by default.
+%%
+%% - `timeout': sets timeout waiting for the commit.
+%%
+%%   + If set to a positive integer, function may return
```
ieQu1, line 842: formatting

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -585,6 +745,362 @@ count(DB) ->
...
+%%     `{error, unrecoverable, timeout}'.
+%%
+%% - `retries': Automatically retry the transaction if commit results
+%% in a recoverable error. This option specifies number of retries. 0
+%% by default. This option has no effect when `sync' = `false'.
+%%
+%% - `retry_interval': set pause between the retries.
```
ieQu1, line 849: unit

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -617,3 +1140,151 @@ call_if_implemented(Mod, Fun, Args, Default) ->
...
+    fun(() -> Ret),
+    transaction_opts(),
+    non_neg_integer()
+) ->
+    transaction_result(Ret).
+trans_maybe_retry(DB, Fun, Opts = #{retry_interval := RetryInterval}, Retries) ->
+    case trans_inner(DB, Fun, Opts) of
+        ?err_rec(Reason) when Retries > 0 ->
+            ?tp(
+                warning,
```
ieQu1, line 1155: lower severity

apps/emqx_durable_storage/src/emqx_ds_tx_conflict_trie.erl
```diff
@@ -0,0 +1,501 @@
...
+-spec topic_filter_to_conflict_domain(emqx_ds:topic_filter()) -> conflict_domain().
+topic_filter_to_conflict_domain(TF) ->
+    tf2cd(TF, []).
+
+%% @doc Create a new conflict tracking trie.
+%%
+%% @param MinSerial Transaction serial at the time of creation.
+%%
+%% @param RotateEvery Automatically rotate the trie when the tracked
+%% serial interval reaches this value. Atom `infinity' disable
```
ieQu1, line 61: disables

apps/emqx_durable_storage/asn.1/DSBuiltinSLSkipstreamV1.asn
```diff
@@ -0,0 +1,17 @@
+--------------------------------------------------------------------------
```
ieQu1, line 1: For future

apps/emqx_durable_storage/asn.1/DSBuiltinSLReference.asn
```diff
@@ -0,0 +1,13 @@
+--------------------------------------------------------------------------
```
ieQu1, line 1: For future

apps/emqx_durable_storage/asn.1/DSBuiltinMetadata.asn
```diff
@@ -0,0 +1,28 @@
+--------------------------------------------------------------------------
```
ieQu1, line 1: For future

apps/emqx_durable_storage/asn.1/DSBuiltinStorageLayer.asn
```diff
@@ -0,0 +1,22 @@
+--------------------------------------------------------------------------
```
ieQu1, line 1: For future

rebar.config
```diff
@@ -48,7 +48,13 @@
         emqx_exproto_v_1_connection_unary_handler_client,
         emqx_exhook_v_2_hook_provider_client,
         emqx_exhook_v_2_hook_provider_bhvr,
-        'DurableMessage'
+        'DurableMessage',
```
ieQu1, line 51: Omitted due to OTP bug: https://github.com/erlang/otp/issues/9841

apps/emqx_ds_builtin_local/src/emqx_ds_builtin_local.erl
```diff
@@ -138,6 +153,7 @@ close_db(DB) ->
 
 -spec add_generation(emqx_ds:db()) -> ok | {error, _}.
 add_generation(DB) ->
+    %% FIXME: execute this in the serializer's context
```
ieQu1, line 156: Remove this comment. Transactions no longer "guess" the generation.

apps/emqx_durable_storage/src/emqx_ds_storage_layer_kv.erl
```diff
@@ -0,0 +1,180 @@
...
+    emqx_ds:topic()
+) ->
+    {ok, binary()} | undefined.
+
+-callback next(
+    emqx_ds_storage_layer:dbshard(),
+    emqx_ds_storage_layer:generation_data(),
+    Iter,
+    pos_integer(),
+    %% FIXME: remove
```
ieQu1, line 98: Currently calls to kv layouts are still dispatched via emqx_ds_storage_layer and carry some extra parameters. Fixing it is not super trivial and not very useful, except from aesthetics perspective, so there are some useless parameters.

apps/emqx_durable_storage/src/emqx_ds_msg_serializer.erl
```diff
@@ -47,20 +47,26 @@ check_schema(v1) ->
     ok;
 check_schema(asn1) ->
     ok;
+check_schema(verbatim) ->
+    ok;
 check_schema(_) ->
     {error, "Unknown schema type"}.
 
--spec serialize(schema(), emqx_types:message()) -> binary().
+-spec serialize(schema(), tuple()) -> binary().
```
ieQu1, line 55: Type is wrong.

apps/emqx_durable_storage/src/emqx_ds_msg_serializer.erl
```diff
@@ -47,20 +47,26 @@ check_schema(v1) ->
...
 serialize(v1, Msg) ->
     serialize_v1(Msg);
 serialize(asn1, Msg) ->
-    serialize_asn1(Msg).
+    serialize_asn1(Msg);
+serialize(verbatim, Blob) ->
+    Blob.
 
--spec deserialize(schema(), binary()) -> emqx_types:message().
+-spec deserialize(schema(), binary()) -> tuple().
```
ieQu1, line 63: Return type is wrong

apps/emqx_durable_storage/mix.exs
```diff
@@ -11,10 +11,16 @@ defmodule EMQXDurableStorage.MixProject do
...
-          compile_opts: [:per, :noobj, outdir: ~c"gen_src"]}
-      ],
+      asn1_srcs: for i <- ["DurableMessage",
+                           "DSBuiltinMetadata",
+                           "DSBuiltinSLReference",
+                           "DSBuiltinSLSkipstreamV1",
+                           "DSBuiltinSLSkipstreamKV",
+                           "DSBuiltinStorageLayer"
+                          ],
+               do: %{src: "./asn.1/" ++ i ++ ".asn",
```
thalesmg, line 21:
```suggestion
               do: %{src: "./asn.1/#{i}.asn",
```

apps/emqx_ds_builtin_local/src/emqx_ds_builtin_local.erl
```diff
@@ -546,10 +614,58 @@ do_delete_next(
...
+
+otx_commit_tx_batch(DBShard, SerCtl, Serial, Batches) ->
+    case otx_get_tx_serial(DBShard) of
+        SerCtl ->
+            do_commit_tx_batches(DBShard, Serial, Batches);
+        Val ->
+            ?err_unrec({serial_mismatch, SerCtl, Val})
+    end.
+
+otx_cfg_flush_interval(_DB) ->
```
ieQu1, line 642: FIXME

apps/emqx_durable_storage/src/emqx_ds_optimistic_tx.erl
```diff
@@ -0,0 +1,472 @@
...
+    ?tp(emqx_ds_optimistic_tx_commit_begin, #{db => DB, ctx => Ctx, ops => Ops}),
+    #kv_tx_ctx{leader = Leader} = Ctx,
+    Ref = monitor(process, Leader),
+    %% Note: currently timer is not canceled when commit abort is
+    %% signalled via the monitor message (i.e. when the leader process
+    %% is down). In this case the caller _will_ receive double `DOWN'.
+    %% There's no trivial way to fix it. It's up to the client to
+    %% track the pending transactions and ignore unexpected commit
+    %% notifications.
+    TRef = emqx_ds_lib:send_after(Timeout, self(), tx_timeout_msg(Ref)),
```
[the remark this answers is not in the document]
ieQu1: `timer:apply_after` is implemented as a singleton gen_server, which, on top of that, spawns an extra worker process for each MFA it executes.
OTP efficiency guide doesn't recommend using `timer` module, and I agree. That's what makes this problem non-trivial to solve. This problem is not worth risk overloading `timer_server`.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -585,6 +741,365 @@ count(DB) ->
...
+%% Default = 1s.
+%%
+%% == Return values ==
+%%
+%% - When transaction doesn't have any side effects (writes, deletes
+%% or asserts), `{nop, Ret}' is returned regardless of other factors.
+%% `Ret' is the return value of the transaction fun.
+%%
+%% - When `timeout' option is set `async', this function returns
+%% `{async, Ref, Ret}' tuple where `Ref' is a reference. Result of the
```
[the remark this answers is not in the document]
ieQu1: `Ret` is the return value of the transaction fun. I'll update the comment.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -617,3 +1139,151 @@ call_if_implemented(Mod, Fun, Args, Default) ->
...
+is_topic_filter([]) ->
+    true;
+is_topic_filter(['#']) ->
+    true;
+is_topic_filter(['+' | Rest]) ->
+    is_topic_filter(Rest);
+is_topic_filter([B | Rest]) when is_binary(B) ->
+    is_topic_filter(Rest);
+is_topic_filter(_) ->
+    false.
```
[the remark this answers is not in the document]
ieQu1: Some day I will vanquish '' from EMQX code entirely; there's absolutely no reason not to use <<>>. 
In this particular instance (key-value storage) there's no need to keep the compatibility with the common topic parser, since "topics" in this API are not following MQTT standard: they are arbitrary binaries that may not be proper unicode, or may contain slashes.
I'll add a comment.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -378,14 +521,25 @@ register_backend(Name, Module) ->
 %% @doc Different DBs are completely independent from each other. They
 %% could represent something like different tenants.
 -spec open_db(db(), create_db_opts()) -> ok.
-open_db(DB, Opts = #{backend := Backend}) ->
+open_db(DB, UserOpts) ->
+    Opts = #{backend := Backend} = set_db_defaults(UserOpts),
+    %% Sanity checks:
+    case Opts of
+        #{store_ttv := true, append_only := true} ->
+            %% Key-value pairs don't have a builtin timestamp field,
```
ieQu1, line 529: This comment is no longer accurate.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -585,6 +741,365 @@ count(DB) ->
...
+%% - `retry_interval': set pause between the retries in milliseconds.
+%% Default = 1s.
+%%
+%% == Return values ==
+%%
+%% - When transaction doesn't have any side effects (writes, deletes
+%% or asserts), `{nop, Ret}' is returned regardless of other factors.
+%% `Ret' is the return value of the transaction fun.
+%%
+%% - When `timeout' option is set `async', this function returns
```
ieQu1, line 853: It's incorrect, sync is now a separate parameter

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -585,6 +741,365 @@ count(DB) ->
...
+%%
+%% @param Fun Fold function that takes the following parameters:
+%%
+%% - Slab: a tuple consisting of shard and generation
+%%
+%% - Stream: stream identifier
+%%
+%% - DSKey: message key in the stream
+%%
+%% - Payload: either `#message' record or key-value pair {K, V}
```
ieQu1, line 997: TTV

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -158,7 +160,10 @@
         %% writes are in general unsafe but require much less resources, i.e. with RocksDB
         %% non-durable (WAL-less) writes do not usually involve _any_ disk I/O.
         %% Default: `true'.
-        durable => boolean()
+        durable => boolean(),
+        %% Whether the argument is a list of cooked batches instead of
+        %% a single batch. Default: `false'.
+        list => boolean()
```
ieQu1, line 166: Remove. This is a remaining of the first prototype

apps/emqx_durable_storage/src/emqx_ds_tx_conflict_trie.erl
```diff
@@ -0,0 +1,501 @@
...
+            case lists:prefix(Prefix, Key) of
+                true ->
+                    Acc = Fun(Key, Val, Acc0),
+                    do_fold_keys_with_prefix(Fun, Prefix, It, Acc);
+                false ->
+                    Acc0
+            end
+    end.
+
+wildcard_prefix(L) ->
```
ieQu1, line 206: This name is confusing: it returns a non-wildcard prefix of a wildcard domain.

apps/emqx_durable_storage/src/emqx_ds_optimistic_tx.erl
```diff
@@ -0,0 +1,519 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+%% @doc This module serves the same role as `emqx_ds_buffer', but it's
```
ieQu1, line 5: This comment is misleading: buffer serves a different function.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -305,12 +353,93 @@
...
+
+-type transaction_opts() :: #{
+    db := db(),
+    %% Specifies shard of transaction. If set to `{auto, Term}' then
+    %% the shard is derived automatically by calling
+    %% `shard_of(DB, Term)'.
+    shard := shard() | {auto, _},
+
+    %% If not specified, the last generation will be used:
+    generation := generation(),
```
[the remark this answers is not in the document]
ieQu1: The comment is not accurate

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -617,3 +1139,151 @@ call_if_implemented(Mod, Fun, Args, Default) ->
...
+fold_streams(Fun, Acc0, [{Slab, Stream, It} | Rest], Ctx) ->
+    Acc = fold_iterator(Fun, Acc0, Slab, Stream, It, Ctx),
+    fold_streams(Fun, Acc, Rest, Ctx).
+
+-spec fold_iterator(
+    fold_fun(Acc), Acc, slab(), stream(), iterator(), fold_ctx()
+) -> Acc.
+fold_iterator(Fun, Acc0, Slab, Stream, It0, Ctx) ->
+    #fold_ctx{db = DB, batch_size = BatchSize} = Ctx,
+    case next(DB, It0, BatchSize) of
```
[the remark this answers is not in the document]
ieQu1: Fixed.

apps/emqx_ds_builtin_raft/src/emqx_ds_replication_layer.erl
```diff
@@ -575,6 +649,57 @@ shards_of_batch(DB, [Operation | Rest], Acc) ->
...
+    emqx_ds_storage_layer_ttv:prepare_tx(DBShard, Generation, SerialBin, Ops, Opts).
+
+otx_commit_tx_batch({DB, Shard}, SerCtl, Serial, Batches) ->
+    Command = #{
+        ?tag => commit_tx_batch,
+        ?prev_serial => SerCtl,
+        ?serial => Serial,
+        ?batches => Batches
+    },
+    ra_command(DB, Shard, Command, 0).
```
ieQu1, line 694: Fixme: use non-throwing version.

apps/emqx_ds_builtin_raft/src/proto/emqx_ds_otx_proto_v1.erl
```diff
@@ -0,0 +1,35 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2024-2025 EMQ Technologies Co., Ltd. All Rights Reserved.
```
ieQu1, line 2: Copyright years

apps/emqx_durable_storage/include/emqx_ds_builtin_tx.hrl
```diff
@@ -0,0 +1,25 @@
...
+%% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-ifndef(EMQX_DS_BUILTIN_TX_HRL).
+-define(EMQX_DS_BUILTIN_TX_HRL, true).
+
+-record(kv_tx_ctx, {
+    shard :: emqx_ds:shard(),
+    leader :: pid(),
+    serial :: term(),
+    %% Current generation. Used to guard against committing
```
ieQu1, line 11: Fixme: comment is wrong. There's no "current generation" in this API.

apps/emqx_durable_storage/include/emqx_ds_builtin_tx.hrl
```diff
@@ -0,0 +1,25 @@
...
+    %% transactions that span multiple generations
+    generation :: emqx_ds:generation(),
+    opts :: emqx_ds:transaction_opts()
+}).
+
+-record(ds_tx, {
+    ctx :: emqx_ds_optimistic_tx:ctx(),
+    ops :: emqx_ds:tx_ops() | emqx_ds:tx_ops(),
+    from :: pid() | reference(),
```
ieQu1, line 20: Remove reference. Aliases aren't supported.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -585,6 +748,393 @@ count(DB) ->
...
+%% in a recoverable error. This option specifies number of retries. 0
+%% by default. This option has no effect when `sync' = `false'.
+%%
+%% - `retry_interval': set pause between the retries in milliseconds.
+%% Default = 1s.
+%%
+%% == Return values ==
+%%
+%% - When transaction doesn't have any side effects (writes, deletes
+%% or asserts), `{nop, Ret}' is returned regardless of other factors.
```
ieQu1, line 858: reads too

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -585,6 +748,393 @@ count(DB) ->
...
+    end.
+
+%% @doc Add a transaction precondition that asserts presence of the
+%% given value in the topic.
+%%
+%% This operation is considered side effect.
+%%
+%% @param Topic is used the key
+%%
+%% @param Expected value (binary) or atom `_' that matches presence of
```
ieQu1, line 952: `@param Topic`

apps/emqx_durable_storage/src/emqx_ds_lib.erl
```diff
@@ -51,6 +59,69 @@ terminate(Module, Reason, Misc) when Reason =:= shutdown; Reason =:= normal ->
...
+cancel_timer(TRef, TimeoutMsg) ->
+    _ = erlang:cancel_timer(TRef),
+    receive
+        TimeoutMsg ->
+            ok
+    after 0 ->
+        ok
+    end.
+
+%% @doc A non-throwing version of ets:delete
```
ieQu1, line 80: ets:delete/1

apps/emqx_durable_storage/src/emqx_ds_msg_serializer.erl
```diff
@@ -47,20 +47,26 @@ check_schema(v1) ->
     ok;
 check_schema(asn1) ->
     ok;
+check_schema(verbatim) ->
```
ieQu1, line 50: Investigate removing msg_serializer from the TTV layouts. In the ideal world payload (de)serialization should happen on the frontend, not in the storage layout.

apps/emqx_durable_storage/src/emqx_ds_optimistic_tx.erl
```diff
@@ -0,0 +1,618 @@
...
+%% spawned on the leader node.
+%%
+%% Optimistic transaction leader tracks recent writes and verifies
+%% transactions from the clients. Transactions containing reads that
+%% may race with the recently updated topics are rejected. Valid
+%% transactions are "cooked" and added to the buffer.
+%%
+%% Buffer is periodically flushed in a single call the backend.
+%%
+%% Leader also keeps the transaction serial.
```
ieQu1, line 16:
```suggestion
%% Leader is also tasked with incrementing the transaction serial for pending transactions.
```

apps/emqx_ds_builtin_raft/src/emqx_ds_replication_layer.erl
```diff
@@ -575,6 +649,57 @@ shards_of_batch(DB, [Operation | Rest], Acc) ->
...
+
+otx_get_tx_serial(DB, Shard) ->
+    emqx_ds_storage_layer_ttv:get_read_tx_serial({DB, Shard}).
+
+otx_get_leader(DB, Shard) ->
+    emqx_ds_builtin_raft_sup:get_gvar(DB, ?gv_otx_leader_pid(Shard), undefined).
+
+otx_become_leader(DB, Shard) ->
+    Command = #{?tag => new_otx_leader, ?otx_leader_pid => self()},
+    %% FIXME: this is probably not the right way to do it:
```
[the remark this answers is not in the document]
keynslug: If the leader for a shard exists and there is a shard's Ra server on the local node, it publishes this information to `ra_leaderboard`. Then the leader becomes the first element of `emqx_ds_builtin_raft_shard:servers(DB, Shard, leader_preferred)`. In this case doing this should be good enough I think (if I correctly understand the intent, I'll study the changes in the meantime):
```
    Servers = emqx_ds_builtin_raft_shard:servers(DB, Shard, leader_preferred),
    case ra:process_command(Servers, Command, 5_000) of
```
I.e. if the leader is known it will be contacted first, otherwise command will reach a leader once it's known.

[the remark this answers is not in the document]
ieQu1: >  otherwise command will reach a leader once it's known.

This isn't desirable in this case: these commands are sent from the OTX leader; for simplicity, its commands shouldn't outlive the lifetime of the local leader.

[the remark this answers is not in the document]
keynslug: > its commands shouldn't outlive the lifetime of the local leader.

This looks achievable with `ra:pipeline_command(LocalServer, Command, CorrelationRef, normal)`. This is asynchronous API, the caller is expected to receive `{ra_event, _FromId, {rejected, {not_leader, Leader, CorrelationRef}}}` if `LocalServer` is not the leader anymore. Otherwise, if it is it will try to commit and send `{ra_event, LocalServer, {applied, [{CorrelationRef, Reply}]}}` once it's applied.

There are some rough edges in this API, at least in my opinion. For example the caller won't get anything in response if leader is not known, the command will just be silently dropped (which actually does not follow the behavior stated in the `ra:pipeline_command/4` function doc).

Also I believe it's still possible in theory for the command to be committed to the log by the local leader, but be applied _after_ another leader is elected (for whatever reason).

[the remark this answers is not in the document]
ieQu1: What about this approach https://github.com/emqx/emqx/pull/15208/commits/3bb629a8894a2dab3ee580d72d5953663a6c57e8 ?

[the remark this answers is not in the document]
keynslug: > 3bb629a8894a2dab3ee580d72d5953663a6c57e8

This will indeed only ask the leader to commit and apply a command only if there's known leader for the shard.

A couple of things to be aware of:
1. `Leader` is not guaranteed to be equal to `emqx_ds_builtin_raft_shard:local_server(DB, Shard)`. If the leader needs to be contacted only if it's local to the node, I guess one more comparison / filtering needs to be made.
2. In an unlikely scenario where the leader loses leadership exactly before the command arrives in its message queue, it will buffer the command and forward to the new leader once it's elected, and this new leader will happily commit and apply it. Note that this buffering / forwarding is part of `ra:process_command` internal logic, but not `ra:pipeline_command`. 🫠

apps/emqx_ds_backends/test/emqx_ds_backends_SUITE.erl
```diff
@@ -978,6 +978,442 @@ t_sub_catchup_unrecoverable(Config) ->
...
+                    ]}
+                ),
+                do_commit_tx(DB, Tx5, Ops5)
+            ),
+
+            %% Verify that side effects of failed transactions weren't
+            %% applied:
+            ?assertEqual(
+                [{Topic1, 0, <<"payload1">>}],
+                emqx_ds:dirty_read(DB, ['#'])
```
ieQu1, line 1247: I think it's better to use transaction read here to rule out bad data lingering in the buffer.

apps/emqx_durable_storage/src/emqx_ds_optimistic_tx.erl
```diff
@@ -0,0 +1,620 @@
...
+%% @doc This module implements a per-shard singleton process
+%% facilitating optimistic transactions. This process is dynamically
+%% spawned on the leader node.
+%%
+%% Optimistic transaction leader tracks recent writes and verifies
+%% transactions from the clients. Transactions containing reads that
+%% may race with the recently updated topics are rejected. Valid
+%% transactions are "cooked" and added to the buffer.
+%%
+%% Buffer is periodically flushed in a single call the backend.
```
ieQu1, line 14: to the backend

apps/emqx_durable_storage/src/emqx_ds_tx_conflict_trie.erl
```diff
@@ -0,0 +1,509 @@
...
+    true;
+is_dirty(
+    CD,
+    Serial,
+    #conflict_tree{trie = Trie, old_trie = OldTrie}
+) ->
+    check_dirty(CD, Serial, Trie) orelse
+        check_dirty(CD, Serial, OldTrie).
+
+%% @doc This function is used to reduce size of the trie by removing old
```
zmstone, line 115: please add some info about the practical implication to transactions after rotation happens when older conflict set is dropped.

zmstone:
Q1: what's the prefix-conflict detection for? retained message reads and writes are non-wildcards.
Q2: if a transaction tries to write `t/#`, and a lot others conflict with it (e.g. `t/1`, `t/2`, ..., `t/10000`): will `t/#` unluckily fall into a infinite retry loop ?

ieQu1:
> Q1: what's the prefix-conflict detection for? retained message reads and writes are non-wildcards. Q2: if a transaction tries to write `t/#`, and a lot others conflict with it (e.g. `t/1`, `t/2`, ..., `t/10000`): will `t/#` unluckily fall into a infinite retry loop ?

It's mainly for session storage. For example, when a client connects with CleanSession=true, the entire session state (encoded as `[<<"d">>, SessionId, '#']`) is dropped. It must create a conflict with another channel that attempts to update an existing session, so either channel's operations are rejected and session state doesn't get partially updated.

> Q2: if a transaction tries to write t/#, and a lot others conflict with it (e.g. t/1, t/2, ..., t/10000): will t/# unluckily fall into a infinite retry loop ?

Currently OTX doesn't reject transactions that only do writes. Transactions can be rejected if they contain topic deletions, preconditions or transactional reads. 

The fact that topic deletions and preconditions may create conflicts is motivated by performance (batching) and design simplification, rather than OTX logic. If need arises, this limitation can be lifted.

apps/emqx_ds_builtin_local/src/emqx_ds_builtin_local.erl
```diff
@@ -546,6 +601,50 @@ do_delete_next(
             Error
     end.
 
+%%================================================================================
+%% Optimistic TX
+%%================================================================================
+
+-define(serial_key, <<"emqx_ds_builtin_local_tx_serial">>).
```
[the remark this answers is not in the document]
ieQu1: This is a DB-global variable, meaning it leaves in the namespace shared with all other DS-related modules. Short names might cause a collision. 
It is written only when the OTX leader process flushes a batch of transactions, which is not very frequent.

apps/emqx_durable_storage/src/emqx_ds_tx_conflict_trie.erl
ieQu1: Fixed: https://github.com/emqx/emqx/pull/15208/commits/4cab80ae3e762bdce862e8b7fea8fe72b78c92a9

zmstone:
> It's mainly for session storage. For example, when a client connects with CleanSession=true, the entire session state (encoded as [<<"d">>, SessionId, '#']) is dropped. It must create a conflict with another channel that attempts to update an existing session, so either channel's operations are rejected and session state doesn't get partially updated.

Could you explain why it is `[<<"d">>, SessionId, '#']` but not `[<<"d">>, SessionId]` ?

ieQu1:
> Could you explain why it is `[<<"d">>, SessionId, '#']` but not `[<<"d">>, SessionId]`

To avoid write amplification. When we update some tiny object, like a replay position for one stream or a sequence number, we don't want to overwrite everything, like positions in all the other streams, session metadata, etc.

zmstone:
> To avoid write amplification.

This I can agree, but do we need fine-grained locks for each tiny object (which implies concurrent updates from different processes)? If not, should it be enough to lock with `[<<"d">>, SessionId]` for a tiny object update?

zmstone:
thanks, 
i am just trying to understand why the need of `#` symbol,
since it does not support `+`, `a/b/#` and `a/b` seem equivalent.

zmstone (dismissed):

ieQu1:
> i am just trying to understand why the need of # symbol,

Write to `a/b` conflicts with read of `a/b`, `a/#` and `#`. Write to `a/b/#` also conflicts with read of a/b/c/...

keynslug (commented):
👍🏼

apps/emqx_ds_builtin_raft/src/emqx_ds_replication_layer.erl
```diff
@@ -1335,6 +1558,25 @@ clientid_size(ClientID) when is_binary(ClientID) ->
...
+
+%% @doc This internal function is used by the OTX leader process to
+%% communicate with the Raft machine.
+-spec local_raft_leader(emqx_ds:db(), emqx_ds:shard()) ->
+    ra:server_id() | unknown.
+local_raft_leader(DB, Shard) ->
+    emqx_ds_builtin_raft_shard:server_info(
+        leader,
+        emqx_ds_builtin_raft_shard:local_server(DB, Shard)
+    ).
```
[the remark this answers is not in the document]
ieQu1: I don't remember exactly why I'd rolled back the changes from the previous discussion, but some higher level API was causing problems during the initialization, IIRC it was checking leader state via `shard_sentinel` process that is started later than `optimistic_tx` process.

apps/emqx_durable_storage/src/emqx_ds_optimistic_tx.erl
```diff
@@ -0,0 +1,619 @@
...
+                info;
+            false ->
+                %% Sleep some to prevent a hot restart loop
+                timer:sleep(1_000),
+                error
+        end,
+    ?tp(Level, ds_otx_terminate, #{state => State, reason => Reason}).
+
+handle_event(info, {'EXIT', _, _Reason}, _State, _Data) ->
+    {stop, shutdown};
```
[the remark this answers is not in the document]
ieQu1: `gen_statem`'s `postpone` action postpones everything, including `EXIT`s from the supervisor. We don't want that.

apps/emqx_ds_builtin_raft/src/emqx_ds_replication_layer.erl
```diff
@@ -1097,7 +1249,46 @@ apply(
...
+        ?batches := Batches,
+        ?otx_leader_pid := From
+    },
+    State = #{db_shard := DBShard, tx_serial := ExpectedSerial, otx_leader_pid := Leader}
+) ->
+    case From of
+        Leader when SerCtl =:= ExpectedSerial ->
+            ok = emqx_ds_storage_layer_ttv:commit_batch(DBShard, Batches, #{durable => false}),
+            emqx_ds_storage_layer_ttv:set_read_tx_serial(DBShard, Serial),
+            {State#{tx_serial := Serial}, ok};
```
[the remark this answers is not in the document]
ieQu1: Good catch, fixed.

keynslug (approved):

PR #15214, opened 2025-05-14
chore: remove dead code

savonarola:
Release version: 5.10

## Summary

Removed unused macros. In some cases, I left the unused occurrences to avoid confusion, e.g., to avoid breaks in named sequences, like
```erlang
-define(ONE, 1).
-define(TWO, 2).
%% What is 3 ?
-define(FOUR, 4).
```
Or when it seems that a macro can be helpful in the near future.

## PR Checklist


apps/emqx_gateway/src/emqx_gateway_cm.erl
```diff
@@ -450,7 +449,7 @@ takeover_session(GwName, ClientId) ->
             }),
             lists:foreach(
                 fun(StalePid) ->
-                    catch discard_session(GwName, ClientId, StalePid)
```
thalesmg, line 453: Nit: this may actually change the behavior, instead of just refactoring.

apps/emqx_utils/src/bpapi/emqx_bpapi_trans.erl
```diff
@@ -20,8 +20,6 @@
 
 -export([parse_transform/2, format_error/1]).
 
-%%-define(debug, true).
```
thalesmg, line 23: Maybe leave this one as a hint that it may be enabled for debugging?

savonarola: ✅

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7445)

PR #15231, opened 2025-05-19
ci: trigger update-emqx-docs workflow when docker images are ready

id:
Release version: 5.10.0

## PR Checklist

### DIFF
--- a/.github/workflows/_push-entrypoint.yaml
+++ b/.github/workflows/_push-entrypoint.yaml
@@ -255,6 +255,7 @@ jobs:
     uses: ./.github/workflows/update-emqx-docs.yaml
     with:
       tag: ${{ github.ref_name }}
+    secrets: inherit
 
   update_public_broker:
     if: needs.prepare.outputs.release == 'true'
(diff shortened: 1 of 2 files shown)


.github/workflows/_push-entrypoint.yaml
```diff
@@ -245,6 +246,16 @@ jobs:
       builder: ${{ needs.init.outputs.BUILDER }}
       ct-matrix: ${{ needs.prepare.outputs.ct-matrix }}
 
+  update-emqx-docs:
+    if: needs.prepare.outputs.release == 'true'
```
thalesmg, line 250: Should we also guard that the event type is `tag`?

.github/workflows/_push-entrypoint.yaml
```diff
@@ -245,6 +246,16 @@ jobs:
...
+    needs:
+      - init
+      - prepare
+      - build_and_push_docker_images
+    uses: ./.github/workflows/update-emqx-docs.yaml
+    with:
+      tag: ${{ github.ref_name }}
+
```
thalesmg, line 258: This may need `secrets: inherit` (not really sure)

thalesmg (approved):

.github/workflows/_push-entrypoint.yaml
id: `release` will not be set to `true` if the workflow is not triggered by tag

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7485)

PR #15238, opened 2025-05-20
fix(connectors api): focus on value when there's an update error

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14292

Release version: 5.10.0

## Summary

The problem was that, during an update, the type is an atom instead of a binary.

image

## PR Checklist

- [na] The changes are covered with new or existing tests (tested manually)
- [na] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files (unreleased feature)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7486)

PR #15258, opened 2025-05-22
Make data backup operations more robust

savonarola:
Fixes [EMQX-14264](https://emqx.atlassian.net/browse/EMQX-14264)

Release version: 5.10

The reported security issue is not confirmed, but we have added a test to verify it explicitly. `erl_tar` itself prevents from unpacking files into dirs lying upper from the current dir: https://github.com/erlang/otp/blob/OTP-27.3.4/lib/stdlib/src/erl_tar.erl#L2002

We also eliminate fuzziness in the API when the data backup file may be interpreted either as the filename in the data backup directory or as the full path to a data backup file. We provide different methods for the two cases.

## PR Checklist


apps/emqx_management/src/emqx_mgmt_data_backup.erl
```diff
@@ -356,10 +383,12 @@ format_error({already_exists, BackupFileName}) ->
 format_error(Reason) ->
     Reason.
 
+-spec format_conf_errors(map()) -> list().
 format_conf_errors(Errors) ->
     Opts = #{print_fun => fun io_lib:format/2},
     maps:values(maps:map(conf_error_formatter(Opts), Errors)).
 
+-spec format_db_errors(map()) -> list().
```
thalesmg, line 391: Nit:

```suggestion
-spec format_db_errors(map()) -> map().
```

thalesmg (approved):

thalesmg (approved):

PR #15113, opened 2025-04-24
feat(sessds): turn QoS0 subscriptions into direct broker subscriptions

keynslug:
Fixes [EMQX-9746](https://emqx.atlassian.net/browse/EMQX-9746).

Release version: 5.10.0

## Summary

1. Introduce concept of _direct_ subscriptions for DS-based persistent sessions.
2. Use `emqx_broker` mechanics for _direct_ subscriptions, similar to in-memory sessions.
3. Avoid setting persistent routes for direct subscriptions.
4. Turn QoS0 subscriptions into direct subscriptions.

## PR Checklist
- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible

[EMQX-9746]: https://emqx.atlassian.net/browse/EMQX-9746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #15261, opened 2025-05-23
feat(sessds): turn QoS0 subscriptions into direct broker subscriptions

keynslug:
Fixes [EMQX-9746](https://emqx.atlassian.net/browse/EMQX-9746).

See also #15113.

Release version: 5.10?

## Summary

This PR introduces _direct_ subscriptions for DS-based persistent sessions. If the client subscribes to a topic with QoS level 0, this subscriptions is treated as _direct_ subscription.

Direct subscriptions:
* Use `emqx_broker` mechanics to work, similar to in-memory sessions.
* Do not store routes in the persistent session router, thus avoiding persistence of matching messages in DS (only if there's no durable subscriptions to the same or encompassing topic though).

Upgrades (turning direct subscription into durable) and downgrades (turning durable subscription into direct) are handled, but slightly differently. Downgrades are processed only during session replays, meaning they are effectively postponed until next client reconnect.

## PR Checklist

[EMQX-9746]: https://emqx.atlassian.net/browse/EMQX-9746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
@@ -125,7 +125,8 @@
 -ifdef(TEST).
 -export([
     runtime_state_invariants/2,
-    offline_state_invariants/2
+    offline_state_invariants/2,
+    assert_runtime_durable_subscription/3
 ]).
 -endif.
 
@@ -1204,9 +1205,8 @@ bq_push(Primary, B = #block{}, BQ) ->
     emqx_persistent_session_ds_fuzzer:model_state(), #{s := map(), scheduler_state_s := t()}
 ) ->
     boolean().
-runtime_state_invariants(ModelState, #{s := S, stream_scheduler_s := SchedS}) ->
-    invariant_new_stream_subscriptions(ModelState, SchedS) and
-        invariant_active_streams_y_ranks(S) and
+runtime_state_invariants(_ModelState, #{s := S, stream_scheduler_s := SchedS}) ->
+    invariant_active_streams_y_ranks(S) and
         invariant_ds_subscriptions(S, SchedS).
 
 -spec offline_state_invariants(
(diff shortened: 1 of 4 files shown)


apps/emqx_utils/src/emqx_utils_maps.erl
```diff
@@ -367,6 +368,13 @@ get_lazy(Key, Map, DefFn) ->
         error -> DefFn()
     end.
 
+%% @doc Group list of maps by value associated with a given key.
+%% Groups appear in the order of the value of the grouping key.
+-spec group_by(K, [#{K := V}]) -> [#{K := V}].
+group_by(K, Maps) ->
+    Values = lists:usort([maps:get(K, M) || M <- Maps]),
+    [M || V <- Values, M <- Maps, map_get(K, M) =:= V].
```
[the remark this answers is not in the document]
keynslug: Good point. Moved back to test code in [ed5b648](https://github.com/emqx/emqx/pull/15261/commits/ed5b6483f89a35e11dcf8bef8e96c909bcd9e25d).

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_subs.erl
```diff
@@ -169,30 +260,90 @@ on_unsubscribe(SessionId, TopicFilter, S0) ->
...
+            {{downgrade, DesiredMode, TopicFilter, Sub0}, S};
+        direct ->
+            %% Restore the subscription in the broker:
+            ok = add_route(direct, SessionId, TopicFilter, SubOpts);
+        durable ->
+            %% Nothing to do:
+            ok
+    end.
+
+downgrade_subscription(TopicFilter, SState, SessionId, S0) ->
```
ieQu1, line 321: `downgrade_subscription` -> `change_to_direct`

Terms `upgrade`/`downgrade` are ambiguous, and may collide with possible future usecases where the state should be updated according to the code change.

keynslug: Changed in [5c22ce5](https://github.com/emqx/emqx/pull/15261/commits/5c22ce53578c0f52e7e9c595776816d8f36c9781).

apps/emqx_durable_storage/test/emqx_ds_test_helpers.erl
```diff
@@ -175,6 +175,14 @@ message_diff_options(Fields) ->
         compare_fun => fun(M1, M2) -> message_eq(Fields, M1, M2) end
     }.
 
+%% @doc Group list of maps by value associated with a given key.
+%% Groups appear in the order of the value of the grouping key, relative order
+%% inside each group is preserved.
+-spec group_by(K, [#{K := V}]) -> [#{K := V}].
```
ieQu1, line 181: group_maps_by

apps/emqx/test/emqx_persistent_session_ds_fuzzer.erl
```diff
@@ -607,7 +607,11 @@ next_state(S, Ret, {call, ?MODULE, connect, [_, Opts]}) ->
     S#{
         conninfo := Ret,
         connected := true,
-        conn_opts := Opts
+        conn_opts := Opts,
+        subs := maps:map(
```
ieQu1, line 611: I don't get what is this code doing

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_stream_scheduler.erl
```diff
@@ -1210,23 +1216,29 @@ runtime_state_invariants(ModelState, #{s := S, stream_scheduler_s := SchedS}) ->
...
+%% Verify that each active durable topic subscription has a new stream watch:
 invariant_new_stream_subscriptions(#{subs := Subs}, #s{
     new_stream_subs = Watches, sub_metadata = SubStates
 }) ->
-    ExpectedTopics = lists:sort(maps:keys(Subs)),
+    ExpectedTopics = lists:sort(
+        [
+            Topic
+         || {Topic, SubOpts = #{qos := QoS}} <- maps:to_list(Subs),
+            QoS > ?QOS_0 orelse maps:get(durable, SubOpts, false)
```
ieQu1, line 1227: I think this invariant can be changed to "for each model subscription there should be *either* a direct or durable broker subscription". This should eliminate the need for `durable` flag in the model.

PR #15271, opened 2025-05-27
chore(deps): pin unicode_util_compat

savonarola:
Pin dependency to avoid conflict


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7586)

PR #15284, opened 2025-05-28
fix(new mix build): fix `emqx_durable_storage/mix.exs`

thalesmg:
This affects only the new mix build.

Set of asn1 source names changed.


thalesmg:
@keynslug cherry-picked the commit from your branch, thank you.

PR #15309, opened 2025-05-30
fix(new mix build): add missing `mix.exs` for `emqx_gateway_nats`

thalesmg:
Only affects new mix build.


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7653)

PR #15311, opened 2025-05-31
refactor: completely random jwk to sign login token

zmstone:
Release version: 5.10.1

## Summary

There is no reason to hash anything for the one-time jwk.
The full token is stored in DB for validation later, never needed any secret.

## PR Checklist

- [~] Change log: no user experience change


apps/emqx_dashboard/src/emqx_dashboard_token.erl
```diff
@@ -89,12 +85,11 @@ create_tables() ->
...
-%%--------------------------------------------------------------------
-%% jwt apply
-do_sign(#?ADMIN{username = Username} = User, Password) ->
+-spec sign(User :: dashboard_user()) ->
+    {ok, dashboard_user_role(), Token :: binary()}.
+sign(#?ADMIN{username = Username} = User) ->
     ExpTime = jwt_expiration_time(),
-    Salt = salt(),
-    JWK = jwk(Username, Password, Salt),
+    JWK = jwk(),
```
savonarola, line 92: Curious: why do we create a token as JWT if we cannot verify it?

zmstone: i guess it's just to take advantage of the JSON nature of the token, so to retrieve the expire time.

for frontend and scripts too, client can know when to refresh a token when it's close to expire.

savonarola (dismissed):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7761)

PR #15313, opened 2025-06-02
ci: github event payload is passed as is on workflow_call

id:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8131)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8131)

PR #15318, opened 2025-06-02
refactor(dashboard): simplify dispatch management

savonarola:
Fixes [EMQX-14339](https://emqx.atlassian.net/browse/EMQX-14339)

Release version: 5.10.x

## Summary

The main purpose of the PR is to get rid of the quite complex logic of caching the cowboy dispatch term in a file.
We also make some fixes and some refactorings to simplify module interactions (see comments and individual commits).

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files — **invisibe except in some extreme cases**

[EMQX-14339]: https://emqx.atlassian.net/browse/EMQX-14339?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard.erl
@@ -10,8 +10,8 @@
     stop_listeners/1,
     stop_listeners/0,
     list_listeners/0,
-    update_dispatch/0,
-    listeners_status/0
+    listeners_status/0,
+    regenerate_dispatch_after_config_update/0
 ]).
 
 %% Authorization
@@ -21,8 +21,6 @@
 -include_lib("emqx/include/http_api.hrl").
 -include_lib("emqx/include/emqx_release.hrl").
 -include_lib("snabbkaffe/include/snabbkaffe.hrl").
-
--dialyzer({[no_opaque, no_match, no_return], [start_listeners/1]}).
 
 -define(EMQX_MIDDLE, emqx_dashboard_middleware).
 
@@ -46,7 +44,7 @@ stop_listeners() ->
 -spec start_listeners(listener_configs()) ->
     ok | {error, [listener_name()]}.
 start_listeners(Listeners) ->
-    {ok, _} = application:ensure_all_started(minirest),
+    %% NOTE
     %% Before starting the listeners, we do not generate the full dispatch upfront,
     %% because the listeners may fail to start and the generation may appear useless.
     InitDispatch = init_dispatch(),
@@ -67,6 +65,7 @@ start_listeners(Listeners) ->
                         ]),
                         {[Name | OkAcc], ErrAcc};
                     {error, _Reason} ->
+                        %% NOTE
                         %% Don't record the reason because minirest already does(too much logs noise).
                         {OkAcc, [Name | ErrAcc]}
                 end
@@ -74,7 +73,7 @@ start_listeners(Listeners) ->
             {[], []},
             listeners(ensure_ssl_cert(Listeners))
         ),
-    ok = update_dispatch(OkListeners),
+    ok = emqx_dashboard_dispatch:regenerate_dispatch(OkListeners),
     case ErrListeners of
         [] ->
             ok;
@@ -98,12 +97,6 @@ stop_listeners(Listeners) ->
         listeners(Listeners)
     ).
 
--spec update_dispatch() -> {ok, [listener_name()]}.
-update_dispatch() ->
-    #{started := Started} = listeners_status(),
-    ok = update_dispatch(Started),
-    {ok, Started}.
-
 -spec listeners_status() -> #{started := [listener_name()], stopped := [listener_name()]}.
 listeners_status() ->
     ListenerNames = [
@@ -112,6 +105,11 @@ listeners_status() ->
     ],
     {Started, Stopped} = lists:partition(fun is_listener_started/1, ListenerNames),
     #{started => Started, stopped => Stopped}.
+
+-spec regenerate_dispatch_after_config_update() -> ok.
+regenerate_dispatch_after_config_update() ->
+    #{started := Listeners} = listeners_status(),
+    ok = emqx_dashboard_dispatch:regenerate_dispatch_after_config_update(Listeners).
 
 %%--------------------------------------------------------------------
 %% internal
@@ -123,22 +121,6 @@ is_listener_started(Name) ->
     catch
         error:badarg -> false
     end.
-
-update_dispatch(ListenerNames) ->
-    {Time, ok} = timer:tc(fun() ->
-        lists:foreach(
-            fun(Name) ->
-                ok = minirest:update_dispatch(Name)
-            end,
-            ListenerNames
-        )
-    end),
-    ?tp(error, regenerate_dispatch, #{
-        elapsed => erlang:convert_time_unit(Time, microsecond, millisecond),
-        i18n_lang => emqx:get_config([dashboard, i18n_lang]),
-        listeners => ListenerNames
-    }),
-    ok.
 
 init_dispatch() ->
     static_dispatch() ++ dynamic_dispatch().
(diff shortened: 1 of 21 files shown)


apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -58,21 +65,66 @@ start_listeners(Listeners) ->
...
                         {OkAcc, [Name | ErrAcc]}
                 end
             end,
             {[], []},
             listeners(ensure_ssl_cert(Listeners))
         ),
+    ok = emqx_dashboard_dispatch:regenerate_dispatch(OkListeners),
     case ErrListeners of
         [] ->
-            optvar:set(emqx_dashboard_listeners_ready, OkListeners),
```
savonarola, line 70: This didn't seem to be working correctly.
The function is parametrized with `Listeners`, but the variable is not.
So, it looks like a channel into the `emqx_dashboard_listener` module for passing the latest started listener names to regenerate the dispatch on application start.

The problem is that its value was also used to generate dispatches after config updates.
So in the scenario:
* We disable some listeners (e.g., `http`)
* We enable it, and it is passed to the `start_listeners` fun.
* `emqx_dashboard_listeners_ready` gets set to `['http:dashboard']`
* In subsequent config updates, only `http` gets regenerated.

We try to eliminate the mutual coupling of `emqx_dashboard` and `emqx_dashboard_listener(_config)`. Now, `emqx_dashboard` is self-contained and does not need `emqx_dashboard_listener(_config)` to be initialized properly.

apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -58,21 +65,66 @@ start_listeners(Listeners) ->
...
                     {error, _Reason} ->
+                        %% NOTE
                         %% Don't record the reason because minirest already does(too much logs noise).
                         {OkAcc, [Name | ErrAcc]}
                 end
             end,
             {[], []},
             listeners(ensure_ssl_cert(Listeners))
         ),
+    ok = emqx_dashboard_dispatch:regenerate_dispatch(OkListeners),
```
savonarola, line 76: Even if only some listeners started, we do not want have them unconfigured.

apps/emqx_dashboard/src/emqx_dashboard_listener_config.erl
```diff
@@ -134,26 +131,6 @@ remove_sensitive_data(Conf0) ->
...
-    NewHttps = get_listener(https, NewConf),
-    {StopHttp, StartHttp} = diff_listeners(http, OldHttp, NewHttp, SwaggerSupport),
-    {StopHttps, StartHttps} = diff_listeners(https, OldHttps, NewHttps, SwaggerSupport),
-    Stop = maps:merge(StopHttp, StopHttps),
-    Start = maps:merge(StartHttp, StartHttps),
-    delay_job({update_listeners, Stop, Start}).
-
-%% in post_config_update, the config is not yet persisted to persistent_term
-%% so we need to delegate the listener update to the gen_server a bit later
-delay_job(Msg) ->
```
savonarola, line 153: This looks quite hairy (what if config update lasts longer?), we try to make this a bit more accurate (see `emqx_dashboard_dispatch:regenerate_dispatch_after_config_update/1`)

apps/emqx_dashboard/src/emqx_dashboard_sup.erl
```diff
@@ -20,7 +20,8 @@ init([]) ->
     ok = emqx_dashboard_desc_cache:init(),
     {ok,
         {{one_for_one, 5, 100}, [
-            ?CHILD(emqx_dashboard_listener, brutal_kill),
```
savonarola, line 23: `emqx_dashboard_listener(_config)` does not become busy for several seconds anymore

apps/emqx_dashboard/test/emqx_dashboard_https_SUITE.erl
```diff
@@ -129,69 +149,103 @@ t_compatibility_ssl_cert(_Config) ->
...
+        <<"verify_peer">>
     ),
-    emqx_common_test_helpers:load_config(emqx_dashboard_schema, VerifyPeerConf1),
+    {ok, _} = emqx:update_config([<<"dashboard">>], maps:get(<<"dashboard">>, VerifyPeerConf1)),
+    ok = emqx_dashboard:stop_listeners(),
     ?assertMatch({error, [?NAME]}, emqx_dashboard:start_listeners()),
+
     %% verify_peer and cacertfile is ok.
     VerifyPeerConf2 = emqx_utils_maps:deep_put(
-        [dashboard, listeners, https, cacertfile],
```
savonarola, line 185: This test did not seem to be working as expected, `cacertfile` should be inside `ssl_options` map:  `[dashboard, listeners, https, ssl_options, cacertfile]` 🤔

apps/emqx_dashboard/test/emqx_dashboard_https_SUITE.erl
savonarola, line 1: We refactor the suite to use new `emqx_cth_suite` helpers.

apps/emqx_dashboard/src/emqx_dashboard_dispatch.erl
```diff
@@ -0,0 +1,119 @@
...
+do_regenerate_dispatch(ListenerNames) ->
+    {Time, ok} = timer:tc(fun() ->
+        lists:foreach(
+            fun(Name) ->
+                ok = minirest:update_dispatch(Name)
+            end,
+            ListenerNames
+        )
+    end),
+    ?tp(error, regenerate_dispatch, #{
```
keynslug: Q: Why error?

apps/emqx_dashboard/src/emqx_dashboard_dispatch.erl
```diff
@@ -0,0 +1,119 @@
...
+        elapsed => erlang:convert_time_unit(Time, microsecond, millisecond),
```
keynslug: Nit. `elapsed_ms`?

apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -10,43 +10,50 @@
...
+-spec start_listeners(listener_configs()) ->
+    ok | {error, [listener_name()]}.
 start_listeners(Listeners) ->
-    {ok, _} = application:ensure_all_started(minirest),
-    SwaggerSupport = emqx:get_config([dashboard, swagger_support], true),
-    InitDispatch = dispatch(),
+    %% NOTE
+    %% Before starting the listeners, we do not generate the full dispatch upfront,
+    %% because the listeners may fail to start and the generation may appear useless.
+    InitDispatch = init_dispatch(),
```
keynslug, line 50: Just to discuss. As an occasional EMQX user / operator / integration tester, I'd prefer EMQX Dashboard listener available = EMQX API usable. As far as I understand, this is not exactly the case right now because of this short phase where the listener starts with not-yet-initialized dispatch. Now that this "regeneration" is much more effective, this sort of optimization also feels a bit unnecessary, especially since we don't really expect listeners to fail to start.

keynslug (commented):
🚀

apps/emqx_dashboard/src/emqx_dashboard.erl
savonarola: I mostly documented the original intentions, and had similar thoughts while doing that 😅 

The thing is, the `minirest` API itself also separates starting a dummy listener and generating the real dispatch further.

I would like to address both `minirest` and the starting process in a separate PR.

zmstone (dismissed):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7764)

PR #15320, opened 2025-06-03
ci: remove actions/prepare-jmeter from dependabot config

id:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7691)

PR #15361, opened 2025-06-10
fix(frame): handle malformed v5 user-property length for pair

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14323

Release version: 5.10.1

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_frame.erl
+++ b/apps/emqx/src/emqx_frame.erl
@@ -692,13 +692,6 @@ parse_utf8_pair(<<LenK:16/big, Key:LenK/binary, Rest/binary>>, _StrictMode) ->
         parsed_key => Key,
         remaining_bytes_length => byte_size(Rest)
     });
-parse_utf8_pair(Bin, _StrictMode) when
-    4 > byte_size(Bin)
-->
-    ?PARSE_ERR(#{
-        cause => user_property_not_enough_bytes,
-        total_bytes => byte_size(Bin)
-    });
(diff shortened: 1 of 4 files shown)


apps/emqx/src/emqx_frame.erl
```diff
@@ -685,12 +685,24 @@ parse_utf8_pair(<<LenK:16/big, _Key:LenK/binary, LenV:16/big, Rest/binary>>, _St
...
+        cause => malformed_user_property_missing_value,
+        parsed_key_length => LenK,
+        parsed_key => Key,
+        remaining_bytes_length => byte_size(Rest)
+    });
 parse_utf8_pair(Bin, _StrictMode) when
     4 > byte_size(Bin)
 ->
     ?PARSE_ERR(#{
         cause => user_property_not_enough_bytes,
```
keynslug: Q: What do you think about just dropping this weird guard, replacing the cause with `malformed_user_property` and calling it a day? Introducing two extra clauses feels a little redundant to me here.

thalesmg: I was going for a more "backwards compatible" change (in the sense of preserving existing errors), but I agree that this one is particularly weird, specially since there was no catch-all clause before this fix.  :thinking:

zmstone (approved):

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-7765)

PR #15377, opened 2025-06-12
feat(config handler): allow saving namespaced configs in mnesia

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14389

Release version: 6.0.0

## Summary

In order to support per-namespace user roles with scoped resources and permissions, we'll need to store namespaced configurations in mnesia.  Mnesia was chosen because we want to avoid generating large or multiple `cluster.hocon` files as the number of namespaces and their configs might be large.

Here, in order to persist the config changes in mnesia, one calls `emqx:update_config(KeyPath, Req, #{namespace => Namespace, ...}).`, where `Namespace` is the binary name for the namespace.  This config will use the existing `emqx_config_handler` machinery (i.e., type check, call pre/post config update hooks), but read and write configs to mnesia instead of persistent term and/or `cluster.hocon`.

Later, for this milestone, we'll want tenant admin users to have their independent, namespaced configurations with resouces that are independent (i.e., they may have connectors/actions/sources/rules with the same names without clashes).  The `{pre,post}_config_update` hooks for such config root keys will use the newly introduced arity that passes down the extra context (namespace) so they may act in the appropriate configuration.

## PR Checklist

- [na] ~~Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files~~ (this is an internal change that should not be externally visible right now) 

### DIFF
--- a/apps/emqx/include/emqx.hrl
+++ b/apps/emqx/include/emqx.hrl
@@ -99,15 +99,6 @@
 -define(KIND_REPLICATE, replicate).
 -define(KIND_INITIATE, initiate).
 
--define(CONFIG_TAB, emqx_config).
-
--record(?CONFIG_TAB, {
-    %% {Namespace, RootKey}
-    root_key,
-    raw_value,
-    extra = #{}
-}).
-
 %%--------------------------------------------------------------------
 %% Client Attributes
 %%--------------------------------------------------------------------
--- a/apps/emqx/include/emqx.hrl
+++ b/apps/emqx/include/emqx.hrl
@@ -99,6 +99,16 @@
 -define(KIND_REPLICATE, replicate).
 -define(KIND_INITIATE, initiate).
 
+-define(CONFIG_TAB, emqx_config).
+
+-record(?CONFIG_TAB, {
+    %% {Namespace, RootKey}
+    root_key,
+    raw_value,
+    checked_value,
+    extra = #{}
+}).
+
 %%--------------------------------------------------------------------
 %% Client Attributes
 %%--------------------------------------------------------------------
--- a/apps/emqx/include/emqx.hrl
+++ b/apps/emqx/include/emqx.hrl
@@ -105,7 +105,6 @@
     %% {Namespace, RootKey}
     root_key,
     raw_value,
-    checked_value,
     extra = #{}
 }).
 
--- a/apps/emqx/src/emqx_config.erl
+++ b/apps/emqx/src/emqx_config.erl
@@ -593,9 +593,10 @@ clear_invalid_namespaced_config(Namespace, RootKeyBin) when is_binary(Namespace)
             case map_size(Errors) == 0 of
                 true ->
                     persistent_term:erase(?INVALID_NS_CONF_PT_KEY(Namespace)),
-                    emqx_corrupt_namespace_config_checker:check(Namespace);
+                    emqx_corrupt_namespace_config_checker:clear(Namespace);
                 false ->
-                    persistent_term:put(?INVALID_NS_CONF_PT_KEY(Namespace), Errors)
+                    persistent_term:put(?INVALID_NS_CONF_PT_KEY(Namespace), Errors),
+                    emqx_corrupt_namespace_config_checker:check(Namespace)
             end
     end.
 
--- a/apps/emqx/src/emqx_config.erl
+++ b/apps/emqx/src/emqx_config.erl
@@ -22,7 +22,6 @@
     fill_defaults/2,
     fill_defaults/3,
     save_configs/5,
-    save_configs_mnesia/5,
     save_to_app_env/1,
     save_to_config_map/2,
     save_to_override_conf/3,
@@ -43,8 +42,6 @@
 -export([
     get/1,
     get/2,
-    get_mnesia/2,
-    get_mnesia/3,
     find/1,
     find_raw/1,
     put/1,
@@ -57,8 +54,6 @@
 -export([
     get_raw/1,
     get_raw/2,
-    get_raw_mnesia/2,
-    get_raw_mnesia/3,
     put_raw/1,
     put_raw/2
 ]).
@@ -92,7 +87,17 @@
 -export([ensure_atom_conf_path/2]).
 -export([load_config_files/2]).
 -export([upgrade_raw_conf/2]).
--export([seed_defaults_for_all_roots_mnesia/2, erase_namespace_configs_mnesia/1]).
+
+%% Namespaced configs
+-export([seed_defaults_for_all_roots_namespaced/2, erase_namespaced_configs/1]).
+-export([
+    get_namespaced/2,
+    get_namespaced/3,
+    get_raw_namespaced/2,
+    get_raw_namespaced/3,
+    save_configs_namespaced/5,
+    save_configs_namespaced_tx/5
+]).
 
 -ifdef(TEST).
 -export([erase_all/0, backup_and_write/2, cluster_hocon_file/0, base_hocon_file/0]).
@@ -103,8 +108,8 @@
 
 -define(CONF, conf).
 -define(RAW_CONF, raw_conf).
+-define(NS_CONF(NS), {ns_conf, NS}).
 -define(PERSIS_SCHEMA_MODS, {?MODULE, schema_mods}).
--define(PERSIS_SCHEMA_MODS_MNESIA, {?MODULE, schema_mods, mnesia}).
 -define(PERSIS_KEY(TYPE, ROOT), {?MODULE, TYPE, ROOT}).
 -define(ZONE_CONF_PATH(ZONE, PATH), [zones, ZONE | PATH]).
 -define(LISTENER_CONF_PATH(TYPE, LISTENER, PATH), [listeners, TYPE, LISTENER | PATH]).
@@ -175,7 +180,7 @@ get_root([RootName | _]) ->
     #{RootName => do_get(?CONF, [RootName], #{})}.
 
 get_root_mnesia([RootName | _], Namespace) ->
-    #{bin(RootName) => do_get_mnesia([RootName], Namespace, {value, #{}})}.
+    #{bin(RootName) => do_get(?NS_CONF(Namespace), [RootName], #{})}.
 
 %% @doc For the given path, get raw root value enclosed in a single-key map.
 %% key is ensured to be binary.
@@ -183,26 +188,9 @@ get_root_raw([RootName | _]) ->
     #{bin(RootName) => get_raw([RootName], #{})}.
 
 get_root_raw_mnesia([RootName | _], Namespace) ->
-    #{bin(RootName) => do_get_raw_mnesia([RootName], Namespace, {value, #{}})}.
+    #{bin(RootName) => do_get_raw_namespaced([RootName], Namespace, {value, #{}})}.
 
-do_get_mnesia([Root | KeyRest] = KeyPath, Namespace, DefaultAction) ->
-    RootBin = bin(Root),
-    case mnesia:dirty_read(?CONFIG_TAB, {Namespace, RootBin}) of
-        [#?CONFIG_TAB{checked_value = #{} = RawConfig}] when DefaultAction == error ->
-            deep_get_mnesia(Root, KeyRest, RawConfig);
-        [#?CONFIG_TAB{checked_value = #{} = RawConfig}] ->
-            {value, Default} = DefaultAction,
-            emqx_utils_maps:deep_get(KeyRest, RawConfig, Default);
-        _ ->
-            case DefaultAction of
-                error ->
-                    error({config_not_found, KeyPath});
-                {value, Default} ->
-                    Default
-            end
-    end.
-
-do_get_raw_mnesia([Root | KeyRest] = KeyPath, Namespace, DefaultAction) ->
+do_get_raw_namespaced([Root | KeyRest] = KeyPath, Namespace, DefaultAction) ->
     RootBin = bin(Root),
     case mnesia:dirty_read(?CONFIG_TAB, {Namespace, RootBin}) of
         [#?CONFIG_TAB{raw_value = #{} = RawConfig}] when DefaultAction == error ->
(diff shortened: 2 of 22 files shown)


thalesmg:
- [x] store checked config in persistent term instead of mnesia (functions may be stored in checked config)

apps/emqx/include/emqx.hrl
```diff
@@ -99,6 +99,15 @@
...
 
+-define(CONFIG_TAB, emqx_config).
+
+-record(?CONFIG_TAB, {
+    %% {Namespace, RootKey}
+    root_key,
+    raw_value,
+    extra = #{}
+}).
+
```
ieQu1: emqx.hrl is included in almost every module. Are these macros so important that we need them everywhere? 
Also, CONFIG_TAB is a rather common name, it may collide with others.
I suggest putting this in a different header.

thalesmg:
- [x] test restarting a node and verifying that it loads checked configs correctly
- [x] don't execute transaction when replicating cluster rpc entries

apps/emqx/src/emqx_config.erl
```diff
@@ -145,16 +173,60 @@
...
+                error ->
+                    error({config_not_found, KeyPath});
+                {value, Default} ->
+                    Default
+            end
+    end.
+
+deep_get_namespaced(Root, KeyPath, Config) ->
+    try
+        emqx_utils_maps:deep_get(KeyPath, Config)
```
savonarola, line 228: not directly related to the PR, but it's surprising that a generic module `emqx_utils_maps`  raises `config_not_found`

apps/emqx/src/emqx_config.erl
```diff
@@ -302,6 +388,24 @@ put_raw(KeyPath0, Config) ->
...
+        Config
+    ).
+
+-spec put_namespaced(binary(), emqx_utils_maps:config_key_path(), term()) -> ok.
+put_namespaced(Namespace, KeyPath, Config) when is_binary(Namespace) ->
+    [_RootName | KeyPathRest] = KeyPath,
+    Putter = fun(_Path, RootValue, Value) ->
+        emqx_utils_maps:deep_put(KeyPathRest, RootValue, Value)
+    end,
+    do_put(?NS_CONF(Namespace), Putter, KeyPath, Config).
```
savonarola, line 411: Do we need to update `put_config_post_change_actions`?

apps/emqx/src/emqx_config_handler.erl
```diff
@@ -409,9 +450,18 @@ apply_pre_config_update(Module, Callback, 4, #{
...
+    Module:Callback(ConfKeyPath, UpdateReq, OldRawConf, ClusterRPCOpts);
+apply_pre_config_update(Module, Callback, 5, #{
+    conf_key_path := ConfKeyPath,
+    update_req := UpdateReq,
+    namespace := Namespace,
+    old_raw_conf := OldRawConf,
+    cluster_rpc_opts := ClusterRPCOpts
+}) ->
+    ExtraContext = #{namespace => Namespace},
+    Module:Callback(ConfKeyPath, UpdateReq, OldRawConf, ClusterRPCOpts, ExtraContext);
```
savonarola: As far as I understand, we introduce a new callback here (of arity 5), not to clash with existing callbacks, because the existing handlers do not expect to receive namespace config updates.

We probably need to add the new `@callback`s to the behaviour.

But currently, there are very few implementations of the arity 4 callback, i.e. few handlers respect `ClusterRPCOpts`.

So instead of adding a new callback I would prefer to
* merge `ClusterRPCOpts` and `ExtraContext`;
* give a more generic name to the resulting merged context, because the EMQX core where the current module lives does not know about clustered configs. `ClusterRPCOpts` actually contains just a flag indicating the reaction to errors during config update.
* update the few existing arity-4 handlers to take namespace info from the context into account (namely, ignore updates with nonempty namespace).


Thus we will avoid adding a new callback.

apps/emqx_conf/src/emqx_conf_app.erl
```diff
@@ -20,6 +20,7 @@
 
 start(_StartType, _StartArgs) ->
     ok = mria:wait_for_tables(emqx_cluster_rpc:create_tables()),
+    _ = emqx_config:create_tables(),
```
savonarola, line 23: `emqx_conf` seems to require `emqx` started, which creates tables. So why do we need to create tables here?

apps/emqx/test/emqx_config_handler_SUITE.erl
```diff
@@ -7,30 +7,244 @@
...
+        false ->
+            emqx:get_raw_config(KeyPath)
+    end.
+
+get_raw_config(TCConfig, KeyPath, Default) when is_list(TCConfig) ->
+    case is_namespaced(TCConfig) of
+        true ->
+            emqx_config:get_raw_namespaced(KeyPath, ?NS, Default);
+        false ->
+            emqx:get_raw_config(KeyPath, Default)
```
savonarola, line 117: Nit
M.b., it would be nice for a potential user of the namespace configs to have a uniform API with regular configs, i.e., to have the respective facade functions in the `emqx` module also.

apps/emqx/src/emqx_config.erl
```diff
@@ -331,6 +435,7 @@ init_load(SchemaMod, Conf) when is_list(Conf) orelse is_binary(Conf) ->
     save_to_app_env(AppEnvs),
     ok = save_to_config_map(CheckedConf, RawConf3),
     maybe_init_default_zone(),
+    load_namespaced_config(SchemaMod),
```
savonarola: Previously, a node wouldn't start if the config was invalid. Now, the config check for NS configs is delayed until the node starts.

At the same time, failing here may be frequent because we regularly update the schema, breaking backward compatibility. 

For regular configs, fixes are relatively easy: start node, see config error, edit the config file.

But for mnesia configs, things may be more complicated.

* We probably need to ensure we have a comprehensive error message: which NS-es have an invalid config, and some instructions on how to fix. Currently, we will probably have a node crash from hocon's throw (is this true?).
* We need a tool to fix an invalid config. If the config becomes invalid, we need to drop/update it. But to even access it, we need mnesia, and so the running cluster, which does not start because of bad configs... E.g., we may add some startup flags, like "ignore bad NS configs", "drop bad NS configs", etc.
* We need some tests for starting with invalid configs.

apps/emqx_conf/src/emqx_conf_app.erl
thalesmg: `emqx_conf` is actually [started before `emqx`](https://github.com/emqx/emqx/blob/a97dd238936f874b5dc0a7ecadfeefca491030e8/apps/emqx_machine/src/emqx_machine_boot.erl#L191-L193), I think.

apps/emqx/test/emqx_config_handler_SUITE.erl
thalesmg: I agree.  However, since there are 2 arities (one for erroring on missing values, another for default values), and the namespaced variants require an extra argument for the namespace, I ended up introducing these `_namespaced` variants.

Unless we went with something like `get_raw_config({Namespace, KeyPath})` etc., then we could maintain the same API.  :thinking:

apps/emqx/src/emqx_config.erl
thalesmg: AFAIK, `zones` won't be part of the allowed namespaced keys (at least in the short/medium term), so `put_config_post_change_actions` should be a no-op.

If we ever add `zones` to the allowed set, then we should consider that, indeed.

apps/emqx/src/emqx_config_handler.erl
thalesmg: Great suggestion, I've just pushed a commit that does this.

apps/emqx/test/emqx_config_handler_SUITE.erl
thalesmg: :checkered_flag:

apps/emqx/src/emqx_config.erl
thalesmg: Excellent point.  I need to think about such cases.

apps/emqx/src/emqx_config.erl
```diff
@@ -380,6 +488,125 @@ fill_defaults_for_all_roots(SchemaMod, RawConf0) ->
...
+    RootKeyAtom = atom(RootKeyBin),
+    RawConf = #{RootKeyBin => RawConf0},
+    case check_config_namespaced(SchemaMod, RawConf, AllowedNSRoots) of
+        {ok, #{} = CheckedRoot} ->
+            CheckedConf = maps:get(RootKeyAtom, CheckedRoot, #{}),
+            clear_invalid_namespaced_config(Namespace, RootKeyBin),
+            put_namespaced(Namespace, [RootKeyAtom], CheckedConf);
+        {error, HoconErrors} ->
+            %% TODO FIXME: check this when (re)starting emqx and raise alarms, log when
+            %% logger is ready.
```
thalesmg: fixme

apps/emqx/src/emqx_config.erl
thalesmg: I changed it so that the node starts up and raises corrupt namespace config errors as alarms, and prevent NS MQTT clients from connecting.

I think that we could add the idea of running `jq` programs to fix NS configs _en masse_ in a future follow up?

thalesmg:
I was going to add a boot/`bats` test now, but realized it can't be done right now: we need at least one allowed root key that gets injected in the release build, and there are none yet.

I'll revisit this test once we add one of the allowed root keys (`connectors`, `actions`, `sources` and/or `rule_engine.rules` (not a root key, but another thing to consider...)).

apps/emqx/src/emqx_config.erl
```diff
@@ -380,6 +489,127 @@ fill_defaults_for_all_roots(SchemaMod, RawConf0) ->
...
+        {ok, #{} = CheckedRoot} ->
+            CheckedConf = maps:get(RootKeyAtom, CheckedRoot, #{}),
+            clear_invalid_namespaced_config(Namespace, RootKeyBin),
+            put_namespaced(Namespace, [RootKeyAtom], CheckedConf);
+        {error, HoconErrors} ->
+            mark_namespaced_config_invalid(Namespace, RootKeyBin, HoconErrors)
+    end.
+
+mark_namespaced_config_invalid(Namespace, RootKeyBin, HoconErrors) when is_binary(Namespace) ->
+    PrevErrors = persistent_term:get(?INVALID_NS_CONF_PT_KEY(Namespace), #{}),
```
savonarola, line 577: Curious: why use PT and not an ets?

apps/emqx/src/emqx_corrupt_namespace_config_checker.erl
```diff
@@ -0,0 +1,118 @@
...
+
+%%------------------------------------------------------------------------------
+%% Internal fns
+%%------------------------------------------------------------------------------
+
+raise_alarms(NamespaceErrors) when is_list(NamespaceErrors) ->
+    lists:foreach(fun raise_alarms/1, NamespaceErrors);
+raise_alarms({Namespace, HoconErrorsPerRoot0}) ->
+    HoconErrorsPerRoot = format_errors(HoconErrorsPerRoot0),
+    emqx_alarm:safe_activate(
```
savonarola: If the number of Namespaces is huge, won't be EMQX or Dashboard overflooded with alarma?

apps/emqx/src/emqx_config.erl
```diff
@@ -380,6 +489,127 @@ fill_defaults_for_all_roots(SchemaMod, RawConf0) ->
...
+clear_invalid_namespaced_config(Namespace, RootKeyBin) when is_binary(Namespace) ->
+    case get_namespace_config_errors(Namespace) of
+        undefined ->
+            ok;
+        #{} = Errors0 ->
+            Errors = maps:remove(RootKeyBin, Errors0),
+            case map_size(Errors) == 0 of
+                true ->
+                    persistent_term:erase(?INVALID_NS_CONF_PT_KEY(Namespace)),
+                    emqx_corrupt_namespace_config_checker:check(Namespace);
```
savonarola: It's a but confusing IMO: why we run "checker" after clearing the wrong config or replacing with a valid? Shouldn't it already be ok?

savonarola (approved):

apps/emqx/src/emqx_config.erl
thalesmg: Since we check it when each namespaced client connects, I thought it would be more lightweight to read: https://github.com/emqx/emqx/blob/4d91e7461c8e021c1f1e85ccfaa2703e26e5d8e2/apps/emqx_mt/src/emqx_mt_hookcb.erl#L52

apps/emqx/src/emqx_corrupt_namespace_config_checker.erl
thalesmg: Yes, I think so...  
I'm not sure how we could handle such a case.  :crying_cat_face:

apps/emqx/src/emqx_config.erl
thalesmg: True.  This is a cast, because I didn't want to make the call from the config process to the alarm process.
But I guess we could then introduce a cast that clears the alarm, which is basically this does.

... and, actually, we should "check" in the other branch as well, so that the alarm message is updated (one root key was fixed, but others remain) :melting_face: :see_no_evil:

apps/emqx/src/emqx_corrupt_namespace_config_checker.erl
thalesmg: Maybe we should raise one fat alarm with all namespaces?  :thinking:

thalesmg: Changed to raise a single alarm for all namespaces.

:checkered_flag:

apps/emqx/src/emqx_config.erl
thalesmg: Fixed to re-check if there are still errors, clear the NS otherwise.

:checkered_flag:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8204)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8204)

PR #15389, opened 2025-06-16
test(mgmt cli): attempt to fix flaky test

thalesmg:
https://github.com/emqx/emqx/actions/runs/15684118558/job/44183628374?pr=15382#step:5:590

```
%%% emqx_mgmt_cli_SUITE ==> t_authz: FAILED
%%% emqx_mgmt_cli_SUITE ==>
Failure/Error: ?assertMatch([ _ ], gen_server : call ( Pid , list_authz_cache ))
  expected: = [ _ ]
       got: []
      line: 258
```

### DIFF
--- a/apps/emqx_management/test/emqx_mgmt_cli_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_cli_SUITE.erl
@@ -247,6 +247,9 @@ t_listeners(_Config) ->
 t_authz(_Config) ->
     %% authz cache-clean all         # Clears authorization cache on all nodes
     ?assertMatch(ok, emqx_ctl:run_command(["authz", "cache-clean", "all"])),
+    %% "cache-clean all" relies on an inserted timestamp in persistent term; we need to
+    %% sleep to avoid considering a freshly inserted cache entry as stale.
+    ct:sleep(1),
     ClientId = "authz_clean_test",
     ClientIdBin = list_to_binary(ClientId),
     %% authz cache-clean <ClientId>  # Clears authorization cache for given client


apps/emqx_management/test/emqx_mgmt_cli_SUITE.erl
```diff
@@ -255,15 +255,15 @@ t_authz(_Config) ->
     {ok, _} = emqtt:connect(C),
     {ok, _, _} = emqtt:subscribe(C, <<"topic/1">>, 1),
     [Pid] = emqx_cm:lookup_channels(ClientIdBin),
-    ?assertMatch([_], gen_server:call(Pid, list_authz_cache)),
+    ?retry(100, 10, ?assertMatch([_], gen_server:call(Pid, list_authz_cache))),
```
savonarola: I also investigated this case a bit, but couldn't quickly figure out the reason why we have empty here. `emqtt:subscribe` is synchronous and populates the cache, how can we get empty 🤔 🤔 🤔

thalesmg: Good question.  :thinking:

thalesmg: ~~Maybe the cache TTL is too tight, and expires just before we ask for it?~~ then we would never get a non-empty result.  :thinking:

thalesmg: https://github.com/emqx/emqx/blob/f1c14e7014e66c07446181732ee92a45d88dc063/apps/emqx_management/test/emqx_mgmt_cli_SUITE.erl#L249

https://github.com/emqx/emqx/blob/f1c14e7014e66c07446181732ee92a45d88dc063/apps/emqx/src/emqx_authz_cache.erl#L169


The "drain cache" operation seems to be just a timestamp that goes into PT.  So, if the test is quick, the timestamp of the just inserted cache might be in the same millisecond and then the fresh entry be considered stale.  :thinking:

thalesmg: Yep, adding a 1 ms sleep seems to stabilize the test.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8168)

PR #15396, opened 2025-06-17
fix(broker): avoid needless `emqx_shared_sub` call on subscriber down

keynslug:
Addresses https://github.com/emqx/emqx/discussions/14011.

Release version: 5.10.1?

## Summary

Server `emqx_shared_sub` manages its own monitors and runs cleanups. Moreover, during high disconnect rate it can be overloaded with DOWN signals, and `unsubscribe/3` call can easily timeout.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_shared_sub_SUITE.erl
+++ b/apps/emqx/test/emqx_shared_sub_SUITE.erl
@@ -792,6 +792,8 @@ t_qos1_random_dispatch_if_all_members_are_down(Config) when is_list(Config) ->
     [Pid1, Pid2] = emqx_shared_sub:subscribers(Group, Topic),
     ?assert(is_process_alive(Pid1)),
     ?assert(is_process_alive(Pid2)),
+    ?retry(100, 10, ?assertEqual(disconnected, get_channel_info(conn_state, Pid1))),
+    ?retry(100, 10, ?assertEqual(disconnected, get_channel_info(conn_state, Pid2))),
 
     {ok, _} = emqtt:publish(ConnPub, Topic, <<"hello11">>, 1),
     ?retry(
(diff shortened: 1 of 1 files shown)


apps/emqx/src/emqx_broker.erl
```diff
@@ -527,6 +525,16 @@ subscriber_down(SubPid) ->
     ),
     ets:delete(?SUBSCRIPTION, SubPid).
 
+do_unsubscribe_down(Topic, SubPid, SubOpts) ->
+    true = ets:delete(?SUBOPTION, {Topic, SubPid}),
+    case Topic of
+        B when is_binary(B) ->
```
zmstone, line 531: nit: maybe add a IS_SHARED_TOPIC macro in emqx_mqtt.hrl.

apps/emqx/src/emqx_broker.erl
```diff
@@ -210,24 +210,23 @@ unsubscribe(Topic) when ?IS_TOPIC(Topic) ->
 do_unsubscribe(Topic, SubPid, SubOpts) ->
     true = ets:delete(?SUBOPTION, {Topic, SubPid}),
     true = ets:delete_object(?SUBSCRIPTION, {SubPid, Topic}),
-    do_unsubscribe2(Topic, SubPid, SubOpts).
+    case Topic of
+        B when is_binary(B) ->
```
zmstone, line 214: nit: maybe add a IS_SHARED_TOPIC macro in emqx_mqtt.hrl.

apps/emqx/src/emqx_broker.erl
keynslug: To keep most common case clauses first, I guess then we'd then need `IS_REGULAR_TOPIC`, which is just `is_binary(X)`?

apps/emqx/src/emqx_broker.erl
keynslug: I'll try to make it look nicer in the followup PRs.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8165)

PR #15401, opened 2025-06-18
feat(snowflake): support new streaming api

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14388

Release version: 6.0.0

## Summary

N.B.: Snowpipe Streaming APIs are still on preview only for AWS accounts, at the time of writing.

## PR Checklist

- [x] The changes are covered with new or existing tests (locally ran test suite against real snowflake)

### DIFF
--- a/apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_channel_client.erl
+++ b/apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_channel_client.erl
@@ -283,27 +283,27 @@ do_append_rows(State0, Records) ->
         ?write_pool_id := WritePoolId
     } = State0,
     JWTToken = emqx_connector_jwt:ensure_jwt(JWTConfig),
+    Headers = [
+        {<<"Content-Type">>, <<"application/json">>},
+        {<<"Accept">>, <<"application/json">>},
+        {<<"User-Agent">>, <<"emqx">>},
+        {<<"Authorization">>, [<<"Bearer ">>, JWTToken]}
+    ],
+    Body = lists:map(fun(R) -> [emqx_utils_json:encode(R), $\n] end, Records),
+    AppendRowsPath = iolist_to_binary([
+        AppendRowsPath0,
+        <<"?continuationToken=">>,
+        uri_quote(ContinuationToken)
+    ]),
+    Req = {AppendRowsPath, Headers, Body},
+    ?tp(debug, "snowflake_streaming_append_rows_request", #{
+        write_pool_id => WritePoolId,
+        channel => ChannelName,
+        path => AppendRowsPath,
+        records => Records
+    }),
+    Resp = ?MODULE:do_append_rows({WritePoolId, Id}, Req, RequestTTL, MaxRetries),
     maybe
-        Headers = [
-            {<<"Content-Type">>, <<"application/json">>},
-            {<<"Accept">>, <<"application/json">>},
-            {<<"User-Agent">>, <<"emqx">>},
-            {<<"Authorization">>, [<<"Bearer ">>, JWTToken]}
-        ],
-        Body = lists:map(fun(R) -> [emqx_utils_json:encode(R), $\n] end, Records),
-        AppendRowsPath = iolist_to_binary([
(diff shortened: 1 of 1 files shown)


thalesmg:
Todo:
- [x] classify ehttpc results
- [x] open channel when client restarts
- [x] actually use allocation apis
- [x] try again to use "normal" JWT instead of rowset token

apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
```diff
@@ -1943,3 +1954,19 @@ clean_aggregated_upload_work_dir() ->
     Dir = filename:join([emqx:data_dir(), bridge]),
     _ = file:del_dir_r(Dir),
     ok.
+
+%% https://stackoverflow.com/a/34883331/2708711
+all_atoms() ->
+    do_all_atoms(_N = 0, []).
+
+idx_to_atom(N) ->
+    binary_to_term(<<131, 75, N:24>>).
```
keynslug, line 1963: While unlikely, this might break in the future because it is not part of documented API contract. I'd personally lean to the safer side and maybe suggest to trace `list_to_atom/1` and friends in a comment.

apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_channel_client.erl
```diff
@@ -0,0 +1,329 @@
...
+            <<"?continuationToken=">>,
+            uri_quote(ContinuationToken)
+        ]),
+        Req = {AppendRowsPath, Headers, Body},
+        ?tp(debug, "snowflake_streaming_append_rows_request", #{
+            write_pool_id => WritePoolId,
+            channel => ChannelName,
+            path => AppendRowsPath,
+            records => Records
+        }),
```
keynslug: Nit. Those expressions are basically infallible, and having them inside `maybe` adds extra cognitive load.

keynslug (approved):

apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
thalesmg: Indeed, it _could_ change in the future.  Though, tracing all such calls may not help, I think, if the atoms come from lazy loading module code while in test (interactive code loading).  Which seemed to be the case while I was debugging: it seemed to break when it decided to load `ssl` and some atom literals were created.

apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_channel_client.erl
thalesmg: Fair.  I think these are inside the `maybe` as a sub-product of previous refactoring.

apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
keynslug: >  Though, tracing all such calls may not help, I think, if the atoms come from lazy loading module code while in test

Yes that was another concern. 🫠 This sounds like valid atom table growth, which would result in false negative in this case.

keynslug (approved):

apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
keynslug: ...But that admittedly is a separate concern. Overall, I think we should only really care about atom table exhaustion that is a product of runtime `erlang:*_to_atom/1,2` calls. So comparing atom table size has potential for false negatives = flaky testcases.

Admittedly, you can also (unwillingly) create new atoms through:
* `binary_to_term/1`ing serialized data, e.g. loaded from storage.
* Programmatically construct, compile and load modules.

...But both of those can be safely ignored for now I think.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8177)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8177)

PR #15406, opened 2025-06-19
feat(ds): do not create static keys in LTS for levels with zero threshold

savonarola:
Needed for https://github.com/emqx/emqx-plugin-ecq/pull/3

### DIFF
--- a/apps/emqx_durable_storage/src/emqx_ds_lts.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_lts.erl
@@ -530,7 +530,7 @@ do_lookup_topic_key(Trie, State, [Tok | Rest], Varying) ->
 do_topic_key(Trie, _, _, State, [], _Parent, Tokens, Varying) ->
     %% We reached the end of topic. Assert: Trie node that corresponds
     %% to EOT cannot be a wildcard.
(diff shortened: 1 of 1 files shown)


apps/emqx_durable_storage/src/emqx_ds_lts.erl
```diff
@@ -542,9 +542,19 @@ do_topic_key(Trie, _, _, State, [], _Parent, Tokens, Varying) ->
...
     % TODO: it's not necessary to call it every time.
     Threshold = ThresholdFun(Depth, Parent),
-    {NChildren, IsWildcard, NextState} = trie_next_(Trie, State, Tok),
+    %% If the threshold is 0, we do not want to create a new static node for this level.
+    CreateTok =
+        case Threshold of
+            0 -> ?PLUS;
+            _ -> Tok
+        end,
+    {NChildren, IsWildcard, NextState} = trie_next_(Trie, State, Tok, CreateTok),
```
savonarola, line 551: Looks a bit ugly, but the logic here is quite critical & complex, so I tried not to change it significantly.

ieQu1:
Here's an alternative implementation: https://github.com/emqx/emqx/pull/15409

PR #15407, opened 2025-06-19
perf(broker): try to minimize number of subscriber shards

keynslug:
Part of EMQX-14333.

Release version: 6.0?

## Summary

Number of shards per topic is not limited now, and each shard will hold no more that 1024 (`emqx_broker_helper:shard_capacity/0`) subscribers.

Downside is: subscriber table may now hold more than 1024 + 32 * C entries per topic, where C is number of schedulers, but in the general case number of entries now scales better with number of subscribers.

Also instrument core dispatch/delivery hot-paths with few latency metrics, disabled by default to have zero runtime impact.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_broker.erl
+++ b/apps/emqx/src/emqx_broker.erl
@@ -117,11 +117,13 @@ create_tabs() ->
     ok = emqx_utils_ets:new(?SUBOPTION, [ordered_set | TabOpts]),
 
     %% Subscription: SubPid -> TopicFilter1, TopicFilter2, TopicFilter3, ...
-    %% duplicate_bag: o(1) insert
+    %% duplicate_bag: O(1) insert
     ok = emqx_utils_ets:new(?SUBSCRIPTION, [duplicate_bag | TabOpts]),
 
-    %% Subscriber: Topic -> SubPid1, SubPid2, SubPid3, ...
-    %% bag: o(n) insert:(
+    %% Subscriber: Topic -> SubPid1, SubPid2, SubPid3, ..., ShardPid1, ...
+    %% bag: O(n) insert
+    %% However, the factor is small: high-fanout topics are _sharded_ into
+    %% buckets of constant length (e.g. 1024).
     ok = emqx_utils_ets:new(?SUBSCRIBER, [bag | TabOpts]).
 
 %%------------------------------------------------------------------------------
@@ -165,9 +167,7 @@ with_subid(SubId, SubOpts) ->
     maps:put(subid, SubId, SubOpts).
 
 do_subscribe(Topic, SubPid, SubOpts) when is_binary(Topic) ->
-    %% FIXME: subscribe shard bug
-    %% https://emqx.atlassian.net/browse/EMQX-10214
-    I = emqx_broker_helper:get_sub_shard(SubPid, Topic),
+    I = emqx_broker_helper:assign_sub_shard(Topic),
     true = ets:insert(?SUBOPTION, {{Topic, SubPid}, with_shard_idx(I, SubOpts)}),
     Sync = call(pick({Topic, I}), {subscribe, Topic, SubPid, I}),
     case Sync of
@@ -217,8 +217,8 @@ do_unsubscribe(Topic, SubPid, SubOpts) ->
 -spec do_unsubscribe_regular(emqx_types:topic(), pid(), emqx_types:subopts()) ->
     ok.
 do_unsubscribe_regular(Topic, SubPid, SubOpts) ->
-    _ = emqx_broker_helper:reclaim_seq(Topic),
     I = maps:get(shard, SubOpts, 0),
+    _ = emqx_broker_helper:unassign_sub_shard(Topic, I),
     case I of
         0 -> emqx_exclusive_subscription:unsubscribe(Topic, SubOpts);
         _ -> ok
--- a/apps/emqx/src/emqx_broker.erl
+++ b/apps/emqx/src/emqx_broker.erl
@@ -11,6 +11,7 @@
 -include("emqx_external_trace.hrl").
 
 -include("logger.hrl").
+-include("emqx_instr.hrl").
 -include("types.hrl").
 -include("emqx_mqtt.hrl").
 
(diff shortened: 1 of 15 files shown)


zmstone:
> Also instrument core dispatch/delivery hot-paths with few latency metrics, disabled by default to have zero runtime impact.


maybe disable by persistent_term?

keynslug:
> maybe disable by persistent_term?

In its current shape, this is pretty much useless outside synthetic performance benchmark scenarios, and having separate builds for those is manageable, albeit admittedly cumbersome. Also apparently in bright hot code paths impact of just a couple of `persistent_term:get/2`s is noticeable. That said, maybe do it later, when need arises + convenience outweighs these small penalties?

rebar.config.erl
```diff
@@ -158,6 +158,7 @@ common_compile_opts(Vsn) ->
         {d, 'EMQX_RELEASE_EDITION', ee}
     ] ++
         [{d, 'EMQX_BENCHMARK'} || os:getenv("EMQX_BENCHMARK") =:= "1"] ++
+        [{d, 'EMQX_BROKER_INSTR'} || os:getenv("EMQX_BROKER_INSTR") =:= "1"] ++
```
thalesmg, line 161: This needs to be added to `mix.exs` as well, in its `erlc_options/1`.

apps/emqx/src/emqx_broker_helper.erl
```diff
@@ -76,17 +83,61 @@ lookup_subid(SubPid) when is_pid(SubPid) ->
...
+assign_sub_shard(Topic) ->
+    Ix = ets:lookup_element(?HELPER, Topic, 2, 0),
+    Shard = {Topic, Ix},
+    case ets:update_counter(?HELPER, Shard, {2, 1}, {'_', 0}) of
+        N when N =< ?SHARD_CAPACITY ->
+            Ix;
+        _ ->
+            %% Subject to races.
+            _ = ets:update_counter(?HELPER, Topic, {2, 1, Ix + 1, Ix + 1}, Shard),
+            Ix
```
thalesmg, line 112: Q: shouldn't this sub-shard index be bumped in this case?  i.e., shouldn't it be the value returned by this last `update_counter`?

apps/emqx/src/emqx_broker.erl
```diff
@@ -168,9 +167,7 @@ with_subid(SubId, SubOpts) ->
     maps:put(subid, SubId, SubOpts).
 
 do_subscribe(Topic, SubPid, SubOpts) when is_binary(Topic) ->
-    %% FIXME: subscribe shard bug
-    %% https://emqx.atlassian.net/browse/EMQX-10214
```
JimMoen, line 172: was it fixed in this PR?

apps/emqx/src/emqx_broker_helper.erl
keynslug: At this point we already bumped `{Topic, Ix}` counter up = reserved space in `Ix`th shard, and during _unassign_ will need to decrement it.

apps/emqx/src/emqx_broker.erl
keynslug: I haven't completely understood the issue outlined in the ticket, but as far as I could tell it no longer applies.

> there were 5 {shard, _} tuple in subscribers list.

Specifically, this is no longer the case.

apps/emqx/test/emqx_broker_helper_SUITE.erl
```diff
@@ -65,16 +58,134 @@ t_shard_seq(_) ->
     emqx_broker_helper:reclaim_seq(TestTopic),
```
thalesmg: Both `{reclaim,create}_seq` seem to be only used here now.

apps/emqx/src/emqx_broker_helper.erl
```diff
@@ -18,7 +20,10 @@
     register_sub/2,
     lookup_subid/1,
     lookup_subpid/1,
-    get_sub_shard/2,
+    assign_sub_shard/1,
+    assigned_sub_shards/1,
+    unassign_sub_shard/2,
+    shard_capacity/0,
     create_seq/1,
     reclaim_seq/1
```
thalesmg: Seems to be dead code now

thalesmg (approved):
:fire: 

Nit: would be nice to have a function (module?) doc explaining the shard assign and (specially) unassign/refill logic in a bit more detail.

apps/emqx/test/emqx_broker_helper_SUITE.erl
keynslug: ✂️

apps/emqx/src/emqx_broker_helper.erl
keynslug: ✂️

keynslug:
> Nit: would be nice to have a function (module?) doc explaining the shard assign and (specially) unassign/refill logic in a bit more detail.

Please take a look at [bb1f85e](https://github.com/emqx/emqx/pull/15407/commits/bb1f85e38c53df51ba6e403e0338ca107fee64d4). I might have unknowingly left out important details, so suggestions are always welcome.

apps/emqx/src/emqx_broker_helper.erl
```diff
@@ -76,25 +80,98 @@ lookup_subid(SubPid) when is_pid(SubPid) ->
...
 
--spec shards_num() -> pos_integer().
-shards_num() ->
-    %% Dynamic sharding later...
-    ets:lookup_element(?HELPER, shards, 2).
+-doc """
+Unassign `Topic` from the given shard `Ix`, where `Ix` was obtained before by
+calling `assign_sub_shard(Topic)`.
+
+\"Unassignment\" is performed in such a way that further assignments will tend
```
thalesmg, line 119: Femto note: with triple quotes, it's not necessary to escape single double-quotes.  Though, it doesn't hurt to leave the escaping here either because it'll render the same (by `IEx.Helpers.h/1`, for example).

thalesmg (approved):
:rocket: :shipit:

apps/emqx/src/emqx_broker_helper.erl
keynslug: Makes perfect sense, I guess my syntax highlighter misled me. 🥲

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8197)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8198)

PR #15430, opened 2025-06-25
chore(new mix build): fix rebar3/mix build discrepancies

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14412
Fixes https://emqx.atlassian.net/browse/EMQX-14413

Release version: 6.0.?

## Summary

Also addresses circular dependency of apps when using `emqx_connector_examples` and `emqx_{action,connector}_info`.

On the one hand, `emqx_{action,connector}_info` and `emqx_connector_examples` are
referenced by several bridges.  These modules used to live in the `emqx_connector` and
`emqx_bridge` applications.

On the other hand, `emqx_bridge` and `emqx_connector` apps have all `emqx_bridge_*` apps
as dependencies, creating a circular dependency of such apps.  Thus, the behaviour modules
were moved to a new application to avoid this problem.

## PR Checklist

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -220,7 +220,7 @@ defmodule EMQXUmbrella.MixProject do
   def common_dep(:jsx), do: {:jsx, github: "talentdeficit/jsx", tag: "v3.1.0", override: true}
   # in conflict by emqtt and hocon
   def common_dep(:getopt), do: {:getopt, "1.0.2", override: true}
-  def common_dep(:telemetry), do: {:telemetry, "1.3.0", override: true}
+  def common_dep(:telemetry), do: {:telemetry, "1.3.0", manager: :rebar3, override: true}
   # in conflict by grpc and eetcd
   def common_dep(:gpb), do: {:gpb, "4.21.1", override: true, runtime: false}
   def common_dep(:ra), do: {:ra, github: "emqx/ra", tag: "v2.15.2-emqx-3", override: true}


apps/emqx_rule_engine/mix.exs
```diff
@@ -5,9 +5,9 @@ defmodule EMQXRuleEngine.MixProject do
   def project do
     [
       app: :emqx_rule_engine,
-      version: "0.1.0",
+      version: "5.2.9",
       build_path: "../../_build",
-      erlc_options: UMP.erlc_options(),
+      erlc_options: [:compressed, {:parse_transform} | UMP.strict_erlc_options()],
```
keynslug: Nit. I suspect this `compressed` is some legacy leftover that doesn't make sense anymore. I'd suggest to drop it (perhaps in a followup).

apps/emqx_auth/mix.exs
```diff
@@ -5,13 +5,13 @@ defmodule EMQXAuth.MixProject do
...
       app: :emqx_auth,
-      version: "0.1.0",
+      version: "0.4.7",
       build_path: "../../_build",
       compilers: Mix.compilers() ++ [:copy_srcs],
       # used by our `Mix.Tasks.Compile.CopySrcs` compiler
       extra_dirs: extra_dirs(),
       # config_path: "../../config/config.exs",
-      erlc_options: UMP.erlc_options(),
+      erlc_options: [{:parse_transform} | UMP.strict_erlc_options()],
```
keynslug: I haven't found any evidence (neither in `rebar3` nor `compiler` code) this is actually a valid option that does anything. Removing them also does not seem to cause any trouble. While you're on it, perhaps it's better to remove it from `rebar.config`s and simplify `mix.exs`s instead?

apps/emqx_dashboard/mix.exs
```diff
@@ -20,15 +20,18 @@ defmodule EMQXDashboard.MixProject do
...
 
   def deps() do
     [
       {:pot, "1.0.2"},
       {:emqx_ctl, in_umbrella: true},
+      {:emqx_utils, in_umbrella: true},
       {:emqx, in_umbrella: true},
-      UMP.common_dep(:minirest),
+      ## Probably not actually needed?
+      {:emqx_bridge_http, in_umbrella: true},
```
keynslug: Almost definitely, such dependency does not make sense.

apps/emqx_gateway/mix.exs
```diff
@@ -21,7 +21,20 @@ defmodule EMQXGateway.MixProject do
...
+    [
+      extra_applications: [
+        :emqx_auth_http,
+        :emqx_auth_jwt,
+        :emqx_auth_ldap,
+        :emqx_auth_mnesia,
+        :emqx_auth_mongodb,
+        :emqx_auth_mysql,
+        :emqx_auth_postgresql,
+        :emqx_auth_redis
```
keynslug, line 33: Q: Trying to get better understanding: why is needed?

apps/emqx_gateway_lwm2m/mix.exs
```diff
@@ -18,7 +18,7 @@ defmodule EMQXGatewayLwm2m.MixProject do
   end
 
   def application do
-    [extra_applications: UMP.extra_applications()]
+    [extra_applications: [:xmerl | UMP.extra_applications()]]
```
keynslug, line 21: Q: Isn't it just a dependency? Or is it something to do with the fact that it's an OTP application?

apps/emqx_schema_registry/mix.exs
```diff
@@ -21,23 +21,29 @@ defmodule EMQXSchemaRegistry.MixProject do
   end
 
   def application do
-    [extra_applications: UMP.extra_applications(), mod: {:emqx_schema_registry_app, []}]
+    [
+      extra_applications: UMP.extra_applications(),
+      included_applications: [:emqx_rule_engine],
```
keynslug, line 26: Q: Why `emqx_rule_engine` is an "included application"?

keynslug (commented):
> On the other hand, emqx_bridge and emqx_connector apps have all emqx_bridge_* apps as dependencies, creating a circular dependency of such apps.

This sounds wrong. I'd honestly advise against separating behavior modules that make perfect sense in their respective apps.

To understand better: how is this circular dependency expressed? I do not see anything in `emqx_bridge.app.src` / `emqx_connector.app.src`, do you mean `emqx_bridge_enterprise` module?

keynslug:
Btw, hands down really amazing work!

thalesmg:
> This sounds wrong. I'd honestly advise against separating behavior modules that make perfect sense in their respective apps.

The problem is that rebar3 is sloppier than mix in regards to referencing behaviours (and other things).  So I think it's more correct now.  :thinking: 
 
> To understand better: how is this circular dependency expressed? I do not see anything in `emqx_bridge.app.src` / `emqx_connector.app.src`, do you mean `emqx_bridge_enterprise` module?

As I described in the description, the behaviours lived in `emqx_bridge` and `emqx_connector` apps.  These behaviours are used in `emqx_bridge_*` apps (as in, they have `-behaviour(xxx).` in their modules).  `emqx_bridge` and `emqx_connector` have all `emqx_bridge_*` apps [injected as dependencies](https://github.com/emqx/emqx/blob/64da249cd0a6ab1412c0a300a6d3ae2db68b1c1b/apps/emqx_machine/src/emqx_machine_boot.erl#L171-L179), to ensure correct boot order.  This injection is handled by `emqx_machine`, that's why it doesn't appear in `.app.src`.
So, in order to reference the behaviours, `emqx_bridge_*` apps need to reference `emqx_bridge` and `emqx_connector` as dependencies.  That's when the circularity appear.

apps/emqx_rule_engine/mix.exs
thalesmg: Yes, some stuff seem outdated, but I just wanted to make both builds equal for now.

apps/emqx_gateway/mix.exs
thalesmg: This is to match gateway's [`.app.src`](https://github.com/emqx/emqx/blob/64da249cd0a6ab1412c0a300a6d3ae2db68b1c1b/apps/emqx_gateway/src/emqx_gateway.app.src#L13-L20)

apps/emqx_gateway_lwm2m/mix.exs
thalesmg: It's because it's an OTP app, so this is the way to declare this application depends on it being started.

apps/emqx_schema_registry/mix.exs
thalesmg: This matches the `.app.src` from schema registry, and, IIRC, schema registry has [some calls](https://github.com/emqx/emqx/blob/64da249cd0a6ab1412c0a300a6d3ae2db68b1c1b/apps/emqx_schema_registry/src/emqx_schema_registry_app.erl#L15) to `emqx_rule_engine` stuff in there.

Those calls changed recently after Ilia refactored some things, maybe it's not necessary to include it anymore.  :thinking:

thalesmg: Also, we want to just "include" so that we may reference those modules in there, but not require `emqx_rule_engine` app to be started, since the latter depends on schema registry... :sweat_smile: :speak_no_evil:

apps/emqx_auth/mix.exs
thalesmg: Weird, I wonder where this came from.  :thinking: 

I'll push a commit removing those and see if things keep working.

apps/emqx_schema_registry/mix.exs
keynslug: So far this looks like a regular dependency of `emqx_schema_registry` on `emqx_rule_engine`.

> the latter depends on schema registry

Once again I can't find where where this dependency is declared. 🫠

apps/emqx_gateway/mix.exs
keynslug: I mean, sometimes intra-umbrella dependencies go into `deps()`, sometimes into `extra_applications`. So I'm asking because it's not apparent what is the pattern. 🙃

thalesmg: Right.  Both would indeed work, but I went with `extra_applications` here because we just want to start those apps before the gateway, but we don't directly reference modules or call them in gateway's code (I think).

apps/emqx_schema_registry/mix.exs
thalesmg: > Once again I can't find where where this dependency is declared. 🫠

Huh, right, it's not written anywhere.  :melting_face: 

IIRC, the problem of making it "official" and adding `emqx_schema_registry` to `emqx_rule_engine` deps is that then a cycle is introduced:

```
** (Mix) Could not sort dependencies. The following dependencies form a cycle: ekka, emqx_schema_registry, emqx, emqx_modules, emqx_rule_engine, mnesia_rocksdb, emqx_conf, emqx_auth, mria, sext
```

... yet, adding it to `extra_applications` seems to work.  :thinking: 

I'll push one commit adding it there.

thalesmg: Those circularity problems seem to come from build-time dependencies (referencing code and behaviours) and runtime deps (starting apps in the correct order).

keynslug:
> `emqx_bridge` and `emqx_connector` have all `emqx_bridge_*` apps [injected as dependencies](https://github.com/emqx/emqx/blob/64da249cd0a6ab1412c0a300a6d3ae2db68b1c1b/apps/emqx_machine/src/emqx_machine_boot.erl#L171-L179), to ensure correct boot order

Oh god. No wonder I missed it. 🫠

>  So I think it's more correct now. 🤔

Yes. It's now clearer now, thanks for the explanation. We're in slight disagreement because of naming and/or scope then I guess. In my head, dependency chain should ideally look like this:
```
emqx_bridge (logic, types and behaviours common for all bridges)
     ↖(depends on)– emqx_bridge_* (specific bridge implementation)
                         ↖(depends on) emqx_bridges (application that knows about all bridges + relevant logic)
```

In this PR, `emqx_connector_info` is now part of what I think `emqx_bridge` application should be, because it only separates behaviours (but not common types and logic).

This would also help to get rid of this another very special way of "declaring dependencies" (which in this case I guess is actually "unbreaking the broken dependency chain") on `emqx_machine` level. Unless of course I'm missing important details about its purpose.

Not suggesting to do it in this PR of course. Perhaps just renaming (to something like `emqx_gen_bridge`) to avoid doing it in followup PRs.

thalesmg:
> Yes. It's now clearer now, thanks for the explanation. We're in slight disagreement because of naming and/or scope then I guess. In my head, dependency chain should ideally look like this:
> 
> ```
> emqx_bridge (logic, types and behaviours common for all bridges)
>      ↖(depends on)– emqx_bridge_* (specific bridge implementation)
>                          ↖(depends on) emqx_bridges (application that knows about all bridges + relevant logic)
> ```

I think this is already the state in this PR, _modulo_ application names?  (with the added detail that `emqx_bridges` in your terminology also depends on `emqx_bridge`).


> In this PR, `emqx_connector_info` is now part of what I think `emqx_bridge` application should be, because it only separates behaviours (but not common types and logic).
> 
> This would also help to get rid of this another very special way of "declaring dependencies" (which in this case I guess is actually "unbreaking the broken dependency chain") on `emqx_machine` level. Unless of course I'm missing important details about its purpose.

I think this dynamic injection comes from the time when we had the distinct builds (CE and EE).  Another reason, IIRC, is to avoid forgetting to add a bridge app as dependency to `emqx_bridge` and having issues with start order.  We had such issues in the past...


> Not suggesting to do it in this PR of course. Perhaps just renaming (to something like `emqx_gen_bridge`) to avoid doing it in followup PRs.

Do you mean renaming `emqx_connector_info` to `emqx_gen_bridge`?

apps/emqx_schema_registry/mix.exs
keynslug: Ah... 😣 Shame mix doesn't tell what exactly the cycle is. I suspect that it might no longer be there once `emqx_bridge_http` is removed from `emqx_schema_registry` deps. Not advocating to change anything though.

thalesmg: > Ah... 😣 Shame mix doesn't tell what exactly the cycle is. 

It does, as in the example pasted above:

```
** (Mix) Could not sort dependencies. The following dependencies form a cycle: ekka, emqx_schema_registry, emqx, emqx_modules, emqx_rule_engine, mnesia_rocksdb, emqx_conf, emqx_auth, mria, sext
```

Yet I tried adding it as an `extra_dep` (but not _a_ dep), and it worked: https://github.com/emqx/emqx/pull/15430/commits/1e9a2a5478eddc48b997c85adbbf5cb2c93cc9f4

keynslug:
> Do you mean renaming `emqx_connector_info` to `emqx_gen_bridge`?

Yes.

> Another reason, IIRC, is to avoid forgetting to add a bridge app as dependency to emqx_bridge and having issues with start order.

Right. To clarify: you mean "applications are started in incorrect order" or "we forgot to include bridge in the release"? The latter sounds like a general problem, I'd call to introduce (or perhaps implement myself) a general solution.

thalesmg:
> Right. To clarify: you mean "applications are started in incorrect order" or "we forgot to include bridge in the release"? The latter sounds like a general problem, I'd call to introduce (or perhaps implement myself) a general solution.

The former.  Each bridge app must be started before `emqx_bridge` app is started (using current terminology).

apps/emqx_schema_registry/mix.exs
keynslug: > It does, as in the example pasted above:

I mean, it's just a list of applications, you can't easily tell where the cycle is. Confusingly enough, it also includes `sext` which depends only on `stdlib` and `kernel`.

thalesmg: Ah, I see.  Indeed, it may be tricky to correctly tell the user (from mix's point of view) who is the culprit for introducing the cycle, I guess.  
One has to rely on recent changes to deps to know who last introduced one of the apps in the cycle to their deps.

thalesmg:
> The former. Each bridge app must be started before `emqx_bridge` app is started (using current terminology).

... and the automatic injection relies on all bridge apps following the `emqx_bridge_*` naming convention.

apps/emqx_schema_registry/mix.exs
thalesmg: Huh.  Compilation works, but something broke in the tests, and of course it's not reproducible locally...  Great rebar3 stuff as usual.  :melting_face:  :melting_face:  :melting_face: 

I'll drop this change for now.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8185)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8185)

PR #15431, opened 2025-06-25
feat(new mix build): port xref check to mix task

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14421

Release version: 6.0.?

## PR Checklist

- [na] The changes are covered with new or existing tests (manually tested)
- [na] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files

### DIFF
--- a/apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
+++ b/apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
@@ -1,0 +1,193 @@
+defmodule Mix.Tasks.Emqx.Xref do
+  use Mix.Task
+
+  alias Mix.Tasks.Emqx.Ct, as: ECt
+  alias EMQXUmbrella.MixProject, as: UMP
+
+  @requirements ["compile", "loadpaths"]
+
+  @xref :xref
+
+  @default_checks [
+    :undefined_function_calls,
+    :undefined_functions,
+    :locals_not_used,
+    :deprecated_function_calls,
+    # :warnings_as_errors,
+    :deprecated_functions,
+  ]
+
+  @excluded_apps MapSet.new([
+    :emqx_mix_utils,
+  ])
+
+  @exclusions MapSet.new([
+    EMQXUmbrella.MixProject,
+  ])
+
+  @queries [
+    {"E || \"mnesia\":\"dirty_delete.*\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"transaction\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"async_dirty\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"clear_table\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"create_table\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"delete_table\"/\".*\" : Fun", []}
+  ]
+
+  @impl true
+  def run(args) do
+    start_xref()
+    set_library_path()
+    set_default()
+    add_umbrella_app_dirs()
+    check_failures = run_checks()
+    query_failures = run_queries()
+    stop_xref()
+    case {check_failures, query_failures} do
+      {[], []} ->
+        Mix.shell().info([
+          :green,
+          "No xref issues found"
+        ])
+        :ok
+
+      _ ->
+        Mix.shell().info([
+          :red,
+          "Xref encountered issues:\n\n",
+          inspect(%{
+            check_failures: check_failures,
+            query_failures: query_failures,
+          }, pretty: true),
+        ])
+    end
+
+    :ok
+  end
+
+  def run_checks() do
+    exclusions = gather_exclusions()
+
+    Enum.reduce(@default_checks, [], fn check, acc ->
+      {:ok, results} = :xref.analyse(@xref, check)
+
+      mods = Enum.flat_map(results, fn
+        {m, _f, _a} ->
+          [m]
+
+        {{ms, _fs, _as}, {_mt, _ft, _at}} ->
+          [ms]
+
+        _ ->
+          []
+      end)
+      mod_ignores =
+        mods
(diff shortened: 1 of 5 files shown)


apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
```diff
@@ -0,0 +1,193 @@
...
+        acc
+      else
+        [{check, results} | acc]
+      end
+    end)
+  end
+
+  def run_queries() do
+    Enum.flat_map(@queries, fn {query, expected} ->
+      {:ok, result} = :xref.q(@xref, to_charlist(query))
```
keynslug, line 107: Perhaps define them as charlist literals?

apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
```diff
@@ -0,0 +1,193 @@
...
+
+      _ ->
+        Mix.shell().info([
+          :red,
+          "Xref encountered issues:\n\n",
+          inspect(%{
+            check_failures: check_failures,
+            query_failures: query_failures,
+          }, pretty: true),
+        ])
```
keynslug, line 48: Usually the expectation is to terminate with non-zero exit code in this case. Is it handled elsewhere?

apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
```diff
@@ -0,0 +1,193 @@
...
+    :undefined_functions,
+    :locals_not_used,
+    :deprecated_function_calls,
+    # :warnings_as_errors,
+    :deprecated_functions,
+  ]
+
+  @excluded_apps MapSet.new([
+    :emqx_mix_utils,
+  ])
```
keynslug: Nit. Feels like too many ways to exclude something. Could it perhaps be moved on app's `mix.exs` level, as something like `xref_ignore: true`?

apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
```diff
@@ -0,0 +1,193 @@
...
+  @xref :xref
+
+  @default_checks [
+    :undefined_function_calls,
+    :undefined_functions,
+    :locals_not_used,
+    :deprecated_function_calls,
+    # :warnings_as_errors,
+    :deprecated_functions,
+  ]
```
keynslug: Q: Why is it named "default" but does not seem to be changeable?

apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
```diff
@@ -0,0 +1,193 @@
...
+  ])
+
+  @queries [
+    {"E || \"mnesia\":\"dirty_delete.*\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"transaction\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"async_dirty\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"clear_table\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"create_table\"/\".*\" : Fun", []},
+    {"E || \"mnesia\":\"delete_table\"/\".*\" : Fun", []}
+  ]
```
keynslug: Arguable, but: to me, this (checks and queries) needs to be defined at the top level `mix.exs`, if we think about it as library code (as it says in the filepath).

apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
thalesmg: In Elixir, it's more usual and convenient to both read and write strings (binaries), and charlists are mostly for interfacing with Erlang code (such as this).

IMO, since this is not a hot path nor it's likely to gain much in terms of speed, having the declaration being more convenient to read and write (without the `~c` sigils) is nicer.

apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
thalesmg: Aha, forgot that detail.  :see_no_evil:

apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
thalesmg: This app is a very special case (it's a build-time only dependency), so I think that for now this should suffice.

But, if we happen to have more apps in the future that have a good reason to be wholly ignored from Xref, I agree that we should support an attribute to indicate it should be ignored.

apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
thalesmg: At first, while porting it, I thought there could be some customization at some point, but in the end I think there's no need.  I'll rename this attribute.

apps/emqx_mix_utils/lib/mix/tasks/emqx.xref.ex
thalesmg: IMO, I felt it was more convenient to have everything in one place, since we are not constrained by how rebar3 does it.  But I agree that having it in the main `mix.exs` makes it easier to find.

> as library code (as it says in the filepath).

Not sure I understand this.  Do you mean the `/lib/`?  This is just Mix's convention:  `lib` is the Mix equivalent of Rebar3's `src`.  Since `Mix.Task`s need to be compiled, they need to be `.ex` files in the compiler path (`lib`) instead of scripts (`.exs` anywhere).

This is meant to serve only our build system, not to be a general library to be used in other repos.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8182)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8182)

PR #15442, opened 2025-06-27
test(s3tables): attempt to stabilize flaky suite

thalesmg:


apps/emqx/test/emqx_common_test_helpers.erl
```diff
@@ -1260,7 +1260,12 @@ call_janitor() ->
 
 call_janitor(Timeout) ->
     Janitor = get_or_spawn_janitor(),
-    ok = emqx_test_janitor:stop(Janitor, Timeout),
+    try
+        emqx_test_janitor:stop(Janitor, Timeout)
+    catch
+        K:E:S ->
+            ct:pal("janitor crashed:\n  ~p", [{K, E, S}])
+    end,
```
keynslug: I feel there's a risk of hiding issues introduced by testcase authors that might pop up later in unrelated testcases or even test suites. After all, `end_per_testcase is crashed` is more visible that a log entry hiding inside green testcase output.

thalesmg: Fair.  I reverted this and now only ignore 5xx errors in the s3tables suite to have the desired effect.

thalesmg:
Seems a bit more stable now; ran 14 times without this suite breaking.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8195)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8195)

PR #15450, opened 2025-06-30
ci: run push entrypoint on `release-60`

thalesmg:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8199)

PR #15421, opened 2025-06-24
feat(broker): allow async shard dispatch through broker pool

keynslug:
Part of EMQX-14333.

Release version: 6.0?

## Summary

On top of #15407.

Add feature flag that enables dispatching of subscriber shards through the broker pool. This feature is hidden and disabled by default.

* Impact on high-fanout performance varies, see [report](https://emqx.atlassian.net/wiki/spaces/HOTPOT/pages/1815413979/In-mem+Fanout+Performance).
* No decisive negative impact on 1-to-1 performance, however it proved to be very hard to measure. Work in progress.

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible

PR #15451, opened 2025-06-30
feat(socket): implement `esockd_socket` connection module

keynslug:
Part of EMQX-14333.

Release version: 6.0.0

## Summary

This PR introduces `tcp_backend` listener option, exclusively to TCP listeners, and employs emqx/esockd#204 with a separate connection module named `emqx_socket_connection` when `tcp_backend` is `socket`. This shows consistently better wide-fanout performance, especially when combined with async shard dispatch introduced in emqx/emqx#15421, see [report](https://emqx.atlassian.net/wiki/spaces/HOTPOT/pages/1815413979/In-mem+Fanout+Performance).

This new connection module is identical to `emqx_connection` module in most of aspects, but made separate primarily because of:
* Socket-based connections use substantially different connection loop, and integrating it into `emqx_connection` might have made the latter too bloated, harder to test and reason about.
* Combined module would have larger footprint, which would lead to a bunch of "micro-deoptimizations" affecting every listener, admittedly pretty minor ones.

Obvious downside is the need to keep changes to connection modules in sync. This requires additional development processes, or perhaps some CI magic, otherwise code will eventually diverge. Ideas are welcome.

Notes:
* Commit ed96d999565a0346d7c986bda486f2326c17b457 is part of `iovec()` experiments that I decided to include in this PR, because it shows small consistent improvements in average message processing latency (~2% less) and VM reductions (~3.5% less).
* While TLS connections are currently not supported under `esockd_socket` listeners, there's [ongoing](https://github.com/erlang/otp/pull/9398) [work](https://github.com/erlang/otp/pull/9879) in Erlang/OTP to support familiar SSL API with `socket` sockets.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_frame.erl
+++ b/apps/emqx/src/emqx_frame.erl
@@ -56,8 +56,8 @@
 -type parse_state_initial() :: #options{}.
 
 -type parse_result() ::
-    %% Need more bytes out of stream, `0` means it's unclear how much more.
-    {_NeedMore :: non_neg_integer(), parse_state()}
+    %% Need more bytes out of stream.
+    {some_more | _NBytesMore :: pos_integer(), parse_state()}
     %% There's a full packet.
     | {emqx_types:packet(), binary(), parse_state_initial()}.
 
@@ -144,7 +144,7 @@ parse(
 ) ->
     parse_body_frame(Bin, Header, Need, Body, Options);
 parse(<<>>, State) ->
-    {0, State}.
+    {some_more, State}.
 
 %% @doc Parses _complete_ binary frame into a single `#mqtt_packet{}`.
 -spec parse_complete(iodata(), parse_state_initial()) ->
@@ -170,7 +170,7 @@ parse_complete(
     end.
 
 parse_remaining_len(<<>>, Header, Mult, Length, Options) ->
-    {_NeedMore = 0, #remlen{hdr = Header, len = Length, mult = Mult, opts = Options}};
+    {some_more, #remlen{hdr = Header, len = Length, mult = Mult, opts = Options}};
 parse_remaining_len(<<0:8, Rest/binary>>, Header, 1, 0, Options) ->
     Packet = parse_bodyless_packet(Header),
     {Packet, Rest, Options};
--- a/apps/emqx/src/emqx_frame.erl
+++ b/apps/emqx/src/emqx_frame.erl
@@ -74,8 +74,6 @@
 
 -define(MULTIPLIER_MAX, 16#200000).
 
--dialyzer({no_match, [serialize_utf8_string/3]}).
-
 %% @doc Describe state for logging.
 describe_state(Options = #options{}) ->
     #{
@@ -777,9 +775,11 @@ serialize_pkt(Packet, #{version := Ver, max_size := MaxSize, strict_mode := Stri
     end.
 
 -spec serialize(emqx_types:packet()) -> iodata().
-serialize(Packet) -> serialize(Packet, ?MQTT_PROTO_V4, false).
+serialize(Packet) ->
+    serialize(Packet, ?MQTT_PROTO_V4, false).
 
-serialize(Packet, Ver) -> serialize(Packet, Ver, false).
+serialize(Packet, Ver) ->
+    serialize(Packet, Ver, false).
 
 -spec serialize(emqx_types:packet(), emqx_types:proto_ver(), boolean()) -> iodata().
 serialize(
@@ -791,33 +791,43 @@ serialize(
     Ver,
     StrictMode
 ) ->
-    serialize(
-        Header,
-        serialize_variable(Variable, Ver, StrictMode),
-        serialize_payload(Payload),
-        StrictMode
-    ).
-
-serialize(
-    #mqtt_packet_header{
-        type = Type,
-        dup = Dup,
-        qos = QoS,
-        retain = Retain
-    },
-    VariableBin,
-    PayloadBin,
-    _StrictMode
-) when
-    ?CONNECT =< Type andalso Type =< ?AUTH
-->
-    Len = iolist_size(VariableBin) + iolist_size(PayloadBin),
+    VariableBin = serialize_variable(Variable, Ver, StrictMode),
+    PayloadBin = serialize_payload(Payload),
+    RemLen = iolist_size(VariableBin) + iolist_size(PayloadBin),
     [
-        <<Type:4, (flag(Dup)):1, (flag(QoS)):2, (flag(Retain)):1>>,
-        serialize_remaining_len(Len),
+        serialize_header(Header),
+        serialize_remaining_len(RemLen),
         VariableBin,
         PayloadBin
     ].
+
+-compile(
+    {inline, [
+        serialize_header/1,
+        serialize_payload/1,
+        serialize_remaining_len/1,
+        serialize_variable_byte_integer/1
+    ]}
+).
+
+%% erlfmt-ignore
+-define(bool(B), (case B of true -> 1; false -> 0; undefined -> 0 end):1).
+
+%% erlfmt-ignore
+-define(utf8string(X, STRICT),
+    (begin
+        true = (case STRICT of
+            true -> byte_size(unicode:characters_to_binary(X));
+            false -> byte_size(X)
+        end =< 16#ffff),
+        byte_size(X)
+    end):16/big-unsigned-integer, X/bytes
+).
+
+serialize_header(
+    #mqtt_packet_header{type = Type, dup = Dup, qos = QoS, retain = Retain}
+) when ?CONNECT =< Type andalso Type =< ?AUTH ->
+    <<Type:4, ?bool(Dup), QoS:2, ?bool(Retain)>>.
 
 serialize_variable(
     #mqtt_packet_connect{
@@ -859,19 +869,25 @@ serialize_variable(
             KeepAlive:16/big-unsigned-integer
         >>,
         serialize_properties(Properties, ProtoVer, StrictMode),
-        serialize_utf8_string(ClientId, StrictMode),
+        <<?utf8string(ClientId, StrictMode)>>,
         case WillFlag of
             true ->
                 [
                     serialize_properties(WillProps, ProtoVer, StrictMode),
-                    serialize_utf8_string(WillTopic, StrictMode),
+                    <<?utf8string(WillTopic, StrictMode)>>,
                     serialize_binary_data(WillPayload)
                 ];
             false ->
                 <<>>
         end,
-        serialize_utf8_string(Username, true, StrictMode),
-        serialize_utf8_string(Password, true, StrictMode)
+        case Username of
+            undefined -> <<>>;
+            _ -> <<?utf8string(Username, StrictMode)>>
+        end,
+        case Password of
+            undefined -> <<>>;
+            _ -> <<?utf8string(Password, StrictMode)>>
+        end
     ];
 serialize_variable(
     #mqtt_packet_connack{
@@ -893,7 +909,7 @@ serialize_variable(
     StrictMode
 ) ->
     [
-        serialize_utf8_string(TopicName, StrictMode),
+        <<?utf8string(TopicName, StrictMode)>>,
         case PacketId of
             undefined -> <<>>;
             _ -> <<PacketId:16/big-unsigned-integer>>
@@ -1036,9 +1052,9 @@ serialize_property('Payload-Format-Indicator', Val, _StrictMode) ->
 serialize_property('Message-Expiry-Interval', Val, _StrictMode) ->
     <<16#02, Val:32/big>>;
 serialize_property('Content-Type', Val, StrictMode) ->
-    <<16#03, (serialize_utf8_string(Val, StrictMode))/binary>>;
+    <<16#03, ?utf8string(Val, StrictMode)>>;
 serialize_property('Response-Topic', Val, StrictMode) ->
-    <<16#08, (serialize_utf8_string(Val, StrictMode))/binary>>;
+    <<16#08, ?utf8string(Val, StrictMode)>>;
 serialize_property('Correlation-Data', Val, _StrictMode) ->
     <<16#09, (byte_size(Val)):16, Val/binary>>;
 serialize_property('Subscription-Identifier', Val, _StrictMode) ->
@@ -1046,11 +1062,11 @@ serialize_property('Subscription-Identifier', Val, _StrictMode) ->
 serialize_property('Session-Expiry-Interval', Val, _StrictMode) ->
     <<16#11, Val:32/big>>;
 serialize_property('Assigned-Client-Identifier', Val, StrictMode) ->
-    <<16#12, (serialize_utf8_string(Val, StrictMode))/binary>>;
+    <<16#12, ?utf8string(Val, StrictMode)>>;
 serialize_property('Server-Keep-Alive', Val, _StrictMode) ->
     <<16#13, Val:16/big>>;
 serialize_property('Authentication-Method', Val, StrictMode) ->
-    <<16#15, (serialize_utf8_string(Val, StrictMode))/binary>>;
+    <<16#15, ?utf8string(Val, StrictMode)>>;
 serialize_property('Authentication-Data', Val, _StrictMode) ->
     <<16#16, (iolist_size(Val)):16, Val/binary>>;
 serialize_property('Request-Problem-Information', Val, _StrictMode) ->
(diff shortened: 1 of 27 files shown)


rel/i18n/emqx_schema.hocon
```diff
@@ -613,6 +613,17 @@ Note, the choice of `parse_unit` affects the interpretation of the `active_n` se
 fields_mqtt_opts_parse_unit.label:
 """MQTT Message Parse Unit"""
 
+fields_mqtt_opts_tcp_backend.desc:
+"""Indicates which TCP backend should be used by the listener.
+
+- `gen_tcp`: Standard backend, in use since EMQX 5.0 release.
+
+- `socket`: Experimental backend, looking to improve message latency and compute resource usage.
+    Note that some `tcp_options` settings will have no effect when using this backend, e.g.: `high_watermark` and `send_timeout_close`."""
```
qzhuyan: not suggested for IPv6 as otp doc said it is not fully tested.

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1248 @@
...
+-include("emqx_instr.hrl").
+-include_lib("snabbkaffe/include/snabbkaffe.hrl").
+
+-ifdef(TEST).
+-compile(export_all).
+-compile(nowarn_export_all).
+-endif.
+
+-elvis([{elvis_style, used_ignored_variable, disable}]).
+-elvis([{elvis_style, invalid_dynamic_call, #{ignore => [emqx_connection]}}]).
```
thalesmg: Stale directive?  Or perhaps:

```suggestion
-elvis([{elvis_style, invalid_dynamic_call, #{ignore => [emqx_socket_connection]}}]).
```
?

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1248 @@
...
+%% @doc Gather socket statistics, for `emqx_congestion` alarms.
+-spec sockstats([atom()], state()) -> emqx_types:stats().
+sockstats(Keys, #state{socket = Socket, sockstate = SS}) ->
+    #{counters := Counters} = socket:info(Socket),
+    lists:map(
+        fun
+            (S = recv_oct) -> {S, maps:get(read_byte, Counters, 0)};
+            (S = recv_cnt) -> {S, maps:get(read_pkg, Counters, 0)};
+            (S = send_oct) -> {S, maps:get(write_byte, Counters, 0)};
+            (S = send_cnt) -> {S, maps:get(write_pkg, Counters, 0)};
```
thalesmg, line 203: Just curiosity: what does `pkg` stand for here?  I'd expect something like `pkt` (short for _packet_).  :thinking:

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1248 @@
...
+    case SS of
+        idle ->
+            handle_data_ready(Socket, State);
+        #congested{handle = Handle} ->
+            handle_send_ready(Socket, SS, State);
+        _ ->
+            handle_data_ready(Socket, State)
+    end;
+handle_msg({'$socket', _Socket, abort, {_Handle, Reason}}, State) ->
+    handle_info({sock_error, Reason}, State);
```
thalesmg: Q: could this be a stale message if `_Handle` in `#congested.handle` is different?

rel/i18n/emqx_schema.hocon
keynslug: Well, it says "experimental". Note that the same doc also _does not_ say the same about `{inet_backend, socket}` in `gen_tcp`, which suggests that relevant API surface is well-tested.

apps/emqx/src/emqx_socket_connection.erl
keynslug: Hmm, IIUC _any_ abort basically means someone closed the socket / there was a socket error. Maybe I'm missing something?

apps/emqx/src/emqx_socket_connection.erl
keynslug: Yeah, don't know to be honest. The choice of naming baffles me as well.
https://github.com/emqx/otp/blob/a7cd578faeeda57177fcabd45b65f765b10ef1b4/erts/emulator/nifs/unix/unix_socket_syncio.c#L5360-L5361

apps/emqx/src/emqx_socket_connection.erl
thalesmg: I was asking if it is possible that a stale `abort` arrives here, but is no longer relevant, so it wouldn't be necessary to log it?  e.g. the handle changed, and the `Handle` in this message has been abandoned.

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1248 @@
...
+terminate(
+    Reason,
+    State = #state{channel = Channel}
+) ->
+    try
+        Channel1 = emqx_channel:set_conn_state(disconnected, Channel),
+        emqx_congestion:cancel_alarms(?MODULE, State),
+        emqx_channel:terminate(Reason, Channel1),
+        close_socket_ok(State),
+        ?TRACE("SOCKET", "emqx_connection_terminated", #{reason => Reason})
```
thalesmg, line 623:
```suggestion
        ?TRACE("SOCKET", "emqx_socket_connection_terminated", #{reason => Reason})
```
?

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1248 @@
...
+    catch
+        E:C:S ->
+            ?tp(warning, unclean_terminate, #{exception => E, context => C, stacktrace => S})
```
thalesmg, line 626: Nit: maybe add the connection module as metadata, to differentiate between this module and `emqx_connection`?

Though, seeing how many tracepoints and logs would need to change, maybe it could be added to the process logger metadata.  :thinking:

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1248 @@
...
+            {ok, Msgs, State};
+        true when SS =/= closed ->
+            request_more_data(Socket, More, Msgs, State);
+        _ ->
+            {ok, Msgs, State}
+    end.
+
+-compile({inline, [request_more_data/4]}).
+request_more_data(Socket, More, Acc, State) ->
+    %% TODO: `{otp, select_read}`.
```
thalesmg, line 761: Q: should this be addressed here?  Or maybe in a follow up?

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1248 @@
...
+                exception => Reason,
+                stacktrace => Stacktrace
+            }),
+            erlang:error(?FRAME_SERIALIZE_ERROR)
+    end.
+
+%%--------------------------------------------------------------------
+%% Send data
+
+-spec send(non_neg_integer(), iodata(), state()) -> {ok, state()}.
```
thalesmg: Nit: typespec does not cover `{ok, {sock_error, _}, state()}`.

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1248 @@
...
+    Handle = make_ref(),
+    case socket:send(Socket, IoData, [], Handle) of
+        ok ->
+            {ok, State};
+        {select, {_Info, Rest}} ->
+            %% Partially accepted, renew deadline.
+            {ok, queue_send(Handle, Rest, State)};
+        {select, _Info} ->
+            %% Totally congested, keep the deadline.
+            NState = State#state{sockstate = SS#congested{sendq = IoData}},
```
thalesmg: Q: don't we need to update/set `#congested.handle` to the freshly created handle here?

thalesmg (dismissed):
:fire: :shipit:

apps/emqx/src/emqx_socket_connection.erl
keynslug: `sock_error` later forces the connection process to close and abandon the socket. But in theory we may get 2 aborts per each outstanding select (one for `recv`, one for `send`). So, spot on, I'll double-check.

apps/emqx/src/emqx_socket_connection.erl
keynslug: Good find! That's most likely a bug.

apps/emqx/src/emqx_socket_connection.erl
keynslug: I was thinking about a followup. It's not entirely clear this will prove useful, and at very least need experimentation.

apps/emqx/src/emqx_socket_connection.erl
keynslug: Good point, did the latter.

apps/emqx/src/emqx_frame.erl
```diff
@@ -56,7 +56,9 @@
 -type parse_state_initial() :: #options{}.
 
 -type parse_result() ::
-    {more, parse_state()}
+    %% Need more bytes out of stream, `0` means it's unclear how much more.
```
qzhuyan: it is counter-intuitive, maybe error prone.

More = 0, is nomore to me. 

I prefer type `unknown | non_neg_integer()`

apps/emqx/test/emqx_client_SUITE.erl
```diff
@@ -165,53 +207,47 @@ t_will_message(_Config) ->
...
     {ok, C2} = emqtt:start_link([
         {clean_start, true},
         {clientid, <<"c2">>}
     ]),
     {ok, _} = emqtt:connect(C2),
 
-    ok = emqtt:publish(C2, nth(2, ?TOPICS), <<"qos 0">>, 0),
-    {ok, _} = emqtt:publish(C2, nth(3, ?TOPICS), <<"qos 1">>, 1),
-    {ok, _} = emqtt:publish(C2, nth(4, ?TOPICS), <<"qos 2">>, 2),
+    ok = emqtt:publish(C2, <<"TopicA/B">>, <<"qos 0">>, 0),
```
qzhuyan, line 230: nit, honestly I think old code is more readable

apps/emqx/src/emqx_listeners.erl
```diff
@@ -454,13 +460,35 @@ do_start_listener(quic, Name, Id, #{bind := Bind} = Opts) ->
...
+                socket -> ConnMod = emqx_socket_connection
+            end,
+            esockd:reset_options(
+                {Id, ListenOn},
+                esockd_opts(Id, Type, Name, ConnMod, NewConf)
+            );
+        _Different ->
+            %% TODO
+            %% Again, we're not strictly required to drop live connections in this case.
+            {error, not_supported}
```
zmstone, line 480: maybe add some description to the config schema desc?
e.g. when choose to use `socket` (instead of `gen_tcp`), the socket options cannot be updated from Dashboard or API.

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1267 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2018-2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+%% This module interacts with the transport layer of MQTT
+%% Transport: esockd_socket.
+%%
+%% NOTE
```
zmstone, line 8: nit: add same note to emqx_connection too ?

apps/emqx/test/emqx_client_SUITE.erl
keynslug: Interesting! To me, it was always an additional cognitive load to understand how `nth(3, ?TOPICS)` and `nth(4, ?TOPICS)` relate to `nth(6, ?WILD_TOPICS)` in one testcase, or why `nth(7, ?WILD_TOPICS)` and `nth(1, ?WILD_TOPICS)` were specifically chosen for "overlapping subscriptions" testcase, etc.

apps/emqx/src/emqx_frame.erl
keynslug: While I agree about explicitness / error-proneness, `unknown` and `0` have no practical difference for callers, at least for now: i.e. there's no such state as "we don't need to read byte stream anymore". And using `0` avoids an extra mapping step in `emqx_socket_connection`.

> I prefer type `unknown | non_neg_integer()`

Perhaps you meant `unknown | pos_integer()`?

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1267 @@
...
+        true when SS =/= closed ->
+            request_more_data(Socket, More, Msgs, State);
+        _ ->
+            {ok, Msgs, State}
+    end.
+
+-compile({inline, [request_more_data/4]}).
+request_more_data(Socket, More, Acc, State) ->
+    %% TODO: `{otp, select_read}`.
+    case sock_async_recv(Socket, More) of
```
zmstone, line 762: will this return `{complete, ComplteInfo}`?

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1267 @@
...
+    end,
+    handle_info({sock_closed, Reason}, ensure_close_socket(Reason, State));
+%% handle QUIC control stream events
+handle_info({quic, Event, Handle, Prop}, State) when is_atom(Event) ->
+    case emqx_quic_stream:Event(Handle, Prop, State) of
+        {{continue, Msgs}, NewState} ->
+            {ok, Msgs, NewState};
+        Other ->
+            Other
+    end;
```
zmstone:
```suggestion
```

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1267 @@
...
+    ?BROKER_INSTR_SETMARK(t0_deliver, {_Msg#message.extra, ?BROKER_INSTR_TS()}),
+    Delivers = [Deliver | emqx_utils:drain_deliver(ActiveN)],
+    with_channel(handle_deliver, [Delivers], State);
+handle_msg({connack, ConnAck}, State) ->
+    handle_outgoing(ConnAck, State);
+handle_msg({close, Reason}, State) ->
+    %% @FIXME here it could be close due to appl error.
+    ?TRACE("SOCKET", "socket_force_closed", #{reason => Reason}),
+    handle_info({sock_closed, Reason}, close_socket(State));
+handle_msg({event, connected}, State = #state{channel = Channel}) ->
```
zmstone, line 585: this module seems to lack of the trigger for `{event, connected | disconnected | running}`

apps/emqx/src/emqx_frame.erl
keynslug: Addressed in 566e73af8f148c9d128b238c05f38f63e22af0bb.

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -0,0 +1,1269 @@
...
+        connmod => ?MODULE
+    }),
+    ShutdownPolicy = emqx_config:get_zone_conf(Zone, [force_shutdown]),
+    _ = emqx_utils:tune_heap_size(ShutdownPolicy),
+    _ = set_tcp_keepalive(Listener),
+    case sock_async_recv(Socket, 0) of
+        {ok, Data} ->
+            NState = start_idle_timer(State),
+            handle_recv({recv_more, Data}, Parent, NState);
+        {select, _SelectInfo} ->
```
zmstone, line 363: so there is no need to kee SelectInfo for the next call ?

apps/emqx/src/emqx_listeners.erl
keynslug: Most of TCP options can be updated I think, it's `tcp_backend` that cannot be changed without running `restart_listener/4`. I'll clarify the latter, unless I'm missing something.

apps/emqx/src/emqx_socket_connection.erl
keynslug: Only when running on Windows I believe. What do you suggest to do about it?

apps/emqx/src/emqx_socket_connection.erl
keynslug: Aren't those emitted on the channel level?
* https://github.com/emqx/emqx/blob/12dfd49e8b13bc442b1ebace1683c0312405aaed/apps/emqx/src/emqx_channel.erl#L1341-L1342
* https://github.com/emqx/emqx/blob/12dfd49e8b13bc442b1ebace1683c0312405aaed/apps/emqx/src/emqx_channel.erl#L1098

apps/emqx/src/emqx_socket_connection.erl
keynslug: Currently, `socket` keeps no special context in `recv`'s `select_info()`s, it's just `{select_info, recv, Handle}`. Supposedly, one can just `socket:recv(Socket, 0, ...)` again once requested number of bytes has arrived, signaled through `{$socket, Socket, select, ...}`.

Connection loop needs to disambiguate `recv` and `send` "selects" though, and for that we only keep `send` handles (in `#congested{}` structure) but not `recv` handles.

apps/emqx/src/emqx_socket_connection.erl
zmstone: maybe do not allow this backend on Window when checking schema.

zmstone (approved):

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8448)

PR #15462, opened 2025-07-01
chore(new mix build): more fixes and cleanups

thalesmg:
Part of https://emqx.atlassian.net/browse/EMQX-10365

### DIFF
--- a/apps/emqx_utils/mix.exs
+++ b/apps/emqx_utils/mix.exs
@@ -15,7 +15,7 @@ defmodule EMQXUtils.MixProject do
       lockfile: "../../mix.lock",
       elixir: "~> 1.14",
       start_permanent: Mix.env() == :prod,
-      deps: UMP.common_deps() ++ deps()
+      deps: UMP.common_deps() ++ deps() ++ test_deps()
     ]
   end
 
@@ -32,5 +32,15 @@ defmodule EMQXUtils.MixProject do
       UMP.common_dep(:snabbkaffe),
       {:erlang_qq, github: "k32/erlang_qq", tag: "1.0.0", override: true}
     ]
+  end
+
+  defp test_deps() do
+    if UMP.test_env?() do
+      [
+        UMP.common_dep(:cowboy)
+      ]
+    else
+      []
+    end
   end
 end


apps/emqx_bpapi/src/emqx_bpapi.app.src
keynslug, line 1: Since there's just a couple of files, perhaps easier to put them under `emqx/src` and `emqx/include`? And `emqx/README-bpapi.md`. Unless I'm missing why this needs to be a separate application.

Otherwise, I think it's missing `mria` and `mnesia` dependencies.

apps/emqx_mix_utils/lib/mix/tasks/emqx.ct.ex
```diff
@@ -187,19 +187,19 @@ defmodule Mix.Tasks.Emqx.Ct do
...
+  #     ["apps", app_name | _] = Path.split(suite_path)
+  #     String.to_atom(app_name)
+  #   end)
+  #   |> Enum.uniq()
+  #   |> Enum.flat_map(fn app ->
+  #     {:ok, mods} = :application.get_key(app, :modules)
+  #     mods
+  #   end)
+  #   |> Code.ensure_all_loaded()
+  # end
```
keynslug, line 202: Nit. Those few commented out blocks would probably welcome a comment explaining their purpose.

keynslug (commented):
Solid job!

apps/emqx_bpapi/src/emqx_bpapi.app.src
thalesmg: That's basically where they were before... :sweat_smile: 

Similar to the need to extract connector behaviours into their own application (https://github.com/emqx/emqx/commit/59e39623ed3dbabc10f106a9f2b2ff323faeecd4) so as to avoid circular build dependencies:

Previously:

- `emqx_bpapi` lived in `emqx` app.
- It was referenced/used by multiple `_proto_vX` modules in different applications.
- ... including modules in `emqx_durable_storage`, on which `emqx` app depends on; hence a cycle.

thalesmg: But good point about the deps, I'll fix them (I'll add just `mria`, because it's up to it to start `mnesia` as part of its initialization).

apps/emqx_mix_utils/lib/mix/tasks/emqx.ct.ex
thalesmg: I just left them here for now because I thought they might be useful while the build migration is ongoing.  We can have another clean up after it's a bit more stable to remove them later, IMO.

apps/emqx_bpapi/src/emqx_bpapi.app.src
thalesmg: :checkered_flag:

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8200)

PR #15468, opened 2025-07-02
chore(new mix build): more fixes

thalesmg:
Part of https://emqx.atlassian.net/browse/EMQX-10365

### DIFF
--- a/apps/emqx/test/emqx_limiter_shared_SUITE.erl
+++ b/apps/emqx/test/emqx_limiter_shared_SUITE.erl
@@ -10,6 +10,8 @@
 -include_lib("eunit/include/eunit.hrl").
 -include_lib("common_test/include/ct.hrl").
 
+-import(emqx_common_test_helpers, [on_exit/1]).
+
 %%--------------------------------------------------------------------
 %% Setups
 %%--------------------------------------------------------------------
@@ -27,7 +29,7 @@ end_per_suite(Config) ->
 init_per_testcase(_TestCase, Config) ->
     Config.
 
-end_per_testcase(_TestCase, Config) ->
+end_per_testcase(_TestCase, _Config) ->
     Groups = emqx_limiter_registry:list_groups(),
     lists:foreach(
         fun(Group) ->
@@ -35,7 +37,8 @@ end_per_testcase(_TestCase, Config) ->
         end,
         Groups
     ),
-    Config.
+    emqx_common_test_helpers:call_janitor(),
+    ok.
 
 %%--------------------------------------------------------------------
 %% Tests
@@ -172,6 +175,11 @@ t_change_options_from_unlimited(_) ->
     {true, _Client2} = emqx_limiter_client:try_consume(Client1, 1).
 
 t_concurrent(_) ->
+    %% This test fails if log level is debug, so it's important to ensure the correct
+    %% level is set now.
+    #{level := PrevLevel} = logger:get_primary_config(),
+    on_exit(fun() -> emqx_logger:set_primary_log_level(PrevLevel) end),
+    ok = emqx_logger:set_primary_log_level(warning),
     ok = test_concurrent(33333, 1000),
     ok = test_concurrent(333, 1000).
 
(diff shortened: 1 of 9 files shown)


apps/emqx/test/emqx_limiter_shared_SUITE.erl
```diff
@@ -172,6 +175,11 @@ t_change_options_from_unlimited(_) ->
     {true, _Client2} = emqx_limiter_client:try_consume(Client1, 1).
 
 t_concurrent(_) ->
+    %% This test fails if log level is debug, so it's important to ensure the correct
+    %% level is set now.
+    #{level := PrevLevel} = logger:get_primary_config(),
+    on_exit(fun() -> emqx_logger:set_primary_log_level(PrevLevel) end),
+    ok = emqx_logger:set_primary_log_level(warning),
```
keynslug, line 182: Nit. This doesn't feel like a good practice in general, I suspect there are more test assertions sensitive to increased timings, due to massive volume of logging. Perhaps there should be another assertion in `emqx_cth_suite:verify_clean_suite_state/1` instead?

apps/emqx_mix_utils/lib/mix/tasks/emqx.ct.ex
```diff
@@ -45,6 +45,8 @@ defmodule Mix.Tasks.Emqx.Ct do
 
     EMQX.Mix.Utils.clear_screen()
 
+    :logger.set_primary_config(:level, :warning)
```
keynslug: Nit. OTP default is actually `notice`, `warning` is `emqx_conf`s default.

keynslug (commented):
🚀

apps/emqx/test/emqx_limiter_shared_SUITE.erl
thalesmg: In this case, it was failing not because another test, but because somewhere along running the CT tasks the level was being set to `:debug`.  I suspect it might come from some app inadvertently [starting `Logger`](https://github.com/elixir-lang/elixir/blob/7b20c281d521aa7aa2ad2baa1e9ae6c579d79d0c/lib/logger/lib/logger/app.ex#L35) (maybe during `compile.all`?), but I couldn't spot exactly.

Since this particular test does depend on the assumption that log level is not debug, I made this assumption explicit here.

apps/emqx_mix_utils/lib/mix/tasks/emqx.ct.ex
thalesmg: :checkered_flag:

apps/emqx/test/emqx_limiter_shared_SUITE.erl
keynslug: > it might come from some app inadvertently [starting Logger](https://github.com/elixir-lang/elixir/blob/7b20c281d521aa7aa2ad2baa1e9ae6c579d79d0c/lib/logger/lib/logger/app.ex#L35)

Ouch! 🫠  Well, to me this sounds like something that definitely should not affect any test code then to be honest, and needs to be fixed on the build system level.

> depend on the assumption that log level is not debug

I'm trying to say that there might be few more _implicit_ assumptions this particular test case relies on, as well as other test cases in the code base. Enforcing them on testcase-by-testcase basis might prove ugly in the long term.

thalesmg: > Ouch! 🫠 Well, to me this sounds like something that definitely should not affect any test code then to be honest, and needs to be fixed on the build system level.

That's why I also set it in `emqx.ct`, to be safe, since it's part of the build system.

> I'm trying to say that there might be few more implicit assumptions this particular test case relies on, as well as other test cases in the code base. Enforcing them on testcase-by-testcase basis might prove ugly in the long term.

Fair.  I think that after [this fix](https://github.com/emqx/emqx/pull/15468/commits/3f2a943fe1e78c6714317b26a86f6a1ce8273da8), this shouldn't be needed.  No harm in leaving it here either, I believe.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8201)

PR #15473, opened 2025-07-03
chore: namespaced admins

thalesmg:
Fixes <issue-or-jira-number>

Release version:

## PR Checklist


apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -293,8 +290,14 @@ listeners() ->
 
 api_key_authorize(Req, HandlerInfo, Key, Secret) ->
     case emqx_mgmt_auth:authorize(HandlerInfo, Req, Key, Secret) of
```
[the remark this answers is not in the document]
thalesmg: Completely wrong. :joy_cat: 

Not only 3-tuples are handled [here](https://github.com/thalesmg/emqx/blob/a9e3330df4adac30128abea69a75002298c05799/apps/emqx_dashboard/src/emqx_dashboard.erl#L301), but also adding a `{error, Reason}` clause (which [already exists](https://github.com/thalesmg/emqx/blob/a9e3330df4adac30128abea69a75002298c05799/apps/emqx_dashboard/src/emqx_dashboard.erl#L309)) wouldn't handle 3-tuples.

apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -310,6 +313,24 @@ api_key_authorize(Req, HandlerInfo, Key, Secret) ->
             )
     end.
 
+jwt_token_bearer_authorize(Req, HandlerInfo, Token) ->
+    case emqx_dashboard_admin:verify_token(Req, HandlerInfo, Token) of
+        {ok, #{actor := Username} = ActorContext} ->
+            AuthnMeta = #{
```
qzhuyan, line 321: AuthnMeta deserve a type spec for it ? too imporant that `namespace` cannot be escaping.

what is the meaning of `<< >>`  namespace? could this be used to escape?

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -77,6 +75,9 @@
 
 -define(USERNAME_ALREADY_EXISTS_ERROR, <<"username_already_exists">>).
 
+-define(role, role).
+-define(namespace, namespace).
```
qzhuyan: make sure all the places use this `?namespace` ?

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -696,27 +728,41 @@ add_default_user(Username, Password) ->
...
+    Role;
+role(Role) when is_binary(Role) ->
+    Role.
 
-role(Data) ->
-    emqx_dashboard_rbac:role(Data).
+namespace_of(#?ADMIN{extra = #{?namespace := Namespace}}) when is_binary(Namespace) ->
+    Namespace;
+namespace_of(#?ADMIN{}) ->
+    undefined.
```
qzhuyan, line 768: I understand this is for compatibility, but I think it worth to have some hardenings:

return `undefined` only when `extra` is unset or `extra = #{?namespace := undefined}`

apps/emqx_dashboard/src/emqx_dashboard.erl
thalesmg: We [do have a typespec](https://github.com/thalesmg/emqx/blob/a9e3330df4adac30128abea69a75002298c05799/apps/emqx_dashboard/src/emqx_dashboard.erl#L39-L43) for it, but indeed I forgot to add `namespace` to it.

> what is the meaning of << >> namespace? could this be used to escape?

Not sure how it would escape anything.  If namespace is a binary, it's already namespaced.

We should add an extra check when attempting to create an user to ensure the namespace does exist.  But I think that can be added in a future PR.

apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -132,25 +132,31 @@ read(Name) ->
...
 
-do_update(Name, Enable, ExpiredAt, Desc, Role) ->
+do_update(Name, Enable, ExpiredAt, Desc, #{role := Role, namespace := Namespace}) ->
     case mnesia:read(?APP, Name, write) of
         [] ->
             mnesia:abort(not_found);
         [App0 = #?APP{enable = Enable0, extra = Extra0}] ->
-            #{desc := Desc0} = Extra = normalize_extra(Extra0),
+            #{desc := Desc0} = Extra0 = normalize_extra(Extra0),
+            Extra = emqx_utils_maps:put_if(
```
qzhuyan, line 149: it looks like other places handles ?namespace = `undefined` why we need `put_if` ?

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
thalesmg: Extra is always set, AFAIK.

For example:

- https://github.com/emqx/emqx/blob/e2f28ec606f5968098428ebc71a5feea12a746ea/apps/emqx_dashboard/src/emqx_dashboard_admin.erl#L225-L231
- https://github.com/emqx/emqx/blob/e2f28ec606f5968098428ebc71a5feea12a746ea/apps/emqx_dashboard/src/emqx_dashboard_admin.erl#L305

apps/emqx_management/src/emqx_mgmt_auth.erl
thalesmg: I was just avoiding writing it if there's no namespace (essentially, "global" namespace).

apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -271,14 +281,14 @@ force_create_app(App) ->
...
-            mnesia:abort(name_already_existed);
+            mnesia:abort(name_already_exists);
         [] ->
             case mnesia:match_object(?APP, #?APP{api_key = ApiKey, _ = '_'}, read) of
                 [] ->
                     ok = mnesia:write(App),
                     to_map(App);
                 _ ->
-                    mnesia:abort(api_key_already_existed)
+                    mnesia:abort(api_key_already_exists)
```
qzhuyan, line 291: Q: what if namespace is different?

thalesmg: Not sure I understand: different from what, in this context?

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
thalesmg: :checkered_flag:

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -311,25 +325,32 @@ force_add_user(Username, Password, Role, Desc) ->
...
                 pwdhash = hash(Password),
                 role = Role,
                 description = Desc,
-                extra = #{password_ts => erlang:system_time(second)}
+                extra = Extra#{password_ts => erlang:system_time(second)}
             },
             mnesia:write(Admin),
-            ?SLOG(info, #{msg => "dashboard_sso_user_added", username => Username, role => Role}),
-            flatten_username(#{username => Username, role => Role, description => Desc});
+            ?SLOG(info, #{msg => "dashboard_sso_user_added", username => Username, ?role => Role}),
```
qzhuyan:
```suggestion
            ?SLOG(info, #{msg => "dashboard_sso_user_added", username => Username, ?role => Role, ?namespace => Namespace}),
```

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -411,15 +433,26 @@ verify_password_expiration(ValidityDuration, #?ADMIN{extra = Extra} = User) ->
...
+                end,
+            mnesia:write(Admin#?ADMIN{role = Role, description = Desc, extra = NewExtra}),
             {
                 role(Admin) =:= Role,
-                flatten_username(#{username => Username, role => Role, description => Desc})
+                flatten_username(#{
+                    username => Username,
+                    role => Role,
+                    description => Desc,
+                    ?namespace => maps:get(?namespace, Extra, undefined)
```
qzhuyan: Q:   what about `NewExtra` ?

thalesmg: Good catch.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -507,7 +540,7 @@ to_external_user(UserRecord) ->
     flatten_username(#{
         username => Username,
         description => Desc,
-        role => ensure_role(Role),
+        ?role => ensure_role(Role),
         mfa => format_mfa(Extra)
```
qzhuyan, line 547: Q: what about namespace?

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -696,27 +728,41 @@ add_default_user(Username, Password) ->
...
+    Role.
 
-role(Data) ->
-    emqx_dashboard_rbac:role(Data).
+namespace_of(#?ADMIN{extra = #{?namespace := Namespace}}) when is_binary(Namespace) ->
+    Namespace;
+namespace_of(#?ADMIN{}) ->
+    undefined.
 
 flatten_username(#{username := ?SSO_USERNAME(Backend, Name)} = Data) ->
```
qzhuyan, line 770: flatten_username need to handle the namespace which is in `Data` or in field `mfa` of Data? why is that?

thalesmg: Note that this predates any changes herein.

`mfa` is a field for multi factor authentication, apparently, so it has no relation to namespaces, AFAIK.

thalesmg: ... and, looking at its definition now, seems like the purpose of flatten_username is just to transform the `username` based on whether it's an SSO user.  So, doesn't look like it needs to be touched here.  :thinking:

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
qzhuyan: Oh I thought we upgrade from old version DB which has no `Extra`.

return undefined only when extra has no  ?namespace or Extra is #{?namespace := undefined}

apps/emqx_management/src/emqx_mgmt_auth.erl
qzhuyan: If user create an APP when there is App already created by same ApiKey but in another namespace, expecially in 
`undefined` namespace. The call will be rejected here.

Do we support one apikey used for different namespace?

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -115,18 +114,31 @@ add_default_user() ->
...
+        {ok, ParsedRole} ?= parse_role(Role0),
+        #{?role := Role} = ParsedRole,
+        Extra = parsed_role_to_extra(ParsedRole),
+        do_add_user(Username, Password, Role, Desc, Extra)
     end.
 
-do_add_user(Username, Password, Role, Desc) ->
-    Res = mria:sync_transaction(?DASHBOARD_SHARD, fun add_user_/4, [Username, Password, Role, Desc]),
+parsed_role_to_extra(#{?namespace := Ns}) when is_binary(Ns) ->
+    #{?namespace => Ns};
```
qzhuyan, line 132: Q: we need `password_ts` or not?

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -411,15 +436,26 @@ verify_password_expiration(ValidityDuration, #?ADMIN{extra = Extra} = User) ->
...
 %% @private
-update_user_(Username, Role, Desc) ->
+update_user_(Username, Role, Desc, Extra) ->
     case mnesia:wread({?ADMIN, Username}) of
         [] ->
             mnesia:abort(<<"username_not_found">>);
         [Admin] ->
-            mnesia:write(Admin#?ADMIN{role = Role, description = Desc}),
+            #?ADMIN{extra = Extra0} = Admin,
+            NewExtra =
```
qzhuyan, line 445: I think we need to update `password_ts` if namespace is changed.

apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -457,37 +474,34 @@ get_role(#{role := Role}) ->
 get_role(_Desc) ->
     ?ROLE_API_DEFAULT.
 
+namespace_of(#{namespace := Namespace}) when is_binary(Namespace) ->
```
qzhuyan: nit, dup code, I see it somewhere else.

qzhuyan:
nothing blocks the merge,  I think

qzhuyan (approved):

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
thalesmg: `add_user_/5` in this module takes care of injecting that attribute.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
thalesmg: Note: `password_ts` was _not_ changed prior to the changes herein.

Looking at the attribute name and [its usage](https://github.com/emqx/emqx/blob/e2f28ec606f5968098428ebc71a5feea12a746ea/apps/emqx_dashboard/src/emqx_dashboard_admin.erl#L453), it seems to me that its intention is to track when the _password_ changes, rather than when the _user_ changes.  :thinking:

apps/emqx_management/src/emqx_mgmt_auth.erl
thalesmg: It may _seem_ duplicate, but it's an intentional separation of concerns/domains.

- Here, in `emqx_mgmt_auth`, we deal with `#?APP{}` records (API keys).
- In `emqx_dashboard_admin`, we deal with `#?ADMIN{}` records (dashboard users).
- In `emqx_dashboard_token`, we deal with `#?ADMIN_JWT{}` records (JWT tokens).

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
thalesmg: I'll add in a follow up, thanks!

apps/emqx_management/src/emqx_mgmt_auth.erl
thalesmg: It seems from the previous state of this code that it's intentional: `api_key` seems to be under a "unique" constraint (if this was a SQL db), because it's what is used to create the bearer token.  So it makes sense to be unique.

Namespace is simply an attribute of the role.  It's perfectly acceptable to have multiple API keys that have roles in the same DB.  Creating another API key referencing the same namespace as another API key won't be blocked by this code.

Note that we are not using this table to _declare_ namespaces.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
thalesmg: > return undefined only when extra has no ?namespace or Extra is #{?namespace := undefined}

Isn't that functionally equivalent?  It's `undefined` unless it's a binary value (which is the only type that makes sense for a namespace).  :thinking:

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
thalesmg: To be fixed here: https://github.com/emqx/emqx/pull/15490

apps/emqx_dashboard/src/emqx_dashboard.erl
thalesmg: Created a ticket to remember going back to this later: https://emqx.atlassian.net/browse/EMQX-14485

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8250)

PR #15483, opened 2025-07-07
fix(new mix build): use only `PROFILE` env var

thalesmg:


id (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8245)

PR #15485, opened 2025-07-07
fix(new mix build): add missing callback dependency

thalesmg:


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8246)

PR #15490, opened 2025-07-09
feat(connectors): support namespaced connectors

thalesmg:
Part of https://emqx.atlassian.net/browse/EMQX-14385

Release version: 6.0.0

## PR Checklist


apps/emqx_connector/src/emqx_connector.erl
```diff
@@ -291,28 +345,11 @@ remove(ConnectorType, ConnectorName) ->
             {error, Reason}
     end.
 
-update(ConnectorType, ConnectorName, RawConf) ->
```
thalesmg, line 294: This was only used in tests...  :melting_face:

apps/emqx_connector/src/emqx_connector.erl
```diff
@@ -291,28 +345,11 @@ remove(ConnectorType, ConnectorName) ->
...
-            );
-        Error ->
-            Error
-    end.
-
 %%----------------------------------------------------------------------------------------
 %% Data backup
 %%----------------------------------------------------------------------------------------
 
+%% TODO: namespace
```
thalesmg, line 343: To be done in a follow up PR.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -485,6 +485,8 @@ filter_result(#{} = Result) ->
         end,
         Result
     );
+filter_result(Results) when is_list(Results) ->
+    lists:map(fun filter_result/1, Results);
```
keynslug: Nit. Not this PR's change, but: function name (`filter_result/1`) sounds slightly misleading compared to what it actually does.

apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -287,6 +288,11 @@ return_unauthorized(Code, Message) ->
         },
         #{code => Code, message => Message}}.
 
+get_namespace(#{auth_meta := #{namespace := Namespace}} = _Request) when is_binary(Namespace) ->
+    Namespace;
+get_namespace(#{} = _Request) ->
+    undefined.
```
keynslug: Q: Is `undefined` essentially a separate namespace, e.g. separate from any other possible `binary()` namespace? Shouldn't it be named `default` or something similar, to make it clearer that it's perfectly valid namespace? `undefined` might be a little vague in this context.

apps/emqx_connector/src/emqx_connector_resource.erl
```diff
@@ -251,14 +291,22 @@ get_temp_conf(TypeAtom, CheckedConf) ->
...
-%% just for perform_connector_changes/1
-remove(Type, Name, _Conf, _Opts) ->
-    ?SLOG(info, #{msg => "remove_connector", type => Type, name => Name}),
-    emqx_resource:remove_local(resource_id(Type, Name)).
+    remove(_Namespace = undefined, Type, Name, #{}, #{}).
+remove(Namespace, Type, Name) ->
+    remove(Namespace, Type, Name, #{}, #{}).
+remove(Type, Name, Conf, Opts) ->
+    remove(_Namespace = undefined, Type, Name, Conf, Opts).
+remove(Namespace, Type, Name, _Conf, _Opts) ->
```
keynslug:
```suggestion
    remove(Namespace, Type, Name, #{}, #{}).

remove(Type, Name, Conf, Opts) ->
    remove(_Namespace = undefined, Type, Name, Conf, Opts).

remove(Namespace, Type, Name, _Conf, _Opts) ->
```

apps/emqx_connector/src/emqx_connector_resource.erl
```diff
@@ -99,26 +123,37 @@ connector_hookpoint_to_connector_id(_) ->
     {error, bad_connector_hookpoint}.
 
 restart(Type, Name) ->
-    emqx_resource:restart(resource_id(Type, Name)).
+    restart(_Namespace = undefined, Type, Name).
+restart(Namespace, Type, Name) ->
+    ConnResId = resource_id(Namespace, Type, Name),
+    emqx_resource:restart(ConnResId).
```
keynslug:
```suggestion
restart(Type, Name) ->
    restart(_Namespace = undefined, Type, Name).

restart(Namespace, Type, Name) ->
    ConnResId = resource_id(Namespace, Type, Name),
    emqx_resource:restart(ConnResId).
```

However, from the security perspective, generally speaking, it might be better to make callers aware of namespaces, by removing `restart/2` variant and its friends.

apps/emqx_connector/src/emqx_connector_api.erl
```diff
@@ -419,24 +434,29 @@ schema("/connectors_probe") ->
...
-    of
+lookup_from_all_nodes(Namespace, ConnectorType, ConnectorName, SuccCode) ->
+    Nodes = emqx_bpapi:nodes_supporting_bpapi_version(?BPAPI_NAME, 2),
+    Res = emqx_connector_proto_v2:lookup(
+        Nodes,
+        Namespace,
+        ConnectorType,
+        ConnectorName
+    ),
+    case is_ok(Res) of
```
keynslug, line 460: Q. To clarify: does this mean that regular non-namespaced lookups will behave weirdly in heterogenous / not-fully-upgraded clusters?

apps/emqx_connector/src/emqx_connector_api.erl
```diff
@@ -832,3 +858,13 @@ maybe_focus_on_request_connector(Reason0, Type0, Name0) ->
         _ ->
             Reason0
     end.
+
+get_raw_config(Namespace, KeyPath, Default) when is_binary(Namespace) ->
+    emqx:get_raw_config({Namespace, KeyPath}, Default);
```
keynslug: This API feels overloaded, especially since 2 variants are backed by 2 different mechanisms. Or, since there are already `get_namespaced/2` APIs IIUC, my suggestion to use them directly and to drop variants with namespaced keys (like that one), basically to make API surface smaller, more straightforward and harder to (accidentally) misuse. Which to me also sounds better from security perspective.

apps/emqx_connector/src/emqx_connector.erl
```diff
@@ -560,3 +599,18 @@ multi_validate_connector_names(Conf) ->
                 bad_connectors => BadConnectors
             }}
     end.
+
+ssl_cert_dir(Type, Name, #{namespace := Namespace}) when is_binary(Namespace) ->
```
keynslug:
```suggestion
ssl_cert_dir(Namespace, Type, Name) when is_binary(Namespace) ->
```

To make it more explicit and, more importantly, uniform with the rest of functions in this module.

apps/emqx_connector/src/emqx_connector.erl
```diff
@@ -223,19 +262,21 @@ list() ->
...
 lookup(Type, Name) ->
-    RawConf = emqx:get_raw_config([connectors, Type, Name], #{}),
-    lookup(Type, Name, RawConf).
-
-lookup(Type, Name, RawConf) ->
-    case emqx_resource:get_instance(emqx_connector_resource:resource_id(Type, Name)) of
+    lookup(_Namespace = undefined, Type, Name).
+lookup(Namespace, Type, Name) ->
+    RawConf = get_raw_config(Namespace, [connectors, Type, Name], #{}),
+    do_lookup(Namespace, Type, Name, RawConf).
```
keynslug: Nit.
```suggestion
    lookup(_Namespace = undefined, Type, Name).

lookup(Namespace, Type, Name) ->
    RawConf = get_raw_config(Namespace, [connectors, Type, Name], #{}),
    do_lookup(Namespace, Type, Name, RawConf).
```

But again, gently suggesting to just make callers aware of namespaces by dropping `lookup/2`.

apps/emqx/src/emqx_schema_hooks.erl
```diff
@@ -7,9 +7,9 @@
 -type hookpoint() :: atom().
 
 -callback injected_fields() ->
-    #{hookpoint() => [hocon_schema:field()]}.
+    #{hookpoint() => [hocon_schema:field() | binary()]}.
```
keynslug, line 10: This change sounds unclear without an explanation (e.g. a comment): what does "injecting a string into the schema" can mean, when the function is clearly for "injecting fields"? Perhaps there's some clearer way to express it?

apps/emqx_dashboard/src/emqx_dashboard.erl
thalesmg: Yes, it means the "global" namespace (which is the only one we have today), as namespaces must have binary names (from the multi-tenancy feature).  This builds on recent changes ([[1]](https://github.com/emqx/emqx/pull/15473), [[2]](https://github.com/emqx/emqx/pull/15377)) for the same epic.

Giving it a more semantic value does make sense.

apps/emqx_connector/src/emqx_connector_resource.erl
thalesmg: I grouped what are basically clauses with default values together, since they defer to the next line(s) immediately, and `erlfmt` seemed to accept it.  :thinking:

apps/emqx_connector/src/emqx_connector_api.erl
thalesmg: Yes.  We don't support changing configurations during rolling upgrades, so it should not be a big problem to have a transient state during rolling upgrades, specially since it's only "visual"?

apps/emqx_connector/src/emqx_connector_api.erl
thalesmg: The idea here is to be convenient: we need to check if the API to be called is namespaced or not.  We do need to change the way `get_namespaced/2` APIs are called because it non-namespaced/namespaced calls are different (tuple argument or not).

apps/emqx/src/emqx_schema_hooks.erl
thalesmg: Good question.  This is a schema thing being injected, though it's not a whole field, but only the root key names as binaries.  :thinking:

apps/emqx_connector/src/emqx_connector_resource.erl
thalesmg: I was avoid dropping those arities because it will probably require touching several test suites that (wrongly, mostly?) use those APIs directly and older BPAPI calls, but it's probably best to do it now anyway.

apps/emqx_connector/src/emqx_connector_api.erl
keynslug: I understand the idea, but in this case it makes this API (`emqx:get_raw_config/1`) too broad. Now for example Dialyzer won't help if someone _accidentally_ passes a tuple as a first argument, where it might have highlighted it as a security bug otherwise.

Strictly IMO, but sometimes convenience might hurt because there's now seemingly 2 (or more) different ways to do similar thing and then you have to understand if they are indeed identical and decide which one to pick (e.g. as someone who sees those APIs for the first time).

I might be missing some details to grasp the rest of the comment, isn't this `get_raw_config/3` here is already overloaded, so it's a matter of changing the body to `emqx_config:get_raw_namespaced(KeyPath, Namespace, Default);`?

apps/emqx_connector/src/emqx_connector_api.erl
keynslug: I don't have good answer to that question. 🌝 I'm inclined towards "it's not a big problem" just as you, but felt it might be important to highlight that somewhere.

apps/emqx_connector/src/emqx_connector_resource.erl
keynslug: Yeah, `erlfmt` seems okay with it, so please take as a minor nit. My internal parser had hard time understanding that these are _different functions_ rather than _separate clauses_ because they have identical names and are grouped together. 🫠

apps/emqx/src/emqx_schema_hooks.erl
thalesmg: Maybe the callback should be more generally called `injected_values`, to indicate it's simply a list of values.

keynslug: Right, I see now, it's not actually about _schema_ per se. In this case it's basically a whitelist, so it sounds slightly misleading to carry it through `injected_fields`, as is renaming it to `injected_values` to be honest. I guess to make it make more sense here, we'd need something like a whole _separate schema_ for namespaced roots, consisting of a small subset of roots of the global schema. Due to complexity, this does not sound like a good idea in the context of this PR, and I don't have a better one. Maybe @savonarola might have some?

apps/emqx_dashboard/src/emqx_dashboard_api.erl
thalesmg: :checkered_flag:

apps/emqx_connector/src/emqx_connector_api.erl
thalesmg: IIUC that the suggestion is to remove the convenience from `emqx:_` helpers and let call sites use the `emqx_config:_` fns directly, then this is in conflict with [the original suggestion](https://github.com/emqx/emqx/pull/15377#discussion_r2178003322).  :melting_face: 

@savonarola WDYT?

apps/emqx_dashboard/src/emqx_dashboard.erl
thalesmg: :checkered_flag:

apps/emqx_connector/src/emqx_connector.erl
thalesmg: :checkered_flag:

apps/emqx_connector/src/emqx_connector_api.erl
keynslug: Heh I see. 🫠 To clarify I'm not suggesting to remove `emqx:get_raw_config/1`. Having _separate_ `emqx:get_namespaced_config/1,2` and `emqx:get_namespaced_raw_config/1,2` should be fine I think.

thalesmg: Hmmm it still sounds like reverting the convenience changes above, because we already have `emqx_config:get{,_raw}_namespaced` fns.  :thinking: 

... and just increasing the arity of `emqx:get_config` and similar can be more confusing due to the variants with and without default values.

keynslug: I'm confused. 🫠 I'll try to reiterate.

My suggestion is to remove `emqx:get_raw_config({Namespace, KeyPath}, ...)` overload, in other words disallow tuples as first argument.

I don't think it conflicts with the original suggestion of
> to have the respective facade functions in the emqx module

...Which in this case amounts to introducing `emqx:get_namespaced_config/1,2` and `emqx:get_raw_namespaced_config/1,2` _facades_, which call respective `emqx_config` module API functions.

apps/emqx_connector/src/emqx_connector_api.erl
thalesmg: Actually, non-namespaced lookups should _not_ behave "weirdly", given that we don't support changing config during rolling upgrades.  Since there shouldn't be any config changes, and assuming the config/state was not inconsistent before the rolling upgrade, the connectors should all be the same.  Granted, the only thing it won't be able to notice is if the status on one or more old nodes changes (e.g., becomes disconnected).

apps/emqx_connector/src/emqx_connector_api.erl
thalesmg: > My suggestion is to remove emqx:get_raw_config({Namespace, KeyPath}, ...) overload, in other words disallow tuples as first argument.

That was my understanding of your suggestion.  Indeed it conflicts with the [original suggestion](https://github.com/emqx/emqx/pull/15377#discussion_r2178003322) that prompted its introduction.

> ...Which in this case amounts to introducing emqx:get_namespaced_config/1,2 and emqx:get_raw_namespaced_config/1,2 facades, which call respective emqx_config module API functions.

That is a different proposition, I think.  I wonder what @savonarola thinks about that, since I've already discussed this with him and it seemed fine to have the tuple argument for convenience.

apps/emqx_connector/src/emqx_connector_api.erl
keynslug: Right. Should be less of a concern then, thanks for clarifying!

apps/emqx_connector/src/emqx_connector_api.erl
thalesmg: Changed this anyway.  I hope @savonarola doesn't disagree.  :sweat:

apps/emqx/include/emqx_config.hrl
```diff
@@ -0,0 +1,9 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-ifndef(EMQX_CONFIG_HRL).
+-define(EMQX_CONFIG_HRL, true).
+
+-define(global_ns, global).
```
keynslug, line 7: Minor nit / discussion. To me, "global" sounds like a space where everything should be visible, including things from non-"global" spaces.

keynslug (approved):
🔥 LGTM besides https://github.com/emqx/emqx/pull/15490#discussion_r2204301051, guess it's fine to sort this one out in a followup.

thalesmg:
> 🔥 LGTM besides [#15490 (comment)](https://github.com/emqx/emqx/pull/15490#discussion_r2204301051), guess it's fine to sort this one out in a followup.

Indeed, my understanding is that this is a separate thing that we can deal in a follow up.

Though, the particular solution of defining a schema module whose roots is just a subset of the main schema might be a little tricky because, at least in theory, we _could_ have different schema modules (as we had with CE), and whatever is set at build time should be filtered.  If we drop that assumption, things could be simplified for this proposed approach.

Though I also prefer to do it separately than here, whatever solution we choose.  :sweat_smile:

apps/emqx/include/emqx_config.hrl
thalesmg: Hmmm I see how it may also be interpreted that way.

How about `no_ns, no_ns`?

keynslug: Could work. Another example I have is from k8s: there's `default` namespace when not explicitly mentioned.

thalesmg: Could work too.

Thinking about the multi-tenancy feature, though, I feel like `default` gives the impression of _a_ namespace, while the "global" / "default" namespace is actually the _lack of a namespace_.  Which is the current state of affairs, before this feature lands.  Managed namespaces are declared explicitly.  So I'd favor a name that conveys better the idea "lack of namespace", "no namespace", etc., if a good one does exist.  :thinking:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8427)

apps/emqx_connector/src/emqx_connector.erl
thalesmg: https://github.com/emqx/emqx/pull/15609

PR #15491, opened 2025-07-09
chore(new mix build): make custom compilers less verbose by default

thalesmg:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8248)

PR #15492, opened 2025-07-10
feat(bigquery): check for table/dataset access on each health check

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14475

Release version: 6.0.0

## PR Checklist

- [na] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files (unreleased feature)

### DIFF
--- a/apps/emqx_bridge_bigquery/src/emqx_bridge_bigquery_impl.erl
+++ b/apps/emqx_bridge_bigquery/src/emqx_bridge_bigquery_impl.erl
@@ -50,10 +50,12 @@
 
 %% Fields
 -define(client, client).
+-define(dataset, dataset).
 -define(insert_all_path, insert_all_path).
 -define(installed_channels, installed_channels).
 -define(project_id, project_id).
 -define(request_ttl, request_ttl).
+-define(table, table).
 
 -type connector_config() :: #{}.
 -type connector_state() :: #{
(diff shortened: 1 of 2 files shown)


apps/emqx_bridge_bigquery/src/emqx_bridge_bigquery_impl.erl
```diff
@@ -175,10 +188,26 @@ on_remove_channel(_ConnResId, ConnState, _ChanResId) ->
...
+    #{
+        ?table := Table,
+        ?dataset := Dataset,
+        ?request_ttl := RequestTTL
+    } = maps:get(ChanResId, InstalledChannels),
+    Opts = #{?request_ttl => RequestTTL},
+    case get_table(Client, ProjectId, Dataset, Table, Opts) of
+        {ok, _} ->
+            ?status_connected;
+        {error, #{status_code := 404}} ->
```
qzhuyan, line 203: nit, I guess the restapi response contains error messages, No? 

A bit risky that we may return misleading info to user.

thalesmg: Not sure I understand.  Here, we are attempting to `GET` the table directly, so if the API returns 404, it means "not found".

It's true that BigQuery may return 404 if the service account being used does not have enough permissions, even if it in fact exists.  Though, there's nothing we can do about that, and the [error message just says](https://emqx.atlassian.net/browse/EMQX-14472?focusedCommentId=55867) what the status code already conveys.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8249)

PR #15501, opened 2025-07-11
feat(new mix build/ct): add support for ct specs

thalesmg:
```
mix ct --spec apps/emqx_bridge_s3tables/test/some.spec
```


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8276)

PR #15510, opened 2025-07-14
fix(new mix build): print whole xref output

thalesmg:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8299)

PR #15512, opened 2025-07-14
test(s3tables): disable flaky scenario

thalesmg:
Even though now the retry/restart workarounds seem to be working properly, this test case is still very flaky and may fail 5 times in a row in CI.  I believe we have a net gain disabling it, even though we'll stop testing this very particular scenario.

e.g.:

https://github.com/emqx/emqx/actions/runs/16274083953/job/45949204895?pr=15490


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8300)

PR #15518, opened 2025-07-16
fix(shared-sub): stop monitoring remote subscribers to avoid races

keynslug:
Fixes [EMQX-14477](https://emqx.atlassian.net/browse/EMQX-14477).

Release version: 6.0.0-M1.202507

## Summary

Prior to these changes, multiple EMQX nodes could in general react simultaneously to a single shared subscriber going down. If, due to load skew, remote node reacted to this event earlier than local node, leftover state started to accumulate. This would have eventually manifested in global routing table inconsistencies.

This PR changes `emqx_shared_sub` such that there's no monitoring of subscribers anymore. Local node is the only node responsible for cleaning up after local subscribers going down, as signaled by `emqx_broker_helper` that is now the component responsible for monitoring subcribers. In case whole node goes down, responsibility for cleaning up is shifted to `emqx_router_helper` and now follows the common logic. This breaks dependency chain a little, will be refactored in a follow up PR.

## PR Checklist

[EMQX-14477]: https://emqx.atlassian.net/browse/EMQX-14477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_broker_helper.erl
+++ b/apps/emqx/src/emqx_broker_helper.erl
@@ -267,7 +267,6 @@ collect_messages(Reg, Down) ->
 %% Collect both register_sub and 'DOWN' messages in a loop.
 %% There is no other message sent to this process, so the
 %% 'receive' should not have to scan the mailbox.
-%% TODO: Double-check.
 collect_messages(Reg, Down, 0) ->
     {Reg, Down};
 collect_messages(Reg, Down, N) ->
--- a/apps/emqx/src/emqx_broker_helper.erl
+++ b/apps/emqx/src/emqx_broker_helper.erl
@@ -255,10 +255,10 @@ code_change(_OldVsn, State, _Extra) ->
 %%--------------------------------------------------------------------
 
 collect_and_handle(Reg0, Down0) ->
-    {Reg, Down} = collect_messages(Reg0, Down0),
+    {Regs, Down} = collect_messages(Reg0, Down0),
     %% handle register before handle down to avoid race condition
     %% because down message is always the last one from a process
-    ok = handle_register(Reg),
+    ok = handle_registrations(Regs),
     ok = handle_down(Down).
 
 collect_messages(Reg, Down) ->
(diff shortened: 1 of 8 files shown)


apps/emqx/src/emqx_broker_helper.erl
```diff
@@ -267,6 +267,7 @@ collect_messages(Reg, Down) ->
 %% Collect both register_sub and 'DOWN' messages in a loop.
 %% There is no other message sent to this process, so the
 %% 'receive' should not have to scan the mailbox.
+%% TODO: Double-check.
```
thalesmg: Q: stale comment?

apps/emqx/src/emqx_shared_sub.erl
```diff
@@ -410,43 +420,43 @@ subscribers(Group, Topic) ->
...
-    {ok, update_stats(#state{pmon = PMon})}.
-
+    %% There may be leftovers from older incarnation of this node, clean them up:
+    cleanup_node(node()),
+    %% Let `emqx_stats` periodically update stats instead of relying on mnesia events:
+    ok = emqx_stats:update_interval(?MODULE, fun ?MODULE:update_stats/0),
+    {ok, #state{pmon = pmon_new()}}.
+
+%% Deprecated, left for backwards compatibility.
+%% Should be removed in the next release.
```
thalesmg, line 454: Q: if we no longer run a transaction with this, and it's not a RPC target, why do we need to keep it?

I know the comment on the `export` directive says it's an "RPC target", but I couldn't find any references to `init_monitors` other than in this module.  :thinking:

thalesmg (commented):
(still reading)

apps/emqx/src/emqx_shared_sub.erl
keynslug: Hmm, good point. My thought was there still may be replicants running older release that can hit this RPC, but:
* It's super unlikely, since it's initialization routine (perhaps, if there's a crash).
* Setting up process monitors in a process executing transaction on a core does not make sense anyway. 🫠

keynslug: ...However, it still may be safer to keep it. WDYT?

keynslug:
@thalesmg Made few small last minute changes, sorry about disrupting the review.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8342)

PR #15525, opened 2025-07-17
feat(schema registry): forbid removal of serde if referenced

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14443

## Summary

image

## PR Checklist


zmstone:
how about the usages in rules ?

zmstone (approved):

thalesmg:
> how about the usages in rules ?

The original ticket only complained about later editing the dependent schema validation/message transformation, which breaks.

I think rules will just fail during runtime, as they do today; shouldn't prevent them from being edited, I think.

EDIT: just tested, indeed one can create/edit rules referencing anything.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8434)

PR #15534, opened 2025-07-18
test(raft): mark test cases as flaky

thalesmg:
These two test cases (`t_rebalance`, `t_crash_restart_recover`) have been failing a lot in CI.

Having a proper fix for the cause of their flakiness is obviously the ideal correct solution.  However, until that happens, we could reduce the time that is wasted re-running CI by retrying only them a bit instead of the whole workflow run, which is quite slow.

Ex: https://github.com/emqx/emqx/actions/runs/16377217654/job/46282627060#step:5:5939


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8435)

PR #15544, opened 2025-07-22
feat: datalayers using Arrow Flight SQL driver

JimMoen:
Fixes [EMQX-13706](https://emqx.atlassian.net/browse/EMQX-13706)

Release version: 6.0.0

## PR Checklist

[EMQX-13706]: https://emqx.atlassian.net/browse/EMQX-13706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.ci/docker-compose-file/docker-compose-toxiproxy.yaml
```diff
@@ -57,8 +57,10 @@ services:
       - 4000:4000
       - 4001:4001
       # Datalayers
+      - 8360:8360
       - 8361:8361
       - 8362:8362
+      - 8363:8363
```
thalesmg: IMO it's better to avoid adding more ports to bind on the host machine.

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
```diff
@@ -0,0 +1,571 @@
...
+    ok.
+
+on_get_channels(InstId) ->
+    emqx_bridge_v2:get_channels_for_connector(InstId).
+
+on_get_status(InstId, _State) ->
+    health_check(InstId).
+
+on_get_channel_status(InstId, _ChannelId, _State) ->
+    %% TODO: Check the status for each channel.
```
thalesmg, line 235: Q: should this be done in this PR, or in a follow up?

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
```diff
@@ -0,0 +1,571 @@
...
+            }),
+            {error, Reason}
+    end.
+
+on_stop(InstId, _State) ->
+    ?SLOG(info, #{
+        msg => "stoping_datalayers_arrow_flight_connector",
+        connector => InstId
+    }),
+    %% TODO: close all channels prepared statements
```
thalesmg, line 152: Q: should this be done in this PR, or in a follow up?

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers.erl
```diff
@@ -136,15 +191,30 @@ fields(action) ->
...
         emqx_bridge_datalayers_connector:precision_field()
     ];
+fields(action_parameters_arrow_flight) ->
+    [
+        {sql,
+            mk(
+                emqx_schema:template(),
+                #{desc => ?DESC(sql_template), default => ?DEFAULT_SQL, format => <<"sql">>}
+            )},
+        emqx_bridge_v2_schema:undefined_as_null_field()
```
JimMoen: nit: remove this option for datalayers data integration.
should always be `null` for fields not found.

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
```diff
@@ -0,0 +1,571 @@
...
+    Query = [InsertPart, <<" values ">> | lists:join($,, Rows)] ++ [<<" on ">>, OnClause],
+    Query;
+render_batch_sql_row({InsertPart, RowTemplate}, Querys, ChannelState) ->
+    Rows = [render_row(RowTemplate, Msg, ChannelState) || {_, Msg} <- Querys],
+    Query = [InsertPart, <<" values ">> | lists:join($,, Rows)],
+    Query.
+
+render_row(RowTemplate, Data, ChannelState) ->
+    RenderOpts =
+        case maps:get(undefined_vars_as_null, ChannelState, false) of
```
JimMoen: rm it.

scripts/ct/run.sh
```diff
@@ -327,7 +330,7 @@ else
     if [ -z "${REBAR3CT:-}" ]; then
         docker exec -e IS_CI="$IS_CI" \
                     -e PROFILE="$PROFILE" \
-                    -e SUITEGROUP="${SUITEGROUP:-}" \
+                    -e SUITES="${SUITES:-}" \
```
thalesmg: this probably shouldn't be touched here.  it's used by `find-suites.sh` in the makefile

JimMoen: checked the script, `SUITEGROUP` should be kept here. 👍
add `SUITES` for running sepecific (or just single) SUITE.
e.g. `env SUITES='foo_SUITE.erl' ./scripts/ct/run.sh --app apps/bar --keep-up`

.ci/docker-compose-file/docker-compose-toxiproxy.yaml
JimMoen: Datalayers do not support listening on both tcp and tls ports simultaneously.

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
JimMoen: in a follow up

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
JimMoen: in a follow up

.ci/docker-compose-file/docker-compose-toxiproxy.yaml
thalesmg: I don't think what I said is related to that. :sweat_smile: 

It was about binding ports on the host machine (specially when using `run.sh` to test locally).

scripts/ct/run.sh
thalesmg: If you are running specific tests locally, I suggest simply entering a shell inside the container and running them directly.

i.e.

```sh
env DOCKER_CT_RUNNER_IMAGE=ghcr.io/emqx/emqx-builder/5.5-4:1.18.3-27.3.4.2-1-ubuntu24.04 PROFILE=emqx-enterprise scripts/ct/run.sh --app apps/emqx_bridge_datalayers --attach
make apps/emqx_bridge_datalayers-ct
```

mix.exs
```diff
@@ -286,6 +286,14 @@ defmodule EMQXUmbrella.MixProject do
     # TODO: {:proper, "1.5.0"}, when it's published to hex.pm
     do: {:proper, github: "proper-testing/proper", tag: "v1.5.0", override: true}
 
+  def common_dep(:datalayers),
```
thalesmg: Nit: this is not a "common" dependency right now, since it's only used by datalayers; so I suggest moving it to datalayer's `mix.exs`.

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_connector_SUITE.erl
thalesmg, line 1: Please don't add those `_connector_SUITE`s with those "lifecycle" tests which perform direct calls to the module APIs.  This makes it harder to later do refactorings.  It's preferable to use only the HTTP APIs to create/edit/delete the connectors/actions/rules.

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_SUITE.erl
```diff
@@ -1045,6 +1046,8 @@ t_write_failure(Config) ->
     ),
     ok.
 
+%% XXX: this required the previous test to be run first.
+%% otherwise the table `mqtt` won't be created.
```
thalesmg, line 1050: We shouldn't depend on a specific order tests are executed.  Better to do the required setup/teardown for each test case as appropriate.

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers.erl
```diff
@@ -29,32 +29,60 @@
...
-                    Method, datalayers, datalayers, ParamsExample
+                    Method, datalayers, datalayers, ParamsExampleInflux
+                )
+            }
+        },
+        #{
+            <<"datalayers_arrow_flight">> => #{
+                summary => <<"Datalayers Action by Arrow Flight SQL Driver">>,
+                value => emqx_bridge_v2_schema:action_values(
+                    Method, datalayers, datalayers_arrow_flgiht, ParamsExampleArrowFlight
```
thalesmg:
```suggestion
                    Method, datalayers, datalayers_arrow_flight, ParamsExampleArrowFlight
```

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
```diff
@@ -0,0 +1,566 @@
...
+    %% TODO: Check the status for each channel.
+    %% e.g.
+    %% - check the prepared statements
+    %% - check sql database/table existed
+    health_check(InstId).
+
+health_check(InstId) ->
+    case do_health_check(InstId) of
+        true -> ?status_connected;
+        false -> ?status_disconnected
```
thalesmg: Please add a more descriptive reason for the disconnected state.  We have an [ongoing ticket](https://emqx.atlassian.net/browse/EMQX-14024) to try to improve such descriptions, so it's easier for the user to know what's going on.

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
```diff
@@ -0,0 +1,566 @@
...
+) ->
+    {ok, _Result} | {error, _Reason}.
+do_query(
+    InstId, QueryMode, [{ChannelId, _Data} | _] = Querys, #{channels := Channels} = State, InitArgs
+) ->
+    ChannelState = maps:get(ChannelId, Channels),
+    EnablePrepared = enable_prepared(State, ChannelState),
+    case get_template(EnablePrepared, ChannelState) of
+        undefined ->
+            {error, no_template_fonud};
```
thalesmg:
```suggestion
            {error, {unrecoverable_error, no_template_found}};
```

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
```diff
@@ -0,0 +1,566 @@
...
+pick_and_do(InstId, FuncMode, DriverFunArgs) ->
+    try
+        ecpool:pick_and_do(
+            InstId, {?MODULE, call_driver, [FuncMode, DriverFunArgs]}, no_handover
+        )
+    of
+        {error, _Reason} = Err ->
+            Err;
+        Result ->
+            Result
```
thalesmg: Nit:

```suggestion
```

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
```diff
@@ -0,0 +1,566 @@
...
+    map_size(Templates) == 0
+->
+    ChannelState;
+init_prepare(ChannelState = #{}) ->
+    case prepare_sql(ChannelState) of
+        {ok, PrepStatements} when is_map(PrepStatements) ->
+            %% each client holds a prepared statements reference
+            ChannelState#{prepared_refs => PrepStatements};
+        _Error ->
+            ChannelState#{prepared_refs => {error, 'TODO_ERRORS'}}
```
thalesmg: Lingering `todo`

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_connector.erl
```diff
@@ -158,14 +180,23 @@ fields("connector") ->
...
+        {driver_type,
+            mk(enum([?DATALAYERS_DRIVER_TYPE_INFLUX, ?DATALAYERS_DRIVER_TYPE_ARROW_FLIGHT]), #{
+                required => false,
+                default => ?DATALAYERS_DRIVER_TYPE_INFLUX,
+                desc => ?DESC("driver_type")
+            })},
+        {database, mk(binary(), #{required => true, desc => ?DESC("database")})},
+        {username, mk(binary(), #{desc => ?DESC("username")})},
+        {password, emqx_schema_secret:mk(#{desc => ?DESC("password")})},
+        {enable_prepared, fun enable_prepared/1}
```
thalesmg: Q: why not use a simple map for the field schema here?

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_SUITE.erl
```diff
@@ -0,0 +1,461 @@
...
+    <<
+        "INSERT INTO mqtt_arrow.common_test(ts, msgid, clientid, topic, qos, payload) VALUES "
+        "(${timestamp}, ${id}, ${clientid}, ${topic}, ${qos}, ${payload})"
+    >>.
+
+parse_and_check_connector(ConfigString, Name) ->
+    {ok, RawConf} = hocon:binary(ConfigString, #{format => map}),
+    hocon_tconf:check_plain(emqx_connector_schema, RawConf, #{required => false, atom_key => false}),
+    #{<<"connectors">> := #{<<"datalayers">> := #{Name := Config}}} = RawConf,
+    Config.
```
thalesmg: Note that the function here does not assert that it processed the correct config root.

Please use `emqx_bridge_v2_testlib:parse_and_check_connector`.

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_SUITE.erl
```diff
@@ -0,0 +1,461 @@
...
+
+parse_and_check_action(ConfigString, Name) ->
+    {ok, RawConf} = hocon:binary(ConfigString, #{format => map}),
+    hocon_tconf:check_plain(emqx_bridge_v2_schema, RawConf, #{required => false, atom_key => false}),
+    #{<<"actions">> := #{<<"datalayers">> := #{Name := Config}}} = RawConf,
+    Config.
```
thalesmg: Note that the function here does not assert that it processed the correct config root.

Please use `emqx_bridge_v2_testlib:parse_and_check`.

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_SUITE.erl
```diff
@@ -0,0 +1,461 @@
...
+    emqx_bridge_v2_testlib:create_bridge_api(Config, Overrides).
+
+query_resource(Config, Request) ->
+    Type = ?config(bridge_type, Config),
+    Name = ?config(bridge_name, Config),
+    BridgeV2Id = id(Type, Name),
+    ConnectorResId = emqx_connector_resource:resource_id(Type, Name),
+    emqx_resource:query(BridgeV2Id, Request, #{
+        timeout => 1_000, connector_resource_id => ConnectorResId
+    }).
```
thalesmg: Please avoid calling `query` directly.  Prefer to exercise it via rules, as an MQTT client, whenever possible.

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_SUITE.erl
```diff
@@ -0,0 +1,461 @@
...
+    ProxyPort = ?config(proxy_port, Config),
+    ProxyHost = ?config(proxy_host, Config),
+    ProxyName = ?config(proxy_name, Config),
+    {ok, _} = create_bridge(Config),
+    ConnectorId = emqx_connector_resource:resource_id(datalayers, ?config(connector_name, Config)),
+    ?assertEqual({ok, connected}, emqx_resource_manager:health_check(ConnectorId)),
+    emqx_common_test_helpers:with_failure(down, ProxyName, ProxyHost, ProxyPort, fun() ->
+        ?assertEqual({ok, disconnected}, emqx_resource_manager:health_check(ConnectorId))
+    end),
+    ok.
```
thalesmg: See `emqx_bridge_v2_testlib:t_on_get_status`, which does this in a more standardized way.

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_connector_SUITE.erl
JimMoen: seems using `emqx_bridge_v2_testlib:t_*` functions is enough?

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_SUITE.erl
JimMoen: marked a TODO to fix the flaky test.
[EMQX-14576](https://emqx.atlassian.net/browse/EMQX-14576)

.ci/docker-compose-file/docker-compose-toxiproxy.yaml
JimMoen: 🤔 
This seems necessary for now. Perhaps we could add container IP detection to the script?
Something like this:
https://github.com/datalayers-io/datalayers-adapter-erl/blob/main/.github/workflows/_test_entrypoint.yaml#L117-L120

I'd try this in the follow-up PR.

scripts/ct/run.sh
JimMoen: It still runs all suites in the application.
You still need to set the `SUITES` environment variable. If that's the case, why not put it outside?
I'm trying this to run a single SUITE file:
https://github.com/emqx/emqx/blob/master/Makefile#L126

.ci/docker-compose-file/docker-compose-toxiproxy.yaml
thalesmg: Why is this necessary?

I run all tests within `run.sh` without the need to bind ports in the host machine.  As long as you run the tests _inside docker_ too, the containers can communicate without reserving ports on the host.  :thinking:

scripts/ct/run.sh
thalesmg: Not necessary to put it outside if you go into the container as I mentioned above.

```sh
env DOCKER_CT_RUNNER_IMAGE=ghcr.io/emqx/emqx-builder/5.5-4:1.18.3-27.3.4.2-1-ubuntu24.04 PROFILE=emqx-enterprise scripts/ct/run.sh --app apps/emqx_bridge_datalayers --attach

# you can specify the test suites here, of course; this is to be executed inside the `erlang` container
make ct-suite SUITES=apps/emqx/test/emqx_SUITE.erl
```

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_connector_SUITE.erl
thalesmg: yes, as long as it does use the HTTP APIs, and does not make direct calls to the impl modules, or uses `emqx_resource:query` and others without good reasons

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8618)

PR #15550, opened 2025-07-23
set 1 replica as default in helm chart

id:
clustered setup requires a license

Fixes an issue when all nodes except one will be crashing if a user deployes helm chart as is.

Release version: 5.9.2, 5.10.1, 6.0.0-M2.202508, 6.0.0

## PR Checklist


id:
See #15553 instead.

PR #15557, opened 2025-07-23
test(takeover): mark test cases as flaky

thalesmg:
These test cases have been failing a lot in CI.

Having a proper fix for the cause of their flakiness is obviously the ideal correct solution.  However, until that happens, we could reduce the time that is wasted re-running CI by retrying only them a bit instead of the whole workflow run.


keynslug:
Hopefully, this will help: https://github.com/emqx/emqx/pull/15565.

PR #15587, opened 2025-07-25
feat(snowflake): detach streaming mode to separate connector type

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14523

## Summary

The new Snowflake Snowpipe Streaming APIs do not require ODBC connections, which simplify the setup required to use Snowflake integration.  Since the connection is so different from the aggregated mode from existing integration, it's cleaner to make a new connector type than to graft streaming as an Action mode onto the old connector.

The previously added Streaming Action mode was only released in 6.0.0-M1, an internal release.  So, although this is a breaking schema change (we're dropping `mode = streaming` from `snowflake` Action type), it should be fine, as it's not publicly released.

A couple common functions used in both connectors are moved to `emqx_bridge_snowflake_lib` for shared usage.

## PR Checklist

- [x] The changes are covered with new or existing tests (ran test suites against real snowflake)
- [na] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files (previous streaming mode was only in internal release)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8488)

PR #15594, opened 2025-07-28
feat(trace): expose "max number of traces" as config option

keynslug:
Fixes [EMQX-14467](https://emqx.atlassian.net/browse/EMQX-14467).

Release version: 6.0.0

## Summary

This PR also includes:
* Refactoring effort to make core, data layer and API concerns slightly better separated into their own modules.
* Attempt to avoid unconstrained atom creation through reusing integer "slots" for `logger`-level handler IDs.
    Notably, the latter _does not_ cover Traces CLI.

## PR Checklist

[EMQX-14467]: https://emqx.atlassian.net/browse/EMQX-14467?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_trace/emqx_trace.erl
+++ b/apps/emqx/src/emqx_trace/emqx_trace.erl
@@ -265,28 +265,20 @@ list_traces(Enable) ->
 -spec create(trace()) ->
     {ok, trace()}
     | {error,
-        {duplicate_condition, iodata()}
-        | {already_existed, iodata()}
+        {already_existed, iodata()}
+        | {duplicate_condition, iodata()}
+        | {max_limit_reached, non_neg_integer()}
         | {bad_type, any()}
         | iodata()}.
 create(Trace) ->
     maybe
         {ok, TraceRecord} ?= mk_trace_record(Trace),
-        true ?= mnesia:table_info(?TRACE, size) < max_traces(),
         ok ?= insert_new_trace(TraceRecord),
         {ok, format(TraceRecord)}
     else
         {error, Reason} ->
-            {error, Reason};
-        false ->
-            {error,
-                "The number of traces created has reached the maximum"
-                " please delete the useless ones first"}
+            {error, Reason}
     end.
-
--spec max_traces() -> non_neg_integer().
-max_traces() ->
-    emqx_config:get([trace, max_traces]).
 
 -spec delete(Name :: binary()) -> ok | {error, not_found}.
 delete(Name) ->
@@ -427,7 +419,7 @@ code_change(_, State, _Extra) ->
 
 insert_new_trace(Trace) ->
     case transaction(fun emqx_trace_dl:insert_new_trace/1, [Trace]) of
-        {ok, _} ->
+        ok ->
             %% We call this to ensure the trace is active when we return
             check();
         {error, _} = Error ->
--- a/apps/emqx/src/emqx_trace/emqx_trace.erl
+++ b/apps/emqx/src/emqx_trace/emqx_trace.erl
@@ -272,7 +272,7 @@ list_traces(Enable) ->
 create(Trace) ->
     maybe
         {ok, TraceRecord} ?= mk_trace_record(Trace),
-        true ?= mnesia:table_info(?TRACE, size) < ?MAX_SIZE,
+        true ?= mnesia:table_info(?TRACE, size) < max_traces(),
         ok ?= insert_new_trace(TraceRecord),
         {ok, format(TraceRecord)}
     else
@@ -283,6 +283,10 @@ create(Trace) ->
                 "The number of traces created has reached the maximum"
                 " please delete the useless ones first"}
     end.
+
+-spec max_traces() -> non_neg_integer().
+max_traces() ->
+    emqx_config:get([trace, max_traces]).
 
 -spec delete(Name :: binary()) -> ok | {error, not_found}.
 delete(Name) ->
--- a/apps/emqx/src/emqx_trace/emqx_trace.erl
+++ b/apps/emqx/src/emqx_trace/emqx_trace.erl
@@ -647,8 +647,12 @@ validate_time_range(_) ->
 
 validate_name(undefined) ->
     {error, "name required"};
-validate_name(Name) ->
-    case re:run(Name, ?NAME_RE) of
+validate_name(Name) when is_binary(Name) ->
+    case
+        Name =:= unicode:characters_to_binary(Name, utf8) andalso
+            re:run(Name, ?NAME_RE)
+    of
+        false -> {error, "Name is not a UTF-8 string"};
         nomatch -> {error, "Name should be " ?NAME_RE};
         _ -> ok
     end.
--- a/apps/emqx/src/emqx_trace/emqx_trace.erl
+++ b/apps/emqx/src/emqx_trace/emqx_trace.erl
@@ -756,3 +756,48 @@ maybe_migrate_trace(Fields) ->
             {atomic, ok} = mnesia:transform_table(?TRACE, TransFun, Fields, ?TRACE),
             ok
     end.
+
+%% Tests
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+
+-define(DEFAULT_MS, 60 * 15000).
+
+find_closest_time_empty_test() ->
+    ?assertEqual(?DEFAULT_MS, find_closest_time([], now_second())).
+
+find_closest_time_disabled_test() ->
+    Now = now_second(),
+    Traces = [
+        #emqx_trace{name = <<"disable">>, start_at = Now + 1, end_at = Now + 2, enable = false}
+    ],
+    ?assertEqual(?DEFAULT_MS, find_closest_time(Traces, Now)).
+
+find_closest_time_ends_soon_test() ->
+    Now = now_second(),
+    Traces = [
+        #emqx_trace{name = <<"running">>, start_at = Now, end_at = Now + 10, enable = true}
+    ],
+    ?assertEqual(10000, find_closest_time(Traces, Now)).
+
+find_closest_time_starts_soon_test() ->
+    Now = now_second(),
+    Traces = [
+        #emqx_trace{name = <<"waiting">>, start_at = Now + 2, end_at = Now + 10, enable = true}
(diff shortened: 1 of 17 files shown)


apps/emqx/src/emqx_trace/emqx_trace_freelist.erl
```diff
@@ -0,0 +1,96 @@
...
+first([]) ->
+    {error, full}.
+
+-spec to_list(t()) -> [integer()].
+to_list([A, B | Rest]) ->
+    lists:seq(A, B) ++ to_list(Rest);
+to_list([]) ->
+    [].
+
+-spec occupy(integer(), t()) -> t().
```
savonarola, line 47: Nit 
a bit confusing that occupying an absent value is not a error, meaning that it's the client of this module should ensure consistency 🤔

savonarola (approved):

savonarola (approved):

changes/ee/feat-15594.en.md
```diff
@@ -0,0 +1,3 @@
+Exposed maximum number of traces allowed to exist in the cluster simultaneously as a configuraion option `trace.max_traces`. This limit does not apply to node-local traces managed through `emqx ctl traces`.
```
thalesmg:
```suggestion
Exposed maximum number of traces allowed to exist in the cluster simultaneously as a configuration option `trace.max_traces`. This limit does not apply to node-local traces managed through `emqx ctl traces`.
```

apps/emqx/src/emqx_trace/emqx_trace_handler.erl
```diff
@@ -146,16 +117,30 @@ uninstall(HandlerId) ->
     [
         #{
             name => binary(),
-            type => topic | clientid | ip_address,
+            type => topic | clientid | ip_address | ruleid,
```
thalesmg, line 120: Nit suggestion: maybe we could make this a separate type, since it's referenced in at least 3 places here.

keynslug:
Sorry @savonarola for another review request, had to do a rebase and [mildly refactor few test suites](https://github.com/emqx/emqx/pull/15594/commits/6031f77a10483e2371947f0476665e15d5aa30e7).

savonarola (dismissed):

apps/emqx_management/src/emqx_mgmt_cli.erl
```diff
@@ -1183,7 +1186,8 @@ indent_print({Key, Val}) ->
     emqx_ctl:print("  ~-16s: ~w~n", [Key, Val]).
 
 name(Filter) ->
-    iolist_to_binary(["CLI-", Filter]).
+    Name = iolist_to_binary(["CLI-", Filter]),
+    Name = unicode:characters_to_binary(Name, utf8).
```
thalesmg, line 1190: Q: is this intended to be an assertion that both results yield in the same value?

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -242,34 +261,36 @@ log_filter([{Id, FilterFun, Filter, Name} | Rest], Log0) ->
...
-            {error,
-                "The number of traces created has reached the maximum"
-                " please delete the useless ones first"}
+    maybe
+        {ok, TraceRecord} ?= mk_trace_record(Trace),
+        ok ?= insert_new_trace(TraceRecord),
+        {ok, format(TraceRecord)}
+    else
+        {error, Reason} ->
+            {error, Reason}
```
thalesmg: Nit: this clause is not needed, since there's no transformation on the error value.

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -320,21 +341,28 @@ trace_file_detail(File) ->
...
+        enable => Enable,
+        name => Name,
+        type => Type,
+        filter => Filter,
+        start_at => StartAt,
+        end_at => EndAt,
+        payload_encode => PayloadEncode,
+        payload_limit => maps:get(payload_limit, Extra, ?MAX_PAYLOAD_FORMAT_SIZE),
+        formatter => maps:get(formatter, Extra, text)
+    }.
```
thalesmg, line 362: Q: should `namespace` be extracted from `Extra` to match the `trace()` type, if the value is available?

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -575,112 +602,80 @@ classify_by_time(
...
+            payload_encode = maps:get(payload_encode, Trace, text),
+            extra = maps:merge(
+                #{formatter => text},
+                maps:with([formatter, payload_limit, namespace], Trace)
+            )
+        }),
+        ok ?= validate_time_range(Record),
+        {ok, Record}
+    else
+        Error -> Error
```
thalesmg: Nit: this `else` clause is not required.

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -575,112 +602,80 @@ classify_by_time(
...
+validate_filter(clientid, _ClientId = <<_/binary>>) ->
+    ok;
+validate_filter(topic, Topic) ->
+    validate_topic(Topic);
+validate_filter(ip_address, Addr) ->
+    validate_ip_address(Addr);
+validate_filter(ruleid, _RuleId = <<_/binary>>) ->
+    ok;
+validate_filter(_, _) ->
+    {error, "type=[topic,clientid,ip_address] required"}.
```
thalesmg:
```suggestion
    {error, "type=[topic,clientid,ip_address,ruleid] required"}.
```

apps/emqx_management/src/emqx_mgmt_cli.erl
keynslug: Yes. Idea is to assert it's not `{error, ...}` early. But in fact it's not an expected error condition, hence emitting `badmatch` looks fine here.

apps/emqx/src/emqx_trace/emqx_trace.erl
keynslug: Hmm, after taking another look there's rather awkward inconsistency between logic and typespecs. For example, I can't seem to find where `{?global_ns | binary(), ruleid()}` filter variant is constructed and/or used.

On the other hand, there's `namespace` key-value pair but it's enforced only for `ruleid` traces. This probably means that namespaced users are able to trace any client or topic. Which can be considered a security concern, depending on how we think about user namespaces.

I do agree that it's a blemish on the module API, however the existing behavior is not affected AFAICT. How about we discuss the general problem in this thread but tackle it later in a followup PR?

apps/emqx/src/emqx_trace/emqx_trace_handler.erl
keynslug: Agreed. I'm thinking of refining the shape of the outer type further in a followup PR.

apps/emqx/src/emqx_trace/emqx_trace.erl
thalesmg: I also found the way the trace API is tied together to be a bit confusing.  :sweat: 

> I can't seem to find where {?global_ns | binary(), ruleid()} filter variant is constructed and/or used.

https://github.com/emqx/emqx/blob/0286ba2e334cb0b5a40d0478f328aa7e51478508/apps/emqx/src/emqx_trace/emqx_trace_handler.erl#L214

> Which can be considered a security concern, depending on how we think about user namespaces.

From what I understood from discussions with @zmstone (please correct me if I misunderstood), the purpose of namespaces is more to avoid accidental mutation.  Reading stuff from other namespaces (or the _global_ one) is fine, does not need to be strictly forbidden.

> I do agree that it's a blemish on the module API, however the existing behavior is not affected AFAICT. How about we discuss the general problem in this thread but tackle it later in a followup PR?

Sure, just thought it would be appropriate to add the new `namespace` root key here, while the change is red hot.

thalesmg (approved):

apps/emqx/src/emqx_trace/emqx_trace.erl
zmstone: Admin/API user/key namespace for GA is only to be used by trusted logins, meaning all users can trace all MQTT client namespaces.
Another reason is MQTT client namespace extraction is flexible, it does not always match admin/API user/key namespace.

Longer term goal for multi-tenancy is to replace the server-less edition, for this, we will need a global flag to enforce the use of TLS SNI for MQTT user namespace, so admin/API namespace and MQTT client namespace can be aligned. --- after this is in place, we can implement trace filter to drop traces emitted from other namespaces (other than the namespace of the dashboard/API user/key which created the trace).

keynslug: Understood, thanks.

> [emqx/apps/emqx/src/emqx_trace/emqx_trace_handler.erl](https://github.com/emqx/emqx/blob/0286ba2e334cb0b5a40d0478f328aa7e51478508/apps/emqx/src/emqx_trace/emqx_trace_handler.erl#L214)

Right. It's not what I was wondering about though, I was mostly confused that it's permitted [here](https://github.com/emqx/emqx/blob/d79d019d5935c2b956140bd330b94c39b0c6239a/apps/emqx/src/emqx_trace/emqx_trace_handler.erl#L37) and [here](https://github.com/emqx/emqx/blob/d79d019d5935c2b956140bd330b94c39b0c6239a/apps/emqx/src/emqx_trace/emqx_trace_handler.erl#L273) but never actually used.

What's more, if `filter` is `{global, <<"NS">>}` in a `tracer()` passed to `install/4` then the line you're linked to will produce `{global, {global, <<"NS">>}}` (if I'm following the code correctly).

Anyway, let's tackle these in a followup, going to be quick.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8485)

apps/emqx/src/emqx_trace/emqx_trace.erl
keynslug: Drafted a followup #15620.

PR #15609, opened 2025-07-29
feat(config backup): support exporting/importing namespaced config

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14383

## Summary

Continuation of the https://emqx.atlassian.net/browse/EMQX-14303 epic.

Here, we export namespaced configs as separate hocon files in the backup tarball, one directory for each namespace.  Filtering by root keys when exporting should keep working as it already does for the global namespace.

## PR Checklist

- [na] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files (part of a bigger epic, not officially released)


apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
```diff
@@ -181,6 +182,13 @@ do_check_rbac(#{?role := ?ROLE_SUPERUSER, ?namespace := Namespace}, _Req, ?TRACE
 ->
     %% Used by rule simulation API.
     true;
+do_check_rbac(
+    #{?role := ?ROLE_SUPERUSER, ?namespace := Namespace}, _Req, ?DATA_BACKUP_API(_, _)
+) when
+    is_binary(Namespace)
+->
+    %% Configuration backup export/import.
```
zmstone, line 190: we should only allow global admin for this API.
for namespace admin, we should add an API to export/import configs in their own ns.

thalesmg: :checkered_flag:

thalesmg:
> this may come in the future: support export/import config for a given ns by a global admin.

I think we could add this when needed by adding an extra body or query param argument with the namespace.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8496)

PR #15623, opened 2025-07-31
ci: use emqx/macos-erlang action in package-macos

id:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8491)

PR #15624, opened 2025-07-31
250731 sync r510 to r60

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8498)

PR #15629, opened 2025-08-01
feat(namespace): delete namespaced resources on namespace deletion

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14548

Release version: 6.0.0

## Summary

Once a managed namespace is deleted, we delete its resources:
- Connectors
- Sources
- Actions
- Rules
- Dashboard users (and their tokens)
- API Keys

## PR Checklist

- [na] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files (part of a larger epic)


zmstone (commented):
hocon schema modules must define a behavior callback `roots/0`.
can we add a `dependencies/0` callback which references to other schema modules ?

apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -30,6 +32,8 @@
 
 -export([backup_tables/0, validate_mnesia_backup/1]).
 
+-export([delete_all_keys_from_namespace/1]).
```
zmstone, line 35: nit:
```suggestion
-export([delete_all_keys_for_namespace/1]).
```

apps/emqx_mt/src/emqx_mt_state.erl
```diff
@@ -56,10 +58,12 @@
...
 ]).
 
-%% Internal exports for `emqx_mt_config`.
+%% Internal exports for `emqx_mt_config{,_janitor}`.
 -export([
     tables_to_backup/0,
-    ensure_ns_added/1
+    ensure_ns_added/1,
+    clear_tombstone/1,
+    is_tombstoned/1
```
zmstone, line 66: this API should be called before creating a namespace ?

apps/emqx_mt/src/emqx_mt_config_janitor.erl
```diff
@@ -0,0 +1,182 @@
...
+        {ok, _} ->
+            ok;
+        {error, Reason} ->
+            ?SLOG(error, #{
+                msg => "mt_failed_to_delete_namespaced_config",
+                namespace => Namespace,
+                root_key => RootKey,
+                reason => Reason
+            }),
+            ok
```
zmstone, line 176: failed to deleted should not result in tombstone clear?

thalesmg: I was wondering what should we do in this case?  :thinking:

apps/emqx_mt/src/emqx_mt_state.erl
thalesmg: it's checked directly in the transaction to create the namespace.

https://github.com/thalesmg/emqx/blob/16990652a1730d1ab56ef96e0891124cc2a83550/apps/emqx_mt/src/emqx_mt_state.erl#L545-L563

apps/emqx_management/src/emqx_mgmt_auth.erl
thalesmg: "for" here gives me the impression of "on behalf of".

IMO, "from" still sounds better than "for" for this use case.

maybe it could be "of namespace" :thinking:

thalesmg:
> hocon schema modules must define a behavior callback `roots/0`. can we add a `dependencies/0` callback which references to other schema modules ?

IIUC, even if we added such `-callback dependencies() -> [{root(), [module()]}].`, we'll still need to register them in the dep registry so we have the "total" order, right?  I'm trying to see which benefits moving this to hocon would give us.  :thinking:

zmstone:
> > hocon schema modules must define a behavior callback `roots/0`. can we add a `dependencies/0` callback which references to other schema modules ?
> 
> IIUC, even if we added such `-callback dependencies() -> [{root(), [module()]}].`, we'll still need to register them in the dep registry so we have the "total" order, right? I'm trying to see which benefits moving this to hocon would give us. 🤔

I think it's perhaps better to keep the schema information (dependency of configs) in the schema modules.
The code to build the dependency tree is slightly a separate matter, maybe should be a part of `hocon` as well, but `emqx_conf` works too.

apps/emqx_mt/src/emqx_mt_state.erl
zmstone: > this API should be called before creating a namespace ?

I mean, should not allow to add back a namespace which is being deleted.

thalesmg: > I mean, should not allow to add back a namespace which is being deleted.

yes, I got that.  the code I linked above handles that, right?

apps/emqx_mt/src/emqx_mt_config_janitor.erl
zmstone: but what should we do after we ignore it?

keeping the namespace (and tombstone) should at least allow a retry later.
if retry fails, some manual clean-up will be needed.

otherwise, if resources or db records orphanated, it would be hard to notice since the namespace would be gone?

thalesmg: that's what I was asking :sweat_smile:

thalesmg:
> approved. but would be nice to have a follow-up fix or tests to cover partial delete failure.

fair, I'll do it in a follow up

apps/emqx_mt/src/emqx_mt_config_janitor.erl
thalesmg: Huh.  Detecting update failures might be trickier than expected.... :melting_face: 

https://github.com/emqx/emqx/blob/b757cbfa1c3bfb7cacfec3d86c0c31774bc240c8/apps/emqx/src/emqx_config_handler.erl#L357-L360

As long as the config handler throws errors, the config won't be saved, at least.  :sweat_smile:

thalesmg:
> fair, I'll do it in a follow up

:checkered_flag: 

https://github.com/emqx/emqx/pull/15651

apps/emqx_mt/src/emqx_mt_config_janitor.erl
thalesmg: Addressed in https://github.com/emqx/emqx/pull/15651

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8990)

PR #15630, opened 2025-08-01
ci: fix app vsn check script when application is deleted

thalesmg:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8508)

PR #15640, opened 2025-08-05
refactor: delete 'bridges' schema from schema JSON for dashboard

zmstone:
To facilitate dashboard UI components, the backend exposes below API specs in JSON format

- hotconf
- bridges
- actions
- connectors

'bridges' is deprecated since EMQX 5.3.
This commit deletes the schema dump of 'bridges'

The deprecated bridge (v1) API is not deleted in this commit though.

### DIFF
--- a/apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
+++ b/apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
@@ -1106,9 +1106,6 @@ api_spec_schemas(Root) ->
             Schemas
     end.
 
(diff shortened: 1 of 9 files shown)


keynslug:
I guess this also need to be removed.
https://github.com/emqx/emqx/blob/e288c337c00c6288a5d391493573fc0b6001aa01/scripts/test/emqx-smoke-test.sh#L93

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8617)

PR #15645, opened 2025-08-05
ci(eunit): workaround `eunit`'s duplication of modules

thalesmg:
`eunit` is buggy when enumerating modules: the docs says that `{application, _}` uses the application's `.app` file to discover modules.  However, if we use just that, `eunit` sees both `mod` and `mod_test` in `.app`, and then decides to also add `mod_test` _again_, duplicating the tests being run.  So, as a workaround to this bug, we need to explicitly enumerate all modules.

Also, just enumerating all modules yield the same behavior.  We need to check if both `mod` and `mod_tests` exist, and remove the latter, since `eunit` will **always** add `mod_tests` given `mod`.

### DIFF
--- a/apps/emqx_mix_utils/lib/mix/tasks/emqx.eunit.ex
+++ b/apps/emqx_mix_utils/lib/mix/tasks/emqx.eunit.ex
@@ -142,8 +142,23 @@ defmodule Mix.Tasks.Emqx.Eunit do
   end
 
   defp discover_tests(%{cases: [], modules: []} = _opts) do
+    ## `eunit` is problematic here: the docs says that `{application, _}` uses the
+    ## application's `.app` file to discover modules.  However, if we use just that,
+    ## `eunit` sees both `mod` and `mod_test` in `.app`, and then decides to also add
+    ## `mod_test` _again_, duplicating the tests being run.  So, as a workaround to this
+    ## bug, we need to explicitly enumerate all modules.
+    ##
+    ## Also, just enumerating all modules yield the same behavior.  We need to check if
+    ## both `mod` and `mod_tests` exist, and remove the latter, since `eunit` will
+    ## **always** add `mod_tests` given `mod`.
     Mix.Dep.Umbrella.cached()
-    |> Enum.map(&{:application, &1.app})
+    |> Enum.map(fn dep ->
+      modules =
+        dep.opts[:app_properties][:modules]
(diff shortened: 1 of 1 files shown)


apps/emqx_mix_utils/lib/mix/tasks/emqx.eunit.ex
```diff
@@ -152,4 +167,30 @@ defmodule Mix.Tasks.Emqx.Eunit do
...
+
+    Enum.reduce(test_modules, normal_modules, fn test_mod, acc ->
+      base_mod =
+        test_mod
+        |> to_string()
+        |> String.replace(re, "")
+        |> String.to_atom()
+
+      if base_mod in acc do
+        # don't add `mod_test`, otherwise `eunit` will duplicate it
```
zmstone:
```suggestion
        # don't add `mod_tests`, otherwise `eunit` will duplicate it
```

apps/emqx_mix_utils/lib/mix/tasks/emqx.eunit.ex
```diff
@@ -142,8 +142,23 @@ defmodule Mix.Tasks.Emqx.Eunit do
   end
 
   defp discover_tests(%{cases: [], modules: []} = _opts) do
+    ## `eunit` is buggy here: the docs says that `{application, _}` uses the application's
```
zmstone: It's perhaps not a bug for eunit.
In rebar3 convention, test modules reside in `test` dir, but not `ebin`, and test modules are not added to the `modules` list in `.app`

thalesmg: reworded.

zmstone: still claimed it being a 'bug' in the following lines and the PR description.

thalesmg: IMO, it still is a bug... the docs say it uses the module list provided explicitly or from `.app`, though that's not entirely true, since it unconditionally adds `mod_tests` (if it exists) for each `mod`, be it in the module list or not.  Therefore, it **does not** entirely respect the module/application list given as argument.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8548)

PR #15649, opened 2025-08-05
feat: add `alloydb`, `cockroachdb` and `redshift` postgres flavors

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14543

These 3 Connectors & Actions are simple shims over the existing Postgres Connector.

Release version: 6.0.0

## PR Checklist


rel/i18n/emqx_bridge_cockroachdb_connector_schema.hocon
```diff
@@ -0,0 +1,7 @@
+emqx_bridge_cockroachdb_connector_schema {
+  config_connector {
+    label: """CockroachDB Connector Configuration"""
+    desc:
+    """Configuration for a connector to CockroachDB."""
```
ieQu1, line 5: Perhaps elaborate the documentation, e.g. provide a link to the official site.

thalesmg: Yes, that sort of detail will be added to https://github.com/emqx/emqx-docs/ as usual.

savonarola (approved):

savonarola:
Nit: having a separate app for a "bridge alias" seems to be an overkill 😅

thalesmg:
> Nit: having a separate app for a "bridge alias" seems to be an overkill 😅

Fair.  Currently I don't think we have a more lightweight way to have the same impl module serve different brandings.  We also do this with Matrix and Timescale.  Maybe we could look into supporting that.  :thinking:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8560)

changes/ee/feat-15649.en.md
```diff
@@ -0,0 +1 @@
+Added new Connectors and Actions that appends data to AWS AlloyDB, CockroachDB and AWS Redshift.
```
lina50, line 1: Is it Google AlloyDB instead of AWS AlloyDB?

thalesmg:
> Nit: having a separate app for a "bridge alias" seems to be an overkill 😅

Though, a counterpoint is that, if any small differences are needed to be made between those shims, we'll need a whole app anyway (as they'll probably appear).  :sweat_smile: 

And, sure enough, soon they appeared: 
- https://github.com/emqx/emqx/pull/15675
- https://github.com/emqx/emqx/pull/15676

PR #15650, opened 2025-08-05
feat(trace): implement trace log rotation

keynslug:
Fixes [EMQX-14537](https://emqx.atlassian.net/browse/EMQX-14537).

Release version: 6.0.0

## Summary

TBA

## PR Checklist

- [ ] Schema changes are intentionally breaking
    Trace Stream API is expanded in `position` query parameter, formally making it backward incompatible. However, code gracefully handles non-zero integer values.


apps/emqx/src/emqx_trace/emqx_trace_dl.erl
```diff
@@ -113,11 +114,20 @@ delete(Name) ->
...
+%% Deprecated since 5.10.0
 -spec get_trace_filename(Name :: binary()) -> {ok, string()}.
 get_trace_filename(Name) ->
     case mnesia:read(?TRACE, Name, read) of
         [] -> mnesia:abort(not_found);
-        [#?TRACE{start_at = Start}] -> {ok, emqx_trace:filename(Name, Start)}
+        [#?TRACE{start_at = Start}] -> {ok, emqx_trace:log_filename(Name, Start)}
+    end.
+
+%% Introduced in 6.0.0
```
thalesmg: Micro nit:

```suggestion
-doc #{since => <<"6.0.0">>}.
```

apps/emqx/src/emqx_trace/emqx_trace_dl.erl
```diff
@@ -113,11 +114,20 @@ delete(Name) ->
     end.
 
 %% Introduced in 5.0
+%% Deprecated since 5.10.0
```
thalesmg: Micro nit:

```suggestion
-doc #{deprecated => <<"5.10.0">>}.
```

apps/emqx/src/emqx_trace/emqx_trace_handler.erl
```diff
@@ -150,9 +174,219 @@ filesync(HandlerId) ->
...
+            Error
+    end.
+
+%%
+
+-opaque log_fragment() :: #{
+    %% Log fragment index:
+    i := non_neg_integer(),
+    %% Inode number from the filesystem:
+    node := integer(),
```
thalesmg: Nit (since `node` is so commonly used for erlang nodes):

```suggestion
    %% Inode number from the filesystem:
    inode := integer(),
```

apps/emqx_management/src/proto/emqx_mgmt_trace_proto_v3.erl
```diff
@@ -0,0 +1,60 @@
...
+    node(),
+    emqx_trace:name(),
+    start | {cont, emqx_trace:cursor()},
+    pos_integer() | undefined
+) ->
+    {ok, binary(), {cont | eof | retry, emqx_trace:cursor()}}
+    | {error, _}
+    | {badrpc, _}.
+stream_trace_log(Node, Name, Position, Limit) ->
+    rpc:call(Node, emqx_mgmt_api_trace, read_trace_file, [Name, Position, Limit], ?RPC_TIMEOUT).
```
thalesmg, line 60: Q: is there a particular reason to use `rpc:*` instead of `erpc:*`?

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -261,22 +304,136 @@ clear() ->
...
+    {error, Reason};
+stream_log_cursor(_Name, _Cursor, _Limit) ->
+    {error, bad_cursor}.
+
+log_filepath(Name, Start) ->
+    filename:join(trace_dir(), log_filename(Name, Start)).
+
+log_filename(Name, Start) ->
+    [Time, _] = string:split(calendar:system_time_to_rfc3339(Start), "T", leading),
+    lists:flatten(["trace_", binary_to_list(Name), "_", Time, ".log"]).
```
thalesmg, line 436: Q: is there any constraint on the name?  If it can be any string, maybe we should call `emqx_utils:safe_filename` on the result.

thalesmg (approved):

apps/emqx_management/src/proto/emqx_mgmt_trace_proto_v3.erl
keynslug: No. But `rpc` nowadays is basically `erpc` with trivial error mapping, so it's not that much of a benefit as it was before.

apps/emqx/src/emqx_trace/emqx_trace.erl
keynslug: Yes, and it's actually both [here](https://github.com/emqx/emqx/blob/5d731e884a28374eb335ceb8e535e8484e918fb3/apps/emqx/src/emqx_trace/emqx_trace.erl#L763-L764) and in the API handler for some weird reason.

apps/emqx/src/emqx_trace/emqx_trace_handler.erl
keynslug: Good point.

rel/i18n/emqx_mgmt_api_trace.hocon
```diff
@@ -110,10 +110,18 @@ max_response_bytes.desc:
...
 """Maximum response bytes"""
 
-current_trace_offset.desc:
-"""Offset from the current trace position."""
-current_trace_offset.label:
-"""Offset from the current trace position."""
+current_trace_cursor.desc:
+"""Cursor for the current trace position."""
+current_trace_cursor.label:
+"""Cursor for the current trace position."""
```
zmstone:
```suggestion
"""Trace Cursor""
```

rel/i18n/emqx_mgmt_api_trace.hocon
```diff
@@ -110,10 +110,18 @@ max_response_bytes.desc:
...
+
+current_trace_stream_hint.desc:
+"""Hint on the state of the current log stream.
+* `eof`: Stream has reached the end of the log.
+* `retry`: Stream is temporarily unavailable, will recover quickly."""
+
+current_trace_stream_hint.label:
+"""Current trace stream hint."""
```
zmstone:
```suggestion
"""Current Trace Stream Hint"""
```

apps/emqx_management/src/emqx_mgmt_api_trace.erl
```diff
@@ -374,14 +374,26 @@ fields(position) ->
...
+                    description => ?DESC(current_trace_cursor),
                     in => query,
                     required => false,
                     default => 0
                 }
             )}
+    ];
+fields(stream_hint) ->
+    [
+        {hint,
```
zmstone, line 388: Q: how should dashboard make use of this hint?

apps/emqx_management/src/emqx_mgmt_api_trace.erl
```diff
@@ -374,14 +374,26 @@ fields(position) ->
...
+            hoconsc:mk(
+                atom(),
```
zmstone: enum ?

apps/emqx_management/src/emqx_mgmt_api_trace.erl
keynslug: My rough idea was that the dashboard decides to wait a bit before requesting the next portion depending on the presence and value of this field. The primary goal here is not to enter tight loop if it's `retry`, which means there's likely an ongoing rotation.

And perhaps to show a visual cue "there is more" while it's not `eof`, or maybe enter "tail -f" mode if there was an `eof`. This is just a pure guess though.

zmstone: should perhaps notify frontend team to make such changes.

keynslug: Sure, noted. Just to clarify however, looking at how the dashboard uses this API, things are expected to work (more or less) as before without the dashboard knowing about meaning of this field.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9040)

PR #15664, opened 2025-08-08
test(crl): attempt to fix flaky test

thalesmg:
https://github.com/emqx/emqx/actions/runs/16833710032/job/47688370724?pr=15663#step:5:261

### DIFF
--- a/apps/emqx/test/emqx_crl_cache_SUITE.erl
+++ b/apps/emqx/test/emqx_crl_cache_SUITE.erl
@@ -488,6 +488,18 @@ ensure_ssl_manager_alive() ->
 now_ms() ->
     erlang:system_time(millisecond).
 
+%% Since some events in this suite might come from different processes, relying on time
+%% ordering might be flaky.
+sort_trace(Trace) ->
+    lists:sort(
+        fun(#{?snk_kind := K1} = E1, #{?snk_kind := K2} = E2) ->
+            S1 = maps:get(?snk_span, E1, undefined),
+            S2 = maps:get(?snk_span, E2, undefined),
+            {K1, S1} =< {K2, S2}
+        end,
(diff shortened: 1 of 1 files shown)


apps/emqx/test/emqx_crl_cache_SUITE.erl
```diff
@@ -488,6 +488,18 @@ ensure_ssl_manager_alive() ->
...
+%% ordering might be flaky.
+sort_trace(Trace) ->
+    lists:sort(
+        fun(#{?snk_kind := K1} = E1, #{?snk_kind := K2} = E2) ->
+            S1 = maps:get(?snk_span, E1, undefined),
+            S2 = maps:get(?snk_span, E2, undefined),
+            {K1, S1} =< {K2, S2}
+        end,
+        Trace
+    ).
```
keynslug, line 501: I guess `emqx_ds_test_helpers:group_maps_by(?snk_kind, Trace)` might help here, especially because `lists:sort/2` is not a stable sort I believe.

Also `=<` is unusual for a comparison function, I wonder why it's not `<`.

ieQu1: =< _is_ the usual ordering function https://www.erlang.org/doc/apps/stdlib/lists#ordering_function

ieQu1 (approved):

apps/emqx/test/emqx_crl_cache_SUITE.erl
keynslug: Oh right, thanks for the update. And turns out the sort is actually stable.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8614)

PR #15668, opened 2025-08-10
refactor(sessds): Split out persistent maps to a untility module

ieQu1:
Fixes <issue-or-jira-number>

Release version: 6.0.0

## Summary

Generalize persistent map (create a "DS ORM").

## PR Checklist


apps/emqx/test/emqx_persistent_session_ds_state_fuzzer.erl
```diff
@@ -356,14 +351,6 @@ sample(Size) ->
 precondition(_, _) ->
     true.
 
-postcondition(S, {call, ?MODULE, iterate_sessions, [_]}, Result) ->
```
ieQu1, line 359: I removed this check for simplicity: listing the sessions is a dirty operation, so it tends to create unnecessary problems for the test. Iteration is currently done using a common DS API that has its own tests, so there's no need to test it here.

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state_v2.erl
```diff
@@ -433,15 +433,7 @@ open_tx(ClientId, Verify) ->
                 ?ranks => emqx_ds_pmap:tx_restore(?MODULE, ?top_ranks, ClientId),
                 ?awaiting_rel => emqx_ds_pmap:tx_restore(?MODULE, ?top_awaiting_rel, ClientId)
             },
-            case Verify of
```
ieQu1, line 442: This was added before `optimistic_tx` module learned to track reads. Now it's redundand.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9000)

PR #15693, opened 2025-08-13
fix(postgres): use allocation apis to avoid leaking resource pool

thalesmg:
Possibly fixes https://emqx.atlassian.net/browse/EMQX-14600

## Summary

I could not reproduce the problem locally.  I suspect it might have been some kind of race while starting the pool, and the `close_connections` functions might have crashed or was not called.

## PR Checklist


qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8638)

PR #15712, opened 2025-08-15
fix: upgrade to Erlang 27.3.4.2-2 for unzip badarg issue

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8789)

PR #15713, opened 2025-08-15
refactor: do not catch exception during config init_load

zmstone:
previsouly the assumption about logger being not ready during initial config load is not entirely true.
the logger config is maybe not in sync with other nodes in the cluster but should be ready to at least log app start errors because the default log config is a part of the kernel app.

Release: 5.8.8


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8699)

PR #15714, opened 2025-08-15
test(oracle): refactor suites to standardize drop bridges v1 apis

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14606

**N.B.**: this <ins>**does not introduce new test cases**</ins>, only refactors existing suites to adhere to a more standardized way to test bridges that use the test library and uses HTTP APIs as much as possible, and at the same time stop using bridges V1 APIs so we may drop it.

Release version: 6.0.0

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
+++ b/apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
@@ -742,6 +742,9 @@ get_action_api(Config) ->
     ct:pal("get action (http) result:\n  ~p", [Res]),
     Res.
 
+get_action_api2(Config) ->
+    simplify_result(get_action_api(Config)).
+
 get_action_metrics_api(Config) ->
     ActionName = ?config(action_name, Config),
     ActionType = ?config(action_type, Config),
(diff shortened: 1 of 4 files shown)


apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
```diff
@@ -742,6 +742,9 @@ get_action_api(Config) ->
     ct:pal("get action (http) result:\n  ~p", [Res]),
     Res.
 
+get_action_api2(Config) ->
```
keynslug, line 745: Nit. In my opinion, it's preferable to just let the user code call `Mod:simplify_result(Mod:get_action_api(Config))` or have a local facade function, instead of introducing "library" functions with such obscure names.

thalesmg: I was doing that in other suites, but decided to add this upgraded function, and add a local facade that drops the `2.  The name is indeed obscure, sadly the other name was already taken and would require touching lots of suites simultaneously to fix...

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9015)

PR #15721, opened 2025-08-18
fix(mqtt_bridge): update emqtt

savonarola:
Follow-up to #15603

(emqtt was bumped but not to the latest)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8657)

PR #15723, opened 2025-08-18
ci: rename runner macos-14-arm64 to macos-14

zmstone:
macos-14-arm64 is self-hosted runner. now we can use github hosted runner for macos-14.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8700)

PR #15732, opened 2025-08-18
test(http): refactor suites to standardize and drop bridges v1 apis

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14582

Release version: 6.0.0

## Summary

**N.B.**: this <ins>**does not introduce new test cases**</ins>, only refactors existing suites to adhere to a more standardized way to test bridges that use the test library and uses HTTP APIs as much as possible, and at the same time stop using bridges V1 APIs so we may drop it.

## PR Checklist


thalesmg:
CI failure unrelated to the changes herein.

apps/emqx_connector/test/emqx_connector_api_SUITE.erl
```diff
@@ -434,9 +434,12 @@ resume_connector_resource(ConnectorResID, Config) ->
 http_connector_config() ->
     http_connector_config(_Overrides = #{}).
 http_connector_config(Overrides) ->
-    Params0 = emqx_bridge_http_v2_SUITE:make_connector_config([{http_server, #{port => 18083}}]),
-    Params = emqx_utils_maps:deep_merge(Params0, Overrides),
-    emqx_bridge_v2_testlib:parse_and_check_connector(<<"http">>, <<"x">>, Params).
+    emqx_bridge_http_SUITE:connector_config(
```
keynslug: Should be moved to a CT helper / testlib? Those inter-suite dependencies are somewhat headache-inducing. 🫠

apps/emqx_rule_engine/test/emqx_rule_engine_api_rule_apply_SUITE.erl
```diff
@@ -63,8 +63,8 @@ end_per_suite(Config) ->
     ok = emqx_cth_suite:stop(Apps),
     ok.
 
-init_per_testcase(_Case, Config) ->
-    emqx_bridge_http_test_lib:init_http_success_server(Config).
+init_per_testcase(TestCase, TCConfig) ->
+    emqx_bridge_http_SUITE:init_per_testcase(TestCase, TCConfig).
```
keynslug: Same, it could be pretty painful to change something in `emqx_bridge_http_SUITE` test suite and then investigate failures here. In this case I think it would be more preferable even just to copy-paste the relevant code, to be honest. 🫣

apps/emqx_rule_engine/test/emqx_rule_engine_api_rule_apply_SUITE.erl
```diff
@@ -63,8 +63,8 @@ end_per_suite(Config) ->
...
 
 end_per_testcase(_TestCase, _Config) ->
     ok = emqx_bridge_http_connector_test_server:stop(),
```
keynslug, line 97: Q: Curious why `init_per_testcase/2` changed by this one did not?

apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
```diff
@@ -832,6 +832,9 @@ probe_bridge_api(Kind, BridgeType, BridgeName, BridgeConfig) ->
     ct:pal("bridge probe (~s, http) result:\n  ~p", [Kind, Res]),
     Res.
 
+probe_bridge_api2(TCConfig, Overrides) ->
```
keynslug: Again, perhaps give a more apparent (even if wordy) name? `probe_bridge_api_simplify`?

apps/emqx_connector/test/emqx_connector_api_SUITE.erl
thalesmg: I think it's fair for a bridge's suite to be responsible for providing facilities to create connector configurations without having to go through creating a testlib that is not so widely used.  Specially if they are standardized (as I'm trying to do).
IMO, for such use cases, it's not that headache-inducing.  :thinking:

apps/emqx_rule_engine/test/emqx_rule_engine_api_rule_apply_SUITE.erl
thalesmg: Because the desired cleanup is already here.

apps/emqx_rule_engine/test/emqx_rule_engine_api_rule_apply_SUITE.erl
thalesmg: :checkered_flag:

apps/emqx_connector/test/emqx_connector_api_SUITE.erl
keynslug: My point is that the usual expectation (for me at least) is that a test suite is mostly self-contained, which means its local functions (`connector_config` in this case) is safe to touch without affecting any other test suite. In this case it's not true, and it's not obvious: after all we export all test suite functions because it's easy not because we think that all of them are part of the module contract. If it's shared code, why not put it in a testlib? It might eventually become more widely used. If it's not worth it in your opinion, then once again, having a copy of this function locally sounds better to me.

thalesmg: For this pure function that simply constructs a valid HTTP connector config, it'll be quite fine to change it (to add a new field, say) and let other suites pick that change up.  :thinking:

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9018)

PR #15729, opened 2025-08-18
sync `release-510` to `release-60`

thalesmg:

PR #15733, opened 2025-08-18
fix(conf app): return exit code 1 when config fails to load

thalesmg:
Only when syncing a recent change to `release-60` it was detected that the exit code when config fails to load is 0.

See:
- https://github.com/emqx/emqx/pull/15729
- https://github.com/emqx/emqx/actions/runs/17050064723/job/48336139780?pr=15729#step:7:11

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_app.erl
+++ b/apps/emqx_conf/src/emqx_conf_app.erl
@@ -28,12 +28,25 @@
 
 start(_StartType, _StartArgs) ->
     ok = mria:wait_for_tables(emqx_cluster_rpc:create_tables()),
-    ok = init_conf(),
+    try
+        ok = init_conf()
+    catch
+        C:E:St ->
+            ?SLOG(error, #{
+                msg => "failed_to_load_config", error => C, reason => E, stacktrace => St
+            }),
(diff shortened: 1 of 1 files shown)


apps/emqx_conf/src/emqx_conf_app.erl
```diff
@@ -28,13 +28,25 @@
 
 start(_StartType, _StartArgs) ->
     ok = mria:wait_for_tables(emqx_cluster_rpc:create_tables()),
-    ok = init_conf(),
+    try
+        ok = init_conf()
+    catch
+        C:E:St ->
+            %% logger is not quite ready.
+            io:format(standard_error, "Failed to load config~n~p~n~p~n~p~n", [C, E, St]),
```
zmstone:
```suggestion
            ?SLOG(#{msg => "failed_to_load_config", error => C, reason => E, stacktrace => St}),
```

thalesmg: :checkered_flag:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8702)

PR #15755, opened 2025-08-20
build(mix): add deps helper

thalesmg:
To prevent unexpected issues with changing order of application compilation, and to also introduce a helper to refer to common dependencies.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9027)

PR #15757, opened 2025-08-20
ci: do not update public broker if not latest

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-8801)

PR #15780, opened 2025-08-27
test: use per test case client ID

zmstone:
The test become flaky after the client ID registration throttling.


ieQu1 (approved):

zmstone:
cluster-link failed test is disabled in merge base.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9091)

PR #15782, opened 2025-08-27
move enterprise releases back to emqx.git

id:


.github/workflows/build_packages_cron.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
         profile:
           - ['emqx', 'master']
           - ['emqx', 'release-58']
+          - ['emqx-enterprise', 'release-58']
```
thalesmg, line 26: Q: does this change take effect if not present on `master`?

thalesmg (approved):

zmstone (approved):

.github/workflows/build_packages_cron.yaml
id: Right, it does not.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9092)

PR #15784, opened 2025-08-27
250827 sync `release-510` to `release-60`

zmstone:
Fixes <issue-or-jira-number>

Release version:

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -58,7 +58,8 @@
 %% Test/debug interface
 -export([
     all_channels/0,
-    all_client_ids/0
+    all_client_ids/0,
+    do_unregister_channel/1
 ]).
 
 %% Client management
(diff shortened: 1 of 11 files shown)


apps/emqx/src/emqx_cm.erl
```diff
@@ -278,14 +279,48 @@ set_chan_stats(ClientId, ChanPid, Stats) when ?IS_CLIENTID(ClientId) ->
...
+open_session(CleanStart, ClientInfo = #{clientid := ClientId}, ConnInfo, MaybeWillMsg) ->
+    Pids = emqx_cm_registry:lookup_all_channels(ClientId),
+    {LocalAlive, LocalDown, Remote} = lists:foldl(
+        fun(Pid, {LA, LD, R}) ->
+            IsLocal = (node(Pid) =:= node()),
+            case {IsLocal, IsLocal andalso is_process_alive(Pid)} of
+                {true, true} ->
+                    {LA + 1, LD, R};
+                {true, false} ->
+                    {LA, LD + 1, R};
```
thalesmg, line 291: IIUC, if the pid is not local, we increment the `LocalDown` count here?  Shouldn't it be `Remote`?

thalesmg (approved):

id (dismissed):

apps/emqx/src/emqx_cm.erl
zmstone: if pid is not local, then `IsLocal` is `false` hence it will be `{false, false}`

id (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9044)

PR #15772, opened 2025-08-25
fix: crash congestion when username is unicode char

zhongwencool:
Fixes <issue-or-jira-number>

```
2025-08-19T09:55:39.227861+00:00 [warning] clientid: f25899f5-1166-444b-bd46-5fdad28daffd, msg: unclean_terminate, peername: 192.168.2.3:37220, username: 矿卡#11, context: badarg, stacktrace: [{erlang,list_to_binary,[[99,111,110,110,95,99,111,110,103,101,115,116,105,111,110,47,“f25899f5-1166-444b-bd46-5fdad28daffd”,47,“矿卡#11”]],[{error_info,#{module => erl_erts_errors}}]},{emqx_congestion,do_cancel_alarm_congestion,4,[{file,“emqx_congestion.erl”},{line,106}]},{lists,foreach_1,2,[{file,“lists.erl”},{line,2310}]},{emqx_connection,terminate,2,[{file,“emqx_connection.erl”},{line,621}]},{proc_lib,wake_up,3,[{file,“proc_lib.erl”},{line,340}]}], exception: error
[...]

PR #15785, opened 2025-08-27
fix: crash congestion when username is unicode char

zmstone:
port #15772 to release-58

Release: v/e5.8.8


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9093)

PR #15703, opened 2025-08-14
feat(namespace): add api to bulk-transform namespaced configurations

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14463

Release version: 6.0.0

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files

PR #15789, opened 2025-08-28
Implement Message Queues

savonarola:
Fixes [EMQX-14402)](https://emqx.atlassian.net/browse/EMQX-14402)

Release version: 6.0.0

## Summary

Implement Message queues.

See [README.md](https://github.com/savonarola/emqx/blob/250811-message-queues-ds-subs-api/apps/emqx_mq/README.md) for high-level design.

Also see [slides](https://github.com/savonarola/emqx-slides/blob/message-queue/main.pdf) from the demo for the detailed data flow explanation.

## PR Checklist


rel/i18n/emqx_mq_api.hocon
```diff
@@ -0,0 +1,25 @@
+emqx_mq_api {
+    message_queues_list {
+        desc: """~
+            List all message queues
```
ieQu1:
```suggestion
            List all message queues.
```

rel/i18n/emqx_mq_api.hocon
```diff
@@ -0,0 +1,25 @@
...
+            The message queues are sorted in lexicographical order,
+            so this API can be used for prefix based fuzzy search.~"""
```
ieQu1:
```suggestion
            so this API can be used for prefix-based fuzzy search.~"""
```

apps/emqx_mq/src/emqx_mq_utils.erl
```diff
@@ -0,0 +1,22 @@
...
+
+-export([
+    merge_maps/1
+]).
+
+%%--------------------------------------------------------------------
+%% API
+%%--------------------------------------------------------------------
+
+merge_maps(Maps) ->
```
ieQu1: Please add a typespec. Also move to `emqx_utils_maps` ?

apps/emqx_mq/src/emqx_mq_gc/emqx_mq_gc_worker.erl
```diff
@@ -0,0 +1,161 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+-module(emqx_mq_gc_worker).
+
+-moduledoc "The module is responsible for garbage collection of Message "
+"Queue data.".
```
ieQu1: Nit:
```suggestion
-moduledoc """"
The module is responsible for garbage collection of Message
Queue data.
""".
```

apps/emqx_mq/src/emqx_mq_gc/emqx_mq_gc_worker.erl
```diff
@@ -0,0 +1,161 @@
...
+        type => worker,
+        modules => [?MODULE]
+    }.
+
+%%--------------------------------------------------------------------
+%% gen_server callbacks
+%%--------------------------------------------------------------------
+
+init([]) ->
+    erlang:process_flag(trap_exit, true),
```
ieQu1, line 62: If you `trap_exit`, then you must handle `{'EXIT', SupPid, shutdown}` message accordingly: https://erlang.org/documentation/doc-4.9.1/doc/design_principles/sup_princ.html#shutdown

apps/emqx_mq/src/emqx_mq_gc/emqx_mq_gc_worker.erl
ieQu1: Also please describe how the GC logic works, it's not super obvious from the code.

apps/emqx_mq/src/emqx_mq_gc/emqx_mq_gc.erl
```diff
@@ -0,0 +1,93 @@
...
+handle_call(_Request, _From, State) ->
+    {reply, {error, unknown_request}, State}.
+
+handle_cast(_Request, State) ->
+    {noreply, State}.
+
+handle_info(#gc{}, State) ->
+    ok = start_gc(),
+    erlang:send_after(gc_interval(), self(), #gc{}),
+    {noreply, State}.
```
ieQu1, line 69: I think we should invest in a common "crontab" server, otherwise code like this will keep reproducing...

In the simplest case it could look like as a module implementing `gen_server` behavior, stored in `emqx_util`, that can be parameterized with task definitions. This way it can be plugged into a any supervisor, thus making lifetime of the cronjobs trivially correct, but this is open for discussion.

apps/emqx_mq/src/emqx_mq.app.src
```diff
@@ -0,0 +1,10 @@
+{application, emqx_mq, [
+    {description, "Message Queue"},
+    {vsn, "0.1.0"},
+    {registered, []},
+    {mod, {emqx_mq_app, []}},
+    {applications, [kernel, stdlib, gproc, emqx, emqx_durable_storage]},
```
ieQu1: Business applications should have `emqx_ds_backends` as a dependency rather then `durable_storage`.

apps/emqx_mq/asn.1/ConsumerState.asn
```diff
@@ -0,0 +1,37 @@
...
+    bufferProgress StreamBufferProgress
+  }
+
+  ConsumerStateV1 ::= SEQUENCE {
+    progress SEQUENCE OF ShardProgress
+  }
+
+  ConsumerState ::= CHOICE {
+    v1 ConsumerStateV1,
+    ...
```
ieQu1: If you intend to make your schema extendable then it's almost always necessary to add tags manually.

apps/emqx_mq/asn.1/ConsumerClaim.asn
```diff
@@ -0,0 +1,24 @@
...
+    lastSeenTimestamp INTEGER
+  }
+
+  ConsumerClaimV1 ::= CHOICE {
+    live ConsumerClaimLive,
+    tombstone ConsumerClaimTombstone
+  }
+
+  ConsumerClaim ::= CHOICE {
+    v1 ConsumerClaimV1,
```
ieQu1: Same

apps/emqx_mq/README.md
```diff
@@ -0,0 +1 @@
+# MQTT Message Queue
```
ieQu1, line 1: I don't think this follows our common application README template

apps/emqx_durable_storage/src/emqx_ds_buffer.erl
```diff
@@ -302,6 +302,7 @@ do_flush(
             S#s{
                 n = 0,
                 n_bytes = 0,
+
```
ieQu1: Please revert

apps/emqx/test/emqx_channel_hookpoints_SUITE.erl
```diff
@@ -1,8 +1,8 @@
 %%--------------------------------------------------------------------
-%% Copyright (c) 2018-2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
```
ieQu1: Why change copyright years?

apps/emqx/test/emqx_channel_hookpoints_SUITE.erl
```diff
@@ -48,10 +48,44 @@ t_delayed_puback(_Config) ->
     ),
     emqtt:disconnect(ConnPid).
 
+t_deliver_via_handle_info({init, Config}) ->
```
ieQu1, line 53: Please add a comment describing what property is being verified, and how; what is given and what is expected.

apps/emqx_mq/asn.1/ConsumerClaim.asn
savonarola: Consumer persistence is mostly outdated, this branch does not include moving to pmaps yet

apps/emqx_mq/mix.exs
```diff
@@ -0,0 +1,44 @@
...
+      start_permanent: Mix.env() == :prod,
+      deps: deps()
+    ]
+  end
+
+  def asn1_srcs() do
+    "./asn.1/*.asn"
+    |> Path.wildcard()
+    |> Enum.map(fn src ->
+      %{src: src, compile_opts: [:per, :noobj, outdir: ~c"gen_src"]}
```
ieQu1: 1. I don't think per is a good option. It's not extendable. It's for cases where format of the data is fixed and size of encoded binaries is of utmost importance.
2. I prefer when each ASN.1 schema file is listed in the `asn1_srcs` individually, rather than using wildcards and `Enum.map`. This way it's harder to accidentally change compile options for the *existing* schemas, and render old data unreadable. This kind of mistake would be catastrophic, but hard to find in the tests.

apps/emqx_mq/test/emqx_mq_consumer_SUITE.erl
```diff
@@ -0,0 +1,57 @@
...
+-include_lib("emqx/include/asserts.hrl").
+-include("../src/emqx_mq_internal.hrl").
+
+all() ->
+    emqx_common_test_helpers:all(?MODULE).
+
+init_per_suite(Config) ->
+    Apps = emqx_cth_suite:start(
+        [
+            {emqx_durable_storage, #{override_env => [{poll_batch_size, 1}]}},
```
ieQu1: Note: this options will soon be moved elsewhere. Is there a particular reason to set it?

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_db.erl
```diff
@@ -0,0 +1,304 @@
...
+                {error, unrecoverable, {active_consumer, ExistingConsumerRef}}
+        end
+    end),
+    get_result(TxRes).
+
+-spec find_consumer(emqx_mq_types:mq(), timestamp()) ->
+    {ok, consumer_ref()} | not_found | queue_removed.
+find_consumer(#{topic_filter := _MQTopic} = MQ, TS) ->
+    LeadershipTopic = topic_leadership(MQ),
+    case emqx_ds:dirty_read(?MQ_CONSUMER_DB, LeadershipTopic) of
```
ieQu1: It always makes sense to specify the shard when you're looking for a single entry, this way you'll send exactly one request instead of broadcasting it to all shards.

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_db.erl
```diff
@@ -0,0 +1,304 @@
...
+        n_shards => NSites * ?SHARDS_PER_SITE,
+        replication_options => #{},
+        n_sites => NSites,
+        replication_factor => ?REPLICATION_FACTOR
+    }.
+
+tx_opts(#{topic_filter := MQTopic, id := MQId}) ->
+    #{
+        db => ?MQ_CONSUMER_DB,
+        shard => {auto, {MQTopic, MQId}},
```
ieQu1: I'm not sure whether this API will always accept arbitrary terms instead of being limited to just binaries. The current relaxed behavior is sort of accidental. 

If you want to shard by a tuple that's fine, but then you're better off implementing sharding function yourself.

apps/emqx/test/emqx_channel_hookpoints_SUITE.erl
savonarola: The file appeared in 2023, so 2018 was a typo. Changed to 2023-2025

apps/emqx_durable_storage/src/emqx_ds_buffer.erl
savonarola: Before moving to "in review", I also additionaly review the PR and try to remove all debug leftovers.

apps/emqx_mq/src/emqx_mq_gc/emqx_mq_gc_worker.erl
savonarola: 🤔 isn't this document related to the OTP process design in general?
`gen_server` handles `'EXIT'` from parent explicitly and calls terminate callback https://github.com/erlang/otp/blob/5082a16e7852e92e6052940bacc9ea4f34366cdd/lib/stdlib/src/gen_server.erl#L2378. So we should never receive such a message...

apps/emqx_mq/test/emqx_mq_consumer_SUITE.erl
savonarola: The reason is to have buffers both full and not full during the tests. With the default setting, the DS delivers large batches, so we either have buffers filled at once of have to configure large buffers and prepare a lot of messages.

savonarola:
I plan to add a bit more tests to improve coverage, before merging or in a follow-up.

apps/emqx_mq/README.md
```diff
@@ -0,0 +1,169 @@
...
+A Message Queue is a collection of messages with the following properties:
+
+- Message Queue is identified by `topic/filter` from the user's perspective.
+- Has an explicit lifecycle.
+- Is automatically replenished with published messages matched with the topic filter during the queue's lifetime.
+- Clients may subscribe to the queue's `$q/topic/filter` to consume messages from the queue cooperatively.
+- A Queue is capped by time or size.
+- A Queue is not strictly ordered.A queue may have "Last-Value" semantics. When enabled,
+- Each message should have `mq-key` user property set to be saved in the queue.
+- Messages overwrite previous messages with the same key in the same topic.
```
thalesmg: Nit:
```suggestion
- A Queue is not strictly ordered. A queue may have "Last-Value" semantics. When enabled:
  - Each message should have `mq-key` user property set to be saved in the queue.
  - Messages overwrite previous messages with the same key in the same topic.
```

apps/emqx_mq/README.md
```diff
@@ -0,0 +1,169 @@
...
+## Internal Implementation
+
+Message Queue is designed to have little interaction with other parts of EMQX.
+Message Queue application
+* Uses Durable Storage to store messages and queue state.
+* Uses Mnesia table to maintain message queue index.
+* Is integrated via hooks into the EMQX core.
+
+The major components are:
+* Local subcsription registry ([emqx_mq_sub_registry.erl](./src/emqx_mq_sub_registry.erl), [emqx_mq_sub.erl](./src/emqx_mq_sub.erl), [emqx_mq_sub_buffer.erl](./src/emqx_mq_sub_buffer.erl)). These are data structures
```
thalesmg:
```suggestion
* Local subscription registry ([emqx_mq_sub_registry.erl](./src/emqx_mq_sub_registry.erl), [emqx_mq_sub.erl](./src/emqx_mq_sub.erl), [emqx_mq_sub_buffer.erl](./src/emqx_mq_sub_buffer.erl)). These are data structures
```

apps/emqx_mq/mix.exs
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def deps() do
+    [
+      {:emqx, in_umbrella: true},
+      {:emqx_utils, in_umbrella: true},
+      {:emqx_durable_storage, in_umbrella: true},
+      UMP.common_dep(:minirest),
+      UMP.common_dep(:gproc)
+    ]
```
thalesmg:
```suggestion
    UMP.deps([
      {:emqx, in_umbrella: true},
      {:emqx_utils, in_umbrella: true},
      {:emqx_durable_storage, in_umbrella: true},
      :minirest,
      :gproc
    ])
```

apps/emqx_mq/README.md
```diff
@@ -0,0 +1,169 @@
...
+### Publishing
+
+* A client publishes a message to `some/topic`.
+* An MQ hook is triggered to handle the message publication.
+* The hook looks up in the MQ registry if there are any Message Queues whose topic filter matches the message topic.
+* If yes, the hook writes the message to the corresponding Message Queues.
+
+### Subscribing/Consuming
+
+* A client subscribes to `some/topic`.
```
thalesmg:
```suggestion
* A client subscribes to `$q/some/topic`.
```

?

apps/emqx_mq/src/emqx_mq_types.erl
```diff
@@ -0,0 +1,67 @@
...
+}.
+-type ack() :: ?MQ_ACK | ?MQ_NACK | ?MQ_REJECTED.
+-type mq_topic() :: binary().
+-type mqid() :: binary().
+-type consumer_sup_id() :: mq_topic().
+
+-type dispatch_strategy() :: random | least_inflight | round_robin.
+
+-type mq() :: #{
+    id => mqid(),
```
thalesmg, line 30: Nit: of all the fields, only the id is optional?  :sweat_smile: :speak_no_evil:

apps/emqx_mq/asn.1/MessageQueue.asn
```diff
@@ -0,0 +1,45 @@
+MessageQueue DEFINITIONS ::=
+BEGIN
+
+  StreamBufferProgress ::= SEQUENCE {
+    it [0] OCTET STRING,
```
thalesmg, line 5: Q: would something like this work to [avoid explicit tags](https://asn1.io/analyzer/)?

```suggestion
MessageQueue DEFINITIONS AUTOMATIC TAGS ::=
BEGIN

  StreamBufferProgress ::= SEQUENCE {
    it OCTET STRING,
```

apps/emqx_mq/src/emqx_mq.erl
```diff
@@ -0,0 +1,225 @@
...
+                {error, Reason} ->
+                    {error, Reason};
+                {ok, NewSub} ->
+                    ok = emqx_mq_sub_registry:update(SubscriberRef, NewSub),
+                    ok;
+                {ok, NewSub, Result} ->
+                    ok = emqx_mq_sub_registry:update(SubscriberRef, NewSub),
+                    {ok, Result};
+                _ ->
+                    error({invalid_return_value_from_handler, Handler})
```
thalesmg: Nit: return the invalid value as well?

apps/emqx_mq/src/emqx_mq.erl
```diff
@@ -0,0 +1,225 @@
...
+        {ok, Messages} ->
+            {ok, Acc#{deliver => delivers(SubscriberRef, Messages) ++ Delivers}};
+        {error, recreate} ->
+            ok = recreate_sub(SubscriberRef, ClientInfo)
+    end;
+on_client_handle_info(_ClientInfo, _Message, Acc) ->
+    ?tp_debug(mq_on_client_handle_info, #{message => _Message}),
+    {ok, Acc}.
+
+on_session_disonnected(ClientInfo, #{subscriptions := Subs} = _SessionInfo) ->
```
thalesmg:
```suggestion
on_session_disconnected(ClientInfo, #{subscriptions := Subs} = _SessionInfo) ->
```

apps/emqx_mq/src/emqx_mq.erl
```diff
@@ -0,0 +1,225 @@
...
+    emqx_mq_message_db:insert(MQ, [Message]).
+
+delivers(SubscriberRef, Messages) ->
+    lists:map(
+        fun(Message0) ->
+            Message1 = emqx_message:set_headers(
+                #{?MQ_HEADER_SUBSCRIBER_ID => SubscriberRef}, Message0
+            ),
+            %% Override QoS to 1 to require ack from the client
+            Message = Message1#message{qos = ?QOS_1},
```
thalesmg, line 218: Q: if QoS is 2, do we need to downgrade it too, even though it has `PUBREC` / `PUBCOMP`?

apps/emqx_mq/src/emqx_mq_api.erl
```diff
@@ -0,0 +1,308 @@
...
+                hoconsc:ref(emqx_dashboard_swagger, cursor),
+                hoconsc:ref(emqx_dashboard_swagger, limit)
+            ],
+            responses => #{
+                200 => emqx_dashboard_swagger:schema_with_example(
+                    hoconsc:array(ref(emqx_mq_schema, message_queues_api_get)),
+                    get_message_queues_example()
+                ),
+                400 => emqx_dashboard_swagger:error_codes(
+                    ['BAD_REQUEST'], <<"Bad cursor">>
```
thalesmg: [Apparently](https://github.com/emqx/emqx/pull/15703#discussion_r2278704005), we started setting the example values in i18n too.

```suggestion
                    ['BAD_REQUEST'], ?DESC("bad_cursor_error")
```

apps/emqx_mq/src/emqx_mq_api.erl
```diff
@@ -0,0 +1,308 @@
...
+            responses => #{
+                200 => emqx_dashboard_swagger:schema_with_example(
+                    ref(emqx_mq_schema, message_queue),
+                    get_message_queue_example()
+                ),
+                404 => emqx_dashboard_swagger:error_codes(
+                    ['NOT_FOUND'], <<"Message queue not found">>
+                ),
+                400 => emqx_dashboard_swagger:error_codes(
+                    ['INVALID_CREDENTIAL'], <<"Invalid message queue">>
```
thalesmg:
```suggestion
                    ['BAD_REQUEST'], <<"Invalid message queue">>
```
?

apps/emqx_mq/src/emqx_mq_api.erl
```diff
@@ -0,0 +1,308 @@
...
+            summary => <<"Delete message queue">>,
+            description => ?DESC(message_queues_delete),
+            parameters => [topic_filter_param()],
+            responses => #{
+                204 => <<"Operation success">>,
+                404 => emqx_dashboard_swagger:error_codes(
+                    ['NOT_FOUND'], <<"Message queue not found">>
+                ),
+                400 => emqx_dashboard_swagger:error_codes(
+                    ['INVALID_CREDENTIAL'], <<"Invalid message queue">>
```
thalesmg:
```suggestion
                    ['BAD_REQUEST'], <<"Invalid message queue">>
```
?

apps/emqx_mq/src/emqx_mq_api.erl
```diff
@@ -0,0 +1,308 @@
...
+
+'/message_queues'(get, #{query_string := QString}) ->
+    EncodedCursor = maps:get(<<"cursor">>, QString, undefined),
+    Limit = maps:get(<<"limit">>, QString),
+    case decode_cursor(EncodedCursor) of
+        {ok, Cursor} ->
+            {MessageQueues, CursorNext} = get_message_queues(Cursor, Limit),
+            case CursorNext of
+                undefined ->
+                    {200, #{data => MessageQueues, meta => #{hasnext => false}}};
```
thalesmg: Nit (here and below in the file): maybe use helpers from `apps/emqx_utils/include/emqx_http_api.hrl`?

apps/emqx_mq/src/emqx_mq.erl
```diff
@@ -0,0 +1,225 @@
...
+-include_lib("snabbkaffe/include/snabbkaffe.hrl").
+
+-export([register_hooks/0, unregister_hooks/0]).
+
+-export([
+    on_message_publish/1,
+    on_session_subscribed/3,
+    on_session_unsubscribed/2,
+    on_session_resumed/2,
+    on_session_disonnected/2,
```
thalesmg:
```suggestion
    on_session_disconnected/2,
```

apps/emqx_mq/src/emqx_mq_config.erl
```diff
@@ -0,0 +1,45 @@
...
+    UpdatedMessageQueue = hocon_tconf:check_plain(Schema, UpdatedMessageQueueRaw, #{
+        atom_key => true
+    }),
+    UpdatedMessageQueue.
+
+%%------------------------------------------------------------------------------
+%% Internal functions
+%%------------------------------------------------------------------------------
+
+binary_key_map(Map) ->
```
thalesmg, line 38: Nit: there is `emqx_utils_maps:binary_key_map/1`.

apps/emqx_mq/src/emqx_mq_gc/emqx_mq_gc_worker.erl
```diff
@@ -0,0 +1,180 @@
...
+    NowMS = now_ms(),
+    RetentionPeriod = emqx_config:get([mq, regular_queue_retention_period]),
+    TimeThreshold = NowMS - RetentionPeriod,
+    maybe_create_new_generation(SlabInfo, TimeThreshold),
+    ExpiredSlabInfo =
+        lists:filtermap(
+            fun({Slab, #{until := Until}}) ->
+                case is_number(Until) andalso slab_time_to_ms(Until) =< TimeThreshold of
+                    true ->
+                        {true, {Slab, #{finished_ago => NowMS - Until}}};
```
thalesmg: If `Until` is in microseconds, shouldn't we use the converted value here when subtracting?

```suggestion
                case is_number(Until) andalso slab_time_to_ms(Until) =< TimeThreshold of
                    true ->
                        {true, {Slab, #{finished_ago => NowMS - slab_time_to_ms(Until)}}};
```

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer.erl
```diff
@@ -0,0 +1,393 @@
...
+                        error => _Error, mq_topic_filter => _MQTopicFilter
+                    }),
+                    not_found
+            end
+    end.
+
+try_register_consumer(#{id := Id, topic_filter := MQTopicFilter} = _MQ) ->
+    case emqx_mq_registry:find(MQTopicFilter) of
+        {ok, #{id := Id}} ->
+            case global:register_name(global_name(Id), self()) of
```
thalesmg, line 378: N.B.: the default resolution function used by `global` is [`random_exit_name/3`](https://www.erlang.org/doc/apps/kernel/global.html#random_exit_name/3), which will kill one of the consumers with `kill`, giving it no chance to run `terminate`.

apps/emqx_mq/src/emqx_mq_internal.hrl
```diff
@@ -0,0 +1,63 @@
...
+%% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+-ifndef(EMQX_MQ_INTERNAL_HRL).
+-define(EMQX_MQ_INTERNAL_HRL, true).
+
+-include_lib("snabbkaffe/include/snabbkaffe.hrl").
+
+% -define(tp_debug(KIND, EVENT), ?tp_ignore_side_effects_in_prod(KIND, EVENT)).
+-define(tp_debug(KIND, EVENT), ?tp(warning, KIND, EVENT)).
```
thalesmg:
```suggestion
-define(tp_debug(KIND, EVENT), ?tp_ignore_side_effects_in_prod(KIND, EVENT)).
%% -define(tp_debug(KIND, EVENT), ?tp(warning, KIND, EVENT)).
```

?

apps/emqx_mq/asn.1/MessageQueue.asn
savonarola: This is intentional, after some discussion with Dmitrii. Explicit tags allow to remove fields in the future.

apps/emqx_mq/src/emqx_mq.erl
savonarola: Other value is a logical error, it should never happen, but if it does we better to note it earlier.
Let's just remove this case to avoid confusion.

apps/emqx_mq/src/emqx_mq.erl
savonarola: Yes, message queues works with qos1 only

apps/emqx_mq/src/emqx_mq_api.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_config.erl
savonarola: This version is intentionally not deep

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer.erl
savonarola: I think that's ok in init

apps/emqx_mq/src/emqx_mq_types.erl
savonarola: Yes, not to make a separate or nested type for mq creation (id is autoassigned).

apps/emqx_mq/README.md
savonarola: ✅

apps/emqx_mq/README.md
savonarola: ✅

apps/emqx_mq/mix.exs
savonarola: ✅

apps/emqx_mq/README.md
savonarola: ✅ Clarified

apps/emqx_mq/src/emqx_mq.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_api.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_api.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_api.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_gc/emqx_mq_gc_worker.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_internal.hrl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer.erl
thalesmg: The process chosen to be killed might not be the one initializing...  :sweat_smile:

thalesmg: as we discussed, progress might be lost due to the likely scenario where this arises

There's a split into cliques A and B.
A and B have different consumer processes registered.
The split heals.
global chooses one of the consumers to be kill'ed

savonarola: That will cause some progress loss in case of cluster heal, I think this is acceptable so far.

savonarola:
The coverage is much better now (see updated codecov comment)

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_server.erl
```diff
@@ -0,0 +1,582 @@
...
+                {ok, [{ds_ack, MessageId}], State};
+            ?MQ_REJECTED ->
+                {ok, [], redispatch_on_reject(State1, SubscriberRef, MessageId)}
+        end
+    else
+        _ ->
+            ?tp(error, mq_consumer_handle_ack_unknown_subscriber, #{
+                subscriber_ref => SubscriberRef, message_id => MessageId, ack => Ack
+            }),
+            State0
```
thalesmg: This return value does not match the `{ok, [event()], t()}` type.

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_server.erl
```diff
@@ -0,0 +1,582 @@
...
+%%--------------------------------------------------------------------
+
+add_new_messages(#state{messages = Messages0} = State, MessagesWithIds) ->
+    {MessageIds, Messages} = lists:foldl(
+        fun({MessageId, Message}, {IdsAcc, MessagesAcc}) ->
+            {[MessageId | IdsAcc], MessagesAcc#{MessageId => Message}}
+        end,
+        {[], Messages0},
+        MessagesWithIds
+    ),
```
thalesmg: Micro suggestion (unless it's important to use the reverse the order of `MessagesWithIds` in `MessageIds`).

```suggestion
    {MessageIds, Messages} = lists:mapfoldl(
        fun({MessageId, Message}, Acc) ->
            {MessageId, Acc#{MessageId => Message}}
        end,
        Messages0,
        MessagesWithIds
    ),
```

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_server.erl
```diff
@@ -0,0 +1,582 @@
...
+        #subscriber_timeout{subscriber_ref = SubscriberRef}
+    ),
+    SubscriberData#{timeout_tref => TimeoutTRef}.
+
+update_last_ack(SubscriberData0) ->
+    SubscriberData0#{last_ack_ts => now_ms_monotonic()}.
+
+cancel_subscriber_timeout(#{timeout_tref := TimeoutTRef} = SubscriberData) ->
+    emqx_utils:cancel_timer(TimeoutTRef),
+    SubscriberData#{timeout_tref => undefined}.
```
thalesmg: Nit:

```suggestion
    SubscriberData#{timeout_tref := TimeoutTRef}.

update_last_ack(SubscriberData0) ->
    SubscriberData0#{last_ack_ts := now_ms_monotonic()}.

cancel_subscriber_timeout(#{timeout_tref := TimeoutTRef} = SubscriberData) ->
    emqx_utils:cancel_timer(TimeoutTRef),
    SubscriberData#{timeout_tref := undefined}.
```

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_server.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_server.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_server.erl
savonarola: This is also an unreachable code branch, so I decided to just drop it

apps/emqx_mq/src/perf/emqx_mq_perf.erl_
thalesmg, line 1: Maybe rename this file to `.erl` and wrap the module in something like `-ifdef(MQ_PERF).`?

apps/emqx_mq/src/emqx_mq_registry.erl
```diff
@@ -0,0 +1,264 @@
...
+
+mq_ets_record_stream(Key) ->
+    fun() ->
+        case next_key(Key) of
+            '$end_of_table' ->
+                [];
+            NextKey ->
+                [ets:lookup(?MQ_REGISTRY_INDEX_TAB, NextKey) | mq_ets_record_stream(NextKey)]
+        end
+    end.
```
thalesmg, line 243: Q: wouldn't `emqx_utils_stream:ets` fit here?

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_streams.erl
```diff
@@ -0,0 +1,501 @@
...
+        case emqx_mq_consumer_stream_buffer:handle_ack(SB, StreamMessageId) of
+            {ok, SB1} ->
+                State1 = State0#{
+                    shards => Shards0#{Shard => ShardState0#{stream_buffer => SB1}}
+                },
+                State = save_progress_update(State1, Shard),
+                CS#cs{state = State};
+            finished ->
+                Streams = maps:remove(Stream, Streams0),
+                ShardState = ShardState0#{status => finished, generation => Generation},
```
thalesmg: Q: should this drop the other map keys to make the term slimmer and closer to the typespec?  (also more similar to what is done a dozen lines below).

```suggestion
                ShardState = #{status => finished, generation => Generation},
```

apps/emqx_mq/src/emqx_mq_message_db.erl
```diff
@@ -0,0 +1,314 @@
...
+]).
+
+-define(MQ_MESSAGE_DB_APPEND_RETRY, 5).
+-define(MQ_MESSAGE_DB_LTS_SETTINGS, #{
+    %% "topic/TOPIC/key/СOMPACTION_KEY"
+    lts_threshold_spec => {simple, {100, 0, 0, 100, 0, 100}}
+}).
+-define(MQ_MESSAGE_DB_TOPIC(MQ_TOPIC, MQ_ID, KEY), [
+    <<"topic">>, MQ_TOPIC, MQ_ID, <<"key">>, KEY
+]).
```
thalesmg: IIUC:

```suggestion
    %% "topic/MQ_TOPIC/MQ_ID/key/СOMPACTION_KEY"
    lts_threshold_spec => {simple, {100, 0, 0, 100, 0}}
}).
-define(MQ_MESSAGE_DB_TOPIC(MQ_TOPIC, MQ_ID, KEY), [
    <<"topic">>, MQ_TOPIC, MQ_ID, <<"key">>, KEY
]).
```

?

apps/emqx_mq/src/emqx_mq_message_db.erl
```diff
@@ -0,0 +1,314 @@
...
+
+-spec regular_db_slab_info() -> #{emqx_ds:slab() => emqx_ds:slab_info()}.
+regular_db_slab_info() ->
+    emqx_ds:list_slabs(?MQ_MESSAGE_REGULAR_DB).
+
+-spec initial_generations(emqx_mq_types:mq() | emqx_mq_types:mq_handle()) ->
+    #{emqx_ds:slab() => emqx_ds:generation()}.
+initial_generations(MQ) ->
+    SlabInfo = emqx_ds:list_slabs(db(MQ)),
+    maps:from_list(lists:reverse(lists:sort(maps:keys(SlabInfo)))).
```
thalesmg, line 209: Since keys in a map are unique and don't have order:

```suggestion
    maps:from_list(maps:keys(SlabInfo)).
```

?

apps/emqx_mq/src/emqx_mq_message_db.erl
```diff
@@ -0,0 +1,314 @@
...
+                        Shard, fun(Msgs) -> [{Topic, Value} | Msgs] end, [{Topic, Value}], Acc
+                    )
+            end
+        end,
+        #{},
+        Messages
+    ),
+    groupped_by_shard_to_list(ByShard).
+
+groupped_by_shard_to_list(ByShard) ->
```
thalesmg:
```suggestion
grouped_by_shard_to_list(ByShard) ->
```

apps/emqx_mq/src/emqx_mq_state_storage.erl
```diff
@@ -0,0 +1,539 @@
...
+    TransOpts = trans_opts(Id),
+    TxRes = emqx_ds:trans(
+        TransOpts,
+        fun() ->
+            case open_mq_state_tx(Id) of
+                undefined ->
+                    not_found;
+                MQStateRec0 ->
+                    MQ0 = emqx_ds_pmap:collection_get(?col_mq, ?mq_key, MQStateRec0),
+                    MQ = maps:merge(MQ0, MQFields),
```
thalesmg, line 236: Q: should we filter out the immutable fields (id, topic_filter, is_lastvalue, ...) here?

apps/emqx_mq/src/emqx_mq_state_storage.erl
```diff
@@ -0,0 +1,539 @@
...
+                }}
+        },
+    {ok, Bin} = 'MessageQueue':encode('ShardProgress', Rec),
+    Bin;
+pmap_encode_val(?pn_shard_progress, _Key, #{
+    status := finished,
+    generation := Generation
+}) ->
+    Rec = {v1, {finished, #'FinishedShardProgress'{generation = Generation}}},
+    {ok, Bin} = 'MessageQueue':encode('ShardProgress', Rec),
```
thalesmg: Q: should `Rec` be a `#ShardProgress{}`, similar to the clause above?

apps/emqx_mq/src/emqx_mq_sub_buffer.erl
```diff
@@ -0,0 +1,61 @@
...
+new() ->
+    gb_trees:empty().
+
+-spec add(t(), emqx_types:message()) -> t().
+add(Tree, Msg) ->
+    {Slab, SlabMessageId} = emqx_message:get_header(?MQ_HEADER_MESSAGE_ID, Msg),
+    gb_trees:insert({SlabMessageId, Slab}, Msg, Tree).
+
+-spec take(t(), non_neg_integer()) -> {[{emqx_mq_types:message_id(), emqx_types:message()}], t()}.
+take(Tree, N) ->
```
thalesmg: Nit:

```suggestion
take(Tree, N) when N >= 0 ->
```

apps/emqx_mq/src/emqx_mq_sup.erl
```diff
@@ -0,0 +1,100 @@
...
+    ChildSpecs = [],
+    {ok, {SupFlags, ChildSpecs}}.
+
+consumer_sup_child_spec() ->
+    #{
+        id => ?CONSUMER_SUP,
+        start => {?MODULE, start_consumer_sup, []},
+        restart => permanent,
+        shutdown => 5000,
+        type => supervisor,
```
thalesmg:
```suggestion
        shutdown => infinity,
        type => supervisor,
```

apps/emqx_mq/src/emqx_mq_sup.erl
```diff
@@ -0,0 +1,100 @@
...
+        modules => [?MODULE]
+    }.
+
+gc_sup_child_spec() ->
+    #{
+        id => ?GC_SUP,
+        start => {?MODULE, start_gc_sup, []},
+        restart => permanent,
+        shutdown => 5000,
+        type => supervisor,
```
thalesmg:
```suggestion
        shutdown => infinity,
        type => supervisor,
```

apps/emqx_mq/src/emqx_mq_registry.erl
savonarola: `emqx_utils_stream:ets` was here originally. But it is for reading the whole table with select/match; here we need to start from a particular key. Maybe we could somehow use `emqx_utils_stream:ets` (because it is just head and tail generator), but IMO that would be unobvious.

apps/emqx_mq/src/emqx_mq_message_db.erl
savonarola: If, say, for a shard `<<"0">>` we have slabs `{<<"0">>, 2}, {<<"0">>, 1}`, we want `{<<"0">> => 1}` in the result. We make sort so that slabs with smaller gens override those with greater gens.

apps/emqx_mq/src/emqx_mq_message_db.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_state_storage.erl
savonarola: That's quite funny because these lines are covered 🙃

The reason the code works correctly is that 
`{v1, {finished, ...}}` and `#'ShardProgress'{progress = {finished, ...}}` have the same tuple structure `{..., {finished, ...}}` and the first item (the name of the root record) is ignored by the generated encoder because it is passed explicitly.

savonarola: ✅

apps/emqx_mq/src/perf/emqx_mq_perf.erl_
savonarola: ✅

apps/emqx_mq/src/emqx_mq_consumer/emqx_mq_consumer_streams.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_state_storage.erl
savonarola: This module is considered to be low-level, not for direct usage by other parts of the app; the filtering is in `emqx_mq_registry` module.

apps/emqx_mq/src/perf/emqx_mq_perf.erl_
thalesmg, line 1: Now this file can be deleted?

apps/emqx_mq/src/emqx_mq_sub.erl
```diff
@@ -0,0 +1,538 @@
...
+    NPublish = max_inflight(MQ) - map_size(Inflight0),
+    {MessagesWithIds, Buffer} = emqx_mq_sub_buffer:take(Buffer0, NPublish),
+    {Inflight, Messages0} = lists:foldl(
+        fun({MessageId, Message}, {InflightAcc, Msgs}) ->
+            {InflightAcc#{MessageId => Message}, [Message | Msgs]}
+        end,
+        {Inflight0, []},
+        MessagesWithIds
+    ),
+    Messages = lists:reverse(Messages0),
```
thalesmg, line 229: Micro nit: sounds like `lists:mapfoldl` would simplify things slightly.

apps/emqx_mq/src/emqx_mq_message_db.erl
```diff
@@ -0,0 +1,314 @@
...
+mq_message_topic(#{topic_filter := TopicFilter, id := Id} = _MQ, Key) ->
+    ?MQ_MESSAGE_DB_TOPIC(TopicFilter, Id, Key).
+
+db(#{is_lastvalue := true} = _MQ) ->
+    ?MQ_MESSAGE_LASTVALUE_DB;
+db(#{is_lastvalue := false} = _MQ) ->
+    ?MQ_MESSAGE_REGULAR_DB.
+
+insert_generation(?MQ_MESSAGE_LASTVALUE_DB) ->
+    1;
```
thalesmg, line 312: Q: shouldn't this be dynamic (i.e., depend on the current generation value)?

savonarola: No, last-value queues do not switch generations, because their data is removed with topic del opts.

savonarola:
Will address the nit-s in a follow-up.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9408)

PR #15795, opened 2025-08-28
chore: update ELP helper script to match mix build

savonarola:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9314)

PR #15809, opened 2025-08-29
test: fix flaky test

zmstone:
1. use unique test client id per test case/function
2. ensure non-clean session is gone after test completed


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9097)

PR #15818, opened 2025-09-02
fix(acl): `{allow|deny, all}` should match ANY topic

zmstone:
Fixes [EMQX-14665](https://emqx.atlassian.net/browse/EMQX-14665)

Release version: 5.8.9

## Summary

Previously, {allow/deny, all} rule is internally translated to match topic '#'.
However, according to MQTT spec, dollar sign prefixed topic should not match + or #.
This makes the `all` rule faile to match topics like $testtopic/1

This commit changes to use a special internal value for the 'all' rule so that ANY topic will be matched.

## PR Checklist


apps/emqx_auth/src/emqx_authz/emqx_authz_rule.erl
```diff
@@ -140,7 +141,7 @@ compile(Permission, Who, Action, TopicFilters) ->
 compile({Permission, all}) when
     ?IS_PERMISSION(Permission)
 ->
-    {Permission, all, all, [compile_topic(<<"#">>)]};
+    {Permission, all, all, [?ANY_TOPIC]};
 compile({Permission, Who, Action, TopicFilters}) when
```
JimMoen, line 147: should `'all'` also be supported in specific rule?
It only supports `topic() | topic_filter() | {eq, topic() | topic_filter()}`.


The existing topicfilter list cannot handle the following usage.
`{allow, {username, "foo"}, publish, all}.`
`{deny, {username, "bar"}, subscrube, all}.`

zmstone: make sense. I'll add it.

zmstone: Actually, a second thought:
It's possible for acl file because `all` is an Erlang atom.

Supporting it would cause confusion as we also have raw rules in JSON format, from which, we cannot tell if `all` is a topic or a special wildcard.

JimMoen: they should be different in json
```json
{
...
  "topics": ["t/1", "foo/bar"], // current supported

  "topics": "all" // means the atom 'all'
...
}
```

zmstone: the problem is, it allows both `topic` and `topics`
I'll try to add a special handling for `topics` to support `all`, but not `topic`.

zmstone: Unfortunately, raw rules' `{Permission, Who, Action, [Topic]}` is mapped to `[{Permission, Who, Action, Topic}]`.
So there is no `topics` exposed in API response, and no `topics` accepted from API request.

I will not try to expand the scope of 5.8 for raw rules.

apps/emqx_auth/test/emqx_authz/emqx_authz_rule_SUITE.erl
```diff
@@ -71,7 +71,11 @@ t_compile(_) ->
     % Some of the following testcase are relying on the internal representation of
     % `emqx_template:t()`. If the internal representation is changed, these testcases
     % may fail.
-    ?assertEqual({deny, all, all, [['#']]}, emqx_authz_rule:compile({deny, all})),
+    ?assertEqual({deny, all, all, ['ALL_TOPICS']}, emqx_authz_rule:compile({deny, all})),
+    ?assertEqual({deny, all, all, ['ALL_TOPICS']}, emqx_authz_rule:compile({deny, all, all, all})),
+    ?assertEqual(
+        {deny, all, subscribe, ['ALL_TOPICS']}, emqx_authz_rule:compile({deny, all, subscribe, all})
```
thalesmg, line 77: Nit: use `?ALL_TOPICS` from `emqx_authz.hrl`?

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9396)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10530)

PR #15819, opened 2025-09-02
250902 sync r58 to r59

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9172)

PR #15828, opened 2025-09-02
build(asn1): recompile if compile opts change or output is missing

thalesmg:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9332)

PR #15826, opened 2025-09-02
fix(kafka consumer/source): treat group acl denied as healthy connection

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14668

Release version: 6.0.0

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [na] The changes are covered with new or existing tests
  - tested manually; didn't add automated tests just yet to avoid conflicts with bridge v1 refactorings in master
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files

PR #15834, opened 2025-09-03
fix(kafka consumer/source): treat group acl denied as healthy connection (r58)

thalesmg:
Port of https://github.com/emqx/emqx/pull/15826

Fixes https://emqx.atlassian.net/browse/EMQX-14668

Release version: 5.8.x


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9224)

PR #15841, opened 2025-09-03
feat(mt): always count sessions if fewer than 1 k

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14591

Release version: 6.0.0

## Summary

The original motivation for using `set` for `?COUNTER_TAB` was to be able to use
`mnesia:dirty_update_counter`.  But this function apparently works with `ordered_set` too:

https://github.com/emqx/otp/blob/7ebce9b1e8bb6b10637410600d1d65a5a9b79fc7/lib/mnesia/src/mnesia.erl#L2769

Since `ets:select` is efficient on `ordered_set` ETS tables when the key is partially
bound (such as fixing the namespace), we changed the types of those tables here for more
efficient scans/counts.

This shouldn't be a problem for rolling upgrades: the old table will not be created in the
new nodes even though it's part of the schema table, and the data in the old table does
not need to be migrated because the clients connected to the old node will simply
disconnect and eventually reconnect to new nodes, at which point they'll populate the new
counter table.

## PR Checklist

### DIFF
--- a/changes/ee/feat-15841.en.md
+++ b/changes/ee/feat-15841.en.md
@@ -1,0 +1,3 @@
+Improved the refresh rate of the session count for namespaced sessions.  Now, when there are less than 1000 connections a given namespace, its count is updated on demand, otherwise, it's updated every 5 seconds.
+
+During a rolling upgrade with versions prior to 6.0, the namespace session count will be inconsistent as we changed the internal tables that are used to track them.  As the upgrade progresses, clients connected to nodes running older versions will be disconnected and reconnect to new nodes, and the count shall eventually become consistent at the end of the upgrade when there are no longer old nodes.


apps/emqx_mt/src/emqx_mt_state.erl
```diff
@@ -321,6 +321,10 @@ with_ccache(Ns) ->
     case lookup_counter_from_cache(Ns) of
         false ->
             update_ccache(Ns);
+        Cnt when Cnt < 1_000 ->
```
zmstone, line 324: Should also check the total size of the ~cache~ `emqx_mt_counter` table.
Otherwise ets select can be very expensive when there are a lot of namespaces

zmstone: Actually, it would be nice to have some tests to see how `do_count_clients/1` behaves when there are large amount of namespaces.

thalesmg: After our discussion, we decided on changing the table types to `ordered_set`, so that `ets:select` with partially bound key patterns will be efficient.

https://github.com/emqx/emqx/pull/15841/commits/099d74b855f3b58dbe249a8696a127b01d31efbc

zmstone:
please add a release note to describe this change and the impact on user experience during rolling upgrade.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9336)

PR #15843, opened 2025-09-04
fix(mix): fix bad automatic merge resolution (r58)

thalesmg:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9228)

PR #15845, opened 2025-09-04
feat(mqtt connector): add static clientid & username & password tuples

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14679

Release version: 5.8.9

## Summary

Some servers require unique usernames for each clientid, hence the need to change the previous simple clientid into a clientid & username & password tuple.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt.app.src
+++ b/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt.app.src
@@ -1,7 +1,7 @@
 %% -*- mode: erlang -*-
 {application, emqx_bridge_mqtt, [
     {description, "EMQX MQTT Broker Bridge"},
-    {vsn, "0.2.9"},
+    {vsn, "0.2.10"},
     {registered, []},
     {applications, [
         kernel,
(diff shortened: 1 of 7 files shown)


apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt.app.src
```diff
@@ -1,7 +1,7 @@
 %% -*- mode: erlang -*-
 {application, emqx_bridge_mqtt, [
     {description, "EMQX MQTT Broker Bridge"},
-    {vsn, "0.2.9"},
```
zmstone, line 4: more like a 0.3 ?

zmstone:
please notify frontend team to make the changes in dashboard.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt.app.src
thalesmg: those internal versions don't matter much, I think; no need to follow any kind of semantic versioning, since the goal is just to track changes between releases

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9267)

PR #15852, opened 2025-09-05
ci: fix performance test workflow

id:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9265)

PR #15862, opened 2025-09-08
chore: fix flaky test

savonarola:
This is an attempt to actually fix flaky test `emqx_mgmt_api_clients_SUITE.t_list_clients_v2_limit`

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -934,18 +934,21 @@ do_list_clients_v2(Nodes, Cursor = #{type := ?CURSOR_TYPE_ETS, node := Node}, QS
         end
     end;
 do_list_clients_v2(Nodes, _Cursor = #{type := ?CURSOR_TYPE_DS, iterator := Iter0}, QString0, Acc0) ->
-    #{limit := Limit} = Acc0,
+    #{limit := Limit, remaining := Remaining0} = Acc0,
     {Rows0, Iter} = emqx_persistent_session_ds_state:session_iterator_next(Iter0, Limit),
     NewCursor = next_ds_cursor(Iter),
     Rows1 = check_for_live_and_expired(Rows0),
-    Rows = run_filters(Rows1, QString0),
+    Rows2 = run_filters(Rows1, QString0),
+    Rows = lists:sublist(Rows2, Remaining0),
     Acc1 = maps:update_with(rows, fun(Rs) -> [{undefined, Rows} | Rs] end, Acc0),
     Acc = #{n := N} = maps:update_with(n, fun(N) -> N + length(Rows) end, Acc1),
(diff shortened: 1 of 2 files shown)


apps/emqx_durable_storage/src/emqx_ds_optimistic_tx.erl
```diff
@@ -394,10 +394,9 @@ handle_event(info, {'EXIT', _, Reason}, _State, _Data) ->
         normal -> keep_state_and_data;
         _ -> {stop, shutdown}
     end;
-handle_event(enter, _OldState, ?leader(?pending), D0 = #d{flush_interval = T}) ->
-    %% Schedule unconditional flush after the given interval:
+handle_event(enter, _OldState, ?leader(?pending), D0) ->
     D = D0#d{entered_pending_at = erlang:monotonic_time(millisecond)},
-    {keep_state, D, {state_timeout, T, ?timeout_flush}};
```
savonarola: We only enter `?leader(?pending)` from `?leader(?idle)` setting early timeout explicitly

apps/emqx_durable_storage/src/emqx_ds_optimistic_tx.erl
```diff
@@ -574,8 +571,29 @@ finalize_add_pending(_, D = #d{n_items = N, max_items = MaxItems}, Actions) when
     handle_flush(D, Actions);
 finalize_add_pending(?idle, D = #d{idle_flush_interval = IdleInterval}, Actions) ->
     %% Schedule early flush if the shard is idle:
-    Timeout = {timeout, IdleInterval, ?timeout_flush},
+    Timeout = {state_timeout, IdleInterval, ?timeout_flush},
```
savonarola: The original problem which made `emqx_mgmt_api_clients_SUITE.t_list_clients_v2_limit` flaky was the following:
* Several clients disconnect almost at once.
* They try to clear will timers and do this within `idle_flush_interval`.
* Will timers are stored in the DS, so two tx-es with deletions arrive into this process.
* One moves the process `idle` -> `pending` and schedules early flush.
* The second arrives into `pending` state and **cancels** early flush (`{timeout, ..., ...}` actions are cancelled by any other event, is this intentional? 🤔).
* The flush happens only after `state_timeout` which is significant: 500ms.

So in the test, some clients were alive quite long after disconnect command, causing flakiness.

Also, it seems a bit counter-intuitive that we may have large commit times when the load is quite small (just a few writes).

ieQu1: This logic works as intended. Please revert changes in this module. Tuning flushing logic is planned, but it requires proper benchmarking and prototyping.

savonarola:
Symptomatic fixes were merged as https://github.com/emqx/emqx/commit/add2eb776b3a99e786737623a021c0772cc70be7

PR #15873, opened 2025-09-09
Use tail recursion for message processing

ieQu1:
Fixes [EMQX-14689](https://emqx.atlassian.net/browse/EMQX-14689)

Release version: 6.0.0

## Summary

Optimize hot code paths related to message transformation and delivery. 

## PR Checklist

- [x] Not visible to the user. ~Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files~
- [x] No schema change ~Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)~

[EMQX-14689]: https://emqx.atlassian.net/browse/EMQX-14689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/emqx/src/emqx_connection.erl
```diff
@@ -904,10 +904,35 @@ handle_outgoing(Packets, State = #state{channel = _Channel}) ->
     Res.
 
 do_handle_outgoing(Packets, State) when is_list(Packets) ->
-    N = length(Packets),
-    send(N, [serialize_and_inc_stats(State, Packet) || Packet <- Packets], State);
+    BufSize = 16#100_000,
```
keynslug: Q: Perhaps worth a comment explaining how this value was chosen?

apps/emqx/src/emqx_connection.erl
keynslug, line 1: Would be good to have it in `emqx_socket_connection` also. Porting should be a matter of copy and paste, more or less.

apps/emqx/src/emqx_connection.erl
ieQu1: 1MB was chosen for no particular reason, based on some non-scientific benchmarks. Previously the limit was infinite.

ieQu1: Added a TODO.

apps/emqx/src/emqx_connection.erl
ieQu1: Done.

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9412)

PR #15877, opened 2025-09-09
ci(mix): make args parsing stricter

thalesmg:
So we can detect call errors more readily.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9394)

PR #15899, opened 2025-09-11
250911 delete acl cache when disconnected

zmstone:
Fixes [EMQX-14702](https://emqx.atlassian.net/browse/EMQX-14702)

Release version: 5.8.9, 5.9.2, 5.10.1, 6.0.0

## PR Checklist

[EMQX-14702]: https://emqx.atlassian.net/browse/EMQX-14702?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -2918,6 +2918,7 @@ ensure_disconnected(
         clientinfo = ClientInfo
     }
 ) ->
+    ok = emqx_authz_cache:empty_authz_cache(),
     NConnInfo = ConnInfo#{disconnected_at => erlang:system_time(millisecond)},
     ok = run_hooks('client.disconnected', [ClientInfo, Reason, NConnInfo]),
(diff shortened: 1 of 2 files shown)


apps/emqx/src/emqx_channel.erl
```diff
@@ -2918,6 +2918,7 @@ ensure_disconnected(
         clientinfo = ClientInfo
     }
 ) ->
+    ok = emqx_authz_cache:empty_authz_cache(),
```
qzhuyan, line 2921: if session is  terminated, process will exit, this call is unnecessary I think.

just a idea for fine tune, 
Is there a way to call this only when the connection is disconnected but session is not terminated?

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9496)

PR #15900, opened 2025-09-11
[r59] perf: copy strings to avoid retaining large packet binaries

zmstone:
Fixes [EMQX-14703](https://emqx.atlassian.net/browse/EMQX-14703)
Ported from r510 PR: https://github.com/emqx/emqx/pull/15907

Release version: 5.9.2

## Summary

Previously, UTF-8 fields parsed from MQTT packets (e.g. client ID, username, password, and especially topic) were sub-binaries of the original packet buffer. Because these values may live as long as the session (e.g. stored in caches), the entire packet binary could be kept alive unnecessarily, leading to excessive RAM usage.

This change ensures all parsed UTF-8 strings are copied into standalone binaries, allowing the original packet binary to be released promptly.

## Performance Results (larger scale test is yet to be done)

In dev environment, I tested with **10,000 clients**, each publishing once per second to a randomized (fixed-length) topic
 with QoS 0.

### Test 1: 1 KB payload, 90-byte topic
- **Before patch**  
  - beam.smp RES: ~3.2 GB  
  - beam.smp CPU: ~248%  
- **After patch**  
  - beam.smp RES: ~1.9 GB (↓ ~40%)  
  - beam.smp CPU: ~207% (↓ ~16%)  

Significant improvement in both memory footprint and CPU usage.  

### Test 2: 2-byte payload, 270-byte topic
- **With patch**  
  - beam.smp RES: ~2.1 GB  
  - beam.smp CPU: ~270%  
- **Without patch**  
  - beam.smp RES: ~2.1 GB  
  - beam.smp CPU: ~276%  

No measurable CPU penalty for copying larger topic strings.  
Memory use remains stable and predictable.

## PR Checklist

[EMQX-14703]: https://emqx.atlassian.net/browse/EMQX-14703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_frame.erl
+++ b/apps/emqx/src/emqx_frame.erl
@@ -71,6 +71,7 @@
 -define(SERIALIZE_ERR(Reason), ?THROW_SERIALIZE_ERROR(Reason)).
 
 -define(MULTIPLIER_MAX, 16#200000).
+-define(SMALL_BINARY, 64).
 
 -dialyzer({no_match, [serialize_utf8_string/3]}).
 
@@ -356,7 +357,7 @@ do_parse_connect(
         _ ->
             ?PARSE_ERR(#{
                 cause => malformed_connect,
-                unexpected_trailing_bytes => size(Rest7)
+                unexpected_trailing_bytes => byte_size(Rest7)
             })
     end;
 do_parse_connect(_ProtoName, _IsBridge, _ProtoVer, Bin, _StrictMode) ->
@@ -384,7 +385,7 @@ parse_packet(
     Header = #mqtt_packet_header{type = ?PUBLISH, qos = QoS},
     #options{strict_mode = StrictMode, version = Ver}
 ) ->
-    {TopicName, Rest} = parse_utf8_string(Bin, StrictMode, _Cause = invalid_topic),
+    {TopicName, Rest} = parse_utf8_string(Bin, StrictMode, _Cause = invalid_topic, publish_topic),
     {PacketId, Rest1} =
         case QoS of
             ?QOS_0 -> {undefined, Rest};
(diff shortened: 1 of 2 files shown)


apps/emqx/src/emqx_frame.erl
```diff
@@ -713,7 +713,7 @@ parse_optional(Bin, _F, false) ->
 parse_utf8_string(<<Len:16/big, Str:Len/binary, Rest/binary>>, true) ->
     {validate_utf8(Str), Rest};
 parse_utf8_string(<<Len:16/big, Str:Len/binary, Rest/binary>>, false) ->
-    {Str, Rest};
+    {binary:copy(Str), Rest};
```
keynslug: Just thinking out loud.

I suspect this might have an uncontrollable, too broad impact on performance.

It's easy to imagine scenarios where common workloads will have (slightly?) larger memory footprint on average, because those workloads do not usually hold onto binaries for long but will still require more copying. This is of course a speculation and needs careful performance analysis, but it still feels safer to target specific cases (e.g. constructing conninfo / will message) on upper levels.

Simplicity of this general approach is admirable however.

zmstone: This is the reason why the PR is still in draft mode.
I'll run some tests to compare CPU usage.

zmstone: It's quite certain though, that client ID and username copy is necessary when there is a Will payload, because they continue to live after Will is deleted.

zmstone: did some tests, test result is added in PR description

zmstone: Requested QA team to conduct larger scale tests:

1. CONNECT requests with/without Will message.
2. PUBLISH with long topic, small payload.
3. PUBLSH with short topic, large payload.

changes/ee/fix-15900.en.md
thalesmg, line 1: `perf-*.en.md`?

apps/emqx/src/emqx_frame.erl
keynslug: Interesting results! Thanks for sharing.

keynslug: > I tested with 10,000 clients, each publishing once per second to a randomized (fixed-length) topic
with QoS 0.

So there was no routing / hooks? Is it essentially more memory-effective authz cache that gave that much of an improvement?

zmstone: messages are all dropped due to no subscriber. no auth, no acl rules (allow all), no rules, no hooks.

keynslug: I see. I guess it should be the cache then since it's enabled by default and populated even if no authz is configured IIUC.

qzhuyan:
I think these are small copies, erlang allocators could handle it well.

qzhuyan:
note1: if it is about the sub binary that keep refcnt binary unreleased, maybe the next GC will shrink the mem usage.  so we may comparing the result after global gc call? 
note2:  OTP27 says about behaviour changes, unsure about how much.

qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9645)

PR #15902, opened 2025-09-11
upgrade to emqtt 1.13.8

zmstone:
This included fixes for MQTT client to address issues in mqtt bridges.

- Respect PINGRESP timeout
- Fix hanging connection when TLS alert happens during `mqtt_connect`.
- Fix hanging connection when TLS disconnect happens during `mqtt_connect`

Fixes <issue-or-jira-number>

Release version: 5.8.9

## PR Checklist

### DIFF
--- a/changes/ee/fix-15902.en.md
+++ b/changes/ee/fix-15902.en.md
@@ -1,0 +1,5 @@
+Upgrade MQTT client library to 1.13.8
+
+This improves MQTT bridge connectivity with:
+- Connector will automatically reconnect when peer broker does not reply PINGRESP.
+- Bridge over TLS failure is more promptly handled if connection breaks while waiting for CONNACK.
--- a/apps/emqx/rebar.config
+++ b/apps/emqx/rebar.config
@@ -47,7 +47,7 @@
             {meck, "0.9.2"},
             {proper, "1.4.0"},
             {bbmustache, "1.10.0"},
-            {emqtt, {git, "https://github.com/emqx/emqtt", {tag, "1.13.7"}}}
+            {emqtt, {git, "https://github.com/emqx/emqtt", {tag, "1.13.8"}}}
         ]},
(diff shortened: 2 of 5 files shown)


thalesmg (dismissed):

savonarola (dismissed):

changes/ee/fix-15902.en.md
```diff
@@ -0,0 +1,6 @@
+Upgrade MQTT client library to 1.13.8
+
+This fixes auto reconnect of MQTT bridge when:
```
thalesmg: I'm not sure it manifests in any way for bridges, as we don't use emqtt's reconnect there.  :thinking:

thalesmg: The only thing is that if such failures happen, they take longer than needed to actually shut down the client (~~not sure which timeout is responsible for that~~ probably `connack_timeout`).

zmstone: All the fixed issues seem to cause TCP or SSL connection to hang preventing the resource manager to detect connection state hence no health-check retry ?

thalesmg: > they take longer than needed to actually shut down the client

IIRC they do shut down, probably due to `connack_timeout` timer.  (at least the ones I patched due to using emqtt as a client directly in crl/ocsp suites, not caught in bridge suites)

zmstone: How about this:

```
This improves MQTT bridge connectivity with:
- Connector will automatically reconnect when peer broker does not reply PINGRESP.
- Bridge over TLS failure is more promptly handled if connection breaks while waiting for CONNACK.
```

thalesmg: SGTM

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9539)

PR #15909, opened 2025-09-12
fix(mysql): fix connectors getting stuck in "connecting" state

savonarola:


apps/emqx_mysql/src/emqx_mysql.erl
```diff
@@ -237,19 +239,30 @@ mysql_function(_) ->
...
+                        {error, Reason} ->
+                            %% do not log error, it is logged in prepare_sql_to_conn
+                            {?status_connecting, emqx_utils:readable_error_msg(Reason)}
+                    end;
+                false ->
+                    {?status_connecting, <<"some_connections_are_unhealthy">>}
             end;
-        false ->
-            ?status_connecting
+        {error, timeout} ->
```
savonarola, line 269: Currently, in EMQX, MySQL driver is configured with infinity timeout.
So, if a tcp connection breaks ungracefully, any request (including health checks) just makes workers stuck.
The subsequent healthcheck requests also fail with timeout.
So in case of healthcheck timeout we report disconnected status to trigger reconnection.

rebar.config
```diff
@@ -89,7 +89,7 @@
     {gen_rpc, {git, "https://github.com/emqx/gen_rpc", {tag, "3.4.1"}}},
     {grpc, {git, "https://github.com/emqx/grpc-erl", {tag, "0.6.12"}}},
     {minirest, {git, "https://github.com/emqx/minirest", {tag, "1.4.4"}}},
-    {ecpool, {git, "https://github.com/emqx/ecpool", {tag, "0.6.1"}}},
+    {ecpool, {git, "https://github.com/emqx/ecpool", {tag, "0.6.1-hotfix"}}},
```
savonarola, line 92: https://github.com/emqx/ecpool/pull/59

apps/emqx_mysql/src/emqx_mysql.erl
thalesmg: Strange that the mysql driver does not attempt to reconnect or die if the tcp connection breaks :thinking:

savonarola: Its just a guess and an attempt to fix a problem of a concrete customer.

The thing is that the connection may break without any fin/rst packets, traffic just stops to arrive. A reconnect solves the problem, but the connector stucks in `connecting` state till the traffic unblocks.

PR #15920, opened 2025-09-14
[r60] upgrade mogodb driver to 3.1.0

zmstone:
This removes elrang-pbkdf2 (which is not used anyway) dependency from the release


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9492)

PR #15930, opened 2025-09-15
refactor: do not log each and every function call permission from remsh

zmstone:
Release version: 6.0.0

## Summary

Do not audit-log each and every `module:function` call in restricted shell.
Otherwise it may result in too many logs if for example a loop like below is evaluated in remote console:
```
lists:foreach(fun(_) -> module:function() end, lists:seq(1,000000)).
```

## PR Checklist


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9595)

PR #15953, opened 2025-09-17
build(fix): remove duplicate top-level dep

thalesmg:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9590)

PR #15965, opened 2025-09-18
feat(mq): move mq ds db configuration to the common durable_storage section

savonarola:
Fixes [EMQX-14733](https://emqx.atlassian.net/browse/EMQX-14733)

Release version: 6.0.0

## PR Checklist

~~- [ ] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)~~ Not, but the feature is not released

[EMQX-14733]: https://emqx.atlassian.net/browse/EMQX-14733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_ds_schema.erl
+++ b/apps/emqx/src/emqx_ds_schema.erl
@@ -7,10 +7,14 @@
 
 %% API:
 -export([schema/0]).
--export([db_config_messages/0, db_config_sessions/0, db_config_timers/0, db_config_shared_subs/0]).
-
-%% API for external apps using DS databases:
--export([db_config/1, db_schema/4]).
+-export([
+    db_config_messages/0,
+    db_config_sessions/0,
+    db_config_timers/0,
+    db_config_shared_subs/0,
+    db_config_mq_states/0,
+    db_config_mq_messages/0
+]).
 
 %% Behavior callbacks:
 -export([fields/1, desc/1, namespace/0]).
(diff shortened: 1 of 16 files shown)


apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -103,10 +113,24 @@ schema() ->
...
+                #{}
+            )},
+        {mq_messages,
+            db_schema(
+                [builtin_raft, builtin_local],
+                ?IMPORTANCE_MEDIUM,
+                ?DESC(mq_messages),
+                #{}
             )}
-    ] ++ emqx_schema_hooks:list_injection_point('durable_storage', []).
```
savonarola, line 107: We cannot actually inject anything meaningful here, without private schema and config helpers.

apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -103,10 +113,24 @@ schema() ->
...
+                ?DESC(mq_states),
+                #{}
+            )},
+        {mq_messages,
+            db_schema(
+                [builtin_raft, builtin_local],
+                ?IMPORTANCE_MEDIUM,
+                ?DESC(mq_messages),
+                #{}
             )}
```
zmstone, line 130: is it possible to configure builtin_raft for one and builtin_local for another ?

savonarola: Yes we may, the DBs are used independently

savonarola:
@zmstone the schema helper already generates raft backend as default. IMO if we are to hide local backend, we better do this uniformly for all databases

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9644)

PR #15967, opened 2025-09-19
fix(audit): Clean audit logs via dirty operation

HJianBo:
Before this fix, the audit process would start a huge transaction if the emqx_audit table stored a large number of items. It's hard to execute, which causes the emqx_audit to be cleared.

### How to reproduce this bug

Whenever a large number of calls are executed on the remote_console, for example:
```
spawn_link(fun() -> lists:foreach(fun(_) -> ets:info(emqx_audit, size) end, lists:seq(1, 200000)) end).
```
200K records are temporarily stored in the emqx_audit table. This type of operation is common when running a maintenance script on a production system.

It casued the emqx_audit can't be cleared
```
mnesia:info().
---> Processes holding locks <---
Lock: {{emqx_audit,1757942217435345},write,{tid,460,<0.4084.0>}}
Lock: {{emqx_audit,1757942220881245},write,{tid,460,<0.4084.0>}}
Lock: {{emqx_audit,1757942235399604},write,{tid,460,<0.4084.0>}}
...

299 transactions committed, 49 aborted, 6 restarted, 200293 logged to disc
19911 held locks, 0 in queue; 1 local transactions, 0 remote
0 transactions waits for other nodes: []
```

Release version: e5.8.9

## PR Checklist

- ~For internal contributor: there is a jira ticket to track this change~
- ~The changes are covered with new or existing tests~
- ~Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)~

### DIFF
--- a/changes/ee/fix-15967.en.md
+++ b/changes/ee/fix-15967.en.md
@@ -1,0 +1,1 @@
+Avoid a rapid increase in memory usage caused by Mnesia transaction blocking when cleaning a large number of audit logs.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9638)

PR #15977, opened 2025-09-19
fix(ns config): use cluster-wide config change when bulk importing configs

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14728

Fixes https://emqx.atlassian.net/browse/EMQX-14736

Release version: 6.0.0

## Summary

The wrong config update function was being used: it was altering only the local node...

Also, made an improvement to resource manager: when adding channels, we now add a trigger for channel health check, instead of waiting for connector health check to run (which might take a long time, depending on configuration) or manually kicking it off (like we do in `emqx_bridge_v2_api:wait_for_ready`).  This should improve feedback when importing configurations.

## PR Checklist


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9663)

PR #15986, opened 2025-09-22
fix(trace): avoid cleaning rotated log fragments

keynslug:
Part of [EMQX-14716](https://emqx.atlassian.net/browse/EMQX-14716).

Release version: 6.0.0

## Summary

See individual commits.

## PR Checklist


apps/emqx/test/emqx_trace_SUITE.erl
```diff
@@ -345,6 +345,63 @@ t_client_huge_payload_truncated(_Config) ->
     ?assertMatch({match, _}, re:run(Bin3, ReN(TruncatedBytes2), [unicode])),
     ok.
 
+%% Verify that cleanup routing does not touch log files of existing traces:
```
zmstone: routing? rutine?

keynslug: "Routine". 🫠 Thanks, fixed.

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -688,22 +690,33 @@ stop_trace(#{id := HandlerID, dst := Basename, filter := {Type, Filter}} = Handl
...
+                StaleFiles
+            );
+        _Error ->
             ok
     end.
 
+trace_file_is_stale("zip", _TraceFiles) ->
+    false;
+trace_file_is_stale(Filename, TraceFiles) ->
+    lists:all(fun(Basename) -> string:prefix(Filename, Basename) =:= nomatch end, TraceFiles).
```
zmstone: maybe add a is_active_trace/1 which parses the filename and lookup in the table ?

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -688,22 +690,33 @@ stop_trace(#{id := HandlerID, dst := Basename, filter := {Type, Filter}} = Handl
...
+                    file:delete(filename:join(TraceDir, Filename))
+                end,
+                StaleFiles
+            );
+        _Error ->
             ok
     end.
 
+trace_file_is_stale("zip", _TraceFiles) ->
+    false;
```
zmstone, line 726: maybe not the scope of this pr: zip dir is only cleaned up in the `terminate` callback, if the system is never gracefully shutdown. there could be files never gets deleted.

apps/emqx/src/emqx_trace/emqx_trace.erl
keynslug: That was actually the approach I tried first. It became quite hairy, mostly because of non-trivial regexp-based parsing and correct matching parsed dates against `started_at`. Of course O(N²) is not great, what do you think about [7b66600](https://github.com/emqx/emqx/pull/15986/commits/7b66600447d79ae37ddbb47344717df74c265368)?

apps/emqx/src/emqx_trace/emqx_trace.erl
keynslug: Good point. Looks like a [minor change](https://github.com/emqx/emqx/pull/15986/commits/da0fa13740bea9bc89d961e91bcd8498e03e1333), now included in this PR.

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -688,22 +690,49 @@ stop_trace(#{id := HandlerID, dst := Basename, filter := {Type, Filter}} = Handl
...
+clean_zip_dir() ->
+    file:del_dir_r(zip_dir()).
+
+trace_file_is_stale("zip", _TraceFileSet) ->
+    false;
+trace_file_is_stale(Filename, TraceFileSet) ->
+    %% NOTE: Basename should be the nearest predecessor of `Filename` if trace exists.
+    case gb_sets:next(gb_sets:iterator_from(Filename, TraceFileSet, reversed)) of
+        {Basename, _It} ->
+            string:prefix(Filename, Basename) =:= nomatch;
```
thalesmg, line 731: Tiny simplification (for OTP 27+):

```suggestion
    case gb_sets:smaller(Filename, TraceFileSet) of
        {found, Basename} ->
            string:prefix(Filename, Basename) =:= nomatch;
```

apps/emqx/src/emqx_trace/emqx_trace.erl
```diff
@@ -688,22 +690,49 @@ stop_trace(#{id := HandlerID, dst := Basename, filter := {Type, Filter}} = Handl
...
+            clean_stale_trace_files(TraceDir, Files);
+        _Error ->
             ok
     end.
 
+clean_stale_trace_files(TraceDir, Files) ->
+    %% NOTE: Cleaning both stale and foreign files.
+    TraceFileSet = lists:foldl(
+        fun(#?TRACE{name = Name, start_at = StartAt}, Acc) ->
+            %% NOTE: If `Name` is unique = filename is unique, `gb_sets:insert/2` is safe.
```
zmstone, line 707: `gb_sets:from_list` deduplicates.

keynslug: The same with `gb_sets:add/2`, but this is basically an attempt to avoid extra work.

apps/emqx/src/emqx_trace/emqx_trace.erl
keynslug: Turns out it's not the same, but I guess the comment was not entirely correct. We need "less than or equal" not "strictly less" provided by `gb_sets:smaller/2`.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9687)

PR #15988, opened 2025-09-22
feat: support specifying `zone_override` in authn results

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14627

Release version: 6.1.0

## PR Checklist


apps/emqx/src/emqx_connection.erl
```diff
@@ -1264,6 +1256,30 @@ wait_for_quic_stream_close(
...
+init_zone_specific_state(Zone, Opts, #state{} = State0) ->
+    #state{
+        transport = Transport,
+        socket = Socket
+    } = State0,
+    FrameOpts = #{
+        strict_mode => emqx_config:get_zone_conf(Zone, [mqtt, strict_mode]),
+        max_size => emqx_config:get_zone_conf(Zone, [mqtt, max_packet_size])
+    },
+    Parser = init_parser(Transport, Socket, FrameOpts),
```
zmstone, line 1270: max_packet_size has to match what's set to listener tcp options.

to keep it simple, we should not change parser (and document it)

changes/ee/feat-15988.en.md
```diff
@@ -0,0 +1,5 @@
+Added support for overriding zone using authentication results.  If a backend returns a `zone_override` attribute in its successful authentication result, it'll replace the connecting client's zone.
```
zmstone, line 1:
```suggestion
Added support for overriding zone using authentication results.  If a backend returns a `zone_override` attribute in its successful authentication result, it'll replace the connecting client's zone.

This can be useful to set per-zone rate limiters at runtime.
```

apps/emqx/src/emqx_connection.erl
zmstone: sorry, maybe just document that: when listener's `parse_unit` is set to `frame`, `max_packet_size` will not change after zone override.

thalesmg: :checkered_flag:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10585)

PR #16010, opened 2025-09-24
fix(fallback actions): provide metadata to republish function

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14757

Fixes https://github.com/emqx/emqx/discussions/16005

Release version: 5.9.2, 5.10.2, 6.0.1

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -2642,10 +2642,25 @@ trigger_fallback_action(Id, #{kind := reference, type := Type, name := Name}, Re
     end;
 trigger_fallback_action(Id, #{kind := republish, args := #{?COMPUTED := Args}}, Req, _QueryOpts) ->
     try
-        _ = emqx_rule_actions:republish(Req, Req, Args),
+        Env =
+            case Req of
+                #{metadata := #{rule_id := _}} ->
+                    Req;
+                #{} ->
+                    %% Rule metadata might be missing if the originating rule SQL is
+                    %% strict.  We inject the action id in its stead, since
+                    %% `emqx_rule_actions:republish` expects it.
+                    maps:update_with(
+                        metadata,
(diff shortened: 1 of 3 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -2642,10 +2642,25 @@ trigger_fallback_action(Id, #{kind := reference, type := Type, name := Name}, Re
     end;
 trigger_fallback_action(Id, #{kind := republish, args := #{?COMPUTED := Args}}, Req, _QueryOpts) ->
     try
-        _ = emqx_rule_actions:republish(Req, Req, Args),
+        Env =
+            case Req of
+                #{metadata := #{rule_id := _}} ->
+                    Req;
+                #{} ->
+                    %% Rule metadata might be missing if the originating rule SQL is
```
zmstone, line 2650: what does strict mean here ?

thalesmg: Does not output everything (`*`, or at least `metadata`).

thalesmg: Actually, explicitly outputting `metadata` does not work, because the key becomes a binary... :melting_face: :see_no_evil:

zmstone: would this maps:update_with override user's metadata ?

thalesmg: it's called only if the key is absent or without the `rule_id` sub-key.

thalesmg: also, if the user were to add a field called `<<"metadata">>`, it won't be an atom key.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9730)

PR #16016, opened 2025-09-25
ci: run perf test nightly on cron packages

id:
Trigger performance test every night at 01:00 UTC, use packages from the latest run of `build_packages_cron.yaml`


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10587)

PR #16021, opened 2025-09-25
fix(dsraft): use Cluster ID as part of Ra server UID

keynslug:
Fixes [EMQX-14756](https://emqx.atlassian.net/browse/EMQX-14756).

Release version: 6.0.0

## Summary

See individual commits.

## PR Checklist

    Layout of data directory slightly changed, consider part of DS-related breaking changes.


savonarola:
Nit: we probably cannot call ids predictable after the latest commits in the PR

keynslug:
> Nit: we probably cannot call ids predictable after the latest commits in the PR

Sort of, _predictable UID_ here means that it's generated solely from the DB shard and the current DS Cluster ID, the latter being opaque but stable.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9729)

PR #16031, opened 2025-09-26
feat(ds-otx): make CBM responsible for locating OTX leaders

keynslug:
Fixes <issue-or-jira-number>

Release version: 6.0.0

## Summary

TBA

## PR Checklist


ieQu1:
It looks like this is already merged. Also, this solution is overcomplicated, why does it pass the CBM to resolve the PID, instead of resolving name to PID at the backend level and passing the PID to the OTX?

PR #16063, opened 2025-09-30
fix(resource manager): return consistent reply type when channel health check aborts (r60)

thalesmg:
The expected result type is `#{status := resource_status(), error := term()}`.

Release version: 6.0.1, 6.1.0


qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9844)

PR #16064, opened 2025-09-30
fix(resource manager): return consistent reply type when channel health check aborts (r58)

thalesmg:
The expected result type is `#{status := resource_status(), error := term()}`.

Release version: 5.8.9, 5.9.2, 5.10.2


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9769)

PR #16068, opened 2025-09-30
fix(kafka & gcp pubsub consumers): add deprecated field

thalesmg:
This field was kept in the code as hidden field for backwards compatibility, but long deprecated in practice since the transition to connectors + sources.

Release version: 6.0.1, 6.1.0


thalesmg:
On second thought, the correct path to migrate is to create new kafka source and rule pairs, and the sources must not have the topic mapping field.

PR #16078, opened 2025-10-02
docs: hide durable queues config doc

zmstone:
Durable storage for shared-subscription was not documented and probably no one ever tried it out.
Given that this schema is deleted in v6, we hide it from newer releases in v5.


savonarola (dismissed):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9774)

PR #16080, opened 2025-10-02
feat(mq): add configuration switch for disabling MQ

savonarola:
Fixes [EMQX-14794](https://emqx.atlassian.net/browse/EMQX-14794)

Release version: 6.0.1

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

[EMQX-14794]: https://emqx.atlassian.net/browse/EMQX-14794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #16087, opened 2025-10-03
chore(mq): add changelog

savonarola:
Follow-up for #16080

Since the feature is released we should add change logs.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9863)

PR #16105, opened 2025-10-09
perf(ds): Optimize session connect

ieQu1:
Fixes <issue-or-jira-number>

Release version: 6.1.0

## Summary

Optimize connection latency

https://snapshots.raintank.io/dashboard/snapshot/601oITctp0Uu1RCh2VKtLdGT6HcURPRG

## PR Checklist

### DIFF
--- a/changes/ee/fix-16105.en.md
+++ b/changes/ee/fix-16105.en.md
@@ -1,1 +1,2 @@
-Improve the performance of the connection process of the durable sessions.
+Durable storage performance optimization.
+In particular, this fix reduces the latency of `CONNACK` for clients using durable session.


changes/ee/fix-16105.en.md
```diff
@@ -0,0 +1 @@
+Improve the performance of the connection process of the durable sessions.
```
zmstone: maybe make it more clear from end-user pov. what difference is it to be expected, would be even better how much different etc.

apps/emqx_durable_storage/src/emqx_ds_optimistic_tx.erl
```diff
@@ -556,18 +573,27 @@ do_cook_dirty_appends(DB, Shard, CBM, Gen, GenData, PresumedCommitSerial, Pendin
...
+        entered_pending_at = EP, flush_interval = FlushInterval, n_items = N, max_items = MaxItems
+    }
+) ->
+    D =
+        case EP of
+            _ when is_integer(EP) ->
+                D0;
+            undefined ->
+                %% Schedule unconditional flush after FlushInterval:
+                erlang:send_after(FlushInterval, self(), ?timeout_flush),
```
keynslug, line 588: Nit. Perhaps save on setting up an ERTS timer if `FlushInterval` is 0, and just send a message? As 0 is the default now.

keynslug (dismissed):

apps/emqx_durable_storage/src/emqx_ds_optimistic_tx.erl
ieQu1: That is regular flush timer, not idle flush which became 0 by default. 
I don't think removing the regular flush timer when idle flush timeout = 0 is entirely fair, otherwise in the situation where the message queue length is very long (and `after` clause is never reached) the server may never flush.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9954)

Issue #15605, opened 2025-07-29
MQTT Backend Has Over 1000 Retained Messages, But I Only Receive 1000


AK-DSB:
### What happened?

<img width="1976" height="1361" alt="Image" src="https://github.com/user-attachments/assets/937ec6b0-0a09-4bc1-b87e-723fe07101ab" />

<img width="2416" height="1529" alt="Image" src="https://github.com/user-attachments/assets/5d6e81c6-9484-44a1-ac5a-ee9a9ab0ee72" />

<img width="1799" height="927" alt="Image" src="https://github.com/user-attachments/assets/b46d307a-4b59-42f2-b532-3f9a7b43f905" />
The MQTT backend has over 1000 retained messages, but after subscribing to these topics, I only receive 1000 messages. Every time I run it, it's always 1000 messages. Why is this? Is there a way to get all retained messages every time I run it?

### What did you expect to happen?

To be able to receive all retained messages after each run.

### EMQX version

<details>

```console
[$ ./bin/emqx_ctl broker
# paste output here](image: emqx/emqx-enterprise:5.9.0)
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

---
*This issue was automatically translated from Chinese.*

id:
@AK-DSB can you confirm that there is indeed > 1000 topics matching the `winclient/+/isonline` pattern?

You could use [/topics](https://docs.emqx.com/en/enterprise/v5.10/admin/api-docs.html#tag/Topics/paths/~1topics/get) API with pagination.

AK-DSB:
> [@AK-DSB](https://github.com/AK-DSB) can you confirm that there is indeed > 1000 topics matching the `winclient/+/isonline` pattern?
> 
> You could use [/topics](https://docs.emqx.com/en/enterprise/v5.10/admin/api-docs.html#tag/Topics/paths/~1topics/get) API with pagination.

I'm sure, the last picture shows the retained message has 1503, only 3 message are system retain message, the others are winclient/+/isonline pattern, and I test the bug is about the image version, when I use image: emqx/emqx-enterprise:5.8.7, everything is ok, but  image: emqx/emqx-enterprise:5.9.0 or latest is not ok

id:
[...]

PR #16109, opened 2025-10-10
fix(retainer): continue delivering messages after dispatch rate limit

thalesmg:

## Summary

Previously, if at any point a retained message dispatch would hit the rate limit, the dispatching for that client would stop before all existing retained messages could be delivered (and likely several other clients that might've had their requests in the mailbox just after this one).

Now, instead of dispatching all retained messages through a single pool of workers, each client that subscribes to a topic filter containing retained messages will spawn an ephemeral worker that does the dispatching.  If the latter hits the shared rate limit, it sleeps for a while before trying again, until either the parent channel dies, or it manages to deliver everything retained.

In order to avoid inconsistencies when reindexing, we introduce a monotonically increasing incarnation number that is bumped when a full reindex occurs, and track which readers are using said incarnation number.  When a reindex occurs, the reindex process eventually calls `emqx_retainer_proto_v2:wait_dispatch_complete`, which will block until all ephemeral workers depending on previous incarnation numbers terminate.

## PR Checklist


apps/emqx/priv/bpapi.versions
```diff
@@ -82,7 +82,6 @@
 {emqx_prometheus,2}.
 {emqx_resource,1}.
 {emqx_resource,2}.
-{emqx_retainer,1}.
```
thalesmg, line 85: Apparently, even in [e5.0.0](https://github.com/emqx/emqx/blob/71f8e1035854a425900f9b10bbb2eb39af0e5fd8/apps/emqx_retainer/src/emqx_retainer_mnesia.erl#L630) we didn't use this version.

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -107,131 +168,35 @@ terminate(_Reason, #{pool := Pool, id := Id}) ->
...
-deliver(Messages, Pid, Topic, Limiter) ->
-    case erlang:is_process_alive(Pid) of
-        false ->
-            no_receiver;
-        _ ->
-            BatchSize = emqx_conf:get([retainer, flow_control, batch_deliver_number], 0),
-            NMessages = filter_delivery(Messages, Topic),
-            case BatchSize of
-                0 ->
-                    deliver_to_client(NMessages, Pid, Topic),
```
thalesmg, line 175: This was also fixed here: if `BatchSize` was zero, we didn't check the limiter to see if we could dispatch before sending the messages.

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -107,131 +168,35 @@ terminate(_Reason, #{pool := Pool, id := Id}) ->
...
-            dispatch_with_cursor(Context, Messages1, Cursor1, Pid, Topic, Limiter1);
-        {false, Limiter1} ->
-            ok = delete_cursor(Context, Cursor0),
-            {ok, Limiter1};
-        no_receiver ->
-            ?tp(debug, retainer_dispatcher_no_receiver, #{topic => Topic}),
-            ok = delete_cursor(Context, Cursor0),
-            {ok, Limiter0}
+notify_callers(Callers0, Incarnation) ->
+    case maps:take(Incarnation, Callers0) of
```
thalesmg: We should actually check if there even older incarnations before notifying them.

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -589,20 +602,29 @@ do_populate_index_meta(ConfigIndices) ->
...
+                    read -> {ReadIndices, Incarnation0};
+                    write -> {WriteIndices, Incarnation0}
+                end;
+            [] ->
+                {[], 0}
+        end,
+    case ReturnIncarnation of
+        true -> {Indices, Incarnation};
+        false -> Indices
     end.
```
keynslug, line 628: Nit. Having extra parameter affecting return type doesn't seem to make a lot of sense. Perhaps just always return `{Indices, Incarnation}` for simplicity?

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
```diff
@@ -0,0 +1,309 @@
...
+
+-define(DELAY_MS, 100).
+
+-type limiter() :: emqx_limiter_client:t().
+-type message() :: emqx_types:message().
+-type topic() :: emqx_types:topic().
+
+-type continuation() ::
+    {?at_once, [message()]}
+    | {?with_cursor, [message()], emqx_retainer:cursor()}.
```
keynslug, line 59: Tiny nit. In this case it'd probably be handier to just have records `#at_once{}` / `#with_cursor{}`.

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
```diff
@@ -0,0 +1,309 @@
...
+
+-type state() :: #{
```
keynslug, line 61: Q: Why not record? Would give tighter compile-time guarantees without those atoms-as-macros.

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
```diff
@@ -0,0 +1,309 @@
...
+-define(mref, mref).
+-define(parent, parent).
+-define(start_dispatch, start_dispatch).
+-define(stop, stop).
+-define(topic, topic).
+-define(undefined, undefined).
+-define(with_cursor, with_cursor).
+
+-define(DELAY(MSGS, CURSOR, LIMITER), {?delay, MSGS, CURSOR, LIMITER}).
+-define(DELAY_DELIVER(MSGS, LIMITER), {?delay, MSGS, undefined, LIMITER}).
```
keynslug, line 49: Tiny nit. This also looks essentially like a record declaration.

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
```diff
@@ -0,0 +1,309 @@
...
+
+-type delay() :: ?DELAY(_RestMessages, _Cursor, limiter()).
+
+%%------------------------------------------------------------------------------
+%% API
+%%------------------------------------------------------------------------------
+
+start(Topic, Clientid, Parent) ->
+    Opts = #{?topic => Topic, ?clientid => Clientid, ?parent => Parent},
+    gen_server:start(?MODULE, Opts, []).
```
keynslug, line 78: Since it starts out of `emqx_retainer` supervision tree, this process may potentially outlive `emqx_retainer` application (for example, during graceful node shutdown). Because it's essentially part of it and depends on `emqx_retainer` resources, it might be a good idea to put it there.

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
```diff
@@ -0,0 +1,309 @@
...
+            deliver_in_batches(RestMsgs, BatchSize, Pid, Topic, Limiter1);
+        {false, Limiter1, Reason} ->
+            ?tp(retained_dispatch_failed_for_rate_exceeded_limit, #{}),
+            ?SLOG_THROTTLE(
+                warning,
+                #{
+                    msg => retained_dispatch_failed_for_rate_exceeded_limit,
+                    reason => Reason,
+                    topic => Topic,
+                    dropped_count => BatchActualSize
```
keynslug, line 260: Q: Aren't messages preserved now instead of being dropped?

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
```diff
@@ -0,0 +1,309 @@
...
+    case emqx_limiter_client:try_consume(Limiter0, BatchActualSize) of
+        {true, Limiter1} ->
+            ok = deliver_to_client(Batch, Pid, Topic),
+            deliver_in_batches(RestMsgs, BatchSize, Pid, Topic, Limiter1);
+        {false, Limiter1, Reason} ->
+            ?tp(retained_dispatch_failed_for_rate_exceeded_limit, #{}),
+            ?SLOG_THROTTLE(
+                warning,
+                #{
+                    msg => retained_dispatch_failed_for_rate_exceeded_limit,
```
keynslug, line 257: Q: Perhaps should now be `"retained_msg_dispatch_throttled_due_to_quote_exceeded"`? E.g. similar to what `emqx_esockd_limiter` prints.

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
```diff
@@ -0,0 +1,309 @@
...
+-spec deliver([message()], pid(), topic(), limiter()) ->
+    {ok, limiter()} | ?stop | delay().
+deliver(Messages, Pid, Topic, Limiter) ->
+    case erlang:is_process_alive(Pid) of
+        false ->
+            %% Parent died
+            ?tp(debug, retainer_dispatcher_no_receiver, #{topic => Topic}),
+            ?stop;
+        _ ->
+            BatchSize = emqx_conf:get([retainer, flow_control, batch_deliver_number], 0),
```
zmstone, line 234: we have a default mailbox size based kill, this default batch size should be limited to prevent the process from getting killed.

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
```diff
@@ -0,0 +1,309 @@
...
+    ?topic := emqx_types:topic()
+}.
+
+-type delay() :: ?DELAY(_RestMessages, _Cursor, limiter()).
+
+%%------------------------------------------------------------------------------
+%% API
+%%------------------------------------------------------------------------------
+
+start(Topic, Clientid, Parent) ->
```
zmstone, line 76: this process should maybe also have a max age limit and count limit.

e.g.  by default stop dispatching after 1m, and stop dispatching after 10000 messages.

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
```diff
@@ -0,0 +1,309 @@
...
+                _ ->
+                    deliver_in_batches(NMessages, BatchSize, Pid, Topic, Limiter)
+            end
+    end.
+
+deliver_in_batches([], _BatchSize, _Pid, _Topic, Limiter) ->
+    {ok, Limiter};
+deliver_in_batches(Msgs, BatchSize, Pid, Topic, Limiter0) ->
+    {BatchActualSize, Batch, RestMsgs} = take(BatchSize, Msgs),
+    case emqx_limiter_client:try_consume(Limiter0, BatchActualSize) of
```
keynslug, line 248: General question: now, in the dispatcher-per-client, why do we have _shared_ dispatch rate limit? Documentation says it's about shielding clients from storms of retained messages, but in this case exclusive limiter would be better choice. On the other hand now that there's no single dispatcher anymore, Mnesia might get a lot more pressure. Perhaps we need to clarify documentation?

zmstone: It's a global limiter because we need to throttle the load of ets table traversal.
In the old times, many clients can start scanning the table concurrently, causes memory use to surge, and clients in the middle of such scan will likely timeout (iirc, it's a synced call when handling subscription, so they may never get back SUBACK).

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
zmstone: iiuc, it's called from the `clinet.subscribe` hook, so it's actually started by the session process.

we may link here, since emqx_connection never traps exit.

apps/emqx_retainer/src/emqx_retainer.erl
```diff
@@ -104,18 +104,22 @@
 -callback delete_cursor(backend_state(), cursor()) -> ok.
 -callback clean(backend_state()) -> ok.
 -callback size(backend_state()) -> non_neg_integer().
+-callback current_index_incarnation(backend_state()) -> integer().
```
savonarola, line 107: nit: make optional?

thalesmg:
Thanks for the reviews.

It sounds like it's better to instead use a different approach, probably what @savonarola suggested:

> Maybe we do not need any limiters at all in this case. Instead, we need a few workers (maybe much fewer than there are dispatchers currenty, 1, 2 or 4) that just receive dispatch requests from sessions, queue them and fulfil at workers' maximum configured pace.

... coupled with the incarnation counter so that we can reindex more safely.

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
thalesmg: I was avoiding to change all other call sites to drop the incarnation, since basically only one place is interested in it.

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
thalesmg: Map states are easier to inspect, if needed.

thalesmg:
I'll make another fix attempt in a separate branch/PR.

apps/emqx_retainer/src/emqx_retainer_ephemeral_dispatcher.erl
thalesmg: each zone has potentially different max mailbox sizes, so it's hard to pick which capped size to use

thalesmg:
https://github.com/emqx/emqx/pull/16125

PR #16112, opened 2025-10-11
feat(exhook): allow gRPC hooks to modify topic filters on subscribe/u…

### LINKED ISSUE #16102: Extend gRPC Hook Extension limited functionality

Allow gRPC methods to return the same values that the corresponding extension [callbacks](https://docs.emqx.com/en/emqx/latest/extensions/hooks.html#callback-function) are allowed to return.

### Why is this needed?

Currently, according to the [gRPC Hook Extension docs](https://docs.emqx.com/en/emqx/latest/extensions/exhook.html), only the `OnClientAuthenticate`, `OnClientCheckAcl` (seems like an error in the docs since `OnClientCheckAcl` has been replaced with `OnClientAuthorize` in a previous version) and `OnMessagePublish` are allowed to return data to EMQX. This seems needlessly limiting for a lot of possible usecases for the gRPC Hook Extension feature. 

In my case i would want to use the gRPC Hook Extension feature for extending the existing ACL model implemented by EMQX. Specifically i would like to tweak the set of topics matched by a wildcard subscription. For this the `OnClientSubscribe` and `OnClientUnsubscribe` methods seem  like they would do the trick. An extension written in Erlang can return new topic filters from the [callbacks](https://docs.emqx.com/en/emqx/latest/extensions/hooks.html#callback-function) `client.subscribe` and `client.unsubscribe`, but it is not possible to return new topic filters from the corresponding gRPC methods to achieve the same thing. This would force me to write an extension in Erlang, which would limit maintainability because that language is not widely used and increase implementation effort. A microservice implementing the gRPC interface written in e.g. go would not have those drawbacks, but instead increase the latency of publish and subscribe actions a bit.

MorganaFuture:
Fixes #16102

Release version: 7.0.0

Introduced in: N/A

## Summary

`client.subscribe` is dispatched with `emqx_exhook:cast/2`, which discards the server's
response, so a gRPC hook can observe a SUBSCRIBE but never act on it — unlike an Erlang
callback such as `emqx_rewrite:rewrite_subscribe/4`, whose `{ok, TopicFilters}` the channel
honours.

The new RPC is `OnClientSubscribeRewrite`, returning `ValuedResponse` with a `topic_filters`
value, registered under the hook name `client.subscribe.rewrite`. `OnClientSubscribe` keeps
its `EmptySuccess` return type, so a server generated from an earlier .proto keeps compiling
and keeps being called as before — that is the answer to the backward-compatibility question
in review: no change to the published `emqx.exhook.v3` contract.

On the reason-code comments: the hook runs after `check_sub_authzs`/`check_sub_caps` and its
result reaches the session directly, so `emqx_exhook_handler` re-establishes what the broker
otherwise guarantees:

* one filter back per requested filter, in the same order — SUBACK carries one reason code
  per requested topic filter, ordered as received [MQTT-3.9.3-1], and `do_gen_reason_codes/3`
  pairs the two lists positionally with no clause for a longer one;
* `subopts` are range-checked, since `qos` becomes the granted QoS;
* a rewritten filter is re-authorized and re-checked against the zone's subscribe
  capabilities, so a hook cannot grant what the client could not have requested itself;
* an unchanged filter keeps its original term instead of being re-parsed, so a shared
  subscription stays keyed by its `#share{}` record and still matches on unsubscribe.

A response breaking any of these is discarded whole and the requested filters are used
unchanged.

`emqx_channel.erl` is untouched — the change there in the previous revision was an identity
no-op. `client.unsubscribe` is dropped: #16102 does not ask for it, and UNSUBACK reason codes
come straight from `post_process_unsubscribe/4` with no re-derivation step, so a hook there
could change the reason code count in the packet itself.

The RPC deliberately has no digit suffix: `OnClientSubscribeV2` generates the client stub
`on_client_subscribe_v2` while `grpc_stream:init/2` resolves the server handler as
`on_client_subscribe_v_2`, so every call dies in `list_to_existing_atom/1`.

## PR Checklist

- [x] Change log added: `changes/ce/feat-16112.en.md`

### DIFF
--- a/apps/emqx_exhook/priv/protos/exhook.proto
+++ b/apps/emqx_exhook/priv/protos/exhook.proto
@@ -48,6 +48,15 @@ service HookProvider {
 
   rpc OnClientSubscribe(ClientSubscribeRequest) returns (EmptySuccess) {};
 
+  // Same event as `OnClientSubscribe`, but the server may rewrite the topic
+  // filters. Register it under the `client.subscribe.rewrite` hook name.
+  //
+  // Do not rename this to a digit suffix such as `OnClientSubscribeV2`: the
+  // generated client stub would be `on_client_subscribe_v2` while the server
+  // dispatch in `grpc_stream:init/2` resolves `on_client_subscribe_v_2`, and
+  // every call fails with `badarg` in `list_to_existing_atom/1`.
+  rpc OnClientSubscribeRewrite(ClientSubscribeRequest) returns (ValuedResponse) {};
+
   rpc OnClientUnsubscribe(ClientUnsubscribeRequest) returns (EmptySuccess) {};
 
   rpc OnSessionCreated(SessionCreatedRequest) returns (EmptySuccess) {};
@@ -314,7 +323,7 @@ message LoadedResponse {
 }
 
 // Responsed by `ClientAuthenticateRequest` `ClientAuthorizeRequest`
-// `MessageIngressRequest` `MessagePublishRequest`
+// `MessageIngressRequest` `MessagePublishRequest` `ClientSubscribeRequest`
 
 message ValuedResponse {
 
@@ -341,7 +350,26 @@ message ValuedResponse {
 
     // Message result, used on the 'message.*' hooks
     Message message = 4;
+
+    // Rewritten topic filters, used on the 'client.subscribe.rewrite' hook
+    TopicFilters topic_filters = 5;
   }
+}
+
+// The topic filters a `client.subscribe.rewrite` hook wants the client to be
+// subscribed to, in place of the ones it requested.
+//
+// `filters` MUST hold exactly as many entries as the request did, in the same
+// order: SUBACK carries one reason code per requested topic filter, ordered as
+// the filters were received (MQTT-3.9.3-1). A response of a different length is
+// rejected as a whole and the requested filters are used unchanged.
+//
+// A rewritten filter is re-authorized and re-checked against the subscribe
+// capabilities of the zone, so a hook cannot grant more than the client could
+// have asked for itself.
+message TopicFilters {
+
+  repeated TopicFilter filters = 1;
 }
 
 // no Response by other Requests
@@ -373,6 +401,7 @@ message HookSpec {
   //   "client.connected",    "client.disconnected"
   //   "client.authenticate", "client.authorize"
   //   "client.subscribe",    "client.unsubscribe"
+  //   "client.subscribe.rewrite"
   //
   //   "session.created",      "session.subscribed"
   //   "session.unsubscribed", "session.resumed"
--- a/apps/emqx_exhook/src/emqx_exhook_handler.erl
+++ b/apps/emqx_exhook/src/emqx_exhook_handler.erl
@@ -164,15 +164,14 @@ on_client_subscribe(ClientInfo, Props, TopicFilters) ->
         {StopOrOk, #{topic_filters := RespTfs}} ->
             case rewrite_topicfilters(ClientInfo, TopicFilters, RespTfs) of
                 {ok, TopicFilters} ->
-                    %% Nothing was rewritten. Leave the accumulator alone so the
-                    %% rest of the `client.subscribe' chain still runs, as it did
-                    %% while this hook was fire-and-forget -- but only when the
-                    %% server said to continue. Returning the filters unchanged is
-                    %% a reasonable way for a server to say "I have looked at this
-                    %% and no later hook should touch it", and `ignore' would
-                    %% silently downgrade that to "continue": `emqx_hooks' treats
-                    %% any term other than `stop' or `{stop, Acc}' as continue.
-                    %% `{stop, TopicFilters}' keeps both properties.
+                    %% Nothing was rewritten, so the accumulator is left alone
+                    %% and the rest of the `client.subscribe' chain runs -- but
+                    %% only when the server said to continue. Returning the
+                    %% filters unchanged is how a server says "I have looked at
+                    %% this and no later hook should touch it", and `ignore'
+                    %% would downgrade that to "continue", because `emqx_hooks'
+                    %% treats any term other than `stop' or `{stop, Acc}' as
+                    %% continue. `{stop, TopicFilters}' keeps both properties.
                     unchanged_topicfilters(StopOrOk, TopicFilters);
                 {ok, NTopicFilters} ->
                     {StopOrOk, NTopicFilters};
--- a/apps/emqx_exhook/src/emqx_exhook_handler.erl
+++ b/apps/emqx_exhook/src/emqx_exhook_handler.erl
@@ -164,10 +164,16 @@ on_client_subscribe(ClientInfo, Props, TopicFilters) ->
         {StopOrOk, #{topic_filters := RespTfs}} ->
             case rewrite_topicfilters(ClientInfo, TopicFilters, RespTfs) of
                 {ok, TopicFilters} ->
-                    %% Nothing was rewritten. Leave the accumulator alone so the rest
-                    %% of the `client.subscribe' chain still runs, as it did while
-                    %% this hook was fire-and-forget.
-                    ignore;
+                    %% Nothing was rewritten. Leave the accumulator alone so the
+                    %% rest of the `client.subscribe' chain still runs, as it did
+                    %% while this hook was fire-and-forget -- but only when the
+                    %% server said to continue. Returning the filters unchanged is
+                    %% a reasonable way for a server to say "I have looked at this
+                    %% and no later hook should touch it", and `ignore' would
+                    %% silently downgrade that to "continue": `emqx_hooks' treats
+                    %% any term other than `stop' or `{stop, Acc}' as continue.
+                    %% `{stop, TopicFilters}' keeps both properties.
+                    unchanged_topicfilters(StopOrOk, TopicFilters);
                 {ok, NTopicFilters} ->
                     {StopOrOk, NTopicFilters};
                 {error, _Reason} ->
@@ -176,6 +182,11 @@ on_client_subscribe(ClientInfo, Props, TopicFilters) ->
         ignore ->
             ignore
     end.
+
+unchanged_topicfilters(stop, TopicFilters) ->
+    {stop, TopicFilters};
+unchanged_topicfilters(_Ok, _TopicFilters) ->
+    ignore.
 
 on_client_unsubscribe(ClientInfo, Props, TopicFilters) ->
     {UserProps, SystemProps} = format_props(Props),
--- a/apps/emqx_exhook/src/emqx_exhook_handler.erl
+++ b/apps/emqx_exhook/src/emqx_exhook_handler.erl
@@ -45,6 +45,7 @@
     stringfy/1,
     merge_responsed_bool/2,
     merge_responsed_message/2,
+    merge_responsed_topicfilters/2,
     assign_to_message/2,
     clientinfo/1,
     request_meta/0
@@ -159,7 +160,22 @@ on_client_subscribe(ClientInfo, Props, TopicFilters) ->
         props => SystemProps,
         topic_filters => topicfilters(TopicFilters)
     },
-    emqx_exhook:cast('client.subscribe', Req).
+    case emqx_exhook:call_fold('client.subscribe', Req, fun merge_responsed_topicfilters/2) of
+        {StopOrOk, #{topic_filters := RespTfs}} ->
+            case rewrite_topicfilters(ClientInfo, TopicFilters, RespTfs) of
+                {ok, TopicFilters} ->
+                    %% Nothing was rewritten. Leave the accumulator alone so the rest
+                    %% of the `client.subscribe' chain still runs, as it did while
+                    %% this hook was fire-and-forget.
+                    ignore;
+                {ok, NTopicFilters} ->
+                    {StopOrOk, NTopicFilters};
+                {error, _Reason} ->
+                    ignore
+            end;
+        ignore ->
+            ignore
+    end.
 
 on_client_unsubscribe(ClientInfo, Props, TopicFilters) ->
     {UserProps, SystemProps} = format_props(Props),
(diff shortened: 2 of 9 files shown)


apps/emqx/src/emqx_channel.erl
```diff
@@ -887,8 +887,15 @@ process_subscribe(SubPkt = ?SUBSCRIBE_PACKET(PacketId, _Properties, _TopicFilter
     ),
     case Pipe of
         {ok, NPkt = ?SUBSCRIBE_PACKET(_PacketId, TFChecked), Channel} ->
+            %% Run hooks and extract topic filters from the returned packet
+            ?SUBSCRIBE_PACKET(_, _, ModifiedTopicFilters) = run_sub_hooks(NPkt, Channel),
+            %% Extract only the topic filters (without reason codes) for subscription
+            TopicFiltersForSub = [
+                TopicFilter
+             || {TopicFilter, ?RC_SUCCESS} <- ModifiedTopicFilters
+            ],
```
zmstone: According to MQTT spec:

> The Payload contains a list of Reason Codes. Each Reason Code corresponds to a Topic Filter in the SUBSCRIBE packet being acknowledged. The order of Reason Codes in the SUBACK packet MUST match the order of Topic Filters in the SUBSCRIBE packet [MQTT-3.9.3-1].

There are two things we should address here:
1. The hook should not alter the total number of topics (no more, no less than the originally requested topics).
2. This part should make sure the reason codes in  SUBACK is ordered as the received topics order.

apps/emqx_exhook/priv/protos/exhook.proto
```diff
@@ -46,9 +46,9 @@ service HookProvider {
 
   rpc OnClientAuthorize(ClientAuthorizeRequest) returns (ValuedResponse) {};
 
-  rpc OnClientSubscribe(ClientSubscribeRequest) returns (EmptySuccess) {};
+  rpc OnClientSubscribe(ClientSubscribeRequest) returns (ValuedResponse) {};
 
-  rpc OnClientUnsubscribe(ClientUnsubscribeRequest) returns (EmptySuccess) {};
+  rpc OnClientUnsubscribe(ClientUnsubscribeRequest) returns (ValuedResponse) {};
```
zmstone: why should we change unsubscribe API signature ?

apps/emqx_exhook/priv/protos/exhook.proto
```diff
@@ -46,9 +46,9 @@ service HookProvider {
 
   rpc OnClientAuthorize(ClientAuthorizeRequest) returns (ValuedResponse) {};
 
-  rpc OnClientSubscribe(ClientSubscribeRequest) returns (EmptySuccess) {};
+  rpc OnClientSubscribe(ClientSubscribeRequest) returns (ValuedResponse) {};
```
zmstone: is this a non-backward compatible change?
if yes, is it possible to keep it backward compatible ?

apps/emqx_exhook/priv/protos/exhook.proto
```diff
@@ -322,9 +322,17 @@ message ValuedResponse {
 
     // Message result, used on the 'message.*' hooks
     Message message = 4;
+
+    // TopicFilters result, used on the 'client.subscribe', 'client.unsubscribe' hooks
+    TopicFilters topic_filters = 5;
```
zmstone: due to the fact that MQTT broker must return the exact same number of reason codes as requested topic filters, we'll perhaps need to introduce an extra list, or an extra field for TopicFilters to indicate which topics are additional to subscribe (but never send a reason code back to the client for the extra).

There is maybe another kind of requirement: for some subscriptions, EMQX should reply success, but never actually subscribe. this can be useful if client is very much assertive about reason codes.

zmstone:
Thank you for the PR @MorganaFuture 

Just so you know, there are alternative solutions to manipulate MQTT topics.
- [Auto-subscribe](https://docs.emqx.com/en/emqx/latest/messaging/mqtt-auto-subscription.html#auto-subscribe)
- [Topic-rewrite](https://docs.emqx.com/en/emqx/latest/messaging/mqtt-topic-rewrite.html) -- can rewrite subscribe topics
- [Republish Action](https://docs.emqx.com/en/emqx/latest/data-integration/rule-get-started.html#add-republish-action)
- [Message trasnformation](https://docs.emqx.com/en/emqx/latest/data-integration/message-transformation.html#message-transformation) -- like topic rewrite, but only on the publishing side.

zmstone:
Hi @MorganaFuture 

Do you wish to continue working on this ?
My comments may not seem so constructive. This is because I do not have the answers to them myself.

MorganaFuture:
@zmstone hi!
Yea,  I will look at this tomorrow

zmstone:
> @zmstone hi! Yea, I will look at this tomorrow

Plase rebase to branch `dev-70`

apps/emqx/src/emqx_channel.erl
```diff
@@ -2742,14 +2749,33 @@ do_check_sub_caps(ClientInfo, [TopicFilter = {{_Topic, _SubOpts}, _OtherRC} | Mo
...
+%% Replaces all successful subscriptions with hook results, preserves failures
+merge_modified_topic_filters(TopicFilters0, _OriginalSuccessful, NTopicFilters) when
+    is_list(NTopicFilters)
+->
+    %% Hooks can completely change topics and options
+    %% We replace all successful subscriptions with the hook results
+    %% and append any non-SUCCESS results from the original list
+    Failures = [TF || TF = {_, RC} <- TopicFilters0, RC =/= ?RC_SUCCESS],
+    Successes = [{TF, ?RC_SUCCESS} || TF <- NTopicFilters],
+    Successes ++ Failures;
```
JimMoen: The order of TopicFilters and their RCs should not be changed.
See also [MQTT-3.9.3-1]

apps/emqx/src/emqx_channel.erl
MorganaFuture: gone, emqx_channel.erl is untouched now. the count and the order are enforced in
emqx_exhook_handler instead, hook response is dropped if it doesn't match.

apps/emqx_exhook/priv/protos/exhook.proto
MorganaFuture: yes it was. kept OnClientSubscribe on EmptySuccess and added
OnClientSubscribeRewrite instead, opt-in via the client.subscribe.rewrite hook
name. old servers unaffected.

apps/emqx_exhook/priv/protos/exhook.proto
MorganaFuture: dropped the unsubscribe half. post_process_unsubscribe/4 puts one reason code per
hook filter straight into UNSUBACK, nothing re-derives it there.

apps/emqx_exhook/priv/protos/exhook.proto
MorganaFuture: went strict 1:1 for now, no added topics at all in this one.

the separate field sounds right for those. emqx_auto_subscribe already does the
out-of-band half, posts {subscribe, ...} to the channel and it never reaches the
SUBACK. would rather do it on top once this lands, unless you want it here.

left the success-without-subscribe one alone, no idea either what it should do
about a later PUBLISH, or session.subscribed.

apps/emqx/src/emqx_channel.erl
MorganaFuture: gone, emqx_channel.erl is untouched now. the order comes from requiring the hook
to answer in the same positions.

MorganaFuture:
rebased onto dev-70 and reworked.

hook has to return one filter per requested filter in the same order now,
response dropped otherwise. compatibility done with a new RPC instead of
changing the existing one. dropped the unsubscribe half.

apps/emqx_exhook/priv/protos/exhook.proto
MorganaFuture: taking this one back rather than leaving it with you: this pr stays strict 1:1, no added topics at all. if the extra-topics field is still wanted i will open it separately on top, so nothing here waits on that decision.

MorganaFuture:
pushed a fix for something the move to call_fold introduced.

a server that does not advertise client.subscribe answers a bare `ignore`, and call_fold was folding that into the same branch as a real error, so failed_action decided what happened next. with the default `deny` the chain stopped, and every server behind the one that simply does not implement the hookpoint stopped receiving it, with nothing in the log. the client sees no difference either, since deny_action_result for client.subscribe round-trips the request.

that branch is shared by every hookpoint, so message.publish had the same hole. `ignore` now just moves on to the next server. t_call_fold_skips_servers_without_the_hookpoint covers it.

happy to split that into its own pr if you would rather keep this one to the subscribe rewrite.

apps/emqx_exhook/src/emqx_exhook_handler.erl
```diff
@@ -462,6 +478,131 @@ topicfilters(Tfs) when is_list(Tfs) ->
...
+rewrite_topic(_Topic, _SubOpts, Name) ->
+    throw({invalid_topic_filter, #{topic => Name}}).
+
+%% A hook that only rewrites the topic can leave `subopts' unset, in which case
+%% the client's own options stand. Whatever it does set is range-checked: these
+%% become the granted QoS in the SUBACK, and an out-of-range value is either an
+%% illegal reason code or, on MQTT 3.1.1, not serializable at all.
+rewrite_subopts(SubOpts, undefined) ->
+    SubOpts;
+rewrite_subopts(SubOpts, NSubOpts) when is_map(NSubOpts) ->
```
zmstone, line 567: A partially-populated `SubOpts` silently zeroes the fields the hook did not set, and on the wire that
case is indistinguishable from the hook deliberately sending zeros.

`SubOpts.qos` is `uint32 qos = 1` — a proto3 scalar, so it has no field presence. A server that wants
to change only `nl` builds a `SubOpts` with `nl = 1`; `qos`, `rh` and `rap` go on the wire as their
proto3 defaults and gpb decodes them as `0`. This clause then rebuilds the whole map from them, so a
client's QoS 2 subscription comes back as QoS 0 and the SUBACK grants 0. No error, no log — the
subscription just stops delivering at the level the client asked for.

The `subopt/4` `Min` default below cannot help: with proto3 scalars the key is always present in the
decoded map, so the default never applies.

The comment above this function says *"A hook that only rewrites the topic can leave `subopts' unset,
in which case the client's own options stand."* That is true, and it works, because `subopts` is a
message field and message fields do have presence — `maps:get(subopts, RespTf, undefined)` correctly
sees `undefined`. But it reads as though partial population is equally safe, and it is the opposite:
it is the one shape that loses data quietly.

Two ways out, and the choice is worth stating:

* Document it as a hard requirement — in `exhook.proto` next to `SubOpts subopts = 3`, not only here:
  if you set `subopts` at all, you must set every field, because unset means zero. A server author
  reads the .proto, not this module.
* Or give the fields presence (`optional uint32 qos = 1` in proto3), so an unset field really is
  absent and `subopt/4`'s default becomes meaningful. That changes the published
  `emqx.exhook.v3` contract, which the PR body is careful to avoid, so probably not — but it is the
  only version that makes the current code's intent true.

Either way this deserves a test: a hook that sets only `nl` on a QoS 2 subscription, asserting
whichever behaviour you decide is correct.

apps/emqx_exhook/src/emqx_exhook_handler.erl
```diff
@@ -159,7 +160,22 @@ on_client_subscribe(ClientInfo, Props, TopicFilters) ->
         props => SystemProps,
         topic_filters => topicfilters(TopicFilters)
     },
-    emqx_exhook:cast('client.subscribe', Req).
+    case emqx_exhook:call_fold('client.subscribe', Req, fun merge_responsed_topicfilters/2) of
+        {StopOrOk, #{topic_filters := RespTfs}} ->
+            case rewrite_topicfilters(ClientInfo, TopicFilters, RespTfs) of
+                {ok, TopicFilters} ->
+                    %% Nothing was rewritten. Leave the accumulator alone so the rest
```
zmstone: This discards the server's `STOP_AND_RETURN`.

`StopOrOk` is bound from the response type, and the branch below honours it — `{StopOrOk,
NTopicFilters}`. Here it is dropped and replaced with `ignore`, so a server that returns
`STOP_AND_RETURN` with the filters unchanged has its decision silently downgraded to "continue", and
the rest of the `client.subscribe` chain runs after all.

Returning the same filters is a reasonable way for a hook to say "I have looked at this and no later
hook should touch it". Under this clause that is exactly what does not happen.

The comment explains the goal — keep the chain running as it did while the hook was fire-and-forget —
and that goal is right for a server that answers `CONTINUE`. It is the `STOP` case that gets caught by
the same net. `{stop, TopicFilters}` would preserve both properties: the accumulator is unchanged and
the stop is honoured.

Worth checking what `emqx_hooks:run_fold/3` does with `{stop, Acc}` where `Acc` equals the input
before changing anything — if that path is what the comment was avoiding, say so here, because the
current code makes the two cases indistinguishable.

apps/emqx_exhook/src/emqx_exhook_server.erl
```diff
@@ -222,6 +217,22 @@ resolve_hookspec(HookSpecs) when is_list(HookSpecs) ->
         HookSpecs
     ).
 
+%% @private
+%% `client.subscribe.rewrite' is the `client.subscribe' hookpoint served by the
+%% `OnClientSubscribeRewrite' RPC, which -- unlike `OnClientSubscribe' -- may rewrite
+%% the topic filters. It is registered under a distinct name so that a server
+%% generated from an older .proto keeps answering the RPC it was built with.
+resolve_hook_name(<<"client.subscribe.rewrite">>) ->
```
zmstone, line 252: A server that registers both `client.subscribe` and `client.subscribe.rewrite` gets whichever appears
last in its `HookSpec` list, with nothing to say so.

Both names resolve to the same hookpoint atom `'client.subscribe'`, and `resolve_hookspec/1`
accumulates with `Acc#{Name => Opts}`, so the second entry overwrites the first. Order in the list
then decides whether the RPC called is `OnClientSubscribe` or `OnClientSubscribeRewrite` — and the
list comes from the server, so the outcome depends on how a third party built its `HookSpec`.

This is a plausible thing for a server to do while migrating: keep the old registration and add the
new one, expecting the broker to prefer the richer RPC. Today it might get either.

Suggest making it explicit — reject the combination with a clear error, or define a precedence
(`rewrite` wins) and note it in the .proto next to the `OnClientSubscribeRewrite` comment. Silent
last-one-wins is the outcome to avoid, since a server author has no way to observe which was chosen
short of watching which RPC arrives.

apps/emqx_exhook/src/emqx_exhook_handler.erl
MorganaFuture: Fixed. `{stop, TopicFilters}` rather than `ignore` when nothing was rewritten: `do_run_fold/3` returns the accumulator on `{stop, Acc}` and continues on anything else, so this keeps the accumulator untouched and honours the stop. CONTINUE still answers `ignore`, so the fire-and-forget behaviour is unchanged there. `t_subscribe_unchanged_filters_still_stop` pins both answers.

apps/emqx_exhook/src/emqx_exhook_server.erl
MorganaFuture: Went with precedence rather than rejecting the combination, since keeping the old registration while adding the new one is the natural migration step. `resolve_hookspec/1` now prefers the valued RPC in either list order and logs which one it took, so the choice is observable. `t_duplicate_subscribe_registration_prefers_rewrite` pins both orders plus the two single-registration cases.

apps/emqx_exhook/src/emqx_exhook_handler.erl
MorganaFuture: You are right, and I do not want to guess at the fix here.

The three `SubOpts` scalars are plain proto3, so gpb hands us `qos = 0` whether the server set it or left it out. There is no way to tell the two apart on the decode side, so anything I do inside `rewrite_subopts/2` is a guess about intent. Treating 0 as "unset" is wrong, since 0 is a legal value for all four.

Two options that actually carry presence, both wire-compatible in the sense that field numbers and encoding do not change:

1. Mark the four fields `optional` in `SubOpts`. Smallest diff, but `SubOpts` is shared with `session.subscribed` and `session.unsubscribed`, so the generated Erlang changes for those paths too and every server built against the current .proto sees the same message type change shape.
2. Give `OnClientSubscribeRewrite` its own message with `optional` scalars, leaving `SubOpts` untouched. Contained to the new RPC, at the cost of a second message that means almost the same thing.

I lean towards 2 because this RPC is new and nothing else has to move, but it is your call which one fits the .proto's direction. Whichever we pick, the unset fields should fall back to the client's own value rather than to zero, which is what the `Min` default in `subopt/4` was reaching for and cannot do today.

If you would rather not extend the proto at all, the honest alternative is to document that `subopts`, when present, replaces all four options and the hook owns them - and to reject a `SubOpts` whose fields would silently downgrade the client's request. That keeps the wire format but pushes the burden onto every hook author.

zmstone (dismissed):

MorganaFuture:
elvis `consistent_variable_casing` failed because `recheck_topicfilter/2` introduced `QoS` into a module that spells the variable `Qos` everywhere else. Renamed mine to match the file rather than touch the four existing uses. 71/71 locally.

zmstone:
please rebase after https://github.com/emqx/emqx/pull/18707 is merged

MorganaFuture:
Rebased onto dev-70 after #18707, no conflicts. `emqx_exhook_SUITE` 34/34 on the new base.

Also rewrote the comments this PR adds so they describe the current code rather than what it replaced.

Your approval was dismissed by my own lint-only push before the rebase, so this needs another look when you have a moment.

zmstone (approved):

hjianbo:
Thanks for the contribution! I have one design suggestion regarding `client.subscribe.rewrite`.

This introduces a special hook name for the same event, and future valued hooks would require more special names such as `client.connect.rewrite`. It also forces all `client.subscribe` calls through `call_fold`, changing the existing `cast` behavior even for legacy providers that still return `EmptySuccess`.

Could we negotiate the response type through `HookSpec` instead?

```protobuf
message HookSpec {
  string name = 1;
  repeated string topics = 2;
  bool valued_response = 3;
}
```

And add:

```protobuf
rpc OnClientSubscribeValued(ClientSubscribeRequest) returns (ValuedResponse);
rpc OnClientUnsubscribeValued(ClientUnsubscribeRequest) returns (ValuedResponse);
```

With `valued_response = false`, keep the existing RPC and `cast` path; with `true`, use the valued RPC and `call_fold`. This keeps hook names stable, avoids future special-name proliferation, and remains backward compatible.

PR #16123, opened 2025-10-14
fix(cluster): update `ekka` to 0.23.2

keynslug:
Release version: 6.0.1

## Summary

This update includes `mria` 0.8.18 that in turn fixes a bug in `mria_status` logic. This bug could have been the cause of stuck and/or incomplete cluster joins in core-replicant clusters.

## PR Checklist

- [ ] For internal contributor: there is a jira ticket to track this change
- [ ] The changes are covered with new or existing tests
    Tested manually using https://github.com/emqx/emqx-operator/pull/1155 facilities.
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible

PR #16124, opened 2025-10-14
chore: refine changelog entry

keynslug:
Followup to #16123.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-9946)

PR #16125, opened 2025-10-14
fix(retainer): continue delivering messages after dispatch rate limit

thalesmg:

New fix attempt after https://github.com/emqx/emqx/pull/16109
 

## PR Checklist


apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -162,43 +270,46 @@ delete_cursor(Context, Cursor) ->
...
+            ?tp(debug, retainer_dispatcher_no_receiver, #{topic => Topic}),
             no_receiver;
         _ ->
+            %% todo: cap at max process mailbox size to avoid killing; but each zone might
+            %% have different max mailbox sizes...
             BatchSize = emqx_conf:get([retainer, flow_control, batch_deliver_number], 0),
             NMessages = filter_delivery(Messages, Topic),
             case BatchSize of
                 0 ->
-                    deliver_to_client(NMessages, Pid, Topic),
```
thalesmg, line 175: This was also fixed here: if `BatchSize` was zero, we didn't check the limiter to see if we could dispatch before sending the messages.

apps/emqx/priv/bpapi.versions
```diff
@@ -82,7 +82,6 @@
 {emqx_prometheus,2}.
 {emqx_resource,1}.
 {emqx_resource,2}.
-{emqx_retainer,1}.
```
thalesmg, line 85: Apparently, even in [e5.0.0](https://github.com/emqx/emqx/blob/71f8e1035854a425900f9b10bbb2eb39af0e5fd8/apps/emqx_retainer/src/emqx_retainer_mnesia.erl#L630) we didn't use this version.

apps/emqx_retainer/src/emqx_retainer_sup.erl
```diff
@@ -81,12 +81,13 @@ init([?worker_sup]) ->
 %%--------------------------------------------------------------------
 
 start_dispatcher() ->
+    NumWorkers = max(1, min(4, emqx_vm:schedulers())),
```
ieQu1, line 84: Why 4?

thalesmg: Following the suggestion here: https://github.com/emqx/emqx/pull/16109#pullrequestreview-3334883940

ieQu1: This deserves a comment explaining the reason, otherwise it's a magic number.

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -107,26 +163,90 @@ terminate(_Reason, #{pool := Pool, id := Id}) ->
 cast(Msg) ->
     gen_server:cast(worker(), Msg).
 
--spec dispatch(pid(), topic(), limiter()) -> {ok, limiter()}.
-dispatch(Pid, Topic, Limiter) ->
+-spec handle_dispatch_request(pid(), topic(), state()) -> state().
+handle_dispatch_request(Pid, Topic, State0) ->
+    #{dispatch_requests := Reqs0} = State0,
+    Reqs = queue:in({Pid, Topic}, Reqs0),
```
savonarola, line 169: We probably need limit the queue? And also cleanup on `Pid` exit?

zmstone, line 169: the old impl was maybe a brutal cut-off, but it's more fair in a sense that: if there are many clients coming continuously, they still get the chance to receive some retained messages.
with this new impl, if the queue continues to grow, the tail of the queue will never get a chance to receive any messages in time.

so we need some kind of (preferably  configurable) fairness guarantee, e.g. like Ilia advised, limit the total size of the queue, force the old ones to drop out.

i'd personally prefer a roudrobin scheduling over a set of ongoing cursors (with limited number). a cursor can be closed when

1. it has reached end of table (completed)
2. it is too old (configurable age limit)
3. the owner session is disconnected/dead

thalesmg: It'll be dropped when the queue get to that pid, since we check if pid is still alive on deliver.

thalesmg: ... we could drop it a bit earlier, just before creating the cursors, as an optimization, true.

PR #16132, opened 2025-10-16
feat: add centralized certificate management api

thalesmg:
Part of https://emqx.atlassian.net/browse/EMQX-14804

Release version: 6.1.0

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_certs.erl
+++ b/apps/emqx_conf/src/emqx_conf_certs.erl
@@ -42,7 +42,7 @@
 -define(FILENAME_ACC_KEY, "acc-key.pem").
 -define(FILENAME_KEY_PASSWORD, "key-password").
 
--define(BPAPI, emqx_mgmt_api_certs).
+-define(BPAPI, emqx_conf_certs).
 
 -type maybe_namespace() :: emqx_config:maybe_namespace().
 -type file_kind() ::
@@ -109,7 +109,7 @@ list_bundles(Namespace) ->
     ok | {error, [#{node := node(), kind := file | rpc, reason := term()}]}.
 delete_bundle(Namespace, BundleName) ->
     Nodes = emqx_bpapi:nodes_supporting_bpapi_version(?BPAPI, 1),
-    Res = emqx_mgmt_api_certs_proto_v1:delete_bundle(Nodes, Namespace, BundleName),
+    Res = emqx_conf_certs_proto_v1:delete_bundle(Nodes, Namespace, BundleName),
     NodeRes = lists:zip(Nodes, Res),
     Errors = lists:filtermap(
         fun
@@ -135,7 +135,7 @@ delete_bundle(Namespace, BundleName) ->
     ok | {error, [#{node := node(), kind := file | rpc, reason := term()}]}.
 add_managed_files(Namespace, BundleName, Files) ->
     Nodes = emqx_bpapi:nodes_supporting_bpapi_version(?BPAPI, 1),
-    Res = emqx_mgmt_api_certs_proto_v1:add_managed_files(
+    Res = emqx_conf_certs_proto_v1:add_managed_files(
         Nodes, Namespace, BundleName, Files
     ),
     NodeRes = lists:zip(Nodes, Res),
--- a/apps/emqx/priv/bpapi.versions
+++ b/apps/emqx/priv/bpapi.versions
@@ -17,6 +17,7 @@
 {emqx_conf,3}.
 {emqx_conf,4}.
 {emqx_conf,5}.
+{emqx_conf_certs,1}.
 {emqx_connector,1}.
 {emqx_connector,2}.
 {emqx_dashboard,1}.
@@ -49,7 +50,6 @@
 {emqx_management,5}.
 {emqx_metrics,1}.
 {emqx_metrics,2}.
-{emqx_mgmt_api_certs,1}.
 {emqx_mgmt_api_clients,1}.
 {emqx_mgmt_api_plugins,1}.
 {emqx_mgmt_api_plugins,2}.
--- a/apps/emqx_conf/src/proto/emqx_conf_certs_proto_v1.erl
+++ b/apps/emqx_conf/src/proto/emqx_conf_certs_proto_v1.erl
@@ -1,9 +1,7 @@
 %%--------------------------------------------------------------------
 %% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
 %%-------------------------------------------------------------------
--module(emqx_mgmt_api_certs_proto_v1).
-
-%% FIXME: move to conf_certs_proto...
+-module(emqx_conf_certs_proto_v1).
 
 -behaviour(emqx_bpapi).
 


apps/emqx_management/src/emqx_mgmt_api_certs.erl
```diff
@@ -0,0 +1,337 @@
...
+        get => #{
+            tags => ?TAGS,
+            description => ?DESC("ns_file_list"),
+            parameters => [param_path_ns(), param_path_bundle_name()],
+            responses =>
+                #{
+                    200 =>
+                        emqx_dashboard_swagger:schema_with_examples(
+                            array(ref(file_out)),
+                            example_file_list()
```
thalesmg: fix type and example

apps/emqx_management/src/emqx_mgmt_api_certs.erl
```diff
@@ -0,0 +1,343 @@
...
+            parameters => [param_path_ns(), param_path_bundle_name()],
+            responses =>
+                #{
+                    204 => <<"">>,
+                    500 => internal_error(?DESC("internal_error"))
+                }
+        }
+    }.
+
+fields(file_in) ->
```
savonarola: Maybe provide multipart/form-data endpoint?

apps/emqx_conf/include/emqx_conf_certs.hrl
```diff
@@ -0,0 +1,14 @@
...
+%%--------------------------------------------------------------------
+-ifndef(EMQX_CONF_CERTS_HRL).
+-define(EMQX_CONF_CERTS_HRL, true).
+
+-define(FILE_KIND_KEY, key).
+-define(FILE_KIND_CHAIN, chain).
+-define(FILE_KIND_CA, ca).
+%% ACME account key
+-define(FILE_KIND_ACC_KEY, acc_key).
+-define(FILE_KIND_KEY_PASSWORD, key_password).
```
zmstone, line 12:
```suggestion
-define(FILE_KIND_KEY_PASSWORD, key_password).
%% ACME account key password is not supported so far.
```

apps/emqx_conf/src/emqx_conf_certs.erl
```diff
@@ -0,0 +1,266 @@
...
+filename_to_kind(?FILENAME_CA) ->
+    {ok, ?FILE_KIND_CA};
+filename_to_kind(?FILENAME_ACC_KEY) ->
+    {ok, ?FILE_KIND_ACC_KEY};
+filename_to_kind(?FILENAME_KEY_PASSWORD) ->
+    {ok, ?FILE_KIND_KEY_PASSWORD};
+filename_to_kind(_) ->
+    error.
+
+escape_name(Name) ->
```
zmstone, line 244: maybe this should reside in emqx_mt app e.g.
emqx_mt:escape_ns
to be able to reverse, there should be a emqx_mt:unescape_ns when e.g. to be able to find the namespace from dir name.

zmstone:
This a great base, however not very useful without supporting it from [SSL options config schema](https://emqx.atlassian.net/wiki/spaces/EMQX/pages/1671823504/Proposal+Centralized+TLS+Certificate+Management+for+EMQX#SSL-Configuration-Options)
 
Will you add it in a followup PR ?

apps/emqx_conf/src/emqx_conf_certs.erl
thalesmg: this is also applied to the bundle name, because it might have the same problems.

apps/emqx_management/src/emqx_mgmt_api_certs.erl
thalesmg: I followed the original proposal which is designed as a one-by-one upload.  multipart/form-data would be useful if we wanted to upload multiple files at once, right?

@zmstone WDYT?

thalesmg:
> This a great base, however not very useful without supporting it from [SSL options config schema](https://emqx.atlassian.net/wiki/spaces/EMQX/pages/1671823504/Proposal+Centralized+TLS+Certificate+Management+for+EMQX#SSL-Configuration-Options)
> 
> Will you add it in a followup PR ?

Yes, that's the idea.  Instead of having one big PR that is both harder to read and requires a lot of rework due to potential change requests, it felt better to first check the API is acceptable before building upon it.

apps/emqx_management/src/emqx_mgmt_api_certs.erl
zmstone: yes, good point.
one can upload single file using multipart, no problem.
e.g. if a cert is expire, one can just upload the cert file without having to re-upload key.

apps/emqx_conf/src/emqx_conf_certs.erl
zmstone: for bundle name, we can restrict it to a safe char-set.

apps/emqx_management/src/emqx_mgmt_api_certs.erl
thalesmg: > e.g. if a cert is expire, one can just upload the cert file without having to re-upload key.

the current design of uploading one-by-one already allows for this, without need for multipart.  :thinking:

apps/emqx_conf/src/emqx_conf_certs.erl
thalesmg: the problem with moving this to `emqx_mt` is that then there would be a call from `emqx_conf` to `emqx_mt`, the latter being one of the terminal apps, and the former one of the initial.

thalesmg: changed the way nss are escaped/quoted

apps/emqx_conf/src/emqx_conf_certs.erl
```diff
@@ -0,0 +1,226 @@
...
+    DataDir = emqx:data_dir(),
+    filename:join([DataDir, certs2, global]);
+base_dir(Namespace0) when is_binary(Namespace0) ->
+    DataDir = emqx:data_dir(),
+    Namespace = escape_name(Namespace0),
+    filename:join([DataDir, certs2, ns, Namespace]).
+
+dir(Namespace, BundleName0) ->
+    BaseDir = base_dir(Namespace),
+    BundleName = escape_name(BundleName0),
```
zmstone: we need to escape namespace due to compatibility (old version mqtt ns has no char-set restriction).
for bundle name, we can restrict char-set to make it easier.

apps/emqx_management/src/emqx_mgmt_api_certs.erl
```diff
@@ -0,0 +1,344 @@
...
+    }.
+
+%%-------------------------------------------------------------------------------------------------
+%% Internal fns
+%%-------------------------------------------------------------------------------------------------
+
+handle_list_bundles(Namespace) ->
+    case emqx_conf_certs:list_bundles(Namespace) of
+        {ok, Bundles} ->
+            Res = bundles_out(Bundles),
```
zmstone, line 317: escape namespace to list, and unescape before return?

apps/emqx_management/src/emqx_mgmt_api_certs.erl
zmstone: it's good to have multipart upload so one curl command or one request from dashboard can be used to upload multiple files.

I mean supporting multi-part does not mean one cannot upload single file anymore.

apps/emqx_conf/src/emqx_conf_certs.erl
thalesmg: charset is already restricted in last commits.

apps/emqx_management/src/emqx_mgmt_api_certs.erl
thalesmg: there's no need to escape/unescape namespace here. 

https://github.com/emqx/emqx/pull/16132/commits/aa1a8b81037d8540bb816c4c292a1e5179129226

zmstone (approved):

apps/emqx_management/src/emqx_mgmt_api_certs.erl
thalesmg: added multi-part support in the last two commits

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10602)

PR #16137, opened 2025-10-17
fix(retainer): continue delivering messages after hitting dispatch rate limit

thalesmg:

## Summary

Additionally to / as consequence of the fix:

 Ports the `client.handle_info` hook from 6.0 here.  Also, to make hook API more
extensible, changes the signature of the following hook callbacks to accept a more open
context map:

- `client.handle_info`
- `client.timeout`
- `message.delivered`
- `message.acked`

The context map, at this moment, has two keys which provide lazy access to channel and
session data: `chan_info_fn` and `session_info_fn`, which behave as `emqx_channel:info/2`
and `emqx_session:info/2` with the channel and session in the function's closure,
respectively.

This way, if we ever need to thread new data to one of these callbacks, we can simply add
it to the context map, and callbacks that are not interested in this new piece of
information do not need to be changed.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds.erl
@@ -697,7 +697,7 @@ handle_info(
             }
     end;
 handle_info(Msg, Session, _ClientInfo) ->
-    ?tp(warning, ?sessds_unknown_message, #{message => Msg}),
+    ?tp(debug, ?sessds_unknown_message, #{message => Msg}),
     Session.
 
 %%--------------------------------------------------------------------


apps/emqx/src/emqx_channel.erl
```diff
@@ -1642,6 +1649,27 @@ handle_info({disconnect, ReasonCode, ReasonName, Props}, Channel) ->
     handle_out(disconnect, {ReasonCode, ReasonName, Props}, Channel);
 handle_info({puback, PacketId, PubRes, RC}, Channel) ->
     do_finish_publish(PacketId, PubRes, RC, Channel);
+handle_info(
```
savonarola: 👍 Looks great!
But now _retaining_ leaked into the channel module. Could we just move all the retained message-related functions to the `retainer` app?
Seems that it would be enough for this to
* backport `channel.info` callback from 6.0
* additionally pass `fun(Prop) -> session:info(Prop, Session) end` lambda to this callback to figure out inflight info.

thalesmg: Thanks for the review!  :beers: 

I'll look into backport that callback.

> additionally pass fun(Prop) -> session:info(Prop, Session) end lambda to this callback to figure out inflight info.

Do you think this is necessary if we pass the whole session to the callback, and the called stuff is already using the session API (nothing internal)?

savonarola: For simplicity, yes, we may pass the session itself. My idea was to hide from the callback everything but lazy info.

apps/emqx/src/emqx_hookpoints.erl
```diff
@@ -41,7 +39,7 @@
     'client.subscribe',
     'client.unsubscribe',
     'client.timeout',
-    'client.monitored_process_down',
```
thalesmg, line 44: Checked with @savonarola , we have deprecated this callback and it's currently not used anywhere.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1761,7 +1776,8 @@ handle_timeout(
             handle_out(disconnect, ?RC_NOT_AUTHORIZED, Channel2)
     end;
 handle_timeout(TRef, Msg, Channel) ->
-    case emqx_hooks:run_fold('client.timeout', [TRef, Msg], []) of
+    HookContext = mk_common_hook_context(Channel),
+    case emqx_hooks:run_fold('client.timeout', [TRef, Msg, HookContext], []) of
```
savonarola: Nit: probably not a strictly necessary hook, because apps plugged into the channels by hooks may use `send_after` instead of `start_timer` to send messages to themselves.

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
```diff
@@ -625,7 +646,16 @@ reindex(NewIndices, Force, StatusFun) when
 
             %% Wait for all dispatch operations to be completed to avoid
             %% inconsistent results.
-            true = wait_dispatch_complete(?REINDEX_DISPATCH_WAIT),
+            DispatchesCompleted = wait_dispatch_complete(?REINDEX_DISPATCH_WAIT),
```
savonarola, line 649: Reindex is an administrative operation, so waiting for a little bit IMO is not a big deal. So could we simplify the things in the following way:
* Do not monitor clients in the dispatcher.
* Send incarnation together with cursor to the channels.
* Send it back to the dispatcher when querying for the next batches.
* On reindex, synchronously call the dispatchers passing the increased incarnation.
* The dispatchers set up timer and after ?REINDEX_DISPATCH_WAIT 
   * respond ok to the reindexer
   * start to just drop `next` requests with old incarnation.

In other words, just unconditionally give the maximum time for the clients to complete and then reject those who did not manage to complete.

WDYT?

apps/emqx/src/emqx_channel.erl
thalesmg: Right, but this was already here, I just added the new context.

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
thalesmg: It sounds it could work well, though it might complicate `emqx_retainer_SUITE:t_reindex` check for correctness.  :thinking: 

- If we set a low test-only `REINDEX_DISPATCH_WAIT` interval, the test might become flaky.
- If we keep the default 30 s, the test case will take a lot longer.

savonarola: Will a few seconds suffice?

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -235,3 +577,207 @@ take(N, [H | T], Count, Acc) ->
...
+
+do_take_delivers(0, RestDelivers, Acc) ->
+    {lists:reverse(Acc), RestDelivers};
+do_take_delivers(_N, [] = RestDelivers, Acc) ->
+    {lists:reverse(Acc), RestDelivers};
+do_take_delivers(N, [Deliver | RestDelivers], Acc) ->
+    do_take_delivers(N - 1, RestDelivers, [Deliver | Acc]).
+
+maybe_nudge_next_retained_batch(#message{} = Msg, deliver, _HookContext) ->
+    case pop_last_retained_batch_msgid(Msg) of
```
savonarola: We should take into account that durable sessions returns fake messages from puback, so `Msg`'s id won't be valid in the callbacks.

At the same time, seems that durable session is able to handle all the buffering and itself through `emqx_persistent_session_ds_inflight`, so we probably may just return the whole received buffers as delivers, and the ds session will pipe them to the client respecting the receive-maximum.

thalesmg: > We should take into account that durable sessions call deliver callbacks with a fake message, so Msg's id won't be valid in the callbacks.

Good catch.  I'll have to think what to do in this case...  Maybe check the last msgid in message acked? 
Though relying on that won't be quite enough if only QoS 0 messages are flying around.  :thinking: 

> At the same time, seems that durable session is able to handle all the buffering and itself through emqx_persistent_session_ds_inflight, so we probably may just return the whole received buffers as delivers, and the ds session will pipe them to the client respecting the receive-maximum.

This kinda breaks the session abstraction (changing the popping behavior based on the impl), though it could be the simplest solution.  
On the one hand, we currently we have only two and it's not quite extensible, so it might be fine.  On the other hand, this also makes the retainer depend on implementation details of the DS session, which will break retainer if it no longer accepts all the transient messages at some point.

thalesmg: Wait, `message.delivered` is not called by DS session nor with a fake message.  :thinking: 

Only `delivery.dropped.#` seems to be.

changes/ee/breaking-16137.en.md
```diff
@@ -0,0 +1,5 @@
+The signature of the `message.delivered`, `message.acked` and `client.timeout` callbacks has changed.  Plugins that rely on these callbacks will need to update their code to continue functioning properly.
```
zmstone: this make it non-suitable for 5.9 release.

we need to find a new way to ensure backward compatibility.
e.g. hook registration with a version number ?

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
savonarola: >This kinda breaks the session abstraction

yes 😓 yet this is more like a workaround over an already diverged behaviour 😅 

we may isolate this a bit e.g. in the following way:
* indroduce some special info key for `emqx_session`s like `emqx_session:info(available_inflight, Sess)` which will always return `infinity` for durable session impl and actual available room for mem session
* ask for the next batch not when the messages were acked, but when we managed to push everything into the inflight. In this case we will not need deliver/ack hooks at all.

changes/ee/breaking-16137.en.md
savonarola: In 5.9, we may e.g. stash additional context in the PD

```erlang
emqx_hooks:stash_context('message.delivered', Ctx) %% put(?HOOK_CTX_KEY('message.delivered'), Ctx),
emqx_hooks:run_fold('message.delivered', ...),
emqx_hooks:unstash_context('message.delivered'), %% erase(?HOOK_CTX_KEY('message.delivered')),
```

And inside the hook
```erlang
Ctx = emqx_hooks:context('message.delivered') %% get(?HOOK_CTX_KEY('message.delivered'))
```

A bit ugly but IMO acceptable for support releases

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
thalesmg: Still, we currently don't need this since `message.delivered` is not fed fake messages, right?

apps/emqx_retainer/src/emqx_retainer_mnesia.erl
thalesmg: :checkered_flag:

changes/ee/breaking-16137.en.md
thalesmg: :checkered_flag:

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
savonarola: `message.delivered` is functional, but `message.acked` is not for DS sessions
As far as I understand we need `message.delivered` to request the next batch when we pushed everything to the client.
But we need to know when and how many messages to push into the inflight, and we use `message.acked` for this for mem sessions. In DS we do not have this callback, so we need to push the whole batch and wait till it is delivered.
Do I understand correctly?

thalesmg: I added a similar check to `message.acked` because, if at some point the inflight becomes full when there are still retained messages to deliver, we need to start poking the retained batch again when there's room, which happens typically when the `message.acked` hook is triggered, if I understood correctly.  

The message output from `session_ds:puback` (for example) is indeed fake, but should not matter for poking the retained batch because it does not rely on the message id: https://github.com/thalesmg/emqx/blob/2bf2713200412285e7530ae3a0c22d8d2a1637c5/apps/emqx_retainer/src/emqx_retainer_dispatcher.erl#L724-L732

I'm thinking we might not need this mechanism of checking the last message id... We could just push retained messages if there are any and there is room.  :thinking:

thalesmg: I've just pushed a commit removing this last id tracking thing, and also testing the blocked delivery when using DS.  The test cases suggest it seems to be working without it, unless there's something else I'm missing.

savonarola: > but should not matter for poking the retained batch because it does not rely on the message id

Yes I was also thinking that the message ids themselves should not matter.

Should we handle the situation when there are only qos0 messages in the batch? In this case there will be no acks, neither fake or real.

thalesmg: > Should we handle the situation when there are only qos0 messages in the batch? In this case there will be no acks, neither fake or real.

I think this is already accounted for by nudging the process of pushing deliveries both after delivery and after ack.

If we have only QoS 0 messages in retained batch, and...

1) ... inflight is full, then that mean _some other messages with QoS > 0 are inflight_, and they will nudge the process at some point.
2) ... inflight is not full, then the QoS 0 deliveries should be pushed, and since they do not take up inflight space, they should eventually all be pushed when nudging in `message.delivery`.

savonarola (dismissed):

apps/emqx/src/emqx_channel.erl
```diff
@@ -3329,6 +3363,13 @@ disconnect_attrs(Reason, Channel) ->
     emqx_external_trace:disconnect_attrs(Reason, Channel).
 -endif.
 
+mk_common_hook_context(Channel) ->
+    Session = Channel#channel.session,
+    #{
+        chan_info_fn => fun(Prop) -> emqx_channel:info(Prop, Channel) end,
```
zmstone, line 3374: avoid anonymous func.
`{fun ?MODULE:info/2, [Channel]}`

apps/emqx/src/emqx_channel.erl
```diff
@@ -3329,6 +3363,13 @@ disconnect_attrs(Reason, Channel) ->
...
+        session_info_fn => fun(Prop) -> emqx_session:info(Prop, Session) end
```
zmstone, line 3375: `{fun emqx_session:info/2, [Session]}`

apps/emqx/src/emqx_channel.erl
```diff
@@ -3192,6 +3210,22 @@ run_hooks(Name, Args, Acc) ->
     ok = emqx_metrics:inc(Name),
     emqx_hooks:run_fold(Name, Args, Acc).
 
+-compile({inline, [run_hook_with_context/3]}).
+run_hook_with_context(Name, Args, Channel) ->
+    HookContext = mk_common_hook_context(Channel),
```
zmstone, line 3220: we will need some perf test to check if this will work well.
the session object for example is non-trivial term to copy into pd for each message.

zmstone: Maybe this:
For hook callbacks registered as `{M, F, Args}`, we try to call `M:callback_version(F)`, if there this remote call is `undef`, call with the original args.
Otherwise call with version indicated args.

savonarola: Does putting a term into PD cares about the term itself at all? AFAIK it just puts a eterm reference into the dict hash https://github.com/erlang/otp/blob/aec7d4e7971a65437e01b641a5ce97aff385573f/erts/emulator/beam/erl_process_dict.c#L628, without copying 🤔

apps/emqx/src/emqx_channel.erl
savonarola: Probably we may just pass session and channel then

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -7,17 +7,28 @@
...
 -export([
     start_link/2,
     dispatch/1,
+    next/1,
+    %% RPC target (v2)
     wait_dispatch_complete/1,
     worker/0
 ]).
 
+%% Hooks
```
zmstone, line 24: we need a uml-like call sequence diagram to describe how dispatcher and client process interact

apps/emqx/src/emqx_channel.erl
zmstone: for some reason i had the impression that pd put is a deep copy like ets. i must be wrong then.

apps/emqx/src/emqx_channel.erl
zmstone: actually make sense to use anonymous func, otherwise may lead to a large term for reporting.
e.g. in crash report, or if run process_info/0 from remote_console.

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -101,110 +274,227 @@ terminate(_Reason, #{pool := Pool, id := Id}) ->
...
     Mod = emqx_retainer:backend_module(Context),
-    State = emqx_retainer:backend_state(Context),
+    BackendState = emqx_retainer:backend_state(Context),
     case emqx_topic:wildcard(Topic) of
         true ->
-            {ok, Messages, Cursor} = Mod:match_messages(State, Topic, undefined),
-            dispatch_with_cursor(Context, Messages, Cursor, Pid, Topic, Limiter);
+            {ok, Messages, Cursor} = Mod:match_messages(BackendState, Topic, undefined),
+            Incarnation = cursor_index_incarnation(Cursor),
+            DispatchCursor = #dispatch_cursor{
```
zmstone, line 350: record a timestamp ? when a message is 5s newer than this timestamp, do not dispatch.

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -101,110 +274,227 @@ terminate(_Reason, #{pool := Pool, id := Id}) ->
     gproc_pool:disconnect_worker(Pool, {Pool, Id}).
 
 %%%===================================================================
-%%% Internal functions
+%%% Internal exports (only for mocking)
+%%%===================================================================
+
+retry_ttl() ->
+    %% todo: make configurable?
```
zmstone: maybe add a jira ticket ref here, otherwise this todo will be forgotten.

changes/ee/fix-16137.en.md
```diff
@@ -0,0 +1 @@
+Fixed an issue where not all retained messages would be delivered if a subscriber hit the retained message dispatch rate limit.
```
zmstone, line 1: please also mention the limits. the ttl and maximum number of pending?

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
thalesmg: this is first of all a question, if it _should be_ configurable. :sweat_smile:

zmstone: yes

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
thalesmg: documented the interaction in a comment.

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
thalesmg: :checkered_flag:

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
thalesmg: :checkered_flag:

changes/ee/fix-16137.en.md
thalesmg: :checkered_flag:

thalesmg:
Thanks for the reviews, @zmstone and @savonarola .  I've added a few commits to address the latest comments.

rel/i18n/emqx_retainer_schema.hocon
```diff
@@ -69,4 +69,10 @@ msg_expiry_interval_override.desc:
 allow_never_expire.desc:
 """If true, retained messages set to never expire (i.e., whose `Message-Expiry-Interval = 0`) are not affected by the expiry time override.  This configuration only takes effect when `msg_expiry_interval_override` is set."""
 
+dispatch_retry_ttl.label:
+"Dispatch retry TTL"
+
+dispatch_retry_ttl.desc:
+"""If the dispatch rate limit is reached while iterating over retained topics, then the dispatcher processes retry the dispatching at a later time.  This option controls how old an iteration may get before being dropped in the event of retries."""
```
zmstone: Maybe simplify it:

This option specifies the maximum amount of time EMQX will spend delivering retained messages to a single subscription. Once the time limit is reached, the dispatch process is stopped — even if additional retained messages match the subscription.

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
```diff
@@ -101,114 +322,241 @@ terminate(_Reason, #{pool := Pool, id := Id}) ->
...
+            ensure_delayed_retry(State, ?DELAY_MS)
     end.
 
-match_next(_Context, _Topic, undefined) ->
+enqueue_retry(Retry, ToRetry0) ->
+    ToRetry1 =
+        case queue:len(ToRetry0) >= ?MODULE:max_queue_size() of
+            true ->
+                %% log?
+                ?tp("retainer_retry_request_dropped_due_to_overflow", #{}),
```
zmstone, line 449: maybe send a message to the channel owner channel to log this event at warning level.

rel/i18n/emqx_retainer_schema.hocon
thalesmg: :checkered_flag:

apps/emqx_retainer/src/emqx_retainer_dispatcher.erl
thalesmg: :checkered_flag:

thalesmg:
Failing CI tests are unrelated to the changes herein: https://github.com/emqx/emqx/pull/16167

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10160)

PR #16152, opened 2025-10-22
feat(mq): implement message count/total byte size limits for queues

savonarola:
Fixes [EMQX-14799](https://emqx.atlassian.net/browse/EMQX-14799)

See https://emqx.atlassian.net/wiki/x/HwBZe.

For more details of how limiting works, also see `apps/emqx_mq/src/emqx_mq_message_quota_index.erl`.

Performance comparison for 1000 queues and 2000 inserts/sec 

image

Commands:
```
# for unlimited queues
./emqtt_bench pub -q 1 -t 'qul50/%rand_1000/%c' -i 5 -c 10000 -I 5000 -L 10000000 --max-random-wait 1000
# for limited queues
./emqtt_bench pub -q 1 -t 'ql50/%rand_1000/%c' -i 5 -c 10000 -I 5000 -L 10000000 --max-random-wait 1000
```

Release version: 6.0.1

## PR Checklist

[EMQX-14799]: https://emqx.atlassian.net/browse/EMQX-14799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_mq/src/emqx_mq_message_quota_buffer.erl
+++ b/apps/emqx_mq/src/emqx_mq_message_quota_buffer.erl
@@ -237,4 +237,4 @@ buffer_max_size() ->
     emqx:get_config([mq, quota, buffer_max_size]).
 
 buffer_flush_interval() ->
-    emqx:get_config([mq, quota, buffer_flush_interval]).
+    round(emqx:get_config([mq, quota, buffer_flush_interval]) * (0.9 + 0.2 * rand:uniform())).
--- a/apps/emqx_mq/test/emqx_mq_gc_SUITE.erl
+++ b/apps/emqx_mq/test/emqx_mq_gc_SUITE.erl
@@ -121,7 +121,7 @@ t_limited_regular(_Config) ->
     emqx_mq_test_utils:populate(200, #{
         topic_prefix => <<"tc/">>, payload_prefix => <<"payload-">>, different_clients => true
     }),
-    ok = emqx_mq_message_quota_buffer:flush(),
+    ct:sleep(1100),
     ?assertWaitEvent(emqx_mq_gc:gc(), #{?snk_kind := mq_gc_done}, 1000),
 
     %% Check that only the last 100 + threshold messages are available
@@ -184,7 +184,7 @@ t_limited_lastvalue(_Config) ->
     emqx_mq_test_utils:populate_lastvalue(80, #{topic_prefix => <<"tc/1/">>}),
     %% Publish 3rd portion of 80 messages to the queue
     emqx_mq_test_utils:populate_lastvalue(80, #{topic_prefix => <<"tc/3/">>}),
-    ok = emqx_mq_message_quota_buffer:flush(),
+    ct:sleep(1100),
 
     %% Run GC
     ?assertWaitEvent(emqx_mq_gc:gc(), #{?snk_kind := mq_gc_done}, 1000),


rel/i18n/emqx_mq_schema.hocon
```diff
@@ -113,6 +113,26 @@ emqx_mq_schema {
...
+    max_shard_message_count {
+        desc: """~
+            The maximum number of messages in a shard for the Message Queue. The number of shards is configured
+            for all Message Queues in the Durable Storage settings."""
+        label: "Max Shard Message Count"
+    }
+
+    max_shard_message_bytes {
+        desc: """~
+            The maximum number of bytes in a shard for the Message Queue. The number of shards is configured
```
zmstone, line 132: should mention that the replication factor is not relevant to this limit? i.e. if a queue has 3 replicas, it does not mean the limit must be x3

apps/emqx_mq/src/emqx_mq_schema.erl
```diff
@@ -65,6 +65,44 @@ fields(mq) ->
                     <<"regular">> => false,
                     <<"lastvalue">> => #{}
                 }
+            })},
+        {quota,
+            mk(ref(quota), #{
+                required => true,
+                desc => ?DESC(quota),
+                default => #{},
+                importance => ?IMPORTANCE_HIDDEN
```
zmstone, line 74: add comments to explain why they are hidden.

changes/ee/feat-16152.en.md
```diff
@@ -0,0 +1 @@
+Introduced individual limits for message queues, allowing configuration for max message count and size.
```
zmstone: maybe also mention the newly added metrics

apps/emqx_mq/src/emqx_mq_message_quota_index.erl
```diff
@@ -0,0 +1,546 @@
...
+Deletion of a message is the reverse operation of insertion,
+i.e. decreasing the value of the corresponding bucket by the message size and by 1 for the message count.
+
+Because of the insertion logic, we always know exactly from which bucket to delete the message.
+
+#### Compaction
+
+After each operation, we "compact" the index.
+* We squash consecutive buckets that are in common smaller than the threshold.
+* We remove tail buckets that are responsible for data that exceeds the limits.
```
zmstone: tail is a bit ambiguous here, maybe the 'oldest' ?

apps/emqx_mq/src/emqx_mq_message_quota_buffer.erl
```diff
@@ -0,0 +1,240 @@
...
+    {message_queue_len, InboxSize} = process_info(self(), message_queue_len),
+    ok = emqx_mq_metrics:set_quota_buffer_inbox_size(WorkerId, InboxSize).
+
+pool_size() ->
+    emqx:get_config([mq, quota, buffer_pool_size]).
+
+buffer_max_size() ->
+    emqx:get_config([mq, quota, buffer_max_size]).
+
+buffer_flush_interval() ->
```
zmstone, line 239: maybe add some random delay here to get away from unfortunate situation that nodes accidentally always flush around the same moment (i.e. conflict).

apps/emqx_mq/src/emqx_mq_registry.erl
```diff
@@ -205,7 +209,7 @@ delete_all() ->
 
 -doc """
 Update the MQ by its topic filter.
-`is_lastvalue` is not allowed to be updated.
+`is_lastvalue` and is not allowed to be updated.
```
savonarola: finish the rephrasing

apps/emqx_mq/src/emqx_mq_registry.erl
```diff
@@ -105,11 +108,12 @@ create(#{topic_filter := TopicFilter, is_lastvalue := IsLastValue} = MQ0) when
...
                 {error, Reason}
         catch
-            Class:Reason ->
+            Class:Reason:StackTrace ->
                 ?tp(error, mq_registry_create_mq_state_error, #{
                     mq => MQ,
                     class => Class,
-                    reason => Reason
+                    reason => Reason,
+                    stacktrace => StackTrace
```
savonarola: remove Stacktrace

apps/emqx_mq/src/emqx_mq_message_quota_index.erl
```diff
@@ -0,0 +1,546 @@
...
+
+-type t() :: #index{}.
+-type timestamp_us() :: non_neg_integer().
+-type update_type() :: add | delete.
+-type update_values() :: #{bytes => non_neg_integer(), count => non_neg_integer()}.
+-type update() :: #{
+    type := update_type(),
+    timestamp_us := timestamp_us(),
+    values := update_values()
+}.
```
keynslug: Nit. Since buffers will hold potentially huge amount of these records, perhaps makes sense to minimize memory overhead per entry? E.g. `update() :: {timestamp_us(), integer(), integer()}` where negative integer means deletion.

apps/emqx_mq/src/emqx_mq_message_quota_index.erl
```diff
@@ -0,0 +1,546 @@
...
+
+#### Insertion
+
+The "insertion" of a message in the index is accounting for message size in the index, i.e.
+increasing the value of the corresponding bucket by the message size and by 1 for the message count.
+
+* If the message's timestamp is greater than the start of the leading bucket, we insert it into the leading bucket.
+  If the leading bucket is full, we create a new bucket at the head of the list.
+  We take the maximum seen timestamp value + 1 as the timestamp for the new bucket start.
+  We do this to make sure that none of the existing insertions should have been done in the new bucket
```
keynslug, line 39: Q: Does this essentially mean "concurrent insertions could end up in the new bucket"? Or (more likely) I miss / misunderstand something? Somewhat unclear what's the exact effect of using `MaxTs + 1`, AFAICS currently broker's millisecond timestamp is used to construct `update()`s, which likely means `MaxTs + 1` would be the same for messages arriving in the same millisecond.

savonarola: The logic behind MaxTs is the following.

Assume we have
* nodes A, B, and maybe more
* a single bucket in index starting from ts = 0
* time on node B is 1s ahead from A.

**Without `MaxTs`**
* B inserts some messages with ts=2000 into the bucket;
* Then A inserts some messages with ts=1000 (its time is lagging) into the bucket and sees that threshold is reached;
* A starts a new bucket starting from its local time 1000, we have two buckets now `0 - 1000`, `1001 -`
* When it comes time to delete B's messages from the index we cannot do it correctly because they are in the first bucket whose bounds do not correspond to the actual messages' timestamps.

**With `MaxTs` but without + 1**
* B inserts some messages with ts=2000 into the bucket;
* A inserts some messages with ts=1000 into the bucket and sees that threshold is reached;
* A starts a new bucket starting from MaxTs, (i.e. from 2000 because B updated it), we have two buckets now `0 - 1999`, `2000 -`
* B inserts some more messages with ts=2000, they get into the new bucket.
* When it comes time to delete B's messages with ts=2000 we are not sure from which bucket to delete them — either from first or the second.

**With `MaxTs` and +1**
* B inserts some messages with ts=2000 into the bucket;
* A inserts some messages with ts=1000 into the bucket and sees that threshold is reached;
* A starts a new bucket starting from MaxTs, (i.e. from 2001), we have two buckets now `0 - 2000`, `2001 -`
* A,B or other nodes insert some more messages with ts<=2000, they get into the old (first) bucket.
* A,B or other nodes insert some messages with ts>=2001, they get into the new bucket.
* When it comes time to delete B's messages with ts 2000 we are sure they are from the first bucket.

**In other words, using `MaxTs+1` for the lower bound of the new bucket prevents the situation when we have messages with the same timestamp in different buckets.**

The downside is the following:
* B inserts some messages with ts=2000 into the bucket and sees that threshold is reached;
* B starts a new bucket starting from MaxTs, (i.e. from 2001), we have two buckets now `0 - 2000`, `2001 -`
* B still inserts some more messages with ts=2000 into the bucket, they get into the old (first) bucket.
* Moreover, A for a second (because it's time is lagging) still inserts the messages into the first `0 - 2000` bucket.
This leads to the situations when the size of the first bucket exceeds the threshold and when we have more excess data after the GC. We ignore this so far, because for hot queues the impact of delayed GC is significantly higher anyway.

apps/emqx_mq/src/emqx_mq_message_quota_index.erl
savonarola: ✅

apps/emqx_mq/src/emqx_mq_message_quota_buffer.erl
savonarola: ✅

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10156)

PR #16161, opened 2025-10-27
fix(resource): Handle already started and present errors in ensure_child

zhongwencool:
Added explicit handling for {already_started, _} and already_present errors in ensure_child/5 to prevent unnecessary error propagation when a child process is already running or present.

Having multiple processes create the same resource concurrently is fine if at least one succeeds. This should not cause a crash.

```
2025-10-27T02:05:37.744917+00:00 [error] crasher: initial call: my_plugin_receiver_parser:init/1, pid: <0.4940.0>, registered_name: [], exit: {{handshake_callback_failed,{already_present,[{emqx_resource_manager_sup,ensure_child,5,[{file,"emqx_resource_manager_sup.erl"},{line,21}]},{emqx_resource_manager,create,5,[{file,"emqx_resource_manager.erl"},{line,269}]},{emqx_resource_manager,create_and_return_data,5,[{file,"emqx_resource_manager.erl"},{line,262}]},{my_plugin_resource,create,5,[{file,"xyz_resource.erl"},{line,27}]},
```

Fixes <issue-or-jira-number>

5.10.2

Release version:

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_manager_sup.erl
+++ b/apps/emqx_resource/src/emqx_resource_manager_sup.erl
@@ -15,6 +15,10 @@
 
 ensure_child(ResId, Group, ResourceType, Config, Opts) ->
     case supervisor:start_child(?MODULE, child_spec(ResId, Group, ResourceType, Config, Opts)) of
+        {error, {already_started, _}} ->
+            ok;
+        {error, already_present} ->
+            ok;
         {error, Reason} ->
             %% This should not happen in production but it can be a huge time sink in
             %% development environments if the error is just silently ignored.


apps/emqx_resource/src/emqx_resource_manager_sup.erl
```diff
@@ -15,6 +15,10 @@
 
 ensure_child(ResId, Group, ResourceType, Config, Opts) ->
     case supervisor:start_child(?MODULE, child_spec(ResId, Group, ResourceType, Config, Opts)) of
+        {error, {already_started, _}} ->
+            ok;
+        {error, already_present} ->
+            ok;
```
thalesmg, line 21: I think those cases should not be silently ignored:  the resource id could exist with a different configuration than the one being passed here, hence it would lead to a silent "corruption" of the expected config.

If the plugin is starting the same resource id concurrently, it would be better to handle this error there than ignoring it for all resources, I believe.

PR #16178, opened 2025-10-28
feat(streams): put up basic application skeleton

keynslug:
Part of [EMQX-14857](https://emqx.atlassian.net/browse/EMQX-14857).

## Summary

TBA

## PR Checklist

[EMQX-14857]: https://emqx.atlassian.net/browse/EMQX-14857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rel/i18n/emqx_streams_schema.hocon
+++ b/rel/i18n/emqx_streams_schema.hocon
@@ -1,0 +1,6 @@
+emqx_streams_schema {
+
+enable.desc:
+"""Enable Message Streams feature."""
+
+}


apps/emqx_streams/src/emqx_streams_schema.erl
```diff
@@ -0,0 +1,41 @@
...
+
+%% `hocon_schema' API
+-export([namespace/0, roots/0, fields/1, desc/1, tags/0]).
+
+%%------------------------------------------------------------------------------
+%% `hocon_schema' APIs
+%%------------------------------------------------------------------------------
+
+namespace() ->
+    durable_streams.
```
savonarola: Nit: may be "durable" prefix is redundant? Otherwise we take some kind of an obligation to consistently call "durable" everything that uses durable storage (even if there is no non-durable impl).

keynslug: Good point. I have a similar gripe, but on the other hand _streams_ sounds too general and maybe not too future-proof. Anyway, let's go with a simpler naming.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10611)

PR #16195, opened 2025-10-30
ci(docker): don't push when not publishing

thalesmg:
Trying to fix this: https://github.com/emqx/emqx/actions/runs/18907693639/job/53973612772?pr=16181#step:10:657

Simply setting `DOCKER_PUSH=false` and `DOCKER_LOAD=true` doesn't work currently, [because](https://github.com/emqx/emqx/actions/runs/18949722926/job/54111948131?pr=16195#step:10:59):

```
ERROR: failed to build: docker exporter does not currently support exporting manifest lists
```


thalesmg:
https://github.com/emqx/emqx/pull/16197

PR #16197, opened 2025-10-30
ci: fix docker workflow

id:
Now "build" jobs are using matrix to produce separate images for amd64 and arm64, and "publish" job publishes multi arch image to docker hub via local registry.

Also update UI tests to use official selenium docker image, since google does not publish official arm64 packages for Chrome.

Successful "Build and push docker images" workflow run: https://github.com/emqx/emqx/actions/runs/18957496878
image

### DIFF
--- a/.github/workflows/build_and_push_docker_images.yaml
+++ b/.github/workflows/build_and_push_docker_images.yaml
@@ -54,45 +54,8 @@ permissions:
   contents: read
 
 jobs:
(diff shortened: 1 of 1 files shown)


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10258)

PR #16190, opened 2025-10-29
test(dsraft): attempt to stabilize flaky test

thalesmg:
https://github.com/emqx/emqx/actions/runs/18918819109/job/54011970231?pr=16189

```
%%% emqx_ds_builtin_raft_SUITE ==> t_rebalance: OK
Testing apps.emqx_ds_builtin_raft.emqx_ds_builtin_raft_SUITE: Stopping test case repeat operation: {flaky,3}
Killed
make: *** [Makefile:143: apps/emqx_ds_builtin_raft-ct] Error 137
```

PR #16203, opened 2025-10-31
test(dsraft): another attempt to stabilize flaky test

thalesmg:
Sequence to https://github.com/emqx/emqx/pull/16190

I vaguely remember seeing memory issues with snabbkaffe tracepoints in hot loops.  The
last ran test case just before the VM is killed is `t_rebalance`:

https://github.com/emqx/emqx/actions/runs/18918819109/job/54011970231?pr=16189
```
%%% emqx_ds_builtin_raft_SUITE ==> t_rebalance: OK
Testing apps.emqx_ds_builtin_raft.emqx_ds_builtin_raft_SUITE: Stopping test case repeat operation: {flaky,3}
Killed
make: *** [Makefile:143: apps/emqx_ds_builtin_raft-ct] Error 137
```

... Which suggests the problematic case is `t_rebalance_chaotic_converges`, which is
indeed wrapped in a check trace.

See also the individual commits.

### DIFF
--- a/apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
+++ b/apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
@@ -1092,7 +1092,7 @@ sort_canonical_forms(Msgs) ->
 
 %%
 
-suite() -> [{timetrap, {seconds, 120}}].
+suite() -> [{timetrap, {seconds, 120}}, {ct_hooks, [emqx_cth_ct_hook_flaky]}].
 
 all() ->
     Broken = [


thalesmg:
After current fixes, 16 consecutive successful runs: https://github.com/emqx/emqx/actions/runs/19036419365/job/54380816834?pr=16203

apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
```diff
@@ -250,121 +250,119 @@ t_rebalance(Config) ->
...
-                        n_shards_online(Node, ?DB),
-                        "Each node is now responsible for 3/4 of the shards"
-                    )
-                )
-             || Node <- Nodes
-            ],
+    %% 2. Start filling the storage:
+    emqx_ds_raft_test_helpers:apply_stream(?DB, Nodes, Stream),
+    timer:sleep(5000),
+    emqx_ds_raft_test_helpers:verify_stream_effects(
```
keynslug, line 315: The trouble is `?defer_assert` that is used extensively throughout the test case (in `verify_stream_effects/4` also) expects that we're inside snabbkaffe two-phase test run.

keynslug:
Interesting find! I'm now curious to see if this can be done without having to lose snabbkaffe tracing. My suggestion is to let it hang for a couple of days, feel free to ping me then.

thalesmg:
> Interesting find! I'm now curious to see if this can be done without having to lose snabbkaffe tracing. My suggestion is to let it hang for a couple of days, feel free to ping me then.

in that case, let's add the retry hook: https://github.com/emqx/emqx/pull/16215

PR #16204, opened 2025-11-01
ci: remove unused permissions from the docker workflow

id:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10262)

PR #16211, opened 2025-11-03
feat(namespace): add initial namespaced channel metrics

thalesmg:
Part of https://emqx.atlassian.net/browse/EMQX-14274

Release version: 6.1.0

## Summary

To facilitate per-tenant billing, we will need to support isolated metrics collection per tenant.

This PR adds the following namespaced metrics:

- Messages received
  - Count
  - Bytes
- Messages sent
  - Count
  - Bytes

The following metrics are also part of the epic, but will be added in follow ups.

- Number of sessions
- Messages received
  - Rate
- Messages sent
  - Rate
- Data integration
  - Number of actions triggered
- DB recrods
  - Number of AuthN records
  - Number of AuthZ records

See also individual commits for details.

Follow up: https://github.com/emqx/emqx/pull/16346

## PR Checklist

### DIFF
--- a/apps/emqx_mt/mix.exs
+++ b/apps/emqx_mt/mix.exs
@@ -7,7 +7,7 @@ defmodule EMQXMT.MixProject do
       app: :emqx_mt,
       version: "6.0.1",
       build_path: "../../_build",
-      erlc_options: UMP.erlc_options(),
+      erlc_options: UMP.strict_erlc_options(),
       erlc_paths: UMP.erlc_paths(),
       deps_path: "../../deps",
       lockfile: "../../mix.lock",


apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -597,6 +601,10 @@ handle_msg({event, disconnected}, State = #state{channel = Channel}) ->
     ClientId = emqx_channel:info(clientid, Channel),
     emqx_cm:set_chan_info(ClientId, info(State)),
     {ok, State};
+%% TODO: handle `{event, {zone_changed, NewZone}}`
```
thalesmg, line 604: https://emqx.atlassian.net/browse/EMQX-14869

apps/emqx/src/emqx_metrics.erl
```diff
@@ -21,30 +22,42 @@
...
-    dec/1,
-    dec/2,
-    set/2
+    all_global/0,
+    val_global/1,
+    inc_global/1,
+    inc_global/2,
+    dec_global/1,
+    dec_global/2,
+    set_global/2
```
zmstone, line 38: maybe we can keep the old APIs unchanged ?

apps/emqx/src/emqx_metrics.erl
```diff
@@ -389,6 +464,21 @@ handle_call({set_type_to_counter, Keys}, _From, State) ->
         Keys
     ),
     {reply, ok, State};
+handle_call(#register_namespace{namespace = Namespace}, _From, State) ->
+    case persistent_term:get(?PT_KEY(Namespace), undefined) of
+        undefined ->
+            CRef = counters:new(?MAX_SIZE, [write_concurrency]),
```
zmstone, line 475: this is probably micro optimization, but we should perhaps try to limit the size for less memory consumption.
same for the default, i don't think we have 1024 counters.

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -1230,6 +1238,16 @@ inc_qos_stats(Type, Packet) ->
...
 
+inc_metrics(Name, State, Val) ->
+    case State#state.namespace of
+        ?global_ns ->
+            emqx_metrics:inc(?global_ns, Name, Val);
+        Namespace ->
+            emqx_metrics:inc(?global_ns, Name, Val),
+            _ = emqx_metrics:inc_safe(Namespace, Name, Val),
+            ok
+    end.
```
zmstone, line 1249:
```suggestion
    emqx_metrics:inc(?global_ns, Name, Val),
    case State#state.namespace of
        ?global_ns ->
            ok;
        Namespace ->
            _ = emqx_metrics:inc_safe(Namespace, Name, Val),
            ok
    end.
```

apps/emqx/src/emqx_metrics.erl
```diff
@@ -138,207 +159,270 @@ create(Type, Name) ->
...
+
+-spec inc(maybe_namespace(), metric_name(), pos_integer()) -> ok.
+inc(Namespace, Name, Value) ->
+    update_counter(Namespace, Name, Value).
+
+-spec inc_safe(maybe_namespace(), metric_name()) -> ok | {error, not_found}.
+inc_safe(Namespace, Name) ->
+    inc_safe(Namespace, Name, 1).
+
+-spec inc_safe(maybe_namespace(), metric_name(), pos_integer()) -> ok | {error, not_found}.
```
zmstone, line 246: i don't see any caller handles the error, maybe always return ok.

changes/ee/feat-16211.en.md
```diff
@@ -0,0 +1,10 @@
+Added initial support for namespaced metrics.
+
+- Messages received
+  - Count
+  - Bytes
+- Messages sent
+  - Count
+  - Bytes
+
+Clients in managed namespaces will bump the namespaced metrics above, as well as continue to bump the usual global metrics.
```
zmstone, line 10: also add how it's exposed to users for observability?

zmstone (commented):
maybe add a new prometheus scraping endpoint for the tenant metrics

one endpoint for all tenants
another one for one specific tenant.

apps/emqx/src/emqx_metrics.erl
thalesmg: It's unchanged, it's only the name that's clearer that it only affects the global NS.  I changed the name exactly to avoid subtle issues due to extending only the arity with the namespace argument without changing the name.

apps/emqx/src/emqx_metrics.erl
thalesmg: I copied the existing initialization for the global counters.  :thinking:

apps/emqx/src/emqx_metrics.erl
thalesmg: no caller uses it _now_, but returning something here gives the caller a choice

thalesmg:
> maybe add a new prometheus scraping endpoint for the tenant metrics
> 
> one endpoint for all tenants another one for one specific tenant.

added a endpoint in last few commits; passing a namespace as a query parameter returns data only from that ns, otherwise returns data from all.

apps/emqx/src/emqx_metrics.erl
zmstone: we can optimize global counters size too (i don't think there are 1024 global counters).
Even less for namespace counters.
Given it's per namespace, maybe worth optimize.

```
2> C = counters:new(1024, [write_concurrency]).
{write_concurrency,#Ref<0.2769774019.1654784001.198075>}
3> counters:info(C).
#{memory => 172064,size => 1024} <- 172KB
4> D = counters:new(10, [write_concurrency]).
{write_concurrency,#Ref<0.2769774019.1654784001.198126>}
5> counters:info(D).
#{memory => 2720,size => 10} <- 2KB
```

apps/emqx/src/emqx_metrics.erl
zmstone: i meant add new api for namespace metrics, not to add `global | namespace` as argument (i.e. not just change arity).

apps/emqx/src/emqx_metrics.erl
thalesmg: given it's a (micro?) optimization, maybe it's best left for another time?

specially since it basically renders `reserved_idx` useless/obsolete and will require something else more complicated in its place to allow dynamic registration of metrics?  Or, at least, coming up with the new limit.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10621)

PR #16230, opened 2025-11-05
251103 i18n docs

zmstone:
this concludes all

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
@@ -714,9 +714,12 @@ trans_description(Spec, Hocon, Options) ->
     end.
 
 -ifdef(TEST).
+%% Do not raise error in tests because there are schema defined in tests.
 missing_i18n_ref(Text) ->
     to_bin(Text).
 -else.
+%% Fail in production, if any translation is missing, the node will not boot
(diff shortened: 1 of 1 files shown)


apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
```diff
@@ -714,15 +713,14 @@ trans_description(Spec, Hocon, Options, Module) ->
...
-maybe_warn_missing_desc(Hocon, Text, Module) ->
-    case os:getenv("WARN_MISSING_DESC") of
-        "1" when Text =/= <<>> ->
-            io:format(user, "~p Missing-api-translation: ~s~n", [Module, Text]);
-        "1" ->
-            io:format(user, "~p Missing-api-translation: ~0p~n", [Module, Hocon]);
-        _ ->
-            ok
-    end.
+-ifdef(TEST).
```
savonarola, line 716: shouldn't be vice versa?
Fail in tests, but tolerate in prod.

zmstone: this is because there are schema defined in test suites.

we fail in production code because if any translation is missing, it won't boot, so at least the smoke tests will catch it.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10406)

PR #16254, opened 2025-11-07
revert "fix(retainer): change retainer iteration flow to allow iteration when hitting rate limit" (r59)

thalesmg:
Performance issues were found due to this patch, so we'll revert it to unblock the release.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10467)

PR #16261, opened 2025-11-10
fix(KafkaConsumer): check leader connection only for assigned partitions [v5]

zmstone:
Fixes [EMQX-14891](https://emqx.atlassian.net/browse/EMQX-14891)

Release version: 5.10.3

Previously, Kafka source connector health-check checks partition leader connectivity for all partitions. In case of cluster, only a subset of partitions will be assigned for each EMQX node, this causes the leader connections for unassigned partitions idling. Kafka close idle connections after a while (10m by default), this leads to false connectivity alarms.

## PR Checklist

[EMQX-14891]: https://emqx.atlassian.net/browse/EMQX-14891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/ee/fix-16261.en.md
+++ b/changes/ee/fix-16261.en.md
@@ -1,0 +1,5 @@
+Previously, the Kafka source connector performed health checks by verifying partition leader connectivity for all partitions.
+In a clustered deployment, each EMQX node is assigned only a subset of partitions, causing leader connections for unassigned partitions to remain idle.
+Since Kafka closes idle connections after a timeout (10 minutes by default), this behavior could trigger false connectivity alarms.
+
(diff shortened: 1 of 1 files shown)


id:
Let's move this to 6.1.0 unless we really have to fix 5.10.x.

zmstone:
> Let's move this to 6.1.0 unless we really have to fix 5.10.x.

it's needed for 5.8, 5.9, and 5.10, i have created https://github.com/emqx/emqx/pull/16263 for it, will forward merge to 5.10.
will send another pr to release-60

PR #16271, opened 2025-11-11
chore: upgrade to brod 4.4.7

zmstone:
No user impact.
- Bring dependency of dependency to consitency
- brod-4.4.7 added a `warning` level log message in consumer group member process which can help to troubleshoot in some (probably rare) cases. (only observed in CI though)


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10627)

PR #16272, opened 2025-11-11
feat(mnesia authz): namespace acl rules

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14811

Release version: 6.1.0

## PR Checklist

### DIFF
--- a/apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
+++ b/apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
@@ -930,11 +930,25 @@ get_namespace(#{query_string := QueryString} = Req) ->
             {ok, QSNamespace}
     end.
 
+validate_managed_namespace(#{resolved_ns := ?global_ns} = Req, _Meta) ->
+    {ok, Req};
+validate_managed_namespace(#{resolved_ns := Namespace} = Req, _Meta) ->
+    Res = emqx_hooks:run_fold('namespace.resource_pre_create', [#{namespace => Namespace}], #{
+        exists => false
+    }),
+    case Res of
+        #{exists := false} ->
+            ?BAD_REQUEST(<<"Managed namespace not found">>);
+        #{exists := true} ->
+            {ok, Req}
+    end.
+
 table(?global_ns) -> ?ACL_TABLE;
 table(_Namespace) -> ?ACL_NS_TABLE.
 
 filter(Req0, Meta) ->
     maybe
         {ok, Req1} ?= is_configured_authz_source(Req0, Meta),
-        resolve_namespace(Req1, Meta)
+        {ok, Req2} ?= resolve_namespace(Req1, Meta),
+        validate_managed_namespace(Req2, Meta)
     end.
--- a/apps/emqx/src/emqx_hookpoints.erl
+++ b/apps/emqx/src/emqx_hookpoints.erl
@@ -68,7 +68,8 @@
     'tls_handshake.psk_lookup',
     'config.zones_updated',
     'api_actor.pre_create',
-    'namespace.delete'
+    'namespace.delete',
+    'namespace.resource_pre_create'
 ]).
 
 -define(HOOKPOINTS, (?MQTT_CLIENT_LIFECYCLE_HOOKPOINTS ++ ?MANAGEMENT_HOOKPOINTS)).
@@ -288,6 +289,15 @@ when
 %% Executed out of channel process context
 %% Implementations must be idempotent.
 -callback 'namespace.delete'(emqx_config:namespace()) -> callback_result().
+
+%% NOTE
+%% Executed out of channel process context
+%% Called before create a namespaced resource (e.g.: mnesia authz rules) as a way for
+%% initial applications to check if a namespace exists, since that information lives in
+%% the terminal `emqx_mt` application.
+%% Should return `true` if managed namespace exists, `false` otherwise.
+-callback 'namespace.resource_pre_create'(#{namespace => emqx_config:namespace()}) ->
+    fold_callback_result(#{exists := boolean()}).
 
 %%-----------------------------------------------------------------------------
 %% API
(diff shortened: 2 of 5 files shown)


apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
```diff
@@ -811,29 +860,49 @@ ensure_rules_len(Rules, MaxLen) ->
...
+            {403, <<"User not authorized to operate on requested namespace">>}
+    end.
+
+get_namespace(#{query_string := QueryString} = Req) ->
+    ActorNamespace = emqx_dashboard:get_namespace(Req),
+    case maps:get(<<"ns">>, QueryString, ActorNamespace) of
+        QSNamespace when QSNamespace /= ActorNamespace andalso ActorNamespace /= ?global_ns ->
+            {error, not_authorized};
+        QSNamespace ->
+            {ok, QSNamespace}
```
zmstone, line 930: is ns existence checked during admin/app key authentication?

thalesmg: during admin/app key creation

zmstone: one should not be able to create acl rules for non-existing ns?

thalesmg: if the admin user creating the rule exists, it means that the ns exists

the admin/app key lifetimes are tied to the ns': we delete them when we delete the ns

zmstone: can global admin create acl records for a non-existing tenant?

thalesmg: Yes.  Wouldn't admins want to setup acls before creating the ns and thus allowing clients to connect?

If we wanted to be super strict with this, we'll need a way to call the terminal app where nses are kept from this initial app.

zmstone: i’m concerned about orphan namespaces in acl due to typos etc. 

maybe an alert to remind admins that there are acl records created for non-existing namespaces?

apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
```diff
@@ -491,14 +498,139 @@ is_configured_authz_source(Params, _Meta) ->
         end
     ).
 
-users(get, #{query_string := QueryString}) ->
+users(get, Req) ->
```
savonarola, line 542: Nit: we probably may put `Namespace` into bindings via a filter (avoiding nesting many methods into `with_namespace` fun)

apps/emqx_auth_mnesia/src/emqx_authz_mnesia.erl
```diff
@@ -124,75 +148,110 @@ init_tables() ->
...
     ets:fun2ms(
-        fun(#emqx_acl{who = {?ACL_TABLE_CLIENTID, Clientid}, rules = Rules}) ->
+        fun(#?ACL_TABLE{who = {?ACL_TABLE_CLIENTID, Clientid}, rules = Rules}) ->
+            [{clientid, Clientid}, {rules, Rules}]
+        end
+    );
+list_clientid_rules(Namespace) when is_binary(Namespace) ->
+    ets:fun2ms(
+        fun(#?ACL_NS_TABLE{who = ?WHO_NS(Ns, {?ACL_TABLE_CLIENTID, Clientid}), rules = Rules}) when
+            Ns == Namespace
```
savonarola: As far as I understand, such `ets:fun2ms` will produce a match spec with Ns constraint in Guard, not in MatchHead, which will result into full scans.
We probably should compose the matchspec manually to take advantage of ets ordering.

apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
savonarola: Like https://github.com/emqx/emqx/blob/816788fdbfaee7ee7ada1b619606a36ec9ab6eb1/apps/emqx_mq/src/emqx_mq_api.erl#L341

apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
```diff
@@ -512,12 +644,35 @@ users(get, #{query_string := QueryString}) ->
...
+            fun ?MODULE:format_result/1
+        )
+    of
+        {error, page_limit_invalid} ->
+            {400, #{code => <<"INVALID_PARAMETER">>, message => <<"page_limit_invalid">>}};
+        {error, Node, Error} ->
+            Message = list_to_binary(
+                io_lib:format("bad rpc call ~p, Reason ~p", [Node, Error])
+            ),
+            {500, #{code => <<"NODE_DOWN">>, message => Message}};
```
savonarola: Nit: maybe also update to `emqx_http_api.hrl` helpers

apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
thalesmg: I think the main problem is inverting the calls from this initial app to the terminal mt app; will need to inject it somewhere to avoid calling it directly.  :thinking:

apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
thalesmg: :checkered_flag:

apps/emqx_auth_mnesia/src/emqx_authz_mnesia.erl
thalesmg: :checkered_flag:

apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
thalesmg: :checkered_flag:

apps/emqx_auth_mnesia/src/emqx_authz_api_mnesia.erl
thalesmg: :checkered_flag:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10626)

PR #16300, opened 2025-11-17
Dev/ds backup

ieQu1:
Fixes <issue-or-jira-number>

Release version:

## Summary

There are several challenges to taking a backup of the DS:

1. Data consistency
2. Sharding
3. Side effects

### Data consistency

Problem: information in different durable storages can be implicitly related: for example, `sessions` DB stores iterators pointing at `messages` DB. 
This observation leads to the first constraint imposed on the design: backups must include all DBs.
Partial backups are useless. 

In addition, the order of taking the backup must be followed.
Let us again consider the `sessions` - `messages` DB pair:
if backup of `messages` DB doesn't contain data that has been already read by `sessions`, then the restored state is inconsistent.
There's no general solution to this problem, as data is sharded in multiple dimensions and we don't have the ability to take a consistent snapshot of all shards.
But if we make a reasonable assumption that there are no circular logical references between the DBs (i.e. `messages` never contain iterators of `sessions`), then this problem is solved by imposing a deterministic order of taking the backup, defined by the business logic level.
In our example the order is the following: 

1. `sessions`, `shared_subs` 
2. `messages`

There are two possible APIs that can do that:

1. Add an integer `backup_priority` option to the DB settings. 
2. Create a backup API that lets the user pass the list of DBs to back up.

The procedure of taking the backup should look like this on the high level:

```

              shard1                               shard1
  take       /      \    DB1          take        /      \   DB2
  DB1--->-----shard2--->-snapshot---->DB2------>---shard2--->snapshot
  snapshot   \      /    ready        snapshot    \      /   ready
              shard3      \                        shard3      \
                           "---> backup ---------------.        \
                                snapshots               \        \
                                                         \        "---> backup -------.
                                                          \            snapshots       \
                                                           "---------------------------->backup ready
``` 

Sequence should be enforced when taking snapshot: 
backup must not proceed to the next DB until snapshots of all shards of the current DB are taken.

Taking backups of the shard snapshots (read: copying SST files) may be done asynchronously.  

### Sharding

Problem: Raft backend is designed to shard data across different sites.
This helps with horizontal scalability: to handle more data one can add more sites, 
but it creates a challenge for the backup, 
as no single site can have the entire dataset.

This can be addressed by having two types of backups: local and remote.

Local backups rely on a distributed file system (NFS, SMB, HDFS, ...) to gather data from all sites in one place.

Remote backups involve may involve an additional step of transferring (`rsync`) the local backup to the remote host. 
We do not consider this type of backup for now.

### Side effects

Problem: sometimes it's not sufficient to simply restore the DS.
For example, durable sessions make changes in the routing table.

Restoring the backup must involve user code that performs all necessary side effects.

### Restoration with sites

Problem: site IDs must be taken into consideration when restoring a backup.

Solution: TODO

## PR Checklist

### DIFF
--- a/apps/emqx_durable_storage/asn.1/BackupMetadata.asn
+++ b/apps/emqx_durable_storage/asn.1/BackupMetadata.asn
@@ -1,0 +1,10 @@
+-- This schema specifies binary encoding of shard backup metadata.
+
+DSBackupInfo DEFINITIONS AUTOMATIC TAGS ::=
+BEGIN
+  ShardInfo ::= SEQUENCE {
+    vsn INTEGER,
+    committedSerial OPTION INTEGER,
+    timestamp OPTION INTEGER
+  }
+END
--- a/apps/emqx_durable_storage/src/emqx_ds_backup.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_backup.erl
@@ -1,0 +1,69 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_ds_backup).
+-moduledoc """
+This process supervises backup collection and management.
+
+DS backup includes schema and all durable storage databases.
+""".
+
+-behavior(gen_server).
+
+%% API:
+-export([start_link/0]).
+
+%% behavior callbacks:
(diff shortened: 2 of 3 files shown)


ieQu1:
Solution 2: create a new generation in messages DB after restoring the backup. The biggest problem is message "overwrite" that could happen if messages go missing during backup and get overwritten. But if we create a new generation, session can detect missing messages, and new messages will go to the new generation.

PR #16314, opened 2025-11-19
chore(namespaced resources api): redo global admin ux

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14807

Release version: 6.1.0

## Summary

Now, global admin users will see resources from all namespaces (by default) when listing namespaced resources (connectors/sources/actions/rules).  They may focus on one particular namespace when performing CRUD operations by passing the `ns=NS` query parameter.  If they want to list only the global namespace resources, they omit `ns` and pass `only_global=true` query parameter.  Namespaced resources now return the `namespace` field to denote where they come from, with `namespace` being `null` for global resources to distinguish them from a potential namespace called `"global"`. 

## PR Checklist


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10769)

PR #16320, opened 2025-11-19
test(takeover): mark test case as flaky

thalesmg:
https://github.com/emqx/emqx/actions/runs/19505457236/job/55839596466?pr=16316#step:5:281


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10680)

PR #16326, opened 2025-11-20
MQTT Streams

savonarola:
Fixes [EMQX-14842](https://emqx.atlassian.net/browse/EMQX-14842)

Release version: 6.1.0

## Summary

See `apps/emqx_streams/README.md`. 

The write path, schema, HTTTP API and config are mostly similar to the MQs.
The read (subscribe) path is much simpler.

## PR Checklist

[EMQX-14842]: https://emqx.atlassian.net/browse/EMQX-14842?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10868)

apps/emqx_streams/src/emqx_streams_api.erl
```diff
@@ -0,0 +1,401 @@
...
+    {ok, undefined};
+decode_cursor(EncodedCursor) ->
+    try
+        {ok, emqx_base62:decode(EncodedCursor)}
+    catch
+        _:_ ->
+            bad_cursor
+    end.
+
+add_message_stream(NewStreamRaw) ->
```
ieQu1, line 369: Input validation?

apps/emqx_streams/src/emqx_streams_api.erl
```diff
@@ -0,0 +1,401 @@
...
+
+get_message_stream(TopicFilter) ->
+    case emqx_streams_registry:find(TopicFilter) of
+        not_found ->
+            not_found;
+        {ok, Stream} ->
+            {ok, emqx_streams_config:stream_to_raw_get(Stream)}
+    end.
+
+update_message_stream(TopicFilter, UpdatedStreamRaw) ->
```
ieQu1, line 386: Input validation?

savonarola: It's a private function, and it is called with data validated with api schema.
Also, converters from raw (e.g. `emqx_streams_config:stream_update_from_raw_put/1`) also make validation with schema, so if some invalid data sneaks into the input of the function, the first line will crash.

apps/emqx_streams/src/emqx_streams_api.erl
savonarola: See https://github.com/emqx/emqx/pull/16326/files/1a000501451c2bfce75006c7e2e9b5f909288c1c#r2609690337

apps/emqx_extsub/README.md
```diff
@@ -1,8 +1,136 @@
...
+    {ok, state()} | ignore.
+```
+
+If the handler wants to handle the topic, then it should return `{ok, state()}`. If the handler is not interested in the topic, it should return `ignore`.
+
+The `subscribe_ctx()` contains callback functions that the handler may use to send messages to self (optionally with delay).
+
+`handle_subscribe` callback is called when the client subscribes to a topic. For a single-topic handler, state() is always `undefined`. For a multi-topic handler, state() is either `undefined` if no topic has been handled yet, or the state of the handler for the previous topics.
+
+The subscribe
```
keynslug, line 71: Nit. Some leftoveers?

apps/emqx_extsub/README.md
```diff
@@ -1,8 +1,136 @@
...
+
+#### `handle_delivered` callback
+
+```erlang
+-callback handle_delivered(state(), ack_ctx(), emqx_types:message(), emqx_extsub_types:ack()) ->
+    state().
+```
+
+`handle_delivered` callback is called when a message provided by the handler has been delivered to the client.
+For QoS 0 messages this happens when the message is pushed to the inflight. For QoS 1 and 2 messages, this happens when the message is acknowledged by the client (if the underlying session supports acknowledgements; otherwise the callback is called when the message was delivered to the inflight).
```
keynslug, line 100: Nit. QoS 0 does not hit session's inflight AFAICT, perhaps "sent to the connection"?

apps/emqx_streams/src/emqx_streams_gc/emqx_streams_gc.erl
```diff
@@ -0,0 +1,96 @@
...
+    case is_responsible() of
+        true ->
+            ?tp_debug(streams_gc_starting, #{}),
+            ok = emqx_streams_sup:start_gc();
+        false ->
+            ok
+    end.
+
+gc_interval() ->
+    emqx_config:get([streams, gc_interval]).
```
keynslug, line 88: Nit. Should probably go to `emqx_streams_config`?

apps/emqx_streams/src/emqx_streams_api.erl
```diff
@@ -0,0 +1,401 @@
...
+decode_cursor(EncodedCursor) ->
+    try
+        {ok, emqx_base62:decode(EncodedCursor)}
+    catch
+        _:_ ->
+            bad_cursor
+    end.
+
+add_message_stream(NewStreamRaw) ->
+    NewStream = emqx_streams_config:stream_from_raw_post(NewStreamRaw),
```
keynslug, line 370: Q: Wouldn't using `translate_body => {true, atom_keys}` in `api_spec/0` give this essentially for free?

apps/emqx_streams/src/emqx_streams_message_db.erl
```diff
@@ -0,0 +1,699 @@
...
+    dirty_index/2
+]).
+
+-define(STREAMS_MESSAGE_DB_APPEND_RETRY, 3).
+-define(STREAMS_MESSAGE_DB_QUOTA_INDEX_APPEND_RETRY, 3).
+-define(STREAMS_MESSAGE_DB_DELETE_RETRY, 1).
+-define(STREAMS_MESSAGE_DB_DELETE_RETRY_DELAY, 1000).
+-define(STREAMS_MESSAGE_DB_LTS_SETTINGS, #{
+    %% "topic/STREAM_TOPIC/STREAM_ID/key/СOMPACTION_KEY"
+    lts_threshold_spec => {simple, {100, 0, 0, 100, 0}}
```
keynslug, line 71: Nit. Perhaps put it next to `?STREAMS_MESSAGE_DB_TOPIC` macrodef?

apps/emqx_streams/src/emqx_streams_message_db.erl
```diff
@@ -0,0 +1,699 @@
...
+db(Stream) ->
+    case emqx_streams_prop:is_append_only(Stream) of
+        true ->
+            ?STREAMS_MESSAGE_REGULAR_DB;
+        false ->
+            ?STREAMS_MESSAGE_LASTVALUE_DB
+    end.
+
+insert_generation(?STREAMS_MESSAGE_LASTVALUE_DB) ->
+    1.
```
keynslug, line 665: Nit. Having this indirection makes not much sense, maybe just put it behind a macrodef?

apps/emqx_streams/src/emqx_streams_message_db.erl
```diff
@@ -0,0 +1,699 @@
...
+
+retry_interval(?STREAMS_MESSAGE_LASTVALUE_DB) ->
+    emqx:get_config([durable_storage, streams_messages, transaction, flush_interval], 100) * 2.
```
keynslug, line 668: Perhaps choice of this particular value as interval here worth briefly documenting?

apps/emqx_streams/src/emqx_streams_registry.erl
```diff
@@ -0,0 +1,370 @@
...
+delete(TopicFilter) ->
+    ?tp_debug(mq_registry_delete, #{topic_filter => TopicFilter}),
+    Key = make_key(TopicFilter),
+    case mnesia:dirty_read(?STREAMS_REGISTRY_INDEX_TAB, Key) of
+        [] ->
+            not_found;
+        [#?STREAMS_REGISTRY_INDEX_TAB{} = Rec] ->
+            ok = mria:dirty_delete_object(Rec),
+            Stream = record_to_stream(Rec),
+            emqx_streams_message_db:drop(Stream)
```
keynslug, line 176: Q: Shouldn't we do it the other way around? Otherwise we can report error to the user but: refreshing the page will make it look like deletion succeeded, but there will still be leftovers in the DS DB (which AFAICT GC won't notice).

apps/emqx_streams/src/emqx_streams_message_db.erl
```diff
@@ -0,0 +1,699 @@
...
+
+%%--------------------------------------------------------------------
+%% Internal functions
+%%--------------------------------------------------------------------
+
+key(#{key_expression := KeyExpression} = Stream, Message) ->
+    Bindings = #{message => message_to_map(Message)},
+    case emqx_variform:render(KeyExpression, Bindings, #{eval_as_string => true}) of
+        {error, Reason} ->
+            ?tp(warning, streams_message_db_key_expression_error, #{
```
keynslug, line 548: Q: Does it need throttling?

apps/emqx_streams/src/emqx_streams_prop.erl
```diff
@@ -0,0 +1,90 @@
...
+quota_index_opts(#{
+    limits := #{
+        max_shard_message_count := MaxShardMessageCount,
+        max_shard_message_bytes := MaxShardMessageBytes
+    }
+}) ->
+    maps:from_list(
+        limit_to_index_opt(bytes, MaxShardMessageBytes) ++
+            limit_to_index_opt(count, MaxShardMessageCount)
+    ).
```
keynslug, line 64: Tiny nit. Feels a bit too complicated for such trivial thing, maybe teach `emqx_mq_quota_buffer` to work with `#{bytes => unlimited, ...`?

changes/ee/feat-16326.en.md
```diff
@@ -0,0 +1,10 @@
+Implemented Message Streams.
+
+Message Streams are collections of messages identified by `topic/filter`.
+Message Streams have an explicit lifecycle and are automatically replenished with published messages matched with the stream's topic filter during the stream's lifetime.
+
+Streams allow repetitive and consumption of messages with ordering guarantees.
+
+Clients may subscribe to a special topic of the form `$s/<timestamp>/topic/filter` to consume messages from the stream
+starting from the given timestamp. The timestamp may be a unix timestamp in microseconds or take one of special values,
+`earliest` or `latest`.
```
keynslug, line 10:
```suggestion
Message Streams are durable collections of messages identified by a topic filter.
They have an explicit lifecycle, and any published message that matches the Stream’s topic filter is automatically appended to the stream.

Streams allow consumption of messages with ordering guarantees and can be consumed multiple times.

To consume messages from a stream, clients can subscribe to a special topic of the form
`$s/<timestamp>/topic/filter`, where `topic/filter` refers to an existing Stream. Subscribing with a timestamp allows consumption to begin at a specific point in time. The timestamp may be a Unix timestamp in microseconds or one of two special values: `earliest` or `latest`.
```

keynslug (approved):
🚀 

Mostly nits. Feel free to push them into followups if needed.

savonarola:
Thanks, @keynslug, will address the comments now in follow-ups

apps/emqx_streams/src/emqx_streams_api.erl
savonarola: We need non-atom bodies in the controller as well to pass to update config.

PR #16331, opened 2025-11-21
251119 fix kafka consumer group health check v5

zmstone:
Fixes [EMQX-14891](https://emqx.atlassian.net/browse/EMQX-14891)

Release version: 5.8.9, 5.9.3, 5.10.3

### This PR fixes two issues
1. False connectivity alarm for Kafka consumer.
2. Kafka down may cause group subscriber supervisor to reach max intensity, then cause consumers to stop.

Issue 2 was discovered during manual test of the fix for issue 1.

### The failed attempt
Issue 1 was previously attempted by https://github.com/emqx/emqx/pull/16263, but it was not sufficient.
Because the old connectivity check was checking **shared** payload connections managed by `brod_client`,
However, since `brod-4.2.0` partition consumers no longer use the shared connections, they establish their own connections instead. Hence the shared connections will idle then eventually closed by Kafka server.

### The new fix

For issue 1: exposed a health-check API from `brod`. EMQX no longer has to check individual partition workers.
For issue 2: changed supervisor to use `brod_supervisor3` which supports delayed restart so the children's restarts will never cause supervisor to reach max restart intensity.

## PR Checklist

[EMQX-14891]: https://emqx.atlassian.net/browse/EMQX-14891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
+++ b/apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl
@@ -553,6 +553,61 @@ op_bridge_api(Kind, Op, BridgeType, BridgeName) ->
     ct:pal("bridge op result:\n  ~p", [Res]),
     Res.
 
+%% Assert bridge start operation with retry logic for kafka_consumer rebalancing
+assert_bridge_start(Kind, Type, Name) when Type =:= <<"kafka_consumer">> ->
+    %% For kafka_consumer, retry when encountering "Consumer group rebalancing" errors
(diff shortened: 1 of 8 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10771)

PR #16336, opened 2025-11-24
251124 ensure cleanup is not killed

zmstone:
Release version: 5.8.9, 5.9.3, 5.10.3

## Summary

The issue is discovered in test cases.
The effect of a dangling resource varies per resource type.
For example, for Kafka producer, it means a terminated child reference in supervisor.

To user experience, it mostly just a timeout when clicking "test-connectivity" or "disable" (connector) from the dashboard.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/test/emqx_resource_tests.erl
+++ b/apps/emqx_resource/test/emqx_resource_tests.erl
@@ -52,7 +52,7 @@ cleanup_by_agent_test_() ->
     ].
 
 test_normal_completion() ->
-    LogCalls = ets:new(log_calls, [set, public]),
+    LogCalls = ets:new(log_calls, [bag, public]),
     LogFn = fun(Level, LogData) ->
         ets:insert(LogCalls, {Level, LogData}),
         ok
@@ -72,7 +72,7 @@ test_normal_completion() ->
     ets:delete(LogCalls).
 
 test_abnormal_completion() ->
-    LogCalls = ets:new(log_calls, [set, public]),
+    LogCalls = ets:new(log_calls, [bag, public]),
     LogFn = fun(Level, LogData) ->
         ets:insert(LogCalls, {Level, LogData}),
         ok
@@ -95,7 +95,7 @@ test_abnormal_completion() ->
     ets:delete(LogCalls).
 
 test_parent_dies_before_task() ->
-    LogCalls = ets:new(log_calls, [set, public]),
+    LogCalls = ets:new(log_calls, [bag, public]),
     LogFn = fun(Level, LogData) ->
         ets:insert(LogCalls, {Level, LogData}),
         ok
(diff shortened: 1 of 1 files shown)


apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -598,19 +604,76 @@ force_kill(ResId, MRef0) ->
                 {'DOWN', MRef, process, Pid, _} ->
                     ok
             end,
-            try_clean_allocated_resources(ResId),
-            ok
+            LogFn = fun(Level, LogData) -> ?SLOG(Level, LogData#{resource_id => ResId}) end,
+            ok = cleanup_by_agent(
```
keynslug, line 608: Q: Simply submit to a `emqx_pool` instead? Or is it potentially too long running?

zmstone: yes, it can take too long, blocking the pool workers is perhaps not a good idea.

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -598,19 +604,76 @@ force_kill(ResId, MRef0) ->
...
+%% but the agent should continue to run (not linked) to ensure the task is complete.
+cleanup_by_agent(LogFn, TaskFn, Timeout) ->
+    Parent = self(),
+    {Agent, _} = erlang:spawn_monitor(fun() -> run_agent(LogFn, Parent, TaskFn, Timeout) end),
+    %% Agent has a 60s timeout to complete,
+    %% the caller (cowboy handler) is unlikely to wait this long, so there is no 'after' clause here.
+    receive
+        {'DOWN', _, process, Pid, {cleanup_result, normal}} when Pid =:= Agent ->
+            LogFn(debug, #{msg => "resource_cleanup_done"});
+        {'DOWN', _, process, Pid, Reason} when Pid =:= Agent ->
```
thalesmg: Q: why not bind the returned monitor reference?  It might have a higher chance of triggering the ref trick optimization, instead of using a guard to match on the pid?

apps/emqx_resource/test/emqx_resource_tests.erl
```diff
@@ -37,3 +37,193 @@ is_dry_run_test_() ->
...
+        {"abnormal completion", fun test_abnormal_completion/0},
+        {"parent dies before task completes", fun test_parent_dies_before_task/0},
+        {"timeout scenario", fun test_timeout/0},
+        {"orphaned agent - parent dies then task completes", fun test_orphaned_agent/0}
+    ].
+
+test_normal_completion() ->
+    LogCalls = ets:new(log_calls, [set, public]),
+    LogFn = fun(Level, LogData) ->
+        ets:insert(LogCalls, {Level, LogData}),
```
thalesmg: If the key is the level, we won't be able to see multiple messages of the same level.  e.g.: `cleanup_agent_is_orphanated` followed by `orphanated_cleanup_agent_completed`.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10786)

PR #16345, opened 2025-11-25
feat(mnesia authn): add namespace support

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14810

Release version: 6.1.0

## PR Checklist

### DIFF
--- a/apps/emqx_gateway/src/emqx_gateway_api_authn.erl
+++ b/apps/emqx_gateway/src/emqx_gateway_api_authn.erl
@@ -34,6 +34,8 @@
 
 -define(TAGS, [<<"Gateway Authentication">>]).
 
+-define(NAMESPACE, ?global_ns).
+
 %%--------------------------------------------------------------------
 %% minirest behaviour callbacks
 %%--------------------------------------------------------------------
@@ -184,7 +186,6 @@ schema("/gateways/:name/authentication/users") ->
                 desc => ?DESC(list_users),
                 summary => <<"List users for gateway authenticator">>,
                 parameters => lists:flatten([
-                    ns_qs_param(),
                     params_gateway_name_in_path(),
                     params_paging_in_qs(),
                     params_fuzzy_in_qs()
@@ -205,7 +206,6 @@ schema("/gateways/:name/authentication/users") ->
                 desc => ?DESC(add_user),
                 summary => <<"Add user for gateway authenticator">>,
                 parameters => lists:flatten([
-                    ns_qs_param(),
                     params_gateway_name_in_path()
                 ]),
                 'requestBody' => emqx_dashboard_swagger:schema_with_examples(
@@ -233,7 +233,6 @@ schema("/gateways/:name/authentication/users/:uid") ->
                 desc => ?DESC(get_user),
                 summary => <<"Get user info for gateway authenticator">>,
                 parameters => lists:flatten([
-                    ns_qs_param(),
                     params_gateway_name_in_path(),
                     params_userid_in_path()
                 ]),
@@ -253,7 +252,6 @@ schema("/gateways/:name/authentication/users/:uid") ->
                 desc => ?DESC(update_user),
                 summary => <<"Update user info for gateway authenticator">>,
                 parameters => lists:flatten([
-                    ns_qs_param(),
                     params_gateway_name_in_path(),
                     params_userid_in_path()
                 ]),
@@ -277,7 +275,6 @@ schema("/gateways/:name/authentication/users/:uid") ->
                 desc => ?DESC(delete_user),
                 summary => <<"Delete user for gateway authenticator">>,
                 parameters => lists:flatten([
-                    ns_qs_param(),
                     params_gateway_name_in_path(),
                     params_userid_in_path()
                 ]),
@@ -341,9 +338,6 @@ params_fuzzy_in_qs() ->
                 }
             )}
     ].
-
-ns_qs_param() ->
-    {ns, mk(binary(), #{in => query, required => false})}.
 
 %%--------------------------------------------------------------------
 %% schemas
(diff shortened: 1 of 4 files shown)


apps/emqx_gateway/src/emqx_gateway_http.erl
```diff
@@ -572,6 +576,60 @@ to_list(B) when is_binary(B) ->
 sum_cluster_connections(List) ->
     sum_cluster_connections(List, 0, 0).
 
+parse_namespace(#{query_string := QueryString} = Req) ->
```
zmstone, line 579: So far, we do not support namespace for gateway clients.
We can probably just return 403 if API is called with NS

apps/emqx_gateway/src/emqx_gateway_api_authn.erl
```diff
@@ -90,63 +83,31 @@ authn(delete, #{bindings := #{name := Name0}}) ->
...
-            id := AuthId,
-            chain_name := ChainName
-        }
-    ) ->
-        emqx_authn_api:delete_user(ChainName, AuthId, UserId)
-    end).
+users_insta(get, #{bindings := #{uid := UserId}} = Req) ->
+    #{
+        gateway := #{id := AuthId, chain_name := ChainName},
+        resolved_ns := Namespace
```
zmstone, line 98: we don't support namespace for gateways for now. maybe hardcode it to global ns for now.

apps/emqx_gateway/src/emqx_gateway_http.erl
thalesmg: :checkered_flag:

apps/emqx_gateway/src/emqx_gateway_api_authn.erl
thalesmg: :checkered_flag:

apps/emqx_auth/test/data/user-credentials-plain-ns.json
```diff
@@ -0,0 +1,13 @@
+[
+    {
+        "user_id":"myuser1",
+        "password":"password1",
+        "is_superuser": true
+    },
+    {
+        "user_id":"myuser2",
+        "password":"password2",
+        "is_superuser": false,
```
zmstone, line 10: just a note: please check if superuser in a namespace will break jail.

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
```diff
@@ -1692,3 +1733,47 @@ authn_settings_example() ->
 
 authn_cache_status_example() ->
     emqx_auth_cache_schema:metrics_example().
+
+user_out(#{namespace := ?global_ns} = User) ->
+    User#{namespace := null};
```
zmstone, line 1738: maybe `maps:remove(namespace, User);` ?

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
```diff
@@ -838,23 +847,37 @@ listener_authenticator_position(
...
 
-authenticator_users(post, #{bindings := #{id := AuthenticatorID}, body := UserInfo}) ->
-    add_user(?GLOBAL, AuthenticatorID, UserInfo);
+authenticator_users(post, #{bindings := #{id := AuthenticatorID}, body := UserInfo} = Req) ->
+    ActorNamespace = emqx_dashboard:get_namespace(Req),
+    case UserInfo of
+        #{<<"namespace">> := UserNamespace} when
+            ActorNamespace /= ?global_ns andalso UserNamespace /= ActorNamespace
+        ->
+            ?FORBIDDEN(<<"User not authorized to operate on requested namespace">>);
```
savonarola, line 856: Why only post is protected from accessing by not authorised admins? Why are put and delete not checked?

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
```diff
@@ -883,44 +906,56 @@ listener_authenticator_users(get, #{
...
         end
     );
-listener_authenticator_user(get, #{
-    bindings := #{
-        listener_id := ListenerID,
-        id := AuthenticatorID,
-        user_id := UserID
-    }
-}) ->
+listener_authenticator_user(
```
savonarola, line 927: @zmstone could we already drop per-listener API? It is has not been exposed for long and supporting those `with_chain(...)` is a burden.

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
thalesmg: Put and delete are already checked by the filter function.  `POST` is special because the namespace is in the `user_info` rather than in the query param.

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
thalesmg: We may do that in a follow up, if @zmstone gives the green light.

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
thalesmg: I left the key to make the response consistent.  I guess, in this particular case, the choice between `null` and an absent key are mostly stylistic?  :thinking:

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
zmstone: ok. we can do it in a follow-up.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10806)

PR #16346, opened 2025-11-25
feat: add more namespaced metrics (part II)

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14274

Release version: 6.1.0

## Summary

Continuation to https://github.com/emqx/emqx/pull/16211

The following metrics are added in this PR:

- Number of sessions
- Data integration
  - Number of actions triggered
- DB recrods
  - Number of AuthN records
  - Number of AuthZ records

The following metrics were previously deemed necessary, but, after further discussions and refinements, we decided not to add them in EMQX:

- Messages received
  - Rate
- Messages sent
  - Rate

See also individual commits for details.

## PR Checklist

### DIFF
--- a/apps/emqx_auth/src/emqx_authn/emqx_authn_schema.erl
+++ b/apps/emqx_auth/src/emqx_authn/emqx_authn_schema.erl
@@ -228,7 +228,12 @@ fields("settings") ->
                     importance => ?IMPORTANCE_LOW,
                     default => emqx_auth_cache_schema:default_config()
                 }
-            )}
+            )},
+        {builtin_record_count_refresh_interval,
+            hoconsc:mk(emqx_schema:timeout_duration_ms(), #{
+                default => <<"1h">>,
+                importance => ?IMPORTANCE_HIDDEN
+            })}
     ];
 fields("metrics_status_fields") ->
     [
--- a/apps/emqx_auth/src/emqx_authz/emqx_authz_schema.erl
+++ b/apps/emqx_auth/src/emqx_authz/emqx_authz_schema.erl
@@ -124,7 +124,14 @@ injected_fields(AuthzSchemaMods) ->
     }.
 
 authz_fields() ->
-    sources_fields() ++ node_cache_fields().
+    [
+        {builtin_record_count_refresh_interval,
+            hoconsc:mk(emqx_schema:timeout_duration_ms(), #{
+                default => <<"1h">>,
+                importance => ?IMPORTANCE_HIDDEN
+            })}
+        | sources_fields() ++ node_cache_fields()
+    ].
 
 sources_fields() ->
     AuthzSchemaMods = source_schema_mods(),
--- a/apps/emqx_auth_mnesia/src/emqx_auth_mnesia_bookkeeper.erl
+++ b/apps/emqx_auth_mnesia/src/emqx_auth_mnesia_bookkeeper.erl
@@ -1,0 +1,156 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2025 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_auth_mnesia_bookkeeper).
+
+-behaviour(gen_server).
+
+%% API
+-export([
+    start_link/0,
+
+    tally_authn_now/0,
+    tally_authz_now/0
+]).
+
+%% `gen_server' API
+-export([
+    init/1,
+    terminate/2,
+    handle_continue/2,
+    handle_call/3,
+    handle_cast/2,
+    handle_info/2
+]).
+
+%%------------------------------------------------------------------------------
+%% Type declarations
+%%------------------------------------------------------------------------------
+
+-include("emqx_auth_mnesia_internal.hrl").
+
+-define(authz_tref, authz_tref).
+-define(authn_tref, authn_tref).
+
+%% Calls/casts/infos/continues
+-record(tally_authn, {}).
+-record(tally_authz, {}).
+
+%%------------------------------------------------------------------------------
+%% API
+%%------------------------------------------------------------------------------
+
+start_link() ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, _Opts = #{}, []).
+
+%% For debug/manual ops
+tally_authn_now() ->
+    gen_server:call(?MODULE, #tally_authn{}, infinity).
+
+tally_authz_now() ->
+    gen_server:call(?MODULE, #tally_authz{}, infinity).
+
+%%------------------------------------------------------------------------------
+%% `gen_server' API
+%%------------------------------------------------------------------------------
+
+init(_Opts) ->
+    State = #{
+        ?authz_tref => undefined,
+        ?authn_tref => undefined
+    },
+    {ok, State, {continue, #tally_authn{}}}.
+
+terminate(_Reason, _State) ->
+    ok.
+
+handle_continue(#tally_authn{}, State0) ->
+    State = handle_tally({?AUTHN_NS_TAB, ?AUTHN_NS_COUNT_TAB}, State0),
+    {noreply, State, {continue, #tally_authz{}}};
+handle_continue(#tally_authz{}, State0) ->
+    State = handle_tally({?AUTHZ_NS_TAB, ?AUTHZ_NS_COUNT_TAB}, State0),
+    {noreply, State}.
+
+handle_call(#tally_authz{}, _From, State0) ->
+    State = handle_tally({?AUTHZ_NS_TAB, ?AUTHZ_NS_COUNT_TAB}, State0),
+    {noreply, State};
+handle_call(#tally_authn{}, _From, State0) ->
+    State = handle_tally({?AUTHN_NS_TAB, ?AUTHN_NS_COUNT_TAB}, State0),
+    {noreply, State};
+handle_call(Call, _From, State) ->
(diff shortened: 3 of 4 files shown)


apps/emqx_auth_mnesia/src/emqx_authz_mnesia.erl
```diff
@@ -268,6 +269,16 @@ record_count(Namespace) when is_binary(Namespace) ->
     ],
     ets:select_count(?ACL_NS_TABLE, MS).
 
+-spec record_count_per_namespace() -> #{emqx_config:namespace() => non_neg_integer()}.
+record_count_per_namespace() ->
+    ets:foldl(
```
zmstone: we'll likely need some index for this one.

apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
```diff
@@ -331,6 +333,31 @@ list_users(QueryString0, #{user_group := UserGroup}) ->
...
+            #?NS_TAB{user_id = ?NS_KEY(Namespace, '_', '_'), _ = '_'},
+            [],
+            [true]
+        }
+    ],
+    ets:select_count(?NS_TAB, MS).
+
+-spec record_count_per_namespace() -> #{emqx_config:namespace() => non_neg_integer()}.
+record_count_per_namespace() ->
+    ets:foldl(
```
zmstone: this can be slow.
maybe create a counter index.

apps/emqx_prometheus/src/emqx_prometheus.erl
```diff
@@ -830,6 +847,63 @@ metric_data_ns1(MetricKAtom, NsToMetrics, Mode) ->
...
+        case emqx_mt_state:count_clients(Namespace) of
+            {ok, N} ->
+                Acc#{Namespace => [{emqx_sessions_count, N}]};
+            {error, _} ->
+                %% race?
+                Acc#{Namespace => [{emqx_sessions_count, 0}]}
+        end
+    end,
+    InitAcc = #{},
+    emqx_mt_state:fold_known_nss(FoldFn, InitAcc);
```
zmstone, line 866: maybe simply report table size.
scan is too costly

thalesmg: table size does not fit this.  we need per-ns counts to add the labels.

thalesmg: and note that we scan `?NS_TAB`, not `?COUNTER_TAB`.

apps/emqx_auth_mnesia/src/emqx_authz_mnesia.erl
thalesmg: this is a bit tricky, because we do dirty operations on the table, specially delete...

thalesmg: :checkered_flag:

apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
thalesmg: :checkered_flag:

apps/emqx_prometheus/src/emqx_prometheus.erl
zmstone: yes, but there can be up to 1000 ns, if implicitly created, no limit ?

apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
```diff
@@ -224,26 +230,45 @@ import_users({PasswordType, Filename, FileData}, State, Opts) ->
...
+    IncBucketFn = fun(Bucket, Acc) ->
+        N = maps:get(Bucket, Acc, 0),
+        maps:put(Bucket, N + 1, Acc)
+    end,
+    FoldFn = fun(User, Acc0) ->
+        Return = insert_user(User, Opts),
+        case Return of
+            {success, Ns} ->
+                PerNs0 = maps:get(per_ns, Acc0),
+                PerNs = maps:update_with(Ns, fun(N) -> N + 1 end, 1, PerNs0),
```
savonarola: Nit: this seems to be the same as `IncBucketFn`, m.b. rename `IncBucketFn -> IncKey` and use also here

apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
```diff
@@ -224,26 +230,45 @@ import_users({PasswordType, Filename, FileData}, State, Opts) ->
 do_import_users([], _Opts) ->
     {error, empty_users};
 do_import_users(Users, Opts) ->
+    IncBucketFn = fun(Bucket, Acc) ->
```
savonarola: nit: m.b. make a regular function since it is not a closure

apps/emqx_mt/src/emqx_mt_state.erl
```diff
@@ -273,6 +274,21 @@ The third argument is the number of namespaces to return.
 list_managed_ns_details(LastNs, Limit) ->
     do_list_ns_details(?CONFIG_TAB, LastNs, Limit).
 
+fold_known_nss(Fn, Acc) ->
```
savonarola, line 277: Nit M.b. just provide a `emqx_utils_stream` stream of namespaces

savonarola: Then me may efficiently truncate it if needed and, e.g., introduce a reasonable limit in metrics controller (to address 
https://github.com/emqx/emqx/pull/16346/files#r2565506845)

apps/emqx_auth_mnesia/src/emqx_authz_mnesia.erl
```diff
@@ -252,21 +258,15 @@ list_clientid_rules(Namespace) when is_binary(Namespace) ->
...
-    ],
-    ets:select_count(?ACL_NS_TABLE, MS).
+    try
+        ets:lookup_element(?NS_COUNT_TAB, Namespace, 2, 0)
+    catch
+        error:badarg -> 0
+    end.
+
+-spec record_count_per_namespace() -> #{emqx_config:namespace() => non_neg_integer()}.
+record_count_per_namespace() ->
```
savonarola, line 256: Nit: Maybe also add a limit argument: give counts for the N first namespaces

apps/emqx_prometheus/src/emqx_prometheus.erl
savonarola: As far a I understand we also should avoid making the size of prometheus endpoint http response too large

apps/emqx_auth_mnesia/src/emqx_authz_mnesia.erl
thalesmg: my understanding is that we want to provide metrics for _all_ namespaces for billing.  :thinking:

apps/emqx_prometheus/src/emqx_prometheus.erl
thalesmg: > yes, but there can be up to 1000 ns, if implicitly created, no limit ?

still, the table size is useless if we want to report per namespace...  we do need to go through them, unless we no longer want to report the metrics?

> As far a I understand we also should avoid making the size of prometheus endpoint http response too large

well, it have already increased a lot then after the [previous pr](https://github.com/emqx/emqx/pull/16211)....  :sweat_smile: 

if we no longer want to expose all namespace metrics in prometheus, then we'll need to roll back and then just expose metrics for a single namespace in a separate api.

apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
```diff
@@ -103,6 +108,7 @@ create_tables() ->
         {attributes, record_info(fields, ?NS_TAB)},
         {storage_properties, [{ets, [{read_concurrency, true}]}]}
     ]),
+    ok = emqx_utils_ets:new(?NS_COUNT_TAB, [ordered_set, public]),
```
zmstone: who is the owner of this ets table ?

zmstone: for existing records loaded from dis, is this counter table bootstrapped ?

thalesmg: good catch.  it's currently the application controller, but it should be the application's top supervisor.

https://github.com/emqx/emqx/blob/c8c599147640d01c53a5a5f9775690e60f0de087/apps/emqx_auth_mnesia/src/emqx_auth_mnesia_app.erl#L14-L15

apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
thalesmg: :checkered_flag:

apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
thalesmg: :checkered_flag:

apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
thalesmg: :checkered_flag:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10826)

PR #16356, opened 2025-11-26
test: refactor emqx-boot.bats to pytest test_emqx_boot.py

zmstone:
refactor: use pytest for boot tests


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10841)

PR #16231, opened 2025-11-06
chore: Log cache cleanup only when records are deleted

zhongwencool:
Updated the cleanup function to log cache cleanup events only if records were actually deleted, reducing unnecessary debug logs.

A large number of useless log prints are extremely annoying.

```
2025-11-06T02:11:47.365519+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authz_cache, num_deleted: 0
2025-11-06T02:11:47.365910+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authn_cache, num_deleted: 0
2025-11-06T02:12:47.372490+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authz_cache, num_deleted: 0
2025-11-06T02:12:47.372661+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authn_cache, num_deleted: 0
2025-11-06T02:13:47.373763+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authn_cache, num_deleted: 0
2025-11-06T02:13:47.373763+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authz_cache, num_deleted: 0
2025-11-06T02:14:47.375182+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authz_cache, num_deleted: 0
2025-11-06T02:14:47.375810+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authn_cache, num_deleted: 0
2025-11-06T02:15:47.385230+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authn_cache, num_deleted: 0
2025-11-06T02:15:47.385230+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authz_cache, num_deleted: 0
2025-11-06T02:16:47.387658+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authn_cache, num_deleted: 0
2025-11-06T02:16:47.388297+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authz_cache, num_deleted: 0
2025-11-06T02:17:47.392900+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authn_cache, num_deleted: 0
2025-11-06T02:17:47.393277+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authz_cache, num_deleted: 0
2025-11-06T02:18:47.396505+00:00 [debug] msg: auth_cache_cleanup, name: emqx_authn_cache, num_deleted: 0
[...]

PR #16357, opened 2025-11-26
chore: Log cache cleanup only when records are deleted

zmstone:
Updated the cleanup function to log cache cleanup events only if records were actually deleted, reducing unnecessary debug logs.
port https://github.com/emqx/emqx/pull/16231 to v6


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10840)

PR #16368, opened 2025-11-29
chore: adapt for OTP 28 changes

zmstone:
Release: 6.1.0

### DIFF
--- a/apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_streaming_impl.erl
+++ b/apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_streaming_impl.erl
@@ -462,7 +462,6 @@ run_streaming_action(Batch, ActionResId, _ActionState) ->
         handover
     ).
 
--dialyzer({nowarn_function, get_streaming_hostname/2}).
 get_streaming_hostname(SetupPoolId, SetupPoolState) ->
     #{
         jwt_config := JWTConfig,
--- a/apps/emqx_connector_jwt/src/emqx_connector_jwt.erl
+++ b/apps/emqx_connector_jwt/src/emqx_connector_jwt.erl
@@ -19,7 +19,8 @@
 ]).
 
 -type jwt() :: binary().
--type wrapped_jwk() :: fun(() -> jose_jwk:key()).
+%% Ideally, it should be emqx_secret:t(jwk()) instead of emqx_secret:t(_)
+-type wrapped_jwk() :: emqx_secret:t(_).
 -type jwk() :: jose_jwk:key().
 -type duration() :: non_neg_integer().
 


apps/emqx/src/emqx_ds_schema.erl
```diff
@@ -694,7 +694,7 @@ to_size_limit(In) when is_integer(In), In > 0 ->
 to_size_limit(In) when In =:= infinity; In =:= "infinity"; In =:= <<"infinity">> ->
     {ok, infinity};
 to_size_limit(In) when is_list(In); is_binary(In) ->
-    RE = "^([0-9]+)(.[0-9]{1,3})? *([kKMGTPEZYRQ]i?B)?$",
+    RE = "^([0-9]+)(\\.[0-9]{1,3})? *([kKMGTPEZYRQ]i?B)?$",
```
ieQu1, line 697: Hah, good catch.

zmstone: this needs back-port

apps/emqx_rule_engine/test/emqx_rule_engine_api_2_SUITE.erl
```diff
@@ -2674,8 +2674,7 @@ t_direct_dispatch_empty_string(_Config) ->
             ?assertReceive({deliver, <<"rep">>, _}),
             ok
         end,
-        fun(Trace) ->
-            ?assertMatch([_], ?of_kind("bad_direct_dispatch_resolved_value", Trace)),
+        fun(_Trace) ->
```
ieQu1: Minor nit: `fun(_) -> ok end` can be replaced with `[]`.

apps/emqx/src/emqx_ds_schema.erl
zmstone: https://github.com/emqx/emqx/pull/16388

apps/emqx/src/emqx_ocsp_cache.erl
```diff
@@ -266,10 +266,15 @@ read_server_cert(ServerCertPemPath0) ->
...
-                {ok, ServerCertPem} ->
-                    [{'Certificate', ServerCertDer, _} | _] =
-                        public_key:pem_decode(ServerCertPem),
-                    public_key:der_decode('Certificate', ServerCertDer);
+                {ok, ServerCertPem} when byte_size(ServerCertPem) > 0 ->
+                    case public_key:pem_decode(ServerCertPem) of
+                        [{'Certificate', ServerCertDer, _} | _] ->
+                            public_key:der_decode('Certificate', ServerCertDer);
+                        [] ->
+                            error({empty_pem_file, ServerCertPemPath})
```
thalesmg: Since we're no longer being explosive about the match value, maybe add a clause for things other than a certificate as the first entry?

apps/emqx/src/emqx_ocsp_cache.erl
```diff
@@ -367,10 +374,19 @@ do_http_fetch_and_cache(ListenerID, Params) ->
...
+    IssuerPemPathBin = to_bin(IssuerPemPath),
     IssuerPem =
-        case file:read_file(IssuerPemPath) of
-            {ok, IssuerPem0} -> IssuerPem0;
-            {error, Error0} -> error({bad_issuer_pem_file, Error0})
+        %% Check if IssuerPemPath is PEM content (starts with "-----BEGIN") or a file path
+        case binary:match(IssuerPemPathBin, <<"-----BEGIN">>) of
+            {0, _} ->
+                %% It's PEM content, use it directly
+                IssuerPemPathBin;
```
thalesmg: From the code, `IssuerPemPath` has always been a path and seemingly has not changed.  Why add this?

apps/emqx/src/emqx_ocsp_cache.erl
```diff
@@ -486,7 +502,13 @@ ensure_timer(ListenerID, Message, State, Timeout) ->
     }.
 
 build_ocsp_request(IssuerPem, ServerCert) ->
-    [{'Certificate', IssuerDer, _} | _] = public_key:pem_decode(IssuerPem),
+    IssuerDer =
+        case public_key:pem_decode(IssuerPem) of
+            [{'Certificate', Der, _} | _] ->
+                Der;
+            [] ->
+                error({empty_issuer_pem, "No certificates found in issuer PEM data"})
```
thalesmg: Add a clause for non-certificate first element?

apps/emqx_mix_utils/lib/mix/tasks/emqx.ct.ex
```diff
@@ -459,25 +460,77 @@ defmodule Mix.Tasks.Emqx.Ct do
...
+          yecc_match or leex_match
+        rescue
+          _ -> false
+        end
+
+      _ ->
+        false
+    end
+  rescue
+    _ -> false
```
thalesmg: Maybe all this could be replaced with a simple exclusion list that is removed from the enumeration of all modules?

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -362,12 +362,24 @@ run_loop(
         {ok, Data} ->
             NState = start_idle_timer(State),
             handle_recv({recv_more, Data}, Parent, NState);
-        {select, _SelectInfo} ->
+        {completion, {completion_info, _, _}} ->
```
keynslug: Q: Are those changes because of Dialyzer complaints?

Since:
1. Windows [is not supported](https://github.com/emqx/emqx/blob/483d4bc1ec16fbbaca4f6e803058d7aa6b7b677e/apps/emqx/src/emqx_schema.erl#L639-L640) currently.
2. Impossible to get `select_read` since it needs to be explicitly enabled on the socket.
3. Should be impossible to get partial data because we explicitly request 0 bytes.

Otherwise, extra match on `recv` is pointless since we do not call any other `socket` functions apart from `socket:recv/4` here.

zmstone: I'll have to ask @id , this change was before I took over

id: Yes, these were added to satisfy dialyzer. Please feel free to revert.

zmstone: dialyzer is actually complaining about `{error, {closed, _}}` not possible to match because OTP 28 no longer returns `{error, {closed, Data}}`, just `{error, Reason}`. I have made a fix-up to the big commit with `-dialyzer(nowarn_function, ...)`

apps/emqx_bridge/src/emqx_bridge_v2.erl
```diff
@@ -887,7 +887,7 @@ Throws `{invalid_id, Id}` if the input `Id` has an invalid format.
 
 See [`emqx_resource:parse_channel_id/1`].
 """.
--doc #{equiv => fun emqx_resource:parse_channel_id/1}.
+-doc #{equiv => "emqx_resource:parse_channel_id/1"}.
```
thalesmg: It seems this [should be](https://www.erlang.org/doc/system/documentation#doc-metadata):

```suggestion
-doc #{equiv => emqx_resource:parse_channel_id/1}.
```

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_client.erl
```diff
@@ -301,7 +297,9 @@ parse_jwt_config(ResourceId, #{
         sub => ServiceAccountEmail,
         aud => Aud,
         kid => KId,
-        alg => Alg
+        alg => Alg,
+        % 30 seconds grace period
+        grace_period => 30000
```
thalesmg: 1) Maybe there's no need to do this: default is 5 s, and this one does not involve HTTP requests.
2) Grace period is in [seconds](https://github.com/emqx/emqx/blob/b376c7696567b65aad65b804a6c8bb3284c8b8aa/apps/emqx_connector_jwt/src/emqx_connector_jwt.erl#L147).

apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_lib.erl
```diff
@@ -93,7 +94,9 @@ jwt_config(ResId, Params) ->
         sub => Sub,
         aud => <<"unused">>,
         kid => <<"unused">>,
-        alg => <<"RS256">>
+        alg => <<"RS256">>,
+        table => emqx_connector_jwt,
+        grace_period => 30000
```
thalesmg: 1) Maybe there's no need to do this: default is 5 s, and this one does not involve HTTP requests.
2) Grace period is in [seconds](https://github.com/emqx/emqx/blob/b376c7696567b65aad65b804a6c8bb3284c8b8aa/apps/emqx_connector_jwt/src/emqx_connector_jwt.erl#L147).

apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_streaming_impl.erl
```diff
@@ -462,6 +462,7 @@ run_streaming_action(Batch, ActionResId, _ActionState) ->
         handover
     ).
 
+-dialyzer({nowarn_function, get_streaming_hostname/2}).
```
thalesmg: Curious as to why this is needed now.  :thinking:

apps/emqx_rule_engine/test/emqx_rule_engine_api_2_SUITE.erl
```diff
@@ -2674,10 +2674,7 @@ t_direct_dispatch_empty_string(_Config) ->
             ?assertReceive({deliver, <<"rep">>, _}),
             ok
         end,
-        fun(Trace) ->
-            ?assertMatch([_], ?of_kind("bad_direct_dispatch_resolved_value", Trace)),
-            ok
-        end
```
thalesmg, line 2680: Why remove this assertion?

apps/emqx_machine/test/emqx_machine_tests.erl
```diff
@@ -24,7 +24,7 @@ sorted_reboot_apps_test_() ->
 sorted_reboot_apps_cycle_test() ->
     Apps = [{1, [2]}, {2, [1, 3]}],
     ?assertError(
-        {circular_application_dependency, [[1, 2, 1], [2, 1, 2]]},
+        {circular_application_dependency, [[2, 1, 2], [1, 2, 1]]},
```
thalesmg, line 27: How did this change if the code didn't change?  :exploding_head:

apps/emqx_rule_engine/test/emqx_rule_engine_api_2_SUITE.erl
thalesmg: I see it seems to be the change in the placeholder behavior.

Then it'd be better to keep the assertion and use a bad template that provokes it (something that doesn't resolve to a boolean).

zmstone: because it no longer fails now.
`re:split`, when arg1 is `<<>>`, it no longer returns `[<<>>]`, but `[]`, which allows fallback to `Default`
https://github.com/emqx/emqx/pull/16368/files#diff-cd7d85754cf60324aa8e1c9ff44a2cbb9c6cfc2461ed83a0e7b741e793baedc2R305-R308

ieQu1 (dismissed):

apps/emqx_machine/test/emqx_machine_tests.erl
id: dunno, it was failing with the original order

apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_streaming_impl.erl
zmstone: it's because type `emqx_secret:t/1` is an opaque but type `local_jwt_config()`'s `jwk` field expects otherwise.
but i can just fixed it by using `emqx_secret:t(_)` for `jwk`

zmstone: likely a dialyzer bug.

apps/emqx_bridge/src/emqx_bridge_v2.erl
zmstone: mix does not agree. deleted it.

apps/emqx_machine/test/emqx_machine_tests.erl
zmstone: non-deterministic ordering in the circular report.

apps/emqx_bridge/src/emqx_bridge_v2.erl
thalesmg: or maybe just keep it as a charlist then.

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10906)

PR #16377, opened 2025-12-02
perf(metrics): optimize counter allocation

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14895

Release version: 6.1.0

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_metrics.erl
+++ b/apps/emqx/src/emqx_metrics.erl
@@ -87,14 +87,15 @@
 %% Type declarations
 %%------------------------------------------------------------------------------
 
--opaque metric_idx() :: 1..1024.
+-define(MAX_SIZE, 256).
+-define(RESERVED_IDX, 128).
+
+-opaque metric_idx() :: 1..?MAX_SIZE.
 
 -type metric_name() :: atom() | string() | binary().
 
 -type maybe_namespace() :: emqx_config:maybe_namespace().
 
--define(MAX_SIZE, 1024).
--define(RESERVED_IDX, 512).
 -define(TAB, ?MODULE).
 -define(SERVER, ?MODULE).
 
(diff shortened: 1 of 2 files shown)


thalesmg:
The only places that seem to rely on `emqx_metrics:{create,ensure}_metrics` are:
- https://github.com/emqx/emqx/blob/50ca4971e3517b62a857c1ba9b3b703b1a5303c7/apps/emqx_exhook/src/emqx_exhook_server.erl#L233-L238
- https://github.com/emqx/emqx/blob/50ca4971e3517b62a857c1ba9b3b703b1a5303c7/apps/emqx_auth/src/emqx_authz/emqx_authz.erl#L109-L110

Exhook registers and bumps those metrics, but they are seemingly not read anywhere.  At least explicitly, unless we are currently relying on exporting all `emqx_metrics` to read exhook metrics.  This will also eliminate the `list_to_atom` issue we have there as well.

Perhaps we could drop the `emqx_metrics` usage from `emqx_exhook` application, and move the authz metrics to `emqx_metrics:reserved_idx/1`?

apps/emqx/src/emqx_metrics.erl
```diff
@@ -87,14 +87,15 @@
 %% Type declarations
 %%------------------------------------------------------------------------------
 
--opaque metric_idx() :: 1..1024.
+-define(MAX_SIZE, 256).
```
keynslug: Q: Why not `128`?

thalesmg: just maintaining the same relationship between max and reserved that has existed before.

thalesmg: ~~well, apparently 128 max is too tight~~ if reserved = max, then the places where dynamic metrics are created break:

https://github.com/emqx/emqx/actions/runs/19867134808/job/56934214755?pr=16377#step:5:160

https://github.com/emqx/emqx/pull/16377#issuecomment-3602810769

keynslug:
> Perhaps we could drop the `emqx_metrics` usage from `emqx_exhook` application, and move the authz metrics to `emqx_metrics:reserved_idx/1`?

Interesting, thanks for digging into this. 👍🏼 Yeah, exhook seemingly does something wrong indeed. In general, allocating "resources" from a limited pool in random places sounds risky so I support your general idea.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10863)

PR #16388, opened 2025-12-04
fix(emqx_ds_schema): escape dot in bytes size regex pattern

zmstone:
No significant user experience impact unless someone is actually using a invalid number for size configs for DS.

e.g. `foo_bar_size = 2a0KB` instead of `2.0KB`.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10932)

PR #16393, opened 2025-12-05
fix(clink): anticipate no routerepl mqtt client after suspend

keynslug:
Release version: 6.0.2

## Summary

This can happen because route-syncer is generally unable to handle suspend request before draining operation queue, which can cause one more attempt to do a batch sync.

## PR Checklist

### DIFF
--- a/apps/emqx_cluster_link/src/emqx_cluster_link_routerepl.erl
+++ b/apps/emqx_cluster_link/src/emqx_cluster_link_routerepl.erl
@@ -44,6 +44,26 @@
     terminate/2
 ]).
 
+-export_type([
+    actor/0,
+    routeid/0
+]).
+
+%% Name of the target cluster for route replication:
+-type cluster() :: emqx_cluster_link_schema:cluster().
+
+%% Replication actor:
+%% * `node()` for regular routes.
+%% * `'ps-routes-v1'` for persistent routes.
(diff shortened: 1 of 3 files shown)


apps/emqx_cluster_link/src/emqx_cluster_link_routerepl.erl
```diff
@@ -44,6 +44,26 @@
...
+%% * `node()` for regular routes.
+%% * `'ps-routes-v1'` for persistent routes.
+%% See also: `emqx_cluster_link_extrouter`.
+-type actor() :: atom().
+
+%% Route ID:
+%% A pair of {Topic, Route ID} uniquely identifies a single actor's route.
+%% This is needed for example to disambiguate different local routes that intersect
+%% to the same topic filter for a partical link, according to configured link filters.
+-type routeid() :: binary().
```
zmstone, line 65: Tuple in comment, but binary in code ?

keynslug: Was trying to say that basically `unique_route() :: {topic(), routeid()}`, where the second element is currently a binary string.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11891)

PR #16401, opened 2025-12-08
refactor: delete system_monitor app and sysmon_top config

zmstone:
system_monitor is a hidden feature since long ago, time to remove it.
this also gets rid of the indirect dependency of `supervisor3`


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10929)

PR #16406, opened 2025-12-08
251208 stop using tag prefix

zmstone:
Going forward, we'll tag release without `v` or `e` prefixes.
Tags `e6.0.0` and `e6.0.1` are already pushed with the new format tag (on the same commits) `6.0.0` and `6.0.1`.


qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-10954)

PR #16413, opened 2025-12-09
perf: no need to double monitor MQTT session processes

zmstone:
Fixes [EMQX-14961](https://emqx.atlassian.net/browse/EMQX-14961)

Release version: 6.0.2, 6.1.1

## Summary

This is a follow-up improvement after #14508.
When MQTT session process is already monitored by `emqx_cm`, there is no need for `emqx_broker_helper` to monitor it again.

We cannot get rid of the monitoring completely because there are non-session subscribers:

- Gateway processes
- QUIC streams
- Durable session direct routes

The improvement is NOT expected to be significant, but:

- Less RAM usage as the number of monitor references is cut in half (for RAM sessions)
- Lower CPU usage spike when large amount of sessions expire around the same moment (as 1/2 of the DOWN message handling)

## PR Checklist

[EMQX-14961]: https://emqx.atlassian.net/browse/EMQX-14961?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_session_mem.erl
+++ b/apps/emqx/src/emqx_session_mem.erl
@@ -308,19 +308,7 @@ subscribe(
     Session = #session{clientid = ClientId, subscriptions = Subs}
 ) ->
     IsNew = not maps:is_key(TopicFilter, Subs),
-    %% No need for broker to monitor self() process because
-    %% it's monitored by emqx_cm
-    %% and emqx_cm:clean_down/1 calls emqx_broker_helper:clean_down/1
-    %% when the process is DOWN
-    Monitor =
-        case emqx_cm:is_monitored(self()) of
-            true ->
-                %% gen_tcp/ssl/socket connections
-                no_monitor;
-            false ->
-                %% quic stream
-                monitor
-        end,
+    Monitor = maybe_mointor(),
     case IsNew andalso is_subscriptions_full(Session) of
         false ->
             ok = emqx_broker:subscribe(TopicFilter, ClientId, SubOpts, Monitor),
@@ -328,6 +316,22 @@ subscribe(
             {ok, Session1};
         true ->
             {error, ?RC_QUOTA_EXCEEDED}
+    end.
+
+%% Returns 'no_monitor' if self() is monitored by emqx_cm.
+%% Otherwise `monitor`.
+%%
+%% When 'no_monitor', it means there is no need for broker to monitor self() process,
+%% because emqx_cm:clean_down/1 calls emqx_broker_helper:clean_down/1
+%% when the process is DOWN
+maybe_mointor() ->
+    case emqx_cm:is_monitored(self()) of
+        true ->
+            %% gen_tcp/ssl/socket connections
+            no_monitor;
+        false ->
+            %% quic stream
+            monitor
     end.
 
 is_subscriptions_full(#session{max_subscriptions = infinity}) ->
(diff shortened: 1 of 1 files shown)


apps/emqx/src/emqx_broker_helper.erl
```diff
@@ -268,43 +273,52 @@ collect_messages(Reg, Down) ->
 %% There is no other message sent to this process, so the
 %% 'receive' should not have to scan the mailbox.
 collect_messages(Reg, Down, 0) ->
-    {Reg, Down};
+    {lists:reverse(Reg), lists:reverse(Down)};
```
keynslug, line 276: Q: Why do we need to reverse the lists now?

apps/emqx/src/emqx_broker_helper.erl
```diff
@@ -268,43 +273,52 @@ collect_messages(Reg, Down) ->
...
 handle_register({SubId, SubPid}) ->
-    record_subscription(SubId, SubPid),
-    monitor_subscriber(SubId, SubPid).
+    ok = monitor_subscriber(SubPid),
+    insert_tables(SubId, SubPid).
+
+insert_tables(SubId, SubPid) ->
+    %% Must insert pid to ID first to ensure no leak
+    true = insert_pid_to_id(SubId, SubPid),
+    true = insert_id_to_pid(SubId, SubPid).
```
keynslug, line 297: Nit. I think it's essentially the same piece of information as kept here:
https://github.com/emqx/emqx/blob/2902ecbd349ccc6972925d880aaed2542e304874/apps/emqx/src/emqx_cm.erl#L199

There are few functions that rely on contents of `?SUBID` table through `lookup_subpid/1`, but they are more or less unimportant from what I can see. Dropping this table altogether could probably be a subject of another followup?

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -308,9 +308,22 @@ subscribe(
...
+                %% gen_tcp/ssl/socket connections
+                no_monitor;
+            false ->
+                %% quic stream
+                monitor
+        end,
     case IsNew andalso is_subscriptions_full(Session) of
         false ->
-            ok = emqx_broker:subscribe(TopicFilter, ClientId, SubOpts),
+            ok = emqx_broker:subscribe(TopicFilter, ClientId, SubOpts, Monitor),
```
keynslug, line 314: NIt. There's a call to `subscribe/3` in `resume/2`, perhaps worth to clarify in a comment why it uses `monitor` flavor?

apps/emqx/src/emqx_broker_helper.erl
zmstone: the tests asserts order, but in practice, it's unlikely to matter.

apps/emqx/src/emqx_broker_helper.erl
zmstone: Yes. I'll make follow-up changes to drop the public APIs first (no support for subpid lookup from subid), then we'll see if we can drop the SUBID table.

apps/emqx/src/emqx_session_mem.erl
zmstone: that's what I missed. good catch, i'll fix it.

zmstone: fixed in latest commit.

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11465)

PR #16429, opened 2025-12-12
refactor: delete unused code

zmstone:
emqx_broker:get_subopts was only used in tests.
it is now deleted as a pre-step to decommission the SUBID table


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11468)

PR #16430, opened 2025-12-12
refactor: Remove SubId-to-Pid lookup from subscription checking

zmstone:
This refactoring removes the dependency on `emqx_broker_helper:lookup_subpid`
when checking if a subscriber has subscribed to a given topic.

**Changes:**
- `emqx_broker:subscribed/2` now only accepts process IDs (Pid), removing
  the SubId overload that previously used `lookup_subpid`
- `emqx:subscribed/2` (public API) maintains backward compatibility by
  handling `SubId` lookups via `emqx_cm:lookup_channels` instead of
  `emqx_broker_helper:lookup_subpid`
- Updated `emqx_persistent_session_ds_subs` (test code) to use the public API

**Impact:**
- Internal usage of `emqx_broker:subscribed/2` with SubId was only in tests,
  which have been updated
- The public `emqx:subscribed/2` API remains backward compatible for plugins
  and external code that pass SubId (client ID)
- This change is a step toward eventually removing the `SUBID` ETS table, as
  subscription checking no longer depends on the SubId→Pid mapping

**Note:** The `lookup_subpid` function is still used by `emqx_broker:subscriptions/1`,
more fixes will follow.

### DIFF
--- a/apps/emqx/src/emqx.erl
+++ b/apps/emqx/src/emqx.erl
@@ -167,7 +167,13 @@ subscriptions(SubPid) when is_pid(SubPid) ->
 subscribed(SubPid, Topic) when is_pid(SubPid) ->
     emqx_broker:subscribed(SubPid, iolist_to_binary(Topic));
 subscribed(SubId, Topic) when is_atom(SubId); is_binary(SubId) ->
-    emqx_broker:subscribed(SubId, iolist_to_binary(Topic)).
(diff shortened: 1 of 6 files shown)


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11469)

PR #16451, opened 2025-12-18
chore(mq,streams): update README

savonarola:
Fixes [EMQX-14802](https://emqx.atlassian.net/browse/EMQX-14802)

Release version: 6.1.0

## PR Checklist

[EMQX-14802]: https://emqx.atlassian.net/browse/EMQX-14802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_mq/README.md
+++ b/apps/emqx_mq/README.md
@@ -98,13 +98,13 @@ Message Queues can be capped by time or size individually. The capping is config
 * `max_shard_message_count` is the maximum number of messages in a shard for the Message Queue. The number of shards is configured for all Message Queues in the Durable Storage settings. So the total count limit for the Message Queue is `n_shards * max_shard_message_count`.
 * `max_shard_message_bytes` is the maximum number of bytes in a shard for the Message Queue. Note that this limit is not aware of the replication factor. So the total byte limit for the Message Queue is `n_shards * replication_factor * max_shard_message_bytes`
 
-Also the limit is soft. The limit threshold 10% and is applied to the queues during the GC process. That means that immediately after GC the queue may exceed the limit by up to 10%, but between GC runs the amount of excess data may be larger.
+Also the limit is soft. The limit is applied to the queues during the GC process with 10% threshold. That means that immediately after GC the queue may exceed the limits by up to 10%, but between GC runs the amount of excess data may be larger.
 
 The threshold and some other settings are configured via the `quota` section of the global Message Queue configuration, but they are not publicly exposed currently.
 
 ## Queue auto creation
 
-Message Queues are be automatically created when subscribing to a queue topic `$q/some/topic`. The auto creation is configured via the `auto_create` section of the global Message Queue configuration. By default, the queues are auto created as last-value queues.
+Message Queues are automatically created when subscribing to a queue topic `$q/some/topic`. The auto creation is configured via the `auto_create` section of the global Message Queue configuration. By default, the queues are auto created as last-value queues.
 ```hocon
 ...
 auto_create {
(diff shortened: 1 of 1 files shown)


apps/emqx_mq/README.md
```diff
@@ -83,6 +83,48 @@ The following diagram shows the data flow between the Message Queue components:
...
+    "max_shard_message_count": 1000,
+    "max_shard_message_bytes": "100MB"
+}
+...
+```
+
+* `max_shard_message_count` is the maximum number of messages in a shard for the Message Queue. The number of shards is configured for all Message Queues in the Durable Storage settings. So the total count limit for the Message Queue is `n_shards * max_shard_message_count`.
+* `max_shard_message_bytes` is the maximum number of bytes in a shard for the Message Queue. Note that this limit is not aware of the replication factor. So the total byte limit for the Message Queue is `n_shards * replication_factor * max_shard_message_bytes`
+
+Also the limit is soft. The limit threshold 10% and is applied to the queues during the GC process. That means that immediately after GC the queue may exceed the limit by up to 10%, but between GC runs the amount of excess data may be larger.
```
thalesmg: maybe

```suggestion
Also the limit is soft. The queues limit threshold is increased by 10% during the GC process. That means that immediately after GC the queue may exceed the limit by up to 10%, but between GC runs the amount of excess data may be larger.
```

?

apps/emqx_mq/README.md
```diff
@@ -83,6 +83,48 @@ The following diagram shows the data flow between the Message Queue components:
...
+
+The threshold and some other settings are configured via the `quota` section of the global Message Queue configuration, but they are not publicly exposed currently.
+
+## Queue auto creation
+
+Message Queues are be automatically created when subscribing to a queue topic `$q/some/topic`. The auto creation is configured via the `auto_create` section of the global Message Queue configuration. By default, the queues are auto created as last-value queues.
```
thalesmg:
```suggestion
Message Queues are automatically created when subscribing to a queue topic `$q/some/topic`. The auto creation is configured via the `auto_create` section of the global Message Queue configuration. By default, the queues are auto created as last-value queues.
```

apps/emqx_mq/README.md
```diff
@@ -83,6 +83,48 @@ The following diagram shows the data flow between the Message Queue components:
...
+}
+...
+```
+One may change the auto creation settings to regular queues, or disable the auto creation. Obviously, one may not autocreate queues as regular and last-value at the same time.
+
+## QoS handling
+
+When delivering messages to the subscribers, the QoS level is overridden to 1 to require PUBACK from the client.
+
+With one exception, when publishing messages to the Message Queue, the messages are stores synchronously to the Message Queue database. That means that the publisher will not receive PUBACK from the broker till the message is stored.
```
thalesmg:
```suggestion
With one exception, when publishing messages to the Message Queue, the messages are stored synchronously to the Message Queue database. That means that the publisher will not receive PUBACK from the broker till the message is stored.
```

apps/emqx_mq/README.md
```diff
@@ -83,6 +83,48 @@ The following diagram shows the data flow between the Message Queue components:
...
+
+The exception is when a message is published with QoS0 to a unlimited regular queue. In this case, the message is stored asynchronously via dirty append mechanism and the publisher will receive PUBACK from the broker immediately.
```
thalesmg: Q: should the publisher receive a `PUBACK` if it published a message with QoS 0?  :thinking:

savonarola: Of course no 🙈

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11085)

PR #16466, opened 2025-12-21
251220 selective ct tests

zmstone:
Release version: 5.8, 5.9 and 5.10

## Summary

 Adds dependency analysis for EMQX apps and integrates it into CI to run tests only for changed apps and their transitive dependents, reducing CI runtime and resource usage.

### Changes

- `scripts/gen_deps.escript`: Analyzes BEAM files (via xref) and `-include_lib()` directives to generate "used-by" relationships with transitive closure. Outputs to deps.txt in format: `app: user1 user2` (or `all`/`none`).
- `deps.txt`: Dependency graph file committed to git, used by CI to determine which apps to test.
- `scripts/find-apps.sh`: Added `--base-ref` option to detect changed apps via git diff and lookup transitive users from `deps.txt`. Falls back to all apps if `rebar.config` or `.github/.ci` change.
- CI workflows:
  - `_pr_entrypoint.yaml`: Uses `--base-ref` to generate selective test matrix
  - `static_checks.yaml`: Validates `deps.txt` is up to date
- `scripts/test-find-apps.sh`: Test script validating the functionality

### Benefits

- Faster CI: only tests affected apps
- Reduced resource usage
- Automatic dependency discovery
- CI ensures deps.txt stays current

### Drawbacks

- Incomplete code coverage report


scripts/gen_deps.escript
```diff
@@ -0,0 +1,383 @@
...
+                error ->
+                    Acc
+            end
+        end,
+        sets:new(),
+        Lines
+    ).
+
+parse_include_lib_line(Line) when is_binary(Line) ->
+    % Match: -include_lib("emqx/include/...") or -include_lib("emqx_*/include/...")
```
savonarola: We may also probably obtain structured includes from `abstract_code` chunks of beams if we compile them with debug_info. Thus we may work only with beams.

savonarola:
🚀 IMO the functionality of deps.txt generation got too complex to live in a script, maybe move it to a utils/a separate lib as regular code and add some simple tests

scripts/gen_deps.escript
zmstone: I doubt it's going to be faster than grep.
Or do you think it's more reliable ?

savonarola:
Seems that find apps script runs just a couple of commands and almost all of its code is data structure and string manipulation implemented in shell. Maybe we could consider implementing it in smth like python.

scripts/gen_deps.escript
savonarola: > it's more reliable

Yes, without ad-hoc code parsing.

zmstone:
> Seems that find apps script runs just a couple of commands and almost all of its code is data structure and string manipulation implemented in shell. Maybe we could consider implementing it in smth like python.

I plan to do that when porting this to v6

scripts/gen_deps.escript
zmstone: done.

deps.txt
```diff
@@ -0,0 +1,105 @@
...
+emqx_management: all
+emqx_message_transformation: emqx_prometheus
+emqx_mix_utils: none
+emqx_modules: emqx_node_rebalance emqx_telemetry
+emqx_mongodb: emqx_auth_mongodb emqx_bridge_mongodb
+emqx_mysql: emqx_auth_mysql emqx_bridge_mysql
+emqx_node_rebalance: none
+emqx_opentelemetry: all
+emqx_oracle: all
+emqx_plugins: all
```
savonarola, line 92: Not to fix here, just to note: there is probably cyclic usage of `emqx <-> emqx_plugin` causing `emqx_plugin` (and some other apps) to have`all` value.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11205)

PR #16473, opened 2025-12-23
test: fix flaky tests

zmstone:
The publishes are sent by 3 MQTT clients connected to 3 nodes and collected by 2 subscribers.
Ordering can be quite loose


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11196)

PR #16487, opened 2025-12-26
ci: ensure python venv

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11257)

Issue #16386, opened 2025-12-04
Subscriptions lost after connector reconnection in EMQX bridge rules


oscarsnav-ext-inditex:
### Summary:
Edge brokers lose topic subscriptions after remote cluster node failures; rules stop working until manual restart or connector disable/enable.

### Context:
We have multiple edge EMQX single-node brokers (v6.0.1) connected to a remote continental EMQX 3-node cluster deployed in AKS using the EMQX Operator (v5.8.2).
The edge brokers have a rule that uses an MQTT bridge source as data input (emqx_continental_lc_in).

The source have the following configuration:

``` json
{
    "error": "",
    "name": "emqx_continental_lc_in",
    "status": "connected",
    "type": "mqtt",
    "rules": [
        "from_emqx_continental_lc_in"
    ],
    "node_status": [
        {
            "node": "emqx@55653",
            "status": "connected",
            "status_reason": ""
        }
    ],
    "status_reason": "",
    "connector": "continental_mqtt_connector",
    "created_at": 1764695446767,
    "description": "Connector to gather centralized messages from eMQX",
    "enable": true,
    "last_modified_at": 1764758241257,
    "parameters": {
        "no_local": false,
        "qos": 1,
        "topic": "$queue/lc/55653/in/#"
    },
    "resource_opts": {
        "health_check_interval": "15s",
        "health_check_interval_jitter": "0ms",
        "health_check_timeout": "60s"
    }
}
```
 
The connector have the following configuration

``` json
{
    "name": "continental_mqtt_connector",
    "status": "connected",
    "type": "mqtt",
    "sources": [
        "emqx_continental_lc_in",
        "source_continental_emqx_location"
    ],
    "node_status": [
        {
            "node": "emqx@55653",
            "status": "connected"
        }
    ],
    "actions": [
        "republish_to_continental"
    ],
    "bridge_mode": false,
    "clean_start": true,
    "clientid_prefix": "edge_55653_",
    "connect_timeout": "10s",
    "description": "Connector to external EMQX",
    "enable": true,
    "keepalive": "60s",
    "max_inflight": 32,
    "password": "******",
    "pool_size": 4,
    "proto_ver": "v5",
    "resource_opts": {
        "health_check_interval": "15s",
        "health_check_timeout": "10s",
        "start_after_created": true,
        "start_timeout": "6s"
    },
    "retry_interval": "15s",
    "server": "100.64.239.251:1883",
    "ssl": {
        "ciphers": [],
[...]

PR #16507, opened 2026-01-06
fix(mqtt source): re-subscribe on reconnect

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-15006

Fixes https://github.com/emqx/emqx/issues/16386

Release version: 6.1.1

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files

PR #16508, opened 2026-01-06
fix(mqtt source): re-subscribe on reconnect (r58)

thalesmg:
Port of https://github.com/emqx/emqx/pull/16507

Fixes https://emqx.atlassian.net/browse/EMQX-15006

Fixes https://github.com/emqx/emqx/issues/16386

Release version: 5.8.10, 5.9.3, 5.10.3

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_ingress.erl
+++ b/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_ingress.erl
@@ -152,15 +152,7 @@ on_reconnect(ClientPid, ReconnectContext) ->
         ingress_config := IngressConfig,
         pool_size := PoolSize
     } = ReconnectContext,
-    EmqttInfo = emqtt:info(ClientPid),
-    IsSessionPresent = proplists:get_value(session_present, EmqttInfo) band 1 == 1,
-    Res =
-        case IsSessionPresent of
-            true ->
-                ok;
-            false ->
-                subscribe_channel_helper(ClientPid, WorkerName, IngressConfig, WorkerIdx, PoolSize)
-        end,
+    Res = subscribe_channel_helper(ClientPid, WorkerName, IngressConfig, WorkerIdx, PoolSize),
     ?tp(debug, "mqtt_source_reconnected", #{chan_res_id => _ChanResId}),
     Res.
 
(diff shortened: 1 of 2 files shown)


apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_ingress.erl
```diff
@@ -113,29 +124,75 @@ should_subscribe(RemoteTopic, WorkerIdx, PoolSize, Name, LogWarn) ->
...
+on_reconnect(ClientPid, ReconnectContext) ->
+    #{
+        chan_res_id := _ChanResId,
+        worker_name := WorkerName,
+        worker_idx := WorkerIdx,
+        ingress_config := IngressConfig,
+        pool_size := PoolSize
+    } = ReconnectContext,
+    EmqttInfo = emqtt:info(ClientPid),
+    IsSessionPresent = proplists:get_value(session_present, EmqttInfo) band 1 == 1,
```
zmstone: maybe don't check, e.g. if reconfigured with a different sub topic?

zmstone: and a Q: how to make sure old topics are unsubscribed (after update topics) when session present

changes/ee/fix-16507.en.md
```diff
@@ -0,0 +1 @@
+Previously, when an MQTT Source's Connector recovered afeter losing its connection, topics would not be re-subscribed and the Source would stop working until the Connector itself was restarted.  Now, the Source will re-subscribe upon reconnect if there is no session present.
```
ieQu1:
```suggestion
Previously, when an MQTT Source's Connector recovered after losing its connection, topics would not be re-subscribed and the Source would stop working until the Connector itself was restarted.  Now, the Source will re-subscribe upon reconnect if there is no session present.
```

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_ingress.erl
thalesmg: ~~if we don't check, we might duplicate shared subscriptions, right?~~ apparently I've misinterpreted the spec, they don't get duplicated.

also, I don't quite follow this: 

> if reconfigured with a different sub topic?

this callback is called when the existing emqtt client disconnects and then reconnects, not when reconfiguring the source.

> and a Q: how to make sure old topics are unsubscribed (after update topics) when session present

also doesn't apply to _reconnect_.  reconfiguring involves removing old source (which unsubscribes) and then adding a new source.

thalesmg: related to this, though, is a different problem that apparently has always been present that I noticed: if the remote MQTT broker goes away (indefinitely, or for a really long time) and the user wants to remove a source and/or its connector, unsubscribe fails and hence aborts the source removal...  not sure how we want to handle this.

- if the failure is transient, and sessions are being used, then we want to postpone the removal so that removing the source unsubscribes to the topic.
- if the failure is permanent, we want to remove the source regardless of not being able to unsubscribe.

:thinking: 

(however we want to handle this, the fix should be done in a different PR, IMO)

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11400)

PR #16511, opened 2026-01-07
feat: Support insert data into IoTDB Table model

HJianBo:
Fixes https://emqx.atlassian.net/browse/EMQX-15000

Release version: 5.10.3

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
+++ b/apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
@@ -61,6 +61,7 @@
         pool_size := pos_integer(),
         iotdb_version => atom(),
         protocol_version => atom(),
+        sql := #{dialect := tree | table, database => binary()},
         request => undefined | map(),
         atom() => _
     }.
@@ -70,6 +71,8 @@
         driver := driver(),
         channels := map(),
         iotdb_version := atom(),
+        sql_dialect := tree | table,
+        database => binary(),
         atom() => _
     }.
 
@@ -149,12 +152,11 @@ fields("connection_fields") ->
                     default => ?VSN_1_3_X
                 }
             )},
-        {sql_dialect,
+        {sql,
             mk(
-                hoconsc:union(fun sql_dialect_union_selector/1),
+                hoconsc:union(fun sql_union_selector/1),
                 #{
-                    desc => ?DESC(emqx_bridge_iotdb, "config_sql_dialect"),
-                    default => tree
+                    desc => ?DESC("config_sql_dialect")
                 }
             )},
         {authentication,
@@ -165,13 +167,34 @@ fields("connection_fields") ->
                 }
             )}
     ];
+fields(sql_dialect_tree) ->
+    [
+        {dialect,
+            mk(
+                tree,
+                #{
+                    desc => ?DESC("config_sql_dialect"),
+                    required => true,
+                    default => tree
+                }
+            )}
+    ];
 fields(sql_dialect_table) ->
     [
+        {dialect,
+            mk(
+                table,
+                #{
+                    desc => ?DESC("config_sql_dialect"),
+                    required => true,
+                    default => table
+                }
+            )},
         {database,
             mk(
                 binary(),
                 #{
-                    desc => ?DESC(emqx_bridge_iotdb, "config_database"),
+                    desc => ?DESC("config_database"),
                     required => true,
                     validator => fun emqx_schema:non_empty_string/1
                 }
@@ -200,12 +223,11 @@ fields("config_thrift") ->
                         default => ?PROTOCOL_V3
                     }
                 )},
-            {sql_dialect,
+            {sql,
                 mk(
-                    hoconsc:union(fun sql_dialect_union_selector/1),
+                    hoconsc:union(fun sql_union_selector/1),
                     #{
-                        desc => ?DESC(emqx_bridge_iotdb, "config_sql_dialect"),
-                        default => tree
+                        desc => ?DESC("config_sql_dialect")
                     }
                 )},
             {'zoneId',
@@ -248,14 +270,14 @@ fields("put_" ++ Driver) ->
 fields("get_" ++ Driver) ->
     emqx_bridge_schema:status_fields() ++ fields("post_" ++ Driver).
 
-sql_dialect_union_selector(all_union_members) ->
-    [tree, ref(?MODULE, sql_dialect_table)];
-sql_dialect_union_selector({value, Value}) ->
-    case is_map(Value) of
-        true ->
-            [ref(?MODULE, sql_dialect_table)];
-        _ ->
-            [tree]
+sql_union_selector(all_union_members) ->
+    [ref(?MODULE, sql_dialect_tree), ref(?MODULE, sql_dialect_table)];
+sql_union_selector({value, Value}) ->
+    case maps:get(<<"dialect">>, Value, <<"tree">>) of
+        <<"tree">> ->
+            [ref(?MODULE, sql_dialect_tree)];
+        <<"table">> ->
+            [ref(?MODULE, sql_dialect_table)]
     end.
 
 common_fields(Driver) ->
@@ -274,8 +296,10 @@ desc(authentication) ->
     ?DESC("config_authentication");
 desc(connector_resource_opts) ->
     "Connector resource options";
+desc(sql_dialect_tree) ->
+    ?DESC("config_sql_dialect");
 desc(sql_dialect_table) ->
-    ?DESC(emqx_bridge_iotdb, "config_sql_dialect");
+    ?DESC("config_sql_dialect");
 desc(Struct) when is_list(Struct) ->
     case string:split(Struct, "_") of
         ["config", _] ->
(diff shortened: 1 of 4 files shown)


apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb.erl
```diff
@@ -51,7 +51,9 @@ fields(action_config) ->
     emqx_resource_schema:override(
         emqx_bridge_v2_schema:make_producer_action_schema(
             mk(
-                ref(?MODULE, action_parameters),
+                hoconsc:union([
+                    ref(?MODULE, action_parameters_tree), ref(?MODULE, action_parameters_table)
```
thalesmg: I think we might need a union member selector function to prettify schema errors, otherwise a schema failure might result in a big error message that QA will complain about (e.g., if someone `parameters.data` to an integer instead of the correct type when performing an HTTP request).

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb.erl
```diff
@@ -268,6 +229,117 @@ resource_creation_opts() ->
             )}
     ].
 
+action_parameters(tree) ->
+    [
+        {write_to_table,
+            mk(
+                hoconsc:union([false]),
```
thalesmg:
```suggestion
                false,
```

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb.erl
```diff
@@ -268,6 +229,117 @@ resource_creation_opts() ->
...
+                    validator => fun emqx_schema:non_empty_array/1
+                }
+            )}
+    ] ++
+        action_parameters_common();
+action_parameters(table) ->
+    [
+        {write_to_table,
+            mk(
+                hoconsc:union([true]),
```
thalesmg:
```suggestion
                true,
```

apps/emqx_bridge_iotdb/test/emqx_bridge_iotdb_impl_SUITE.erl
```diff
@@ -479,7 +477,7 @@ t_device_id(Config) ->
 t_template(Config) ->
     %% Create without data  configured
     ?assertMatch(
-        {error, #{reason := empty_array_not_allowed}},
+        {error, #{reason := matched_no_union_member}},
```
thalesmg: This new error indicates the parameter type is no longer being resolved, so the assertion is no longer verifying the empty array behavior itself, it seems.

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
```diff
@@ -642,36 +676,44 @@ on_add_channel(
...
+        WriteToTable = maps:get(write_to_table, Parameter, false),
+        SqlDialect = maps:get(sql_dialect, OldState0, tree),
+        DeviceId =
+            case WriteToTable of
+                true ->
+                    maps:get(table, Parameter, <<>>);
+                false ->
+                    maps:get(device_id, Parameter, <<>>)
+            end,
+        ok ?= check_channel_exists(ChannelId, Channels),
```
thalesmg: This is already enforced by `emqx_resource_manager`.

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
```diff
@@ -642,36 +676,44 @@ on_add_channel(
...
+        ok ?= check_write_to_table(WriteToTable, SqlDialect),
+        ok ?= check_restapi_version(Version),
+        {ok, DeviceIdTemplate} ?= preproc_device_id(WriteToTable, DeviceId),
```
thalesmg: This uses a `maybe` expression, but this function (and a few others) throw errors instead of returning `{error, _}`?

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
```diff
@@ -921,33 +1030,182 @@ normalize_thrift_timeout(Timeouts) ->
...
+
+do_validate_measurements_consistency(Used, More) ->
+    case lists:all(fun(M) -> M =:= Used end, More) of
+        false ->
+            ?SLOG(warning, #{
+                msg => "ignore_inconsistent_column_names_in_batch",
+                hint => "Use the first record's column names to insert into the table",
+                ignored => More,
+                used => Used
+            });
```
thalesmg: - Since this is called for every single element in a (potentially large) batch, maybe throttle this log?
- Adding the action resource id would help users identify this logs not only comes from an IoTDB action, but also which one specifically.

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
```diff
@@ -1000,45 +1281,117 @@ proc_record_data(
...
+            ?SLOG(debug, #{exception => Error, reason => Reason}),
+            {error, {invalid_data, Reason}}
+    end;
+proc_record_data_for_table([], _Msg, MeasurementAcc, ValueAcc) ->
+    {ok, lists:reverse(MeasurementAcc), lists:reverse(ValueAcc)}.
+
+init_connector_state(Config) ->
+    case parse_sql_dialect(maps:get(sql_dialect, Config, tree)) of
+        {table, <<>>} ->
+            throw({failed_to_start_iotdb_bridge, "database is required when sql_dialect is table"});
```
thalesmg: If this is required when `sql_dialect` is table, why not make it required in the schema?

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
```diff
@@ -157,6 +165,17 @@ fields("connection_fields") ->
...
             )}
     ];
+fields(sql_dialect_table) ->
+    [
+        {database,
+            mk(
+                binary(),
+                #{
+                    desc => ?DESC(emqx_bridge_iotdb, "config_database"),
+                    required => false
```
thalesmg: Since we enforce that `database` is non-empty when using this union type, we should enforce it here:


```suggestion
                    required => true,
                    validator => fun emqx_schema:non_empty_string/1
```

hjianbo:
Thanks for the suggestions @thalesmg. I’ve addressed all the comments. Please take a look when you have a moment.

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb.erl
```diff
@@ -268,6 +227,144 @@ resource_creation_opts() ->
...
+            (_) -> false
+        end,
+        emqx_placeholder:preproc_tmpl(TableName)
+    ),
+    case HasVar of
+        true ->
+            {error, <<"Table name cannot contain variables in table model">>};
+        false ->
+            ok
+    end.
```
keynslug: Since it's both typed and documented as just binary string / table name, I find it strange to have this particular validation rule. To me, there are 2 options: either validate according to the IoTDB table name syntax, or don't validate at all, leaving this up to IoTDB to reject invalid requests.

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
```diff
@@ -180,6 +200,14 @@ fields("config_thrift") ->
                         default => ?PROTOCOL_V3
                     }
                 )},
+            {sql_dialect,
```
keynslug: Expressiveness of `..., sql_dialect { database = "foo"}, ...` is somewhat lacking I'd say, especially when documentation says there should be a _table_ dialect. Perhaps better to 1) use `enum([tree, table])` here instead of a union, 2) inline `database` field making it optional, 3) add a validator instead for the whole schema in this module, 4) wire it up into `emqx_bridge_iotdb_connector_info`? Admittedly sounds like a hassle, I'll leave it to @thalesmg to comment.

thalesmg: I think that's a fair assessment.  That would make the schema semantics clearer, I believe.

hjianbo: The purpose of this structure is to facilitate frontend interaction. It allows the UI to conditionally display the Database input field when table is selected and hide it for tree mode. This leverages the standard oneOf syntax based on the Enum, rather than relying on custom logic to check the sql_dialect value.

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb.erl
hjianbo: Thanks @keynslug I'll remove this validation.

apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
hjianbo: The Dashboard has already implemented the logic to hide or show input fields based on this oneOf structure. 

E.g, even without any specific adaptation for the new IoTDB parameters, the Dashboard can already handle the showing and hiding of this field.

Table Model:
image

Tree Model:
image

keynslug: Oh nice! I think having clear configuration schema is as much important, to be honest. I wonder then if we can achieve both by allowing user to choose either `sql {dialect = tree}` or `sql { dialect = table, database = "foo" }`, through a bit more complex (than it is right now) union in the schema?

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11699)

PR #16477, opened 2025-12-23
251222 selective ct tests v6

zmstone:
Release version 6.0, 6.1

## Summary

Adds dependency analysis for EMQX apps and integrates it into CI to run tests only for changed apps and their transitive dependents, reducing CI runtime and resource usage.

### Changes

- `apps/emqx_mix_utils/lib/mix/tasks/emqx.gen_deps.ex`: Elixir Mix task that analyzes `.beam` files (via `xref`) and `-include_lib()` directives to generate "used-by" relationships with transitive closure. Outputs to `deps.txt` in format: `app: user1 user2` (or `all`/`none`).
- `apps/emqx_mix_utils/lib/emqx/mix/deps.ex`: Core dependency analysis module that performs xref queries and include directive parsing.
- `deps.txt`: Dependency graph file committed to git, used by CI to determine which apps to test.
- `scripts/find_apps.py`: Python script that detects changed apps via git diff and looks up transitive users from `deps.txt`. Falls back to all apps if `mix.exs` or `.github/.ci` change. Replaces the previous bash implementation.
- CI workflows:
  - `_pr_entrypoint.yaml`: Uses `--base-ref` to generate selective test matrix based on changed apps
  - `_push-entrypoint.yaml`: Updated to use Python script for matrix generation
  - `static_checks.yaml`: Validates `deps.txt` is up to date by regenerating and comparing
- `scripts/test/test_find_apps.py`: Pytest test suite validating the functionality with comprehensive test cases
- `Makefile`: Added test execution for `emqx_mix_utils` tests in eunit target

### Benefits

- Faster CI: only tests affected apps and their dependents
- Reduced resource usage
- Automatic dependency discovery via `.beam` file analysis
- CI ensures `deps.txt` stays current
- Python implementation is more maintainable than bash
- Comprehensive test coverage with `pytest`

### Drawbacks

- Incomplete code coverage report (only changed apps are tested)

PR #16516, opened 2026-01-07
feat(aggregator): add aggregated delivery finish metrics

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14936

Release version: 6.1.1

## PR Checklist

### DIFF
--- a/apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
+++ b/apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
@@ -30,8 +30,6 @@
     terminate/2
 ]).
 
--export([mk_delivery_finished_callback_for_action/1, on_delivery_finished/2]).
-
 %% For testing only
 -export([where/1]).
 
@@ -481,21 +479,6 @@ invoke_delivery_finished_callback(#st{delivery_finished_callback = {Fn, Args}} =
                 stacktrace => Stacktrace
             }),
             ok
-    end.
-
-mk_delivery_finished_callback_for_action(ActionResId) ->
-    {fun ?MODULE:on_delivery_finished/2, [ActionResId]}.
-
--spec on_delivery_finished(ok | {skipped, _} | {error, _}, emqx_resource:action_resource_id()) ->
-    ok.
-on_delivery_finished(Result, ActionResId) ->
-    case Result of
-        ok ->
-            emqx_resource_metrics:aggregated_upload_success_inc(ActionResId);
-        {skipped, _} ->
-            ok;
-        {error, _} ->
-            emqx_resource_metrics:aggregated_upload_failure_inc(ActionResId)
     end.
 
 enqueue_status_error({upload_failed, Error}, St = #st{errors = QErrors}) ->
--- a/apps/emqx_resource/src/emqx_resource_metrics.erl
+++ b/apps/emqx_resource/src/emqx_resource_metrics.erl
@@ -75,6 +75,8 @@
     aggregated_upload_failure_get/1
 ]).
 
+-export([mk_delivery_finished_callback_for_action/1, on_delivery_finished/2]).
+
 -define(TELEMETRY_PREFIX, emqx, resource).
 
 -spec events() -> [telemetry:event_name()].
@@ -513,6 +515,27 @@ aggregated_upload_failure_inc(ID, Val) ->
 
 aggregated_upload_failure_get(ID) ->
     emqx_metrics_worker:get(?RES_METRICS, ID, 'aggregated_upload.failure').
+
+-doc """
+Helper for aggregated actions to add a callback to bump their metrics.
+""".
+mk_delivery_finished_callback_for_action(ActionResId) ->
+    {fun ?MODULE:on_delivery_finished/2, [ActionResId]}.
+
+-doc """
+Callback handler for aggregated actions to bump their metrics when delivery finishes.
+""".
+-spec on_delivery_finished(ok | {skipped, _} | {error, _}, emqx_resource:action_resource_id()) ->
+    ok.
+on_delivery_finished(Result, ActionResId) ->
+    case Result of
+        ok ->
+            emqx_resource_metrics:aggregated_upload_success_inc(ActionResId);
+        {skipped, _} ->
+            ok;
+        {error, _} ->
+            emqx_resource_metrics:aggregated_upload_failure_inc(ActionResId)
+    end.
 
 inc_actions_executed(#{namespace := Namespace}) ->
     emqx_metrics:inc(Namespace, 'actions.executed');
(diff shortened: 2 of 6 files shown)


apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -456,8 +463,55 @@ handle_delivery_exit(Buffer, Error, St = #st{name = Name}) ->
...
+
+mk_delivery_finished_callback_for_action(ActionResId) ->
+    {?MODULE, on_delivery_finished, [ActionResId]}.
+
+-spec on_delivery_finished(ok | {skipped, _} | {error, _}, emqx_resource:action_resource_id()) ->
+    ok.
+on_delivery_finished(Result, ActionResId) ->
+    case Result of
+        ok ->
+            emqx_resource_metrics:aggregated_upload_success_inc(ActionResId);
```
keynslug: Nit. Ideally I'd avoid adding runtime dependency on `emqx_resource` here. I don't have an idea where to put instead however, but maybe you have? (Perhaps `emqx_resource_metrics` itself?)

apps/emqx_bridge/src/emqx_bridge_v2_api.erl
```diff
@@ -1668,53 +1670,61 @@ format_metrics(#{
...
+        'dropped.resource_stopped' => 0,
+        'matched' => 0,
+        'queuing' => 0,
+        'queuing_bytes' => 0,
+        'retried' => 0,
+        'late_reply' => 0,
+        'failed' => 0,
+        'inflight' => 0,
+        'success' => 0,
+        'aggregated_upload_success' => 0,
```
zmstone: aggregated_upload.success|failure ?

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -456,8 +463,55 @@ handle_delivery_exit(Buffer, Error, St = #st{name = Name}) ->
...
+    ok;
+invoke_delivery_finished_callback(#st{delivery_finished_callback = {M, F, Args}} = St, Result) ->
+    try
+        _ = apply(M, F, [Result | Args]),
+        ok
+    catch
+        Kind:Error:Stacktrace ->
+            ?tp(warning, "aggregated_delivery_finish_callback_exception", #{
+                action => St#st.name,
+                error => {Kind, Error},
```
zmstone:
```suggestion
                exception => Kind,
                cause => Error,
```

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -456,8 +463,55 @@ handle_delivery_exit(Buffer, Error, St = #st{name = Name}) ->
...
+    is_function(Fn, 1)
+->
+    try
+        _ = Fn(Result),
+        ok
+    catch
+        Kind:Error:Stacktrace ->
+            ?tp(warning, "aggregated_delivery_finish_callback_exception", #{
+                action => St#st.name,
+                error => {Kind, Error},
```
zmstone:
```suggestion
                exception => Kind
                cause => Error,
```

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -456,8 +463,55 @@ handle_delivery_exit(Buffer, Error, St = #st{name = Name}) ->
...
+                stacktrace => Stacktrace
+            }),
+            ok
+    end;
+invoke_delivery_finished_callback(#st{delivery_finished_callback = Fn} = St, Result) when
```
zmstone: there is maybe no need for two clause.
supply `{Func, Args}`, and apply like this: `apply(Func, [Result | Args])`.

For exported function `{fun ?MODULE:finished_cb/2, [myarg]}`.
Also makes anonymous (tests?) functions easier.

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
thalesmg: Why is having a runtime dependency on `emqx_resource` a problem here?  

`emqx_resource_metrics` is a module from the same application as `emqx_resource`.  What would be different?  :thinking:

keynslug: Because there's no need to have a dependency. It will make dependency graph more complex than it should be, for example for the purpose of https://github.com/emqx/emqx/pull/16477.

The difference in this case would be in `emqx_bridge_s3_connector` (which already naturally depends on `emqx_resource`).
```
delivery_finished_callback =>
            emqx_resource_metrics:mk_delivery_finished_callback_for_action(ActionResId)
```

thalesmg: `emqx_resource_metrics` already contains the metrics for resources, I think it's already aptly located.  :thinking: 

Since connectors are resources, it's weird to me to impose a restriction on connectors not being able to refer to `emqx_resource`...

keynslug: This application is neither a connector nor resource, that's the point. It's a supporting library.

My proposal is I believe much less involved than c61412d7420d6e59c80878bef93f0264d5862775.

1. Move `-export([mk_delivery_finished_callback_for_action/1, on_delivery_finished/2]).` to `emqx_resource_metrics` module.
2. Change call sites accordingly, in `emqx_bridge_s3_connector` and friends.

apps/emqx_resource/src/emqx_resource_metrics.erl
```diff
@@ -90,7 +96,10 @@ events() ->
             received,
             retried_failed,
             retried_success,
-            success
+            success,
+            %% These live in the `emqx_connector_aggregator` application.
+            aggregated_upload_success,
+            aggregated_upload_failure
```
keynslug: To clarify further (sorry for not being clear enough), this was not about where _metrics_ live, it is about where functions `mk_delivery_finished_callback_for_action/1` and `on_delivery_finished/2` live.

1. Metrics should live here, it is perfectly good place for them.
2. The proposal is to also move these 2 functions here (see https://github.com/emqx/emqx/pull/16516#discussion_r2672160295). This will I believe help to avoid having dependency on either `emqx_resource` or `emqx_telemetry` in `emqx_connector_aggregator` application.

If you still feel this is unclear and/or unreasonable, feel free to revert 3053f881e15a50a3d3ff3621beca42314982e21a.

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
thalesmg: It's still a library _for connectors_, so I still think it shouldn't be a problem.

Anyway, moved those 2 functions to `emqx_resource_metrics`.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11865)

PR #16520, opened 2026-01-08
refactor: move module `emqx_auth_http_utils` to `emqx_auth`

zmstone:
so `emqx_auth_http` is not a dependency for other apps.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11476)

PR #16525, opened 2026-01-08
chore: remove version prefix in shell prompt

zmstone:
Since we no longer use prefix for git tags.
This is only a display change when EMQX is started in console mode or the prompt in remote console. Should not impact any user interface contact.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11479)

PR #16524, opened 2026-01-08
Rule engine function base64 encode/decode support `no_padding` and `urlsafe` options

zmstone:
Fixes [EMQX-15010](https://emqx.atlassian.net/browse/EMQX-15010)

Release version: 6.0.2, 6.1.1

## Summary

now one can pass 'no_padding' or 'urlsafe' or both to base64_encode and base64_decode functions

doc change: https://github.com/emqx/emqx-docs/pull/3323

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

[EMQX-15010]: https://emqx.atlassian.net/browse/EMQX-15010?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #16527, opened 2026-01-08
feat: move base64 encode/decode functions to variform bif

zmstone:
A follow up of PR https://github.com/emqx/emqx/pull/16524.
Move them to varifrom bif functions so it can be used in auth configs.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11481)

PR #16530, opened 2026-01-08
Refactor: const module for `sni_fun`

zmstone:
So far, no hot-upgrade on `release-61` yet, so we can still mutate the const module.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11878)

PR #15489, opened 2025-07-09
fix(schema): sso oidc issuer

qzhuyan:
Fixes #15459 

Release version:

## PR Checklist

- [ ] For internal contributor: there is a jira ticket to track this change
- [ ] The changes are covered with new or existing tests
- [ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [ ] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #16540, opened 2026-01-09
fix(oidc-sso): prevent trailing slash normalization on issuer URL

### LINKED ISSUE #16503: EMQX adds trailing slash to OIDC Issuer URL
### What happened?

When entering a base URL (e.g. `https://auth.example.com`) as an "Issuer URL" in EMQX for OIDC configuration, it automatically appends a trailing slash.

The auth service I am using (Pocket ID) returns the `issuer` without a trailing slash in the `/.well-known/openid-configuration` response. This then causes EMQX to perform strict validation, and fail with an `issuer_mismatch` error, due to the URLs not matching.

For it to use the URL that was entered (`https://auth.example.com`) instead of appending a trailing slash.

- Create a new OIDC configuration from `/#/sso/oidc`
- Enter the issuer URL as `https://auth.example.com`

I noticed this was also mentioned as an issue in https://github.com/emqx/emqx/issues/16438

In that issue, they mentioned they used "Traefik's rewritebody plugin to rewrite the OIDC discovery document, adding a trailing slash to the issuer field for requests from EMQX's IP", however that's not really feasible in my scenario. And this should likely be fixed within EMQX instead.

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
sysdescr  : EMQX Enterprise
version   : 6.0.0
datetime  : 2026-01-05T17:23:50.368731247+00:00
uptime    : 1 days, 2 hours, 38 minutes, 23 seconds
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
cat: /etc/os-release: No such file or directory
$ uname -a
Linux Tatooine 4.4.302+ #86003 SMP Wed Oct 22 13:30:41 CST 2025 x86_64 GNU/Linux synology_apollolake_918+
```

</details>

### Log files

<details>

```log
emqx  | 2026-01-05T17:23:18.704768+00:00 [error] Metadata load failed for issuer https://auth.example.com/. Retrying in 8716 ms. Error Details: {configuration_load_failed,{issuer_mismatch,<<104,116,116,112,115,58,47,47,97,117,116,104,46,118,111,107,101,46,100,101,118>>}}
```

</details>

zmstone:
Fixes #16503 [EMQX-15016](https://emqx.atlassian.net/browse/EMQX-15016)

Release version: 6.0.2, 6.1.1

## Summary

EMQX was automatically adding a trailing slash to OIDC issuer URLs during config save, causing issuer mismatch errors when the provider's discovery document returns the issuer without a trailing slash.

The schema change is not perfectly compatible because previously `http` scheme is automatically added if missing.
For example, if configured with `example.com`, it's converted to `http://example.com`.
I believe using the `url` type was mostly for validation, but not conversion, when it was first introduced in https://github.com/emqx/emqx/pull/15489

## PR Checklist

[EMQX-15016]: https://emqx.atlassian.net/browse/EMQX-15016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc.erl
+++ b/apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc.erl
@@ -12,7 +12,8 @@
 -export([
     namespace/0,
     fields/1,
-    desc/1
+    desc/1,
+    validate_issuer_url/1
 ]).
 
 -export([
@@ -53,11 +54,12 @@ fields(oidc) ->
         [
             {issuer,
                 ?HOCON(
-                    emqx_schema:url(),
+                    binary(),
                     #{
                         desc => ?DESC(issuer),
                         required => true,
-                        example => <<"https://issuer.com">>
+                        example => <<"https://issuer.com">>,
+                        validator => fun ?MODULE:validate_issuer_url/1
                     }
                 )},
             {clientid,
@@ -270,6 +272,16 @@ convert_certs(_Dir, Conf) ->
 %%------------------------------------------------------------------------------
 %% Internal functions
 %%------------------------------------------------------------------------------
+
+validate_issuer_url(Value) ->
+    maybe
+        #{scheme := Scheme, host := _Host} ?= uri_string:parse(Value),
+        true ?= (Scheme =:= <<"http">> orelse Scheme =:= <<"https">>),
+        ok
+    else
+        _ ->
+            throw(invalid_issuer_url)
+    end.
 
 save_jwks_file(Dir, Content) ->
     case filelib:is_file(Content) of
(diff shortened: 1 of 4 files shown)


apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc.erl
```diff
@@ -53,11 +54,12 @@ fields(oidc) ->
         [
             {issuer,
                 ?HOCON(
-                    emqx_schema:url(),
+                    binary(),
```
zmstone, line 57: the `emqx_shema:url()` type has an implicit converter which normalizes uri (append a trailing slash).

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11486)

PR #16545, opened 2026-01-11
260110 allow hash `#` in Erlang cookie

zmstone:
Fixes [EMQX-14424](https://emqx.atlassian.net/browse/EMQX-14424)

Release version: 6.0.3, 6.1.1

## Summary

Fixed `node.cookie` handling of `#` character. Previously, if the cookie contained `#`, only the prefix before `#` would take effect.
For example, if `abc#d` was configured, only `abc` was used as the cookie.

Added validation to reject problematic characters: backslash, single quote, double quote, and space.

### Compatibility Issue (just a corner case)

- If `#` is used in `node.cookie` in older version: re-configure the cookie with truncated value before upgrade.
- If `#` is used in `EMQX_NODE__COOKIE`: truncate the value before starting the new version node.

## PR Checklist

[EMQX-14424]: https://emqx.atlassian.net/browse/EMQX-14424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_schema.erl
+++ b/apps/emqx_conf/src/emqx_conf_schema.erl
@@ -1776,7 +1776,7 @@ validate_tls_stateless_tickets_seed(_) ->
 validate_cookie("") ->
     throw("Cookie must be non-empty string");
 validate_cookie(Cookie) when is_list(Cookie) ->
-    case length(Cookie) >= 255 of
+    case length(Cookie) > 255 of
         true ->
             throw("Cookie cannot be more than 255 bytes");
         false ->
--- a/apps/emqx_conf/test/emqx_conf_schema_tests.erl
+++ b/apps/emqx_conf/test/emqx_conf_schema_tests.erl
@@ -720,7 +720,7 @@ validate_cookie_test_() ->
             )
         end},
         {"cookie too long", fun() ->
-            LongCookie = lists:duplicate(255, $a),
+            LongCookie = lists:duplicate(256, $a),
             Conf = io_lib:format("node {cookie = \"~s\", data_dir = \".\"}", [LongCookie]),
             with_file(
                 "emqx-conf-cookie-long.hocon",


apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -1768,3 +1772,40 @@ validate_tls_stateless_tickets_seed(Seed) when is_binary(Seed) ->
     end;
 validate_tls_stateless_tickets_seed(_) ->
     {error, "tls_stateless_tickets_seed must be a string value"}.
+
+validate_cookie("") ->
+    throw("Cookie must be non-empty string");
+validate_cookie(Cookie) when is_list(Cookie) ->
+    case length(Cookie) >= 255 of
+        true ->
+            throw("Cookie cannot be more than 255 bytes");
```
thalesmg: To make the check consistent with the message:

```suggestion
    case length(Cookie) > 255 of
        true ->
            throw("Cookie cannot be more than 255 bytes");
```

apps/emqx_conf/test/emqx_conf_schema_tests.erl
```diff
@@ -705,6 +705,110 @@ node_role_conf(Role0) ->
...
+                fun() ->
+                    ?assertMatch(
+                        {error, #{reason := "Cookie must be non-empty string"}},
+                        emqx_hocon:load_and_check(emqx_conf_schema, "emqx-conf-cookie-empty.hocon")
+                    )
+                end
+            )
+        end},
+        {"cookie too long", fun() ->
+            LongCookie = lists:duplicate(255, $a),
```
thalesmg:
```suggestion
            LongCookie = lists:duplicate(256, $a),
```

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11884)

PR #16546, opened 2026-01-11
chore: delete appup scripts

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11488)

PR #16456, opened 2025-12-18
feat: support tls 1.3 session resumption (5.10)

zmstone:
Fixes [EMQX-14979](https://emqx.atlassian.net/browse/EMQX-14979)

Release version: 5.10.3

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

[EMQX-14979]: https://emqx.atlassian.net/browse/EMQX-14979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #16547, opened 2026-01-12
refactor(ssl): no tls 1.2 session resumption by default

zmstone:
Fixes [EMQX-15018](https://emqx.atlassian.net/browse/EMQX-15018)

Release version: 6.2.0

## Summary

Prior to 5.10.3 (6.0.2, and 6.1.1), `reuse_sessions` option is not configurable for server due a bug in `emqx_tls_lib` (see commit cfd29240b88). After the fix, it's possbible to turn off.

See PR: [5.10.3 pr #16456](https://github.com/emqx/emqx/pull/16456)  and [6.0.2 pr #16461](https://github.com/emqx/emqx/pull/16461)

This PR changes the default value from schema, so to disable it.

**Reason**

TLS 1.2 session state cache is local to each node, so it's not very useful in clustered environments.
Additionally, the number of session tickets is limited by `session_cache_server_max` (default = 1000).

The cost of enabling session reuse is that each TLS connection will call the gen_server
`ssl_server_session_cache` to track the session ID, even when the client has no intention
to resume later (there is no way for the server to know this beforehand).

EMQX's defaults are tuned for large-scale clusters with large numbers of connections.
We believe the cost to support TLS 1.2 session reuse by default is too high, hence it is disabled by default.

**Note**

OTP can be further optimized to generate a random ID (0-32 bytes according to spec).
Currently there is always a `gen_server:call` to generate session ID even when `reuse_sessions` is set to `false`.

## PR Checklist

[EMQX-15018]: https://emqx.atlassian.net/browse/EMQX-15018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/changes/ee/perf-16547.en.md
+++ b/changes/ee/perf-16547.en.md
@@ -1,0 +1,4 @@
+Disable TLS 1.2 session reuse by default to reduce TLS handshake overhead.
+
+The TLS 1.2 session cache size is limited to 1000 entries, and the cache is local to each node.
+This makes the reuse rate very low, especially when large numbers of connections connect to a large cluster.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12404)

PR #16550, opened 2026-01-12
refactor: do not cache subscribe authz (ACL) result

zmstone:
Fixes [EMQX-14701](https://emqx.atlassian.net/browse/EMQX-14701)

Release version: 6.2.0

## Summary

The per-client authz (ACL) cache is not very useful because 'SUBSCRIBE' requests are mostly sent only once, also, session takeover does not inherit the cache

## PR Checklist

[EMQX-14701]: https://emqx.atlassian.net/browse/EMQX-14701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_authz_cache.erl
+++ b/apps/emqx/src/emqx_authz_cache.erl
@@ -30,6 +30,7 @@
 ]).
 
 -define(KEY(Topic, QoS, Retain), {?MODULE, Topic, QoS, Retain}).
+-define(KEY_PAT, {?MODULE, _, _, _}).
 -type authz_result() :: allow | deny.
 -type system_time() :: integer().
 -type retain_flag() :: 0 | 1.
@@ -174,7 +175,7 @@ map_authz_cache(Fun) ->
 map_authz_cache(Fun, Dict) ->
     [
         Fun(R)
-     || R = {{?MODULE, _T, _QoS, _Retain}, _Authz} <- Dict
+     || R = {?KEY_PAT, _Authz} <- Dict
     ].
 
 foreach_authz_cache(Fun) ->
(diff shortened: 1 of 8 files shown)


apps/emqx/src/emqx_authz_cache.erl
```diff
@@ -293,3 +311,9 @@ if_expired(TTL, CachedAt, Fun) ->
         true -> Fun(true);
         false -> Fun(false)
     end.
+
+retain_flag(true) -> 1;
+retain_flag(false) -> 0.
```
keynslug, line 316: Q. Why do we need this conversion? IIUC both smallints and atoms are one-word immediate terms anyway, and it probably doesn't make any difference when it comes to hashing pdict keys.

apps/emqx/test/emqx_authz_cache_SUITE.erl
```diff
@@ -30,6 +30,8 @@ t_cache_exclude(_) ->
     {ok, _} = emqtt:connect(Client),
     {ok, _, _} = emqtt:subscribe(Client, <<"nocache/+/#">>, 0),
     emqtt:publish(Client, <<"nocache/1">>, <<"{\"x\":1}">>, 0),
+    %% wait for async publish to be processed
+    ct:sleep(100),
```
keynslug, line 34: Nit. Perhaps this?
```suggestion
    {ok, #{}}  = emqtt:publish(Client, <<"nocache/1">>, <<"{\"x\":1}">>, 1),
```

keynslug (approved):
🚀

apps/emqx/src/emqx_authz_cache.erl
zmstone: it makes a mall difference when inspecting the pd; also having a boolean in the key always seem odd to me.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12401)

PR #16551, opened 2026-01-12
feat(streams): handle $stream prefix

savonarola:
Fixes [EMQX-15015](https://emqx.atlassian.net/browse/EMQX-15015)

Release version: 6.1.1

## PR Checklist

[EMQX-15015]: https://emqx.atlassian.net/browse/EMQX-15015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_streams/src/emqx_streams_extsub_handler.erl
+++ b/apps/emqx_streams/src/emqx_streams_extsub_handler.erl
@@ -139,7 +139,7 @@ DS streams are explicity called `DS streams' here.
     ds_stream :: emqx_ds:stream()
 }).
 
--define(START_FROM_USER_PROP, <<"start_from">>).
+-define(START_FROM_USER_PROP, <<"$stream.start-from">>).
 
 %%------------------------------------------------------------------------------------
 %% ExtSub Handler callbacks
@@ -177,7 +177,7 @@ handle_subscribe(
                     {ok, schedule_check_stream_status(Handler)}
             end;
         ?err_unrec(Reason) ->
-            ?tp(error, streams_extsub_handler_subscribe_error, #{
+            ?tp(warning, streams_extsub_handler_subscribe_error, #{
                 reason => Reason,
                 topic_filter => TopicFilter,
                 recoverable => false
--- a/changes/ee/fix-16551.en.md
+++ b/changes/ee/fix-16551.en.md
@@ -1,1 +1,8 @@
-Introduce new subscription API for streams. Now, a client can subscribe to a stream using `$stream` prefix: `$stream/<topic_filter>`. Start time can be specified using `start_from` user subscription property.
+Introduce new subscription API for streams. Now, a client can subscribe to a stream using `$stream` prefix: `$stream/<topic_filter>`. Start time can be specified using `$stream.start-from` user subscription property.
+
+Examples:
+```
+$ mqttx sub -t '$stream/t/#' -up '$stream.start-from: earliest'
+$ mqttx sub -t '$stream/t/#' -up '$stream.start-from: latest'
+$ mqttx sub -t '$stream/t/#' -up '$stream.start-srom: 1768290505406468'
+```
--- a/apps/emqx_mix_utils/lib/emqx/gen_deps/db.ex
+++ b/apps/emqx_mix_utils/lib/emqx/gen_deps/db.ex
@@ -134,7 +134,8 @@ defmodule Emqx.GenDeps.DB do
     Enum.reduce(erl_files ++ hrl_files, [], fn file_path, acc ->
       case find_include_directives(file_path, %{app: app, file: file_path}) do
         {:ok, deps, warn} ->
-          Enum.each(warn, &Mix.shell().info/1)
+          shell = Mix.shell()
+          Enum.each(warn, &shell.info/1)
           deps ++ acc
 
         {:error, error} ->
--- a/apps/emqx_streams/test/emqx_streams_SUITE.erl
+++ b/apps/emqx_streams/test/emqx_streams_SUITE.erl
@@ -141,7 +141,7 @@ t_read_earliest(Config) ->
             emqx_streams_test_utils:emqtt_sub(
                 CSub,
                 [<<"$stream/t/#">>],
-                [{<<"start_from">>, <<"earliest">>}]
+                [{<<"$stream.start-from">>, <<"earliest">>}]
             )
     end,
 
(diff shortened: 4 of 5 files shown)


changes/ee/fix-16551.en.md
```diff
@@ -0,0 +1 @@
+Introduce new subscription API for streams. Now, a client can subscribe to a stream using `$stream` prefix: `$stream/<topic_filter>`. Start time can be specified using `start_from` user subscription property.
```
zmstone: maybe provide a mqttx CLI example.
`-up '$stream.start-from: latest'`

apps/emqx_streams/src/emqx_streams_extsub_handler.erl
```diff
@@ -129,6 +139,8 @@ DS streams are explicity called `DS streams' here.
     ds_stream :: emqx_ds:stream()
 }).
 
+-define(START_FROM_USER_PROP, <<"start_from">>).
```
zmstone: This is our first System defined property ?
I propose to use `$` prefix for the property name (following the MQTT convention).

```suggestion
-define(START_FROM_USER_PROP, <<"$stream.Start-From">>).
```

changes/ee/fix-16551.en.md
savonarola: ✅

apps/emqx_extsub/src/emqx_extsub_handler_registry.erl
```diff
@@ -153,11 +162,11 @@ recreate(
     ByRef = maps:remove(HandlerRef, ByRef0),
     Registry = Registry0#registry{by_ref = ByRef, by_topic_cbm = ByTopicCBM},
     lists:foldl(
```
keynslug: Nit. `maps:fold/3`?

apps/emqx_streams/src/emqx_streams_extsub_handler.erl
```diff
@@ -137,85 +149,63 @@ handle_subscribe(
...
-        {ok, Handler} ->
-            {ok, schedule_check_stream_status(Handler)};
+    case check_stream_subscribe_topic_filter(SubscribeCtx, TopicFilter) of
+        {ok, SubscribeParams} ->
+            %% Topic filter is formally valid, try to subscribe
+            case subscribe(Handler1, SubscribeParams) of
+                {ok, Handler} ->
+                    {ok, schedule_check_stream_status(Handler)};
+                ?err_unrec(Reason) ->
+                    ?tp(error, streams_extsub_handler_subscribe_error, #{
```
keynslug, line 162: Q. Does `topic_filter_already_present` case end up here? If so, it isn't really an _error_, or?

apps/emqx_streams/src/emqx_streams_extsub_handler.erl
```diff
@@ -137,85 +149,63 @@ handle_subscribe(
...
+                        topic_filter => TopicFilter,
+                        recoverable => true
+                    }),
+                    %% Since the topic filter is formally valid, we add it to the unknown topic filters
+                    %% to be checked later.
+                    Handler = add_unknown_stream(Handler1, SubscribeParams),
+                    {ok, schedule_check_stream_status(Handler)}
+            end;
         ?err_unrec(Reason) ->
             ?tp(error, streams_extsub_handler_subscribe_error, #{
```
keynslug: Nit. This is also essentially just a protocol violation?

apps/emqx_streams/src/emqx_streams_extsub_handler.erl
```diff
@@ -845,19 +822,72 @@ check_active_streams_status(
...
+    maybe
+        {ok, StartFrom, TopicFilter} ?= split_topic_filter(TopicFilter0),
+        {ok, #subscribe_params{
+            partition = <<"all">>,
+            start_from = StartFrom,
+            topic_filter = TopicFilter,
+            full_topic_filter = FullTopicFilter
+        }}
+    end;
+check_stream_subscribe_topic_filter(Ctx, <<"$stream/", TopicFilter/binary>> = FullTopicFilter) ->
```
keynslug, line 855: Random idea. Perhaps "lock" the connection into either `$s` or `$stream` protocol flavor on first subscribe? To basically avoid dealing with situations "same stream but different subscribe requests"?

keynslug (commented):
👍🏼

apps/emqx_streams/src/emqx_streams_extsub_handler.erl
savonarola: Yes, mini-protocol violation, like `$s/foobar/t/#` with `foobar` in place of start time.

apps/emqx_streams/src/emqx_streams_extsub_handler.erl
savonarola: `topic_filter_already_present` should not normally happen because extsub prevents sending duplicate topics to the handlers.

apps/emqx_streams/src/emqx_streams_extsub_handler.erl
savonarola: I experimented with different approches and came to the variant where we minimally surprise (I hope) a client.

In session callbacks, we cannot "forbid" subscription to a topic, so a client may subscribe to `$s/123/t/#`, `$s/latest/t/#` and `$stream/t/#` and have all the topics in the session even if we ignored some of them. Also a client in the current model may subsribe to the same stream but different partitions (although this is hidden). So we tend to make _all_ these subscriptions effective, following the logic "one subscribed topic in session == one stream subscription".

To consistently forbid confusing multiple subscriptions, I plan to add an acl/`client.subscribe` hook, to avoid  duplicating topics even get into the session. But it may also happen that we earlier deprecate some ways of subscription.

apps/emqx_extsub/src/emqx_extsub_handler_registry.erl
savonarola: ✅

apps/emqx_streams/src/emqx_streams_extsub_handler.erl
savonarola: Changed to warning, since that is not an EMQX's problem.

apps/emqx_streams/src/emqx_streams_extsub_handler.erl
keynslug: > we cannot "forbid" subscription to a topic, so a client may subscribe to `$s/123/t/#`

Right, it's possible to just ignore such subscribe requests once we know the client knows `$stream` "protocol". Another point being: `$s` is probably going to be phased out of the documentation IIUC, so it might be not that surprising to a client that knows nothing about `$s`.

I see your point however, not trying to argue.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11890)

PR #16554, opened 2026-01-13
style: rm stale comment

thalesmg:


thalesmg:
https://github.com/emqx/emqx/pull/16558

PR #16539, opened 2026-01-09
feat(schema registry): add sparkplugb state for aliases

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14748

Release version: 6.0.2, 6.1.1

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #16558, opened 2026-01-13
refactor(schema registry spb): remove need for ets table

thalesmg:
Follow up to https://github.com/emqx/emqx/pull/16539

Release version: 6.0.2, 6.1.1

## PR Checklist


apps/emqx_schema_registry/src/emqx_schema_registry_spb_gc.erl
```diff
@@ -0,0 +1,94 @@
...
+
+%%------------------------------------------------------------------------------
+%% `gen_server' API
+%%------------------------------------------------------------------------------
+
+init(_Opts) ->
+    State = #{
+        ?clients => #{}
+    },
+    {ok, State}.
```
zmstone: This process is not commonly required (only when spb message with alias received), so it can hibernate by default.
```suggestion
    {ok, State, hibernate}.
```

zmstone (commented):
Maybe using ETS table for registry does not solve the fallback action issue after all.

1. Race condition: If cleanup is done right after disconnected (or killed), async fallback action may not be able to lookup from ETS.
2. Fallback actions might already have the alias: I did not go though the code paths, but "action"s, take inputs from rule engine select results, if decode is already done (alias already mapped) in select results, why load alias mappings again later ?

apps/emqx_schema_registry/src/emqx_schema_registry_spb_state.erl
```diff
@@ -155,8 +170,10 @@ gather_aliases(Metrics) ->
         Metrics
     ).
 
+%% Called in channel process context.
 insert_alias_mapping(Key, Mapping) ->
-    ets:insert(?TAB, {Key, Mapping}).
+    emqx_schema_registry_spb_gc:register_client(self()),
+    ets:insert(?TAB, ?ENTRY(Key, Mapping, self())).
```
zmstone: If ETS is a must:
- Use owner `Pid` as key: this gives fast GC and owner lookup
- A index from the current `Key` to `list(Pid)` (bag) to support make non-owner lookup (if necessary, see my previous comment).

thalesmg:
> Race condition: If cleanup is done right after disconnected (or killed), async fallback action may not be able to lookup from ETS.

good point.  I don't have a good idea to prevent this case: maybe give the cleanup a grace period before running?

> Fallback actions might already have the alias: I did not go though the code paths, but "action"s, take inputs from rule engine select results, if decode is already done (alias already mapped) in select results, why load alias mappings again later ?

true, if the input has already gone through the decoding, then it's already done.  the fallback action might need the mapping if it itself publishes to a `NDATA`/`DDATA` topic: https://github.com/thalesmg/emqx/blob/5cfb77e77492696464a1327b82e87a70fd169d1c/apps/emqx_schema_registry/test/emqx_schema_registry_spb_SUITE.erl#L565-L569

thalesmg:
> maybe give the cleanup a grace period before running?

actually, that might not help: the fallback action might execute quite some time after the client is gone.  :thinking:

apps/emqx_schema_registry/src/emqx_schema_registry_spb_gc.erl
thalesmg: :checkered_flag:

apps/emqx_schema_registry/src/emqx_schema_registry_spb_state.erl
thalesmg: removed ets tables after we refined the requirements further.

thalesmg:
removed ets tables after we refined the requirements further.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11527)

PR #16579, opened 2026-01-16
ci: fix app version check when there are no changes

thalesmg:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11906)

PR #16581, opened 2026-01-16
feat(schema registry spb): introduce helper function to zip property and data sets

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14872

Release version: 6.1.1

## PR Checklist

### DIFF
--- a/apps/emqx_schema_registry/src/emqx_schema_registry_serde.erl
+++ b/apps/emqx_schema_registry/src/emqx_schema_registry_serde.erl
@@ -31,7 +31,7 @@
     rsf_avro_decode/1,
     rsf_schema_encode_and_tag/1,
     rsf_schema_decode_tagged/1,
-    rsf_spb_zip_propsets/1
+    rsf_spb_zip_kvs/1
 ]).
 
 %% Tests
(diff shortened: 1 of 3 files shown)


apps/emqx_schema_registry/src/emqx_schema_registry_serde.erl
```diff
@@ -166,6 +167,35 @@ rsf_schema_decode_tagged([RegistryName, Data | Args]) ->
             error(Reason)
     end.
 
+-doc """
+- Expects a valid, decoded sparkplugb message.
+
+- `properties` (and any nested `PropertySet` values) have their `keys` and `values` fields
+  removed and have a `_kvs` key added, whose value is the values of the two former fields
+  zipped together (_à la_ `maps:from_list(lists:zip(Keys, Values))`).  Values that have
+  the `PropertySet` or `PropertySetList` types are recursively transformed like this.
```
keynslug: Q: Why is extra `_kvs` nesting? A measure to preserve information? To me it sounds perfectly fine to do a lossy transformation in a rule function call, because it's always possible to have something like `select spb_decode(payload) as spb_payload, spb_zip_propsets(spb_payload) as decoded from "${t}"` so that the original is still in the evaluation context.

thalesmg: To account for other potential keys coming from extensions in the schema and clobbering due to column/property name clashes.  

[`PropertySet`](https://github.com/emqx/emqx/blob/d7eb33d508e17db359515b001a63edd80d985b7b/apps/emqx_schema_registry/priv/sparkplug_b.proto#L166), [`Row`](https://github.com/emqx/emqx/blob/d7eb33d508e17db359515b001a63edd80d985b7b/apps/emqx_schema_registry/priv/sparkplug_b.proto#L131) and [`DataSet`](https://github.com/emqx/emqx/blob/d7eb33d508e17db359515b001a63edd80d985b7b/apps/emqx_schema_registry/priv/sparkplug_b.proto#L138) have those, and I don't know when they might crop up.

Also, `DataSet` has a couple [other](https://github.com/emqx/emqx/blob/d7eb33d508e17db359515b001a63edd80d985b7b/apps/emqx_schema_registry/priv/sparkplug_b.proto#L134) [fields](https://github.com/emqx/emqx/blob/d7eb33d508e17db359515b001a63edd80d985b7b/apps/emqx_schema_registry/priv/sparkplug_b.proto#L136) that could be clobbered if the property/column name clashes with it.

apps/emqx_schema_registry/src/emqx_schema_registry_serde.erl
```diff
@@ -30,7 +30,8 @@
     rsf_avro_encode/1,
     rsf_avro_decode/1,
     rsf_schema_encode_and_tag/1,
-    rsf_schema_decode_tagged/1
+    rsf_schema_decode_tagged/1,
+    rsf_spb_zip_propsets/1
```
zmstone: since it works for both propertyset and dataset, make sense to name it `spb_zip_kvs`?

thalesmg: :checkered_flag:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11907)

PR #15917, opened 2025-09-13
feat: add 'remote' as a shortcut for 'remote_console'

zmstone:
And use 'remote' in usage info.

PR #16601, opened 2026-01-21
feat: add a remote_console alias 'remote'

zmstone:
Port https://github.com/emqx/emqx/pull/15917 to v5, but no change in the usage print (i.e. hidden).


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11826)

PR #16597, opened 2026-01-20
fix(auth): fix SQL query rendering for mysql, postgresql authn/authz.  

savonarola:
Fixes [EMQX-15032](https://emqx.atlassian.net/browse/EMQX-15032), #16437

Release version: 6.2.0

## Summary

The PR fixes issues with SQL query rendering in authn/authz in a situation when they contain invalid `${foobar}` or quoted  `${$}{foobar}` varibles. The issues were caused on two levels:

1. pre-rendering invalid variables`${foobar}` as quoted variables for sql backends. See [fix(auth): correctly prepare template of sql arguments](https://github.com/emqx/emqx/pull/16597/commits/3b98d8f99dc5f7f6b06383fe0386eda7dee39239) commit. The prerendering happens as follows:
   * We render query `select ${username} = ${bar}` into a statement with placeholders and row template: `select ? = ?` + `["${username}", "${bar}"]`
   * Then, in row template we quote unallowed variables`["${username}", "${bar}"] -> ["${username}", "${$}{bar}"]`.
   * When making a query, we render the row template and pass it to the query with placeholders: `select ? = ?` is called with `["actual username", "${bar}"]`.
  
An issue was that when converting `["${username}", "${bar}"] -> ["${username}", "${$}{bar}"]` we treated row template as iolist-represented string, not a list, which surprisingly worked, but only for allowed variables.

2.  The `emqx_posgresql` and `emqx_mysql` connectors _additionally parse_ passed sql queries (`select ${username} = ${bar}` -> `select ? = ?` ). This resulted into invalid queries when there were quoted vars: `select ${username} = '${$}{username}'` was translated into `select ? = '${username}'` by auth and then into `select ? = '?'` by connector. For auth, we introduce minimal mysql and postgresql connectors with minimal logic to fix this.
 
We also
* improve schema helpers and reduce boilerplate
* break dependency of auth from bridges

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

[...]

PR #16605, opened 2026-01-21
chore: remove redundant logging

savonarola:
Follow-up to https://github.com/emqx/emqx/pull/16597


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12409)

PR #16607, opened 2026-01-21
feat(retainer): support iteration resumption after session takeover/resume

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14838

Release version:

## PR Checklist


apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state.erl
```diff
@@ -69,6 +69,7 @@
     fold_awaiting_rel/3,
     n_awaiting_rel/1
 ]).
+-export([get_runtime_data/1, put_runtime_data/2]).
```
savonarola, line 72: Nit: isn't there a confusion in naming 🤔 If the data is runtime, then why do we persist it?

apps/emqx/src/emqx_channel.erl
```diff
@@ -1505,13 +1505,31 @@ handle_call(
...
+    }
+) ->
+    %% RPC Called by `emqx_cm_proto_v4`, only when using in-memory sessions
+    %% NOTE
+    %% Ensure channel has enough time left to react to takeover end call. At the same
+    %% time ensure that channel dies off reasonably quickly if no call will arrive.
+    Interval = interval(expire_takeover, Channel),
+    NChannel = reset_timer(expire_session, Interval, Channel),
+    ok = emqx_cm:unregister_channel(ClientId),
+    PreTerminateCtx = #{},
```
savonarola, line 1530: Maybe pass some channel identifiers into the context? Otherwise we imply using process dictionary for handling this callback.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1505,13 +1505,31 @@ handle_call(
...
+    PreTerminateState = emqx_hooks:run_fold('session.pre_terminate', [PreTerminateCtx], #{}),
```
savonarola, line 1531: Nit: lmo it's a bit confusing that `session.*` hook called directly from channel, not from session module.

savonarola: Also it more looks like `pre_takeover', because the new 'session.pre_terminate' is not called on session terminate.

savonarola:
Maybe we could implement that a bit less invasively (without adding a new protocol, new takeover functions, and without new uses of pd for indirect argument passing 🤔) 

Currently, when a user subscribes to a topic, session places `sub_props` into `SubOpts` of subscriptions map `#{TopicFilter => SubOpts}`. These `sub_props` travel with the session to the new channel during the takeover and the subscriptions get resumed with the initial `sub_props` sent by the client. We may use the same mechanism for external subscriptions, e.g.

* Introduce a single extsub callback, like
  ```erlang
  -callback to_subopt(state(), emqx_extsub_types:topic_filter()) :: ignore | {ok, binary()}
  ```
* Inroduce a fold hook `session.save_subopts` or something which is called by session with `TopicFilter, SubOpts` arguments;
* Before takeover or non-final termination (persisting), `emqx_session` calls the hook for the topics.
* `extsub` subscribes to `session.save_subopts` and populates SubOpts for topic filters handled externally with `extsub => #{Mod => ToSubOptsResult}`.
* These enriched subopts travel or get restored on session resumption.
* On restoration part, we do not need any additional hooks because `handle_subscribe` already receives subotps. We only need hide other module's states (`extsub => #{Mod => ToSubOptsResult}` --> `extsub => ToSubOptsResult`).
* Extsub handler sees that in subopts there is a serialized subopt state and resumes instead of starting from clean.

apps/emqx/src/emqx_channel.erl
thalesmg: funnily enough, I started by calling it `pre_takeover`, but renamed it to `pre_terminate`, kinda prompted by the need of calling it at different times depending on session impl.

thalesmg:
@savonarola very interesting idea.  It does sound more elegant, indeed.  I'll experiment and see how it turns out, thank you for the insight.

thalesmg:
https://github.com/emqx/emqx/pull/16637

PR #16610, opened 2026-01-22
fix(coap): RFC-compliant parsing, observe sequencing, and tests

zhongwencool:
- Enforce RFC 7252/7959/7641 rules in `emqx_coap_frame` (token length, option parsing/encoding, block sizes, reserved values),
  and classify format vs request errors.
  - `emqx_coap_channel`/`emqx_coap_proxy_conn` now handle per-message parse errors (Reset or 4.xx) without dropping listeners;
  connection mode rejects missing token/clientid with 4.00.
  - Observe sequencing uses monotonic 24-bit values and updates on retransmission (`emqx_coap_observe_res`, `emqx_coap_session`,
  `emqx_coap_transport`).
  - CoAP transport/publish tweaks (initial retry interval, observe resend updates, expiry interval handling).
  - Tests: add CoAP CT suites/helpers, reduce mocks; trim/reshape LwM2M CT cases.

image

Fixes <issue-or-jira-number>

Release version: 6.1.1

## PR Checklist

### DIFF
--- a/apps/emqx_gateway_coap/test/emqx_coap_SUITE.erl
+++ b/apps/emqx_gateway_coap/test/emqx_coap_SUITE.erl
@@ -1110,7 +1110,14 @@ disconnection(Channel, Token, ShortenParamName) ->
     },
     URI = compose_uri(Prefix, Queries, ShortenParamName),
     Req = make_req(delete),
-    {ok, deleted, _} = do_request(Channel, URI, Req).
+    case do_request(Channel, URI, Req) of
+        {ok, deleted, _} ->
+            ok;
+        {error, timeout} ->
+            %% Use a fresh socket if the channel got closed during gateway updates.
+            {ok, deleted, _} = er_coap_client:request(delete, URI),
+            ok
+    end.
 
 observe(Channel, Token, Observe) ->
     observe(Channel, Token, Observe, false).
--- a/apps/emqx_gateway_coap/test/emqx_coap_gateway_SUITE.erl
+++ b/apps/emqx_gateway_coap/test/emqx_coap_gateway_SUITE.erl
@@ -21,29 +21,35 @@ t_schema_and_gateway_paths(_) ->
     _ = emqx_coap_schema:namespace(),
     _ = emqx_coap_schema:desc(other),
     ok = meck:new(emqx_gateway_utils, [passthrough, no_history, no_link]),
-    ok = meck:expect(emqx_gateway_utils, normalize_config, fun(_) -> [] end),
+    ok = meck:new(emqx_gateway_utils_conf, [passthrough, no_history, no_link]),
+    ok = meck:expect(
+        emqx_gateway_utils_conf,
+        to_rt_listener_configs,
+        fun(_, _, _, _) -> [#{original_listener_config => #{}}] end
+    ),
     ok = meck:expect(
         emqx_gateway_utils,
         start_listeners,
-        fun(_, _, _, _) -> {error, {bad_listener, #{}}} end
+        fun(_) -> {error, {bad_listener, #{original_listener_config => #{}}}} end
     ),
     ?assertThrow(
         {badconf, _},
         emqx_gateway_coap:on_gateway_load(
-            #{name => coap, config => #{}},
+            #{name => coap, config => #{connection_required => false}},
             #{}
         )
     ),
     ok = meck:expect(
         emqx_gateway_utils,
-        update_gateway,
-        fun(_, _, _, _, _) -> erlang:error(update_failed) end
+        update_gateway_listeners,
+        fun(_, _, _) -> {error, update_failed} end
     ),
     {error, update_failed} =
         emqx_gateway_coap:on_gateway_update(
-            #{},
-            #{name => coap, config => #{}},
+            #{connection_required => false},
+            #{name => coap, config => #{connection_required => false}},
             #{ctx => #{gwname => coap, cm => self()}}
         ),
+    meck:unload(emqx_gateway_utils_conf),
     meck:unload(emqx_gateway_utils),
     ok.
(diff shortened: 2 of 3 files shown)


apps/emqx_gateway_coap/test/emqx_coap_SUITE.erl
```diff
@@ -663,6 +991,96 @@ t_connectionless_pubsub(_) ->
     do(Fun),
     update_coap_with_connection_mode(true).
 
+t_frame_encode_decode_options(Config) ->
+    emqx_coap_protocol_SUITE:t_frame_encode_decode_options(Config).
```
savonarola: The target methods (`emqx_coap_protocol_SUITE:Method/1`) are context-free and ignore `Config`, so we just call the same tests once more here, why? 🤔

apps/emqx_gateway_coap/test/emqx_coap_protocol_SUITE.erl
```diff
@@ -0,0 +1,1077 @@
...
+
+-compile(export_all).
+-compile(nowarn_export_all).
+
+-include("emqx_coap.hrl").
+-include_lib("eunit/include/eunit.hrl").
+-include_lib("common_test/include/ct.hrl").
+-include_lib("emqx/include/emqx.hrl").
+
+-define(CONF_DEFAULT, <<
```
savonarola: nit: maybe use `~b"""..."""` literals

apps/emqx_gateway_coap/src/emqx_coap_channel.erl
```diff
@@ -443,32 +413,18 @@ check_token(
...
     end.
-
-enrich_conninfo(
-    {Queries, _Msg},
-    Channel = #channel{
-        keepalive = KeepAlive,
-        conninfo = ConnInfo
-    }
-) ->
+enrich_conninfo({Queries, _Msg}, Channel) ->
```
savonarola, line 447: nit: maybe highlight the changes that are not just syntactical ones, its challenging to find out what changed except the syntax.

apps/emqx_gateway_coap/test/emqx_coap_protocol_SUITE.erl
```diff
@@ -0,0 +1,1077 @@
...
+    ok = meck:new(emqx_gateway_ctx, [passthrough]),
+    ok = meck:new(emqx_hooks, [passthrough]),
+    ok = meck:expect(
+        emqx_gateway_ctx,
+        authenticate,
+        fun(_, ClientInfo) -> {ok, ClientInfo#{auth_expire_at => undefined}} end
+    ),
+    ok = meck:expect(emqx_gateway_ctx, open_session, fun(_, _, _, _, _, _) -> {ok, #{}} end),
+    ok = meck:expect(emqx_gateway_ctx, connection_expire_interval, fun(_, _) -> 1 end),
+    ok = meck:expect(emqx_hooks, run, fun(_, _) -> ok end),
```
savonarola: Curious: why do we need to stub hooks? `authentication` is stubbed already.

apps/emqx_gateway_coap/test/emqx_coap_SUITE.erl
zhongwencool: remove...

apps/emqx_gateway_coap/test/emqx_coap_protocol_SUITE.erl
zhongwencool: done

apps/emqx_gateway_coap/src/emqx_coap_observe_res.erl
```diff
@@ -84,25 +87,33 @@ subscriptions(Manager) ->
...
+    Res#{seq_id := max(Now, SeqId + 1)}.
+
+%% RFC 7641 Section 4.4: derive Observe value from a local timestamp.
+current_value() ->
+    observe_value(current_seq()).
+
+observe_value(SeqId) ->
+    SeqId band ?MAX_SEQ_ID.
+
+current_seq() ->
```
zhongwencool, line 118: [4.4](https://datatracker.ietf.org/doc/html/rfc7641#section-4.4).  Reordering

   Because messages can get reordered, the client needs a way to
   determine if a notification arrived later than a newer notification.
   For this purpose, the server MUST set the value of the Observe Option
   of each notification it sends to the 24 least significant bits of a
   strictly increasing sequence number.  The sequence number MAY start
   at any value and MUST NOT increase so fast that it increases by more
   than 2^23 within less than 256 seconds.

   The sequence number selected for a notification MUST be greater than
   that of any preceding notification sent to the same client with the
   same token for the same resource.  The value of the Observe Option
   MUST be current at the time of transmission; if a notification is
   retransmitted, the server MUST update the value of the option to the
   sequence number that is current at that time before retransmission.

   Implementation Note:  A simple implementation that satisfies the
      requirements is to obtain a timestamp from a local clock.  The
      sequence number then is the timestamp in ticks, where 1 tick =
      (256 seconds)/(2^23) = 30.52 microseconds.  It is not necessary
      that the clock reflects the current time/date.

      Another valid implementation is to store a 24-bit unsigned integer
      variable per resource and increment this variable each time the
      resource undergoes a change of state (provided that the resource
      changes its state less than 2^23 times in the first 256 seconds
      after every state change).  This removes the need to update the
      value of the Observe Option on retransmission when the resource
      state did not change.

   Design Note:  The choice of a 24-bit option value and a time span of
      256 seconds theoretically allows for a notification rate of up to
      65536 notifications per second.  Constrained nodes often have
      rather imprecise clocks, though, and inaccuracies of the client
      and server side may cancel out or add in effect.  Therefore, the
      maximum notification rate is reduced to 32768 notifications per
      second.  This is still well beyond the highest known design




Hartke                       Standards Track                   [Page 16]

[RFC 7641](https://datatracker.ietf.org/doc/html/rfc7641)               Observing Resources in CoAP        September 2015


      objective of around 1 kHz (most CoAP applications will be several
      orders of magnitude below that) but allows total clock
      inaccuracies of up to -50/+100%.

savonarola (dismissed):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-11934)

PR #16614, opened 2026-01-22
Fix various errors related to durable sessions and storage

ieQu1:
Fixes [EMQX-15052](https://emqx.atlassian.net/browse/EMQX-15052)

Release version: 6.1.1

## Summary

- [x] emqx_ds_storage_skipstream_lts_v2: fix boundary condition in scan_stream. Make sure messages with the timestamp that was used to create iterator with is included in the beam sent to the subscriber.
- [x] emqx_channel: fix serialization of session state access during channel shutdown
- [x] emqx_ds_pmap: improve optional access sequence validation
- [x] emqx_ds_builtin_raft: fix potential inconsistencies related to the storage layer schema
- [x] emqx_ds_builtin_local: fix configuration update
- [x] emqx_ds_builtin_raft: parallel stopping of DBs and shards
- [x] emqx_ds_builtin_local: parallel stopping of DBs and shards
- [x] emqx_ds_builtin_raft: async stopping of OTX leader process on membership change. Link ra leader and OTX process.

## PR Checklist

[EMQX-15052]: https://emqx.atlassian.net/browse/EMQX-15052?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_liveness.erl
+++ b/apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_liveness.erl
@@ -66,7 +66,8 @@ start_link(DB) ->
 
 -spec notify_shard_up(emqx_ds:db(), emqx_ds:shard()) -> ok.
 notify_shard_up(DB, Shard) ->
-    emqx_ds_builtin_raft_liveness_proto_v1:multicast_shard_up(nodes(), DB, Shard, -1).
+    Nodes = emqx:cluster_nodes(running),
+    emqx_ds_builtin_raft_liveness_proto_v1:multicast_shard_up(Nodes, DB, Shard, -1).
 
 %%================================================================================
 %% Internal exports


apps/emqx_streams/src/emqx_streams_message_db.erl
```diff
@@ -620,7 +620,9 @@ do_delete(DB, Topics, Retries, Slabs0, _Errors0) ->
                 db => DB,
                 shard => Shard,
                 generation => Generation,
-                sync => false
+                sync => false,
+                retries => ?STREAMS_MESSAGE_DB_DELETE_RETRY,
+                retry_interval => ?STREAMS_MESSAGE_DB_DELETE_RETRY_DELAY
```
ieQu1, line 625: Note: async transaction can return error tuple if transaction fails at pre-commit.

apps/emqx_durable_storage/src/emqx_ds.erl
```diff
@@ -821,18 +837,47 @@ update_db_config(DB, Patch) ->
...
+        ),
+    case maps:get(errors, Opts, report) of
+        report ->
+            Result;
+        crash ->
+            case Result of
+                {Slabs, []} ->
+                    Slabs;
+                {_Slabs, Errors} ->
+                    exit({list_slabs, DB, Errors})
```
keynslug: Nit. `crash` is interpreted slightly differently across the module, it's `erlang:exit/1` here and `erlang:error/1` in `fold_topic/4`.

apps/emqx_durable_storage/include/emqx_ds_pmap.hrl
```diff
@@ -34,14 +34,20 @@
...
+-define(pmap_id, '_pmap_id').
+
+%% Uncomment to enable checking of sequential access.
+%% -define(CHECK_SEQNO, true).
 
 %% Enable this flag if you suspect some code breaks the sequence:
 -ifndef(CHECK_SEQNO).
 -define(set_dirty, ?collection_dirty := true).
 -define(unset_dirty, ?collection_dirty := false).
+-define(new_pmap_collection, ?pmap_id => 0).
```
keynslug, line 46: Q: Why this field is still initialized even if `CHECK_SEQNO` is undefined?

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -475,28 +530,31 @@ init({ShardId, Options}) ->
...
+                cf_refs = CFRefs
+            },
+            {ok, S};
+        Schema ->
+            Shard = open_shard(ShardId, DB, CFRefs, Schema),
+            CurrentGenId = maps:get(current_generation, Schema),
+            S = #s{
+                shard_id = ShardId,
+                db = DB,
+                db_opts = filter_layout_db_opts(Options),
```
keynslug, line 549: Nit. This call nowadays always produce empty map, perhaps time to drop the field and the corresponding logic?

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -767,15 +850,42 @@ open_generation(ShardId, DB, CFRefs, GenId, GenSchema) ->
...
+            db = DB,
+            db_opts = filter_layout_db_opts(Options),
+            cf_refs = CFRefs,
+            cf_need_flush = CurrentGenId,
+            schema = Schema,
+            shard = Shard
+        },
+        commit_metadata(S),
+        {ok, S}
+    end.
```
keynslug, line 874: Q: Does it make sense to unify it with the portion of logic in `init/1` to avoid duplication and thus the need to keep them identical?

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_machine.erl
```diff
@@ -295,19 +395,23 @@ apply(
             since => Since
         }
     ),
-    Result = emqx_ds_storage_layer:add_generation(DBShard, Since),
+    Result =
+        case Vsn of
+            0 -> emqx_ds_storage_layer:add_generation(DBShard, Since);
+            1 -> emqx_ds_storage_layer:add_generation(DBShard, Since, Prototype)
+        end,
```
keynslug: Q: No need for a case expression? Or is it still possible to see `v0` log entries here? Brief reading of `ra` implementation tentatively suggests that it's not.

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft.erl
```diff
@@ -902,6 +906,22 @@ do_new_kv_tx_ctx_v1(DB, Shard, Generation, Options) ->
...
+""".
+-spec propagate_schema(emqx_ds:db(), emqx_ds:shard(), ra:server_id()) -> ok | emqx_ds:error(_).
+propagate_schema(DB, Shard, Leader) ->
+    maybe
+        %% v0 used a different mechanism for schema updates, we need
+        %% at least v1:
+        ok ?= wait_for_upgrade(DB, Shard, Leader, 1),
+        SiteSchema = emqx_dsch:get_db_schema(DB),
+        Command = emqx_ds_builtin_raft_machine:update_schema(SiteSchema, emqx_ds:timestamp_us()),
+        Result = ra_command(DB, Shard, Command, 5),
```
keynslug: Wouldn't essentially the same outcome be achieved by emitting `{append, update_schema(...)}` Ra effect inside `apply(_RaftMeta, {machine_version, 0, 1}, ...)` in `emqx_ds_builtin_raft_machine`? But without the need to orchestrate it like that on this level. AFAICT the promise is good enough: the entry should be appended to the log by the leader once (at least), right after `{machine_version, 0, 1}` log entry.

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_machine.erl
```diff
@@ -237,8 +271,56 @@ init(#{db := DB, shard := Shard, schema := Schema}) ->
...
+    ?tp(
+        debug,
+        ds_ra_update_schema,
+        #{
+            shard => DBShard,
+            schema => Schema
+        }
+    ),
+    State = #{latest := Latest} = safe_update_latest(NewLatest, State0#{schema := Schema}),
+    ok = emqx_ds_storage_layer:ensure_schema(DBShard, Schema, Latest),
```
keynslug, line 315: Call to `emqx_ds_builtin_raft:update_shards_schema/5` seems to eventually end up here, but `emqx_ds_storage_layer:ensure_schema/3` is a no-op if the schema already in place on the storage layer. Yet, Ra machine state will contain the new schema, and it would only apply if the storage layer is wiped clean. Is this intentional / am I missing something?

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_aux.erl
```diff
@@ -0,0 +1,249 @@
...
+-define(stopped, stopped).
+-record(starting, {starter :: pid(), retry = 0 :: non_neg_integer()}).
+-record(running, {pid :: pid()}).
+-record(stopping, {pid :: pid(), stopper :: pid()}).
+
+-type otx_state() :: ?stopped | #starting{} | #running{} | #stopping{}.
+
+%% AUX server state:
+-record(aux, {
+    name,
```
keynslug: Q: Unused?

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_aux.erl
```diff
@@ -0,0 +1,249 @@
...
+manage_otx(DB, Shard, Server, State, #cast_otx_started{result = {ok, Pid}}, #starting{}) ->
+    %% Sucessfully started, monitor the server:
+    Effects = [{monitor, process, aux, Pid}],
+    %% Check if the server state changed in the meantime:
+    case State of
+        leader -> ok;
+        _ -> ra:cast_aux_command(Server, #cast_stop_otx{})
+    end,
+    {ok, #running{pid = Pid}, Effects};
+manage_otx(DB, Shard, Server, leader, #cast_otx_started{result = Err}, #starting{}) ->
```
keynslug, line 178: Q: Possible to get `{error, {already_started, ...}}` here?

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_aux.erl
```diff
@@ -0,0 +1,249 @@
...
+                emqx_ds_builtin_raft_db_sup:stop_otx_leader(DB, Shard)
+            end
+        ),
+    {ok, #stopping{pid = Pid, stopper = AsyncStopper}, []};
+manage_otx(DB, Shard, _Server, _State, #cast_stop_otx{}, ?stopped) ->
+    emqx_dsch:gvar_unset_all(DB, Shard, ?gv_sc_leader),
+    {ok, ?stopped, []};
+manage_otx(_DB, _Shard, _Server, _State, #cast_stop_otx{}, S) ->
+    %% Alredy in transitional state, ignore the request. If server is
+    %% starting then `post_start_otx' will shut it down.
```
keynslug: Q: What is `post_start_otx`?

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_aux.erl
```diff
@@ -0,0 +1,249 @@
...
+-include("emqx_ds_builtin_raft.hrl").
+-include_lib("emqx_durable_storage/include/emqx_ds.hrl").
+-include_lib("snabbkaffe/include/trace.hrl").
+
+%%================================================================================
+%% Type declarations
+%%================================================================================
+
+-define(stopped, stopped).
+-record(starting, {starter :: pid(), retry = 0 :: non_neg_integer()}).
```
keynslug: Nit. `retry` looks unused also.

apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
```diff
@@ -669,8 +671,8 @@ t_rebalance_tolerate_lost(Config) ->
     %% Messages can now again be persisted successfully.
     {Msgs1, MsgStream1} = emqx_utils_stream:consume(50, MsgStream),
     {Msgs2, _MsgStream} = emqx_utils_stream:consume(50, MsgStream1),
-    ?assertEqual(ok, ?ON(N2, emqx_ds_test_helpers:dirty_append(?DB, Msgs1))),
-    ?assertEqual(ok, ?ON(N3, emqx_ds_test_helpers:dirty_append(?DB, Msgs2))),
+    ?assertEqual(ok, ?ON(N2, emqx_ds_test_helpers:dirty_append(AppendOpts, Msgs1))),
```
keynslug, line 799: Q: Naturally wondering why it needs generous number of retries?

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
ieQu1: It may be accessed by v0 state machine. I don't want to touch any of v0 logic in this PR.

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_aux.erl
ieQu1: Fixed

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft.erl
ieQu1: Leader's schema should be propagated every time a node takes over. This is the new way of dealing with inconsistent cluster configurations, where leader's configuration applies to all its followers. 
Maybe in the future we can avoid redundant operations, but probably not in scope of this PR.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
ieQu1: Since ra may replay old logs, I didn't try to change any old behaviors of the storage layer (and if accidentally changed anything, then it's a bug). As such, my goal was to keep the old code paths intact. This, obviously, leads to some duplication and unused fields.

apps/emqx_durable_storage/include/emqx_ds_pmap.hrl
ieQu1: I don't think it's possible to write a macro that is used inside a map, but doesn't add any fields there. 
On the other hand, I don't want to disengage [small map optimizations]( https://erlang.org/documentation/doc-14.0-rc2/doc/efficiency_guide/maps.html#using-maps-as-an-alternative-to-records).

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_machine.erl
ieQu1: > and it would only apply if the storage layer is wiped clean

Storage layer's schema field becomes completely irrelevant in v1, but it's used by v0. I tried to keep v0 behavior intact.

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_aux.erl
ieQu1: Fixed. `..._sup:start_otx_leader` now translates it to `{ok, _}`

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_aux.erl
ieQu1: Fixed.

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_aux.erl
ieQu1: It was a function that got removed during refactoring. Fixed.

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_machine.erl
ieQu1: Indeed. Fixed.

apps/emqx_durable_storage/src/emqx_ds.erl
ieQu1: Fixed.

apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
```diff
@@ -1261,7 +1386,7 @@ end_per_testcase(TCName, Config) ->
     ok = snabbkaffe:stop(),
     Result = emqx_common_test_helpers:end_per_testcase(?MODULE, TCName, Config),
     catch emqx_ds:drop_db(TCName),
-    emqx_cth_suite:clean_work_dir(?config(work_dir, Config)),
+    %% emqx_cth_suite:clean_work_dir(?config(work_dir, Config)),
```
ieQu1: FIXME: rollback

apps/emqx/priv/bpapi.versions
```diff
@@ -85,7 +86,6 @@
 {emqx_plugins,3}.
 {emqx_prometheus,1}.
 {emqx_prometheus,2}.
-{emqx_prometheus,3}.
```
ieQu1, line 88: I'm not sure why automation decided to delete this version, but this BPAPI is empty (?)

apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
ieQu1: Upgrading RFSM takes a significant time. Previously `emqx_ds:wait_for_db` didn't check leader state, so this testcase's reliance on timings broke down. I've since improved `emqx_ds:wait_for_db`, so this retry technically isn't needed any longer. But I think it's worth keeping it, since transaction retries are normal for DS.

apps/emqx/priv/bpapi.versions
ieQu1: Rolled back.

zmstone (approved):

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_liveness.erl
```diff
@@ -0,0 +1,178 @@
...
+%% API functions
+%%================================================================================
+
+-spec start_link(emqx_ds:db()) -> {ok, pid()}.
+start_link(DB) ->
+    gen_server:start_link(?via(DB), ?MODULE, DB, []).
+
+-spec notify_shard_up(emqx_ds:db(), emqx_ds:shard()) -> ok.
+notify_shard_up(DB, Shard) ->
+    emqx_ds_builtin_raft_liveness_proto_v1:multicast_shard_up(nodes(), DB, Shard).
```
keynslug: I guess the risk here is rather small, but: shouldn't this be limited to this cluster nodes only? As far as I understand, running `emqx ctl cluster leave` currently causes the node to form its own separate cluster while still being connected across Erlang distribution.

ieQu1: Thanks, this is reasonable. I will fix that.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12202)

PR #16616, opened 2026-01-22
feat(oidc): add role and namespace expressions

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-15019

Release version: 6.2.0

## PR Checklist

### DIFF
--- a/rel/i18n/emqx_dashboard_sso_oidc.hocon
+++ b/rel/i18n/emqx_dashboard_sso_oidc.hocon
@@ -46,7 +46,7 @@ role_expr.desc: """~
 
   Any result that is not a single value out of the above will forbid the user from being created.
 
-  If not set, will create a viewer user, or retain the existing role, unless the user exists and was created by OIDC.~"""
+  If not set, will create a viewer user, or retain the existing role, if the user exists.~"""
 
 namespace_expr.desc: """~
   A [`jq`](https://jqlang.org/manual/) program to construct the Dashboard user namespace.
@@ -55,7 +55,7 @@ namespace_expr.desc: """~
 
   Any result other than the described above will forbid the user from being created.
 
-  If not set, will create a user in the global namespace, unless the user exists and was created by OIDC.~"""
+  If not set, will create a user in the global namespace or retain existing namespace, if the user exists.~"""
 
 dashboard_addr.desc:
 """The address of the EMQX Dashboard."""


apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc.erl
```diff
@@ -86,6 +86,26 @@ fields(oidc) ->
...
                 )},
+            {role_source,
+                ?HOCON(
+                    hoconsc:enum([userinfo, id_token]),
+                    #{desc => ?DESC(role_source), default => <<"userinfo">>}
+                )},
+            {role_expr,
+                ?HOCON(
+                    binary(),
+                    #{desc => ?DESC(role_expr), default => <<"\"viewer\"">>}
```
zmstone: maybe add a validator to parse jq expression make sure no syntax error.

rel/i18n/emqx_dashboard_sso_oidc.hocon
```diff
@@ -20,6 +20,39 @@ name_var_source.desc: """~
...
+  - `userinfo`: Utilizes the response from the `/userinfo` endpoint.
+  - `id_token`: Utilizes the claims contained in the access token.~"""
+
+namespace_source.desc: """~
+  From which source to extract information to construct the Dashboard user namespace.  Defaults to `userinfo`.
+
+  - `userinfo`: Utilizes the response from the `/userinfo` endpoint.
+  - `id_token`: Utilizes the claims contained in the access token.~"""
+
+role_expr.desc: """~
```
zmstone, line 40: maybe clarify what happens if the extracted role is different from the one stored in EMQX database (bootstrapped previously, or manually set by admin).

thalesmg: I don't think that situation is possible, since the users that come from different backends, such as OIDC, are "different", even if the visual username is the same.

thalesmg: https://github.com/emqx/emqx/pull/16616/changes/b6dcf25cde44f415f587234d69ccafcd6e82ae95

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc.erl
thalesmg: :checkered_flag:

rel/i18n/emqx_dashboard_sso_oidc.hocon
```diff
@@ -46,7 +46,7 @@ role_expr.desc: """~
 
   Any result that is not a single value out of the above will forbid the user from being created.
 
-  Defaults to `"viewer"` (will create a viewer user).~"""
+  If not set, will create a viewer user, or retain the existing role, unless the user exists and was created by OIDC.~"""
```
zmstone: I don't follow the "unless the user exists and was created by OIDC." part.

IIUC:

> If not set, EMQX will create a `veiwer`  user, or retain the existing role if the user was previously created. If set, EMQX always stores the latest role returned from this extraction after a successful login.

rel/i18n/emqx_dashboard_sso_oidc.hocon
```diff
@@ -55,7 +55,7 @@ namespace_expr.desc: """~
 
   Any result other than the described above will forbid the user from being created.
 
-  Defaults to `null` (will create a user in the global namespace).~"""
+  If not set, will create a user in the global namespace, unless the user exists and was created by OIDC.~"""
```
zmstone: same here, maybe make it more clear ?

rel/i18n/emqx_dashboard_sso_oidc.hocon
thalesmg: Users created by SSO backends are independent from each other and from "normal" users, even if their visual usernames are the same.

e.g.: (visual) username `a` may be created multiple times: one normal user, one from ldap, one from oidc, ....

thalesmg: maybe "via OIDC" would be clearer.

zmstone: Thanks.
1. SSO users never clash with EMQX managed users, this is great info, and it maybe makes more sense to document as a separate statement. 
2. It's worth clarifying that EMQX applies the latest mapped role from OIDC response, and stores it for display purpose (so one can view the mapped result from dashboard).

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12410)

PR #16650, opened 2026-01-28
refactor: hide single-value root config schema

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12039)

PR #16656, opened 2026-01-29
chore(sysmon): include process label in sysmonitor reports

keynslug:

## Summary

This PR introduces few cosmetic improvements to system monitor reports on events such as `busy_port`, `long_schedule`, etc.
* Include [process label](https://www.erlang.org/doc/apps/stdlib/proc_lib.html#set_label/1).
* Include more compact "initial call" information tuple.
* Strip few unnecessary information tuples.
* Simplify related module interfaces and test suites.

## PR Checklist

### DIFF
--- a/apps/emqx/include/logger.hrl
+++ b/apps/emqx/include/logger.hrl
@@ -18,12 +18,12 @@
             logger:log(
                 Level,
                 (Data),
-                emqx_logger:attach_label((begin
+                (begin
                     Meta
                 end)#{
                     mfa => {?MODULE, ?FUNCTION_NAME, ?FUNCTION_ARITY},
                     line => ?LINE
-                })
+                }
             );
         false ->
             ok
--- a/apps/emqx/src/emqx_logger.erl
+++ b/apps/emqx/src/emqx_logger.erl
@@ -24,9 +24,7 @@
     error/3,
     critical/1,
     critical/2,
-    critical/3,
-    %% Metadata:
-    attach_label/1
+    critical/3
 ]).
 
 %% Configs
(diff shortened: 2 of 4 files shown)


apps/emqx/src/emqx_vm.erl
```diff
@@ -55,30 +51,25 @@
...
+    suspending
+    | ?PROCESS_GC_KEYS
 ]).
 
 -define(PROCESS_GC_KEYS, [
     memory,
     total_heap_size,
     heap_size,
-    stack_size,
-    min_heap_size
```
zmstone, line 81: Q: this was not valid?

apps/emqx/src/emqx_connection.erl
```diff
@@ -387,6 +387,7 @@ run_loop(
 ) ->
     Peername = emqx_channel:info(peername, Channel),
     emqx_logger:set_metadata_peername(esockd:format(Peername)),
+    proc_lib:set_label({Listener, Peername}),
```
zmstone, line 390: maybe set clientid in emqx_channel after it's parsed | overridden

apps/emqx/src/emqx_vm.erl
keynslug: It's just that `min_heap_size` does not give any usable information IMO, it's usually just a constant.

savonarola (approved):

apps/emqx/src/emqx_connection.erl
keynslug: > 31fd7c6f293810a7e1e84ea18384ab7fe3e31b07

Decided to go with the approach where Client ID gets both into log metadata and process label. At most 8 + 64 bytes of extra memory (for sane Client IDs) per connection still sounds more reasonable than trying to make sure every logging / tracing code path (`?SLOG`, `?TRACE`, `emqx_trace`, snabbkaffe's `?tp`) is aware of the new semantics of process labels.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12043)

PR #16671, opened 2026-02-02
fix(channel): record disconnected_at at takeover/discard begin

zmstone:
Release version: 5.9.3, 5.10.4

## Summary

During session takeover or discard, disconnected_at was being recorded too late
(in ensure_disconnected), after the new session's connected_at was already set.
This caused a race condition where disconnected_at > connected_at, making it
difficult to track client presence state externally.

This fix records disconnected_at immediately when takeover begins or when discard
is received, ensuring it's always earlier than the new session's connected_at.

Related to PR #14799.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_takeover_SUITE.erl
+++ b/apps/emqx/test/emqx_takeover_SUITE.erl
@@ -43,7 +43,9 @@ tc_v5_only() ->
         t_takeover_before_session_expire_willdelay0,
         t_takeover_session_then_normal_disconnect,
         t_takeover_session_then_abnormal_disconnect,
-        t_takeover_session_then_abnormal_disconnect_2
+        t_takeover_session_then_abnormal_disconnect_2,
+        t_disconnected_at_before_connected_at_on_takeover,
+        t_disconnected_at_before_connected_at_on_discard
     ].
 
 init_per_suite(Config) ->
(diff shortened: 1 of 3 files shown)


HJianBo (dismissed):

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -1153,3 +1155,148 @@ make_client_id(Case, Config) ->
...
+        ],
+        %% Get connected event for new session
+        ConnectedEvents = [
+            E
+         || E = {connected, Id, _} <- AllEvents,
+            Id =:= ClientId
+        ],
+
+        ?assert(length(DisconnectedEvents) >= 1, "Should have at least one disconnected event"),
+        ?assert(length(ConnectedEvents) >= 1, "Should have at least one connected event"),
```
thalesmg: Given that the ETS tables opts above do not specify `{keypos, 2}` or similar, then in the whole table there'll be _at most_ one event of each kind (`disconnected`, `connected`) for all clients, right?

thalesmg (approved):

savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13455)

PR #16672, opened 2026-02-02
260202 ensure pid is printed in logs

zmstone:
Release version: 5.8.10, 5.10.4

## Summary

Pid is useful to correlate log entries.
For example, when a process is killed due to heap-size limit exceeded, the kill log does may only contain a Pid,
There are maybe some other logs emitted by the killed process before, but we cannot be sure which one is which.

## PR Checklist


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13291)

PR #16677, opened 2026-02-02
ci: avoid building twice, increase timeout to reduce test flakiness

thalesmg:
https://github.com/emqx/emqx/actions/runs/21594546474/job/62234664534?pr=16673#step:7:103

pytest job now runs in ~ 1 m 20 s vs. ~ 9 m.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12083)

PR #16683, opened 2026-02-03
fix(crl): support https distribution points in CRL cache

zmstone:
Fixes  https://github.com/emqx/emqx/discussions/16667  [EMQX-15079](https://emqx.atlassian.net/browse/EMQX-15079)

Release version: 6.2.0

## Summary

This PR adds HTTPS support for CRL Distribution Point URLs in EMQX CRL cache handling.

## PR Checklist

[EMQX-15079]: https://emqx.atlassian.net/browse/EMQX-15079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_ssl_crl_cache.erl
+++ b/apps/emqx/src/emqx_ssl_crl_cache.erl
@@ -153,11 +153,11 @@ delete({der, CRLs}) ->
     ssl_manager:delete_crls({?NO_DIST_POINT, CRLs});
 delete(URI) ->
     case uri_string:normalize(URI, [return_map]) of
-        #{scheme := "http", path := _} ->
+        #{scheme := Scheme, path := _} when Scheme =:= "http"; Scheme =:= "https" ->
             Key = cache_key(URI),
             ssl_manager:delete_crls(Key);
         _ ->
-            {error, {only_http_distribution_points_supported, URI}}
+            {error, {only_http_and_https_distribution_points_supported, URI}}
     end.
 
 %%--------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


thalesmg:
IIRC, forbiding HTTPS in DPs is by design.  :thinking: 

From [rfc5280](https://datatracker.ietf.org/doc/html/rfc5280):

> When certificates include a cRLDistributionPoints extension with an
   https URI or similar scheme, circular dependencies can be introduced.
   The relying party is forced to perform an additional path validation
   in order to obtain the CRL required to complete the initial path
   validation!  Circular conditions can also be created with an https
   URI (or similar scheme) in the authorityInfoAccess or
   subjectInfoAccess extensions.  At worst, this situation can create
   unresolvable dependencies.

>   CAs SHOULD NOT include URIs that specify https, ldaps, or similar
   schemes in extensions.  CAs that include an https URI in one of these
   extensions MUST ensure that the server's certificate can be validated
   without using the information that is pointed to by the URI.  Relying
   parties that choose to validate the server's certificate when
   obtaining information pointed to by an https URI in the
   cRLDistributionPoints, authorityInfoAccess, or subjectInfoAccess
   extensions MUST be prepared for the possibility that this will result
   in unbounded recursion.

Seems like [other places](https://www.sysadmins.lv/blog-en/designing-crl-distribution-points-and-authority-information-access-locations.aspx) also explicitly do not support it or advise against it:

> NEVER use HTTPS protocol for CRT/CRL file retrieval, because CryptoAPI will permanently fail to fetch this URL.

thalesmg:
Also, this restriction is in OTP (from whence the cache module was "forked"): https://github.com/erlang/otp/blob/4a6bef88743d9f05e6250168c87ed8d89a0162ce/lib/ssl/src/ssl_crl_cache.erl#L155

zmstone:
> CAs SHOULD NOT include URIs that specify https, ldaps, or similar
schemes in extensions. 

`SHOULD NOT` is not `MUST NOT`.

> CAs that include an https URI in one of these extensions MUST ...

See, https can be used, just make sure do not create loop.

> > NEVER use HTTPS protocol for CRT/CRL file retrieval, because CryptoAPI will permanently fail to fetch this URL.

This is limitation of CryptoAPI, whatever that is.

> Also, this restriction is in OTP (from whence the cache module was "forked"):

OTP should fix it.

zmstone:
About potential security concern: What if an attacker sends a certificate which causes EMQX to run in infinite recursion of CRL download.

- There is no pre-verification CRL-fetch in `ssl`. CRL check (fetch) is only triggered after the cert is already verified by trusted CAs
- The CRL download client `httpc` does not enable `verify_peer`.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12419)

PR #16684, opened 2026-02-03
fix(mqtt): allow `client_attrs_init` to use CONNECT password

zmstone:
Release version: 6.0.2, 6.1.1, 6.2.0

## Summary

- Make `mqtt.client_attrs_init` expressions able to use CONNECT password via `${password}`.
- Inject CONNECT password into the temporary variform rendering context in `emqx_channel` during initial attribute setup.
- Remove the temporary `password` key right after rendering so it is not persisted in `clientinfo`.
- Add `t_client_attr_from_password` in `emqx_client_SUITE` to verify attr initialization from password and ensure `password` is not retained in `clientinfo`

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -2022,11 +2022,10 @@ maybe_set_client_initial_attrs(ConnPkt, #{zone := Zone} = ClientInfo) ->
         Inits ->
             UserProperty = get_user_property_as_map(ConnPkt),
             Password = get_connect_password(ConnPkt),
-            ClientInfo1 = initialize_client_attrs(
-                Inits,
-                ClientInfo#{user_property => UserProperty, password => Password}
-            ),
-            {ok, maps:without([user_property, password], ClientInfo1)}
+            RenderCtx = ClientInfo#{user_property => UserProperty, password => Password},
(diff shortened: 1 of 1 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12086)

apps/emqx/src/emqx_channel.erl
```diff
@@ -2021,10 +2021,17 @@ maybe_set_client_initial_attrs(ConnPkt, #{zone := Zone} = ClientInfo) ->
             {ok, ClientInfo};
         Inits ->
             UserProperty = get_user_property_as_map(ConnPkt),
-            ClientInfo1 = initialize_client_attrs(Inits, ClientInfo#{user_property => UserProperty}),
-            {ok, maps:remove(user_property, ClientInfo1)}
+            Password = get_connect_password(ConnPkt),
+            ClientInfo1 = initialize_client_attrs(
```
keynslug: Nit. Perhaps make `initialize_client_attrs/2` just return attrs instead of enriching the client info? To avoid this slightly awkward `maps:without/2`.

zmstone: done

PR #15839, opened 2025-09-03
fix(gpb): support binary map keys with unions

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14407

Release version: 6.0.0

## Summary

See https://github.com/emqx/gpb/pull/1

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #16685, opened 2026-02-03
fix(schema registry): add protobuf option hash to ensure code is regenerated, add workaround for spb alias tracking

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-15080

Release version:
6.0.3, 6.1.1, 6.2.0

## Summary

In https://github.com/emqx/emqx/pull/15839 , we changed `emqx_schema_registry_serde:base_protobuf_opts/0`, namely, adding `{maps_key_type, binary}`, and thus removing the corresponding call to `emqx_utils_maps:binary_key_map`.  For EMQX nodes that started at 6.0.0 and later, this is fine.

However, if the node was running a version prior to 6.0.0 that also uses OTP 27 (the version used by 6.0.x), then any protobuf schema code that is cached will contain the old gpb options (without `maps_key_type`), but the new EMQX will lack the `binary_key_map` call.  

Thus, one side-effect of this is that the new sparkplug B (spb) feature of alias mapping won’t work in those conditions, as the code expects binary keys, and the old generated gpb modules output atom keys.

By adding a hash of the gpb options used to the code cache key used, the code will be regenerated when those change.

## PR Checklist

### DIFF
--- a/apps/emqx_schema_registry/src/emqx_schema_registry_spb_state.erl
+++ b/apps/emqx_schema_registry/src/emqx_schema_registry_spb_state.erl
@@ -72,14 +72,14 @@ parse_spb_topic(Topic) ->
 register_aliases(#message{payload = Payload}, BirthMsg) ->
     try emqx_schema_registry_serde:rsf_spb_decode([Payload]) of
         #{<<"metrics">> := Metrics} ->
-            Mapping = gather_aliases(Metrics),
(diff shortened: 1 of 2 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12126)

PR #16694, opened 2026-02-04
Feat/quic peercert

mikekelly:
Release version: 6.2.0

Summary

Previously, emqx_quic_stream:peercert/1 always returned nossl, making it impossible to use peer_cert_as_username (e.g. cn, dn) on QUIC mTLS listeners. This was noted in the code with a @todo comment referencing lack of msquic support.

Now that quicer:peercert/1 is available, this PR implements the missing function clause to extract the DER-encoded peer certificate from the QUIC connection handle. When quicer:peercert/1 returns {ok, DerCert}, the certificate is passed through to the existing certificate processing pipeline. On error (e.g. client did not present a certificate), it falls back to nossl as before.

Key changes:
  - apps/emqx/src/emqx_quic_stream.erl — added a peercert({quic, Conn, _, _}) clause that delegates to quicer:peercert(Conn)
  - apps/emqx/test/emqx_quic_multistreams_SUITE.erl — added t_peercert to the misc test group, covering all three branches (cert returned, error, non-QUIC socket)

No schema changes. No breaking changes — the only behavioral difference is that QUIC connections with client certificates will now populate peercert instead of returning nossl.

### DIFF
--- a/apps/emqx/test/emqx_quic_multistreams_SUITE.erl
+++ b/apps/emqx/test/emqx_quic_multistreams_SUITE.erl
@@ -135,7 +135,8 @@ groups() ->
             t_conn_resume,
             t_conn_without_ctrl_stream,
             t_probe_client_conn,
-            t_client_probe_conn
+            t_client_probe_conn,
+            t_peercert
         ]}
     ].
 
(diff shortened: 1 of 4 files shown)


savonarola:
Thanks for the contribution!

@qzhuyan What do you think?

qzhuyan:
thanks for the PR, I am checking.

apps/emqx/test/emqx_quic_multistreams_SUITE.erl
```diff
@@ -1799,6 +1800,24 @@ t_client_probe_conn(Config) ->
     ),
     ok.
 
+t_peercert(_Config) ->
```
qzhuyan: again, thanks for the PR! 

I prefer have a real test instead of using the mock.

I could update this test but I need to push to  your branch, is that ok for you?

mikekelly: sure thing, fire away

mikekelly:
bump @savonarola @qzhuyan

qzhuyan:
> bump @savonarola @qzhuyan


@mikekelly 
I am busy with other stuffs, will fix it today or during the weekends.

zmstone (approved):

mikekelly:
Thanks guys 👍

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16030)

PR #16695, opened 2026-02-04
fix(rule engine): ensure metrics exist when creating/updating rule (r58)

thalesmg:
Fixes (?) https://emqx.atlassian.net/browse/EMQX-15088

Release version:
5.8.10, 5.10.4

## Summary

Attempt to fix/mitigate errors such as:

```
Reason: {{badkey,'actions.success'},[{erlang,map_get,['actions.success',#{}],[{error_info,#{module => erl_erts_errors}}]},{emqx_metrics_worker,idx_metric,4,[{file,"emqx_metrics_worker.erl"},{line,683}]},{emqx_metrics_worker,inc,4,[{file,"emqx_metrics_worker.erl"},{line,322}]},{emqx_rule_runtime,do_eval_action_reply_to,2,[{file,"emqx_rule_runtime.erl"},{line,819}]}...
```

Although, at the time of writing, it is not known how the rule metrics disappear from the map stored in the persistent term...  Maybe the metrics worker process is dying (though that would entirely remove it from PT).

## PR Checklist

- [~] The changes are covered with new or existing tests
- [~] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files  (unsure if it really fixes, since it's not known how it's being triggered)

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_engine.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_engine.erl
@@ -318,6 +318,7 @@ load_hooks_for_rule(#{from := Topics}) ->
 maybe_add_metrics_for_rule(Id) ->
     case emqx_metrics_worker:has_metrics(rule_metrics, Id) of
         true ->
+            _ = emqx_metrics_worker:ensure_metrics(rule_metrics, Id, ?METRICS, ?RATE_METRICS),
             ok = reset_metrics_for_rule(Id);
         false ->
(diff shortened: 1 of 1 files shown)


zmstone:
if it's a race condition when rule is deleted, would it result in stale metrics crated ?

thalesmg:
> if it's a race condition when rule is deleted, would it result in stale metrics crated ?

if it's a race between update & delete, or reset & delete, then yes, it could be possible.

PR #16699, opened 2026-02-04
fix(metrics worker): avoid crashing if name, id or metric is unknown (r58)

thalesmg:
Fixes/mitigates https://emqx.atlassian.net/browse/EMQX-15088

Release version:
5.8.10, 5.10.4

## Summary

Attempt to fix/mitigate errors such as:

```
Reason: {{badkey,'actions.success'},[{erlang,map_get,['actions.success',#{}],[{error_info,#{module => erl_erts_errors}}]},{emqx_metrics_worker,idx_metric,4,[{file,"emqx_metrics_worker.erl"},{line,683}]},{emqx_metrics_worker,inc,4,[{file,"emqx_metrics_worker.erl"},{line,322}]},{emqx_rule_runtime,do_eval_action_reply_to,2,[{file,"emqx_rule_runtime.erl"},{line,819}]}...
```

Although, at the time of writing, it is not known how the rule metrics disappear from the map stored in the persistent term...  Maybe the metrics worker process is dying (though that would entirely remove it from PT).

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_runtime.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_runtime.erl
@@ -81,7 +81,7 @@ apply_rule(Rule = #{id := RuleID}, Columns, Envs) ->
         },
         debug
     ),
-    ok = emqx_metrics_worker:inc(rule_metrics, RuleID, 'matched'),
+    ok = metric_inc(RuleID, 'matched'),
     clear_rule_payload(),
     try
         do_apply_rule(Rule, add_metadata(Columns, #{rule_id => RuleID}), Envs)
@@ -191,7 +191,7 @@ do_apply_rule(
                     trace_rule_sql(
                         "SQL_yielded_result", #{result => FinalCollection}, debug
                     ),
-                    ok = emqx_metrics_worker:inc(rule_metrics, RuleId, 'passed')
+                    ok = metric_inc(RuleId, 'passed')
             end,
             NewEnvs = maps:merge(ColumnsAndSelected, Envs),
             {ok, [handle_action_list(RuleId, Actions, Coll, NewEnvs) || Coll <- FinalCollection]};
@@ -214,7 +214,7 @@ do_apply_rule(
     case evaluate_select(Fields, Columns, Conditions) of
         {ok, Selected} ->
             trace_rule_sql("SQL_yielded_result", #{result => Selected}, debug),
-            ok = emqx_metrics_worker:inc(rule_metrics, RuleId, 'passed'),
+            ok = metric_inc(RuleId, 'passed'),
             {ok, handle_action_list(RuleId, Actions, Selected, maps:merge(Columns, Envs))};
         false ->
             trace_rule_sql("SQL_yielded_no_result"),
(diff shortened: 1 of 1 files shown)


apps/emqx_utils/include/logger.hrl
```diff
@@ -0,0 +1,132 @@
...
+    end
+).
+
+-define(AUDIT_HANDLER, emqx_audit).
+-define(TRACE_FILTER, emqx_trace_filter).
+-define(OWN_KEYS, [level, filters, filter_default, handlers]).
+
+%% Internal macro
+-define(_DO_TRACE(Tag, Msg, Meta),
+    case persistent_term:get(?TRACE_FILTER, []) of
```
zmstone: maybe a separate PR: put the trace module in this PT, so there is no static dependency from `emqx_utils` to `emqx_trace`

zmstone: for v6.

apps/emqx/include/logger.hrl
```diff
@@ -14,119 +14,4 @@
...
-%% Only for compile pass, ce edition will not call it
--define(AUDIT(_L_, _M_), _ = {_L_, _M_}).
--endif.
-
-%% print to 'user' group leader
--define(ULOG(Fmt, Args), io:format(user, Fmt, Args)).
--define(ELOG(Fmt, Args), io:format(standard_error, Fmt, Args)).
-
--endif.
+-include_lib("emqx_utils/include/logger.hrl").
```
zmstone: maybe don't do this in v5

apps/emqx_utils/src/emqx_metrics_worker.erl
```diff
@@ -259,12 +260,46 @@ inc(Name, Id, Metric) ->
 
 -spec inc(handler_name(), metric_id(), metric_name(), integer()) -> ok.
 inc(Name, Id, Metric, Val) ->
```
zmstone, line 261: catch exception in caller and log there ?

apps/emqx_utils/src/emqx_metrics_worker.erl
```diff
@@ -259,12 +260,46 @@ inc(Name, Id, Metric) ->
...
-    counters:add(get_ref(Name, Id), idx_metric(Name, Id, counter, Metric), Val).
+    case get_counter_ref_and_idx(Name, Id, Metric) of
+        {ok, CounterRef, Idx} ->
+            counters:add(CounterRef, Idx, Val);
+        {error, Reason} ->
+            ?SLOG_THROTTLE(
```
keynslug: My 2 cents: don't think it's a good idea to introduce logging on the library level, especially stateful (e.g. throttled) logging. I'd say it's better to perhaps have a wrapper on a application level somewhere, e.g. `emqx_rule_action_metrics:inc/4` that does this.

apps/emqx/include/logger.hrl
thalesmg: we need to do this now, otherwise can't log in `emqx_utils`...

thalesmg: mix won't build with `emqx_utils` referencing an `emqx` include (it's not as lax as rebar3)

apps/emqx_utils/include/logger.hrl
thalesmg: I just moved the existing code without a single change.

apps/emqx/include/logger.hrl
thalesmg: reverted it.

apps/emqx_utils/src/emqx_metrics_worker.erl
thalesmg: reverted this.

apps/emqx_utils/src/emqx_metrics_worker.erl
thalesmg: :checkered_flag:

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13294)

PR #16700, opened 2026-02-05
test(greptimedb): stabilize boolean variants assertions

zmstone:
## Summary
Refactor t_boolean_variants in GreptimeDB bridge CT to reduce flakiness by publishing all boolean variants first and asserting rows via clientid + int_key lookup, instead of per-variant table clearing and immediate query assertions.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12135)

PR #16707, opened 2026-02-05
feat: add azure event grid data integration

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14829

Release version:
6.0.3, 6.1.1, 6.2.0

## Summary

This is supposed to be a branded MQTT connector/action/source.  Currently, it has no distinguishing features.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_azure_event_grid/src/emqx_bridge_azure_event_grid_impl.erl
+++ b/apps/emqx_bridge_azure_event_grid/src/emqx_bridge_azure_event_grid_impl.erl
@@ -1,0 +1,179 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_bridge_azure_event_grid_impl).
+
+-behaviour(emqx_resource).
+
+-include_lib("emqx/include/logger.hrl").
+-include_lib("snabbkaffe/include/trace.hrl").
+-include_lib("emqx_resource/include/emqx_resource.hrl").
+-include_lib("emqx/include/emqx_trace.hrl").
+
+%% `emqx_resource` API
+-export([
+    resource_type/0,
+    callback_mode/0,
+
+    on_start/2,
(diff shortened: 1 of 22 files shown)


apps/emqx_bridge_azure_event_grid/src/emqx_bridge_azure_event_grid_impl.erl
```diff
@@ -0,0 +1,179 @@
...
+%% API
+%%------------------------------------------------------------------------------
+
+%%------------------------------------------------------------------------------
+%% Internal fns
+%%------------------------------------------------------------------------------
+
+%% Azure Event Grid requires SNI to be set when connecting.  We copy it from `server` if
+%% unset and TLS is enabled.
+maybe_set_sni(#{ssl := #{enable := true} = SSLOpts0} = ConnConfig0) when
```
zmstone, line 171: maybe this should be made a common logic for regular mqtt bridges too.

thalesmg: event grid _forces_ this to be the case; other brokers might not impose this restriction.  so it might not be quite general?

zmstone: Adding SNI for servers which do not check SNI most of the time is harmless.

The upside is: almost all shared brokers enforce SNI. Before this is released, people will have to use the regular mqtt connector to connect event grid or other brokers.
Often we get issue reports like: i can connect using mqttx, mosquitto, and python script, etc., but cannot connect from EMQX, then I'd reply: on, could you try fill the hostname (no scheme, no :port), and turns out that's the exact reason.

Maybe fix in another JIra/PR.

thalesmg: https://emqx.atlassian.net/browse/EMQX-15113

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12423)

PR #16721, opened 2026-02-09
fix: avoid duplicate QoS2 delivery after await_rel timeout

zmstone:
Release version: 6.2.0

## Summary

Handle QoS 2 retransmitted `PUBLISH` packets with `DUP=1` safely when `await_rel_timeout` state has already expired.

Before this change, after the broker expired `awaiting_rel` (default 300s), a retried QoS 2 `PUBLISH` with the same packet id could be treated as a new publish and be delivered to subscribers again. The fix updates `emqx_session_mem:publish/3` to treat such retransmissions as handshake retries only: return `PUBREC` and (re)track `awaiting_rel` without re-publishing the application message. A regression test was added in `emqx_channel_await_rel_expire_SUITE` to validate no duplicate delivery after expiry while still acknowledging with `PUBREC`.

## PR Checklist


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12425)

PR #16725, opened 2026-02-10
fix(config): disable tcp congestion alarm by default

zmstone:
Releases: 6.2.0

## Summary

Disable TCP connection congestion alarm by default by changing `conn_congestion.enable_alarm` default from `true` to `false`.

To preserve existing congestion alarm test behavior, suites that validate congestion alarm activation now explicitly set `conn_congestion.enable_alarm = true` in their test config. This keeps runtime defaults quieter while keeping alarm-specific tests intentional and explicit.

### Reasons

- High false alarm rate (temp network glitch causes alarm to go off).
- Not actionable (there is typically not much one can do to to resolve this alarm).

## PR Checklist


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12429)

PR #16727, opened 2026-02-10
ci: skip job if matrix is empty

thalesmg:
https://github.com/emqx/emqx/actions/runs/21869383620?pr=16723

> Error when evaluating 'strategy' for job 'ct'. emqx/emqx/.github/workflows/run_test_cases.yaml@51e685e92addfb2504ad6c38dbc0b5fb21081d2a (Line: 144, Col: 9): matrix must define at least one vector


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12179)

PR #16735, opened 2026-02-11
feat(plugin): support HTTP API callbacks from plugins

zmstone:

## Summary

This change allows plugins to expose HTTP callbacks via `/api/v5/plugin_api/:plugin/[...]`.

User-facing behavior:
- Requests to plugin API endpoints are routed only when the target plugin is running.
- If the plugin, callback module, or callback function is missing, the API returns `404`.
- Callback timeouts return `503`; callback crashes and invalid responses return `500` with clear error messages.

Code highlights:
- Add plugin API behaviour and callback resolver in `emqx_plugins`. `behaviour` is optional, but callback is a must.
- Add dashboard endpoint module for plugin API request forwarding.
- Add coverage in plugin and dashboard suites for success/not-found/auth/crash paths.

## PR Checklist

### DIFF
--- a/apps/emqx_plugins/test/emqx_plugins_test_helpers.erl
+++ b/apps/emqx_plugins/test/emqx_plugins_test_helpers.erl
@@ -12,6 +12,8 @@
 -include_lib("snabbkaffe/include/snabbkaffe.hrl").
 
 -define(PACKAGE_SUFFIX, ".tar.gz").
+-define(DOWNLOAD_RETRIES, 3).
+-define(HTTP_TIMEOUT, 30_000).
 
 get_demo_plugin_package(
     #{
@@ -24,7 +26,7 @@ get_demo_plugin_package(
 ) ->
     TargetName = lists:flatten([ReleaseName, "-", PluginVsn, ?PACKAGE_SUFFIX]),
     FileURI = lists:flatten(lists:join("/", [GitUrl, ReleaseTag, TargetName])),
-    {ok, {_Status, _Headers, PluginBin}} = httpc:request(FileURI),
+    PluginBin = download_plugin_package(FileURI, ?DOWNLOAD_RETRIES),
     Pkg = filename:join([
         WorkDir,
         TargetName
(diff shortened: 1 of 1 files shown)


apps/emqx_plugins/src/emqx_plugins_api_endpoint.erl
```diff
@@ -0,0 +1,156 @@
...
+        body => maps:get(body, Params, #{})
+    },
+    AuthMeta = maps:get(auth_meta, Params, #{}),
+    Context = #{
+        auth_meta => AuthMeta,
+        namespace => emqx_dashboard:get_namespace(Params)
+    },
+    case Plugin of
+        undefined ->
+            {404, #{code => <<"NOT_FOUND">>, message => <<"Plugin API Not Found">>}};
```
savonarola: nit: use `emqx_http_api.hrl` helpers

apps/emqx_plugins/src/emqx_plugin_api.erl
```diff
@@ -0,0 +1,25 @@
...
+    PathRemainder :: [binary()],
+    Request :: map(),
+    Context :: map()
+) ->
+    callback_response().
+
+-type callback_response() ::
+    {ok, Status :: pos_integer(), Headers :: map() | [{binary(), iodata()}], Body :: term()}
+    | {error, Code :: atom() | binary() | string(), Msg :: iodata()}
+    | {error, Status :: pos_integer(), Headers :: map() | [{binary(), iodata()}], Body :: term()}.
```
savonarola: nit: maybe add more descriptive types, `header_name()`, etc.

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -121,6 +122,20 @@ plugin_schema(NameVsn) ->
 plugin_i18n(NameVsn) ->
     ?CATCH(emqx_plugins_fs:read_i18n(NameVsn)).
 
+-spec resolve_api_callback(binary() | string() | atom()) ->
```
savonarola: Nit: the new API looks a bit orthogonal to the existing ones. The endpoint and the new methods operate with module names, not `name_vsn`s. Also a new callback convention is introduced, while there is an existing module and convention for talking with plugins: https://github.com/emqx/emqx/blob/a6d5022aeaff55443f68e5c1d382305274961ca3/apps/emqx_plugins/src/emqx_plugins_apps.erl#L118 It also contains now duplicated code for checking callback presence, catching exceptions, etc.

Maybe we could
* hide the details and expose a single function from `emqx_plugins` instead, like `emqx_plugins:handle_api_call(name_vsn(), request(), timeout())` and make call dispatching inside `emqx_plugins` app, not in the dashboard controller.
* follow the existing convention, assume `PLUGIN_app:on_handle_api_call(...)` callback existence and call it via `emqx_plugins_apps:on_handle_api_call`, as for the other callbacks.

apps/emqx_plugins/src/emqx_plugins_api_endpoint.erl
```diff
@@ -0,0 +1,156 @@
...
+
+-define(DEFAULT_TIMEOUT, 5000).
+
+api_spec() ->
+    emqx_dashboard_swagger:spec(?MODULE, #{check_schema => false}).
+
+paths() ->
+    ["/plugin_api/:plugin/[...]"].
+
+schema("/plugin_api/:plugin/[...]") ->
```
savonarola, line 25: Nit: all the plugin-related endpoints operate with `name_vsn()`.

zmstone: These are management APIs for admins.

Runtime REST API users should not have to care about version, compatibility is a responsibility for the plugin itself.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12176)

PR #16747, opened 2026-02-12
fix(plugin_api): refine api-doc for plugin_api endpoint

zmstone:
Follow-up fix for #16735.
Only change with no end-user impact and no changelog needed.
This adjusts the plugin API swagger requestBody schema to plain OpenAPI object format and added more descrions.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12187)

PR #16746, opened 2026-02-12
perf: reduce memsup process scan overhead

zmstone:
Release version: 5.8.10, 5.10.4

## Summary
- set `-os_mon memsup_system_only true` in `apps/emqx/etc/vm.args.cloud` so releases default to system-wide memory stats only
- reduce per-process memsup scan overhead on running nodes

## PR Checklist
- [ ] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

```
  GET /api/v5/prometheus/stats
    → emqx_prometheus_api:stats/2
    → emqx_prometheus:collect/1
    → emqx_prometheus:vm_data/1
    → emqx_mgmt:vm_stats/0
    → emqx_mgmt:get_sys_memory/0
    → emqx_mgmt_cache:get_sys_memory/0       (3-sec timeout)
    → emqx_mgmt_cache:refresh_sys_memory/2
    → load_ctl:get_sys_memory/0
    → lc_lib:get_sys_memory/0
    → lc_lib:do_get_sys_memory_usage/1
    → memsup:get_system_memory_data()         ← BOTTLENECK
```

PR #16757, opened 2026-02-13
perf(os_mon): set memsup_system_only by default

zmopish:
Forwport of #16746 to release-60.

## Summary

Set memsup_system_only=true by default to reduce os_mon overhead.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12205)

PR #16758, opened 2026-02-13
Merge plugins-60 to release-60

zmstone:
Will NOT diverge with `plugins-*` branches after all.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12193)

PR #16765, opened 2026-02-13
test(audit): make t_http_api deterministic on release-60

zmstone:
## Summary
Fix flaky assertion in `emqx_audit_api_SUITE:t_http_api`.

The test previously queried audit logs with `limit=1` and assumed the latest record was always `/configs/global_zone`, which can race with other config audit entries (for example `/configs/log`).

This change makes the test deterministic by:
- capturing start timestamp before issuing the config update
- querying audit API with filters: `from=rest_api`, `operation_id=/configs/global_zone`, and `gte_created_at`
- polling briefly until the matching entry appears


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12207)

PR #16775, opened 2026-02-15
fix(schema_validation): fail fast on hanging JSON schema refs

zmstone:
Fixes https://emqx.atlassian.net/browse/EMQX-14514

## Summary

This PR adds a bounded timeout around schema-registry schema checks used by schema validation, so publish processing fails fast when a JSON Schema `$ref` endpoint is unresponsive.

It also adds a deterministic CT case that reproduces the hanging `$ref` scenario with a local non-responsive endpoint and verifies both timeout-path behavior and validation metric consistency.

## PR Checklist

### DIFF
--- a/apps/emqx_schema_validation/src/emqx_schema_validation.erl
+++ b/apps/emqx_schema_validation/src/emqx_schema_validation.erl
@@ -232,6 +232,9 @@ evaluate_schema_check(Check, Validation, #message{payload = Data}) ->
             false
     end.
 
+%% Run schema check in a spawned process with a timeout.
+%% JSON schemas may contain external `$ref` URLs that trigger HTTP fetches
+%% which can hang indefinitely, so we need the timeout to fail fast.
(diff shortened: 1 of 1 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12213)

apps/emqx_schema_validation/src/emqx_schema_validation.erl
```diff
@@ -223,6 +232,14 @@ evaluate_schema_check(Check, Validation, #message{payload = Data}) ->
             false
     end.
 
+schema_check_with_timeout(SerdeName, Data, ExtraArgs) ->
+    emqx_utils:nolink_apply(
```
savonarola, line 239: Q: shouldn't we run with spawn only when schema check potentially involves external interactions? Are there scenarios where we just check against a locally-defined schema? In these cases spawning a process may be an overkill 🤔

zmstone:
this is more complex than i anticipated.
will try to fix in a different pr.

PR #16777, opened 2026-02-15
fix(gateway): include hook context for session.created

zhongwencool:
## Summary
- Add hook context when firing `session.created` in gateway session manager.
- Update related gateway test coverage.

 Ensure hooks receive complete context for downstream logic to prevent crash.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12231)

PR #16779, opened 2026-02-15
fix(mqtt): classify malformed first packets as invalid CONNECT

zmstone:
Fixes EMQX-14889

Release version:
6.0.3, 6.1.2, 6.2.0

## Summary

Improve malformed first-packet handling before MQTT CONNECT is established:

- Classify first-packet non-CONNECT cases as `invalid_connect_packet` instead of generic frame errors.
- Keep shutdown behavior strict (close connection), but use clearer shutdown counting semantics for pre-CONNECT failures.
- Add protocol resemblance hints in error metadata (`resemble_protocol`) and detect PROXY protocol v2 signatures.
- Keep existing failure log message names and enrich reason maps for better diagnostics.

Also adds/updates integration and unit coverage for:

- first packet gate behavior on both `gen_tcp` and `socket` backends,
- idle-state malformed frame handling,
- protocol resemblance detection branches.

## PR Checklist


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12214)

apps/emqx/src/emqx_connection.erl
```diff
@@ -776,6 +776,14 @@ next_incoming_msgs(Packets) ->
     lists:foldl(Fun, [], Packets).
 
 parse_incoming(Data, State = #state{parser = Parser}) ->
+    parse_incoming_after_gate(
```
keynslug: Nit. Perhaps do it in the catch clauses (error handling branches) instead? So it doesn't make every packet parsing call slightly more complex, by forcing to compute extra conditionals and small heap allocations. E.g. by providing something like `emqx_frame:guess_first_packet_protocol/1` instead.

apps/emqx/test/emqx_listeners_SUITE.erl
```diff
@@ -81,6 +81,70 @@ t_restart_listeners(_) ->
     ok = emqx_listeners:restart(),
     ok = emqx_listeners:stop().
 
+t_first_packet_not_connect_plaintext(_Config) ->
```
keynslug: FWIW I tend to put similar testcases to `{group, misbehaving}` in `emqx_client_SUITE`.

apps/emqx/src/emqx_frame.erl
```diff
@@ -175,6 +176,82 @@ parse_complete(
             parse_packet_complete(Rest2, Header, Options)
     end.
 
+-spec guess_first_packet_protocol(binary()) -> map().
+guess_first_packet_protocol(<<Type:4, _Flags:4, _/binary>> = Data) ->
+    {Preview, PreviewEncoding} = format_data_prefix(Data, 32),
+    #{
+        packet_type => Type,
+        packet_type_name => emqx_packet:type_name(Type),
```
keynslug, line 184: Q: Is there a reason to report `packet_type`? As both `packet_type_name` and received prefix seem to provide the same amount information. Perhaps just `packet_type => emqx_packet:type_name(Type)`?

keynslug (approved):

apps/emqx/src/emqx_frame.erl
zmstone: make sense

PR #16784, opened 2026-02-15
fix(plugins): avoid self node config fetch on startup

zmstone:
Release version:
- 6.0.3
- 6.1.1
- 6.2.0

## Summary

This change reduces noisy plugin startup warnings by avoiding local-node config fetch in the plugin cluster sync path.

During startup, plugin config is now fetched only from peer nodes, so single-node deployments no longer emit repeated `config_not_found_on_node` warnings for self.

## PR Checklist


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12219)

PR #16786, opened 2026-02-16
feat(variform): add jq function to variform bif

zmstone:

## Summary

Add `jq` as a supported built-in function in variform expressions.

This PR adds `jq/2` and `jq/3` to `emqx_variform_bif` so variform expressions can evaluate jq filters against JSON input (binary JSON text or Erlang terms). It keeps variform-style argument validation and propagates jq runtime errors as `{jq_exception, Reason}`.

Tests were added for both direct BIF usage and expression-level rendering paths.

## PR Checklist

### DIFF
--- a/apps/emqx_utils/src/emqx_variform_bif.erl
+++ b/apps/emqx_utils/src/emqx_variform_bif.erl
@@ -75,6 +75,8 @@
     base64_decode/1,
     base64_decode/2,
     base64_decode/3,
+    jq1/2,
+    jq1/3,
     json_value/2,
     jwt_value/2
 ]).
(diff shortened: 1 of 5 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12221)

apps/emqx_utils/src/emqx_variform_bif.erl
```diff
@@ -75,6 +75,8 @@
     base64_decode/1,
     base64_decode/2,
     base64_decode/3,
+    jq1/2,
+    jq1/3,
```
keynslug, line 79: Q: Why not `jq`?

apps/emqx_utils/src/emqx_variform_bif.erl
```diff
@@ -680,6 +682,41 @@ base64_decode(Data, Opt1, Opt2) ->
     Options = base64_opts(Options0, Opt2),
     base64:decode(Data, Options).
 
+-spec jq1(binary(), binary() | term()) -> term().
+jq1(_FilterProgram, NULL) when ?IS_NULL(NULL) ->
+    ?BADARG();
+jq1(NULL, _JSON) when ?IS_NULL(NULL) ->
+    ?BADARG();
+jq1(FilterProgram, JsonBin) when is_binary(FilterProgram), is_binary(JsonBin) ->
+    jq_decode_first(jq:process_json(FilterProgram, JsonBin));
```
keynslug, line 691: General question: it makes much harder to reason about computation complexity of variform expressions (e.g. how soon computation terminates), okay to sacrifice this property?

apps/emqx_utils/src/emqx_variform_bif.erl
zmstone: jq returns an array, jq1 returns the first element (if exists).

apps/emqx_utils/src/emqx_variform_bif.erl
zmstone: you mean do not accept binary input ?

apps/emqx_utils/src/emqx_variform_bif.erl
zmstone: this is mainly to differ from the jq function in rule-engine, which returns list.

apps/emqx_utils/src/emqx_variform_bif.erl
keynslug: Sorry, I meant in general, disregard the specific line this comment is left on. It's more about for example an ability to express loops with arbitrary termination conditions.

apps/emqx_utils/src/emqx_variform_bif.erl
keynslug: I see. To me, this restriction sounds a bit artificial and counter-intuitive. I guess mostly because as a user, I'm not really fully aware of jq computation model, to me it's always "one input → one output". And when it suddenly loses information by throwing some results away, this may be puzzling.

Perhaps another option is to `lists:map(fun emqx_utils_json:decode/1, Outputs)` when there's more than one output? It's apparently not consistent either, but might be more intuitive?

PR #16788, opened 2026-02-16
chore(i18n): deduplicate rel i18n keys

zmstone:
## Summary
Remove duplicated i18n key definitions in rel/i18n HOCON files.

- Keep a single definition per key path
- Preserve intended wording (for example, Action wording for Pulsar `desc_name`)
- This is internal i18n housekeeping with no feature behavior change


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12222)

PR #16801, opened 2026-02-18
refactor: do not include_lib in emqx.hrl

zmstone:
`include_lib` makes `include("emqx.hrl")` not possible to use for plugins.

This PR only adds `include_lib` in the modules which require `#message{}` record.

### DIFF
--- a/apps/emqx/include/emqx.hrl
+++ b/apps/emqx/include/emqx.hrl
@@ -39,13 +39,11 @@
 
 -record(subscription, {topic, subid, subopts}).
 
--include_lib("emqx_utils/include/emqx_message.hrl").
-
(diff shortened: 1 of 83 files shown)


ieQu1:
There's a surprising number of changes. So the problem is that `emqx.hrl` cannot contain `-include_lib("emqx_utils/include/emqx_message.hrl").`? It sounds like a build system problem.

zmstone:
> There's a surprising number of changes. So the problem is that `emqx.hrl` cannot contain `-include_lib("emqx_utils/include/emqx_message.hrl").`? It sounds like a build system problem.

Try to build plugins independently inside

PR #16805, opened 2026-02-19
feat(authz): support hook-controlled authz cache skip

zmstone:

## Summary

This PR adds an optional cache-control signal for `client.authorize` hook results so dynamic authorization decisions can skip authz cache writes.

Changes:
- Extend `emqx_access_control:authorize_hook_result()` with optional `no_cache` and `cache` fields.
- Update authz cache write path in `emqx_access_control` to honor hook cache policy.
- Add tests in `emqx_access_control_SUITE` to verify both default cache store and no-cache behavior.

This branch intentionally excludes plugin changes.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_access_control.erl
+++ b/apps/emqx/src/emqx_access_control.erl
@@ -20,10 +20,9 @@
         result := authz_result(),
         from := term(),
         %% Optional cache control for authorization result.
-        %% no_cache=true (or cache=false) tells EMQX not to store this result
+        %% is_cacheable=false tells EMQX not to store this result
         %% in authz cache for the current topic/action.
-        no_cache => boolean(),
-        cache => boolean()
+        is_cacheable => boolean()
     }.
 
 -type authn_result() :: #{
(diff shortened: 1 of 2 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12232)

apps/emqx/src/emqx_access_control.erl
```diff
@@ -246,9 +253,12 @@ do_authorize(ClientInfo, Action, Topic) ->
                 'client.check_authz_complete',
                 [ClientInfo, Action, Topic, deny, unknown_return_format]
             ),
-            deny
+            {deny, true}
```
thalesmg, line 256:
```suggestion
            {deny, false}
```
Otherwise, we'll cache the failure, which could be transient?

zmstone: hard to justify.
the cache would act as a throttle for transient errors.

PR #16807, opened 2026-02-19
chore(ci): measure ct excution times

savonarola:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12233)

PR #16808, opened 2026-02-20
fix: use safe binary_to_term for three external decode paths

zmstone:
## Summary
- add `[safe]` to `term_decode/1` in rule engine
- add `[safe]` to ExProto `to_pid/1` decoding path
- add `[safe]` to Syskeeper frame marshaller binary decoding
- bump app versions for `emqx_gateway_exproto` and `emqx_bridge_syskeeper` as required on `release-58`

## Notes
- internal hardening change with no end-user feature impact


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13299)

PR #16809, opened 2026-02-20
docs(exproto): deprecate

zmstone:
## Summary
- add optional `config_schema_importance` to gateway definitions
- let `emqx_gateway_schema` apply per-gateway schema importance when building `gateway` fields
- set ExProto gateway config schema importance to `?IMPORTANCE_NO_DOC` so it is excluded from default docgen output

## Notes
- internal-only change
- no changelog entry required


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12234)

zmstone:
Deprecation will be announced in emqx-docs.

PR #16839, opened 2026-02-27
260226  `emqx_username_quota` plugin 1.1.0

zmstone:
### Overview

- When plugin boots up, backfill quota state from already connected sessions.
  - Previously assumed plugin is installed before node starts.
- Store quota overrides in DB
  - Previously used a static whitelist config for a list of usernames.
- Refactored snapshot management to use 'green' and 'blue' tables to minimize the gap between snapshot builds.

### Details
- Replaced the static `username_white_list` configuration with per-username quota overrides managed via REST API.
  Overrides support custom limits (positive integer), unlimited sessions (`"nolimit"`), or connection blocking (`0`).
- Added REST API endpoints for managing overrides:
  `POST /quota/overrides`, `DELETE /quota/overrides`, and `GET /quota/overrides`.
- The `GET /quota/usernames` and `GET /quota/usernames/:username` responses now include a `limit` field
  showing the effective quota for each username.
- Config validation now rejects `max_sessions_per_username` values less than 1
  instead of silently falling back to the default.
- Replaced synchronous snapshot rebuild with async background build using blue/green ETS table swapping,
  eliminating data gaps during rebuilds.
- Added mandatory `used_gte` query parameter to `GET /quota/usernames` to filter by minimum session count
  and reduce snapshot size. Providing both `used_gte` and `cursor` returns 400.
- Added `DELETE /quota/snapshot` endpoint to force an immediate snapshot rebuild.
- Snapshot ownership is now core-only; `GET /quota/usernames` returns 404 `NOT_AVAILABLE` on replicant nodes.
- Replaced `snapshot_refresh_interval_ms` config with `snapshot_min_age_ms` (default 300000, clamped to 120000–900000)
  controlling the minimum age before a snapshot can be rebuilt.
- Invalid or unavailable cursor now returns 400 `INVALID_CURSOR` instead of silently restarting from page 1.
- 503 responses now include `snapshot_build_in_progress: true` when no snapshot is available yet
  (e.g. first request after startup or force-rebuild).
- Cursor encoding now uses URL-safe base64 without padding for cleaner query strings.
- On plugin startup, existing local sessions are bootstrapped into quota state with
  replication-watermark-based backpressure to avoid overloading core nodes.

### DIFF
--- a/plugins/emqx_username_quota/src/emqx_username_quota_cluster_watch.erl
+++ b/plugins/emqx_username_quota/src/emqx_username_quota_cluster_watch.erl
@@ -136,12 +136,11 @@ finish_bootstrap(Count, _BatchCount, BatchSeq) ->
 
 process_channel(Rest, Count, BatchCount, BatchSeq, ClientId, ChanPid, ClientInfo) ->
     Username = emqx_utils_conv:bin(maps:get(username, ClientInfo, <<>>)),
-    ClientIdBin = emqx_utils_conv:bin(ClientId),
-    case Username =:= <<>> orelse ClientIdBin =:= <<>> of
+    case Username =:= <<>> of
         true ->
             bootstrap_loop(Rest, Count, BatchCount, BatchSeq);
         false ->
-            emqx_username_quota_pool:add(Username, ClientIdBin, ChanPid),
+            emqx_username_quota_pool:add(Username, ClientId, ChanPid),
             after_add(Rest, Count + 1, BatchCount + 1, BatchSeq)
     end.
 
--- a/plugins/emqx_username_quota/src/emqx_username_quota_state.erl
+++ b/plugins/emqx_username_quota/src/emqx_username_quota_state.erl
@@ -28,10 +28,13 @@
 -include("emqx_username_quota.hrl").
 
 -doc """
-Create the two core Mria tables used by username quota state.
+Create the 3 core Mria tables used by username quota state.
 ?RECORD_TAB stores one row per active session, keyed by {Username, ClientId, Pid}.
+
 ?COUNTER_TAB stores per-node counters, keyed by {Username, Node};
 global count per username is derived by summing all node counters.
+
+?OVERRIDE_TAB stores the per-username quota overrides, keyed by Username.
 """.
 create_tables() ->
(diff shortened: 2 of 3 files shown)


plugins/emqx_username_quota/src/emqx_username_quota_cluster_watch.erl
```diff
@@ -83,6 +103,91 @@ terminate(_Reason, _State) ->
...
+    {ok, Count};
+finish_bootstrap(Count, _BatchCount, BatchSeq) ->
+    case wait_for_replication(BatchSeq + 1) of
+        ok -> {ok, Count};
+        timeout -> {error, replication_timeout, Count}
+    end.
+
+process_channel(Rest, Count, BatchCount, BatchSeq, ClientId, ChanPid, ClientInfo) ->
+    Username = emqx_utils_conv:bin(maps:get(username, ClientInfo, <<>>)),
+    ClientIdBin = emqx_utils_conv:bin(ClientId),
```
thalesmg: Q: is it possible for the clientid to be empty?  When the broker generates one for the client, it seems to be stored in the channel table as the key (it is kept as `""` in the `ConnInfo` element, though).

plugins/emqx_username_quota/src/emqx_username_quota_state.erl
```diff
@@ -10,21 +10,29 @@
...
-%% @doc Create the two core Mria tables used by username quota state.
-%% ?RECORD_TAB stores one row per active session, keyed by {Username, ClientId, Pid}.
-%% ?COUNTER_TAB stores per-node counters, keyed by {Username, Node};
-%% global count per username is derived by summing all node counters.
+-doc """
+Create the two core Mria tables used by username quota state.
+?RECORD_TAB stores one row per active session, keyed by {Username, ClientId, Pid}.
+?COUNTER_TAB stores per-node counters, keyed by {Username, Node};
+global count per username is derived by summing all node counters.
+""".
```
thalesmg, line 38: Q: does `?OVERRIDE_TAB` need a description?

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12517)

PR #16840, opened 2026-02-27
feat: add basic a2a registry functionality

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-15142

Release version:
6.2.0

## Summary

Here we implement the basics of a2a registry.  The feature is disabled by default.  We use the existing retainer implementation to piggy-back on all the complexities it already handles to avoid re-implementing it, and to retain its semantics regarding message handling and administrative operations.

We validate the consistency between publishing clientid and the target discovery topic id, the ids themselves, and the payload schema (if enabled).  If any validation fails, we forbid the publishing (the retained message does not get stored).

We don't implement any extra indices at the moment.  That's deferred to a later time when we have concrete queries to be made.

Schema was converted to draft 06 by an llm...  because `jesse` does not support draft 2020-12, the original schema format.

To be done in follow ups (referencing the [current EIP](https://github.com/emqx/eip/blob/1a4589ac5a899cddcf6bf7c1c00f7e2acacccdb0/active/0033-agent-registry.md)):

- [A2A Authorization Policy Levels](https://github.com/emqx/eip/blob/1a4589ac5a899cddcf6bf7c1c00f7e2acacccdb0/active/0033-agent-registry.md?plain=1#L419)
- [Message-Layer Trust](https://github.com/emqx/eip/blob/1a4589ac5a899cddcf6bf7c1c00f7e2acacccdb0/active/0033-agent-registry.md?plain=1#L433)
- Some CLI admin commands (register from file, stats)
- HTTP CRUD API

## PR Checklist

### DIFF
--- a/apps/emqx_a2a_registry/src/emqx_a2a_registry_internal.hrl
+++ b/apps/emqx_a2a_registry/src/emqx_a2a_registry_internal.hrl
@@ -14,5 +14,7 @@
 
 -define(A2A_SCHEMA_REGISTRY_SERDE_NAME, <<"__sdKn7w9B3UuI7LAmkQSHP__a2a_agent_card">>).
 
+-define(SEGMENT_ID_RE, <<"^[A-Za-z0-9._-]+$">>).
+
 %% END ifndef(__EMQX_A2A_REGISTRY_HRL__)
 -endif.
--- a/apps/emqx_a2a_registry/src/emqx_a2a_registry.erl
+++ b/apps/emqx_a2a_registry/src/emqx_a2a_registry.erl
@@ -5,7 +5,15 @@
 
 %% API
 -export([
-    ensure_card_schema_registered/0
+    ensure_card_schema_registered/0,
+
+    discovery_topic/3,
+
+    list_cards/0,
+    list_cards/1,
+    delete_card/1,
+
+    lookup_agent_status/1
 ]).
 
 %%------------------------------------------------------------------------------
@@ -13,6 +21,7 @@
 %%------------------------------------------------------------------------------
 
 -include("emqx_a2a_registry_internal.hrl").
+-include_lib("emqx_utils/include/emqx_message.hrl").
 
 %%------------------------------------------------------------------------------
 %% API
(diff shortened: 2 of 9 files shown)


apps/emqx_a2a_registry/src/emqx_a2a_registry_internal.hrl
```diff
@@ -0,0 +1,20 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-ifndef(__EMQX_A2A_REGISTRY_HRL__).
+-define(__EMQX_A2A_REGISTRY_HRL__, true).
+
+-define(A2A_TOPIC_NS, <<"a2a">>).
```
zmstone:
```suggestion
-define(A2A_TOPIC_NS, <<"$a2a">>).
```

apps/emqx_a2a_registry/src/emqx_a2a_registry_internal.hrl
```diff
@@ -0,0 +1,20 @@
...
+-define(A2A_TOPIC_V1, <<"v1">>).
+-define(A2A_TOPIC_DISCOVERY, <<"discovery">>).
+
+-define(A2A_PROP_STATUS_KEY, <<"a2a-status">>).
+-define(A2A_PROP_ONLINE_VAL, <<"online">>).
+-define(A2A_PROP_OFFLINE_VAL, <<"offline">>).
+
+-define(A2A_SCHEMA_REGISTRY_SERDE_NAME, <<"__sdKn7w9B3UuI7LAmkQSHP__a2a_agent_card">>).
```
zmstone, line 15: why the magic prefix.
it's a standard, so we could just use `a2a_agent_card`? maybe add `_v1` suffix.
i don't think we'd have to worry about clasing.

thalesmg: to avoid clashing with user-managed serdes, just like we do with sparkplug b's serde.

thalesmg: ... and this name is not exposed, so no reason not to use it, either.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12443)

PR #16842, opened 2026-02-28
fix(plugins): reduce plugin config not-found log to debug level

zmstone:
Releases: 6.0.3, 6.1.2, 6.2.0
## Summary

When a node starts up and tries to fetch plugin config from peer nodes,
it previously logged a warning even when all peers simply didn't have the config yet
(e.g., first node to load the plugin). This was confusing and noisy.

Now config-not-found from all peers is logged at debug level, while genuine errors
(RPC failures, timeouts) remain as warnings. Also simplified the internal error
representation by removing the redundant `config_not_found_on_node` tuple.

## PR Checklist
- [x] Change log for changes visible by users has been added to `changes/ee/fix-16842.en.md`


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12468)

PR #16847, opened 2026-03-02
fix(message_transformation): handle unicode in prettify_operation

zmstone:
## Summary

Fix crash in `prettify_operation/1` when variform expressions contain non-ASCII
unicode characters (e.g. Chinese characters in string literals).

`emqx_variform:compile/1` converts the expression binary to a unicode codepoint
list via `unicode:characters_to_list/1`. When `decompile/1` returns this list,
`iolist_to_binary/1` crashes with `badarg` because codepoints > 255 are not
valid iolist elements. Replaced with `unicode:characters_to_binary/1`.

Fix versions: 6.0.3, 6.1.2, 6.2.0

https://emqx.atlassian.net/browse/EEC-1297

### DIFF
--- a/apps/emqx_message_transformation/src/emqx_message_transformation.erl
+++ b/apps/emqx_message_transformation/src/emqx_message_transformation.erl
@@ -177,7 +177,7 @@ prettify_operation(Operation0) ->
     %% TODO: remove injected bif module
     Operation = maps:update_with(
         value,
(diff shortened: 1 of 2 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12471)

PR #16848, opened 2026-03-02
fix(dsraft): stabilize test suite + fix minor issues

keynslug:
Release version: 6.1.2, 6.2.0

## Summary

This PR focuses on the `emqx_ds_builtin_raft` test suites stability, and addresses several issues uncovered in the process.
1. Correct error mapping: recoverability and various edge cases, e.g. OTX shutdown or leader stepdown.
2. Shard membership force-forgets: correct criteria to pick Ra server to address force-forgets to, more straightforward behavior.

## PR Checklist

### DIFF
--- a/apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
+++ b/apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
@@ -351,10 +351,12 @@ t_replication_transfers_snapshots(Config) ->
                 )
             end},
             {"no incomplete snapshot transfers", fun(Trace) ->
-                ?strict_causality(
-                    #{?snk_kind := dsraft_snapshot_read_started},
-                    #{?snk_kind := dsraft_snapshot_accept_complete},
-                    Trace
+                ?assert(
+                    ?strict_causality(
+                        #{?snk_kind := dsraft_snapshot_read_started},
+                        #{?snk_kind := dsraft_snapshot_accept_complete},
+                        Trace
+                    )
                 )
             end}
         ]


apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
```diff
@@ -291,60 +290,72 @@ t_replication_transfers_snapshots('end', Config) ->
...
+            ),
+
+            {NodeOffline, Nodes}
         end,
-        []
+        [
+            {"snapshot taken on live nodes", fun({NodeOffline, Nodes}, Trace) ->
+                NodesLive = Nodes -- [NodeOffline],
+                Events = ?of_kind("dsrepl_snapshot_write", Trace),
+                ?assertMatch([[_], [_]], [?of_node(N, Events) || N <- NodesLive])
```
thalesmg, line 345: Would it make sense to also `?assertMatch([], ?of_node(NodeOffline, Events))`?

keynslug: In my opinion, not much. Mostly because it's not what the testcase is concerned about: even if it's nearly impossible for `NodeOffline` to take snapshots (in the current implementation), it sounds fine if it does, for example, at the end of the test case sequence.

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_server_snapshot.erl
```diff
@@ -63,7 +63,7 @@ prepare(Index, State) ->
 ) ->
     ok | {ok, _BytesWritten :: non_neg_integer()} | {error, ra_snapshot:file_err()}.
 write(Dir, Meta, MachineState, Sync) ->
-    ?tp(dsrepl_snapshot_write, #{meta => Meta, state => MachineState}),
+    ?tp("dsrepl_snapshot_write", #{meta => Meta, state => MachineState}),
```
ieQu1: 1. I am not sure that changing atoms to strings is a good idea. E.g. if you want to filter tracepoints on prefix, snabbkaffe supports filtering by process metadata domain.
2. I would prefer dsraft prefix to dsrepl to be more precise and leave room for alternative replication implementations.

apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
```diff
@@ -291,60 +290,72 @@ t_replication_transfers_snapshots('end', Config) ->
...
+            end},
+            {"single snapshot transfer to the node that was offline", fun({NodeOffline, _}, Trace) ->
+                ?assertMatch(
+                    [#{?snk_meta := #{node := NodeOffline}}],
+                    ?of_kind("dsrepl_snapshot_accept_complete", Trace)
+                )
+            end},
+            {"no incomplete snapshot transfers", fun(Trace) ->
+                ?strict_causality(
```
ieQu1: Note: `?strict_causality` succeeds and returns `false` in case when neither causes nor effects were found in the trace. This may lead to false negatives when names or structure of trace events changes. If you are sure that the testcase emits such pairs of events, it's always advisable to wrap this macro in `?assert()`.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12694)

PR #16857, opened 2026-03-03
build: add warning as error compile option

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12521)

PR #16861, opened 2026-03-04
Sync GitHub action versions and strict erlc options everywhere

zmstone:
And strict erlc options everywhere.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12522)

PR #16863, opened 2026-03-04
fix(resource): add warning log for async reply to expired request

zmstone:

## Summary

- Included a nit-comment fix from previous PRs

- Add a warning log when an async reply arrives after the buffered request has already expired in `emqx_resource_buffer_worker`.

  Previously this path only incremented a metric counter, which made it hard to diagnose from logs why replies were dropped. This change keeps existing behavior and adds explicit warning visibility for debugging.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -1748,6 +1748,15 @@ handle_async_reply1(
             IsAcked andalso
                 begin
                     emqx_resource_metrics:late_reply_inc(Id),
+                    ?SLOG(
+                        warning,
+                        #{
+                            msg => "async_reply_to_expired_request",
+                            resource_id => Id,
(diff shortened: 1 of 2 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1748,6 +1748,15 @@ handle_async_reply1(
             IsAcked andalso
                 begin
                     emqx_resource_metrics:late_reply_inc(Id),
+                    ?SLOG(
```
keynslug, line 1751: Q: No need to throttle?

zmstone: good question.
but it was explicitly requested NOT to throttle.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12544)

PR #16877, opened 2026-03-06
build: allow make docker in git worktree

zmstone:
There is no need for it to be a git repo when building docker image (`COPY . /emqx`).
The `build` script already set and exports `PKG_VSN`, and the docker build `--build-arg` already passes `PKG_VSN`,
So all needed is to export `PKG_VSN` before `make ${PROFILE}-tgz`

This makes possible to build image from git worktree.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12562)

PR #16878, opened 2026-03-06
feat: add Gateway API HTTPRoute and TLSRoute to emqx-enterprise chart

Rory-Z:
## Summary
- Add Kubernetes Gateway API `HTTPRoute` and `TLSRoute` templates to the emqx-enterprise Helm chart as an alternative to traditional Ingress resources
- HTTPRoute supports dashboard (port 18083) and WebSocket (port 8083) routing
- TLSRoute supports MQTTS (port 8883) and WSS (port 8084) TLS passthrough routing
- All routes are disabled by default and require `parentRefs` to reference a Gateway
- Fail at Helm render time if a WS/WSS route references Service ports disabled by `service.wsEnabled`

## Test plan
- [x] `helm template` renders correctly with all routes enabled
- [x] `helm template` renders correctly with all routes disabled (default)
- [x] `helm template` fails when `httpRoute.ws` or `tlsRoute.wss` is enabled while `service.wsEnabled=false`
- [x] Verify HTTPRoute attaches to Gateway and routes dashboard traffic
- [x] Verify TLSRoute attaches to Gateway and passes through MQTTS traffic

Validated on Kind Kubernetes v1.31.2 with Envoy Gateway v1.8.3 and Gateway API v1.5.1: both routes reported `Accepted=True` and `ResolvedRefs=True`; dashboard `/status` returned HTTP 200; MQTTS completed a TLS 1.3 handshake and MQTT 5 CONNECT with CONNACK reason 0. The two invalid WS/WSS combinations fail locally with their expected Helm template errors.

### DIFF
--- a/deploy/charts/emqx-enterprise/templates/httproute.yaml
+++ b/deploy/charts/emqx-enterprise/templates/httproute.yaml
@@ -1,0 +1,72 @@
+{{- if .Values.httpRoute.dashboard.enabled }}
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: {{ printf "%s-%s" (include "emqx.fullname" .) "dashboard" }}
+  labels:
+    app.kubernetes.io/name: {{ include "emqx.name" . }}
+    helm.sh/chart: {{ include "emqx.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    {{- with .Values.httpRoute.dashboard.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.httpRoute.dashboard.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- with .Values.httpRoute.dashboard.parentRefs }}
+  parentRefs:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.httpRoute.dashboard.hostnames }}
+  hostnames:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  rules:
+    - matches:
+        - path:
+            type: {{ .Values.httpRoute.dashboard.pathType | default "PathPrefix" }}
+            value: {{ .Values.httpRoute.dashboard.path | default "/" }}
+      backendRefs:
+        - name: {{ include "emqx.fullname" . }}
+          port: {{ .Values.service.dashboard }}
+---
+{{- end }}
+{{- if .Values.httpRoute.ws.enabled }}
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: {{ printf "%s-%s" (include "emqx.fullname" .) "ws" }}
+  labels:
+    app.kubernetes.io/name: {{ include "emqx.name" . }}
+    helm.sh/chart: {{ include "emqx.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    {{- with .Values.httpRoute.ws.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.httpRoute.ws.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- with .Values.httpRoute.ws.parentRefs }}
+  parentRefs:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.httpRoute.ws.hostnames }}
+  hostnames:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  rules:
+    - matches:
+        - path:
+            type: {{ .Values.httpRoute.ws.pathType | default "PathPrefix" }}
+            value: {{ .Values.httpRoute.ws.path | default "/mqtt" }}
+      backendRefs:
+        - name: {{ include "emqx.fullname" . }}
+          port: {{ .Values.service.ws }}
+---
+{{- end }}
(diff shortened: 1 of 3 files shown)


deploy/charts/emqx-enterprise/templates/httproute.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - matches:
+        - path:
+            type: {{ .Values.httpRoute.dashboard.pathType | default "PathPrefix" }}
+            value: {{ .Values.httpRoute.dashboard.path | default "/" }}
+      backendRefs:
+        - name: {{ include "emqx.fullname" . }}
+          port: {{ .Values.service.dashboard }}
+---
+{{- end }}
+{{- if .Values.httpRoute.ws.enabled }}
```
zmstone, line 37: The `ws`/`wss` ports on the chart's Service are conditional: `templates/service.yaml` only renders them inside `{{- if .Values.service.wsEnabled }}` (default `true`). If a user has `service.wsEnabled: false` and enables `httpRoute.ws` (or `tlsRoute.wss`), this renders an HTTPRoute whose `backendRefs.port` doesn't exist on the Service — no render-time error, just a route the Gateway controller marks unresolved at runtime.

Suggest guarding: `{{- if and .Values.httpRoute.ws.enabled .Values.service.wsEnabled }}` — or better, `{{- if .Values.httpRoute.ws.enabled }}{{- if not .Values.service.wsEnabled }}{{ fail "httpRoute.ws requires service.wsEnabled=true" }}{{- end }}` so the misconfiguration fails loudly at template time. (The comment in values.yaml next to `wsEnabled` already notes this coupling for the ingress; same applies here.)

deploy/charts/emqx-enterprise/templates/tlsroute.yaml
```diff
@@ -0,0 +1,64 @@
...
+  hostnames:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  rules:
+    - backendRefs:
+        - name: {{ include "emqx.fullname" . }}
+          port: {{ .Values.service.mqttssl }}
+---
+{{- end }}
+{{- if .Values.tlsRoute.wss.enabled }}
```
zmstone, line 33: Same `service.wsEnabled` coupling as the `httpRoute.ws` comment: the `wss` Service port only exists when `service.wsEnabled=true`, so this TLSRoute can reference a nonexistent port with no render-time error. Suggest the same guard/`fail`. (The `mqtts` block above is fine — the `mqttssl` port is unconditional.)

deploy/charts/emqx-enterprise/templates/tlsroute.yaml
```diff
@@ -0,0 +1,64 @@
+{{- if .Values.tlsRoute.mqtts.enabled }}
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
zmstone: Worth a doc note in `values.yaml` for operators: `HTTPRoute` `gateway.networking.k8s.io/v1` requires the Gateway API **standard**-channel CRDs (>= v1.0), while `TLSRoute` `v1alpha2` only ships in the **experimental** channel — on a cluster with only standard-channel CRDs, enabling `tlsRoute.*` makes `helm install/upgrade` fail at apply time. Also worth one sentence that the referenced Gateway listener must be `mode: Passthrough` for these TLSRoutes (the chart terminates TLS at EMQX). Nothing to change in the templates — just document it next to the `tlsRoute` values so the failure mode isn't a surprise.

deploy/charts/emqx-enterprise/templates/httproute.yaml
```diff
@@ -0,0 +1,72 @@
...
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    {{- with .Values.httpRoute.dashboard.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- with .Values.httpRoute.dashboard.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- with .Values.httpRoute.dashboard.parentRefs }}
```
keynslug, line 19: Nit. Makes sense to emit a warning in NOTES.txt if there are zero `parentRefs`?

deploy/charts/emqx-enterprise/templates/httproute.yaml
```diff
@@ -0,0 +1,72 @@
...
+  {{- with .Values.httpRoute.dashboard.hostnames }}
+  hostnames:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  rules:
+    - matches:
+        - path:
+            type: {{ .Values.httpRoute.dashboard.pathType | default "PathPrefix" }}
+            value: {{ .Values.httpRoute.dashboard.path | default "/" }}
+      backendRefs:
```
keynslug, line 32: Q: Makes sense to support attaching user-defined `filters` here?

deploy/charts/emqx-enterprise/templates/httproute.yaml
Rory-Z: Fixed in eb983b3e0f. The chart now fails at template time when httpRoute.ws.enabled=true and service.wsEnabled=false, and values.yaml documents the coupling.

deploy/charts/emqx-enterprise/templates/tlsroute.yaml
Rory-Z: Fixed in eb983b3e0f with the same fail-fast validation for tlsRoute.wss.enabled.

deploy/charts/emqx-enterprise/templates/tlsroute.yaml
Rory-Z: Updated in eb983b3e0f for the current API: TLSRoute now uses v1, and values.yaml documents that it requires Gateway API standard-channel CRDs v1.5.0+ and a Passthrough TLS listener.

deploy/charts/emqx-enterprise/templates/httproute.yaml
Rory-Z: I would keep this out of the initial scope. parentRefs is optional in the API and the routes are disabled by default; adding the chart first NOTES warning for a valid manifest would add noise. The Gateway controller status remains the source of attachment feedback.

deploy/charts/emqx-enterprise/templates/httproute.yaml
Rory-Z: I would leave filters out of the initial scope. Filter types and support vary by Gateway implementation, and we can add this backward-compatibly once there is a concrete use case and test coverage.

keynslug (approved):

PR #16881, opened 2026-03-06
build: include plugins dir in mix format scope

zmstone:
## Summary

- Add `plugins/*/mix.exs` to `.formatter.exs` so `mix format` covers plugin mix files.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12563)

PR #16882, opened 2026-03-06
Plugin: MQTT egress bridge with disk queue

zmstone:
The native MQTT bridge (does not offer disk buffer).
This plugin is to the rescue.

### Pros

- Disk queue

### Cons

- No rule engine support (with basic topic templating, but no payload transformation)

## Future work

- ~Config management UI aid~
- ~UI for observability~ [done]
- ~Metrics scraping~ [done]

### DIFF
--- a/plugins/emqx_bridge_mqtt_dq/test/emqx_bridge_mqtt_dq_SUITE.erl
+++ b/plugins/emqx_bridge_mqtt_dq/test/emqx_bridge_mqtt_dq_SUITE.erl
@@ -1080,7 +1080,7 @@ t_sync_bridges_disable_bridge(Config) ->
     setup_config([DisabledBridge]),
     ok = emqx_bridge_mqtt_dq_app:sync_bridges(),
     ?assertEqual([], bridge_children(emqx_bridge_mqtt_dq_sup)),
-    ?assertNot(filelib:is_dir(QueueDir)),
+    ?assertNot(filelib:is_dir(filename:join(QueueDir, "dis"))),
     Snapshot = emqx_bridge_mqtt_dq_metrics:snapshot(),
     ?assertEqual(false, maps:is_key(<<"dis">>, maps:get(bridges, Snapshot, #{}))),
     ?assertEqual(


plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq.erl
```diff
@@ -0,0 +1,106 @@
...
+    emqx_bridge_mqtt_dq_buffer:enqueue(Pid, Item, no_ack);
+do_enqueue(Pid, Item, _QoS, Bridge, BridgeName, BufferIndex) ->
+    Alias = alias([reply]),
+    emqx_bridge_mqtt_dq_buffer:enqueue(Pid, Item, Alias),
+    Timeout = maps:get(enqueue_timeout_ms, Bridge, 5000),
+    receive
+        {Alias, ok} -> ok
+    after Timeout ->
+        unalias(Alias),
+        ?LOG(error, #{
```
thalesmg: maybe check mailbox once more after unaliasing to cover race condition, and clean mailbox?

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_app.erl
```diff
@@ -0,0 +1,214 @@
...
+        error ->
+            maybe_disable_and_purge_bridge(Sup, Id, DisabledBridges)
+    end.
+
+maybe_disable_and_purge_bridge(Sup, Id, DisabledBridges) ->
+    case maps:find(Id, DisabledBridges) of
+        {ok, BridgeConfig} ->
+            disable_and_purge_bridge(Sup, Id, BridgeConfig);
+        error ->
+            stop_bridge_child(Sup, Id)
```
thalesmg: Q: shouldn't running bridges that are no longer part of the config be purged too?

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_bridge_sup.erl
```diff
@@ -0,0 +1,40 @@
...
+-module(emqx_bridge_mqtt_dq_bridge_sup).
+
+-behaviour(supervisor).
+
+-export([start_link/1]).
+-export([init/1]).
+
+start_link(BridgeConfig) ->
+    #{name := Name} = BridgeConfig,
+    RegName = list_to_atom("emqx_bridge_mqtt_dq_bridge_" ++ binary_to_list(Name)),
```
thalesmg: Q: does this need to be registered?  maybe we could avoid creating atoms for this supervisor, since it doesn't seem to be needed.  otherwise, if it is, maybe `gproc` could be used.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer_sup.erl
```diff
@@ -0,0 +1,35 @@
...
+-module(emqx_bridge_mqtt_dq_buffer_sup).
+
+-behaviour(supervisor).
+
+-export([start_link/1]).
+-export([init/1]).
+
+start_link(BridgeConfig) ->
+    #{name := Name} = BridgeConfig,
+    RegName = list_to_atom("emqx_bridge_mqtt_dq_buf_" ++ binary_to_list(Name)),
```
thalesmg: Q: does this need to be registered?  maybe we could avoid creating atoms for this supervisor, since it doesn't seem to be needed.  otherwise, if it is, maybe `gproc` could be used.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_conn_sup.erl
```diff
@@ -0,0 +1,33 @@
...
+-export([start_link/1, reg_name/1]).
+-export([init/1]).
+
+start_link(BridgeConfig) ->
+    #{name := Name} = BridgeConfig,
+    supervisor:start_link({local, reg_name(Name)}, ?MODULE, [BridgeConfig]).
+
+-spec reg_name(binary()) -> atom().
+reg_name(BridgeName) ->
+    list_to_atom("emqx_bridge_mqtt_dq_conn_" ++ binary_to_list(BridgeName)).
```
thalesmg: maybe use `gproc`, or "anonymous" processes?

zmstone: ~no need for registration at all. removed.~

zmstone: fixed to use gproc.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq.erl
```diff
@@ -0,0 +1,120 @@
...
+pub_props(Headers) ->
+    maps:with(
+        [
+            'Payload-Format-Indicator',
+            'Message-Expiry-Interval',
+            'Response-Topic',
+            'Correlation-Data',
+            'User-Property',
+            'Content-Type'
+        ],
```
keynslug: Q: What's the purpose of whitelisting properties?

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq.erl
```diff
@@ -0,0 +1,120 @@
...
+        maps:get(properties, Headers, #{})
+    ).
+
+maybe_forward(#{name := Name, enable := true, filter_topic := Filter} = Bridge, Topic, QoS, Item) ->
+    case emqx_topic:match(Topic, Filter) of
```
keynslug, line 55: Q: Single topic filter per bridge is enough to satisfy requirements, as I understand?

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_app.erl
```diff
@@ -0,0 +1,222 @@
...
+
+sync_existing_child(Sup, Id, DesiredSpecs, DisabledBridges) ->
+    case maps:find(Id, DesiredSpecs) of
+        {ok, DesiredSpec} ->
+            maybe_restart_child(Sup, Id, DesiredSpec);
+        error ->
+            maybe_disable_and_purge_bridge(Sup, Id, DisabledBridges)
+    end.
+
+maybe_disable_and_purge_bridge(Sup, Id, DisabledBridges) ->
```
keynslug: Nit. Naming is slightly off IIUC, it says "maybe" but the end result is always destructive.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_app.erl
```diff
@@ -0,0 +1,222 @@
...
+                msg => "mqtt_dq_queue_dir_not_writable",
+                bridge => Name,
+                dir => Dir,
+                reason => Reason
+            }),
+            {error, {queue_dir_not_writable, Name, Dir, Reason}}
+    end.
+
+check_one_queue_dir(Dir) ->
+    case filelib:ensure_dir(filename:join(Dir, "dummy")) of
```
keynslug: Q: Why not just a single `file:write_file/2`?

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
```diff
@@ -0,0 +1,345 @@
...
+    gen_server:start_link(?MODULE, {BridgeConfig, Index}, []).
+
+-spec enqueue(pid(), queue_item(), no_ack | reference()) -> ok.
+enqueue(Pid, Item, AckAlias) ->
+    Pid ! {enqueue, Item, AckAlias},
+    ok.
+
+-spec get_pid(binary(), non_neg_integer()) -> pid().
+get_pid(BridgeName, Index) ->
+    persistent_term:get({?MODULE, BridgeName, Index}).
```
keynslug: Nit. Perhaps `gproc` is a better option for such things, as it's a process registry and tracks liveness.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
```diff
@@ -0,0 +1,345 @@
...
+    #{
+        name := BridgeName,
+        queue_dir := QueueDirBase,
+        seg_bytes := SegBytes,
+        max_total_bytes := MaxTotalBytes,
+        pool_size := PoolSize,
+        max_inflight := MaxInflight
+    } = BridgeConfig,
+    QueueDir = queue_dir(QueueDirBase, Index),
+    ok = filelib:ensure_dir(filename:join(QueueDir, "dummy")),
```
keynslug: Q: What's the purpose of this?

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
```diff
@@ -0,0 +1,345 @@
...
+    {noreply, State};
+handle_info(_Info, State) ->
+    {noreply, State}.
+
+terminate(_Reason, #{queue := Q, bridge_name := BridgeName, index := Index}) ->
+    ok = emqx_bridge_mqtt_dq_metrics:set_buffered(BridgeName, Index, 0),
+    ok = emqx_bridge_mqtt_dq_metrics:set_buffered_bytes(BridgeName, Index, 0),
+    _ = replayq:close(Q),
+    _ = persistent_term:erase({?MODULE, BridgeName, Index}),
+    ok.
```
keynslug, line 159: Q: Not going to be called? The process does not trap exits AFAICS.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_connector.erl
```diff
@@ -0,0 +1,575 @@
...
+    {reply, ok, update_metrics(State)};
+handle_call(status, _From, State) ->
+    {reply, #{connected => maps:get(connected, State, false)}, State};
+handle_call(_Request, _From, State) ->
+    {reply, {error, unknown_call}, State}.
+
+handle_info({publish_batch, Items, From, Ref}, State) ->
+    {noreply, accept_batch(Items, From, Ref, State)};
+handle_info({async_pub_ack, SeqNo, Result}, State) ->
+    {noreply, handle_async_ack(SeqNo, Result, State)};
```
keynslug, line 128: Wondering if there should be some measure against accumulating "stuck" entries in the inflight, which for whatever reason never get their ack from the broker.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq.erl
```diff
@@ -0,0 +1,120 @@
...
+    case wait_enqueue_ack(Alias, Timeout) of
+        ok ->
+            ok;
+        timeout ->
+            ?LOG(error, #{
+                msg => "mqtt_dq_enqueue_timeout",
+                bridge => BridgeName,
+                index => BufferIndex
+            }),
+            ok
```
keynslug, line 94: General question: so it's mostly about backpressure, since it doesn't affect PUBACKs getting back to the client in any way? If this is about durability, I wonder why it doesn't behave similarly to MQ, i.e.:
1. Client never gets PUBACK if the message was not durably persisted.
2. Client gets negative PUBACK if queue overflows.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq.erl
zmstone: yes, we can extend it in the future if required.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
zmstone: hmm, i remember using gproc, maybe forgot to push

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq.erl
zmstone: no good reason, really.

zmstone: maybe claude-code is being preservative

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
zmstone: ah, it was for bridge supervisor lookup, not for buffer worker. now fixing.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq.erl
zmstone: This plugin prioritizes availability over consistency.
I should consult the stakeholders if they want it otherwise.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_app.erl
zmstone: `file:write_file/2` does not create dir for you, returns `enoent`.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
zmstone: to ensure dir exists

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_connector.erl
zmstone: will add in a different PR.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_app.erl
```diff
@@ -0,0 +1,222 @@
...
+        start => {emqx_bridge_mqtt_dq_bridge_sup, start_link, [BridgeConfig]},
+        restart => permanent,
+        shutdown => infinity,
+        type => supervisor,
+        modules => [emqx_bridge_mqtt_dq_bridge_sup]
+    }.
+
+start_bridge(BridgeConfig) ->
+    ChildSpec = bridge_child_spec(BridgeConfig),
+    case supervisor:start_child(emqx_bridge_mqtt_dq_sup, ChildSpec) of
```
thalesmg: nit: maybe use `start_child/2` defined below in the file.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
keynslug: Mostly wondering if it's still necessary, as there is `emqx_bridge_mqtt_dq_app:check_queue_dirs/0` already.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_app.erl
keynslug: Ah sorry, misread the code. FWIW there's `filelib:ensure_path/1` nowadays.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
zmstone: check_queue_dirs only done at boot, the plugin can be reconfigured at runtime.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq.erl
zmstone: deleted.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq.erl
zmstone: no feedback yet, but will refactor if they want to reject PUBLISH if failed to enqueue.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
ieQu1: You can use `filelib:ensure_path` instead of using "dummy"

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_config.erl
```diff
@@ -0,0 +1,354 @@
...
+
+parse(RawConfig) when is_map(RawConfig) ->
+    RawConfig1 = resolve_envs(RawConfig),
+    RawRemotes = maps:get(<<"remotes">>, RawConfig1, #{}),
+    Remotes = parse_remotes(RawRemotes),
+    RawBridges = maps:get(<<"bridges">>, RawConfig1, #{}),
+    Bridges = parse_bridges(RawBridges, Remotes),
+    #{bridges => Bridges};
+parse(_) ->
+    #{bridges => []}.
```
thalesmg: maybe should log some error; otherwise, this will silently drop all bridges on a parse error

(similar for remotes?)

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_config.erl
thalesmg, line 1: Q: why this plugin and the others don't use hocon schemas?

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_connector.erl
```diff
@@ -0,0 +1,647 @@
...
+        proto_ver => ProtoVer,
+        clean_start => true,
+        keepalive => Keepalive,
+        max_inflight => MaxInflight,
+        force_ping => true,
+        connect_timeout => 10
+    },
+    ConnOpts1 = maybe_add_credentials(ConnOpts, Username, Password),
+    ConnOpts2 = maybe_add_ssl(ConnOpts1, Host, SslConf),
+    process_flag(trap_exit, true),
```
thalesmg: duplicated

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_buffer.erl
zmstone: fixed to call ensure_path

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_config.erl
zmstone: this plugin has no avro schema yet because the config editor in dashboard does not support map of objects yet.
plugin in general do not support hocon schema because it's non-standard and hard to people to adopt (so picked avro schema instead)

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_connector.erl
```diff
@@ -0,0 +1,647 @@
...
+%%
+%% Backlog size is bounded by buffer-side flow control:
+%% each buffer sends at most MAX_INFLIGHT_BATCHES (2) batches before
+%% waiting for acks, each batch has at most max_inflight items.
+%% With K buffers per connector (buffer_pool_size / pool_size),
+%% the worst-case backlog is K * 2 * max_inflight items.
+accept_batch(Items, From, Ref, State) ->
+    {Entries, State1} = assign_seqnos(Items, Ref, State),
+    SeqNos = [SeqNo || {SeqNo, _, _, _} <- Entries],
+    Pending = maps:from_list([{S, []} || S <- SeqNos]),
```
thalesmg: tiny nit:


```suggestion
    Pending = maps:from_keys(SeqNos, []),
```

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_config.erl
thalesmg: it seems to re-implement facilities that come from hocon (parsing raw configs, validation, parsing types like byte size, defaults, ...), and this module does not seem to depend on avro directly for the parse path.

also, there is a `plugins/emqx_bridge_mqtt_dq/priv/config.hocon` file, so hocon is still used, at least under the hood...  so maybe it could be leveraged to avoid re-inventing those features?

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_config.erl
zmstone: deleted the non-map clauses, will let it crash.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_config.erl
zmstone: config.hocon can be json inside as content, `hocon` app will load either format (JSON | HOCON).
but it has to be an avro json for avoid schema checks.

avro schema check is not just for backend, it is also for frontend to render the config editor UI.
--- another reason why avro was picked in the first place, because there are avro typescript libs, but nothing for our home-made hocon schema (JSON dump).

I agree it makes sense to support hocon schema in plugins, but not in the scope of this plugin.

thalesmg: I'm still confused what this has to do with avro.

This module has a `parse/1` function that receives an Erlang map (hence potentially already parsed from avro, or from HTTP API) of configs, and does parsing, validation, and eventually produces a map with atom keys.  All this is quite close to what hocon already does.

My original question was:  why not define a hocon schema and just use hocon's machinery to do all this, instead of creating those `get_nested`, `to_boolean`, converter logic for things like sni, etc. ?

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_connector.erl
```diff
@@ -0,0 +1,646 @@
...
+    end.
+
+handle_item_result(SeqNo, BatchRef, Item, Retries, Result, State) ->
+    case should_retry(Result) of
+        false ->
+            #{bridge_name := BridgeName} = State,
+            ok = emqx_bridge_mqtt_dq_metrics:incr_bridge_publish(BridgeName, 1),
+            finish_item(SeqNo, BatchRef, remove_inflight(SeqNo, State));
+        true ->
+            case should_retry_again(Retries) of
```
thalesmg, line 283: Nit: naming is a bit confusing




```suggestion
    case is_success(Result) of
        true ->
            #{bridge_name := BridgeName} = State,
            ok = emqx_bridge_mqtt_dq_metrics:incr_bridge_publish(BridgeName, 1),
            finish_item(SeqNo, BatchRef, remove_inflight(SeqNo, State));
        false ->
            case should_retry(Retries) of
```

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_metrics.erl
```diff
@@ -0,0 +1,511 @@
...
+    Acc#{connectors := Connectors#{{BridgeName, Index} => ConnectorMetrics1}};
+fold_row({{connector, BridgeName, Index, connected}, Value}, #{connectors := Connectors} = Acc) ->
+    ConnectorMetrics0 = maps:get(
+        {BridgeName, Index}, Connectors, #{backlog => 0, inflight => 0, connected => 0}
+    ),
+    ConnectorMetrics1 = ConnectorMetrics0#{connected => Value},
+    Acc#{connectors := Connectors#{{BridgeName, Index} => ConnectorMetrics1}};
+fold_row({{connector, BridgeName, Index, inflight}, Value}, #{connectors := Connectors} = Acc) ->
+    ConnectorMetrics0 = maps:get(
+        {BridgeName, Index}, Connectors, #{backlog => 0, inflight => 0, connected => 0}
```
thalesmg, line 213: nit: maybe define `empty_metrics/0` and `empty_buffer_metrics/0` to replace those repeated default values above and below this line.

plugins/emqx_bridge_mqtt_dq/src/emqx_bridge_mqtt_dq_api.erl
```diff
@@ -0,0 +1,495 @@
...
+    {ok, 200,
+        #{
+            <<"content-type">> => <<"text/html; charset=utf-8">>,
+            <<"cache-control">> => <<"no-store, no-cache, must-revalidate">>,
+            <<"pragma">> => <<"no-cache">>,
+            <<"expires">> => <<"0">>
+        },
+        ui_html()};
+handle(_Method, _Path, _Request) ->
+    {error, not_found}.
```
thalesmg, line 40: Q: the readme suggests configuring bridges via the dashboard, but there seems to be only `get` endpoints?

zmstone: config update is a part of `emqx_plugins` framework.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12693)

PR #16886, opened 2026-03-08
ci: handle empty workflow matrices in GitHub Actions

zmstone:
## Summary
- make `run_test_cases` skip `eunit_and_proper` cleanly when `ct-matrix` is empty
- simplify `static_checks` to always run with the `emqx-enterprise` profile
- remove the unused `ct-matrix` input from the `static_checks` reusable workflow callers


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12567)

PR #16887, opened 2026-03-08
feat: add MQTT subscription filters

zmstone:
Fixes [EMQX-15161](https://emqx.atlassian.net/browse/EMQX-15161)

Release version: 6.2.0

## Summary

- Adds optional MQTT subscription filters controlled by `mqtt.subscription_filter`.
- When enabled, a subscription such as `sensor/+/temperature?location=roomA&value>25` is validated and stored against the base topic filter while the predicate is evaluated from MQTT 5 `User-Property` entries during delivery.
- When disabled, `?` remains part of the topic filter text so existing behavior is preserved.
- Includes schema, example config, and coverage for packet validation, channel/session handling, schema translation, and MQTT v5 protocol behavior.

## PR Checklist

[EMQX-15161]: https://emqx.atlassian.net/browse/EMQX-15161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds/DurableSession.asn
+++ b/apps/emqx/src/emqx_persistent_session_ds/DurableSession.asn
@@ -24,20 +24,23 @@ BEGIN
     gt(1),
     lt(2),
     gte(3),
-    lte(4)
+    lte(4),
+    ...
   }
 
   FilterClause ::= SEQUENCE {
     op [0] FilterOp,
     key [1] OCTET STRING,
     stringValue [2] OCTET STRING OPTIONAL,
-    numberValue [3] REAL OPTIONAL
+    numberValue [3] REAL OPTIONAL,
+    ...
   }
 
   SubscriptionFilter ::= SEQUENCE {
     raw [0] OCTET STRING,
     source [1] OCTET STRING,
-    clauses [2] SEQUENCE OF FilterClause
+    clauses [2] SEQUENCE OF FilterClause,
+    ...
   }
 
   SubState ::= SEQUENCE {


apps/emqx/src/emqx_channel.erl
```diff
@@ -3114,8 +3124,28 @@ maybe_shutdown(Reason, _Intent = shutdown, Channel) ->
...
+                    sub_filter_ast => AST,
+                    sub_filter_enabled => true,
+                    sub_filter_source => TopicFilter
+                }}
+            );
+        {error, malformed_filter} ->
+            error({invalid_subscription_filter, TopicFilter})
+    end;
+parse_raw_topic_filter(TopicFilter, Mode) ->
+    emqx_topic:parse(emqx_subscription_filter:normalize_unsubscribe(TopicFilter, Mode)).
```
thalesmg: nit: this is called `normalize_unsubscribe`, but `parse_raw_topic_filter{,s}` is called during `do_enrich_subscribe` only, apparently?

apps/emqx/src/emqx_session.erl
```diff
@@ -490,6 +490,13 @@ enrich_message(
 ) ->
     _ = emqx_session_events:handle_event(ClientInfo, {dropped, Msg, no_local}),
     [];
+enrich_message(ClientInfo, MsgIn, SubOpts = #{sub_filter_ast := AST}, UpgradeQoS) ->
+    case emqx_subscription_filter:match_message(AST, MsgIn) of
+        true ->
+            enrich_message(ClientInfo, MsgIn, clean_sub_filter_subopts(SubOpts), UpgradeQoS);
```
thalesmg, line 501: Q: shouldn't check if `sub_filter_enabled := true` too?

apps/emqx/src/emqx_subscription_filter.erl
```diff
@@ -0,0 +1,188 @@
...
+-type ast() :: [clause()].
+
+-export_type([mode/0, ast/0]).
+
+-spec split_topic_filter(
+    emqx_types:topic() | emqx_types:share(), mode()
+) -> {emqx_types:topic() | emqx_types:share(), no_filter | {filter, binary()}}.
+split_topic_filter(#share{group = Group, topic = TopicFilter}, Mode) ->
+    {BaseTopic, Filter} = split_topic_filter(TopicFilter, Mode),
+    {#share{group = Group, topic = BaseTopic}, Filter};
```
thalesmg: maybe:


```suggestion
split_topic_filter(#share{topic = TopicFilter} = SharedSub, Mode) ->
    {BaseTopic, Filter} = split_topic_filter(TopicFilter, Mode),
    {SharedSub#share{topic = BaseTopic}, Filter};
```

?

apps/emqx/src/emqx_subscription_filter.erl
```diff
@@ -0,0 +1,188 @@
...
+    {TopicFilter, no_filter};
+split_topic_filter(TopicFilter, enable) when is_binary(TopicFilter) ->
+    case binary:match(TopicFilter, <<"?">>) of
+        nomatch ->
+            {TopicFilter, no_filter};
+        {Pos, 1} ->
+            {
+                binary:part(TopicFilter, 0, Pos),
+                {filter, binary:part(TopicFilter, Pos + 1, byte_size(TopicFilter) - Pos - 1)}
+            }
```
thalesmg, line 55: this makes it impossible to use `?` in other segments, even though they wouldn't/shouldn't affect the filter.

e.g.: `{3, 1} = binary:match(<<"a/b?c/d?e">>, <<"?">>)`, when we actually wanted `{7, 1}`.

not sure if intentional.

apps/emqx/src/emqx_subscription_filter.erl
```diff
@@ -0,0 +1,188 @@
...
+            {ok, Number}
+    catch
+        error:badarg ->
+            {error, bad_number}
+    end;
+maybe_number(Value) ->
+    maybe_number(emqx_utils_conv:bin(Value)).
+
+trim(Bin) ->
+    unicode:characters_to_binary(string:trim(Bin)).
```
thalesmg: maybe assert it's not `{error | incomplete, _, _}` so it doesn't blow up in strange ways deeper in `parse_clauses`?

apps/emqx/src/emqx_session.erl
```diff
@@ -490,6 +490,13 @@ enrich_message(
...
+        false ->
+            []
```
thalesmg, line 504: Q: should it emit some metrics or events like `emqx_session_events:handle_event`?

apps/emqx/src/emqx_subscription_filter.erl
zmstone: scanning from left to right is intentional

apps/emqx/src/emqx_subscription_filter.erl
zmstone: changed to a guard assert, Bin is always a binary, and string:trim should always return binary too.

apps/emqx/src/emqx_session.erl
zmstone: sub_filter_ast's existence already indicates it's enabled ?

thalesmg: then why does `sub_filter_enabled` exist?

apps/emqx/src/emqx_metrics.erl
```diff
@@ -579,6 +579,7 @@ reserved_idx('delivery.dropped.too_large') -> 63;
 reserved_idx('delivery.dropped.qos0_msg') -> 64;
 reserved_idx('delivery.dropped.queue_full') -> 65;
 reserved_idx('delivery.dropped.expired') -> 66;
+reserved_idx('delivery.dropped.subscription_filter') -> 102;
```
thalesmg: position of this clause makes further updates to this function clunkier...  would have to scan all clauses and find the max, instead of just checking the last clause.  😓

apps/emqx/test/emqx_persistent_session_SUITE.erl
```diff
@@ -456,6 +456,96 @@ t_connect_session_expiry_interval_qos2(Config) ->
...
+The session subscribes with a filter, goes offline, accumulates one non-matching
+message followed by one matching message, and then reconnects. Only the matching
+message should be replayed, and subsequent matching live messages should still
+be delivered.
+""".
+t_connect_session_expiry_interval_subscription_filter(Config) ->
+    case ?config(persistence, Config) of
+        false ->
+            {skip, persistence_disabled};
+        ds ->
```
thalesmg: nit: feels like this belongs to the `init` setup?

rel/config/examples/mqtt.conf.example
```diff
@@ -39,6 +39,9 @@ mqtt {
     ## Whether to enable support for MQTT exclusive subscription
     exclusive_subscription = false
 
+    ## Whether to enable support for EMQX subscription filters (`topic?key=value`)
+    subscription_filter = disable
```
ieQu1: I think `subscription_filter` name is too generic for this feature. If I read the description correctly, it's about letting subscribers filter messages via user properties, but `mqtt.subscription_filter` title doesn't capture that.

apps/emqx/test/emqx_persistent_session_SUITE.erl
```diff
@@ -456,6 +456,96 @@ t_connect_session_expiry_interval_qos2(Config) ->
...
+            OldMode = emqx:get_config([mqtt, subscription_filter], disable),
+            emqx_config:put([mqtt, subscription_filter], enable),
+            [{old_subscription_filter, OldMode} | Config]
+    end;
+t_connect_session_expiry_interval_subscription_filter('end', Config) ->
+    emqx_config:put([mqtt, subscription_filter], ?config(old_subscription_filter, Config)).
+
+-doc """
+Verify that a durable session keeps replay progress when subscription filtering
```
ieQu1:
```suggestion
Verify that a persistent session keeps replay progress when subscription filtering
```
Note: this test suite is not only about durable sessions.

apps/emqx/src/emqx_session.erl
```diff
@@ -490,6 +490,14 @@ enrich_message(
 ) ->
     _ = emqx_session_events:handle_event(ClientInfo, {dropped, Msg, no_local}),
     [];
+enrich_message(ClientInfo, MsgIn, SubOpts = #{sub_filter_ast := AST}, UpgradeQoS) ->
```
ieQu1, line 498: This part looks extremely dangerous for durable sessions, which rely on deterministic nature of this function. Since this code adds an external call to `emqx_subscription_filter:match_message`, it means the latter must be totally deterministic too (i.e., it can't change between versions).

ieQu1: Also please don't abuse the "ASN.1 schema supports arbitrary subscription properties" loophole. It was *not* left to inject new properties without recording their types in the schema, it was added there due to loose typing at the time. The proper way to add a subscription property is by adding it to DurableSession.asn.

apps/emqx/src/emqx_subscription_filter.erl
```diff
@@ -0,0 +1,188 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2017-2026 EMQ Technologies Co., Ltd. All Rights Reserved.
```
ieQu1: Wrong date?

apps/emqx/src/emqx_subscription_filter.erl
```diff
@@ -0,0 +1,188 @@
...
+mk_clause(<<"=">>, Key, Value) ->
+    {ok, {eq, Key, Value}};
+mk_clause(<<">">>, Key, Value) ->
+    mk_numeric_clause(gt, Key, Value);
+mk_clause(<<"<">>, Key, Value) ->
+    mk_numeric_clause(lt, Key, Value);
+mk_clause(<<">=">>, Key, Value) ->
+    mk_numeric_clause(gte, Key, Value);
+mk_clause(<<"<=">>, Key, Value) ->
+    mk_numeric_clause(lte, Key, Value).
```
ieQu1, line 155: Is there an (in)formal specification of the filter language anywhere? If grammar was defined in BNF, we could publish the BNF as a reference for the users. Not perfect, but passable. 
But here grammar is defined in ad-hoc code, making it unclear e.g. how technical writers can describe the filter language to the users.

apps/emqx/test/emqx_mqtt_protocol_v5_SUITE.erl
```diff
@@ -913,6 +913,90 @@ t_publish_properties(Config) ->
     ?assertEqual(Properties, maps:get(properties, Msg1)),
     ok = emqtt:disconnect(Client1).
 
+t_subscription_filter(init, Config) ->
```
ieQu1, line 916: Can you please add a short human-readable comment describing the testcase; e.g. what property does it verify?

apps/emqx/src/emqx_subscription_filter.erl
```diff
@@ -0,0 +1,188 @@
...
+-include_lib("emqx_utils/include/emqx_message.hrl").
+
+-export([
+    split_topic_filter/2,
+    validate_subscription/2,
+    normalize_topic_filter/2,
+    match_message/2
+]).
+
+-type mode() :: enable | disable.
```
ieQu1: `boolean()`?

ieQu1:
I think this PR may be missing tests related to AuthZ. The new syntax of the topics creates non-trivial interactions with AuthZ flows concerning topics. It is necessary to create a regression test verifying that manipulations with filters don't let the client bypass the auth.

apps/emqx/src/emqx_subscription_filter.erl
```diff
@@ -0,0 +1,188 @@
...
+        {Pos, 1} ->
+            {
+                binary:part(TopicFilter, 0, Pos),
+                {filter, binary:part(TopicFilter, Pos + 1, byte_size(TopicFilter) - Pos - 1)}
+            }
+    end.
+
+-spec validate_subscription(emqx_types:topic() | emqx_types:share(), mode()) ->
+    {ok, #{
+        mode := plain | filtered,
```
ieQu1: It's a bit confusing that this module also declares a type called `mode()`, which also acts as a boolean of sorts, but with different values.

rel/config/examples/mqtt.conf.example
zmstone: good point, maybe `mqtt.subscription_message_filter` ?

apps/emqx/src/emqx_session.erl
zmstone: i try to add filter to the durable schema.

apps/emqx/src/emqx_persistent_session_ds/DurableSession.asn
```diff
@@ -19,6 +19,27 @@ BEGIN
     val [1] OCTET STRING
   }
 
+  FilterOp ::= ENUMERATED {
+    eq(0),
+    gt(1),
+    lt(2),
+    gte(3),
+    lte(4)
+  }
```
ieQu1: `, ...`?

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12907)

PR #16888, opened 2026-03-08
chore: generate new default certs

zmstone:
Release version: 6.2.0

Generated new default certificates.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12556)

PR #16897, opened 2026-03-10
fix(username-quota): in case of 503, there is no need for retry cursor

zmstone:
## Summary

Username-quota plugin.
In version 1.1.1, we changed for first core node to always the owner of the snapshot,
meaning the next retry will likely go to the owner without any help from the cursor.
so the `retry_cursor` in 503 is now obsolete.

## PR Checklist


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12696)

PR #16634, opened 2026-01-26
260126 external schema reg support get

zmstone:
Fixes [EMQX-15017](https://emqx.atlassian.net/browse/EMQX-15017)

Release version: 6.2.0

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

[EMQX-15017]: https://emqx.atlassian.net/browse/EMQX-15017?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #16903, opened 2026-03-11
fix(schemareg): use url-safe base64 encoding for GET query string payload

zmstone:
Fix for #16634: the GET method support added in that PR used standard base64 encoding for the payload query parameter, but standard base64 contains `+`, `/`, and `=` characters that are not URL-safe. This changes it to use url-safe base64 encoding (`#{mode => urlsafe, padding => false}`).

The test server handler is also updated to decode url-safe base64 for GET requests.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12572)

PR #16904, opened 2026-03-11
fix(plugins): enforce single enabled version per plugin

zmstone:
Fixes 

## Summary

Prevent multiple versions of the same plugin from being enabled or started at the same time.

- Normalize plugin state so only the latest enabled version of a plugin remains enabled in configuration.
- Return explicit management API errors when another version of the same plugin is already loaded or running.
- Add plugin tests covering version normalization and per-version running status detection.

## PR Checklist

### DIFF
--- a/apps/emqx_plugins/src/emqx_plugins.erl
+++ b/apps/emqx_plugins/src/emqx_plugins.erl
@@ -1294,6 +1294,10 @@ version_selection_helpers_case() ->
     ),
     ?assertEqual(
         <<"demo-1.0.0">>,
+        emqx_plugins_utils:latest_name_vsn(<<"demo-1.0.0">>, <<"demo-1.0.0">>)
+    ),
+    ?assertEqual(
+        <<"demo-2.0.0">>,
         emqx_plugins_utils:latest_name_vsn(<<"demo-2.0.0">>, <<"demo-1.0.0">>)
     ).
 


apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -976,3 +982,45 @@ add_health_status(StatusInfo, #{health_status := HealthStatus}) ->
     StatusInfo#{health_status => HealthStatus};
 add_health_status(StatusInfo, _) ->
     StatusInfo.
+
+-ifdef(TEST).
```
savonarola, line 1000: Nit: do we really need these tests 🤔

apps/emqx_plugins/src/emqx_plugins_info.erl
```diff
@@ -165,8 +154,115 @@ check_plugin(PluginInfo, NameVsn) ->
     }}.
 
 configured() ->
-    emqx_conf:get([?CONF_ROOT, states]).
+    lists:map(fun normalize_state_item/1, emqx_conf:get([?CONF_ROOT, states])).
+
```
savonarola, line 158:
```suggestion

-doc """
If `NameVsn` plugin has an enabled configuration we still consider it enabled
only if its `Vsn` is greater than any other enabled version of the same plugin.
""".
```

apps/emqx_plugins/src/emqx_plugins_info.erl
```diff
@@ -165,8 +154,115 @@ check_plugin(PluginInfo, NameVsn) ->
     }}.
 
 configured() ->
```
savonarola, line 156: Nit: maybe

```suggestion
configured() ->
  configured(emqx_conf:get([?CONF_ROOT, states])).

configured(States) ->
```

and get rid of mocking in eunit

apps/emqx_plugins/src/emqx_plugins_info.erl
```diff
@@ -165,8 +154,115 @@ check_plugin(PluginInfo, NameVsn) ->
...
+    NameVsnBin = bin(NameVsn),
+    ExactStates = [
+        Enabled
+     || #{name_vsn := NV, enable := Enabled} <- Configured, bin(NV) =:= NameVsnBin
+    ],
+    case ExactStates of
+        [] ->
+            not_configured;
+        _ ->
+            case lists:any(fun(Enabled) -> Enabled =:= true end, ExactStates) of
```
savonarola:
```suggestion
            %% Only single state in ExactStates is expected actually
            case lists:any(fun(Enabled) -> Enabled =:= true end, ExactStates) of
```

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -894,6 +910,99 @@ for_plugins(ActionFun) ->
...
+        fun
+            (#{name_vsn := NameVsn, enable := true}, Acc) ->
+                Name = plugin_name(NameVsn),
+                NameVsnBin = bin(NameVsn),
+                case maps:get(Name, Acc, undefined) of
+                    undefined ->
+                        Acc#{Name => NameVsnBin};
+                    Existing ->
+                        case is_newer_name_vsn(NameVsnBin, Existing) of
+                            true -> Acc#{Name => NameVsnBin};
```
savonarola: Nit: a bit difficult to perceive which argument is greater of which. M.b. would be nice to have smth. like:
```suggestion
                        case compare_vsn(NameVsnBin, Existing) of
                            gt -> Acc#{Name => NameVsnBin};
```

zmstone: will try NOT to swap the order of the args, but keep using the existing `newer | older | same` return values.

apps/emqx_plugins/src/emqx_plugins_info.erl
zmstone: tried to add more details.

apps/emqx_plugins/src/emqx_plugins_info.erl
zmstone: unsure, if one configures (manually) multiple states of the same name-vsn ? 
changed it to use `lists:member(true, ExactStates)`

zmstone:
@savonarola since your last review, i found another issue: the change was too strict, resulted to disallow multiple versions of the same plugins to even install.
now multiple versions are allowed to be installed, but only one version can be activated.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12807)

PR #16914, opened 2026-03-13
fix(mqtt): add peerport to clientinfo

zmstone:
Fixes https://emqx.atlassian.net/browse/EMQX-15177

Release version:
5.8.10, 5.9.3, 5.10.4

## Summary

Expose `peerport` in MQTT `clientinfo` so hooks and integrations can consume the remote port without unpacking `peername`.

The change updates `emqx_channel:init/2` to persist the port alongside the existing peer address fields and extends `emqx_channel_SUITE` to cover the new field.

## PR Checklist

### DIFF
--- a/changes/ee/fix-16914.en.md
+++ b/changes/ee/fix-16914.en.md
@@ -1,5 +1,5 @@
-Fixed an issue where the MQTT client information exposed to hooks and other integrations did not include the client's remote port.
+Added `${peerport}` to the client information available in authentication and authorization request templates.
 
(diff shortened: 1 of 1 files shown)


zmstone:
Closing this for `release-58`.

After reviewing the change against the bugfix-release scope, making `${peerport}` available in authentication and authorization request templates would require adding a new allowed template variable as well as providing its runtime value. That is feature work rather than a pure bugfix on `5.8`.

The smaller `clientinfo.peerport` change by itself does not make `${peerport}` usable in authn/authz templates, so the PR no longer matches the intended user-facing behavior.

We can revisit this in a later minor release and discuss `5.10` separately.

PR #16915, opened 2026-03-13
fix(authz): allow propagating peerport to HTTP authorization templates

zmstone:
Fixes https://emqx.atlassian.net/browse/EMQX-15177

Release version: 5.10.4

## Summary

The original report was that `peerport` was not available for HTTP authorization request templates.
This change exposes `peerport` in MQTT client info before `client_attrs_init` expressions are evaluated, so users can initialize a client attribute from `peerport` and reference that client attribute in HTTP authorization templates.
This does not add a new direct HTTP template variable; it enables the reported use case through the existing `client_attrs_init` path.
The change also extends `emqx_channel_SUITE` expectations for client info to cover the new field.

## PR Checklist

### DIFF
--- a/changes/ee/fix-16915.en.md
+++ b/changes/ee/fix-16915.en.md
@@ -1,1 +1,1 @@
-Made `peerport` available in `mqtt.client_attrs_init` expressions so it can be used when initializing client attributes.
+Made `peerport` available in `mqtt.client_attrs_init` and `${peerport}` in authentication/authorization request templates.


qzhuyan:
Will be more confident to have a test case cover it with auth.

zmstone:
the reporter thought `peerport` is the listener's port.
what they actually want is an identifier of listener which accepted the client. for that, there is `${listener}`.

PR #16917, opened 2026-03-13
fix(mqtt-bridge-dq): add node name (hash) to client id prefix

zmstone:
fix client id clash when clustered.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12701)

PR #16926, opened 2026-03-16
feat(limiter): add `delivery_messages` and `delivery_bytes` limiter kinds

thalesmg:
Fixes <issue-or-jira-number>

Release version:
6.1.2, 6.2.0

## Summary

We already have _outgoing_ rate limit on message number and size.  Here, _outgoing_ is from the point of view of the client: we rate limit the `PUBLISH` packets in the publisher client context, hence the limiter tokens consumed are from this publisher.

The new requirement is to impose rate limiting on the _incoming_ message rate and size: the _deliveries_ a subscriber client channel receives.

For this, we introduce here two new limiter kinds: `delivery_messages` and `delivery_bytes`.  They are checked before the delivery is turned into a packet, and before executing the `'message.delivered'` hook, on a one-by-one basis.  If a delivery hits the limit, we drop it and it does not get sent on the wire.

On the one hand, this may pose a problem if some user sets a finite `delivery_{messages,bytes}` limit, but sets `mqtt.retry_interval = infinity` (which, unfortunately, is the default value).  In this case, the channel will get stuck if a QoS 1 or 2 delivery is dropped.  On the other hand, if `mqtt.retry_interval` is finite, the client channel will eventually recover and deliver all QoS > 0 messages naturally, and drop QoS 0 ones without retrying.

If unspecified, the default values are unlimited, thus keeping backwards compatibility.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -1431,7 +1431,48 @@ wrap_list(X) ->
 %% return list(emqx_types:packet())
 do_deliver({pubrel, PacketId}, Channel) ->
     {[?PUBREL_PACKET(PacketId, ?RC_SUCCESS)], Channel};
-do_deliver(
+do_deliver({_PacketId, Msg} = Publish, Channel0) ->
+    #channel{quota = Limiter0} = Channel0,
+    #message{payload = Payload, topic = Topic} = Msg,
+    Result = emqx_limiter_client_container:try_consume(
+        Limiter0,
+        [
+            {delivery_bytes, iolist_size(Payload)},
+            {delivery_messages, 1}
+        ]
+    ),
+    case Result of
+        {true, Limiter} ->
+            Channel = Channel0#channel{quota = Limiter},
+            do_deliver1(Publish, Channel);
+        {false, Limiter, Reason} ->
+            ?SLOG_THROTTLE(
+                warning,
+                #{
+                    msg => cannot_deliver_from_topic_due_to_quota_exceeded,
+                    reason => Reason
+                },
+                #{topic => Topic, tag => "QUOTA"}
+            ),
+            ?tp("chan_delivery_rate_limit", #{limiter => Limiter, reason => Reason}),
+            Channel = Channel0#channel{quota = Limiter},
+            {[], Channel}
+    end;
+do_deliver([Publish], Channel) ->
+    do_deliver(Publish, Channel);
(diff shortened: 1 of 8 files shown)


rel/i18n/emqx_limiter_schema.hocon
```diff
@@ -67,6 +67,30 @@ For example: `100MB/60m`: Once every 60 minutes, up to 100 megabytes can be sent
...
+"""Data Delivery Burst"""
+
+delivery_messages_rate.desc:
+"""Todo."""
+
+delivery_messages_rate.label:
+"""Message Delivery Rate"""
+
+delivery_messages_burst.desc:
+"""Todo."""
```
thalesmg, line 89: This is an intentional placeholder.

I'm looking for feedback on the actual implementation first, and will fix this once consensus is reached.

apps/emqx/src/emqx_channel.erl
```diff
@@ -1431,7 +1431,48 @@ wrap_list(X) ->
...
+    ),
+    case Result of
+        {true, Limiter} ->
+            Channel = Channel0#channel{quota = Limiter},
+            do_deliver1(Publish, Channel);
+        {false, Limiter, Reason} ->
+            ?SLOG_THROTTLE(
+                warning,
+                #{
+                    msg => cannot_deliver_from_topic_due_to_quota_exceeded,
```
thalesmg, line 1452: todo: register this throttled log

thalesmg:
Looking into how these could be implemented, but there might be a few concerns.

For DS sessions, I think the proposed changes (buffering until budget becomes available) could break protocol, in particular ordering of messages from the same publisher clientid.  Not to mention introduce a kind of extra state machine for the channel/session?

The order breaking could happen like:

1) Publisher `A` publishes message `1` on topic `t`, which subscriber `B` subscribes to.  DS sessions are in use.
2) `B` receives the deliver, but it's rate limited, so it gets buffered.
3) `A` publishes a new `2` message on `t`.  This time, `B` receives the delivery but is not rate limited, so `2` goes into the wire.
4) Later, the buffered `1` message is unblocked and sent on the wire, after `2`.

thalesmg:
Also, the current implementation here is quite simple and relies only on the existing `retry_interval` timer.  What would be the gains of the other proposed implementation, given it seems to (apparently) introduce far more complexity and a new timer?

thalesmg:
https://github.com/emqx/emqx/pull/16929

PR #16929, opened 2026-03-17
feat(limiter): add `delivery_messages` and `delivery_bytes` limiter kinds (mem session only)

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-15195

Release version:
6.1.1, 6.2.0

## Summary

We already have _outgoing_ rate limit on message number and size.  Here, _outgoing_ is from the point of view of the client: we rate limit the `PUBLISH` packets in the publisher client context, hence the limiter tokens consumed are from this publisher.

The new requirement is to impose rate limiting on the _incoming_ message rate and size: the _deliveries_ a subscriber client channel receives.

For this, we introduce here two new limiter kinds: `delivery_messages` and `delivery_bytes`.  They are only supported for memory sessions for now.  The limiters are checked when dequeuing a message from `mqueue` and when delivering messages to the session.  If the limit is hit, QoS 0 messages are dropped, QoS > 0 are queued in `mqueue`, and a retry is scheduled by the channel.  The retry time is derived from the limiter's configuration.

If unspecified, the default values are unlimited, thus keeping backwards compatibility.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -145,9 +145,12 @@
     ((ConnState == connected) orelse (ConnState == reauthenticating))
 ).
 
+%% Used by mem sessions
+-define(RETRY_DEQUEUE_TIMER, retry_dequeue).
+
 %% Timers implemented by sessions
 -define(IS_COMMON_SESSION_TIMER(N),
-    ((N == retry_delivery) orelse (N == expire_awaiting_rel))
+    ((N == retry_delivery) orelse (N == expire_awaiting_rel) orelse (N == ?RETRY_DEQUEUE_TIMER))
 ).
 %% Timers implemented by sessions that need to be handled only when the client is connected
 -define(IS_COMMON_SESSION_ONLINE_TIMER(N),
@@ -157,7 +160,6 @@
 -define(chan_terminating, chan_terminating).
 -define(normal, normal).
 -define(RAND_CLIENTID_BYTES, 16).
--define(RETRY_DEQUEUE_TIMER, retry_dequeue).
 
 -dialyzer({no_match, [shutdown/4, ensure_timer/2, interval/2]}).
 
@@ -1841,20 +1843,24 @@ handle_timeout(
 handle_timeout(
     _TRef,
     TimerName,
-    Channel = #channel{session = Session, clientinfo = ClientInfo}
+    Channel0 = #channel{session = Session, clientinfo = ClientInfo}
 ) when ?IS_COMMON_SESSION_TIMER(TimerName) ->
     %% NOTE
     %% Responsibility for these timers is smeared across both this module and the
     %% `emqx_session` module: the latter holds configured timer intervals, and is
     %% responsible for the actual timeout logic. Yet they are managed here, since
     %% they are kind of common to all session implementations.
+    Channel1 = clean_timer(TimerName, Channel0),
+    stash_limiter_ctx(Channel1),
     case emqx_session:handle_timeout(ClientInfo, TimerName, Session) of
         {ok, Publishes, NSession} ->
-            NChannel = Channel#channel{session = NSession},
-            handle_out(publish, Publishes, clean_timer(TimerName, NChannel));
+            Channel2 = pop_limiter_ctx(Channel1),
+            Channel = Channel2#channel{session = NSession},
+            handle_out(publish, Publishes, clean_timer(TimerName, Channel));
         {ok, Publishes, Timeout, NSession} ->
-            NChannel = Channel#channel{session = NSession},
-            handle_out(publish, Publishes, reset_timer(TimerName, Timeout, NChannel))
+            Channel2 = pop_limiter_ctx(Channel1),
+            Channel = Channel2#channel{session = NSession},
+            handle_out(publish, Publishes, reset_timer(TimerName, Timeout, Channel))
     end;
 handle_timeout(
     _TRef,


rel/i18n/emqx_limiter_schema.hocon
```diff
@@ -70,4 +70,28 @@ bytes_burst.label:
 mqtt.desc:
 """MQTT related limiters."""
 
+delivery_bytes_rate.desc:
+"""Todo."""
```
thalesmg: Intentional placeholder; will be fixed once implementation is agreed upon.

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -490,22 +493,34 @@ dequeue(ClientInfo, Session = #session{inflight = Inflight, mqueue = Q}) ->
...
+do_dequeue(Msg, ClientInfo = #{zone := Zone}, Cnt, Msgs, Q, PrevQ) ->
+    case emqx_message:is_expired(Msg, Zone) of
+        true ->
+            _ = emqx_session_events:handle_event(ClientInfo, {expired, Msg}),
+            dequeue(ClientInfo, Cnt, Msgs, Q);
+        false ->
+            case try_consume_delivery_rate_limit(Msg) of
+                ok ->
+                    dequeue(ClientInfo, acc_cnt(Msg, Cnt), [Msg | Msgs], Q);
+                {error, Reason} ->
```
zmstone, line 520: check if Msg is QoS 0, drop it if yes.

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -884,6 +919,66 @@ publish_will_message_now(#session{} = Session, #message{} = WillMsg) ->
...
+%%--------------------------------------------------------------------
+
+try_consume_delivery_rate_limit(Msg) ->
+    case emqx_session:get_context() of
+        #{limiter := Limiter0} = Ctx0 ->
+            #message{payload = Payload} = Msg,
+            Res = emqx_limiter_client_container:try_consume(
+                Limiter0,
+                [
+                    {delivery_bytes, iolist_size(Payload)},
```
zmstone: emqx_message:estimate_size ?

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -884,6 +919,66 @@ publish_will_message_now(#session{} = Session, #message{} = WillMsg) ->
...
+
+ensure_retry_dequeue_timer(_, #{retry_dequeue_tref := TRef} = Ctx0) when is_reference(TRef) ->
+    Ctx0;
+ensure_retry_dequeue_timer({failed_to_consume_from_limiter, LimiterId}, Ctx0) ->
+    try emqx_limiter_registry:get_limiter_options(LimiterId) of
+        #{interval := Interval, capacity := Capacity} ->
+            %% Finite capacity
+            Time = max(1, Interval div Capacity),
+            TRef = emqx_utils:start_timer(Time, {emqx_session, ?RETRY_DEQUEUE_TIMER}),
+            Ctx0#{retry_dequeue_tref => TRef};
```
zmstone: maybe session should just put retry instruction in the context, and channel should take care of the timer start/cancel.
e.g.
`Ctx0#{deliver_retry => RemainingMs}`

zmstone: this way, the timer reference can live in channel state

apps/emqx/src/emqx_channel.erl
```diff
@@ -849,22 +855,28 @@ process_pubrel(
...
+            Channel = pop_limiter_ctx(Channel0),
             handle_out(publish, Publishes, Channel#channel{session = NSession});
         {error, ?RC_PROTOCOL_ERROR} ->
+            Channel = pop_limiter_ctx(Channel0),
             handle_out(disconnect, ?RC_PROTOCOL_ERROR, Channel);
         {error, ?RC_PACKET_IDENTIFIER_IN_USE} ->
+            Channel = pop_limiter_ctx(Channel0),
             ok = inc_metrics('packets.pubcomp.inuse', Channel),
-            {ok, Channel};
+            {ok, Channel0};
```
zmstone: why return `Channel0`, but not `Channel` here?

thalesmg: typo.

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -884,6 +925,64 @@ publish_will_message_now(#session{} = Session, #message{} = WillMsg) ->
...
+    ?SLOG_THROTTLE(
+        warning,
+        #{
+            msg => cannot_deliver_from_topic_due_to_quota_exceeded,
+            reason => Reason
+        },
+        #{topic => Topic, tag => "QUOTA"}
+    ).
+
+ensure_retry_dequeue_timer(_, #{?DELIVER_RETRY_TIMER := TRef} = Ctx0) when is_reference(TRef) ->
```
zmstone: it should not check ref here ?

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -609,7 +648,9 @@ mark_begin_deliver(Msg) ->
 handle_timeout(ClientInfo, retry_delivery, Session) ->
     retry(ClientInfo, Session);
 handle_timeout(ClientInfo, expire_awaiting_rel, Session) ->
-    expire(ClientInfo, Session).
+    expire(ClientInfo, Session);
+handle_timeout(ClientInfo, ?DELIVER_RETRY_TIMER, Session) ->
```
zmstone: should delete the timer from `timers` map.
otherwise if the limiter still limits in the following `dequeue` call, it won't start a new timer ?

rel/i18n/emqx_limiter_schema.hocon
```diff
@@ -70,4 +70,28 @@ bytes_burst.label:
...
+
+delivery_bytes_burst.desc:
+"""Todo."""
+
+delivery_bytes_rate.label:
+"""Data Delivery Burst"""
+
+delivery_messages_rate.desc:
+"""Todo."""
+
```
zmstone: duplicates line 76~80

apps/emqx/src/emqx_session_mem.erl
thalesmg: why not?  we avoid the recomputing of the time by checking.

zmstone: i mean, it's now an integer, isn't it.

apps/emqx/src/emqx_session_mem.erl
thalesmg: I think it will.  I don't see `ensure_timer/3` not starting a timer.  :thinking: 

https://github.com/emqx/emqx/blob/daebe2bac2b008045e98f4c9cab35c6ca86fe529/apps/emqx/src/emqx_channel.erl#L1886-L1888

rel/i18n/emqx_limiter_schema.hocon
thalesmg: not duplicated.  should be burst.

:checkered_flag:

apps/emqx/src/emqx_session_mem.erl
thalesmg: oh, right, the guard is wrong.  :see_no_evil:

thalesmg: :checkered_flag:

apps/emqx/src/emqx_session_mem.erl
thalesmg: :checkered_flag:

thalesmg:
I'll add the changelog in a follow up, to avoid losing approval.

https://github.com/emqx/emqx/pull/16959

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12912)

PR #16931, opened 2026-03-18
feat(emqx_username_quota): track quota exceeded stats

zmstone:
## Summary
- add `emqx_username_quota_stats` to own ETS-based quota exceeded stats per username
- move quota exceeded warning logging into the stats process with simple throttling
- cover the new stats behavior in plugin CT

## Verification
- `env SUITES=test/emqx_username_quota_SUITE.erl make plugins/emqx_username_quota-ct`


plugins/emqx_username_quota/src/emqx_username_quota_sup.erl
```diff
@@ -25,6 +25,13 @@ init([]) ->
     MFA = {PoolModule, start_link, []},
     SupArgs = [PoolModule, PoolType, MFA],
     ChildSpecs = [
+        #{
+            id => emqx_username_quota_stats,
+            start => {emqx_username_quota_stats, start_link, []},
+            type => worker,
+            restart => permanent,
+            shutdown => 1_000
+        },
```
thalesmg, line 34: since this supervisor is `one_for_all` and this new process seems to be only for simple stats bookkeeping, maybe it would be better to move it under another supervisor, so that if it crashes it doesn't restart the other core processes?

plugins/emqx_username_quota/src/emqx_username_quota_stats.erl
```diff
@@ -0,0 +1,145 @@
...
+-doc """
+Record one quota exceeded event for Username.
+""".
+record_quota_exceeded(Username) when is_binary(Username) ->
+    case whereis(?SERVER) of
+        undefined ->
+            ok;
+        _Pid ->
+            gen_server:cast(?SERVER, {record_quota_exceeded, Username})
+    end.
```
thalesmg: nit: cast does not care if the process does not exist

```elixir
iex(1)> :gen_server.cast(:i_dont_exist, :hello)
:ok
```


```suggestion
    gen_server:cast(?SERVER, {record_quota_exceeded, Username}).
```

plugins/emqx_username_quota/src/emqx_username_quota_stats.erl
```diff
@@ -0,0 +1,145 @@
...
+
+-doc """
+Return the stored quota exceeded statistics for Username.
+""".
+get(Username) when is_binary(Username) ->
+    case ets:lookup(?STATS_TAB, Username) of
```
thalesmg: maybe wrap in a `try` to avoid the race when table being re-created?

plugins/emqx_username_quota/src/emqx_username_quota_stats.erl
```diff
@@ -0,0 +1,145 @@
...
+reset() ->
+    case whereis(?SERVER) of
+        undefined ->
+            ok;
+        _Pid ->
+            gen_server:call(?SERVER, reset)
+    end.
+
+init([]) ->
+    ok = emqx_utils_ets:new(?STATS_TAB, [named_table, ordered_set, public, {keypos, 2}]),
```
thalesmg:
```suggestion
    ok = emqx_utils_ets:new(?STATS_TAB, [named_table, ordered_set, public, {keypos, #?STATS_TAB.username}]),
```
?

thalesmg (dismissed):

plugins/emqx_username_quota/src/emqx_username_quota_sup.erl
zmstone: it's a simple enough gen_server, not expected to restart at all, so the risk is low.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12809)

PR #16943, opened 2026-03-19
feat(dashboard_sso): add force MFA support for SSO users

JimMoen:
Release version: 5.10

## Summary

Add per-backend `force_mfa` configuration for SSO (OIDC/SAML/LDAP), forcing SSO users to complete TOTP MFA setup/verification before receiving a dashboard token, regardless of IDP-side settings.

### MFA three-state model
- `not_configured`: user has no MFA secret → force setup (return QR code + temp token)
- `enabled`: user has MFA configured → require TOTP verification
- `admin_disabled`: admin explicitly disabled MFA for this user → skip MFA (preserves secret with `disabled_by => admin` marker)

### SSO integration
- Each SSO backend (OIDC/SAML/LDAP) can independently set `force_mfa: true/false` in its config
- SSO login callbacks (`ensure_user_exists`) now call `emqx_dashboard_sso_mfa:check_sso_mfa/2` instead of directly signing tokens
- MFA setup secret stored in ETS temp token (not written to user record until setup completes)

### New API endpoints
- `POST /sso/mfa/setup` — complete MFA setup with temp token + TOTP code
- `POST /sso/mfa/verify` — verify TOTP for users with MFA already enabled

### Admin MFA control
- `emqx_dashboard_admin:admin_disable_mfa/1` — disable MFA for a user (preserves secret)
- `emqx_dashboard_admin:admin_enable_mfa/1` — re-enable MFA

### Key files
- `apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa.erl` — core SSO MFA logic
- `apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl` — setup/verify endpoints
- `apps/emqx_dashboard/src/emqx_dashboard_mfa.erl` — temp token mechanism (ETS), admin disable/enable
- `apps/emqx_dashboard/src/emqx_dashboard_admin.erl` — MFA state management
- `apps/emqx_dashboard_sso/src/emqx_dashboard_sso_schema.erl` — `force_mfa` config field

### Testing
- `emqx_dashboard_sso_mfa_SUITE` with 19 test cases: temp token lifecycle, check_sso_mfa flow for all MFA states, admin control, and API endpoint tests.

## PR Checklist
- [ ] Schema changes are backward compatible or intentionally breaking

Schema changes: backward compatible — new optional `force_mfa` boolean field per SSO backend config, defaults to false.

### DIFF
--- a/apps/emqx_dashboard/include/emqx_dashboard.hrl
+++ b/apps/emqx_dashboard/include/emqx_dashboard.hrl
@@ -19,6 +19,34 @@
     description :: binary(),
     role = ?ROLE_DEFAULT :: dashboard_user_role(),
     extra = #{} :: map()
+    %% Extensible metadata map. Known keys:
+    %%
+    %%   mfa_state => #{mechanism => totp, secret => binary(), ...}
+    %%              | disabled
+    %%     MFA configuration for this user.
+    %%     - Map with `mechanism` and `secret`: MFA is enabled or pending setup.
+    %%       May also contain `first_verify_ts` after first successful TOTP verify.
+    %%     - `disabled` atom: admin has explicitly exempted this user from MFA.
+    %%     - Key absent: MFA never configured (default).
+    %%
+    %%   mfa_pending => #{type => setup | challenge,
+    %%                    token => binary(),
+    %%                    secret => binary(),        %% setup only
+    %%                    timestamp => integer()}
+    %%     Short-lived SSO MFA session token for TOTP setup or verification.
+    %%     Only used when MFA is required after SSO login.
+    %%     For regular (non-SSO) login, no pending state is required.
+    %%
+    %%   sso_code => #{code => binary(),
+    %%                 payload => map(),
+    %%                 exptime => integer()}
+    %%     One-time SSO login code for secure redirect. Created after
+    %%     SAML/OIDC callback; exchanged by frontend via POST /sso/token_exchange.
+    %%     TTL: 60 seconds. Consumed after single use.
+    %%
+    %%   login_lock => integer()
+    %%     Timestamp until which login is locked after too many failed attempts.
+    %%
 }).
 
 -type dashboard_user() :: #?ADMIN{}.
--- a/apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
@@ -5,6 +5,8 @@
 %% @doc Multifactor authentication interface.
 -module(emqx_dashboard_mfa).
 
+-include("emqx_dashboard.hrl").
+
 -export([
     init/1,
     verify/2,
@@ -15,14 +17,31 @@
     supported_mechanisms/0
 ]).
 
--export_type([mfa_state/0, mechanism/0]).
+%% SSO MFA temporary tokens — stored in extra map of emqx_admin record
+-export([
+    create_setup_token/2,
+    create_verify_token/1,
+    verify_temp_token/2,
+    peek_temp_token/2,
+    generate_token/1
+]).
+
+-export_type([mfa_state/0, mechanism/0, temp_token_purpose/0]).
 
 -type mechanism() :: totp.
--type totp_state() :: #{mechanism := totp, secret := binary(), first_verify_ts => integer()}.
+-type totp_state() :: #{
+    mechanism := totp,
+    secret := binary(),
+    first_verify_ts => integer()
+}.
 -type mfa_state() :: disabled | totp_state().
+-type temp_token_purpose() :: {setup, binary()} | verify.
 
 -define(TOTP_KEY_BYTES, 20).
 -define(NO_FIRST_VERIFY_TS, 0).
+%% Temporary token validity: 5 minutes
+-define(TEMP_TOKEN_TTL_SEC, 300).
+-define(TOKEN_BYTES, 32).
 
 %% @doc Translate binary format mechanism name to atom.
 -spec mechanism(binary()) -> mechanism().
@@ -104,6 +123,82 @@ is_need_setup_error(#{mechanism := totp, error := missing_mfa_token, secret := S
 is_need_setup_error(_) ->
     false.
 
+%%--------------------------------------------------------------------
+%% SSO MFA temporary tokens — stored in extra map of emqx_admin record
+%%
+%% Tokens are stored as mfa_pending key in the emqx_admin record's
+%% extra map. This ensures SSO MFA works correctly in clustered
+%% deployments without a separate Mnesia table.
+%%--------------------------------------------------------------------
+
+%% @doc Create a short-lived temporary token for SSO MFA setup flow.
+%% The TOTP secret is stored in the pending record (not in user's enabled MFA state).
+-spec create_setup_token(binary(), binary()) -> binary().
+create_setup_token(Username, TotpSecret) ->
+    Token = generate_token(<<"mfa_">>),
+    Pending = #{
+        type => setup,
+        token => Token,
+        secret => TotpSecret,
+        timestamp => erlang:system_time(second)
+    },
+    {ok, ok} = emqx_dashboard_admin:set_mfa_pending(Username, Pending),
+    Token.
+
+%% @doc Create a short-lived temporary token for SSO MFA verify flow.
+-spec create_verify_token(binary()) -> binary().
+create_verify_token(Username) ->
+    Token = generate_token(<<"mfa_">>),
+    Pending = #{
+        type => challenge,
+        token => Token,
+        timestamp => erlang:system_time(second)
+    },
+    {ok, ok} = emqx_dashboard_admin:set_mfa_pending(Username, Pending),
+    Token.
+
+%% @doc Verify and consume a temporary token.
+%% Looks up the pending token by username (O(1) ets:lookup) and compares.
+-spec verify_temp_token(dashboard_username(), binary()) ->
+    {ok, term(), temp_token_purpose()} | {error, term()}.
+verify_temp_token(SsoUsername, Token) ->
+    case emqx_dashboard_admin:get_mfa_pending(SsoUsername) of
+        {ok, #{token := StoredToken, type := Type, timestamp := Timestamp} = Pending} when
+            StoredToken =:= Token
+        ->
+            case is_token_valid(Timestamp) of
+                true ->
+                    {ok, ok} = emqx_dashboard_admin:clear_mfa_pending(SsoUsername),
+                    {ok, SsoUsername, to_purpose(Type, Pending)};
+                false ->
+                    {error, token_expired}
+            end;
+        _ ->
+            {error, invalid_token}
+    end.
+
+%% @doc Peek at a temporary token without consuming it.
+-spec peek_temp_token(dashboard_username(), binary()) ->
+    {ok, term(), temp_token_purpose()} | {error, term()}.
+peek_temp_token(SsoUsername, Token) ->
+    case emqx_dashboard_admin:get_mfa_pending(SsoUsername) of
+        {ok, #{token := StoredToken, type := Type, timestamp := Timestamp} = Pending} when
+            StoredToken =:= Token
+        ->
+            case is_token_valid(Timestamp) of
+                true ->
+                    {ok, SsoUsername, to_purpose(Type, Pending)};
+                false ->
+                    {error, token_expired}
+            end;
+        _ ->
+            {error, invalid_token}
+    end.
+
+%%--------------------------------------------------------------------
+%% Internal functions
+%%--------------------------------------------------------------------
+
 verify_totp(Token, Secret) ->
     try
         case pot:valid_totp(Token, Secret) of
@@ -122,3 +217,19 @@ totp_error(Reason) ->
         mechanism => totp,
         error => Reason
     }.
+
+%% @doc Generate a random token with a prefix.
+%% Format: <<Prefix, RandomHex:64>>
+-spec generate_token(binary()) -> binary().
+generate_token(Prefix) ->
+    Hex = binary:encode_hex(crypto:strong_rand_bytes(?TOKEN_BYTES), lowercase),
+    <<Prefix/binary, Hex/binary>>.
+
+is_token_valid(Timestamp) when is_integer(Timestamp) ->
+    Now = erlang:system_time(second),
+    (Now - Timestamp) < ?TEMP_TOKEN_TTL_SEC;
+is_token_valid(_) ->
+    false.
+
+to_purpose(setup, #{secret := Secret}) -> {setup, Secret};
+to_purpose(challenge, _Pending) -> verify.
--- a/apps/emqx_dashboard_sso/src/emqx_dashboard_sso_api.erl
+++ b/apps/emqx_dashboard_sso/src/emqx_dashboard_sso_api.erl
@@ -9,6 +9,7 @@
 -include_lib("hocon/include/hoconsc.hrl").
 -include_lib("emqx/include/logger.hrl").
 -include_lib("emqx/include/emqx_api_key_scopes.hrl").
+-include_lib("emqx_dashboard/include/emqx_dashboard.hrl").
 
 -import(hoconsc, [
     mk/2,
@@ -33,7 +34,8 @@
     running/2,
     login/2,
     sso/2,
-    backend/2
+    backend/2,
+    token_exchange/2
 ]).
 
 -export([sso_parameters/1, login_meta/4]).
@@ -57,7 +59,8 @@ paths() ->
         "/sso",
         "/sso/:backend",
         "/sso/running",
-        "/sso/login/:backend"
+        "/sso/login/:backend",
+        "/sso/token_exchange"
     ].
 
 schema("/sso/running") ->
@@ -95,11 +98,33 @@ schema("/sso/login/:backend") ->
             parameters => backend_name_in_path(),
             'requestBody' => login_union(),
             responses => #{
-                200 => emqx_dashboard_api:fields([role, token, version, license]),
+                200 => hoconsc:union([
+                    ref(login_success_response),
+                    ref(mfa_setup_response),
+                    ref(mfa_verify_response)
+                ]),
                 %% Redirect to IDP for saml
                 302 => response_schema(302),
                 401 => response_schema(401),
                 404 => response_schema(404)
+            },
+            security => []
+        }
+    };
+schema("/sso/token_exchange") ->
+    #{
+        'operationId' => token_exchange,
+        post => #{
+            tags => [?TAGS],
+            desc => ?DESC(token_exchange),
+            'requestBody' => ref(token_exchange_request),
+            responses => #{
+                200 => hoconsc:union([
+                    ref(login_success_response),
+                    ref(mfa_setup_response),
+                    ref(mfa_verify_response)
+                ]),
+                400 => response_schema(400)
             },
             security => []
         }
@@ -152,7 +177,35 @@ fields(backend_status) ->
                         desc => ?DESC(last_error)
                     }
                 )}
-        ].
+        ];
+fields(login_success_response) ->
+    emqx_dashboard_api:fields([role, token, version, license]) ++
+        [
+            {username, mk(binary(), #{desc => <<"Username">>})},
+            {backend, mk(binary(), #{desc => <<"SSO backend type">>})}
+        ];
+fields(mfa_setup_response) ->
+    [
+        {action, mk(binary(), #{desc => <<"MFA action: mfa_setup">>, example => <<"mfa_setup">>})},
+        {setup_token, mk(binary(), #{desc => <<"Temporary setup token for MFA binding">>})},
+        {mechanism, mk(binary(), #{desc => <<"MFA mechanism">>, example => <<"totp">>})},
+        {username, mk(binary(), #{desc => <<"Username">>})},
+        {backend, mk(binary(), #{desc => <<"SSO backend type">>})}
+    ];
+fields(mfa_verify_response) ->
+    [
+        {action,
+            mk(binary(), #{desc => <<"MFA action: mfa_verify">>, example => <<"mfa_verify">>})},
+        {verify_token, mk(binary(), #{desc => <<"Temporary verify token for MFA verification">>})},
+        {username, mk(binary(), #{desc => <<"Username">>})},
+        {backend, mk(binary(), #{desc => <<"SSO backend type">>})}
+    ];
+fields(token_exchange_request) ->
+    [
+        {code, mk(binary(), #{desc => <<"One-time SSO code">>, required => true})},
+        {username, mk(binary(), #{desc => ?DESC(sso_username), required => true})},
+        {backend, mk(binary(), #{desc => ?DESC(sso_backend), required => true})}
+    ].
 
 %%--------------------------------------------------------------------
 %% API
@@ -161,6 +214,32 @@ fields(backend_status) ->
 running(get, _Request) ->
     {200, emqx_dashboard_sso_manager:running()}.
 
+token_exchange(post, #{
+    body := #{<<"code">> := Code, <<"username">> := Username, <<"backend">> := BackendBin}
+}) ->
+    case parse_backend(BackendBin) of
+        {ok, BackendAtom} ->
+            case emqx_dashboard_sso_code:exchange_code(BackendAtom, Username, Code) of
+                {ok, #{action := <<"login">>}} ->
+                    %% Sign JWT now at exchange time
+                    SsoUsername = ?SSO_USERNAME(BackendAtom, Username),
+                    case emqx_dashboard_admin:lookup_user(SsoUsername) of
+                        [User] ->
+                            {ok, Role, Token} = emqx_dashboard_token:sign(User),
+                            {200, login_meta(Username, Role, Token, BackendAtom)};
+                        [] ->
+                            {400, #{code => ?BAD_REQUEST, message => <<"User not found">>}}
+                    end;
+                {ok, Payload} ->
+                    %% MFA setup/verify — return as-is
+                    {200, Payload};
+                {error, _Reason} ->
+                    {400, #{code => ?BAD_REQUEST, message => <<"Invalid or expired SSO code">>}}
+            end;
+        {error, _} ->
+            {400, #{code => ?BAD_REQUEST, message => <<"Unknown SSO backend">>}}
+    end.
+
 login(post, #{bindings := #{backend := Backend}, body := Body} = Request) ->
     minirest_handler:update_log_meta(#{log_from => dashboard, log_source => Backend}),
     case emqx_dashboard_sso_manager:lookup_state(Backend) of
@@ -168,7 +247,23 @@ login(post, #{bindings := #{backend := Backend}, body := Body} = Request) ->
             {404, #{code => ?BACKEND_NOT_FOUND, message => <<"Backend not found">>}};
         State ->
             case emqx_dashboard_sso:login(provider(Backend), Request, State) of
+                {ok, login} ->
+                    ?SLOG(info, #{
+                        msg => "dashboard_sso_login_successful",
+                        request => emqx_utils:redact(Request)
+                    }),
+                    Username = maps:get(<<"username">>, Body),
+                    minirest_handler:update_log_meta(#{log_source => Username}),
+                    SsoUsername = ?SSO_USERNAME(Backend, Username),
+                    case emqx_dashboard_admin:lookup_user(SsoUsername) of
+                        [User] ->
+                            {ok, Role, Token} = emqx_dashboard_token:sign(User),
+                            {200, login_meta(Username, Role, Token, Backend)};
+                        [] ->
+                            {400, #{code => ?BAD_REQUEST, message => <<"User not found">>}}
+                    end;
                 {ok, Role, Token} ->
+                    %% Legacy path — kept for backends that still sign directly
                     ?SLOG(info, #{
                         msg => "dashboard_sso_login_successful",
                         request => emqx_utils:redact(Request)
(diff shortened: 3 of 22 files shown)


apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
```diff
@@ -104,6 +128,90 @@ is_need_setup_error(#{mechanism := totp, error := missing_mfa_token, secret := S
...
+            ok
+    end.
+
+%% @doc Create a short-lived temporary token for SSO MFA setup flow.
+%% The TOTP secret is stored in the token's ETS entry so it is NOT written
+%% to the user record until setup is completed successfully.
+-spec create_setup_token(binary(), binary()) -> binary().
+create_setup_token(Username, TotpSecret) ->
+    create_temp_token(Username, {setup, TotpSecret}).
+
```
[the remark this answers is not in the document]
JimMoen: Fixed in 91f5fddd: defined temp_token_purpose() type as {setup, binary()} | verify and updated create_temp_token/2 spec.

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
```diff
@@ -104,6 +128,90 @@ is_need_setup_error(#{mechanism := totp, error := missing_mfa_token, secret := S
...
+            ets:delete(?TEMP_TOKEN_TAB, Token),
+            Now = erlang:system_time(second),
+            case Now =< ExpTime of
+                true ->
+                    {ok, Username, Purpose};
+                false ->
+                    {error, token_expired}
+            end;
+        [] ->
+            {error, invalid_token}
```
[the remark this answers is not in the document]
JimMoen: Fixed in 91f5fddd: added ets:whereis guard — returns {error, invalid_token} when table doesn't exist instead of crashing. Test: t_verify_token_no_ets_table.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa.erl
```diff
@@ -0,0 +1,123 @@
...
+            %% Admin has disabled MFA for this user, skip MFA
+            emqx_dashboard_token:sign(User);
+        not_configured ->
+            %% User has never configured TOTP, initialize and guide setup
+            init_and_create_setup(Username);
+        enabled ->
+            %% User has TOTP configured and enabled, require verification
+            VerifyToken = emqx_dashboard_mfa:create_verify_token(format_username(Username)),
+            {mfa_verify, VerifyToken}
+    end.
```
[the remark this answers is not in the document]
JimMoen: Fixed in b46a7e16: format_username/1 now encodes backend as <<"sso:backend:name">>. Added parse_token_username/1 to recover {Backend, Name}. Deleted find_sso_user. Tests: t_parse_token_username, t_format_username_roundtrip.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_saml.erl
```diff
@@ -355,6 +355,18 @@ gen_redirect_response(DashboardAddr, Username) ->
             Target = login_redirect_target(DashboardAddr, Username, Role, Token),
             Response = {302, maps:merge(?RESPHEADERS, #{<<"location">> => Target}), ?REDIRECT_BODY},
             {redirect, Username, Response};
+        {mfa_setup, SetupToken, QRInfo} ->
+            Target = mfa_redirect_target(
+                DashboardAddr, Username, SetupToken, QRInfo, <<"mfa_setup">>
```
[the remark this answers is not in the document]
JimMoen: Fixed in 1ed2894c: SAML redirect now passes #{} instead of QRInfo. Secret retrieved securely via new POST /sso/mfa/setup-info endpoint (non-consuming peek). Test: t_mfa_setup_info_api.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc_api.erl
```diff
@@ -205,9 +205,13 @@ ensure_user_exists(_Cfg, <<"undefined">>) ->
...
     case emqx_dashboard_admin:lookup_user(?BACKEND, Username) of
         [User] ->
-            case emqx_dashboard_token:sign(User) of
+            case emqx_dashboard_sso_mfa:check_sso_mfa(User, ?BACKEND) of
                 {ok, Role, Token} ->
                     {ok, login_redirect_target(Cfg, Username, Role, Token)};
+                {mfa_setup, SetupToken, QRInfo} ->
+                    {ok, mfa_redirect_target(Cfg, Username, SetupToken, QRInfo, <<"mfa_setup">>)};
+                {mfa_verify, VerifyToken} ->
+                    {ok, mfa_redirect_target(Cfg, Username, VerifyToken, #{}, <<"mfa_verify">>)};
```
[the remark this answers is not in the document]
JimMoen: Fixed in 1ed2894c: same fix as SAML — OIDC redirect passes empty map instead of QRInfo.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl
```diff
@@ -0,0 +1,222 @@
...
+        {ok, {Backend, User}} ->
+            SsoUsername = ?SSO_USERNAME(Backend, Username),
+            case emqx_dashboard_admin:get_mfa_state(SsoUsername) of
+                {ok, #{mechanism := totp} = MfaState} ->
+                    case emqx_dashboard_mfa:verify(MfaState, TotpCode) of
+                        ok ->
+                            sign_and_respond(User, Username, Backend);
+                        {ok, NewState} ->
+                            {ok, ok} = emqx_dashboard_admin:set_mfa_state(SsoUsername, NewState),
+                            sign_and_respond(User, Username, Backend);
```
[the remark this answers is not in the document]
JimMoen: Fixed in 9cef6731 + elvis refactor: extracted verify_and_respond/6 and persist_state_and_respond/5 with explicit error handling and SLOG logging. Also fixed finalize_mfa_setup similarly.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_api.erl
```diff
@@ -171,6 +171,32 @@ login(post, #{bindings := #{backend := Backend}, body := Body} = Request) ->
...
+                    }),
+                    Username = maps:get(<<"username">>, Body),
+                    {200, #{
+                        action => <<"mfa_setup">>,
+                        setup_token => SetupToken,
+                        secret => maps:get(secret, QRInfo, <<>>),
+                        mechanism => maps:get(mechanism, QRInfo, totp),
+                        username => Username,
+                        backend => Backend
+                    }};
```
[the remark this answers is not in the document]
JimMoen: Fixed in ecb75ab8: /sso/login/:backend 200 response updated to hoconsc:union of login_success_response, mfa_setup_response, mfa_verify_response. Secret field removed from mfa_setup_response (moved to /sso/mfa/setup-info).

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl
```diff
@@ -0,0 +1,222 @@
...
+
+find_sso_user(_Username, []) ->
+    {error, not_found};
+find_sso_user(Username, [Backend | Rest]) ->
+    case emqx_dashboard_admin:lookup_user(Backend, Username) of
+        [User] ->
+            {ok, {Backend, User}};
+        [] ->
+            find_sso_user(Username, Rest)
+    end.
```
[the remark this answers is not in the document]
JimMoen: Fixed in b46a7e16: find_sso_user replaced by resolve_sso_user using parse_token_username to extract backend from token. No hardcoded backend list.

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
```diff
@@ -104,6 +130,120 @@ is_need_setup_error(#{mechanism := totp, error := missing_mfa_token, secret := S
...
+%%--------------------------------------------------------------------
+%% SSO MFA temporary tokens
+%%--------------------------------------------------------------------
+
+%% @doc Ensure the ETS table for temporary MFA tokens exists.
+-spec ensure_temp_token_tab() -> ok.
+ensure_temp_token_tab() ->
+    case ets:whereis(?TEMP_TOKEN_TAB) of
+        undefined ->
+            ?TEMP_TOKEN_TAB = ets:new(?TEMP_TOKEN_TAB, [
```
zmstone: this will not work when clustered.

zmstone:
I think we can simplify this design and avoid the new ETS-backed temp-token flow entirely.

After SSO succeeds, instead of creating a server-side temp token, could we issue a short-lived  (expires in 60s) **pre-auth bearer token** that carries MFA metadata, e.g. user identity, backend, `action = setup|verify`, expiry Then:

- if MFA is not needed, issue the normal dashboard JWT as today
- if MFA setup is needed, return the pre-auth token plus the TOTP secret/QR info
- if MFA verify is needed, return the pre-auth token and require it on `/sso/mfa/verify`
- only the MFA completion endpoints should accept this token
- all normal dashboard APIs must reject it
- after TOTP succeeds, exchange it for the real dashboard JWT

Why I think this is a better fit:

- it avoids the new local ETS state entirely
- no risk of losing tokens because the ETS owner process exits
- no cross-node / load-balancer issue where login lands on node A and MFA completion lands on node B
- it reuses the existing token model instead of introducing a second transient auth mechanism

The important caveat is that this **cannot** be treated as a normal bearer token with extra metadata; it has to be a restricted “login not finished yet” token class. If normal APIs accept it, MFA is effectively bypassed.

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
JimMoen: Agreed — ETS is node-local, so SSO MFA flow will break when the setup/verify request lands on a different node than the login.

Fixed by replacing ETS with an independent Mnesia table `emqx_admin_mfa` (mria `ram_copies`), aligned with the e4.4 enterprise `mqtt_admin_mfa` design:

- MFA state (secret, enabled, admin_disabled) stored in `emqx_admin_mfa` table instead of `emqx_admin.extra` map
- Pending tokens stored as `pending_token`/`pending_timestamp` fields in the same record (one per user, auto-overwrite)
- Table created in `emqx_dashboard_admin:create_tables/0`, managed by supervisor — no process lifetime issues
- Token lookup via `mnesia:dirty_select` matching `pending_token` field (same pattern as e4.4 `verify_mfa_token/1`)
- No GC needed — each user has exactly one pending token slot, overwritten on each login attempt

This also resolves the ETS table ownership and expired token accumulation concerns.

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
```diff
@@ -104,6 +130,120 @@ is_need_setup_error(#{mechanism := totp, error := missing_mfa_token, secret := S
...
+                set,
+                public,
+                named_table,
+                {read_concurrency, true}
+            ]),
```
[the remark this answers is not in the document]
JimMoen: Fixed — ETS replaced with Mnesia `emqx_admin_mfa` table (mria `ram_copies`), created in `emqx_dashboard_admin:create_tables/0` under supervisor. No process lifetime issues.

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
```diff
@@ -122,3 +262,11 @@ totp_error(Reason) ->
         mechanism => totp,
         error => Reason
     }.
+
+-spec create_temp_token(binary(), temp_token_purpose()) -> binary().
+create_temp_token(Username, Purpose) ->
+    ensure_temp_token_tab(),
+    Token = binary:encode_hex(crypto:strong_rand_bytes(?TEMP_TOKEN_BYTES)),
+    ExpTime = erlang:system_time(second) + ?TEMP_TOKEN_TTL_SEC,
```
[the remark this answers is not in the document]
JimMoen: Fixed — with Mnesia, each user has exactly one record with a single `pending_token` slot. New logins overwrite the old token, so there is no accumulation. No GC needed.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl
```diff
@@ -0,0 +1,296 @@
...
+            tags => [?TAGS],
+            desc => <<"Verify SSO MFA TOTP code and issue JWT">>,
+            'requestBody' => ref(mfa_verify_request),
+            responses => #{
+                200 => emqx_dashboard_api:fields([token, version, license]),
+                400 => response_schema(400),
+                401 => response_schema(401)
+            },
+            security => []
+        }
```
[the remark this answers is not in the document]
JimMoen: Fixed — `/sso/mfa/setup` and `/sso/mfa/verify` 200 response now uses `ref(emqx_dashboard_sso_api, login_success_response)` which includes all fields (token, version, license, username, role, backend).

JimMoen:
Thanks for the suggestion. I evaluated the stateless pre-auth JWT approach — it works for the **verify** flow but not for **setup**.

### Setup flow requires server-side secret storage

The setup flow has two steps:
1. SSO login → generate TOTP secret → return QR info to frontend
2. User scans QR, submits TOTP code → server calls `pot:valid_totp(Code, Secret)` → persists MFA state

In step 2, the server needs the TOTP secret. If it's embedded in the JWT, the secret travels through the client (headers/body), which is the same exposure we specifically avoided by introducing `POST /sso/mfa/setup-info` (to keep the secret off redirect URLs). If the secret is **not** in the JWT, then the server must store it somewhere — bringing us back to server-side state.

### Verify flow: stateless possible but with tradeoffs

The verify flow (TOTP already bound) doesn't need secret exchange, so a JWT could work. However:
- **One-time consumption** is lost — a stateless JWT can't be invalidated without a server-side blacklist, which reintroduces state
- **Two mechanisms** for the same flow (JWT for verify, server-state for setup) adds complexity

### What we did instead

Replaced the standalone ETS temp-token table with a **Mnesia table** `emqx_admin_mfa` (`disc_copies`, mria-synced), aligned with the e4.4 enterprise `mqtt_admin_mfa` design:

| Concern | Resolution |
|---------|-----------|
| Cross-node / LB routing | Mnesia disc_copies + mria sync — any node can verify |
| ETS owner process exit | Table created in `create_tables/0`, supervisor-managed |
| Token accumulation | One `pending_token` slot per user, auto-overwritten on new login |
| Token replay | Consumed after use (pending fields cleared in transaction) |
| Setup secret | Stored in Mnesia record until setup completes |
| Auth chain isolation | MFA endpoints use `security => []`, Bearer token path untouched |

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
zmstone: there should be no need for a new table, see comment below.

Also, hot-upgrade requirement would not allow us to create a new table.

zmstone:
The QR code itself is the TOTP secret. Meaning it's not an issue just to provide TOTP secret in the pre-auth token.

JimMoen:
You're right that the TOTP secret exposure is not a concern — the QR code is the secret, and Keycloak's `UpdateTotp.java` does exactly this (sends secret to frontend as a form field, gets it back with the TOTP code). So I retract that argument.

However, after researching industry practices, I believe the **token revocation / one-time guarantee** is the real issue for our specific flow, not secret exposure.

### The problem: our MFA endpoints are unauthenticated

Unlike Keycloak's TOTP setup (which happens inside an authenticated session), our SSO MFA endpoints use `security => []` — the pre-auth token is the **only** credential. This means an intercepted pre-auth JWT gives the attacker everything needed to complete setup.

**Concrete scenario:**
1. SSO login → pre-auth JWT issued (contains secretA, 60s TTL)
2. Attacker intercepts JWT (XSS, browser extension, shared network)
3. User abandons setup, re-triggers SSO → gets new JWT with secretB → completes setup with secretB
4. Attacker submits old JWT (secretA) + valid TOTP code within 60s
5. Server: JWT signature valid ✅, TOTP code valid against secretA ✅ → overwrites MFA record with secretA
6. **User locked out**: authenticator app has secretB, server has secretA

With server-side state, step 5 fails because the old token was invalidated when the new one was created.

If instead we refuse to regenerate within TTL: attacker intercepts JWT → user can't re-setup for 60s → attacker completes setup first → attacker controls the account's MFA.

### Industry alignment

- **IETF draft-contario-totp-secure-enrollment-02**: Even for improving QR code enrollment, the proposed scheme uses a **one-time nonce** (server-side state) to ensure only one enrollment succeeds. Quote: _"only one request for the secret key is honored"_
- **Multiple best-practice guides** (WorkOS, DEV.to, ASP.NET Core guide): recommend server-side pending state during setup, with "verify before persisting" pattern
- **Medium JWT-vs-DB comparison**: acknowledges JWT works for OTP but notes _"Even though it's a JWT, treat the OTP inside as one-use. After successful verification, you should still mark it server-side"_ — which reintroduces state

### Current implementation

The pushed commits replace ETS with a Mnesia `emqx_admin_mfa` table (`disc_copies`, mria-synced):
- One `pending_token` slot per user — new login overwrites old, invalidating intercepted tokens
- Token consumed after use (pending fields cleared in transaction)
- No separate temp table; MFA state and pending token share one record per user
- All 53 test cases pass (cli + ldap + mfa + oidc + saml suites)
- Zero dialyzer warnings (after spec fixes)

Happy to discuss further if you see a way to get one-time guarantee without server state.

zmstone:
Take OIDC as an example. Today's flow:

  1. POST /sso/login/oidc (security => []) → 302 redirect to IdP
  2. User authenticates at IdP
  3. IdP redirects browser to GET /sso/oidc/callback (security => []) → EMQX exchanges the auth code, creates a bearer token, and redirects to dashboard/?login_meta=<base64 {token, role, ...}>

  With MFA, only step 3 changes:

  3. GET /sso/oidc/callback → EMQX exchanges the auth code, sees MFA is required → instead of a full session token, returns a short-lived MFA-pending token (which embeds the TOTP secret for first-time enrollment) in login_meta
  4. Dashboard decodes the token, renders QR code (first time) or TOTP input
  5. POST /mfa/verify (security => [#{bearerAuth => []}]) with the short-lived token + TOTP code → returns the real session token

This is secure because the short-lived MFA-pending token is only issued after the OIDC auth code exchange and userinfo retrieval succeed — there is no way to obtain it without completing the SSO flow first.

zmstone:
> Concrete scenario:
> 1. SSO login → pre-auth JWT issued (contains secretA, 60s TTL)
> 2. Attacker intercepts JWT (XSS, browser extension, shared network)
> 3. User abandons setup, re-triggers SSO → gets new JWT with secretB → completes setup with secretB
> 4. Attacker submits old JWT (secretA) + valid TOTP code within 60s
> 5. Server: JWT signature valid ✅, TOTP code valid against secretA ✅ → overwrites MFA record with secretA
> 6. User locked out: authenticator app has secretB, server has secretA

If the attacker has XSS or a malicious browser extension, they can steal *any* token.

apps/emqx_dashboard/include/emqx_dashboard.hrl
```diff
@@ -19,6 +19,27 @@
...
+    %%     - Map with `mechanism` and `secret`: MFA is enabled or pending setup.
+    %%       May also contain `first_verify_ts` after first successful TOTP verify.
+    %%     - `disabled` atom: admin has explicitly exempted this user from MFA.
+    %%     - Key absent: MFA never configured (default).
+    %%
+    %%   mfa_pending => #{type => setup | challenge,
+    %%                    token => binary(),
+    %%                    secret => binary(),        %% setup only
+    %%                    timestamp => integer()}
+    %%     Short-lived SSO MFA session token for TOTP setup or verification.
```
zmstone, line 36:
```suggestion
    %%     Short-lived SSO MFA session token for TOTP setup or verification.
    %%     Only used when MFA is required after SSO login.
    %%     For regular (non-SSO) login, no pending state is required.
```

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
```diff
@@ -15,14 +17,38 @@
...
+
+-export_type([mfa_state/0, mechanism/0, temp_token_purpose/0]).
 
 -type mechanism() :: totp.
--type totp_state() :: #{mechanism := totp, secret := binary(), first_verify_ts => integer()}.
+-type totp_state() :: #{
+    mechanism := totp,
+    secret := binary(),
+    first_verify_ts => integer(),
+    disabled_by => admin
```
zmstone: is disable but keep secret a hard requirement? 

I'd like to keep using the `totp_state() :: disabled` type for disabled state.
i.e. always re-init the secret.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl
```diff
@@ -0,0 +1,291 @@
...
+    };
+schema("/sso/mfa/setup") ->
+    #{
+        'operationId' => mfa_setup,
+        post => #{
+            tags => [?TAGS],
+            desc => <<"Complete SSO MFA TOTP setup: bind TOTP and issue JWT">>,
```
zmstone: add i18n ref

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl
```diff
@@ -0,0 +1,291 @@
...
+        "/sso/mfa/setup",
+        "/sso/mfa/verify"
+    ].
+
+schema("/sso/mfa/setup_info") ->
+    #{
+        'operationId' => mfa_setup_info,
+        post => #{
+            tags => [?TAGS],
+            desc => <<"Retrieve TOTP setup info (secret, mechanism) for QR code display">>,
```
zmstone: use i18n ref

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl
```diff
@@ -0,0 +1,291 @@
...
+        post => #{
+            tags => [?TAGS],
+            desc => <<"Verify SSO MFA TOTP code and issue JWT">>,
+            'requestBody' => ref(mfa_verify_request),
+            responses => #{
+                200 => ref(emqx_dashboard_sso_api, login_success_response),
+                400 => response_schema(400),
+                401 => response_schema(401)
+            },
+            security => []
```
zmstone, line 99: add comment to explain why this endpoint has no security.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl
```diff
@@ -0,0 +1,291 @@
...
+        post => #{
+            tags => [?TAGS],
+            desc => <<"Complete SSO MFA TOTP setup: bind TOTP and issue JWT">>,
+            'requestBody' => ref(mfa_setup_request),
+            responses => #{
+                200 => ref(emqx_dashboard_sso_api, login_success_response),
+                400 => response_schema(400),
+                401 => response_schema(401)
+            },
+            security => []
```
zmstone, line 82: add comment to explain why this endpoint has no security

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl
```diff
@@ -0,0 +1,291 @@
...
+        }
+    }.
+
+fields(mfa_setup_info_request) ->
+    [
+        {setup_token,
+            mk(binary(), #{
+                desc => <<"Temporary setup token from SSO login">>,
```
zmstone: use i18n ref for all desc

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl
```diff
@@ -0,0 +1,291 @@
...
+        'operationId' => mfa_setup_info,
+        post => #{
+            tags => [?TAGS],
+            desc => <<"Retrieve TOTP setup info (secret, mechanism) for QR code display">>,
+            'requestBody' => ref(mfa_setup_info_request),
+            responses => #{
+                200 => ref(mfa_setup_info_response),
+                401 => response_schema(401)
+            },
+            security => []
```
zmstone, line 65: add comment to explain why this endpoint has no secruity

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
```diff
@@ -104,6 +130,110 @@ is_need_setup_error(#{mechanism := totp, error := missing_mfa_token, secret := S
...
+%% @doc Verify and consume a temporary token.
+%% Returns {ok, Username, Purpose} on success; the token is cleared after use.
+%% For setup tokens, Purpose is {setup, TotpSecret}.
+%% For verify tokens, Purpose is verify.
+-spec verify_temp_token(binary()) ->
+    {ok, term(), temp_token_purpose()} | {error, term()}.
+verify_temp_token(Token) ->
+    case emqx_dashboard_admin:find_by_pending_token(Token) of
+        {ok, Username, #{type := Type, timestamp := Timestamp} = Pending} ->
+            case is_token_valid(Timestamp) of
```
zmstone, line 169: verify before find.

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
```diff
@@ -104,6 +130,110 @@ is_need_setup_error(#{mechanism := totp, error := missing_mfa_token, secret := S
...
+    Token.
+
+%% @doc Verify and consume a temporary token.
+%% Returns {ok, Username, Purpose} on success; the token is cleared after use.
+%% For setup tokens, Purpose is {setup, TotpSecret}.
+%% For verify tokens, Purpose is verify.
+-spec verify_temp_token(binary()) ->
+    {ok, term(), temp_token_purpose()} | {error, term()}.
+verify_temp_token(Token) ->
+    case emqx_dashboard_admin:find_by_pending_token(Token) of
```
zmstone: the find is a ets table scan, can become an attack point.
we can perhaps create a cluster-shared secret (for v5, we can use erlang cookie) to sign/salt the token so to fence attacking tokens.

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
JimMoen: No, keeping the secret is not required. We've simplified it:
- `disable_mfa` now stores `mfa_state => disabled` (the atom, no secret).
- `admin_enable_mfa` clears `mfa_state` entirely (not reinit), so the
  next SSO login with `force_mfa=true` triggers a fresh TOTP setup.
The full admin lifecycle for SSO users is:
1. User logs in via SSO, `force_mfa=true` → TOTP setup → bound
2. Admin disables MFA → `mfa_state => disabled` → user skips MFA on login
3. Admin re-enables → `mfa_state` removed → next login triggers setup again
`DELETE /users/:username/mfa` also supports `?reset=true` for hard reset
(removes mfa_state entirely without going through the disabled state),
which covers the "user lost their phone" scenario.

zmstone (dismissed):

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_saml.erl
```diff
@@ -392,9 +424,9 @@ is_msie(Headers) ->
...
-    LoginMeta = emqx_dashboard_sso_api:login_meta(Username, Role, Token, saml),
-    <<DashboardAddr/binary, "/?login_meta=", (base64_login_meta(LoginMeta))/binary>>.
-
-base64_login_meta(LoginMeta) ->
-    base64:encode(emqx_utils_json:encode(LoginMeta)).
+code_redirect_target(DashboardAddr, Username, Payload) ->
+    SsoUsername = ?SSO_USERNAME(saml, Username),
+    Code = emqx_dashboard_sso_code:create_code(SsoUsername, Payload),
+    LoginMeta = #{code => Code, username => Username},
+    MetaBin = base64:encode(emqx_utils_json:encode(LoginMeta)),
```
zmstone: encode with urlsafe option

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc_api.erl
```diff
@@ -228,8 +256,10 @@ ensure_user_exists(Cfg, Username) ->
 make_callback_url(#{config := #{dashboard_addr := Addr}}) ->
     list_to_binary(binary_to_list(Addr) ++ ?BASE_PATH ++ ?CALLBACK_PATH).
 
-login_redirect_target(#{config := #{dashboard_addr := Addr}}, Username, Role, Token) ->
-    LoginMeta = emqx_dashboard_sso_api:login_meta(Username, Role, Token, oidc),
+code_redirect_target(#{config := #{dashboard_addr := Addr}}, Username, Payload) ->
+    SsoUsername = ?SSO_USERNAME(?BACKEND, Username),
+    Code = emqx_dashboard_sso_code:create_code(SsoUsername, Payload),
+    LoginMeta = #{code => Code, username => Username},
     MetaBin = base64:encode(emqx_utils_json:encode(LoginMeta)),
```
zmstone: encode with urlsafe option.

apps/emqx_dashboard/src/emqx_dashboard_mfa.erl
```diff
@@ -122,3 +234,50 @@ totp_error(Reason) ->
...
+                _ ->
+                    false
+            end;
+        _ ->
+            false
+    end;
+verify_signed_token(_, _) ->
+    false.
+
+compute_hmac(Payload) ->
```
zmstone: there is maybe no need to sign now. the signature was to avoid table scan triggered by untrusted tokens, now it's just a lookup, which is probably cheaper than sha256

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13516)

PR #16951, opened 2026-03-20
refactor(logger): do not log 'domain' meta to final JSON log field

zmstone:
Release version: 6.2.0

'domain' in log meta is mostly for log handlers to filter logs, end users, it duplicates with 'tag'
so we remove it starting from 6.2


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12906)

PR #16952, opened 2026-03-20
chore: unhide root schema docs for rule_engine, message_transformation, and schema_validation

zmstone:
Release versions: 5.8.10, 5.9.3, 5.10.4
These roots were marked as `IMPORTANCE_HIDDEN`, which excluded them
from the generated schema documentation (`schema-en.json`).

Since users may need to reference these configurations in the docs,
unhide them so they appear in the schema doc dump.

Follows the same approach as commit 69377e237e which unhid `rule_engine`
for v6.0+ (EMQX-14408). This backports the same fix to `release-58`
and extends it to `message_transformation` and `schema_validation`.


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13303)

PR #16957, opened 2026-03-20
refactor(sub-filter): drop message in session process instead of broker

zmstone:
## Summary

- Revert broker-level subscription message filtering (commit 345dd24818) so that filtering only happens in the session process.
- The feature is disabled by default, so the extra ETS lookup in the broker hot path does not justify the cost.
- Durable sessions already filter in the session process; this keeps behavior consistent.

Release version: 6.2.0


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12908)

Issue #16950, opened 2026-03-20
Expose min_bytes and sleep_timeout brod consumer options in Kafka source configuration


meowcakes:
### What would you like to be added or enhanced?

Description:

  The Kafka consumer source (sources.kafka_consumer) does not expose brod's min_bytes or sleep_timeout consumer options. The current ConsumerConfig in emqx_bridge_kafka_impl_consumer.erl only passes four options to brod:

  ConsumerConfig = [
      {begin_offset, BeginOffset},
      {max_bytes, MaxBatchBytes},
      {max_wait_time, MaxWaitTime},
      {offset_reset_policy, OffsetResetPolicy}
  ],

  This results in brod using its defaults of min_bytes = 0 and sleep_timeout = 1000. With min_bytes = 0, the Kafka broker returns fetch responses immediately even when there is no data. Brod then sleeps for sleep_timeout (1
  second) before issuing the next fetch. This creates up to 1 second of latency for new messages arriving on an idle partition.

  Setting min_bytes = 1 would cause the broker to hold the fetch request in purgatory until at least 1 byte of data is available (bounded by max_wait_time), which is true long-polling. Combined with sleep_timeout = 0, this
  would reduce consumer latency to near-zero for new messages.

  Proposed change:

  Expose min_bytes and sleep_timeout as configurable parameters in the Kafka consumer source schema.

  ConsumerConfig = [
      {begin_offset, BeginOffset},
      {max_bytes, MaxBatchBytes},
      {max_wait_time, MaxWaitTime},
      {min_bytes, MinBytes},
      {sleep_timeout, SleepTimeout},
      {offset_reset_policy, OffsetResetPolicy}
  ],

### Why is this needed?

We use the Kafka consumer source to bridge messages from Kafka back to MQTT device inbox topics. These are interactive state updates where users expect near-real-time feedback. The current
  ~1 second worst-case latency from the polling behavior is too high.

zmstone:
@meowcakes thank you for reporting this.
we'll change the default to use min_bytes = 1 and sleep_timeout = 10 in 5.8, 5.9, 5.10, 6.0, 6.1, and 6.2
but only add new options in 6.2

zmstone:
closing this issue.

release eta:

- 6.2.0 20260331
- 5.8.10 20260416
- 6.1.2 20260423
- 5.10.4 20260514

PR #16961, opened 2026-03-21
fix(kafka-consumer): ensure long poll

zmstone:
Fixes  https://github.com/emqx/emqx/issues/16950

Release version: 5.8.10, 5.9.3, 5.10.4

## Summary

- Ensure Kafka consumer fetch requests wait briefly for data instead of returning empty batches immediately when the topic is idle.
- Set `min_bytes = 1` so the broker can long-poll up to the configured `max_wait_time`, which avoids the extra idle delay caused by immediate empty fetches.
- Added an end-user-facing changelog entry for the polling behavior improvement.

## PR Checklist


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13304)

PR #16963, opened 2026-03-22
test(mgmt): fix flaky t_persistent_sessions5

zmstone:
## Summary

- Fix flaky `t_persistent_sessions5` by also asserting `count=0` in `init_per_testcase`, not just empty data.
- The persistent session bookkeeper's cached disconnected session count can be stale from a previous test, causing the total count to overshoot (e.g. 5 instead of expected 4).
- The retry now waits for the bookkeeper to refresh its cached count (100ms interval) before proceeding.

Release version: 6.2.0


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12916)

PR #16964, opened 2026-03-22
build: pin jq v0.3.16 for macos-15

zmstone:


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12917)

PR #16969, opened 2026-03-23
fix(emqx_management): preserve active plugin version on delete

zmstone:
Fixes EMQX-15193

## Summary

- Fix the plugin delete API so uninstalling an inactive conflicting version does not stop and purge the active older version.
- Keep delete idempotent by logging and continuing on non-fatal disable/uninstall failures, while still failing if stopping a running plugin fails.
- Add a regression test covering install old version, start it, install conflicting new version, delete the new version, and verify the old version remains running.

## PR Checklist
  This regression has not been released yet, so no changelog entry is needed.

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
@@ -759,78 +759,7 @@ delete_package(NameVsn) ->
 
 %% For RPC plugin delete
 delete_package(NameVsn, _Opts) ->
-    _ = emqx_plugins:forget_allowed_installation(NameVsn),
-    case has_other_active_version(NameVsn) of
-        true ->
-            ok = emqx_plugins:delete_state(NameVsn),
-            ok = emqx_plugins:purge(NameVsn),
-            _ = emqx_plugins:delete_package(NameVsn),
-            ok;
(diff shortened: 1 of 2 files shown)


apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -754,22 +754,86 @@ describe_package(NameVsn) ->
...
+-spec delete_package(name_vsn(), map()) -> ok | {error, any()}.
 delete_package(NameVsn, _Opts) ->
     _ = emqx_plugins:forget_allowed_installation(NameVsn),
-    case emqx_plugins:ensure_stopped(NameVsn) of
-        ok ->
-            _ = emqx_plugins:ensure_disabled(NameVsn),
-            _ = emqx_plugins:ensure_uninstalled(NameVsn),
+    case has_other_active_version(NameVsn) of
+        true ->
+            ok = emqx_plugins:delete_state(NameVsn),
```
savonarola: Nit: I would rather incapsulate this into `emqx_plugins` module to reduce granularity of operations

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-12984)

PR #16979, opened 2026-03-24
fix(bridge): support MQTT queue ingress over subids

zmstone:
Fixes EMQX-15138

Release version:
6.1.2, 6.2.0

## Summary

This updates MQTT ingress bridge routing so remote queue subscriptions using `$queue/{name}/{bind-filter}` work correctly when the remote broker supports MQTT 5 Subscription Identifiers.

The connector now keeps a dedicated index for Subscription Identifiers alongside the existing topic index. Ingress subscriptions use Subscription Identifiers when available, dispatch incoming publishes by Subscription Identifier first, and fall back to topic matching otherwise. Queue subscriptions are never indexed by topic because queue deliveries arrive on the original publish topic rather than the `$queue/...` filter.

For brokers that reject Subscription Identifiers with SUBACK `0xA1`, normal topic subscriptions retry without the identifier while `$queue/...` subscriptions remain rejected because they cannot be routed safely without it.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
+++ b/apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
@@ -331,10 +331,6 @@ get_source_metrics_api(Config) ->
 simple_create_rule_api(TCConfig) ->
     emqx_bridge_v2_testlib:simple_create_rule_api(TCConfig).
 
-mq_ingress_test_callback(Msg, Pid) ->
-    Pid ! {mq_ingress_consumed, Msg},
-    ok.
-
 start_client(TCConfigOrNode) ->
     start_client(TCConfigOrNode, _Opts = #{}).
 
@@ -485,94 +481,66 @@ t_mqtt_conn_bridge_ingress_shared_subscription(TCConfig) ->
     ok.
 
 -doc "MQTT ingress dispatches queue deliveries by Subscription Identifier.".
-t_mqtt_conn_bridge_ingress_subid_dispatch(_TCConfig) ->
-    SubscriptionIdToHandlerIndex = ets:new(?MODULE, [set, public]),
-    TopicToHandlerIndex = emqx_topic_index:new(),
-    BridgeName = <<"bridge:mqtt:test">>,
-    Conf =
-        emqx_bridge_mqtt_ingress:config(
+t_mqtt_conn_bridge_ingress_subid_dispatch(TCConfig) ->
+    {201, _} = create_connector_api(TCConfig, #{}),
+    {201, _} =
+        create_source_api(TCConfig, #{
+            <<"parameters">> => #{
+                <<"topic">> => <<"$queue/orders/t/#">>,
+                <<"qos">> => 1
+            }
+        }),
+    ConnResId = connector_resource_id(TCConfig),
+    {ok, _, #{state := #{installed_channels := InstalledChannels}}} =
+        emqx_resource:get_instance(ConnResId),
+    ?assertMatch(
+        [
             #{
-                ingress_list => [
+                remote := #{topic := <<"$queue/orders/t/#">>},
+                ingress_list := [
                     #{
-                        server => <<"127.0.0.1:1883">>,
-                        remote => #{
-                            topic => <<"$queue/orders/t/#">>,
-                            qos => 1,
-                            no_local => false
-                        },
-                        on_message_received => {?MODULE, mq_ingress_test_callback, [self()]}
+                        remote := #{topic := <<"$queue/orders/t/#">>},
+                        subscription_id := 1
                     }
                 ]
-            },
-            BridgeName,
-            SubscriptionIdToHandlerIndex,
-            TopicToHandlerIndex
-        ),
-    ?assertMatch(
-        #{
-            ingress_list := [
-                #{
-                    remote := #{topic := <<"$queue/orders/t/#">>},
-                    subscription_id := 1
-                }
-            ]
-        },
-        Conf
+            }
+        ],
+        maps:values(InstalledChannels)
     ),
-
-    ok =
-        emqx_bridge_mqtt_ingress:handle_publish(
-            #{
-                dup => false,
-                payload => <<"hello">>,
-                properties => #{'Subscription-Identifier' => 1},
-                qos => 1,
-                retain => false,
-                topic => <<"t/1">>
-            },
-            BridgeName,
-            SubscriptionIdToHandlerIndex,
-            TopicToHandlerIndex
-        ),
-    ?assertReceive(
-        {mq_ingress_consumed, #{topic := <<"t/1">>, payload := <<"hello">>}},
-        1_000
-    ),
+    #{topic := RepublishTopic} = simple_create_rule_api(TCConfig),
+    C = start_client(TCConfig),
+    {ok, _, [_]} = emqtt:subscribe(C, RepublishTopic, [{qos, 1}]),
+    {ok, _} = emqtt:publish(C, <<"orders/t/1">>, <<"hello">>, [{qos, 1}]),
+    {publish, #{topic := RepublishTopic, payload := PayloadBin}} =
+        ?assertReceive({publish, _}, 3_000),
+    ?assertMatch(#{<<"payload">> := <<"hello">>}, emqx_utils_json:decode(PayloadBin)),
+    ?assertNotReceive({publish, _}),
     ok.
 
 -doc "MQTT ingress rejects queue subscriptions when Subscription Identifiers are unavailable.".
 t_mqtt_conn_bridge_rejects_queue_source_without_subscription_identifier(TCConfig) ->
-    TopicToHandlerIndex = emqx_topic_index:new(),
-    State = #{
-        installed_channels => #{},
-        pool_name => connector_resource_id(TCConfig),
-        pool_size => 1,
-        subscription_id_to_handler_index => undefined,
-        topic_to_handler_index => TopicToHandlerIndex,
-        server => <<"127.0.0.1:1883">>
-    },
-    ChannelConfig = #{
-        hookpoints => [],
-        parameters => #{
-            topic => <<"$queue/orders/t/#">>,
-            qos => 1
+    {201, _} = create_connector_api(TCConfig, #{
+        <<"proto_ver">> => <<"v3">>
+    }),
+    {400, #{
+        <<"message">> := #{
+            <<"kind">> := <<"validation_error">>,
+            <<"reason">> := <<"queue subscriptions require connector proto_ver = v5">>
         }
-    },
-    ?assertError(
-        {unrecoverable_error, subscription_identifier_required_for_queue_subscription},
-        emqx_bridge_mqtt_connector:on_add_channel(
-            connector_resource_id(TCConfig),
-            State,
-            <<"source:mqtt:test">>,
-            ChannelConfig
-        )
-    ),
+    }} =
+        create_source_api(TCConfig, #{
+            <<"parameters">> => #{
+                <<"topic">> => <<"$queue/orders/t/#">>,
+                <<"qos">> => 1
+            }
+        }),
     ok.
 
-t_mqtt_conn_bridge_ingress_retries_without_subid_on_a1() ->
-    [{cluster, true}].
-
 -doc "MQTT ingress retries normal topic subscriptions without Subscription Identifier on SUBACK 0xA1.".
+t_mqtt_conn_bridge_ingress_retries_without_subid_on_a1() ->
+    [{matrix, true}].
+t_mqtt_conn_bridge_ingress_retries_without_subid_on_a1(matrix) ->
+    [[?cluster]];
 t_mqtt_conn_bridge_ingress_retries_without_subid_on_a1(TCConfig) ->
     Self = self(),
     Nodes = get_config(nodes, TCConfig),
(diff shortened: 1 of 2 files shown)


apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -166,14 +174,20 @@ on_add_channel(
     %% Add ingress channel
     RemoteParams0 = maps:get(parameters, ChannelConfig),
     {LocalParams, RemoteParams} = take(local, RemoteParams0, #{}),
+    ok = ensure_queue_subscription_supported(RemoteParams, SubscriptionIdToHandlerIndex),
```
savonarola, line 189: Do we intentionally crash with `badmatch` here if ensure fails?

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -99,16 +100,22 @@ on_start(ResourceId, #{server := Server} = Conf) ->
         config => emqx_utils:redact(Conf)
     }),
     TopicToHandlerIndex = emqx_topic_index:new(),
-    StartConf = Conf#{topic_to_handler_index => TopicToHandlerIndex},
+    SubscriptionIdToHandlerIndex = maybe_new_subscription_id_index(Conf),
```
savonarola, line 103: Is it intended to create `ets` in the context of the process calling this callback?

zmstone: tbh, i did not try to reason, the topic to handler index table created at line 102 is also owned by the same caller.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
zmstone: will try to change it to throw.

zmstone: done, changed to `error(...)`

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -519,7 +545,42 @@ mk_ingress_config(
...
+
+maybe_new_subscription_id_index(#{proto_ver := v5}) ->
+    ets:new(?MODULE, [set, public, {read_concurrency, true}]);
+maybe_new_subscription_id_index(_Conf) ->
+    undefined.
+
+maybe_delete_subscription_id_index(undefined) ->
+    ok;
+maybe_delete_subscription_id_index(SubscriptionIdToHandlerIndex) ->
+    ets:delete(SubscriptionIdToHandlerIndex).
```
keynslug: Nit. `undefined` is already matched at the calling site.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -519,7 +545,42 @@ mk_ingress_config(
...
+        false ->
+            ok;
+        true ->
+            {error, {unrecoverable_error, subscription_identifier_required_for_queue_subscription}}
+    end.
+
+is_queue_topic(<<"$queue/", _/binary>>) ->
+    true;
+is_queue_topic(_) ->
+    false.
```
keynslug, line 585: It does not sound entirely fair: here we are assuming that the remote broker is EMQX with Queue feature enabled. If it's not EMQX or EMQX with Queues disabled, shouldn't we treat it just as a regular subscription? E.g. can be unexpected for the user to get an error if the remote is not even EMQX.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -622,16 +685,17 @@ mk_emqtt_client_opts(
     ClientId = maps:get(clientid, ClientidInfo),
     Username = maps:get(username, ClientidInfo, undefined),
     Password = maps:get(password, ClientidInfo, undefined),
-    ClientOpts1 = ClientOpts0#{
-        clientid := ClientId,
-        msg_handler => mk_client_event_handler(Name, TopicToHandlerIndex)
+    ClientOpts1 = ClientOpts0#{clientid := ClientId},
+    ClientOpts1a = ClientOpts1#{
+        msg_handler =>
+            mk_client_event_handler(Name, SubscriptionIdToHandlerIndex, TopicToHandlerIndex)
```
keynslug: Nit. Why separate oddly-named binding?

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_ingress.erl
```diff
@@ -150,56 +195,111 @@ on_reconnect(ClientPid, ReconnectContext) ->
...
 unsubscribe_remote_topics(
-    Pid, IngressList, WorkerIdx, PoolSize, Name, ChannelId, TopicToHandlerIndex
+    Pid,
+    IngressList,
+    WorkerIdx,
+    PoolSize,
+    Name,
+    ChannelId,
+    SubscriptionIdToHandlerIndex,
+    TopicToHandlerIndex
```
keynslug: Nit. Perhaps just collapse into the callsite for simplicity?

keynslug (commented):
Great idea! 💡

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
zmstone: given it's a $ prefixed topic, the risk is low.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
zmstone: there are too callers, simplified the caller instead.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
zmstone: fixed.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_ingress.erl
zmstone: unsure what this means, could you explain ?

keynslug: Sorry, I meant to say that having separate `unsubscribe_remote_topics` function does not seem to give nothing besides inflating the code, so it sounds natural to drop it and inline into the only callsite.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
keynslug: I see your point. Introducing such specifics into a generic MQTT bridge that irreversibly affects interoperability still sounds questionable to me. For example, a natural followup question is "why `$queue` but not `$stream`?".

A couple of other options to perhaps make this "smoother", i.e. start with a more permissive approach.
1. Move this behind some connector schema option that basically says "remote is EMQX 6"?
2. Turn this precondition into a set of hints on the Dashboard side (e.g. suggest to enable v5 if user puts `$queue/...` into the topic field) and a highlight in the relevant documentation chapter?

Anyway, agree that impact is pretty slim, no strong opinions here.

zmstone: will consider add an source config in future version.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -315,6 +348,8 @@ on_stop(ResourceId, State) ->
             _ ->
                 #{}
         end,
+    SubscriptionIdToHandlerIndex = maps:get(subscription_id_to_handler_index, StateMap, undefined),
+    ok = maybe_delete_subscription_id_index(SubscriptionIdToHandlerIndex),
```
thalesmg: maybe we could use the allocation api already in place for this new resource

apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
```diff
@@ -668,7 +838,7 @@ t_resubscribe_on_fast_failure() ->
 t_resubscribe_on_fast_failure(matrix) ->
     [
         [?cluster, ProtoVer, CleanStart]
-     || ProtoVer <- [?proto_v3, ?proto_v5],
```
thalesmg, line 671: Q: why not keep testing for proto_v3 here?

apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
```diff
@@ -450,6 +484,142 @@ t_mqtt_conn_bridge_ingress_shared_subscription(TCConfig) ->
...
+            <<"source:mqtt:test">>,
+            ChannelConfig
+        )
+    ),
+    ok.
+
+t_mqtt_conn_bridge_ingress_retries_without_subid_on_a1() ->
+    [{cluster, true}].
+
+-doc "MQTT ingress retries normal topic subscriptions without Subscription Identifier on SUBACK 0xA1.".
```
thalesmg:
```suggestion
-doc "MQTT ingress retries normal topic subscriptions without Subscription Identifier on SUBACK 0xA1.".
t_mqtt_conn_bridge_ingress_retries_without_subid_on_a1() ->
    [{matrix, true}].
t_mqtt_conn_bridge_ingress_retries_without_subid_on_a1(matrix) ->
    [[?cluster]];
```

apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
```diff
@@ -450,6 +484,142 @@ t_mqtt_conn_bridge_ingress_shared_subscription(TCConfig) ->
...
+    ?assertError(
+        {unrecoverable_error, subscription_identifier_required_for_queue_subscription},
+        emqx_bridge_mqtt_connector:on_add_channel(
+            connector_resource_id(TCConfig),
+            State,
+            <<"source:mqtt:test">>,
+            ChannelConfig
+        )
+    ),
+    ok.
```
thalesmg, line 537: having tests call directly the connection callback functions makes it harder to later refactor stuff, as they depend on internals.  

it's always best to use the APIs (`create_x_api`).

since this case is entirely controlled by configuration (simply set `proto_ver /= v5`), there seems to be no reason to break this convention.

apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
```diff
@@ -450,6 +484,142 @@ t_mqtt_conn_bridge_ingress_shared_subscription(TCConfig) ->
     ),
     ok.
 
+-doc "MQTT ingress dispatches queue deliveries by Subscription Identifier.".
+t_mqtt_conn_bridge_ingress_subid_dispatch(_TCConfig) ->
+    SubscriptionIdToHandlerIndex = ets:new(?MODULE, [set, public]),
+    TopicToHandlerIndex = emqx_topic_index:new(),
+    BridgeName = <<"bridge:mqtt:test">>,
+    Conf =
+        emqx_bridge_mqtt_ingress:config(
```
thalesmg: same comment about depending on implementation internals: this seems like it could just use the normal APIs to create the connector and source, and then use normal publishing to see it work?  if one wants to see the connector state has the expected subscription_id, still create it via the APIs, and then consult it via `emqx_resource:get_instance/1`.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
zmstone: what do you mean?

zmstone: emqx_resource:allocate_resource ?

apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
zmstone: will restore

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
thalesmg: yes, it's currently being used for the pool name only: https://github.com/emqx/emqx/blob/f88ea152dcca9f109f968fb98ca58b79e46804ef/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl#L271

then cleared here: https://github.com/emqx/emqx/blob/f88ea152dcca9f109f968fb98ca58b79e46804ef/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl#L324-L330

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13041)

PR #16986, opened 2026-03-26
fix(crl): use verify_none for HTTPS CRL fetches

zmstone:
it's a follow up fix for PR #16683
## Summary

- Fix CRL fetch over HTTPS failing with `unknown_ca` TLS error
- `emqx_crl_cache:http_get/2` now passes `{ssl, [{verify, verify_none}]}` for HTTPS URLs
- `verify_none` is intentional: verifying the CRL server's cert could trigger another CRL fetch, causing infinite recursion
- Added `t_https_crl_fetch` test with a cowboy TLS server to validate the fix


apps/emqx/test/emqx_crl_cache_SUITE.erl
```diff
@@ -962,6 +1002,17 @@ t_filled_cache(Config) ->
 
 %% If the CRL is not cached when the client tries to connect and the
 %% CRL server is unreachable, the client will be denied connection.
```
thalesmg, line 1004: this comment is for the test below the introduced one.

apps/emqx/test/emqx_crl_cache_SUITE.erl
```diff
@@ -212,6 +212,42 @@ init_per_testcase(TestCase, Config) when
...
+    {CrlServerCert, CrlServerKey} = emqx_cth_tls:gen_cert(#{
+        key => ec, issuer => CrlServerCA, subject => #{name => "localhost"}
+    }),
+    {CertFile, KeyFile} = emqx_cth_tls:write_cert(
+        CertDir, "crl-server", {CrlServerCert, CrlServerKey}
+    ),
+    {CACertFile, _} = emqx_cth_tls:write_cert(CertDir, "crl-server-ca", CrlServerCA),
+    application:ensure_all_started(cowboy),
+    ok = application:ensure_started(inets),
+    ok = application:ensure_started(ssl),
```
thalesmg, line 234: it would be better to add these apps to `cth_suite:start`, so that they are teared down by cth_suite when stopping the test suite.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13000)

PR #16991, opened 2026-03-26
fix(test): fix flaky t_reconnect in mqtt action suite

zmstone:
## Summary
- Replace `emqx_utils:drain_down/1` (zero timeout) with a proper receive loop that waits up to 5s for DOWN messages in `t_reconnect`
- `drain_down/1` uses `after 0` so it only collects messages already in the mailbox; after the HTTP kick returns, channel processes may not have terminated yet, causing intermittent assertion failures


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13043)

PR #16992, opened 2026-03-27
feat(gcp wif): add support for specifying `audience` for initial token

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-15205

Release version:
6.2.0

## Summary

Although the initial scope of the WIF auth feature was only for Microsoft Entra ID (which already works), we add the option of specifying `audience` here to expand the possible number of OIDC providers that could be used with this authentication method.  One such example is Auth0, which then can be used by specifying this new parameter.

## PR Checklist

- [x] The changes are covered with new or existing tests (also verified manually with Auth0 and BigQuery)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13046)

PR #16999, opened 2026-03-30
fix(mq): include Subscription-Identifier in queue message deliveries

zmstone:
Fixes EMQX-15211

Release version: 6.2.0

## Summary

When the MQ feature delivers messages to `$queue/` subscribers, the MQTT v5
Subscription-Identifier property was missing from PUBLISH packets. This broke
MQTT bridge ingress routing for queue subscriptions — the connector received
messages but the source statistics stayed at 0 because messages could not be
dispatched to the correct channel.

**Root cause**: `emqx_session:enrich_deliver` looked up subscription options by
the actual message topic (e.g., `t/1`), but the session stores queue
subscriptions by the full queue topic (`$queue/test/t/#`). The lookup failed,
so `SubOpts` was `undefined` and no Subscription-Identifier was added.

**Fix**: Add the original queue subscription topic as a message header
(`mq_sub_topic`) in `emqx_mq:delivers/2`, and use it in `enrich_deliver` to
look up subscription options — following the same pattern as `redispatch_to`.

Key files:
- `apps/emqx_mq/src/emqx_mq.erl` — set `mq_sub_topic` header in deliveries
- `apps/emqx/src/emqx_session.erl` — use `mq_sub_topic` header in `enrich_deliver`

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_session.erl
+++ b/apps/emqx/src/emqx_session.erl
@@ -479,6 +479,9 @@ enrich_deliver(ClientInfo, {deliver, Topic, Msg}, UpgradeQoS, Session) ->
         case Msg of
             #message{headers = #{redispatch_to := ?REDISPATCH_TO(Group, T)}} ->
                 ?IMPL(Session):get_subscription(emqx_topic:make_shared_record(Group, T), Session);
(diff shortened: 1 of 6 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13063)

PR #17000, opened 2026-03-30
fix(mq): handle undefined topic_filter in make_sub_topic

zmstone:
Fixes a regression introduced in #16999.

Release version: 6.2.0

## Summary

When subscribing to a queue without a topic filter (e.g., `$queue/name`),
`emqx_mq_sub:name_topic/1` returns `{Name, undefined}`. The binary
construction in `make_sub_topic/1` crashed with `badarg` because
`undefined` is not a binary, causing all MQ message deliveries in that
session to fail with `hook_callback_exception`.

Fix by pattern-matching the `undefined` case and constructing
`<<"$queue/", Name/binary>>` without the topic filter suffix.

## PR Checklist

### DIFF
--- a/apps/emqx_mq/src/emqx_mq.erl
+++ b/apps/emqx_mq/src/emqx_mq.erl
@@ -374,8 +374,12 @@ make_sub_topic(SubscriberRef) ->
         undefined ->
             undefined;
         Sub ->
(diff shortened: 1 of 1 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13064)

PR #17001, opened 2026-03-30
fix(mq): backport Subscription-Identifier fix for queue deliveries

zmstone:
Backport of #16999 and #17000 from release-62.

Release version: 6.1.2

## Summary

When the MQ feature delivers messages to `$queue/` subscribers, the MQTT v5
Subscription-Identifier property was missing from PUBLISH packets. This broke
MQTT bridge ingress routing for queue subscriptions — the connector received
messages but the source statistics stayed at 0 because messages could not be
dispatched to the correct channel.

**Root cause**: `emqx_session:enrich_deliver` looked up subscription options by
the actual message topic (e.g., `t/1`), but the session stores queue
subscriptions by the full queue topic (`$queue/test/t/#`). The lookup failed,
so `SubOpts` was `undefined` and no Subscription-Identifier was added.

**Fix** (PR #16999): Add the original queue subscription topic as a message header
(`mq_sub_topic`) in `emqx_mq:delivers/2`, and use it in `enrich_deliver` to
look up subscription options.

**Follow-up fix** (PR #17000): Handle the case where the topic filter is `undefined`
(subscribing to `$queue/name` without a topic filter) to avoid `badarg` crash
in `make_sub_topic/1`.

## PR Checklist


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13436)

PR #17012, opened 2026-04-02
fix(authn): let auth chain continue when password is undefined

zmstone:
Fixes EMQX-15214

Release version: 5.8.10, 5.10.4

## Summary

When a client connects without a password (password field absent in CONNECT packet),
password-based authenticators (built-in DB, MySQL, PostgreSQL, MongoDB, Redis, LDAP)
previously returned `{error, bad_username_or_password}`, which stopped the entire
authentication chain.

This was introduced in PR #11490 as a "quick return" optimization, but the choice of
`{error, ...}` over `ignore` was incorrect — it prevents subsequent authenticators
(e.g. HTTP, JWT, or certificate-based) from being tried.

Now these backends return `ignore` when password is `undefined`, consistent with
the "user not found" behavior, allowing the auth chain to continue.

## PR Checklist

### DIFF
--- a/apps/emqx_auth_ldap/src/emqx_auth_ldap.app.src
+++ b/apps/emqx_auth_ldap/src/emqx_auth_ldap.app.src
@@ -1,7 +1,7 @@
 %% -*- mode: erlang -*-
 {application, emqx_auth_ldap, [
     {description, "EMQX LDAP Authentication and Authorization"},
-    {vsn, "0.1.5"},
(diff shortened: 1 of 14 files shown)


savonarola (dismissed):

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13445)

PR #17019, opened 2026-04-07
perf(prometheus): avoid doing further rpc calls while gathering mria stats (r58)

### LINKED ISSUE #16924: prometheus scrap timeout in replicant nodes
## Problem
When there are intense retained messages writes from replicant nodes, the `erpc` calls may start to jam.
This in turn causes the erpc-heavy prometheus metrics scraping to timeout.
Below is the call chain:

```
GET /api/v5/prometheus/stats
  → emqx_prometheus_api:stats/2
    → emqx_prometheus:collect/1
      → prometheus_text_format:format(default)
        → emqx_prometheus:collect_mf/2
          → emqx_prometheus:fetch_from_local_node/1       ← 4-6s
            → mria_data(replicant, Mode)
              → for each metric in [replicant_lag,
                                     replicant_bootstrap_time,
                                     replicant_bootstrap_num_keys,
                                     replicant_message_queue_len,
                                     replicant_stream_count]:
                  → for each shard (16 shards):
                      → mria_status:get_shard_stats(Shard)   ← erpc:call inside
```

## Solution
After https://github.com/emqx/emqx/pull/16913, the performance is better, but there is still room for improvement.
The idea is to add a new API handler which aggregates the metrics and return in one rpc call.

## Notes
- Fix is required in 5.8.10, 5.9.2, 5.10.4, 6.0.3, 6.1.2 and 6.2.0.
- Fallback is required during rolling upgrade (if `erpc:call` raises `undef` exception).

No timeout.

  Use the attached reproducer script: [reproduce-retainer-prom-timeout.sh](https://github.com/user-attachments/files/26030750/reproduce-retainer-prom-timeout.sh)

  ```bash
  ./reproduce-retainer-prom-timeout.sh start
```

  This script starts a 3-container setup with Docker:

  - core: an EMQX Enterprise core node
  - replicant: an EMQX Enterprise replicant node
  - bench: emqtt-bench, publishing retained messages to the replicant at high rate

  ### Topology

  Host ports:

  - 18083 -> core dashboard/API
  - 18084 -> replicant dashboard/API
  - 1883 -> replicant MQTT listener

  The benchmark container connects to the replicant and continuously publishes retained QoS 0 messages with many clients.

  ### What the script does

  1. Starts a core + replicant cluster with static discovery.
  2. Exposes the replicant dashboard API on http://127.0.0.1:18084.
  3. Starts emqtt-bench against the replicant MQTT listener on 127.0.0.1:1883.
  4. Uses many concurrent clients and a very small publish interval to create a retained-message storm.
  5. Monitors the replicant while load is running:
      - periodically calls GET /api/v5/prometheus/stats
      - records scrape latency and timeouts
      - inspects process mailbox growth via emqx eval
      - tails benchmark logs
### Why this reproduces the issue

  Under sustained retained-message load on the replicant, mailbox pressure builds up and Prometheus scraping on the replicant becomes very slow. In the problematic version, repeated calls
  to /api/v5/prometheus/stats may become extremely slow or time out.

  ### Expected reproduction signal

  During the run, check the generated monitor logs under /tmp/retainer-repro-<timestamp>/.

  The main signal is in prom.log. In the problematic case it shows entries like:

  - PROM: status=200 time=7000ms
  - PROM: status=TIMEOUT ...

  mailbox.log may also show many processes with large message_queue_len, often involving erpc:call/5.

  ### Stop the reproducer

  ./scripts/reproduce-retainer-prom-timeout.sh stop

_No response_

### EMQX version

5.8.6

### OS version

_No response_

### Log files

_No response_

### PR DESCRIPTION
**Pull request by thalesmg**
Port of https://github.com/emqx/emqx/pull/17018 to `release-58`.

Fixes https://emqx.atlassian.net/browse/EMQX-15217

Fixes https://github.com/emqx/emqx/issues/16924

Release version:
5.8.10, 5.10.4

## Summary

With https://github.com/emqx/mria/pull/202, we introduce a new stats gathering function that does not perform RPC calls, which currently means that, only for replicants, we don't call `mria_status:get_shard_lag/1` in it.  In `emqx_prometheus`, we introduce a cache process that periodically calls `mria_status:get_shard_lag/1` for all shards and store their value in a cache table.  This table is then read when the Prometheus scraping endpoint is called, thus avoiding having to do one RPC call per shard per API call.

## PR Checklist

### DIFF
--- a/apps/emqx_prometheus/src/emqx_prometheus_cache.erl
+++ b/apps/emqx_prometheus/src/emqx_prometheus_cache.erl
@@ -1,0 +1,124 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_prometheus_cache).
+
+-behaviour(gen_server).
+
+%% API
+-export([
+    start_link/1,
+
+    get_mria_shard_lag/1
+]).
+
+%% `gen_server' API
+-export([
+    init/1,
+    terminate/2,
+    handle_continue/2,
+    handle_call/3,
+    handle_cast/2,
+    handle_info/2
+]).
+
+%%------------------------------------------------------------------------------
+%% Type declarations
+%%------------------------------------------------------------------------------
+
+-include_lib("emqx/include/logger.hrl").
+
+-define(TAB, emqx_prometheus_cache).
+-define(VAL_POS, 2).
+-define(DEFAULT_MRIA_SHARD_LAG, 0).
+
+-define(MRIA_SHARD_LAG_KEY(SHARD), {mria_shard_lag, SHARD}).
+
+-define(mria_refresh_shard_lag_timer, mria_refresh_shard_lag_timer).
+-define(undefined, undefined).
+
+%% Calls/casts/infos/continues
+-record(refresh_mria_shard_lag, {}).
+
+%%------------------------------------------------------------------------------
+%% API
+%%------------------------------------------------------------------------------
+
+start_link(Opts) ->
+    gen_server:start_link({local, ?MODULE}, ?MODULE, Opts, []).
+
+get_mria_shard_lag(Shard) ->
+    try
+        ets:lookup_element(?TAB, ?MRIA_SHARD_LAG_KEY(Shard), ?VAL_POS, ?DEFAULT_MRIA_SHARD_LAG)
+    catch
+        error:badarg ->
+            ?DEFAULT_MRIA_SHARD_LAG
+    end.
+
+%%------------------------------------------------------------------------------
+%% `gen_server' API
+%%------------------------------------------------------------------------------
+
+init(_Opts) ->
+    _ = ets:new(?TAB, [named_table, public, ordered_set, {read_concurrency, true}]),
+    State = #{?mria_refresh_shard_lag_timer => ?undefined},
+    {ok, State, {continue, #refresh_mria_shard_lag{}}}.
+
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: apps/emqx_prometheus/src/emqx_prometheus_cache.erl [hunk 1]
```diff
@@ -0,0 +1,124 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_prometheus_cache).
```
zmstone, line 4: to simplify hot-upgrade, maybe we do not add the cache for v5

thalesmg: the core of the fix is using the cache.  this means we don't want to do anything for v5?

thalesmg: ... or do you mean we should instead drop the lag metric?  that would work as well, regarding performance.  but it's a breaking change, API-wise.

zmstone: new version ekka already optimized the number of rpc calls, right ?

thalesmg: no.  it just introduced a new function that does no erpc calls.  the old function still behaves the same.

zmstone: ok, maybe we skip this optimization for v5

Issue #17022, opened 2026-04-08
EMQX fails to start with IPv6-only node name and HTTP auth cannot reach IPv6 (ULA) endpoint (enetunreach)


Napoleon621:
I am running EMQX Enterprise 6.2.0 on a bare metal Linux machine with IPv6-only networking.

I am trying to configure a password-based HTTP auth backend, where the auth service is exposed via an IPv6 ULA address (fdbd::/16).

However:

EMQX fails to connect to the HTTP auth service with:

[warning] tag: AUTHN/WEBHOOK, msg: start_resource_failed, reason: enetunreach

When attempting to switch the node to IPv6, EMQX fails to start with:

Node 'emqx@::1' not responding to pings.
Make sure '::1' is a resolvable and reachable hostname.

[Environment]
EMQX version: 6.2.0 (Enterprise)
Deployment: Bare metal Linux (no Docker)
Network: IPv6-only
Auth backend: HTTP

[Extra info]

- I can successfully curl the auth service from the host terminal, so i suppose there is no problem for the http auth service and Emqx is running in the same net namespace with root.

- It seems everything works fine on a  ipv4/ipv6 dual stack bare metal machine.

Please help take a look at this issue, really appreciate. 
I can provide any further information if needed.
Thanks.

zmstone:
On IPv6-only hosts, EMQX fails because `cluster.proto_dist` defaults to `inet_tcp` (IPv4). The Erlang distribution layer cannot resolve/connect to `::1` using  IPv4, causing the "not responding to pings" error.

Minimum working IPv6-only config (5 settings):

```
cluster.proto_dist = inet6_tcp                    # CRITICAL
node.name = "emqx@::1"                            # no brackets
dashboard.listeners.http.bind = 18083             # the config parser disregard ip part in 'bind'
dashboard.listeners.http.inet6 = true             # workaround is to explicitly opt-in ipv6
listeners.tcp.default.bind = "[::]:1883"          # brackets in bind
```

zmstone:
I cannot reproduce `enetunreach` error towards a mocked http server though.
Did you use a DNS/hostname or IP in the authentication config?

Napoleon621:
I change domain name to a direct ipv6 address and it worked. 
The http auth service is now successfully connected.

PR #17024, opened 2026-04-08
fix(dashboard): infer inet6 from IPv6 bind address

zmstone:
Fixes https://github.com/emqx/emqx/issues/17022

## Summary

When the dashboard HTTP listener is configured with an IPv6 bind address
(e.g. `bind = "[::]:18083"`), the listener now automatically uses `inet6`
socket option.

Previously, users had to explicitly set both `bind = "[::]:18083"` **and**
`inet6 = true`, otherwise the listener would attempt to open an IPv6
address with an IPv4 socket, which silently falls back to IPv4-only.

The fix adds a clause in `ranch_opts/1` that detects an 8-element IP tuple
(IPv6) in the parsed bind address and injects `inet6` into socket options.

## PR Checklist

~~The changes are covered with new or existing tests~~ Not possible to test in ci, manually tested in docker.

### DIFF
--- a/.github/workflows/build_and_push_docker_images.yaml
+++ b/.github/workflows/build_and_push_docker_images.yaml
@@ -95,9 +95,9 @@ jobs:
           echo "==========="
           echo "_EMQX_DOCKER_IMAGE_TAG=$(head -n 1 .emqx_docker_image_tags)" >> $GITHUB_ENV
 
-      - name: Verify that size of docker image is less than 140 MB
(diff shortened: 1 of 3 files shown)


thalesmg (approved):

qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13441)

PR #16894, opened 2026-03-09
test(datalayers): fix flaky test

thalesmg:
```
%%% emqx_bridge_datalayers_SUITE ==> without_batch.sync_query.apiv1_https.t_const_timestamp: FAILED
%%% emqx_bridge_datalayers_SUITE ==>
Failure/Error: ?assertEqual(<<50,48,50,54,45,48,51,45,48,57,84,49,55,58,52,50,58,52,51,46,57,50,53,55,56,52,48,48,48,43,48,48,58,48,48>>, TimeReturned)
  expected: <<"2026-03-09T17:42:43.925784000+00:00">>
       got: <<"2026-03-09T17:42:43.925784+00:00">>
      line: 637
```

https://github.com/emqx/emqx/actions/runs/22865833638/job/66334377370?pr=16893#step:5:462

PR #17033, opened 2026-04-09
test(datalayers): fix broken case

thalesmg:
datalayers for unknown reasons refuses to return data inserted for data older than 1 month, so we are forced to generate a fresh timestamp here...  the fixed timestamp used previously was from 2026-03-09... hence the need to wrap the whole case in `?retry`...


apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_SUITE.erl
```diff
@@ -578,65 +590,62 @@ t_start_ok_no_subject_tags_write_syntax(Config) ->
...
-                            "foo2=\"${payload.foo}\","
-                            "foo3=\"${payload.foo}somestr\","
-                            "bar=5i,baz0=1.1,baz1=\"a\",baz2=\"ai\",baz3=\"au\",baz4=\"1u\" ",
-                            ConstBin/binary
-                        >>
+    %%
+    %% Also, datalayers for unknown reasons refuses to return data inserted for data older
+    %% than 1 month, so we are forced to generate a fresh timestamp here......  Hence the
+    %% need to wrap the whole case in `?retry`......
+    ?retry(500, 10, begin
```
zmstone: fresh timestamp is for sure not older than 1month, why the retry ?

thalesmg: see the first line of the comment

> If the time is "too round", there may be some formatting issues later.

thalesmg: the pr that introduced the existing comment has an example of such formatting issue:

https://github.com/emqx/emqx/pull/16894

zmstone (dismissed):

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_SUITE.erl
zmstone: `(T div 100) * 100 + 1`

thalesmg: good idea.  I've pushed a commit with this to avoid retrying the whole case

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13493)

PR #17042, opened 2026-04-13
fix(coap): keep DTLS token takeover window aligned with keepalive

HJianBo:
## Summary
- Fix regression introduced after #17030 merge in `connection_required` CoAP mode
- Replace fixed 5s `sock_closed` takeover cleanup window with keepalive-based interval
- This preserves stale-channel cleanup while allowing typical DTLS reconnect/publish flows to reuse token

## Root Cause
- `sock_closed` in connected + connection-required mode used a fixed 5s cleanup timer
- For clients like `coap-client` (new DTLS session per command), the next publish often comes after >5s
- Channel was already cleaned up, so token lookup failed with `Invalid token or clientid in connection mode`

## Changes
- `apps/emqx_gateway_coap/src/emqx_coap_channel.erl`
  - derive grace from `emqx_keepalive:info(check_interval, Keepalive)`
- `apps/emqx_gateway_coap/test/emqx_coap_channel_tests.erl`
  - add regression unit test to assert `sock_closed` cleanup uses keepalive interval

## Verification
- `./rebar3 as test eunit --module=emqx_coap_channel_tests`
- `./rebar3 as test eunit --module=emqx_coap_channel_tests,emqx_coap_proxy_conn_tests,emqx_coap_message_tests`

Both commands pass locally.

### DIFF
--- a/apps/emqx_gateway_coap/src/emqx_coap_channel.erl
+++ b/apps/emqx_gateway_coap/src/emqx_coap_channel.erl
@@ -87,7 +87,6 @@
 -define(INFO_KEYS, [conninfo, conn_state, clientinfo, session]).
 
 -define(DEF_IDLE_SECONDS, 30).
--define(SOCK_CLOSED_TAKEOVER_GRACE_MS, 5000).
 -define(RAND_CLIENTID_BYTES, 16).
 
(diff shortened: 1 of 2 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13498)

PR #17045, opened 2026-04-13
fix(authn): let auth chain continue when password is undefined

zmstone:
Fixes 

Release version:

## Summary

When a client sends a CONNECT packet without a password, password-based authenticators should return `ignore` so the authentication chain can continue instead of failing immediately.

This ports the fix to the v6 release-60 branch for the built-in database, MySQL, PostgreSQL, MongoDB, Redis, and LDAP authenticators, and updates the Mnesia API suite assertions to expect `nomatch` metrics for undefined-password attempts.

## PR Checklist


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13497)

HJianBo (approved):

PR #17053, opened 2026-04-13
feat(emqx_mt): add post_auth_tns_expression to reinit tns after authn

zmstone:
Fixes <issue-or-jira-number>

Release version: 6.1.2, 6.2.1

## Summary

Adds a new multi-tenancy config `multi_tenancy.post_auth_tns_expression` (a [Variform](https://docs.emqx.com/en/emqx/latest/configuration/configuration.html#variform-expressions) expression; empty by default = disabled) to derive `client_attrs.tns` from authentication-response attributes such as a `tag` returned by an HTTP auth backend.

Because the existing `client.authenticate` hook fold short-circuits on `{stop, _}` when the authn chain completes, emqx_mt (which hooks at `?HP_HIGHEST`) cannot see auth-response attributes there. This PR introduces a new hook `client.post_authn` fired from `emqx_channel:do_authenticate/2` after `merge_auth_result/2`. emqx_mt registers on it: when the expression is configured it is evaluated against the merged `ClientInfo`, the rendered value replaces `client_attrs.tns`, and the namespace quota check runs against the new value. Empty render keeps pre-auth tns; render error is logged and treated as no-op.

Key files:
- `apps/emqx/src/emqx_channel.erl` — new hook fired after authn success.
- `apps/emqx/src/emqx_hookpoints.erl` — register `client.post_authn` hookpoint.
- `apps/emqx/src/emqx_metrics.erl`, `rel/i18n/emqx_metrics.hocon` — metric counter.
- `apps/emqx_mt/src/emqx_mt_schema.erl` — new field with Variform compile converter.
- `apps/emqx_mt/src/emqx_mt_config.erl` — `get_post_auth_tns_expression/0`.
- `apps/emqx_mt/src/emqx_mt_hookcb.erl` — register `on_post_authn/1`, shared `decide/3` helper.
- `rel/i18n/emqx_mt_schema.hocon` — description with examples.
- `apps/emqx_mt/test/emqx_mt_SUITE.erl` — six new CT cases covering disabled/tag/coalesce/empty/error/quota paths.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_hookpoints.erl
+++ b/apps/emqx/src/emqx_hookpoints.erl
@@ -36,6 +36,7 @@
     'client.check_authz_complete',
     'client.check_authn_complete',
     'client.authenticate',
+    'client.post_authn',
     'client.subscribe',
     'client.unsubscribe',
     'client.timeout',
@@ -155,6 +156,9 @@ when
     emqx_types:clientinfo(), emqx_access_control:authenticate_hook_result()
 ) ->
     fold_callback_result(emqx_access_control:authenticate_hook_result()).
+
+-callback 'client.post_authn'(emqx_types:clientinfo()) ->
+    fold_callback_result(emqx_types:clientinfo() | {error, term()}).
 
 -callback 'client.subscribe'(emqx_types:clientinfo(), emqx_types:properties(), TopicFilters) ->
     fold_callback_result(TopicFilters)
--- a/apps/emqx/src/emqx_hookpoints.erl
+++ b/apps/emqx/src/emqx_hookpoints.erl
@@ -157,8 +157,17 @@ when
 ) ->
     fold_callback_result(emqx_access_control:authenticate_hook_result()).
 
--callback 'client.post_authn'(emqx_types:clientinfo()) ->
-    fold_callback_result(emqx_types:clientinfo() | {error, term()}).
+-type post_authn_context() :: #{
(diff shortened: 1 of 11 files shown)


apps/emqx/src/emqx_hookpoints.erl
```diff
@@ -156,6 +157,9 @@ when
 ) ->
     fold_callback_result(emqx_access_control:authenticate_hook_result()).
 
+-callback 'client.post_authn'(emqx_types:clientinfo()) ->
+    fold_callback_result(emqx_types:clientinfo() | {error, term()}).
```
thalesmg: IMO, it's more future-proof if we adopt the convention of using map arguments and outputs instead of a single values, in case we need to start passing down other things to the same callback, or even return extra stuff from it.

e.g.:


```suggestion
-type post_authn_in() :: #{
    client_info := emqx_types:clientinfo()
    %% in the future, could add other things like:
    %% conninfo => map(), any() => term()
    %%
    %% ... could return extra stuff as well, such as:
    %% log_metadata => map()
}.
-callback 'client.post_authn'(post_authn_context()) ->
    fold_callback_result(post_authn_context() | {error, term()}).
```

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13524)

PR #17056, opened 2026-04-13
fix(dsraft): make drop DB propagate through cluster correctly

keynslug:
Release version: 6.2.1

## Summary

This PR fixes the `emqx_ds:drop_db/1` behavior so that all relevant data is purged across cluster:
* shard and replication storage,
* shared DB metadata,
* runtime DB schemas.

## PR Checklist

### DIFF
--- a/apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_db_sup.erl
+++ b/apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_db_sup.erl
@@ -12,6 +12,7 @@
 -export([
     start_link_db/4,
     whereis_db/1,
+    drop_db/1,
 
     start_shard/1,
     stop_shard/1,
@@ -23,6 +24,7 @@
     start_otx_leader/2,
     stop_otx_leader/2
 ]).
+
 -export([which_dbs/0, which_shards/1]).
 
 %% behaviour callbacks:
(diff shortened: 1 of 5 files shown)


apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
```diff
@@ -267,13 +267,77 @@ t_shards_allocation(Config) ->
...
+                [n_shards_online(N, ?DB) || N <- Nodes],
+                "Each site is responsible for each shard"
+            ),
+
+            %% Drop the DB.
+            ?assertEqual(ok, ?ON(N1, emqx_ds:drop_db(?DB)))
+        end,
+        [
+            check_membership_consistent(?DB, Nodes),
+            {"no DB metadata is present", fun(_Trace) ->
```
ieQu1: Nit: check stage usually consists of pure functions that don't deal with the runtime state. It looks like the idea here is to let all checks run even if one fails. A more idiomatic approach would be to use `?defer_assert` macro in run stage.

ieQu1 (dismissed):

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_db_sup.erl
```diff
@@ -60,6 +62,16 @@ start_link_db(DB, Create, Schema, RTConf) ->
...
     gproc:where(?name(#?db_sup{db = DB})).
 
+-doc "Drops all shared data belonging to the specified DS DB on the local node".
+-spec drop_db(emqx_ds:db()) -> ok | {error, not_found}.
+drop_db(DB) ->
+    case file:del_dir_r(ra_system_data_dir(DB)) of
+        ok ->
+            ok;
+        {error, _Posix} ->
+            {error, not_found}
```
thalesmg: maybe should match on `enoent` before returning `not_found`?  (e.g., could be a permission issue due to bad deployment/ops)

ieQu1 (approved):

yanzhiemq:
[TestExecution](https://emqx.atlassian.net/browse/QA-14773)

PR #17064, opened 2026-04-14
fix(authn): enforce namespace on authn user collection endpoints

zmstone:
Fixes a namespaced-admin bypass on `/api/v5/authentication/:id/users`.

Release version:
6.1.2, 6.2.1

## Summary

The `GET` and `POST` handlers for `/authentication/:id/users` in
`apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl` used to look at the raw
request body and query string to decide the target namespace. A tenant
(namespaced) dashboard admin could therefore:

- list every global authentication user by omitting the `ns` query
  parameter; and
- create a user (including a superuser) in the global namespace by
  omitting the `namespace` body field.

Both handlers now rely on `resolved_ns` produced by the route's filter
callback. The `GET` handler injects it into the query string before
calling `list_users/3`; the `POST` handler uses it as the effective
namespace and rejects any mismatching `namespace` explicitly set in the
body (global admins can still target any namespace via the body).

In addition, an MQTT authentication user in a non-global namespace can no
longer be marked as `is_superuser`. Both `add_user/3` and
`update_user/5` reject such requests so that explicit ACL rules are
always enforced for tenant clients.

Covered by updates and a new case in
`apps/emqx_auth_mnesia/test/emqx_authn_api_mnesia_SUITE.erl`
(`t_ns_admin_bypass_global`, plus adjustments to `t_authenticator_users`
and `t_authenticator_user`).

## PR Checklist
- [x] Change log for changes visible by users has been added to `changes/ee/fix-17064.en.md`

### DIFF
--- a/apps/emqx_auth/mix.exs
+++ b/apps/emqx_auth/mix.exs
@@ -5,7 +5,7 @@ defmodule EMQXAuth.MixProject do
   def project do
     [
       app: :emqx_auth,
-      version: "6.1.0",
+      version: "6.1.1",
       build_path: "../../_build",
       compilers: Mix.compilers() ++ [:copy_srcs],
       # used by our `Mix.Tasks.Compile.CopySrcs` compiler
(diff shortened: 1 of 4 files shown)


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13525)

PR #17070, opened 2026-04-15
fix(emqx_mt): defer managed-ns gate to post-authn when expression is set

zmstone:
This is a followup fix of PR #17053
Release version: 6.1.2

## Summary

When `multi_tenancy.allow_only_managed_namespaces = true` was combined with `multi_tenancy.post_auth_tns_expression`, the pre-authentication hook rejected any client whose pre-auth `client_attrs.tns` was not already a managed namespace (or was unset), before the authentication chain had a chance to return the attributes used to derive the final namespace.

This made the expression effectively unusable in the scenario it was designed for: deriving the tenant namespace from authentication-response attributes (e.g. `client_attrs.tag`).

Fix: when `post_auth_tns_expression` is configured, skip all tns-based gating in the pre-auth `client.authenticate` hook (managed-ns membership, namespace config errors, quota) and let the `client.post_authn` hook enforce them against the rewritten tns.

- `apps/emqx_mt/src/emqx_mt_hookcb.erl`: short-circuit `on_authenticate/2` when `post_auth_tns_expression` is set.
- `apps/emqx_mt/test/emqx_mt_SUITE.erl`: new regression test `t_post_auth_expression_bypasses_preauth_managed_gate` covering both the managed-ns happy path and the post-authn rejection path.

## PR Checklist

### DIFF
--- a/apps/emqx/include/emqx_release.hrl
+++ b/apps/emqx/include/emqx_release.hrl
@@ -19,4 +19,4 @@
 %% NOTE: Also make sure to follow the instructions in end of
 %% `apps/emqx/src/bpapi/README.md'
 
--define(EMQX_RELEASE_EE, "6.1.2-rc.2").
+-define(EMQX_RELEASE_EE, "6.1.2-rc.3").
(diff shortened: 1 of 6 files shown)


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13526)

PR #17075, opened 2026-04-16
fix(ci): publish :latest docker tag on stable releases

### LINKED ISSUE #17069: emqx-enterprise:latest is pointing to 6.0.0 instead of 6.2.0
whenever you use the tag latest, it installs 6.0.0 instead of 6.2.0

zmstone:
Fixes #17069

Release version:
6.0.3, 6.1.2, 6.2.1

## Summary

After the docker publish workflow was simplified (commit 2f8991fbc8), `publish-docker` only pushed the full `<major>.<minor>.<patch>` multiarch tag, so `emqx/emqx-enterprise:latest` on Docker Hub has not been moved since that refactor — which is what users hit in the linked issue.

This PR:

- Threads `inputs.latest` into `publish-docker` and, when it is `true`, also tags the multiarch manifest with `docker.io/<owner>/<profile>:latest` and `docker.io/emqx/emqx:latest`.
- Fixes `scripts/parse-git-ref.sh` so that only the highest existing stable `X.Y.Z` tag resolves to `latest=true` (via the already-defined `is_latest_stable`). Without this, releasing a patch for an older minor (e.g. `6.0.3` after `6.2.0`) would overwrite `:latest`.

`:<major>.<minor>` tags are intentionally not published — they aren't part of our release convention.

## PR Checklist
- [ ] Change log for changes visible by users has been added to \`changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md\` files

### DIFF
--- a/.github/workflows/build_and_push_docker_images.yaml
+++ b/.github/workflows/build_and_push_docker_images.yaml
@@ -258,6 +258,7 @@ jobs:
 
     env:
       PROFILE: ${{ inputs.profile }}
+      DOCKER_LATEST: ${{ inputs.latest }}
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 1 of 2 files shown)


id (approved):

id:
Consider using https://github.com/docker/metadata-action instead of manually assembling tags.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13583)

PR #17078, opened 2026-04-17
feat(mt): inline managed namespace configs in list details

zmstone:
Fixes EMQX-15219

Release version: 6.1.2, 6.2.1

## Summary

This change extends `GET /api/v5/mt/managed_ns_list_details` to inline each managed namespace's configuration in the list response.

The API now returns the namespace config together with `name` and `created_at`, which removes the extra per-namespace config fetch that management UIs previously needed.

The OpenAPI schema was kept backward compatible by using a managed-namespace-specific response shape for this endpoint, so `/mt/ns_list_details` keeps its original contract.

The suite coverage was updated in `emqx_mt_api_SUITE` to verify both the new inlined config payload and the empty-config case.

## PR Checklist

### DIFF
--- a/apps/emqx_mt/src/emqx_mt_api.erl
+++ b/apps/emqx_mt/src/emqx_mt_api.erl
@@ -192,8 +192,8 @@ schema("/mt/managed_ns_list_details") ->
                 #{
                     200 =>
                         emqx_dashboard_swagger:schema_with_examples(
-                            array(ref(ns_with_details_out)),
-                            example_ns_list_details()
+                            array(ref(managed_ns_with_details_out)),
+                            example_managed_ns_list_details()
                         )
                 }
         }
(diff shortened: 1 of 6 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13578)

apps/emqx_mt/src/emqx_mt_api.erl
```diff
@@ -927,7 +953,12 @@ limiter_config_out(Unit0, LimiterConfig) ->
     ).
 
 ns_details_out(Details0) ->
-    Details = undefined_to_null(Details0),
+    Details1 =
+        case Details0 of
+            #{config := Configs} -> Details0#{config := configs_out(Configs)};
+            _ -> Details0
+        end,
```
thalesmg, line 960: nit:


```suggestion
    Details1 = emqx_utils_maps:update_if_present(config, fun configs_out/1, Details0),
```

thalesmg (approved):

PR #17081, opened 2026-04-17
fix(license): enable swagger schema check for session HWM history API

zmstone:
## Summary

- Fix `function_clause` crash on `GET /license/session_hwm_history?period=daily&limit=30` by switching from a content-type-only custom filter to `check_schema => true`, so swagger coerces query params (`limit` to integer, `period` to atom) per the declared schema.
- Simplify the handler to read already-coerced atom-key values directly.


qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13585)

PR #16724, opened 2026-02-10
fix(rabbitmq): recover from connection/channel pid crashes, avoid channel pid leak (r58)

thalesmg:
Port of #16723 

Fixes https://emqx.atlassian.net/browse/EMQX-15098

Release version:
5.8.10, 5.10.4

## PR Checklist

- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files

PR #17082, opened 2026-04-17
fix(rabbitmq): recover from connection/channel pid crashes

zmstone:
This is a hot-patch PR.

Cherry-picked and squashed from PR #16724 (5.10.4).

- Monitor rabbitmq connection and channel processes, recover on crash
- Fix channel process leak when channel creation fails midway
- Add test cases for crash recovery and channel pid leak scenarios

Fixes https://emqx.atlassian.net/browse/EMQX-15224


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13587)

PR #17084, opened 2026-04-17
fix(mqtt source): add handlers to indices before restarting connector when session resumes (r60)

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-15223

Release version:
6.0.3, 6.1.2, 6.2.0

## Summary

Steps to reproduce:

1) Create a connector, source and rule pointing to a toxiproxy server, which can point to
   the broker itself.
2) Cut the connection with toxiproxy.
3) Send a few QoS 1 messages to the source’s topic.
4) Wait for 2 connector health checks, so that the connector state is wiped out by the
   resource manager restart attempts.
5) Restore the connection in toxiproxy.

The new clients do receive the messages, but the ETS table holding the topic → handler
mapping is empty, and thus triggers no action.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
+++ b/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
@@ -92,15 +92,34 @@ resource_type() -> mqtt.
 callback_mode() -> async_if_possible.
 
 -spec on_start(connector_resource_id(), map()) -> {ok, connector_state()} | {error, term()}.
-on_start(ResourceId, #{server := Server} = Conf) ->
-    ?SLOG(info, #{
-        msg => "starting_mqtt_connector",
-        connector => ResourceId,
+on_start(ConnResId, #{server := Server} = Conf) ->
+    ?tp(info, "starting_mqtt_connector", #{
+        connector => ConnResId,
         config => emqx_utils:redact(Conf)
     }),
     TopicToHandlerIndex = emqx_topic_index:new(),
-    StartConf = Conf#{topic_to_handler_index => TopicToHandlerIndex},
-    case start_mqtt_clients(ResourceId, StartConf) of
+    ok = emqx_resource:allocate_resource(
+        ConnResId,
+        ?MODULE,
+        topic_to_handler_index,
+        TopicToHandlerIndex
+    ),
+    %% Introduced in 6.1
+    %% SubscriptionIdToHandlerIndex = maybe_new_subscription_id_index(Conf),
+    %% ok = emqx_resource:allocate_resource(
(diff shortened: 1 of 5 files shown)


apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -659,6 +669,61 @@ mk_client_event_handler(Name, TopicToHandlerIndex) ->
...
 
+%% If we have `clean_start = false`, then we must add the sources' topics to the topic
+%% handler index before starting the MQTT clients.  Otherwise, upon connecting, there
+%% might be some messages arriving that were queued in the session which will have no
+%% handler assigned, and thus will be lost.
+%%
+%% When the channel is properly added to the state, adding the topics to the indices
+%% should be idempotent.
+maybe_add_sources_with_sessions_to_topic_handler(
+    _ConnResId, #{clean_start := true} = _ConnConfig, _PartialConnState
```
zmstone, line 680: Q: `clean_start` always exists in ConnConfig here ?

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -659,6 +669,61 @@ mk_client_event_handler(Name, TopicToHandlerIndex) ->
...
+) ->
+    ok;
+maybe_add_sources_with_sessions_to_topic_handler(ConnResId, #{} = _ConnConfig, PartialConnState) ->
```
zmstone, line 683:
```suggestion
maybe_add_sources_with_sessions_to_topic_handler(ConnResId, #{clean_start := false} = _ConnConfig, PartialConnState) ->
```

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
thalesmg: yes: https://github.com/emqx/emqx/blob/a3f997122e718bbf568d8d738765b5fd05176eab/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl#L107

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
thalesmg: the first clause already ensures it's not true.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13742)

PR #17087, opened 2026-04-17
fix(pgsql): fix races in raw query execution

savonarola:
Release version: 6.2.0

## PR Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -236,7 +236,7 @@ clean-all:
 	@rm -rf deps
 	@rm -rf _build
 	@rm -f emqx_dialyzer_*_plt*
-	@rm -rf apps/emqx_dashboard/priv
+	@rm -rf apps/emqx_dashboard/priv/{www,*.hocon}
 	@rm -rf apps/*/gen_src/*.{erl,hrl,asn1db}
 
 .PHONY: deps-all
--- a/apps/emqx_auth_postgresql/src/emqx_auth_postgresql_connector.erl
+++ b/apps/emqx_auth_postgresql/src/emqx_auth_postgresql_connector.erl
@@ -31,6 +31,7 @@
 -export([config_fields/0]).
 
 -define(PGSQL_DEFAULT_PORT, 5432).
+-define(QUERY_TIMEOUT, 15_000).
 -define(PGSQL_HOST_OPTIONS, #{
     default_port => ?PGSQL_DEFAULT_PORT
 }).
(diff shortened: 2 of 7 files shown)


apps/emqx_auth_postgresql/src/emqx_auth_postgresql_connector.erl
```diff
@@ -276,30 +277,33 @@ do_check_prepares(
...
-                Class:Reason ->
-                    ?tp(error, "postgresql_auth_connector_query_exception", LogInfo#{
-                        class => Class, reason => Reason
-                    }),
-                    {error, {unrecoverable_error, Reason}}
-            end;
-        {error, disconnected} ->
-            ?tp(warning, "postgresql_auth_connector_query_failed", LogInfo#{reason => disconnected}),
-            {error, {unrecoverable_error, disconnected}}
+    try ecpool_worker:exec(Worker, Fun, ?QUERY_TIMEOUT) of
```
thalesmg: I think we ought to check if `Worker` is really a pid before calling `exec` here, as it could be `false | no_such_pool`, according to the typespecs.

apps/emqx_auth_postgresql/src/emqx_auth_postgresql_connector.erl
```diff
@@ -276,30 +277,33 @@ do_check_prepares(
...
+                reason => Reason
+            }),
+            {error, {unrecoverable_error, format_error(Reason)}};
+        Result ->
+            Result
+    catch
+        exit:{timeout, _} ->
+            ?tp(warning, "postgresql_auth_connector_query_failed", LogInfo#{reason => timeout}),
+            {error, {unrecoverable_error, timeout}};
+        Class:Reason ->
```
thalesmg: nit: we could treat `{noproc, _}` as a recoverable error (worker pid is restarting?)

Makefile
```diff
@@ -236,7 +236,7 @@ clean-all:
 	@rm -rf deps
 	@rm -rf _build
 	@rm -f emqx_dialyzer_*_plt*
-	@rm -rf apps/emqx_dashboard/priv
+	@rm -rf apps/emqx_dashboard/priv/{www,*.hocon}
```
thalesmg, line 239: Q: why not delete the whole `priv`?  unrelated change?

thalesmg (dismissed):

Makefile
savonarola: there is `apps/emqx_dashboard/priv/api-spec.html` commited to the repo now

apps/emqx_auth_postgresql/src/emqx_auth_postgresql_connector.erl
savonarola: ✅

apps/emqx_auth_postgresql/src/emqx_auth_postgresql_connector.erl
savonarola: ✅

thalesmg (approved):

yanzhiemq:
[TestExecution](https://emqx.atlassian.net/browse/QA-14788)

PR #17097, opened 2026-04-20
fix(retainer): restore retainer.enable to decouple retain protocol support from storage

HJianBo:
Fixes N/A

Release version: 5.9

## Summary

This PR restores `retainer.enable` as a real runtime switch for the retainer subsystem on `release-59`.

The key logic change is that retainer process lifecycle is once again controlled by `retainer.enable`, instead of being implicitly coupled to `mqtt.retain_available`.
This matters for deployments that want to keep retained-message protocol support enabled, but do not want EMQX to actually start the retainer subsystem and persist retained data.

The most relevant changes are:
- `apps/emqx_retainer/src/emqx_retainer.erl`: restore `retainer.enable`-driven init and config-update behavior; decouple runtime start/stop from `mqtt.retain_available`
- `apps/emqx_retainer/src/emqx_retainer_schema.erl`: remove both the deprecated marker and the hidden-doc importance on `retainer.enable`, so the option is visible again in schema-generated docs
- `apps/emqx_retainer/test/emqx_retainer_SUITE.erl`: update regression coverage so tests explicitly verify that `mqtt.retain_available` no longer controls retainer lifecycle, and also verify that with `mqtt.retain_available = true`, toggling `retainer.enable` changes whether retained data is actually stored
- `apps/emqx_retainer/test/emqx_retainer_api_SUITE.erl`: add API-level coverage for `PUT /api/v5/mqtt/retainer`, verifying that `enable=false` stops retained storage while `mqtt.retain_available=true` still allows retained publishes through the protocol path
- `changes/ce/fix-17097.en.md`: add user-facing changelog entry

Reasoning:
- `mqtt.retain_available = false` affects protocol-level retained-message behavior and can reject retained publishes on the MQTT path
- some users need a narrower operational control: keep protocol support enabled while disabling retainer storage and worker startup
- restoring `retainer.enable` provides that control without introducing a new configuration surface on `release-59`

## PR Checklist
- ~For internal contributor: there is a jira ticket to track this change~
- [x] Change log for changes visible by users has been added to `changes/ce/(feat|perf|fix|breaking)-<PR-id>.en.md` files

### DIFF
--- a/apps/emqx_retainer/src/emqx_retainer.app.src
+++ b/apps/emqx_retainer/src/emqx_retainer.app.src
@@ -2,7 +2,7 @@
 {application, emqx_retainer, [
     {description, "EMQX Retainer"},
     % strict semver, bump manually!
-    {vsn, "5.1.2"},
+    {vsn, "5.1.3"},
     {modules, []},
     {registered, [emqx_retainer_sup]},
     {applications, [kernel, stdlib, emqx, emqx_ctl]},
--- a/apps/emqx_retainer/src/emqx_retainer.erl
+++ b/apps/emqx_retainer/src/emqx_retainer.erl
@@ -30,8 +30,7 @@
 -export([
     on_session_subscribed/3,
     on_message_publish/1,
-    post_config_update/5,
-    on_config_zones_updated/2
+    post_config_update/5
 ]).
 
 %% Internal APIs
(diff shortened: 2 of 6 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13689)

PR #17105, opened 2026-04-22
Fix influxdb unicode write handling on release-510

HJianBo:
Fixes https://emqx.atlassian.net/browse/EMQX-15228

Release version: 5.10.4

## Summary

This PR fixes Unicode handling for the InfluxDB connector/action when writing Chinese text on `release-510`.

Root cause:
- `write_syntax` parsing converted binaries with `binary_to_list/1`, which breaks UTF-8 text into byte lists.
- the connector value conversion path treated printable lists as generic lists instead of UTF-8 strings, so Chinese tag and field values could be written incorrectly.

What changed:
- parse InfluxDB `write_syntax` binaries as Unicode character lists before template preprocessing.
- convert printable Unicode lists to UTF-8 binaries in the InfluxDB connector before encoding line protocol.
- add a regression test for fixed Chinese text in `write_syntax`.
- add integration tests covering:
  - fixed Chinese written by InfluxDB Action
  - Chinese carried in MQTT payload and written through InfluxDB Action
- simplify rule creation in the integration suite by using `emqx_rule_engine:create_rule/1` directly so the tests do not depend on a fixed dashboard port.

Impact:
- fixed Chinese literals in InfluxDB Action now persist correctly.
- Chinese MQTT payloads routed through InfluxDB Action now persist correctly.
- the regression coverage now exercises API v1/v2 over TCP/TLS and sync/async with/without batching.

## PR Checklist

## Validation
- `./rebar3 eunit --module emqx_bridge_influxdb_tests --name eunit@127.0.0.1 -v`
- Common Test matrix for both Chinese scenarios on:
  - `apiv1_tcp`
  - `apiv1_tls`
  - `apiv2_tcp`
  - `apiv2_tls`
  with sync/async and with/without batching

### DIFF
--- a/changes/ee/fix-17105.en.md
+++ b/changes/ee/fix-17105.en.md
@@ -1,1 +1,1 @@
-Fixed the InfluxDB connector/action to preserve Unicode text, including Chinese characters, when writing values from `write_syntax` literals or MQTT payloads.
+Fixed the InfluxDB connector/action to preserve Unicode text, when writing values from `write_syntax` literals or MQTT payloads.
--- a/changes/ee/fix-17105.en.md
+++ b/changes/ee/fix-17105.en.md
@@ -1,0 +1,1 @@
+Fixed the InfluxDB connector/action to preserve Unicode text, including Chinese characters, when writing values from `write_syntax` literals or MQTT payloads.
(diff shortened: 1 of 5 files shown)


changes/ee/fix-17105.en.md
```diff
@@ -0,0 +1 @@
+Fixed the InfluxDB connector/action to preserve Unicode text, including Chinese characters, when writing values from `write_syntax` literals or MQTT payloads.
```
zmstone:
```suggestion
Fixed the InfluxDB connector/action to preserve Unicode text, when writing values from `write_syntax` literals or MQTT payloads.
```

zmstone (dismissed):

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13741)

PR #17132, opened 2026-04-25
fix(modules): make topic_metrics add/remove idempotent on replicate

zmstone:
Fixes <issue-or-jira-number>

Release version:
6.0.3, 6.1.1, 6.2.0

## Summary

When `add_topic_metrics` / `remove_topic_metrics` runs through cluster RPC, the initiator's `pre_config_update/3` validates against raw config and rejects duplicate-add or missing-remove with `{error, already_existed}` / `{error, not_found}`. On a replicate, that same validation re-runs against the replicate's local raw config; if it has drifted (e.g. transient inconsistency), the same error fires there as a `cluster_rpc_apply_failed`, raising an alarm and stalling commit lag.

This change moves the duplicate / missing checks to the user-facing API (`add_topic_metrics/1`, `remove_topic_metrics/1`) so the initiator still gets the expected error, while `pre_config_update/3` and `post_config_update/5` on replicates apply the change idempotently:

- `pre_config_update` on add: existing topic → `{ok, RawConf}` (no-op) instead of `{error, already_existed}`.
- `pre_config_update` on remove: missing topic → `{ok, RawConf -- [Topic]}` (no-op) instead of `{error, not_found}`.
- `post_config_update`: tolerate `emqx_topic_metrics:register/1` returning `{error, already_existed}` and `deregister/1` returning `{error, topic_not_found}`.

New cluster CT suite `emqx_modules_conf_cluster_SUITE` covers replicate-drift tolerance and that the public API still returns `not_found` for the initiator.

## PR Checklist

### DIFF
--- a/apps/emqx_modules/src/emqx_modules_conf.erl
+++ b/apps/emqx_modules/src/emqx_modules_conf.erl
@@ -58,18 +58,28 @@ topic_metrics() ->
     {ok, emqx_types:topic()}
     | {error, term()}.
 add_topic_metrics(Topic) ->
-    case cfg_update([topic_metrics], ?FUNCTION_NAME, Topic) of
-        {ok, _} -> {ok, Topic};
-        {error, Reason} -> {error, Reason}
+    case lists:member(Topic, topic_metrics()) of
+        true ->
+            {error, already_existed};
+        false ->
+            case cfg_update([topic_metrics], ?FUNCTION_NAME, Topic) of
+                {ok, _} -> {ok, Topic};
+                {error, Reason} -> {error, Reason}
+            end
     end.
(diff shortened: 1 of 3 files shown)


savonarola:
Curious, seems we introduced `pre_config_update/4` and `post_config_update/6` with `#{kind => ?KIND_INITIATE | ?KIND_REPLICATE}` extra info in the last arg exactly for this purpose. Why do we choose to reimplement in an ad-hoc manner?

savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13940)

PR #17140, opened 2026-04-27
fix(crl): accept DER-encoded CRL bodies and warn on undecodable ones

### LINKED ISSUE #17135: CRL HTTP fetch silently accepts empty result on RFC-compliant DER bodies
### What happened?

Configured an SSL listener with `enable_crl_check = true` against a CRL
distribution point that serves an RFC 5280 §5 conforming
`application/pkix-crl` (DER) body. The CRL endpoint returns 200 with a
valid DER-encoded CertificateList, but every TLS handshake fails on the
client with

```
tlsv1 alert internal error
```

and the server-side OTP error is

```
{tls_alert, {bad_crls, "... no_relevant_crls ..."}}
```

There is **no log line** in EMQX indicating that anything went wrong
with the CRL fetch — the listener appears configured correctly, the
URL is reachable, and the response is valid DER, yet every connection
is rejected.

Re-encoding the exact same CRL as PEM at the exact same URL and
restarting EMQX makes the problem disappear. So the failure mode is
specifically: HTTP fetch succeeds, body is valid DER, EMQX silently
caches an empty CRL list, and the OTP SSL stack later reports
`no_relevant_crls`.

The root cause is in `emqx_crl_cache:parse_crls/1`, which only handles
PEM bodies; on a DER body, `public_key:pem_decode/1` returns `[]` (it
does not raise), the surrounding list comprehension yields `[]`, and
`{der, []}` is silently inserted into the cache. Details and a
suggested fix are below.

I expected an SSL listener with `enable_crl_check = true` to fetch
the CRL from the URL in the peer cert's *CRL Distribution Points*
extension, decode it, and validate client certs against it — for
either of the two CRL wire formats commonly served over HTTP:

* PEM (`-----BEGIN X509 CRL-----` … body), and
* DER (the format mandated by RFC 5280 §5 and announced via
  `Content-Type: application/pkix-crl`).

What actually happens with a DER-encoded CRL body is that
`emqx_crl_cache:parse_crls/1` decodes zero CRLs (`public_key:pem_decode/1`
returns `[]` on non-PEM input, no exception is raised, the surrounding
list comprehension yields `[]`), and the empty list is silently
inserted into the cache as `{der, []}`. Subsequent TLS handshakes then
fail with

```
{tls_alert, {bad_crls, "... no_relevant_crls ..."}}
```

with **no log line** indicating that the CRL fetch decoded zero CRLs.
The listener appears configured correctly, the CRL endpoint is
reachable and returns 200, but every handshake is rejected.

This is an interoperability bug because `application/pkix-crl` (DER)
is what RFC 5280 §5 prescribes for CRLs distributed over HTTP, and
many CAs — both commercial and self-hosted — therefore serve CRLs as
DER by default. EMQX's expectation that the body is PEM is
undocumented in the listener schema doc and produces a silent failure
on conforming servers.

### Root cause

`apps/emqx/src/emqx_crl_cache.erl`:

```erlang
parse_crls(Bin) ->
    try
        [CRL || {'CertificateList', CRL, not_encrypted}
                <- public_key:pem_decode(Bin)]
    catch _:_ -> error
    end.
```

`public_key:pem_decode/1` is documented as returning `[]` on input
that contains no PEM-encoded entries; it does *not* raise, so the
`catch` clause is never entered and the comprehension simply yields
`[]`.

The caller, `do_http_fetch_and_cache/2`, treats `[]` as success:

```erlang
do_http_fetch_and_cache(URL, HTTPTimeoutMS) ->
    Resp = ?MODULE:http_get(URL, HTTPTimeoutMS),
    case Resp of
        {ok, {{"_", 200, "_"}, "_", Body}} ->
            case parse_crls(Body) of
                error -> {error, invalid_crl};
                CRLs ->
                    emqx_ssl_crl_cache:insert(to_string(URL), {der, CRLs}),
                    {ok, CRLs}
```

`{der, []}` is inserted into the SSL CRL cache, the OTP SSL stack
later finds zero CRLs for the distribution point, and validation
fails with `revocation_status_undetermined` →
`{bad_crls, no_relevant_crls}`.

`http_get/2` (lines 246–254) doesn't inspect the response
`Content-Type`, so the server's RFC-mandated `application/pkix-crl`
header has no influence on parsing.

### Suggested fix

Treat an empty PEM-decode as a signal to attempt DER decoding. The
diff is small and preserves existing error semantics — malformed
bodies still fall through the `catch` and become `error`:

```diff
--- a/apps/emqx/src/emqx_crl_cache.erl
+++ b/apps/emqx/src/emqx_crl_cache.erl
 parse_crls(Bin) ->
     try
-        [CRL || {'CertificateList', CRL, not_encrypted}
-                <- public_key:pem_decode(Bin)]
+        case [CRL || {'CertificateList', CRL, not_encrypted}
+                     <- public_key:pem_decode(Bin)] of
+            [] ->
+                %% Body wasn't PEM. RFC 5280 §5 mandates
+                %% application/pkix-crl (DER) for CRLs over HTTP,
+                %% so try decoding the body as a single DER
+                %% CertificateList.  der_decode/2 raises on garbage,
+                %% which is caught below and reported as `error`.
+                [public_key:der_decode('CertificateList', Bin)];
+            CRLs ->
+                CRLs
+        end
     catch _:_ -> error
     end.
```

This is sufficient to make EMQX interoperate with RFC-compliant CRL
distribution points without changing the behaviour for PEM bodies.

A more thorough variant would dispatch on `Content-Type`
(`application/pkix-crl` → DER, anything else → PEM), but the patch
above already restores RFC compliance with no behavioural regressions
for existing PEM-CRL deployments.

### Optional: surface the silent-failure case in logs

While here, the empty-list path is worth logging — it currently lets
a misconfiguration sit invisible:

```diff
 do_http_fetch_and_cache(URL, HTTPTimeoutMS) ->
     Resp = ?MODULE:http_get(URL, HTTPTimeoutMS),
     case Resp of
         {ok, {{"_", 200, "_"}, "_", Body}} ->
             case parse_crls(Body) of
-                error -> {error, invalid_crl};
+                error ->
+                    ?SLOG(warning, #{msg => "crl_http_fetch_decode_failed",
+                                     url => URL,
+                                     hint => "body is neither PEM nor DER"}),
+                    {error, invalid_crl};
                 CRLs ->
                     emqx_ssl_crl_cache:insert(to_string(URL),
                                               {der, CRLs}),
                     {ok, CRLs}
```

1. Bring up any HTTP server that serves a DER-encoded CRL with
   `Content-Type: application/pkix-crl` at, say,
   `http://crl.example.test/ca.crl`. A minimal Python reproducer:

   ```python
   import http.server
   class H(http.server.BaseHTTPRequestHandler):
       def do_GET(self):
           with open('crl.der', 'rb') as f:
               body = f.read()
           self.send_response(200)
           self.send_header('Content-Type', 'application/pkix-crl')
           self.send_header('Content-Length', str(len(body)))
           self.end_headers()
           self.wfile.write(body)
   http.server.HTTPServer(('0.0.0.0', 8080), H).serve_forever()
   ```

   with `crl.der` produced by

   ```sh
   openssl ca -gencrl -out crl.der -outform DER -config openssl.cnf
   ```

2. Issue a server certificate whose CRL Distribution Points
   extension points at that URL.

3. Configure an EMQX listener:

   ```hocon
   listeners.ssl.default {
     bind = "0.0.0.0:8883"
     ssl_options {
       certfile  = "/path/server.pem"
       keyfile   = "/path/server.key"
       cacertfile = "/path/ca.pem"
       verify = verify_peer
       fail_if_no_peer_cert = true
       enable_crl_check = true
     }
   }
   ```

4. Connect with

   ```sh
   openssl s_client -cert client.pem -key client.key \
       -CAfile ca.pem -connect <emqx>:8883 -servername <name>
   ```

**Expected**: the CRL is fetched, decoded, and the client cert is
validated against it. Handshake completes (or is rejected with
`certificate revoked` if the client cert is on the CRL).

**Actual**: handshake fails with `tlsv1 alert internal error`. The
EMQX log shows nothing diagnostic about the CRL fetch. Re-encoding
the same CRL as PEM at the same URL and restarting EMQX makes the
problem disappear.

* RFC 5280 §5 (Internet X.509 PKI CRL Profile) defines the wire
  format for CRLs distributed over HTTP as `application/pkix-crl`,
  which is DER. Any RFC-compliant CRL distribution point will
  trigger this bug.
* OTP `public_key:pem_decode/1` is documented as returning `[]`
  rather than raising on non-PEM input, which is why the existing
  `try/catch` in `parse_crls/1` doesn't catch the failure.
* The fix above keeps the existing `catch` in place so that
  truly malformed bodies (neither PEM nor DER) still surface as
  `{error, invalid_crl}`. Adding the `?SLOG(warning, …)` line in
  `do_http_fetch_and_cache/2` makes that case visible in logs
  instead of silently rejecting handshakes.
* Workaround for users until fixed: configure your CA to serve
  CRLs as PEM at the URL baked into the cert's CRL DP. If the CA
  can only emit DER, run a small wrapper that fetches DER and
  re-encodes via `openssl crl -inform DER -outform PEM`.

### EMQX version

EMQX 5.10.3 (current latest at time of writing). The same code is
present on `master` and `release-5.10`.

### OS version

<details>

```console
PRETTY_NAME="Ubuntu 24.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.4 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo

Linux tio.bevemyr.com 6.8.0-107-generic #107-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar 13 19:51:50 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
```

</details>

### Log files

<details>
n/a — the bug is a *silent* failure: there is no log line indicating
that the CRL fetch decoded zero CRLs, which is part of why the
report is filed. The only observable symptom is the TLS handshake
alert `{bad_crls, "... no_relevant_crls ..."}` on the client side.

</details>

zmstone:
Fixes #17135

## Summary

`emqx_crl_cache:parse_crls/1` only decoded PEM bodies. RFC 5280 §5 mandates `application/pkix-crl` (DER) for CRLs distributed over HTTP, and many CAs serve CRLs that way by default. On a DER body, `public_key:pem_decode/1` returns `[]` (it does not raise), so the empty list was silently cached as `{der, []}` and every TLS handshake on `enable_crl_check = true` listeners then failed with `{bad_crls, no_relevant_crls}` — with no log line indicating that the CRL fetch decoded zero CRLs.

- `parse_crls/1`: when the PEM-decode comprehension yields `[]`, fall back to `public_key:der_decode('CertificateList', Bin)` inside the existing `try/catch`. On valid DER, the original body is stored as a single CRL; on garbage, the catch still surfaces the failure as `error`.
- `do_http_fetch_and_cache/2`: when `parse_crls/1` returns `error`, emit a `?SLOG(warning, ...)` with the URL and a hint that the body is neither PEM nor DER, so the misconfiguration is visible instead of silent.

CT covers the DER-body path (`t_refresh_der_response`) and updates `t_refresh_invalid_response` to assert that an undecodable body now surfaces as `crl_refresh_failure` with `error => invalid_crl` rather than a silent empty insert.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_crl_cache.erl
+++ b/apps/emqx/src/emqx_crl_cache.erl
@@ -205,6 +205,11 @@ do_http_fetch_and_cache(URL, HTTPTimeoutMS) ->
         {ok, {{_, 200, _}, _, Body}} ->
             case parse_crls(Body) of
                 error ->
+                    ?SLOG(warning, #{
+                        msg => "crl_http_fetch_decode_failed",
+                        url => URL,
+                        hint => "body is neither PEM nor DER"
+                    }),
                     {error, invalid_crl};
                 CRLs ->
                     %% Note: must ensure it's a string and not a
@@ -222,7 +227,17 @@ do_http_fetch_and_cache(URL, HTTPTimeoutMS) ->
 
 parse_crls(Bin) ->
     try
-        [CRL || {'CertificateList', CRL, not_encrypted} <- public_key:pem_decode(Bin)]
+        case [CRL || {'CertificateList', CRL, not_encrypted} <- public_key:pem_decode(Bin)] of
+            [] ->
+                %% Body wasn't PEM. RFC 5280 §5 mandates application/pkix-crl
+                %% (DER) for CRLs distributed over HTTP, so try decoding the
+                %% body as a single DER CertificateList. der_decode/2 raises
+                %% on garbage, which is caught below and reported as `error`.
+                _ = public_key:der_decode('CertificateList', Bin),
+                [Bin];
+            CRLs ->
+                CRLs
+        end
     catch
         _:_ ->
             error
--- a/apps/emqx/test/emqx_crl_cache_SUITE.erl
+++ b/apps/emqx/test/emqx_crl_cache_SUITE.erl
@@ -583,6 +583,8 @@ t_refresh_request_error(_Config) ->
     emqx_config_handler:stop(),
     ok.
 
+%% A body that is neither PEM nor DER must be reported as a decode failure
+%% so the listener does not silently cache an empty CRL list.
 t_refresh_invalid_response(_Config) ->
     meck:expect(
         emqx_crl_cache,
@@ -597,17 +599,51 @@ t_refresh_invalid_response(_Config) ->
     ?check_trace(
         ?wait_async_action(
             ?assertEqual(ok, emqx_crl_cache:refresh(URL)),
-            #{?snk_kind := crl_cache_insert},
+            #{?snk_kind := crl_refresh_failure},
             5_000
         ),
         fun(Trace) ->
             ?assertMatch(
-                [#{crls := []}],
-                ?of_kind(crl_cache_insert, Trace)
+                [#{error := invalid_crl} | _],
+                ?of_kind(crl_refresh_failure, Trace)
             ),
+            ?assertEqual([], ?of_kind(crl_cache_insert, Trace)),
             ok
         end
     ),
+    ok = snabbkaffe:stop(),
+    emqx_config_handler:stop(),
+    ok.
+
+%% RFC 5280 §5 mandates application/pkix-crl (DER) over HTTP. EMQX must
+%% decode such bodies and cache the CRL, not silently insert {der, []}.
+t_refresh_der_response(Config) ->
+    CRLDer = ?config(crl_der, Config),
+    meck:expect(
+        emqx_crl_cache,
+        http_get,
+        fun(_URL, _HTTPTimeout) ->
+            {ok, {{"HTTP/1.0", 200, 'OK'}, [], CRLDer}}
+        end
+    ),
+    emqx_config_handler:start_link(),
+    {ok, _} = emqx_crl_cache:start_link(),
+    URL = "http://localhost/crl.der",
+    URLBin = iolist_to_binary(URL),
+    Ref = get_crl_cache_table(),
+    ?check_trace(
+        ?wait_async_action(
+            ?assertEqual(ok, emqx_crl_cache:refresh(URL)),
+            #{?snk_kind := crl_cache_insert},
+            5_000
+        ),
+        fun(Trace) ->
+            Inserts = [maps:with([url, crls], E) || E <- ?of_kind(crl_cache_insert, Trace)],
+            ?assertEqual([#{url => URL, crls => [CRLDer]}], Inserts),
+            ok
+        end
+    ),
+    ?assertEqual([{URLBin, [CRLDer]}], ets:tab2list(Ref)),
     ok = snabbkaffe:stop(),
     emqx_config_handler:stop(),
     ok.
--- a/changes/ee/fix-17140.en.md
+++ b/changes/ee/fix-17140.en.md
@@ -1,0 +1,5 @@
+Fixed a silent failure when EMQX fetched a Certificate Revocation List (CRL) over HTTP from a server that returns a DER-encoded body (`Content-Type: application/pkix-crl`, the format mandated by RFC 5280 §5).
+
+Previously, EMQX only decoded PEM-encoded CRL bodies; a DER body was silently treated as zero CRLs and cached as an empty list, causing every TLS handshake on `enable_crl_check = true` listeners to fail with `bad_crls, no_relevant_crls` and no log line indicating what went wrong.
+
+EMQX now decodes both PEM and DER CRL bodies. When a fetched body is neither, a warning is logged with the URL so the misconfiguration is visible.


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13922)

PR #17145, opened 2026-04-27
Support authorization source preconditions

hjianbo:
Fixes https://emqx.atlassian.net/browse/EMQX-15237

Release version:
6.3.0

## Summary

Authentication already supports call conditions, but ACL authorization sources could not be conditionally selected. This change lets deployments route different business functions to different authorization backends using Variform expressions over client and request variables.

Key changes:

* Add a common `precondition` option to EMQX authorization sources.
* Compile source preconditions when authorization sources are created or updated, and keep compiled expressions in the shared authorization chain state.
* Evaluate preconditions before calling backend modules; sources are skipped when the expression does not evaluate to `true`.
* Keep backend source modules isolated from the new common field by removing `precondition` before calling backend `create`, `update`, and `destroy` callbacks.
* Add coverage for invalid expressions, runtime Variform errors, source skipping behavior, and HTTP API create/update validation.

The schema change is backward compatible: `precondition` defaults to an empty value, which preserves existing behavior and avoids adding precondition state on the authorization hot path.

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ee/feat-17145.en.md`

### DIFF
--- a/apps/emqx_auth/src/emqx_authz/emqx_authz.erl
+++ b/apps/emqx_auth/src/emqx_authz/emqx_authz.erl
@@ -370,7 +370,7 @@ ensure_deleted(#{enable := false}, _) ->
 ensure_deleted(SourceState, #{clear_metric := ClearMetric}) ->
     Type = type(SourceState),
     Module = authz_module(Type),
-    Module:destroy(SourceState),
+    Module:destroy(cleanup_precondition(SourceState)),
     ClearMetric andalso emqx_metrics_worker:clear_metrics(authz_metrics, Type).
 
 %% Called for both sources and raw sources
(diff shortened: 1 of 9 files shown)


rel/i18n/emqx_authz_schema.hocon
```diff
@@ -54,6 +54,12 @@ enable.desc:
 enable.label:
 """enable"""
 
+precondition.label:
+"""Precondition"""
+
+precondition.desc:
+"""A Variform expression to evaluate with client and request information before calling this authorization source. Available variables include client information as well as request bindings such as <code>action</code>, <code>topic</code>, <code>qos</code>, and <code>retain</code>. The source is skipped when the expression does not evaluate to true."""
```
zmstone: add the available client information examples.

apps/emqx_auth/src/emqx_authz/emqx_authz.erl
```diff
@@ -557,6 +610,18 @@ do_authorize(
             {Matched, Type}
     end.
 
+check_precondition(undefined, _Client, _Action, _Topic) ->
+    true;
+check_precondition(Precondition, Client0, Action, Topic) ->
+    Client = emqx_auth_template:rename_client_info_vars(Client0),
+    Vars = (emqx_authz_utils:vars_for_rule_query(Client, Action))#{topic => Topic},
```
zmstone: doc says it supports QoS retain flag etc. but here only topic is added.

zmstone: maybe only add topic and action, do not add qos and retain flag.

zmstone (commented):
for ACL, a user will very much likely need a new function in emqx_variform_bif to match topic against a filter.

e.g.

`topic_match(topic, topic_join(clientid, '#')))` for pre-condition.

apps/emqx_utils/src/emqx_variform_bif.erl
```diff
@@ -409,6 +412,64 @@ nth(N, List) when is_integer(N) andalso is_list(List) ->
...
+    end.
+
+topic_match(NULL, _) when ?IS_NULL(NULL) ->
+    ?BADARG();
+topic_match(_, NULL) when ?IS_NULL(NULL) ->
+    ?BADARG();
+topic_match(Topic0, Filter0) ->
+    Topic = any_to_str(Topic0),
+    Filter = any_to_str(Filter0),
+    match_topic_words(topic_words(Topic), topic_words(Filter)).
```
keynslug: Q: Why not just `emqx_topic:match(Topic, Filter)`?

apps/emqx_utils/src/emqx_variform_bif.erl
```diff
@@ -61,6 +61,9 @@
 %% Array functions
 -export([nth/2]).
 
+%% Topic functions
+-export([topic_join/1, topic_join/2, topic_match/2]).
```
zmstone: add a topic_split/1 function.
and add roundtrip tests (using variform, not internal function calls).

e.g.
assert expression `str_eq(topic_join(topic_split('topic/1'), 'topic/1')` results in `true`

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16052)

PR #17152, opened 2026-04-28
feat: support config distribution port connect options for perf tunings

qzhuyan:
Fixes [EMQX-15239](https://emqx.atlassian.net/browse/EMQX-15239)

Support config erlang `port` port options for distribution port and have default buffer size to 1 MB. 

Without this feature, erlang dist port has performance issues where the erlang port buffer is too small (default 1460 Bytes, some platform 9KB) even dist port buffer  (vmarg +zdbbl )is set to much larger (e.g 32MB). 

This issue was found in emqx 4.4, now ports the solution to 6.0. 

note: user could check dist port port options in use with 
```
rp([recon:port_info(element(6,X)) || X<-ets:tab2list(sys_dist)]).
```
If the buffer is too small, enlarge it with 
```
inet:setopts(Socket, [{buffer, 1024*1024}])
```

NOTE1: It may solve lots of issues of 'erpc timeout', 'mnesia congestions' , it may also help to support more core nodes. 
NOTE2: buffer location map:
```
=================
dist port buffer (+zdbbl)
-----------------------
erl port buffer (inet opt: buffer). <--- this PR targeting. 
------------------------
Linux socket buffer (inet opt: sndbuf, rcvbuf)
=================
```
Release version:

## PR Checklist

[EMQX-15239]: https://emqx.atlassian.net/browse/EMQX-15239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_schema.erl
+++ b/apps/emqx_conf/src/emqx_conf_schema.erl
@@ -42,6 +42,8 @@
 
 -define(DEFAULT_NODE_NAME, <<"emqx@127.0.0.1">>).
 
+-define(DEFAULT_KERNEL_INET_DIST_CONNECT_OPTIONS, <<"[{buffer,1048576}]">>).
+
 %% Static apps which merge their configs into the merged emqx.conf
 %% The list can not be made a dynamic read at run-time as it is used
 %% by nodetool to generate app.<time>.config before EMQX is started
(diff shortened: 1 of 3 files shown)


apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -508,6 +510,16 @@ fields("node") ->
                     'readOnly' => true
                 }
             )},
+        {"dist_connect_options",
+            sc(
+                string(),
```
zmstone: binary()

apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -42,6 +42,8 @@
 
 -define(DEFAULT_NODE_NAME, <<"emqx@127.0.0.1">>).
 
+-define(DEFAULT_KERNEL_INET_DIST_CONNECT_OPTIONS, "[{buffer,1024*1204}]").
```
zmstone: does vm.args support 1024*1024?

should it be 1048576 instead of?

qzhuyan: right, won't work after I changed it to string ... thanks.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13944)

PR #17165, opened 2026-04-29
feat(resource): add buffer worker dispatch strategy

hjianbo:
Fixes https://emqx.atlassian.net/browse/EMQX-15257

Release version:
6.0.3, 6.1.2, 6.2.1

## Summary

This PR adds a configurable resource buffer worker dispatch strategy for actions.

Previously, resource buffer workers used the caller process as the default pick key when a query did not provide an explicit `pick_key`. In workloads where a small number of clients publish a large amount of messages, this can concentrate traffic on a single buffer worker even when `worker_pool_size` is greater than one.

The new `resource_opts.dispatch_strategy` option supports:

* `per_clientid` - the default behavior, preserving the existing caller-process based dispatch when no explicit `pick_key` is provided.
* `random` - use a random pick key for each query when no explicit `pick_key` is provided, allowing such traffic to spread across buffer workers.

The schema change is exposed through the common resource schema and the bridge v2 action resource options, so it is available from configuration and HTTP API action payloads. Runtime query option collection now carries this setting into `emqx_resource_buffer_worker`, while explicit `pick_key` values continue to take precedence.

Tests cover schema acceptance/rejection and runtime dispatch across multiple buffer workers.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/schema/emqx_bridge_v2_schema.erl
+++ b/apps/emqx_bridge/src/schema/emqx_bridge_v2_schema.erl
@@ -467,6 +467,7 @@ common_action_resource_opts_subfields() ->
         batch_size,
         batch_time,
         buffer_mode,
+        buffer_worker_dispatch_strategy,
         buffer_seg_bytes,
         health_check_interval,
         health_check_interval_jitter,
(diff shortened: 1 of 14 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14103)

PR #17173, opened 2026-04-29
fix(mgmt): block api key from leaking or restoring dashboard accounts via data backup

zmstone:
Fixes <issue-or-jira-number>

Release version: 6.0.4

## Summary

Port #17169 to release-60.

API keys with the `system` scope could previously call `POST /data/export` and `POST /data/import` to read or write the `dashboard_users` and `api_keys` mnesia table sets. That bypassed the existing `?SCOPE_DENIED` guard on `/users` and `/api_key`: a system-scoped API key could export the cluster, edit the tarball, and re-import it to install crafted dashboard accounts or new super-admin API keys.

This PR adds two narrow guards keyed on the caller's `auth_type`:

- `POST /data/export`: when called via API key, the `dashboard_users` and `api_keys` table sets are silently filtered out of the resulting archive (also when the caller explicitly listed them in `table_sets`).
- `POST /data/import`: when called via API key, the uploaded tarball is inspected locally; if it contains either sensitive table set, the request is rejected with `403 FORBIDDEN`. API-key callers must therefore upload and import on the same node.

Dashboard bearer-token (JWT login) callers are **unaffected** and continue to be able to back up and restore the full database, including dashboard users and API keys. The new test `t_import_dashboard_token_allows_sensitive_tables` pins this regression contract.

**Most relevant files:**
- `apps/emqx_management/src/emqx_mgmt_api_data_backup.erl` — branches on the `auth_meta` map populated by `emqx_dashboard:authorize/2`.
- `apps/emqx_management/src/emqx_mgmt_data_backup.erl` — new `sensitive_table_set_names/0` and `peek_sensitive_table_sets/1` helpers.

**Tests** (`apps/emqx_management/test/emqx_mgmt_api_data_backup_SUITE.erl`):
- `t_export_api_key_omits_sensitive_tables` — API-key export's archive must not contain `mnesia/emqx_admin` or `mnesia/emqx_app`.
- `t_import_api_key_blocks_sensitive_tables` — API-key import of a tarball containing those tables returns `403 FORBIDDEN`.
- `t_import_dashboard_token_allows_sensitive_tables` — same tarball imports successfully via dashboard bearer token.

Existing import tests use fixtures that contain `emqx_admin`/`emqx_app`; they have been switched to dashboard bearer-token auth so they continue to verify the full-import path.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_data_backup.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_data_backup.erl
@@ -228,7 +228,8 @@ field_filename(IsRequired, Meta) ->
 
 data_export(post, #{body := Params0} = Req) ->
     Namespace = emqx_dashboard:get_namespace(Req),
-    Params = emqx_utils_maps:put_if(Params0, <<"namespace">>, Namespace, is_binary(Namespace)),
+    Params1 = emqx_utils_maps:put_if(Params0, <<"namespace">>, Namespace, is_binary(Namespace)),
+    Params = maybe_filter_export_params(Params1, auth_meta(Req)),
     maybe
         ok ?= emqx_mgmt_data_backup:validate_export_root_keys(Params),
         {ok, Opts} ?= emqx_mgmt_data_backup:parse_export_request(Params),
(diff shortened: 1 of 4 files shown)


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13963)

PR #17174, opened 2026-04-30
fix: clarify namespace change guard

zmopish:
## Summary
- replace one-armed maybe guards with explicit namespace-change checks
- document why namespace moves are rejected for dashboard users and API keys

## Testing
- ./rebar3 as test compile
- git diff --check

Internal-only readability/comment refactor; no user-facing behavior change.


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13970)

Issue #17159, opened 2026-04-29
[Bug] CONNECT with invalid UTF-8 or null-char ClientId accepted with rc=0x00 (MQTT-1.5.3-1/2)


LoboQ1ng:
### What happened?

EMQX accepts CONNECT packets whose Client Identifier contains either:

1. **Invalid UTF-8 byte sequences** (e.g., lone continuation byte `\xff\xfe`)
2. **The Unicode null character U+0000** (encoded as `\x00` inside a length-prefixed string)

In both cases EMQX returns `CONNACK[return_code=0x00]` (Connection Accepted).

MQTT 3.1.1 §1.5.3 mandates:

> **[MQTT-1.5.3-1]** A UTF-8 encoded string MUST NOT include an encoding of the
> null character U+0000. If a receiver (Server or Client) receives a Control
> Packet containing U+0000 it MUST close the Network Connection.
>
> **[MQTT-1.5.3-2]** A UTF-8 encoded string MUST NOT include encodings that are
> not well-formed Unicode. If a receiver receives a malformed UTF-8 sequence it
> MUST close the Network Connection.

### What did you expect to happen?

It was expected to close the connection, but it was accepted instead.
Three other brokers close the connection on both inputs:

| Broker    | Invalid UTF-8 ClientId | Null-char ClientId |
| --------- | ---------------------- | ------------------ |
| **EMQX**  | ❌ CONNACK rc=0x00      | ❌ CONNACK rc=0x00  |
| Mosquitto | ✅ closes               | ✅ closes           |
| HiveMQ    | ✅ closes               | ✅ closes           |
| NanoMQ    | ✅ closes               | ✅ closes           |

### How can we reproduce it (as minimally and precisely as possible)?

No external dependencies — raw TCP only.

```python
import socket, struct

def make_connect(client_id_bytes: bytes, host="127.0.0.1", port=1884) -> bytes:
    """Build a raw MQTT 3.1.1 CONNECT packet with an arbitrary ClientId payload."""
    proto_name  = b'\x00\x04MQTT'
    proto_level = b'\x04'
    conn_flags  = b'\x02'           # CleanSession=1
    keep_alive  = b'\x00\x3c'
    cid_field   = struct.pack('>H', len(client_id_bytes)) + client_id_bytes
    body        = proto_name + proto_level + conn_flags + keep_alive + cid_field
    return bytes([0x10, len(body)]) + body

def test(label, client_id_bytes, host="127.0.0.1", port=1884):
[...]

PR #17185, opened 2026-04-30
feat(mqtt): enforce strict parser mode by default

zmstone:
Fixes https://github.com/emqx/emqx/issues/17159

Release version: 6.3.0

## Summary

Flips the default of `mqtt.strict_mode` from `false` to `true` so EMQX rejects malformed MQTT packets out of the box (invalid UTF-8 in client ID / topic, reserved-bit violations, bad packet IDs, etc.) instead of silently accepting them.

- `apps/emqx/src/emqx_schema.erl`: `mqtt.strict_mode` default → `true`.
- `apps/emqx/src/emqx_frame.erl`: parser-internal `?DEFAULT_OPTIONS` `strict_mode` → `true`, aligning the macro with the schema for any `initial_parse_state/0` caller that does not pass an explicit option.
- `rel/i18n/emqx_schema.hocon`: description updated to reflect the new default and the migration knob (`mqtt.strict_mode = false`).

This is a breaking behavior change: clients that today send malformed packets will be disconnected on the new default. Operators who need the older behavior for non-conformant clients can set `mqtt.strict_mode = false` globally or per-zone (per-listener through the listener's zone).

A handful of tests were updated to be explicit about strict_mode where they had been implicitly relying on the lenient default. Two new CTs cover the new default: `t_default_parse_state_is_strict` (frame-level) and `t_default_strict_mode_is_true` (schema/zone propagation).

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_frame.erl
+++ b/apps/emqx/src/emqx_frame.erl
@@ -65,7 +65,7 @@
 -type serialize_opts() :: options().
 
 -define(DEFAULT_OPTIONS, #{
-    strict_mode => false,
+    strict_mode => true,
     max_size => ?MAX_PACKET_SIZE,
     version => ?MQTT_PROTO_V4
 }).
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -4052,7 +4052,7 @@ mqtt_general() ->
             sc(
                 boolean(),
                 #{
-                    default => false,
+                    default => true,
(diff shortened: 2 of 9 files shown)


savonarola (approved):

zmstone:
needs this fix for CI: https://github.com/emqx/paho.mqtt.testing/pull/16

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16053)

PR #17200, opened 2026-05-04
feat(plugins): TTL and sha256-pinning for install allowlist + zip-slip fix

zmstone:
Release version: 5.8.10, 5.10.4, 6.0.3, 6.1.2, 6.2.1

Clarification on CVE-2026-44725:
The zip-slip can only happen if dashboard admin account is compromised + a stale allow list entry is left behind.
See more details in security advisory: https://github.com/emqx/emqx/security/advisories/GHSA-cp9x-5qwc-fj6r

## Summary

Two changes to the plugin install path:

- **Allowlist TTL + optional sha256 pinning** (`feat-17200`): `emqx ctl plugins allow <name-vsn>` entries now expire 5 minutes after issue. An optional `sha256:<HEX>` argument pins the allowlist entry to a specific package digest; uploads whose contents do not match are rejected with `403 Forbidden`. Existing behavior (no pinning) is preserved when the `sha256:` argument is omitted.
- **Zip-slip fix on install** (`fix-17200`): a tarball whose entry name contained `..` segments could write files outside the plugin install directory. The install path now refuses to extract any entry that would escape the install directory.

Most of the change is in `apps/emqx_plugins/src/emqx_plugins.erl` (TTL store, sha256 verification, tar entry safety check) and `apps/emqx_plugins/src/emqx_plugins_cli.erl` / `apps/emqx_management/src/emqx_mgmt_cli.erl` for the new CLI argument. New BPAPI proto module `emqx_plugins_proto_v4` carries the allowlist update across the cluster.

## PR Checklist
- [ ] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to \`changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md\` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17201, opened 2026-05-05
feat(plugins): TTL and sha256-pinning for install allowlist + zip-slip fix

zmstone:
Fixes <issue-or-jira-number>

Port of #17200 (release-58) to release-60, plus the auto-revoke follow-up that's parallel to #17202 on release-58.

## Summary

Three changes to the plugin install path:

- **Allowlist TTL + optional sha256 pinning** (`feat-17201`): `emqx ctl plugins allow <name-vsn>` entries now expire 5 minutes after issue. An optional `sha256:<HEX>` argument pins the allowlist entry to a specific package digest; uploads whose contents do not match are rejected with `403 Forbidden`. Existing behavior (no pinning) is preserved when the `sha256:` argument is omitted.
- **Auto-revoke on successful install** (also `feat-17201`): a successful HTTP install immediately calls `disallow_installation` cluster-wide, so the same grant cannot be reused for a subsequent (potentially different) tarball. The TTL still applies as a backstop; this closes the common-path window down to "single successful upload" instead of 5 minutes.
- **Zip-slip fix on install** (`fix-17201`): a tarball whose entry name contained `..` segments could write files outside the plugin install directory. The install path now refuses to extract any entry that would escape the install directory.

The release-60 plugin code was refactored relative to release-58, so this is a manual port rather than a cherry-pick. The TTL/sha256 logic in `emqx_plugins.erl` and the CLI/API plumbing are line-for-line equivalent; the zip-slip safety check was applied to `emqx_plugins_fs.erl` (which is where `write_tar_file_content` / `delete_tar_file_content` live in v6).

New BPAPI proto module `emqx_plugins_proto_v4` carries the sha256 argument across the cluster; the existing `/1` (no-hash) path keeps using v3 so rolling upgrades remain safe.

## PR Checklist

### DIFF
--- a/apps/emqx/priv/bpapi.versions
+++ b/apps/emqx/priv/bpapi.versions
@@ -78,6 +78,7 @@
 {emqx_plugins,1}.
 {emqx_plugins,2}.
 {emqx_plugins,3}.
+{emqx_plugins,4}.
 {emqx_prometheus,1}.
 {emqx_prometheus,2}.
 {emqx_resource,1}.
--- a/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_plugins.erl
@@ -558,7 +558,9 @@ upload_install(post, #{}) ->
 install_package_on_nodes(NameVsn, Bin) ->
     case emqx_plugins:is_allowed_installation(NameVsn, Bin) of
         ok ->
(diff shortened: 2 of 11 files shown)


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14107)

PR #17202, opened 2026-05-05
fix(plugins): revoke install allowlist entry after successful HTTP install

zmstone:
Fixes <issue-or-jira-number>

Release version: 5.8.11, 5.10.4

Follow-up to #17200.

## Summary

A successful plugin install via \`POST /api/v5/plugins/install\` (and the dashboard upload that wraps it) now calls \`emqx_plugins_proto_v3:disallow_installation\` cluster-wide as part of the same request. Without this, the allow entry that authorized the install lived on for up to 5 minutes (the TTL added in #17200) — a window in which a compromised admin/API credential could re-use the same grant against a *different* tarball.

The 5-minute TTL still applies as a backstop; this change closes the common-path window down to "single successful upload" instead of "5 minutes." On install failure the entry is left in place so the operator can retry without re-issuing \`emqx ctl plugins allow\`.

Touches one file: \`apps/emqx_management/src/emqx_mgmt_api_plugins.erl\` (\`do_install_package\` and a small helper).

## PR Checklist


apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -453,7 +453,9 @@ upload_install(post, #{}) ->
 do_install_package(NameVsn, FileName, Bin) ->
     case emqx_plugins:is_allowed_installation(NameVsn, Bin) of
         ok ->
-            do_install_package(FileName, Bin);
+            Result = do_install_package(FileName, Bin),
+            ok = forget_allow_after_install(NameVsn, Result),
```
savonarola, line 457: maybe run before do_install_package?

savonarola (approved):

apps/emqx_management/src/emqx_mgmt_api_plugins.erl
zmstone: this is intentional to allow retry.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-13997)

PR #17217, opened 2026-05-07
fix(api_key): make bootstrap-file scope handling lenient

JimMoen:
The bootstrap-file loader treated unknown scope names as a fatal error
that aborted the entire file load, and the regex for the optional 4th
`:scopes` segment used `.+` which prevented matching a trailing colon
with no scope content. Both behaviors made the file fragile to typos
and to the documented `key:secret:role:` form for an intentional empty
scope list.

Changes:

* Regex: switch the 4th capture group from `(?::(.+))?` to
  `(?::(.*))?` so a trailing colon matches with an empty Group 4
  instead of forcing PCRE to backtrack and steal the role into the
  scope segment (which then surfaced as a confusing
  `{"invalid_role", <<>>}` error).

* Lenient validation: `parse_bootstrap_scopes/4` now uses a new
  `emqx_mgmt_api_key_scopes:filter_valid_scopes/1` helper that
  partitions the parsed scopes into kept and rejected sets. Unknown
  names are dropped with a `bootstrap_file_unknown_scopes_dropped`
  warning log; the entry is still created with whatever valid scopes
  remain. The HTTP create/update path keeps using the strict
  `validate_scopes/1` so client typos still fail loudly there.

* Empty scope list: `parse_scopes_str(<<>>) → []` (was
  `undefined`). A line ending in `:` now stores
  `extra#{scopes => []}` — an explicit "deny all mapped paths"
  marker that the operator can later edit through the dashboard, in
  contrast to the 3-segment line which still stores no `scopes` key
  at all (back-compat all-allow).

* Runtime: `check_path_in_scopes/2` no longer rejects every path
  outright when `Scopes = []`. Mapped paths remain denied, but
  unmapped public endpoints stay reachable so a key bootstrapped with
  no usable scopes can still observe the node and be reconfigured.

Tests cover happy path (single/multi/3-seg/4-seg-empty), partial-typo
filtering, full-typo deny-all, runtime allow-unmapped/deny-mapped, and
a path-to-scope cache guard so the runtime test cannot silently turn
green if cache population regresses. The i18n description is updated
to document the 4-segment format and the lenient policy.Fixes <issue-or-jira-number>

Release version:
5.8.10, 5.10.4

will be port to =>
6.0.3, 6.1.2, 6.2.0

## PR Checklist

- ~[ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files~ fix the in developing feature in this release.

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl
@@ -30,6 +30,7 @@ affecting user-facing API key configurations.
     init_cache/0,
     clear_cache/0,
     validate_scopes/1,
+    filter_valid_scopes/1,
     is_denied_scope/1
 ]).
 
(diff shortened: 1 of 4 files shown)


zmstone (approved):

zmstone:
rebase for green ci.

zmstone (approved):

hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14021)

Issue #12501, opened 2024-02-10
Direct support for lets encrypt


burner-:
### What would you like to be added or enhanced?

EMQX dashboard should have direct support for lets encrypt as dashboard https conector and at mqtt conntectors. 

### Why is this needed?

Currenty it is possible cofigure admin panel behind ssl but it need quite lot of manual configuration. First touch of prodcut gives me feel that security as default is not in culture of this product development. It is more like addon. Also that default admin password give vibes that there is lot of work at developin proper security culture for this product developers. 

These days many modern softwares what include web managmend have also just one click checkbox what allow enable lets encrypt for managment console and automatically keep care of getting certs and so on. That encourage to use encryption when ever it is possible (when service is in public network). 
Encryption and other basic security shoud be default. Not addon what needs hours of manual configuration and set up scripts.

zmstone:
Thank you @burner- 
We might consider adding an ACME client in EMQX to make things easier.

Out of curiosity many choose to terminate TLS at LB though, that's not the case for you ?

burner-:
I can do it with BL but its not a point. Point is to try engourage devs to do product what engourage users to not shoot to their own leg with security. 
But honestly deeper I dive more worried I come security culture of EMQX. I am sorry to say that looks that lack of fast way to get TLS is just tip of iceberg at bad choises. If you want I can open different ticket from each one but at this point "basic things what should be clear without saying that not to do with this way" what I hit in first hour of use:
a) 0.0.0.0 binded admin console port by default. Admin console should bind only localhost by default because it have default password set. 
b) no ssl by default at admin console. Even self signed cert would be better than http as default. 
c) all listeners are enabled and bind to 0.0.0.0 by default without any authentication.
d) authentication allow policy by default. All listeners are open if no authentication is set. Good practise is that by default those are closed and admin can set anonymous authentication method what opens them for all if really wanted. 
e) default password at admin console. Yes this is not so bad if admin console is binded to localhost. Anyway default password usually just brings false sense of security. That why usually it is keeped as bad design and recommended way is just remove login screen when no password is set. This is more like tweak compared to a-d

At that point I just shutted down that server and continue my holiday. I did not want to do full security review for that in my holiday what it clearly needs before it is safe to connect it back to internet. I think I return to this later but clearly there is place that your dev team do internal review for best practices. I dont want to be mean or anything but it is just big surprise to find so many "student solution" in same product what is so popular and targeted to be enterprise.
[...]

PR #17219, opened 2026-05-07
feat(acme): ACME client plugin for managed cert bundles

zmstone:
Fixes: https://github.com/emqx/emqx/issues/12501
Release version:
6.1.2

## Summary

Adds an ACME client plugin (`emqx_acme`) under `plugins/emqx_acme/` that issues and renews TLS certificates from any RFC 8555 ACME CA (e.g. Let's Encrypt) into an EMQX managed certificate bundle. The plugin can also rewrite SSL/WSS listener configs cluster-wide so listeners flip from the default localhost cert to the ACME-issued cert without manual config edits.

image

### Issuance flow

1. Lowest-name node in `emqx:running_nodes/0` wins the deterministic election for the periodic timer; manual `emqx_acme_issuer:issue/0` runs on the receiving node.
2. Capture pre-state: `bundle_in_use` if both `chain` and `key` exist in `emqx_managed_certs:list_managed_files/2`, otherwise `bundle_empty`.
3. Run ACME (account, order, HTTP-01 challenge served by an internal Cowboy listener, finalize, download).
4. `emqx_managed_certs:add_managed_files/3` writes `acc_key`/`key`/`chain` to the bundle (cluster-wide via BPAPI).
5. **Scenario A** (`bundle_in_use`): no listener config change. Erlang ssl's PEM cache invalidates on `mtime` change every ~2 minutes (`ssl-11.5.3/src/ssl_pem_cache.erl:62`), so the next handshake serves the new cert without restarting the listener.
6. **Scenario B** (`bundle_empty`): for each existing listener in `listener_ids`, `emqx_conf:update([listeners, T, N], #{enable => true, ssl_options => #{managed_certs => #{bundle_name => CertBundleName}}}, #{override_to => cluster})`. The cluster config update fires `emqx_listeners:post_config_update/5` on every node which restarts the listener with the new managed-cert opts. Listener entries that don't currently exist (e.g. user wants the plugin to issue first and create the listener later) are silently skipped.

Boot precondition: the listeners named in `listener_ids` must be bootable on their own at first boot — either `enable: false` or with the static `etc/certs/cert.pem` shipped with the installation. The plugin never seeds a dummy cert; that's the operator's responsibility, and is essentially free because every install ships the localhost cert.

### Key files

- `plugins/emqx_acme/src/emqx_acme_config.erl` — config schema + parser (`parse/1` exported).
- `plugins/emqx_acme/src/emqx_acme_issuer.erl` — gen_server orchestrating ACME flow + Scenario A/B branch.
- `plugins/emqx_acme/src/emqx_acme_challenge.erl` — Cowboy handler for `/.well-known/acme-challenge/:token`.
- `plugins/emqx_acme/src/emqx_acme_api.erl` — plugin REST handlers (`GET /status`, `POST /issue`, `POST /renew`) under `/api/v5/plugin_api/emqx_acme-<vsn>/`.

## PR Checklist


plugins/emqx_acme/src/emqx_acme_issuer.erl
```diff
@@ -0,0 +1,686 @@
...
+%%--------------------------------------------------------------------
+
+%% Spawn a monitored worker to run the actual ACME flow. The worker
+%% reports its result back as a `worker_done` cast; if it crashes the
+%% gen_server picks that up via the matching DOWN message.
+spawn_worker(Action) ->
+    Self = self(),
+    {Pid, Ref} = spawn_monitor(fun() ->
+        Result = run_action(Action),
+        gen_server:cast(Self, {worker_done, self(), Result})
```
savonarola: nit: why not use `'DOWN'` for the result also?

plugins/emqx_acme/src/emqx_acme_challenge.erl
```diff
@@ -0,0 +1,112 @@
...
+            Req = cowboy_req:reply(404, #{}, <<"not found">>, Req0),
+            {ok, Req, State}
+    end.
+
+%%--------------------------------------------------------------------
+%% Internal
+%%--------------------------------------------------------------------
+
+ensure_tab() ->
+    case ets:whereis(?CHALLENGE_TAB) of
```
savonarola: nit: `emqx_utils_ets:new()`

plugins/emqx_acme/src/emqx_acme_challenge.erl
```diff
@@ -0,0 +1,112 @@
...
+-define(LISTENER_NAME, emqx_acme_challenge).
+-define(CHALLENGE_TAB, emqx_acme_challenge_tab).
+
+%%--------------------------------------------------------------------
+%% API
+%%--------------------------------------------------------------------
+
+-spec start(pos_integer()) -> ok | {error, term()}.
+start(Port) ->
+    ensure_tab(),
```
savonarola: nit: creating tab in the context of a caller looks a bit fragile

savonarola (dismissed):

savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14169)

PR #17224, opened 2026-05-08
fix sts bug  in the feature

baidu-qian:
修复在开启外部ssl证书的开关时，/tmp/ssl没有被sts中的emqx服务识别的问题。


zmstone:
one should be able to set config in values:

```
  emqxConfig:
    EMQX_LISTENERS__SSL__DEFAULT__SSL_OPTIONS__CERTFILE: /tmp/ssl/tls.crt
    EMQX_LISTENERS__SSL__DEFAULT__SSL_OPTIONS__KEYFILE:  /tmp/ssl/tls.key
```

this is more flexible so you do not need to change the chart template.
also one is free to change listener name, or use the certs for `dashboard` or `wss` listener

PR #17227, opened 2026-05-08
fix(config): name the file when cluster.hocon save fails

zmstone:
Fixes <issue-or-jira-number>

Release version:
6.0.3, 6.1.1, 6.2.0

## Summary

When `cluster.hocon` (or its directory) is read-only, immutable, or otherwise
unwritable — read-only filesystem, file owned by root, single-file Docker
bind-mount, `chattr +i`, etc. — saving a config change via the dashboard or
REST API previously returned an opaque HTTP 400 with body
`{config_update_crashed,{badmatch,{error,ebusy}}}` and only logged a badmatch
crash that did not name the file. Operators had to trace through the badmatch
stacktrace to figure out which file was unwritable.

This PR makes the failure self-explanatory:

- `apps/emqx/src/emqx_config_backup_manager.erl`: replace the hard
  `ok = file:rename(...)` match with a case clause; on failure log the file
  path + posix reason via a structured `failed_to_save_conf_file` error and
  return `{error, #{filename, reason}}`. The hint enumerates the common
  operator-side causes (read-only directory, single-file Docker bind-mount,
  immutable attribute). The temporary file is also cleaned up on failure.
- `apps/emqx/src/emqx_config.erl`: `save_configs/5` now throws a structured
  map on file save failure; `safe_handle_update_request` catches the throw
  and returns `{error, Reason}` so the API renders a clean 400 body that
  names both the file and the reason.
- The previously-silent failure of the temporary-file write (e.g. read-only
  directory) is also now surfaced — it used to return HTTP 200 even though
  the change was not persisted to disk.
- New regression test `t_cluster_hocon_save_failure` in
  `apps/emqx/test/emqx_config_SUITE.erl`.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_config.erl
+++ b/apps/emqx/src/emqx_config.erl
@@ -1003,10 +1003,10 @@ save_configs(AppEnvs, Conf, RawConf, OverrideConf, Opts) ->
         ok ->
             save_to_app_env(AppEnvs),
             save_to_config_map(Conf, RawConf);
-        {error, #{filename := Filename, reason := Reason}} ->
+        {error, #{filename := FileName, reason := Reason}} ->
             throw(#{
                 msg => failed_to_save_conf_file,
-                filename => Filename,
+                filename => FileName,
                 reason => Reason
             })
     end.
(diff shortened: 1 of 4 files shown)


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14116)

PR #17235, opened 2026-05-08
feat(api_key_scopes): add 4 login-only scope macros and admin-only subset list

JimMoen:
Introduce four new scope name macros for dashboard login users only:

  * user_management
  * mfa_management
  * sso_management
  * api_key_management

API keys MUST NOT hold any of these — schema validation in API key
create/update handlers and the bootstrap-file loader reject them.

Among these four, three are admin-only (user_management, sso_management,
api_key_management) — only login users with role=administrator may hold
them. mfa_management is intentionally available to any login user role
including viewer; for non-admin holders it acts as a self-exemption key
allowing the user to bypass force_mfa and admin_required locks on their
own MFA only (non-admins with mfa_management still cannot manage other
users MFA).

The macros and the LOGIN_ONLY_SCOPES / ADMIN_ONLY_SCOPES helper lists
land here as the foundation. Subsequent commits wire them into the
scope catalogue, scopes/0 callbacks, schema validation, and the new
login-side scope check in emqx_dashboard_rbac.Fixes <issue-or-jira-number>

Release version:
5.10.4

will port to => 6.0.3, 6.1.2, 6.2.0

## PR Checklist


apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -195,7 +230,7 @@ fields(List) ->
     [field(Key) || Key <- List, field_filter(Key)].
 
 user_fields() ->
-    fields([username, role, description, backend]) ++ ee_user_fields().
+    fields([username, role, description, backend, scopes]) ++ ee_user_fields().
```
hjianbo, line 236: This adds `scopes` to the user schema/response for every edition, but the enforcement path is still EE-only: the non-EE branch in `emqx_dashboard_token` skips `check_login_user_scopes/2` and just refreshes the JWT. In CE a caller can set/read `scopes` and get a false sense of restricted access while requests are authorized only by RBAC. If dashboard user scopes are EE-only, please guard the API/schema/storage paths with the same edition condition; otherwise run the login-user scope check outside the EE branch.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -249,6 +284,13 @@ field(new_pwd) ->
 field(role) ->
     {role,
         mk(binary(), #{desc => ?DESC(role), default => ?ROLE_DEFAULT, example => ?ROLE_DEFAULT})};
+field(scopes) ->
+    {scopes,
+        mk(hoconsc:array(binary()), #{
```
hjianbo, line 292: `to_external_user/1` now returns `scopes => null` when the key is absent, but this schema declares `scopes` as `array(binary())` only and is used by the `/users` responses. Generated clients will expect an array while the response can legally be JSON null. Please split request/response schema or make the response field nullable/union so the OpenAPI contract matches the body.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -443,38 +514,279 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
 change_mfa(delete, #{bindings := #{username := Username0}} = Req) ->
     Username = username(Req, Username0),
     LogMeta = #{msg => "dashboard_user_mfa_disable", username => Username},
-    case emqx_dashboard_admin:disable_mfa(Username) of
+    case authorize_mfa_change(Req, Username, disable) of
```
hjianbo, line 563: The new handler-level matrix is not reached for one important SSO self-disable path. `emqx_dashboard_rbac:check_rbac/5` still rejects viewer `DELETE /users/:self/mfa` when the current SSO backend has `force_mfa=true` before this code executes. That ignores the new `force_mfa_snapshot` semantics and also blocks users with `mfa_management` from using the self-exemption documented in the changelog. Please let self `DELETE /users/:username/mfa` pass role RBAC and enforce the lock policy here, or make the RBAC check use the same snapshot/scope-aware predicate.

apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
```diff
@@ -20,13 +20,27 @@
...
+        <<"/api_key">> => ?SCOPE_API_KEY_MGMT,
+        <<"/api_key/:name">> => ?SCOPE_API_KEY_MGMT
+    }.
 
 api_spec() ->
     emqx_dashboard_swagger:spec(?MODULE, #{check_schema => true, translate_body => true}).
 
 paths() ->
-    ["/api_key", "/api_key/:name", "/api_key/scopes"].
+    ["/api_key", "/api_key/:name", "/api_key_scopes"].
```
hjianbo, line 43: This removes the existing `GET /api_key/scopes` route and only registers `GET /api_key_scopes`. Existing clients calling `/api_key/scopes` will now match `/api_key/:name` and receive a lookup for an API key named `scopes` instead of the catalogue. Unless this is intentionally a breaking API change for this release, please keep `/api_key/scopes` as an alias while adding the new top-level route.

hjianbo (commented):
Reviewed the latest update and left a few inline comments on the remaining behavioral/API-contract issues.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -443,38 +514,279 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
...
+                msg => "set_user_scopes_failed",
+                username => Username,
+                reason => Reason
+            }),
+            {error, Reason}
+    end.
+
+%% --- MFA self-lock authorization ---
+%%
+%% SPEC-dashboard-user-scopes.md sec 6.3 / 6.5. Decision matrix:
```
zmstone: ?

apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
JimMoen: "/api_key_scopes" was new added in this developing branch. It havn't been released.
no breaking change.

apps/emqx/include/emqx_api_key_scopes.hrl
```diff
@@ -29,3 +29,63 @@
 %% Endpoints that API Keys must never access (dashboard login/SSO/
 %% API Key self-management).  Not exposed to users.
 -define(SCOPE_DENIED, <<"$denied">>).
+
+%% ── Login-user-only scopes (since 5.x) ─────────────────────────────
```
zmstone:
```suggestion
%% ── Login-user-only scopes (since 5.10.4) ─────────────────────────────
```

apps/emqx/src/emqx_scope_catalogue.erl
```diff
@@ -0,0 +1,163 @@
...
+  * `api_key_management` (administrator-only)
+
+API keys MUST NOT hold any of the four login-only scopes — schema
+validation in emqx_mgmt_api_api_keys rejects API key creation /
+update if the scope list contains any login-only scope; the
+bootstrap-file loader filters such names with a warning.
+""".
+-spec login_user_scope_catalogue() ->
+    [#{name := binary(), desc := binary(), admin_only := boolean()}].
+login_user_scope_catalogue() ->
```
zmstone: nit: your agent is from UK ?
```suggestion
login_user_scope_catalog() ->
```

apps/emqx/src/emqx_scope_catalogue.erl
```diff
@@ -0,0 +1,163 @@
...
+    %% Tag each API-key catalogue entry as not admin-only and append
+    %% the four new entries.
+    [Entry#{admin_only => false} || Entry <- scope_catalogue()] ++
```
zmstone: add a function, admin_only_scope_catalog

zmstone: and ++ common_scope_catalog()

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -298,6 +360,111 @@ clear_login_lock(Username) ->
...
+%%     by force_mfa_snapshot. Recorded independently from policy so
+%%     that the field remains a clean audit trail of direct admin
+%%     decisions even if the policy is later turned off.
+%%
+%%   scopes :: [binary()]
+%%     Login user's scope list. undefined (key absent) means "fall
+%%     back to role default" — administrator gets implicit full
+%%     access; viewer/publisher get implicit empty (no new-scope
+%%     privileges). Explicit [] means "all mapped paths denied,
+%%     unmapped paths still allowed" (consistent with API key empty
```
zmstone, line 373: the existence of unmapped path should be a static check error at build time?

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -298,6 +360,111 @@ clear_login_lock(Username) ->
...
+%%     mechanism in this feat; future work).
+%%
+%%   admin_required :: boolean()
+%%     True iff an admin actively required MFA on a user NOT covered
+%%     by force_mfa_snapshot. Recorded independently from policy so
+%%     that the field remains a clean audit trail of direct admin
+%%     decisions even if the policy is later turned off.
+%%
+%%   scopes :: [binary()]
+%%     Login user's scope list. undefined (key absent) means "fall
```
zmstone, line 369: this fallback rule is convoluted.

to keep it simple, undefined should fallback to role default regardless of which role.

apps/emqx_dashboard/test/emqx_dashboard_user_scopes_SUITE.erl
```diff
@@ -0,0 +1,568 @@
...
+        <<"description">> => <<"test">>,
+        <<"scopes">> => [<<"bogus_scope_name">>]
+    },
+    ?assertMatch(
+        {ok, 400, _},
+        request_api(post, api_path(["users"]), auth_header(Token), Body)
+    ).
+
+%%--------------------------------------------------------------------
+%% MFA self-lock matrix (SPEC sec 6.3, the 7-line decision table)
```
zmstone: this pr included some SPEC references not committed to git.
leaving them will only lead to confusion.

clarify in self-containing code comments.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -192,6 +204,8 @@ do_disable_mfa(Username) ->
 
 %% @doc Enable MFA state.
 %% Return error if it's already enabled.
+%% Called from CLI (emqx ctl admins mfa <user> enable totp), so the
```
zmstone, line 215: maybe rename it to avoid future abuse.
`enable_mfa_from_cli` ?

apps/emqx/src/emqx_scope_catalogue.erl
```diff
@@ -0,0 +1,163 @@
...
+%% @doc Static catalogue of EMQX permission scopes.
+%%
+%% This module hosts the human-readable descriptions for the global
+%% scope vocabulary. It lives in `emqx' (the base app) because the
+%% scope name macros it references are already in
+%% `emqx/include/emqx_api_key_scopes.hrl' and the data is consumed
+%% by both `emqx_management' (API-key authorisation, /api_key_scopes
+%% endpoint) and `emqx_dashboard' (login user scope schema validation,
+%% /user_scopes endpoint). Putting it here keeps the data above the
+%% sibling apps in the dependency graph and avoids the implicit
```
savonarola: Nit: moving dashboard/mgmt code into core app also breaks dependencies, core emqx app now depends on management and dashboard apps (implicitly).
We may move to this common logic into the utils map, like `apps/emqx_utils/include/emqx_http_api.hrl` or `apps/emqx_utils/include/emqx_ds_dbs.hrl`. Or move in a new scope-dedicated app.

apps/emqx/src/emqx_scope_catalogue.erl
```diff
@@ -0,0 +1,163 @@
...
+        #{
+            name => ?SCOPE_CLUSTER_OPERATIONS,
+            desc => <<
+                "Cluster management, node operations, "
+                "load rebalancing, node eviction, and multi-tenancy"
+            >>
+        },
+        #{
+            name => ?SCOPE_SYSTEM,
+            desc => <<
```
savonarola: Should descs be localized?

apps/emqx/src/emqx_scope_catalogue.erl
```diff
@@ -0,0 +1,163 @@
...
+        #{
+            name => ?SCOPE_LICENSE,
+            desc => <<"License management">>
+        }
+    ].
+
+-doc """
+Catalogue of all scopes that a dashboard LOGIN USER may hold.
+
+Comprises the 10 API-key catalogue scopes plus the four login-only
```
savonarola: Nit: do we need to specify the exact counts? These numbers may become outdated soon.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -443,38 +514,284 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
...
+    case Caller of
+        undefined ->
+            %% No bearer token caller (shouldn't happen — bearer-only
+            %% endpoint, but guard anyway).
+            {deny, 401, 'UNAUTHORIZED', <<"Bearer auth required">>};
+        _ ->
+            authorize_mfa_change_with_caller(Caller, TargetUsername, Op)
+    end.
+
+authorize_mfa_change_with_caller(Caller, TargetUsername, Op) ->
```
savonarola, line 775: IMO having additional condition based authorizations makes the whole authorization more fragile. Bugs may be introduced and unnoted, agents may sweep out important conditions, etc.

To protect ourselves and the code from possible issues, maybe we may try to keep the things more declarative: just have different routes for changing own creds and changing arbitrary user's creds. 

Changing arbitrary user's creds may be available only to the admins; while changing own creds to any user. Changing own creds should not contain a user argument to check at all (`/users/:username/mfa` -> `/me/mfa`), so another user cannot slip in.

savonarola: @zmstone wdyt?

savonarola: IMO having different rotes for other parts of the (quite complex) decision matrix would also simplify things, like  having `/me/init_mfa` for the initial setup, etc.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -306,43 +348,68 @@ logout(_, #{
             {401, ?WRONG_TOKEN_OR_USERNAME, <<"Ensure your token & username">>}
     end.
 
+user_scopes(get, _Request) ->
+    {200, #{scopes => emqx_scope_catalogue:login_user_scope_catalogue()}}.
+
 users(get, _Request) ->
     {200, filter_result(emqx_dashboard_admin:all_users())};
```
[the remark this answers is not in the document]
JimMoen: Fixed in 5bf0c03 (now part of the subsequent restructure). `emqx_dashboard_admin:to_external_user/1` now projects the effective scope list for every entry returned by `GET /users`; the same projection is used by POST / PUT responses so all three paths match the `user_fields()` schema.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -443,38 +510,267 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
...
+finalise_update_user(Username, Scopes, Result) ->
+    case maybe_set_user_scopes(Username, Scopes) of
+        ok ->
+            {200, filter_result(Result)};
+        {error, <<"username_not_found">> = Reason} ->
+            {404, ?USER_NOT_FOUND, Reason};
+        {error, Reason} ->
+            {400, ?BAD_REQUEST, Reason}
+    end.
+
```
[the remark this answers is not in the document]
JimMoen: Fixed in 5bf0c03. After `maybe_set_user_scopes/2` succeeds, the handler re-reads the user record and projects through `emqx_dashboard_admin:to_external_user/1`, so the PUT body now reflects the persisted state including the effective scopes list.

changes/ee/feat-17235.en.md
```diff
@@ -0,0 +1,5 @@
+Introduced fine-grained scope-based access control for dashboard login users. Each user record now carries an optional `scopes` field; when set, requests are authorised against the same path-to-scope catalogue used for API keys, layered on top of the existing role-based check. Four new scopes (`user_management`, `mfa_management`, `sso_management`, `api_key_management`) cover dashboard-only endpoints and are admin-only except `mfa_management`, which any role may hold for self-exemption from forced MFA. API keys cannot hold any of the four login-only scopes, and the `publisher` API-key role is now constrained to `[publish]` only — both checks apply to the HTTP API and to bootstrap-file loading (incompatible scopes are dropped with a warning).
+
+Two new public catalogue endpoints expose the scope vocabulary for UI consumption: `GET /api_key_scopes` (10 entries) and `GET /user_scopes` (14 entries with an `admin_only` flag). Both are top-level paths so they cannot be intercepted by sibling wildcard routes, and both are accessible to any bearer-authenticated caller.
+
```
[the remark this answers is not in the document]
JimMoen: The tightening was reverted in 49ca5c9 — `GET /sso/running` is back to public (`security => []`) because the dashboard login page needs to probe it before authentication to render the SSO button list. No changelog entry needed since there is no net behaviour change versus the release-510 base.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
JimMoen: Fixed in f2218de. Under CE the `scopes` field is now stripped from the schema (`field_filter/1`), the request body (handler's `maybe_set_user_scopes/2` is a noop in CE), and the response projection (`ee_user_extra/1` returns `#{}` in CE) — CE clients can neither see nor set scopes, removing the false sense of restricted access.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
JimMoen: Fixed in 5bf0c03. The `null` sentinel was removed; `to_external_user/1` now projects through `emqx_dashboard_admin:effective_scopes_of_admin/1`, which expands the `undefined` storage value into the role-default scope list (administrator → common + login-only; viewer → common only). The response is always an array, matching the `array(binary())` schema.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
JimMoen: Fixed in 30b193e. `emqx_dashboard_rbac:check_rbac/5` no longer consults the live SSO `force_mfa` flag for viewer `DELETE /users/:self/mfa`; the policy decision moved entirely into `emqx_dashboard_api:authorize_mfa_change/3`. The handler now uses an `admin_override` tri-state (`mfa_required | mfa_exempted | undefined`) which superseded the `force_mfa_snapshot` + `admin_required` pair (the snapshot field was removed in 6065a37) — `mfa_management` scope holders can now self-exempt as documented.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
JimMoen: There are two gates already in place:

* `scripts/check-api-scopes.sh` runs in CI as a build-time check and fails the build if any `minirest_api` module is missing `scopes/0`.
* `t_all_endpoints_covered_by_scopes` in `emqx_mgmt_api_key_scopes_SUITE` walks every declared path from the loaded apps and asserts each one either maps to a scope or appears in `IntentionallyUnmapped` (public endpoints).

Both must pass before merge, so an unmapped path can't slip through.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
JimMoen: Simplified in 5bf0c03. `effective_scopes_of_admin/1` now uses a single role-default fallback regardless of which role:

```erlang
case scopes_of(Username) of
    undefined -> role_default_scopes(Role);   %% admin: full; viewer: common
    Scopes when is_list(Scopes) -> Scopes
end
```

The previous two-tier `if snapshot then ... else ...` branch was dropped when `force_mfa_snapshot` itself was removed in 6065a37.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
JimMoen: Cleaned up in f75c684 — all references to the unmerged SPEC and to internal commit / branch names have been removed from in-tree comments. Doc strings now stand alone.

apps/emqx/include/emqx_api_key_scopes.hrl
JimMoen: Applied in f75c684.

apps/emqx/src/emqx_scope_catalogue.erl
JimMoen: Renamed in f75c684 — `catalogue → catalog` across module name, exported functions, hocon entries, comments and the changelog. The agent's defaults are British English; switched to US for this codebase.

apps/emqx/src/emqx_scope_catalogue.erl
JimMoen: Done in f75c684. `emqx_scope_catalog` now exposes:

```erlang
common_scope_catalog/0      %% shared between API keys and login users
admin_only_scope_catalog/0  %% login-only scopes, with admin_only flag
scope_catalog/0             %% == common_scope_catalog/0 (API-key view)
login_user_scope_catalog/0  %% common ++ admin_only_scope_catalog/0
```

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
JimMoen: Renamed in f75c684 — `emqx_dashboard_admin:enable_mfa/2` is now `enable_mfa_from_cli/2`. The only caller is `emqx_dashboard_sso_cli`, where the admin-by-CLI assumption is now explicit in the name.

apps/emqx/src/emqx_scope_catalogue.erl
JimMoen: Moved in 360db9b. `emqx_scope_catalog` and the scope macro header now live under `apps/emqx_utils/`. The base `emqx` app no longer carries any dashboard/management feature data.

apps/emqx/src/emqx_scope_catalogue.erl
JimMoen: Done in 12a07fa. `emqx_scope_catalog` now returns `?DESC(Id)` handles instead of hard-coded binaries; the strings live in `rel/i18n/emqx_scope_catalog.hocon`. The two catalog HTTP handlers (`api_key_scopes` and `user_scopes`) resolve the handle through `emqx_dashboard_swagger:get_i18n/4` before serializing the response.

apps/emqx/src/emqx_scope_catalogue.erl
JimMoen: Removed in f0f8339. Hard-coded scope counts ("10", "14", "four") are gone from comments, doc strings and the changelog; CT assertions that referenced the counts now derive them from `length(emqx_scope_catalog:...)` so the tests stay correct as the catalog evolves.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
JimMoen: Good design point. Splitting `/users/:username/mfa` into `/me/mfa` (self) and `/users/:other/mfa` (admin) would let the routing layer carry the self-vs-other distinction so handlers can stop checking it.

The change requires frontend coordination (emqx-dashboard5 currently calls `/users/<self>/mfa` for both setup and disable) and would also rework the SSO username encoding in the path. I'd prefer to track that as a follow-up PR rather than expand this one further.

For now the condition-based authorization is pinned by the matrix in `emqx_dashboard_user_scopes_SUITE` (8 cases for the IsSelf × HasMfaMgmt × Locked × FirstSetup truth table) and `emqx_dashboard_rbac_SUITE`.

JimMoen: Agreed — `/me/init_mfa` and friends would let the routes encode the operation type instead of overloading POST/DELETE on the same path. I'll roll all of that into the follow-up that introduces `/me/*`.

apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
```diff
@@ -28,6 +30,86 @@ check_rbac(Req, Username, Extra) ->
...
+check_login_user_scopes_for_path(Username, Path) ->
+    %% Self-targeted user endpoints (own change_pwd / own MFA) bypass
+    %% the scope check — they are governed by RBAC's self-check (above)
+    %% and, for MFA, by emqx_dashboard_api:authorize_mfa_change/3
+    %% (force_mfa snapshot, admin_required, mfa_management self-
+    %% exemption). Subjecting them to user_management / mfa_management
+    %% scope here would lock viewers out of changing their own
+    %% password / setting up their own MFA — exactly the opposite of
+    %% what the scope was designed to gate (managing OTHER users).
+    case is_self_user_endpoint(Path, Username) of
```
hjianbo: This self-endpoint bypass is broader than the comment describes. It allows every `/users/<self>/...` path to skip the login-user scope check, including `PUT /users/<self>`. A superuser with explicit `scopes = []` can therefore update its own record and restore `user_management`, even though the explicit scope list was meant to restrict user-management access. Please narrow the bypass to only the self-service paths that need it, e.g. `/users/<self>/change_pwd` and `/users/<self>/mfa`.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -443,38 +518,285 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
...
+                <<"Non-administrator users cannot hold admin-only scopes: ">>, Names
+            ]),
+            {error, Msg}
+    end.
+
+%% Run the validate → update_user → set_scopes pipeline. Mirrors
+%% create_user/5 above, also kept as a helper to stay within elvis's
+%% nesting cap.
+update_user(Username, Role, Desc, Scopes) ->
+    case validate_login_user_scopes(Role, Scopes) of
```
hjianbo: This validates only the scopes submitted in the current request. If an administrator with persisted admin-only scopes is demoted to `viewer` via `PUT /users/:username` without a `scopes` field, `maybe_set_user_scopes(undefined)` preserves the old scope list, so the viewer can keep scopes such as `user_management` or `api_key_management`. Role updates should validate the final persisted scope set, or clear/recompute incompatible scopes when the role changes.

apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
```diff
@@ -277,7 +291,7 @@ api_key_by_name(put, #{bindings := #{name := Name}, body := Body}) ->
     Desc = maps:get(<<"desc">>, Body, undefined),
     Role = maps:get(<<"role">>, Body, ?ROLE_API_DEFAULT),
     Scopes = maps:get(<<"scopes">>, Body, undefined),
-    case validate_scopes(Scopes) of
+    case validate_scopes(Role, Scopes) of
```
hjianbo: The same stale-scope problem exists for API keys. When a key is updated to role `publisher` without a `scopes` field, this validates `undefined` as acceptable and `emqx_mgmt_auth:maybe_set_scopes/2` keeps the previous persisted scopes. Runtime RBAC still limits publisher requests today, but the stored config and API response can violate the invariant that publisher keys may only hold `publish`. Please validate against the final persisted scopes, or reset scopes to a publisher-compatible value when the role changes.

hjianbo (commented):
Reviewed the latest head and left three inline comments on remaining authorization/persistence edge cases.

apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
JimMoen: Good catch — fixed in 6281b5563f.

The self-bypass is now constrained to the `change_pwd` and `mfa` self-service endpoints via a small whitelist (`is_self_service_endpoint/2` in `emqx_dashboard_rbac.erl`); every other handler now goes through the normal scope check, so a viewer hitting `/users/:self` no longer escapes the `user_management` requirement.

Coverage:
- New CT case `t_check_login_user_scopes_self_user_record_not_bypassed` in `emqx_dashboard_rbac_SUITE` exercises a self-targeted PUT against a non-whitelisted endpoint and asserts the regular scope check still applies.
- Existing self-rotate / self-pwd cases continue to pass, confirming the whitelist preserves the intended self-service paths.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
JimMoen: Fixed in 43eee0c3f7.

`update_user/4` in `emqx_dashboard_api.erl` now computes an *effective* scope list — the request body's `scopes' field when supplied, otherwise the persisted scopes via `emqx_dashboard_admin:scopes_of/1` — and feeds that into `validate_login_user_scopes/2`. The persistence layer still receives the original (possibly `undefined`) value so partial-update semantics are preserved: callers that did not touch `scopes' do not have their scopes rewritten.

Net effect for the scenario you described: a `PUT /users/:username` that demotes an administrator with persisted `user_management` (or any other admin-only scope) to `viewer` without supplying `scopes` is now rejected with 400, and the persisted role/scopes remain intact.

Coverage in `emqx_dashboard_user_scopes_SUITE`:
- `t_role_demotion_with_persisted_admin_scopes_is_rejected` — demote with persisted `user_management`, asserts 400 + persisted state unchanged.
- `t_role_demotion_with_compatible_persisted_scopes_succeeds` — demote with persisted `mfa_management` (any-role compatible), asserts 200 + role updated.

Side note: while addressing this, I also tightened the related break-glass invariant for `dashboard.default_username` in f4c33dc17a — the default administrator can no longer be deleted, role-changed, or assigned an explicit scope list via the REST API, regardless of how many other admins exist. This closes a related lockout scenario (the previous "last-man-standing" delete rule could be evaded by first adding a throwaway admin).

apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
JimMoen: Fixed in 43eee0c3f7 (same commit as the dashboard-user counterpart).

`api_key_by_name(put, ...)` in `emqx_mgmt_api_api_keys.erl` now resolves the effective scope list before calling `validate_scopes/2`: when the request body omits `scopes`, the persisted scopes are read via `emqx_mgmt_auth:read/1` (the existing reader) and validated against the new role. The downstream `emqx_mgmt_auth:update/6` call still receives the original (possibly `undefined`) value so partial-update semantics are preserved.

Net effect: changing an admin-role key with persisted non-`publish` scopes to `publisher` without resetting `scopes` is now rejected with 400 publisher-only error, and the stored config + GET response remain consistent with the publisher-only-`publish` invariant — defense-in-depth on top of the runtime RBAC check you mentioned.

Coverage in `emqx_mgmt_api_api_keys_SUITE`:
- `t_ee_publisher_role_change_with_persisted_admin_scopes_is_rejected` — admin → publisher with persisted `[connections, monitoring]`, asserts 400 + persisted role/scopes unchanged.
- `t_ee_publisher_role_change_with_compatible_persisted_scopes_succeeds` — admin → publisher with persisted `[publish]`, asserts 200 + role updated.

ieQu1 (dismissed):

hjianbo (commented):
Reviewed the latest head and left two inline comments on the remaining edge cases.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -306,64 +351,126 @@ logout(_, #{
...
+    case is_default_admin_modification(Username, Role, Scopes) of
+        ok ->
+            update_user(Username, Role, Desc, Scopes);
+        {error, Msg} ->
+            {400, ?NOT_ALLOWED, Msg}
     end;
 user(delete, #{bindings := #{username := Username}} = Req) ->
-    DefaultUsername = emqx_dashboard_admin:default_username(),
-    case Username == DefaultUsername of
+    case is_default_admin(Username) of
```
hjianbo, line 435: This checks the raw path username before applying `username(Req, Username)`, unlike the PUT path above. For an SSO request such as `DELETE /users/admin?backend=ldap`, the actual target is `{ldap, <<"admin">>}`, but this branch compares `<<"admin">>` with `dashboard.default_username` and rejects it as the local break-glass account. Please resolve the target username first and only protect the local default admin.

changes/ee/feat-17235.en.md
```diff
@@ -0,0 +1,11 @@
...
+- Self-service on a user's own record now respects scopes. Only the dedicated change-password and MFA self endpoints still bypass scope checks; other operations such as `PUT /users/:self` are subject to the user's scopes. A viewer without `user_management` can still change their own password and manage their own MFA, but can no longer edit other profile fields on their own record without being granted the relevant scope.
+- `PUT /users/:username` and `PUT /api_keys/:name` validate role changes against the effective persisted scopes when the request body omits the `scopes` field. Demoting a user or changing an API key role is rejected if the persisted scopes are incompatible with the new role, ensuring role and scopes stay consistent even across partial updates.
```
hjianbo: The actual API route is `PUT /api_key/:name` (singular), as registered in `emqx_mgmt_api_api_keys:paths/0`. Please update the changelog path so the release note matches the HTTP API.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
JimMoen: Good catch — fixed in `d0759940b5`.

The DELETE branch now resolves the SSO target via `username/2` before consulting `is_default_admin/1`, mirroring the PUT path. `is_default_admin/1` also gained a binary guard so SSO targets (tuples) are categorically never matched even if a future caller forgets to resolve them first.

Coverage in `emqx_dashboard_user_scopes_SUITE`:
- `t_default_admin_protection_does_not_apply_to_sso_users` issues `DELETE /users/admin?backend=ldap` and asserts the response is not the break-glass `default administrator` rejection (it goes on to fail downstream because the SSO backend isn't started in this suite — only the dispatch path is under test).
- `t_default_admin_cannot_be_deleted` continues to pass, confirming the local-target protection is intact.

changes/ee/feat-17235.en.md
JimMoen: Fixed in `65c91542c2` — changelog now says `PUT /api_key/:name` to match the route registered in `emqx_mgmt_api_api_keys:paths/0`.

savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14100)

PR #17242, opened 2026-05-09
fix(ds_local): keep meta_worker tick non-periodic

zmstone:
Release version: 5.8.10

## Summary

Follow-up to PR #17228. The periodic-tick portion of that fix made
`erlang:send_after/3` deliver `{?MODULE, tick}` so it would match the
handler clause and the timer would actually fire every 100 ms.

That broke `emqx_ds_builtin_local_SUITE:t_store_batch_fail` under
cover-compiled enterprise CI — caught when syncing forward to
release-510 (PR #17241): `meck:unload` could not purge the
cover-compiled `emqx_ds_storage_layer` beam while the worker was
actively calling into it every 100 ms.

The pre-existing `tick` atom (which falls through to the no-op
catchall and stops the timer after one iteration) is restored. The
badarith guard in `tick/2` — which is the change `t_state_fuzz` on
master needed — is kept.

## PR Checklist


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14084)

PR #17244, opened 2026-05-10
refactor(relup): official release package for relup target

zmstone:
Release version: 5.10.4

## Summary

Replaces the in-tree relup framework with a CLI-only operator flow. The target EMQX release tarball is staged out-of-band by the operator; the plugin verifies it, extracts the runtime subdirs, applies the matching `.relup` hop, and writes a version marker so the next `bin/emqx` start execs into the new tree.

PR #17263 already landed the mechanical vendoring of the plugin into `plugins/emqx_relup/` and PR #17260 landed the removal of the legacy `install_upgrade.escript` / versioned-bin-copy machinery. After rebasing this branch on top of those, the diff here is just the framework refactor plus a CLI surface for upgrade logs.

## Two commits

1. `refactor(relup): official tgz package as upgrade target` - the framework refactor (see below).
2. `feat(relup): expose upgrade-log API via CLI` - wires the existing `emqx_relup_log` mnesia table (per-node audit trail) to three new CLI commands.

## Operator workflow (new model)

1. Install the plugin: download `emqx_relup-<vsn>.tar.gz` from `https://packages.emqx.io/emqx-plugins/e<EMQX-VSN>/`, drop it in the plugin install dir, run `emqx ctl plugins install/enable/start emqx_relup-<vsn>` on every node.
2. `emqx ctl relup list-supported-paths` - confirm the `{From, Target}` hop is in the plugin's `priv/relup/` catalog.
3. `scp` the EMQX target tarball + `<tarball>.sha256` sidecar to every node (anywhere readable; no staging dir).
4. `emqx ctl relup upgrade <TarballPath>` on each node. The handler verifies the sha256, extracts the tarball, reads the target version from `releases/emqx_vars` (`REL_VSN`), and runs the catalog's matching code-change + post-upgrade-callbacks.
5. Verify with `emqx ctl status`, `cat <RootDir>/relup/version`, and `emqx ctl relup status` / `emqx ctl relup logs`. Restart the node; the `bin/emqx` wrapper detects `<RootDir>/relup/version` and execs into the deployed tree.
6. Manual `rm` once the cluster is fully on the target version.

### Framework

- **Lookup is tarball-driven, not sidecar-or-CLI-driven.** Target version is read from the tarball's own `releases/emqx_vars` (`REL_VSN`) after sha256 verification. No `.target_vsn` sidecar, no CLI version arg, no version-from-filename parsing.
- **`<tarball>.sha256` enforced before extraction** (bare digest and `sha256sum <digest>  <filename>` formats both accepted).
- **Deploy footprint is filtered.** Only `bin/`, `erts-*/`, `lib/`, `releases/` from the tarball end up under `<RootDir>/relup/<TargetVsn>/`. `data/`, `etc/`, `log/`, `plugins/` stay rooted at the original install (the `bin/emqx` wrapper preserves `BASE_RUNNER_ROOT_DIR` across re-exec for them). This also prevents a deploy-dir-shaped relup loop.
- **`.relup` catalog is source-controlled** under `plugins/emqx_relup/priv/relup/`, one file per `{From, Target}` hop. Validated at app start; malformed entries are logged at `warning` and skipped, not fatal.
- **REST API removed.** `apps/emqx_management/src/emqx_mgmt_api_relup.erl` and the v1 proto module are gone. Dashboard surface is also dropped (frontend will hide the page in a coming release). Exempted in `?FORCE_DELETED_APIS` + `?FORCE_DELETED_MODULES` to keep bpapi static checks happy.
- **No `deploy_inplace=true`.** Only the independent `<RootDir>/relup/<TargetVsn>/` deploy remains.

### CLI

- `emqx ctl relup upgrade <TarballPath>` - apply a hop.
- `emqx ctl relup list-supported-paths` - print supported `{From, Target}` hops from the catalog.
- `emqx ctl relup status` - `in-progress` or `idle`.
- `emqx ctl relup logs` - print this node's upgrade history.
- `emqx ctl relup logs-clear` - wipe this node's log table.

The log table (`emqx_relup_log`) is `{local_content, true}` mnesia, persisted on disk. Nothing in the plugin deletes it on stop/unload, so plugin uninstall keeps the rows; reinstall re-attaches.

### Packaging

- `<RootDir>/relup/` shipped empty by default. relx overlay (`{mkdir, "relup"}`), deb `rules`, and rpm `spec` all create the dir; ownership comes from the existing `chown emqx:emqx` postinst step (deb) and `%files` attr (rpm).

## Test coverage

- `plugins/emqx_relup/test/emqx_relup_handler_SUITE` - catalog validation (well-formed / invalid-term / unparseable).
- `plugins/emqx_relup/test/emqx_relup_round_trip_SUITE` - full `check_and_unpack -> perform_upgrade -> permanent_upgrade` against a forged target tarball + sha256 sidecar + no-op `.relup`. Negative cases: missing/mismatched sha256, missing `releases/emqx_vars`, no catalog entry, os/arch mismatch. Positive case for `sha256sum`-format sidecar. Deploy-only-runtime-dirs assertion (noise dirs in the tarball don't leak into the deploy tree).
- `plugins/emqx_relup/test/fixtures/make-fake-target.sh` - forges a no-op target tarball for end-to-end smoke testing without a real release build.

Plus a manual test plan (uncommitted) covers the bash-side `bin/emqx` re-exec path that CT cannot reach.

## Breaking changes

`changes/ee/breaking-17244.en.md` covers the REST API removal and the new CLI-only operator surface.

## PR Checklist


terry-xiaoyu (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14198)

PR #17247, opened 2026-05-10
fix(plugins): log method/path on API callback crash, hint timeout config

zmstone:

## Summary

When a plugin's REST API callback crashes or runs over its time budget, the gateway now logs the failing API call's method and path along with the configured timeout — so on a busy node the offending call is identifiable instead of just `plugin_api_callback_crash, reason: timeout`.

Two specific fixes in `apps/emqx_plugins/src/emqx_plugins.erl` `handle_api_call/3`:

- `emqx_utils:nolink_apply/2` exits with the bare atom `timeout`, not `{timeout, _}`. The old `exit:{timeout, _}` clause never matched, so every timeout fell through to the generic crash log. Now caught as `exit:timeout` with a dedicated `plugin_api_callback_timeout` warning that names `plugins.api_endpoint.timeout` as the knob to raise.
- The crash log now includes `method`, `path`, and `timeout_ms` alongside the existing `plugin/class/reason/stacktrace` fields.

No behavior change for synchronous handlers that complete on time; only the failure-path logs and the 503 timeout response are affected.

## PR Checklist


savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14119)

PR #17254, opened 2026-05-12
chore(deps): bump lc to 0.3.7

zmstone:
Release version: 6.0.3

## Summary

Bump `lc` from 0.3.4 to 0.3.7.

Picks the most constraining memory reading among cgroup v2, cgroup v1, and host `/proc/meminfo`. Smallest non-zero total wins, with larger usage ratio breaking ties.

Before this fix the reading could be misleading in two ways inside containers:

- a tight cgroup limit was masked by the host view (e.g. host >70% while cgroup limit <10%, or the reverse), because `lc` was taking `lists:max/1` over the three readings;
- after the 0.3.6 attempt to prefer cgroup, hosts where a cgroup is mounted with no memory limit set returned the cgroup v1 sentinel (~8 EiB), collapsing the reported usage ratio to ~0%.

0.3.7 picks the smallest non-zero total instead, so the binding limit is always selected — a real cgroup cap is at most node memory; an unlimited cgroup's sentinel is far larger than node memory.

Affects overload-protection thresholds and the broker's reported memory metric inside containers.

Upstream PR: https://github.com/emqx/lc/pull/23

## PR Checklist
- [x] Change log added under `changes/ee/fix-17254.en.md`


savonarola (dismissed):

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14178)

PR #17258, opened 2026-05-13
fix(mqttsn): handle duplicate CONNECT without crashing the connection process

zmstone:

## Summary

A second CONNECT received on an already-connected MQTT-SN session caused
the connection process to exit with a `case_clause`. The channel returned
`{error, unexpected_connect, Channel}` from
`emqx_mqttsn_channel:handle_in/2` (`conn_state = connected`), but
`emqx_gateway_conn:with_channel/3` has no matching clause for that shape,
so the process crashed and the client's session was lost.

This change makes the channel react like the regular MQTT channel: log a
warning and call `handle_out(disconnect, ?SN_RC_NOT_SUPPORTED, Channel)`,
which sends an `SN_DISCONNECT` and shuts the session down via the normal
`{close, Reason}` path.

Scope is intentionally narrow:
* `apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl` — replace the
  unhandled `{error, ...}` return with `handle_out(disconnect, ...)`.
* `apps/emqx_gateway_mqttsn/test/emqx_sn_protocol_SUITE.erl` — new
  `t_duplicate_connect` regression test: two CONNECTs on one UDP socket,
  expect `SN_DISCONNECT` and no crash.

Note: per-peer UDP processes are tracked by `esockd_udp` (linked, not
supervised with a restart spec), so this crash does not contribute to any
OTP restart-intensity budget — but it still kills the targeted client's
session and produces noisy crash logs, which is the behaviour the fix
addresses. `emqx_stomp_channel.erl:430` has the same `{error, unexpected_connect, _}`
shape; left for a separate change.

## PR Checklist

### DIFF
--- a/apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
+++ b/apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
@@ -524,7 +524,8 @@ handle_in(
     ?SN_CONNECT_MSG(_Flags, _ProtoId, _Duration, _ClientId),
     Channel = #channel{conn_state = connected}
 ) ->
-    {error, unexpected_connect, Channel};
+    ?SLOG(warning, #{msg => "receive_connect_packet_in_connected_state"}),
+    handle_out(disconnect, ?SN_RC_NOT_SUPPORTED, Channel);
 handle_in(
     ?SN_WILLTOPIC_EMPTY_MSG,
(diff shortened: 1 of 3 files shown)


hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14179)

PR #17259, opened 2026-05-13
refactor: drop unused package management scripts and commands

zmstone:
Drops the OTP `release_handler` / appup-based hot-upgrade plumbing that has been advertised but non-functional on v6.

Removes:
- `bin/install_upgrade.escript` (entire file, 503 lines).
- `bin/emqx` ctl commands: `versions`, `unpack`, `install`, `uninstall`, `upgrade`, `downgrade` (196 lines).
- `mix.exs`: the `&copy_escript(&1, "install_upgrade.escript")` release step and the versioned-binary copy loop (`bin/emqx-<vsn>`, `bin/emqx_ctl-<vsn>`, `bin/install_upgrade.escript-<vsn>`).

release-60 has no `rebar.config.erl` (build is mix-driven), so that hunk from the source commit is dropped.


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14180)

PR #17263, opened 2026-05-13
chore(relup): vendor emqx-relup plugin into plugins/

zmstone:
Vendor the emqx-relup plugin (https://github.com/emqx/emqx-relup, tag 0.2.2) into the monorepo as an in-tree umbrella plugin under `plugins/emqx_relup/`. Standalone, mechanical move; runtime modules are unchanged from upstream modulo `erlfmt`.

This is split out from PR #17244 so the vendoring lands first as a small, easy-to-review changeset; the framework refactor (CLI redesign, upload-staged target, REST API removal, etc.) will rebase on top once this merges.

## What's in this PR

- New umbrella plugin at `plugins/emqx_relup/` (handler, main, cli, sup, app, libs, file_utils, utils, plus rebar.config and emqx_relup.app.src).
- `rebar.config.erl`: drop the external `{emqx_relup, {git, ...}}` dep.
- `build`: drop `make_relup()` and the `relup` artifact step.
- `Makefile`: drop `\$(REL_PROFILES:%=%-relup)` target.

No user-facing behavior change. No changelog entry by design (the plugin still installs and behaves as before).

## Follow-up

PR #17244 will be rebased on top of this once merged, leaving only the refactor commits to review there.


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14123)

PR #17265, opened 2026-05-14
chore(relup): vendor emqx-relup plugin into plugins/emqx_relup

zmstone:
Port of #17263 from release-510 onto release-60.

## Summary

Moves the `emqx-relup` plugin source into the monorepo as an umbrella plugin at `plugins/emqx_relup/`. Drops the relup-target plumbing in `Makefile` and `build` (the `make_relup` helper and the `relup` artifact case), which was the last consumer of the external `emqx-relup` git dep — both belonged to the previous design where the upgrade target was bundled into the plugin tarball at build time.

The runtime modules (handler, main, cli, sup, app, libs, file_utils, utils) come verbatim from the source PR. The upstream `rebar.config` is replaced by a `mix.exs` modeled on `plugins/emqx_username_quota/mix.exs`, since release-60 builds plugins through mix rather than rebar3. The `emqx_plugrel` metadata block in the upstream rebar.config maps to the `:emqx_plugin` keyword list in `mix.exs` (authors, builder, repo, functionality, compatibility, description, hidden). The application metadata that lived in `emqx_relup.app.src` is now generated from the `mix.exs` `application/0` callback.

Build verification: `make plugin-emqx_relup` produces `_build/plugins/emqx_relup-0.2.2.tar.gz` plus its `.sha256` sidecar; full `make` release assembly is unaffected.

This is preparation for refactoring the framework so that the upgrade target tarball is staged out-of-band by the operator rather than packaged into the plugin's `priv` dir.

## PR Checklist


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14182)

PR #17267, opened 2026-05-14
feat(conf): default node.max_ports to auto

zmstone:
Fixes <issue-or-jira-number>

Release version: 6.3.0

## Summary

- `node.max_ports` schema changed from a fixed integer to `union(auto, range(1024, 134217727))`, with default `auto`.
- `auto` resolves to `cores * 65536` for hosts with up to 8 logical CPUs (via `tr_vm_args_max_ports/1`), falling back to the historical `1048576` for larger hosts. 65536 is 2^16, the next power of two the VM would round up to anyway, so the configured and effective `+Q` match.
- `+P` (Erlang process limit) is derived as `2 * max_ports` to leave headroom for TLS/QUIC workloads where one connection can spawn multiple processes.
- The hidden `node.process_limit` setting is no longer deprecated; it is reinstated as an override that wins only when strictly larger than the derived value (smaller user values are ignored to keep the process table from under-sizing the port table).
- New `node.schedulers` setting (default `auto`) caps `+S` at `logical_processors_available`, so a container limited via `--cpuset-cpus` or k8s CPU requests doesn't spawn scheduler OS threads it can't run in parallel. A user-specified positive integer overrides the auto-detected value.

Most files touched:
- `apps/emqx_conf/src/emqx_conf_schema.erl` — schema + translation
- `rel/i18n/emqx_conf_schema.hocon` — `max_ports`/`process_limit`/`schedulers` descriptions

## PR Checklist

### DIFF
--- a/changes/ee/feat-17267.en.md
+++ b/changes/ee/feat-17267.en.md
@@ -1,3 +1,0 @@
-The `node.max_ports` config now defaults to `auto`, which scales the Erlang VM port limit (`+Q`) with the number of logical CPU cores: 64000 ports per core for up to 8 cores, and 1048576 (the historical fixed default) above that. Explicit integer values are still accepted.
-
-The Erlang process limit (`+P`) is now derived as `round(1.2 * max_ports)` instead of the previous `2 * max_ports`, reducing memory used by the process table on small deployments. The hidden `node.process_limit` setting is reinstated as an override: when set to a value larger than the derived limit, it is respected; smaller values are ignored so the process table never under-sizes the port table.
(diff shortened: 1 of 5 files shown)


changes/ee/feat-17267.en.md
qzhuyan, line 1: The operator need take extra care about this change. 

I personlly think it is better to not make this change considering all the deplyed nodes.


If we really want this, I think it is safer to say this is a  breaking change insead. As old setup which defaults to 1M ports will  stop work when reach the new limit and restart the VM will be the only solution.

changes/ee/feat-17267.en.md
```diff
@@ -0,0 +1,3 @@
+The `node.max_ports` config now defaults to `auto`, which scales the Erlang VM port limit (`+Q`) with the number of logical CPU cores: 64000 ports per core for up to 8 cores, and 1048576 (the historical fixed default) above that. Explicit integer values are still accepted.
+
+The Erlang process limit (`+P`) is now derived as `round(1.2 * max_ports)` instead of the previous `2 * max_ports`, reducing memory used by the process table on small deployments. The hidden `node.process_limit` setting is reinstated as an override: when set to a value larger than the derived limit, it is respected; smaller values are ignored so the process table never under-sizes the port table.
```
qzhuyan: 2 x max_port was designed for TLS  and QUIC I remember.

changes/ee/feat-17267.en.md
zmstone: this is required for small deployments optimized to consume less than 100MB RAM on a fresh boot.

zmstone: will provide more context in another PR.

qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16058)

PR #17271, opened 2026-05-15
build(docker): apply Debian security upgrades and drop unused libgnutls30t64

zmstone:
Fixes <issue-or-jira-number>

## Summary

Hardens the runtime stage of the official EMQX docker image so it passes customer image-scanner checks against `emqx/emqx:6.2.0`:

- `apt-get upgrade` is run in the runtime stage so the image picks up the latest Debian security patches.
- `libgnutls30t64` is purged. EMQX talks TLS via OpenSSL through Erlang/OTP and never links GnuTLS, so it was only present as a transitive dependency of `curl` and showed up as a scanner finding. The removal is guarded with `dpkg-query` so a future base image that doesn't ship the package will not break the build.
- `curl` is replaced by a pinned, SHA256-verified statically-linked binary from [stunnel/static-curl](https://github.com/stunnel/static-curl) 8.20.0 (OpenSSL build, HTTP/2 + HTTP/3, no librtmp, no GnuTLS). The fetch happens in a discardable `curl_static` build stage based on `BUILD_FROM` because the emqx-builder image already ships `curl`, `xz` and `tar`; only the verified binary is copied into the runtime image. Container healthchecks that call `curl` continue to work unchanged.

Relevant file: `deploy/docker/Dockerfile`.

Verified locally by building `emqx/emqx-enterprise` for amd64, booting it, running `debsecan --suite trixie` against installed packages, and confirming `curl https://...` works from inside the container and successfully fetches the dashboard API at `http://localhost:18083/api/v5/status` (HTTP 200).

## PR Checklist

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -2,6 +2,30 @@ ARG BUILD_FROM=ghcr.io/emqx/emqx-builder/6.1-4:1.18.3-27.3.4.2-8-debian13
 ARG RUN_FROM=debian:13-slim
 ARG SOURCE_TYPE=src # tgz
 
+# Fetch a self-contained `curl` binary from stunnel/static-curl. Debian's stock
+# curl drags libgnutls30t64 in via librtmp1 (libcurl4t64 -> librtmp1 -> gnutls),
+# which is what makes scanners flag the four GnuTLS CVEs even though EMQX
+# itself never links GnuTLS. The static binary is built with OpenSSL and no
+# RTMP, so we get `curl` back without re-introducing GnuTLS. This stage is
+# discarded after the COPY in the runtime stage; only the verified binary
+# ships in the final image. Based on BUILD_FROM (the emqx-builder image)
+# because it already ships curl + xz + tar; no apt-install needed.
+FROM ${BUILD_FROM} AS curl_static
+ARG TARGETARCH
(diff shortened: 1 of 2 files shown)


deploy/docker/Dockerfile
```diff
@@ -47,10 +71,28 @@ ENV LC_ALL=C.UTF-8
...
+# Runtime image hardening:
+#   * apt-get upgrade applies Debian security patches (e.g. patched libssl3t64
+#     clears CVE-2026-28387/28388/28389/28390).
+#   * libgnutls30t64 is removed: EMQX talks TLS via OpenSSL through Erlang/OTP
+#     and never links GnuTLS, so dropping it clears scanner findings for
+#     CVE-2026-42010/33845/42011/3833 without waiting for Debian patches.
+#     Guarded by dpkg-query so the build still works if a future base image
+#     does not ship libgnutls30t64.
+#   * curl ships as a statically-linked binary (COPY above) instead of via
+#     apt, because Debian's curl pulls libgnutls30t64 in via librtmp1.
```
zmstone: already documented above.
```suggestion
```

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14390)

PR #16942, opened 2026-03-19
feat(api_key): add scope-based permission control for API keys

JimMoen:
Release version: 5.10

## Summary

Add scope-based permission control for API keys, allowing keys to be restricted to specific API path categories based on OpenAPI tags.

### Core scope mechanism
- New module `emqx_mgmt_api_key_scopes` classifies API paths by OpenAPI tags with `persistent_term` caching
- `emqx_mgmt_auth` extended: `create/update/authorize` flow supports scopes
- Three-state semantics: key without scopes = full access (backward compatible); `[]` = deny all scoped paths; list = allow only listed scopes
- Bootstrap file parsing extended to support optional scopes field (`key:secret:role:scopes`)
- New endpoint `GET /api_key/scopes` lists available scopes

### Preset scope groups (display-layer aliases)
- 8 preset groups bundle ~55 OpenAPI tags: `connections`, `data_integration`, `access_control`, `gateways`, `monitoring`, `system`, `security`, `extensions`
- `expand_groups/1` resolves group names to individual tag lists
- Groups are pure UI convenience — API key `scopes` field always stores individual tag names
- `GET /api_key/scopes` returns both groups and scopes for Dashboard consumption

### Key files
- `apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl` — scope definitions, cache, preset groups
- `apps/emqx_management/src/emqx_mgmt_auth.erl` — authorize with scopes, bootstrap parsing
- `apps/emqx_management/src/emqx_mgmt_api_api_keys.erl` — API schema and endpoints

### Testing
- `emqx_mgmt_api_key_scopes_SUITE` with 18 test cases: unit (cache, path matching, validate), integration (authorize with/without/empty scopes), API (CRUD with scopes), and preset group coverage test that fails CI when any OpenAPI tag is uncovered.

## PR Checklist
- [x] For internal contributor: there is a jira ticket to track this change
- [x] The changes are covered with new or existing tests
- [ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [ ] Schema changes are backward compatible or intentionally breaking

Schema changes: backward compatible — new optional `scopes` field in API key, defaults to undefined (full access).

PR #17273, opened 2026-05-15
fix(dashboard): scope-classify pre-login and api-key meta endpoints

zmstone:
Release version: 5.10.4

## Summary

The collector in `emqx_mgmt_api_key_scopes` emits a `path_missing_from_scopes_map` warning for any path returned by `paths/0` that is missing from the same module's `scopes/0` map. Seven endpoints intentionally relied on the fail-open fallback for unmapped paths and produced this warning on every boot:

- `emqx_dashboard_api`: `/login`, `/logout`, `/user_scopes`
- `emqx_dashboard_sso_api`: `/sso/running`, `/sso/login/:backend`, `/sso/token_exchange`
- `emqx_mgmt_api_api_keys`: `/api_key_scopes`

A new `?SCOPE_PUBLIC` sentinel lets each module explicitly declare a path as intentionally unscoped. The collector recognises the sentinel, preserves the existing fail-open runtime semantics (the path stays out of the cache), and skips the warning. Paths genuinely forgotten from `scopes/0` still warn.

The pre-existing `t_all_endpoints_covered_by_scopes` allow-list was a hardcoded duplicate of the same intent; it now derives its public allow-list from each module's `scopes/0`. A new `t_init_cache_no_missing_path_warnings` case asserts every declared path is in its module's scope map (real scope or `?SCOPE_PUBLIC`), so the next added path will fail the test instead of silently producing a boot warning.

Relevant files: `apps/emqx_utils/include/emqx_api_key_scopes.hrl`, `apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl`, the three API modules above, and the existing CT suite `apps/emqx_management/test/emqx_mgmt_api_key_scopes_SUITE.erl`.

## PR Checklist


apps/emqx_utils/include/emqx_api_key_scopes.hrl
```diff
@@ -30,6 +30,16 @@
...
 
+%% Sentinel for paths that are intentionally unscoped: pre-login entry
+%% points (/login, /sso/login/:backend, ...) and meta endpoints that
+%% only return static catalog data (/user_scopes, /api_key_scopes).
+%% Modules using the map form of scopes/0 must declare such paths with
+%% this value rather than omitting them, so that genuinely forgotten
+%% paths still produce a startup warning. The collector treats this
+%% value as: do not insert into the runtime cache (preserves fail-open
+%% semantics) and do not emit path_missing_from_scopes_map.
+-define(SCOPE_PUBLIC, public).
```
JimMoen: Naming/type inconsistency: `?SCOPE_DENIED` is a binary (`<<"$denied">>`), and every user-visible scope macro elsewhere in this file (and in the surrounding `?SCOPE_CONNECTIONS`, `?SCOPE_DATA_INTEGRATION`, ...) is also a binary. `?SCOPE_PUBLIC` breaks that convention by being an atom (`public`).

Two practical concerns:

1. **Subtle silent bugs if ever flowed into a runtime comparison.** Today the collector is the only consumer and matches the sentinel explicitly, so the type difference is invisible. But if a future change ever puts `?SCOPE_PUBLIC` next to other scopes in a `lists:member(Scope, KeyScopes)` or similar (where `KeyScopes` are binaries), the atom-vs-binary comparison would silently always be `false` — easy to miss, no compile-time signal.

2. **Reader load.** The hrl already establishes a strong "scope name = binary, prefixed `$` if internal" convention (`?SCOPE_DENIED` = `<<"$denied">>`). A reader scanning the file will assume `?SCOPE_PUBLIC` follows the same shape until they reach the definition.

Two ways out:

- **Preferred:** make it a binary in the same `$`-prefixed style: `-define(SCOPE_PUBLIC, <<"$public">>).` — keeps type uniformity across the whole macro family, and the collector's `?SCOPE_PUBLIC -> InnerAcc;` clause keeps working unchanged.
- **Alternative:** if the atom is intentional (e.g. to make it impossible to express via the JSON API which only carries binaries), call that out explicitly in the docstring above — something like "Atom rather than binary so it cannot be expressed in a JSON payload and is strictly a compile-time sentinel." That defuses the surprise on first read and documents the design intent for future contributors.

Either is fine — the current code works — but the type inconsistency on its own is worth a one-line answer in the comment block.

JimMoen:
## Follow-up question on the two-layer guard design

This PR strengthens the CT-side guard significantly: `t_all_endpoints_covered_by_scopes` now derives its public allow-list from `scopes/0` instead of a hardcoded list, and the new `t_init_cache_no_missing_path_warnings` enforces per-path coverage on every module that returns a map from `scopes/0`. Both excellent.

However, the **pre-compile shell guard** `scripts/check-api-scopes.sh` was not touched. After reading it end-to-end (release-510, blob `7cf1f3be7d`, unchanged since `b69a7eaf12` introduced it in PR #16942), it does **only** an existence check — it greps each `minirest_api` module for an `-export([... scopes/0 ...])` or a `scopes() ->` definition, and that's it. It cannot detect that the 7 endpoints this PR fixes were missing from their module's `scopes/0` map, because all three owning modules **do** export `scopes/0`. The script returns OK regardless of map content.

That leaves the project with this layering today:

| Guard | Stage | Catches |
|---|---|---|
| `check-api-scopes.sh` | pre-compile (`_pr_entrypoint.yaml:115-117`) | module has no `scopes/0` at all |
| `t_all_endpoints_covered_by_scopes` | CT (post-compile, slow) | path in `paths/0` not covered by aggregated `scopes/0` map |
| `t_init_cache_no_missing_path_warnings` (new) | CT (post-compile, slow) | same as above, per-module |

In practice this means: if I open a PR that adds a new endpoint to `paths/0` but forgets to add it to `scopes/0`, **the pre-compile shell guard does not warn me**. I find out only when the CT job runs, often 5-10 minutes into CI. With `?SCOPE_PUBLIC` introduced here, the cost of "remembering to map every path" goes up — every new pre-login or meta endpoint now has to be explicitly declared `?SCOPE_PUBLIC`, otherwise CT fails late.

A few options to close the gap, ranked by my preference:

1. **Promote `t_init_cache_no_missing_path_warnings` into a standalone escript and run it as an early CI step** (after compile, before CT). It can directly call `Mod:paths()` / `Mod:scopes()` on the compiled beams, give a clean "module X path Y missing scope" error, and fail fast. EMQX already has `scripts/gen-api-scopes-doc.escript` doing similar beam-level introspection — the coverage check could live next to it in ~10 lines. This is the cheapest path to fast feedback without duplicating logic.

2. **Document the layering explicitly** in the script's header comment, e.g.:
   ```
   ## NOTE: This is an EXISTENCE check only. Per-path coverage of scopes/0
   ## is verified by t_init_cache_no_missing_path_warnings in
   ## emqx_mgmt_api_key_scopes_SUITE — that test is the source of truth for
   ## "every path in paths/0 has a scope".
   ```
   So future contributors don't expect this script to catch the case this PR fixes.

3. **Leave it as is** — accept that the shell guard is just a coarse sanity check and CT is the real gate. This is fine if you're OK with the 5-10 minute feedback delay on missing-scope mistakes.

Strengthening `check-api-scopes.sh` itself to do per-path checking is **not** in this list — statically parsing Erlang map literals (especially with hrl macros expanded) from shell is fragile, and you already have a perfectly good escript runner for this kind of introspection.

Not blocking on this PR — happy to file a follow-up issue if helpful. Wanted to surface it here because PR #16942 introduced the shell guard with the intent of catching missing scopes, and this PR is the natural moment to look at whether that intent has been fully realised.

zmstone:
Thanks @JimMoen -- you're right that the shell guard couldn't catch this case and CT-only feedback is too slow. Went with your option 1.

`scripts/check-api-scope-coverage.escript` does the beam-level coverage check the way you described (mirrors `gen-api-scopes-doc.escript`). It's wired into the compile job right after `make ${PROFILE}-compile test-compile`, so a missing path now fails CI within seconds of compile finishing instead of mid-CT.

`scripts/check-api-scopes.sh` keeps its existence-only role at the sanity-check stage; added a header comment pointing to the new escript + the CT case so future contributors don't expect this script to do more than it does.

Verified the escript by temporarily removing two `?SCOPE_PUBLIC` entries from `emqx_dashboard_api` and recompiling -- the escript correctly reported:
```
ERROR: 2 path(s) missing from scopes/0 map (would warn at boot):
  emqx_dashboard_api -> /login
  emqx_dashboard_api -> /logout
```
and exited 1. Clean run after restore.

JimMoen (approved):
LGTM.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14400)

PR #17276, opened 2026-05-15
build(docker): harden runtime image and ship static curl

zmstone:
Fixes <issue-or-jira-number>

Release version: 5.8.10

## Summary

Port of #17271 to release-58. Hardens the runtime stage of the official EMQX docker image so it passes customer image-scanner checks:

- `apt-get upgrade` is run in the runtime stage so the image picks up the latest Debian security patches.
- `libgnutls30t64` is purged. EMQX talks TLS via OpenSSL through Erlang/OTP and never links GnuTLS, so it was only present as a transitive dependency of `curl` and showed up as a scanner finding. The removal is guarded with `dpkg-query` so a future base image that doesn't ship the package will not break the build.
- `curl` is replaced by a pinned, SHA256-verified statically-linked binary from [stunnel/static-curl](https://github.com/stunnel/static-curl) 8.20.0 (OpenSSL build, HTTP/2 + HTTP/3, no librtmp, no GnuTLS). The fetch happens in a discardable `curl_static` build stage based on `BUILD_FROM` because the emqx-builder image already ships `curl`, `xz` and `tar`; only the verified binary is copied into the runtime image. Container healthchecks that call `curl` continue to work unchanged.

Relevant file: `deploy/docker/Dockerfile`.

Verified locally by building `emqx-enterprise:5.8.10-ge1e5f53c` for amd64, booting it, running `debsecan --suite trixie` against installed packages, and confirming `curl http://localhost:18083/api/v5/status` returns HTTP 200 from inside the container.

## PR Checklist

### DIFF
--- a/changes/ee/fix-17276.en.md
+++ b/changes/ee/fix-17276.en.md
@@ -1,0 +1,12 @@
+Hardened the official EMQX docker image to clear image-scanner findings:
+
+- Applied Debian security upgrades during the runtime image build, so the
+  image picks up the latest patched `libssl3t64`.
+- Removed the unused `libgnutls30t64` package. EMQX talks TLS via OpenSSL
+  through Erlang/OTP and never links GnuTLS, so it was only present as a
+  transitive dependency of `curl` and showed up in scanner reports.
+- Replaced the Debian `curl` package — which would have transitively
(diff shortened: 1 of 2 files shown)


emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14387)

PR #17285, opened 2026-05-17
refactor(authn): resolve provider optional-callback dispatch at registration

zmstone:
Each authenticator now caches its provider's optional-callback set in the #authenticator{} record when the authenticator is created. emqx_authn_chains:call_authenticator/4 consults that map at dispatch time instead of calling erlang:function_exported/3 on every client login. The provider is force-loaded once during probing so the behaviour is correct under -mode interactive as well as -mode embedded.

### DIFF
--- a/apps/emqx_auth/src/emqx_authn/emqx_authn.erl
+++ b/apps/emqx_auth/src/emqx_authn/emqx_authn.erl
@@ -12,6 +12,7 @@
     get_enabled_authns/0,
 
     register_provider/2,
+    register_provider/3,
     deregister_provider/1
 ]).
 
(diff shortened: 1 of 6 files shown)


savonarola:
The callback modules are our own, so maybe we could simplify:
* Extend `emqx_authn_chains:register_provider(Type, Module)` to `emqx_authn_chains:register_provider(Type, Module, Opts)` with optional `Opts :: #{methods := [{atom(), arity()}]}` or even just `Opts :: #{support_user_operations := boolean()}`.
* When creating authenticator, populate `#authenticator.support_user_operations` from provider info.
* In `call_authenticator` just check the flag.

So authn backends just inform themselves whether they support the operations on users.

zmstone:
Adopted your simpler suggestion — providers now declare `support_user_operations` at registration time via `register_provider/3`, and `call_authenticator/4` just checks the flag. The `function_exported` probe is gone entirely.

`emqx_authn_mnesia` and `emqx_authn_scram_mnesia` are the only two providers that flag `support_user_operations => true`. SCRAM gains a stub `import_users/2` returning `{error, unsupported_operation}` because SCRAM credentials genuinely can't be bulk-imported (no plaintext password to derive salts from).

Also rebased to `release-63` and changed the PR base accordingly — this is paired with the upcoming `-mode interactive` default which only ships in 6.3.

savonarola (dismissed):

savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16062)

PR #17304, opened 2026-05-19
test(bridge): widen recovery retry budget for slow TLS bridge variants

zmstone:
## Summary

`emqx_bridge_v2_testlib:t_on_get_status/2` waits up to 200ms × 100 = 20s for the resource to come back up after the toxiproxy is restored. The `arrow_tls` group of `emqx_bridge_datalayers_arrow_SUITE` occasionally exceeds that budget — TLS handshake plus arrow-flight connection setup pushes recovery past 20s under load, while plain TCP and HTTP variants of the same test pass.

Bumping the budget to 200 attempts (40s) in both retries inside `t_on_get_status` (initial stabilization + post-recovery). `?retry` exits on first success, so this is free for fast variants.

No production code changes, no user-facing behavior changes.

### DIFF
--- a/apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_SUITE.erl
+++ b/apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_SUITE.erl
@@ -331,7 +331,15 @@ t_create_via_http(Config) ->
     ok.
 
 t_on_get_status(Config) ->
-    emqx_bridge_v2_testlib:t_on_get_status(Config),
+    %% Shorten health_check_interval so recovery fits the 20s retry budget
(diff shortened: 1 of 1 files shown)


thalesmg:
IMO, I think it's simpler/better to adjust the problematic suite instead to use shorter health check intervals (e.g.: https://github.com/emqx/emqx/blob/d5259fdedd6d30dd557cfc06e6d749fb50992c6c/apps/emqx_bridge/test/emqx_bridge_v2_testlib.erl#L2607-L2632 ), which would tend to make other test cases a bit faster as a side effect, instead of changing the shared test case.

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14614)

PR #17306, opened 2026-05-19
fix(backup): skip read-only roots in import schema check

zmstone:
Fixes <issue-or-jira-number>

Release version: 5.8.11

## Summary

The cluster configuration import was failing with a `required_field: node.cookie` schema-check error when the exported `cluster.hocon` contained any partial sub-tree of the `node` root (for example a single overridden `node.dist_listen_options`).

`emqx_mgmt_data_backup:validate_cluster_hocon/1` does a shallow `maps:merge(emqx:get_raw_config([]), RawConf1)` between the running raw config and the imported cluster.hocon. A partial `node` map in the imported file replaces the entire running `node` map, dropping required-no-default fields like `node.cookie`.

Read-only roots (`node`, `rpc`) are not applied by the importer anyway. They are now dropped from the imported config before the merge so the schema check sees the running node's own values for those roots.

A new eunit `import_skip_root_keys_test/0` guards that the skip-list stays disjoint from `?CONF_KEYS`. A new CT test `t_import_cluster_hocon_partial_node` exercises this path and reproduces the original failure without the fix.

Targeted at `release-58` so it can be propagated forward via the usual sync PRs.

## PR Checklist


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14490)

PR #17309, opened 2026-05-19
Sanitize PROXY-Protocol v2 cert TLVs before they enter client identity

zmstone:
Release version:
6.0.3, 6.1.2, 6.2.1

## Summary

When a listener has `proxy_protocol = true`, the SSL_CN / SSL_SUBJECT sub-TLVs of an incoming PROXY-Protocol v2 frame are attacker-controlled bytes. Those bytes were copied verbatim into `ClientInfo#{cn, dn}` and surfaced as the `${cert_common_name}` / `${cert_subject}` template variables.

An operator who follows the documented pattern of forwarding mTLS identity from an upstream L4 load balancer (TLS terminated upstream, identity carried in PP2) and uses `${cert_common_name}` or `${cert_subject}` in an `emqx_auth_http` / `emqx_authz_http` header template was exposed to CRLF injection — a CR/LF in the TLV split the outgoing HTTP request to the auth backend and let an attacker smuggle headers that could coerce an `allow` decision.

This is the same byte class EMQX already rejects on MQTT-ingested clientid / username / password via `emqx_frame:validate_utf8/1` (`strict_mode = true` by default). The PROXY-Protocol ingestion path skipped that check — asymmetric validation.

### Mitigation

1. **Primary**: validate CN / DN at the peer-cert ingestion boundary in `emqx_channel:set_peercert_infos/3`. Bytes in the ranges `0x00-0x1F` or `0x7F-0x9F` cause the connection to be rejected before any MQTT packet is parsed and a warning is logged (`peer_certificate_field_rejected`). This matches the existing MQTT-validation semantics and shields all downstream consumers — including `emqx_auth_http`, `emqx_authz_http`, and HTTP bridges/actions templated from `clientinfo.cn` — since the offending bytes never enter `ClientInfo` in the first place.

2. **Defense in depth**: `emqx_auth_http_utils:generate_request/2` now refuses to emit a request when a rendered header name or value contains CR, LF, or NUL, regardless of where the bytes came from. If a future render path forgets to sanitize, this second wall still prevents request smuggling. The check is intentionally narrow (just the bytes that split a request line) so it does not change behavior for clean templates that interpolate UTF-8 or other variables such as `${username}`.

### Files

- `apps/emqx_utils/src/emqx_utils.erl` — new `is_mqtt_safe_utf8/1` helper.
- `apps/emqx/src/emqx_channel.erl` — validate CN / DN in `set_peercert_infos/3`.
- `apps/emqx_auth/src/emqx_auth_http_utils.erl` — reject CR/LF/NUL in rendered HTTP header name/value pairs.
- Tests: `apps/emqx/test/emqx_channel_SUITE.erl` and `apps/emqx_auth/test/emqx_auth_http_utils_tests.erl`.

### Backports

The same vulnerability is present on v5 (reporter named 5.8.6) and the active v6 branches. This PR targets `release-60`. Forward backports to `release-61` / `release-62` and a back-port to the active v5 branch should follow.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -305,7 +305,10 @@ set_peercert_infos(NoSSL, ClientInfo, _) when
 ->
     ClientInfo#{username => undefined};
 set_peercert_infos(PeerCert, ClientInfo, Zone) ->
-    {DN, CN} = {esockd_peercert:subject(PeerCert), esockd_peercert:common_name(PeerCert)},
+    DN = esockd_peercert:subject(PeerCert),
+    CN = esockd_peercert:common_name(PeerCert),
+    ok = validate_peercert_string(dn, DN, PeerCert),
+    ok = validate_peercert_string(cn, CN, PeerCert),
     PeercetAs = fun(Key) ->
         case get_mqtt_conf(Zone, Key) of
             cn -> CN;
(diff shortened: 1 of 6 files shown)


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14411)

PR #17311, opened 2026-05-20
fix(docker): avoid empty node host on lookup failure

hjianbo:
Fixes emqx/emqx-customer-support#1

Release version:
6.0.3

## Summary

The Docker entrypoint previously derived the automatic node host only from `hostname -i`. In container environments where the hostname cannot be resolved, that command can return no IPv4 address, causing the entrypoint to export an invalid node name such as `emqx@`. Erlang distribution then fails during `net_kernel` startup with `nodistribution`.

This change keeps the existing `hostname -i` behavior for normal environments, falls back to the first IPv4 address reported by `hostname -I` when hostname resolution fails, and exits with a clear error if no local IP can be determined while auto-generating the node name. Explicit `EMQX_NODE_NAME`, `EMQX_NODE__NAME`, or `EMQX_HOST` settings keep their current precedence.

A lightweight pytest regression covers the fallback path and the empty-host guard, and the PR entrypoint workflow now runs that test with the existing script pytest checks.

## PR Checklist

### DIFF
--- a/.github/workflows/_pr_entrypoint.yaml
+++ b/.github/workflows/_pr_entrypoint.yaml
@@ -78,13 +78,13 @@ jobs:
         run: ./scripts/shellcheck.sh
       - name: Run shell tests
         run: scripts/shelltest/run_tests.sh
-      - name: Run pytest find-apps tests
+      - name: Run pytest script tests
         run: |
           python3 -m venv venv
           source venv/bin/activate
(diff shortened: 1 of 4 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14412)

PR #17313, opened 2026-05-20
Fix cluster sync raw config diff noise

hjianbo:
Fixes https://emqx.atlassian.net/browse/EMQX-15291

Release version: 5.10

### Problem details

`conf cluster_sync status` could report many inconsistent keys even when all nodes had the same effective running configuration and the same cluster transaction id.

The false positives were caused by comparing raw configuration maps directly. Raw configuration in EMQX is the unparsed, untranslated representation kept by the configuration subsystem, and it is not guaranteed to be canonical across different loading paths.

One reproducible path is:

1. Start one core node.
2. Import a backup containing cluster configuration, for example a rule with a republish action.
3. Start another core node and let it join the cluster.
4. Run `conf cluster_sync status`.

The first node applies the imported configuration through the online update/import path. Its raw config keeps the sparse user-provided representation, such as a republish action with only `topic` under `args`.

The joining node first syncs `cluster.hocon` from an existing core node and then loads it through the boot-time `emqx_config:init_load/1` path. That path fills schema defaults into the runtime raw config. As a result, the same rule can contain materialized default fields on the joining node, for example `enable`, `name`, `qos`, `retain`, `payload`, `mqtt_properties`, `user_properties`, and `direct_dispatch`.

Both raw representations translate to the same checked config, so the nodes are effectively synchronized. However, the old status command compared those raw/default representation differences as if they were real cluster-sync inconsistencies. This also affected other default/materialized roots observed in the Jira investigation, such as `actions`, `license`, and `rule_engine`.

### Fix approach

This PR changes `conf cluster_sync status` to compare normalized effective configuration instead of raw maps directly.

For each running node, the command now fetches raw config, fills defaults, and validates it through the schema to obtain the checked configuration used by EMQX at runtime. The consistency decision is made from that normalized checked config, so semantically equivalent raw representations no longer produce false inconsistent-key reports.

For diagnostic output, the command still fetches raw config for the affected root key, but it prunes the raw payload using the checked-config diff before printing. This keeps useful raw details for real changes while suppressing raw-only differences such as omitted defaults or materialized defaults.

The warning itself is still emitted whenever the checked config reports a real difference, even when the corresponding raw payload is unavailable or pruned empty, such as virtual/default entries like `zones.default`. The HOCON payload is printed only when there is meaningful raw diff to show.

The change also fixes the original crash in raw diff pruning when a nested key exists on one node but is missing on the other.

## Testing

- `BUILD_WITHOUT_QUIC=1 ERL_FLAGS='-kernel prevent_overlapping_partitions false' ./rebar3 ct --name 'test@127.0.0.1' -v --readable=true --suite apps/emqx_conf/test/emqx_conf_cluster_sync_import_SUITE.erl --case t_status_after_import_then_join_ignores_raw_only_diffs`
- `BUILD_WITHOUT_QUIC=1 BUILD_WITHOUT_ROCKSDB=1 ./rebar3 eunit --module=emqx_conf_cli`
- `git diff --check`
- `BEFORE_REF=$(git merge-base HEAD ce/release-510) AFTER_REF=HEAD ./scripts/check-changes-filename-pattern.sh`

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_cli.erl
+++ b/apps/emqx_conf/src/emqx_conf_cli.erl
@@ -852,12 +852,12 @@ print_inconsistent_conf(New = #{}, Old = #{}, Options) ->
         changed := Changed
     } = emqx_utils_maps:diff_maps(New, Old),
     RemovedFmt = "~ts(~w)'s ~s has deleted certain keys, but they are still present on ~ts(~w).~n",
-    print_inconsistent(Removed, RemovedFmt, Options),
+    print_inconsistent(Removed, RemovedFmt, removed, Options),
     AddedFmt = "~ts(~w)'s ~s has new setting, but it has not been applied to ~ts(~w).~n",
-    print_inconsistent(Added, AddedFmt, Options),
+    print_inconsistent(Added, AddedFmt, added, Options),
     ChangedFmt =
         "~ts(~w)'s ~s has been updated, but the changes have not been applied to ~ts(~w).~n",
-    print_inconsistent(Changed, ChangedFmt, Options);
+    print_inconsistent(Changed, ChangedFmt, changed, Options);
 %% authentication rewrite topic_metrics is list(not map).
 print_inconsistent_conf(New, Old, Options) ->
     #{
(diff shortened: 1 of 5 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14429)

PR #17315, opened 2026-05-20
fix: extend control-character rejection beyond peer cert CN / DN

zmstone:
Fixes <issue-or-jira-number>

Release version:
6.0.3, 6.1.2, 6.2.1

## Summary

Follow-up to #17309. That PR rejected control characters in `cn` / `dn` at the peer-cert ingestion boundary. This PR closes three sibling holes through which the same byte-class could still flow into HTTP authentication, authorization, and bridge templates.

### Changes

1. **`peersni` validated at ingestion** (`apps/emqx/src/emqx_channel.erl`). `peersni` is set in `ConnInfo` by the TCP / WebSocket / SocketConnection layer and is sourced from the TLS handshake SNI or, when PROXY-Protocol v2 is in use, from the `pp2_authority` TLV (attacker-controlled). After `take_conn_info_fields/3` moves it into `ClientInfo`, `init/2` now runs it through `emqx_utils:is_mqtt_safe_utf8/1` and exits the connection with `{shutdown, clientinfo_field_invalid}` on bad bytes — same semantics as the existing `cn` / `dn` check.

2. **`client_attrs` values vetted before being stored** (`apps/emqx/src/emqx_channel.erl`). `initialize_client_attrs/2` evaluates Variform expressions defined in `mqtt.client_attrs_init`. The rendered binaries land in `ClientInfo.client_attrs` and are then exposed to templates as `${client_attrs.<name>}` (the documented `tns` tenant-namespace pattern is one example). Values that contain control bytes are dropped with a warning, not stored — so a template like `${client_attrs.tns}` cannot carry injected bytes into an outbound HTTP header even if the Variform expression pulled bytes from a non-strict-mode MQTT v5 `user_property`. Dropping the attribute (rather than terminating the connection) avoids punishing operators who write complex Variform expressions while still neutralizing the smuggling primitive.

3. **HTTP bridge connector drops unsafe headers** (`apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl`). `render_headers/3` now consults `emqx_utils:http_header_byte_check/1` on each rendered `{name, value}` pair and drops any pair containing NUL / CR / LF. An MQTT publisher could otherwise smuggle headers into the bridge destination through a templated value such as `${payload.something}` or `${client_attrs.tns}`. Dropping the header (rather than failing the request) keeps the bridge buffer worker progressing while preventing the malicious bytes from reaching the wire.

4. **Shared helper in `emqx_utils`**. The header byte-check, previously private to `emqx_auth_http_utils`, moves to `emqx_utils:http_header_byte_check/1` so both `emqx_auth_http_utils` and `emqx_bridge_http_connector` use the same implementation.

### Tests

- `emqx_channel_SUITE`: `t_peersni_with_crlf_rejected`, `t_peersni_clean_accepted`, `t_client_attrs_with_control_chars_dropped`.
- `emqx_bridge_http_connector` inline eunit: `drop_unsafe_headers_test`.
- Full `emqx_channel_SUITE` (93/93) and inline-eunit / `emqx_auth_http_utils_tests` all green.

### Backports

Same scope as #17309: `release-61`, `release-62`, and the active v5 branch should follow up.

## PR Checklist

### DIFF
--- a/apps/emqx_utils/src/emqx_utils.erl
+++ b/apps/emqx_utils/src/emqx_utils.erl
@@ -896,21 +896,28 @@ is_mqtt_safe_utf8(<<_BadUtf8, _Rest/binary>>) ->
 
 %% @doc Scan a binary for bytes capable of splitting an HTTP/1.1 request line.
 %% Returns `ok' if the binary is safe to emit as an HTTP header name or value,
-%% otherwise `{error, contains_nul | contains_cr | contains_lf}'. Non-binary
+%% otherwise `{error, contains_null | contains_cr | contains_lf}'. Non-binary
 %% inputs are treated as safe; callers that templated something other than a
 %% binary have already failed earlier in a more visible way.
+%%
+%% Implemented with `binary:match/2' (NIF) so the happy path stays cheap; the
+%% per-byte inspection only runs when an offending byte is found.
 -spec http_header_byte_check(binary() | term()) ->
-    ok | {error, contains_nul | contains_cr | contains_lf}.
+    ok | {error, contains_null | contains_cr | contains_lf}.
 http_header_byte_check(Bin) when is_binary(Bin) ->
-    http_header_byte_check_bin(Bin);
+    case binary:match(Bin, [<<$\r>>, <<$\n>>, <<0>>]) of
+        nomatch ->
+            ok;
+        {Pos, _} ->
+            <<_:Pos/binary, Byte, _/binary>> = Bin,
+            {error, control_byte_to_reason(Byte)}
+    end;
 http_header_byte_check(_NonBin) ->
     ok.
 
-http_header_byte_check_bin(<<>>) -> ok;
-http_header_byte_check_bin(<<0, _/binary>>) -> {error, contains_nul};
-http_header_byte_check_bin(<<$\r, _/binary>>) -> {error, contains_cr};
-http_header_byte_check_bin(<<$\n, _/binary>>) -> {error, contains_lf};
-http_header_byte_check_bin(<<_, Rest/binary>>) -> http_header_byte_check_bin(Rest).
+control_byte_to_reason(0) -> contains_null;
+control_byte_to_reason($\r) -> contains_cr;
+control_byte_to_reason($\n) -> contains_lf.
 
 %% @doc Generate random, printable bytes as an ID.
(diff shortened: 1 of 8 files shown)


apps/emqx_utils/src/emqx_utils.erl
```diff
@@ -893,6 +894,24 @@ is_mqtt_safe_utf8(<<_H/utf8, Rest/binary>>) ->
...
+%% binary have already failed earlier in a more visible way.
+-spec http_header_byte_check(binary() | term()) ->
+    ok | {error, contains_nul | contains_cr | contains_lf}.
+http_header_byte_check(Bin) when is_binary(Bin) ->
+    http_header_byte_check_bin(Bin);
+http_header_byte_check(_NonBin) ->
+    ok.
+
+http_header_byte_check_bin(<<>>) -> ok;
+http_header_byte_check_bin(<<0, _/binary>>) -> {error, contains_nul};
```
ieQu1:
```suggestion
http_header_byte_check_bin(<<0, _/binary>>) -> {error, contains_null};
```

apps/emqx_utils/src/emqx_utils.erl
```diff
@@ -893,6 +894,24 @@ is_mqtt_safe_utf8(<<_H/utf8, Rest/binary>>) ->
...
+%% inputs are treated as safe; callers that templated something other than a
+%% binary have already failed earlier in a more visible way.
+-spec http_header_byte_check(binary() | term()) ->
+    ok | {error, contains_nul | contains_cr | contains_lf}.
+http_header_byte_check(Bin) when is_binary(Bin) ->
+    http_header_byte_check_bin(Bin);
+http_header_byte_check(_NonBin) ->
+    ok.
+
+http_header_byte_check_bin(<<>>) -> ok;
```
ieQu1: Perhaps something like
binary:match(Subject, [<<$\r>>, <<$\0>>>, <<$\n>>])

will be faster in the happy case, since it's a nif?
~It will do some extra work after finding a forbidden character, but we probably care less about that case.~

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14417)

PR #17319, opened 2026-05-20
fix(dashboard): set Content-Type to application/json on GET /api/v5/schemas/:name

zmstone:
Fixes <issue-or-jira-number>

## Summary

`GET /api/v5/schemas/{hotconf,actions,connectors}` returned a JSON body but advertised `Content-Type: text/plain; charset=utf-8`, breaking clients that dispatch on the response content type.

The handler pre-encodes the body to a JSON binary. minirest's binary encoder sets no response headers, so cowboy falls back to `text/plain`. Switched the handler return to the `{Status, Headers, Body}` 3-tuple form with an explicit `application/json` header, matching the existing pattern in `emqx_mgmt_api_clients:format_msgs_resp/4` (see comment there: "Make sure minirest won't set another content-type for self-encoded JSON response body").

All three sibling routes (`hotconf`, `actions`, `connectors`) share the same handler, so the fix applies to all three. Updated `apps/emqx_dashboard/test/emqx_dashboard_schema_api_SUITE.erl` to assert the response Content-Type header for each route.

Relevant files:
- `apps/emqx_dashboard/src/emqx_dashboard_schema_api.erl`
- `apps/emqx_dashboard/test/emqx_dashboard_schema_api_SUITE.erl`

## PR Checklist

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_schema_api.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_schema_api.erl
@@ -70,7 +70,10 @@ schema("/schemas/:name") ->
 get_schema(get, #{
     bindings := #{name := Name}
 }) ->
-    {200, gen_schema(Name)};
+    %% The body is already a JSON binary; tell minirest the content type
+    %% explicitly so it doesn't fall back to text/plain.
+    Headers = #{<<"content-type">> => <<"application/json">>},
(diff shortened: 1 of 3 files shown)


JimMoen (approved):
LGTM

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14414)

PR #17322, opened 2026-05-20
fix: extend control-character rejection beyond peer cert CN / DN

zmstone:
Release version:
5.8.11, 5.10.4

## Summary

Follow-up to #17314 on `release-58` (port of upstream #17315 from `release-60`). That earlier PR rejected control characters in `cn` / `dn` at the peer-cert ingestion boundary. This PR closes three sibling holes through which the same byte class could still flow into HTTP authentication, authorization, and bridge templates.

### Changes

1. **`peersni` validated at ingestion** (`apps/emqx/src/emqx_channel.erl`). `peersni` is set in `ConnInfo` by the TCP / WebSocket / connection layer and is sourced from the TLS handshake SNI or, when PROXY-Protocol v2 is in use, from the `pp2_authority` TLV (attacker-controlled). After `take_conn_info_fields/3` moves it into `ClientInfo`, `init/2` now runs it through `emqx_utils:is_mqtt_safe_utf8/1` and exits the connection with `{shutdown, clientinfo_field_invalid}` on bad bytes — same semantics as the existing `cn` / `dn` check.

2. **`client_attrs` values vetted before being stored** (`apps/emqx/src/emqx_channel.erl`). `initialize_client_attrs/2` evaluates Variform expressions defined in `mqtt.client_attrs_init`. The rendered binaries land in `ClientInfo.client_attrs` and are then exposed to templates as `${client_attrs.<name>}` (the documented `tns` tenant-namespace pattern is one example). Values that contain control bytes are dropped with a warning, not stored — so a template like `${client_attrs.tns}` cannot carry injected bytes into an outbound HTTP header even if the Variform expression pulled bytes from a non-strict-mode MQTT v5 `user_property`. Dropping the attribute (rather than terminating the connection) avoids punishing operators who write complex Variform expressions while still neutralizing the smuggling primitive.

3. **HTTP bridge connector drops unsafe headers** (`apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl`). `render_headers/3` now consults `emqx_utils:http_header_byte_check/1` on each rendered `{name, value}` pair and drops any pair containing NUL / CR / LF. An MQTT publisher could otherwise smuggle headers into the bridge destination through a templated value such as `${payload.something}` or `${client_attrs.tns}`. Dropping the header (rather than failing the request) keeps the bridge buffer worker progressing while preventing the malicious bytes from reaching the wire.

4. **Shared helper in `emqx_utils`**. The header byte-check, previously private to `emqx_auth_utils` (the 5.x location; called `emqx_auth_http_utils` on 6.x), moves to `emqx_utils:http_header_byte_check/1` so both `emqx_auth_utils` and `emqx_bridge_http_connector` use the same implementation. The happy-path scan is backed by `binary:match/2` (NIF). The NUL reason atom is renamed `contains_nul` -> `contains_null` for spelling consistency.

### Tests

- `emqx_channel_SUITE`: `t_peersni_with_crlf_rejected`, `t_peersni_clean_accepted`, `t_client_attrs_with_control_chars_dropped` (all pass).
- `emqx_bridge_http_connector` inline eunit: `drop_unsafe_headers_test`.
- `emqx_auth_utils_tests` updated for the renamed reason atom.

### Backports

Stacked port of upstream #17315 onto `release-58`; same scope as the prior #17314 port. The v5 file layout differs from v6 — the HTTP request generator and its header validator live in `emqx_auth_utils` (v5) rather than `emqx_auth_http_utils` (v6) — so the shared-helper switchover applied there.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
+++ b/apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
@@ -831,15 +831,39 @@ render_request_body(BodyTks, RenderTmplFunc, Msg) ->
     RenderTmplFunc(BodyTks, Msg).
 
 render_headers(HeaderTks, RenderTmplFunc, Msg) ->
-    lists:map(
-        fun({K, V}) ->
-            {
-                render_template_string(K, RenderTmplFunc, Msg),
-                render_template_string(emqx_secret:unwrap(V), RenderTmplFunc, Msg)
-            }
+    lists:filtermap(
+        fun({KTks, VTks}) ->
+            K = render_template_string(KTks, RenderTmplFunc, Msg),
+            V = render_template_string(emqx_secret:unwrap(VTks), RenderTmplFunc, Msg),
+            case safe_http_header(K, V) of
+                true ->
+                    {true, {K, V}};
+                {false, Reason} ->
+                    ?SLOG(warning, #{
+                        msg => "http_header_dropped",
+                        header_name => K,
+                        reason => Reason
+                    }),
+                    false
+            end
         end,
         HeaderTks
     ).
+
+%% Defense-in-depth: a templated header value can interpolate message-level
+%% bytes that originated from an MQTT publisher. Reject CR/LF/NUL so a
+%% malicious payload cannot split the outgoing HTTP request to the bridge
+%% destination.
+safe_http_header(K, V) ->
+    case emqx_utils:http_header_byte_check(K) of
+        ok ->
+            case emqx_utils:http_header_byte_check(V) of
+                ok -> true;
+                {error, Reason} -> {false, Reason}
+            end;
+        {error, Reason} ->
+            {false, Reason}
+    end.
 
 render_template(Template, Msg) ->
     % NOTE: ignoring errors here, missing variables will be rendered as `"undefined"`.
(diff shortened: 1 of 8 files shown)


apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl
```diff
@@ -831,16 +831,40 @@ render_request_body(BodyTks, RenderTmplFunc, Msg) ->
...
+    lists:filtermap(
+        fun({KTks, VTks}) ->
+            K = render_template_string(KTks, RenderTmplFunc, Msg),
+            V = render_template_string(emqx_secret:unwrap(VTks), RenderTmplFunc, Msg),
+            case safe_http_header(K, V) of
+                true ->
+                    {true, {K, V}};
+                {false, Reason} ->
+                    ?SLOG(warning, #{
+                        msg => "http_header_dropped",
```
thalesmg, line 843: nit: maybe add the action res id to help track down in which action it happened?

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14435)

PR #17325, opened 2026-05-20
breaking(relup): use official release tar.gz package as upgrade target

zmstone:
Release version:
6.0.3, 6.1.2, 6.2.1

## Summary

Plugin sync (verbatim from release-510, except mix.exs):
  - source modules (handler, main, cli, app, libs, file_utils, utils, sup)
  - README.md, VERSION (-> 1.0.0)
  - new priv/relup/README.md
  - new test suites (handler + round-trip) and fixture script
  - removed priv/config.hocon (dummy config, unused)
  - mix.exs: drop 'hidden: true' to match upstream's plugrel metadata

REST API removed (#17244 design change):
  - delete apps/emqx_management/src/emqx_mgmt_api_relup.erl
  - delete apps/emqx_management/src/proto/emqx_mgmt_api_relup_proto_v1.erl
  - drop {emqx_mgmt_api_relup,1} from bpapi.versions Hot-upgrade is now driven exclusively through 'emqx ctl relup' on each node. The target tarball lives anywhere readable; a .sha256 sidecar is required; target version is read from the tarball's releases/emqx_vars.

Other touch points:
  - bin/emqx: marker file <RELUP_PATH>/version -> <RELUP_PATH>/current
  - deb/rpm packaging: ship empty relup/.keep directory next to plugins/

## PR Checklist


plugins/emqx_relup/priv/relup/README.md
```diff
@@ -0,0 +1,134 @@
...
+        {restart_application, emqx_something}
+    ],
+    post_upgrade_callbacks => [
+        pr_NNNNN_some_callback,
+        {pr_NNNNN_callback_with_args, ["5.10.5", ChildSpec]}
+    ]
+}.
+```
+
+Both `from_version` and `target_version` are strings (not binaries).
```
thalesmg, line 29: Q: wouldn't it be better to use tuple (or other structured) versions?  That way, they would be directly comparable for ordering (e.g.: we could do `when FromVsn =< {5, 10, 3}` and similar) instead of having to parse the versions every time to safely compare them.

plugins/emqx_relup/priv/relup/README.md
```diff
@@ -0,0 +1,134 @@
...
+| Instruction | Effect |
+|---|---|
+| `{load_module, Mod}` | Replaces the running beam for `Mod`. Skipped if the new beam's md5 matches what's already loaded. Equivalent to `code:soft_purge/1` + `code:load_binary/3`. |
+| `{update, Mod, {advanced, Extra}}` | OTP-style stateful upgrade for processes whose callback module is `Mod`: `sys:suspend` → `load_module` → `sys:change_code(_, Mod, FromVsn, Extra)` → `sys:resume`. Use when a `gen_server` (or other `sys`-aware process) holds state that needs migrating. |
+| `{restart_application, App}` | `application:stop` → purge & delete every module of `App` → `application:ensure_all_started`. Use when stateless restart is acceptable and many modules changed. |
+
+`{apply, _}` is **not** supported. Use `post_upgrade_callbacks`
+instead — see below.
+
+The handler injects `code:add_patha(<RootDir>/data/patches)` after
```
thalesmg: Q: isn't this path already prepended in the [`vm.args`](https://github.com/emqx/emqx/blob/617f1dfde757794e6f0fab77f82770d2f85c2b7a/apps/emqx/etc/vm.args.cloud#L141)?

plugins/emqx_relup/priv/relup/README.md
```diff
@@ -0,0 +1,134 @@
...
+Callbacks must be **idempotent**. The operator may retry a failed
+hop, and a callback that already ran on the previous attempt may
+run again. Guard against double-effects (re-creating a child spec
+that's already running, re-issuing a migration that's already
+applied, etc.).
+
+The reference example is `pr_16802_restart_rabbitmq_connectors/1`
+in `apps/emqx/src/emqx_post_upgrade.erl`: it filters to running
+RabbitMQ connectors before restarting them, so a re-run on a
+healthy cluster is a no-op.
```
thalesmg: well, not exactly a no-op...  it'll restart those again.

plugins/emqx_relup/src/emqx_relup_cli.erl
```diff
@@ -1,17 +1,78 @@
...
+    case emqx_relup_main:upgrade(TarballPath) of
         ok ->
-            emqx_ctl:print("upgraded to ~s successfully", [TargetVsn]);
+            emqx_ctl:print("upgrade complete~n");
         {error, Reason} ->
-            emqx_ctl:print("upgrade failed, reason: ~p", [Reason]);
+            emqx_ctl:print("upgrade failed, reason: ~p~n", [Reason]);
         {error_vm_restarted, Reason} ->
-            emqx_ctl:print("upgrade failed, emqx restarted, reason: ~p", [Reason])
+            emqx_ctl:print("upgrade failed, emqx restarted, reason: ~p~n", [Reason])
```
thalesmg: Q: is there time to print this if the vm has restarted?

plugins/emqx_relup/src/emqx_relup_main.erl
```diff
@@ -94,21 +94,13 @@ init([]) ->
...
-    TargetVsn = emqx_relup_handler:get_target_vsn(),
-    Key = log_upgrade_started(CurrVsn, TargetVsn, Opts),
-    Result = do_upgrade(CurrVsn, TargetVsn, RootDir, Opts),
-    ok = log_upgrade_result(Key, Result),
+    %% target_vsn isn't known until check_and_unpack parses
+    %% releases/emqx_vars from the extracted tarball; do_upgrade/3
+    %% surfaces it so the log row can be stamped.
+    Key = log_upgrade_started(CurrVsn, Opts),
+    {Result, TargetVsn} = do_upgrade(CurrVsn, RootDir, Opts),
+    ok = log_upgrade_result(Key, Result, TargetVsn),
```
thalesmg, line 107: Q: since `do_perform_and_permanent`'s `catch` restarts the vm by calling `init:restart`, perhaps we should delay calling that until after we `log_upgrade_result` followed by a `mnesia:sync_log/0` to ensure the log is persisted?

plugins/emqx_relup/src/emqx_relup_main.erl
```diff
@@ -183,16 +179,33 @@ get_all_upgrade_logs() ->
...
+            {hot_upgraded, TargetVsn};
+        none ->
+            case ets:last(emqx_relup_log) of
+                '$end_of_table' ->
+                    idle;
+                Key ->
+                    case ets:lookup(emqx_relup_log, Key) of
+                        [#emqx_relup_log{status = finished}] -> idle;
+                        [#emqx_relup_log{status = 'in-progress'}] -> 'in-progress'
+                    end
```
thalesmg: micro suggestion:


```suggestion
            case ets:last_lookup(emqx_relup_log) of
                '$end_of_table' ->
                    idle;
                {_, #emqx_relup_log{status = finished}} ->
                    idle;
                {_, #emqx_relup_log{status = 'in-progress'}} ->
                    in-progress'
           end
```

plugins/emqx_relup/src/emqx_relup_main.erl
```diff
@@ -183,16 +179,33 @@ get_all_upgrade_logs() ->
...
             end
     end.
 
+read_current_marker() ->
+    Path = filename:join([code:root_dir(), "relup", "current"]),
+    case file:read_file(Path) of
+        {ok, Bin} -> {ok, string:trim(Bin)};
```
thalesmg: Q: if versions are strings by convention, shouldn't we coerce the binary here to a string to keep it consistent?

plugins/emqx_relup/priv/relup/README.md
zmstone: from and target are always manually crafted, from must match the current running version and target must match the version string in the extracted release file. 
there is no parsing.

plugins/emqx_relup/src/emqx_relup_cli.erl
zmstone: init:restart(), returns immediately without block-wait.

plugins/emqx_relup/priv/relup/README.md
zmstone: will fix.

plugins/emqx_relup/src/emqx_relup_main.erl
zmstone: good point.

plugins/emqx_relup/src/emqx_relup_handler.erl
```diff
@@ -14,43 +14,39 @@
...
+
+check_and_unpack(CurrVsn, RootDir, #{tarball := TarFile} = Opts) ->
     try
+        {ok, UnpackDir} = unpack_release(TarFile),
+        TargetVsn = read_rel_vsn(UnpackDir),
+        ok = assert_no_duplicate_pending(RootDir, TargetVsn),
         ok = assert_not_same_vsn(CurrVsn, TargetVsn),
-        {ok, UnpackDir} = unpack_release(TargetVsn),
         ok = check_write_permission(RootDir),
         ok = check_otp_comaptibility(CurrVsn, RootDir, UnpackDir, TargetVsn),
```
thalesmg: unrelated to changes herein, but:


```suggestion
        ok = check_otp_compatibility(CurrVsn, RootDir, UnpackDir, TargetVsn),
```

plugins/emqx_relup/test/emqx_relup_round_trip_SUITE.erl
```diff
@@ -0,0 +1,411 @@
...
+    cleanup_test_catalog_entries(),
+    ok.
+
+%%==============================================================================
+%% happy path
+%%==============================================================================
+-doc "Forge a fake target tarball + sha256 sidecar + a no-op .relup "
+"catalog entry; drive the three-stage handler API and assert the "
+"target version is read out of `releases/emqx_vars`, the version "
+"marker file is written, and the deploy dir is populated.".
```
thalesmg: nit: use triple-quotes for multi-line docs.

plugins/emqx_relup/src/emqx_relup_cli.erl
thalesmg: so the message _may_ have time to be printed, or maybe it isn't, if the node shuts down fast enough.

plugins/emqx_relup/priv/relup/README.md
thalesmg: I know there is no parsing.  What I suggested was exactly to parse it before passing it down to callbacks, because it's better to deal with structured data.

zmstone: there is no order compare in relup plugin.
for test, target_version can be version strings like `6.1.2-rc.4-ga1b2c3d4e`, hard to make it structured, and not always ordered.

plugins/emqx_relup/src/emqx_relup_cli.erl
zmstone: fixed.

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14612)

PR #17336, opened 2026-05-21
build(refactor): move `parquer` dep to root `mix.exs`

thalesmg:
Even though `parquer` is solely used by the s3tables bridge in 6.0.x, it's used by more apps in later versions.  Syncing `release-60` to `release-61` could fail to correctly update its version at the root `mix.exs` file.  Therefore, we move it to be a `common_dep` clause in the root mix file to facilitate syncing between branches.


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14485)

PR #17342, opened 2026-05-21
fix(backup): skip read-only roots in import schema check (r60)

zmstone:
Fixes <issue-or-jira-number>

Release version: 6.0.3

## Summary

Port of #17306 to `release-60`. Same bug, same fix.

`emqx_mgmt_data_backup:validate_cluster_hocon/1` does a shallow `maps:merge` between the running raw config and the imported cluster.hocon. A partial `node` map in the imported file (for example a single overridden `node.dist_listen_options`) replaces the entire running `node` map, dropping required-no-default fields like `node.cookie`, and the import fails with a `required_field: node.cookie` schema-check error.

Read-only roots (`node`, `rpc`) are not applied by the importer anyway (they are not in `?CONF_KEYS` and no `emqx_config_backup` handler covers them). They are now dropped from the imported config before the merge so the schema check uses the running node's own values for those roots.

- New eunit `import_skip_root_keys_test/0` guards that the skip list stays disjoint from `?CONF_KEYS`.
- New CT `t_import_cluster_hocon_partial_node` reproduces the original failure without the fix.

## PR Checklist


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14486)

PR #17343, opened 2026-05-21
fix(authz): keep `rules` inline when replicating file source

zmstone:

## Summary

introduced in https://github.com/emqx/emqx/pull/16657

Importing a backup that contains a file-type authz source with inline `rules` (i.e. any backup taken from a healthy cluster) used to ship only `path` across the cluster: `validate_cluster_hocon` wrote the ACL file locally and rewrote the source to its post-write form before the config went through `cluster_rpc`. Peer nodes had no such file on disk and failed to apply with:

```
[critical] msg: failed_to_read_acl_file
[error]    msg: cluster_rpc_apply_failed, result: {error,#{reason => failed_to_read_acl_file, kind => validation_error, matched_type => "authz:file"}}
```

The leader's tnx_id advances while peers retry every minute and the `cluster_rpc_apply_failed` alarm stays activated.

### Fix

In `emqx_mgmt_data_backup:validate_cluster_hocon/1`, for every file authz source whose **input** form carried inline `rules`, restore that field before returning the validated config. Each peer's `pre_config_update` then writes its own local copy of `acl.conf` from the replicated content — the same code path the working `/authorization/sources` API has always used.

Sources whose input was a bare `path` (user-specified, e.g. a hand-edited HOCON pointing to a deployment-managed file) are left untouched so the user's intent is preserved.

### Test

New cluster CT `t_import_file_authz_source_on_cluster` in `emqx_mgmt_data_backup_SUITE` sets a file authz source on Core1 (rules-form), exports a backup, mutates the config so re-import is meaningful, imports the backup, and asserts both core nodes end up with the imported `rules` content in their own local `acl.conf` and no `cluster_rpc_apply_failed` alarm fires. Confirmed locally to fail on pristine `release-60` (Core2 keeps the pre-import content) and pass with the fix.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_data_backup.erl
+++ b/apps/emqx_management/src/emqx_mgmt_data_backup.erl
@@ -1144,7 +1144,8 @@ validate_cluster_hocon(RawConf0) ->
     %% the running value and drop required fields without defaults such as
     %% `node.cookie'.
     RawConfFiltered = maps:without(import_skip_root_keys(), RawConf0),
-    %% write ACL file to comply with the schema...
+    %% Write ACL file(s) locally so the `authz:file` schema's path validator
+    %% (which calls file:consult/1) passes during the schema check below.
     RawConf1 = emqx_authz:maybe_write_files(RawConfFiltered),
     maybe
         {ok, RawConf} ?=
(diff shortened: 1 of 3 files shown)


apps/emqx_management/test/emqx_mgmt_data_backup_SUITE.erl
```diff
@@ -597,6 +619,84 @@ t_verify_imported_mnesia_tab_on_cluster(Config) ->
...
+    {ok, _} = ?ON(
+        Core1,
+        emqx_conf:update(
+            [authorization, sources],
+            {replace, [OtherSource]},
+            #{override_to => cluster}
+        )
+    ),
+    %% Import — pre-fix this triggered cluster_rpc_apply_failed on Core2
+    %% because the importer used to ship `path` instead of `rules`.
```
thalesmg, line 670: Q: shouldn't we delete the existing `acl.conf` file created at l. 638 to actually have a missing file at the path?

thalesmg (dismissed):

apps/emqx_management/src/emqx_mgmt_data_backup.erl
```diff
@@ -1166,6 +1175,35 @@ validate_cluster_hocon(RawConf0) ->
 import_skip_root_keys() ->
     [<<"node">>, <<"rpc">>].
 
+restore_inline_authz_rules(
+    #{<<"authorization">> := #{<<"sources">> := NewSources} = NewAuthz} = NewConf,
+    #{<<"authorization">> := #{<<"sources">> := OldSources}}
+) ->
+    Restored = [maybe_restore_source_rules(S, OldSources) || S <- NewSources],
+    NewConf#{<<"authorization">> => NewAuthz#{<<"sources">> => Restored}};
```
thalesmg: nit:


```suggestion
    NewConf#{<<"authorization">> := NewAuthz#{<<"sources">> := Restored}};
```

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14616)

PR #17345, opened 2026-05-21
fix(authz): expand path to rules before replicating file source

zmstone:
Fixes EMQX-15330

Release version:
5.8.11

## Summary

\`emqx ctl conf load\` and \`PUT /api/v5/configs?key=authorization\` used to ship the user-provided \`path\` of a file authz source across the cluster verbatim. Peer nodes had no such file on disk and failed to apply the change with:

\`\`\`
[critical] msg: failed_to_read_acl_file
[error]    msg: cluster_rpc_apply_failed, result: {error,#{reason => failed_to_read_acl_file, kind => validation_error, matched_type => "authz:file"}}
\`\`\`

The leader's tnx_id advances while peers retry every minute indefinitely; the \`cluster_rpc_apply_failed\` alarm stays activated.

### Fix

Expand \`path\` to inline \`rules\` (by reading the file locally on the initiator) in \`emqx_conf_cli:load_config_from_raw/2\` before the config is handed to \`emqx_conf:update\`. Each peer's \`pre_config_update\` then re-runs \`emqx_authz:maybe_write_source_files/1\`, writing its own local copy of \`acl.conf\` from the replicated content — the same code path the working \`/authorization/sources\` API has always used.

### Why only one site here (vs. two on release-60)

This is a port of #17343 (release-60). On release-60 the data-backup importer had the same bug — its \`validate_cluster_hocon\` result (post-write, path-form) is what got passed to \`do_import_conf\`. On release-58 the importer discards that result and feeds \`do_import_conf\` the original rules-form \`RawConf1\`, so the import path is already correct on this branch. Only the CLI / REST \`/configs\` site needs the fix.

### Test

New cluster CT \`t_load_file_authz_with_path_on_cluster\` in \`emqx_mgmt_data_backup_SUITE\` places an ACL file on Core1 only, loads a HOCON pointing at that path via \`emqx_conf_cli:load_config\`, and asserts both core nodes end up with the same content in their own local \`acl.conf\` and no \`cluster_rpc_apply_failed\` alarm fires. Confirmed to fail without the fix (Core2 gets no acl.conf) and pass with it.

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/emqx_conf.app.src
+++ b/apps/emqx_conf/src/emqx_conf.app.src
@@ -1,6 +1,6 @@
 {application, emqx_conf, [
     {description, "EMQX configuration management"},
-    {vsn, "0.4.8"},
+    {vsn, "0.4.9"},
     {registered, []},
     {mod, {emqx_conf_app, []}},
     {applications, [kernel, stdlib]},
--- a/apps/emqx_conf/src/emqx_conf_cli.erl
+++ b/apps/emqx_conf/src/emqx_conf_cli.erl
@@ -351,7 +351,12 @@ load_config_from_raw(RawConf0, Opts) ->
     SchemaMod = emqx_conf:schema_module(),
     RawConf1 = emqx_config:upgrade_raw_conf(SchemaMod, RawConf0),
     case check_config(RawConf1, Opts) of
-        {ok, RawConf} ->
+        {ok, RawConfChecked} ->
+            %% If the user supplied a file authz source with `path`, expand it
+            %% to inline `rules` (by reading the file locally) so cluster_rpc
+            %% replication carries the content rather than a node-local path
+            %% that other nodes can't read.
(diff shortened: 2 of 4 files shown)


zmstone:
Closing: re-examined the import flow on release-58 and it's already correct — the importer discards `validate_cluster_hocon`'s rewritten config and feeds `do_import_conf` the original rules-form `RawConf1`. So the customer's reported `failed_to_read_acl_file` / `cluster_rpc_apply_failed` on import does not reproduce on this branch.

The fix in this PR was for `emqx_conf_cli:load_config_from_raw` (CLI `conf load` / REST `PUT /configs`). CI on the release-60 sibling PR (#17343) revealed that unconditionally expanding `path` → `rules` rewrites the stored path from `${EMQX_ETC_DIR}/acl.conf` to `data_dir/authz/acl.conf` and breaks `t_load_config`. Loading a HOCON with a custom `path` is now expected to be paired with the file existing on every cluster node, same as before.

No user-facing bug remains here to fix.

PR #17348, opened 2026-05-22
fix(conf): suppress noisy cluster sync status diffs

hjianbo:
Fixes https://emqx.atlassian.net/browse/EMQX-15291
Fixes https://emqx.atlassian.net/browse/EMQX-15310

Release version:
6.0.3

## Summary

`emqx ctl conf cluster_sync status` can emit noisy and misleading diagnostics when nodes have the same effective checked configuration but different non-runtime representations. Typical cases include sparse imported raw config on one node and boot-time default-materialized raw config on another node, or generated timestamp metadata refreshed on only some nodes.

This PR ports the cluster sync status fixes from #17313 and #17387 to `release-60`.

The `release-60` port now:

* normalizes each running node's raw config through defaults and schema check before status comparison, with a fallback to checked config if raw normalization fails;
* suppresses raw-only representation differences when printing diagnostics so only real checked config changes remain visible;
* ignores timestamp-only `created_at` and `last_modified_at` metadata differences for actions, sources, bridges, and rule metadata;
* keeps the warning when checked config is genuinely inconsistent, even if the pruned raw payload is empty;
* allows the root raw config path in the v5 proto spec used by status normalization.

Regression coverage includes the import-then-join integration test `emqx_conf_cluster_sync_import_SUITE` from #17313, adapted for `release-60` import default materialization, plus focused CLI/EUnit guards for raw diff pruning and timestamp-only metadata differences.

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_cli.erl
+++ b/apps/emqx_conf/src/emqx_conf_cli.erl
@@ -953,11 +953,32 @@ changed(K, V, Conf) ->
 find_running_confs() ->
     lists:map(
         fun(Node) ->
-            Conf = emqx_conf_proto_v5:get_config(Node, ?global_ns, []),
+            Conf = normalize_running_conf(Node),
             {Node, maps:without(?READONLY_ROOT_KEYS, Conf)}
         end,
         emqx_bpapi:nodes_supporting_bpapi_version(?BPAPI_NAME, 5)
     ).
+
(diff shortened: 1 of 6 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14652)

PR #17349, opened 2026-05-22
fix(clink): stop routerepl manager from trapping exits

keynslug:
Part of [EMQX-15306](https://emqx.atlassian.net/browse/EMQX-15306).

Release version: 6.1.2, 6.2.1

## Summary

This PR ensures that the route replication process tree of a terminating Cluster Link reacts to shutdown requests quickly.

## PR Checklist

[EMQX-15306]: https://emqx.atlassian.net/browse/EMQX-15306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_cluster_link/src/emqx_cluster_link_routerepl.erl
+++ b/apps/emqx_cluster_link/src/emqx_cluster_link_routerepl.erl
@@ -24,7 +24,9 @@
 %% Internal API
 -export([
     start_link_manager/3,
-    start_link_syncer/4
+    start_link_syncer/4,
+    forward_client_disconnected/2,
+    forward_publish/2
 ]).
 
 %% NOTE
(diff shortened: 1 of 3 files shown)


apps/emqx_cluster_link/src/emqx_cluster_link_routerepl.erl
```diff
@@ -285,7 +285,6 @@ callback_mode() ->
 
 init_manager(St = #st{}) ->
     ?tp_routerepl("init", #{actor => St#st.actor}),
-    _ = erlang:process_flag(trap_exit, true),
```
zmstone, line 288: maybe just comment it out and add a comment to explain why. just in case someone adds it back

keynslug: To be clear, this is far from being main contributing factor to the behavior described in EMQX-15306.

Also I just realized that it's not going to be a one-liner, as manager actually owns `emqtt` client process. With this in mind, I'm going to revisit it later and instead focus on more promising approaches first.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14655)

PR #17341, opened 2026-05-21
ci: speed up coverdata to cobertura conversion

zmstone:
## Diagnosis

`tests_passed` spent ~9m 32s of its ~10m 46s wall time in `./scripts/covertool-2.1.0-emqx-1 -cover _build/test/cover ...` (88% of the job). covertool routes every `cover:import/1` and `cover:analyse/3` through OTP's `cover_server` gen_server, so ~11.5M counter updates (121 shard files × ~95k bumps each) are serialised through a single ETS owner.

## Change

Replace covertool with `scripts/cover-to-cobertura.escript`, a thin direct binary parser of the `.coverdata` on-disk format (1-byte size prefix + `binary_to_term`, mirroring `lib/tools/src/cover.erl`'s `get_term/1`). Counters are summed per `{Module, Line}` in plain Erlang maps across N parallel workers, then emitted as line-level Cobertura XML.

Output schema is line-level only (no `<methods>` block); Codecov only consumes `<lines>` for line coverage.

## Local measurement

Downloaded the 121 coverdata files (676 MB) from a recent `tests_passed` run and ran each tool to completion:

| Tool / Workers | Wall time |
| --- | --- |
| covertool (current) | > 13 min (killed) |
| cover-to-cobertura.escript, 1 worker | 40 s |
| cover-to-cobertura.escript, 4 workers | 12 s |
| cover-to-cobertura.escript, 8 workers | 8 s |

The escript defaults to `erlang:system_info(schedulers)` workers, so a 4-vCPU `ubuntu-22.04` GHA runner will use 4. Expected CI cost: ~30–60 s vs the current ~570 s.

## Notes

- `scripts/covertool-2.1.0-emqx-1` (the prebuilt 40 KB escript bundle) is removed; the only consumer was this workflow step.
- Output filename, lookup directory walk (`apps/**/*.erl`), and module-skip behaviour (no source found → skip class) all match covertool.
- The `.coverdata` term layout has been stable in OTP since the early 2010s; the script depends only on the publicly documented `get_term/1` shape.

PR #17359, opened 2026-05-23
ci: run cover-to-cobertura.escript in compile mode

zmstone:
Follow-up to #17341.

The `%%! -mode(compile)` on the shell-args line is ignored — escript only recognises `-mode(compile).` as a module attribute. Without it the script runs in interpret mode, and `fun read_file/2` passed to `lists:foldl/3` blows up at call time:

```
{undef,[{erl_eval,read_file,2,[]}, ...]}
```

### DIFF
--- a/scripts/cover-to-cobertura.escript
+++ b/scripts/cover-to-cobertura.escript
@@ -1,5 +1,4 @@
 #!/usr/bin/env escript
-%%! -mode(compile)
 %%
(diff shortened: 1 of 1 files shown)


hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14617)

PR #17117, opened 2026-04-23
Tighten dashboard MFA reset flow and SSO viewer MFA RBAC

HJianBo:
Fixes 
- https://emqx.atlassian.net/browse/EMQX-15227
- https://emqx.atlassian.net/browse/EMQX-15226

Release version: 5.10.4

## Summary

This PR aligns dashboard MFA reset behavior with the frontend flow, tightens viewer permissions for forced-MFA SSO users, and hardens SSO MFA temporary token handling.

- removes the dedicated hard-reset path on `DELETE /users/:username/mfa?reset=true` and standardizes reset/re-key behavior on `POST /users/:username/mfa`
- updates `apps/emqx_dashboard/src/emqx_dashboard_api.erl`, `apps/emqx_dashboard/src/emqx_dashboard_admin.erl`, `rel/i18n/emqx_dashboard_api.hocon`, and `apps/emqx_dashboard/README.md` so `POST` regenerates the TOTP secret, clears pending MFA state, and invalidates existing dashboard JWTs
- updates dashboard MFA tests in `apps/emqx_dashboard/test/emqx_dashboard_mfa_SUITE.erl` to cover the new reset/re-key semantics and old-query compatibility behavior
- tightens RBAC in `apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl` so a viewer can no longer disable their own MFA when they are an SSO user whose backend has `force_mfa = true`
- adds coverage in `apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl` for SSO viewer deletion with `force_mfa = false/true`, while preserving the existing local-user and setup behavior
- updates `apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa.erl` so existing per-user SSO MFA state is honored even when backend `force_mfa = false`; `force_mfa` now only decides whether `not_configured` users must start setup
- updates `apps/emqx_dashboard/src/emqx_dashboard_mfa.erl` and `apps/emqx_dashboard_sso/src/emqx_dashboard_sso_mfa_api.erl` so SSO MFA temp tokens are only consumed after a successful `setup` or `verify`, incorrect TOTP codes can be retried until TTL / max-attempt invalidation, and failed-attempt accounting is handled atomically
- adds API and concurrency coverage in `apps/emqx_dashboard_sso/test/emqx_dashboard_sso_mfa_SUITE.erl` for retry-after-failure, max-attempt invalidation, one-time token consumption, and atomic failed-attempt accounting
[...]

PR #17067, opened 2026-04-15
feat: port apikey scopes and sso mfa

JimMoen:
port #16942, #16943, #17117, #17122 to `release-60`

Release version:

6.0.3, 6.1.2, 6.2.0

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17361, opened 2026-05-24
feat: port dashboard login-user scopes and bootstrap fixes to release-60

JimMoen:
Fixes [EMQX-15307](https://emqx.atlassian.net/browse/EMQX-15307)

Release version: 6.0.3

## Summary

Ports the dashboard login-user scope model and related fixes from `release-510` to `release-60`. Builds on top of #17067, which previously ported API-key scopes (#16942) and SSO force-MFA (#16943) to this branch.

**Source PRs (release-510):**
- #17117 — `fix(api_key)`: lenient bootstrap-file scope handling
- #17235 — `feat(dashboard)`: fine-grained login-user scopes, default-admin protection, and role/scope consistency on partial updates

### What this PR adds

- **Login-user scopes.** Dashboard users now carry an optional `scopes` field, authorized against the same path-to-scope catalog used for API keys, layered on top of the existing role-based RBAC. Most relevant modules: `emqx_dashboard_admin`, `emqx_dashboard_rbac`, `emqx_dashboard_user_scopes` (new), `emqx_scope_catalog` (new, in `emqx_utils`).
- **Four login-only scopes**: `user_management`, `mfa_management`, `sso_management`, `api_key_management`. API keys cannot hold any of them, and the `publisher` API-key role is constrained to `[publish]` only. Both checks apply to the HTTP API and to bootstrap-file loading — incompatible scopes are dropped with a warning rather than rejecting the whole file (per #17117).
- **Two public catalog endpoints** for UI consumption: `GET /api_key_scopes` and `GET /user_scopes`. Top-level paths so they cannot be shadowed by sibling wildcard routes; accessible to any bearer-authenticated caller.
- **Per-SSO-backend `force_mfa`** is now persisted per user at provision time (`force_mfa_snapshot`) and overridable per-user by an admin via `DELETE`/`POST /users/:username/mfa`. The admin decision trumps live backend policy until changed.
- **Default-admin break-glass protection.** The `dashboard.default_username` user cannot be deleted, demoted from administrator, or have its `scopes` field set; only `description` is mutable. `DELETE /users/:username` on the default user is rejected even when other administrators exist.
- **Role/scope consistency on partial updates.** `PUT /users/:username` and `PUT /api_key/:name` now validate role changes against the effective persisted scopes when the request body omits the `scopes` field, so demotion paths cannot bypass the scope/role constraint.

### Behavior changes operators should be aware of (inherited from #17235)

1. `dashboard.default_username` is now a protected break-glass account.
2. Self-service on a user's own record now respects scopes. Only the dedicated change-password and MFA self endpoints still bypass scope checks; other operations such as `PUT /users/:self` are now subject to the user's scopes. A viewer without `user_management` can still change their own password and manage their own MFA, but can no longer edit other profile fields on their own record without being granted the relevant scope.
3. `PUT /users/:username` / `PUT /api_key/:name` reject role changes that would conflict with the persisted scopes (when the body omits `scopes`).

These are user-visible behavior changes but are forward-only — no schema migration is required, and clients that did not rely on the previous (looser) self-service or default-user semantics are unaffected.

### Verification

Full local CI matrix run before push:

- `make fmt` + `scripts/check-format.sh` — clean
- `scripts/apps-version-check.exs` — pass
- `scripts/check-api-scopes.sh` — all 69 minirest_api modules export `scopes/0`
- `scripts/spellcheck/spellcheck.sh` — pass
- `scripts/check-i18n-style.sh` + `scripts/check_missing_reboot_apps.exs` — pass
- `mix emqx.gen_deps` — `deps.txt` unchanged
- `mix emqx.xref` + `mix dialyzer --mode classic` + `mix emqx.static_checks` — all clean
- Common Test — 183/183 across `dashboard_rbac_SUITE`, `dashboard_admin_SUITE`, `dashboard_mfa_SUITE`, `dashboard_sso_mfa_SUITE`, `dashboard_user_scopes_SUITE`, `mgmt_api_api_keys_SUITE`, `mgmt_api_key_scopes_SUITE`

## PR Checklist

[EMQX-15307]: https://emqx.atlassian.net/browse/EMQX-15307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
```diff
@@ -146,20 +244,19 @@ do_check_rbac(
...
+    %% exemption) is decided in emqx_dashboard_api:authorize_mfa_change/3.
+    %% RBAC must not consult the live backend force_mfa flag here —
+    %% doing so would bypass admin_override and prevent mfa_management
+    %% scope holders from self-exempting.
     case Req of
-        #{bindings := #{username := Username}} ->
-            not is_forced_sso_mfa(maps:get(?backend, Actor, ?BACKEND_LOCAL));
-        _ ->
-            false
+        #{bindings := #{username := Username}} -> true;
```
hjianbo, line 258: P1: This now lets a viewer delete their own MFA without consulting the SSO backend's live `force_mfa` setting. The handler only treats `admin_override == mfa_required` as locked, and after the delete `emqx_dashboard_sso_mfa` classifies `disabled` as `admin_disabled -> {ok, login}`, so an SSO user on a `force_mfa=true` backend can complete MFA once, call `DELETE /users/:self/mfa`, and then log in without MFA. Please add a regression test for that flow and keep live `force_mfa`/disabled-state semantics in the self-disable decision.

apps/emqx_dashboard/src/emqx_dashboard_token.erl
```diff
@@ -265,12 +285,31 @@ check_rbac(Req, HandlerInfo, JWT) ->
...
-            ok = save_new_jwt(JWT),
-            {ok, ActorContextFinal};
+            %% Layer the login-user scope check ON TOP of role-based
+            %% RBAC. Only invoked here (dashboard JWT path) — API key
+            %% authorisation uses its own scope mechanism via
+            %% emqx_mgmt_auth:check_path_in_scopes/2 and must not
+            %% trip on this.
+            #?ADMIN_JWT{extra = Extra, username = Username} = JWT,
+            AdminKey = full_admin_key(Username, Extra),
+            case emqx_dashboard_rbac:check_login_user_scopes(AdminKey, Req) of
```
hjianbo, line 295: P2: `check_login_user_scopes/2` now runs for every successful dashboard JWT authorization, so existing `$denied` routes fail for login users too. In particular, `emqx_dashboard_sso_saml_api` still returns `?SCOPE_DENIED`, and `/sso/saml/metadata` is not marked `security => []`; bearer admins will therefore get 403 for the metadata endpoint. If this endpoint should be public, mark it `security => []` and `?SCOPE_PUBLIC`; if it should require admin auth, map it to `?SCOPE_SSO_MGMT` instead of `$denied`.

hjianbo (commented):
Reviewed the permission and MFA paths. I found two issues that should be addressed before merge.

zmstone (dismissed):

changes/ee/feat-17235.en.md
```diff
@@ -0,0 +1,11 @@
+Introduced fine-grained scope-based access control for dashboard login users. Each user record now carries an optional `scopes` field; when set, requests are authorized against the same path-to-scope catalog used for API keys, layered on top of the existing role-based check. Four new scopes (`user_management`, `mfa_management`, `sso_management`, `api_key_management`) cover dashboard-only endpoints and are admin-only except `mfa_management`, which any role may hold for self-exemption from forced MFA. API keys cannot hold any of the four login-only scopes, and the `publisher` API-key role is now constrained to `[publish]` only — both checks apply to the HTTP API and to bootstrap-file loading (incompatible scopes are dropped with a warning).
```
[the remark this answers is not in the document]
JimMoen: Thanks for flagging — I took a look. The current L1 of `changes/ee/feat-17235.en.md` actually reads:

> Four new scopes (`user_management`, `mfa_management`, `sso_management`, `api_key_management`) cover dashboard-only endpoints and are admin-only except `mfa_management`, which any role may hold for self-exemption from forced MFA.

No "first three" wording exists in this file. I think the wording was transposed from the companion file `changes/ee/feat-17361.en.md`, which did have the "first three" ordering bug — that one is being fixed in a follow-up commit (`64bfe3ee1d`). `feat-17235.en.md` should already be correct; happy to double-check if you're looking at a different revision.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14701)

PR #17366, opened 2026-05-25
fix(redis): validate sentinel auth in dry run

hjianbo:
Fixes https://emqx.atlassian.net/browse/EMQX-15308

Follow-up https://github.com/emqx/emqx/pull/17256

Release version: 5.10.4

## Summary

Redis Sentinel dry runs could reuse the process-global `eredis_sentinel` discovery process that had already been started by an existing connector with valid Sentinel credentials. When a user edited the connector and probed with an invalid Sentinel username, the probe could report success even though saving or reconnecting the connector failed with `WRONGPASS`.

This change adds an explicit Sentinel discovery check in `emqx_redis:on_start/2` before starting the Redis pool. The check uses the current connector config to connect to Sentinel and run `SENTINEL get-master-addr-by-name`, so dry runs and real connector starts validate the current Sentinel credentials instead of relying on stale global Sentinel state.

A regression test creates a valid Sentinel-auth connector first, then dry-runs the same Sentinel with an invalid Sentinel username and expects an authentication error. I verified this with the new regression case, the existing four Redis/Sentinel auth matrix cases, and `erlfmt`.

No changelog is added because this fixes behavior introduced by the previous Redis Sentinel change on `release-510` and should be covered by that release note.

## PR Checklist
- [x] Change log is not added because this fixes behavior introduced by the previous Redis Sentinel PR on `release-510`

### DIFF
--- a/apps/emqx_redis/src/emqx_redis.erl
+++ b/apps/emqx_redis/src/emqx_redis.erl
@@ -34,6 +34,7 @@
 -define(REDIS_HOST_OPTIONS, #{
     default_port => ?REDIS_DEFAULT_PORT
 }).
+-define(SENTINEL_VALIDATION_CLIENT, sentinel_validation_client).
 
 %%=====================================================================
 namespace() -> "redis".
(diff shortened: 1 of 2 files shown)


apps/emqx_redis/src/emqx_redis.erl
```diff
@@ -137,25 +137,92 @@ on_start(InstId, Config0) ->
...
                 {ok, _} ->
-                    {ok, State};
-                {ok, _, _} ->
-                    {ok, State};
+                    ok;
                 {error, Reason} ->
-                    {error, Reason}
+                    validate_sentinel_servers(Rest, Sentinel, Options, [Reason | Errors])
+            after
+                eredis_sentinel_client:stop(Pid)
```
thalesmg: nit: maybe it could be safer to also register this pid as an allocated resource.  

that way, if the sentinel master query happens to take a long time and the process running is killed due to timeout, then we can clean it up in the `on_stop` call that is made by the resource manager.  in such cases, this `after` clause will not be called.

thalesmg (dismissed):

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14532)

PR #17369, opened 2026-05-25
fix(conf): move dashboard listener defaults from emqx.conf to base.hocon

zmstone:
Fixes <issue-or-jira-number>

Release version:
6.0.3, 6.1.1, 6.2.0

## Summary

Field report: `dashboard.listeners.https.ssl_options.{certfile,keyfile}` was hardcoded in the user-editable `etc/emqx.conf` shipped with the release. Because the precedence order is `etc/base.hocon < cluster.hocon < emqx.conf < environment variables`, the hardcoded block silently clobbered every runtime override on restart -- including the `emqx_acme` plugin's "auto-configure dashboard HTTPS on first issuance" feature. `PUT /api/v5/configs/dashboard` would succeed, the runtime config would show the bundle paths, the plugin would log `dashboard_https_enabled`... and then after a restart the listener would serve the default self-signed cert again.

This PR moves the entire `dashboard { listeners { ... } }` block out of `apps/emqx_dashboard/etc/emqx_dashboard.conf` (the per-app fragment that gets merged into `emqx.conf.all`) and into `apps/emqx_conf/etc/base.hocon`. Now `cluster.hocon` -- where dashboard PUTs and ACME-plugin writes land -- correctly overrides the defaults across restarts.

### Files

- `apps/emqx_conf/etc/base.hocon` -- now carries the dashboard listener defaults, with a comment explaining why they live in `base.hocon` rather than `emqx.conf`.
- `apps/emqx_dashboard/etc/emqx_dashboard.conf` -- emptied (matching the style of `apps/emqx_auth/etc/emqx_auth.conf`, `apps/emqx_gateway/etc/emqx_gateway.conf`, etc., which are also empty).
- `apps/emqx_management/test/emqx_mgmt_api_configs_2_SUITE.erl` -- stale comment pointing to `emqx_dashboard.conf` updated to point to the new location in `base.hocon`. No behavioral change in the test itself; it already inlines its own dashboard config.

### Verification

- Rebuilt the release; `_build/emqx-enterprise/rel/emqx/etc/emqx.conf` no longer contains the dashboard block, and `etc/base.hocon` carries the defaults including `http.bind = 18083` and the placeholder cert paths.
- `bin/emqx check_config` passes both with an empty `cluster.hocon` and with a `cluster.hocon` that overrides `dashboard.listeners.https.ssl_options.{certfile,keyfile}`.
- HOCON layering precedence is enforced in `apps/emqx/src/emqx_config.erl:789` via `hocon:files([BaseHocon, ClusterFile | Conf], ...)`, so a value in `cluster.hocon` deterministically overrides one in `base.hocon`.

### Scope / backports

The same `apps/emqx_dashboard/etc/emqx_dashboard.conf` template exists on the older v5 lines (`release-58`, `release-510`) and on `release-61` / `release-62`. This PR only fixes `release-60`; sync-forward should pick it up per the usual workflow. Backports to v5 lines should follow if the maintainers decide the fix is in-scope there.

## PR Checklist


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14618)

PR #17374, opened 2026-05-26
fix(scopes): unify API key and dashboard user scope response as tri-state contract

JimMoen:
Release version: 5.10.4

## Summary

Two unreleased features on `release-510` chose **inconsistent** ways
to project the `scopes` field in their GET response:

- API key scopes (#16942) — omitted the field whenever no explicit
  scopes were set, even though `extra` had no `scopes` key.
- Dashboard user scopes (#17235) — always expanded the role default
  into a concrete 10-or-14-item list whenever no explicit scopes
  were set.

Both behaviours surface the same persisted shape (`extra` with or
without a `scopes` key, or with `scopes => []` / `scopes => [list]`)
but project it differently, which means:

1. The frontend has to special-case two different shapes for the
   same UX concept.
2. The user form's read-modify-write cycle silently sediments role
   default into an explicit list: read the user → form prefills with
   the 10 or 14 effective scopes → save the form unchanged → the
   user is now pinned to that list and will no longer track future
   role-default evolution.

This PR aligns both responses on a single tri-state contract and
makes the "no scopes" state observable only for genuine legacy
upgrade artefacts.

### Tri-state contract

| persisted state              | response `scopes`     |
|------------------------------|-----------------------|
| `extra` has no `scopes` key  | `<<"unset">>` (string)|
| `extra => #{scopes => []}`   | `[]`                  |
| `extra => #{scopes => List}` | `List`                |

The `<<"unset">>` literal is reserved for **records persisted before
#16942 / #17235 landed** that survive an upgrade without ever being
touched by POST, PUT, bootstrap, SSO provisioning, or default-admin
first-boot. Fresh records always carry an explicit `scopes` list.

### Why `<<"unset">>` instead of `null`

`null` is too overloaded in JSON ("missing", "explicitly cleared",
"unknown") and is awkward in dashboard frontends that strictly type
`scopes` as `string[]`. A dedicated string sentinel is unambiguous
and easy to surface to operators (e.g. "this record predates the
scopes feature — review and assign explicit scopes").

The unified response shape `[]` / `["scope", ...]` / `"unset"` is
declared by a dedicated `fields(app_response)` schema for API keys
and a `field(scopes_response)` callback for dashboard users so the
OpenAPI spec accurately reflects the runtime shape (request bodies
still require a plain `array(binary())`, not a union).

### How fresh records avoid `<<"unset">>`

Each creation path materialises role-default scopes at insertion
time so the persisted record is always self-contained:

- **API key POST** (`/api_key`) — `effective_scopes_on_create/2`
  expands `undefined` to `role_default_scopes(Role)`:
    - administrator / viewer → `?GENERIC_SCOPES` (10 management
      scopes, no login-only)
    - publisher → `[<<"publish">>]`
- **Dashboard user POST** (`/users`) — same expansion:
    - administrator → `?GENERIC_SCOPES ++ ?LOGIN_ONLY_SCOPES` (14)
    - viewer → `?GENERIC_SCOPES` (10)
- **Bootstrap loader** — 2-segment (`key:secret`) and 3-segment
  (`key:secret:role`) lines now seed role-default scopes. The
  4-segment form is unchanged.
- **SSO provisioning** — `add_sso_user/4` (called by every SSO
  backend's `ensure_user_exists`) seeds `role_default_scopes(?ROLE_VIEWER)`
  at insertion time. No follow-up call needed.
- **Default admin first-boot** — `do_add_default_user/2` seeds the
  14-scope admin default.

PUT semantics are unchanged: omitted `scopes` still preserves the
persisted state (partial update), explicit `[]` still means deny-all,
explicit list still overwrites.

### Authorization is unchanged

`emqx_mgmt_auth:check_scopes/2` and
`emqx_dashboard_admin:effective_scopes_of_admin/1` still treat a
missing `scopes` key as "role default". The role-default expansion
that previously happened at projection time now happens at insertion
time, so by the time RBAC reads the row there is nothing left to
expand. Same effective decision, different point in the pipeline.

## Commits

- `d43a0f7d21` `fix(scopes): unify API key and dashboard user scope response as tri-state contract`
- `009c4169b0` `fix(scopes): materialise role-default scopes on creation and adopt "unset" sentinel`
- `f287a90944` `fix(scopes): seed role-default scopes for bootstrap, SSO, and default admin`

## Tests

Recompiled all four sources plus four CT SUITEs against OTP 27.
Notable additions:

  - `t_api_post_materialises_default_scopes` /
    `t_post_users_materialises_default_scopes` — verify the mnesia
    row actually carries the seeded list (not just the response).
  - `t_api_legacy_record_shows_unset_sentinel` /
    `t_legacy_record_shows_unset_sentinel` — manually delete the
    `scopes` key from a freshly created row and confirm the
    `<<"unset">>` sentinel is still reachable.
  - `t_bootstrap_file_2_segment_seeds_default_scopes` /
    `t_bootstrap_file_3_segment_publisher_seeds_publish_scope` —
    cover the two new bootstrap paths.
  - `t_default_admin_seeded_with_role_default_scopes` — first-boot
    admin gets the 14-scope admin default.
  - `t_add_sso_user_seeds_role_default_scopes` — fresh SSO viewer
    gets the 10-scope viewer default.

## Frontend impact

The response shape change is documented in
`FRONTEND-scope-changes.md` in the task tree; the frontend PR will
land in `emqx-dashboard5`.

## Why no changelog and no app vsn bump

Both #16942 (API key scopes) and #17235 (dashboard user scopes)
landed on `release-510` after `5.10.3` shipped, so they are not in
any stable release yet. Aligning their response contracts **during**
the same release development cycle, before any stable build exposes
the asymmetry to end users, doesn't require a changelog (the parent
PRs' changelogs cover the user-facing surface) and doesn't require
an app vsn bump (`scripts/apps-version-check.sh` only checks against
the previous release tag, and `emqx_management` / `emqx_dashboard`
were already bumped in the parent PRs).

## PR Checklist

- [x] Change log: N/A — intra-release-cycle alignment of two
      unreleased features; parent PRs' changelogs cover the
      user-facing surface.
- [x] Schema changes are backward compatible or intentionally
      breaking (intentionally breaks the response shape of two
      unreleased features only; described above).

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_admin.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_admin.erl
@@ -144,7 +144,29 @@ add_user(Username, Password, Role, Desc) when is_binary(Username), is_binary(Pas
     end.
 
 do_add_user(Username, Password, Role, Desc) ->
-    Res = mria:sync_transaction(?DASHBOARD_SHARD, fun add_user_/4, [Username, Password, Role, Desc]),
+    do_add_user(Username, Password, Role, Desc, undefined).
+
+%% @doc Internal admin/SSO-user creation entry point.
+%%
+%% `InitialScopes' controls what ends up in the persisted `extra' map's
+%% `scopes' key BEFORE any later `set_user_scopes/2' call:
+%%
+%%   * `undefined' (default, 4-arg version) - no scopes key is written; the
+%%     caller is expected to materialise scopes afterwards (e.g. the API POST
+%%     handler does `maybe_set_user_scopes/2' with role-default scopes).
+%%     This keeps the read-modify-write path of `emqx_dashboard_api' unchanged.
+%%
+%%   * `[binary(), ...]' (5-arg version) - the listed scopes are written
+%%     into the record atomically inside the same mria transaction as the
+%%     row insertion. This is used by paths that do NOT have a follow-up
+%%     `set_user_scopes' step (SSO auto-provisioning, default-admin bootstrap)
+%%     so a new row never appears with a missing scopes key.
+do_add_user(Username, Password, Role, Desc, InitialScopes) ->
+    Res = mria:sync_transaction(
+        ?DASHBOARD_SHARD,
+        fun add_user_/5,
+        [Username, Password, Role, Desc, InitialScopes]
+    ),
     return(Res).
 
 get_mfa_enabled_state(Username) ->
@@ -557,15 +579,17 @@ force_add_user(Username, Password, Role, Desc) ->
     end.
 
 %% @private
-add_user_(Username, Password, Role, Desc) ->
+add_user_(Username, Password, Role, Desc, InitialScopes) ->
     case mnesia:wread({?ADMIN, Username}) of
         [] ->
+            Extra0 = #{password_ts => erlang:system_time(second)},
+            Extra = maybe_seed_initial_scopes(Extra0, InitialScopes),
             Admin = #?ADMIN{
                 username = Username,
                 pwdhash = hash(Password),
                 role = Role,
                 description = Desc,
-                extra = #{password_ts => erlang:system_time(second)}
+                extra = Extra
             },
             mnesia:write(Admin),
             ?SLOG(info, #{msg => "dashboard_sso_user_added", username => Username, role => Role}),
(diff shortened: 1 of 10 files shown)


zmstone:
why is it a feat?

hjianbo:
Required fix: the response schema was not updated for the new tri-state contract.

This PR now returns `scopes: null` when scopes are not explicitly set (dashboard side: `ee_user_extra/1`; API key side: `maybe_add_scopes/2`), but both public schemas still declare `scopes` as `hoconsc:array(binary())`:

- `apps/emqx_dashboard/src/emqx_dashboard_api.erl:294`
- `apps/emqx_management/src/emqx_mgmt_api_api_keys.erl:176`

That means swagger / generated TS types will still expose `string[]`, while the runtime response can be `null`. Frontend code or schema-based clients may consume the field with the wrong type. Please update the response schema to allow `array | null`, or split request and response schemas: request can keep accepting only array/absent, while response allows array/null.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14637)

PR #17376, opened 2026-05-26
test(backup): widen cluster-link export/import coverage

zmstone:
Adds three new CT cases to `emqx_mgmt_data_backup_SUITE` alongside the existing `t_cluster_links`:

- `t_cluster_link_password_round_trip` — verifies a non-empty `password` (typed `emqx_schema_secret`) survives export → wipe → import via `emqx_secret:unwrap/1`. Also pins down current behaviour: the archive's `cluster.hocon` contains the literal password, since export reads the raw override config verbatim. A future move to encrypt/redact secrets-at-rest in backups will break this assertion loudly.
- `t_cluster_link_multi_round_trip` — round-trips three distinct links (mix of SSL on/off, varied topics/pool_size) and asserts all are restored with fields intact and in the original list order (cluster-link uses first-link-matched semantics). Bonus: updates one link post-import and asserts the others are untouched.
- `t_cluster_link_ps_actor_incarnation_preserved` — `ps_actor_incarnation` is the per-link sequence the config handler uses so persistent-routes actors stay in sync after each config change. Verifies a non-zero value survives the round-trip and is not reset to the schema default of 0.

Test-only; no user-visible behaviour change.

### DIFF
--- a/apps/emqx_management/test/emqx_mgmt_data_backup_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_data_backup_SUITE.erl
@@ -554,6 +554,174 @@ t_cluster_links(_Config) ->
     ] = emqx:get_raw_config([cluster, links]),
     ?assertEqual({ok, ?CACERT}, file:read_file(CertPath1)).
 
+-doc """
+The link's `password` field is typed `emqx_schema_secret`; the raw config
+on disk stores it verbatim, but the runtime config wraps it in an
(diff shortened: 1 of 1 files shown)


qzhuyan (approved):

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14654)

PR #17382, opened 2026-05-27
fix(cm_registry): Handle partial network isolation during cleanup

ieQu1:
Fixes [EMQX-15260](https://emqx.atlassian.net/browse/EMQX-15260)

Release version:
5.10.4

## Summary

1. Put more safeguards before triggering channel registry cleanup on behalf of nodes that are left.
2. Trigger registry heal procedure when replicant nodes detect bootstrap of the registry shard.

## PR Checklist

[EMQX-15260]: https://emqx.atlassian.net/browse/EMQX-15260?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx/src/emqx_broker_heal.erl
+++ b/apps/emqx/src/emqx_broker_heal.erl
@@ -26,7 +26,7 @@ Limitations:
 -behaviour(gen_server).
 
 %% API:
--export([start_local/1, on_autoheal/1, consistency_check/0]).
+-export([start_local/1, on_autoheal/1]).
 
 %% behavior callbacks:
 -export([init/1, handle_call/3, handle_continue/2, handle_cast/2, handle_info/2, terminate/2]).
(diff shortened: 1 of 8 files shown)


apps/emqx/src/emqx_broker_heal.erl
```diff
@@ -71,6 +71,38 @@ Start a process that will cause all local channels to renew their registrations
...
+
+This check is not precise and is based on heuristics.
+Due to possible race conditions,
+it assumes that the registry is consistent when it contains the majority of local channels.
+""".
+-spec consistency_check() -> boolean().
+consistency_check() ->
+    MS = {{'$1', '_'}, [], ['$1']},
+    SampleSize = 1000,
+    case ets:select(?CHAN_TAB, [MS], SampleSize) of
```
zmstone: if registry record is lost by any chance, it should be the oldest.
this means, we can check CHAN_CONN_TAB, it's `ordered_set` and keyed by `pid`.

check in a loop, start from `Pid = ets:first(?CHAN_CONN_TAB)`

- Find client id from CHAN_CONN_TAB.
- Check `emqx_cm_registry:lookup_all_channels(ClientId) =/= []`
  - If `true`, consistency is NOT lost, no need for reconcile.
  - If `false`, check `is_process_alive(Pid)`
    - If `true`: consistency is lost, need reconcile;
    - If `false`: this is a race condition, check `ets:next(?CHAN_CONN_TAB, Pid)`

zmstone: a race condition:
for clients in replicant nodes, cm-registry records are replicated back to itself async, so it may result in false positive (of inconsistency).

ieQu1: There's no real way to make this check reliable, as it may race with other node triggering cm registry cleanup on behalf of node executing the check. Because of that, I tried to err on the side of false positivies (for inconsistency). The purpose is just to avoid hammering core nodes if data seems ok.

zmstone: ok, i think CHAN_CONN_TAB will be less likely to give false positives because it's a ordered_set with pid as key.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14649)

PR #17383, opened 2026-05-27
fix(takeover): refresh channel stats after session replay

zmstone:
Release version:
6.0.3, 6.1.2, 6.2.1

## Summary

After a session takeover, the dashboard's client view and REST API
`mqueue_len` / `inflight_cnt` were stale for up to 15 seconds (the
default `mqtt.idle_timeout` / stats-timer interval).

### Root cause

Commit 9c43e0ae15 ("feat(chan): allow return `continue` to
short-circuit control flow back", first shipped in `e5.9.0`) deferred
the takeover replay to a `continue` reply that runs **after** the
`{event, connected}` reply. Consequently `?CHAN_INFO_TAB` was
populated on `{event, connected}` with the pre-replay session
snapshot (`inflight_cnt=0`, `mqueue_len=N`). The actual replay then
moved the N messages from mqueue to inflight, but no event fired to
refresh `?CHAN_INFO_TAB` — so the snapshot stayed stale until the
15s `stats_timer` ticked.

### Fix

Append `?REPLY_EVENT(updated)` to the replies returned from
`emqx_channel:handle_info(continue, _)` post-replay. The catch-all
`handle_msg({event, _Other}, ...)` clause in every connection module
(`emqx_socket_connection`, `emqx_connection`, `emqx_ws_connection`)
already calls `emqx_cm:set_chan_info/2` + `emqx_cm:set_chan_stats/2`,
so this naturally refreshes the snapshot for all listener types.

The same fix path is exercised by both classic (`emqx_session_mem`)
and persistent (`emqx_persistent_session_ds`) takeover, since they
share `emqx_session:replay/3` dispatch.

### Test

`apps/emqx/test/emqx_takeover_SUITE.erl::t_chan_info_refreshed_after_takeover_replay`
reproduces the scenario (client A subscribes & disconnects, N
messages publish into the offline session, client B reconnects with
`clean_start=0` & `auto_ack=false`) and asserts that
`inflight_cnt` reflects N within 5 s — comfortably under the 15 s
stats-timer interval. Verified to fail without the fix
(`inflight_cnt=0` in the polled snapshot) and pass with it.

### Notes

The same regression exists on `release-59` (v5.10) and `release-510`
(v5.11) and should be ported in a follow-up.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_takeover_SUITE.erl
+++ b/apps/emqx/test/emqx_takeover_SUITE.erl
@@ -53,7 +53,8 @@ tc_v5_only() ->
         t_takeover_session_then_abnormal_disconnect,
         t_takeover_session_then_abnormal_disconnect_2,
         t_disconnected_at_before_connected_at_on_takeover,
-        t_disconnected_at_before_connected_at_on_discard
+        t_disconnected_at_before_connected_at_on_discard,
+        t_chan_info_refreshed_after_takeover_replay
     ].
 
 init_per_suite(Config) ->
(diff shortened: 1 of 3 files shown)


apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -918,6 +919,124 @@ t_ongoing_takeover(Config) ->
...
+t_chan_info_refreshed_after_takeover_replay(Config) ->
+    case ?config(persistence_enabled, Config) of
+        true ->
+            {skip, "Classic session only — DS path has its own stats model"};
+        _ ->
+            do_chan_info_refreshed_after_takeover_replay(Config)
+    end.
+
+do_chan_info_refreshed_after_takeover_replay(_Config) ->
+    process_flag(trap_exit, true),
```
thalesmg:
```suggestion
```
?

thalesmg (dismissed):

thalesmg (dismissed):

hjianbo (dismissed):

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14644)

PR #17386, opened 2026-05-27
fix(takeover): refresh channel stats after session replay

zmstone:
Release version: 5.9.3, 5.10.4

## Summary

Port of #17383 to release-59.

After a session takeover, the dashboard's client view and REST API
`mqueue_len` / `inflight_cnt` were stale for up to 15 seconds (the
default `mqtt.idle_timeout` / stats-timer interval).

### Root cause

Commit 9c43e0ae15 ("feat(chan): allow return `continue` to
short-circuit control flow back", first shipped in `e5.9.0`) deferred
the takeover replay to a `continue` reply that runs **after** the
`{event, connected}` reply. Consequently `?CHAN_INFO_TAB` was
populated on `{event, connected}` with the pre-replay session
snapshot (`inflight_cnt=0`, `mqueue_len=N`). The actual replay then
moved the N messages from mqueue to inflight, but no event fired to
refresh `?CHAN_INFO_TAB` — so the snapshot stayed stale until the
15 s `stats_timer` ticked.

### Fix

Append `?REPLY_EVENT(updated)` to the replies returned from
`emqx_channel:handle_info(continue, _)` post-replay. The catch-all
`handle_msg({event, _Other}, ...)` clause in both `emqx_connection`
and `emqx_ws_connection` already calls `emqx_cm:set_chan_info/2` +
`emqx_cm:set_chan_stats/2`, so this naturally refreshes the snapshot
for all listener types.

The same fix path is exercised by both classic (`emqx_session_mem`)
and persistent (`emqx_persistent_session_ds`) takeover, since they
share `emqx_session:replay/3` dispatch.

### Test

`apps/emqx/test/emqx_takeover_SUITE.erl::t_chan_info_refreshed_after_takeover_replay`
reproduces the scenario (client A subscribes & disconnects, N
messages publish into the offline session, client B reconnects with
`clean_start=0` & `auto_ack=false`) and asserts that `inflight_cnt`
reflects N within 5 s — comfortably under the 15 s stats-timer
interval. Verified to fail without the fix (`inflight_cnt=0` in the
polled snapshot) and pass with it.

### Notes

Same regression exists on `release-510` (v5.10) — a follow-up port
will be opened separately.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_takeover_SUITE.erl
+++ b/apps/emqx/test/emqx_takeover_SUITE.erl
@@ -1011,17 +1011,12 @@ wait_for_pub_client_down(#{client := [_SubClient, PubClient]} = CTX) ->
 %% t_takover_in_cluster(_) ->
 %%     todo.
 
-%% Regression test: after a session takeover triggers a replay, the
-%% ?CHAN_INFO_TAB snapshot read by the dashboard and REST API must
-%% reflect the post-replay inflight, not the (zero) pre-replay value.
-%%
-%% Pre-fix, the snapshot would stay stale until the next 15s
-%% stats_timer tick. Post-fix, REPLY_EVENT(updated) appended to the
-%% channel's `continue' reply refreshes the snapshot immediately.
+%% Regression test: chan-info stats reflect post-replay state after
+%% session takeover (instead of waiting up to 15s for the stats_timer).
 t_chan_info_refreshed_after_takeover_replay(Config) ->
     case ?config(persistence_enabled, Config) of
         true ->
-            {skip, "Classic session only — DS path has its own stats model"};
+            {skip, "Classic session only — DS uses seqno_q*/n_streams stats"};
         _ ->
             do_chan_info_refreshed_after_takeover_replay(Config)
     end.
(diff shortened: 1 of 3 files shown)


thalesmg (dismissed):

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -1010,6 +1011,121 @@ wait_for_pub_client_down(#{client := [_SubClient, PubClient]} = CTX) ->
...
+                    false
+            end
+        end,
+        PollTimeout
+    ),
+    ?assertEqual(
+        true,
+        Result,
+        {chan_stats_not_refreshed_post_replay, emqx_cm:get_chan_stats(ClientId, ChanPid)}
+    ),
```
keynslug: Feels a bit too obtuse.

```suggestion
?retry(50, 100, 
    ?assertMatch(
        #{inflight_cnt := N} when N >= NMsgs,
        maps:from_list(emqx_cm:get_chan_stats(ClientId, ChanPid)),
        chan_stats_not_refreshed_post_replay
    )
)
```

Should also be easy to get rid of one more implementation of `?retry(...)` in this module.

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -1010,6 +1011,121 @@ wait_for_pub_client_down(#{client := [_SubClient, PubClient]} = CTX) ->
...
 %%     todo.
 
+%% Regression test: after a session takeover triggers a replay, the
+%% ?CHAN_INFO_TAB snapshot read by the dashboard and REST API must
+%% reflect the post-replay inflight, not the (zero) pre-replay value.
+%%
+%% Pre-fix, the snapshot would stay stale until the next 15s
+%% stats_timer tick. Post-fix, REPLY_EVENT(updated) appended to the
+%% channel's `continue' reply refreshes the snapshot immediately.
+t_chan_info_refreshed_after_takeover_replay(Config) ->
```
keynslug: Nit. This portion of the comment feels better suited for a commit message body.

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -45,7 +45,8 @@ tc_v5_only() ->
         t_takeover_session_then_abnormal_disconnect,
         t_takeover_session_then_abnormal_disconnect_2,
         t_disconnected_at_before_connected_at_on_takeover,
-        t_disconnected_at_before_connected_at_on_discard
+        t_disconnected_at_before_connected_at_on_discard,
+        t_chan_info_refreshed_after_takeover_replay
```
keynslug: Q: Why it's v5 only?

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -1010,6 +1011,121 @@ wait_for_pub_client_down(#{client := [_SubClient, PubClient]} = CTX) ->
...
+%% ?CHAN_INFO_TAB snapshot read by the dashboard and REST API must
+%% reflect the post-replay inflight, not the (zero) pre-replay value.
+%%
+%% Pre-fix, the snapshot would stay stale until the next 15s
+%% stats_timer tick. Post-fix, REPLY_EVENT(updated) appended to the
+%% channel's `continue' reply refreshes the snapshot immediately.
+t_chan_info_refreshed_after_takeover_replay(Config) ->
+    case ?config(persistence_enabled, Config) of
+        true ->
+            {skip, "Classic session only — DS path has its own stats model"};
```
keynslug: Nit. Makes sense to encode it in `groups()`? Since we're already setting up custom matrix there.

keynslug (dismissed):

apps/emqx/test/emqx_takeover_SUITE.erl
zmstone: Good point — the test docstring is now a one-liner. Leaving the longer pre-fix/post-fix explanation in the PR description rather than amending commit history for it.

apps/emqx/test/emqx_takeover_SUITE.erl
zmstone: fixed.

apps/emqx/test/emqx_takeover_SUITE.erl
zmstone: will cover v3 too

zmstone: Updated — test now parameterizes on `mqtt_vsn` (following the existing `t_takeover/1` pattern: Session-Expiry-Interval is only attached for v5; v3 relies on the suite's default `mqtt.session_expiry_interval`). Removed from `tc_v5_only/0`. Both `mqttv3` and `mqttv5` leaves pass.

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14636)

PR #17387, opened 2026-05-28
fix(conf): ignore timestamp metadata in cluster sync status

hjianbo:
Release version:
5.10.4

## Summary

Fixes https://emqx.atlassian.net/browse/EMQX-15310

### Problem details

`emqx ctl conf cluster_sync status` compares each running node's normalized configuration to decide whether a cluster is synchronized. PR #17313 already moved this check away from raw configuration representation differences, but the checked configuration can still contain UI/API generated metadata such as `created_at` and `last_modified_at` for named actions and rules.

In the EMQX-15310 reproduction, a data import rewrote a rule and refreshed an action timestamp on some nodes. The action's effective runtime configuration was otherwise the same, but `cluster_sync status` still reported `actions` as inconsistent because the metadata timestamp differed.

### Fix approach

Normalize the configuration used by `cluster_sync status` before comparison by removing generated timestamp metadata from:

- named `actions`, `sources`, and `bridges`
- `rule_engine.rules.*.metadata`

Real configuration differences remain visible. The regression test covers both cases: timestamp-only action metadata differences are ignored, while an actual action parameter difference still reports `actions` as inconsistent.

There are no schema changes.

### Testing

- `./scripts/erlfmt -c apps/emqx_conf/src/emqx_conf_cli.erl`
- `git diff --check HEAD~2..HEAD`
- `BEFORE_REF=$(git merge-base HEAD ce/release-510) AFTER_REF=HEAD ./scripts/check-changes-filename-pattern.sh`
- `ERL_LIBS=/home/moo/Desktop/Workspace/emqx/_build/test/lib erlc -DTEST -o /tmp/emqx15310_ebin -I apps/emqx_conf/include apps/emqx_conf/src/emqx_conf_cli.erl && ERL_LIBS=/home/moo/Desktop/Workspace/emqx/_build/test/lib erl -noshell -pa /tmp/emqx15310_ebin -eval 'case eunit:test(emqx_conf_cli, [verbose]) of ok -> halt(0); error -> halt(1) end.'`

## PR Checklist

### DIFF
--- a/apps/emqx_conf/src/emqx_conf_cli.erl
+++ b/apps/emqx_conf/src/emqx_conf_cli.erl
@@ -35,6 +35,21 @@
 
 %% All 'cluster.*' keys, except for 'cluster.link', should also be treated as read-only.
 -define(READONLY_ROOT_KEYS, [rpc, node]).
+%% Generated UI/API metadata should not make status report a runtime config mismatch.
+-define(CLUSTER_SYNC_METADATA_ROOTS, [
+    actions,
+    sources,
+    bridges,
+    <<"actions">>,
+    <<"sources">>,
+    <<"bridges">>
+]).
+-define(CLUSTER_SYNC_TIMESTAMP_KEYS, [
+    created_at,
+    last_modified_at,
+    <<"created_at">>,
+    <<"last_modified_at">>
+]).
 
 -dialyzer({no_match, [load/0]}).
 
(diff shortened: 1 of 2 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14634)

PR #17403, opened 2026-05-29
fix: swagger spec corruption

zmstone:
Release version: 5.10.4

bug is caught in emqx-docs.git ci job. not released yet.

PR #17404, opened 2026-05-29
fix: swagger spec corruption

zmstone:
Release version: 6.0.3

Port of https://github.com/emqx/emqx/pull/17403 to release-60. Bug is caught in emqx-docs.git CI; not released yet, so no changelog.

## Summary

`string:titlecase/1` can return chardata (an improper list with a binary tail) even for plain binary input. JSON encoders then serialise that chardata as a JSON array of integers + strings, which crashes downstream OpenAPI tooling (e.g. Redocly's `slugify(tagName)`).

`trans_tags/1` in `emqx_dashboard_swagger.erl` now wraps the titlecased value in `iolist_to_binary/1` so each tag is always a JSON string on the wire.

A regression test in `emqx_dashboard_SUITE:t_swagger_json/1` walks every operation's `tags` array and fails on any non-binary entry.

The v6 branch's `emqx_dashboard_sso` modules already use a consistent `tags => [?TAGS]` shape (where `?TAGS` is a binary), so the SSO callsite tweak from the v5.10 PR is not needed here.

## PR Checklist

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_swagger.erl
@@ -636,7 +636,13 @@ trans_summary(Spec, Hocon, Options) ->
     end.
 
 trans_tags(Spec = #{tags := Tags}) ->
-    Spec#{tags => [string:titlecase(to_bin(Tag)) || Tag <- Tags]};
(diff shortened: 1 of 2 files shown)


qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14673)

PR #17407, opened 2026-05-29
feat(feature gates): implement feature gates

thalesmg:
Release version:
6.3.0

## Summary

Implements https://github.com/emqx/eip/pull/105

(with some modifications from discussions, at the time of writing)

## PR Checklist


apps/emqx_machine/src/emqx_machine_features.erl
```diff
@@ -0,0 +1,390 @@
...
+resolve_dependent_apps(DepFeature, KnownFeatures) ->
+    #{apps := Apps, deps := FeatDeps} = maps:get(DepFeature, KnownFeatures),
+    Apps ++
+        lists:flatmap(
+            fun(FeatDep) -> resolve_dependent_apps(FeatDep, KnownFeatures) end,
+            FeatDeps
+        ).
+
+parse_features(Str0) ->
+    %% todo: check all caps
```
thalesmg: maybe we could just use the same error for this case too.

apps/emqx_machine/src/emqx_machine_features.erl
```diff
@@ -0,0 +1,390 @@
...
+    %% todo: check empty caps ?
```
thalesmg: I don't think it would be actually problematic to have `EMQX_FEATURES=,,,`.  It'll simply be like `essential`.

apps/emqx/priv/bpapi.apps
```diff
@@ -0,0 +1,60 @@
+{emqx,emqx}.
```
zmstone, line 2: maybe add comment how this file is generated and warn not to edit manually

zmstone (commented):
expose an http API (emqx_management) to return the enabled features so the dashboard can decide what to be displayed.

apps/emqx_machine/src/emqx_machine_features.erl
```diff
@@ -0,0 +1,389 @@
...
+            apps => [
+                emqx_schema_registry,
+                %% used by external registries
+                emqx_bridge_http
+            ],
+            deps => []
+        },
+        gateways => #{
+            apps => [emqx_gateway] ++ gateway_dynamic_apps(),
+            deps => [auth]
```
hjianbo, line 140: nit: If authentication is an optional feature for the MQTT protocol, then the Gateway shouldn't strictly depend on it either.

thalesmg: yet, gateway does depend on auth.  :sweat_smile: 

https://github.com/emqx/emqx/blob/7daff925522934ad7b3a4d872075f0d34d3bcd78/apps/emqx_gateway/mix.exs#L26-L33

https://github.com/emqx/emqx/blob/7daff925522934ad7b3a4d872075f0d34d3bcd78/apps/emqx_gateway/mix.exs#L46

so we can't enable gateway without enabling auth.

thalesmg:
> expose an http API (emqx_management) to return the enabled features so the dashboard can decide what to be displayed.

yes, it is yet to be done.  would appreciate feedback on the fundamental approach first, to validate it before going too far.

zmstone:
> > expose an http API (emqx_management) to return the enabled features so the dashboard can decide what to be displayed.
> 
> yes, it is yet to be done. would appreciate feedback on the fundamental approach first, to validate it before going too far.


`GET api/v5/features`

```
{ "data_integration": "disabled",
  "auth": "enabled"
}
```

apps/emqx_machine/src/emqx_machine_features.erl
thalesmg: it might be possible to disentangle them, but best done in separate pr(s)

apps/emqx/priv/bpapi.apps
thalesmg: :checkered_flag:

apps/emqx_bpapi/src/emqx_bpapi.erl
```diff
@@ -177,3 +183,14 @@ update_minimum(API) ->
     end),
     MinVersion = lists:min(mnesia:select(?TAB, MS)),
     mnesia:write(#?TAB{key = {?multicall, API}, version = MinVersion}).
+
+filter_allowed_umbrella_applications(Data0) ->
+    {ok, Owners0} = file:consult(?MODULE:owner_applications_file()),
```
ieQu1, line 188:
```suggestion
    {ok, Owners0} = file:consult(owner_applications_file()),
```
?

apps/emqx_machine/src/emqx_machine.erl
```diff
@@ -227,3 +229,19 @@ get_emqx_vsn() ->
...
+        ?SLOG(notice, Info#{msg => "feature_gates_resolved"}),
+        ok
+    catch
+        exit:#{} = Context ->
+            ?tp(critical, "invalid_feature_specification", Context),
+            exit_loop(1)
+    end.
+
+exit_loop(ExitCode) ->
+    timer:sleep(100),
```
ieQu1, line 245: `  logger_std_h:filesync(Handler),` ?

apps/emqx_machine/src/emqx_machine.erl
```diff
@@ -227,3 +229,19 @@ get_emqx_vsn() ->
...
+    init:stop(ExitCode),
+    exit_loop(ExitCode).
```
ieQu1, line 247: That code looks pretty strange. Please add a comment explaining why is it like that. Use `erlang:halt` instead of looping?

apps/emqx/test/emqx_cth_suite.erl
```diff
@@ -472,6 +472,8 @@ verify_clean_suite_state(#{work_dir := WorkDir}) ->
 clean_suite_state() ->
     _ = emqx_schema_hooks:erase_injections(),
     _ = emqx_config:erase_all(),
+    emqx_machine_features:clear_features(),
```
keynslug, line 475: Should probably be called in `emqx_machine_app:stop/1`?

apps/emqx_machine/src/emqx_machine_boot.erl
```diff
@@ -259,3 +272,28 @@ find_loops(G) ->
...
+    case atom_to_list(App) of
+        "emqx" ->
+            true;
+        "emqx_http_lib" ->
+            false;
+        "emqx_" ++ _ ->
+            true;
+        _ ->
+            false
+    end.
```
keynslug, line 302: Nit. To me it seems odd that with all the static information we have (e.g. reboot lists) there's still a need for this function. I naturally wonder what's missing there.

apps/emqx_machine/src/emqx_machine_features.erl
```diff
@@ -0,0 +1,389 @@
...
+                emqx_bridge,
+                emqx_rule_engine,
+                emqx_postgresql,
+                emqx_mysql,
+                emqx_oracle,
+                emqx_redis,
+                emqx_s3,
+                emqx_mongodb
+            ] ++ data_integration_dynamic_apps(),
+            deps => [schema_registry]
```
keynslug, line 120: I'm somewhat wary of _one more_ place to declare intra-project dependencies (and likely accumulate inconsistencies). If it's about knowing which applications should be started then this information is already there? Or is it about knowing which features were _automatically_ enabled?

apps/emqx_machine/src/emqx_machine_boot.erl
```diff
@@ -171,10 +174,20 @@ is_app(Name) ->
...
+
+%% sorted, full list of apps to reboot; not feature-filtered
+full_sorted_reboot_apps() ->
+    Apps = reboot_apps_with_deps(),
+    sorted_reboot_apps(Apps).
+
+%% internal export only for companion module `emqx_machine_features`
+reboot_apps_with_deps() ->
     RebootApps = reboot_apps(),
     Apps0 = [{App, app_deps(App, RebootApps)} || App <- RebootApps],
```
keynslug, line 191: Q: This suggests that _all_ applications' code are loaded irrespective of the enabled features. Is it a non-goal for this PR to avoid loading anything belonging to disabled features?

apps/emqx_bpapi/src/emqx_bpapi.erl
```diff
@@ -177,3 +183,14 @@ update_minimum(API) ->
...
+    APIsToOwners = maps:from_list(Owners0),
+    lists:filter(
+        fun({API, _Vsn}) ->
+            App = maps:get(API, APIsToOwners),
+            emqx_machine_features:is_umbrella_application_enabled(App)
```
keynslug, line 193: This introduces unexpected runtime dependency. With all other BPAPI-related changes in this PR, to me it honestly suggests that it's time to push responsibility of announcing APIs to the applications, as part of their startup routines. Also partly because right now APIs are announced _before_ applications providing them are even started, which brings its own troubles.

Perhaps a middle ground is to make `emqx_machine` responsible for the announcement itself, and make sure it's done the same way in `emqx_cth_cluster`.

---

I'm not suggesting to address this issue in this PR specifically, but it already touches BPAPI stuff so there's an opportunity to do it nicer way as part of the effort.

apps/emqx_machine/src/emqx_machine_features.erl
```diff
@@ -0,0 +1,389 @@
...
+
+%%------------------------------------------------------------------------------
+%% API
+%%------------------------------------------------------------------------------
+
+info() ->
+    format_info(features()).
+
+%% begin core apps
+is_umbrella_application_enabled(emqx) ->
```
keynslug: Perhaps do `lists:member(App, core_apps()) orelse ...` instead to avoid duplication / accidental inconsistencies?

thalesmg: that would introduce the need to constantly update this function every time an application is created or removed, I think.  :sweat_smile: 

since we have the convention of prefixing our umbrella apps with `emqx_`, this seemed a bit more "robust" (the only known exception being `emqx_http_lib`).

apps/emqx_bpapi/src/emqx_bpapi.erl
thalesmg: yes.  at some point, I attempted to do that initially, but it started to grow a bit.  also, it is a bit more complex, when considering issues with restarting applications, ensuring they clear old entries that belonged to them, and such things.  since the proposal was to try to stay simple if possible, I dropped that idea.

apps/emqx_machine/src/emqx_machine_boot.erl
thalesmg: I didn't change the release to interactive code loading.  the idea was to keep it simple, rather than fiddling with the release in a deeper level.

apps/emqx_machine/src/emqx_machine_features.erl
thalesmg: no, not about restarting _applications_, but just enabling _features_.  restart stuff is left to `mix.exs` and `sorted_reboot_apps`.

this function is to know which features to automatically enable.  for example, if you enable `gateways`, since it depends on the `auth` feature, the latter is automatically enabled.

apps/emqx_machine/src/emqx_machine_boot.erl
thalesmg: reboot lists does not contain only umbrella apps.  there are things like observer_cli, mria, emqx_http_lib, etc.

here, the idea is to filter only those apps in `apps/*`.

apps/emqx/test/emqx_cth_suite.erl
thalesmg: since this is a read only property set only once, it probably could survive application resets.

apps/emqx_machine/src/emqx_machine.erl
thalesmg: this is the same approach we have taken here for a while:

https://github.com/emqx/emqx/blob/68a0704d3c4078b6b26b8b1ae5a1316edbfdf2c9/apps/emqx_conf/src/emqx_conf_app.erl#L43-L46

apps/emqx_machine/src/emqx_machine.erl
thalesmg: not sure what this means.  is this some suggestion to ensure the log is flushed?

apps/emqx_bpapi/src/emqx_bpapi.erl
thalesmg: this takes the same approach for the versions file:

https://github.com/emqx/emqx/blob/68a0704d3c4078b6b26b8b1ae5a1316edbfdf2c9/apps/emqx_bpapi/src/emqx_bpapi.erl#L88

... so that it can be tested and doesn't break this suite, like here:

https://github.com/emqx/emqx/blob/68a0704d3c4078b6b26b8b1ae5a1316edbfdf2c9/apps/emqx_bpapi/test/emqx_bpapi_SUITE.erl#L36-L37

apps/emqx_machine/src/emqx_machine.erl
thalesmg: @zmstone do you recall if there's a specific reason we do it like so instead of `halt/1`?

I suppose the original reason was to have time to flush logs/buffers, and have a more graceful app shutdown?

for the latter, though, it seems in OTP 27 we have some new flush options.

apps/emqx_machine/src/emqx_machine_features.erl
keynslug: Why? Isn't `is_umbrella_application_enabled` already has a clause per element in `core_apps/0` list?

apps/emqx_machine/src/emqx_machine_boot.erl
keynslug: Yes I get the idea. My (mild) concerns are:
1. One more "flavor" of applications to keep track of. E.g. there are "business" applications in the reboot lists, but for some reason it's not the same as "umbrella" applications, and the reason why they are different is not clear. I'd personally welcome if those are unified, of course unless there are strong reasons to distinguish them.
2. Robustness. E.g. what would happen is someone brings `emqx_some_dep` external dependency later into the project.

apps/emqx_machine/src/emqx_machine_features.erl
thalesmg: Oh, sorry, I think I missed your earlier point.

I made one clause for each because I thought it would be better performance-wise to have those clauses for each core app.  If there was a simple way to have a single list expand at compile time to those clauses (as would be with elixir macros), it would indeed be the best of both worlds.

The consistency is tested in `emqx_machine_features_tests` by checking in different scenarios that core apps are always enabled.  (granted, it's somewhat indirect, and a weaker assertion, as it's not bidirectional/if-and-only-if for the core apps)

Hopefully, the core app set won't change so much as the non-core set.

apps/emqx_bpapi/src/emqx_bpapi.erl
keynslug: I see thanks, I can imagine.

If the goal is to stay simple, then the simplest approach to me is to make it operator's concern: basically, say that running heterogeneous clusters is not supported.

Otherwise, the "middle ground" option looks more or less as simple, curious what do you think about it.

apps/emqx_machine/src/emqx_machine_boot.erl
thalesmg: "umbrella applications" are, as mentioned, those under the `apps/*` in our repo, which is nomenclature that mix and (I believe) rebar3 uses.  I don't quite see why they should be the same as reboot lists, as it's more linked to application startup order/inclusion in the release rather than defining what is an umbrella application.

We differentiate umbrella apps because those are the "root" apps under our project that are enabled and started or not.  Their dependencies should be always enabled and not accidentally forbidden from starting, hence the differentiation.

`emqx_some_dep` would be troublesome, indeed...  Ideally, I'd like for a compile-time way to create a list of umbrella apps, which would far more robust.

apps/emqx_bpapi/src/emqx_bpapi.erl
thalesmg: I'm not entirely sure how far we can go with the constraints.  From what I understood, it seems like one use case mentioned in the EIP was for an operator to restart a node with a different set of features.

@zmstone WDYT?

apps/emqx_machine/src/emqx_machine_boot.erl
thalesmg: maybe we could generate such list in a similar fashion to `deps.txt` and use ci to ensure it's consistent

apps/emqx_machine/src/emqx_machine_features.erl
keynslug: > I thought it would be better performance-wise

That's no doubt true. I get the point, but: does it make that much difference here? Assuming that this function gets called between 10 and 100 times during startup, and that's it.

apps/emqx_machine/src/emqx_machine_boot.erl
keynslug: > application startup order/inclusion in the release rather than defining what is an umbrella application

Heh, on the other hand to me this separation is confusing. 😅 E.g. we introduce "features gates" that are designed to enable or disable business features, and there's "business" applications in `reboot_lists.eterm`, but we for some reason still need another concept for this to work.

> maybe we could generate such list

Yeah that would be good. Especially if we can consolidate it with `reboot_lists.eterm`, `?BASIC_REBOOT_APPS` and similar macrodefs and turn into a single source of "what EMQX is comprised of?".

apps/emqx/test/emqx_cth_suite.erl
keynslug: My primary point is: as `emqx_machine` application owns this feature gates state, it should clean it during application stop.

apps/emqx_machine/src/emqx_machine_boot.erl
keynslug: The latter of course looks out of scope of this PR. Regard it more as a discussion.

thalesmg: I honestly am having trouble to see what is confusing with "umbrella apps" and "business apps" in reboot lists.  :sweat_smile: 

"umbrella apps" have already been defined above (and is a common nomenclature in mix/rebar3 lingo).  We are interested in those because here, in our project and repo, we tend to place features in one or more separate apps.  Hence our interest in mapping "features" to one or more umbrella apps.

Reboot lists is what controls what goes into the release, and the order in which they are started (barring some reordering due to declared dependencies).  These may (and currently do) include apps outside our umbrella (e.g.: quicer, observer_cli).  

To me, they seem clearly different, specially since umbrella app is simply a definition.  I really don't see the need to mix them together in the reboot list file.

Could we create a field that lists umbrella in reboot lists?  Maybe we could, but:
- It would likely increase the complexity even further.  We are currently relying on the existing logic from machine boot to construct the sorted reboot app list, we just filter the input to the existing logic.
- If we add a list of umbrella apps to `reboot_lists.eterm`, we then add more possibility of inconsistencies.  Instead of relying on the definition of what is an umbrella app (which is simply the contents of `apps/*` with mix project files, with the caveats already stated before), we add an extra thing to update, and then have to mix that with the original intent of reboot lists: rough ordering of apps.  One place lists the set of umbrella apps, with no ordering, and the other must reference those same apps, but with the ordering semantics.

apps/emqx/test/emqx_cth_suite.erl
thalesmg: it's still immutable, but ok.  changed to clear it when stopping.

apps/emqx_machine/src/emqx_machine_boot.erl
thalesmg: Maybe this could help with my confusion: If you're _not_ suggesting to make reboot lists and umbrella app lists the same thing (in which case I have misunderstood your proposal), could you please sketch the shape of reboot_lists.eterm you envision?  Which fields, a content sample, etc. ?

keynslug: Thanks for the explanation. This is out of scope of this PR so let's keep it as is.

apps/emqx_machine/src/emqx_machine_features.erl
```diff
@@ -0,0 +1,344 @@
...
+    end.
+
+format_info(#{preset := full}) ->
+    KnownFeatures = known_features(),
+    KnownNames = maps:keys(KnownFeatures),
+    #{
+        preset => full,
+        enabled => lists:sort(KnownNames),
+        disabled => [],
+        bundled => [file_transfer, gcp_device, exhook, opentelemetry]
```
zmstone: https://github.com/emqx/eip/pull/105/changes#r3388360145

apps/emqx_machine/src/emqx_machine_features.erl
```diff
@@ -0,0 +1,344 @@
...
+                            hint => did_you_mean(Str, KnownFeatures)
+                        })
+                end
+            end,
+            #{
+                preset => custom,
+                allowed_apps => #{},
+                enabled_features => #{}
+            },
+            string:tokens(Str0, ",")
```
zmstone:
```suggestion
            string:tokens(Str0, ", ")
```

apps/emqx_machine/src/emqx_machine_features.erl
zmstone: which error?

thalesmg: https://github.com/zmstone/eip/blob/7395761c2bbf56d25de4cc7a7dcf2009fcfe6860/active/0037-feature-gates.md?plain=1#L231-L232

apps/emqx_machine/src/emqx_machine.erl
zmstone: that's right, this is the trick to let the vm log everything.
logs, stdio prints.

apps/emqx_bpapi/src/emqx_bpapi.erl
zmstone: it would be too disruptive if one must perform a full cluster restart to change a feature flag.

apps/emqx_machine/src/emqx_machine_features.erl
zmstone: you mean unknown ALLCAPS result in unknown_feature error?
i guess that's fine.

apps/emqx_machine/src/emqx_machine_features.erl
zmstone: true, `string:tokens(",,", ", ")` results in `[]`.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16072)

PR #17420, opened 2026-06-01
Add session buffer observability

### LINKED ISSUE #17396: Observability and byte-volume limits for session mqueue/inflight

Today EMQX bounds session message buffering by *message count*, not by *bytes*:

- `mqueue_len` (default `1000`) caps the number of pending messages; overflow drops messages per the configured discard strategy.
- Process mailbox length limit triggers a force shutdown when exceeded.
- There is **no limit on the byte volume held in the mqueue + inflight window**.

Because Erlang stores large binaries off-heap as refc binaries, a session can hold a small *count* of messages while still pinning a very large amount of memory. With MB-level payloads it is easy to retain hundreds of MB, or even GB, of payload data in a single session without ever tripping the heap-size or mailbox-length safeties.

This becomes acutely visible during **session takeover**: the previous owner ships the full session state (including buffered payloads) to the new owner over the Erlang distribution channel. When many such takeovers happen concurrently (e.g. node rolling restart, network blip, mass reconnect), the distribution port can saturate, producing `busy_dist_port` and, in the worst case, choking the VM.

Proposed scope for this issue (observability only):

1. Expose per-session and aggregate metrics for the **byte size** held in mqueue and inflight (refc binary bytes + term heap bytes).
2. Add an alarm (or a top-N endpoint) so operators can quickly identify sessions retaining unusually large payload volumes.

### Why is this needed?

Operators currently have no first-class signal for \"this session is holding X MB of payload.\" When a takeover storm causes \`busy_dist_port\` or distribution back-pressure, there is no straightforward way to attribute it to specific sessions before or after the fact. Metrics + alarms make this investigable in production without ad-hoc tracing.

### Out of scope / follow-up

A configurable byte-volume cap for mqueue+inflight (and the associated overflow policy) is a behavior change that belongs in the next minor release (target **v6.3**). It will be proposed and discussed as an EIP in https://github.com/emqx/eip, with this issue tracking only the observability piece.

hjianbo:
Fixes #17396

Design discussion: emqx/eip#110

Release version:
6.3.0

## Summary

This PR adds lightweight observability for payload bytes retained by MQTT sessions.

User-facing changes:

- The client list APIs expose `total_payload_bytes` for each session. For in-memory sessions, this is the MQTT payload bytes currently retained by the session mqueue and inflight window. Durable sessions report `0` because their backlog is not retained in the channel process in the same way.
- `sysmon.session.buffered_payload_high_watermark` controls throttled warning logs for a single in-memory session whose `total_payload_bytes` is above the configured threshold. The default is `0`, which disables the warning log.
- `emqx ctl session-top --out <File> [--count <K>] [--sort <mqueue_length|total_payload_bytes>]` scans the cluster on demand and writes a CSV file with `clientid,pid,node,mqueue_length,total_payload_bytes,inflight_count`. `--count` defaults to `10` and is capped at `1000`; `--out` is required to avoid large console output.

Implementation notes:

- In-memory sessions expose payload bytes through the existing channel stats cache. Mqueue payload bytes are derived from current mqueue contents when session stats are refreshed, and inflight payload bytes are derived from the bounded inflight window. This keeps the in-memory mqueue record shape compatible with existing session state; the mqueue part is O(mqueue_len) during stats refresh. APIs and `session-top` read cached stats from `emqx_channel_info` instead of asking every session process to inspect buffered messages.
- The warning check piggybacks on the existing channel stats update path and is O(1) per stats update after the stats are prepared. No EMQX alarms are created.
- `session-top` is handled by the named `emqx_session_buffer_mon` process. Each node scans local `emqx_channel_info` with a small batch sleep and keeps only the bounded top-K rows in memory before the caller merges cluster results.
- The top-K heap skips rows that cannot enter the current top-K set without deleting and reinserting the smallest row, and uses Erlang term ordering for tie-break keys instead of converting keys to binaries on every scanned row.
- This PR intentionally does not add node-level stats, Prometheus gauges, a management top-N endpoint, or periodic background aggregation.

Manual performance check:

- Used `mix run --no-start` to populate a synthetic `emqx_channel_info` cache and call `emqx_session_buffer_mon:scan_local/2` directly. This measures scanner cost over cached channel stats; real-node results can vary with actual ETS row size and node load.
- Direct local scan results after the top-K heap optimization:

| cached rows | top 10 by `total_payload_bytes` | top 1000 by `total_payload_bytes` | top 10 by `mqueue_length` | ETS table size |
|---:|---:|---:|---:|---:|
| 100,000 | 305 ms | 384 ms | 302 ms | 16.8 MB |
| 1,000,000 | 3.40 s | 3.79 s | 3.06 s | 167.9 MB |
| 2,000,000 | 6.40 s | 7.07 s | 6.16 s | 335.7 MB |

- Full async worker path with 2,000,000 cached rows and `--count 1000` wrote 1000 CSV rows in about 7.6 s when sorting by `total_payload_bytes`, and about 6.7 s when sorting by `mqueue_length`. The benchmark data setup took about 10 s and is not part of the runtime command cost.

Checked locally with `git diff --check`, `make test-compile`, `./scripts/apps-version-check.exs`, focused CT cases for message payload size, mqueue payload sizing, memory session stats, management API exposure, `session-top` CSV output, session-buffer warning checks, and session-buffer log throttling, plus the manual scanner benchmark above.

## PR Checklist

### DIFF
--- a/apps/emqx/priv/bpapi.versions
+++ b/apps/emqx/priv/bpapi.versions
@@ -96,6 +96,7 @@
 {emqx_router,1}.
 {emqx_rule_engine,1}.
 {emqx_schema_registry_http_api,1}.
+{emqx_session_buffer,1}.
 {emqx_setopts,1}.
 {emqx_shared_sub,1}.
 {emqx_slow_subs,1}.
--- a/apps/emqx/src/emqx_cm.erl
+++ b/apps/emqx/src/emqx_cm.erl
@@ -275,7 +275,13 @@ set_chan_stats(ClientId, Stats) when ?IS_CLIENTID(ClientId) ->
 set_chan_stats(ClientId, ChanPid, Stats) when ?IS_CLIENTID(ClientId) ->
     Chan = {ClientId, ChanPid},
     try
-        ets:update_element(?CHAN_INFO_TAB, Chan, {3, Stats})
+        case ets:update_element(?CHAN_INFO_TAB, Chan, {3, Stats}) of
+            true ->
+                ok = emqx_session_buffer_mon:maybe_log(ClientId, ChanPid, Stats),
+                true;
+            false ->
+                false
+        end
     catch
         error:badarg -> false
     end.
--- a/apps/emqx/src/emqx_message.erl
+++ b/apps/emqx/src/emqx_message.erl
@@ -184,6 +184,8 @@ estimate_size(#message{topic = Topic, payload = Payload}) ->
     FixedHeaderSize + VarLenSize + TopicLengthSize + TopicSize + PacketIdSize + PayloadSize.
 
 -spec payload_size(emqx_types:message()) -> non_neg_integer().
+payload_size(#message{payload = undefined}) ->
+    0;
 payload_size(#message{payload = Payload}) ->
     iolist_size(Payload).
 
--- a/apps/emqx/src/emqx_message.erl
+++ b/apps/emqx/src/emqx_message.erl
@@ -27,7 +27,8 @@
     topic/1,
     payload/1,
     timestamp/1,
-    timestamp/2
+    timestamp/2,
+    payload_size/1
 ]).
 
 %% Flags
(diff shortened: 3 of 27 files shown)


hjianbo:
Superseded by #17493. Keeping the release-60 PR as the single source because release-60 will sync forward to release-63.

PR #17424, opened 2026-06-01
260601 clean stale cm registry record

zmstone:
Release version: 5.10.4

## Summary

session/channel registry records may leak if the 'delete' operation to the registry table is lost due to for example cluster network partition.
this results in consiquence:

A stale channel/session pid is never deleted even after cluster network split healed.

The fix is to delete stale records (pid is no longer alive) during takeover/kick.

NOTE: for random client IDs, the records may still leak, it will require a gc process to periodically scan the registry.

## PR Checklist


ieQu1 (dismissed):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14670)

PR #17426, opened 2026-06-01
fix(jt808): allow empty registry/authentication URLs when allow_anonymous is true

JimMoen:
## Summary

Fix the JT/T 808 gateway schema validation to allow empty or omitted `registry` and `authentication` URLs when `allow_anonymous` is set to `true`.

## Problem

When updating the JT/T 808 gateway config via `PUT /api/v5/gateways/jt808` with `"registry": ""` and `allow_anonymous = true`, the API returns a 400 error. This is because the `not_empty` validator was applied to both `registry` and `authentication` fields regardless of the `allow_anonymous` setting.

However, when `allow_anonymous = true`, these URLs are completely ignored by the runtime (`emqx_jt808_auth:init/1` hardcodes `registry = undefined, authentication = undefined`), so rejecting empty strings is unnecessarily restrictive.

## Changes

- **Schema fix**: Split `fields_reg_auth_required/1` into two clauses:
  - `fields_reg_auth_required(true)` — keeps `not_empty` validator + `required => true` (for `allow_anonymous = false`)
  - `fields_reg_auth_required(false)` — no validator, `required => false` (for `allow_anonymous = true`)

- **Tests**: 6 new CT test cases covering:
  - `allow_anonymous=true` with empty string URLs → succeeds
  - `allow_anonymous=true` with omitted URLs → succeeds
  - `allow_anonymous=false` with empty string URLs → still fails validation

- **Changelog**: `changes/ee/fix-17426.en.md`

### DIFF
--- a/apps/emqx_gateway_jt808/src/emqx_gateway_jt808.app.src
+++ b/apps/emqx_gateway_jt808/src/emqx_gateway_jt808.app.src
@@ -1,7 +1,7 @@
 %% -*- mode: erlang -*-
 {application, emqx_gateway_jt808, [
     {description, "JT/T 808 Gateway"},
-    {vsn, "0.1.5"},
+    {vsn, "0.1.6"},
     {registered, []},
     {applications, [kernel, stdlib, emqx, emqx_gateway]},
     {env, []},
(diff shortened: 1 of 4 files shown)


hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15518)

PR #17437, opened 2026-06-02
refactor(prometheus): require authentication for scrape endpoints by default

zmstone:
Release version: 6.3.0

## Summary

Flip the schema default for `prometheus.enable_basic_auth` from `false` to
`true` on `release-63` so the Prometheus scrape endpoints under
`/api/v5/prometheus/*` require authentication out of the box. This closes a
secure-by-default gap where a fresh install exposed cluster, ACL/AuthN, and
data-integration metrics without credentials unless the operator explicitly
opted in.

### Behavior change

Operators who scrape these endpoints without credentials will start receiving
`401` after upgrade. They have two paths to choose:

* Recommended: configure the scraper with a dedicated `monitoring`-scope API
  key and use Bearer auth.
* Opt-out: set `prometheus.enable_basic_auth = false` in `cluster.hocon` /
  `emqx.conf` to restore the prior unauthenticated behavior.

Legacy-shape configs (`prometheus.enable = true`, …) are unaffected because the
legacy schema has no `enable_basic_auth` field at all; their runtime read of
`emqx_config:get([prometheus, enable_basic_auth], false)` continues to return
`false`. The legacy-to-recommend converter in
`emqx_prometheus_config:to_recommend_type/1` also still hardcodes `false` so
that operators migrating a legacy config see no behavior change at the API
surface.

### Files touched

* `apps/emqx_prometheus/src/emqx_prometheus_schema.erl` — schema default flip.
* `apps/emqx_prometheus/src/emqx_prometheus_api.erl` — recommend-setting Swagger
  example updated to match the new default.
* `rel/i18n/emqx_prometheus_schema.hocon` — `enable_basic_auth.desc` and the
  `prometheus.desc` blurb note the change and the opt-out.
* `apps/emqx_prometheus/test/emqx_prometheus_api_SUITE.erl` — new CT case
  `t_stats_default_basic_auth_enabled` confirms no-auth → 401, auth → 200, and
  the explicit opt-out path. `t_stats_auth_api` now resets the config at the
  end so it does not leak `enable_basic_auth=true` into subsequent tests.
* `changes/ee/breaking-17437.en.md` — user-facing changelog entry in the
  `breaking` bucket.

## Branch scope

This PR lands on `release-63` only. Earlier maintenance branches
(`release-60`/`release-61`/`release-62`) and the v5 line still default to
`false`. If the change should also land on older minors, that's a separate
port — call it out and we can open follow-up PRs.

## CS heads-up

Worth flagging to the CS team before 6.3.0 ships: any Enterprise customers
running customer-side Prometheus scrapers without credentials will need to
update their scraper config or set the opt-out flag.

## Local validation

* `make fmt-diff` — clean.
* `./scripts/elvis-check.sh \$(git merge-base origin/release-63 HEAD)` — clean.
* Schema default verified end-to-end via an `erl` shell against the freshly
  built beam on the previous (release-60) base — returns `true`. The same
  schema source applies on release-63.
* The `emqx_prometheus` CT suite could not be exercised locally: the host
  already has an EMQX listening on `1883`/`18083` (port collision in
  `init_per_suite`) and the docker runner hit a pre-existing quicer NIF
  mismatch unrelated to this change. CI will exercise the suite.

## PR Checklist

### DIFF
--- a/apps/emqx_prometheus/src/emqx_prometheus_api.erl
+++ b/apps/emqx_prometheus/src/emqx_prometheus_api.erl
@@ -379,7 +379,7 @@ recommend_setting_example() ->
     {Summary, #{
         summary => Summary,
         value => #{
-            enable_basic_auth => false,
+            enable_basic_auth => true,
             push_gateway => #{
                 interval => <<"15s">>,
                 method => put,
--- a/apps/emqx_prometheus/src/emqx_prometheus_schema.erl
+++ b/apps/emqx_prometheus/src/emqx_prometheus_schema.erl
@@ -36,7 +36,7 @@ fields(recommend_setting) ->
             ?HOCON(
                 boolean(),
                 #{
-                    default => false,
+                    default => true,
                     required => true,
                     importance => ?IMPORTANCE_HIGH,
                     desc => ?DESC(enable_basic_auth)
--- a/apps/emqx_prometheus/test/emqx_prometheus_api_SUITE.erl
+++ b/apps/emqx_prometheus/test/emqx_prometheus_api_SUITE.erl
@@ -347,7 +347,47 @@ t_stats_auth_api(_) ->
     Auth = emqx_mgmt_api_test_util:auth_header_(),
     Headers = [Auth | accept_json_header()],
     request_stats(Headers, Auth),
+    {ok, _} = emqx:update_config([prometheus, enable_basic_auth], false),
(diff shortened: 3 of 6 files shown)


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16068)

PR #17442, opened 2026-06-02
fix(backup): dashboard / api-key viewers cannot download backups

zmstone:

## Summary

PRs #17169 / #17173 added a sensitive-table guard so that API-key-authenticated callers cannot `POST /data/export` or `POST /data/import` archives containing `dashboard_users` or `api_keys`. The stored-backup **download** path was missed: an API key with `?SCOPE_SYSTEM`, a dashboard viewer, or a namespaced administrator could `GET /api/v5/data/files/<file>` and download an archive produced by a global administrator. The archive can contain salted password hashes, MFA / TOTP material, API key records, and `cluster.hocon`.

This PR closes the download endpoint with a role gate:

- `GET /data/files/:filename` — now requires the global dashboard administrator (`jwt_token` + role `administrator` + namespace `?global_ns`). API keys (any scope), dashboard viewers, and namespaced administrators are rejected with 403.
- `GET /data/files` (listing) — **unchanged**. Filenames / sizes / timestamps are not credential material; the dangerous operation is the download, which is gated separately. Viewers and namespaced administrators can still see what exists in the backup directory.
- `DELETE /data/files/:filename` — unchanged; no credential bytes leave the node on a delete.

No archive content inspection is performed. A peek per request would pay a tar-table cost and still has to make the same role decision when sensitive content is found, so the role check alone is the simpler invariant.

Relevant files:
- `apps/emqx_management/src/emqx_mgmt_api_data_backup.erl`
- `apps/emqx_management/test/emqx_mgmt_api_data_backup_SUITE.erl`

CT coverage added:
- `t_download_api_key_forbidden`, `t_download_viewer_forbidden`, `t_download_ns_admin_forbidden` — 403 for each non-admin caller.
- `t_download_global_admin_allowed` — 200 for the global administrator.
- `t_list_files_viewer_allowed` — viewer and namespaced administrator can still list.
- Existing `test_file_op(get)` and `upload_backup_test`'s 404 probe switched to dashboard admin auth, since GET is now admin-only.

## PR Checklist
- [x] Change log for changes visible by users has been added to `changes/ee/fix-17440.en.md`


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14713)

PR #17451, opened 2026-06-02
fix(backup): close sensitive-table download bypass

zmstone:
Release version: 5.8.11, 5.9.3, 5.10.4

## Summary

Port of #17442 (release-60 follow-up to #17169 / #17173) to release-58.

PR #17169 added a sensitive-table guard so that API-key-authenticated callers cannot \`POST /data/export\` or \`POST /data/import\` archives containing \`dashboard_users\` or \`api_keys\`. The stored-backup **download** path (\`GET /api/v5/data/files/<file>\`) was missed: an API key with \`?SCOPE_SYSTEM\` or a dashboard viewer could download an archive produced by a dashboard administrator and exfiltrate salted password hashes, MFA / TOTP state, and API key records offline.

This PR closes the download endpoint with a role gate:

- \`GET /data/files/:filename\` — now requires the dashboard administrator role. API keys (any scope) and dashboard viewers are rejected with 403.
- \`GET /data/files\` (listing) — **unchanged**. Filenames / sizes / timestamps are not credential material; the dangerous operation is the download, which is gated separately. Viewers can still see what exists in the backup directory.
- \`DELETE /data/files/:filename\` — unchanged; no credential bytes leave the node on a delete.

No archive content inspection is performed. A peek per request would pay a tar-table cost and still has to make the same role decision when sensitive content is found, so the role gate alone is the simpler invariant.

To carry the role to the data backup handler, \`emqx_dashboard:authorize/2\` now reads the bearer-token JWT's \`extra.role\` and surfaces it in \`auth_meta.role\`. API key callers carry no role, so they fall through to the deny side of the gate naturally.

Relevant files:
- \`apps/emqx_dashboard/src/emqx_dashboard.erl\` — populate \`auth_meta.role\` for bearer-token requests.
- \`apps/emqx_management/src/emqx_mgmt_api_data_backup.erl\` — gate the download.
- \`apps/emqx_management/test/emqx_mgmt_api_data_backup_SUITE.erl\` — new CT cases (\`t_download_api_key_forbidden\`, \`t_download_viewer_forbidden\`, \`t_download_admin_allowed\`, \`t_list_files_viewer_allowed\`); existing \`test_file_op(get)\` and \`upload_backup_test\` 404 probe switched to dashboard admin auth, since GET is now admin-only.

## PR Checklist
- [x] Change log for changes visible by users has been added to \`changes/ee/fix-17451.en.md\`


thalesmg (approved):

hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15520)

PR #17463, opened 2026-06-03
fix(license): API serves monthly session HWM history symmetric with CLI

zmstone:

## Summary

Backport of #17459 to `release-60`. Clean cherry-pick.

The session high-watermark history REST API (`GET /api/v5/license/session_hwm_history`) ignored the `period=monthly` query parameter and always returned daily rows, while the CLI (`emqx ctl license history`) correctly defaulted to monthly.

Two issues:

- The handler read `period` and `limit` from the query string using atom keys, but minirest passes parsed query parameters as binary keys. The atom lookup always missed and fell back to `daily`. Regression from the `check_schema => true` switch.
- The default was `daily` and the limit was hardcoded to 30, both inconsistent with the CLI which defaults to monthly with effectively no row cap.

Fix: read query params with binary keys (values are already coerced to the schema types — atom for the enum, integer for the limit), default to `monthly`, and apply per-period limit defaults (30 for daily, 24 for monthly — matches the 24-month retention window).

i18n description for `param_history_period` / `param_history_limit` updated to match the new defaults.

No changelog: the session HWM history feature has not been released yet.

## PR Checklist

### DIFF
--- a/apps/emqx_license/src/emqx_license_http_api.erl
+++ b/apps/emqx_license/src/emqx_license_http_api.erl
@@ -121,14 +121,13 @@ schema("/license/session_hwm_history") ->
                     hoconsc:mk(hoconsc:enum([daily, monthly]), #{
                         in => query,
                         required => false,
-                        default => daily,
+                        default => monthly,
                         desc => ?DESC("param_history_period")
                     })},
                 {limit,
                     hoconsc:mk(pos_integer(), #{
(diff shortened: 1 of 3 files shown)


keynslug (approved):

yanzhiemq:
[TestExecution](https://emqx.atlassian.net/browse/QA-14733)

PR #17467, opened 2026-06-03
fix(rule_engine): refresh SSRF cache on parent-path config updates

zmstone:
Release version: 6.0.3

## Summary

`PUT /api/v5/configs?mode=merge` and `PUT /api/v5/configs/rule_engine` both update the config tree at the parent `[rule_engine]` path. The framework propagates into sub-handlers using the `propagated_post_config_update` callback name, but `emqx_rule_engine_schema` only exported `post_config_update/5` — so the SSRF policy cache in `persistent_term` was never refreshed when the flag was toggled through these REST endpoints, leaving loopback URLs to slip past the connector-side check.

Fix: export and define `propagated_post_config_update/5` for `[rule_engine, ssrf]` and `[rule_engine, jq_implementation_module]`, both forwarding to the existing `post_config_update/5`. Calls are guarded so that propagation entries with no carried sub-config (e.g. deprecated `jq_implementation_module` absent from the atomized map) are skipped.

Regression test: `apps/emqx_rule_engine/test/emqx_rule_engine_ssrf_SUITE.erl` covers both the buggy parent-path and the working sub-path updates and asserts the post-update cache state.

## PR Checklist
- [ ] no change log since the bug is not released yet.

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_engine_schema.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_engine_schema.erl
@@ -16,6 +16,7 @@
     fields/1,
     desc/1,
     post_config_update/5,
+    propagated_post_config_update/5,
     rule_engine_settings/0
 ]).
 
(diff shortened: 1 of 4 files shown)


qzhuyan (approved):

yanzhiemq:
[TestExecution](https://emqx.atlassian.net/browse/QA-14735)

Issue #17413, opened 2026-05-30
perf(iotdb rest): health check runs `show databases` on every tick


zmstone:
## Summary

The IoTDB connector (REST driver mode) runs `show databases` on every health-check tick via the auth-check path, in addition to the TCP-level `check_ping_restapi`. Returns one row per IoTDB storage group (a.k.a. database).

## Where

`apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl:521-…` (`check_auth_restapi/2`).

```erlang
%% pseudo:
POST /rest/v2/nonQuery  body={"sql":"show databases"}
```

Note: this is only the REST driver. The thrift driver uses `iotdb:ping(Conn)` which is bounded.

## Impact

The row count is operator-controlled (one per storage group / database), not customer-data-proportional, so cost stays small in practice. But:

* The list scales linearly with the number of storage groups the IoT deployment partitions data into. Deployments with many sites / regions can have hundreds.
* It runs on every health-check tick in addition to the cheaper TCP-level ping that already exists.

## Suggested fix

Two options, in preference order:

1. Replace `show databases` with `SELECT 1` via the same REST endpoint. Bounded to a single-row response regardless of storage-group count.
2. Rely on `check_ping_restapi` alone for liveness and move the auth check into startup only. Auth credentials don't drift at runtime; re-checking them on every health tick is the wrong cadence.

## Related

Part of the health-check audit following d94cbde4. YELLOW (bounded but more expensive than necessary).

PR #17474, opened 2026-06-04
perf(iotdb): reduce REST health check cost

hjianbo:
Fixes https://github.com/emqx/emqx/issues/17413

Release version:
6.0.3

## Summary

This ports #17472 to `release-60`.

The IoTDB REST connector previously used `rest/v2/nonQuery` with `show databases` during auth health checks. That operation can be relatively expensive and is only needed to prove credentials are accepted, so the driver now probes `rest/v2/query` with `show version` instead.

The probe remains in the driver health-check path and keeps the existing 401 handling. A trace point records the probe path and SQL so the integration tests can prove the real request shape without introducing a mocked probe server.

Checked on the latest `release-60` base with focused CT against the real IoTDB 1.3 REST service:
`t_restapi_auth_probe_with_real_iotdb` verifies the connector is connected and the health probe uses `rest/v2/query` + `show version`.
`t_restapi_auth_probe_rejects_bad_password` verifies the same probe is still authenticated by IoTDB and bad credentials return `WRONG_LOGIN_PASSWORD`.
Also checked with `erlfmt`, `apps-version-check`, and `git diff --check`.

Added changelog `changes/ee/perf-17474.en.md`. There are no schema changes.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
+++ b/apps/emqx_bridge_iotdb/src/emqx_bridge_iotdb_connector.erl
@@ -80,6 +80,8 @@
 
 -define(CONNECTOR_TYPE, iotdb).
 -define(IOTDB_PING_PATH, <<"ping">>).
+-define(IOTDB_AUTH_PROBE_PATH, <<"rest/v2/query">>).
+-define(IOTDB_AUTH_PROBE_SQL, <<"show version">>).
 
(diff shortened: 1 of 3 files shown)


zmstone (approved):

yanzhiemq:
[TestExecution](https://emqx.atlassian.net/browse/QA-14947)

PR #17476, opened 2026-06-04
fix(acme): unblock plugin config save from the dashboard

zmstone:
Release version: 6.1.2

## Summary

Two issues hit the operator the moment they tried to configure the ACME plugin from the dashboard.

**`function_clause` on saving `acc_key: file://...`**

`acc_key` and `acc_key_password` were declared in the avsc schema as `["null", "string"]` unions. The Avro JSON decoder requires union values to be tagged (`{"string": "..."}`); the dashboard sends bare strings, so `avro_json_decoder:parse_union/4` had no matching clause and the request returned `BAD_CONFIG` with `function_clause` from `decode_avro_json`.

Both fields are now plain `"string"` with default `""`. The parser already maps `""` to "unset" (and still accepts `null` for in-place upgrades), so runtime behaviour is unchanged.

**`issuance_worker_crashed` on a malformed domain like `admin@example`**

The string was passed straight to acme-erlang-client, where `idna:to_ascii/1` raises `exit:{bad_label, _}` from inside a `try` block that only catches `throw`. The exception propagated up and killed the issuance worker, surfacing as `issuance_worker_crashed` in the logs.

- The config parser now rejects entries containing `@`, whitespace, `/`, `\\`, or control codes at save time, returning `{invalid_domain, ...}` so the operator sees the error in the dashboard immediately instead of after the issuance attempt.
- The issuance worker now wraps `run_action/1` in a catch-all, so any remaining unexpected exception (e.g. idna `too_long` on a 100-char label) lands in `/status` as a structured `last_result` and `in_progress` clears for the next attempt — no plugin restart required.

## PR Checklist
- [x] Change log for changes visible by users has been added to `changes/ee/fix-17476.en.md`
- [x] Schema changes are backward compatible or intentionally breaking (avsc default went from `null` to `""`; parser accepts both)

### DIFF
--- a/plugins/emqx_acme/priv/config_schema.avsc
+++ b/plugins/emqx_acme/priv/config_schema.avsc
@@ -140,8 +140,8 @@
     },
     {
       "name": "acc_key",
-      "type": ["null", "string"],
-      "default": null,
+      "type": "string",
+      "default": "",
       "$ui": {
         "component": "input",
         "flex": 12,
@@ -152,8 +152,8 @@
     },
     {
       "name": "acc_key_password",
-      "type": ["null", "string"],
-      "default": null,
+      "type": "string",
+      "default": "",
       "$ui": {
         "component": "input",
         "flex": 12,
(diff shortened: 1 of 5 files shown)


ieQu1 (approved):

hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-14971)

Issue #11671, opened 2023-09-25
The Retain As Published option is not set when bridging


tigercl:
### What happened?

<img width="1265" alt="image" src="https://github.com/emqx/emqx/assets/15085956/b9fba6c6-996f-41a3-b48c-3385211d59f4">

After the bridge mode is turned on, EMQX will set the highest bit of the Protocol Version field in the CONNECT packet to 1, indicating that this is a bridge.

Theoretically, it only works with MQTT 3.1.1. Because in MQTT 5.0 we can set the Retain As Published option when subscribing.

But now, MQTT 5.0 can enable bridge mode, and the corresponding bit in the CONNECT packet will be set to 1.

But regardless of whether the bridge mode is enabled or not, the bridge will not set the Retain As Published option when subscribing, even if the MQTT version is 5.0.

### What did you expect to happen?

When the MQTT version is 5.0, users should no longer be able to enable bridge mode, and the EMQX bridge must also set Retain As Published when subscribing.

### EMQX version

EMQX v5.1.6

### OS version

CentOS 7.9

### Log files

<details>

</details>

tigercl:
您好，您的邮件已经收到，我会及时处理。
如有急事，请电话联系。 by 周子博.

PR #17481, opened 2026-06-04
fix(bridge_mqtt): support retain_as_published on v5 ingress subscriptions

zmstone:

## Summary

Fixes [#11671](https://github.com/emqx/emqx/issues/11671) — *The Retain As Published option is not set when bridging.*

Two related problems with MQTT-bridge ingress under `proto_ver = v5`:

1. **Retain As Published (RAP) not set on remote subscribe.** The bridge issued `emqtt:subscribe/4` without the MQTT 5.0 `rap` subscription option, so the upstream broker always cleared the `retain` flag on forwarded PUBLISHes. The bridge then republished downstream with `retain = 0` even for upstream-retained messages.

2. **`bridge_mode = true` was silently accepted under v5.** The legacy bridge-bit in CONNECT is MQTT 3.1.1-era and has no meaning under MQTT 5.0, where bridge semantics live in subscribe options (RAP, NL).

### Changes

- New per-subscription option `retain_as_published` (boolean, default `false`) on the MQTT source. When `true` and the connector uses `proto_ver = v5`, the upstream broker preserves the original `retain` flag on forwarded PUBLISHes. Defaulting to `false` keeps existing configs behaving identically.
- The bridge subscribe call (and the subscription-identifier retry path) now thread the `rap` subscription option through to `emqtt:subscribe/4`.
- The connector logs a single warning at start when `bridge_mode = true` is combined with `proto_ver = v5`. Schema is not changed to reject the combination — avoids breaking upgrades.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
+++ b/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
@@ -102,6 +102,7 @@ on_start(ConnResId, #{server := Server} = Conf) ->
         connector => ConnResId,
         config => emqx_utils:redact(Conf)
     }),
+    maybe_warn_bridge_mode_v5(ConnResId, Conf),
     TopicToHandlerIndex = emqx_topic_index:new(),
     ok = emqx_resource:allocate_resource(
         ConnResId,
@@ -567,6 +568,18 @@ maybe_new_subscription_id_index(#{proto_ver := v5}) ->
     ets:new(?MODULE, [set, public, {read_concurrency, true}]);
 maybe_new_subscription_id_index(_Conf) ->
     undefined.
+
+maybe_warn_bridge_mode_v5(ConnResId, #{proto_ver := v5, bridge_mode := true}) ->
+    ?tp(mqtt_connector_bridge_mode_v5_warning, #{connector => ConnResId}),
+    ?SLOG(warning, #{
+        msg => "bridge_mode_ignored_for_mqtt_v5",
+        connector => ConnResId,
+        hint =>
+            "bridge_mode is a legacy MQTT 3.1.1-era flag with no effect under MQTT 5.0; "
+            "set retain_as_published on individual subscriptions instead."
+    });
+maybe_warn_bridge_mode_v5(_ConnResId, _Conf) ->
+    ok.
 
 maybe_delete_subscription_id_index(undefined) ->
     ok;
--- a/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
+++ b/apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
@@ -570,13 +570,9 @@ maybe_new_subscription_id_index(_Conf) ->
     undefined.
 
 maybe_warn_bridge_mode_v5(ConnResId, #{proto_ver := v5, bridge_mode := true}) ->
-    ?tp(mqtt_connector_bridge_mode_v5_warning, #{connector => ConnResId}),
-    ?SLOG(warning, #{
-        msg => "bridge_mode_ignored_for_mqtt_v5",
+    ?tp(warning, "bridge_mode_ignored_for_mqtt_v5", #{
         connector => ConnResId,
-        hint =>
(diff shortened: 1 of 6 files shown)


apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
```diff
@@ -568,6 +569,18 @@ maybe_new_subscription_id_index(#{proto_ver := v5}) ->
 maybe_new_subscription_id_index(_Conf) ->
     undefined.
 
+maybe_warn_bridge_mode_v5(ConnResId, #{proto_ver := v5, bridge_mode := true}) ->
+    ?tp(mqtt_connector_bridge_mode_v5_warning, #{connector => ConnResId}),
+    ?SLOG(warning, #{
+        msg => "bridge_mode_ignored_for_mqtt_v5",
+        connector => ConnResId,
```
keynslug:
```suggestion
    ?tp(warning, "bridge_mode_ignored_for_mqtt_v5", #{
        connector => ConnResId,
```

apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
```diff
@@ -1140,3 +1140,109 @@ t_reconnect_with_session(TCConfig) when is_list(TCConfig) ->
...
+    ok = meck:new(emqtt, [passthrough, no_link, no_history]),
+    on_exit(fun() -> catch meck:unload([emqtt]) end),
+    ok = meck:expect(
+        emqtt,
+        subscribe,
+        fun(Pid, Props, Topic, Opts) ->
+            Self ! {subscribe_opts, Opts},
+            meck:passthrough([Pid, Props, Topic, Opts])
+        end
+    ),
```
keynslug: Q: Possible to just peek into `emqx_subopts` table instead (assuming the bridge is connecting to the local node)? To avoid relying on internals.

keynslug (dismissed):

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_ingress.erl
```diff
@@ -119,14 +126,19 @@ maybe_retry_without_subscription_identifier(
 maybe_retry_without_subscription_identifier(
     #{
         subscription_id := _SubscriptionId,
-        remote := #{topic := RemoteTopic, qos := QoS, no_local := NoLocal}
+        remote := #{
+            topic := RemoteTopic,
+            qos := QoS,
+            no_local := NoLocal,
+            retain_as_published := RAP
```
zmstone: maps:get(retain_as_published, Remote, true).

apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
zmstone: Addressed in e95deb7d8a — the helper now waits for the subscription with `?retry` and asserts on `emqx_broker:subscriptions_via_topic/1` subopts, dropping the `emqtt` meck.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_connector.erl
zmstone: Addressed in 32b0bb8b8d — collapsed into a single `?tp(warning, ...)` and shortened the hint now that `retain_as_published` is on by default.

apps/emqx_bridge_mqtt/src/emqx_bridge_mqtt_ingress.erl
zmstone: Done in 605df77afe — both call sites now extract the `remote` map first and do `maps:get(retain_as_published, Remote, true)` so hot-upgrade configs without the key still work.

keynslug (approved):

hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15382)

PR #17487, opened 2026-06-04
feat(dashboard): strengthen password and api-key hashing

zmstone:
Release version: 6.3.0

## Summary

Strengthens dashboard administrator password and API-key secret hashing, and replaces the embarrassingly weak 2-byte salt + single-SHA-256 layout (whose doc comment misleadingly described it as "4 bytes salt").

### Storage format

Self-describing PHC-style: `` `$1$<iterations>$<salt_b64>$<dk_b64>` ``. The leading `1` is the storage schema version; the iteration count is inline so future bumps stay backwards-readable. No mnesia schema changes — the `pwdhash` / `api_secret_hash` fields are already `binary()`.

### Two-tier hashing (same format, different cost)

- **`emqx_dashboard_admin:hash/1`** (dashboard passwords): PBKDF2-HMAC-SHA256, **600,000 iterations** (OWASP 2023), 16-byte random salt. ~500 ms per verify — fine for once-per-login.
- **`emqx_dashboard_admin:hash_api_secret/1`** (API key secrets): same `$1$` format but **iterations = 1**, effectively salted HMAC-SHA256. ~4 µs per verify. API secrets are 32 bytes of `crypto:strong_rand_bytes` (~190 bits of entropy) — a slow KDF would burn CPU on every HTTP API call without strengthening already-uncrackable secrets.

The verifier (`verify_hash/2`) is the same function for both: it reads the iteration count from the stored hash. Constant-time compare via `crypto:hash_equals/2` (OTP 27+).

### Rolling-upgrade safety

Legacy 36-byte hashes (4-byte ASCII-hex salt + raw SHA-256) keep verifying. We deliberately do **not** rehash-on-login: during a rolling upgrade an older node cannot parse the new format, so silently rewriting on login would lock users out at their next request routed to an old node. Legacy hashes are replaced only when the user changes their password or the credential is recreated.

### Callers updated

- `apps/emqx_dashboard/src/emqx_dashboard_admin.erl` — new `hash/1` and `hash_api_secret/1`; rewritten `verify_hash/2` (legacy + v1 branches, returns plain `ok | error`).
- `apps/emqx_management/src/emqx_mgmt_auth.erl` — both `create_app/8` and `add_bootstrap_file/4` now call `hash_api_secret/1`.
- `apps/emqx_dashboard/src/emqx_dashboard_token.erl` — removed the unused `salt/0` (the only caller was the old `hash/1`; grep confirmed nothing else depended on it).
- `apps/emqx_dashboard/test/emqx_dashboard_SUITE.erl` — `t_cli` was peeking into the stored hash as `<<Salt:4, Hash:32>>`; rewritten to verify via `verify_hash/2`.

### Tests

- `t_hash_uses_v1_format` — password hash carries `$1$600000$<salt_b64>$<dk_b64>` with 16-byte salt and 32-byte DK.
- `t_hash_api_secret_uses_iter_1` — API-secret hash carries `$1$1$...` and the unified verifier accepts it.
- `t_verify_v1_hash_roundtrip` — right/wrong password; two hashes of the same password differ (random salt).
- `t_verify_legacy_hash_succeeds` — pre-6.4 36-byte hash still verifies.
- `t_verify_legacy_hash_wrong_password` — wrong password against legacy hash returns `error`.
- `t_verify_rejects_malformed_input` — empty / garbage / truncated v1 / random 36-byte → `error`.

### Hard constraints honoured

- No bcrypt, no Argon2, no new deps.
- Self-describing storage format (version prefix), no tag-less ambiguity.
- Legacy hashes keep working without a forced password reset.
- Constant-time hash comparison.
- `pwdhash` / `api_secret_hash` schema fields unchanged.

## PR Checklist
- [x] Change log for changes visible by users has been added to `` `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` `` files


savonarola (dismissed):

keynslug (dismissed):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16070)

PR #17490, opened 2026-06-04
refactor(utils): use OTP native json module instead of jiffy

zmstone:
Release version: 6.3.1

## Summary

Migrate `emqx_utils_json` from `jiffy` to OTP's native `json` module. Public API is unchanged; all 11 in-module eunit tests pass (3 pre-existing + 8 new), as does `emqx_utils_json_SUITE`.

- `apps/emqx_utils/src/emqx_utils_json.erl` — wraps `json:encode/2` + `json:decode/1,3`. A custom encoder fun handles jiffy-style ejson `{[]}` and `{[{K, V}, ...]}` tuples, `uescape` (`json:encode_binary_escape_all/1`), and `force_utf8` (binary is sanitized to U+FFFD on invalid bytes before encoding). `pretty` is implemented by a post-pass over the compact output to match the existing CLI/HTTP format byte-for-byte (the format the existing `best_effort_json_test/0` asserts). `decode(_, [])` uses a custom `object_push`/`object_finish` decoder to preserve insertion order so the ejson form roundtrips identically.
- `apps/emqx_rule_engine/test/emqx_rule_funcs_SUITE.erl` — two assertions in `t_join_to_string` / `t_join_to_sql_values_string` relied on jiffy's specific map iteration order for a 2-key map (`{"value":2,"key":1}`); updated to OTP json's order (`{"key":1,"value":2}`). Map key order is undefined in JSON, so the previous expectation was brittle.
- `jiffy` stays as a top-level mix dep — a handful of test files (`emqx_connector_aggregator`, `emqx_eviction_agent`, `emqx_node_rebalance`, `emqx_gcp_device`) still call `jiffy:` directly. A follow-up PR can clean those up and drop the dep.

Decode options `return_trailer`, `dedupe_keys`, `copy_strings` are now accepted-and-ignored (no current callers used them); `return_maps` is the default and a no-op.

### DIFF
--- a/apps/emqx_utils/src/emqx_utils_json.erl
+++ b/apps/emqx_utils/src/emqx_utils_json.erl
@@ -96,11 +96,7 @@ encode(Term) ->
 
 -spec encode(json_term(), encode_options()) -> json_text().
 encode(Term, Opts) ->
-    Bin = iolist_to_binary(json:encode(Term, encoder(Opts))),
-    case lists:member(pretty, Opts) of
-        true -> pretty_print(Bin);
-        false -> Bin
-    end.
+    do_encode(Term, encoder(Opts, false), Opts).
 
 -spec encode_proplist(json_term_proplist()) -> json_text().
 encode_proplist(Term) ->
@@ -108,7 +104,16 @@ encode_proplist(Term) ->
 
 -spec encode_proplist(json_term_proplist(), encode_options()) -> json_text().
 encode_proplist(Term, Opts) ->
-    encode(to_ejson(Term), Opts).
+    %% `proplist' mode makes the encoder treat any `[{K, V}, ...]' list as a
+    %% JSON object, so no pre-pass conversion of the term is needed.
+    do_encode(Term, encoder(Opts, true), Opts).
+
+do_encode(Term, Encoder, Opts) ->
+    Bin = iolist_to_binary(json:encode(Term, Encoder)),
+    case lists:member(pretty, Opts) of
+        true -> pretty_print(Bin);
+        false -> Bin
+    end.
 
 -spec safe_encode(json_term()) ->
     {ok, json_text()} | {error, Reason :: term()}.
@@ -281,9 +286,11 @@ json_key(Term) ->
 
 %% Custom encoder fun for `json:encode/2`. Handles:
 %%   * jiffy-style ejson objects `{[]}` and `{[{K, V}, ...]}`.
+%%   * `Proplist' mode (set by `encode_proplist') — also treat any
+%%     `[{K, V}, ...]' list as a JSON object.
 %%   * `uescape` — escape every non-ASCII codepoint as `\uXXXX`.
 %%   * `force_utf8` — replace invalid UTF-8 bytes with U+FFFD instead of crashing.
-encoder(Opts) ->
+encoder(Opts, Proplist) ->
     Uescape = lists:member(uescape, Opts),
     ForceUtf8 = lists:member(force_utf8, Opts),
     fun
@@ -293,6 +300,8 @@ encoder(Opts) ->
             json:encode_key_value_list(L, Enc);
         (B, _Enc) when is_binary(B) ->
             encode_binary(B, Uescape, ForceUtf8);
+        ([{_, _} | _] = L, Enc) when Proplist ->
+            json:encode_key_value_list(L, Enc);
         (Other, Enc) ->
             json:encode_value(Other, Enc)
     end.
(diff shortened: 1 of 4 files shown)


apps/emqx_utils/src/emqx_utils_json.erl
```diff
@@ -247,6 +275,108 @@ json_key(Term) ->
...
+
+%% Pretty-print a compact JSON binary into the format used by the EMQX CLI:
+%%   * 2-space indent per nesting level
+%%   * `\n<indent>` after `{` `[` `,`
+%%   * `\n<indent>` before `}` `]`
+%%   * `" : "` between key and value
+%%   * empty container body keeps a `\n<inner>\n<outer>` pair
+%% This affects the CLI output format, consult the team before changing the format.
+pretty_print(Bin) ->
+    pretty(Bin, 0, <<>>, false, false).
```
keynslug: Q: Replace with `json:format/1`?

apps/emqx_utils/src/emqx_utils_json.erl
```diff
@@ -258,12 +388,14 @@ json_key(Term) ->
...
 
+%% Convert a proplist-with-map term to jiffy-style ejson tuple form so the
+%% encoder fun can recognise it.
 to_ejson([{_, _} | _] = L) ->
     {[{K, to_ejson(V)} || {K, V} <- L]};
 to_ejson(L) when is_list(L) ->
     [to_ejson(E) || E <- L];
 to_ejson(M) when is_map(M) ->
-    maps:map(fun(_K, V) -> to_ejson(V) end, M);
+    {[{K, to_ejson(V)} || K := V <- M]};
```
keynslug: Perhaps just move it into `encoder/1`?

zmstone: Done in eb105dcbae. Dropped `to_ejson/1` entirely; `encode_proplist/2` now runs the encoder in a `Proplist` mode where the fun treats any `[{K, V}, ...]` list as a JSON object (`json:encode_key_value_list/2` recurses through the same fun, so nested proplists inside maps/lists are handled too). This also removes the extra full-term traversal the pre-pass did. Verified behavior-equivalent against the old path and all eunit + `emqx_utils_json_SUITE` cases pass.

apps/emqx_utils/src/emqx_utils_json.erl
zmstone: Switched to `json:format/3` in b2e980d62a — dropped the custom pretty printer entirely (a small `formatter/2` callback keeps the ejson/proplist/`uescape`/`force_utf8` handling). Per your call we're not preserving the old byte-for-byte CLI format; the new output uses `": "`, compact empty/small containers, and a trailing newline. Updated the asserting tests accordingly.

While here I caught two behavior changes from the jiffy→OTP migration and fixed them so the migration stays transparent:
- duplicate object keys: OTP `json:decode` is first-wins, jiffy is last-wins (also the JS/Python/Go convention) — restored last-wins via a custom `object_finish`;
- trailing garbage after a value: `json:decode/3` returns it as a trailer instead of erroring — now rejected again (trailing whitespace still accepted).

apps/emqx_utils/src/emqx_utils_json.erl
```diff
@@ -247,26 +357,91 @@ json_key(Term) ->
...
-to_ejson([{_, _} | _] = L) ->
-    {[{K, to_ejson(V)} || {K, V} <- L]};
-to_ejson(L) when is_list(L) ->
-    [to_ejson(E) || E <- L];
-to_ejson(M) when is_map(M) ->
-    maps:map(fun(_K, V) -> to_ejson(V) end, M);
-to_ejson(T) ->
-    T.
-
 from_ejson(L) when is_list(L) ->
```
keynslug, line 445: Nit. Makes sense to turn it into `proplist_decoders/0`, in similar vein to `to_ejson/1`?

apps/emqx_utils/src/emqx_utils_json.erl
```diff
@@ -118,11 +140,99 @@ decode(Json) ->
...
+    {byte_size(Bin), truncated_json};
+jiffy_compat_error({invalid_byte, Byte}, Stacktrace, _Bin) ->
+    {error_position(Stacktrace), classify_invalid_byte(Byte)};
+jiffy_compat_error({unexpected_sequence, _}, Stacktrace, _Bin) ->
+    {error_position(Stacktrace), invalid_string};
+jiffy_compat_error(Reason, Stacktrace, _Bin) when is_atom(Reason) ->
+    {error_position(Stacktrace), Reason};
+jiffy_compat_error(_Reason, Stacktrace, _Bin) ->
+    {error_position(Stacktrace), invalid_json}.
+
```
keynslug, line 183: Since there's few callsites relying on this "API contract", it's probably just easier to adopt them instead of maintaining this?

changes/ee/perf-17490.en.md
```diff
@@ -0,0 +1 @@
+Migrated the broker's internal JSON encoding/decoding facade from the `jiffy` NIF library to OTP 27's native `json` module. The public API of `emqx_utils_json` is unchanged. The migration reduces NIF dependencies and avoids loading the jiffy port driver at startup for paths that use the broker's JSON facade.
```
keynslug, line 1: > The public API of `emqx_utils_json` is unchanged.

Nit. Likely doesn't make much sense in a changelog entry? Unless it's for plugin maintainers.

qzhuyan, line 1: we may also get some perf degredation according to jiffy's README.

qzhuyan:
I wish we could have a feature toggle to select jiffy or json unless we are pretty happy about the perf of json. 
BTW, jiffy folks suggested jiffy 2.0

zmstone:
> I wish we could have a feature toggle to select jiffy or json unless we are pretty happy about the perf of json. BTW, jiffy folks suggested jiffy 2.0

from jiffy readme, found this: https://github.com/saleyn/glazer

zmstone:
for compatibility and performance, will choose to upgrade to jiffy 2.0 for 6.3

PR #17493, opened 2026-06-05
feat(session): add session buffer observability

### LINKED ISSUE #17396: Observability and byte-volume limits for session mqueue/inflight

hjianbo:
Fixes #17396

Design discussion: emqx/eip#110

Release version: 6.3.0

## Summary

This PR adds lightweight observability for MQTT payload bytes retained by sessions.

- Client APIs expose `total_payload_bytes`. In-memory sessions report payload bytes retained by the mqueue and inflight window; durable sessions report `0` for now.
- `sysmon.session.total_payload_bytes_high_watermark` emits per-client throttled warnings when an in-memory session exceeds the configured threshold. The default `0` disables the warning.
- `emqx ctl session-top` scans cached session statistics across the cluster and writes the top sessions to a new CSV file.

## `session-top` CLI

```text
session-top status       # Show status of the latest session-top scan.
session-top cancel       # Cancel the running session-top scan.
session-top --out <File> # Write cluster top sessions to a new CSV file.
  [--count <K>]          # Result row limit, default: 10, max: 1000.
  [--sort <SortBy>]      # SortBy: mqueue_length | total_payload_bytes.
  [--batch <Size>]       # Rows per local scan batch, default: 1000.
  [--sleep <Ms>]         # Delay between batches, default: 1ms.
```

The CSV columns are `clientid,node,mqueue_length,total_payload_bytes,inflight_count`. The output file must not already exist.

## Implementation Notes

- `emqx_mqueue` caches queued payload bytes. `emqx_session_mem` calculates inflight payload bytes on demand and publishes their sum in cached channel statistics.
- `emqx_session_top_scanner` incrementally scans one local batch per timer tick by using the `emqx_session_tool` accumulator APIs.
- `emqx_session_top_collector` starts one scanner per supported cluster node, merges their local top-K results, and coordinates status and best-effort cancellation from the initiating node.
- CSV generation and file IO are handled by `emqx_mgmt_cli` through the collector completion callback.

## PR Checklist


apps/emqx/src/proto/emqx_session_buffer_proto_v1.erl
```diff
@@ -0,0 +1,22 @@
...
+
+-export([
+    introduced_in/0,
+    local_top/4
+]).
+
+-include_lib("emqx/include/bpapi.hrl").
+
+introduced_in() ->
+    "6.3.0".
```
hjianbo:
```suggestion
    "6.0.3".
```

zmstone:
we cannot hot-upgrade it, so maybe only for release-63

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -1064,6 +1068,18 @@ inflight_insert_ts(#message{extra = #{?INFLIGHT_INSERT_TS := Ts}}) -> Ts.
 without_inflight_insert_ts(#message{extra = Extra} = Msg) ->
     Msg#message{extra = maps:remove(?INFLIGHT_INSERT_TS, Extra)}.
 
+inflight_payload_bytes(Inflight) ->
+    emqx_inflight:fold(
```
zmstone: this could be expensive for each stats collection.
mirror the bytes cache like mqueue in inflight.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,439 @@
...
+-spec update(map() | undefined) -> ok.
+update(Conf0) ->
+    Conf = put_conf(Conf0),
+    gen_server:cast(?MODULE, {update, Conf}).
+
+-spec maybe_log(emqx_types:clientid(), pid(), stats()) -> ok.
+maybe_log(ClientId, ChanPid, Stats) ->
+    Conf = persistent_term:get(?CONF_KEY, ?DEFAULT_CONF),
+    HighWatermark = maps:get(buffered_payload_high_watermark, Conf, 0),
+    TotalPayloadBytes = stat_value(total_payload_bytes, Stats, 0),
```
zmstone: no need to calculate `TotalPyaloadBytes` when disabled.
also change `HighWatermark > 0` to a `is_logging_enabled/1`, avoid `maps:get`, use pattern matching

```
is_loggin_enabled(#{buffered_payload_high_watermark := HWM}) when HWM > 0 -> true;
is_loggin_enabled(_Conf) -> false.
```

apps/emqx/src/emqx_mqueue.erl
```diff
@@ -250,6 +257,10 @@ to_list(MQ, Acc) ->
 -spec dropped(mqueue()) -> count().
 dropped(#mqueue{dropped = Dropped}) -> Dropped.
```
zmstone, line 268: add inline compile option

zmstone: though i'm not sure if inline helps remote calls.

apps/emqx/src/emqx_mqueue.erl
```diff
@@ -279,39 +291,79 @@ in(
...
-    {R, MQ#mqueue{q = Q2, len = Len - 1, p_credit = Cnt - 1}}.
+        0,
+        Q
+    ).
+
+msg_bytes(#message{} = Msg) ->
+    emqx_message:payload_size(Msg).
+
+dec_bytes(#mqueue{bytes_size = BytesSize}, Msg) ->
+    max(BytesSize - msg_bytes(Msg), 0).
```
zmstone: we should trust the algorithm. remove this clamp.
if it ever goes to negative value, we at least know something is wrong.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,439 @@
...
+        node => node(),
+        mqueue_length => stat_value(mqueue_len, Stats, 0),
+        total_payload_bytes => stat_value(total_payload_bytes, Stats, 0),
+        inflight_count => stat_value(inflight_cnt, Stats, 0)
+    }.
+
+keep_top(_Row, 0, _SortBy, TopRows) ->
+    TopRows;
+keep_top(Row, Count, SortBy, TopRows) ->
+    insert_top(Row, Count, top_key(Row, SortBy), TopRows).
```
zmstone: given line 316 is commonly the hottest path,
this top_key builds a new tuple can be a waste if most of them are discarded immediately anyways.

pass SoortBy down to insert_top, and build top_key only when needed.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,439 @@
...
+    Status = #{
+        status => running,
+        pid => Pid,
+        out => maps:get(out, Opts),
+        count => maps:get(count, Opts),
+        sort => maps:get(sort, Opts)
+    },
+    {reply, {ok, Pid}, State#{scan := Scan, top_status => Status}};
+handle_call({run_top, _Opts}, _From, State) ->
+    {reply, {error, busy}, State};
```
zmstone: maybe report something

when the scan started, what's the progress of the scan.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,439 @@
...
+    write_csv/2
+]).
+-endif.
+
+-define(CONF_KEY, {?MODULE, conf}).
+-define(DEFAULT_CONF, #{buffered_payload_high_watermark => 0}).
+-define(LOG_MSG, session_buffer_high_watermark).
+-define(TOP_TIMEOUT, 300000).
+-define(SCAN_BATCH_SIZE, 1000).
+-define(SCAN_SLEEP_MS, 1).
```
zmstone: make batch and sleep configurable.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,439 @@
...
+
+result_rows(Rows, {AccRows, AccBad}) when is_list(Rows) ->
+    {Rows ++ AccRows, AccBad};
+result_rows({badrpc, Reason}, {AccRows, AccBad}) ->
+    {AccRows, [Reason | AccBad]};
+result_rows(Other, {AccRows, AccBad}) ->
+    {AccRows, [Other | AccBad]}.
+
+-spec scan_local(pos_integer(), sort_by()) -> [row()].
+scan_local(Count, SortBy) ->
```
zmstone: the current design only ensures no concurrent initial scan, which means each node can initiate concurrent scan independently.

local scan should be delegated to gen_server (which queues concurrent requests, or returns `busy` when there is ongoing scan).

we should refactor the scan to:

1. scan can be initiated from any node in the cluster, pass init node and timestamp around to cluster nodes
2. each node should allow only one scan at a time (either queue the requests, or reject-try-later)
3. scan can be cancelled from cli

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,439 @@
...
+        case ets:lookup(?CHAN_INFO_TAB, Key) of
+            [{Key = {ClientId, ChanPid}, _Info, Stats}] ->
+                keep_top(row(ClientId, ChanPid, Stats), Count, SortBy, TopRows);
+            _ ->
+                TopRows
+        end,
+    BatchCount1 = BatchCount + 1,
+    case BatchCount1 rem ?SCAN_BATCH_SIZE of
+        0 ->
+            timer:sleep(?SCAN_SLEEP_MS);
```
zmstone: if we design the scan to be handled by gen_server, do not sleep, use timer trigger the next batch.
in between, the batch:
- call garbage_collect()
- erlang:send_after to trigger the next batch
- handle_cast or handle_info handle the cancellation of the in-progress scan

at the end of the scan, send back the top result to the (remote) collector.
monitor the remote collector, abort scan if collector is DOWN.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,439 @@
...
+top_heap_rows(TopRows) ->
+    [Row || {_Key, Row} <- lists:reverse(gb_trees:to_list(TopRows))].
+
+sort_value(Row, mqueue_length) ->
+    maps:get(mqueue_length, Row, 0);
+sort_value(Row, total_payload_bytes) ->
+    maps:get(total_payload_bytes, Row, 0).
+
+write_csv(OutFile, Rows) ->
+    case file:open(OutFile, [write, raw, binary]) of
```
zmstone: check OutFile does not exist.
(do it in the CLI entrypoint too, print error if file already exists).

this should avoid ops typo etc to accidentally overwrite some files (e.g. data/configs/cluster.hocon)

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,439 @@
...
+
+-define(CONF_KEY, {?MODULE, conf}).
+-define(DEFAULT_CONF, #{buffered_payload_high_watermark => 0}).
+-define(LOG_MSG, session_buffer_high_watermark).
+-define(TOP_TIMEOUT, 300000).
+-define(SCAN_BATCH_SIZE, 1000).
+-define(SCAN_SLEEP_MS, 1).
+
+-type stats() :: emqx_types:stats() | map().
+-type sort_by() :: mqueue_length | total_payload_bytes.
```
zmstone: i have drafted another pr attempting to be more generic, https://github.com/emqx/emqx/pull/17712

maybe we can merge it first, then sync it over to dev-63 and add total_payload_bytes on top.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,517 @@
...
+        top_status => #{status => idle}
+    }.
+
+%%--------------------------------------------------------------------
+%% Helpers
+%%--------------------------------------------------------------------
+
+put_conf(Conf0) ->
+    Conf = normalize_conf(Conf0),
+    persistent_term:put(?CONF_KEY, Conf),
```
zmstone: read and compare before put it.

apps/emqx/src/emqx_message.erl
```diff
@@ -182,6 +183,10 @@ estimate_size(#message{topic = Topic, payload = Payload}) ->
     TopicLengthSize = 2,
     FixedHeaderSize + VarLenSize + TopicLengthSize + TopicSize + PacketIdSize + PayloadSize.
 
+-spec payload_size(emqx_types:message()) -> non_neg_integer().
+payload_size(#message{payload = Payload}) ->
+    iolist_size(Payload).
```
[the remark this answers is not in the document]
zmstone: how is it resolved ?
is payload never going to be undefined?

apps/emqx/src/emqx_mqueue.erl
```diff
@@ -97,6 +100,7 @@
     store_qos0 = false :: boolean(),
     max_len = ?MAX_LEN_INFINITY :: count(),
     len = 0 :: count(),
+    bytes_size = 0 :: count(),
```
zmstone: byte_size.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,517 @@
...
+    end.
+
+-spec scan_local(pos_integer(), sort_by()) -> [row()].
+scan_local(Count, _SortBy) when Count =< 0 ->
+    [];
+scan_local(Count, SortBy) ->
+    scan_tool_rows(#{count => Count, sort => SortBy}).
+
+scan_tool_rows(#{count := _Count, sort := _Sort} = Opts) ->
+    session_top_rows(emqx_session_tool:scan(session_tool_opts(Opts))).
```
zmstone: call `scan_acc_new` and `scan_acc`, and trigger batch scans with a timer so `emqx_session_buffer_mon` will only be blocked for one chunk -- possible to abort in the middle and respond to other calls.

apps/emqx/src/emqx_message.erl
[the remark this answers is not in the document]
hjianbo: Yes, except for test case sources, payloads from any other normal sources shouldn't be undefined

apps/emqx/src/emqx_session_buffer_mon.erl
hjianbo: I checked this path again and wanted to confirm the intended layer before changing it.

The CLI path is `session-top` -> `emqx_session_buffer_mon:run_top/1` -> a monitored worker -> `do_run_top/1` -> `emqx_session_tool:cluster_top_by/2` -> `emqx_session_tool_proto_v1:scan/3` -> `erpc:multicall(..., emqx_session_tool, scan, [Opts], Timeout)`.

So the long `cluster_top_by/2` call runs in the monitored worker process, not in the `emqx_session_buffer_mon` gen_server process. The gen_server can still handle `status` and `cancel` while the worker is scanning. `cancel` kills the local worker, but I agree it does not provide a cluster-level cancellation token for remote scans that have already started through `erpc:multicall`.

`scan_acc_new/1` and `scan_acc/1` look like the right primitive for a local gen_server-owned scan. My concern is that using them directly here would either only affect the local scan path, or duplicate the cluster scan/merge/timeout behavior currently owned by `cluster_top_by/2`.

Do you think this PR should extend `emqx_session_tool` with a cluster-level incremental/cancellable scan API and move `session-top` to that, or is there a smaller change you had in mind for this layer?

zmstone: `emqx_session_tool:cluster_top_by` was added for ad-hoc `emqx eval` command for older versions, not intended for `emqx_session_buffer_mon` to call.

The idea is to make `emqx_session_buffer_mon` the task serializer in each node locally, and the runner itself.

1) At any point in time, there is one scan running on each node;
2) CLI -> `emqx_session_buffer_mon` should be able to abort an on-going scans (cluster-wide)
3) CLI -> `emqx_session_buffer_mon` should be able to report scan progress (cluster-wide)
4) Multi-call is to trigger an async scan in all nodes
  - Results can be sent back to the cluster scan initializer
  - Or `emqx_session_buffer_mon` remembers the last scan result locally

zmstone: i.e. call the incremental APIs

> %% Incremental scan engine (for a caller that owns the cursor/state).
-export([
    scan_acc_new/1,
    scan_acc/1,
    scan_acc_rows/1
]).

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,524 @@
...
+
+handle_call({run_top, Opts0}, _From, State = #{top := undefined}) ->
+    Opts = normalize_top_opts(Opts0),
+    case ensure_out_file_absent(maps:get(out, Opts)) of
+        ok ->
+            RunningNodes = emqx:running_nodes(),
+            NodesTotal = length(RunningNodes),
+            Server = self(),
+            {Pid, MRef} =
+                erlang:spawn_monitor(fun() ->
```
zmstone: calling self is a odd design.

the caller only need a handle to receive result, so make_ref should do.
to trigger the scan, send a message to self, or use handle_continue

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,807 @@
...
+    Conf = put_conf(emqx_config:get([sysmon, session], ?DEFAULT_CONF)),
+    {ok, #{
+        conf => Conf,
+        top => undefined,
+        top_status => #{status => idle}
+    }}.
+
+handle_call({run_top, Opts0}, _From, State = #{top := undefined}) ->
+    Opts = normalize_top_opts(Opts0),
+    case ensure_out_file_absent(maps:get(out, Opts)) of
```
zmstone: extract this block to a sub function

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,807 @@
...
+stat_metric_value(_Key, _Row) ->
+    0.
+
+sort_metric(mqueue_length) ->
+    mqueue_len;
+sort_metric(total_payload_bytes) ->
+    total_payload_bytes.
+
+write_csv(OutFile, Rows) ->
+    case file:open(OutFile, [write, exclusive, raw, binary]) of
```
zmstone: to be on the safe side, ensure the file not exist before writing

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -553,12 +556,18 @@ deliver(ClientInfo, Msgs, Session) ->
...
 do_deliver(_ClientInfo, [], Publishes, _CheckLimiter, Session) ->
     {ok, lists:reverse(Publishes), Session};
-do_deliver(ClientInfo, [Msg | More], Acc, CheckLimiter, Session) ->
-    case deliver_msg(ClientInfo, Msg, CheckLimiter, Session) of
-        {ok, [], Session1} ->
-            do_deliver(ClientInfo, More, Acc, CheckLimiter, Session1);
-        {ok, [Publish], Session1} ->
-            do_deliver(ClientInfo, More, [Publish | Acc], CheckLimiter, Session1)
+do_deliver(ClientInfo = #{zone := Zone}, [Msg | More], Acc, CheckLimiter, Session) ->
+    case emqx_message:is_expired(Msg, Zone) of
```
keynslug: One may argue that if a delivery is in the process mailbox, then the broker already "managed to start onward delivery to a matching subscriber". Admittedly, in _rare_ circumstances it's possible for a delivery to take seconds to get here, and this message may indeed expire already with short but reasonable (i.e. not simply 0) expiry interval. I'm honestly inclined to think it's not worth complicating the hot code path however.

That said, maybe worth putting it aside to a separate PR?

apps/emqx/src/emqx_inflight.erl
```diff
@@ -33,18 +36,18 @@
 
 -type max_size() :: pos_integer().
 
--opaque inflight() :: {inflight, max_size(), gb_trees:tree()}.
+-opaque inflight() :: {inflight, max_size(), non_neg_integer(), gb_trees:tree()}.
```
keynslug: Note that it breaks takeover rolling-upgrade-safety. I address this in #17870 for a similar reason, because linked PR slightly changes in-memory session shape. So if we get both merged, shouldn't be a concern.

apps/emqx/src/emqx_inflight.erl
```diff
@@ -123,6 +129,22 @@ window(Inflight = ?INFLIGHT(Tree)) ->
...
+    BytesSize1 = BytesSize - val_bytes(Val),
+    true = BytesSize1 >= 0,
+    BytesSize1.
+
+val_bytes({inflight_data, _Phase, #message{} = Msg, _Timestamp}) ->
+    emqx_message:payload_size(Msg);
+val_bytes(#message{} = Msg) ->
+    emqx_message:payload_size(Msg);
+val_bytes(_Val) ->
+    0.
```
keynslug: This looks a bit questionable. If we want to keep this `emqx_inflight` data structure unaware of what's inside, then "total bytes in-flight" is better tracked on the session level, e.g. in a separate field. Otherwise, it's simpler to make it aware of it, e.g. `insert(PacketId, Msg, Timestamp, Tag)`. This is also highlighted by the test suite that now need its own `-record(inflight_data, ...`.

apps/emqx/src/emqx_message.erl
```diff
@@ -182,6 +183,10 @@ estimate_size(#message{topic = Topic, payload = Payload}) ->
     TopicLengthSize = 2,
     FixedHeaderSize + VarLenSize + TopicLengthSize + TopicSize + PacketIdSize + PayloadSize.
 
+-spec payload_size(emqx_types:message()) -> non_neg_integer().
+payload_size(#message{payload = Payload}) ->
```
keynslug, line 187: Q: Why separate `payload_size/1` if there's already `estimate_size/1`?

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,811 @@
...
+                    do_log(ClientId, ChanPid, Stats, TotalPayloadBytes, HighWatermark);
+                false ->
+                    ok
+            end;
+        false ->
+            ok
+    end.
+
+do_log(ClientId, ChanPid, Stats, TotalPayloadBytes, HighWatermark) ->
+    ?SLOG_THROTTLE(
```
keynslug: If it's broadly throttled, I don't honestly see how operators is going to work with this. Shouldn't it be throttled on per-clientid basis?

apps/emqx/src/emqx_inflight.erl
```diff
@@ -21,6 +23,7 @@
     to_list/1,
     to_list/2,
     size/1,
+    bytes_size/1,
```
keynslug: Another idea: if we know it's not going to be called _too_ often, and the inflight size is usually bounded by a low inflight size, perhaps just compute it on-demand each time? I.e. drop bytesize tracking completely.

keynslug (commented):
Few initial questions / remarks, going to review further tomorrow.

apps/emqx/src/emqx_session_mem.erl
hjianbo: Agreed — removed from this PR.

apps/emqx/src/emqx_inflight.erl
hjianbo: Synced with `dev-63` (#17870). Inflight shape unchanged; bytes tracked on session.

apps/emqx/src/emqx_inflight.erl
hjianbo: Done — inflight stays opaque; bytes on `session.inflight_payload_bytes`.

apps/emqx/src/emqx_session_buffer_mon.erl
hjianbo: Fixed — throttle key is now per-clientid.

apps/emqx/src/emqx_inflight.erl
hjianbo: Kept a session-side counter; no per-stats inflight scan.

apps/emqx/src/emqx_message.erl
hjianbo: Since this PR only monitors total_payload_bytes, it is implemented this way. Initially, I planned to count the entire message, but Stone suggested it wasn't necessary, so we adjusted the scope accordingly.

apps/emqx/src/emqx_mqueue.erl
```diff
@@ -330,11 +353,34 @@ out(MQ = #mqueue{q = Q, p_credit = 0}) ->
...
+            Acc + msg_bytes(Msg)
+        end,
+        0,
+        Q
+    ).
+
+msg_bytes(#message{} = Msg) ->
+    emqx_message:payload_size(Msg);
+msg_bytes(_Msg) ->
+    0.
```
keynslug: Nit. `in/2` only allows `#message{}` records in so it looks unnecessary to be this much defensive here.

apps/emqx/src/emqx_mqueue.erl
```diff
@@ -330,11 +353,34 @@ out(MQ = #mqueue{q = Q, p_credit = 0}) ->
...
+
+dec_bytes(#mqueue{byte_size = BytesSize}, Msg) ->
+    BytesSize1 = BytesSize - msg_bytes(Msg),
+    true = BytesSize1 >= 0,
```
keynslug: Nit. I'd argue it's not worth crashing a session if we got accounting wrong for some reason. Perhaps live it as test-profile-only assertion? That said I'd also suggest to just manually inline this piece of logic into the call sites.

apps/emqx/src/emqx_message.erl
```diff
@@ -182,6 +183,12 @@ estimate_size(#message{topic = Topic, payload = Payload}) ->
     TopicLengthSize = 2,
     FixedHeaderSize + VarLenSize + TopicLengthSize + TopicSize + PacketIdSize + PayloadSize.
 
+-spec payload_size(emqx_types:message()) -> non_neg_integer().
+payload_size(#message{payload = Payload}) when is_binary(Payload); is_list(Payload) ->
```
keynslug:
```suggestion
payload_size(#message{payload = Payload}) when is_binary(Payload) orelse is_list(Payload) ->
```

Nit. Again, not sure there's any point being defensive here.

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -787,6 +793,26 @@ insert_inflight(PktId, Msg, Inflight) ->
...
+        0,
+        Inflight
+    ).
+
+inflight_val_bytes(#inflight_data{message = Msg}) ->
+    emqx_message:payload_size(Msg);
+inflight_val_bytes(#message{} = Msg) ->
+    emqx_message:payload_size(Msg);
+inflight_val_bytes(_Val) ->
+    0.
```
keynslug: Nit. Should be dead code?

apps/emqx/src/emqx_session_mem.erl
```diff
@@ -545,7 +551,7 @@ puback(ClientInfo, PacketId, Session = #session{inflight = Inflight}) ->
     case emqx_inflight:lookup(PacketId, Inflight) of
         {value, #inflight_data{phase = wait_ack, message = #message{qos = ?QOS_1} = Msg}} ->
             Inflight1 = emqx_inflight:delete(PacketId, Inflight),
-            Session1 = Session#session{inflight = Inflight1},
+            Session1 = set_inflight(Session, Inflight1),
```
keynslug: I think this defeats the whole purpose: we keep _cached_ `inflight_payload_bytes` but recompute it on each PUBACK. I imagine it should be:
1. Either we don't cache and compute on-demand, in `info(total_payload_bytes, Session)` clause.
2. Or cache it and adjust it carefully: after all here we have `Msg` already so it's not clear why we need to recompute by scanning whole inflight.

Same applies to most other sites using `set_inflight/2` function.

hjianbo: Thanks so much for your time! I've simplified the implementation now. If we only calculate the payload's iolist_size, it shouldn't be that expensive, so I changed it to calculation on-demand to keep the implementation simpler.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,812 @@
...
+    local_top(Opts);
+local_top(Count, SortBy) ->
+    scan_local(Count, SortBy).
+
+%%--------------------------------------------------------------------
+%% gen_server callbacks
+%%--------------------------------------------------------------------
+
+init([]) ->
+    Conf = put_conf(emqx_config:get([sysmon, session], ?DEFAULT_CONF)),
```
keynslug: Nit. Putting contents of a persistent term into another persistent term looks to me as a premature optimization. Perhaps start with plain `emqx_config:get/2` where this configuration is needed? As far as I understand it's called each time "emit stats" timer fires so not exactly a hot loop?

apps/emqx/src/emqx_mqueue.erl
```diff
@@ -55,6 +57,7 @@
     out/1,
     stats/1,
     dropped/1,
+    bytes_size/1,
```
keynslug: Nit. Perhaps name it `payload_bytes/1` (this and the record field as well)? So it's consistent with how it gets exposed in user interfaces.

apps/emqx/src/emqx_schema.erl
```diff
@@ -1644,6 +1650,18 @@ fields("sysmon") ->
                 }
             )}
     ];
+fields("sysmon_session") ->
+    [
+        {"buffered_payload_high_watermark",
```
keynslug: Nit. Also, maybe `total_payload_bytes_high_watermark`?

apps/emqx/src/emqx_schema.erl
```diff
@@ -3231,6 +3251,11 @@ validate_max_packet_size(Siz) when is_integer(Siz) ->
 validate_max_packet_size(_SizStr) ->
     {error, invalid_packet_size}.
 
+validate_session_buffered_payload_high_watermark(Bytes) when is_integer(Bytes), Bytes >= 0 ->
+    ok;
+validate_session_buffered_payload_high_watermark(_Bytes) ->
+    {error, #{cause => session_buffered_payload_high_watermark_out_of_range, minimum => 0}}.
```
keynslug: Nit. Probably we can generalize it to `validate_positive_bytesize/1`? Schema error should contain the field name anyway so this wordy `cause` looks superfluous.

apps/emqx/test/emqx_cm_SUITE.erl
```diff
@@ -104,6 +104,36 @@ t_get_set_chan_stats(_) ->
...
+    TestPid = self(),
+    ok = meck:new(emqx_session_buffer_mon, [passthrough, no_link]),
+    ok = meck:expect(
+        emqx_session_buffer_mon,
+        maybe_log,
+        fun(ClientId0, ChanPid0, Stats0) ->
+            TestPid ! {session_buffer_check, ClientId0, ChanPid0, Stats0},
+            ok
+        end
+    ),
```
keynslug: Nit. Perhaps `emqx_cth_log_capture` could be used instead of mocking?

apps/emqx/test/emqx_log_throttler_SUITE.erl
```diff
@@ -166,6 +171,29 @@ t_throttle_add_new_msg(_Config) ->
...
+                },
+                5000
+            ),
+            ?assert(emqx_log_throttler:allow(?THROTTLE_SESSION_BUFFER_MSG, <<"c1">>)),
+            ?assertNot(emqx_log_throttler:allow(?THROTTLE_SESSION_BUFFER_MSG, <<"c1">>)),
+            ?assertNot(emqx_log_throttler:allow(?THROTTLE_SESSION_BUFFER_MSG, <<"c2">>))
+        end,
+        []
+    ).
+
```
keynslug: Nit. We probably don't want to add testcase per each throttled message class.

apps/emqx/test/emqx_session_mem_SUITE.erl
```diff
@@ -122,6 +122,35 @@ t_session_stats(_) ->
...
+            })
+        )
+    ),
+    MqueueBytes = emqx_message:payload_size(MqueueMsg),
+    InflightBytes = emqx_message:payload_size(InflightMsg),
+    ?assertEqual(MqueueBytes + InflightBytes, maps:get(total_payload_bytes, Stats)),
+    ?assertNot(maps:is_key(mqueue_bytes, Stats)),
+    ?assertNot(maps:is_key(inflight_bytes, Stats)),
+    ?assertNot(maps:is_key(buffered_bytes, Stats)).
+
```
keynslug: Q: What those assertions are about?

apps/emqx/test/emqx_session_mem_SUITE.erl
```diff
@@ -122,6 +122,35 @@ t_session_stats(_) ->
...
+        emqx_inflight:new(10)
+    ),
+    Stats = maps:from_list(
+        emqx_session_mem:stats(
+            session(#{
+                mqueue => MQueue,
+                inflight => Inflight
+            })
+        )
+    ),
```
keynslug, line 136: This constructs "session-under-test" in a weird way, why not use `emqx_session_mem` APIs instead? E.g. `publish/3` / `puback/3` and `enqueue/3`. This testsuite is supposed to test `emqx_session_mem` API surface after all.

apps/emqx/src/emqx_session_tool.erl
```diff
@@ -316,27 +324,32 @@ maybe_sleep(_SleepMs) ->
     ok.
 
 %% Build the result row. Extras are resolved here, once per winner, by
-%% re-reading the cached info map. If the session has since disconnected
+%% re-reading the cached info/stats maps. If the session has since disconnected
 %% the info row is gone and extras come back empty (best effort).
-resolve_row({Value, ClientId, ChanPid}, Metric, ExtraKeys) ->
+resolve_row({Value, ClientId, ChanPid}, Metric, ExtraKeys, ExtraStats) ->
```
keynslug: I understand this is pre-existing logic, but: this has potential to make results wildly inconsistent (e.g. `#{total_payload_bytes => 42000, inflight_cnt => 0, mqueue_length => 0}`). Those are usually smallints so should be cheap to attach at `heap_offer/3` time as a small tuple? And since these are under API user's control, they supposedly know the consequences of abusing this `extra_*` feature.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,812 @@
...
+    {ok, #{
+        conf => Conf,
+        top => undefined,
+        top_status => #{status => idle}
+    }}.
+
+handle_call({run_top, Opts0}, _From, State = #{top := undefined}) ->
```
keynslug: I think putting cluster-wide collector and local session scanner into a single process complicates things a lot. I'd suggest:
1. To separate into 2 different modules and 2 different processes.
2. Move File IO and CSV aspects into `emqx_mgmt_cli`, for example by allowing to provide a completion callback.

apps/emqx/src/emqx_session_buffer_mon.erl
```diff
@@ -0,0 +1,812 @@
...
+        role := collector | worker,
+        out => file:name_all(),
+        count := pos_integer(),
+        sort := sort_by(),
+        batch_size := pos_integer(),
+        sleep_ms := non_neg_integer(),
+        scan_id := term(),
+        started_at := integer(),
+        initiator := node(),
+        collector := node(),
```
keynslug: Q: Is it possible for those to be different? If not, perhaps better to simplify?

keynslug (commented):
Mostly nits.

apps/emqx/src/emqx_session_tool.erl
```diff
@@ -294,43 +309,57 @@ read_metric(Metric, Stats) when is_list(Stats) ->
...
-            gb_sets:add(Candidate, Heap);
+            heap_add(RankKey, Stats, ExtraStats, Heap);
         false ->
-            {{SmallestValue, _, _}, Heap1} = gb_sets:take_smallest(Heap),
-            {CandValue, _, _} = Candidate,
-            case CandValue > SmallestValue of
-                true -> gb_sets:add(Candidate, Heap1);
+            {{SmallestValue, SmallestClientId, SmallestPid, _}, Heap1} =
+                gb_sets:take_smallest(Heap),
+            SmallestRankKey = {SmallestValue, SmallestClientId, SmallestPid},
```
keynslug: Nit. I'm not sure if it even matters in practice for this tooling, but: if we want to keep more natural ascending ordering by ClientIDs on ties then I guess it's possible to use `{-Value, ClientId, ChanPid}` as key and employ `take_largest/1` instead.

apps/emqx/src/emqx_session_top_collector.erl
```diff
@@ -0,0 +1,395 @@
...
+                msg => session_top_cancel_failed,
+                nodes => Nodes,
+                class => Class,
+                reason => Reason
+            })
+    end.
+
+schedule_timeout(Top = #{scan_id := ScanId}) ->
+    TRef = erlang:send_after(?TOP_TIMEOUT, self(), {top_scan_timeout, ScanId}),
+    Top#{timeout_timer => TRef}.
```
keynslug: Q: Is this timeout necessary? A user has no control over it as it is hard-coded, but it has control over "scanning pace" and I can easily imagine fat nodes with a lot of sessions going over the timeout. On the other hand the user always can cancel an ongoing collection, perhaps removing the hard timeout and let the user decide when they want to cancel is a better option?

apps/emqx/src/emqx_session_top_scanner.erl
```diff
@@ -0,0 +1,176 @@
...
+    #{acc := Acc0} = Scan0,
+    try emqx_session_tool:scan_acc(Acc0) of
+        {continue, Acc1} ->
+            _ = erlang:garbage_collect(),
+            Scan = Scan0#{acc := Acc1},
+            {noreply, schedule_tick(State#{scan := Scan}, sleep_ms(Scan))};
+        {done, Acc1} ->
+            Rows = emqx_session_top_collector:normalize_rows(
+                emqx_session_tool:scan_acc_rows(Acc1)
+            ),
```
keynslug: Nit. As this scan result is going to be sent to potentially remote collector running different EMQX version, it looks like a wrong place to do any sort of data "normalization". Especially since `complete_node/3` calls the same function on the collector's side anyway.

apps/emqx/src/emqx_session_top_scanner.erl
```diff
@@ -0,0 +1,176 @@
...
+-export([
+    init/1,
+    handle_call/3,
+    handle_cast/2,
+    handle_info/2
+]).
+
+-define(DEFAULT_SCAN_BATCH_SIZE, 1000).
+-define(DEFAULT_SCAN_SLEEP_MS, 1).
+-define(SESSION_TOP_EXTRA_STATS, [mqueue_len, total_payload_bytes, inflight_cnt]).
```
keynslug: Nit. Taking forward compatibility into account, this ideally needs to be provided in `start_scan(Opts)`.

apps/emqx/src/emqx_session_top_collector.erl
```diff
@@ -0,0 +1,395 @@
...
+
+top_rows(Rows, SortBy, Count) ->
+    Sorted = lists:sort(
+        fun(RowA, RowB) -> maps:get(SortBy, RowA) >= maps:get(SortBy, RowB) end,
+        Rows
+    ),
+    lists:sublist(Sorted, Count).
+
+normalize_row(#{mqueue_length := _, total_payload_bytes := _, inflight_count := _} = Row) ->
+    Row;
```
keynslug: Nit. This is probably an artifact of calling `normalize_rows/1` twice in different sites?

apps/emqx/src/emqx_session_top_collector.erl
```diff
@@ -0,0 +1,395 @@
...
+    cancel_timeout_timer(Top),
+    {reply, {ok, cancelled}, State#{
+        top := undefined,
+        top_status := cancelled_status(Top, cancelled)
+    }};
+handle_call(cancel, _From, State) ->
+    {reply, {error, not_running}, State};
+handle_call(status, _From, State) ->
+    {Status, State1} = take_status(State),
+    {reply, Status, State1};
```
keynslug: Nit. It's a bit counter-intuitive that function named `status/0` is not idempotent and has side-effects.

keynslug (commented):
Great stuff. Only a handful of minor questions / remarks.

zmstone (approved):

keynslug (approved):

ieQu1 (approved):

PR #17497, opened 2026-06-05
fix(resource): count actions.executed per invocation

zmstone:
Fixes: https://emqx.atlassian.net/browse/EMQX-15328
Release version: 6.1.3, 6.2.2

## Summary

QA-reported metric-correctness fix.  In non-batch mode (`batch_size = 1`) the documented invariant is `actions.executed == actions.messages`.  Today, `actions.executed` falls behind because it was incremented in the buffer-worker telemetry flush handler (one bump per flush window) while `actions.messages` was incremented by the flush's aggregated message count.  K consecutive single-message completions aggregated into one flush would bump `actions.executed` by 1 and `actions.messages` by K.

This PR decouples the two counters:

* `actions.executed` is now bumped once per `on_query` / `on_batch_query` invocation, from `apply_query_fun/9` in `emqx_resource_buffer_worker.erl`.  One execution per batch (regardless of size), one execution per retry, one execution per failure.
* `actions.messages` continues to be derived from the existing telemetry-flush path (renamed helper `inc_actions_messages/2`).

Most relevant files:

* `apps/emqx_resource/src/emqx_resource_metrics.erl` — split helpers; new `inc_actions_executed/2`.
* `apps/emqx_resource/src/emqx_resource_buffer_worker.erl` — call sites that bump per invocation.
* `apps/emqx_bridge/test/emqx_bridge_v2_SUITE.erl` — four new test cases covering single-mode equality, failure counting, batch mode, and namespaced actions.

No behaviour change beyond the metric values.  Metric names and the `actions.messages` semantic are unchanged.

## PR Checklist
- [x] Change log for changes visible by users has been added to `changes/ee/fix-17497.en.md`
- [x] Schema changes are backward compatible or intentionally breaking (no schema change)

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -549,7 +549,7 @@ retry_inflight_sync(Ref, QueryOrBatch, Data0) ->
     IsSimpleQuery = false,
     QueryOpts = #{simple_query => IsSimpleQuery},
     Result = call_query(force_sync, Id, Index, Ref, QueryOrBatch, QueryOpts),
-    {Decision, PostFn, DeltaCounters} =
+    {Decision, PostFn, DeltaCounters0} =
         case QueryOrBatch of
             ?QUERY(ReplyTo, _, HasBeenSent, _ExpireAt, RequestContext, TraceCtx) ->
                 Reply = ?REPLY(ReplyTo, HasBeenSent, Result, RequestContext, TraceCtx),
@@ -557,6 +557,7 @@ retry_inflight_sync(Ref, QueryOrBatch, Data0) ->
             [?QUERY(_, _, _, _, _, _) | _] = Batch ->
                 batch_reply_caller_defer_metrics(Id, Result, Batch, IsSimpleQuery)
         end,
+    DeltaCounters = delta_with_actions_executed(Decision, Result, DeltaCounters0),
     Queries =
         case QueryOrBatch of
             ?QUERY(_, _, _, _, _, _) ->
@@ -792,7 +793,8 @@ do_flush(
     QueryOpts = #{inflight_tid => InflightTID, simple_query => false},
     Result = call_query(async_if_possible, Id, Index, Ref, Request, QueryOpts),
     Reply = ?REPLY(ReplyTo, HasBeenSent, Result, RequestContext, TraceCtx),
-    {Decision, PostFn, DeltaCounters} = reply_caller_defer_metrics(Id, Reply, QueryOpts),
+    {Decision, PostFn, DeltaCounters0} = reply_caller_defer_metrics(Id, Reply, QueryOpts),
+    DeltaCounters = delta_with_actions_executed(Decision, Result, DeltaCounters0),
     PostFn(result_context([Request])),
     Data1 = aggregate_counters(Data0, DeltaCounters),
     case Decision of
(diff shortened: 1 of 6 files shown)


apps/emqx_resource/src/emqx_resource_metrics.erl
```diff
@@ -594,9 +597,17 @@ on_delivery_finished(Result, ActionResId) ->
...
     emqx_metrics:inc_global('actions.messages', Val).
+
+-spec inc_actions_executed(emqx_config:maybe_namespace(), pos_integer()) -> ok.
+inc_actions_executed(?global_ns, 1) ->
+    emqx_metrics:inc_global('actions.executed');
+inc_actions_executed(?global_ns, N) when is_integer(N), N > 1 ->
+    emqx_metrics:inc_global('actions.executed', N);
+inc_actions_executed(Namespace, 1) when is_binary(Namespace) ->
+    emqx_metrics:inc(Namespace, 'actions.executed');
+inc_actions_executed(Namespace, N) when is_binary(Namespace), is_integer(N), N > 1 ->
```
thalesmg: maybe no need to restrict `N > 1`?

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1434,6 +1441,16 @@ collect_rule_trigger_times(?QUERY(_, _, _, _, _, #{rule_trigger_ts := Time}), Ac
 collect_rule_trigger_times(?QUERY(_, _, _, _, _, _), Acc) ->
     Acc.
 
+%% Per-`call_query' delta that bumps `actions.executed' iff the connector callback
+%% was actually invoked.  Bail-outs in `call_query/6' that short-circuit before
+%% `apply_query_fun/9' (resource stopped, not connected, not found) must not count.
```
thalesmg: it's also not invoked if [`pre_query_channel_check`](https://github.com/emqx/emqx/blob/5ae2568002d4442985b2af2337c313645c119cf9/apps/emqx_resource/src/emqx_resource_buffer_worker.erl#L1450) returns an error.

apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1434,6 +1441,16 @@ collect_rule_trigger_times(?QUERY(_, _, _, _, _, #{rule_trigger_ts := Time}), Ac
...
+delta_with_actions_executed(?RESOURCE_ERROR_M(Kind, _), DeltaCounters) when
+    Kind =:= stopped; Kind =:= not_connected; Kind =:= not_found
+->
+    DeltaCounters;
+delta_with_actions_executed(_Result, DeltaCounters) ->
+    DeltaCounters#{actions_executed => 1}.
```
thalesmg, line 1460: also note that metrics are [only added to the state if the decision is final](https://github.com/emqx/emqx/blob/5ae2568002d4442985b2af2337c313645c119cf9/apps/emqx_resource/src/emqx_resource_buffer_worker.erl#L594).

the added test case only tests unrecoverable errors, which is indeed a final decision.

is the intention of this metric to track recoverable attempts as well?

zmstone: there is no explicit requirement in this regard.
i'd say it's acceptable to increment when there is a final result.

thalesmg: just to emphasize: it's currently bumped **only** when the decision is final.

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15669)

PR #17471, opened 2026-06-03
chore: adjust namespaced admin permissions on apis

thalesmg:
Release version:
6.0.3, 6.1.2, 6.2.1

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files

PR #17499, opened 2026-06-05
chore: adjust namespaced admin permissions on apis (v2)

zmstone:
Release version:
6.0.3, 6.1.2, 6.2.1

## Summary

Supersedes #17471 (original author @thalesmg is on vacation). Carries over all of his commits and adds review fixes on top.

### Original change (from #17471)

- `emqx_dashboard_rbac`: drop the clause that let any namespaced administrator `GET` every endpoint. After this PR, namespaced admins fall through to the per-API allowlist (Connector / Bridge-v2 / Rule / Trace / Data Backup / dashboard self-service) or are denied.
- `emqx_mgmt_api_clients`, `emqx_mgmt_api_subscriptions`: wire a `minirest` filter on every endpoint that rejects requests with `403 FORBIDDEN` whenever the authenticated actor is bound to a non-global namespace. This blocks namespaced viewers and namespaced API keys too, not just admins.
- `emqx_utils/include/emqx_http_api.hrl`: new `?FORBIDDEN(Reason)` macro.

### Review fixes added in this PR

1. **Shared filter** — the duplicated `resolve_namespace/2` + `filter/2` pair in both API modules is replaced by `emqx_dashboard:require_global_namespace_filter/2`. Any future global-only endpoint can opt in with a one-line `filter => fun emqx_dashboard:require_global_namespace_filter/2`. The error body now actually mentions namespaces: `"This endpoint is not available to namespaced users"`.
2. **Test expansion** — the namespace-forbidden tests in `emqx_mgmt_api_clients_SUITE` and `emqx_mgmt_api_subscription_SUITE` now drive both the namespaced administrator and the namespaced viewer role, and on the clients suite cover every mutating endpoint the filter is wired on (kickout, bulk kickout, subscribe/unsubscribe single+bulk, set-keepalive).
3. **RBAC regression net** — the deleted `t_namespaced_user_permissions` is restored in inverted form in `emqx_dashboard_admin_SUITE`. It enumerates loaded GET handlers, partitions by RBAC verdict, and asserts the allowed module set is exactly the namespace-scoped whitelist. Catches both accidental leaks (a new global module becomes accessible) and accidental removals (one of the legitimately allowed modules stops working).
4. **Changelog rewrite** — see `changes/ee/breaking-17499.en.md`. The original opened with "Previously, such admins" with no antecedent and only described the admin half of the change. Rewritten to spell out: (a) what namespaced admins lose; (b) that `/clients*` and `/subscriptions` reject every non-global request regardless of role; (c) that other global read-only endpoints (`/nodes`, `/stats`, `/metrics`, `/listeners`, `/banned`, ...) remain reachable by namespaced viewers via the generic viewer-GET rule, with tightening tracked as a follow-up.

### Scope note

Per the original PR conversation ("Todo: remove generic namespaced viewer perms") the viewer-GET bypass on the rest of the global APIs is left for a follow-up so this PR stays focused. The changelog calls that out so customers understand exactly what is and isn't tightened today.

## PR Checklist


zmstone:
will go with this change for 6.0~6.3 https://github.com/emqx/emqx/pull/17501
revisit in future version if to restrict more for namespace admins.

e.g.
do not allow namespace admin to view other namespace clients will have to be fixed by refactoring the `clients` and `clients_v2` APIs to traverse `emqx_mt` tables instead of the global talbes.

PR #17406, opened 2026-05-29
fix(trace): filter events by namespace

thalesmg:
Release version:
6.0.3, 6.1.2, 6.2.1

## Summary

Restricts traces of types topic, clientid and ip address to log only events from the namespace of the admin who created them.  Traces of type ruleid were already namespaced.

If the admin is a global one, then it sees events from all namespaces (except for ruleids, which the previous behavior was already to limit it to events from the global ns only).

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files

PR #17501, opened 2026-06-05
fix: block namespaced users from reading cross-namespace MQTT messages

zmstone:
Release version:
6.0.3, 6.1.2, 6.2.1

## Summary

Part 1 of the namespaced-admin permission split. Targeted at the security-critical leak only: a namespaced dashboard user (administrator or viewer) could read MQTT message content produced outside of their namespace via several REST endpoints.

Part 2 (deferred) — broader restrictions on listing clients/subscriptions metadata — is tracked separately and may slip to a later release.

### What this PR changes

- **MQTT message endpoints** are now rejected with `403 FORBIDDEN` for non-global callers. The minirest filter `emqx_mgmt_api:require_global_namespace_filter/2` is wired on:
  - `GET /clients/:clientid/mqueue_messages`
  - `GET /clients/:clientid/inflight_messages`
  - `GET /mqtt/retainer/messages`
  - `GET|DELETE /mqtt/retainer/message/:topic`
  - `GET /mqtt/delayed/messages`
  - `GET|DELETE /mqtt/delayed/messages/:node/:msgid`
  - `DELETE /mqtt/delayed/messages/:topic`
- **Trace API** is now namespace-scoped per trace, preserving the per-namespace event filtering merged in #17406:
  - `GET /trace` lists only the caller's namespace.
  - `/trace/:name`, `/download`, `/log`, `/log_detail`, `/stop` return `404` for cross-namespace traces (so other-namespace existence is not leaked).
  - `DELETE /trace` (clear-all) is reserved for global admins; namespaced callers get `403`.
  - `emqx_mgmt_cli:traces([\"list\"])` updated for the new request-map handler signature.
- **Backup download** (`GET /data/files/:filename`) was already gated to the global administrator by `emqx_mgmt_api_data_backup:is_global_admin/1`; unchanged.

### What this PR does NOT change

- The broad restriction on `/clients*` and `/subscriptions` (block any non-global request regardless of role) and the RBAC \"namespaced admin can GET anything\" bypass — those are Part 2.

### Verification

- `make` clean, `static_checks` ok, elvis ok on changed files.
- CT:
  - `apps/emqx_management/test/emqx_mgmt_api_clients_SUITE:t_namespaced_user_message_endpoints` ✓
  - `apps/emqx_management/test/emqx_mgmt_api_trace_SUITE:t_namespaced_user_cross_ns_isolation` ✓

## PR Checklist


ieQu1 (dismissed):

apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
```diff
@@ -258,12 +258,32 @@ do_check_rbac(
         #{bindings := #{username := Username}} -> true;
         _ -> false
     end;
+do_check_rbac(
+    #{?role := ?ROLE_SUPERUSER, ?namespace := Namespace, ?actor := Username},
+    Req,
+    ?DASHBOARD_API(post, change_pwd)
+) when is_binary(Namespace) ->
+    %% Namespaced administrators may rotate their own password and may
+    %% reset the password of any other user in their namespace -- it's
```
hjianbo: This RBAC change allows a namespaced admin to call `change_pwd` for another user in the same namespace, but the HTTP handler still calls `emqx_dashboard_admin:change_password(TargetUser, OldPwd, NewPwd)`, which validates `OldPwd` against the target user current password.

So this does not actually implement an admin password reset flow unless the tenant admin already knows the target user old password. I think we should either keep this endpoint self-only for namespaced admins, or add a real trusted reset path and cover both same-namespace success and cross-namespace denial with HTTP tests.

apps/emqx_retainer/src/emqx_retainer_api.erl
```diff
@@ -66,6 +66,7 @@ schema(?PREFIX) ->
 schema(?PREFIX ++ "/messages") ->
     #{
         'operationId' => '/messages',
+        filter => fun emqx_mgmt_api:require_global_namespace_filter/2,
```
hjianbo, line 69: The new global-namespace filter is also applied to retained message APIs, and delayed message APIs get the same treatment in this PR. However, the added tests currently cover only client mqueue/inflight endpoints and trace APIs.

Could we add namespaced admin/viewer HTTP coverage for the retained and delayed endpoints as well? Those are part of the main leak surface fixed by this PR, and tests would help prevent these schema-level filters from being accidentally removed later.

changes/ee/fix-17501.en.md
```diff
@@ -0,0 +1,4 @@
+Block namespaced dashboard users from reading MQTT message content that crosses namespace boundaries.
+
+- Endpoints that return MQTT payloads outside the caller's namespace now reject non-global callers with `403 FORBIDDEN`: `GET /clients/:clientid/mqueue_messages`, `GET /clients/:clientid/inflight_messages`, `GET /mqtt/retainer/messages`, `GET|DELETE /mqtt/retainer/message/:topic`, `GET /mqtt/delayed/messages`, `GET|DELETE /mqtt/delayed/messages/:node/:msgid`, `DELETE /mqtt/delayed/messages/:topic`. Previously a namespaced user could read or delete messages produced by other namespaces.
```
hjianbo: The changelog lists `GET /mqtt/retainer/messages` and `GET|DELETE /mqtt/retainer/message/:topic`, but the filter is attached at the `/mqtt/retainer/messages` path level, so `DELETE /mqtt/retainer/messages` is also blocked for namespaced users.

Could we include `DELETE /mqtt/retainer/messages` in the change note so the release note matches the actual behavior?

hjianbo (approved):

apps/emqx_management/src/emqx_mgmt_api.erl
```diff
@@ -807,6 +811,22 @@ b2i(Bin) when is_binary(Bin) ->
 b2i(Any) ->
     Any.
 
+%% Minirest filter: reject requests whose authenticated actor is bound to a
+%% non-global namespace. Used on endpoints that expose MQTT message content
+%% (mqueue/inflight/retained/delayed payloads, trace logs) which would
+%% otherwise leak across tenants.
+-spec require_global_namespace_filter(Req, map()) ->
```
savonarola, line 818: @zmstone 
Why do we choose to impement this particular limitation with filters? Not. e.g. in `emqx_dashboard_rbac` where we already have some namespace limitations?

zmstone: emqx_dashboard_rbac seem to apply only for token logins
filter applies to api-keys.

the rbac clauses for namespace arguably fit better as filters.

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15022)

apps/emqx_management/src/emqx_mgmt_api.erl
zmstone: i was wrong, rbac apply for both login tokens and api-keys.

i'll see if it's better remover the filters and move them to emqx_dashboard_rbac

PR #17507, opened 2026-06-08
fix(gateway): prevent unauthenticated publish paths

hjianbo:
Release version:
6.2.x

## Summary

Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/74.
Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/75.
Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/82.

This fixes gateway paths that could publish or subscribe before the client had gone through the expected authentication posture.

For MQTT-SN, the idle-state QoS -1 publish path now builds client info with the existing fixed negative-QoS client ID, runs the normal gateway authentication flow, then checks publish authorization before calling the broker. Since MQTT-SN QoS -1 is forwarded as a QoS 0 broker publish, authorization is checked with QoS 0 semantics.

For NATS, the gateway now follows `security_profile` when listener authentication is enabled but no NATS internal authenticator or gateway authenticator is configured. The legacy profile keeps the current anonymous behavior, while the hardened profile advertises `auth_required=true`, rejects PUB/SUB before CONNECT, and rejects CONNECT with `not_authorized`. Explicit anonymous access remains available by setting the listener `enable_authn=false`.

For STOMP, SEND and SUBSCRIBE frames are rejected before CONNECT completes, including transactional SEND frames. Explicit anonymous access after CONNECT remains available when listener authentication is disabled.

For CoAP, connectionless `/ps` publish and observe requests now authenticate before entering pub/sub handling. The request identity is built from the `/ps` URI query when present and is scoped to that request, so a successful connectionless request does not authenticate later requests without credentials. Under the hardened profile, such requests are rejected when no authentication is configured or when credentials are bad; legacy mode and explicit listener `enable_authn=false` preserve the intended anonymous paths.

The tests cover legacy allowed behavior, hardened rejection, bad credential rejection, successful authenticated publish/subscribe, request-scoped connectionless CoAP authentication, explicit listener-authn-disabled anonymous behavior, and the existing authenticated MQTT-SN/NATS/STOMP paths that are runnable in this environment.

## PR Checklist

### DIFF
--- a/apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
+++ b/apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
@@ -264,7 +264,9 @@ enrich_clientinfo(
     {ok, NPacket, NClientInfo} = emqx_utils:pipeline(
         [
             fun maybe_assign_clientid/2,
-            %% FIXME: CALL After authentication successfully
+            %% Mountpoint is evaluated again after successful authentication in
+            %% emqx_gateway_ctx:authenticate/2, because authentication results
+            %% may add client attributes used by mountpoint templates.
             fun fix_mountpoint/2
         ],
         Packet,
@@ -453,7 +455,7 @@ handle_in(
 handle_in(
     Publish =
         ?SN_PUBLISH_MSG(
-            #mqtt_sn_flags{
+            Flags = #mqtt_sn_flags{
                 qos = ?QOS_NEG1,
                 topic_id_type = TopicIdType
             },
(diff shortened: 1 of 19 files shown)


apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -1127,12 +1115,55 @@ check_pub_authz(
...
+            ?tp(info, idle_negative_qos_publish_rejected, #{
+                topic => TopicName,
+                return_code => RC
+            }),
+            ok
+    end.
+
+authenticate_idle_negative_qos(Packet, Channel = #channel{conninfo = ConnInfo}) ->
+    NConnInfo = ConnInfo#{clientid => ?NEG_QOS_CLIENT_ID},
+    {ok, _Packet, NChannel} = enrich_clientinfo(Packet, Channel#channel{conninfo = NConnInfo}),
```
savonarola, line 1161: Should we fix the FIXME in `enrich_clientinfo`: `%% FIXME: CALL After authentication successfully`?

hjianbo: I changed the FIXME into a regular explanatory comment. Mountpoint is now evaluated twice for these gateways: once when constructing the initial clientinfo, and again after successful authentication via `emqx_gateway_ctx:authenticate/2`, so authentication results such as `client_attrs` can be used by mountpoint templates. I also added gateway-specific test coverage for this behavior.

savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15088)

PR #17509, opened 2026-06-08
ci(build): refresh dynlibs in builder before packaging

zmstone:
## Summary

Builder Docker images are not rebuilt on a schedule, so the system libraries they ship (notably openssl / libssl / libcrypto) can be weeks-to-months stale by the time we link or bundle them into release packages. A reporter noticed our zip / tarball packages shipping stale openssl on this account.

Add a small \`scripts/refresh-builder-dynlibs.sh\` that runs \`apt-get update && apt-get upgrade\` (or \`dnf upgrade\` / \`yum update\` depending on the distro family) inside the builder container right before \`make\`, so each package build picks up current security-patched dynlibs from the distro repos.

Failure is non-fatal: if a distro repo is EOL (e.g. CentOS 7) or unreachable, the script logs a warning and proceeds with the image as-is rather than blocking the build. \`REFRESH_DYNLIBS=no\` skips the step for offline / sandboxed local builds.

Only applied on the docker-run path of \`scripts/buildx.sh\` — not the native path, where it would modify the developer's own host.

## PR Checklist
- [ ] Change log for changes visible by users has been added to \`changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md\` files

### DIFF
--- a/scripts/refresh-builder-dynlibs.sh
+++ b/scripts/refresh-builder-dynlibs.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Refresh dynamic libraries (notably openssl) in the builder image before
+# building a package. Builder images are not rebuilt on a schedule, so the
+# libs they ship can be weeks-to-months stale by the time we build against
+# them. Pull current security-patched versions from the distro repos so the
+# package links/bundles fresh shared objects.
+#
+# Failure is non-fatal: if the distro repos are EOL (e.g. CentOS 7) or
+# unreachable, we log a warning and proceed with the image as-is rather
+# than blocking the build.
+#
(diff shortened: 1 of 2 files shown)


scripts/refresh-builder-dynlibs.sh
thalesmg, line 1: should perhaps cleanup the package directories to reduce image size afterwards?

zmstone: the container will exit (deleted) after build.

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15522)

PR #17521, opened 2026-06-09
fix(redis): isolate Sentinel managers per resource

hjianbo:
Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/56

Release version: 5.10.4

## Summary

- Gives each Redis Sentinel connector resource its own runtime Sentinel manager by passing a per-resource `sentinel_manager_ref` to `eredis` and stopping that manager when the resource stops.
- Updates `emqx_redis` to consume the released `eredis_cluster` `0.9.0` tag, which depends on `eredis` `1.3.0`.
- Adds a regression CT case proving two Sentinel resources with different Sentinel endpoints/master credentials can coexist, and fixes the changelog filename for the PR sanity gate.

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ce/fix-17521.en.md`

### DIFF
--- a/apps/emqx_redis/mix.exs
+++ b/apps/emqx_redis/mix.exs
@@ -26,7 +26,7 @@ defmodule EMQXRedis.MixProject do
       {:emqx_connector, in_umbrella: true, runtime: false},
       {:emqx_resource, in_umbrella: true},
       {:eredis_cluster,
-       github: "emqx/eredis_cluster", ref: "34b55acb50f95fd6887c755d2b27b3170502d4e4"}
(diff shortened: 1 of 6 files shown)


hjianbo:
Update: this fix is still needed for r510. Multiple Redis Sentinel resources in the same EMQX node can still interfere with each other when they share the legacy eredis Sentinel singleton. The upstream dependency chain is now available as release tags (`eredis` 1.3.0 and `eredis_cluster` 0.9.0), so this PR can be reopened and updated to consume the released versions instead of temporary refs.

hjianbo:
GitHub does not allow reopening this PR because the head branch was deleted/recreated after it was closed. I restored the branch and continued the same fix in the replacement PR: https://github.com/emqx/emqx/pull/17538

PR #17522, opened 2026-06-09
fix(cm): periodically purge stale rows from global session registry

### LINKED ISSUE #17464: periodic purge of stale session registry
### What happened?

This is a follow up for https://github.com/emqx/emqx/pull/17424
The PR above was not complete.
For random client IDs, the stale registry record may leak forever.

Start a scanner to purge records if any of the conditions is met:
1. it's a local pid, and its not alive (erlang:is_process_alive/1 return false)
2. (core node only): if the pid is not local, but the node is no longer in the cluster (keep the record if the node is stopped, but still in the cluster).

periodic scan, do it at MOST 1 time every 10m, and not exceeding 500 ets:next calls per second, with controlled rate to avoid overloading the system.

meaning it will take 6 hours to complete scanning if system holds 10 million sessions.

_No response_

_No response_

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
# paste output here
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

zmstone:
Release version: 5.10.5

## Summary

Fixes #17464.

PR #17424 stopped some stale rows from accumulating in the global session registry, but the random-clientid case was missed: when a session owner dies without a clean unregister (brief network split, consensus timeout during down-event cleanup), the row stays forever if the same clientid never reconnects.

This change adds a periodic GC sweep on each core node that walks `emqx_session_registry` and removes rows whose owner pid is dead. The sweep is throttled to at most 500 rows/second per node and runs no more often than once every 10 minutes, so it has no measurable impact on broker throughput even on registries with millions of sessions.

## PR Checklist


apps/emqx/src/emqx_cm_registry_keeper.erl
```diff
@@ -141,38 +180,199 @@ terminate(_Reason, _State) ->
...
+        replicant ->
+            {keep, State};
+        _ ->
+            N = node(Pid),
+            #{gone_nodes := Gone, alive_nodes := Alive} = State,
+            case {sets:is_element(N, Gone), sets:is_element(N, Alive)} of
+                {true, _} ->
+                    {remote_orphan, State};
+                {_, true} ->
+                    {keep, State};
```
zmstone: swap the two clauses in case a node was gone then re-joined cluster.

zmstone: Addressed in ad5a789a — split `classify_pid` into core/replicant sub-functions, checked alive before gone, and added `normalize_node_caches/1` to subtract alive from gone after every remote classification so the invariant holds across chunks.

apps/emqx/src/emqx_cm_registry_keeper.erl
```diff
@@ -111,27 +153,25 @@ handle_call(_Request, _From, State) ->
...
+handle_info(start, State0) ->
+    %% ensure_sweep_keys/1 makes the per-sweep counter and node-cache keys
+    %% present, so a hot-loaded beam survives the first tick when the
+    %% pre-upgrade state only had next_clientid+timer_ref.
+    %%
+    %% Scheduled sweeps do NOT emit cm_registry_gc_started: it would just
+    %% be telemetry noise on quiet nodes, and in test mode with hist
+    %% enabled (1s cadence) it would keep unrelated suites' snabbkaffe
+    %% wait-for-silence ticking. cm_registry_gc_finished is emitted only
+    %% when the sweep actually deleted something.
```
thalesmg, line 165: micro nit: maybe move this closer to the actual trace point, since otherwise this comment doesn't appear to make sense at this point.

apps/emqx/src/emqx_cm_registry_keeper.erl
```diff
@@ -141,38 +181,234 @@ terminate(_Reason, _State) ->
...
-    lists:foreach(
-        fun(R) ->
-            case IsExpired(R) of
-                true ->
-                    mria:dirty_delete_object(?CHAN_REG_TAB, R);
-                false ->
-                    ok
-            end
+new_sweep_state(TimerRef) ->
+    (zero_state())#{timer_ref => TimerRef}.
```
thalesmg, line 189: nit:


```suggestion
    (zero_state())#{timer_ref := TimerRef}.
```

ieQu1 (approved):

apps/emqx/src/emqx_cm_registry_keeper.erl
```diff
@@ -141,38 +181,234 @@ terminate(_Reason, _State) ->
...
+do_run_chunk(Cursor, State0) ->
+    {NewNext, State1} = run_chunk_loop(Cursor, ?REGISTRY_GC_CHUNK_SIZE, State0),
+    case NewNext of
+        '$end_of_table' ->
+            maybe_emit_finished(sweep_summary(State1)),
+            TimerRef = send_delay_start(),
+            {noreply, State1#{next_clientid := undefined, timer_ref := TimerRef}};
+        Cid ->
+            %% ensure the next clientid is not in the cache
+            _ = erlang:garbage_collect(),
```
thalesmg, line 280: Q: could you clarify this?  i.e., what does it mean to be "in the cache" here, and how forcing a garbage collection in the current process prevents that?

apps/emqx/src/emqx_cm_registry_keeper.erl
```diff
@@ -141,38 +181,234 @@ terminate(_Reason, _State) ->
...
+run_chunk_loop(ClientId, Budget, State0) ->
+    Rows = mnesia:dirty_read(?CHAN_REG_TAB, ClientId),
+    Next = mnesia:dirty_next(?CHAN_REG_TAB, ClientId),
+    State1 = lists:foldl(
+        fun(R, S) -> process_row(ClientId, R, S) end,
+        State0,
+        Rows
+    ),
+    State2 = bump(scanned, State1),
+    run_chunk_loop(Next, Budget - 1, State2).
```
thalesmg, line 300: nit: this is so similar to `run_full_sweep`, maybe they could be unified with:


```suggestion
    NewBudget = 
        case Budget of
            infinity -> Budget;
            _ -> Budget - 1
        end,
    run_chunk_loop(Next, NewBudget, State2).
```

... and sync full sweep calls this with `infinity` budget.

apps/emqx/src/emqx_cm_registry_keeper.erl
```diff
@@ -141,38 +181,234 @@ terminate(_Reason, _State) ->
...
+                        {remote_orphan, State#{
+                            gone_nodes := sets:add_element(N, Gone)
+                        }};
+                    _ ->
+                        {keep, State#{
+                            alive_nodes := sets:add_element(N, Alive)
+                        }}
+                end
+        end,
+    {Verdict, normalize_node_caches(State1)}.
```
thalesmg, line 400: Q: shouldn't we normalize _before_ classifying the pid?  Otherwise, the cache might be already stale when starting a new chunk.

PR #17524, opened 2026-06-09
test(cm): stabilize open_session throttle flaky case

zmstone:
Release version: 6.0.3

## Summary

Fix the flaky CT case `emqx_cm_SUITE:t_open_session_throttled_on_inflight_local_cleanup`.

The test registers a dead pid via `emqx_cm:register_channel/3` to simulate the in-flight cleanup race the throttle in `emqx_cm:drop_stale_local_rows/2` is meant to detect. `register_channel/3` inserts into `?CHAN_CONN_TAB` and then casts `{registered, ChanPid}` to the `emqx_cm` gen_server, which monitors the pid. Because the pid is already dead, the monitor immediately fires `DOWN`, the `emqx_cm` pool removes the `?CHAN_CONN_TAB` row, and by the time `open_session/3` reaches `do_get_chann_conn_mod/2` the row is gone. The throttle no longer triggers and `open_session/3` returns `{ok, _}` instead of the expected `{error, client_id_unavailable}`.

Suspend `emqx_cm` (`sys:suspend/1`) around `register_channel/3` + `open_session/3` so the `{registered, DeadPid}` cast stays in the mailbox. No monitor is set up, no `DOWN` fires, and the chan-conn row stays in place — the throttle observes the conditions it is designed to detect. `sys:resume/1` runs in the `after` clause so the gen_server is always restored, even if the assertion fails.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_cm_SUITE.erl
+++ b/apps/emqx/test/emqx_cm_SUITE.erl
@@ -579,10 +579,25 @@ t_open_session_throttled_on_inflight_local_cleanup(_) ->
     ChanInfo = ?ChanInfo,
     #{conninfo := ConnInfo} = ChanInfo,
     ?assertReceive({'DOWN', MRef, process, DeadPid, _}),
-    ok = emqx_cm:register_channel(ClientId, DeadPid, ChanInfo#{conn_mod => emqx_connection}),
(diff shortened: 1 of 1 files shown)


thalesmg (approved):

hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15355)

PR #17530, opened 2026-06-09
feat(cluster_link): gate cluster linking behind non-community license

zmstone:
Release version:
6.0.3, 6.1.3, 6.2.2

## Summary

Cluster linking now requires a non-community license. The gate is at the actor layer, not at app boot:

- `emqx_cluster_link_app:start/2` always sets up the broker provider, routing hook and supervisor. Per-link actors are spawned only when `emqx_cluster:is_single_node_mode/0` returns `false`. Under a community license the configured links stay inert and a single warning is logged listing the link names that were left inactive.
- `emqx_cluster_link_config:pre_config_update/3` rejects `create` / `update` (and raw catch-all) requests that would result in any enabled link, returning `{error, single_node_license}`. `delete`/`disable` are always accepted.
- The REST handler maps `single_node_license` to a 400 with a clear message ("Cluster linking is not available under the community (single-node) license. Load a non-community license and retry.") instead of the generic "failed: <reason>" formatting.

After loading a non-community license, the user toggles existing links from disabled to enabled via REST/UI; no node restart is needed.

`emqx_cluster_link` now depends on `cluster_mode` having been written by `emqx_license_checker`. This is enforced by an explicit runtime dependency in `emqx_machine_boot:runtime_deps/0`, so the topsort always boots `emqx_license` first.

New CT suite `emqx_cluster_link_license_SUITE` covers:
- Boot under community with an enabled link configured → no per-link actor; re-boot under normal → actor present.
- `create_link` with `enable = true` under community → `single_node_license`.
- `update_link` to enabled under community → `single_node_license`; `disable` and `delete` still allowed.
- Simulated license upgrade → `update_link` to enabled succeeds and the actor starts.

## PR Checklist

### DIFF
--- a/apps/emqx_cluster_link/test/emqx_cluster_link_license_SUITE.erl
+++ b/apps/emqx_cluster_link/test/emqx_cluster_link_license_SUITE.erl
@@ -19,15 +19,9 @@ all() ->
     emqx_common_test_helpers:all(?MODULE).
 
 init_per_suite(Config) ->
-    %% Disable the default MQTT listener so the suite can run on a host
-    %% where ports 1883/8083 are already taken.
     Apps = emqx_cth_suite:start(
         [
-            {emqx,
-                "listeners.tcp.default.enable = false\n"
-                "listeners.ssl.default.enable = false\n"
-                "listeners.ws.default.enable = false\n"
-                "listeners.wss.default.enable = false\n"},
+            emqx,
             emqx_conf,
             emqx_cluster_link
         ],
--- a/apps/emqx_cluster_link/test/emqx_cluster_link_license_SUITE.erl
+++ b/apps/emqx_cluster_link/test/emqx_cluster_link_license_SUITE.erl
@@ -1,0 +1,160 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+-module(emqx_cluster_link_license_SUITE).
+
+-include_lib("common_test/include/ct.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+-compile(export_all).
+-compile(nowarn_export_all).
+
+%%
+
+suite() ->
+    [{timetrap, {seconds, 30}}].
+
+all() ->
+    emqx_common_test_helpers:all(?MODULE).
+
+init_per_suite(Config) ->
+    %% Disable the default MQTT listener so the suite can run on a host
+    %% where ports 1883/8083 are already taken.
+    Apps = emqx_cth_suite:start(
+        [
+            {emqx,
+                "listeners.tcp.default.enable = false\n"
+                "listeners.ssl.default.enable = false\n"
+                "listeners.ws.default.enable = false\n"
+                "listeners.wss.default.enable = false\n"},
+            emqx_conf,
+            emqx_cluster_link
+        ],
+        #{work_dir => emqx_cth_suite:work_dir(Config)}
+    ),
+    [{suite_apps, Apps} | Config].
+
+end_per_suite(Config) ->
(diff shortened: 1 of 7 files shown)


apps/emqx_cluster_link/test/emqx_cluster_link_license_SUITE.erl
```diff
@@ -0,0 +1,122 @@
...
+init_per_suite(Config) ->
+    %% Disable the default MQTT listener so that the suite can run on a host
+    %% where ports 1883/8083 are already taken (e.g. a developer machine).
+    Apps = emqx_cth_suite:start(
+        [
+            {emqx,
+                "listeners.tcp.default.enable = false\n"
+                "listeners.ssl.default.enable = false\n"
+                "listeners.ws.default.enable = false\n"
+                "listeners.wss.default.enable = false\n"},
```
thalesmg: nit: this is a bit unusual;  we usually start with defaults.

apps/emqx_cluster_link/test/emqx_cluster_link_license_SUITE.erl
```diff
@@ -0,0 +1,122 @@
...
+    ok = restart_cluster_link_app(singleton),
+    ?assertEqual(ok, emqx_cluster_link_config:delete_link(<<"licgate2">>)).
+
+-doc "Disabled links can still be created under community license.".
+t_allow_disabled_link_in_single_mode(_Config) ->
+    ok = restart_cluster_link_app(singleton),
+    Link = link_raw(<<"licgate3">>, false),
+    ?assertMatch({ok, _}, emqx_cluster_link_config:create_link(Link)),
+    ?assertEqual(ok, emqx_cluster_link_config:delete_link(<<"licgate3">>)).
+
```
thalesmg, line 96: q: what happens if a node starts with a configuration with an enabled link but community license?  maybe add a test case?

thalesmg (dismissed):

apps/emqx_cluster_link/test/emqx_cluster_link_license_SUITE.erl
zmstone: refactored the code to NOT start links if running on community license.
after license is loaded, one can toggle enable/disable to start.

ieQu1 (dismissed):

apps/emqx_cluster_link/test/emqx_cluster_link_license_SUITE.erl
```diff
@@ -0,0 +1,154 @@
...
+    case whereis(emqx_cluster_link_sup) of
+        undefined ->
+            [];
+        Pid ->
+            %% Infrastructure children (`metrics', `bookkeeper',
+            %% `{extrouter, gc}') are always present; per-link routerepl
+            %% supervisors are added under the link's cluster name.
+            Infra = [metrics, bookkeeper, {extrouter, gc}],
+            [C || {Id, _, _, _} = C <- supervisor:which_children(Pid), not lists:member(Id, Infra)]
+    end.
```
keynslug: Nit. Looks a bit too invasive and brittle. Perhaps just verify message forwarding resources (were / weren't) started, by inspecting `emqx_cluster_link_mqtt:get_all_resources_local_v1()`?

keynslug (dismissed):

apps/emqx_cluster_link/test/emqx_cluster_link_license_SUITE.erl
zmstone: Switched the SUITE to inspect `emqx_cluster_link_mqtt:get_all_resources_local_v1/0` instead of poking the supervisor — much less brittle. Thanks!

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15387)

PR #17543, opened 2026-06-10
build: drop sfodbc from base docker image, publish as -sf variant

zmstone:
Release version:
5.10.5

## Summary

The EMQX Enterprise docker image carried ~240 MB of unstripped
`libSnowflake.so` since 5.10.x even for users who never touched the
Snowflake bridge. Root cause: the `docker-sfodbc` CI job set
`EMQX_IMAGE_TAG=<base>-sf` in env, but the `build` script does not
read that variable — it computes tags from
`${DOCKER_ORG}/${PROFILE}:${PKG_VSN}${SUFFIX}`. So the job rebuilt
the image with the driver and pushed it under the SAME
`emqx/emqx-enterprise:<vsn>` tag as the base, overwriting it. No
`5.10.x-sf` tag ever shipped.

This PR splits the two images apart:

- `emqx/emqx-enterprise:<vsn>` goes back to being driver-free
  (matches the OSS image size).
- `emqx/emqx-enterprise:<vsn>-sf` is published as a separately
  tagged image with the Snowflake driver layered on top, for
  customers who still rely on it.

The mechanism: a new `EMQX_IMAGE_TAG_SUFFIX` env in `build` that
gets appended to the computed tag (composes with the existing
`-elixir` suffix). The `docker-sfodbc` CI job sets it to `-sf`,
which is what the env var was always meant to do.

## PR Checklist


keynslug (dismissed):

deploy/docker/Dockerfile.sfodbc
```diff
@@ -1,18 +1,17 @@
...
-## PKG_VSN=5.8.1-alpha.2-sf \
 ## EMQX_DOCKERFILE=deploy/docker/Dockerfile.sfodbc \
-## BUILD_FROM=emqx/emqx-enterprise:5.8.1-alpha.2 \
-## EMQX_IMAGE_TAG=emqx/emqx-enterprise:5.8.1-alpha.2-sf \
+## EMQX_IMAGE_TAG_SUFFIX=-sf \
+## BUILD_FROM=emqx/emqx-enterprise:5.10.5 \
 ## ./build emqx-enterprise docker
 
-ARG BUILD_FROM=emqx/emqx-enterprise:5.8.1-alpha.2
+ARG BUILD_FROM=emqx/emqx-enterprise:latest
```
zmstone: 5.10 is never latest

zmstone: Pinned to 5.10.5 in 81bfa7aeca.

id (approved):

PR #17546, opened 2026-06-10
feat(mqtt): cap MQTT 5.0 client-supplied Session-Expiry-Interval

zmstone:
Release version: 6.3.0

## Summary

Adds a new `mqtt.max_session_expiry_interval` config that caps the value an MQTT 5.0 client may request via its `Session-Expiry-Interval` CONNECT property. Before this change the v5 path in `emqx_channel:expiry_interval/2` took the client value verbatim, so any client could ask for effectively-forever session retention.

- New zone field `mqtt.max_session_expiry_interval`: `duration() | infinity`, default `infinity` (preserves existing behavior). Has no effect on MQTT 3.1.1/3.1 clients — their session expiry stays under `mqtt.session_expiry_interval`.
- `emqx_channel:expiry_interval/2` (v5 clause) clamps the client-requested seconds against the config (HOCON duration is ms; converted to seconds for the comparison).
- New `enrich_clamped_session_expiry_interval/2` reflects the clamped value back to the client in the CONNACK `Session-Expiry-Interval` property when (and only when) the server actually modified the requested value, per MQTT 5.0 §3.2.2.3.2.

Key files: `apps/emqx/src/emqx_schema.erl`, `apps/emqx/src/emqx_channel.erl`, `rel/i18n/emqx_schema.hocon`, `apps/emqx/test/emqx_session_expiry_clamp_SUITE.erl`.

## PR Checklist


hjianbo (dismissed):

qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-16075)

PR #17548, opened 2026-06-11
Sync release-58 changes into release-510

hjianbo:
Release version:
5.10.5

## Summary

This PR merges `ce/release-58` into `release-510`.

Related to https://github.com/emqx/emqx-dev-team-tasks/issues/87.

The SAML part updates `esaml` to `v1.1.5` and keeps the XXE CT from the r58 fix so the SAML decoding boundary stays covered. On r510 we run OTP 27, where `xmerl` already rejects XML `ENTITY` declarations by default. Therefore this dependency sync does not represent a new user-visible r510 XXE fix.

The r58 changelog entry for the esaml XXE item is intentionally removed in the follow-up commit. Keeping it in r510 would incorrectly advertise a security fix/change for r510; the CT remains because it is still valuable regression coverage while syncing r58 changes forward.

The merge also carries the r58 backup download RBAC fix and supporting release-58 sync changes. Conflict resolution keeps r510-specific dependencies and behavior, including `oidcc` `3.2.1`.

Follow-up for r510 backup sync compatibility: the r58 guard rejected all API-key backup downloads, while `emqx_backup_sync` exports non-sensitive backups with an API key and then downloads the generated file with the same key. This PR keeps the sensitive-backup protection but narrows the API-key denial to archives that actually contain `dashboard_users` or `api_keys` table sets. Dashboard administrators can still download any backup; API-key callers may download only after the target archive is inspected and proven not to contain those sensitive table sets, and the request fails closed if the archive cannot be inspected.

Checks run: `git diff --check`; `./scripts/apps-version-check.sh`; `rebar3 fmt -c` on changed Erlang/app files; direct `esaml_binding:decode_response/2` checks for POST and Redirect/deflate XXE payloads, both rejecting with `entities_not_allowed`. The local target data-backup CT and `make static_checks` are currently blocked before test execution by the existing local `mria_replica_status_update` compile issue, unrelated to this sync.

## PR Checklist


zmstone (commented):
follow up tasks:

1. Port the commit 597b6e0561de7fdfd8bbfc3a2f37cd65ab699f41 back to reelase-58 (and sync to release-59, then release-510).
2. Add relup instructions.

zmstone (approved):

PR #17559, opened 2026-06-12
fix(swagger): require ?DESC ref for operation summary derivation

zmstone:

## Summary

OpenAPI operations in EMQX rely on a `?DESC(...)` reference in their
`description` field — the swagger generator resolves that reference's
`.label` to produce the operation's `summary`.  Without a summary,
Redoc falls back to the first ~50 characters of the description as the
sidebar / header title, which silently truncates the sentence (e.g.
older builds rendered `GET /banned` as
`List all currently banned client IDs, usernames an`).

The pre-existing boot-time check in
`emqx_dashboard_swagger:enforce_method_desc_policy/1` only caught two
classes of regression: a tagged operation with no description at all,
and a `?DESC` ref whose i18n entry was missing.  A tagged operation
whose `description` was a raw binary slipped through unnoticed: no
summary got derived, and the redoc-side truncation appeared at runtime.

This PR:

- Tightens `enforce_method_desc_policy/1` so a tagged operation must
  either declare `summary => ?DESC(M, S)` explicitly, or carry a
  `?DESC` `desc`/`description` ref that summary can be auto-derived
  from.  Raw binary descriptions on tagged ops now fail at boot with
  `{missing_i18n_ref, missing_desc_ref_for_summary}`.
- Teaches `trans_summary/3` to honour an explicit `summary => ?DESC(...)`
  ref, so operations with genuinely dynamic descriptions (e.g.
  `emqx_mgmt_api_configs` paths that interpolate the requested config
  root) can still pin a static, i18n-driven title.  Literal binary
  summaries remain forbidden in production.
- Adds a CT case `t_swagger_summary_required` that walks every
  production API module via `emqx_dashboard_swagger:spec/2` and asserts
  every tagged operation has a non-empty binary `summary`.  The check
  bypasses the live `/api-docs/swagger.json` endpoint because the test
  build profile bundles test-SUITE modules into the dashboard app's
  module list, polluting the live spec with intentionally permissive
  synthetic operations.

No source-level offenders were found in `release-60`; the existing
operations are all `?DESC`-clean and the user-reported truncation must
have been observed on an older deployment.  The change is a forward
guard against regressions.

## PR Checklist
- [ ] Change log for changes visible by users has been added to \`changes/ee/fix-17559.en.md\`

### DIFF
--- a/apps/emqx_dashboard/test/emqx_dashboard_SUITE.erl
+++ b/apps/emqx_dashboard/test/emqx_dashboard_SUITE.erl
@@ -310,7 +310,14 @@ production_api_modules() ->
     %% `reboot_apps/0' is the canonical EMQX umbrella business-app list
     %% and lets us load each one explicitly here.
     Apps = emqx_machine_boot:reboot_apps(),
-    lists:foreach(fun ensure_loaded/1, Apps),
+    AlreadyLoaded = [App || {App, _, _} <- application:loaded_applications()],
+    NewlyLoaded = [App || App <- Apps, ensure_loaded(App, AlreadyLoaded)],
+    %% `cth_suite' unloads the apps it started; we only need to
+    %% unload the ones we just loaded ourselves, so other code that
+    %% relies on the original load state isn't surprised.
+    emqx_common_test_helpers:on_exit(
+        fun() -> lists:foreach(fun application:unload/1, NewlyLoaded) end
+    ),
     AllModules = lists:flatten([app_modules(App) || App <- Apps]),
     [
         M
@@ -319,11 +326,22 @@ production_api_modules() ->
         implements_minirest_api(M)
     ].
 
-ensure_loaded(App) ->
-    case application:load(App) of
-        ok -> ok;
-        {error, {already_loaded, _}} -> ok;
-        {error, Other} -> ct:pal("Skip ~p: ~p", [App, Other])
+%% Returns true when this call actually loaded the app (so the caller
+%% knows to undo it later).
+ensure_loaded(App, AlreadyLoaded) ->
+    case lists:member(App, AlreadyLoaded) of
+        true ->
+            false;
+        false ->
+            case application:load(App) of
+                ok ->
+                    true;
+                {error, {already_loaded, _}} ->
+                    false;
+                {error, Other} ->
+                    ct:pal("Skip ~p: ~p", [App, Other]),
+                    false
+            end
     end.
 
 app_modules(App) ->
--- a/apps/emqx_dashboard/test/emqx_dashboard_SUITE.erl
+++ b/apps/emqx_dashboard/test/emqx_dashboard_SUITE.erl
@@ -279,6 +279,114 @@ t_swagger_json(_Config) ->
     ?assertEqual([], BadTags, {non_string_tags_in_openapi_spec, BadTags}),
     ok.
 
+-doc """
+Every tagged operation in every production API module must carry a
+non-empty binary `summary'.  Without it, Redoc falls back to the first
+~50 chars of the description for the sidebar title and operation header,
+which silently truncates a complete sentence (e.g. `GET /banned' once
+surfaced "List all currently banned client IDs, usernames an").
+
+This guard is paired with the boot-time enforcement in
+`emqx_dashboard_swagger:enforce_method_desc_policy/1' so a future
+regression cannot ship even if the boot check is loosened.
+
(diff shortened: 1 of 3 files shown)


apps/emqx_dashboard/test/emqx_dashboard_SUITE.erl
```diff
@@ -279,6 +279,114 @@ t_swagger_json(_Config) ->
...
+intentionally permissive synthetic operations.
+""".
+t_swagger_summary_required(_Config) ->
+    Modules = production_api_modules(),
+    ?assert(length(Modules) > 30, {too_few_api_modules, length(Modules)}),
+    Missing = lists:flatmap(fun module_missing_summaries/1, Modules),
+    ?assertEqual([], Missing, {ops_missing_summary, Missing}).
+
+production_api_modules() ->
+    Apps = [App || {App, _, _} <- application:loaded_applications(), is_emqx_app(App)],
```
thalesmg: this might not actually be testing much: umbrella applications are not typically all loaded in tests.

you'll need to do something like this:  https://github.com/emqx/emqx/blob/d758fc0ee306b2aa78d26e2b14097e70e7227eba/apps/emqx_machine/test/emqx_machine_features_tests.erl#L59-L68

thalesmg: e.g., listing "emqx" apps in http bridge suite yields only:

```erlang
  #{loaded =>
        [emqx_plugins,emqx_utils,emqx_bridge_http,emqx_conf,emqx_http_lib,
         emqx_durable_timer,emqx_ctl,emqx_setopts,emqx_ds_backends,
         emqx_management,emqx_connector_jwt,emqx_resource,emqx_bpapi,
         emqx_dashboard,emqx_connector,emqx_ds_builtin_local,emqx_gen_bridge,
         emqx_bridge,emqx_ds_builtin_raft,emqx_rule_engine,
         emqx_durable_storage,emqx]}
```

apps/emqx_dashboard/test/emqx_dashboard_SUITE.erl
```diff
@@ -279,6 +279,122 @@ t_swagger_json(_Config) ->
...
+
+production_api_modules() ->
+    %% Use `emqx_machine_boot:reboot_apps/0' rather than
+    %% `application:loaded_applications/0': the latter only sees apps
+    %% the SUITE happens to have loaded, so a future init_per_suite
+    %% trimming could silently shrink coverage to a handful of apps.
+    %% `reboot_apps/0' is the canonical EMQX umbrella business-app list
+    %% and lets us load each one explicitly here.
+    Apps = emqx_machine_boot:reboot_apps(),
+    lists:foreach(fun ensure_loaded/1, Apps),
```
thalesmg: should cleanup what's loaded, since some stuff might depend on that.


```suggestion
    lists:foreach(fun ensure_loaded/1, Apps),
    on_exit(fun() -> lists:foreach(fun application:unload/1, Apps) end),
```

thalesmg: ... or rather, unload only the apps that were not loaded in the first place; cth_suite will unload the ones it started.

apps/emqx_dashboard/test/emqx_dashboard_SUITE.erl
```diff
@@ -310,7 +310,14 @@ production_api_modules() ->
     %% `reboot_apps/0' is the canonical EMQX umbrella business-app list
     %% and lets us load each one explicitly here.
     Apps = emqx_machine_boot:reboot_apps(),
-    lists:foreach(fun ensure_loaded/1, Apps),
+    AlreadyLoaded = [App || {App, _, _} <- application:loaded_applications()],
+    NewlyLoaded = [App || App <- Apps, ensure_loaded(App, AlreadyLoaded)],
+    %% `cth_suite' unloads the apps it started; we only need to
+    %% unload the ones we just loaded ourselves, so other code that
+    %% relies on the original load state isn't surprised.
+    emqx_common_test_helpers:on_exit(
```
thalesmg, line 318: ~~need to call the janitor when the test ends, in order to take effect (this suite does not seem to do so yet)~~

actually, it'll do the cleanup, just make some noise

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15477)

PR #17564, opened 2026-06-12
Implement security profiles and fail-closed behavior for authz

savonarola:
Fixes  https://github.com/emqx/emqx-dev-team-tasks/issues/112
Addresses https://github.com/emqx/emqx-dev-team-tasks/issues/118

Release version:
6.2.2

## Summary

This PR adds security-profile-aware authorization backend failure handling. The new `authz_backend_failure` policy preserves legacy behavior by ignoring backend failures, while hardened mode fails closed by returning deny.

The policy is applied to authorization backend crashes, query/request failures, invalid backend responses, malformed rule rows, and connector failure paths across HTTP, LDAP, MongoDB, MySQL, PostgreSQL, and Redis authorizers.

**Important**: 
* fail close behaviour _not_ implemented for `emqx_authz_rules` yet. It will be implemented together with interpolation fix (#17596)
* fail close behaviour _not_ implemented for authn/authz hooks, a separate PR will follow (#17589 )

**Important**: the PR is not fully compatible because we change `nomatch` to `ignore` for legacy profile in case of failures. However, this results only in counter changes. I think `ignore` for failures is more semantically correct. `nomatch` means we have **valid set of rules** against which the principal does not match.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_security_profile.erl
+++ b/apps/emqx/src/emqx_security_profile.erl
@@ -47,6 +47,7 @@ Returns policy depending on the current security profile.
     (mqtt_default_bind) -> loopback | any;
     (dashboard_http_default_bind) -> loopback | any;
     (authn_not_configured) -> allow | deny;
+    (authz_backend_failure) -> ignore | deny;
     (dashboard_unchanged_default_credentials) -> allow | deny.
 policy(mqtt_default_bind) ->
     case profile() of
@@ -61,6 +62,11 @@ policy(dashboard_http_default_bind) ->
 policy(authn_not_configured) ->
     case profile() of
         legacy -> allow;
+        hardened -> deny
+    end;
+policy(authz_backend_failure) ->
+    case profile() of
+        legacy -> ignore;
         hardened -> deny
     end;
 policy(dashboard_unchanged_default_credentials) ->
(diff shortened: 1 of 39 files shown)


apps/emqx/src/emqx_security_profile.erl
```diff
@@ -47,6 +47,7 @@ Returns policy depending on the current security profile.
     (mqtt_default_bind) -> loopback | any;
     (dashboard_http_default_bind) -> loopback | any;
     (authn_not_configured) -> allow | deny;
+    (authz_backend_failure) -> ignore | deny;
```
zmstone, line 50: second thought, for this two, maybe deserves a separate config to enable compatibility even in hardened security mode.

reason: these two changes may require changes from integrated components.

savonarola: Added a switch

thalesmg (approved):

apps/emqx_auth/src/emqx_authz/emqx_authz_schema.erl
```diff
@@ -125,6 +125,12 @@ injected_fields(AuthzSchemaMods) ->
 
 authz_fields() ->
     [
+        {ignore_backend_failures,
+            hoconsc:mk(boolean(), #{
+                default => false,
+                desc => ?DESC(ignore_backend_failures),
+                importance => ?IMPORTANCE_HIDDEN
```
zmstone, line 132: do not hide it.
also there should be one for authentication ?

savonarola: There will be a separate PR for authn

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15734)

PR #17568, opened 2026-06-13
fix(deps): bump brod to 4.5.5

zmstone:
Release version: 6.0.3

## Summary

Bumps the Kafka client library `brod` from 4.5.4 to 4.5.5 in the umbrella `mix.exs` pin.

User-visible upstream fix picked up:

- `fix(consumer-group): handle member_id_required error` — `brod` now remembers the member ID returned by the broker when the join response carries the `member_id_required` error code (e.g. against Kafka 2.2.0, which does not support static member instance IDs), so the retry can succeed instead of looping with a fresh empty ID.

The release also contains a `brod_client <-> producers_sup` deadlock fix, but EMQX does not use brod's producer, so that one is not user-visible from EMQX.

No EMQX code changes — the `brod` API is unchanged.

## PR Checklist

### DIFF
--- a/changes/ee/fix-17568.en.md
+++ b/changes/ee/fix-17568.en.md
@@ -1,6 +1,3 @@
 Upgraded the Kafka client library `brod` to 4.5.5.
 
-This release includes two fixes that affect Kafka producer and consumer integrations:
-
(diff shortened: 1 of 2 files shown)


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15529)

PR #17572, opened 2026-06-14
test: de-flake CT suites (shared-sub, clients, login-lock)

zmstone:
## Summary

De-flake three pre-existing timing-sensitive CT tests (surfaced repeatedly during unrelated CI runs). Each fix targets the actual race, not a blanket retry.

- `emqx_shared_sub_SUITE:t_session_takeover` — the final `collect_msgs` assert rejected a late QoS1 **DUP redelivery**, which the test itself notes is expected; ignore DUP redeliveries.
- `emqx_mgmt_api_clients_SUITE:t_subscribe_shared_topic` — the subscribe API returns before the global subscription tables are updated; retry the `ets:tab2list` assertion until the async registration is visible.
- `emqx_dashboard_login_lock_SUITE:t_cancel_lock_with_cli` — the 1s lock duration (kept short for the sibling release test) could expire before the assert under slow CI; use a long duration for this lock-detection test.

Targeted at release-60 so the fixes sync forward. Test-only; no user impact.

### DIFF
--- a/apps/emqx/test/emqx_cm_SUITE.erl
+++ b/apps/emqx/test/emqx_cm_SUITE.erl
@@ -519,8 +519,14 @@ t_clientid_registration_throttled(_) ->
     ChanInfo = ?ChanInfo,
     #{conninfo := ConnInfo} = ChanInfo,
     ?assertReceive({'DOWN', MRef, process, DeadPid, _}),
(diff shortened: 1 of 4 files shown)


JimMoen (approved):

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15530)

PR #17573, opened 2026-06-14
fix(frame): make MQTT v5 user-property parsing O(N) instead of O(N^2)

zmstone:
Release version: 5.8.11

## Summary

Make MQTT v5 `User-Property` parsing scale linearly with the number of entries.

Previously each parsed user-property was appended to the end of the accumulated list via `lists:append/2`, making the parse of a single packet quadratic in the number of user-properties. A CONNECT/PUBLISH/SUBSCRIBE packet carrying many user-properties therefore consumed super-linear scheduler time on the owning connection process.

The parser now prepends each entry (O(1) per entry) and reverses the `'User-Property'` list once when the property map is finalized, preserving wire order. Added a regression test that parses a CONNECT frame with 50k user-properties and asserts both the linear-time behaviour and that count/order are preserved.

Internal tracker: https://github.com/emqx/emqx-dev-team-tasks/issues/142

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_frame.erl
+++ b/apps/emqx/src/emqx_frame.erl
@@ -555,7 +555,7 @@ parse_properties(Bin, ?MQTT_PROTO_V5, StrictMode) ->
     {Len, Rest} = parse_variable_byte_integer(Bin),
     case Rest of
         <<PropsBin:Len/binary, Rest1/binary>> ->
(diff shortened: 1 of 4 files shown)


keynslug (dismissed):
🚀

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15526)

PR #17575, opened 2026-06-14
fix: race in username quota counter causing record/count inconsistency

JimMoen:
Release version: 6.0.3

## Summary

Fixed a race condition in `emqx_username_quota_state:dec_counter/1` where the per-username session counter could be permanently lost.

**Root cause**: `dec_counter/1` does `mria:dirty_update_counter(-1)` and, when the result reaches ≤0, deletes the counter key via `mria:dirty_delete`. Between the update and the delete, a concurrent `add` (session registration) can increment the same counter. The subsequent `dirty_delete` then destroys the counter value, losing the increment permanently. This causes `RECORD_TAB` (actual sessions) to have more entries than `COUNTER_TAB` (tracked counts), making the quota enforcement inaccurate.

**Fix** (3 changes in `emqx_username_quota_state.erl`):

1. **`dec_counter/1`**: Removed `dirty_delete`. Counter keys now stay in the table at 0 or negative values. The reader (`sum_username_counters`) already filters `Count > 0`, so ≤0 values are treated the same as missing keys.

2. **`del/1`**: Swapped order — delete `RECORD_TAB` record first, then call `dec_counter`. If a crash occurs between them, the record is already gone and the counter is slightly higher (safe direction: rejects a few more connections, not fewer).

3. **`delete_client_records/2`**: Same order swap.

**Tests added** (`emqx_username_quota_SUITE.erl`):
- `t_counter_consistency_concurrent` — 40 concurrent add/del with same username, verify counter matches records
- `t_counter_not_deleted_when_zero` — verify counter entry persists at 0 after all sessions removed
- `t_double_unregister_counter_safety` — verify repeated unregister does not break counters

## PR Checklist
- [x] Change log added to `changes/ee/fix-17575.en.md`


JimMoen:
fix https://github.com/emqx/emqx-dev-team-tasks/issues/119

plugins/emqx_username_quota/src/emqx_username_quota_state.erl
```diff
@@ -349,14 +349,8 @@ do_clear_for_node(Node) ->
...
 dec_counter(Username) ->
-    Key = ?COUNTER_KEY(Username, node()),
-    case mria:dirty_update_counter(?COUNTER_TAB, Key, -1) of
-        NewCount when is_integer(NewCount), NewCount =< 0 ->
-            _ = mria:dirty_delete(?COUNTER_TAB, Key),
-            ok;
-        _ ->
-            ok
-    end.
+    _ = mria:dirty_update_counter(?COUNTER_TAB, ?COUNTER_KEY(Username, node()), -1),
```
hjianbo, line 352: Leaving the counter at any negative value can still make quota enforcement undercount later sessions. `sum_username_counters/1` ignores `Count =< 0`, but a later `add/3` only increments the stored value; for example, a stale decrement to `-1` followed by one real registration leaves the raw counter at `0` while `RECORD_TAB` has one session, so `session_count/1` still returns 0. Please prevent negative counters, or only decrement when we know a record was actually removed.

plugins/emqx_username_quota/src/emqx_username_quota_state.erl
```diff
@@ -349,14 +349,8 @@ do_clear_for_node(Node) ->
...
+    ok.
```
hjianbo, line 353: Even if negative counters are avoided, keeping every zero counter forever makes `COUNTER_TAB` grow by every username/node that ever had a session. Snapshot builds scan this table before filtering by positive counts, so high-churn usernames can increase memory use and rebuild latency over time. Please consider a race-safe zero cleanup or compaction path instead of retaining all zero counters permanently.

plugins/emqx_username_quota/changelog.md
```diff
@@ -44,3 +44,12 @@
 - Moved quota exceeded warning logs into the stats worker and throttled them per username.
   A warning is emitted for the first rejection, when the previous warning is older than 1 minute,
   or on every 100th rejection for that username.
+
+## 1.2.1
```
hjianbo, line 48: This changelog introduces plugin version `1.2.1`, but `plugins/emqx_username_quota/VERSION` still contains `1.2.0`, and `mix.exs` reads the package version from that file. Please bump `VERSION` together with this source change; otherwise the packaged plugin metadata and changelog disagree, and the app version check should report this plugin as needing a release version bump.

plugins/emqx_username_quota/src/emqx_username_quota_state.erl
JimMoen: The double-decrement scenario occurs because both `del_client` (hook path) and `del` (pool DOWN) can clean up the same session. Fixed in 2b73187 by deleting MONITOR_TAB before RECORD_TAB and dec_counter in `delete_client_records`. This makes `del/1` find no MONITOR entry and skip, preventing the second decrement. If crash between MONITOR delete and RECORD delete: MONITOR gone → del is no-op → counter not decremented (safe: slightly higher).

plugins/emqx_username_quota/src/emqx_username_quota_state.erl
JimMoen: Growth is bounded: at most one entry per username per node. Snapshot `scan_counters` already has `YieldInterval=1000` to avoid scheduler starvation, and `read_counter/1` + `sum_username_counters/1` filter `Count > 0` so zero entries are only iterated, not counted. For a node with 100k unique usernames, COUNTER_TAB is ~100k entries (~5 MB). If performance becomes a concern, a periodic GC can be added later, but this is not a regression — the old code also accumulated entries for sessions in flight.

plugins/emqx_username_quota/changelog.md
JimMoen: Done — VERSION bumped to 1.2.1 in 2b73187.

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15531)

PR #17581, opened 2026-06-15
fix(jt808): bind authcode to registered phone

hjianbo:
Release version:
6.0.3

## Summary

Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/158.

The JT/T 808 gateway now uses the phone number accepted during authentication as the connection identity. Registration remains optional; when a registration-issued authcode is present, the authentication frame must use the matching phone number before that authcode is accepted.

Temporary `registered_phone` and `authenticated_phone` values are kept in `clientinfo` while processing registration and authentication frames. After successful authentication, they are normalized back to the existing `phone` and `clientid` fields. A subsequent registration frame also clears any stale authentication phone before building the REGISTER_ACK response.

After a client is connected, uplink frames whose header phone differs from the authenticated phone are treated as invalid frames, following the existing `ignore_unsupported_frames` behavior for whether to ignore or disconnect.

Added CT coverage for default auth, anonymous auth, post-auth uplink frame phone mismatch, and REGISTER_ACK phone handling after a failed authentication attempt. Checked with `mix compile`, `erlfmt`, `apps-version-check`, `git diff --check`, and the focused JT808 CT cases.

## PR Checklist


zmstone:
back-port to release-58

zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15533)

PR #17585, opened 2026-06-15
refactor(cm): address review nits on registry keeper

zmstone:
Follow-up to #17522. Code-only cleanup, no behavior change.

## Summary

- `new_sweep_state/1` uses `:=` instead of `=>` so a typo on `timer_ref` would be caught.
- Trim the stale comment above `handle_info(start, _)`; the trace-suppression rationale now lives next to `maybe_emit_finished/1`.
- Spell out the `garbage_collect/0` comment in `do_run_chunk/2` (sub-binaries pinning mnesia row data between chunks).
- Collapse `run_full_sweep/2` into `run_chunk_loop/3` by accepting `infinity` as the budget; the forced-sweep path calls it with `infinity`.
- `classify_remote_pid_on_core/2` only calls `normalize_node_caches/1` in the cache-miss branch where one of the sets actually changes. Cache hits no longer pay `sets:subtract/2` on the hot path.

All 11 cases in `emqx_cm_registry_gc_SUITE` and all 3 in `emqx_cm_registry_keeper_SUITE` pass.

### DIFF
--- a/apps/emqx/src/emqx_cm_registry_keeper.erl
+++ b/apps/emqx/src/emqx_cm_registry_keeper.erl
@@ -121,7 +121,7 @@ purge_loop(ClientId) ->
 handle_call(force_sweep_stale_pids, _From, State0) ->
     State1 = reset_sweep_state(State0, _Forced = true),
     ?tp(info, cm_registry_gc_started, #{forced => true}),
-    State2 = run_full_sweep(undefined, State1),
+    {'$end_of_table', State2} = run_chunk_loop(undefined, infinity, State1),
     Summary = sweep_summary(State2),
     ?tp(info, cm_registry_gc_finished, Summary#{forced => true}),
     {reply, Summary, reset_sweep_counters(State2)};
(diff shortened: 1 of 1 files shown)


apps/emqx/src/emqx_cm_registry_keeper.erl
```diff
@@ -276,12 +272,18 @@ do_run_chunk(Cursor, State0) ->
             TimerRef = send_delay_start(),
             {noreply, State1#{next_clientid := undefined, timer_ref := TimerRef}};
         Cid ->
-            %% ensure the next clientid is not in the cache
+            %% Force a full GC so any sub-binaries pinning mnesia row
+            %% data from this chunk (next_clientid is the most obvious;
+            %% lingering refc-binary references inside the process heap
+            %% are the broader concern) get released before we sleep.
```
thalesmg: since this is a process which is mostly inactive (10 minutes between each run), wouldn't it be somewhat simpler to simply `hibernate` after finishing the scan, since that triggers GC when hibernating and waking up?

thalesmg (dismissed):

apps/emqx/src/emqx_cm_registry_keeper.erl
zmstone: this gc is between the chunks.

thalesmg: yes, I know.  my suggestion/question still stands.  :sweat_smile:

zmstone: yes, make sense to hibernate.
the gc between chunks is still necessary though

thalesmg (approved):

PR #17586, opened 2026-06-15
fix(cm): periodically purge stale rows from global session registry

zmstone:

## Summary

Port of #17522 from release-510.

Adds a slow, throttled background sweep to `emqx_cm_registry_keeper` that periodically purges stale rows from the global session registry. PR #17424 closed the common case (cleanup gated on `mria:is_peer_alive/1`), but that gate is one-shot: a transient peer-core RPC error makes `is_peer_alive` return `{ok, true}` and the cleanup is silently dropped with no retry, and a core that joins after a `{node, down, _}` event never receives a replay. In both cases the registry row could leak forever for a clientid that never reconnects.

Per row, the sweep:
- local pid + not `is_process_alive/1` → unregister;
- remote pid (core role only): cache `mria:is_peer_alive/1` per sweep; `{ok, false}` → unregister. Any other answer (including `{aborted, _}`) is treated as keep, preserving the cluster-consensus safety bias;
- replicants act only on local-dead pids; they skip hist tombstones and remote pids (cores remain the single deleter for those).

The sweep is bounded to ≤ 500 rows/s/node and runs no more often than once per 10 minutes, so it does not measurably affect broker throughput even on registries holding millions of sessions. The existing history-row cleanup continues to run inside the same chunk loop.

Most relevant module: `apps/emqx/src/emqx_cm_registry_keeper.erl`. A `pr_17522_kickoff_registry_keeper/1` post-upgrade hook (`apps/emqx/src/emqx_post_upgrade.erl`) lets a future relup wake a replicant that booted on the pre-fix beam.

## PR Checklist


ieQu1 (dismissed):

zmstone:
force-pushed due to changelog pr number was wrong.

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15561)

PR #17589, opened 2026-06-15
feat(auth): implement fail closed for access control hooks

savonarola:
https://github.com/emqx/emqx-dev-team-tasks/issues/112
Addresses https://github.com/emqx/emqx-dev-team-tasks/issues/118

Release version: 6.2.2

A complement to https://github.com/emqx/emqx/pull/17564

## Summary

This PR adds hardened-profile fail-close behavior for access-control hook callback failures.

In hardened mode, a authentication and authorization hook failures are treated as denial.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_access_control.erl
+++ b/apps/emqx/src/emqx_access_control.erl
@@ -319,9 +319,13 @@ format_retain_flag(false) ->
 
 run_hooks(Name, Args, Acc) when Name == 'client.authenticate'; Name == 'client.authorize' ->
     ok = emqx_metrics:inc_global(Name),
-    {Time, Value} = timer:tc(
-        fun() -> emqx_hooks:run_fold(Name, Args, Acc) end
-    ),
+    {Time, Value} =
+        case emqx_security_profile:policy(access_control_hook_failure) of
+            ignore ->
+                timer:tc(fun() -> {ok, emqx_hooks:run_fold(Name, Args, Acc)} end);
+            interrupt ->
+                timer:tc(fun() -> emqx_hooks:run_fold_strict(Name, Args, Acc) end)
+        end,
     try
         emqx_metrics_worker:observe_hist(
             ?ACCESS_CONTROL_METRICS_WORKER,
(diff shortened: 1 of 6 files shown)


apps/emqx/src/emqx_access_control.erl
```diff
@@ -333,7 +337,15 @@ run_hooks(Name, Args, Acc) when Name == 'client.authenticate'; Name == 'client.a
         _:_ ->
             ok
     end,
-    Value.
+    case Value of
+        {ok, Result} -> Result;
+        {error, _Reason} -> fail_close_result(Name)
```
thalesmg, line 342: q: shouldn't this depend on the security profile?

e.g., if any hooks return `{stop, {error, _}}`, even with the legacy profile, the check here will start to return `deny`.

thalesmg (dismissed):

apps/emqx/src/emqx_access_control.erl
savonarola: With legacy profile, we never hit this clause: https://github.com/emqx/emqx/pull/17589/changes/BASE..6288a88e1071a28a1bdb93975d4741f40a2d2165#diff-8732768172f2c653a41b55cc5a907db46f55fc653b3243a1ffedf6c04d2f8caeR325

zmstone (dismissed):

apps/emqx/src/emqx_access_control.erl
thalesmg: 🤔 
the old hooks can return {stop, {error, _}}, as mentioned.

or is there something else preventing that?

savonarola: If they return `{stop, {error, Err}}`, in both cases (run/run_strict) we will have `{ok, {error, Err}}` in the `Value`

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15738)

PR #17590, opened 2026-06-15
test: de-flake flaky CT cases on release-510

zmstone:
Release version: 5.10.4

## Summary

De-flake pre-existing flaky CT cases on `release-510`, surfaced by the CI runs
for #17585 (https://github.com/emqx/emqx/actions/runs/27550491595) and by this
PR's own CI. None are regressions: #17585 is a code-only cleanup of the registry
keeper, and none of these tests had the stabilizing fixes on `release-510`.

Almost all of them are the same class of bug: a test asserts on broker state
immediately after an action whose effect propagates asynchronously (channel
registration/deregistration, subscription install/remove, cross-node client
listing, durable-session replay, or snabbkaffe trace collection). The fixes wait
for the actual condition (`?retry`, wait-for-no-subscribers, accumulate-until-
complete, filter-by-unique-id) instead of relying on a fixed sleep or window.

Cases fixed:

- **`emqx_cm_SUITE`**: `t_clientid_registration_throttled` and
  `t_open_session_throttled_on_inflight_local_cleanup` - suspend `emqx_cm`
  around register + open_session so the DOWN-driven cleanup cannot race the
  throttle (ports the `release-60` fix 96577bdc34).
- **`emqx_crl_cache_SUITE`**: `t_cache_overflow` - the client-process span
  events and the `emqx_crl_cache` gen_server events interleave; assert each
  stream's order separately.
- **`emqx_dashboard_monitor_SUITE`**: `t_pmap_nodes` - pause the live sampler,
  use deterministic data points, and allow the first bucket to be 1..5 intervals
  from the second (ports the v6 fix).
- **`emqx_mqtt_SUITE`**: `t_message_expiry_interval` - sleep 2s (not 1100ms) past
  a 1s expiry so the wall-clock expiry check is not racy (ports the v6 fix).
- **`emqx_takeover_SUITE`**: `t_kick_session` (connect + subscribe synchronously
  before the kick), `t_takeover_willmsg` (accumulate publishes until all arrive;
  accept a plain socket close as a valid v5 takeover exit), and the shared
  `assert_client_exit` (wait up to 5s for the client EXIT).
- **`emqx_retainer_SUITE`**: `t_store_and_clean` - wait for the unsubscribe to
  propagate and the retained store to clear before re-subscribing.
- **`emqx_mgmt_api_clients_SUITE`**: `t_subscribe_shared_topic`,
  `t_kickout_clients`, `t_persistent_sessions5` - retry the listing/subscription
  assertions until async deregistration and the cluster-wide count settle.
- **`emqx_gateway_api_SUITE`**: `t_clients` - retry the post-kick delete until
  the client is deregistered (ports the v6 fix).
- **`emqx_client_SUITE`**: `t_sock_closed_reason_normal`,
  `t_sock_closed_force_closed_by_client` - per-version unique client ids (avoid
  the reconnect throttle) and assert on this client's trace event rather than
  requiring it to be the only one captured.

Not addressed here (environmental, not test logic): the docker-runner external-
service bridge suites (`emqx_bridge_kinesis`, `emqx_bridge_gcp_pubsub`,
`emqx_bridge_confluent`, `emqx_bridge_kafka`, `emqx_bridge_azure_blob_storage`,
`emqx_bridge_sqlserver`) flake on container/network timing - a different subset
each run - and are unrelated to these changes.

Validation: every fixed case was looped locally (20-30x where fast, fewer for
slow cluster/persistence cases), and a full CI run on this branch passed with no
failures.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_client_SUITE.erl
+++ b/apps/emqx/test/emqx_client_SUITE.erl
@@ -456,7 +456,18 @@ t_sock_closed_reason_normal(_) ->
                     )
                 end,
                 fun(Trace0) ->
-                    ?assertMatch([#{clientid := ClientId}], ?of_kind(sock_closed_normal, Trace0)),
+                    %% Filter by this iteration's unique client id: the captured
+                    %% trace can contain sock_closed events from other clients,
+                    %% so assert on this client's event rather than requiring it
+                    %% to be the only one in the trace.
+                    ?assertMatch(
+                        [#{clientid := ClientId} | _],
+                        [
+                            E
+                         || #{clientid := CId} = E <- ?of_kind(sock_closed_normal, Trace0),
+                            CId =:= ClientId
+                        ]
+                    ),
                     ok
                 end
             )
@@ -482,8 +493,16 @@ t_sock_closed_force_closed_by_client(_) ->
                     )
                 end,
                 fun(Trace0) ->
+                    %% Filter by this iteration's unique client id, see
+                    %% t_sock_closed_reason_normal.
                     ?assertMatch(
-                        [#{clientid := ClientId}], ?of_kind(sock_closed_with_other_reason, Trace0)
+                        [#{clientid := ClientId} | _],
+                        [
+                            E
+                         || #{clientid := CId} = E <-
+                                ?of_kind(sock_closed_with_other_reason, Trace0),
+                            CId =:= ClientId
+                        ]
                     ),
                     ok
                 end
--- a/apps/emqx/test/emqx_client_SUITE.erl
+++ b/apps/emqx/test/emqx_client_SUITE.erl
@@ -438,8 +438,8 @@ t_client_attr_from_user_property(_Config) ->
 
 t_sock_closed_reason_normal(_) ->
     ProtoVers = [v3, v4, v5],
-    [
-        begin
+    lists:foreach(
+        fun(Ver) ->
             %% Use a per-version unique client id: reusing one client id across
             %% the iterations can reconnect before the previous connection's
             %% async deregistration completes, tripping the open_session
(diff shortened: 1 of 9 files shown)


apps/emqx/test/emqx_client_SUITE.erl
```diff
@@ -438,48 +438,76 @@ t_client_attr_from_user_property(_Config) ->
...
-                ?assertMatch([#{clientid := ClientId}], ?of_kind(sock_closed_normal, Trace0)),
-                ok
-            end
-        )
+        begin
+            %% Use a per-version unique client id: reusing one client id across
+            %% the iterations can reconnect before the previous connection's
+            %% async deregistration completes, tripping the open_session
+            %% throttle (rejected as server_unavailable / server_busy).
+            ClientId = iolist_to_binary([atom_to_binary(?FUNCTION_NAME), "-", atom_to_binary(Ver)]),
```
zmstone, line 447: use lists:foreach instead of LC

apps/emqx/test/emqx_client_SUITE.erl
```diff
@@ -438,48 +438,76 @@ t_client_attr_from_user_property(_Config) ->
...
-            fun(Trace0) ->
-                ?assertMatch(
-                    [#{clientid := ClientId}], ?of_kind(sock_closed_with_other_reason, Trace0)
-                ),
-                ok
-            end
-        )
+        begin
+            %% Per-version unique client id, see t_sock_closed_reason_normal.
+            ClientId = iolist_to_binary([atom_to_binary(?FUNCTION_NAME), "-", atom_to_binary(Ver)]),
```
zmstone, line 484: use lists:foreach instead of LC

hjianbo (approved):

qzhuyan (approved):

PR #17595, opened 2026-06-15
fix(s3, s3tables): limit amount of data returned during health checks

### LINKED ISSUE #17411: perf(s3, s3tables): connector health check lists all buckets on every tick
## Summary

The S3 and S3 Tables connectors share `emqx_s3_client:get_status/1` for the connector-level health check, which calls `erlcloud_s3:list_buckets(AWSConfig)` — returning every bucket the AWS account owns on every health-check tick.

This is bounded by AWS to 10000 buckets per call, but a more common cost shape is that it's an account-wide call when a single targeted probe would do.

## Where

* `apps/emqx_bridge_s3/src/emqx_bridge_s3_connector.erl:138` → `emqx_s3_client:get_status/1`
* `apps/emqx_bridge_s3tables/src/emqx_bridge_s3tables_impl.erl:130-133` → same `emqx_s3_client:get_status/1`
* `apps/emqx_s3/src/emqx_s3_client.erl:107-120` — the shared implementation that calls `erlcloud_s3:list_buckets`.

Note: the per-bucket `check_bucket_accessible/2` already uses `[{max_keys, 1}]`. Only the connector-level account-wide call is unbounded.

## Impact

* AWS account with hundreds of buckets returns a list of all of them every tick. Bandwidth and parse cost scale with bucket count.
* List-buckets is also an account-wide operation that requires the credentials to have `s3:ListAllMyBuckets`. Operators following least-privilege would prefer a probe that only needs access to the bucket the bridge actually writes to.

## Suggested fix

Replace the `list_buckets` probe with one of:

1. **`erlcloud_aws:get_caller_identity/1`** (STS) — free of charge, returns only the calling identity. Cleanest "is the credential alive" check. Reference: https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html.
2. **`HEAD /` on the regional S3 endpoint** — bounded TCP + TLS probe, no list at all.

Both eliminate the account-wide list and align with least-privilege configurations.

## Related

Part of the health-check audit following d94cbde4. Filed as a YELLOW (bounded but more expensive than necessary).

thalesmg:
Fixes https://github.com/emqx/emqx/issues/17411

Release version:
6.0.3, 6.1.2, 6.2.1

## PR Checklist

### DIFF
--- a/apps/emqx_s3/src/emqx_s3_client.erl
+++ b/apps/emqx_s3/src/emqx_s3_client.erl
@@ -109,7 +109,8 @@ get_status(S3ClientConfig) ->
         {error, Reason} ->
             {?status_disconnected, emqx_s3_utils:map_error_details(Reason)};
         AWSConfig ->
-            try erlcloud_s3:list_buckets(AWSConfig) of
+            Opts = #{max_buckets => 1},
+            try erlcloud_s3:list_buckets(Opts, AWSConfig) of
                 Props when is_list(Props) ->
                     ?status_connected
             catch
--- a/changes/ee/fix-17595.en.md
+++ b/changes/ee/fix-17595.en.md
@@ -1,0 +1,1 @@
+Limited the amount of data returned during Connector health checks of S3 and S3Tables integrations.  This only has observable effects if the list of existing buckets was really large, in which case the health check will take far less time to execute.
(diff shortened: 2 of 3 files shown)


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15537)

Issue #17505, opened 2026-06-07
MongoDB 8.x: buildInfo called before auth completes in mc_worker driver


Saravana1402:
EMQX version: 5.8.7
MongoDB version: 8.3.2
Error: Command buildInfo requires authentication

need a patch release on 5.8.x.

qzhuyan:
hi could you provide more context?

zmstone:
@Saravana1402 thank you for reporting it.

If I get it correctly, the issue is because MongoDB v8 requires auth for buildInfo query.
We'll think about how to fix it.

zmstone:
fixed. will be included in releases:

- 5.8.11
- 5.10.5
- 6.0.4
- 6.1.3
- 6.2.2
- 6.3.0

PR #17597, opened 2026-06-15
fix(mongodb): bump driver to v3.1.2 for MongoDB 8.0 compatibility

zmstone:
Release version: 5.8.11

## Summary

Fixes a connection failure to MongoDB 8.0+ when authentication is required (#17505).

The `mongodb-erlang` driver's `mc_worker_logic:get_version/3` issued a `buildInfo` command **before** SCRAM authentication, to choose between SCRAM-SHA-1 (server > 2.7) and the legacy MONGODB-CR mechanism. MongoDB 8.0 restricted `buildInfo` to authenticated callers, so the pre-auth probe fails and the connection process crashes. MONGODB-CR was removed in MongoDB 4.0, so the probe only ever selected the SCRAM-SHA-1 path on any supported server.

The driver fix (emqx/mongodb-erlang#58, tag `v3.1.2`) skips the probe and always uses SCRAM-SHA-1. This PR bumps the EMQX dep pin to `v3.1.2`.

### Notes for reviewers / release management

- **Unifies a pre-existing dep-pin discrepancy.** Before this change `apps/emqx_mongodb/rebar.config` pinned `v3.1.0` while `apps/emqx_mongodb/mix.exs` pinned `v3.0.25`, i.e. the rebar3 (canonical) and mix (elixir variant) build paths used different driver commits. Both are now bumped to `v3.1.2`. The only `mc_worker_logic.erl` change between those tags is the buildInfo fix; the `v3.0.25 -> v3.1.x` delta (drop of the `pbkdf2` dep in favour of `crypto:pbkdf2_hmac`, poolboy pinned to the emqx fork tag `1.5.2`) is safe on the OTP 26 builder used by release-58.
- **Hot-upgrade.** The fix is a single stateless dep beam (`mc_worker_logic.beam`); the exported arity and return type are unchanged and there are no call-site changes, so running workers (already past auth) are unaffected and new connections pick up the fix. The next 5.8.x relup catalog should include a `{load_module, mc_worker_logic}` instruction so existing nodes get the fixed module on hot-patch. There is no pending (non-example) `.relup` file in this branch to append to, and the next patch version is not yet fixed — flagging for release management to add the hop when the relup is cut.

## PR Checklist

### DIFF
--- a/apps/emqx_mongodb/mix.exs
+++ b/apps/emqx_mongodb/mix.exs
@@ -25,7 +25,7 @@ defmodule EMQXMongodb.MixProject do
     [
       {:emqx_connector, in_umbrella: true, runtime: false},
       {:emqx_resource, in_umbrella: true},
-      {:mongodb, github: "emqx/mongodb-erlang", tag: "v3.0.25"}
+      {:mongodb, github: "emqx/mongodb-erlang", tag: "v3.1.2"}
     ]
   end
 end
--- a/apps/emqx_mongodb/rebar.config
+++ b/apps/emqx_mongodb/rebar.config
@@ -4,5 +4,5 @@
 {deps, [
     {emqx_connector, {path, "../../apps/emqx_connector"}},
     {emqx_resource, {path, "../../apps/emqx_resource"}},
-    {mongodb, {git, "https://github.com/emqx/mongodb-erlang", {tag, "v3.1.0"}}}
+    {mongodb, {git, "https://github.com/emqx/mongodb-erlang", {tag, "v3.1.2"}}}
 ]}.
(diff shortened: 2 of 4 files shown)


hjianbo (dismissed):

apps/emqx_mongodb/mix.exs
```diff
@@ -25,7 +25,7 @@ defmodule EMQXMongodb.MixProject do
     [
       {:emqx_connector, in_umbrella: true, runtime: false},
       {:emqx_resource, in_umbrella: true},
-      {:mongodb, github: "emqx/mongodb-erlang", tag: "v3.0.25"}
```
thalesmg, line 28: this discrepancy is really weird: it should have been picked up by the ci check.  🤔 
I wonder what's missing so that this slipped through.  🙈

thalesmg (approved):

apps/emqx_mongodb/mix.exs
thalesmg: curiously enough, the script does pick up the discrepancy locally :thinking: 

```
  * mongodb
    * Rebar3 ref: ad3e13a089c2e7a43f23faa4e6ad7feeabef8326
    * Mix ref: 3142f464020918466bc23d5634b37640ee14b075
```

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15563)

PR #17598, opened 2026-06-15
fix(bridge_mongodb): connect to MongoDB 8.0+ requiring auth

### LINKED ISSUE #17505: MongoDB 8.x: buildInfo called before auth completes in mc_worker driver

zmstone:

## Summary

Fixes #17505 — connections to MongoDB 8.0+ fail when authentication is required.

The mongodb-erlang driver runs the `buildInfo` command *before* SCRAM
authentication, in `mc_worker_logic:get_version/3`, to choose between
SCRAM-SHA-1 and the legacy MONGODB-CR mechanism. MongoDB 8.0 restricted
`buildInfo` to authenticated callers, so this pre-auth probe now fails
and the connection can never be established.

### Fix

Bumps the `mongodb` dependency in `apps/emqx_mongodb/mix.exs` from
`v3.1.1` to `v3.1.2`. The new tag patches `get_version/3` to return the
constant `3.0` instead of probing the server. That probe only ever
selected between SCRAM-SHA-1 (version > 2.7) and MONGODB-CR; MongoDB 4.0
removed MONGODB-CR entirely, so the legacy branch was dead code.
Returning `3.0` deterministically selects the SCRAM-SHA-1 path, which all
supported MongoDB versions accept.

Driver PR: https://github.com/emqx/mongodb-erlang/pull/57

### CI coverage (MongoDB 8)

To actually exercise the fix and prevent regression, the single-node TCP
MongoDB CI instance is bumped to 8.0 (new `MONGO_TAG_SINGLE_TCP`; the
replica-set / sharded / single-TLS instances stay on 5 since their init
scripts still use the legacy `mongo` shell). That instance's init script
is ported from `mongo` (removed in MongoDB 6.0) to `mongosh`, validated
end-to-end against a real `mongo:8` container.

MongoDB 5.1 removed the legacy OP_QUERY opcodes, so the suites that force
`use_legacy_protocol = true` (`emqx_authz_mongodb_SUITE`,
`emqx_authn_mongodb_SUITE`, `emqx_bridge_mongodb_SUITE`) can only run
against MongoDB <= 5.0. They are gated behind a runtime wire-version
probe (`isMaster.maxWireVersion <= 13`) so they skip on the 8.0 instance
and still run wherever a legacy-capable server is available.

### Hot-upgrade note (for release management)

The fix was kept minimal on purpose so it can ship as a hot patch:
- `get_version/3` keeps its exported arity; the call site in
  `mc_worker.erl` is unchanged.
- `mc_worker_logic` is a stateless helper — a single hot-loadable
  `mc_worker_logic.beam`, no state migration. Workers already past auth
  are unaffected; new connections pick up the fix.

**The next `.relup` hop catalog entry (`plugins/emqx_relup/priv/relup/`)
should add `{load_module, mc_worker_logic}` to its `code_changes`** so an
in-place upgrade reloads the patched beam. No `.relup` file exists yet
for the upcoming hop, so this is flagged here rather than authored in
this PR (the `{from, target}` versions aren't decided here).

## PR Checklist

### DIFF
--- a/.ci/docker-compose-file/.env
+++ b/.ci/docker-compose-file/.env
@@ -1,6 +1,10 @@
 MYSQL_TAG=8
 REDIS_TAG=7.0
 MONGO_TAG=5
+# Single-node TCP MongoDB is pinned to 8 to cover authenticated connections
+# (issue #17505). Other topologies stay on 5 because their init scripts still
+# use the legacy `mongo` shell, which MongoDB removed in 6.0.
+MONGO_TAG_SINGLE_TCP=8
 PGSQL_TAG=13
 LDAP_TAG=2.4.50
 INFLUXDB_TAG=2.5.0
--- a/.ci/docker-compose-file/docker-compose-mongo-single-tcp.yaml
+++ b/.ci/docker-compose-file/docker-compose-mongo-single-tcp.yaml
@@ -1,7 +1,7 @@
 services:
   mongo_server:
     container_name: mongo
-    image: public.ecr.aws/docker/library/mongo:${MONGO_TAG}
+    image: public.ecr.aws/docker/library/mongo:${MONGO_TAG_SINGLE_TCP}
     restart: always
     networks:
       - emqx_bridge
@@ -15,7 +15,7 @@ services:
       --bind_ip_all
 
   mongo_single_setup:
-    image: public.ecr.aws/docker/library/mongo:${MONGO_TAG}
+    image: public.ecr.aws/docker/library/mongo:${MONGO_TAG_SINGLE_TCP}
     container_name: mongo_single_setup
     networks:
       - emqx_bridge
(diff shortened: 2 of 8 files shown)


qzhuyan (dismissed):

.ci/docker-compose-file/.env
```diff
@@ -1,6 +1,10 @@
 MYSQL_TAG=8
 REDIS_TAG=7.0
 MONGO_TAG=5
+# Single-node TCP MongoDB is pinned to 8 to cover authenticated connections
+# (issue #17505). Other topologies stay on 5 because their init scripts still
+# use the legacy `mongo` shell, which MongoDB removed in 6.0.
+MONGO_TAG_SINGLE_TCP=8
```
thalesmg, line 7: nit: IMO, it would be better to inline the versions in the docker files themselves.  otherwise, it's always confusing to remember where this is set.

thalesmg (approved):

.ci/docker-compose-file/.env
zmstone: is there an example ?

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15580)

.ci/docker-compose-file/.env
thalesmg: simply don't use this file and reference the variable in docker file, just write the version directly

PR #17604, opened 2026-06-16
fix(gbt32960): route downstream command responses to upstream/response

JimMoen:
## Summary

Fixes: [EMQX-15313](https://emqx.atlassian.net/browse/EMQX-15313)

Vehicle responses to platform downstream commands (Parameter Query 0x80, Parameter Setting 0x81, Terminal Control 0x82) were incorrectly routed to `upstream/transparent` when the vehicle response frame used `ack=0xFE`.

**Root cause:** `transform/2` in `emqx_gbt32960_channel.erl` has two clauses:
1. Matches frames with `ack=0xFE`, routes by CMD. CMD 0x80-0xBF fell through to `_ -> upstream/transparent` catch-all
2. Matches response ack codes (0x01-0x07) → `upstream/response`

Since `transform/2` is only called from `handle_in` (vehicle→platform), CMD 0x80-0xBF in this context is always a response to a platform command and should route to `upstream/response`.

**Fix:** Added explicit case branches for these three CMDs in the first clause.

## Changes

- `src/emqx_gbt32960_channel.erl` — added 3 case branches
- `test/emqx_gbt32960_SUITE.erl` — added 3 test cases (t_case26-28) covering ack=0xFE responses

## Notes

- Target release: **6.0.4** — please do not merge yet.

[EMQX-15313]: https://emqx.atlassian.net/browse/EMQX-15313?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_gateway_gbt32960/src/emqx_gbt32960_channel.erl
+++ b/apps/emqx_gateway_gbt32960/src/emqx_gbt32960_channel.erl
@@ -760,7 +760,9 @@ transform(Frame = ?CMD(Cmd), Mountpoint) ->
             ?CMD_PLATFORM_LOGIN -> <<"upstream/plogin">>;
             ?CMD_PLATFORM_LOGOUT -> <<"upstream/plogout">>;
             ?CMD_ACTIVATION -> <<"upstream/activation">>;
-            %CMD_HEARTBEAT, CMD_SCHOOL_TIME ...
+            ?CMD_PARAM_QUERY -> <<"upstream/response">>;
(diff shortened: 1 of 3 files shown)


hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15565)

PR #17605, opened 2026-06-16
fix(oracle): handle prepare checks and large text binds

hjianbo:
Fix version:
6.0.4, 6.1.3, 6.2.2

Fix: [EEC-1322](https://emqx.atlassian.net/browse/EEC-1322)

Driver PR: [emqx/jamdb_oracle#9](https://github.com/emqx/jamdb_oracle/pull/9), released as `emqx/jamdb_oracle@0.4.9.7`.

## Summary

This PR fixes two Oracle action issues found while reproducing EEC-1322.

First, Oracle action prepare checks used to validate SQL templates by executing the rendered user SQL inside an anonymous PL/SQL block followed by rollback. In the reproduced prepare-check case, creating an Oracle action with a normal `INSERT INTO mqtt_test ...` template fired a `BEFORE INSERT` trigger and wrote audit rows before any MQTT message was published. The prepare check now uses `DBMS_SQL.PARSE` so it validates SQL syntax and object references without running the action SQL. Top-level DDL, DCL, and transaction-control statements are rejected during prepare.

Second, the runtime write path could fail after payload size changed from small text to large text. The driver used cursor and bind metadata that had been prepared for smaller values, which could surface as Oracle errors such as `ORA-01461`, `ORA-01013`, or a closed socket. The runtime fix is in the Oracle driver, not in the EMQX connector.

`emqx_oracle` no longer rewrites or wraps user SQL for large payloads. The bumped driver now:

- includes bind format metadata in the cursor cache key, so small and large text bind shapes do not reuse an incompatible cursor;
- computes bind metadata from all rows in a batch, so a later large payload row is encoded with the correct bind format;
- derives named-bind order from the SQL text instead of relying on map iteration order;
- detects large text by encoded byte size, not Erlang list length;
- keeps normal DML and out-parameter statements on the existing driver execution path.

The EEC regression test uses a normal `t_mqtt_msgs` insert with `arrived` before `payload`, so the `NCLOB` payload bind is sent last. This verifies the size-change/cache fix without hiding Oracle's native LOB bind ordering restriction.

## Expected Side Effects

- Oracle action SQL templates that start with DDL, DCL, or transaction-control statements are rejected during prepare. This is intentional because those are not safe action templates.
- The same SQL text may use more than one cached cursor when payload sizes produce different bind formats. This avoids reusing an incompatible cursor, and the existing cursor limit still bounds the cache.
- The driver intentionally does not rewrite SQL or reorder user placeholders. Oracle's native LOB bind restrictions still apply, so users should place large `CLOB`/`NCLOB` bind placeholders after non-LOB placeholders when Oracle requires it.

## PR Checklist

[EEC-1322]: https://emqx.atlassian.net/browse/EEC-1322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/emqx_oracle/src/emqx_oracle.erl
+++ b/apps/emqx_oracle/src/emqx_oracle.erl
@@ -14,6 +14,10 @@
 -define(UNHEALTHY_TARGET_MSG,
     "Oracle table is invalid. Please check if the table exists in Oracle Database."
 ).
+-define(UNSUPPORTED_SQL_STATEMENT_MSG,
+    "unsupported_sql_statement: DDL, DCL, and transaction control statements are not supported "
+    "in Oracle Action SQL templates."
+).
 
 -define(DEFAULT_ROLE, normal).
 -define(SYSDBA_ROLE, sysdba).
@@ -63,6 +67,10 @@
 }).
 
 -type prepares() :: #{atom() => binary()}.
+-type prepare_state() ::
+    prepares()
+    | {error, _Reason :: term(), prepares()}
+    | {error, prepares()}.
 -type params_tokens() :: #{atom() => list()}.
 
 -type state() ::
@@ -70,7 +78,7 @@
         pool_name := binary(),
         installed_channels := map(),
         health_check_timeout := timeout(),
-        prepare_sql := prepares(),
+        prepare_sql := prepare_state(),
         params_tokens := params_tokens(),
         batch_params_tokens := params_tokens()
     }.
@@ -197,6 +205,8 @@ on_get_channel_status(
         {error, undefined_table} ->
             %% return new state indicating that we are connected but the target table is not created
             {?status_disconnected, {unhealthy_target, ?UNHEALTHY_TARGET_MSG}};
+        {error, unsupported_sql_statement} ->
+            {?status_disconnected, {unhealthy_target, ?UNSUPPORTED_SQL_STATEMENT_MSG}};
         {error, _Reason} ->
             %% do not log error, it is logged in prepare_sql_to_conn
             ?status_connecting
@@ -365,6 +375,14 @@ do_check_prepares(
     _ChannelId,
     _PoolName,
     #{
+        prepare_sql := {error, Reason, _Prepares}
+    } = _State
+) ->
+    {error, Reason};
+do_check_prepares(
+    _ChannelId,
+    _PoolName,
+    #{
         prepare_sql := {error, _Prepares}
     } = _State
 ) ->
@@ -387,8 +405,12 @@ do_check_prepares(
                 case ecpool_worker:client(WorkerPid) of
                     {ok, Conn} ->
                         case check_if_table_exists(Conn, SQL, Tokens) of
-                            {error, undefined_table} -> {error, undefined_table};
-                            _ -> ok
+                            {error, undefined_table} ->
+                                {error, undefined_table};
+                            {error, unsupported_sql_statement} ->
+                                {error, unsupported_sql_statement};
+                            _ ->
+                                ok
                         end;
                     _ ->
                         ok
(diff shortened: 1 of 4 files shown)


apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_aggregated_impl.erl
```diff
@@ -358,12 +358,8 @@ insert_report(ActionResId, Opts) ->
 
 connect(Opts) ->
     ConnectStr = conn_str(Opts),
-    %% Note: we don't use `emqx_secret:wrap/1` here because its return type is opaque, and
-    %% dialyzer then complains that it's being fed to a function that doesn't expect
-    %% something opaque...
-    ConnectStrWrapped = fun() -> ConnectStr end,
     DriverOpts = proplists:get_value(driver_options, Opts, []),
-    odbc:connect(ConnectStrWrapped, DriverOpts).
```
zmstone, line 366: why unwrap? the string contains secrets

apps/emqx_oracle/src/emqx_oracle.erl
```diff
@@ -536,79 +561,148 @@ prepare_sql_to_conn(Conn, [{Key, SQL} | PrepareList], TokensMap, Statements) whe
...
+        "  c := dbms_sql.open_cursor; "
+        "  dbms_sql.parse(c, :1, dbms_sql.native); "
+        "  dbms_sql.close_cursor(c); "
+        "exception "
+        "  when others then "
+        "    if dbms_sql.is_open(c) then "
+        "      dbms_sql.close_cursor(c); "
+        "    end if; "
+        "    raise; "
+        "end;",
```
thalesmg, line 577: I know there's a check for side effects before this is called, but maybe we should also attempt a transaction with rollback here, just in case the check is not perfect.

> This procedure parses the given statement in the given cursor. All statements are parsed immediately. <u>In addition, DDL statements are run immediately when parsed.</u>

e.g.: (untested) maybe just wrapping the statement with `begin ... end` would be enough to fool the check.

apps/emqx_oracle/src/emqx_oracle.erl
```diff
@@ -63,14 +67,18 @@
 }).
 
 -type prepares() :: #{atom() => binary()}.
+-type prepare_state() ::
+    prepares()
+    | {error, _Reason :: term(), prepares()}
+    | {error, prepares()}.
```
thalesmg: Q: why two distinct form for the errors?  maybe just keep one to avoid forgetting to handle both.

apps/emqx_oracle/src/emqx_oracle.erl
hjianbo: Thanks, this is a good point.  I checked this against a real Oracle instance to make sure the boundary is clear.

A transaction/rollback wrapper does not protect us from top-level DDL parse side effects.  I tested `DBMS_SQL.PARSE` with `CREATE TABLE ...` followed by `ROLLBACK`, and the table still existed afterwards.  I also tried `SET TRANSACTION READ ONLY` before the parse; Oracle still created the table during parse.  So adding a rollback here would give a false sense of safety.

The current guard is meant to prevent exactly those top-level statements whose side effects happen during parse.  It skips leading whitespace/comments, rejects top-level DDL/DCL/TCL-like statements before calling `DBMS_SQL.PARSE`, and only then uses parse-only validation for normal action DML.  This is why the direct `CREATE TABLE ...` case is rejected before parse.

I also tested the bypass shape mentioned here:

```sql
begin execute immediate 'CREATE TABLE ...'; end;
```

and a few similar forms (`DECLARE ... BEGIN execute immediate ... END`, labelled PL/SQL block, and `CALL proc_that_creates_table()`).  `DBMS_SQL.PARSE` successfully parses them, but it does not execute the PL/SQL body/procedure during prepare, and no table is created.  Those statements could still have side effects when the action is actually executed by runtime traffic, but they do not reintroduce the prepare/status side effect that this PR is fixing.

So I think the current approach is sufficient for this PR's scope: avoid executing user SQL during Oracle action prepare/status checks and reject the known top-level statements that Oracle may execute at parse time.  Adding rollback would not improve that protection for DDL.

changes/ee/fix-17605.en.md
```diff
@@ -0,0 +1 @@
+Fixed an issue where Oracle action validation could cause unintended side effects by running the configured SQL instead of only validating it. Also improved Oracle action support for writing text payloads larger than 4000 bytes when the payload placeholder is the last bind parameter in the SQL template.
```
thalesmg: according to your comment about `begin ... end`, it still can produce side effects if DDL is used.

thalesmg (dismissed):

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15828)

PR #17607, opened 2026-06-16
feat(topic_metrics): v2 support wildcard and prometheus

zmstone:
The existing topic metics has below issues

1. Config based interface (static vs dynamic)
2. No support of topic filters (wildcard)
3. No prometheus integration

This PR introduces topic-metrics v2.

Split the topic metrics implementation into three layers:
- emqx_topic_metrics2: facade module presenting the public API.
- emqx_topic_metrics_registry: persistence helpers (persist_*) that durably record metric registrations.
- cluster_rpc callbacks (do_*_local): pure ETS-only side-effect functions invoked per-node so cluster-wide changes converge through the standard cluster_rpc path rather than ad-hoc multicall fan-out.

Release version: 6.3.0

## PR Checklist


apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,395 @@
...
+-define(EXCEED_LIMIT, 'EXCEED_LIMIT').
+-define(ALREADY_EXISTS, 'ALREADY_EXISTS').
+-define(FORBIDDEN, 'FORBIDDEN').
+
+namespace() -> "topic_metrics2".
+
+api_spec() ->
+    emqx_dashboard_swagger:spec(?MODULE, #{check_schema => true}).
+
+scopes() -> ?SCOPE_CONNECTIONS.
```
zmstone: ?SCOPE_MONITORING

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,395 @@
...
+        ?PREFIX ++ "/:name",
+        ?PREFIX ++ "/:name/reset"
+    ].
+
+schema(?PREFIX) ->
+    #{
+        'operationId' => collections,
+        get => #{
+            tags => ?TAGS,
+            description => <<"List all topic-metric collections visible to the caller.">>,
```
zmstone: use ?DESC for descriptions

apps/emqx_topic_metrics/src/emqx_topic_metrics_hooks.erl
```diff
@@ -0,0 +1,124 @@
...
+
+%% Two indexes: the broad-category counter (in/out/dropped) and a
+%% per-QoS counter. `Idx2 = undefined' means there is no second
+%% counter (the dropped path).
+inc_matches(Topic, Msg, Idx1, Idx2) ->
+    case emqx_topic_metrics_registry:matches_with_record(Topic) of
+        [] ->
+            %% Common case when the feature is loaded but no
+            %% collection is registered — skip the namespace read
+            %% and the foreach.
```
zmstone: get_namespace is cheap, should get namespace and pass it to emqx_topic_metrics_registry:matches_with_record/2 which do not return records for other namespaces.

apps/emqx_topic_metrics/src/emqx_topic_metrics_registry.erl
```diff
@@ -0,0 +1,478 @@
...
+-export([
+    init/1,
+    handle_call/3,
+    handle_cast/2,
+    handle_info/2,
+    terminate/2
+]).
+
+-type owner_ns() :: ?global_ns | binary().
+-type name() :: binary().
```
zmstone: avoid name clash between ns, we should put ns in the db key
```suggestion
-type name() :: {owner_ns(), binary()}.
```

apps/emqx_topic_metrics/src/emqx_topic_metrics_registry.erl
```diff
@@ -0,0 +1,478 @@
...
+        end
+    end,
+    case mria:transaction(?COMMON_SHARD, Fun) of
+        {atomic, ok} -> {ok, ResetTime};
+        {atomic, missing} -> {error, not_found};
+        {aborted, Reason} -> {error, Reason}
+    end.
+
+-spec persist_deregister_all_owned_by(owner_ns() | all) -> {ok, [name()]}.
+persist_deregister_all_owned_by(Scope) ->
```
zmstone:
```suggestion
persist_deregister_all_owned_by(OwerNs) ->
```

apps/emqx_topic_metrics/src/emqx_topic_metrics_hooks.erl
```diff
@@ -0,0 +1,111 @@
...
+on_message_publish(#message{topic = Topic, payload = Payload} = Msg) ->
+    inc_matches(Topic, Msg, [{?IDX_IN, 1}, {?IDX_BYTES_IN, msg_size(Topic, Payload)}]),
+    ok.
+
+on_message_delivered(_ClientInfo, #message{topic = Topic, payload = Payload} = Msg) ->
+    %% Publisher namespace is the same for the delivered hook (it's
+    %% the publisher of the original message, not the subscriber).
+    %% This keeps "messages.out" symmetric with "messages.in" — both
+    %% count namespace-owned activity from the publisher's side.
+    inc_matches(Topic, Msg, [{?IDX_OUT, 1}, {?IDX_BYTES_OUT, msg_size(Topic, Payload)}]),
```
zmstone: change to msg_size(Msg)

apps/emqx_prometheus/src/emqx_prometheus_topic_metrics.erl
```diff
@@ -0,0 +1,186 @@
...
+    ]).
+
+with_node_label(?PROM_DATA_MODE__NODE, Labels) ->
+    Labels;
+with_node_label(?PROM_DATA_MODE__ALL_NODES_AGGREGATED, Labels) ->
+    Labels;
+with_node_label(?PROM_DATA_MODE__ALL_NODES_UNAGGREGATED, Labels) ->
+    [{node, node()} | Labels].
+
+ns_label(?global_ns) -> <<"$global">>;
```
thalesmg: maybe better to _not_ label global metrics?  otherwise, it's impossible to distinguish from a global metric and one from a namespace called `$global`.

apps/emqx_topic_metrics/src/proto/emqx_topic_metrics2_proto_v1.erl
```diff
@@ -0,0 +1,86 @@
...
+
+-spec list([node()], emqx_topic_metrics_registry:owner_ns() | all_ns) ->
+    emqx_rpc:multicall_result().
+list(Nodes, OwnerNs) ->
+    emqx_rpc:multicall(Nodes, emqx_topic_metrics2, list, [OwnerNs]).
+
+-spec lookup([node()], emqx_topic_metrics_registry:name()) ->
+    emqx_rpc:multicall_result().
+lookup(Nodes, Name) ->
+    emqx_rpc:multicall(Nodes, emqx_topic_metrics2, lookup, [Name]).
```
thalesmg: Q: why use `emqx_rpc`/`gen_rpc` for those instead of `erpc`?

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,378 @@
...
+                    case emqx_topic_metrics2:register(BinName, TopicFilter, OwnerNs) of
+                        ok ->
+                            {ok, Rec} = emqx_topic_metrics2:lookup(Name),
+                            {201, view(with_cluster_counters(Rec))};
+                        {error, Reason} ->
+                            err_create(Reason)
+                    end
+            end;
+        {error, Code, Msg} ->
+            {400, #{code => Code, message => Msg}}
```
thalesmg: maybe use `emqx/include/http_api.hrl`?

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,378 @@
...
+api_spec() ->
+    emqx_dashboard_swagger:spec(?MODULE, #{check_schema => true}).
+
+scopes() -> ?SCOPE_MONITORING.
+
+paths() ->
+    [
+        ?PREFIX,
+        ?PREFIX ++ "/:name",
+        ?PREFIX ++ "/:name/reset"
```
thalesmg: nit: writing it out explicitly helps with "grepability".

apps/emqx_topic_metrics/src/emqx_topic_metrics2.erl
```diff
@@ -0,0 +1,217 @@
...
+        ok -> emqx_topic_metrics_schema:validate_topic_filter(TopicFilter);
+        {error, _} = Err -> Err
+    end.
+
+normalize_ns(?global_ns) -> ?global_ns;
+normalize_ns(NS) when is_binary(NS) -> NS.
+
+normalize_scope(all_ns) -> all_ns;
+normalize_scope(?global_ns) -> ?global_ns;
+normalize_scope(NS) when is_binary(NS) -> NS.
```
thalesmg: 🤔 
it doesn't seem to normalize anything?

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,378 @@
...
+        ?global_ns -> ?global_ns;
+        NS when is_binary(NS) -> NS
+    end.
+
+%% Cluster-aggregated list. Short-circuits the RPC when this node is
+%% the only cluster member. A global admin sees every collection; a
+%% namespaced admin only sees collections they own.
+aggregated_list(OwnerNs) ->
+    Scope = list_scope(OwnerNs),
+    case mria:running_nodes() of
```
thalesmg:
```suggestion
    case emqx:running_nodes() of
```

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,378 @@
...
+
+%%--------------------------------------------------------------------
+%% Internal
+%%--------------------------------------------------------------------
+
+actor_ns(Req) ->
+    case emqx_dashboard:get_namespace(Req) of
+        ?global_ns -> ?global_ns;
+        NS when is_binary(NS) -> NS
+    end.
```
thalesmg:
```suggestion
    emqx_dashboard:get_namespace(Req).
```

?

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,378 @@
...
+        {ok, #{metrics := M0} = Existing} -> Acc#{Name => Existing#{metrics => sum_metrics(M0, M)}}
+    end;
+merge_into(_BadEntry, Acc) ->
+    Acc.
+
+sum_metrics(A, B) ->
+    maps:map(fun(K, V) -> V + maps:get(K, B, 0) end, A).
+
+with_cluster_counters(#{name := Name} = Rec) ->
+    case mria:running_nodes() of
```
thalesmg:
```suggestion
    case emqx:running_nodes() of
```

apps/emqx_topic_metrics/src/emqx_topic_metrics_registry.erl
```diff
@@ -0,0 +1,500 @@
...
+-export_type([owner_ns/0, bin_name/0, name/0, record/0]).
+
+%%--------------------------------------------------------------------
+%% Mnesia bootstrap
+%%--------------------------------------------------------------------
+
+create_tables() ->
+    Options = [
+        {type, ordered_set},
+        {rlog_shard, ?COMMON_SHARD},
```
thalesmg: q: is there a specific reason to use the same shard as `emqx` instead of the app having its own?

apps/emqx_topic_metrics/include/emqx_topic_metrics.hrl
```diff
@@ -0,0 +1,66 @@
...
+%% Record stored in ?MRIA_TAB. The mnesia key is the {OwnerNs, BinName}
+%% tuple in `name'. Counter refs are per-node and do NOT live in mria
+%% — every node rehydrates a fresh counter_ref on boot. Reset events
+%% are not persisted; each node emits its own audit log entry from
+%% the cluster_rpc reset callback (see emqx_topic_metrics_registry).
+-record(topic_metric, {
+    name :: {'_' | global | binary(), '_' | binary()},
+    topic_filter :: binary(),
+    create_time :: binary()
+}).
```
thalesmg, line 39: worth adding an `extra = #{} :: map()`?

apps/emqx_topic_metrics/src/emqx_topic_metrics_registry.erl
```diff
@@ -0,0 +1,500 @@
...
+    is_binary(TopicFilter),
+    is_binary(CreateTime),
+    ?IS_NAMESPACE(OwnerNs)
+->
+    Rec = #topic_metric{
+        name = Name,
+        topic_filter = TopicFilter,
+        create_time = CreateTime
+    },
+    Fun = fun() ->
```
thalesmg: maybe move this to a dedicated `_txn` fn and reference it like `fun M:F/N` to avoid sending lambdas as transactions (potential badfun?)?

apps/emqx_topic_metrics/src/emqx_topic_metrics_registry.erl
```diff
@@ -0,0 +1,500 @@
...
+                case scope_match(OwnerNs, N) of
+                    true ->
+                        ok = mnesia:delete({?MRIA_TAB, N}),
+                        {true, N};
+                    false ->
+                        false
+                end
+            end,
+            mnesia:foldl(fun(R, Acc) -> [R | Acc] end, [], ?MRIA_TAB)
+        ),
```
thalesmg: nit: maybe could leverage `mria:match_delete`?

apps/emqx_topic_metrics/src/emqx_topic_metrics_registry.erl
```diff
@@ -0,0 +1,500 @@
...
+%% shows only global-owned rows, a binary shows only that namespace.
+%% The "global admin sees everything" admin-policy decision lives one
+%% layer up (see `emqx_topic_metrics2_api:list_scope/1').
+-spec list(owner_ns() | all_ns) -> [record()].
+list(all_ns) ->
+    [expand(Name, Rec) || {Name, Rec} <- ets:tab2list(?REGISTRY_TAB)];
+list(Scope) ->
+    [
+        expand(Name, Rec)
+     || {{O, _Bin} = Name, Rec} <- ets:tab2list(?REGISTRY_TAB),
```
thalesmg: maybe `ets:match` (or `ets:select`) could avoid stepping through all rows?

apps/emqx_topic_metrics/src/emqx_topic_metrics_sup.erl
```diff
@@ -0,0 +1,26 @@
...
+
+-export([start_link/0]).
+-export([init/1]).
+
+start_link() ->
+    supervisor:start_link({local, ?MODULE}, ?MODULE, []).
+
+init([]) ->
+    {ok,
+        {{one_for_one, 10, 3600}, [
```
thalesmg: nit: use maps for the supervisor flags.

apps/emqx_topic_metrics/src/emqx_topic_metrics_registry.erl
```diff
@@ -0,0 +1,500 @@
...
+
+init([]) ->
+    process_flag(trap_exit, true),
+    ?REGISTRY_TAB = ets:new(?REGISTRY_TAB, [
+        ordered_set, named_table, public, {read_concurrency, true}
+    ]),
+    ?INDEX_TAB = ets:new(?INDEX_TAB, [
+        ordered_set, named_table, public, {read_concurrency, true}
+    ]),
+    ok = mria:wait_for_tables([?MRIA_TAB]),
```
thalesmg: we already wait for tables during app start, before starting the supervisor; maybe this is not needed?

apps/emqx_topic_metrics/src/emqx_topic_metrics_registry.erl
```diff
@@ -0,0 +1,500 @@
...
+%%--------------------------------------------------------------------
+
+init([]) ->
+    process_flag(trap_exit, true),
+    ?REGISTRY_TAB = ets:new(?REGISTRY_TAB, [
+        ordered_set, named_table, public, {read_concurrency, true}
+    ]),
+    ?INDEX_TAB = ets:new(?INDEX_TAB, [
+        ordered_set, named_table, public, {read_concurrency, true}
+    ]),
```
thalesmg: if this process restarts for whatever reason, it'll take the ets tables down with it, and several places that to use those tables could crash noisily.  maybe make the supervisor the table owner?

apps/emqx_topic_metrics/test/emqx_topic_metrics2_api_SUITE.erl
```diff
@@ -0,0 +1,189 @@
...
+
+suite() -> [{timetrap, {seconds, 30}}].
+
+all() -> emqx_common_test_helpers:all(?MODULE).
+
+init_per_suite(Config) ->
+    Apps = emqx_cth_suite:start(
+        [
+            emqx_conf,
+            {emqx, #{override_env => [{boot_modules, [broker]}]}},
```
thalesmg: q: is this option needed?

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,378 @@
...
+        ?PREFIX,
+        ?PREFIX ++ "/:name",
+        ?PREFIX ++ "/:name/reset"
+    ].
+
+schema(?PREFIX) ->
+    #{
+        'operationId' => collections,
+        get => #{
+            tags => ?TAGS,
```
thalesmg, line 70: should those endpoints use a filter to ensure the actor is a global admin (specially if it tries to read/create metrics for namespaces other than its own, if namespaced)?

apps/emqx_topic_metrics/test/emqx_topic_metrics2_api_SUITE.erl
```diff
@@ -0,0 +1,189 @@
...
+    out(emqx_topic_metrics2_api:collections(delete, req(OwnerNs))).
+
+get_one(OwnerNs, Name) ->
+    out(emqx_topic_metrics2_api:collection(get, (req(OwnerNs))#{bindings => #{name => Name}})).
+
+delete_one(OwnerNs, Name) ->
+    out(emqx_topic_metrics2_api:collection(delete, (req(OwnerNs))#{bindings => #{name => Name}})).
+
+reset_one(OwnerNs, Name) ->
+    out(emqx_topic_metrics2_api:reset(put, (req(OwnerNs))#{bindings => #{name => Name}})).
```
thalesmg: it would be great to actually call the HTTP API using a real request, so we don't (accidentally) mock something that does not match a real request behavior.

apps/emqx_topic_metrics/src/emqx_topic_metrics_registry.erl
zmstone: `mria:match_delete/2` doesn't return the deleted keys

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
zmstone: namespaced admins can only get items from their namespace, so there is no need to limit.

however, noticed another issue while thinking about this, the current implementation has isolation in storage and api, but the returned response only has `name`, if a global ns admin try to get all, it wont be able to tell which name is for which ns. so i plan to add a `id` field.

thalesmg: maybe adding `namespace => OwnerNs` would be clear?

also, on the topic of creating a collection.  currently, it seems global admins cannot create a collection for a specific namespace, only for global/all nss?

apps/emqx_topic_metrics/src/emqx_topic_metrics2.erl
```diff
@@ -0,0 +1,217 @@
...
+register(BinName, TopicFilter, OwnerNs) when is_binary(BinName), is_binary(TopicFilter) ->
+    case validate_inputs(BinName, TopicFilter) of
+        ok ->
+            CreateTime = emqx_utils_calendar:now_to_rfc3339(),
+            Name = {check_ns(OwnerNs), BinName},
+            case emqx_topic_metrics_registry:persist_register(Name, TopicFilter, CreateTime) of
+                {ok, new} ->
+                    %% Durable row is in mria; tell every node to
+                    %% materialise the local ETS overlay.
+                    _ = emqx_topic_metrics2_proto_v1:install_local(Name, TopicFilter, CreateTime),
```
thalesmg: nit: maybe log failures?

apps/emqx_topic_metrics/src/emqx_topic_metrics2.erl
```diff
@@ -0,0 +1,217 @@
...
+        ok -> emqx_topic_metrics_schema:validate_topic_filter(TopicFilter);
+        {error, _} = Err -> Err
+    end.
+
+check_ns(?global_ns) -> ?global_ns;
+check_ns(NS) when is_binary(NS) -> NS.
+
+check_scope(all_ns) -> all_ns;
+check_scope(?global_ns) -> ?global_ns;
+check_scope(NS) when is_binary(NS) -> NS.
```
thalesmg: this app's header introduces `?IS_NAMESPACE`, but doesn't seem to be used?

apps/emqx_topic_metrics/test/emqx_topic_metrics2_api_SUITE.erl
```diff
@@ -126,64 +167,115 @@ t_cap(_Config) ->
...
+    Headers = auth_header(Config, Ns),
+    case
+        emqx_mgmt_api_test_util:request_api(
+            Method, Path, _Qs = "", Headers, Body, #{return_all => true}
+        )
+    of
+        {ok, {{_HttpVer, Status, _Reason}, _RespHdrs, RespBody}} ->
+            {Status, decode(RespBody)};
+        {error, {{_HttpVer, Status, _Reason}, _RespHdrs, RespBody}} ->
+            {Status, decode(RespBody)}
```
thalesmg: nit: there is [`simple_request`](https://github.com/emqx/emqx/blob/3ada62fe5e3dc00766ee071c6ccf3294cf804268/apps/emqx_management/test/emqx_mgmt_api_test_util.erl#L362) which already does similar stuff with a convenient api.

apps/emqx_topic_metrics/src/emqx_topic_metrics2.erl
zmstone: done

apps/emqx_topic_metrics/src/emqx_topic_metrics2.erl
zmstone: fixed

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
zmstone: removed id, there is already namespace in the returned object.

apps/emqx_topic_metrics/src/emqx_topic_metrics2.erl
thalesmg: not quite?  if the RPC in this line (and others in this module) fails, it'll fail silently?

thalesmg (dismissed):

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
thalesmg: there still seems to be `id` in the returned objects (l. 342)

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,403 @@
...
+            %% layer, so we surface duplicate-name errors here to
+            %% match operator expectations of POST = create-only.
+            case emqx_topic_metrics2:lookup(Name) of
+                {ok, _} ->
+                    err_create(already_registered);
+                {error, not_found} ->
+                    case emqx_topic_metrics2:register(BinName, TopicFilter, OwnerNs) of
+                        ok ->
+                            {ok, Rec} = emqx_topic_metrics2:lookup(Name),
+                            {201, view(with_cluster_counters(Rec))};
```
thalesmg:
```suggestion
                            ?CREATED(view(with_cluster_counters(Rec)));
```

(and several other places in this file still not using http_api.hrl)

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
zmstone: forgot to push. now pushed.

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,380 @@
...
+            end;
+        {error, Code, Msg} ->
+            ?BAD_REQUEST(Code, Msg)
+    end;
+collections(delete, Req) ->
+    OwnerNs = actor_ns(Req),
+    %% A global admin's "delete all visible" really means "delete
+    %% everything"; a namespaced admin's call is scoped to their ns.
+    ok = emqx_topic_metrics2:deregister_all(list_scope(OwnerNs)),
+    {?NO_CONTENT}.
```
thalesmg:
```suggestion
    ?NO_CONTENT.
```

apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
```diff
@@ -0,0 +1,380 @@
...
+collection(get, #{bindings := #{name := BinName}} = Req) ->
+    OwnerNs = actor_ns(Req),
+    case emqx_topic_metrics2:lookup(BinName, OwnerNs) of
+        {ok, Rec} -> ?OK(view(with_cluster_counters(Rec)));
+        {error, not_found} -> not_found(BinName)
+    end;
+collection(delete, #{bindings := #{name := BinName}} = Req) ->
+    OwnerNs = actor_ns(Req),
+    case emqx_topic_metrics2:deregister(BinName, OwnerNs) of
+        ok -> {?NO_CONTENT};
```
thalesmg:
```suggestion
        ok -> ?NO_CONTENT;
```

thalesmg (dismissed):

thalesmg (approved):

PR #17624, opened 2026-06-17
fix(gcp pubsub consumer): correctly clear optvars when restarting (r58)

thalesmg:
Release version:
5.8.10, 5.10.4

## Summary

when a gcp consumer source is first created with a service account that lacks permissions to create subscriptions, it is marked (using optvar) as unhealthy.

in older ecpool versions (0.5.10), this would make the worker pool startup crash and fail to be added to the connector state.  even though the worker had set the optvar, its cleanup relied on the source being in the state, and hence would not be cleaned up even after restarting or re-creating the source and connector.  we now use the allocation apis to ensure they are cleaned up.

we also change the way workers announce their status for health checks: before, we used their own pids.  that could lead to leaks, however, in case they were forcefully killed by the manager (e.g.: stuck in a call).  we now use their numeric id, which is deterministic.

## PR Checklist


zmstone (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15628)

PR #17638, opened 2026-06-18
fix(cth): make cluster-start cleanup retry-safe

zmstone:
## Summary

Fixes a flake in `emqx_conf_app_SUITE:t_copy_deprecated_data_dir` (and similar cluster tests using the `flaky_tests/0` retry hook).

When a cluster-based test case fails partway through `emqx_cth_cluster:perform/3`, two leftovers survive into the retry:

1. The locally registered peer-controller names. The retry's `emqx_cth_peer:do_start/6` then crashes with `badarg` in `erlang:register/2` on names already bound to the previous attempt's pids.
2. The deterministic per-node `work_dir` is non-empty. The retry's `emqx_cth_suite:verify_clean_suite_state/1` does `{ok, []} = file:list_dir(WorkDir)` and `badmatch`es. This is the original transient `badmatch` from `erpc:call` that triggers the retry in the first place.

This PR adds:

- `emqx_cth_cluster:perform/3` best-effort `stop/1`s the node specs on any startup failure before re-raising, releasing the registered peer-controller names.
- `emqx_cth_cluster:do_perform/3` wipes each node's `work_dir` on a fresh `start` (only) so that `verify_clean_suite_state/1` can match `{ok, []}`. `restart` paths are left alone.
- `emqx_cth_peer:do_start/6` defensively handles a still-bound name by stopping the freshly started peer and erroring with `{already_registered, Node, OldPid}` instead of leaking a peer.

CI/test-only change; no changelog.

### DIFF
--- a/apps/emqx/test/emqx_cth_cluster.erl
+++ b/apps/emqx/test/emqx_cth_cluster.erl
@@ -107,6 +107,24 @@ start(NodeSpecs) ->
 
 perform(Act, NodeSpecs, Opts) ->
     ct:pal("~ping nodes: ~p", [Act, NodeSpecs]),
+    try
+        do_perform(Act, NodeSpecs, Opts)
+    catch
+        Kind:Reason:Stacktrace ->
+            %% A failure partway through startup leaves the already-started peer
+            %% control processes registered locally (by their node name). If we
+            %% leak them, a retry of the same test case (e.g. via the flaky-test
+            %% hook, `emqx_cth_ct_hook_flaky`) reuses the same node names and
+            %% `emqx_cth_peer:do_start/6`'s `erlang:register/2` would crash with
+            %% `badarg`. Best-effort cleanup before re-raising so a retry starts
+            %% from scratch.
+            Nodes = [Node || #{name := Node} <- NodeSpecs],
+            ct:pal("cleaning up partially started nodes after ~p:~p: ~p", [Kind, Reason, Nodes]),
+            catch stop(Nodes),
+            erlang:raise(Kind, Reason, Stacktrace)
+    end.
+
+do_perform(Act, NodeSpecs, Opts) ->
     % 1. Start bare nodes with only basic applications running
     ok = start_nodes_init(NodeSpecs, ?TIMEOUT_NODE_START_MS),
     % 2. Start applications needed to enable clustering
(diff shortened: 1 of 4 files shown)


savonarola (dismissed):

apps/emqx/test/emqx_cth_cluster.erl
```diff
@@ -107,6 +107,33 @@ start(NodeSpecs) ->
...
+do_perform(Act, NodeSpecs, Opts) ->
+    % 0. On a fresh start, make sure each node begins from a clean work dir,
+    % unless the caller opted out via `work_dir_dirty => true`. A flaky-test
+    % retry (or any re-run) reuses the same deterministic work dir, and
+    % leftover data (mnesia/ds/configs/log) would otherwise make
+    % `emqx_cth_suite:verify_clean_suite_state/1` `badmatch` on the next
+    % attempt. `restart` keeps existing data intact by design, so only `start`
+    % cleans. NOTE: must run before `start_nodes_init/2` so we never delete a
+    % peer node's current working directory underneath it.
+    Act =:= start andalso clean_node_work_dirs(NodeSpecs),
```
keynslug: > A flaky-test retry (or any re-run) reuses the same deterministic work dir.

In this case in my opinion it's a wrong place to do it. As `emqx_cth_ct_hook_flaky` now has its own `cleanup_for_retry` I'd suggest to drop it here.

zmstone: Good call — dropped the `clean_node_work_dirs` wipe from `emqx_cth_cluster` in 2306e6340a.

One thing I had to add to make that safe: `emqx_conf_app_SUITE` declared `flaky_tests/0` but never registered `emqx_cth_ct_hook_flaky` via `suite/0`, so its retries are driven purely by the native `{flaky, N}` repeat and the hook's `cleanup_for_retry` never ran for it — which is why the wipe lived in the cluster helper. I've now registered the hook in the suite (matching the other flaky cluster suites), so the hook handles the work-dir cleanup and the helper no longer needs to.

The `perform/3` cleanup-on-failure stays, since the hook doesn't release leaked peer-controller name registrations.

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15774)

PR #17639, opened 2026-06-18
test(stomp): wait for previous session to drain in t_subscribe_inuse

zmstone:
## Summary

`t_subscribe_inuse` calls `with_connection/1` four times sequentially. After
each call the client-side socket is closed, but the gateway's STOMP channel
process tears down the session asynchronously, so the next `with_connection/1`
can start before the previous client has been unregistered from `emqx_cm`.
The strict-singleton match `[#{clientid := ClientId}] = clients()` then fails
when two sessions are briefly listed.

Wrap the two listing matches in `?retry(100, 30, ...)` so they wait up to
3 seconds for exactly one client to be listed.

Verified with 18 consecutive green runs of `t_subscribe_inuse` in a tight
loop (plus a full-suite run, 20/20 green). Test-only change.

### DIFF
--- a/apps/emqx_gateway_stomp/test/emqx_stomp_SUITE.erl
+++ b/apps/emqx_gateway_stomp/test/emqx_stomp_SUITE.erl
@@ -322,7 +322,10 @@ t_subscribe_inuse(_) ->
         fun(Sock) ->
             Setup(Sock),
             %% assert subscription stats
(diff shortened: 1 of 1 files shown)


savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15699)

PR #17640, opened 2026-06-18
test(mgmt): allocate free ports via select_free_port in listeners SUITE

zmstone:
The listener CRUD tests in `emqx_mgmt_api_listeners_SUITE` derived bind ports from `20000 + ?LINE` (per type, e.g. ws:new at 33000 + 305 = 33305). Stable across runs, but not unique across hosts or concurrent CT shards — any other process holding the exact port causes esockd to return `eaddrinuse`, the POST handler turns that into HTTP 400, and the next `maps:keys/1` call on the error tuple crashes with `{badmap,...}`. We hit this in CI as a flake on `t_ws_crud_listeners_by_id`.

Use `emqx_common_test_helpers:select_free_port(tcp)` at the positive bind sites instead. The deliberate negative-path cases (`tcp:bad` / `ssl:bad` / `ws:bad` / `wss:bad`) keep reusing the just-bound port to test conflict handling.

Reproduced locally by holding port 33305 with `nc -l`: the suite went from 22/24 passing to 24/24 with the fix.

### DIFF
--- a/apps/emqx_management/test/emqx_mgmt_api_listeners_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_api_listeners_SUITE.erl
@@ -9,8 +9,7 @@
 -include_lib("eunit/include/eunit.hrl").
 -include_lib("common_test/include/ct.hrl").
 
--define(PORT(Base), (Base + ?LINE)).
--define(PORT, ?PORT(20000)).
+-define(PORT, emqx_common_test_helpers:select_free_port(tcp)).
 
 all() ->
     [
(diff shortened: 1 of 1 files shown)


savonarola (dismissed):

zmstone:
rebased on latest release-60

ieQu1 (approved):

keynslug (approved):

qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15768)

PR #17645, opened 2026-06-18
fix(jwt): suppress empty TE header in JWKS HTTP request

zmstone:
Release version: 5.8.11, 5.10.5

## Summary

EMQX's JWKS HTTP client (used by JWT authentication to fetch signing keys) emitted an RFC 2616-violating empty `TE:` header on every request. This is a long-standing default in Erlang/OTP's `inets` HTTP client, fixed upstream only in inets 9.4.2 / OTP 28.1. Strict identity providers — notably PingFederate — reject such requests with `503` directly, or a TCP reset behind a load balancer, preventing JWKS key refresh from working.

The fix injects an explicit `te: trailers` header at the JWKS call site (only when the user-supplied header map does not already provide one), so the broken inets default never reaches the wire. No other behaviour changes.

### Before / after wire bytes

Captured against a `nc -l` listener on the affected OTP 26 / inets 9.1.0.2 (this branch's toolchain). Trailing `^M$` shows `\r\n` line endings.

Before (empty `te:` value, three spaces after the colon — actually one space and then CRLF):

```
GET /pf/JWKS HTTP/1.1
content-length: 0
te: 
host: 127.0.0.1:18099
connection: keep-alive
```

After the fix (captured by the new CT case against the patched module):

```
GET /jwks.json HTTP/1.1
content-length: 0
te: trailers
host: 127.0.0.1:45381
accept: application/json
connection: keep-alive
```

### Tests

- New `apps/emqx_auth_jwt/test/emqx_authn_jwks_client_SUITE.erl` spawns a raw `gen_tcp` listener, drives `emqx_authn_jwks_client` against it, and asserts the request bytes contain `te: trailers` and do not contain a `te: \r\n` (empty-value) line.

## PR Checklist


qzhuyan (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15630)

PR #17646, opened 2026-06-18
fix(jwt): send explicit TE header in JWKS HTTP client

zmstone:
Release version:
6.0.4, 6.1.3, 6.2.2

## Summary

When EMQX fetches a JWKS for JWT authentication, the underlying `inets`
HTTP client emits a default empty `TE:` header. This is an RFC 2616
violation that was fixed upstream in `inets` 9.4.2 (OTP 28.1). EMQX
release-60 ships OTP 27.x with `inets` 9.3.2, so the broken default
still reaches the wire.

A customer reported that PingFederate rejects the JWKS request: it
returns `503` when fronted by a load balancer, or a TCP RST when hit
directly. Other strict gateways are likely to behave similarly.

This change injects an explicit `TE: trailers` header at the JWKS HTTP
call site (`apps/emqx_auth_jwt/src/emqx_authn_jwks_client.erl`) so the
broken default never reaches the wire. A user-supplied `te` header
takes precedence and is preserved verbatim. The fix is scoped to the
only auth path that hits external IdP endpoints via `httpc:request/4`
(SAML/SSO modules use other HTTP clients and are unaffected).

Two CT cases (`t_jwks_te_header`, `t_jwks_te_header_user_supplied`) use
a raw `gen_tcp` listener to inspect the actual request bytes and assert
that no empty `te:` header is emitted, and that user-supplied `te`
values are preserved.

### Wire capture

Before (OTP 27.3.4.2, `inets` 9.3.2):

```
GET /pf/JWKS HTTP/1.1
content-length: 0
te: 
host: 127.0.0.1:18099
connection: keep-alive
```

After (same OTP, with this fix):

```
GET /pf/JWKS HTTP/1.1
content-length: 0
te: trailers
host: 127.0.0.1:18099
connection: keep-alive
```

The `chore(gcp_pubsub)` commit in this PR is a drive-by version bump
required by `scripts/apps-version-check.exs` on `release-60`; it is
unrelated to the JWT fix.

## PR Checklist


ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15702)

zmstone:
fix ported to v5: https://github.com/emqx/emqx/pull/17645

PR #17648, opened 2026-06-18
sync `release-59` to `release-510`

thalesmg:


apps/emqx_gateway_jt808/src/emqx_jt808_channel.erl
thalesmg, line 1: @hjianbo there were conflicts in this file and in the suite.  could you please help review if the resolution is correct?

hjianbo (approved):

PR #17675, opened 2026-06-22
feat(metrics): increment messages.retained counter

### LINKED ISSUE #17385: [retainer] Expose retained message write counter as a metric
## Current state in `release-62`

- `apps/emqx_prometheus/src/emqx_prometheus.erl:1042` already declares `{emqx_messages_retained, counter, 'messages.retained'}` and `emqx_prometheus.erl:561` collects it.
- However `'messages.retained'` is **not registered** in `apps/emqx/src/emqx_metrics.erl` and is **never incremented** anywhere in the codebase.
- The `?C` macro defaults missing keys to `0`, so `emqx_messages_retained` is currently exposed in Prometheus output as a constant `0`. This is misleading — it looks like a usable metric but is non-functional.

## Proposed change

Expose a monotonic counter of retained message writes (not a rate — monitoring systems can derive rate from the non-negative derivative of the counter, which is the standard Prometheus pattern).

Implementation sketch:
1. Add `'messages.retained'` to `apps/emqx/src/emqx_metrics.erl` (reserved_idx + counter registration + i18n description).
2. Call `emqx_metrics:inc('messages.retained')` from the retainer write path (e.g. `apps/emqx_retainer/src/emqx_retainer_publisher.erl`).
3. No change needed in `emqx_prometheus.erl` — the plumbing is already there.

## Why a counter, not a rate

- A counter preserves precision across scrape intervals.
- It is also the only sane choice if the metric is to be aggregated across a cluster.
- This matches the convention of every other `messages.*` metric already in EMQX.

## User-facing value

- Operators can set the retained-write limit based on measured production load + a buffer, instead of guessing.
- Existing Prometheus/Grafana dashboards can add a `rate(emqx_messages_retained[1m])` panel with no app-side instrumentation in customer services.

### Why is this needed?

EMQX exposes a configurable rate limit for retained message writes (mqtt.retainer.max_publish_rate / equivalent), but provides no metric to observe the actual retained-write rate. Operators are therefore unable to size the limit based on evidence — they either set it conservatively and risk dropping legitimate updates, or set it very high (the reporting user has it at 100,000/s) and lose the protection the limit is intended to provide.

The only retained-related observability today is two gauges (emqx_retained_count, emqx_retained_max) — these reflect the size of the retained store, not the write rate.

sahilpatel09:
Release version:

6.0.4, 6.1.3, 6.2.2

## Summary
Fixes #17385.

EMQX already exposed `emqx_messages_retained` in Prometheus but the metric was never registered as a counter and never incremented — it always showed 0.

This patch:
1. Registers `messages.retained` as a counter metric in `emqx_metrics.erl`
2. Increments it via `inc_global` after a successful retained message store
3. Adds test assertions in `t_store_and_clean` to verify the counter increments on publish and stays unchanged on delete

## PR Checklist

- [X] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

### DIFF
--- a/apps/emqx/src/emqx_metrics.erl
+++ b/apps/emqx/src/emqx_metrics.erl
@@ -617,6 +617,7 @@ reserved_idx('actions.executed') -> 101;
 reserved_idx('delivery.dropped.filter') -> 102;
 reserved_idx('actions.messages') -> 103;
 reserved_idx('client.post_authn') -> 104;
+reserved_idx('messages.retained') -> 105;
 reserved_idx(_) -> undefined.
 
 all_metrics() ->
@@ -711,7 +712,8 @@ message_metrics() ->
         {counter, 'messages.validation_succeeded', ?DESC("messages_validation_succeeded")},
         {counter, 'messages.transformation_failed', ?DESC("messages_transformation_failed")},
         {counter, 'messages.transformation_succeeded', ?DESC("messages_transformation_succeeded")},
-        {counter, 'messages.persisted', ?DESC("messages_persisted")}
+        {counter, 'messages.persisted', ?DESC("messages_persisted")},
+        {counter, 'messages.retained', ?DESC("messages_retained")}
     ].
 
 %% Rule and Action related metrics
--- a/apps/emqx_retainer/src/emqx_retainer_publisher.erl
+++ b/apps/emqx_retainer/src/emqx_retainer_publisher.erl
@@ -56,7 +56,8 @@ store_retained(#message{topic = Topic, payload = Payload} = Msg) ->
             end),
             case WithinLimits of
                 true ->
-                    ?tp(retain_within_limit, #{topic => Topic});
+                    ?tp(retain_within_limit, #{topic => Topic}),
+                    emqx_metrics:inc_global('messages.retained');
                 {false, Reason} ->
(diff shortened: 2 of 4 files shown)


sahilpatel09:
cc @zmstone 
I am new here so please let me know if I am doing something wrong other than the convention of this repo. Thanks!

zmstone:
thank you for the pr.

looks good in general.

one minor: the commit message (and pr titile) is better to be `feat` instead of `fix`.

sahilpatel09:
@zmstone Hi, is this PR still good? I do not know if those failing tests are flaky or real. Do you want me to look into it further? Previously, there was only one test that failed which successfully passed locally in docker environment.

zmstone:
Hi @sahilpatel09 most likely flaky.
there are many flaky fixes merged to dev-xx branches, not reached master yet.
i picked your commit to a new pr targeting `dev-62`.

but if you can rebase this pr to dev-62, we can continue with this one.
otherwise if you are happy with https://github.com/emqx/emqx/pull/17758, we can close this one.

sahilpatel09:
@zmstone Thanks for looking into this. I am completely okay to go ahead with #17758. Please feel free to close this one!

zmstone:
thanks, will be included in release 6.2.2 6.3.0
and will be forwarded to master soon.

PR #17679, opened 2026-06-23
test(prom): de-flake t_listener_shutdown_count

zmstone:
## Summary

\`ConnectRetry\` uses \`emqtt:start_link/1\` per attempt. When the broker rejects CONNECT with \`server_unavailable\` (open-session throttle), the emqtt process exits with \`{shutdown, server_unavailable}\`. The \`try/catch\` handled the synchronous exit from \`gen_statem:call\`, but the async link signal would still kill the test process before the retry could fire.

Fix is to unlink before emqtt:connect and link again once connected.

## PR Checklist


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15784)

PR #17680, opened 2026-06-23
test: de-flake mqtt expiry-interval and durable takeover suites

zmstone:
Release version: 6.0.3

## Summary

Test-only change. De-flakes two unrelated flaky CT failures seen on this branch's CI.

### 1. `emqx_mqtt_SUITE` message-expiry-interval tests

`t_message_expiry_interval` and `t_message_not_expiry_interval` ran three QoS iterations (0/1/2) reusing the **fixed** clientids `Client-Publish` / `Client-Verify` / `Client-Control`, all with `clean_start = false` and `Session-Expiry-Interval = 360`. Because the sessions persist, a resumed `Client-Verify` could carry stale queued/subscription state from a previous iteration (or from the sibling test), and rapid reconnects of the same clientid raced session takeover on slow CI — producing intermittent `should_have_expired` / `no_publish_received` failures.

- Each QoS iteration now gets a **fresh, test-tagged** set of clientids so no session/mqueue/subscription state leaks between iterations or between the two tests.
- Widened the must-receive window in the not-expired case (300ms → 2000ms) to tolerate slow session-resume dispatch on busy runners.

Verified locally: 10/10 consecutive green runs of the two cases.

### 2. `emqx_takeover_SUITE` takenover-exit assertion for durable sessions

When a new client takes over an existing session, in-memory sessions deliver a DISCONNECT with the precise `RC_SESSION_TAKEN_OVER` reason code, so the old client exits with `{shutdown, {disconnected, ?RC_SESSION_TAKEN_OVER, _}}`. For durable (DS) sessions the takeover kick of the old channel is best-effort: the session is taken over regardless, but the broker may step the old channel down with a different reason or slightly later, and may not deliver a clean disconnect to the old socket (a known core-CM race, tracked separately). Asserting the exact reason within the default 1s window made every takenover-asserting case flaky on busy CI — `t_takeover_session_then_abnormal_disconnect_2` failed this way.

- Relaxed `assert_client_exit/3` so that, **for durable sessions only**, it requires the old client to eventually terminate (generous timeout, any reason); the strict `RC_SESSION_TAKEN_OVER` assertion is kept for in-memory sessions.

Verified locally: full `emqx_takeover_SUITE` green (46 ok, 2 skipped).

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_takeover_SUITE.erl
+++ b/apps/emqx/test/emqx_takeover_SUITE.erl
@@ -1126,8 +1126,22 @@ filter_payload(List, Payload) when is_binary(Payload) ->
 
 %% @doc assert emqtt *client* process exits as expected.
 assert_client_exit(Pid, v5, takenover) ->
-    %% @ref: MQTT 5.0 spec [MQTT-3.1.4-3]
-    ?assertReceive({'EXIT', Pid, {shutdown, {disconnected, ?RC_SESSION_TAKEN_OVER, _}}});
+    case emqx_persistent_message:is_persistence_enabled() of
+        false ->
+            %% In-memory sessions deliver a DISCONNECT with the precise
+            %% RC_SESSION_TAKEN_OVER reason code.
+            %% @ref: MQTT 5.0 spec [MQTT-3.1.4-3]
+            ?assertReceive({'EXIT', Pid, {shutdown, {disconnected, ?RC_SESSION_TAKEN_OVER, _}}});
+        true ->
+            %% For durable (DS) sessions the takeover kick of the previously
+            %% connected channel is best-effort: the session is taken over
+            %% regardless, but the broker may step the old channel down with a
+            %% different reason or slightly later, and may not deliver a clean
+            %% DISCONNECT(RC_SESSION_TAKEN_OVER) to the old socket (a known
+            %% core-CM race). Only require that the old client eventually
+            %% terminates, tolerating the exact reason and timing.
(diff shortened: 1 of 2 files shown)


hjianbo (dismissed):

apps/emqx/test/emqx_takeover_SUITE.erl
```diff
@@ -29,8 +29,24 @@ all() ->
     ].
 
 flaky_tests() ->
+    %% These cases assert that an old client is disconnected with reason
+    %% RC_SESSION_TAKEN_OVER when a new client takes over its session. For
+    %% durable (DS) sessions the takeover kick is best-effort and can
+    %% intermittently fail to step the old channel down (a known core-CM race,
+    %% tracked separately), so the assertion flakes on busy CI. Retry these
```
keynslug: Maybe better to relax this assertion instead? Marking half of the test suite as flaky sounds too broad, likely to hide regressions later.

zmstone: Good call — done. Reverted the broad `flaky_tests` marking and instead relaxed `assert_client_exit/3`: for durable (DS) sessions it now only requires the old client to eventually terminate (generous timeout, any reason), since the takeover kick of the old channel is best-effort there. The strict `RC_SESSION_TAKEN_OVER` assertion is kept for in-memory sessions.

keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15780)

PR #17688, opened 2026-06-23
test(bridge_mqtt): de-flake t_reconnect drain

zmstone:
## Summary

\`emqx_utils:drain_down/1\` returns whatever is already in the mailbox (\`after 0\`) and does not wait. After \`kick_clients_http\`, the test process called \`drain_down(2)\` before both DOWN signals had propagated through \`emqx_cm\` + channel teardown, leaving \`DownPids\` with only the one already-arrived pid:

\`\`\`
expected: [Pid1, Pid2]
got: [Pid1]
\`\`\`

Replace \`drain_down/1\` with a local \`wait_for_downs/2\` helper that polls the mailbox until N DOWNs arrive or a 5s deadline elapses (then fails loudly with a useful message).

The second \`drain_down\` use in this suite (line 820) ignores its return value and is left as-is.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_action_SUITE.erl
+++ b/apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_action_SUITE.erl
@@ -340,6 +340,25 @@ get_emqtt_clients(PoolName) ->
         ecpool:workers(PoolName)
     ).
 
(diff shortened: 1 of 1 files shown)


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15831)

PR #17697, opened 2026-06-23
test(mqtt): de-flake t_message_expiry_interval for QoS>0

zmstone:
Release version: 6.0.3

## Summary

Test-only de-flake of `emqx_mqtt_SUITE:t_message_expiry_interval` (no user-facing change).

`message_expiry_interval_init` stops `Client-Verify` and the test publishes immediately afterwards. The broker may not have processed the DISCONNECT yet, so the channel is still `connected` when the message is delivered. A connected session places a QoS>0 message into its **inflight** window (`emqx_session_mem:deliver_msg`) instead of the **mqueue** (the `disconnected` branch of `emqx_channel:handle_deliver`). On resume, `emqx_session_mem:replay/2` resends inflight messages verbatim, **bypassing** the `is_expired` check that `dequeue` applies to mqueue messages — so an already-expired message gets delivered and the test fails with `should_have_expired`.

QoS=0 always lands in the mqueue (expiry-checked), which is why only QoS>0 flaked — most often QoS=2, whose longer PUBLISH/PUBREL handshake widens the race window relative to the disconnect.

Fix: after stopping `Client-Verify`, wait until the broker reports its session as disconnected before publishing, so the queued message deterministically lands in the mqueue and is subject to the expiry check.

Validated by looping `t_message_expiry_interval` 20× and `t_message_not_expiry_interval` 10× locally, all green.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_mqtt_SUITE.erl
+++ b/apps/emqx/test/emqx_mqtt_SUITE.erl
@@ -110,7 +110,36 @@ message_expiry_interval_init(Tag) ->
     emqtt:subscribe(CControl, <<"t/a">>, 1),
     emqtt:subscribe(CVerify, <<"t/a">>, 1),
     emqtt:stop(CVerify),
+    %% Wait until the broker has fully processed Client-Verify's DISCONNECT and
+    %% moved its (retained) session into the `disconnected` state before any
+    %% message is published. Until then the channel is still `connected`, so a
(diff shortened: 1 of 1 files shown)


hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15833)

PR #17698, opened 2026-06-23
fix(session): drop expired in-flight messages at takeover

zmstone:

## Summary

`emqx_session_mem:replay/2` resent every in-flight message verbatim at session
takeover, while `dequeue/4` already dropped queued messages whose
`Message-Expiry-Interval` had elapsed. Because of this asymmetry, an expired
QoS 1 / QoS 2 message that was **in-flight** (sent but not yet acknowledged) when
the client disconnected was redelivered on reconnect, whereas the same message
would have been dropped had it still been in the session queue.

This applies the same expiry check to in-flight messages in `replay/2`
(`apps/emqx/src/emqx_session_mem.erl`). Expired entries are removed from the
in-flight set and emit the `{expired, Msg}` session event (firing the
`delivery.dropped` hook with reason `expired`), mirroring the queue path.
`wait_comp` entries (post-PUBREC) are preserved so the QoS-2 second-half
handshake (PUBREL/PUBCOMP) still completes — the message was already delivered
and acknowledged once, so expiry of the original message must not abort it.

New CT case `t_inflight_expired_dropped_on_takeover` covers the in-flight path;
it fails against the old code and passes with the fix (verified, plus 10x loop).
Existing `t_message_expiry_interval` (queue path) and
`t_puback_not_lost_on_disconnect` (non-expired in-flight replay) still pass.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_session_mem.erl
+++ b/apps/emqx/src/emqx_session_mem.erl
@@ -785,18 +785,38 @@ filter_remote_pendings(ClientInfo, Session, Pendings) ->
 
 -spec replay(emqx_types:clientinfo(), session()) ->
     {ok, replies(), session()}.
-replay(ClientInfo, Session) ->
-    PubsResend = lists:map(
-        fun
-            ({PacketId, #inflight_data{phase = wait_comp}}) ->
-                {pubrel, PacketId};
-            ({PacketId, #inflight_data{message = Msg}}) ->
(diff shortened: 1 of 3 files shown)


zmstone:
inflight messages should be completed even after expired --- this is from mqtt spec.


[MQTT-3.3.2-5] | If the Message Expiry Interval has passed and the Server has not managed to start onward delivery to a matching subscriber, then it MUST delete the copy of the message for that subscriber.

PR #17702, opened 2026-06-24
docs(session): document and test inflight-vs-queued expiry asymmetry

zmstone:
## Summary

Documents and pins the behavior of `emqx_session_mem:replay/2`: inflight messages
are resent verbatim at session takeover while `dequeue/4` drops expired queued
messages.

The asymmetry looks like a bug (and was mistaken for one in a now-closed PR) but
is mandated by [MQTT-3.3.2-5]: an expired message is only deleted if the server
has *not* started onward delivery. An inflight message has already been sent
(awaiting PUBACK/PUBREC/PUBCOMP), so its delivery must complete regardless of
`Message-Expiry-Interval`; a still-queued message has not, so it is dropped.

Changes:
- A comment on `replay/2` explaining the invariant so the asymmetry is not
  re-mistaken for a defect.
- A regression test `t_inflight_retried_after_expiry_on_takeover` asserting that
  an inflight QoS-1 message expired during a disconnect is still retried (as a
  DUP) on reconnect — the inverse of the queued case already covered by
  `t_message_expiry_interval`.

No behavior change, no changelog (internal documentation + test only).

## PR Checklist
- [ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files (N/A — no user-visible change)

### DIFF
--- a/apps/emqx/test/emqx_mqtt_SUITE.erl
+++ b/apps/emqx/test/emqx_mqtt_SUITE.erl
@@ -248,6 +248,66 @@ t_puback_not_lost_on_disconnect(_) ->
     ?assertNot(lists:member(<<"1">>, Msgs1)),
     emqtt:stop(C1).
 
+-doc "An inflight (sent-but-unacked) QoS-1 message is retried at takeover even after its Message-Expiry-Interval elapsed: per [MQTT-3.3.2-5] expiry only drops messages whose onward delivery has not started.".
+t_inflight_retried_after_expiry_on_takeover(_) ->
+    ClientId = <<"inflight_retry_after_expiry">>,
+    {ok, CPub} = emqtt:start_link([
+        {proto_ver, v5},
+        {clientid, <<"inflight_retry_pub">>}
+    ]),
+    {ok, _} = emqtt:connect(CPub),
+    %% Subscriber with a persistent in-mem session, manual acking so the broker
+    %% keeps the delivered message in its inflight set awaiting PUBACK.
+    {ok, C0} = emqtt:start_link([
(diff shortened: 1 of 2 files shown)


apps/emqx/test/emqx_mqtt_SUITE.erl
```diff
@@ -248,6 +248,66 @@ t_puback_not_lost_on_disconnect(_) ->
...
+        [{qos, 1}]
+    ),
+    %% The message reaches the subscriber and enters the session inflight set;
+    %% onward delivery has started, but we deliberately do NOT acknowledge it.
+    receive
+        {publish, #{client_pid := C0, topic := <<"t/a">>}} ->
+            ok
+    after 2000 ->
+        ct:fail(should_receive_publish)
+    end,
```
keynslug: Micronit.
```suggestion
    {publish, #{client_pid := C0}} = ?assertReceive({publish, #{topic := <<"t/a">>}}),
```

keynslug (dismissed):

apps/emqx/test/emqx_mqtt_SUITE.erl
zmstone: Applied, thanks — switched to `?assertReceive` (added the `emqx/include/asserts.hrl` include). Also wrapped the over-long `-doc` line that tripped the elvis line_length sanity check.

ieQu1 (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15834)

PR #17705, opened 2026-06-24
test(bridge_mqtt): fix flaky t_resubscribe_on_fast_failure

zmstone:
Release version: 6.1.2

## Summary

Fixes the flaky `emqx_bridge_mqtt_source_SUITE:t_resubscribe_on_fast_failure`.

The test runs on a 3-node cluster, and each node runs its own MQTT connector pool. When the source broker is restarted, every worker on every node re-runs its reconnect callback for each source channel, emitting a `"mqtt_source_reconnected"` snabbkaffe event:

```
NumNodes * num_sources * pool_size
```

The test only waited for `num_sources * pool_size` events (the comment even said "2 sources with 3 workers each" — it forgot the cluster dimension). Once that many events arrived it published immediately. Any cluster node whose `SUBSCRIBE` to the source broker had not yet completed silently dropped the QoS-1 publish, so fewer republishes than `NumNodes` were produced and `?assertReceivePublish` hit an empty mailbox.

The fix waits for the reconnect events from **all** cluster nodes (`NumNodes * ...`) before publishing, in both restart phases of the test.

Test-only change; no user-facing behavior is affected.

Verified by looping the previously-failing variant (`cluster.proto_v3.clean_start`) — consistently green.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
+++ b/apps/emqx_bridge_mqtt/test/emqx_bridge_mqtt_source_SUITE.erl
@@ -933,8 +933,10 @@ t_resubscribe_on_fast_failure(TCConfig) when is_list(TCConfig) ->
     ct:pal("Restarting source node (1)"),
     {ok, SRef0} = snabbkaffe:subscribe(
         ?match_event(#{?snk_kind := "mqtt_source_reconnected"}),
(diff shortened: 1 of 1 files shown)


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15835)

PR #17706, opened 2026-06-24
test(crl): de-flake t_cache_overflow against async cache events

zmstone:
Release version: 6.0.4

## Summary

De-flake `emqx_crl_cache_SUITE:t_cache_overflow`.

The test drives connections in phases delimited by snabbkaffe markers
(`first_connections`, `first_reconnections`, `first_eviction`,
`second_eviction`, `test_end`) and asserts the events captured in each
window. Two unrelated event streams are mixed in those windows:

* `mqtt_client_connection` span events — emitted **synchronously** by the
  test client processes, so they always fall in the correct window.
* cache mutation events (`new_crl_url_inserted`, `crl_cache_overflow`,
  `crl_cache_ensure_timer`) — emitted **asynchronously** by the
  `emqx_crl_cache` gen_server.

When the gen_server was slow, a cache event for one URL could be emitted
*after* the test process had already advanced to the next phase marker,
landing it in the wrong window and leaving the asserted window short. The
observed CI failure showed exactly this: `Phase1` was missing URL2's
`new_crl_url_inserted` / `crl_cache_ensure_timer`, which had leaked into the
next window.

The fix keeps asserting the connection spans per phase window (reliable),
but asserts the cache-mutation events as a single deterministic stream over
the whole trace. Since all cache events come from the one gen_server, their
global order is fixed (URL1, URL2 fill the cache; URL3 evicts the oldest
URL1; URL1 re-enters and evicts the new oldest URL2), so window placement no
longer matters.

Only `apps/emqx/test/emqx_crl_cache_SUITE.erl` is touched; no production
code changes. Looped locally 20x with no failures.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_crl_cache_SUITE.erl
+++ b/apps/emqx/test/emqx_crl_cache_SUITE.erl
@@ -466,22 +466,12 @@ assert_successful_connection(Config, ClientNum) ->
         end
     ).
 
-trace_between(Trace0, Marker1, Marker2) ->
-    {Trace1, [_ | _]} = ?split_trace_at(#{?snk_kind := Marker2}, Trace0),
-    {[_ | _], [_ | Trace2]} = ?split_trace_at(#{?snk_kind := Marker1}, Trace1),
-    Trace2.
-
 of_kinds(Trace0, Kinds0) ->
     Kinds = sets:from_list(Kinds0, [{version, 2}]),
     lists:filter(
         fun(#{?snk_kind := K}) -> sets:is_element(K, Kinds) end,
         Trace0
     ).
-
-%% The mqtt connection span events, emitted synchronously by the test client
-%% processes.
-conn_trace(Trace) ->
-    of_kinds(Trace, [mqtt_client_connection]).
 
 %% The cache mutation events, emitted by the single emqx_crl_cache gen_server.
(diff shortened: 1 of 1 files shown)


apps/emqx/test/emqx_crl_cache_SUITE.erl
```diff
@@ -796,157 +793,66 @@ t_cache_overflow(Config) ->
...
-                        ?snk_kind := new_crl_url_inserted,
-                        url := URL3
-                    }
+                    #{?snk_span := start, client_num := 3},
+                    #{?snk_span := {complete, ok}, client_num := 3},
+                    #{?snk_span := start, client_num := 3},
+                    #{?snk_span := {complete, ok}, client_num := 3}
                 ],
-                sort_trace(Trace3)
+                conn_trace(trace_between(Trace, first_eviction, second_eviction))
```
thalesmg: I think the original intention was to see that the crl events were caused by the client connections, and which events.

checking just the client connect spans like this is a bit useless, maybe we can drop those entirely.

thalesmg (dismissed):

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15880)

PR #17708, opened 2026-06-24
fix(logger): avoid formatter crash on tuple log fields without chars_limit

zmstone:

## Summary

Some debug-level trace events were being replaced by an OTP `FORMATTER CRASH: <msg>` line, hiding the event content. This was reported on the `authenticator_result` and `authentication_result` authentication trace events — exactly the events that show which authenticator accepted or rejected a connection.

Root cause: `emqx_utils_log:format/3` pattern-matched `chars_limit` (and `depth`, `single_line`) as required config keys, so it died with `function_clause` when the formatter config did not carry `chars_limit`. In the JSON encoder, atoms/numbers/binaries/maps/lists are encoded inline; only **tuple** values fall through to `emqx_utils_log:format("~0p", [Term], Config)`. The authentication trace events carry a tuple-valued `result` field (`{ok, _}` / `{error, _}` / `{stop, _}`), so they were uniquely exposed. OTP's `logger_h_common:try_format/3` then re-emitted the event through the default formatter as `FORMATTER CRASH`.

Changes:
- `apps/emqx_utils/src/emqx_utils_log.erl`: treat `depth` / `single_line` / `chars_limit` as optional, defaulting to the same values `logger_formatter` uses (`unlimited` / `true` / `unlimited`). This is the actual crash site and a shared utility, so the fix covers all caller paths.
- `apps/emqx/src/emqx_logger_jsonfmt.erl`: `add_default_config/1` now also fills `chars_limit => unlimited` (it is the defaults helper, and was the only key it missed).
- Regression tests in `emqx_logger_fmt_SUITE` for both the text and JSON formatters.

The same defect exists on v5 (`emqx_logger_jsonfmt:do_format_msg/3`, inline) and forward; this PR targets `dev-60` and forward-syncs.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_logger_jsonfmt.erl
+++ b/apps/emqx/src/emqx_logger_jsonfmt.erl
@@ -131,7 +131,7 @@ format_msg({Fmt, Args}, _Meta, Config) ->
     emqx_utils_log:format(Fmt, Args, Config).
 
 add_default_config(Config0) ->
-    Default = #{single_line => true},
+    Default = #{single_line => true, chars_limit => unlimited},
     Depth = get_depth(maps:get(depth, Config0, undefined)),
     maps:merge(Default, Config0#{depth => Depth}).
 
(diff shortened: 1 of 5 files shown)


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15836)

PR #17710, opened 2026-06-24
fix(plugins): use fresh_install mode in CLI and add --cluster flag

JimMoen:
Phase 1: Change emqx_plugins_cli:ensure_installed/2 to use ?fresh_install
mode instead of ?normal, matching the API path behavior. This prevents
the 'failed_to_get_plugin_config_from_cluster' warning when installing
a plugin via CLI on a single node, since peer nodes don't have the
plugin config yet.

Phase 2: Add --cluster flag to 'emqx ctl plugins install' command.
When specified, reads the local tarball and distributes it to all
running nodes via erpc:multicall using the new proto v5.

- emqx_plugins_cli.erl: add ensure_installed_cluster/2, use ?fresh_install
- emqx_plugins.erl: add install_package/2 RPC target
- emqx_plugins_proto_v5.erl: new proto module for cluster install
- emqx_mgmt_cli.erl: parse --cluster flag, update help textRelease version:

## PR Checklist


hjianbo (commented):
A few issues to fix before this is ready.

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -665,6 +666,11 @@ get_config(NameVsn, ?CONFIG_FORMAT_MAP, Default) ->
 get_tar(NameVsn) ->
     emqx_plugins_fs:get_tar(NameVsn).
 
+-spec install_package(name_vsn(), binary()) -> ok | {error, term()}.
+install_package(NameVsn, Bin) ->
+    ok = write_package(NameVsn, Bin),
+    ensure_installed(NameVsn, ?fresh_install).
```
hjianbo: install_package/2 writes the tarball before calling fresh_install, but it never removes the tarball/md5 when installation fails. The existing HTTP RPC path deletes the package on failure, and without the same cleanup a bad distributed package can leave remote nodes with a stale tarball that later makes upload_install return ALREADY_INSTALLED. Please reuse the existing cleanup logic or delete the package on non-ok results.

apps/emqx_plugins/src/emqx_plugins_cli.erl
```diff
@@ -128,7 +146,27 @@ do_disallow_installation(NameVsn, LogFun) ->
     ?PRINT(Result, LogFun).
 
 ensure_installed(NameVsn, LogFun) ->
-    ?PRINT(emqx_plugins:ensure_installed(NameVsn), LogFun).
+    ?PRINT(emqx_plugins:ensure_installed(NameVsn, ?fresh_install), LogFun).
```
hjianbo: Switching the plain CLI install path unconditionally to fresh_install bypasses the guard that the fresh_install caller is expected to provide. ensure_installed/2 notes that external layers must reject already installed plugins, and the HTTP API does that before fresh_install, but this CLI path does not. Please add an already-installed check or only use fresh_install for true first-time installs.

apps/emqx_plugins/src/proto/emqx_plugins_proto_v5.erl
```diff
@@ -0,0 +1,36 @@
...
+
+    install_package/3
+]).
+
+-include_lib("emqx/include/bpapi.hrl").
+
+-define(TIMEOUT, 25_000).
+
+introduced_in() ->
+    "5.10.0".
```
hjianbo: This BPAPI is added on the dev-510/5.10.5 line, but introduced_in() says 5.10.0. Since BPAPI modules are frozen from the release declared here, this should use the actual release that introduces v5, e.g. 5.10.5.

zmstone (commented):
need relup instructions to reload ctl registration?

hjianbo (approved):

PR #17711, opened 2026-06-24
fix(authn): consistent namespace handling for built-in users HTTP API

zmstone:
Release version: 6.1.2

## Summary

The built-in database authentication user endpoints resolved the target namespace from **different sources per HTTP verb**: `POST` read it from the request body (`namespace`), while `PUT`/`DELETE`/`GET` only honored the `ns` query parameter. A user created in a namespace via the body therefore could not be found or deleted with the same body field — it appeared to still authenticate after a "successful" delete, and the dashboard listing (which also uses `ns`) showed nothing. This is the root cause of a customer report where a deleted built-in user kept authenticating.

Namespace resolution is now unified in the operation `filter/2`, keyed off the HTTP method:

- `POST` / `PUT` accept the namespace from the request body (`namespace`) **or** the `ns` query parameter — **body wins**.
- `DELETE` accepts the namespace from the `ns` query parameter **or** the request body — **query parameter wins**.
- `GET` / list keep the `ns` query parameter (a GET body has no defined semantics).

Actor authorization is preserved: a namespace-scoped admin may still only operate on its own namespace; only a global admin may target an arbitrary namespace.

Separately, a global admin can now delete a built-in user whose namespace has already been deleted, instead of getting `Managed namespace not found` — so orphaned records can be cleaned up.

### Most relevant changes

- `apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl`: per-method namespace resolution (`requested_namespace/2`), `namespace` added to the update body schema and a small `request_user_delete` body schema (optional `namespace`) so the DELETE body survives validation, and the managed-namespace existence check is skipped for global-admin deletes.
- `apps/emqx_auth_mnesia/test/emqx_authn_api_mnesia_SUITE.erl`: regression tests for body/query precedence on delete, body namespace on update, and deleting an orphaned-namespace record; one obsolete assertion (rejecting `namespace` in an update body) was updated to reflect the new behavior.

> Note: the full CT suite could not be executed in my local environment (the fresh worktree cannot build unrelated enterprise native deps — `aws-lc-sys`/`grpc_plugin` — with the available toolchain). The changed module compiles cleanly and `make fmt`/elvis pass; CI will run the suite.

## PR Checklist

### DIFF
--- a/apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
+++ b/apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
@@ -1785,35 +1785,27 @@ user_out(User) ->
 get_namespace(#{resolved_ns := Namespace}) ->
     Namespace.
 
-parse_namespace(Req, Method) ->
+parse_namespace(Req) ->
     ActorNamespace = emqx_dashboard:get_namespace(Req),
-    case requested_namespace(Method, Req) of
+    case requested_namespace(Req) of
         undefined ->
             {ok, ActorNamespace};
         ActorNamespace ->
             {ok, ActorNamespace};
-        _Other when ActorNamespace =:= ?global_ns ->
+        Other when ActorNamespace =:= ?global_ns ->
             %% Only a global admin may operate on an arbitrary namespace.
-            {ok, _Other};
+            {ok, Other};
         _Other ->
             {error, not_authorized}
     end.
 
-%% Decide where the requested namespace comes from, per HTTP method:
-%%   * POST / PUT carry a body, so the `namespace' body field takes precedence,
-%%     falling back to the `ns' query parameter.
-%%   * DELETE may carry a body too, but the `ns' query parameter wins; the body
-%%     `namespace' is only a fallback.
-%%   * GET (and anything else) only reads the `ns' query parameter, as a GET body
-%%     has no defined semantics.
-%% Returns `undefined' when no namespace was requested, so the caller can default
-%% to the actor's namespace.
-requested_namespace(Method, Req) when Method =:= post; Method =:= put ->
-    first_defined([body_namespace(Req), qs_namespace(Req)]);
-requested_namespace(delete, Req) ->
-    first_defined([qs_namespace(Req), body_namespace(Req)]);
-requested_namespace(_Method, Req) ->
-    qs_namespace(Req).
+%% The requested namespace may be carried in the `ns' query parameter or in the
+%% `namespace' request body field. The query parameter always takes precedence;
+%% the body field is only a fallback (and is the only option for GET, which has
+%% no body). Returns `undefined' when no namespace was requested, so the caller
+%% can default to the actor's namespace.
+requested_namespace(Req) ->
+    first_defined([qs_namespace(Req), body_namespace(Req)]).
 
 qs_namespace(#{query_string := QueryString}) ->
     maps:get(<<"ns">>, QueryString, undefined);
@@ -1829,8 +1821,8 @@ first_defined([undefined | Rest]) -> first_defined(Rest);
 first_defined([Value | _Rest]) -> Value;
 first_defined([]) -> undefined.
 
-resolve_namespace(Req, #{method := Method}) ->
-    case parse_namespace(Req, Method) of
+resolve_namespace(Req, _Meta) ->
+    case parse_namespace(Req) of
         {ok, Namespace} ->
             {ok, Req#{resolved_ns => Namespace}};
         {error, not_authorized} ->
--- a/apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
+++ b/apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
@@ -125,6 +125,7 @@ roots() ->
     [
         request_user_create,
         request_user_update,
+        request_user_delete,
         response_user,
         response_users,
         request_authn_order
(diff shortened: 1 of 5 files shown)


apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
```diff
@@ -1787,17 +1785,52 @@ user_out(User) ->
...
-        QSNamespace ->
-            {ok, QSNamespace}
+    case requested_namespace(Method, Req) of
+        undefined ->
+            {ok, ActorNamespace};
+        ActorNamespace ->
+            {ok, ActorNamespace};
+        _Other when ActorNamespace =:= ?global_ns ->
+            %% Only a global admin may operate on an arbitrary namespace.
+            {ok, _Other};
```
thalesmg: nit:


```suggestion
        Other when ActorNamespace =:= ?global_ns ->
            %% Only a global admin may operate on an arbitrary namespace.
            {ok, Other};
```

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
```diff
@@ -1787,17 +1785,52 @@ user_out(User) ->
...
+%%   * DELETE may carry a body too, but the `ns' query parameter wins; the body
+%%     `namespace' is only a fallback.
+%%   * GET (and anything else) only reads the `ns' query parameter, as a GET body
+%%     has no defined semantics.
+%% Returns `undefined' when no namespace was requested, so the caller can default
+%% to the actor's namespace.
+requested_namespace(Method, Req) when Method =:= post; Method =:= put ->
+    first_defined([body_namespace(Req), qs_namespace(Req)]);
+requested_namespace(delete, Req) ->
+    first_defined([qs_namespace(Req), body_namespace(Req)]);
```
thalesmg: Q: what if both are defined but different? Shouldn't be rejected?
allowing both to diverge sounds like a source of confusion.

apps/emqx_auth_mnesia/test/emqx_authn_api_mnesia_SUITE.erl
```diff
@@ -614,6 +635,60 @@ t_authenticator_user(TCConfig) ->
...
+    {200, _} = create_authenticator(emqx_authn_test_lib:built_in_database_example()),
+    Ns = ?OTHER_NS,
+    User = #{<<"user_id">> => <<"u1">>, <<"password">> => <<"p1">>, <<"namespace">> => Ns},
+    ?assertMatch({201, _}, add_user(User)),
+    ?assertMatch(
+        {200, _},
+        update_user(<<"u1">>, #{<<"password">> => <<"p2">>, <<"namespace">> => Ns}, #{})
+    ),
+    ?assertMatch({204, _}, delete_user(<<"u1">>, #{<<"ns">> => Ns})),
+    ok.
```
thalesmg, line 727: maybe add a clause targeting another ns in the body to see it 404?

changes/ee/fix-17711.en.md
```diff
@@ -0,0 +1,11 @@
+Made namespace handling consistent across the built-in database authentication user HTTP APIs, and allowed cleanup of orphaned records.
+
+Previously, creating a user accepted the target namespace from the request body, while updating, deleting and listing users only honored the `ns` query parameter. As a result, a user created in a namespace via the request body could not be found or deleted with the same body field, appearing as if it still authenticated after deletion.
```
thalesmg: > ... deleting and listing users only honored the `ns` query parameter. As a result, a user created in a namespace via the request body could not ...

`namespace` was not part of the body schema, so this isn't actually a fix.  just a different choice of presenting the api.  sending the namespace in the body for those requests would be an user error.  😅

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
zmstone: decision is to choose qs over body.

thalesmg: yeah, for deletes; and the opposite for other requests.  which adds even more to the confusion...  :sweat:

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
```diff
@@ -1785,35 +1785,27 @@ user_out(User) ->
...
-    first_defined([qs_namespace(Req), body_namespace(Req)]);
-requested_namespace(_Method, Req) ->
-    qs_namespace(Req).
+%% The requested namespace may be carried in the `ns' query parameter or in the
+%% `namespace' request body field. The query parameter always takes precedence;
+%% the body field is only a fallback (and is the only option for GET, which has
+%% no body). Returns `undefined' when no namespace was requested, so the caller
+%% can default to the actor's namespace.
+requested_namespace(Req) ->
+    first_defined([qs_namespace(Req), body_namespace(Req)]).
```
thalesmg: that's not quite what I meant in the original comment.  😅 

it was not about which source is preferred, but about allowing silent inconsistencies when both are defined.

qs may be ns1, body may be ns2.  one of them will be used, the other silently ignored. 

my suggestion was to forbid both to be defined but different.

changes/ee/fix-17711.en.md
zmstone: split the changelog into two parts.

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
zmstone: changed to assert equal if both present.

apps/emqx_auth_mnesia/test/emqx_authn_api_mnesia_SUITE.erl
zmstone: done in t_create_user_unknown_namespace_rejected

apps/emqx_auth/src/emqx_authn/emqx_authn_api.erl
zmstone: changed to reject with 400 if both are provided but not equal

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15859)

PR #17157, opened 2026-04-28
feat(rule engine): add `limit_selects_in_namespace` config to restrict rule execution

thalesmg:
Fixes https://emqx.atlassian.net/browse/EMQX-14809

Release version:
6.0.3, 6.1.2, 6.2.0

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17725, opened 2026-06-25
fix(rule): keep global rules matching namespaced traffic

zmstone:

## Summary

Follow-up to the `rule_engine.limit_selects_in_namespace` feature (#17157), which defaults
to `true`. With it enabled, `emqx_rule_events:restrict_rules_to_namespace/2` kept a rule
only when its namespace **equalled** the publishing message's namespace. A *global* rule
(`namespace = global`) therefore stopped matching messages from clients carrying a tenant
namespace (`client_attrs.tns`) — the regression a customer hit on a 6.0 → 6.2.x upgrade
(a metrics-only global rule silently stopped matching).

This changes the restriction to be asymmetric, matching the feature's real intent:

- **Global rules retain system-wide visibility** and match messages from any namespace
  (restores pre-6.2 behavior; the non-multi-tenant case is unaffected).
- **Namespaced rules stay isolated** to their own namespace (tenant-to-tenant isolation,
  the actual security property, is preserved — `t_limit_selects_in_namespace` still passes).
- A rule that somehow carries no namespace association is treated as global rather than
  being silently dropped.

### Why tests/QA didn't catch the original regression

Existing rule tests create rules without an explicit namespace, but the shared test helper
(`emqx_rule_engine_SUITE:create_rule/1`) stamps `namespace => global`, and all test
publishers are global (no `tns`). So the gate was always `global == global` — a no-op. The
only test exercising the gate with `true` (`t_limit_selects_in_namespace`) uses two
*namespaced* rules; no test covered a global rule against namespaced traffic.

### Files

- `apps/emqx_rule_engine/src/emqx_rule_events.erl` — asymmetric filter in
  `restrict_rules_to_namespace/2`; exported under `TEST`.
- `apps/emqx_rule_engine/test/emqx_rule_events_SUITE.erl` — unit test for the filter matrix
  (global / namespaced / namespace-less rules × global / namespaced messages).
- `apps/emqx_rule_engine/test/emqx_rule_engine_api_2_SUITE.erl` —
  `t_global_rule_matches_namespaced_traffic`: end-to-end regression (metrics-only global
  rule + a `tns`-carrying MQTT v3 publisher, gate on). Verified red before the fix, green
  after.

Operator workaround (still available): `rule_engine.limit_selects_in_namespace = false`.

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/src/emqx_rule_events.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_events.erl
@@ -55,7 +55,8 @@
 -ifdef(TEST).
 -export([
     reason/1,
-    hook_fun/1
+    hook_fun/1,
+    restrict_rules_to_namespace/2
 ]).
 -endif.
 
@@ -1678,7 +1679,13 @@ restrict_rules_to_namespace(EnrichedRules0, Message) ->
     Ns = get_namespace_from_message(Message),
     lists:filter(
         fun
+            (#{rule := #{namespace := ?global_ns}}) ->
+                %% Global rules retain system-wide visibility: they keep matching
+                %% messages from any namespace, preserving pre-6.2 behavior for
+                %% non-multi-tenant deployments.
+                true;
             (#{rule := #{namespace := Ns0}}) ->
+                %% Namespaced rules only match traffic from their own namespace.
                 Ns0 == Ns;
             (_) ->
                 false
--- a/apps/emqx_rule_engine/test/emqx_rule_engine_api_2_SUITE.erl
+++ b/apps/emqx_rule_engine/test/emqx_rule_engine_api_2_SUITE.erl
@@ -3026,3 +3026,82 @@ t_limit_selects_in_namespace(TCConfig0) when is_list(TCConfig0) ->
     emqtt:stop(C1),
     emqtt:stop(C2),
     ok.
+
+-doc """
+Regression test: with `rule_engine.limit_selects_in_namespace = true` (the default), a
+*global* rule must keep matching messages from clients that carry a tenant namespace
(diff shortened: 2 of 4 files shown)


zmstone:
~~the fix is only for a branch which is impossible to reach.~~

apps/emqx_rule_engine/src/emqx_rule_events.erl
```diff
@@ -1678,8 +1679,18 @@ restrict_rules_to_namespace(EnrichedRules0, Message) ->
...
+                %% messages from any namespace, preserving pre-6.2 behavior for
+                %% non-multi-tenant deployments.
+                true;
             (#{rule := #{namespace := Ns0}}) ->
+                %% Namespaced rules only match traffic from their own namespace.
                 Ns0 == Ns;
+            (#{rule := _}) ->
+                %% A rule that somehow carries no namespace association is treated as
+                %% global, so it is never silently dropped.
+                true;
```
thalesmg: this case should be impossible, especially since we initialize the ets table with `emqx_rule_engine:load_rules/0`, which always populates the namespace key.

allowing a corrupt rule to see all namespaces is not a hardened security choice, IMO, since it opens up a way to pry all namespaces, if ever there is such a bug/vulnerability that allows creating such broken rules.  by treating them as "doesn't match", we would take a more conservative instance and avoid this leakage.

zmstone: agree. will revert this part.

thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15884)

Issue #17730, opened 2026-06-25
WS/WSS listener option update rebinds the port even when bind is unchanged (transient eaddrinuse)


zmstone:
## Summary

Updating a WS/WSS listener's transport options unnecessarily **rebinds the listen socket on the same port**, even when the bind address/port did not change (e.g. rotating TLS certificates or changing `verify`). This introduces a transient `eaddrinuse` race and is unnecessary churn.

In `emqx_listeners:do_update_listener/4`, the cowboy (ws/wss) branch does:

```erlang
ok = ranch:suspend_listener(Id),          %% closes the listen socket
ok = ranch:set_transport_options(Id, RanchOpts),
...
ranch:resume_listener(Id)                  %% rebinds the SAME port
```

`resume_listener` reopens the listen socket; if the OS has not released the old one yet, the rebind fails with `eaddrinuse`, failing the config update and dropping the listener. By contrast, TCP/SSL listeners (esockd) update options in place via `esockd:reset_options/2` with no rebind.

## Mitigation in place

#17729 retries `resume_listener` on transient `eaddrinuse`. That removes the flake but keeps the unnecessary rebind.

## Proposed proper fix

Only `suspend`/`resume` (rebind) when the **bind address/port** actually changes. For changes that don't affect the listen socket (TLS cert/key/verify, protocol options), apply `ranch:set_transport_options` + `ranch:set_protocol_options` **without** suspending/resuming, so the listen socket is never rebound. This eliminates the race entirely for the common cert-rotation case and avoids dropping acceptor capacity during the update.

Needs verification that updated TLS options are picked up by new handshakes without a resume in the ranch version in use (`emqx/ranch`).

## Repro / context

- CI flake: `emqx_listeners_SUITE:t_wss_update_opts` — e.g. https://github.com/emqx/emqx/actions/runs/28154468551/job/83381256857
- This is distinct from the `select_free_port` hand-off race fixed in #17694 — here the port is one the listener already owns.

PR #17731, opened 2026-06-25
fix(listeners): retry ws/wss listener rebind on transient eaddrinuse

zmstone:
Release version: 6.0.4

## Summary

Port of #17729 to the 6.0 line.

Updating a WS/WSS listener's transport options (e.g. rotating TLS certificates, changing `verify`) goes through `ranch:suspend_listener` → `ranch:set_transport_options` → `ranch:resume_listener`, where `resume_listener` **rebinds the same port**. If the OS has not released the old listen socket yet, the rebind fails with a transient `eaddrinuse`, which fails the config update and leaves the listener down.

`resume_listener` is now retried a few times with a short backoff on `eaddrinuse`. TCP/SSL listeners are unaffected — they update options in place (`esockd:reset_options`) without rebinding.

The proper fix — avoid rebinding entirely when the bind address/port is unchanged — is tracked in #17730.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_listeners.erl
+++ b/apps/emqx/src/emqx_listeners.erl
@@ -517,7 +517,7 @@ do_update_listener(Type, Name, OldConf, NewConf) when
     end,
     ok = ranch:set_protocol_options(Id, WsOpts),
     %% No-op if the listener was not suspended.
-    ranch:resume_listener(Id);
+    resume_cowboy_listener(Id);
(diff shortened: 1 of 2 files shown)


ieQu1 (approved):

hjianbo (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15881)

PR #17732, opened 2026-06-25
feat(api_keys): accept namespace as a top-level request body field

### LINKED ISSUE #17726: Accept namespace field when creating API keys
## Requirement

When creating an API key via `POST /api/v5/api_key`, allow callers to provide the namespace as a standalone request body field, for example:

```json
{"role":"administrator", "namespace":"ns1"}
```

Currently, namespace selection is only supported when embedded in the role string:

```json
{"role":"ns:ns1::administrator"}
```

## Current behavior

The API key creation schema/handler reads `role` from the request body but does not read a standalone `namespace` field. As a result, a request such as `{"role":"administrator", "namespace":"ns1"}` creates a global API key and returns `"namespace": null`.

Relevant code paths observed in the 6.3-prep line:

- `apps/emqx_management/src/emqx_mgmt_api_api_keys.erl` uses `fields(app)` as the POST body and reads `role` before calling `emqx_mgmt_auth:create(...)`.
- `apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl` parses namespace from role strings in the form `ns:<namespace>::<role>`.
- `apps/emqx_management/src/emqx_mgmt_auth.erl` stores the parsed namespace in API key `extra` and formats it back in responses.
- `apps/emqx_dashboard/test/emqx_dashboard_admin_SUITE.erl` has coverage for creating an API key with `role = ns:ns1::administrator` and receiving `"namespace" := "ns1"`.

## Expected behavior

`POST /api/v5/api_key` should accept `namespace` as an explicit request body field, so API clients do not need to encode namespace into the `role` string manually.

Possible behavior:

- Accept `namespace` together with a plain role, e.g. `{"role":"administrator", "namespace":"ns1"}`.
- Preserve backward compatibility with the existing `ns:<namespace>::<role>` role format.
- Define validation behavior if both forms are provided and disagree.
- Return the effective namespace in the API response.

## Motivation

This is easier for API consumers and avoids exposing the internal role-string namespace encoding as the only way to create namespaced API keys.

zmstone:
Release version: 6.0.4

## Summary

Accept a standalone `namespace` field in the `POST /api_key` and `PUT /api_key/:name` request bodies, as an alternative to encoding the namespace inside the `role` string (`ns:<namespace>::<role>`).

When both forms are supplied they must agree: a bare role plus a namespace field is assembled internally; a matching role-encoded namespace is accepted idempotently; a mismatch is rejected with 400.

Main change: `apps/emqx_management/src/emqx_mgmt_api_api_keys.erl` (`resolve_effective_role/2` and the new `namespace` request-body field).

Closes #17726

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
@@ -9,6 +9,7 @@
 -include_lib("hocon/include/hoconsc.hrl").
 -include_lib("emqx_dashboard/include/emqx_dashboard_rbac.hrl").
 -include_lib("emqx_utils/include/emqx_api_key_scopes.hrl").
+-include_lib("emqx/include/emqx_config.hrl").
 
 -export([api_spec/0, fields/1, paths/0, schema/1, namespace/0]).
 -export([api_key/2, api_key_by_name/2, api_key_scopes/2]).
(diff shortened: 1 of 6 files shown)


apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
```diff
@@ -404,6 +423,44 @@ effective_request_scopes(Name, undefined) ->
...
+%%   * both, encoded X == field X            -> role unchanged (idempotent)
+%%   * both, encoded X /= field Y            -> {error, mismatch}
+resolve_effective_role(Role, undefined) ->
+    {ok, Role};
+resolve_effective_role(_Role, <<>>) ->
+    {error, <<"namespace must not be empty">>};
+resolve_effective_role(Role, Namespace) when is_binary(Namespace) ->
+    case emqx_dashboard_rbac:parse_api_role(Role) of
+        {ok, #{?role := BareRole, ?namespace := ?global_ns}} ->
+            {ok, <<"ns:", Namespace/binary, "::", BareRole/binary>>};
```
thalesmg:
```suggestion
            {ok, emqx_dashboard_rbac:serialize_role(#{?role => BareRole, ?namespace => Namespace})};
```

thalesmg (approved):

PR #17753, opened 2026-06-26
docs: add missing 6.1.3 changelog entry for #17497

zmstone:
PR #17497 (fix `actions.executed` undercounting in non-batch mode) was backported to release-61 in code, but its changelog entry was consolidated into the 6.2-only `feat-17558` entry and therefore dropped from the 6.1.3 release notes. This adds the entry back under 6.1.3 → Observability so the 6.1 line documents the fix it actually ships.

### DIFF
--- a/changes/6.1.3.en.md
+++ b/changes/6.1.3.en.md
@@ -141,6 +141,12 @@
 
 ### Observability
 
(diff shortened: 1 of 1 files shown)


thalesmg (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15903)

PR #17755, opened 2026-06-26
feat(metrics): increment messages.retained counter

### LINKED ISSUE #17385: [retainer] Expose retained message write counter as a metric

zmstone:
Fixes #17385.

Release version: 6.3.0

## Summary

Port of #17675 (targets master) to dev-60. Original author: Sahil Patel.

EMQX already exposed `emqx_messages_retained` in Prometheus but the metric was never registered as a counter and never incremented — it always showed 0. This patch:

1. Registers `messages.retained` as a counter in `emqx_metrics.erl` (using reserved index 412 on dev-60; master's port uses 105 because the index space is different).
2. Increments the counter via `emqx_metrics:inc/1` after a successful retained-message store (master uses `inc_global/1`; that API doesn't exist on the v6.0 line, so renamed to `inc/1`, which is the equivalent).
3. Adds test assertions in `t_store_and_clean` to verify the counter increments on publish and stays unchanged on delete.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_metrics.erl
+++ b/apps/emqx/src/emqx_metrics.erl
@@ -515,6 +515,7 @@ reserved_idx('messages.transformation_succeeded') -> 408;
 reserved_idx('messages.transformation_failed') -> 409;
 reserved_idx('rules.matched') -> 410;
 reserved_idx('actions.executed') -> 411;
+reserved_idx('messages.retained') -> 412;
 reserved_idx(_) -> undefined.
 
 all_metrics() ->
@@ -609,7 +610,8 @@ message_metrics() ->
         {counter, 'messages.validation_succeeded', ?DESC("messages_validation_succeeded")},
         {counter, 'messages.transformation_failed', ?DESC("messages_transformation_failed")},
         {counter, 'messages.transformation_succeeded', ?DESC("messages_transformation_succeeded")},
-        {counter, 'messages.persisted', ?DESC("messages_persisted")}
+        {counter, 'messages.persisted', ?DESC("messages_persisted")},
+        {counter, 'messages.retained', ?DESC("messages_retained")}
     ].
 
 %% Rule and Action related metrics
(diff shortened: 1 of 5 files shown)


zmstone:
Retargeting to dev-62 instead of dev-60 — superseded by #17758.

PR #17764, opened 2026-06-29
fix(plugins): purge stale plugins after cluster rejoin

hjianbo:
Release version:
5.10.5

Fix: https://emqx.atlassian.net/browse/EMQX-14124

## Summary

Fixed stale plugin package state when a node rejoins after missing a cluster-wide plugin uninstall. A node that left before the uninstall could keep local plugin directories; after rejoin, plugin list APIs could still surface that local package while config APIs returned `NOT_FOUND` because the cluster plugin config no longer contained it.

The fix reconciles local plugin packages during `emqx_plugins` startup after configured plugins are installed: packages absent from the current cluster plugin config are purged locally, while other versions of a still-configured plugin app are kept so the force-sync/version-switch path remains available.

The 5.10.4 -> 5.10.5 relup catalog now also loads the plugin management modules and runs the stale-package cleanup during hot upgrade, with catalog README and CT coverage.

Added unit coverage for the purge rules and a cluster regression test for leave, uninstall, rejoin, and cleanup. Also bumped the `emqx_plugins` app version.

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ce/fix-17764.en.md`


apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -862,6 +880,77 @@ put_configured(Configured, ConfLocation) ->
 configured() ->
     get_config_internal(states, []).
 
+configured_name_vsns() ->
+    [bin(NameVsn) || Item <- configured(), NameVsn <- configured_name_vsn(Item)].
+
+configured_name_vsn(#{name_vsn := NameVsn}) ->
```
savonarola: Nit: do we actually need to introduce fuzziness, i.e. ability to work with both atoms and bins?

My idea is that we better avoid adding defensive cases, because the more we add them, the worse is the local understanding of the shape of the data we are working with, and the more defensive cases we have to add further.

zmstone:
What happens if a plugin package is staged (uploaded), but not installed, then the node restart, will the package be deleted ?

> packages absent from the current cluster plugin config are purged locally, while other versions of a still-configured plugin app are kept so the force-sync/version-switch path remains available.

This makes impossible to install a plugin only in one of the nodes in the cluster ?
Maybe we simply fix the list API **not** to return stale plugin.

hjianbo:
@zmstone I checked the current dev-510 paths again.

For a package that is only staged as a tarball and not unpacked/installed, this PR should not delete it. `purge_unconfigured/0` iterates `emqx_plugins_fs:list_name_vsn/0`, which is based on `install_dir/*/release.json`, so tar-only packages are outside this cleanup.

For the "install a plugin only on one node" case, I think we need to distinguish local package/runtime operations from cluster-managed plugin state.

With `emqx ctl plugins install NameVsn`, the package installation itself only runs on the node where the command is executed, so the package files may temporarily exist only on that node. The install flow also records the plugin in `plugins.states` with `enable=false`.

If the operator then runs `emqx ctl plugins start NameVsn`, that starts the plugin only on the current node. It does not change the cluster config to `enable=true`. So yes, `install + start` can temporarily make the package and runtime exist only on one node.

However, that is not a persistent cluster-managed runtime state. When the node restarts, or when a new node joins the cluster, the `emqx_plugins` app starts as part of the normal EMQX application startup/restart flow. During `emqx_plugins` startup, it reads `plugins.states` from cluster config: configured packages may be installed/copied from peers by `ensure_installed()`, but only entries with `enable=true` are auto-started by `ensure_started()`. Since this CLI `install + start` flow leaves the cluster config as `enable=false`, the one-node runtime state is not restored as a cluster state.

The state becomes cluster-managed runtime state only when the operator uses `emqx ctl plugins enable NameVsn`. That updates `plugins.states` to `enable=true` in cluster config. Other running nodes then see the enable transition and call `ensure_started/1`; that path first calls `install(NameVsn, ?normal)`, so nodes without the package will try to copy it from a running peer before starting it.

I also checked this cleanup against the temporary one-node state:

- On the node where CLI install ran, the `NameVsn` is already in `plugins.states`, so `purge_unconfigured/0` keeps it.
- On other nodes that do not have package files yet, there is nothing for this cleanup to remove, because it scans extracted plugin dirs via `install_dir/*/release.json`.
- On node restart or new node join, `emqx_plugins_app:start/2` runs `ensure_installed()` before `purge_unconfigured()`, so configured plugins get a chance to be installed/copied before cleanup. They are then kept because they are configured.

So I don't think this PR breaks the current CLI install/start-on-one-node behavior. It only removes extracted local packages that are absent from cluster config, which is the stale state produced when a node misses a cluster-wide uninstall while it is out of the cluster.

I don't think only fixing the list API is enough, because that would hide the stale package from list output but leave it on disk. It could still affect `describe`, `is_package_present`, and future install/delete/sync paths. If we want to support an intentional persistent local-only plugin mode, we probably need an explicit local-only state/marker; otherwise there is no reliable way to distinguish it from a stale package left behind after cluster-wide uninstall.

zmstone (approved):

PR #17773, opened 2026-06-29
fix(emqx_conf): normalize cluster_rpc abort reasons to {error, _}

zmstone:
Release version: 6.0.4

## Summary

Port of #17770 (dev-58) to dev-60. The bug and the test-isolation gap both exist on dev-60.

### Fix — `emqx_cluster_rpc` failure contract

`emqx_cluster_rpc:multicall/3` is documented to return `{ok, _}` on success and `{error, _}` on failure, and `emqx_conf:update/3` (plus the `emqx_resource`, `emqx_mgmt_api_plugins` and `emqx_rule_engine` config protos) is typed on that contract. But the `init_failure` path in `do_initiate/4` returned the raw mnesia abort reason unwrapped — e.g. `{no_exists, cluster_rpc_mfa}` when the cluster_rpc tables are not yet available during cold start or recovery. That bare tuple then escaped to callers, crashing the request handler with `function_clause` (observed via `emqx_gateway_conf:res/1` in a gateway config teardown).

`wrap_abort/1` normalizes the transaction-abort path: a bare reason becomes `{error, Reason}`, while an already-structured `{error, _}` (a failed MFA result) is left untouched so it is never double-wrapped. `t_commit_fail_test` is updated to expect the normalized shape.

### Test isolation — `emqx_gateway_metrics_SUITE`

`emqx_gateway_metrics:start_link/1` does not clear the broker-wide named ETS table `emqx_gateway_mqttsn_metrics` (`emqx_utils_ets:new/2` is idempotent), so real broker stats from a prior gateway suite in the same CT run could leak into `t_inc_dec`. `init_per_testcase` / `end_per_testcase` now wipe the table so each case starts and leaves a clean slate.

## Differences from #17770

- `emqx_conf.app.src` `vsn`: dev-60 was already at `0.4.9`, bumped to `0.4.10` (vs `0.4.8` → `0.4.9` on dev-58).
- `apps/emqx_conf/mix.exs` `version`: bumped from `6.0.3` → `6.0.4` by the apps-version-check auto-fix (dev-58 doesn't have a `mix.exs` version bump because it's a separate v5 line).
- Changelog renamed `fix-17770.en.md` → `fix-17773.en.md` to match this PR's number.

## PR Checklist

### DIFF
--- a/apps/emqx_conf/mix.exs
+++ b/apps/emqx_conf/mix.exs
@@ -5,7 +5,7 @@ defmodule EMQXConf.MixProject do
   def project do
     [
       app: :emqx_conf,
-      version: "6.0.3",
+      version: "6.0.4",
       build_path: "../../_build",
       # config_path: "../../config/config.exs",
       erlc_options: UMP.erlc_options(),
--- a/apps/emqx_conf/src/emqx_cluster_rpc.erl
+++ b/apps/emqx_conf/src/emqx_cluster_rpc.erl
@@ -505,8 +505,19 @@ do_initiate(MFA, State = #{node := Node}, Count, Failure0) ->
             catch_up(State),
             do_initiate(MFA, State, Count + 1, Failure1);
         {aborted, Error} ->
-            {reply, {init_failure, Error}, State, {continue, ?CATCH_UP}}
+            {reply, {init_failure, wrap_abort(Error)}, State, {continue, ?CATCH_UP}}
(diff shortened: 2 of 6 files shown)


savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15918)

PR #17787, opened 2026-06-30
fix(http_connector): drop request headers from worker-down error reason

zmstone:
Release version: 6.0.4

## Summary

Fixes emqx/emqx-dev-team-tasks#273.

When the HTTP connector's `ehttpc` worker is killed while a request is in flight — for example, by deleting the source while the request has not yet returned — the `gen_server:call` to the dead worker returns an EXIT whose reason carries the original call arguments. Those arguments include the request headers (`Authorization`, `x-api-key`, etc.). The reason then flowed verbatim into several error logs: `http_connector_do_request_failed`, `unrecoverable_resource_error`, and `http_server_query_failed`.

The fix drops the call arguments at the source, in `emqx_bridge_http_connector:transform_result/1`: a new clause matches the `{ehttpc_worker_down, _}` reason and replaces the captured `gen_server:call` args with `'...'`, keeping the worker-down classification and stop reason for diagnostics. Any unrecognized sub-shape falls back to a bare `ehttpc_worker_down` atom. Sanitizing the reason once at this boundary keeps every downstream log consumer safe without logging any header bytes and without relying on per-field redaction.

Relevant files:
- `apps/emqx_bridge_http/src/emqx_bridge_http_connector.erl` — new `transform_result/1` clause and `drop_ehttpc_worker_down_call_args/1` helper.
- `apps/emqx_bridge_http/test/emqx_bridge_http_connector_tests.erl` — unit tests asserting no header bytes survive, including the `{shutdown, ...}`-wrapped variant.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_http/mix.exs
+++ b/apps/emqx_bridge_http/mix.exs
@@ -5,7 +5,7 @@ defmodule EMQXBridgeHTTP.MixProject do
   def project do
     [
       app: :emqx_bridge_http,
-      version: "6.0.3",
+      version: "6.0.4",
       build_path: "../../_build",
       # config_path: "../../config/config.exs",
       erlc_options: UMP.erlc_options(),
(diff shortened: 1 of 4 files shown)


savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15971)

PR #17789, opened 2026-06-30
[v6] chore: add a script to list release versions

zmstone:
Release version: 6.0.4

## Summary

Companion to #17788 (v5). Adds `scripts/rel-versions` for the v6 line.

Given a checkout, the script prints, one per line, the next release version on the current branch's release line and every downstream branch it would forward-sync to.

Example: on a v6.0 topic branch (last reachable GA tag `6.0.3`):

    6.0.4
    6.1.3
    6.2.2

If HEAD is descended from `6.1.2`, the v6.0 line is dropped from the output (`6.1.3`, `6.2.2`). dev-63 currently has no GA tag yet, so the script emits a warning and skips it; once `6.3.0` lands the entry appears automatically.

- Auto-detects the canonical `emqx/emqx` git remote (parses `git remote -v`; doesn't assume `origin`).
- Runs `git fetch <remote> --tags --quiet` before computing so the answers reflect upstream right now.
- Excludes pre-release tags (`-rc.N`, `-alpha`, `-beta`, `-patch.N`) via `--exclude '*-*'`; the version is "next patch after the most recent GA".
- v6 tags use no `e` prefix (`6.0.3`, `6.1.2`, ...); the v5 sibling script in #17788 keeps the `e` prefix.

`shellcheck -S style` clean.

## PR Checklist

### DIFF
--- a/scripts/rel-versions
+++ b/scripts/rel-versions
@@ -1,0 +1,146 @@
+#!/usr/bin/env bash
+# Print the next-to-be-released EMQX version on the current branch and every
+# downstream branch in the forward-sync chain.
+#
+# Example: if HEAD is on a v6.0 topic branch (last reachable GA tag is 6.0.3),
+# this prints
+#
+#     6.0.4
+#     6.1.3
+#     6.2.2
+#
+# computed by running `git describe <remote>/dev-XX --tags --match 'X.Y.*'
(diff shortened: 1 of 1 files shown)


keynslug (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15972)

PR #17739, opened 2026-06-25
fix(redact): address sensitive data leaks

savonarola:
Release version:
6.0.3

## Summary

Addressed sensitive data redaction issues in logs, traces, and audit records.

Resolved:
- https://github.com/emqx/emqx-dev-team-tasks/issues/77
- https://github.com/emqx/emqx-dev-team-tasks/issues/178
- https://github.com/emqx/emqx-dev-team-tasks/issues/216
- https://github.com/emqx/emqx-dev-team-tasks/issues/218
- https://github.com/emqx/emqx-dev-team-tasks/issues/222
- https://github.com/emqx/emqx-dev-team-tasks/issues/224
- https://github.com/emqx/emqx-dev-team-tasks/issues/225
- https://github.com/emqx/emqx-dev-team-tasks/issues/229
- https://github.com/emqx/emqx-dev-team-tasks/issues/232
- https://github.com/emqx/emqx-dev-team-tasks/issues/236
- https://github.com/emqx/emqx-dev-team-tasks/issues/250

## PR Checklist
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17791, opened 2026-06-30
fix(redact): redact jose_jwk records in logs

zmstone:

## Summary

`emqx_utils:redact/1` walked `jose_jwk` records element by element. Because the inner `jose_jwk_kty_oct` tuple was treated as a plain `{Key, Value}` pair whose key is not in the sensitive-key list, the raw HMAC key bytes survived redaction and could appear in `cluster_rpc_apply_result` / `cluster_rpc_apply_ok` debug log lines that dump the JWT authenticator runtime state during config updates.

Changes in `apps/emqx_utils/src/emqx_utils_redact.erl`:

* Add a `do_redact/2` head clause matching the `jose_jwk` record shape, placed before the generic tuple walk, so the whole record is replaced with a `{jose_jwk, '******'}` placeholder before recursion can descend into the key material.
* Mark the `jwk` map key as sensitive (atom / string / binary forms), so non-record JWK values (e.g. a plain map fetched from a JWKS endpoint) are also redacted wholesale.

Tests:

* EUnit cases in `emqx_utils_redact` covering the record shape, the map-key shape, and deeply-nested state maps — asserting the raw bytes never appear in the redacted output.
* CT case `t_redact_hmac_state` in `emqx_authn_jwt_SUITE` that builds a real HMAC authenticator state and asserts `emqx_utils:redact/1` removes the secret.

## PR Checklist

### DIFF
--- a/apps/emqx_auth_jwt/test/emqx_authn_jwt_SUITE.erl
+++ b/apps/emqx_auth_jwt/test/emqx_authn_jwt_SUITE.erl
@@ -275,6 +275,36 @@ t_invalid_signature_trace_keeps_public_jwks(_) ->
     ?assertEqual(ok, emqx_authn_jwt:destroy(State)),
     ok.
 
+%% The HMAC authenticator runtime state carries a `jose_jwk' record holding the
+%% raw key bytes; `emqx_utils:redact/1' (used by config-update / cluster_rpc log
+%% lines that dump such state) must replace it so the secret never reaches a log.
+t_redact_hmac_state(_) ->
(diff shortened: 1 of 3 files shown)


savonarola:
BTW: this was also addressed in https://github.com/emqx/emqx/pull/17739/changes/f8520f79fccc3784b11f1604c704653c0d66b924

savonarola (approved):

emqxqa:
[TestExecution](https://emqx.atlassian.net/browse/QA-15973)

PR #17794, opened 2026-06-30
test(otel): bind-mount config files instead of overlaying /etc/

zmstone:
## Summary

The OTel CT docker-compose mounted the host directory `.ci/docker-compose-file/otel/` onto `/etc/` inside `otel-collector` and `otel-collector-tls` containers. That overlay shadowed the container image's `/etc/` with the host directory. Docker's standard per-container bind mounts for `/etc/hostname`, `/etc/resolv.conf` and `/etc/hosts` then had to create their mountpoint files inside the host `./otel/` directory.

On subsequent CT runs, when Docker's teardown left those stub files behind, the next container start tripped `runc`'s stricter `make_mountpoint` helper with

```
OCI runtime create failed: runc create failed: ...
error mounting "/data/docker/containers/<id>/hostname" to rootfs at "/etc/hostname":
  create mountpoint for /etc/hostname mount: make mountpoint "/etc/hostname": file exists
```

and the test never started. Two recent occurrences on the same runner AMI `ami-024761ae8bd0b667b` (different EC2 instances): https://github.com/emqx/emqx/actions/runs/28435522529/job/84262207448 and https://github.com/emqx/emqx/actions/runs/28356864225/job/84003250328.

Only the OTel test triggers this because it's the only docker-compose in the repo that bind-mounts a host directory at the container's `/etc/`. Other ct-docker tests use single-file or non-`/etc/` mounts.

Mount only the specific config file each collector needs; the container's `/etc/` stays owned by the image so Docker's per-container bind mounts have a clean target. The `./certs:/etc/certs` mount on the TLS variant is unchanged — that lands at `/etc/certs`, not at `/etc/`, and is fine.

Test-only CI change, no user-facing behavior change.

### DIFF
--- a/.ci/docker-compose-file/docker-compose-otel.yaml
+++ b/.ci/docker-compose-file/docker-compose-otel.yaml
@@ -20,9 +20,14 @@ services:
     networks:
       - emqx_bridge
     restart: always
-    command: ["--config=/etc/otel-collector-config.yaml", "${OTELCOL_ARGS}"]
+    command: ["--config=/otel-share/otel-collector-config.yaml", "${OTELCOL_ARGS}"]
     volumes:
-      - ./otel:/etc/
+      # Mount the host's ./otel/ directory at a non-/etc/ path so Docker's
+      # per-container /etc/hostname and /etc/resolv.conf bind mounts don't
+      # have to create their mountpoint files inside ./otel/ on the host
+      # -- which would intermittently trip runc's stricter make_mountpoint
+      # helper with "file exists" on subsequent container starts.
+      - ./otel:/otel-share/
 #    ports:
 #      - "1888:1888"   # pprof extension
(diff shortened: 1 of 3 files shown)


zmstone:
Initial commit broke the test: the original `./otel:/etc/` overlay was doing double duty (deliver the collector config AND expose the file/log + file/metrics exporter outputs back to the host so `emqx_otel_SUITE` could read them via `logs_exporter_file_path/2`). The single-file mount I switched to dropped the second role, so the SUITE got `enoent`.

Force-pushed: same host directory, mount target moved from `/etc/` to `/otel-share/`, and both collector configs updated so their file exporters write to the new path. The container’s `/etc/` is no longer overlaid, so Docker’s per-container `/etc/hostname` and `/etc/resolv.conf` bind mounts have clean targets. The host-side filenames the SUITE reads are unchanged.

thalesmg (approved):

PR #17796, opened 2026-06-30
fix(mqttsn): retire stale disconnected channel on UDP port reuse

zmstone:
Release version: 6.0.4

## Summary

The MQTT-SN gateway indexes channels by the peer UDP `{ip, port}` — the only
addressing UDP offers. When a device disconnects but keeps a non-zero session
expiry (`CleanSession=0`), its channel stays alive in the `disconnected` state
so the session can be resumed. If the OS (on loopback) or a NAT box re-assigns
the same ephemeral source port to a different sender, that sender's `CONNECT`
was routed to the stale `disconnected` channel.

`emqx_mqttsn_channel:handle_in/2` had no clause for a packet arriving in the
`disconnected` state, so the channel crashed with a `function_clause` and the
new device could not connect until the stale session expired.

Crucial changes:

- `apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl`: add `handle_in/2`
  clauses for the `disconnected` state. A `CONNECT` is treated as a takeover —
  the stale channel shuts down with `takeover_by_new_connect` so the connection
  layer can spawn a fresh channel for the new session. Any other packet is
  dropped (`stale_packet_after_disconnect`) and the channel retired.
- `apps/emqx_gateway_mqttsn/test/emqx_sn_protocol_SUITE.erl`: `t_will_case06`
  (which connects with `CleanSession=0`) now retires its lingering session in
  teardown, so the SUITE no longer depends on UDP port-allocation luck. Adds
  `t_takeover_on_udp_port_reuse`, a deterministic regression test that binds a
  second client socket to the first client's source port; it reproduces the
  `function_clause` crash without the fix and passes with it.

The whole `emqx_sn_protocol_SUITE` (70 cases) passes, repeated across runs.

## PR Checklist

### DIFF
--- a/apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
+++ b/apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
@@ -444,6 +444,22 @@ handle_in(
 handle_in(?SN_ADVERTISE_MSG(_GwId, _Radius), Channel) ->
     % ignore
     shutdown(normal, Channel);
+%% A `disconnected' channel is kept alive (indexed by the peer UDP
+%% {ip, port}) so its session can be resumed within the expiry interval.
+%% The OS, or a NAT box, may re-assign the same ephemeral port to a
+%% different sender, in which case that sender's packets are routed here.
+%% Treat a CONNECT as a takeover request: retire this stale channel so the
+%% connection layer can spawn a fresh one for the new session.
+handle_in(
+    ?SN_CONNECT_MSG(_Flags, _ProtoId, _Duration, _ClientId),
+    Channel = #channel{conn_state = disconnected}
+) ->
+    shutdown(takeover_by_new_connect, Channel);
+%% Any other packet on a `disconnected' channel is stale traffic (e.g. a
+%% retransmit from the previous peer of a reused port). Retire the channel
+%% rather than crash; no live session should be addressed here.
+handle_in(_Pkt, Channel = #channel{conn_state = disconnected}) ->
+    shutdown(stale_packet_after_disconnect, Channel);
 %% Ack DISCONNECT even if it is not connected
 handle_in(
     ?SN_DISCONNECT_MSG(_Duration),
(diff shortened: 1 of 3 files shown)


savonarola (dismissed):

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -444,6 +444,22 @@ handle_in(
...
+%% {ip, port}) so its session can be resumed within the expiry interval.
+%% The OS, or a NAT box, may re-assign the same ephemeral port to a
+%% different sender, in which case that sender's packets are routed here.
+%% Treat a CONNECT as a takeover request: retire this stale channel so the
+%% connection layer can spawn a fresh one for the new session.
+handle_in(
+    ?SN_CONNECT_MSG(_Flags, _ProtoId, _Duration, _ClientId),
+    Channel = #channel{conn_state = disconnected}
+) ->
+    shutdown(takeover_by_new_connect, Channel);
```
zmstone, line 457: this means the existing session is lost (shutdown), and new client must reconnect.

@hjianbo maybe we should make possible (from esockd) for a process to unbind from the source tuple, so the session can linger (for future takeover).

hjianbo: One concern/question: since `CleanSession=false` currently maps to a lingering `disconnected` channel/session, this clause matches every `CONNECT` in the `disconnected` state regardless of `ClientId`. If the same client reconnects from the same UDP source tuple before expiry, won't this retire the existing session and force a fresh session instead of resuming it?

Should we distinguish same-`ClientId` reconnects (especially with `CleanSession=false`) from the different-client UDP port-reuse case, or is same-tuple reconnect intentionally allowed to lose the old session until esockd can unbind the source tuple?

hjianbo: Opened https://github.com/emqx/emqx/issues/17809 to track this as a follow-up.

hjianbo (approved):

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
zmstone: good point. will need to see how resume works today.

zmstone: while asleep/awake, NS_CONNECT_MSG gets handled, flips into connected state.

after disconnected, there is no "resume" (as in reactivating an existing process), only takeover.

PR #17808, opened 2026-06-30
refactor(features): make auth part of the core, not a gateable feature

zmstone:
Release version: 6.3.0

## Summary

Make authentication/authorization **part of the core** rather than a gateable feature.

### Problem

`auth` was an optional feature that could be gated off (e.g. `EMQX_FEATURES=ESSENTIAL`). But the default config keeps `listeners.tcp.default.enable_authn = true` and `authorization.no_match = deny`. With `auth` gated off, no authn/authz provider is registered, so the core access control rejects every MQTT `CONNECT` (`no_authn_hooks`) and denies all publish/subscribe (`authorization.no_match`). In other words, a default essential node rejected every client — surfaced by running the Paho interoperability suite against an essential node.

Since a broker is not usable without authn/authz, and the feature gates are not released yet (6.3 GA), there is no reason to keep `auth` gateable.

### Change

- `auth` is removed from `known_features/0`; its apps (`emqx_auth`, the auth drivers, and the `emqx_auth_*` backends, discovered by name) are now returned by `core_apps/0`, so they always start. `core_apps/0` result is cached in `persistent_term` since it is now computed.
- `deps => [auth]` is dropped from `gateways`, `metrics`, and `gcp_device` (auth is always available now).
- The `bundled` feature concept is now unused and is removed entirely: from the `/features` response schema, its i18n description, and the boot test.

After the change, an essential node accepts connections and pub/sub with the default config — verified with Paho `test_connect::test_basic` against essential and no auth overrides.

### Memory (unchanged from a usable essential node)

| Metric | full | essential-interactive |
|---|--:|--:|
| OS RSS | 489 MiB | 177 MiB |
| `erlang:memory(total)` | 248 MB | 109 MB |
| `code` (loaded beams) | 103.7 MB | 20.7 MB |
| loaded modules | 3674 | 907 |
| running apps | 259 | 110 |

Auth is ~11 always-on apps / +65 loaded modules over a hypothetical auth-less essential node (~3 MiB RSS, within jitter). Interactive code loading means the unconfigured auth backends (mongodb/redis/ldap/http) never load their beams, so essential stays far below full (RSS 177 vs 489 MiB, −64%).

## PR Checklist

### DIFF
--- a/apps/emqx_machine/src/emqx_machine_features.erl
+++ b/apps/emqx_machine/src/emqx_machine_features.erl
@@ -17,6 +17,7 @@
 %%------------------------------------------------------------------------------
 
 -define(PT_KEY, {?MODULE, features}).
+-define(CORE_APPS_PT_KEY, {?MODULE, core_apps}).
 -define(ENV_VAR, "EMQX_FEATURES").
 
 %%------------------------------------------------------------------------------
@@ -37,7 +38,9 @@ is_umbrella_application_enabled(Application) when is_atom(Application) ->
 
 %% only for tests
 clear_features() ->
-    persistent_term:erase(?PT_KEY).
+    _ = persistent_term:erase(?PT_KEY),
+    _ = persistent_term:erase(?CORE_APPS_PT_KEY),
+    ok.
 
 %%------------------------------------------------------------------------------
 %% Internal fns
@@ -52,6 +55,16 @@ features() ->
     end.
 
 core_apps() ->
+    case persistent_term:get(?CORE_APPS_PT_KEY, undefined) of
+        undefined ->
+            Apps = base_core_apps() ++ auth_apps(),
+            _ = persistent_term:put(?CORE_APPS_PT_KEY, Apps),
+            Apps;
+        Apps ->
+            Apps
+    end.
+
+base_core_apps() ->
     [
         emqx,
         emqx_machine,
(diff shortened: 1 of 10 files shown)


apps/emqx_machine/src/emqx_machine_features.erl
```diff
@@ -203,18 +206,31 @@ data_integration_dynamic_apps() ->
...
+        emqx_redis,
+        emqx_bridge_http
+    ],
+    Backends = [
         App
-     || App <- RebootApps,
+     || App <- full_sorted_reboot_apps(),
         case atom_to_list(App) of
             "emqx_auth_" ++ _ ->
                 true;
```
thalesmg: I know it was like this before this patch, but I think it would be good to add this extra check:


```suggestion
                emqx_machine_boot:is_umbrella_app(App);
```

apps/emqx_machine/test/emqx_machine_features_tests.erl
```diff
@@ -216,9 +216,11 @@ test_individual_feature(Feature, Info) ->
     ),
     Apps = lists:usort(Apps0 ++ Apps1),
     ?with_features(F, F, begin
-        Enabled = enabled_apps(),
+        Enabled = lists:usort(enabled_apps()),
```
thalesmg: is this needed?  the list should already contain unique items.  🤔

thalesmg (dismissed):

thalesmg (approved):

PR #17813, opened 2026-07-01
feat(rbac): require privilege scopes to stand alone in user and api-key scope lists

zmstone:
Release version: 6.0.4, 6.1.4, 6.2.3

## Summary

Four scopes — `system`, `user_management`, `api_key_management`, `sso_management` — are administrator-equivalent in effect: a caller holding any one of them can reach the areas the other scopes would gate (e.g. `system` covers `/configs/*`, `user_management` can provision another account with any scopes, `api_key_management` can mint a key with any scopes, `sso_management` can rotate the SSO backend). Assigning one of them together with a "restricted" scope list therefore cannot restrict the account.

Dashboard user creation/update (`POST /users`, `PUT /users/:username`) and API-key creation/update (`POST /api_key`, `PUT /api_key/:name`) now reject scope lists that combine any of the four privilege scopes with any non-privilege scope. Callers must supply either a privilege-only list (admin-equivalent) or a non-privilege-only list (genuine least-privilege).

Existing records with a mixed scope set continue to function until the next edit; the next `PUT` against such a record must split the scope list to succeed.

### Implementation notes

- The rule is a schema tightening on writes only; reads are unaffected. The check is applied to the scope list the client explicitly supplied — an omitted `scopes` field (which materialises to the role default, itself a mix of privilege and non-privilege scopes) and an explicit empty list are both treated as the legacy unrestricted / deny-all cases and skip the check.
- `mfa_management` stays out of the privilege group: any login-user role may hold it and it is not administrator-equivalent, so it counts as a non-privilege scope.
- New helpers live in `emqx_scope_catalog` (`check_privilege_scope_mutex/1`, `partition_privilege_scopes/1`) with the `?PRIVILEGE_SCOPES` group defined in `emqx_api_key_scopes.hrl`. Wired into `emqx_dashboard_api:validate_login_user_scopes/3`, `emqx_mgmt_api_api_keys:validate_scopes/3`, and the dashboard CLI create path. The bootstrap-file loader drops the privilege scopes from a mixed line (keeping the more restricted subset) and reports them, consistent with its existing lenient policy.

### Known limitations

- **Namespaced administrators are exempt.** `system`/`user_management`/`api_key_management` are already legitimately part of a namespaced admin's allowed set, and the RBAC dispatch is the authoritative gate for those callers, so the mutex rule (a global-admin concern) is skipped for them. This is documented on `maybe_check_privilege_mutex/2`.

## PR Checklist


apps/emqx_utils/include/emqx_api_key_scopes.hrl
```diff
@@ -126,3 +126,26 @@
     ?SCOPE_API_KEY_MGMT
 ]).
 -define(NS_ADMIN_ALLOWED_SCOPES, ?NS_ADMIN_COMMON_SCOPES ++ ?NS_ADMIN_LOGIN_SCOPES).
+
+%% ── Privilege scope group ──────────────────────────────────────────
+%%
+%% Any one of these scopes is administrator-equivalent in effect:
+%%   * `system' covers `/configs/*' (and `emqx ctl conf load'), which
```
zmstone: API scope has nothing to do with ctl command.
```suggestion
%%   * `system' covers `/configs/*' and `/data/*`, which
```

qzhuyan (approved):

PR #17815, opened 2026-07-01
fix(mqttsn): preserve sessions across UDP proxy reroutes

hjianbo:
Release version: 6.0.4

## Summary

This PR changes MQTT-SN UDP listeners to use `esockd_udp_proxy` with ClientId-based routing, backed by the esockd UDP proxy reroute and owner-aware detach/close callback fix in emqx/esockd#221.

The gateway now treats MQTT-SN ClientId as the logical connection identity instead of relying on the UDP source tuple. This allows asleep sessions to be resumed when a client wakes from a different UDP source tuple, and prevents a reused UDP source tuple from delivering another ClientId's packets into the old session.

`PINGREQ(ClientId)` only reuses existing `asleep` or `awake` channels. It does not migrate a `connected` session from another UDP tuple, so an already connected client continues to receive traffic on its current socket.

When an esockd UDP proxy is rebound from one ClientId to another, MQTT-SN receives an owner-tagged transport detach notification for the old logical connection. EMQX only acts on owner-tagged detach/close notifications from the current UDP proxy, so an old proxy cannot close a newer socket. MQTT-SN's legacy ownerless callbacks are no-ops, while the shared gateway connection process preserves CoAP's legacy bare close behavior. The CoAP UDP proxy close callback also opts in to owner-tagged close notifications. When an asleep session is awakened through a new proxy, the channel retires the previous proxy by ClientId, including the cross-node case where the new UDP tuple is routed to another EMQX node.

Asleep or disconnected channels keep the lingering logical session. Active channels detached by a proxy reroute run the normal logical disconnect path without marking the channel connection closed, so a persistent `CleanSession=false` session remains available for later takeover while the newly rebound proxy stays open. Existing non-UDP-proxy listener behavior remains unchanged.

## PR Checklist

### DIFF
--- a/apps/emqx_gateway/src/bhvrs/emqx_gateway_conn.erl
+++ b/apps/emqx_gateway/src/bhvrs/emqx_gateway_conn.erl
@@ -494,7 +494,7 @@ handle_msg(
     Ctx = ChannMod:info(ctx, Channel),
     ok = emqx_gateway_ctx:metrics_inc(Ctx, 'bytes.received', Oct),
 
-    NState = State#state{socket = NSock},
+    NState = State#state{socket = NSock, sockstate = running},
     {ok, next_incoming_msgs(Packets), NState};
 handle_msg({Inet, _Sock, Data}, State) when
     Inet == tcp;
@@ -541,6 +541,22 @@ handle_msg({inet_reply, _Sock, {error, Reason}}, State) ->
 handle_msg({close, Reason}, State) ->
     ?tp(debug, force_socket_close, #{reason => Reason}),
     handle_info({sock_closed, Reason}, close_socket(State));
+handle_msg(udp_proxy_closed, State = #state{sockstate = closed}) ->
+    {ok, State};
+handle_msg(
+    udp_proxy_closed,
+    State = #state{
+        chann_mod = ChannMod,
+        channel = Channel
+    }
+) ->
+    case ChannMod:info(conn_state, Channel) of
+        ConnState when ConnState == asleep; ConnState == disconnected ->
+            ?tp(debug, udp_proxy_detached, #{conn_state => ConnState}),
+            {ok, State#state{sockstate = closed}};
+        _ConnState ->
+            handle_udp_proxy_closed(State)
+    end;
 handle_msg(udp_proxy_closed, State) ->
     ?tp(debug, udp_proxy_closed, #{reason => normal}),
     handle_info({sock_closed, normal}, close_socket(State));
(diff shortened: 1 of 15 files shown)


apps/emqx_gateway/src/proto/emqx_gateway_cm_proto_v1.erl
```diff
@@ -10,6 +10,7 @@
     introduced_in/0,
 
     get_chan_info/3,
+    get_chan_info/4,
```
zmstone: should introduce v2 instead of modifying v1

apps/emqx_gateway/src/emqx_gateway_cm.erl
```diff
@@ -303,10 +309,10 @@ set_chan_stats(GwName, ClientId, ChanPid, Stats) ->
...
 
 -spec connection_closed(gateway_name(), emqx_types:clientid()) -> true.
-connection_closed(GwName, ClientId) ->
-    %% XXX: Why we need to delete conn_mod tab ???
-    Chan = {ClientId, self()},
-    ets:delete_object(tabname(conn, GwName), Chan).
+connection_closed(_GwName, _ClientId) ->
+    %% Transport close may keep the channel/session alive. The registry/chan/conn/info
+    %% cleanup is done by unregister_channel/2 or the DOWN path in do_unregister_channel/3.
+    true.
```
zmstone, line 315: this impacts all gateways. need a confirmation check (test coverage) for all gateways to ensure no leak.

apps/emqx_gateway/src/bhvrs/emqx_gateway_conn.erl
```diff
@@ -541,9 +542,24 @@ handle_msg({inet_reply, _Sock, {error, Reason}}, State) ->
 handle_msg({close, Reason}, State) ->
     ?tp(debug, force_socket_close, #{reason => Reason}),
     handle_info({sock_closed, Reason}, close_socket(State));
-handle_msg(udp_proxy_closed, State) ->
```
zmstone, line 544: confirm no other gatway emits this ?

hjianbo: Confirmed that CoAP still emits bare `udp_proxy_closed` through its legacy `close/2` callback. The shared connection layer now preserves the old close behavior for that message. MQTT-SN's ownerless `close/2` and `detach/2` are no-ops; its actual esockd 5.17.1 path uses owner-aware `/3` callbacks. Added an end-to-end CoAP regression test and re-ran the full CoAP and MQTT-SN suites.

apps/emqx_gateway/src/emqx_gateway_cm.erl
hjianbo: Kept the explicit no-op and added lifecycle coverage in the shared gateway CM suite. The test verifies that transport close retains the callable chan/conn/info entries, and that `unregister_channel/2` removes all of them; the existing process-DOWN test covers the other cleanup path. The full gateway CM and CoAP suites pass.

apps/emqx_gateway/src/proto/emqx_gateway_cm_proto_v1.erl
hjianbo: Fixed.

zmstone (approved):

PR #17819, opened 2026-07-01
fix(rel): reject alpha/beta dashboard when cutting a final release

zmstone:
## Summary

`scripts/rel/cut.sh` already refuses to tag a final EMQX release without a matching `changes/<tag>.en.md` file and (for `.0` cuts) a BPAPI baseline. This PR adds a matching guard against shipping a pre-release dashboard by accident: for final release cuts, if `EMQX_DASHBOARD_VERSION` in the top-level `Makefile` still ends in `-alpha...` or `-beta...`, `cut.sh` now exits with an error and asks the operator to bump the dashboard version first.

Pre-release EMQX cuts (`rc`, `alpha`, `beta`) continue to allow a pre-release dashboard.

The extraction reads the first `export EMQX_DASHBOARD_VERSION ?= <vsn>` line from the `Makefile` (matching the exact form used today) and only checks for `alpha` / `beta` substrings — the `-1`, `-2`, ... build-number suffixes on final dashboard tags pass through.

## PR Checklist

### DIFF
--- a/scripts/rel/cut.sh
+++ b/scripts/rel/cut.sh
@@ -254,14 +254,14 @@ check_bpapi() {
 ## to ship a pre-release dashboard.
 check_dashboard_version() {
     local dashboard_vsn
-    dashboard_vsn="$(sed -nE 's/^export[[:space:]]+EMQX_DASHBOARD_VERSION[[:space:]]*\??=[[:space:]]*([^[:space:]]+).*/\1/p' Makefile | head -n 1)"
+    dashboard_vsn="$(make -s print-dashboard-version)"
     if [ -z "$dashboard_vsn" ]; then
-        logerr "Could not read EMQX_DASHBOARD_VERSION from Makefile"
+        logerr "Could not read EMQX_DASHBOARD_VERSION via 'make print-dashboard-version'"
         exit 1
     fi
     case "$dashboard_vsn" in
         *alpha*|*beta*)
-            logerr "EMQX_DASHBOARD_VERSION in Makefile is a pre-release ($dashboard_vsn)"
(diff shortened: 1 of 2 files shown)


scripts/rel/cut.sh
```diff
@@ -248,6 +248,30 @@ check_bpapi() {
...
+## Assert that EMQX_DASHBOARD_VERSION in Makefile is a final release,
+## i.e. does not contain 'alpha' or 'beta'. Called only when cutting
+## a final EMQX release; pre-release cuts (rc/alpha/beta) are allowed
+## to ship a pre-release dashboard.
+check_dashboard_version() {
+    local dashboard_vsn
+    dashboard_vsn="$(sed -nE 's/^export[[:space:]]+EMQX_DASHBOARD_VERSION[[:space:]]*\??=[[:space:]]*([^[:space:]]+).*/\1/p' Makefile | head -n 1)"
+    if [ -z "$dashboard_vsn" ]; then
+        logerr "Could not read EMQX_DASHBOARD_VERSION from Makefile"
+        exit 1
```
keynslug: Q: Wouldn't be simpler to add Makefile recipe to just echo dashboard version? This approach seems to rely too much on the Makefile structure.

zmstone:
Good point — pushed 525859ef17 that adds a `print-dashboard-version` recipe to the Makefile and has `cut.sh` call `make -s print-dashboard-version` instead of sed-parsing the export line. Smoke-tested: `make -s print-dashboard-version` prints `2.0.3-1`, and overriding via `EMQX_DASHBOARD_VERSION=2.1.0-alpha.1 make -s print-dashboard-version` prints the override — so the pre-release detection works against the resolved make variable, not the file text.

keynslug (approved):

PR #17827, opened 2026-07-01
test: ESSENTIAL-mode paho interop, authn/authz and cluster smoke tests

zmstone:
Follow-up test coverage for the feature-gates work (auth is now core, #17808), focused on making sure the minimal **ESSENTIAL** preset is fully usable.

## What

- **Paho against ESSENTIAL** (`run_docker_tests.yaml`): the `paho-mqtt-testing` job now also runs the full MQTT interoperability suite once against an `EMQX_FEATURES=ESSENTIAL` node, in addition to the existing FULL runs.

- **Auth/authz smoke test** (`run_docker_tests.yaml` → `essential-auth-smoke-test`): boots a single ESSENTIAL node with built-in-database authentication (users seeded from `bootstrap.csv`) and `acl.conf` file authorization, then uses a real MQTT client to assert valid credentials connect; wrong-password and anonymous are rejected; and per-topic ACL allow/deny works (`no_match = deny`).

- **Cluster smoke test** (new `run_cluster_tests.yaml`, wired into the PR entrypoint): starts a two-node cluster via `start-two-nodes-in-docker.sh` in **both FULL and ESSENTIAL** mode and asserts, at the MQTT level:
  - a subscription on one node is replicated into the cluster-wide routing table and is visible from **both** nodes (cross-node routing);
  - a message published on a separate connection is delivered end to end.

  `start-two-nodes-in-docker.sh` now forwards caller-set `EMQX_*` env vars (e.g. `EMQX_FEATURES`) to both node containers, excluding the node-identity/clustering vars it manages.

All MQTT client checks run in throwaway `python:3-slim` containers sharing the broker network, so there is no dependency on the runner's Python environment.

## Verification

Both `scripts/test/essential-auth-smoke/run.sh` and `scripts/test/cluster-smoke/run.sh` were validated end-to-end locally against a real EMQX docker image (cluster forms, cross-node route replication and delivery confirmed; all authn/authz checks pass). `shellcheck`, `py_compile` and YAML linting are clean.

No user-facing changes; CI/test only.

### DIFF
--- a/scripts/test/cluster-smoke/publisher.py
+++ b/scripts/test/cluster-smoke/publisher.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+"""One-shot MQTT publisher for the cluster smoke test.
+
+Connects through HAProxy (a separate connection from the subscriber, so with
+HAProxy's static round-robin balancing it typically lands on the other node)
+and publishes a payload to the smoke topic.
+"""
+import sys
+import time
+import paho.mqtt.client as mqtt
+
+HOST = sys.argv[1] if len(sys.argv) > 1 else "haproxy"
+PORT = int(sys.argv[2]) if len(sys.argv) > 2 else 1883
+TOPIC = sys.argv[3] if len(sys.argv) > 3 else "cluster/smoke/1"
+PAYLOAD = sys.argv[4] if len(sys.argv) > 4 else "cluster-smoke-payload"
+
+
+def new_client(cid):
+    try:  # paho-mqtt 2.x
+        return mqtt.Client(
+            mqtt.CallbackAPIVersion.VERSION1, client_id=cid, protocol=mqtt.MQTTv5
+        )
+    except AttributeError:  # paho-mqtt 1.x
+        return mqtt.Client(client_id=cid, protocol=mqtt.MQTTv5)
+
+
+c = new_client("cluster-smoke-pub")
+connected = {}
+c.on_connect = lambda cl, u, flags, rc, props=None: connected.update(
+    rc=getattr(rc, "value", rc)
+)
+c.connect(HOST, PORT, 30)
+c.loop_start()
+for _ in range(50):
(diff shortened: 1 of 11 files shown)


hjianbo (dismissed):

scripts/test/cluster-smoke/run.sh
```diff
@@ -0,0 +1,89 @@
...
+
+cleanup() {
+  docker rm -f "${SUB}" >/dev/null 2>&1 || true
+}
+trap cleanup EXIT
+cleanup
+
+## Start the long-lived subscriber (installs paho-mqtt, then subscribes).
+docker run -d --name "${SUB}" --net "${NET}" -v "${HERE}:/smoke:ro" "${PYTHON_IMAGE}" \
+  sh -c "pip install --quiet --disable-pip-version-check --root-user-action=ignore paho-mqtt \
```
thalesmg: better pin a specific paho version to not only avoid breaking due to time passage and new releases, but also to avoid the try...catch to handle multiple potential client versions in the scripts?

scripts/test/cluster-smoke/publisher.py
```diff
@@ -0,0 +1,45 @@
...
+import paho.mqtt.client as mqtt
+
+HOST = sys.argv[1] if len(sys.argv) > 1 else "haproxy"
+PORT = int(sys.argv[2]) if len(sys.argv) > 2 else 1883
+TOPIC = sys.argv[3] if len(sys.argv) > 3 else "cluster/smoke/1"
+PAYLOAD = sys.argv[4] if len(sys.argv) > 4 else "cluster-smoke-payload"
+
+
+def new_client(cid):
+    try:  # paho-mqtt 2.x
```
thalesmg: by pinning the paho version, this shouldn't be necessary.

thalesmg (dismissed):

thalesmg (approved):

PR #17790, opened 2026-06-30
fix(dashboard): redact TOTP secret from login-failed server log

zmstone:
Release version: 6.0.4, 6.1.3, 6.2.2

## Summary

Fixes emqx/emqx-dev-team-tasks#274.

The `dashboard_login_failed` server-log entry included the TOTP shared secret when an MFA-enabled user logged in without supplying a token during first-time setup. The reason map is now passed through `emqx_utils:redact/1` before it reaches `?SLOG`, which replaces the `secret` value with `******`.

The secret is still returned in the 401 response body, where the dashboard needs it to render the authenticator QR code on first-time MFA setup (documented behavior in `apps/emqx_dashboard/README.md`). Only the server-side log copy is removed; the redaction is log-only.

Key change in `apps/emqx_dashboard/src/emqx_dashboard_api.erl`. A new CT case (`emqx_dashboard_mfa_SUITE:t_login_failed_log_redacts_totp_secret`) captures the log event and asserts the secret bytes are absent from the log while still present in the 401 body.

## PR Checklist
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17657, opened 2026-06-19
fix(plugins): redact dashboard credential headers from plugin api callbacks

savonarola:
Fixes emqx/emqx-dev-team-tasks#239

Release version: 6.0.4

## Summary

Remove `authorization` and `cookie` headers from the `ReqInfo` map passed to plugin callbacks (plugins can still identify the caller through the sanitized `auth_meta`).

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17654, opened 2026-06-19
fix(authn_api): redact provider secrets in authenticator creation response

savonarola:
Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/247

Release version: 6.0.3

## Summary

Redact secrets in `POST /authentication`. The deprecated listener-authenticator creation path is also covered.

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17653, opened 2026-06-19
fix(prometheus): redact authorization headers in config api responses

savonarola:
Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/248

Release version: 6.0.3

## Summary

Redact sensitive header values from the Prometheus configuration API responses.

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17857, opened 2026-07-02
fix: backport recent redaction and secret-handling fixes to v5

zmstone:
Release versions: 5.8.11, 5.10.4

## Summary

Backport a batch of recent log-redaction and secret-handling fixes from the v6 line to v5. Sensitive fields are now hidden in the affected logs and API responses.

### Ported cleanly
- #17653 — redact authorization headers in Prometheus config API responses.
- #17654 — redact provider secrets in the authenticator creation response.
- #17739 — broad sweep that redacts sensitive data across authentication/authorization backend traces and related log lines.
- #17787 — drop request headers from the HTTP connector worker-down error reason.
- #17791 — redact `jose_jwk` records (JWT signing key material) in logs.
- #17853 — store additional conventionally-sensitive request headers as secrets and redact iolist-shaped header keys.

### Ported with adjustments
- #17739 — on v5 the template rendering helpers live in `emqx_auth_utils` (v6 renamed the module to `emqx_auth_template`); the redacted render helpers were added there instead. Backend state shapes differ on v5, so the redaction call sites were adapted by hand. The JWT algorithm-allowlist refactor that is entangled with the v6 diff was left out; only the redaction change was taken.

### Skipped (not applicable to v5)
- #17657 — the plugin API-callback endpoint it hardens is v6-only.
- #17790 — the dashboard MFA/TOTP setup flow that could carry the secret is v6-only.
- NATS gateway hunk of #17739 — the NATS gateway is a v6-only app.

## PR Checklist


savonarola:
I started investigating thinking that my ported changes could fail the build; they didn't but the fix was easy so I pushed a commit.

savonarola (approved):

PR #17780, opened 2026-06-30
fix(plugins): purge stale plugins after cluster rejoin

hjianbo:
Release version:
6.0.4

Fix: https://emqx.atlassian.net/browse/EMQX-14124

## Summary

Port of #17764 to dev-60.

When a node rejoins after a plugin was uninstalled while the node was offline, plugin startup now treats the cluster plugin config as authoritative and purges local plugin packages that are no longer configured. The purge keeps alternate local versions of any plugin app that is still configured so force-sync/version switching continues to work.

This adds EUnit coverage for purge selection and a CT case for the cluster rejoin-after-uninstall path. No schema changes.

## PR Checklist
- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ce/fix-17764.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17861, opened 2026-07-03
Revert "fix(plugins): purge stale plugins after cluster rejoin"

hjianbo:
Reverts emqx/emqx#17780


zmstone (approved):

zmstone (changes requested):
need a changelog since 6.1.3 is already released.

zmstone (approved):

PR #17867, opened 2026-07-03
test(crl): accept {error, closed} for revoked-cert rejection

zmstone:
Fixes a flaky `emqx_crl_cache_SUITE` failure:

```
emqx_crl_cache_SUITE ==> t_revoke_then_refresh: FAILED
  expected: {error,{ssl_error,_Sock,{tls_alert,{certificate_revoked,_}}}}
       got: {error,closed}
```

When a revoked client certificate is presented, the server sends a `certificate_revoked` fatal TLS alert. Under a long-standing race, the client can observe the socket close (`{error, closed}`) before it reads the alert — both outcomes mean the connection was refused, which is what the test verifies.

`t_revoked` already tolerated all three outcomes (including `{error, closed}`, with a comment noting the race). `t_revoke_then_refresh` and `do_t_update_listener` still used a strict `?assertMatch` on the alert and therefore flaked. This extracts the tolerant logic into a shared helper `assert_connection_rejected_by_revoked_cert/1` and uses it at all three sites (the helper `ct:fail`s on any other result, so it is not weaker than before for non-revoked outcomes).

Test-only change; no changelog.

Release version: 6.3.0

### DIFF
--- a/apps/emqx/test/emqx_crl_cache_SUITE.erl
+++ b/apps/emqx/test/emqx_crl_cache_SUITE.erl
@@ -473,6 +473,22 @@ update_listener_via_api(ListenerId, NewConfig) ->
     Path = emqx_mgmt_api_test_util:api_path(["listeners", ListenerId]),
     request(put, Path, [], NewConfig).
 
+%% A revoked client certificate must be rejected during the TLS handshake.
+%% The server sends a `certificate_revoked' fatal alert, but under a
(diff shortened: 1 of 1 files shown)


zmstone:
Superseded by https://github.com/emqx/emqx/pull/17868, which targets dev-60 so the fix reaches all active v6 dev branches (dev-60/61/62/63) via the daily sync PRs. Closing this dev-63-only one.

PR #17877, opened 2026-07-06
fix: hardcoded svc.cluster.local in helm chart

### LINKED ISSUE #17876: Helm chart (emqx-enterprise): `EMQX_HOST` hardcodes `svc.cluster.local`, ignoring `clusterDomain`
### What happened?

The `emqx-enterprise` Helm chart exposes a `clusterDomain` value (default `cluster.local`) and uses it for cluster-discovery config in `templates/configmap.yaml` (both `EMQX_CLUSTER__DNS__NAME` and `EMQX_CLUSTER__K8S__SUFFIX` use `{{ .Values.clusterDomain }}`).

However, `templates/StatefulSet.yaml` hardcodes `svc.cluster.local` when setting each pod's `EMQX_HOST`, which becomes the host part of the Erlang node name:

```yaml
- name: EMQX_HOST
  value: $(POD_NAME).{{ include "emqx.fullname" . }}-headless.{{ .Release.Namespace }}.svc.cluster.local
```

On a Kubernetes cluster whose domain is **not** `cluster.local` (kubelet `--cluster-domain`), setting `clusterDomain` fixes discovery but not `EMQX_HOST`, so each node still advertises itself under `...svc.cluster.local`. That FQDN has no DNS records on such a cluster, so:

- Erlang distribution can't resolve the node's own name. `emqx ctl cluster status` fails with `Node 'emqx@...svc.cluster.local' not responding to pings`, even on a single pod.
- Multi-node clusters can't form, because discovery resolves peers under the real domain while node names are stamped with `cluster.local`.

This was hit by a user on a `6.2.1` Helm install (the `not responding to pings` symptom above).

`EMQX_HOST` should use `{{ .Values.clusterDomain }}`, consistent with the ConfigMap, so that setting `clusterDomain` produces a working node name and cluster. With the default value the rendered output is unchanged (`svc.cluster.local`), so the fix is backward compatible.

### How can we reproduce it?

```console
helm template t deploy/charts/emqx-enterprise \
  --set clusterDomain=k8s.corp.internal \
  --show-only templates/StatefulSet.yaml | grep -A1 EMQX_HOST
```

The rendered `EMQX_HOST` still ends in `svc.cluster.local` instead of `svc.k8s.corp.internal`, whereas the ConfigMap's discovery settings correctly render `k8s.corp.internal`.

`clusterDomain` was introduced in `81e7f26e` (`fix: add clusterDomain to enterprise chart`), which updated the ConfigMap but not the StatefulSet's `EMQX_HOST`. The same hardcode is present on `master`.

Proposed one-line fix:

```diff
-            value: $(POD_NAME).{{ include "emqx.fullname" . }}-headless.{{ .Release.Namespace }}.svc.cluster.local
+            value: $(POD_NAME).{{ include "emqx.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
```

Workaround for existing installs until the fix ships — patch `EMQX_HOST` on the StatefulSet to the real domain, then restart (example: release `emqx`, namespace `default`):

```console
kubectl set env statefulset/emqx \
  EMQX_HOST='$(POD_NAME).emqx-headless.default.svc.<your.cluster.domain>'
kubectl rollout restart statefulset/emqx
```

I will open a PR with the one-line fix.

### EMQX version

Enterprise Helm chart. Reproduced on `master` and 6.x (`6.3.0-rc.1`); reported in the field on the `6.2.1` chart. The hardcode predates the `clusterDomain` value, so earlier releases are affected as well.

id:
fixes #17876

6.0.4, 6.1.3, 6.2.2, 6.3.0

## PR Checklist

### DIFF
--- a/.github/workflows/run_helm_tests.yaml
+++ b/.github/workflows/run_helm_tests.yaml
@@ -19,6 +19,16 @@ env:
   EMQX_ENTERPRISE_LICENSE: evaluation
 
 jobs:
+  helm_render_test:
+    runs-on: ${{ github.repository_owner == 'emqx' && 'aws-ubuntu22.04-amd64' || 'ubuntu-22.04' }}
+    defaults:
+      run:
+        shell: bash
+    steps:
+    - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+    - name: check chart rendering
+      run: ./scripts/check-helm-chart.sh
+
   helm_test:
     runs-on: ${{ github.repository_owner == 'emqx' && 'aws-ubuntu22.04-amd64' || 'ubuntu-22.04' }}
     defaults:
@@ -26,6 +36,9 @@ jobs:
         shell: bash
     env:
       EMQX_NAME: ${{ matrix.profile }}
+      # Run the whole suite under a non-default cluster DNS domain so it
+      # exercises the chart's clusterDomain end-to-end.
+      CLUSTER_DOMAIN: k8s.corp.internal
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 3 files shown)


zmstone:
rebase to dev-60

qzhuyan (approved):

PR #17882, opened 2026-07-06
test(sessds): stabilize flaky t_session_replay_retry

zmstone:
Release version: 6.1.3

## Summary

Test-only stabilization of the flaky CT case
`emqx_persistent_session_ds_SUITE:t_session_replay_retry`, which intermittently
failed with `NPubs=9` (expected 10).

The setup-phase wait gave the subscriber only a 5 s per-message gap to receive
all 10 durable publishes. Delivery runs through the persistent-session DS
stream-polling path, whose per-message shard poll can occasionally lag past 5 s
under CI load, so the 10th message arrived late. Every publish was acked before
the wait, so the message is durably stored — this is a delivery-latency race,
not message loss.

`wait_publishes/2` returns as soon as all expected messages arrive, so raising
its per-message-gap timeout costs nothing in the happy path and only stretches
the rare slow tail. This bumps the two waits that return early on success (steps
3 and 6) from 5 s to 30 s, and deliberately leaves the step-5 wait at 5 s: with
the mocked shards still unavailable it is expected to end on the gap timeout, so
a larger value there would only add dead wait time to every run.

Ran the case 10× locally with no failures; happy-path runtime is unchanged.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_persistent_session_ds_SUITE.erl
+++ b/apps/emqx/test/emqx_persistent_session_ds_SUITE.erl
@@ -1298,8 +1298,13 @@ t_session_replay_retry(_Config) ->
                 ClientsPub
             ),
             %% 3. The subscriber receives them, but doesn't ack. This
-            %% sets the stage for message replay:
-            Pubs0 = emqx_common_test_helpers:wait_publishes(NClients, 5_000),
+            %% sets the stage for message replay. Delivery through the
+            %% persistent-session DS stream-polling path can lag under CI load;
+            %% since every message was acked it is durably stored, so nothing is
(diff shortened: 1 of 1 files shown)


thalesmg (approved):

thalesmg:
still happens: https://github.com/emqx/emqx/actions/runs/33096586357/job/98626235934#step:5:537

I suspect it might have something to do with not acking and occasionally 2 messages being assigned to the same shard.  IIRC, in the past there was a more or less similar flakiness I debug, and _I think_ that was the reason in that other PS-related suite.

... though, the sharding keys (clientids) and the number of shards are deterministic, so it actually shouldn't be this case...  :speak_no_evil:

PR #17887, opened 2026-07-07
feat: add sync request plugin

hjianbo:
Release version:
5.10.5

## Summary

Add `emqx_sync_request` as an in-tree runtime plugin under `plugins/`. It exposes a Management API endpoint for synchronous MQTT request/response flows and includes e2e CT coverage, including cross-node request/response behavior.

## PR Checklist


plugins/emqx_sync_request/README.md
```diff
@@ -0,0 +1,70 @@
+# EMQX Sync Request
+
+`emqx_sync_request` publishes one MQTT request through the EMQX Management API
```
zmstone:
```suggestion
`emqx_sync_request` publishes one MQTT request through the EMQX REST API
```

plugins/emqx_sync_request/README.md
```diff
@@ -0,0 +1,70 @@
...
+## Configuration
+
+```hocon
+default_timeout = "10s"
+max_timeout = "60s"
+max_inflight_requests = 10000
+max_payload_size = "64KB"
+```
+
+`max_inflight_requests` limits local HTTP requests waiting for responses.
```
zmstone: per node

plugins/emqx_sync_request/src/emqx_sync_request.erl
```diff
@@ -0,0 +1,683 @@
...
+            TimeoutMs = maps:get(timeout, Req0),
+            Deadline = now_ms() + TimeoutMs,
+            Req = Req0#{
+                req_ref => ReqRef,
+                waiter => self(),
+                deadline => Deadline
+            },
+            true = ets:insert_new(?REQ_TAB, {ReqRef, Req}),
+            Message = make_request_message(Req),
+            case emqx_mgmt:publish(Message) of
```
zmstone: this route to nodes in the cluster.
what it should do instead:

1. lookup local subscriber (including shared), if found, publish to it directly
2. if not found, lookup routes matching this topic, and forward the request to the gen_server running in the remote node, then wait for response.

zmstone: Return 409 (conflict) if more than one subscriber is found for a PUBLISH.
Given that there is correlation data, I don't see a reason why it should use wildcard subscription, so we should directly lookup non-wildcard subscription (or route).

scripts/package-plugin.escript
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env escript
-%%! -pa _build/emqx-enterprise/lib/jsone/ebin
+%%! -pa _build/emqx-enterprise/lib/jsone/ebin -pa _build/test/lib/jsone/ebin
```
savonarola: The script is used for prod releases, is it safe to add test directory here?

plugins/emqx_sync_request/test/emqx_sync_request_SUITE.erl
```diff
@@ -0,0 +1,1383 @@
...
+-include_lib("emqx_utils/include/emqx_api_key_scopes.hrl").
+-include_lib("emqx_utils/include/emqx_message.hrl").
+
+-define(REQ_TOPIC, <<"sync_request/device/1001/request">>).
+-define(RESP_TOPIC, <<"sync_request/device/1001/response">>).
+-define(REQ_ID, <<"request-id-1">>).
+-define(REQ_PAYLOAD, <<"{\"cmd\":\"reboot\"}">>).
+-define(RESP_PAYLOAD, <<"{\"result\":\"ok\"}">>).
+
+all() ->
```
savonarola, line 24: why list explicitly?

plugins/emqx_sync_request/src/emqx_sync_request.erl
```diff
@@ -0,0 +1,968 @@
...
+    gen_server:start_link({local, ?SERVICE}, ?MODULE, [], []).
+
+-spec child_spec() -> supervisor:child_spec().
+child_spec() ->
+    #{
+        id => ?SERVICE,
+        start => {?MODULE, start_link, []}
+    }.
+
+request(Body) ->
```
savonarola, line 99: nit: some public funs have specs and some not

plugins/emqx_sync_request/src/emqx_sync_request.erl
```diff
@@ -0,0 +1,968 @@
...
+    record_request_result(do_request_body(Body)).
+
+do_request_body(Body) when is_map(Body) ->
+    Config = config(),
+    case parse_request(Body, Config) of
+        {ok, Req} ->
+            do_request(Req, Config);
+        {error, Reason} ->
+            {400, error_body(?CODE_BAD_REQUEST, Reason)}
```
savonarola: Nit
Maybe would be nice 
1) to have this module be free of http semantics, move "body" notation, error bodies, http codes etc. to the API endpoint module
2) to have this module provide a transport agnostic api (jsonish data structs with no transport semantics) so that it may be called via othe channels if needed

plugins/emqx_sync_request/src/emqx_sync_request.erl
```diff
@@ -0,0 +1,968 @@
...
+    install_api_dispatch/0,
+    uninstall_api_dispatch/0,
+    request/1,
+    status/0
+]).
+
+%% Plugin callbacks
+-export([
+    on_config_changed/1,
+    on_health_check/0
```
savonarola, line 41: Nit: although 5.10 does not support `on_handle_api_call` for plugins, imo we better implement `emqx_sync_request_api` module and the callback, and manually inject only the connecting stub (dispatching to `on_handle_api_call`) into minirest.
Thus to have the same code structure in later versions.

plugins/emqx_sync_request/src/emqx_sync_request.erl
```diff
@@ -0,0 +1,968 @@
...
+        error -> undefined
+    end.
+
+required(Key, Map) ->
+    case find(Key, Map) of
+        {ok, Value} -> Value;
+        error -> throw({bad_request, iolist_to_binary([atom_to_binary(Key), <<"_required">>])})
+    end.
+
+get(Map, Key, Default) ->
```
savonarola, line 697: Do we actually need to deal with both atoms and binaries in keys?

plugins/emqx_sync_request/test/emqx_sync_request_cluster_SUITE.erl
```diff
@@ -0,0 +1,238 @@
...
+-compile(nowarn_export_all).
+
+-include_lib("common_test/include/ct.hrl").
+-include_lib("eunit/include/eunit.hrl").
+-include_lib("emqx/include/asserts.hrl").
+-include_lib("emqx/include/emqx_mqtt.hrl").
+
+-define(REQ_PAYLOAD, <<"{\"cmd\":\"reboot\"}">>).
+
+all() ->
```
savonarola, line 17: why list manually?

plugins/emqx_sync_request/test/emqx_sync_request_cluster_SUITE.erl
```diff
@@ -0,0 +1,238 @@
...
+    ].
+
+end_per_suite(_Config) ->
+    ok.
+
+t_request_on_one_node_receives_response_from_another_node(Config) ->
+    WorkDir = emqx_cth_suite:work_dir(?FUNCTION_NAME, Config),
+    Nodes =
+        [HttpNode, ResponderNode] =
+        emqx_cth_cluster:start(cluster_specs(), #{work_dir => WorkDir}),
```
savonarola: cluster init and teardown looks like belonging to `init_per_testcase/end_per_testcase/`

plugins/emqx_sync_request/src/emqx_sync_request.erl
```diff
@@ -0,0 +1,968 @@
...
+        exit:{timeout, _} -> Default
+    end.
+
+monitor_waiter(ReqRef, Waiter) ->
+    gen_server:call(?SERVICE, {monitor_waiter, ReqRef, Waiter}, ?TIMEOUT).
+
+forget_monitor(Mon) ->
+    try gen_server:cast(?SERVICE, {forget_monitor, Mon}) of
+        ok -> ok
+    catch
```
savonarola: Why do we need to catch here?

plugins/emqx_sync_request/src/emqx_sync_request_api.erl
```diff
@@ -0,0 +1,168 @@
...
+]).
+
+-export([scopes/0]).
+
+-export([request/2]).
+
+namespace() -> undefined.
+
+api_spec() ->
+    emqx_dashboard_swagger:spec(?MODULE, #{check_schema => false, translate_body => true}).
```
savonarola: Why don't we check schema? AFAIK we canot translate body without checking schema.

scripts/package-plugin.escript
hjianbo: That's correct. My original intention was to build and package the plugins within CT so I could perform end-to-end testing. To work around the packaging failures in ` emqx_sync_request_SUITE:plugin_package/0`, I have updated the CI workflow steps instead

savonarola:
reviewing...

savonarola (approved):

PR #17947, opened 2026-07-14
resume action buffer workers

terry-xiaoyu:
Release version: 5.8.x

If an HTTP Connector is updated (recreated) and had previously sent messages before the update, the Channel status will be set to "blocked" during the update. In this state, all subsequent messages will be buffered until the status is reset to "running" after 15 seconds.

This issue can be reproduced simply by updating the description field of the Connector.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_manager.erl
+++ b/apps/emqx_resource/src/emqx_resource_manager.erl
@@ -1338,11 +1338,11 @@ continue_with_health_check(#data{} = Data0, CurrentState, HCRes) ->
     {NewStatus, Err} = parse_health_check_result(HCRes, Data0),
     IsDryRun = emqx_resource:is_dry_run(ResId),
     _ = maybe_alarm(NewStatus, IsDryRun, ResId, Err, PrevError),
-    ok = maybe_resume_resource_workers(ResId, NewStatus),
     Data1 = Data0#data{
         status = NewStatus, error = Err
     },
     Data = update_state(Data1),
+    ok = maybe_resume_resource_workers(ResId, NewStatus),
     case CurrentState of
         ?state_connected ->
             continue_resource_health_check_connected(NewStatus, Data);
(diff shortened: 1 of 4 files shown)


zmstone (approved):

apps/emqx_resource/src/emqx_resource_manager.erl
```diff
@@ -1921,6 +1923,24 @@ maybe_resume_resource_workers(ResId, ?status_connected) ->
...
+            );
+        false ->
+            ok
+    end.
+
+should_resume_channel_workers(#{status := ?status_connected}) ->
+    true;
+should_resume_channel_workers(_) ->
+    false.
+
```
thalesmg, line 1943: there are no separate workers for each channel id.  `worker_pids(ChannelId)` will always return an empty list.


```suggestion
```

terry-xiaoyu: It's not the Channel, but the Connector that won't have any workers: https://github.com/emqx/emqx/blob/709ca6774573479a6be75c9f83b76a8b0fe478c3/apps/emqx_connector/src/emqx_connector_resource.erl#L367

This is exactly the issue that the current PR aims to fix. Previously, it only called `maybe_resume_resource_workers(ResId, NewStatus)`, where `ResId` is the ID of the Connector. However, since there are no workers, this line of code is actually ineffective.

```erlang
v5.8.11-g0abf56bb(emqx@127.0.0.1)5> emqx_resource_buffer_worker_sup:worker_pids(<<"action:http:http-test-conn:connector:http:http-test-conn">>).
[<0.3787.0>,<0.3788.0>,<0.3789.0>,<0.3790.0>,<0.3791.0>,
 <0.3792.0>,<0.3793.0>,<0.3794.0>,<0.3795.0>,<0.3796.0>,
 <0.3797.0>,<0.3798.0>,<0.3799.0>,<0.3800.0>,<0.3801.0>,
 <0.3802.0>]
v5.8.11-g0abf56bb(emqx@127.0.0.1)6> emqx_resource_buffer_worker_sup:worker_pids(<<"connector:http:http-test-conn">>).
[]
```

terry-xiaoyu: Another strange thing is that both `Authn` and `Authz` have created buffer worker processes, but all their tasks go through `simple_sync_query`, completely bypassing the buffer worker. This could be a problem, leading to unnecessary resource waste:

```erlang
v5.8.11-g0abf56bb(emqx@127.0.0.1)8> emqx_resource_buffer_worker_sup:worker_pids(<<"emqx_authn_http:1040">>).
[<0.8045.0>,<0.8046.0>,<0.8047.0>,<0.8048.0>,<0.8049.0>,
 <0.8050.0>,<0.8051.0>,<0.8052.0>,<0.8053.0>,<0.8054.0>]
```

I will keep the `maybe_resume_resource_workers` line here as it is basically no-op.

terry-xiaoyu:
Created a follow up PR: https://github.com/emqx/emqx/pull/17964

PR #17948, opened 2026-07-14
feat(dynamo): support EC2 IAM role credentials via IMDS

hjianbo:
Release version: 5.10.5

## Summary

Add AWS IAM role credential support to DynamoDB connectors. When both credential fields are omitted, EMQX obtains temporary credentials from an ECS task role or EC2 instance metadata.

Explicit credentials must be provided as a complete access-key/secret-key pair. IAM role credentials remain unresolved in each worker so every DynamoDB request can use erlcloud metadata caching and refresh temporary credentials before expiry.

During a 5.10.4 to 5.10.5 hot upgrade, running DynamoDB connector resources are restarted to adopt the new credential handling; stopped resources remain stopped.

This schema change is backward compatible: existing explicit credential configurations continue to work unchanged.

## PR Checklist


deps.txt
```diff
@@ -23,7 +23,7 @@ emqx_bridge_couchbase: none
 emqx_bridge_datalayers: none
 emqx_bridge_disk_log: none
 emqx_bridge_doris: none
-emqx_bridge_dynamo: none
+emqx_bridge_dynamo: all
```
zmstone: how did this happen?

apps/emqx_bridge/src/schema/emqx_bridge_enterprise.erl
```diff
@@ -173,6 +173,7 @@ fields(bridges) ->
                 hoconsc:map(name, ref(emqx_bridge_dynamo, "config")),
                 #{
                     desc => <<"Dynamo Bridge Config">>,
+                    validator => fun emqx_bridge_dynamo_connector:credentials_validator/1,
```
zmstone: maybe not worth it to introduce static func call dependency (deps.txt).
do this: do not force credentials validation when loading config, move the credential resolution to runtime. see comment below.

apps/emqx_bridge_dynamo/src/emqx_bridge_dynamo_connector_info.erl
```diff
@@ -30,6 +30,7 @@ config_schema() ->
             hoconsc:map(name, hoconsc:ref(emqx_bridge_dynamo, "config_connector")),
             #{
                 desc => <<"DynamoDB Connector Config">>,
+                validator => fun emqx_bridge_dynamo_connector:credentials_validator/1,
```
zmstone, line 33: we can keep this validator to validate API calls

apps/emqx_bridge_dynamo/src/emqx_bridge_dynamo_connector.erl
```diff
@@ -94,14 +93,24 @@ on_start(
         default_port => DefaultPort
     }),
 
+    ClientConfig = #{
+        host => Host,
+        port => Port,
+        scheme => Scheme
+    },
+    ClientConfig1 =
```
zmstone, line 109: validate keys here, return error tuple

apps/emqx_bridge_dynamo/src/emqx_bridge_dynamo_connector_client.erl
```diff
@@ -69,26 +73,46 @@ init(#{
 }) ->
     %% TODO: teach `erlcloud` to to accept 0-arity closures as passwords.
     SecretAccessKey = to_str(emqx_secret:unwrap(Secret)),
-    erlcloud_ddb2:configure(AccessKeyID, SecretAccessKey, Host, Port, Scheme),
```
zmstone, line 72: question: why stop calling this func?

hjianbo: `erlcloud_ddb2:configure/5` only stores the config in the caller process dictionary, and the config-less DDB APIs read it back implicitly. We removed that hidden state intentionally: the worker now keeps an unresolved `AWSConfig` in its gen_server state and passes it explicitly to every DDB call, allowing each request to run `erlcloud_aws:update_config/1` and refresh IAM-role credentials. Calling `configure/5` here would be redundant and would reintroduce the process-dictionary dependency.

zmstone (dismissed):

deps.txt
hjianbo: This was caused by adding a validator in emqx_bridge_enterprise that directly called emqx_bridge_dynamo_connector. The dependency checker therefore required emqx_bridge_dynamo: all. The cross-application validator has now been removed and deps.txt is restored to none; validation stays in the Dynamo-owned connector schema and at runtime.

apps/emqx_bridge/src/schema/emqx_bridge_enterprise.erl
hjianbo: Done. The cross-application validator was removed from the legacy bridge schema. on_start/2 now validates the key pair, and IAM metadata credentials are resolved by the client at runtime. The connector-owned API schema keeps its validator for API requests, as discussed above.

apps/emqx_bridge_dynamo/src/emqx_bridge_dynamo_connector.erl
```diff
@@ -325,6 +342,20 @@ connect(Opts) ->
...
+credentials_validator(Config) ->
+    Invalid =
+        credential_present(<<"aws_access_key_id">>, aws_access_key_id, Config) xor
+            credential_present(<<"aws_secret_access_key">>, aws_secret_access_key, Config),
+    case Invalid of
+        false -> ok;
+        true -> {error, <<"aws_access_key_id and aws_secret_access_key must be provided together">>}
+    end.
+
+credential_present(BinaryKey, AtomKey, Config) ->
```
thalesmg: at this point, the config is an atom key map, cannot be a binary key map.

apps/emqx_bridge_dynamo/src/emqx_bridge_dynamo_connector.erl
```diff
@@ -325,6 +342,20 @@ connect(Opts) ->
...
+    Value0 = maps:get(BinaryKey, Config, maps:get(AtomKey, Config, undefined)),
+    Value = emqx_schema_secret:source(Value0),
```
thalesmg: why source instead of just unwrap?  it'll just return file://... , which could be empty.

apps/emqx_bridge_dynamo/test/emqx_bridge_dynamo_SUITE.erl
```diff
@@ -373,6 +402,263 @@ directly_get_field(Key, Field) ->
...
+            " aws_secret_access_key = \"secret_key\""
+            " }"
+        >>,
+    ?assertMatch(
+        #{<<"aws_access_key_id">> := <<"access_key">>},
+        parse_and_check(ValidLegacyConfig, <<"dynamo">>, <<"test">>)
+    ),
+    LegacyConfigWithoutCredentials =
+        binary:replace(
+            binary:replace(
```
thalesmg: it would be better to use a map as the legacy config and update it, instead of relying on binary replacements.

apps/emqx_bridge_dynamo/test/emqx_bridge_dynamo_SUITE.erl
```diff
@@ -373,6 +402,263 @@ directly_get_field(Key, Field) ->
...
+    ),
+    ?assertEqual(
+        InvalidCredentialsError,
+        emqx_bridge_dynamo_connector:on_start(
+            <<"dynamo-invalid-secret-key-only">>,
+            RuntimeConfig#{aws_secret_access_key => <<"secret_key">>}
+        )
+    ).
+
+assert_invalid_credentials(Check) ->
```
thalesmg: nit: there's a helper fns above, please move it there to keep the module's organization.

plugins/emqx_relup/test/emqx_relup_handler_SUITE.erl
```diff
@@ -178,6 +178,28 @@ t_5105_catalog_covers_saml_xxe_backup_plugin_api_postgresql_oracle_and_lwm2m_fix
     ),
     ?assert(is_before({load_module, emqx_mgmt_data_backup_proto_v2}, AnnounceBPAPI, CodeChanges)).
 
+-doc "The 5.10.4 -> 5.10.5 hop reloads DynamoDB modules before restarting "
+"running DynamoDB connector resources.".
```
thalesmg: nit: it's customary to use triple double-quotes for multi-line docstrings.

apps/emqx_bridge_dynamo/src/emqx_bridge_dynamo_connector.erl
hjianbo: Updated. This callback runs both before emqx_config atomizes checked keys, where the raw config uses binary keys, and in probe/runtime paths, where the config uses atom keys. I normalized atom-keyed credential configs to binary keys and documented both phases, including the outer map invocation from hoconsc:map(name, ...).

apps/emqx_bridge_dynamo/src/emqx_bridge_dynamo_connector.erl
hjianbo: Changed to emqx_secret:unwrap/1 so file-backed secrets are validated using their actual contents. Added coverage for both non-empty and empty secret files.

apps/emqx_bridge_dynamo/test/emqx_bridge_dynamo_SUITE.erl
hjianbo: Changed the legacy config fixture to a map and now derive the variants with maps:remove/2 and maps:without/2.

apps/emqx_bridge_dynamo/test/emqx_bridge_dynamo_SUITE.erl
hjianbo: Moved assert_invalid_credentials/1 to the helper functions section.

plugins/emqx_relup/test/emqx_relup_handler_SUITE.erl
hjianbo: Changed the multi-line documentation to a triple-quoted docstring.

ieQu1 (approved):

zmstone:
also ported to dev-58: https://github.com/emqx/emqx/pull/17948

zmstone:
fixes: https://github.com/emqx/emqx/issues/17963

PR #17949, opened 2026-07-14
fix(agent): validate stream and kv tool formats

### LINKED ISSUE #17942: [6.3.0]  Agent accepts unsupported stream tool format "xml" and reports it as running
### What happened?

  When creating a Stream Agent tool with an unsupported format value such as `xml`, EMQX accepts the configuration and returns HTTP 201
  Created.

  The tool is stored with `format: "xml"` and its runtime status is reported as `running` with no error:

  {
    "status": "running",
    "error": null
  }

  According to the implementation, Stream and KV tools support only `json` and `binary`. Unsupported formats should be rejected or
  reported with an `invalid_format` error.

  The `format` field should accept only the supported values `json` and `binary`.

  When an unsupported value such as `xml` is submitted, the API should reject the request with HTTP 400 and an `invalid_format` error.
  The invalid tool configuration should not be stored, and no running tool or runtime status should be created.

#### 1. Create a regular stream

  curl -u admin:public \
    -X POST \
    http://10.42.6.148:18083/api/v5/streams \
    -d '{
      "name": "pr17662-xml-check",
      "topic_filter": "pr17662/xml/#",
      "is_lastvalue": false
    }'

  Expected result:

  HTTP 200 OK

  #### 2. Create an Agent tool with an unsupported format

  curl -i -u admin:public \
    -X POST \
    http://10.42.6.148:18083/api/v5/plugin_api/emqx_agent/tools \
    -d '{
      "type": "stream__write",
      "id": "pr17662-xml-tool",
      "desc": "invalid xml format",
      "stream": "pr17662-xml-check",
      "format": "xml"
    }'

  Actual result:

  HTTP/1.1 201 Created
  {}

  #### 3. Verify the stored tool configuration

  curl -u admin:public \
    http://10.42.6.148:18083/api/v5/plugin_api/emqx_agent/tools/stream__write/pr17662-xml-tool

  Actual result:

  {
    "desc": "invalid xml format",
    "format": "xml",
    "id": "pr17662-xml-tool",
    "stream": "pr17662-xml-check",
    "type": "stream__write"
  }

  #### 4. Check the runtime tool status

  curl -u admin:public \
    http://10.42.6.148:18083/api/v5/plugin_api/emqx_agent/tools/statuses

  Actual result:

  {
    "stream__write@pr17662-xml-tool": {
      "status": "running",
      "error": null
    }
  }

  ### Expected Behavior

  The format field should accept only:

  json
  binary

  An unsupported value such as xml should be rejected, or the tool should enter a failed state with an invalid_format error.

  ### Actual Behavior

  The API accepts format: "xml", returns HTTP 201, stores the configuration, and reports the tool status as running with no error.

_No response_

### EMQX version

  Environment:
  - EMQX 6.3.x
  - emqx_agent-1.0.0
  - Plugin git ref: `a7e22ba2...`

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

savonarola:
Release version: 6.3.0

Fixes #17942

## Summary

- Require stream and KV tool formats to be explicitly configured as `json` or `binary`, rejecting missing and unsupported values.
- Validate tool-specific configuration through optional tool callbacks (`pre_config_update` alternative).

## PR Checklist

- [ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files -- **unreleased feature**
- [x] Schema changes are backward compatible or intentionally breaking (described in the summary)

### DIFF
--- a/plugins/emqx_agent/src/tools/emqx_agent_tool_kv.erl
+++ b/plugins/emqx_agent/src/tools/emqx_agent_tool_kv.erl
@@ -159,14 +159,16 @@ tool_format(#{<<"type">> := Type, <<"format">> := Format}) when
         (Format =:= <<"json">> orelse Format =:= <<"binary">>)
 ->
     {ok, Format};
+tool_format(#{<<"type">> := Type, <<"format">> := Format}) when
+    Type =:= ?WRITE_TYPE; Type =:= ?READ_TYPE; Type =:= ?READ_ALL_TYPE
+->
+    {error, {invalid_format, Format}};
 tool_format(#{<<"type">> := Type}) when
     Type =:= ?WRITE_TYPE; Type =:= ?READ_TYPE; Type =:= ?READ_ALL_TYPE
 ->
-    {ok, <<"json">>};
+    {error, {missing_field, <<"format">>}};
 tool_format(#{<<"type">> := Type}) when Type =:= ?DEL_TYPE; Type =:= ?CLEAR_TYPE ->
-    {ok, undefined};
-tool_format(#{<<"format">> := Format}) ->
-    {error, {invalid_format, Format}}.
+    {ok, undefined}.
 
 display_name(Desc, ?WRITE_TYPE) -> <<Desc/binary, " — KV Write">>;
 display_name(Desc, ?READ_TYPE) -> <<Desc/binary, " — KV Read">>;
--- a/plugins/emqx_agent/src/tools/emqx_agent_tool_stream.erl
+++ b/plugins/emqx_agent/src/tools/emqx_agent_tool_stream.erl
@@ -235,10 +235,12 @@ tool_format(#{<<"type">> := Type, <<"format">> := Format}) when
         (Format =:= <<"json">> orelse Format =:= <<"binary">>)
 ->
     {ok, Format};
+tool_format(#{<<"type">> := Type, <<"format">> := Format}) when
+    Type =:= ?WRITE_TYPE orelse Type =:= ?READ_TYPE
+->
+    {error, {invalid_format, Format}};
 tool_format(#{<<"type">> := Type}) when Type =:= ?WRITE_TYPE orelse Type =:= ?READ_TYPE ->
-    {ok, <<"json">>};
-tool_format(#{<<"format">> := Format}) ->
-    {error, {invalid_format, Format}}.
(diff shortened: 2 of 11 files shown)


plugins/emqx_agent/src/tools/emqx_agent_tool_kv.erl
```diff
@@ -159,14 +159,16 @@ tool_format(#{<<"type">> := Type, <<"format">> := Format}) when
         (Format =:= <<"json">> orelse Format =:= <<"binary">>)
 ->
     {ok, Format};
+tool_format(#{<<"type">> := Type, <<"format">> := Format}) when
+    Type =:= ?WRITE_TYPE; Type =:= ?READ_TYPE; Type =:= ?READ_ALL_TYPE
+->
+    {error, {invalid_format, Format}};
```
thalesmg, line 165: maybe explicitly list the allowed formats rather than the forbidden ones?


```suggestion
tool_format(#{<<"type">> := Type, <<"format">> := Format}) when
    Type =/= <<"json">> andalso Type =/= <<"binary">>
->
    {error, {invalid_format, Format}};
```

plugins/emqx_agent/src/tools/emqx_agent_tool_stream.erl
```diff
@@ -235,10 +235,12 @@ tool_format(#{<<"type">> := Type, <<"format">> := Format}) when
         (Format =:= <<"json">> orelse Format =:= <<"binary">>)
 ->
     {ok, Format};
+tool_format(#{<<"type">> := Type, <<"format">> := Format}) when
+    Type =:= ?WRITE_TYPE orelse Type =:= ?READ_TYPE
```
thalesmg, line 239: same here (list allowed rather than forbidden)

thalesmg (approved):

plugins/emqx_agent/src/tools/emqx_agent_tool_kv.erl
savonarola: This case serves like a guard 🤔 Without explicit formats, if we add new formats, we don't have to change it.

PR #17957, opened 2026-07-15
fix(rule engine): trigger global rules on all events

### LINKED ISSUE #17953: [6.3.0]$events/client/connack rules do not match when rule_engine.limit_selects_in_namespace = true
### What happened?

$events/client/connack rules do not match when rule_engine.limit_selects_in_namespace = true

When a rule belongs to global instead of any specific namespace, the event $events/client/connack should also be matched.

### Steps to Reproduce

  1. Create a rule:

  SELECT *
  FROM "$events/client/connack"

  2. Configure the rule with a republish action.
  3. Establish a new MQTT connection.
  4. Observe the rule metrics.

  ### Actual Result

  The MQTT client receives a successful CONNACK:CONNACK (0)

  However, the rule is not matched:

  {
    "matched": 0,
    "actions.total": 0
  }

  The equivalent rule using:

  SELECT *
  FROM "$events/client/connected"

  is matched successfully under the same configuration.

  ### Expected Result

  $events/client/connack should be matched in the same way as $events/client/connected when the client belongs to the global namespace.

  After changing: rule_engine.limit_selects_in_namespace = false

  the client.connack rule matched successfully:

  {
    "matched": 1,
    "actions.total": 1,
    "actions.success": 1
  }

  ### Suspected Cause

  client.connected resolves the namespace from ClientInfo and falls back to the global namespace when no
  namespace is present.

  client.connack resolves the namespace through the hook context. The hook context contains namespace =>
  undefined, which is used directly instead of falling back to the global namespace. As a result, the global
  rule is not found when namespace-limited rule selection is enabled.

_No response_

### EMQX version

6.3.0.

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

thalesmg:
Fixes https://github.com/emqx/emqx/issues/17953

The fix in `98d29c753c921dc1584b28880047bbbd012ca3e5` only addressed `message.publish` events, with other events not triggering global rules when `limit_selects_in_namespace = true`.

Release version:
6.0.5, 6.1.4, 6.2.3, 6.3.0

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/mix.exs
+++ b/apps/emqx_rule_engine/mix.exs
@@ -5,7 +5,7 @@ defmodule EMQXRuleEngine.MixProject do
   def project do
     [
       app: :emqx_rule_engine,
-      version: "6.0.3",
+      version: "6.0.4",
       build_path: "../../_build",
       erlc_options: UMP.strict_erlc_options(),
       erlc_paths: UMP.erlc_paths(),
--- a/apps/emqx_rule_engine/src/emqx_rule_engine.erl
+++ b/apps/emqx_rule_engine/src/emqx_rule_engine.erl
@@ -32,6 +32,7 @@
     get_rules_from_all_namespaces/0,
     get_rules_for_topic/1,
     get_enriched_rules_with_matching_event/2,
+    get_enriched_rules_with_matching_event_including_global/2,
     get_enriched_rules_with_matching_event_all_namespaces/1,
     get_rule_ids_by_action/1,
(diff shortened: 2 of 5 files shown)


qzhuyan (approved):

zmstone:
NOTE: 6.1.3 and 6.2.2 are already released . so I updated the description, changed release versions to 6.0.5, 6.1.4, 6.2.3, 6.3.0

PR #17961, opened 2026-07-15
fix(kafka, pulsar connectors): switch to `connecting` on health check timeouts

### LINKED ISSUE #17916: Finite `health_check_timeout` defeats Kafka producer internal buffering during hang-style outages; auto-retry kills wolff producers (regression of #14048)
### What happened?

A customer reported that their Kafka producer action, configured with a disk buffer specifically to survive Kafka outages, failed **all** messages during an outage test instead of buffering them. Their action metrics showed the signature clearly: `matched: 50000, success: 1331, failed: 48669, queuing: 0, dropped: 0`. They simulated the outage by blocking network connectivity to Kafka (packet drop, not connection refused). They had also set the action's `resource_opts.health_check_timeout` to 10s, but the same happens with the default 60s — the outage just needs to last a couple of minutes.

I reproduced this end-to-end on `emqx/emqx-enterprise:6.2.2` (details below). The cascade is:

1. Kafka becomes unreachable in a *hang* fashion (firewall DROP / security group / VM freeze / k8s network partition). Established connections go silent; new TCP connects hang until `connect_timeout`.
2. Initially everything works as designed: connector and action go `connecting`, messages keep flowing into wolff's replayq, `queuing` grows. 👍
3. After `request_timeout` (= `max(2 * metadata_request_timeout, 30s)`, see `emqx_bridge_kafka_impl_producer:on_start/2`) the wolff partition producers declare their connections dead and start requesting reconnects from the shared `wolff_client` process. Each such request makes `wolff_client` do a **synchronous, serialized** metadata refresh: a request timeout on the dead metadata connection, then fresh connects to each bootstrap host at up to `connect_timeout` (5s) each.
4. EMQX health checks (`on_get_status`, `on_get_channel_status`) are `gen_server` calls into that same busy `wolff_client` process **with an `infinity` call timeout** (`wolff_client:safe_call/2`). They queue behind the reconnect storm and stall for longer than any finite `health_check_timeout`.
5. `emqx_resource_manager` maps the timed-out health check to `{disconnected, resource_health_check_timed_out}` (`worker_resource_health_check/1`, `worker_channel_health_check/3`).
6. Channel `disconnected` → `emqx_resource_buffer_worker` rejects every message with `{unrecoverable_error, <<"channel: ... not operational">>}` → the action `failed` counter increments and **nothing reaches the buffer**.
7. Connector `disconnected` → `auto_retry` → `start_resource` → `emqx_resource:call_start/3` → `clean_allocated_resources/1` → Kafka `on_stop` → **the wolff producers and client are killed** (`wolff:stop_and_delete_supervised_producers`, `stop_and_delete_supervised_client`). With `buffer.mode = memory` (the default) this **destroys all data buffered so far**. With `disk`, the replayq segments survive and are re-sent after recovery (with lost ack callbacks, so `success` under-counts; delivery is at-least-once).
8. Every subsequent retry fails (`start_resource_failed`, Kafka still unreachable), so the action keeps failing all traffic until Kafka returns.

This directly contradicts the driver's own design. PR #14048 (EMQX-13338, released in 5.8.2) is titled *"do not consider action disconnected once created"* precisely because the resource manager killing wolff producers loses buffered data — and the code comments in `emqx_bridge_kafka_impl_producer` (`on_start`, `on_get_status`, `on_get_channel_status`) all warn about this. The `resource_opts.health_check_timeout` option introduced in 5.10.0 re-opened a path to `disconnected` at the resource-manager level that bypasses the driver's guarantee entirely. Before 5.10 this failure mode could not happen (no health check timeout existed, which is equivalent to `infinity`).

Note the asymmetry that makes this easy to miss in tests: if Kafka dies with a fast *connection refused* (e.g. `docker stop`), the health check returns an error quickly, status degrades only to `connecting`, and buffering works. Only hang-style outages (packet drop) trigger the cascade — which is unfortunately what real-world network partitions, security-group changes and abrupt VM/node losses look like.

Per #14048: once a Kafka producer action has been created, a Kafka outage — regardless of how the network fails — should leave the action in `connecting`, keep accepting messages into wolff's buffer (up to the configured per-partition limit), and drain the buffer after Kafka recovers. The wolff producers must not be killed while they hold buffered data.

docker-compose with Kafka in KRaft mode, EMQX Enterprise 6.2.2, and an iptables sidecar sharing Kafka's netns to simulate the network partition:

```yaml
services:
  kafka:
    image: apache/kafka:3.9.0
    environment:
      KAFKA_NODE_ID: 1
      KAFKA_PROCESS_ROLES: broker,controller
      KAFKA_LISTENERS: PLAINTEXT://:9092,CONTROLLER://:9093
      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://kafka:9092
      KAFKA_CONTROLLER_LISTENER_NAMES: CONTROLLER
      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: CONTROLLER:PLAINTEXT,PLAINTEXT:PLAINTEXT
      KAFKA_CONTROLLER_QUORUM_VOTERS: 1@localhost:9093
      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
  blackhole:
    image: alpine:3.20
    network_mode: "service:kafka"
    cap_add: [NET_ADMIN]
    command: sh -c "apk add --no-cache iptables >/dev/null && sleep infinity"
  emqx:
    image: emqx/emqx-enterprise:6.2.2
    ports: ["18083:18083"]
```

1. Create a topic with 3 partitions; create a `kafka_producer` connector (`bootstrap_hosts = kafka:9092`) and an action with `parameters.buffer.mode = disk`, `query_mode = async`; add a rule `SELECT * FROM "t/repro"` → the action. To reproduce the customer's setting use `resource_opts = {health_check_interval = 15s, health_check_timeout = 10s}` on the action — but the default 60s fails the same way, just a bit later.
2. Publish continuously (~5 msg/s) to `t/repro`; confirm `success` tracks `matched`.
3. Blackhole Kafka:
   ```sh
   docker compose exec blackhole iptables -A INPUT  ! -i lo -p tcp --dport 9092 -j DROP
   docker compose exec blackhole iptables -A OUTPUT ! -o lo -p tcp --sport 9092 -j DROP
   ```
4. Watch action status + metrics.

Observed timeline (5 msg/s, timestamps relative to the DROP):

| t | status | metrics |
|---|--------|---------|
| +10s…+70s | `connecting` (both levels) | `queuing` grows steadily, `failed: 0` — **buffering works as designed** |
| ~+2min | action flips `disconnected` | every message → `failed` ("channel not operational"), `queuing` frozen |
| ~+2min20s | connector flips `disconnected` | **wolff producers + client killed** (`wolff_producers_sup` `shutdown_error` in logs), then `start_resource_failed` every 15s |
| heal | back to `connected` | disk replayq drains to Kafka; the 506 messages failed during the window are lost permanently |

Control experiment: same run with `resource_opts.health_check_timeout = infinity` on **both** connector and action — 5-minute blackhole, status stayed `connecting` the whole time, `queuing` grew to 1181, `failed: 0`, and after healing the queue fully drained (verified against topic end offsets). None of the failure signatures appear in the logs.

Key code locations (release-62 / master):

- `emqx_resource_manager:worker_resource_health_check/1` and `worker_channel_health_check/3` — `exit:timeout` → `{?status_disconnected, <<"resource_health_check_timed_out">>}`.
- `emqx_resource:call_start/3` → `clean_allocated_resources/1` → `call_stop` → Kafka `on_stop` kills producers/client on the very first `auto_retry` after entering `disconnected`.
- `wolff_client:safe_call/2` — `gen_server:call(Pid, Call, infinity)`; `get_metadata/6` probes bootstrap hosts sequentially, each bounded only by `connect_timeout`.
- `emqx_bridge_kafka_impl_producer:on_get_status/2` / `on_get_channel_status/3` — deliberately never return `disconnected` (comments explain why), i.e. the *only* remaining source of `disconnected` after creation is the health-check-timeout mapping.

Workaround (validated): set `resource_opts.health_check_timeout = infinity` on **both** the Kafka connector and the action. `infinity` is accepted since the field was introduced in 5.10.0. Setting it on only one level is not enough — either level flipping to `disconnected` breaks buffering (the channel via fail-fast, the connector via the producer kill).

Possible directions for a real fix (in rough order of preference):

1. For resources that manage their own buffers (`simple_async_internal_buffer` / bypass query modes), map a health-check timeout to `connecting` instead of `disconnected` — a slow health check is not evidence that the resource is dead, and for these drivers the cost of a false `disconnected` is data loss.
2. Make the Kafka health check non-blocking: read cached connection state instead of issuing synchronous metadata refreshes through the (possibly busy) `wolff_client` process, so it completes in bounded time regardless of outage type.
3. At minimum, don't `clean_allocated_resources` on `auto_retry` when the previous incarnation of the resource is still running and holds buffered data — killing wolff producers destroys the memory-mode buffer outright.

There are similar field reports for other bridge types showing `resource_health_check_timed_out` together with `clean_allocated_resources` churn in the same logs, so the resource-manager side of this is not Kafka-specific — Kafka (and other internal-buffer drivers like Pulsar) just pay the highest price for it.

### EMQX version

<details>

Reproduced on `emqx/emqx-enterprise:6.2.2` (docker). The interaction exists in every version since 5.10.0 (`resource_opts.health_check_timeout` introduced); the relevant code is unchanged on `release-62`, `release-63` and `master` as of this report.

</details>

### OS version

<details>

Docker Desktop 29.4.0 on macOS (containers: official images, Linux). Not OS-specific.

</details>

### Log files

<details>

```
08:27:55 [warning] msg: health_check_failed, status: connecting, resource_id: connector:kafka_producer:repro
08:28:15 [warning] msg: failed_to_fetch_metadata, topic: repro-topic, errors: [#{reason => connection_timed_out,host => <<"kafka:9092">>}]   (repeats every ~5s)
08:28:48 [error] msg: producer_is_still_disconnected_after_retry, topic: repro-topic, partition: 2, attempts: 10
08:29:40 [error] msg: unrecoverable_resource_error, reason: {unrecoverable_error,<<"channel: \"action:kafka_producer:repro:connector:kafka_producer:repro\" not operational">>}
08:30:00 [error] Supervisor: {local,wolff_producers_sup}. Context: shutdown_error. Reason: killed. Offender: pid=<0.5760.0>.
08:30:00 [error] crasher: initial call: wolff_producer:init/1, exit: {killed,...}   (one per partition producer)
08:30:05 [error] Supervisor: {local,wolff_client_sup}. Context: shutdown_error. Reason: killed.
08:30:10 [warning] msg: start_resource_failed, reason: [#{reason => connection_timed_out,host => <<"kafka:9092">>}], resource_id: connector:kafka_producer:repro   (repeats every 15s until Kafka returns)
```

</details>

thalesmg:
Fixes https://github.com/emqx/emqx/issues/17916

Release version:
6.0.4, 6.1.4, 6.2.3, 6.3.0

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource.erl
+++ b/apps/emqx_resource/src/emqx_resource.erl
@@ -82,6 +82,8 @@
     %% get the callback mode of a specific module
     get_callback_mode/1,
     get_resource_type/1,
+    get_resource_health_check_timeout_status/1,
+    get_channel_health_check_timeout_status/1,
     get_callback_mode/2,
     get_query_opts/2,
     %% start the instance
@@ -163,7 +165,9 @@
     query_mode/1,
     on_format_query_result/1,
     callback_mode/1,
-    query_opts/1
+    query_opts/1,
+    resource_health_check_timeout_status/0,
+    channel_health_check_timeout_status/0
 ]).
 
 -type namespace() :: binary().
@@ -255,6 +259,18 @@
 
 %% Used for tagging log entries.
 -callback resource_type() -> atom().
+
+%% When a health check times out, set this status for the resource.
+%%
+%% Most modules should not tamper with this; exceptions are those which must not destroy
+%% their queues (kafka and pulsar, currently)
+-callback resource_health_check_timeout_status() -> health_check_status().
+
+%% When a health check times out, set this status for the channel.
+%%
+%% Most modules should not tamper with this; exceptions are those which must not destroy
+%% their queues (kafka and pulsar, currently)
+-callback channel_health_check_timeout_status() -> health_check_status().
 
 -elvis([{elvis_style, no_match_in_condition, disable}]).
 
@@ -532,6 +548,22 @@ get_callback_mode(Mod) ->
 -spec get_resource_type(module()) -> resource_type().
 get_resource_type(Mod) ->
     Mod:resource_type().
+
+get_resource_health_check_timeout_status(Mod) ->
+    case erlang:function_exported(Mod, resource_health_check_timeout_status, 0) of
+        true ->
+            Mod:resource_health_check_timeout_status();
+        _ ->
+            ?status_disconnected
+    end.
+
+get_channel_health_check_timeout_status(Mod) ->
+    case erlang:function_exported(Mod, channel_health_check_timeout_status, 0) of
+        true ->
+            Mod:resource_health_check_timeout_status();
+        _ ->
+            ?status_disconnected
+    end.
 
 -spec get_callback_mode(module(), resource_state()) -> callback_mode() | undefined.
 get_callback_mode(Mod, State) ->
--- a/apps/emqx_resource/src/emqx_resource.erl
+++ b/apps/emqx_resource/src/emqx_resource.erl
@@ -560,7 +560,7 @@ get_resource_health_check_timeout_status(Mod) ->
 get_channel_health_check_timeout_status(Mod) ->
     case erlang:function_exported(Mod, channel_health_check_timeout_status, 0) of
         true ->
-            Mod:resource_health_check_timeout_status();
+            Mod:channel_health_check_timeout_status();
         _ ->
             ?status_disconnected
     end.
--- a/apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_SUITE.erl
+++ b/apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_SUITE.erl
@@ -375,3 +375,9 @@ t_dynamic_topics(TCConfig) ->
 
 t_disallow_disk_mode_for_dynamic_topic(TCConfig) ->
     emqx_bridge_kafka_action_SUITE:?FUNCTION_NAME(TCConfig).
+
+t_resource_health_check_timeout_status(TCConfig) when is_list(TCConfig) ->
+    emqx_bridge_kafka_action_SUITE:?FUNCTION_NAME(TCConfig).
+
+t_channel_health_check_timeout_status(TCConfig) when is_list(TCConfig) ->
+    emqx_bridge_kafka_action_SUITE:?FUNCTION_NAME(TCConfig).
--- a/apps/emqx_bridge_confluent/test/emqx_bridge_confluent_producer_SUITE.erl
+++ b/apps/emqx_bridge_confluent/test/emqx_bridge_confluent_producer_SUITE.erl
@@ -367,3 +367,9 @@ t_dynamic_topics(TCConfig) ->
 
 t_disallow_disk_mode_for_dynamic_topic(TCConfig) ->
     emqx_bridge_kafka_action_SUITE:?FUNCTION_NAME(TCConfig).
+
+t_resource_health_check_timeout_status(TCConfig) when is_list(TCConfig) ->
+    emqx_bridge_kafka_action_SUITE:?FUNCTION_NAME(TCConfig).
+
+t_channel_health_check_timeout_status(TCConfig) when is_list(TCConfig) ->
+    emqx_bridge_kafka_action_SUITE:?FUNCTION_NAME(TCConfig).
--- a/apps/emqx_bridge_kafka/mix.exs
+++ b/apps/emqx_bridge_kafka/mix.exs
@@ -5,7 +5,7 @@ defmodule EMQXBridgeKafka.MixProject do
   def project do
     [
       app: :emqx_bridge_kafka,
-      version: "6.0.3",
+      version: "6.0.4",
       build_path: "../../_build",
       erlc_options: UMP.erlc_options(),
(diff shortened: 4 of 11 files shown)


apps/emqx_resource/src/emqx_resource.erl
```diff
@@ -533,6 +549,22 @@ get_callback_mode(Mod) ->
...
+        true ->
+            Mod:resource_health_check_timeout_status();
+        _ ->
+            ?status_disconnected
+    end.
+
+get_channel_health_check_timeout_status(Mod) ->
+    case erlang:function_exported(Mod, channel_health_check_timeout_status, 0) of
+        true ->
+            Mod:resource_health_check_timeout_status();
```
zmstone:
```suggestion
            Mod:channel_health_check_timeout_status();
```

zmstone (approved):

PR #17964, opened 2026-07-16
perf(auth): don't spawn buffer workers for authenticator/authorizer resources

terry-xiaoyu:
Authenticator and authorizer resources always query through `emqx_resource:simple_sync_query/2', which calls the resource callback directly and bypasses the buffer worker processes (see the note in `emqx_resource_buffer_worker:simple_sync_query/3').  Spawning a buffer worker pool for them only creates idle workers that are never used.

Set `spawn_buffer_workers => false' in the authn/authz default resource opts, the same way connectors already do in `emqx_connector_resource'.

Release version:

## PR Checklist


zmstone (approved):

terry-xiaoyu:
No such problem in emqx newer than 6.x.

PR #17939, opened 2026-07-14
test(license): stabilize session HWM history checks

hjianbo:
Release version: 6.0.4, 6.1.3, 6.2.2
Fix https://github.com/emqx/emqx/actions/runs/29067650276/job/86716415411?pr=17921

## Summary

Stabilize session HWM history API checks by asserting the expected historical periods in order. The periodic resource sampler can legitimately add a current-day record while the suite runs.

## PR Checklist

- [x] The changes are covered with new or existing tests
- [ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [ ] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #17968, opened 2026-07-16
test(license): de-flake session hwm history http api suite

zmstone:
Release version: 6.0.3

## Summary

Test-only change to de-flake `emqx_license_http_api_SUITE`'s session HWM
history cases.

The `t_session_hwm_history_*` cases seed a fixed set of historical rows into
`emqx_license_session_hwm` and then assert on the exact response of the
`session_hwm_history` HTTP API. Meanwhile the background sampler
(`emqx_license_resources`) fires a periodic `update_resources` tick that calls
`emqx_license_session_hwm:observe/2` with the current wall-clock time and the
live session count. When such a tick landed between setup and the GET, it wrote
an extra `{period => today, high_watermark => 0}` row, which — sorted newest
first — was prepended to the response and broke the exact-match assertions.

The fix suspends `emqx_license_resources` in `session_hwm_setup/1` (guarded by
`whereis/1`) and resumes it in `session_hwm_teardown/1`, so no live row can be
written during the assertions. Suspending only pauses the timer-driven resource
refresh; nothing else depends on a synchronous reply from that process. With the
table now deterministic, the assertions are restored to strict exact-list
matches (dropping the earlier period-filtering workaround), so an unexpected row
would now fail the test instead of being silently ignored.

Product behavior is unchanged.

## PR Checklist

### DIFF
--- a/apps/emqx_license/test/emqx_license_http_api_SUITE.erl
+++ b/apps/emqx_license/test/emqx_license_http_api_SUITE.erl
@@ -378,6 +378,12 @@ validate_setting(Res, ExpectLow, ExpectHigh, DynMax) ->
 %%------------------------------------------------------------------------------
 
 session_hwm_setup(Config) ->
+    %% Suspend the background license sampler for the duration of the test so it
+    %% cannot write a live current-day row into the table while we assert on a
+    %% fixed set of seeded rows. Without this, a sampler tick landing between
+    %% setup and the GET prepends an extra {period => today, high_watermark => 0}
+    %% row and flakes the exact-match assertions.
+    ok = suspend_license_sampler(),
     %% Pin timezone so the period labels are host-independent.
     OrigTz = emqx_config:get([license, high_watermark_timezone], system),
     emqx_config:put([license, high_watermark_timezone], <<"+00:00">>),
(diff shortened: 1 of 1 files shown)


keynslug:
Note that there's already #17939 merged.

zmstone:
> Note that there's already #17939 merged.

this fix is on top of it

apps/emqx_license/test/emqx_license_http_api_SUITE.erl
```diff
@@ -378,6 +378,12 @@ validate_setting(Res, ExpectLow, ExpectHigh, DynMax) ->
 %%------------------------------------------------------------------------------
 
 session_hwm_setup(Config) ->
+    %% Suspend the background license sampler for the duration of the test so it
+    %% cannot write a live current-day row into the table while we assert on a
+    %% fixed set of seeded rows. Without this, a sampler tick landing between
+    %% setup and the GET prepends an extra {period => today, high_watermark => 0}
+    %% row and flakes the exact-match assertions.
+    ok = suspend_license_sampler(),
```
keynslug, line 386: In my opinion #17939 is still a better approach. Why do you think it needs changing?
1. It does not depend on race avoidance and obscure implementation details.
2. Note that `suspend_license_sampler()` may be called when it's already too late, it's not deterministic that test suite starts executing `t_session_hwm_history_default_monthly` in time.

Issue #17483, opened 2026-06-04
HTTP connector enable/disable does not re-apply SSRF validation for existing denied targets


yanzhiemq:
### Summary

After `rule_engine.ssrf` is enabled, existing HTTP connectors that point to a denied target can still be disabled and re-enabled successfully. The enable/disable path does not appear to re-apply SSRF validation for HTTP connector `url` fields.

This differs from MQTT connectors, where the same policy-transition scenarios are rejected correctly on later enable/disable operations.

### Affected Area

- Rule Engine / Data Integration
- HTTP connector
- `rule_engine.ssrf`
- `PUT /api/v5/connectors/{id}/enable/{true|false}`

### Found In

`6.1.2-rc.3-g8554445f`

### Reproduction Scenario 1: Connector Created While SSRF Is Disabled

1. Disable SSRF:

   ```json
   {
     "ssrf": {
       "enable": false,
       "allow_cidrs": [],
       "deny_cidrs": ["127.0.0.0/8"],
       "deny_hosts": []
     }
   }
   ```

2. Create an HTTP connector with a loopback target:

   ```json
   {
     "type": "http",
     "name": "manual_ssrf_toggle",
     "enable": true,
     "url": "http://127.0.0.1:18083",
     "headers": {},
     "pool_type": "hash",
     "pool_size": 1,
     "resource_opts": {
       "start_after_created": false,
       "start_timeout": "5s",
       "health_check_interval": "15s"
     }
   }
   ```

3. Enable SSRF with default deny CIDRs, including `127.0.0.0/8`.

4. Disable the connector:

   ```http
   PUT /api/v5/connectors/http%3Amanual_ssrf_toggle/enable/false
   ```

5. Re-enable the connector:

   ```http
   PUT /api/v5/connectors/http%3Amanual_ssrf_toggle/enable/true
   ```

### Actual Result

The HTTP connector can be re-enabled successfully:

```text
HTTP 204
```

### Expected Result

The disable or re-enable operation should re-apply the active SSRF policy and reject the connector because `127.0.0.1` matches `127.0.0.0/8`.

### Reproduction Scenario 2: Connector Created While Target Is Allowlisted

1. Enable SSRF and allow loopback:

   ```json
   {
     "ssrf": {
       "enable": true,
       "allow_cidrs": ["127.0.0.0/8"],
       "deny_cidrs": ["127.0.0.0/8"],
       "deny_hosts": []
     }
   }
   ```

[...]

PR #17970, opened 2026-07-16
fix(connector): limit SSRF checks to http/mqtt connectors at create/update

### LINKED ISSUE #17940: [6.3.0]When SSRF protection is enabled, residual connectors remain after deleting a webhook.
### What happened?

When SSRF protection is enabled, residual connectors remain after deleting a webhook.

When SSRF protection is enabled, deletion should be prohibited if a matching entry is found.

1. Create a webhook with URL: [http://127.0.0.1]

Image

2. Enable SSRF protection

Image

3. Click Delete on the webhook page

Image

4. A 500 error is displayed, but the webhook is deleted successfully
5. Check connectors: the connector remains residual, while the action and rule are removed

https://github.com/user-attachments/assets/f39449d3-87bf-409f-84ca-de37d15f87aa

Image

_No response_

### EMQX version

6.3.0

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

zmstone:
Release version: 6.0.4

## Summary

Fixes #17940.

With SSRF protection enabled, deleting a webhook whose connector's address was denied by the policy returned a 500 and left a residual connector while its action and rule were removed. The trigger was an *unrelated* connector: EMQX re-validates the whole `connectors` config subtree on every change, and ~20 connector schemas (greptimedb, kafka, mysql, influxdb, …) enforced the SSRF check on their `server`/`servers` field via the shared `emqx_schema:servers_sc/2`. Once a stored sibling address became denied, that whole-subtree re-validation threw, so creating, updating, enabling **or deleting** any connector could fail.

### Change

Rather than police SSRF across every connector type and operation, this scopes it back:

- Remove the `ssrf_check` option from the ~20 connector host-option definitions, so the shared `servers_sc`/`parse_server` schema check no longer enforces SSRF. Whole-config re-validation therefore never fails on a now-denied sibling, and deletes are unaffected.
- SSRF is now enforced only for **HTTP** (`url`) and **MQTT** (`server`) connectors, via `emqx_connector_resource:parse_confs/3`, which runs when a connector is **created or updated**. Creating/updating such a connector with a denied address is still rejected.
- Enabling/disabling and deleting connectors are no longer subject to the policy. This reverts the re-enable re-validation previously added for #17483.

### Behavior change / scope

- Connector types other than HTTP/MQTT (greptimedb, kafka, mysql, influxdb, cassandra, pulsar, tdengine, …) are **no longer** SSRF-checked.
- Re-enabling a connector whose stored address is now denied is allowed again.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_azure_event_hub/src/emqx_bridge_azure_event_hub.erl
+++ b/apps/emqx_bridge_azure_event_hub/src/emqx_bridge_azure_event_hub.erl
@@ -485,4 +485,4 @@ server_name_indication_converter(SNI, _HoconOpts) ->
     SNI.
 
 host_opts() ->
-    #{default_port => 9093, ssrf_check => true}.
+    #{default_port => 9093}.
--- a/apps/emqx_bridge_bigquery/src/emqx_bridge_bigquery_impl.erl
+++ b/apps/emqx_bridge_bigquery/src/emqx_bridge_bigquery_impl.erl
@@ -109,7 +109,7 @@ on_start(ConnResId, ConnConfig0) ->
     ),
     {Transport, HostPort} = emqx_bridge_gcp_pubsub_client:get_transport(bigquery),
     #{hostname := Host, port := Port} = emqx_schema:parse_server(HostPort, #{
-        default_port => 443, ssrf_check => true
+        default_port => 443
     }),
     ConnConfig = ConnConfig1#{
         jwt_opts => #{
--- a/apps/emqx_bridge_cassandra/mix.exs
+++ b/apps/emqx_bridge_cassandra/mix.exs
@@ -5,7 +5,7 @@ defmodule EMQXBridgeCassandra.MixProject do
   def project do
     [
       app: :emqx_bridge_cassandra,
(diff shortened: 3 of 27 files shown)


thalesmg (approved):
I wonder if instead we could leave the checks on all connectors, and fix the validation somehow.  Maybe extend hocon so that a check has some extra context (e.g.: a validator receives this extra context that can contain anything, such as "this is an update" or "enable check X"), or, more quick and dirty, thread that down using `try put(...), ... after erase(...) end`.

zmstone:
> I wonder if instead we could leave the checks on all connectors, and fix the validation somehow. Maybe extend hocon so that a check has some extra context (e.g.: a validator receives this extra context that can contain anything, such as "this is an update" or "enable check X"), or, more quick and dirty, thread that down using `try put(...), ... after erase(...) end`.

I'd go with the straightforward solution to implement checks on all connectors.

Too much magics for hocon schema makes it hard to maint

Issue #17913, opened 2026-07-09
[6.3.0] MQTT v3 CONNECT succeeds when password is provided without username


Le9Wang:
### What happened?

In EMQX Enterprise 6.3.0-beta.1, an MQTT v3 client can connect successfully when a password is provided without a username. The failing test case is `test_v3_connect_password_on_no_username`, which expects the connection to be rejected according to the protocol requirement that it is not valid to supply a password without supplying a user name. The issue also exists on mqtt v3.1.1.

Observed test failure:

```python
mqtt_client.username = None
mqtt_client.password = "Abcdefg12345"
response = await mqtt_client.connect().wait(timeout_ms=2_000)
assert response != "Success", f"Connection should fail, but succeeded with reason {response}."
```

Failure output:

```text
E       AssertionError: Connection should fail, but succeeded with reason Success.
E       assert ReasonCode(Connack, 'Success') != 'Success'

regs/tests_protocol/src/tests_protocol/mqtt/mqttv3/test_v3_connect.py:187
```

### What did you expect to happen?

The broker should reject the MQTT v3 CONNECT request when a password is supplied but the username is omitted.

### How can we reproduce it (as minimally and precisely as possible)?

1. Use EMQX Enterprise 6.3.0-beta.1.
2. Run the MQTT v3 protocol test case `test_v3_connect_password_on_no_username`.
3. Configure the client with `username = None` and `password = "Abcdefg12345"`.
4. Attempt to connect.
5. Observe that the broker returns `Success` instead of rejecting the connection.

Minimal reproduction snippet:

```python
mqtt_client.username = None
mqtt_client.password = "Abcdefg12345"
response = await mqtt_client.connect().wait(timeout_ms=2_000)
```

### Anything else we need to know?

Issue title should be prefixed with `[6.3.0]`.

This failure was observed in the following environment:
- Product: EMQX Enterprise
- Test priority: P1
- Test file: `regs/tests_protocol/src/tests_protocol/mqtt/mqttv3/test_v3_connect.py`
- Test name: `test_v3_connect_password_on_no_username`

### EMQX version

<details>

```console
sysdescr  : EMQX Enterprise
version   : 6.3.0-beta.1-gc8404573
datetime  : 2026-07-09T06:16:15.734072919+00:00
uptime    : 4 hours, 3 minutes, 6 seconds
```

</details>

### Log files

<details>

</details>

savonarola:
[...]

PR #17971, opened 2026-07-16
fix(mqtt): reject v3.1.1 CONNECT with password but no username

zmstone:
Release version: 6.3.0

## Summary

Fixes https://github.com/emqx/emqx/issues/17913.

Enforce **[MQTT-3.1.2-22]** ("If the User Name Flag is set to 0, the
Password Flag MUST be set to 0") for MQTT v3.1.1 (protocol level 4)
unconditionally, i.e. regardless of `mqtt.strict_mode`.

Change in `apps/emqx/src/emqx_frame.erl`,
`validate_connect_password_flag/4`: the v3.1.1 clause now fires
independently of the strict-mode argument. It throws exactly when the
username flag is clear and the password flag is set; the other three flag
combinations parse cleanly. v3.1 (level 3) is unchanged (its spec does not
couple the flags), and v5 is unchanged.

### Context on dev-63 / strict mode

On this branch `mqtt.strict_mode` already defaults to `true`, so a CONNECT
that sets the password flag while clearing the username flag is already
rejected by default. This change closes the remaining gap: the check was
previously gated on strict mode, so a listener with `strict_mode = false`
would still accept such packets. With this change the [MQTT-3.1.2-22]
violation is rejected in both modes.

Note: a CONNECT that sets the username flag with an empty username string
plus a password remains protocol-legal and is still accepted; that is not
a [MQTT-3.1.2-22] violation.

Tests (`emqx_frame_SUITE`): `t_invalid_password_flag` now asserts rejection
in both non-strict and strict mode; added `t_valid_password_flag_v4`
(password + username still parses) and `t_password_flag_v3_no_username`
(v3.1 still accepted) to confirm the fix is scoped. Full suite: 58 ok.

Note: the older v6 lines (dev-60/61/62) still default `strict_mode` to
`false`; if this behavior tightening is wanted there too, it can be
backported. This PR targets dev-63 (the line the report was filed against,
6.3.0-beta.1).

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_frame.erl
+++ b/apps/emqx/src/emqx_frame.erl
@@ -1344,16 +1344,15 @@ validate_connect_will(false, WillRetain, _) when WillRetain -> ?PARSE_ERR(invali
 validate_connect_will(_, _, _) -> ok.
 
 -compile({inline, [validate_connect_password_flag/4]}).
-%% MQTT-v3.1
-%% Username flag and password flag are not strongly related
+%% MQTT-v3.1: username and password flags are not strongly related; no check.
 %% https://public.dhe.ibm.com/software/dw/webservices/ws-mqtt/mqtt-v3r1.html#connect
-validate_connect_password_flag(true, ?MQTT_PROTO_V3, _, _) ->
+validate_connect_password_flag(_, ?MQTT_PROTO_V3, _, _) ->
     ok;
-%% MQTT-v3.1.1-[MQTT-3.1.2-22]
-validate_connect_password_flag(true, ?MQTT_PROTO_V4, UsernameFlag, PasswordFlag) ->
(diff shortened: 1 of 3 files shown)


keynslug (approved):

zmstone:
Closing: on reflection we are keeping bug-for-bug compatibility here.

Accepting a v3.1.1 CONNECT with the password flag set but the username flag clear is nonconformant per [MQTT-3.1.2-22], but it is not a security concern, and hard-rejecting it would block some old buggy clients from upgrading. On dev-63 `mqtt.strict_mode` already defaults to `true`, which rejects this shape by default for anyone who wants the strict behavior, while lenient mode keeps the compatible path. That is the right tradeoff, so we will not enforce it unconditionally.

PR #17974, opened 2026-07-16
fix(logging): redact raw MQTT packet data by default

savonarola:
Release version: 6.0.4

Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/269 

## Summary
- Redact raw MQTT packet data in connection and parser-error logs by default.
- Add listener-scoped `allow_log_packet_data_from` CIDR allowlists for temporary trusted-client diagnostics.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_packet_data_logger.erl
+++ b/apps/emqx/src/emqx_packet_data_logger.erl
@@ -1,0 +1,46 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+-module(emqx_packet_data_logger).
+
+-export([add_packet_data/5]).
+
+-define(REDACTED, <<"******">>).
+
+-spec add_packet_data(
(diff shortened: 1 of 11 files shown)


apps/emqx/src/emqx_packet_data_logger.erl
```diff
@@ -0,0 +1,46 @@
...
+add_raw_packet_data(Log, Key, Data, hex) ->
+    Log#{Key => binary_to_list(binary:encode_hex(Data)), type => "hex"}.
+
+add_redacted_packet_data(Log, Key, raw) ->
+    Log#{Key => ?REDACTED};
+add_redacted_packet_data(Log, Key, hex) ->
+    Log#{Key => ?REDACTED, type => "hidden"}.
+
+is_allowed(Channel) ->
+    ClientInfo = emqx_channel:info(clientinfo, Channel),
```
zmstone, line 33: maybe set a process dict flag at init state instead of calling channel lookup repeatedly

savonarola: This happens only inside trace/debug macro calls, does not hit performance outside debugging

zmstone (approved):

PR #17998, opened 2026-07-17
feat(topic_metrics): add ns query param to topic_metrics2 :name endpoints

zmstone:
Release version: 6.3.0

## Summary

Follow-up to #17607 (topic-metrics v2). A global admin can list all
collections (each namespaced row carrying its `namespace`), but the
per-collection detail endpoints could not address a namespaced
collection, so the dashboard could not fetch/refresh an individual
namespaced row.

This adds an optional `ns` query parameter to the three per-collection
endpoints in `emqx_topic_metrics2_api.erl`:

- `GET  /mqtt/topic_metrics2/:name`
- `DELETE /mqtt/topic_metrics2/:name`
- `PUT  /mqtt/topic_metrics2/:name/reset`

Resolution rules (`resolve_ns/1` + `with_ns/2`):

- No `ns` param → the actor's own namespace (unchanged behavior).
- Global admin → may target any namespace named in `ns`.
- Namespaced admin naming their own ns → allowed (no-op).
- Namespaced admin naming a foreign ns → `403 Forbidden` (no namespace
  escalation).

The query-param mechanics mirror the sibling `emqx_prometheus_api`
(`{ns, mk(binary(), #{in => query, required => false})}` read from
`#{query_string := Qs}`). The list and collection-level bulk
`POST`/`DELETE` endpoints are unchanged — list already shows a global
admin everything and the bulk delete already scopes correctly.

The 403 namespace-escalation guard is covered by a dedicated test
(`t_namespaced_admin_foreign_ns_forbidden`), alongside cases for a
global admin targeting a namespace and a namespaced admin naming its own
namespace.

## PR Checklist

### DIFF
--- a/apps/emqx_topic_metrics/test/emqx_topic_metrics2_api_SUITE.erl
+++ b/apps/emqx_topic_metrics/test/emqx_topic_metrics2_api_SUITE.erl
@@ -252,6 +252,51 @@ t_namespace_isolation(Config) ->
     ?assertEqual(lists:sort(Keys), lists:usort(Keys)),
     ?assertEqual(4, length(AllForGlobal)).
 
+-doc "Global admin GETs/resets/deletes a namespaced collection via `?ns='.".
+t_global_targets_namespace_via_ns_param(Config) ->
+    %% acme owns a collection.
+    {201, _} = create(Config, ?NS_ACME, <<"n">>, <<"acme/n/#">>),
+    %% Global admin, no ns -> its own (global) namespace, which has no `n'.
+    ?assertMatch({404, _}, get_one(Config, ?global_ns, <<"n">>)),
+    %% Global admin with ?ns=acme -> reaches acme's collection.
+    {200, One} = get_one_ns(Config, ?global_ns, <<"n">>, ?NS_ACME),
+    ?assertMatch(#{<<"name">> := <<"n">>, <<"namespace">> := ?NS_ACME}, One),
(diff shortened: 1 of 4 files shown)


apps/emqx_topic_metrics/test/emqx_topic_metrics2_api_SUITE.erl
```diff
@@ -286,6 +342,17 @@ request(Config, Ns, Method, PathParts, Body) ->
         auth_header => auth_header(Config, Ns)
     }).
 
+%% Like request/5 but appends `?ns=<TargetNs>' to the URL, keeping the
+%% actor's own token from `Ns'.
+request_ns(Config, Ns, Method, PathParts, TargetNs, Body) ->
+    Url = emqx_mgmt_api_test_util:api_path(PathParts) ++ "?ns=" ++ b2l(TargetNs),
```
thalesmg: there's already a `query_params` parameter in simple_request that handles this.

thalesmg (dismissed):

apps/emqx_topic_metrics/test/emqx_topic_metrics2_api_SUITE.erl
zmstone: Good catch, switched to `query_params => #{<<"ns">> => TargetNs}`. Thanks!

PR #18005, opened 2026-07-17
fix(audit): correctly redact cli audit logs

savonarola:
Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/223

Release version: 6.0.4

## Summary

Fix CLI audit records that could persist caller-provided API secrets.

CLI handlers now define `<handler>_audit_args/1` callbacks that receive the original command arguments before audit persistence. If such a handler is absent, all the arguments are redacted.

## PR Checklist


apps/emqx_ctl/src/emqx_ctl.erl
```diff
@@ -341,17 +342,73 @@ safe_to_existing_atom(Str) ->
...
+            case prune_unnecessary_log(Log1) of
+                false ->
+                    ok;
+                {ok, Log2} ->
+                    apply_audit_command(Log2, Mod, Fun, Level, From)
+            end
+    end.
+
+apply_audit_args_callback(undefined, Log) ->
+    Log;
```
zmstone: when does this happen?

apps/emqx_ctl/src/emqx_ctl.erl
```diff
@@ -341,17 +342,73 @@ safe_to_existing_atom(Str) ->
...
+apply_audit_args_callback({CliModule, CliFunction} = CliHandler, Log = #{args := Args}) ->
+    case audit_args_function(CliModule, CliFunction) of
+        undefined ->
+            Log;
```
zmstone: and here ?
maybe should redact Args completely.

apps/emqx_ctl/src/emqx_ctl.erl
```diff
@@ -412,6 +469,9 @@ eval_erl([String]) ->
     {ok, Value} = eval_expr(Parsed),
     print("~p~n", [Value]).
 
+eval_erl_audit_args(Args) ->
```
zmstone, line 473: document the intent to NOT mask anything for eval command.
maybe add a warning message in the help text that the args of eval commands will be logged to audit logs without redaction.

apps/emqx_ctl/src/emqx_ctl.erl
savonarola: ✅ good catch

apps/emqx_ctl/src/emqx_ctl.erl
savonarola: Unified the paths for unknown commands. Now both an unknown atom and a not registered command result into early return and we don't get here.

zmstone (approved):

PR #18006, opened 2026-07-19
fix(topic_metrics): reject shared-subscription topic filters

### LINKED ISSUE #17965: [6.3.0]When configuring the topic $share/group/sensor/1 in the topic metrics collection, metrics cannot be counted normally.
### What happened?

When configuring the topic $share/group/sensor/1 in the topic metrics collection, metrics cannot be counted normally.

We should explicitly disallow creating collections with topics starting with $share/.

1. Create a collection with the topic set to $share/group/sensor/1

Image

2. Subscribe to the topic $share/group/sensor/1
3. send a mesage

Image

No metrics will be counted for shared topics. However, if I create a collection for the topic sensor/1, messages can be captured normally.
Image

Image

_No response_

### EMQX version

6.3.0

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

zmstone:
Release version: 6.3.0

## Summary

Fixes #17965.

Topic metrics v2 (`apps/emqx_topic_metrics`) accepted a collection whose `topic_filter` was a shared-subscription filter (`$share/<group>/<topic>` or the `$queue/<topic>` alias). Such a filter is a valid *subscription* filter but never matches a published `#message.topic` — the share prefix lives on the subscription, not on the message — so the collection was created but counted nothing.

Crucial changes:

* `emqx_topic_metrics_schema:validate_topic_filter/1` now rejects shared-subscription filters after the existing `emqx_topic:validate/1` check. Both the REST create path and the facade `register/3` (config / import / rehydrate) funnel through this validator, so both now return an error. The REST handler maps it to `400 BAD_TOPIC_FILTER` with a specific message explaining shared filters are not allowed.
* i18n (`rel/i18n/emqx_topic_metrics2_api.hocon`): the `topic_filter` field doc now states shared filters are not allowed and points users at the plain topic filter.

Deliveries to shared subscribers are unaffected: a plain-topic collection (e.g. `sensor/1`) still counts deliveries to a `$share/group/sensor/1` subscriber under `messages.out`, because the `message.delivered` hook sees the real topic. This is verified by a new end-to-end CT (`t_shared_subscription_delivery_counted`) using real MQTT clients — no topic stripping was needed.

Tests: `$share/` and `$queue/` create → 400; `sensor/1`, `sensor/#`, `sensor/+/temp` still 201; facade `register/3` rejects shared filters; publish-in + shared-delivery-out counting.

## PR Checklist
- [ ] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files (N/A — topic metrics v2 is not yet released)

### DIFF
--- a/apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
+++ b/apps/emqx_topic_metrics/src/emqx_topic_metrics2_api.erl
@@ -347,8 +347,13 @@ parse_create(#{<<"name">> := Name, <<"topic_filter">> := TF}) ->
     case emqx_topic_metrics_schema:validate_name(Name) of
         ok ->
             case emqx_topic_metrics_schema:validate_topic_filter(TF) of
-                ok -> {ok, Name, TF};
-                {error, _} -> {error, ?BAD_TOPIC_FILTER, <<"Invalid topic filter">>}
+                ok ->
+                    {ok, Name, TF};
+                {error, #{cause := shared_topic_filter_not_allowed}} ->
+                    {error, ?BAD_TOPIC_FILTER,
+                        <<"Shared subscription topic filters ($share/, $queue/) are not allowed">>};
+                {error, _} ->
+                    {error, ?BAD_TOPIC_FILTER, <<"Invalid topic filter">>}
             end;
         {error, _} ->
             {error, ?BAD_NAME, <<"Invalid name">>}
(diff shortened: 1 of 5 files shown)


keynslug (approved):

zmstone:
rebased for synced flaky fixes in base.

Issue #17946, opened 2026-07-14
[6.3.0]Uploading a namespace backup changes the import result


lina50:
### What happened?

  The same namespace backup produces different results depending on whether it is uploaded first.

### What did you expect to happen?

The two outcomes are expected to be consistent.

#### Reproduction

  1. Export a backup as a namespace user:

  POST /api/v5/data/export

  2. Import the exported file directly as a global administrator:

  POST /api/v5/data/import
  {
    "filename": "<namespace-backup>"
  }

  Result:

  HTTP 400 — import failed

  3. Download the same namespace backup.
  4. Upload it as a global administrator:

  POST /api/v5/data/files

  Result: HTTP 204 — upload succeeded

  5. Import the uploaded file as the same global administrator:

  POST /api/v5/data/import
  {
    "filename": "<uploaded-backup>"
  }

  Result: HTTP 204 — import succeeded

  This will cause confusion for users.

  Direct import:
  Namespace backup → Global import → HTTP 400

  Upload first:
  Namespace backup → Upload to global scope → Global import → HTTP 204

The fundamental reason for the different results is that the upload operation changes the storage scope of the backup file.
  The workflow differs as follows:
  Namespace A export
          ↓
  The file is actually stored at:
  ${EMQX_BACKUP_DIR}/ns/<namespace-A>/<filename>

  When the global administrator directly executes:

  POST /api/v5/data/import
  {
    "filename": "<filename>"
  }

  The import operation uses the global scope based on the global administrator’s identity and looks for the file at:

  ${EMQX_BACKUP_DIR}/<filename>

  The file is not located there, so the request returns HTTP 400.

  The upload workflow is different:

  POST /api/v5/data/files

  After the global administrator uploads the file, the code executes:

  Namespace = emqx_dashboard:get_namespace(Req)

  For a global administrator, Namespace is global, so the file is written to:

  ${EMQX_BACKUP_DIR}/<filename>

  When the file is imported again, the global scope can find it, so the request returns HTTP 204.

### EMQX version

6.3.0

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

PR #18008, opened 2026-07-19
fix(data_backup): honor namespace scope on import and upload

zmstone:
Release version: 6.0.3

## Summary

Data backup **import** (`POST /data/import`) and **upload** (`POST /data/files`) ignored the namespace scope, unlike list / download / delete / export. As reported in #17946, a global administrator could not import a namespaced backup directly — the import looked only in the global scope and returned `400` — while uploading the same file first appeared to work because upload silently re-homed it into the global scope.

Both operations now resolve the scope through the existing guarded `op_namespace/2` helper (used by list / download / delete), and both endpoints declare the optional `namespace` query parameter:

- `emqx_mgmt_api_data_backup:data_import/2` and `data_files/2` (upload) capture `query_string` and use `op_namespace(Req, QS)` instead of `get_namespace(Req)`.
- The `namespace` query parameter is added to the `/data/import` and `/data/files` (POST) schemas.

Net effect: a global administrator manages a namespace's backups end-to-end with `?namespace=<ns>`; without the parameter they stay on the global scope. Namespaced administrators remain confined to their own namespace — `op_namespace/2` ignores the parameter for them.

Most relevant file: `apps/emqx_management/src/emqx_mgmt_api_data_backup.erl`.

## PR Checklist


apps/emqx_management/src/emqx_mgmt_api_data_backup.erl
```diff
@@ -263,8 +269,8 @@ data_export(post, #{body := Params0} = Req) ->
             {500, #{code => 'INTERNAL_ERROR', message => Msg}}
     end.
 
-data_import(post, #{body := #{<<"filename">> := Filename} = Body} = Req) ->
-    Namespace = emqx_dashboard:get_namespace(Req),
+data_import(post, #{body := #{<<"filename">> := Filename} = Body, query_string := QS} = Req) ->
+    Namespace = op_namespace(Req, QS),
```
thalesmg, line 273: nit: req already contains qs.

thalesmg (approved):
I wonder if we shouldn't perhaps add a marker to the exported file indicating which namespace it should go into.  That way, the file would be self contained.

Issue #17966, opened 2026-07-16
GCP PubSub consumer: sporadic messages delayed by exactly `ack_deadline` (60s) due to abandoned synchronous pull requests


id:
### What happened?

In a production deployment using the GCP PubSub consumer source (PubSub outbox topic → MQTT republish) with sporadic traffic:

- ~90% of messages are republished to MQTT in under 2 s after being published to PubSub;
- ~5–10% are delayed by almost exactly **~61 s**;
- occasionally a message takes ~2 minutes or more (perceived as "never delivered" by an application with a 2-minute timeout).

The delayed deliveries quantize precisely at the subscription's `ackDeadlineSeconds` (60 s — EMQX's default) plus normal processing time. The same pattern reproduces on independent clusters, so it is not node- or environment-specific.

#### Root cause analysis

The affected messages are ones whose *first* delivery attempt from Pub/Sub was assigned to a pull request that EMQX had effectively abandoned. Chain:

1. The consumer worker long-polls `subscriptions:pull` with only `maxMessages` in the body (no `returnImmediately`) — `emqx_bridge_gcp_pubsub_consumer_worker:body(State, pull)` — so Pub/Sub holds quiet pulls open server-side (on the order of 90 s).
2. Pulls are async ehttpc requests bounded by `request_ttl` (default **45 s**), i.e. shorter than the server-side hold. On a quiet subscription every pull cycle therefore ends with client-side abandonment — and HTTP/1.1 has no request cancellation.
3. ehttpc never expires a *sent* async request (`ehttpc:is_sent_req_expired/2` returns `false` for async); the abandoned pull lingers until the zombie-connection check force-kills the whole connection ~`max_inactive` (10 s) after the deadline (`force_reconnecting_zombie_http_connection`). NAT/proxy middleboxes may also silently kill the byte-idle held connection even earlier.
4. Pub/Sub starts a message's lease the moment it commits it to a pull response, with no client receive-confirmation. A message committed to a dying/dead pull connection counts as an outstanding delivery and becomes deliverable again only after the full ack deadline — 60 s, which EMQX itself sets on the subscription (`body(State, create)` / `body(State, patch_subscription)`).
5. The redelivery is consumed by a live pull and republished normally → observed +61 s end-to-end. Two consecutive losses → ~+121 s ("never delivered" within a 2-minute window).

[...]

PR #18017, opened 2026-07-20
fix(gcp_pubsub): keep an in-flight pull open until its lease expires

zmstone:
Release version: 6.0.4, 6.1.4, 6.2.3, 6.3.0

## Summary

A GCP PubSub consumer source could repeatedly delay delivery of messages it had already
been handed by the server. This surfaced as a flaky
`emqx_bridge_gcp_pubsub_consumer_SUITE:t_cluster_subscription`, but the cause is in the
consumer worker, not the test.

`do_pull_async/1` in `emqx_bridge_gcp_pubsub_consumer_worker.erl` sends the pull request
with a TTL of `ack_deadline` — there is a comment there explaining that the pull must be
allowed to live as long as the lease. But the timer that *cancels* that request was armed
from `pull_retry_interval`, which `emqx_bridge_gcp_pubsub_impl_consumer.erl` derives from
`resource_opts.request_ttl`. Those are unrelated knobs, so the timer always pre-empted the
intended lifetime: 1s vs 10s in CT, 45s vs 60s with default settings.

That matters because:

* A pull is a long poll — with no `returnImmediately` it legitimately stays open until a
  message arrives.
* `ehttpc` never expires a request once it has been *sent*
  (`is_sent_req_expired/2`), so `request_ttl` only bounds the pending-queue wait. This
  timer is the only thing bounding the long poll.
* `ehttpc:cancel/2` drops the request *without* invoking the reply callback, so a response
  the server has already produced is discarded.

So every `pull_retry_interval` the worker tore down a healthy long poll. When the server
had just leased a message into that response, the response was thrown away but the lease
stood, making the message invisible until `ack_deadline` expired — and the next pull raced
the same short timer again.

Measured at the nginx in front of the emulator, with the source idle:

| | idle pull cadence | pulls in run | `t_cluster_subscription` |
|---|---|---|---|
| before | 2/sec (1/sec/node) | 200 | fails |
| after | 1 per 10s (= `ack_deadline`) | 16 | passes |

Publishing into that churn (a 30s idle window before the publish) reproduced the failure
100% of the time on the current code, and passes with this change.

The fix arms the timer with `ack_deadline` while a pull is in flight, and keeps
`pull_retry_interval` for the retry path where nothing is in flight and there is nothing to
lose by retrying soon. The stuck-pull cancellation added in bbe42e123f (#17966) is
preserved — it is still needed, since `ehttpc` will not time the request out on its own.

`t_no_premature_pull_cancel` covers the invariant: it fails without this change and passes
with it. Full consumer suite: 27 passed, 0 failed (1 pre-existing user-skip).

bbe42e123f has not been released yet, so this needs no release note of its own; the
existing `changes/ee/fix-17982.en.md` is annotated with this PR instead.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_gcp_pubsub/test/emqx_bridge_gcp_pubsub_consumer_SUITE.erl
+++ b/apps/emqx_bridge_gcp_pubsub/test/emqx_bridge_gcp_pubsub_consumer_SUITE.erl
@@ -1210,6 +1210,42 @@ t_timeout_during_pull(TCConfig) ->
     ),
     ok.
 
+-doc """
+Checks that a healthy in-flight pull request is left alone until its lease expires.
+
+A pull is a long poll that legitimately stays open until a message arrives.  Cancelling it
+sooner discards whatever the server has already leased into its response, and those
+messages then stay invisible until the lease (`ack_deadline`) expires.
+""".
+t_no_premature_pull_cancel(TCConfig) ->
+    AckDeadline = <<"10s">>,
+    %% A short `request_ttl' drives `pull_retry_interval': if the in-flight pull were tied to
+    %% it, it would be cancelled every second.  It must not be while a pull is in flight.
+    RequestTTL = <<"1s">>,
+    ?check_trace(
+        begin
+            {201, _} = create_connector_api(TCConfig, #{}),
+            {201, _} = create_source_api(TCConfig, #{
+                <<"parameters">> => #{<<"ack_deadline">> => AckDeadline},
+                <<"resource_opts">> => #{<<"request_ttl">> => RequestTTL}
+            }),
+            {ok, _} = snabbkaffe:block_until(
+                ?match_event(#{?snk_kind := "gcp_pubsub_consumer_worker_subscription_ready"}),
+                10_000
+            ),
+            %% Idle well within `ack_deadline': nothing should be cancelled.
+            timer:sleep(3_000),
+            ok
+        end,
+        fun(Trace) ->
+            ?assertEqual(
+                [], ?of_kind("gcp_consumer_worker_pull_stream_cancelled", Trace)
+            ),
+            ok
+        end
+    ),
+    ok.
+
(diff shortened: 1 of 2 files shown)


apps/emqx_bridge_gcp_pubsub/test/emqx_bridge_gcp_pubsub_consumer_SUITE.erl
```diff
@@ -1210,6 +1210,36 @@ t_timeout_during_pull(TCConfig) ->
...
+
+A pull is a long poll that legitimately stays open until a message arrives.  Cancelling it
+sooner discards whatever the server has already leased into its response, and those
+messages then stay invisible until the lease (`ack_deadline`) expires.
+""".
+t_no_premature_pull_cancel(TCConfig) ->
+    ?check_trace(
+        begin
+            {201, _} = create_connector_api(TCConfig, #{}),
+            {201, _} = create_source_api(TCConfig, #{}),
```
thalesmg: maybe set the desired timeouts explicitly so it doesn't depend on suite defaults?

changes/ee/fix-17982.en.md
```diff
@@ -1 +1,2 @@
-GCP PubSub Consumer now uses HTTP2 and cancels its pull request when it reaches the timeout.  This signals more cleanly to the GCP server that it may lease the messages to a new pull request, reducing tail latencies.
+[#18017](https://github.com/emqx/emqx/pull/18017) GCP PubSub Consumer now uses HTTP2 and cancels its pull request when it reaches the timeout.  This signals more cleanly to the GCP server that it may lease the messages to a new pull request, reducing tail latencies.
```
thalesmg: undo this change?

thalesmg (commented):
I don't think this will resolve the flakiness, as the emulator does not support http2, so the requests are not actually cancelled.

thalesmg (approved):

PR #18027, opened 2026-07-21
fix(mqtt): close connection on shared subscription when disabled

### LINKED ISSUE #3640: How to force close a connection when receiving invalid share subscriptions ?
The spec indicates that  `0x9E Shared Subscriptions not supported` is a kind of `protocol error`. 
According to the error handling chapter :https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#S4_13_Errors

 `When a Server detects a Malformed Packet or Protocol Error, and a Reason Code is given in the specification, it MUST close the Network Connection [MQTT-4.13.1-1]`.

I make a share subscription to a broker(with setting `mqtt.shared_subscription=false`) and got Subscribe Failed, but the connection is not closed.

zmstone:
Release version: 6.3.0

## Summary

Fixes #3640.

When `mqtt.shared_subscription = false` and a client sends a SUBSCRIBE containing a shared topic
filter (`$share/...` / `$queue/...`), EMQX now closes the network connection instead of only
returning `0x9E Shared Subscriptions not supported` in the SUBACK and keeping the session alive.

This follows MQTT 5.0 [MQTT-4.13.1-1]: when the server detects a protocol error for which a reason
code is specified, it must close the network connection.

* `emqx_channel:check_sub_caps/2` short-circuits the SUBSCRIBE pipeline with
  `{error, {disconnect, ?RC_SHARED_SUBSCRIPTIONS_NOT_SUPPORTED}, Channel}` if any topic filter in
  the packet was rejected with that reason code. The existing pipeline handling in
  `process_subscribe/2` + `handle_out(disconnect, ...)` already sends DISCONNECT then closes for
  MQTT v5, and closes the socket for v3.1/v3.1.1.
* Scope is limited to shared subscriptions; wildcard-not-supported (`0xA2`) and exclusive
  subscription behavior are unchanged.
* Only reachable when an operator explicitly opts out of shared subscriptions (the capability is
  enabled by default).

New suite `apps/emqx/test/emqx_shared_sub_disabled_SUITE.erl` covers: v5 gets DISCONNECT `0x9E` and
the channel is gone; v3.1.1 gets the socket closed; a mixed normal+shared SUBSCRIBE still closes;
normal subscriptions are unaffected; and shared subscriptions still work with the default config.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_shared_sub_disabled_SUITE.erl
+++ b/apps/emqx/test/emqx_shared_sub_disabled_SUITE.erl
@@ -11,6 +11,7 @@
 -include_lib("emqx/include/asserts.hrl").
 -include_lib("eunit/include/eunit.hrl").
 -include_lib("common_test/include/ct.hrl").
+-include_lib("snabbkaffe/include/snabbkaffe.hrl").
 
 -define(SHARED_TOPIC, <<"$share/g/t/1">>).
 -define(NORMAL_TOPIC, <<"t/1">>).
@@ -51,7 +52,7 @@ t_v5_shared_sub_disabled(_Config) ->
         {'DOWN', MRef, process, C,
             {shutdown, {disconnected, ?RC_SHARED_SUBSCRIPTIONS_NOT_SUPPORTED, _}}}
     ),
-    ?assertEqual([], emqx_cm:lookup_channels(clientid(v5))).
+    assert_channel_gone(clientid(v5)).
 
 -doc """
 An MQTT v3.1.1 client attempting a shared subscription while shared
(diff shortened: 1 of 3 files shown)


apps/emqx/test/emqx_shared_sub_disabled_SUITE.erl
```diff
@@ -0,0 +1,114 @@
...
+    ok = disable_shared_subscription(),
+    C = connect(v5),
+    MRef = erlang:monitor(process, C),
+    _ = catch emqtt:subscribe(C, ?SHARED_TOPIC, 0),
+    ?assertReceive({disconnected, ?RC_SHARED_SUBSCRIPTIONS_NOT_SUPPORTED, _}),
+    ?assertReceive(
+        {'DOWN', MRef, process, C,
+            {shutdown, {disconnected, ?RC_SHARED_SUBSCRIPTIONS_NOT_SUPPORTED, _}}}
+    ),
+    ?assertEqual([], emqx_cm:lookup_channels(clientid(v5))).
```
thalesmg: this may require a retry to avoid flakiness, since cleanup is async?

thalesmg (dismissed):

apps/emqx/test/emqx_shared_sub_disabled_SUITE.erl
zmstone: good catch

PR #18033, opened 2026-07-21
perf(essential): skip per-connection stats reporting when client-info API is off

zmstone:
## Summary

Follow-up to the QA feature-comparison (emqx/emqx-qa#1509), which showed `ESSENTIAL` and `FULL` within ~3% CPU because feature-gating only stops *apps* from booting — the core `emqx` hot path still does all of its observability/API bookkeeping regardless of which features are on.

This PR skips one piece of that bookkeeping — **per-connection stats reporting** — when the feature it feeds (the dashboard/management client-info API) is disabled.

## What it does

- **`emqx_features`** (new, base `emqx` app): reads boot-time capability booleans from `persistent_term`, defaulting to `true` (so `FULL` / unset behaves exactly as today). No upward dependency on `emqx_machine`.
- **`emqx_machine_features`**: after resolving `EMQX_FEATURES`, publishes the `observability` and `client_info` capabilities into `persistent_term`.
- **Connection modules** (`emqx_connection`, `emqx_socket_connection`, `emqx_ws_connection`): each connection periodically reports its stats into `?CHAN_INFO_TAB` (via `emqx_cm:set_chan_stats`), which only feeds the `GET /clients` stats columns. When `client_info` is disabled, `maybe_set_chan_stats/2` skips both the `stats/1` gathering and the ETS write.

### Why gate the push, not the timer

The `emit_stats` timer is intentionally left running. Besides reporting stats it also drives **send-queue congestion recovery** (`check_sendq_congestion`, `congested → running`), **connection-congestion alarms**, and (in `emqx_socket_connection`) the **send-timeout check**. Disabling the whole timer would suppress those — notably, a quiet sendq-congested connection could fail to recover and keep buffering QoS0. So only the stats push is gated; all other timer duties are unaffected.

## Why it is safe

`?CHAN_INFO_TAB`'s stats column is read only by the management/dashboard client-info API. Core paths (takeover, kick/discard, routing) use the pid-registry tables; node eviction/rebalance read the *info* column, not the *stats* column. Rate limiting and flow control are independent mechanisms (own token buckets / socket backpressure) and read none of this.

## Benchmark

Single node, `emqtt_bench` (1:1 pub/sub, QoS 1, 256-byte payload). Broker CPU measured inside EMQX via `recon:scheduler_usage/1`. The effect was isolated on **one running node** by flipping the `client_info` capability at runtime (same apps, same connections, timer firing in both phases):

**12 000 connections, `idle_timeout=2s` (amplified push cadence to lift the signal above noise):**

| stats push | `set_chan_stats`/s | broker CPU |
|---|---|---|
| on | 7 956 | 2.84 cores |
| gated (this PR) | 0 | 2.32 cores (**−18%**) |

This gives a per-push cost of **~6.6×10⁻⁵ cores** (the `stats/1` computation + the `?CHAN_INFO_TAB` write). The saving scales with the report rate ≈ `connections / idle_timeout` (plus socket-event-driven reports). At the **default** `idle_timeout`, e.g. 16 000 connections ≈ 1 600 reports/s ≈ **0.1 cores** — a few percent of broker CPU at these loads, more for connection-heavy deployments, and negligible at very high message rates (consistent with QA's ~3% overall gap). Keeping the timer running for congestion handling costs ~0.004 cores, so correctness here is essentially free. Throughput is unaffected.

End-to-end in `ESSENTIAL` mode (`client_info` resolved off), `emqx_cm:set_chan_stats` drops to **0 calls/s** across all connection modules while the timers keep firing.

## Notes

- The default v6 TCP listener uses `emqx_socket_connection`; the gate is applied there as well as in `emqx_connection`/`emqx_ws_connection`.

### DIFF
--- a/apps/emqx_machine/src/emqx_machine_features.erl
+++ b/apps/emqx_machine/src/emqx_machine_features.erl
@@ -280,19 +280,25 @@ cache_features() ->
 
 %% Push boot-time capability flags down into the base `emqx` application so the
 %% hot path can gate optional book-keeping without depending on `emqx_machine`.
-%% `observability` follows the metrics/dashboard-monitor sinks; `client_info`
-%% follows the dashboard/management client-info REST API.
+%% Keyed on the governance features rather than concrete app names, so moving or
+%% renaming an app within a feature does not change the resolved capabilities.
+%% `observability` follows the metrics/dashboard-monitor sinks (the `metrics`
+%% feature, or `dashboard` for the dashboard monitor); `client_info` follows the
+%% dashboard/management client-info REST API (the `dashboard` feature, which
+%% owns `emqx_management`).
 publish_capabilities(Features) ->
-    Observability = any_app_enabled([emqx_prometheus, emqx_dashboard], Features),
-    ClientInfo = any_app_enabled([emqx_management, emqx_dashboard], Features),
+    Observability = any_feature_enabled([metrics, dashboard], Features),
+    ClientInfo = any_feature_enabled([dashboard], Features),
     ok = emqx_features:set_capability(observability, Observability),
     ok = emqx_features:set_capability(client_info, ClientInfo),
     ok.
 
-any_app_enabled(_Apps, #{preset := full}) ->
+any_feature_enabled(_Features, #{preset := full}) ->
     true;
-any_app_enabled(Apps, #{allowed_apps := Allowed}) ->
-    lists:any(fun(App) -> is_map_key(App, Allowed) end, Apps).
+any_feature_enabled(Features, #{enabled_features := Enabled}) ->
+    lists:any(fun(Feature) -> is_map_key(Feature, Enabled) end, Features);
+any_feature_enabled(_Features, _Other) ->
+    false.
 
 resolve_dependent_apps(Feature, KnownFeatures) ->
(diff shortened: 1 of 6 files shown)


qzhuyan (dismissed):

apps/emqx_machine/src/emqx_machine_features.erl
```diff
@@ -275,8 +275,25 @@ cache_features() ->
...
+    ok = publish_capabilities(Features),
     Features.
 
+%% Push boot-time capability flags down into the base `emqx` application so the
+%% hot path can gate optional book-keeping without depending on `emqx_machine`.
+%% `observability` follows the metrics/dashboard-monitor sinks; `client_info`
+%% follows the dashboard/management client-info REST API.
+publish_capabilities(Features) ->
+    Observability = any_app_enabled([emqx_prometheus, emqx_dashboard], Features),
+    ClientInfo = any_app_enabled([emqx_management, emqx_dashboard], Features),
```
thalesmg: maybe just check for the feature itself (dashboard, metrics) rather than specific apps?

they are more semantic, and would allow renaming/moving apps around.

thalesmg (dismissed):

qzhuyan (approved):

PR #18034, opened 2026-07-21
perf(essential): skip global message metrics counters when observability is off

zmstone:
## Summary

Follow-up to the QA feature-comparison (emqx/emqx-qa#1509). Companion to the `emqx_features` boot-time capability mechanism (merged in #18033). This one skips the **global per-packet `emqx_metrics` counters** on the hot path when observability is off.

## What it does

When the observability sinks (Prometheus/`metrics`, `$SYS`, dashboard monitor) are disabled, the connection modules skip the per-packet global `emqx_metrics` updates that only feed those sinks:

- `emqx_metrics:inc_recv/2` and `inc_sent/2` — several `counters:add` per PUBLISH in each direction (doubled under a tenant namespace).
- `bytes.received` / `bytes.sent` (`inc_metrics/3`).

The `observability` capability is read once at connection init and cached in connection state (`#conf` for `emqx_connection`/`emqx_socket_connection`, `#state` for `emqx_ws_connection`), so the hot path adds only a cached record read, not a `persistent_term` lookup per packet.

Gated in **all three** connection modules — `emqx_connection`, `emqx_socket_connection` (the default v6 TCP/TLS listener), and `emqx_ws_connection` — so the optimization actually engages for real traffic.

## What is deliberately kept

- **`recv_pkt`** — drives keepalive detection; always tracked.
- **Per-connection process-dictionary counters** (`recv_msg`/`send_msg`/qos) — cheap process-local writes feeding the per-client stats display.
- Rate limiting and flow control are unaffected — they use their own token buckets / socket backpressure and read none of these counters.

## Benchmark

Single node, `emqtt_bench` QoS 1 publishers, 16-byte payload, ~57k msg/s. Broker CPU via `recon:scheduler_usage/1`. Isolated on one FULL node by flipping `observability` and reconnecting (the flag is cached per connection):

| observability | `inc_recv`/s | broker CPU |
|---|---|---|
| on | 57 361 | 9.09 cores |
| off (this PR) | 0 | 8.86 cores (**−2.6%**) |

`inc_recv` drops to 0, confirming the gate engages on the TCP path. The saving is per-packet, so it scales with message rate (~2.6% here at ~57k msg/s; Phase B actually sustained a higher rate for lower CPU) and is negligible at low message rates — consistent with QA's ~3% overall ESSENTIAL-vs-FULL gap.

### DIFF
--- a/apps/emqx/src/emqx_connection.erl
+++ b/apps/emqx/src/emqx_connection.erl
@@ -100,7 +100,11 @@
     %% Forced GC thresholds, `false` if disabled
     force_gc :: false | {_EachNMessages :: pos_integer(), _EachNBytes :: pos_integer()},
     %% Forced shutdown policy
-    force_shutdown :: emqx_types:oom_policy()
+    force_shutdown :: emqx_types:oom_policy(),
+    %% Whether observability sinks (Prometheus/metrics, `$SYS`, dashboard
+    %% monitor) are enabled. When `false`, the global message counters that
+    %% only feed those sinks are skipped on the hot path.
+    observability = true :: boolean()
 }).
 
 -record(thresholds, {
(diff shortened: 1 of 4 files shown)


zmstone:
too little gain.

PR #18035, opened 2026-07-21
perf(essential): skip client-info table when client-info API is off [RFC]

zmstone:
## Draft / RFC — one of three ESSENTIAL-mode hot-path perf explorations

Follow-up to emqx/emqx-qa#1509. Companion to the `emqx_features` boot-time capability mechanism. **This is the most invasive of the three and is an RFC** — it changes how eviction/rebalance enumerate channels. Opening it draft to settle the direction before investing further.

### The idea

`?CHAN_INFO_TAB` (`emqx_cm`) is the single largest per-connection ETS record — a full info map per client — and it exists **only** as a read model for the client-info REST API (`GET /clients`). Core paths (takeover, kick/discard, routing) use the pid-registry tables, never this one. So when the client-info API is disabled we should be able to skip populating it and reclaim that per-connection memory at high client counts (the QA runs already show ESSENTIAL ~1–1.4 GB lower).

### What this PR does

- Gates `insert_channel_info` on `client_info_enabled()` — the info map is not stored when the API is off. `set_chan_info` / `set_chan_stats` then naturally no-op.
- Guards the one internal reader: `emqx_channel` reads its own `sockinfo` back from the table on a keepalive-interval change (`maps:get(sockinfo, emqx_cm:get_chan_info(...), #{})`), which would crash on the now-empty table — it is skipped when the API is off.
- Keeps eviction/rebalance working: **`emqx_cm:all_channels_stream/1`** (used by `emqx_eviction_agent`, a core app) needs the full info map, so when the table is disabled it now enumerates the always-present pid registry (`?CHAN_CONN_TAB`) and fetches each channel's info **live** via `Mod:call(Pid, info)`.

### ⚠️ Open question / why this is an RFC

The live fallback issues **one process call per channel**. That is fine for the rare, deliberate eviction/rebalance enumeration, but it is O(N) synchronous calls — unacceptable if anything enumerates frequently, and slow at 100k+ channels. Options to resolve before this is mergeable:

1. **Keep a minimal always-on record** (conninfo + clientinfo + conn_state only) and gate only the heavy/mutable fields + stats — no live fallback needed, smaller memory win.
2. **Bulk live gather** for eviction (batch `erlang:process_info` / a dedicated cast) instead of per-channel `call`.
3. Accept the live fallback but assert eviction is the only caller and cap concurrency.

I lean toward **(1)** as the real design, with this PR showing the end-to-end gating. Feedback welcome.

### Validation TODO (draft)
- [ ] CI build (warnings_as_errors) + existing CT (esp. `emqx_eviction_agent`, `emqx_node_rebalance`, `emqx_mgmt_api_clients`)
- [ ] Confirm `GET /clients` works in FULL; eviction still enumerates in ESSENTIAL
- [ ] Measure per-connection memory delta at high client counts
- [ ] Decide on the open question above before un-drafting

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -1726,9 +1726,16 @@ handle_cast(
     NKeepAlive = emqx_keepalive:update(Zone, Interval, KeepAlive),
     NConnInfo = maps:put(keepalive, Interval, ConnInfo),
     NChannel = Channel#channel{keepalive = NKeepAlive, conninfo = NConnInfo},
-    SockInfo = maps:get(sockinfo, emqx_cm:get_chan_info(ClientId), #{}),
-    ChanInfo1 = info(NChannel),
-    emqx_cm:set_chan_info(ClientId, ChanInfo1#{sockinfo => SockInfo}),
+    %% Refresh the client-info read model only when it is maintained; when the
+    %% client-info API is disabled the info table is not populated.
+    case emqx_features:client_info_enabled() of
+        true ->
+            SockInfo = maps:get(sockinfo, emqx_cm:get_chan_info(ClientId), #{}),
+            ChanInfo1 = info(NChannel),
+            emqx_cm:set_chan_info(ClientId, ChanInfo1#{sockinfo => SockInfo});
+        false ->
+            ok
+    end,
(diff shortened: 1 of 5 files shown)


zmstone:
client-info and chan-info are useful troubleshooting tables

PR #18075, opened 2026-07-23
fix(resource): age out queued requests when inflight window is saturated

zmstone:
Release versions:
5.10.5

## Summary

When a remote system accepts async requests but silently stops responding (e.g. a remote MQTT broker that stops sending PUBACK), the action's sent-but-unacknowledged requests saturate the buffer worker's inflight window and `flush` can no longer run. Since `request_ttl` was only checked at flush time or when a reply arrives, queued requests were never sieved for expiry: they accumulated uncounted (`queuing` kept growing, zero `expired`/`dropped` counted) until the buffer eventually overflowed (`dropped.queue_full`).

Changes:

- `emqx_resource_buffer_worker`: add a periodic sweep (every `resume_interval`, which is bounded by `request_ttl`) that, **while the inflight window is full**, drops queued (not yet sent) requests past their TTL and counts them as `dropped.expired`. Requests already sent (inflight) are deliberately **not** expired: once delivery has started, the request must be settled by its reply — counted as `late_reply` when the reply arrives past the TTL. For MQTT, an unacknowledged QoS 1/2 PUBLISH must be completed or resent, never abandoned [MQTT-4.4.0-1]; message expiry only applies before delivery has started [MQTT-3.3.2-5].
- `emqx_resource_buffer_worker`: re-arm the `blocked` state resume timer when nothing is retriable and the inflight table is full; previously the worker went dormant in this situation.
- `emqx_bridge_mqtt`: pass the action's `request_ttl` as the async publish timeout instead of `infinity`, so messages waiting *unsent* in the emqtt client's internal pending queue (when its own inflight window to the remote broker is full) are dropped once past TTL, instead of being delivered stale after connectivity recovers. PUBLISHes already sent to the remote broker are unaffected.

Note: `emqx_resource_buffer_worker.erl` is shared with the v6 branches, which have the same issue; this PR targets 5.10 — a forward port to v6 should follow separately.

## PR Checklist


zmstone:
we'll live with the fact that expired req drops are driven by non-timer events for now.

PR #18077, opened 2026-07-23
fix(cluster): refuse join before node boot is complete

zmstone:
Release version:
6.0.4
6.1.4
6.2.3

## Summary

A `cluster join` request (CLI or dashboard invite RPC) reaching a node that is still starting its applications makes mria restart underneath the in-progress boot loop; whichever mria-backed app is being started at that moment crashes with a `mria_schema` noproc and takes the whole node down. Observed in the two-node cluster smoke test under `EMQX_FEATURES=ESSENTIAL` on dev-63 (e.g. PR #18059), where `emqx_retainer` is among the last apps to start and was reliably mid-init when the join arrived:

```
[notice] Application: mria. Exited: stopped. Type: temporary.
[error]  crasher: emqx_retainer:init/1 ... {noproc,{gen_server,call,[mria_schema,{create_table,...
[critical] failed_to_start_app ... emqx_retainer -> node down
```

The race is not specific to feature gates (they only widen the window), so the fix targets dev-60 and forward-syncs up the chain.

Change:

- `emqx_machine:start/0` marks boot as in progress; `emqx_machine_boot:post_boot/0` clears the mark only after the boot app list has started **and** `start_autocluster/0` has registered the ekka join callbacks — that registration is what makes a join safe (it stops the apps before mria restarts), so the mark models the real invariant.
- `emqx_cluster:join/1` refuses while the mark is set:
  ```
  Failed to join the cluster: "This node has not fully booted yet. Please retry after it is started."
  ```
  Both `emqx ctl cluster join` and the dashboard invite RPC funnel through `emqx_cluster:join/1`, so both are covered. Environments that boot without `emqx_machine` (common tests) never set the mark and are unaffected.
- The two-node smoke test now retries `emqx ctl cluster join` until it is accepted, so CI exercises the refusal path end-to-end and the node-crash flake is gone.

Verified: compiles clean on the 6.0 base; single-node `EMQX_FEATURES=ESSENTIAL` boot clean with the guard in place (6.3 build); new CT case `t_cluster_join_refused_while_booting` asserts the refusal while booting and that the command proceeds to the normal peer check afterwards.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_cluster.erl
+++ b/apps/emqx/src/emqx_cluster.erl
@@ -10,7 +10,8 @@
     force_leave/1,
     ensure_normal_mode/0,
     ensure_singleton_mode/0,
-    is_single_node_mode/0
+    is_single_node_mode/0,
+    set_booting/1
 ]).
 
 %% RPC callback functions
@@ -29,6 +30,16 @@
 -endif.
 
 join(PeerNode) ->
+    case is_booting() of
+        true ->
+            {error,
+                "This node has not fully booted yet. "
+                "Please retry after it is started."};
+        false ->
+            do_join(PeerNode)
+    end.
+
+do_join(PeerNode) ->
     %% Local node starts with default license (singleton mode),
     %% But the peer node(s) may have a license which allows local node to join.
     %% So we do not check the license locally.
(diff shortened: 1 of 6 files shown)


thalesmg (approved):
I wonder if classy won't take care of these cases already

zmstone:
> I wonder if classy won't take care of these cases already

that'll be a question for dev-63

PR #17944, opened 2026-07-14
feat(http): add OAuth2 client-credentials support to HTTP connector and auth

terry-xiaoyu:
Adds shared OAuth2 client-credentials token management for HTTP-based connectors and HTTP authentication/authorization.

- Introduces `emqx_connector_oauth2`, which fetches, caches, and proactively refreshes access tokens per resource, restores registrations after manager restarts, and avoids cross-resource fetch blocking.
- Adds an `oauth2` configuration block to the HTTP connector, `authn_http`, and `authz_http`, injects Bearer tokens into outgoing requests, and includes token availability in connector health checks.
- Validates token endpoints and token responses, supports TLS options for HTTPS token endpoints, and rejects conflicting user-provided `Authorization` headers.

Release version: 6.0.4

Fix: N/A

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (described in the summary)

PR #18079, opened 2026-07-23
refactor(kafka): reuse shared OAuth2 token manager

hjianbo:
Release version:

6.1.4

Follow-up: https://github.com/emqx/emqx/pull/17944

## Summary

Reuse `emqx_connector_oauth2` for Kafka producer and consumer OAuth2 client-credentials authentication, and remove the Kafka-specific token cache.

OAuth2 registrations are keyed by the connector resource ID, preventing concurrent probes from overwriting or unregistering each other. Failed consumer starts also unregister their OAuth2 configuration.

Kafka configuration field names remain unchanged, including independent request and connect timeouts. The `grant_type` field remains required in the schema and keeps its `client_credentials` default for API compatibility; connector API responses now materialize that default instead of returning an empty field. OAuth token responses may use either `Bearer` or `JWT` as `token_type`. Token endpoint failures use the shared manager sanitized error shape.

Confluent logical-cluster and identity-pool values continue to be passed only as SASL OAUTHBEARER extensions and are not sent to the token endpoint.

Inline OAuth2 TLS certificates and private keys are persisted under the connector certificate directory, matching connector-level TLS handling and preventing raw private-key material from being returned by the API.

## PR Checklist

- [x] No changelog is required because this refactors unreleased OAuth2 support
- [x] The OAuth2 SSL `enable` default is intentionally disabled and backward compatible for this unreleased feature


apps/emqx_connector_oauth2/src/emqx_connector_oauth2_schema.erl
```diff
@@ -84,7 +84,7 @@ fields(client_credentials) ->
             })},
         {ssl,
             mk(ref(emqx_schema, "ssl_client_opts"), #{
-                default => #{<<"enable">> => true},
+                default => #{<<"enable">> => false},
```
thalesmg, line 87: strange default?  shouldn't use https by default?

apps/emqx_bridge_kafka/test/emqx_bridge_kafka_action_SUITE.erl
```diff
@@ -2429,6 +2431,26 @@ t_oauth_client_credentials_authn(TCConfig0) ->
...
     ?assertMatch({201, #{<<"status">> := <<"connected">>}}, CreateConnectorFn()),
+    #{connector_type := ConnectorType, connector_name := ConnectorName} =
+        emqx_bridge_v2_testlib:get_common_values(TCConfig),
+    ?assertMatch(
+        {200, #{
+            <<"authentication">> := #{<<"grant_type">> := <<"client_credentials">>}
+        }},
+        emqx_bridge_v2_testlib:simplify_result(
+            emqx_bridge_v2_testlib:get_connector_api(ConnectorType, ConnectorName)
+        )
```
thalesmg: nit: maybe make a `get_connector_api` small helper in this module, similar to other suites

apps/emqx_bridge_kafka/src/emqx_bridge_kafka.erl
```diff
@@ -691,6 +685,27 @@ kafka_connector_config_fields() ->
...
+    auth_union_members();
+auth_union_member_selector({value, Value}) when is_map(Value) ->
+    case emqx_utils_maps:binary_key_map(Value) of
+        #{<<"mechanism">> := Mechanism} when Mechanism =:= oauth; Mechanism =:= <<"oauth">> ->
+            [ref(auth_oauth_client_credentials)];
+        _ ->
+            auth_union_members()
+    end;
+auth_union_member_selector({value, _Value}) ->
+    auth_union_members().
```
thalesmg: this seems strange:  why not narrow down to the specific type or throw an error if there is no matching union member?

apps/emqx_auth_http/test/emqx_authz_http_SUITE.erl
```diff
@@ -950,6 +986,65 @@ setup_handler_and_config(TCConfig, Handler, Config) ->
...
+                {raise, Class, Reason} -> erlang:raise(Class, Reason, []);
+                _ -> Result
+            end
+        end
+    ),
+    try
+        Test()
+    after
+        meck:unload(emqx_resource)
+    end.
```
thalesmg: nit: we have `emqx_common_test_helpers:with_mock`

apps/emqx_bridge_http/test/emqx_bridge_http_SUITE.erl
```diff
@@ -440,6 +440,38 @@ t_oauth2_client_credentials(TCConfig) ->
...
+    {201, #{<<"status">> := <<"connected">>}} =
+        create_connector_api(TCConfig, #{<<"oauth2">> => OAuth2}),
+    #{
+        connector_type := Type,
+        connector_name := Name
+    } = emqx_bridge_v2_testlib:get_common_values(TCConfig),
+    {200, #{<<"oauth2">> := #{<<"ssl">> := SavedSSL}}} =
+        emqx_bridge_v2_testlib:simplify_result(
+            emqx_bridge_v2_testlib:get_connector_api(Type, Name)
+        ),
```
thalesmg: nit: please create a `get_connector_api` mini helper in the module, similar to other suites

apps/emqx_bridge_kafka/test/emqx_bridge_kafka_source_SUITE.erl
```diff
@@ -1961,3 +1961,156 @@ t_msk_iam_authn(TCConfig) ->
...
+        ?assertMatch(
+            {200, #{
+                <<"status">> := <<"connected">>,
+                <<"authentication">> := #{
+                    <<"grant_type">> := <<"client_credentials">>
+                }
+            }},
+            emqx_bridge_v2_testlib:simplify_result(
+                emqx_bridge_v2_testlib:get_connector_api(ConnectorType, ConnectorName)
+            )
```
thalesmg: nit: please create mini helper similar to others

apps/emqx_bridge_kafka/test/emqx_bridge_kafka_source_SUITE.erl
```diff
@@ -1961,3 +1961,156 @@ t_msk_iam_authn(TCConfig) ->
...
+    ResourceIdB = ?PROBE_ID_NEW(),
+    ok = emqx_bridge_kafka_oauth_authn:register(ResourceIdA, Auth),
+    ok = emqx_bridge_kafka_oauth_authn:register(ResourceIdB, Auth),
+    CallbackB = emqx_bridge_kafka_oauth_authn:mk_token_callback(ResourceIdB),
+    ok = emqx_bridge_kafka_oauth_authn:unregister(ResourceIdA),
+    ?assertEqual(
+        {ok, #{token => <<"token-b">>}},
+        CallbackB(#{client_id => <<"probing_brod_consumers">>})
+    ),
+    ok = emqx_bridge_kafka_oauth_authn:unregister(ResourceIdB),
```
thalesmg: I'm a bit confused as to what this tests:  the handler above always returns token-b unconditionally, so this seems like it's just testing the mock?

maybe add a doc annotation explaining what is going on?

apps/emqx_auth/src/emqx_authn/emqx_authn_utils.erl
```diff
@@ -49,7 +49,11 @@ create_resource(Module, ResourceConfig, #{resource_id := ResourceId} = State, Me
                 ResourceConfig,
                 ?DEFAULT_RESOURCE_OPTS(OwnerId)
             ),
-        ok = start_resource_if_enabled(State, Mechanism, Backend)
+        ok =
+            start_created_resource(
+                ResourceId,
+                fun() -> start_resource_if_enabled(State, Mechanism, Backend) end
+            )
```
thalesmg: nit:


```suggestion
            remove_resources_on_exception(
                ResourceId,
                fun() -> start_resource_if_enabled(State, Mechanism, Backend) end
            )
```

apps/emqx_bridge_kafka/test/emqx_bridge_kafka_source_SUITE.erl
hjianbo: Thanks, you're right. The previous test was poorly structured and its intent was not clear enough. I replaced it with two successive consumer connector probes using different OAuth2 client IDs, and added a short doc annotation.

Both probes use brod's shared `probing_brod_consumers` client ID, but the token endpoint must receive one request for each OAuth2 configuration. With the old Kafka-client-ID-keyed cache, the second probe would reuse the first probe's token and never request a token for `client-b`.

apps/emqx_connector_oauth2/src/emqx_connector_oauth2_schema.erl
hjianbo: This default does not mean that HTTP is preferred or that HTTPS is disabled. The transport is determined by the `token_endpoint` scheme.

For an `https://` endpoint, `http_options/4` always builds TLS options with `SSL#{enable => true}`. The schema default only keeps an omitted SSL configuration from being materialized as explicitly enabled; HTTPS endpoints still use TLS, and users can configure CA, verification, certificates, and other TLS options when needed.

thalesmg (approved):

PR #18085, opened 2026-07-23
feat(kafka): add max_batch_age and max_retries producer options (wolff 4.2.1)

zmstone:
Release version:
6.0.4, 6.1.4, 6.2.3, 6.3.0

## Summary

Bump wolff 4.1.10 → 4.2.1 and integrate its two new producer options end-to-end for the Kafka, Confluent, and Azure Event Hubs actions:

- `max_batch_age` (default `infinity`): maximum duration a message may stay in the producer buffer before it is dropped instead of being sent. Applies both to still-queued messages (even while disconnected) and to messages that were in-flight when the connection dropped.
- `max_retries` (default `infinity`, wolff option name `max_retry`): maximum number of retries after Kafka error responses (e.g. `not_leader_for_partition`) before a batch is dropped. Connection-loss resends are not counted against this limit; they are bounded by `max_batch_age`.

Additionally exposes two previously hard-coded / derived wolff settings:

- Action `parameters.reconnect_delay` (default `2s`, the previous hard-coded value): wolff producer `reconnect_delay_ms`, the delay before reconnecting after a connection loss.
- Connector `request_timeout` (default `30s`): the wolff client request timeout, previously derived as `max(2 * metadata_request_timeout, 30s)`.

All new fields are added in `emqx_bridge_kafka` and flow into the Confluent and Azure Event Hubs schemas via their existing field derivation; each bridge gets its own i18n descs. `producers_config/5` / `on_start` pass them through to wolff.

Drop-outcome accounting (`emqx_bridge_kafka_impl_producer`), designed so each message is counted exactly once per counter family:

- `message_expired` ack → reply `{error, request_expired}`: the internal-buffer reply path concludes the rule action (and triggers fallback actions) without bumping resource metrics; `dropped` and `dropped.expired` are bumped by the new `[wolff, dropped_expired]` telemetry subscription. The parallel `[wolff, dropped]` event is intentionally not subscribed, so expiry drops do not double-count `dropped` (same division of labor as `buffer_overflow_discarded`, and same reply reason as the Pulsar connector uses for its expiry drops).
- `max_retry_exceeded` ack → reply `{error, max_retry_exceeded}`: counted as `failed` via the reply path, mirroring `message_too_large`; wolff's parent `dropped` counter remains unmapped, keeping `dropped >= dropped.expired` coherent. The final failed attempt's `[wolff, retried_failed]` event bumps only `retried`/`retried.failed` (as before), so no double count there either.
- The sync send path (`wolff:send_sync2/4`) maps the same two drop reasons; wolff 4.2.1 fixed the stale `offset_reply()` type spec that previously made this fail dialyzer.

Also rides along:
- wolff 4.1.11's `max_linger_time` restore for memory-mode buffers (same fix that went to v5 in #18082), and the corresponding `max_linger_time`/`max_linger_bytes` i18n description updates from #18101.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_tests.erl
+++ b/apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_tests.erl
@@ -183,5 +183,15 @@ aeh_producer_test_() ->
                         }
                     }
                 )
+            )},
+        {"reconnect_delay defaults to 2s and is settable",
+            ?_assertMatch(
+                #{<<"parameters">> := #{<<"reconnect_delay">> := <<"1500ms">>}},
+                check_action(#{<<"parameters">> => #{<<"reconnect_delay">> => <<"1500ms">>}})
+            )},
+        {"connector request_timeout defaults to 30s",
+            ?_assertMatch(
+                #{<<"request_timeout">> := <<"30s">>},
+                check_connector(#{})
             )}
     ].
--- a/apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_tests.erl
+++ b/apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_tests.erl
@@ -184,7 +184,12 @@ aeh_producer_test_() ->
                     }
                 )
             )},
-        {"reconnect_delay defaults to 2s and is settable",
+        {"reconnect_delay defaults to 2s",
+            ?_assertMatch(
+                #{<<"parameters">> := #{<<"reconnect_delay">> := <<"2s">>}},
+                check_action(#{})
+            )},
+        {"reconnect_delay is settable",
             ?_assertMatch(
                 #{<<"parameters">> := #{<<"reconnect_delay">> := <<"1500ms">>}},
                 check_action(#{<<"parameters">> => #{<<"reconnect_delay">> => <<"1500ms">>}})
--- a/apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_tests.erl
+++ b/apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_tests.erl
@@ -156,5 +156,32 @@ aeh_producer_test_() ->
             ?_assertThrow(
                 ?action_validation_error("Expected: no_compression" ++ _, <<"gzip">>),
                 check_action(#{<<"parameters">> => #{<<"compression">> => <<"gzip">>}})
+            )},
+        {"max_batch_age and max_retries default to infinity",
+            ?_assertMatch(
+                #{
+                    <<"parameters">> := #{
+                        <<"max_batch_age">> := <<"infinity">>,
+                        <<"max_retries">> := <<"infinity">>
+                    }
+                },
+                check_action(#{})
+            )},
+        {"max_batch_age and max_retries are settable",
+            ?_assertMatch(
+                #{
+                    <<"parameters">> := #{
+                        <<"max_batch_age">> := <<"500ms">>,
+                        <<"max_retries">> := 3
+                    }
+                },
+                check_action(
+                    #{
+                        <<"parameters">> => #{
+                            <<"max_batch_age">> => <<"500ms">>,
+                            <<"max_retries">> => 3
+                        }
+                    }
+                )
             )}
     ].
--- a/apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
+++ b/apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
@@ -581,8 +581,6 @@ do_send_msg(async, KafkaTopic, KafkaMessage, Producers, AsyncReplyFn) ->
     %% See emqx_resource_buffer_worker:simple_async_internal_buffer
     {ok, Pid}.
 
-%% Wolff producer never gives up retrying
-%% so there can only be 'ok' results.
 on_kafka_ack(_Partition, Offset, ReplyFnAndArgs) when is_integer(Offset) ->
     %% `emqx_rule_runtime:inc_action_metrics/2' is embedded inside reply function
     emqx_resource:apply_reply_fun(ReplyFnAndArgs, ok);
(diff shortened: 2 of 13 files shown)


apps/emqx_bridge_kafka/src/emqx_bridge_kafka_impl_producer.erl
```diff
@@ -559,10 +558,16 @@ render_timestamp(Template, Message) ->
...
-        ok
+        case wolff:send_sync2(Producers, KafkaTopic, [KafkaMessage], SyncTimeout) of
+            {_Partition, message_expired} ->
+                %% dropped because it stayed in the buffer longer than 'max_batch_age'
+                {error, request_expired};
+            {_Partition, max_retry_exceeded} ->
+                %% dropped after 'max_retries' Kafka error responses
+                {error, max_retry_exceeded};
+            {_Partition, _Offset} ->
+                ok
```
zmstone: {_Partition, Offset} when is_integer(Offset) -> ok;
{_Partition, message_expired} -> {error, request_expired};
{_Partition, Reason} -> {error, Reason}

changes/ee/feat-18085.en.md
```diff
@@ -0,0 +1,19 @@
...
+  The default `infinity` keeps the previous behavior (retry forever).
+
+- `reconnect_delay`: delay before the producer attempts to reconnect after a connection loss.
+  The default `2s` keeps the previous (hard-coded) behavior.
+
+Also added a `request_timeout` option to the corresponding connectors: the maximum duration to wait for a reply from Kafka for a pending request before the connection is considered stale and re-established.
+The default `30s` matches the previous behavior for default configurations (it used to be derived as at least 30 seconds).
+
+Additionally, `max_linger_time` is honored again for memory-mode buffers (upgraded the Kafka client library wolff to 4.2.0). When less than a full batch is buffered, the producer waits up to `max_linger_time` for more messages before forming a produce request, reducing the produce request rate at high message rates; sending is immediate whenever a full batch's worth of data is available, and the publish path is never delayed. The default `max_linger_time = 0` keeps the previous send-immediately behavior.
```
zmstone: keep it short.

rel/i18n/emqx_bridge_azure_event_hub.hocon
```diff
@@ -18,6 +18,12 @@ min_metadata_refresh_interval.desc:
 min_metadata_refresh_interval.label:
 """Min Metadata Refresh Interval"""
 
+request_timeout.desc:
+"""Maximum duration to wait for a reply from Azure Event Hubs for a pending request.
```
zmstone, line 22: add a note: the risk of setting it too small is Event Hubs may ingest the batch but delay the ack, causing EMQX to retry the already accepted batch after reconnect, i.e. overwhelming data volume for downstream consumers.

rel/i18n/emqx_bridge_azure_event_hub.hocon
```diff
@@ -105,16 +111,40 @@ partition_count_refresh_interval.label:
 """Partition Count Refresh Interval"""
 
 max_linger_time.desc:
-"""Maximum duration for a per-partition producer to wait for messages in order to collect a batch to buffer.
-The default value `0` means no wait. For non-memory buffer mode, it's advised to configure at least `5ms` for less IOPS."""
+"""Maximum duration a per-partition producer waits to accumulate more messages into a larger batch.
+The default value `0` means no wait (messages are sent as soon as they arrive).
+The wait ends early as soon as a full batch's worth of data has accumulated, so a busy producer is never delayed.
+When buffering to disk, the wait happens before writing to the buffer; configuring at least `5ms` is advised there to reduce disk IOPS."""
```
zmstone, line 119: add a note: the default value is optimized for messaging latency, if acceptable, a delay is encouraged to help lower the number of requests per second towards Event Hubs.

rel/i18n/emqx_bridge_azure_event_hub.hocon
```diff
@@ -105,16 +111,40 @@ partition_count_refresh_interval.label:
...
+Each dropped message counts towards the <code>dropped.expired</code> metric.
+The default value <code>infinity</code> means messages never expire (retained until sent or the buffer overflows)."""
+
+max_batch_age.label: "Max Batch Age"
+
+max_retries.desc:
+"""Maximum number of times to retry sending a message batch after Azure Event Hubs responds with a retryable error (for example, when the partition leader has changed).
+After the initial attempt and this many retries have all failed, the batch is dropped and each of its messages counts towards the <code>failed</code> metric.
+Reconnect-triggered resends after a connection loss are not counted against this limit; they are bounded by <code>max_batch_age</code> instead.
+The default value <code>infinity</code> means retry forever."""
```
zmstone: do not repeat default value in desc, default value is a part of the schema.

rel/i18n/emqx_bridge_azure_event_hub.hocon
```diff
@@ -105,16 +111,40 @@ partition_count_refresh_interval.label:
...
-"""Maximum number of bytes for a per-partition producer to wait for messages in order to collect a batch to buffer."""
+"""Maximum number of bytes a per-partition producer accumulates while waiting before it stops waiting and sends the batch."""
 
 max_linger_bytes.label: "Max Linger Bytes"
 
+max_batch_age.desc:
+"""Maximum duration a message may stay in the producer buffer before it is dropped instead of being sent to Azure Event Hubs.
+A message batch is dropped only when all of its messages have stayed in the buffer longer than this duration; this applies both to messages still queued (including while the connection is down) and to messages awaiting acknowledgement when the connection is lost.
+Each dropped message counts towards the <code>dropped.expired</code> metric.
+The default value <code>infinity</code> means messages never expire (retained until sent or the buffer overflows)."""
```
zmstone: do not repeat default value in desc, default value is a part of the schema.

apps/emqx_bridge_azure_event_hub/test/emqx_bridge_azure_event_hub_tests.erl
```diff
@@ -156,5 +156,42 @@ aeh_producer_test_() ->
...
+                check_action(
+                    #{
+                        <<"parameters">> => #{
+                            <<"max_batch_age">> => <<"500ms">>,
+                            <<"max_retries">> => 3
+                        }
+                    }
+                )
+            )},
+        {"reconnect_delay defaults to 2s and is settable",
```
thalesmg: nit: this doesn't test the default, despite the title.

apps/emqx_bridge_kafka/test/emqx_bridge_kafka_action_SUITE.erl
```diff
@@ -689,6 +713,174 @@ t_smoke_metrics(TCConfig) ->
...
+being sent on reconnect.  Each dropped message bumps `dropped` and
+`dropped.expired` exactly once and must NOT be counted as `failed` or
+`success` (the rule action is concluded via the `request_expired` reply,
+which bypasses resource metrics).  Fresh messages produced after recovery
+are delivered normally.
+""".
+t_max_batch_age_expiry_drop() ->
+    [{matrix, true}].
+t_max_batch_age_expiry_drop(matrix) ->
+    [[?tcp_plain, ?no_auth]];
```
thalesmg: this one would be good to exercise both sync and async too, since they use different paths.

I wonder if it would feasible to exercise max_retries_exceeded too (preferably with as little mocking as possible)

```suggestion
    [
        [?tcp_plain, ?no_auth, Sync]
     || Sync <- [?sync, ?async]
    ];
```

thalesmg (dismissed):

thalesmg (approved):

PR #18097, opened 2026-07-24
fix(datalayers): handle missing prepared ref after ecpool reconnect

JimMoen:
Follow-up to #18014.

After Datalayers server restart, the old prepared statement references in `prepared_refs` map become stale. When ecpool workers reconnect, their new client pids are not in the old `prepared_refs` map, causing write failures.

### Fixes
- `call_driver`: when client pid is not found in `prepared_refs`, call `prepare_sql_to_conn` on-the-fly to create a new prepared statement
- `prune_dead_refs`: filter dead worker pids from `prepared_refs` on each query to prevent stale entry accumulation

### Tests
- `t_reconnect_and_write`: toxiproxy disconnect + reconnect verifies writes succeed after recovery
- `t_prune_dead_refs`: unit test for dead pid cleanup behavior

### DIFF
--- a/apps/emqx_bridge_datalayers/mix.exs
+++ b/apps/emqx_bridge_datalayers/mix.exs
@@ -5,7 +5,7 @@ defmodule EMQXBridgeDatalayers.MixProject do
   def project do
     [
       app: :emqx_bridge_datalayers,
-      version: "6.0.2",
+      version: "6.0.3",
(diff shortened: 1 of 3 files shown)


zmstone:
squash the commits?

zmstone (approved):

PR #18117, opened 2026-07-25
fix(multi_tenancy): purge built-in authn/authz data on namespace delete

### LINKED ISSUE #17816: [6.3.0]When deleting namespaces, the corresponding authentication and authorization data are not removed.
### What happened?

When deleting namespaces, the corresponding authentication and authorization data are not removed.

When a namespace is deleted, the corresponding authentication and authorization data shall also be removed.

1. Create a namespace named ns1.
2. Add an authenticated user test to ns1.
3. Delete namespace ns1 (the prompt indicates relevant data will be removed upon deletion).

Image

4. Recreate namespace ns1.
5. Check the list of authenticated users; user test still exists.

_No response_

### EMQX version

6.1.3

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

zmstone:
Release version: 6.1.4, 6.2.3, 6.3.0

## Summary

Fixes #17816.

Deleting a managed namespace did not remove its built-in database authentication users (password-based and SCRAM) or authorization rules, so recreating a namespace with the same name showed the stale users and rules.

The `'namespace.delete'` hook (run by the namespace cleanup janitor) already has callbacks from the dashboard and management apps to clean their namespaced data, but no auth app registered one. This PR:

- Adds `purge_namespace/1` to `emqx_authn_mnesia` and `emqx_authn_scram_mnesia`, deleting all of the namespace's users across all user groups with mria dirty deletes (global-namespace records are never touched).
- Adds `emqx_auth_mnesia_hookcb`, registered at `emqx_auth_mnesia` app start, whose `'namespace.delete'` callback runs the two new purges plus the existing `emqx_authz_mnesia:purge_rules/1`.
- Adds a new `emqx ctl mt purge_ns <namespace>` CLI command (JSON output) that deletes a namespace and purges all of its data synchronously, without checking whether the namespace exists. It is idempotent — a last resort to manually clean up leftover data when a previous (API-triggered) deletion was interrupted halfway.

The purge operates on the mnesia tables directly (rather than iterating configured authenticators) since the tables are the source of truth for the namespaced data; dirty deletes replicate cluster-wide via mria.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_hookcb.erl
+++ b/apps/emqx_management/src/emqx_mgmt_hookcb.erl
@@ -32,8 +32,26 @@ unregister_hooks() ->
 
 on_namespace_delete(Namespace) ->
     ok = emqx_mgmt_auth:delete_all_keys_from_namespace(Namespace),
+    ok = purge_builtin_auth_data(Namespace),
     ok.
 
 %%------------------------------------------------------------------------------
 %% Internal fns
 %%------------------------------------------------------------------------------
+
+-doc """
+Deletes all built-in database authentication users (password-based and SCRAM) and
+authorization rules belonging to the deleted namespace.
+
+The built-in database tables exist only while the `emqx_auth_mnesia` application is
+running; its top supervisor creates them.
+""".
+purge_builtin_auth_data(Namespace) ->
+    case is_pid(whereis(emqx_auth_mnesia_sup)) of
+        true ->
+            ok = emqx_authn_mnesia:purge_namespace(Namespace),
+            ok = emqx_authn_scram_mnesia:purge_namespace(Namespace),
+            ok = emqx_authz_mnesia:purge_rules(Namespace);
+        false ->
+            ok
+    end.
--- a/apps/emqx_mt/src/emqx_mt_cli.erl
+++ b/apps/emqx_mt/src/emqx_mt_cli.erl
@@ -24,9 +24,13 @@ mt(["purge_ns", Ns0]) ->
     Ns = unicode:characters_to_binary(Ns0),
     case emqx_mt_config:force_purge_ns(Ns) of
         ok ->
-            emqx_ctl:print("ok~n");
+            print_json(#{result => <<"ok">>, namespace => Ns});
         {error, cleanup_incomplete} ->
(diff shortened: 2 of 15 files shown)


apps/emqx_management/src/emqx_mgmt_hookcb.erl
```diff
@@ -32,8 +32,26 @@ unregister_hooks() ->
...
+
+The built-in database tables exist only while the `emqx_auth_mnesia` application is
+running; its top supervisor creates them.
+""".
+purge_builtin_auth_data(Namespace) ->
+    case is_pid(whereis(emqx_auth_mnesia_sup)) of
+        true ->
+            ok = emqx_authn_mnesia:purge_namespace(Namespace),
+            ok = emqx_authn_scram_mnesia:purge_namespace(Namespace),
+            ok = emqx_authz_mnesia:purge_rules(Namespace);
```
thalesmg: it's a bit strange for this cleanup to belong to emqx_management instead of emqx_auth_mnesia.

if we move them to the proper application, then it shouldn't be needed to check for the existence of the supervisor.

apps/emqx_mt/src/emqx_mt_cli.erl
```diff
@@ -0,0 +1,41 @@
...
+unload() ->
+    ok = emqx_ctl:unregister_command(mt).
+
+mt(["purge_ns", Ns0]) ->
+    Ns = unicode:characters_to_binary(Ns0),
+    case emqx_mt_config:force_purge_ns(Ns) of
+        ok ->
+            emqx_ctl:print("ok~n");
+        {error, cleanup_incomplete} ->
+            emqx_ctl:print("Some cleanup steps failed (see logs); re-run the command to retry.~n")
```
thalesmg: q:  weren't we aiming to adopt json responses for CLI commands?  since this is a new module and functions, maybe we could adopt that here.

apps/emqx_management/src/emqx_mgmt_hookcb.erl
zmstone: changed to register namespace delete hook callback.

thalesmg (approved):

Issue #17727, opened 2026-06-25
Date-based / daily rotation for disk_log-based file outputs (trace + DiskLog connector)


zmstone:
## Summary

EMQX's `disk_log`-backed file outputs — the trace log handler and the **DiskLog / FileLog connector** (action sink) — rotate **only by size** today (`max_file_size` × `rotation_count`). Operators who need a per-calendar-day file have to bolt `logrotate` on top, but that interacts badly with `disk_log`'s `.idx` / `.siz` sidecar files (the `.siz` counter doesn't get reset by external `copytruncate`, so EMQX re-rotates on its own counter and the names drift away from the intended one-file-per-day pattern).

Expose end-of-day (EoD) rotation, with the date encoded in the filename via a template. Keep size-based rotation as a within-day safety cap.

## Customer use case (the trigger for this issue)

A customer is configuring a DiskLog → FileLog connector to capture only MQTT messages and needs the trace file rotated once per calendar day, kept for 30 days, with date-stamped names like `mqtt-trace-20260624.log`. They tried a `logrotate` rule with `copytruncate`; that's incompatible with `disk_log`'s `.siz` sidecar (EMQX's internal counter doesn't see the truncate and rotates early, drifting filenames).

## Constraints we can't change

- **The `.N` suffix on rotated files** comes from Erlang/OTP's `disk_log` wrap mode. We can change the base file path (and put a date in it), but every rotated file is mechanically `<base>.<N>`. The active file is never the un-suffixed base. So a day's files will look like:

  ```
  mqtt-trace-20260624.log.1     ← today's active slot
  mqtt-trace-20260624.log.2     ← previous within-day rotation
  mqtt-trace-20260624.log.3
  mqtt-trace-20260624.log.idx   ← disk_log sidecar (index)
  mqtt-trace-20260624.log.siz   ← disk_log sidecar (size counter)
  mqtt-trace-20260623.log.1     ← yesterday's last slot
  ```

  Documentation must call this out so operators size `max_file_number` for the worst-case day, and so they're not confused by the sidecars.

- **The action/connector framework has no native EoD event.** Callbacks today are request-driven (`on_query`, `on_batch_query`) or periodic status (`on_get_status`). There is no "cron" / "scheduled tick" hook surfaced through `emqx_resource_manager`. Any EoD rotation has to find its trigger elsewhere — see Implementation paths below.

## Proposed behaviour

Add a rotation-strategy option to the trace handler AND the DiskLog connector schema:

[...]

PR #18119, opened 2026-07-25
feat(disk_log): support daily/hourly rotation with date-stamped filenames

zmstone:
Release version:
6.3.0

## Summary

Implements date-based (daily / hourly) rotation for the Disk Log connector, per #17727.

- New optional `rotation` block in the connector config: `period = none | hour | day` (default `none`, backward compatible), `retention_period` (duration, default `infinity`), and `timezone` (`UTC`, `local`, or a fixed offset such as `+02:00`, resolved via `emqx_utils_calendar`).
- When `period` is `day`/`hour`, the log is opened with the current period's date stamp inserted into the filename. The stamp is always `YYYYMMDDHH` — daily files carry a literal `00` hour (`mqtt-trace.log` → `mqtt-trace-2026062400.log`), so the name format stays the same if `period` is later switched between `day` and `hour`. Files on disk are `mqtt-trace-2026062400.log.N` plus the `.idx`/`.siz` bookkeeping files, as imposed by OTP `disk_log` wrap mode — documented in the schema descriptions.
- Period boundary detection piggy-backs on `on_get_status/2` (path 1 recommended in the issue; no resource-framework change). Since health check results cannot carry updated connector state, the currently open file is read from `disk_log:info/1` and compared against the expected dated path computed from "now"; on mismatch the log is closed and reopened with the new dated path.
- Size-based rotation (`max_file_size` / `max_file_number`) is unchanged and keeps acting as a within-period cap.
- Retention: after each period rotation (and on connector start), date-stamped file sets from periods older than `retention_period` are deleted (only files whose name matches the `<base>-<stamp><ext>.<N|idx|siz>` shape). No-op when `infinity`.
- Writes hitting the brief close/reopen window during a period rotation now map to a recoverable error, so the buffer worker retries them instead of dropping.

Decisions taken:

- The existing flat `max_file_size` / `max_file_number` fields stay where they are (the connector shipped in 6.0, so moving them under `rotation{}` would break released configs); the `rotation` block is purely additive.
- The trace log handler is **not** covered here: it rotates via `logger_disk_log_h` handler config, a different mechanism, and will be handled as a follow-up if still desired.

Main files: `apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl`, `emqx_bridge_disk_log_connector_schema.erl`, `rel/i18n/emqx_bridge_disk_log_connector_schema.hocon`.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl
+++ b/apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl
@@ -32,6 +32,8 @@
     on_batch_query/3
 ]).
 
+-export([timezone_offset_seconds/1]).
+
 -ifdef(TEST).
 -export([flush/1, get_wrap_logs/2]).
 -endif.
@@ -46,19 +48,29 @@
 -define(disk_log, disk_log).
 
 -define(filepath, filepath).
+-define(conn_config, conn_config).
 -define(installed_actions, installed_actions).
 
 -define(template, template).
 -define(write_mode, write_mode).
 
+-define(SECONDS_PER_DAY, 86400).
+
+-type period() :: none | hour | day.
+-type rotation() :: #{
+    period := period(),
+    retain_period_for_days => pos_integer(),
+    timezone := binary()
+}.
 -type connector_config() :: #{
     filepath := binary(),
     max_file_size := pos_integer(),
-    max_file_number := pos_integer()
+    max_file_number := pos_integer(),
+    rotation => rotation()
 }.
 -type connector_state() :: #{
-    deleteme := true,
     ?filepath := binary(),
+    ?conn_config := connector_config(),
     ?installed_actions := #{action_resource_id() => action_state()}
 }.
 
@@ -94,10 +106,14 @@ callback_mode() ->
     {ok, connector_state()} | {error, _Reason}.
 on_start(ConnResId, ConnConfig) ->
     #{filepath := FilepathBin} = ConnConfig,
+    NowS = now_s(),
+    ActiveFilepath = active_filepath(ConnConfig, NowS),
     maybe
-        ok ?= do_open_log(ConnResId, ConnConfig),
+        ok ?= do_open_log(ConnResId, ConnConfig#{filepath := ActiveFilepath}),
+        ok = sweep_expired_period_files(ConnConfig, NowS),
         ConnState = #{
             ?filepath => FilepathBin,
+            ?conn_config => ConnConfig,
             ?installed_actions => #{}
         },
         {ok, ConnState}
@@ -112,13 +128,13 @@ on_stop(ConnResId, _ConnState) ->
     ok.
 
 -spec on_get_status(connector_resource_id(), connector_state()) ->
-    ?status_connected | ?status_disconnected.
-on_get_status(ConnResId, #{?filepath := Filepath} = _ConnState) ->
+    ?status_connected | ?status_disconnected | {?status_disconnected, binary()}.
+on_get_status(ConnResId, #{?conn_config := ConnConfig} = _ConnState) ->
     case disk_log:info(ConnResId) of
         {error, no_such_log} ->
             ?status_disconnected;
         LogInfo when is_list(LogInfo) ->
-            check_file_status(Filepath, LogInfo, ConnResId)
+            check_period_and_file_status(ConnResId, ConnConfig, LogInfo)
     end.
 
 -spec on_get_channels(connector_resource_id()) ->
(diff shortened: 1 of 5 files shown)


rel/i18n/emqx_bridge_disk_log_connector_schema.hocon
```diff
@@ -18,4 +18,24 @@ emqx_bridge_disk_log_connector_schema {
   """Maximum Number of Files"""
   max_file_number.desc:
   """Maximum number of log files to be used.  Once the maximum number of files is reached and a new rotation is required, the oldest such file is truncated and used as the new current file."""
```
zmstone: update this description.

First, this number of files applies to each rotation period.
Second, I believe the oldest file is deleted, but not truncated.

rel/i18n/emqx_bridge_disk_log_connector_schema.hocon
zmstone: this desc should be updated too

zmstone: Updated: the description now mentions the date-stamped format (`mqtt-trace-2026062400.log.N`).

rel/i18n/emqx_bridge_disk_log_connector_schema.hocon
zmstone: Updated: it now says the limit applies to each rotation period's file set separately. On the delete-vs-truncate point: OTP wrap logs reuse the oldest file in place — `disk_log_1:new_ext_file/2` calls `open_truncate/1` on it — so the file is not deleted; I worded it as "the oldest file's contents are discarded and the file is reused as the new current file".

apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl
```diff
@@ -261,6 +275,172 @@ maybe_rotate(ConnResId, ConnConfig) ->
...
+%% `siz').  Anything else is left alone.
+is_expired_period_file(File, Root, Ext, CutoffKey) ->
+    FileBin = unicode:characters_to_binary(File),
+    Prefix = <<Root/binary, "-">>,
+    PrefixSize = byte_size(Prefix),
+    ExtSize = byte_size(Ext),
+    maybe
+        <<Prefix:PrefixSize/binary, Rest/binary>> ?= FileBin,
+        {Stamp, <<Ext:ExtSize/binary, $., Suffix/binary>>} ?=
+            string:take(Rest, lists:seq($0, $9)),
```
zmstone: add is_decimal_str for string:take(..., lists:seq($0,$9))

apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl
```diff
@@ -261,6 +275,172 @@ maybe_rotate(ConnResId, ConnConfig) ->
...
+
+is_period_key(Stamp) ->
+    byte_size(Stamp) =:= 10.
+
+is_wrap_log_suffix(<<"idx">>) ->
+    true;
+is_wrap_log_suffix(<<"siz">>) ->
+    true;
+is_wrap_log_suffix(Suffix) ->
+    Suffix =/= <<>> andalso {Suffix, <<>>} =:= string:take(Suffix, lists:seq($0, $9)).
```
zmstone: is_decimal_str for string:take

apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl
zmstone: Done — the stamp is matched as a fixed 10-byte segment now and validated with `is_decimal_str/1` (which also let `is_period_key/1` fold away).

apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl
zmstone: Done — `is_wrap_log_suffix/1` now delegates to `is_decimal_str/1`.

apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl
```diff
@@ -94,10 +107,14 @@ callback_mode() ->
...
 on_start(ConnResId, ConnConfig) ->
     #{filepath := FilepathBin} = ConnConfig,
+    NowS = now_s(),
+    ActiveFilepath = active_filepath(ConnConfig, NowS),
     maybe
-        ok ?= do_open_log(ConnResId, ConnConfig),
+        ok ?= do_open_log(ConnResId, ConnConfig#{filepath := ActiveFilepath}),
+        ok = sweep_expired_period_files(ConnConfig, NowS),
         ConnState = #{
             ?filepath => FilepathBin,
```
thalesmg: nit: this one is kinda redundant now that we retain the whole config map in the state.

apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl
```diff
@@ -261,6 +278,173 @@ maybe_rotate(ConnResId, ConnConfig) ->
...
+check_period_and_file_status(ConnResId, ConnConfig, LogInfo0) ->
+    case maybe_rotate_period(ConnResId, ConnConfig, LogInfo0) of
+        not_rotated ->
+            check_file_status(open_filepath(LogInfo0), LogInfo0, ConnResId);
+        rotated ->
+            case disk_log:info(ConnResId) of
+                LogInfo when is_list(LogInfo) ->
+                    check_file_status(open_filepath(LogInfo), LogInfo, ConnResId);
+                {error, no_such_log} ->
+                    ?status_disconnected
```
thalesmg:
```suggestion
                    {?status_disconnected, no_such_log}
```

apps/emqx_bridge_disk_log/src/emqx_bridge_disk_log_connector.erl
```diff
@@ -261,6 +278,173 @@ maybe_rotate(ConnResId, ConnConfig) ->
...
+    case open_filepath(LogInfo) =:= ExpectedFilepath of
+        true ->
+            not_rotated;
+        false ->
+            rotate_period(ConnResId, ConnConfig, ExpectedFilepath, NowS)
+    end.
+
+rotate_period(ConnResId, ConnConfig, NewFilepath, NowS) ->
+    maybe
+        ok ?= stop_disk_log(ConnResId),
```
thalesmg, line 315: I wonder if concurrent writes during this rotation could interact badly with this (unless there's a specific error that inserting returns, which could be used to mark the error as recoverable)

thalesmg: apparently a new clause was added below, so should be already covered.

apps/emqx_bridge_disk_log/test/emqx_bridge_disk_log_SUITE.erl
thalesmg, line 1: nit: would be interesting to have a case where entries are concurrently being written (sync and async) while the rotation happens.

thalesmg (dismissed):

thalesmg (approved):

Issue #17856, opened 2026-07-02
Support per-namespace authentication and authorization chains


id:
### Use case

Multi-tenant deployment where each namespace maps to a tenant with its own auth service. The tenant admin wants EMQX to hold no user data: each tenant's HTTP auth endpoint decides connect/pub/sub for its own clients.

### Current limitations

- An authentication chain holds at most one authenticator per `mechanism:backend` pair — a second `password_based:http` with a different URL and a namespace `precondition` is impossible.
- `authentication` / `authorization` are not namespaced config roots. Namespaced roots are hardcoded to `actions`, `connectors`, `rule_engine`, `sources` (`emqx_conf_app:ensure_allowed_namespaced_root_keys/0`); managed-namespace configs only cover `session` and `limiter` (`emqx_mt_config`).
- The HTTP authenticator hostname is not templatable — `${client_attrs.tns}` in the URL authority is rejected at validation (`emqx_auth_http_utils:parse_url/1`, fixed connection pool per authenticator). Verified on 6.2.1.

### Workaround today

Single global HTTP authenticator with `url = "http://gw/authn/${client_attrs.tns}"` (path/query/headers/body are templatable; requires `tns` to be set pre-auth via `mqtt.client_attrs_init`) plus path-based routing to per-tenant services in the customer's ingress. Works, but forces the customer to run and maintain a routing layer, and namespace admins cannot self-manage auth config.

### Proposed feature

Add `authentication` and `authorization` to namespaced config roots, so a managed namespace can define its own chain/sources (e.g. its own `password_based:http` authenticator), resolved after `client_attrs.tns` extraction, falling back to the global chain when the namespace defines none.

### Alternative (smaller): multiple authenticators of the same type

Allow multiple authenticators of the same `mechanism:backend` in the global chain, disambiguated by `precondition` — e.g. chain entries keyed by index or a user-supplied id instead of `mechanism:backend`.

### Alternative: templated hostname

Allow template variables in the HTTP authenticator/authorizer hostname, e.g. `url = "https://${client_attrs.tns}.auth.example.com/authn"`. Removes the need for a customer-side routing layer entirely. Implies: per-rendered-host connection pooling (today the pool binds to a fixed `request_base` at creation), per-host TLS verification, and a hostname allowlist/suffix constraint since `tns` is client-derived input (SSRF risk otherwise).

zmstone:
> we'll implement "Alternative: templated hostname"

[...]

PR #18130, opened 2026-07-27
feat(auth): support templated hostname in HTTP authn/authz URL

zmstone:
Release version:
6.3.0

## Summary

Implements the "templated hostname" alternative chosen in #17856: the HTTP authenticator and HTTP authorizer URL host part may now contain template variables, e.g. `https://${client_attrs.tns}.auth.example.com/authn`, enabling per-tenant auth endpoints without a customer-side routing layer.

Crucial changes in logic:

- A new `hostname_resolution` config (enum `static | dynamic`, default `static`) selects the request path. `static` is today's behavior byte-for-byte: fixed `request_base` + `ehttpc` pool via `emqx_bridge_http_connector`. `dynamic` resolves the hostname per request and is **required** when the URL host contains template placeholders (a templated host with `static` resolution is a config error); it also works with a literal hostname, for per-request DNS resolution without a persistent pool. `emqx_auth_http_utils:parse_url_template/2` classifies the URL accordingly. Only the host may be templated: the scheme must be literal `http`/`https`, the port (if any) a literal integer, and userinfo/fragments remain rejected.
- In dynamic mode, no connector resource is created. `emqx_auth_http_utils:one_off_request/5` renders the host per request (when templated), then issues the request via `hackney` with connect/recv timeouts from `connect_timeout`/`request_timeout` and the authenticator's/authorizer's `ssl` options applied against the rendered host (peer verification included; SNI derives from the rendered host unless explicitly configured — covered by tests). Responses are normalized to the same shapes the `ehttpc` path produces, so response handling is unchanged.
- Connection reuse in dynamic mode respects the existing `pool_size` config: requests go through a shared hackney pool (named `authn` / `authz`) capped at `pool_size` idle connections (hackney pools key cached sockets by host, so per-tenant hosts share the cap); `pool_size = 0` uses no pool at all. Pipelining options (`enable_pipelining`, `max_inactive`) do not apply in dynamic mode — hackney does one in-flight request per connection; the docs say so.
- Since a rendered host derives from client-controlled input, a templated-host config **must** set the new `allowed_hosts` field (exact hostnames or `*.suffix` wildcard patterns). At request time the rendered host is charset-validated (plain DNS name/IPv4 label rules, so no port/path/userinfo injection) and checked against the list; rendering failures (e.g. unset `client_attrs`), malformed hostnames, and hosts outside the list fail closed without any request being sent. `allowed_hosts` is not required for a literal host in dynamic mode (admin-controlled, not client-derived).
- Host template variables are folded into the auth-cache key template, so cached decisions cannot leak across rendered hosts.
- Updating an authenticator/authorizer between static and dynamic resolution tears down/recreates the connector resource appropriately; status APIs report `connected` for resource-less (dynamic) instances, as they already do for resource-less backends.
- `oauth2` is rejected in combination with dynamic resolution (its token flow lives in the connector resource), and the SCRAM REST API authenticator explicitly keeps rejecting templated hosts.

Most relevant files: `apps/emqx_auth/src/emqx_auth_http_utils.erl` (URL classification, allowlist, one-off request, hackney pool), `apps/emqx_auth_http/src/emqx_authn_http.erl`, `apps/emqx_auth_http/src/emqx_authz_http.erl` (state building and query dispatch), schema + i18n for `hostname_resolution`, `allowed_hosts`, `pool_size`.

## PR Checklist


apps/emqx_auth/src/emqx_auth_http_utils.erl
```diff
@@ -200,4 +452,138 @@ templates_test_() ->
         )
     ].
 
+parse_url_template_test_() ->
```
zmstone: move eunit test cases  to test dir eunit module

apps/emqx_auth/src/emqx_auth_http_utils.erl
```diff
@@ -11,15 +11,50 @@
...
+-ifdef(TEST).
+%% Internal functions exported for tests in test/emqx_auth_http_utils_tests.erl.
+-export([format_url/4]).
+-endif.
 
 -define(DEFAULT_HTTP_REQUEST_CONTENT_TYPE, <<"application/json">>).
 
+%% Rendered host must be a plain DNS name or IPv4 address:
+%% non-empty labels of [a-z0-9_-] separated by dots.
+-define(HOSTNAME_RE, "^[a-z0-9_-]+(\\.[a-z0-9_-]+)*$").
```
savonarola: `_` is not a valid hostname character (https://www.rfc-editor.org/info/rfc1034/#section-3.5)

apps/emqx_auth/src/emqx_auth_http_utils.erl
```diff
@@ -72,16 +146,258 @@ parse_url(Url) ->
...
+    Scheme = parse_scheme(SchemeBin, Url),
+    #{
+        scheme => Scheme,
+        host_template => Host,
+        port => emqx_maybe:define(Port, default_port(Scheme)),
+        path => emqx_utils_uri:path(Parsed),
+        query => emqx_maybe:define(emqx_utils_uri:query(Parsed), <<>>)
+    }.
+
+one_off_static_url(
```
savonarola: `Url` seems not needed by this fun 🤔 It may simply throw `invalid_url`/`no_host`/... etc. and the caller may attach the Url to the message

apps/emqx_auth/src/emqx_auth_http_utils.erl
```diff
@@ -59,9 +94,48 @@ transform_header_name(Headers) ->
...
+
+Returns `{pooled, {RequestBase, Path, Query}}` when the authority is fixed
+and `Resolution` is `static`, `{static_host, #{scheme, host, port, path, query}}`
+when the authority is fixed but `Resolution` is `dynamic`, or
+`{templated_host, #{scheme, host_template, port, path, query}}` whenever the
+host is templated (regardless of `Resolution`; callers decide whether that is
+allowed). Only the host may be templated: the scheme must be literal
+`http`/`https` and the port, if present, must be a literal integer.
+""".
+-spec parse_url_template(binary(), hostname_resolution()) ->
```
savonarola: Nit. this fun is a bit difficult to reason about 🤔 
* it returns three different shapes
* the return values are tagged with notation of the caller (`pooled`) 

Also not quite clear why it allows static resolution + templated host althougth this is obviously an invalid pair

maybe add a more unified return types like

```
url_template() :: 
  #{
        scheme := http | https,
        host := {template, binary()} | {static, binary()},
        port := inet:port_number(),
        path := request_path(),
        query := request_query()
    }.
parse_url_template(binary(), hostname_resolution()) :: url_template().
%% throws on templated host + static resolution;
%% maybe treats constant host + dynamic resolution as template
```

Convert `#{host := {static, ...}}` to legacy tuples only on passing to the ehttpc resource.

apps/emqx_auth/src/emqx_auth_http_utils.erl
```diff
@@ -11,15 +11,50 @@
...
+%% `host_template` rendered per request and checked against `allowed_hosts`.
+%% `pool` is a hackney pool name, or `false` for no connection reuse.
+-type one_off_base() :: #{
+    scheme := http | https,
+    port := inet:port_number(),
+    connect_timeout := timeout(),
+    ssl_opts := [ssl:tls_client_option()],
+    pool := atom() | false,
+    static_host => binary(),
+    host_template => emqx_template:t(),
```
savonarola: Nit: mb make less fuzzy,

```erlang
host => {static, binary()} | {template, emqx_template:t(), [allowed_host()]} 
```
or just remove `static_host` at all, treat is as a constant template which is a no-op to interpolate.

savonarola (commented):
I'd propose to get rid of `emqx_bridge_http_connector`, and use hackney requests for all cases.

Using bridges' resources was always an overcomplicated compromise, now we add more complexity and potentially diverging behavior.

zmstone:
> I'd propose to get rid of emqx_bridge_http_connector, and use hackney requests for all cases.

Maybe in the next version (v7). 
For static hosts, the decision must be test result based.
I have made it possible to configure `hostname_resolution = dynamic` even for static hosts, so it should be easy to compare performance side-by-side with just a config change.

savonarola (approved):

Issue #17825, opened 2026-07-01
feat(gcp_pubsub): support fully-qualified topic paths for cross-project topics


zmstone:
## Summary

The GCP Pub/Sub connector (producer and consumer) currently assumes the Pub/Sub topic lives in the same GCP project as the service account credentials — the project ID is read from the service account JSON and prepended to the topic name when building REST API paths. In practice, GCP service accounts are frequently granted access to Pub/Sub topics in other projects (a common multi-team / shared-topic setup), and users have to work around this by creating one connector per project just to switch service accounts.

Request: accept fully-qualified topic paths in the form `projects/<project-id>/topics/<topic-name>` (the canonical Pub/Sub resource name). When such a value is provided, the connector should use its `<project-id>` when constructing the REST path instead of the service account's own project. Bare `<topic-name>` continues to work and picks up the service account's project as today.

## Where the assumption lives

All code paths that hard-code the project from the service account, on `master`:

- Producer publish (`apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_impl_producer.erl:455-456`):

    ```erlang
    publish_path(#{project_id := ProjectId}, #{pubsub_topic := PubSubTopic}) ->
        <<"/v1/projects/", ProjectId/binary, "/topics/", PubSubTopic/binary, ":publish">>.
    ```

- Producer topic health check (`apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_client.erl:317-320`):

    ```erlang
    pubsub_get_topic(Topic, ClientState, ReqOpts) ->
        #{project_id := ProjectId} = ClientState,
        Method = get,
        Path = <<"/v1/projects/", ProjectId/binary, "/topics/", Topic/binary>>,
    ```

- Consumer worker — subscription create / topic resource strings (`apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_worker.erl:730, :743, :768`):

    ```erlang
    TopicResource = <<"projects/", ProjectId/binary, "/topics/", PubSubTopic/binary>>,
    %% and
    <<"projects/", ProjectId/binary, "/subscriptions/", SubscriptionId/binary>>.
    ```

## Proposed behaviour

- The `pubsub_topic` action config accepts either:
  - Bare topic name: `my-topic` — resolved against the service account's project (today's behaviour).
[...]

PR #18137, opened 2026-07-28
feat(gcp_pubsub): support fully-qualified cross-project topic paths

zmstone:
Release version:

6.0.4, 6.1.4, 6.2.3, 6.3.0

## Summary

Implements https://github.com/emqx/emqx/issues/17825.

The GCP Pub/Sub producer and consumer topic configuration now accepts a fully-qualified
topic path `projects/<project-id>/topics/<topic-name>` in addition to a bare topic name.
A fully-qualified path targets the topic in `<project-id>` (which may differ from the
service account's own project) when building the REST resource paths; the service
account still authenticates the requests. Bare topic names keep resolving against the
service account's project, as before.

Crucial points:

* A single resolver, `emqx_bridge_gcp_pubsub_client:resolve_topic/2`, returns the
  `{ProjectId, TopicName}` pair used by all three path-building sites: producer publish
  path, producer topic health check (`pubsub_get_topic`), and the consumer worker's
  topic resource (subscription create/patch bodies).
* The consumer **subscription** is still created in the service account's project; only
  the topic reference may point to another project (per the issue, moving the
  subscription is out of scope).
* The consumer subscription id derivation maps slashes in a fully-qualified topic to
  `-`, since slashes cannot appear in subscription ids (percent-encoding does not help,
  as the server decodes the path). Ids for bare topic names are unchanged.
* Schema validation accepts a bare name or exactly `projects/<id>/topics/<name>`, and
  rejects other slash-containing values with a clear error. This is backward compatible:
  previously-working configurations are unaffected (other multi-slash values could never
  have worked at runtime).

## PR Checklist


apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_worker.erl
```diff
@@ -753,12 +753,18 @@ do_get_subscription(State) ->
...
     %% doesn't...  We currently validate bridge names to not include that character.  The
     %% exception is the prefix from the probe API.
     BridgeName1 = to_bin(BridgeName0),
     BridgeName = binary:replace(BridgeName1, <<":">>, <<"-">>),
+    %% A fully-qualified `projects/<project-id>/topics/<topic-name>' topic contains
+    %% slashes, which cannot appear in a subscription id (percent-encoding them does not
+    %% help: the server decodes the path back to slashes); map them to `-' to keep the
+    %% id unique per project + topic.  Bare topic names cannot contain slashes and are
+    %% thus unaffected.
+    Topic = binary:replace(Topic0, <<"/">>, <<"-">>, [global]),
```
thalesmg: what arrived here is what resolve_topic calls topic name.  maybe use that function to extract the topic name?

thalesmg (dismissed):

thalesmg (approved):

Issue #17747, opened 2026-06-26
Use emqx_ctl join cluster report 500 INTERNAL_ERROR: error


ShuLian1984:
### What happened?

I am deploying EMQX 5.8.9 on ubuntu 22.04 and set discovery_strategy manual ，All node are allowed to communicate with each other on all ports.
- create 3 independent core node in the beginning. Running normally.
- use `emqx_ctl cluster join` commad add frist core or second core node to master node(10.10.91.224), dashboard received "500 INTERNAL_ERROR: error" and error log in emqx.log.
- and chrome console error message: "GET http://10.10.91.224:18083/api/v5/monitor_current 500 (Internal Server Error)"
- Running emqx_ctl cluster status repeatedly show all nodes as running without problem

help, any one tell me how to fix it?

<img width="1071" height="919" alt="Image" src="https://github.com/user-attachments/assets/393b589b-4e01-4a37-96e4-41db3bfbb4e9" />

### What did you expect to happen?

Running normally. not any error message

### How can we reproduce it (as minimally and precisely as possible)?

install emqx
```bash
curl -s https://assets.emqx.com/scripts/install-emqx-deb.sh | sudo bash
apt update && apt install emqx=5.8.9
```
and  not modify any other files exclude `/etc/emqx/emqx.conf`, `/etc/emqx/emqx.conf` content:
```bash
listeners {
  ssl {
    default {
      enabled = false
    }
  }
  tcp {
    default {
      acceptors = 16
      access_rules = [
        "allow all"
      ]
      bind = "0.0.0.0:1883"
      enable = true
      enable_authn = true
      max_conn_rate = infinity
      max_connections = infinity
      mountpoint = ""
      proxy_protocol = true
      proxy_protocol_timeout = "3s"
      tcp_options {
        active_n = 100
        backlog = 1024
        buffer = "4KB"
        high_watermark = "1MB"
        keepalive = none
        nodelay = true
        nolinger = false
        reuseaddr = true
        send_timeout = "15s"
        send_timeout_close = true
      }
      zone = default
    }
  }
  ws {
    default {
      enabled = false
    }
  }
  wss {
    default {
      enabled = false
    }
  }
}

node {
  # modify to node ip
  name = "emqx@10.10.91.224"
  cookie = "4yiulncsRnx2iseQnemyxrg"
  data_dir = "/var/lib/emqx"
}

cluster {
  name = "emqx"
  discovery_strategy  =  manual
}

dashboard {
    listeners {
      http {
[...]

PR #18139, opened 2026-07-28
fix: handle RPC transport failures in BPAPI proto callers (dev-58 audit findings 4-13)

zmstone:
Release versions:
5.8.12, 5.9.4, 5.10.5

## Summary

Caller-side fixes for the open findings (4-13) of the RPC error-handling audit that followed #17747 (fixed by #18107): several BPAPI proto callers did not handle the transport-failure shape their proto can actually produce, so a target node going down (or a remote crash) between the `running_nodes` check and the RPC escaped as an HTTP 500 or crashed a server process.

- `emqx_bridge_api` / `emqx_bridge_v2_api` / `emqx_connector_api`: the shared `is_ok/1` helper had no `{badrpc, _}` clause; per-node start/stop/restart operations now return 503 with a clear message instead of crashing in `lists:filter/2`.
- `emqx_delayed_api`: DELETE of a delayed message on an unreachable node returns 400 (like the GET sibling) instead of a `case_clause` 500.
- `emqx_mgmt_api_configs`: the text/plain `GET /configs` path returns a structured `BAD_NODE` error (like the JSON sibling) instead of crashing in `hocon_pp:do/2`.
- `emqx_mgmt_api_relup`: a `{badrpc, _}` result is no longer serialized into a 200 response body; the cluster-wide status endpoint no longer asserts an empty bad-nodes list.
- `emqx_mgmt_api_trace`: trace log streaming handles all `{badrpc, _}` shapes (previously only `nodedown`), returning 503.
- `emqx_retainer_mnesia`: reindexing tolerates nodes that are unreachable during the wait-dispatch-complete multicall instead of a `badmatch` crash.
- `emqx_ft`: the export-file download endpoint catches all erpc transport errors and remote crashes (503 instead of 500); the assembler converts erpc raises from the fragment `pread` proto into `{error, _}` so it stops cleanly instead of crashing the gen_statem.

No proto modules were touched and no proto versions bumped; erpc-backed protos keep their transport, callers' catches were widened instead.

Audit finding 14 (`emqx_cm_proto_v2/v3:takeover_finish` misleading spec + `wrap_rpc/1` "dead" badrpc clause) is intentionally not included: `emqx_cm:wrap_rpc/1` is shared by rpc:call-backed protos for which the clause is live, and the spec fix would touch a frozen proto module.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge.app.src
+++ b/apps/emqx_bridge/src/emqx_bridge.app.src
@@ -1,7 +1,7 @@
 %% -*- mode: erlang -*-
 {application, emqx_bridge, [
     {description, "EMQX bridges"},
-    {vsn, "0.2.10"},
+    {vsn, "0.2.11"},
     {registered, [emqx_bridge_sup]},
     {mod, {emqx_bridge_app, []}},
     {applications, [
--- a/apps/emqx_bridge/src/emqx_bridge_api.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_api.erl
@@ -1047,6 +1047,10 @@ is_ok(OkResult = {ok, _}) ->
     OkResult;
 is_ok(Error = {error, _}) ->
     Error;
+is_ok({badrpc, _} = Error) ->
+    %% Single-node operations use `rpc:call' which returns `{badrpc, Reason}'
+    %% when the target node is unreachable or the remote call fails.
+    {error, Error};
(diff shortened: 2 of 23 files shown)


zmstone:
fixed in v6, leaving v5 alone

v6 pr for ref: https://github.com/emqx/emqx/pull/18287

PR #18146, opened 2026-07-28
fix(mgmt): canonicalize request path in API scope lookup

zmstone:
Release version: 6.0.4

## Summary

Match dashboard and management-API requests against the scope catalog using the same request-path canonicalization the HTTP router applies when selecting a handler — per-segment percent-decoding and `.`/`..` resolution — so equivalent forms of a path are authorized consistently.

The change is localized to the path-to-scope matcher (`apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl`): concrete request-path segments are canonicalized with the router's own decoder (`cow_uri:urldecode/1`) before being compared to the route templates. Decoding is done per segment so an encoded separator stays within its segment.

## PR Checklist

### DIFF
--- a/apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
+++ b/apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
@@ -129,19 +129,24 @@ serialize_role(#{?role := Role, ?namespace := Namespace}) when is_binary(Namespa
     <<"ns:", Namespace/binary, "::", Role/binary>>.
 
 check_login_user_scopes_strict(Username, Path) ->
-    %% Always work on the effective scope list (role-default expanded)
-    %% so administrators with no explicit scopes implicitly hold the
-    %% full catalog and viewers implicitly hold the common scopes.
-    %% Explicit [] is honoured as "no permissions".
-    Scopes = emqx_dashboard_admin:effective_scopes_of(Username),
     case emqx_mgmt_api_key_scopes:classify_path(Path) of
         %% Explicitly public endpoint — allow regardless of scopes.
-        public -> true;
-        %% Path maps to no known scope: fail closed rather than open, so
-        %% a catalog gap cannot silently grant a scope-restricted user
-        %% access to an unmapped management endpoint.
-        not_found -> false;
-        {scope, PathScope} -> lists:member(PathScope, Scopes)
+        public ->
+            true;
+        %% Path maps to no known scope. Fail closed only for users that
+        %% carry an explicit scope list (deliberately restricted), so a
+        %% catalog gap cannot silently grant them an unmapped endpoint.
+        %% Users with no explicit scopes are not scope-restricted and stay
+        %% governed by role-based RBAC alone, so they are not locked out.
+        not_found ->
+            emqx_dashboard_admin:scopes_of(Username) =:= undefined;
+        {scope, PathScope} ->
+            %% Work on the effective scope list (role-default expanded) so
+            %% administrators with no explicit scopes implicitly hold the
+            %% full catalog and viewers implicitly hold the common scopes.
+            %% Explicit [] is honoured as "no permissions".
+            Scopes = emqx_dashboard_admin:effective_scopes_of(Username),
+            lists:member(PathScope, Scopes)
     end.
 
 %% Whitelist of self-service paths that may skip the login-user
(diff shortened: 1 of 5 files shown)


zmstone:
Note on the design, for reviewers:

**Why the login-user scope check normalizes a concrete path instead of using the route template.**
The API-key scope check resolves against `maps:get(path, HandlerInfo)` — the route template (e.g. `/clients/:clientid`) that cowboy already matched — so it is canonical by construction and needs no path matching of its own (see the comment on `emqx_mgmt_auth:check_scopes/2`). The login-user predicate (`emqx_dashboard_rbac:check_login_user_scopes/2`) cannot simply do the same, because the *same* predicate also implements the self-service exemption (a user may hit their own `/users/<self>/change_pwd` and `/mfa` without the user-management scope). That exemption needs the **concrete username** from the request; the template only carries the `:username` placeholder. So the predicate operates on the concrete request path, and therefore must canonicalize it to agree with the path the router dispatched on. (Unifying the two onto `HandlerInfo` — template for the scope lookup, request bindings + operation id for the self-service check — is a reasonable larger refactor, but out of scope for this fix / backport.)

**Why the normalization is hand-rolled rather than taken from a library.**
It must reproduce `cowboy_router`'s canonicalization *exactly*, or a residual mismatch remains:
- `cowboy_router:split_path/1` does precisely what we need (per-segment `cow_uri:urldecode/1` + `remove_dot_segments`), but it is not exported.
- `uri_string:normalize/1` (stdlib) does RFC 3986 syntax normalization: it removes dot-segments, but it only percent-decodes *unreserved* characters and leaves *reserved* ones encoded — e.g. `uri_string:normalize(<<"/a%2Fb">>)` stays `<<"/a%2Fb">>`, whereas the router's `cow_uri:urldecode(<<"a%2Fb">>)` yields `<<"a/b">>`. Using it would leave the matcher out of step with dispatch on reserved-character encodings.

So we reuse the router's own exported decoder (`cow_uri:urldecode/1`) per segment and copy `remove_dot_segments` verbatim from `cowboy_router`, which guarantees byte-for-byte agreement with routing.

savonarola:
My ideas are the following:

## Additionally get rid of blacklist logic and permissive defaults

`path_to_scope(Path)` maps **both** `?SCOPE_PUBLIC`  and unknown paths to `undefined` which is treated as allow in `check_login_user_scopes_strict`. We probably should allow only paths that exactly matched`?SCOPE_PUBLIC`, and forbid anything that did not match to a known scope.

## General considerations

### We must not have the same endpoints for self-service and management

Users should go to and endpoint like `/change_my_pwd` (without any user argument) to change own password, not to administrative `/users/<self>/change_pwd`.

### We must not have parallel path resolution

I remember advocating against such a model multiple times: applying authorization to paths that are then **independently** resolved to actual controllers, for the exact reason that came up here 🫠 

>we cannot simply do the same, because the same predicate also implements the self-service exemption: a user may access their own `/users/<self>/change_pwd` and `/mfa` without the user-management scope.

Every administrative action ultimately goes to a concrete module, a concrete handler function, and some inspectable request data, so we should still be able to apply any imaginable filter 🤔

The parallel resolution is a very probable source of **never-ending security bugs**, we should probably priritize getting rid of it.

zmstone:
Thanks — addressed the first point (permissive default / blacklist) in dfd476e:

- Added `emqx_mgmt_api_key_scopes:classify_path/1` → `{scope, Name} | public | not_found`.
- `check_login_user_scopes_strict/2` now **allows only** exact `?SCOPE_PUBLIC` matches and **denies** anything that maps to no known scope (fail closed), instead of collapsing both to `undefined` → allow.
- `path_to_scope/1` keeps its two-way contract (delegates to `classify_path/1`) so the API-key path (`emqx_mgmt_auth:check_path_in_scopes/2`, which has its own documented fail-open for unmapped status/prometheus endpoints) is unchanged. Happy to tighten that side too if you'd prefer symmetry.
- One behavioural note: unmapped now denies for *all* login users, including unrestricted admins. `t_all_endpoints_covered_by_scopes` guarantees no in-tree endpoint is unmapped, so no functional impact there; but an out-of-tree plugin endpoint lacking a `scopes/0` mapping would now be unreachable via dashboard login rather than silently allowed. Flagging in case that matters for the plugin story.

The two **General considerations** — a dedicated self-service endpoint (e.g. `/change_my_pwd` with no user argument) and removing the parallel path→controller resolution — I agree are the durable fix, but they're larger and out of scope for this backport. We'll take them up for 6.3. (This PR targets dev-60; the same change is backported to 5.10 in #18147.)

savonarola (dismissed):

savonarola (approved):

PR #18147, opened 2026-07-28
fix(mgmt): canonicalize request path in API scope lookup

zmstone:
Release version: 5.10.5

## Summary

Match dashboard and management-API requests against the scope catalog using the same request-path canonicalization the HTTP router applies when selecting a handler — per-segment percent-decoding and `.`/`..` resolution — so equivalent forms of a path are authorized consistently.

The change is localized to the path-to-scope matcher (`apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl`): concrete request-path segments are canonicalized with the router's own decoder (`cow_uri:urldecode/1`) before being compared to the route templates. Decoding is done per segment so an encoded separator stays within its segment.

Backport of emqx/emqx#18146 to the 5.10 line.

## PR Checklist


zmstone:
Note on the design, for reviewers:

**Why the login-user scope check normalizes a concrete path instead of using the route template.**
The API-key scope check resolves against `maps:get(path, HandlerInfo)` — the route template (e.g. `/clients/:clientid`) that cowboy already matched — so it is canonical by construction and needs no path matching of its own (see the comment on `emqx_mgmt_auth:check_scopes/2`). The login-user predicate (`emqx_dashboard_rbac:check_login_user_scopes/2`) cannot simply do the same, because the *same* predicate also implements the self-service exemption (a user may hit their own `/users/<self>/change_pwd` and `/mfa` without the user-management scope). That exemption needs the **concrete username** from the request; the template only carries the `:username` placeholder. So the predicate operates on the concrete request path, and therefore must canonicalize it to agree with the path the router dispatched on. (Unifying the two onto `HandlerInfo` — template for the scope lookup, request bindings + operation id for the self-service check — is a reasonable larger refactor, but out of scope for this fix / backport.)

**Why the normalization is hand-rolled rather than taken from a library.**
It must reproduce `cowboy_router`'s canonicalization *exactly*, or a residual mismatch remains:
- `cowboy_router:split_path/1` does precisely what we need (per-segment `cow_uri:urldecode/1` + `remove_dot_segments`), but it is not exported.
- `uri_string:normalize/1` (stdlib) does RFC 3986 syntax normalization: it removes dot-segments, but it only percent-decodes *unreserved* characters and leaves *reserved* ones encoded — e.g. `uri_string:normalize(<<"/a%2Fb">>)` stays `<<"/a%2Fb">>`, whereas the router's `cow_uri:urldecode(<<"a%2Fb">>)` yields `<<"a/b">>`. Using it would leave the matcher out of step with dispatch on reserved-character encodings.

So we reuse the router's own exported decoder (`cow_uri:urldecode/1`) per segment and copy `remove_dot_segments` verbatim from `cowboy_router`, which guarantees byte-for-byte agreement with routing.

savonarola (approved):

PR #18149, opened 2026-07-28
fix(authz): authorize delayed message replay

savonarola:
Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/111

Release version: 6.2.3

## Summary

The implementation appeared to be more extensive than it seemed because of current design limitations and several issues.

The major design limitation is:

- we must preserve the *exact* authorization context against the *exact* message that is later published
- however:
  - authorization happens against the raw packet, so there is no message yet, and it is impossible to "match" the context against the message
  - `message.publish`, the first hook where we are able to access a message, is *far* from the place where we perform authorization, so the message may undergo multiple transformations
 
### We solve this with the following pieces

- Change the order of constructing the message and performing authorization.
- We introduce a new fold callback, `message.ingress`, which we call after message construction and before authorization.
- We introduce the `emqx_authz_context` module, which builds the context, a projection of `clientinfo()`, for authorization.
- We introduce the `emqx_message_ingress` module, which encapsulates the `message.ingress` and `client.authorize` hooks (`emqx_access_control` invocation), mounting, and message publishing.

### Delayed authz design

- In `message.ingress`, `emqx_delayed` handles `$delayed/...` topics. It strips the prefix and puts the delayed metadata and authz context (former `clientinfo`) into the header. `emqx_access_control` does not know about delayed messages anymore.
- In `message.publish`, `emqx_delayed` sees the delayed metadata in the headers and stores the message together with the authz context.
- On timeout, `emqx_delayed` performs authorization and full ban checks using the saved authz context.

### Notes

* _Why do we need `emqx_authz_context` and an authz context?_ Client info is arbitrarily large, may contain sensitive data, and **contains a password**. In `emqx_delayed`, we store messages in Mnesia (on disk), so blindly saving client info is highly insecure.
* Since we use the authz context for delayed publishing in hardened mode, we must use it for regular authorization as well. Since the authz context is a projection, we must make sure that authorization backends have all the variables they use available. This expanded the changes, since we added tests for each backend.
* _Why do we need `emqx_message_ingress`?_ In `emqx_message_ingress`, we have methods that unite the `message.ingress` hook with access control, as well as mounting and publishing. By providing these functions, we localize important transformations, such as mounting, in a single place. By maximizing the use of `emqx_message_ingress` functions, we can be more certain that a message being published *correctly corresponds* to the one that was authorized. This was actually done incorrectly in some gateways.

### Incompatibilities.

They are mostly related to code that seems to be simply incorrect or insecure.

* Previously, `message.publish` hooks with priorities above the old delayed-hook priority ran both when a delayed message was scheduled and when it was replayed. The delayed hook now runs earlier and intercepts scheduling before those hooks, so they run only during replay.
* `client.authorize` and `client.check_authz_complete` will not receive the full `clientinfo()` anymore in hardened mode.
* Specifically, `OnClientAuthorize` receives an empty `password` string in hardened mode instead of the client's password.
* Several gateways had incorrect mounting logic: they mounted topics before authorization. Therefore, with `authorization.include_mountpoint = true`, ACL checks were run against topics mounted twice.
* Some gateways did not mount will messages correctly.
* Direct `emqx:publish/1` calls and management API publishes using `$delayed/...` bypass `message.ingress` and can fail to schedule the delayed message.
* Pending delayed messages created before the upgrade are dropped when replayed in hardened mode because they do not contain an authorization context. Legacy mode continues to replay them.

## PR Checklist


zmstone:
please rebase it to dev-63
also commits worth squash a bit?

apps/emqx_gateway_jt808/include/emqx_jt808.hrl
```diff
@@ -196,8 +196,8 @@
 
 %% Default Configs
 -define(DEFAULT_MOUNTPOINT, <<"jt808/${clientid}/">>).
--define(DEFAULT_UP_TOPIC, <<?DEFAULT_MOUNTPOINT/binary, "${phone}/up">>).
--define(DEFAULT_DN_TOPIC, <<?DEFAULT_MOUNTPOINT/binary, "${phone}/dn">>).
+-define(DEFAULT_UP_TOPIC, <<"${phone}/up">>).
```
zmstone, line 199: this is a breaking change for people who configured a mountpoint.
another reason to target dev-63 and add breaking- changelog

apps/emqx_modules/src/emqx_delayed.erl
```diff
@@ -130,36 +133,89 @@ create_tables() ->
...
+            {stop, {error, Reason}}
+    end;
+on_message_ingress(_Ctx, Msg) ->
+    {ok, Msg}.
+
 on_message_publish(
     Msg = #message{
         id = Id,
-        topic = <<"$delayed/", Topic/binary>>,
+        headers = #{?DELAYED_HEADER := #{delay := Delay}},
```
zmstone, line 155: will rule-engine republish to `$delayed/` topic end up here?

apps/emqx_modules/src/emqx_delayed.erl
```diff
@@ -463,27 +519,74 @@ do_publish(Key = {Ts, _Id}, Now, Acc) when Ts =< Now ->
...
+                false -> emqx:publish(remove_delayed_header(Msg));
+                true -> ignore_delayed_message_publish("client is banned", ClientId, Topic)
+            end;
+        deny ->
+            ignore_delayed_message_publish("authorization denied", ClientId, Topic)
+    end.
+
+-compile({inline, [ignore_delayed_message_publish/3]}).
+ignore_delayed_message_publish(Reason, ClientId, Topic) ->
+    ?tp(notice, ignore_delayed_message_publish, #{
```
zmstone, line 572: log warning ?

apps/emqx_exhook/src/emqx_exhook.erl
```diff
@@ -76,6 +76,8 @@ deny_action_result('client.authenticate', _) ->
     #{result => false};
 deny_action_result('client.authorize', _) ->
     #{result => false};
+deny_action_result('message.ingress', Req) ->
+    Req#{message => {error, not_authorized}};
```
zmstone, line 80: why not #{result => false}?

apps/emqx_exhook/src/emqx_exhook.erl
```diff
@@ -76,6 +76,8 @@ deny_action_result('client.authenticate', _) ->
...
 deny_action_result('message.publish', Msg) ->
     %% TODO: Not support to deny a message
```
zmstone, line 82: fail-close here too? since breaking change anyway.

apps/emqx/src/emqx_authz_context.erl
```diff
@@ -0,0 +1,69 @@
...
+    client_attrs => emqx_types:client_attrs(),
+    cn => binary(),
+    dn => binary(),
+    listener => atom(),
+    peername => emqx_types:peername(),
+    peerport => inet:port_number()
+}.
+-type t() :: legacy() | restricted().
+
+-define(RESTRICTED_KEYS, [
```
zmstone, line 40: need to double check if this covers all they keys used in all acl checks.

my agent scan found `now_time` is used in `emqx_authz_client_info`.

which I do not quite follow: it oddly takes `auth_expire_at` for `NowMs`.

apps/emqx/src/emqx_message_ingress.erl
```diff
@@ -0,0 +1,84 @@
...
+        {ok, NMsg} -> do_authorize(AuthzContext, NMsg);
+        {error, _} = Error -> Error
+    end.
+
+-spec authorize(
+    emqx_types:clientinfo(),
+    emqx_types:message()
+) -> {allow, emqx_types:message()} | deny.
+authorize(ClientInfo, Msg) ->
+    AuthzContext = emqx_authz_context:make(ClientInfo),
```
zmstone, line 52: this is hot path building a new map for each publish.

do I understand it correctly that we only build this context for external callback hooks?
For internal (code in this monorepo), we can audit the code the ensure no unrestricted keys got abused ?

savonarola:
closed in favor of https://github.com/emqx/emqx/pull/18246, will address the comments there.

Issue #18158, opened 2026-07-29
[6.1.4]In the Authentication and Authorization module, the certificate path for OAuth2 authentication is displayed incorrectly.


lina50:
### What happened?

In the Authentication and Authorization module, the certificate path for OAuth2 authentication is displayed incorrectly.

### What did you expect to happen?

 It should be displayed as a file path, just like in the HTTP connector.

### How can we reproduce it (as minimally and precisely as possible)?

 It should be displayed as a file path, just like in the HTTP connector.

<img width="1416" height="842" alt="Image" src="https://github.com/user-attachments/assets/62138bda-34ee-48c1-867d-961af0ae25af" />

### EMQX version

6.1.4-rc.1-g481774ec

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

PR #18161, opened 2026-07-29
fix(auth): persist HTTP OAuth2 TLS certificates

hjianbo:
Release version:

6.1.4

Fix: N/A

Follow-up: #18079

## Summary

Persist inline TLS certificates configured under `oauth2.ssl` for HTTP authentication and authorization as files in their existing mutable certificate directories.

Keep existing top-level TLS handling and error behavior unchanged. Add end-to-end coverage for both configuration paths.

## PR Checklist

- [x] The changes are covered with new tests
- [x] No changelog is required because this fixes the unreleased OAuth2 support before 6.1.4
- [x] No schema changes

### DIFF
--- a/apps/emqx_auth/src/emqx_auth_utils.erl
+++ b/apps/emqx_auth/src/emqx_auth_utils.erl
@@ -5,7 +5,8 @@
 -module(emqx_auth_utils).
 
 -export([
-    cached_simple_sync_query/4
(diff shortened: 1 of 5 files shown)


hjianbo:
Fix https://github.com/emqx/emqx/issues/18158#event-28636623187

zmstone (approved):

PR #18166, opened 2026-07-29
fix(cluster): allow replicant node to join while booting

### LINKED ISSUE #18157: [6.1.4]Manual cluster discovery no longer works for Core + Replicant clusters
### What happened?

Manual cluster discovery worked for Core + Replicant deployments before PR #18077. After PR #18077, a Replicant node cannot join a cluster during boot, resulting
  in a deadlock:

  1. The Replicant cannot fully boot because no Core node is available.
  2. `emqx ctl cluster join` cannot be executed because the node has not fully booted.
  3. The node can never join the cluster.

  Manual discovery should remain supported for Core + Replicant clusters.

  A Replicant should be able to bootstrap through an explicit manual join operation without requiring static, etcd, or another automatic discovery strategy.

Deploy a cluster with 2 Core nodes and 1 Replicant

 The two Core nodes were deployed successfully.

`root@emqx-0:/opt/emqx# ./bin/emqx ctl cluster status
Cluster status: #{running_nodes => ['emqx@emqx-0.hs','emqx@emqx-1.hs'],
                  stopped_nodes => []}
root@emqx-0:/opt/emqx#`

  The Replicant reports:

  no_available_core_node_at_same_version
  known_cores: #{}

  Trying to manually join the cluster:

`emqx@emqxrn-0:/opt/emqx$ ./bin/emqx ctl cluster join emqx@emqx-0.hs
Failed to join the cluster: "This node has not fully booted yet. Please retry after it is started."`

_No response_

### EMQX version

  - EMQX Enterprise: `6.1.4-rc.1-g481774ec`
  - Branch: `release-61`
  - OS: Ubuntu 22.04
  - Topology: 2 Core + 1 Replicant
  - Discovery strategy: `manual`

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

zmstone:
Release version:
6.0.4, 6.1.4, 6.2.3, 6.3.0

## Summary

Fixes #18157 — a regression from #18077.

#18077 made `emqx_cluster:join/1` refuse to join while node boot is in progress, because a join
restarts mria under the boot app-start loop and crashes whichever application is starting. That is
correct for core nodes, which boot standalone. But a replicant node cannot complete boot without a
core node to replicate from: with manual cluster discovery, the operator's `emqx ctl cluster join`
is what provides the core, so refusing it deadlocks the replicant (it can never get a core, so it
can never boot; it can never boot, so it can never join).

This PR scopes the refusal to core nodes (`is_booting() andalso mria_rlog:role() =:= core`),
restoring the pre-#18077 behavior for replicants while keeping the #18077 protection for cores.

Safety of the replicant exemption: mid-boot, the ekka `stop`/`start` callbacks are not yet
registered (`start_autocluster/0` runs later), and `mria_lib:exec_callback/2` is a no-op for
unregistered callbacks — so a replicant's mid-boot join does not stop/restart the emqx application
list (the #18077 crash mechanism). Verified end-to-end on a local 1 core + 1 replicant manual-
discovery cluster built from this branch: the replicant stalled with
`no_available_core_node_at_same_version`, the mid-boot `cluster join` succeeded, the replicant then
completed boot and appeared in `cluster status`, with no crash reports in its log. The peer-side
`can_i_join` license/permission check still applies to replicant joins.

Tests: kept `t_cluster_join_refused_while_booting` (nodes run as core in CT, so it now also
documents that the refusal is core-only) and added
`t_cluster_join_allowed_while_booting_on_replicant` (mocks `mria_rlog:role/0` to `replicant` and
asserts the join proceeds past the boot guard to the peer check).

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_cluster.erl
+++ b/apps/emqx/src/emqx_cluster.erl
@@ -30,7 +30,7 @@
 -endif.
 
 join(PeerNode) ->
-    case is_booting() of
+    case is_booting() andalso mria_rlog:role() =:= core of
         true ->
             {error,
                 "This node has not fully booted yet. "
@@ -55,9 +55,9 @@ Mark boot as in progress (`true`) or complete (`false`).
 
 Called by `emqx_machine`: set at boot entry and cleared once the boot app list
 has been started and the ekka join callbacks are registered. While the flag is
-set, `join/1` refuses to run: joining makes mria restart, and doing that under
-a boot in progress crashes whichever mria-backed application is starting at
-that moment, taking the node down.
+set, `join/1` on a core node refuses to run: joining makes mria restart, and
+doing that under a boot in progress crashes whichever mria-backed application
+is starting at that moment, taking the node down.
 
 Environments that boot without `emqx_machine` (e.g. common tests) never set
 the flag, so `join/1` is not restricted there.
--- a/apps/emqx_management/test/emqx_mgmt_cli_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_cli_SUITE.erl
@@ -147,6 +147,27 @@ t_cluster_join_refused_while_booting(_Config) ->
     ),
     ok.
 
+-doc """
+Verifies that on a replicant node `cluster join' is NOT refused while boot is
+in progress: a replicant cannot finish booting without a core node, so a
+mid-boot join is what bootstraps it under manual cluster discovery. The
+command must proceed past the boot guard to the peer check.
+""".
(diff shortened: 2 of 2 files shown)


ieQu1:
I think comments are excessive. Maybe remove them, except for the one explaining what the new testcase does?

Note that 6.3 logic will be almost entirely different.

ieQu1 (approved):

PR #18183, opened 2026-07-29
fix(prometheus): tolerate absent tables in namespaced metrics collection

zmstone:
Release version:
6.1.5, 6.2.3, 6.3.0

## Summary

Fixes the product robustness issue flagged as out-of-scope in #18179: the Prometheus collector crash-loops with `{badarg, ets:first(emqx_mt_ns)}` on every scrape of the namespaced stats when the `emqx_mt` application is not running, because `emqx_mt_state:fold_known_nss/2` does not guard against its table being absent and `emqx_mt` is not a dependency of `emqx_prometheus`.

Guard at the source, so any caller is safe and the semantics stay correct (no namespaces / no records exist when the owning feature is not running):

* `emqx_mt_state:fold_known_nss/2` returns the accumulator unchanged when `emqx_mt_ns` does not exist.
* `emqx_mt_state:is_known_ns/1` returns `false` in the same situation (reached via `emqx_mt_state:count_clients/1` on the same collect path when a specific namespace is queried).
* `emqx_authz_mnesia:record_count_per_namespace/0` and `emqx_authn_mnesia:record_count_per_namespace/0` (the sibling authz/authn gatherers on the same collect path) return an empty map when their count tables are absent, i.e. when `emqx_auth_mnesia` is not running — it is not a dependency of `emqx_prometheus` either.

Tests: a new CT case scrapes `/prometheus/namespaced_stats` with `emqx_mt` and `emqx_auth_mnesia` not started and asserts a 200 response with the namespaced session/authz/authn metrics omitted; the existing `t_namespaced_stats` case now also asserts that `emqx_sessions_count` is still emitted per namespace when `emqx_mt` is running; eunit covers `fold_known_nss/2` and `is_known_ns/1` with the table absent.

## PR Checklist

https://claude.ai/code/session_01CfQYgWJcuCABsSHMrVK4s9

### DIFF
--- a/apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
+++ b/apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
@@ -380,7 +380,12 @@ record_count(Namespace) when is_binary(Namespace) ->
 
 -spec record_count_per_namespace() -> #{emqx_config:namespace() => non_neg_integer()}.
 record_count_per_namespace() ->
-    maps:from_list(ets:tab2list(?AUTHN_NS_COUNT_TAB)).
+    try
+        maps:from_list(ets:tab2list(?AUTHN_NS_COUNT_TAB))
+    catch
+        %% `emqx_auth_mnesia' is not running: the table does not exist.
+        error:badarg -> #{}
+    end.
 
 %%--------------------------------------------------------------------
 %% QueryString to MatchSpec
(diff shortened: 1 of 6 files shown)


thalesmg (approved):

hjianbo (approved):

zmstone:
will not merge this to 6.1.4

PR #18188, opened 2026-07-30
fix(plugins): [defense-in-depth] plugin framework against malicious packages

JimMoen:
Fixes three plugin framework security issues in a single PR, applying partial restrictions on plugin code execution at the framework level (defense-in-depth).

## 1. CLI install bypasses the allow gate

`emqx ctl plugins install <name-vsn>` did not check `is_allowed_installation/1`, so a tarball that landed in the install dir (cluster replication, manual placement) could be installed without a prior `plugins allow` grant — bypassing the gate enforced by the HTTP upload path.

The CLI path now enforces the same allow gate (both `install` and `install --cluster`): refuses without a grant, verifies the local tarball against the allow entry's sha256 binding when present, consumes the grant cluster-wide on success, retains it on failure so the admin can retry.

## 2. Plugin API responses can set browser-sensitive headers

A plugin callback's returned HTTP headers were passed through to the Dashboard origin unchecked, allowing a malicious plugin to set `set-cookie`, `access-control-allow-origin`, `location`, `content-security-policy`, etc.

Plugin API responses are now restricted to an allow-list of safe headers (content metadata, caching, etag/validation, correlation ids, rate-limit headers); browser-sensitive headers (auth, cookies, CORS, redirects, security policy) and any custom header without the `x-plugin-` prefix are stripped case-insensitively.

## 3. Unbounded plugin package decompression

Package upload and cluster sync accepted tarballs with no limits on size, decompressed size, file count, path depth or extraction time — enabling DoS via oversized / high-compression-ratio / deeply-nested tarballs.

New `plugins.package_limits` configuration bounds the extraction pipeline (package write, tar table scan, timed extraction, cluster sync RPC timeout). Tar entries escaping the install dir (path traversal) are also rejected.

### Default values — open for discussion

| Config | Default | Rationale |
|---|---|---|
| `max_package_size` | `10MB` | Beam plugin packages are small (e.g. a typical plugin tarball is ~400KiB); 10MB only excludes packages with huge static assets |
| `max_decompressed_size` | `50MB` | Still generous vs. real plugin content |
| `max_file_count` | `10000` | |
| `max_path_depth` | `32` | |
| `max_extraction_time_ms` | `60s` | Also used as the RPC timeout for cluster package copying |

The `10MB` / `50MB` defaults in particular are a judgement call — please comment if they should be larger/smaller.

## Test plan

- `t_cli_install_requires_allow`: CLI install denied without grant, succeeds with grant, grant consumed on success / retained on failure
- `t_cli_install_sha256_binding`: CLI install denied on sha256 mismatch, succeeds with the matching sha256
- `t_plugin_api_forbidden_headers_filtered`: unit + HTTP-level tests that headers outside the allow-list are stripped (incl. mixed-case) and allowed / `x-plugin-`-prefixed headers pass through
- `t_package_size_limit` / `t_package_file_count_limit` / `t_package_path_depth_limit` / `t_package_decompressed_size_limit`: each limit rejects offending packages with no partial writes
- `t_tar_path_traversal`: zip-slip entries reject the whole package
- Full `apps/emqx_plugins-ct` (54 cases) and eunit (25) pass; `static_checks`, `elvis` clean

### DIFF
--- a/apps/emqx_plugins/mix.exs
+++ b/apps/emqx_plugins/mix.exs
@@ -25,6 +25,11 @@ defmodule EMQXPlugins.MixProject do
     UMP.deps([
       {:emqx, in_umbrella: true},
       :erlavro
-    ])
+    ]) ++
+      if UMP.test_env?() do
+        [{:cth_readable, "1.5.1"}]
+      else
+        []
+      end
   end
 end
--- a/apps/emqx_plugins/src/emqx_plugins.erl
+++ b/apps/emqx_plugins/src/emqx_plugins.erl
@@ -8,40 +8,40 @@
 -include_lib("emqx/include/logger.hrl").
 -include_lib("snabbkaffe/include/trace.hrl").
 
-%% Browser/security-sensitive response headers that plugin API callbacks
-%% are not allowed to set. All names must be lower-case binary because
+%% Response headers that plugin API callbacks are allowed to set. An
+%% allow-list is used instead of a deny-list because a deny-list is doomed to
+%% be incomplete — every new browser security mechanism adds another header
+%% to deny. Plugins that need a custom header must prefix it with
+%% `x-plugin-'. All names must be lower-case binary because
 %% `normalize_headers/1' lower-cases keys.
--define(PLUGIN_API_FORBIDDEN_HEADERS, [
-    %% auth
-    <<"authorization">>,
-    <<"www-authenticate">>,
-    <<"proxy-authenticate">>,
-    %% cookies
-    <<"cookie">>,
-    <<"cookie2">>,
-    <<"set-cookie">>,
-    <<"set-cookie2">>,
-    %% CORS
-    <<"access-control-allow-origin">>,
-    <<"access-control-allow-credentials">>,
-    <<"access-control-allow-methods">>,
-    <<"access-control-allow-headers">>,
-    <<"access-control-expose-headers">>,
-    <<"access-control-max-age">>,
-    <<"access-control-request-method">>,
-    <<"access-control-request-headers">>,
-    %% redirects
-    <<"location">>,
-    <<"refresh">>,
-    %% security/policy
-    <<"content-security-policy">>,
-    <<"content-security-policy-report-only">>,
-    <<"strict-transport-security">>,
-    <<"x-frame-options">>,
-    <<"x-content-type-options">>,
-    <<"referrer-policy">>,
-    <<"permissions-policy">>
+-define(PLUGIN_API_ALLOWED_HEADERS, [
+    %% content metadata
+    <<"content-type">>,
+    <<"content-language">>,
+    <<"content-disposition">>,
+    <<"content-range">>,
+    <<"accept-ranges">>,
+    %% caching
+    <<"cache-control">>,
+    <<"expires">>,
+    <<"pragma">>,
+    %% entity/validation
+    <<"etag">>,
+    <<"last-modified">>,
+    <<"age">>,
+    <<"vary">>,
+    %% misc safe
+    <<"retry-after">>,
+    %% correlation ids
+    <<"x-request-id">>,
+    <<"x-correlation-id">>,
+    %% rate limiting
+    <<"x-ratelimit-limit">>,
+    <<"x-ratelimit-remaining">>,
+    <<"x-ratelimit-reset">>
 ]).
+
+-define(PLUGIN_API_CUSTOM_HEADER_PREFIX, <<"x-plugin-">>).
 
 -ifdef(TEST).
 -include_lib("eunit/include/eunit.hrl").
@@ -695,19 +695,16 @@ map_plugin_api_result(_Other) ->
     {500, #{code => <<"INTERNAL_ERROR">>, message => <<"Invalid Plugin API Response">>}}.
 
 filter_plugin_api_headers(Headers) ->
-    maps:filter(fun(K, _V) -> not is_forbidden_header(K) end, Headers).
+    maps:filter(fun(K, _V) -> is_allowed_header(K) end, Headers).
 
-is_forbidden_header(Header) when is_binary(Header) ->
-    is_forbidden_header_lower(to_bin(string:lowercase(Header)));
-is_forbidden_header(_) ->
-    true.
+is_allowed_header(Header) when is_binary(Header) ->
+    Lower = to_bin(string:lowercase(Header)),
+    lists:member(Lower, ?PLUGIN_API_ALLOWED_HEADERS) orelse is_custom_header(Lower);
+is_allowed_header(_) ->
+    false.
 
-is_forbidden_header_lower(<<"access-control-", _/binary>>) ->
-    true;
-is_forbidden_header_lower(Lower) when is_binary(Lower) ->
-    lists:member(Lower, ?PLUGIN_API_FORBIDDEN_HEADERS);
-is_forbidden_header_lower(_) ->
-    true.
+is_custom_header(Header) ->
+    nomatch =/= string:prefix(Header, ?PLUGIN_API_CUSTOM_HEADER_PREFIX).
 
 normalize_headers(Headers) when is_map(Headers) ->
     maps:from_list([{to_bin(K), iolist_to_binary(V)} || {K, V} <- maps:to_list(Headers)]);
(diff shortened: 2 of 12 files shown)


JimMoen:
close emqx/emqx-dev-team-tasks#241
close emqx/emqx-dev-team-tasks#243
close emqx/emqx-dev-team-tasks#245

zmstone:
fix it to dev-63

apps/emqx_plugins/src/emqx_plugins_cli_utils.erl
```diff
@@ -236,19 +237,55 @@ do_disallow_installation(NameVsn, LogFun) ->
     ?PRINT(Result, LogFun).
 
 ensure_installed(NameVsn, LogFun) ->
+    %% The CLI path must enforce the same allow gate as the HTTP upload path:
+    %% a tarball that landed in the install dir (e.g. via cluster replication
+    %% or manual placement) must not be installable without an explicit
+    %% `plugins allow' grant from the admin.
+    case emqx_plugins:is_allowed_installation(NameVsn) of
```
zmstone, line 244: check sha256 sum too (if allow list has it)

apps/emqx_plugins/src/emqx_plugins_fs.erl
```diff
@@ -169,38 +189,65 @@ is_tar_present(NameVsn) ->
...
-                            ok = delete_tar_file_content(install_dir(), TarContent),
-                            {error, Reason}
-                    end;
-                {error, Reason} ->
-                    {error, Reason}
+    maybe
+        ok ?= check_package_file_size(TarGz),
+        ok ?= check_tar_table(TarGz),
+        {ok, TarContent} ?= extract_tarball(TarGz),
+        ok ?= check_decompressed_size(TarContent),
```
zmstone: check size after decompressed is maybe too late. 
i wonder if it's possible to cap a memory usage limit while extracting.

apps/emqx_plugins/src/emqx_plugins_fs.erl
```diff
@@ -169,38 +189,65 @@ is_tar_present(NameVsn) ->
...
-                                msg => "failed_to_read_after_install", reason => Reason
-                            }),
-                            ok = delete_tar_file_content(install_dir(), TarContent),
-                            {error, Reason}
-                    end;
-                {error, Reason} ->
-                    {error, Reason}
+    maybe
+        ok ?= check_package_file_size(TarGz),
+        ok ?= check_tar_table(TarGz),
```
zmstone: this should also be time bound, a killable tmp worker

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -8,8 +8,44 @@
 -include_lib("emqx/include/logger.hrl").
 -include_lib("snabbkaffe/include/trace.hrl").
 
+%% Browser/security-sensitive response headers that plugin API callbacks
+%% are not allowed to set. All names must be lower-case binary because
+%% `normalize_headers/1' lower-cases keys.
+-define(PLUGIN_API_FORBIDDEN_HEADERS, [
```
zmstone: this is doomed to be incomplete. use allow-list instead.

```
-define(PLUGIN_API_ALLOWED_HEADERS, [
    <<"content-type">>, <<"content-language">>, <<"content-disposition">>,
    <<"content-range">>, <<"accept-ranges">>,
    <<"cache-control">>, <<"expires">>, <<"pragma">>,
    <<"etag">>, <<"last-modified">>, <<"age">>, <<"vary">>,
    <<"retry-after">>,
    <<"x-request-id">>, <<"x-correlation-id">>,
    <<"x-ratelimit-limit">>, <<"x-ratelimit-remaining">>, <<"x-ratelimit-reset">>
]).
```

and if plugin wants to set arbitrary header, they must prefix with `x-plugin-`
`-define(PLUGIN_API_CUSTOM_HEADER_PREFIX, <<"x-plugin-">>).`

JimMoen:
**On the plugin trust chain**

The fixes in this PR are defense-in-depth: the framework constrains plugin behavior even after a plugin is allowed and loaded. But a more fundamental question remains — **how can users be sure the plugin package they install has not been tampered with?**

`allow_installation` answers "who may install", not "whether the package is trustworthy". The sha256 check only guarantees the uploaded bytes match what was allowed at that moment; it cannot prove the origin. An attacker can tamper with an official package, re-pack and redistribute it; a user who believes it is official allows and installs it, and the malicious code runs with the framework's full trust.

Worth evaluating separately: plugin package signing (public-key verification), binding the allow grant to official release hashes, and runtime sandboxing / capability limits.

apps/emqx_plugins/src/emqx_plugins.erl
```diff
@@ -1591,4 +1643,62 @@ ensure_no_other_version_active_rejects_running_case() ->
         ensure_no_other_version_active(Plugins, <<"demo-2.0.0">>)
     ).
 
+map_plugin_api_result_filters_forbidden_headers_test() ->
```
zmstone: move test cases to _test moduels

changes/ce/fix-18188.en.md
```diff
@@ -0,0 +1,5 @@
...
+
+* Fixed an authorization bypass where `emqx ctl plugins install` could install a plugin package without a prior `plugins allow` grant. The CLI now enforces the same allow gate as the HTTP upload API (including the sha256 check when the allow entry records one), and consumes the grant upon successful installation.
+* Plugin API callback responses are restricted to an allow-list of safe response headers; browser-sensitive headers (such as `set-cookie`, `location`, `access-control-*`, `content-security-policy`, and other authentication or security policy headers) and custom headers without the `x-plugin-` prefix are stripped.
+* Added the `plugins.package_limits` configuration to bound plugin package extraction: `max_package_size` (default `10MB`), `max_decompressed_size` (default `50MB`), `max_file_count` (default `10000`), `max_path_depth` (default `32`), and `max_extraction_time_ms` (default `60s`, also used as the RPC timeout for cluster package copying). Packages violating these limits are rejected before or during extraction. Tar entries escaping the install directory (path traversal) are also rejected.
```
zmstone:
```suggestion
Hardened the plugin framework package and runtime integrity checks.

* Now `emqx ctl plugins install` must be done after `emqx ctl plugins allow`, enforcing the same allow gate as the HTTP upload API. 
* Plugin API callback responses are restricted to an allow-list of safe response headers; browser-sensitive headers (such as `set-cookie`, `location`, `access-control-*`, `content-security-policy`, and other authentication or security policy headers) and custom headers without the `x-plugin-` prefix are stripped.
* Added the `plugins.package_limits` configuration to bound plugin package extraction: `max_package_size` (default `10MB`), `max_decompressed_size` (default `50MB`), `max_file_count` (default `10000`), `max_path_depth` (default `32`), and `max_extraction_time_ms` (default `60s`, also used as the RPC timeout for cluster package copying). Packages violating these limits are rejected before or during extraction. Tar entries escaping the install directory (path traversal) are also rejected.
```

apps/emqx_plugins/src/emqx_plugins_fs.erl
```diff
@@ -150,16 +153,34 @@ get_tar(NameVsn) ->
     TarGz = tar_file_path(NameVsn),
     case file:read_file(TarGz) of
         {ok, Content} ->
-            {ok, Content};
+            check_tar_binary_size(Content);
```
savonarola: I am not sure that I get the logic: why do we check size of a file that is 
a) already written somewhere to disc
b) even read to memory?

apps/emqx_plugins/src/emqx_plugins_fs.erl
```diff
@@ -150,16 +153,34 @@ get_tar(NameVsn) ->
...
+                        {error, _} = Err ->
+                            Err
+                    end;
                 Err ->
                     Err
             end
     end.
 
+check_tar_binary_size(Content) ->
+    case byte_size(Content) > max_package_size() of
```
zmstone: this size is already checked before file:read_file, no need to check again.

apps/emqx_plugins/src/emqx_plugins_fs.erl
```diff
@@ -169,38 +190,65 @@ is_tar_present(NameVsn) ->
...
 
--spec write_tar(name_vsn(), iodata()) -> ok.
+-spec write_tar(name_vsn(), iodata()) -> ok | {error, map()}.
 write_tar(NameVsn, Content) ->
-    TarFilePath = tar_file_path(NameVsn),
-    ok = filelib:ensure_dir(TarFilePath),
-    ok = file:write_file(TarFilePath, Content),
-    MD5 = emqx_utils:bin_to_hexstr(crypto:hash(md5, Content), lower),
-    ok = file:write_file(md5sum_file_path(NameVsn), MD5).
+    case iolist_size(Content) > max_package_size() of
```
savonarola, line 199: Also looks a bit strange without at least a check for the total number of packages.

apps/emqx_plugins/src/emqx_plugins_fs.erl
```diff
@@ -214,6 +262,177 @@ install_from_local_tar(NameVsn, InstallValidator) ->
             }}
     end.
 
+%% Read the tar entry table (names and metadata only, no content) and reject
+%% packages with too many entries, too-deep paths, or a total declared size
+%% beyond the limit before extraction. The table read itself runs in a
+%% time-bounded worker because reading a crafted archive can be slow.
+check_tar_table(TarGz) ->
```
savonarola: nit: `check_` prefix seems to be too generic (check against what?)  M.b. `validate_decompressed_storage_limits(TarGz)`? `validate_max_file_count`/`validate_max_path_deps`/`validate_max_decompressed_size` ...

apps/emqx_plugins/src/emqx_plugins_fs.erl
```diff
@@ -276,6 +495,24 @@ lib_dir(NameVsn) ->
 install_dir() ->
     emqx_config:get([?CONF_ROOT, install_dir], "").
 
+%% Defaults must mirror emqx_plugins_schema. The two-arg form is used so
```
savonarola, line 502: We probably should either check total quota for file count/occupied size or at least just max package count

apps/emqx_plugins/mix.exs
savonarola, line 1: Why do we need this change?

apps/emqx_plugins/src/emqx_plugins_fs.erl
JimMoen: The `package_limits` bounds target untrusted package *content* (the #243 DoS vector). A package-count quota guards a different vector — filling the install dir with many packages — which requires the same admin authorization as writing into the install dir directly (upload already needs an `allow` grant, and the install dir is operator-managed). I would prefer to keep this PR scoped to content bounds and track `max_package_count` as a follow-up; happy to add it here if you feel otherwise.

apps/emqx_plugins/src/emqx_plugins_fs.erl
JimMoen: Same as above: a package-count quota is a reasonable hardening but guards an admin-privilege vector, while `package_limits` here bounds untrusted package content. Suggest tracking `max_package_count` as a follow-up.

apps/emqx_plugins/mix.exs
JimMoen: `cth_readable` is the Common Test output hook used by EMQX test infra. The root Makefile compiles it only in its `test-compile` target; app-level CT targets like `apps/emqx_plugins-ct` do not, so the suite fails at init unless the app declares the dep itself — same pattern as `plugins/emqx_offline_messages/mix.exs`.

zmstone (approved):

Issue #17930, opened 2026-07-13
[6.3.0]Issue with RBAC permission updates


lina50:
### What happened?

Creating an admin user or API key with default scopes results in an error when editing the user or API key later.

### What did you expect to happen?

Update succeeded.

### How can we reproduce it (as minimally and precisely as possible)?

1. Create an admin user with scopes left blank, saving succeeds.

<img width="1438" height="850" alt="Image" src="https://github.com/user-attachments/assets/d530b382-9338-4a70-8d53-7d52e658f079" />
2. All permissions are displayed.
<img width="1199" height="286" alt="Image" src="https://github.com/user-attachments/assets/65dfe3f5-80e8-44ae-821e-2f63210d4aac" />
3.Editing the Note triggers an error.

<img width="1438" height="898" alt="Image" src="https://github.com/user-attachments/assets/630dd60d-97ec-48bc-9439-6649395366fc" />

The root cause should be that this PR distinguishes privilege scopes.https://github.com/emqx/emqx/pull/17813

### EMQX version

[6.3.0-beta.2-g025bb991 

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

`{"code":"BAD_REQUEST","message":"Privilege scopes cannot be combined with other scopes. Privilege scopes present: system, user_management, sso_management, api_key_management. Other scopes present: connections, publish, data_integration, access_control, gateways, monitoring, cluster_operations, audit, license, mfa_management. Assign either privilege scopes alone (administrator-equivalent), or non-privilege scopes alone."}`

zmstone:
fixed in https://github.com/emqx/emqx/pull/18009

PR #18196, opened 2026-07-30
fix(mgmt): accept unset-equivalent scopes on API key write

### LINKED ISSUE #18195: [6.1.4]Failed to modify the API key scope.
### What happened?

When we previously addressed this issue ：https://github.com/emqx/emqx/issues/17930, we submitted PR #18009 (https://github.com/emqx/emqx/pull/18009/changes), which modified only the user-related
  code and overlooked that API keys had the same issue.

 Updating the API key scope does not produce an error.

1 .Create an API key with the scope left blank.

Image

Image

2. Do nothing and update it directly; the update fails.

Image

_No response_

### EMQX version

6.1.4-rc.1-g663ae3ef

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

zmstone:
Release version:
6.0.4, 6.1.4, 6.2.3, 6.3.0

## Summary

Fixes #18195: creating an API key with the scope left blank and then updating it (even unchanged) failed with 400. This is the API-key counterpart of #18009 (dashboard users, #17930) — the same read-modify-write bug: GET returns the materialized role-default scope list (or the `unset` sentinel for legacy records), and re-submitting that value was rejected because the privilege-scope mutex was misapplied to the role-default mix and the request schema did not accept the sentinel.

* `emqx_mgmt_auth:write_scope_intent/2` (new, mirrors `emqx_dashboard_api:write_scope_intent/2`) normalizes the `scopes` request value to `keep | unset | {set, List}`: the `unset` sentinel and a list whose set equals the role default (order-insensitive) are both "no explicit scopes".
* `emqx_mgmt_api_api_keys` create/update paths run validation per intent: unset-equivalent writes skip the privilege mutex (as omitted scopes already did) and persist with the `scopes` field cleared (`maybe_set_scopes(Extra, unset)` removes the field), so such keys keep forward-compatible implicit scopes instead of a frozen list. Explicit lists are validated and stored verbatim; the publisher-only-`publish` invariant is unchanged.
* The `scopes` request schema is now `hoconsc:union([unset, hoconsc:array(binary())])`, matching the response shape so a GET value round-trips verbatim.

## PR Checklist

### DIFF
--- a/apps/emqx_management/test/emqx_mgmt_api_api_keys_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_api_api_keys_SUITE.erl
@@ -1460,6 +1460,10 @@ t_ee_scopes_update_unchanged_roundtrip(_Config) ->
     %% Order-insensitive: a permuted role-default list is also unset-equivalent.
     ?assertMatch({ok, _}, update_app(Name, #{scopes => lists:reverse(Default)})),
     ?assertMatch({ok, #{<<"scopes">> := <<"unset">>}}, read_app(Name)),
+    %% From the unset state, an explicit non-default list still takes effect.
+    Explicit = [?SCOPE_CONNECTIONS, ?SCOPE_MONITORING],
+    ?assertMatch({ok, _}, update_app(Name, #{scopes => Explicit})),
+    ?assertMatch({ok, #{<<"scopes">> := Explicit}}, read_app(Name)),
     delete_app(Name).
 
 -doc """
--- a/apps/emqx_management/test/emqx_mgmt_api_api_keys_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_api_api_keys_SUITE.erl
@@ -38,6 +38,11 @@
     t_ee_api_key_rejects_mixed_privilege_via_post,
     t_ee_api_key_rejects_mixed_privilege_via_put,
     t_ee_api_key_privilege_mutex_message,
+    %% Unset-equivalent scopes on create/update (#18195)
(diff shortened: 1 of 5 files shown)


apps/emqx_management/test/emqx_mgmt_api_api_keys_SUITE.erl
```diff
@@ -1435,6 +1440,69 @@ t_ee_api_key_privilege_mutex_message(_Config) ->
...
+    ?assertMatch({ok, _}, update_app(Name, #{scopes => Default})),
+    %% The role-default list is stored as "no explicit scopes".
+    ?assertMatch({ok, #{<<"scopes">> := <<"unset">>}}, read_app(Name)),
+    %% Round-trip once more with the `unset` sentinel GET now returns.
+    ?assertMatch({ok, _}, update_app(Name, #{scopes => <<"unset">>})),
+    ?assertMatch({ok, #{<<"scopes">> := <<"unset">>}}, read_app(Name)),
+    %% Order-insensitive: a permuted role-default list is also unset-equivalent.
+    ?assertMatch({ok, _}, update_app(Name, #{scopes => lists:reverse(Default)})),
+    ?assertMatch({ok, #{<<"scopes">> := <<"unset">>}}, read_app(Name)),
+    delete_app(Name).
```
thalesmg, line 1467: nit: maybe add an extra assertion trying to set an explicit non-default list after state is unset?

thalesmg (dismissed):

PR #18222, opened 2026-07-31
fix(mgmt): namespaced admin API key defaults exclude publish scope

### LINKED ISSUE #18220: [6.1.4]Namespace API key exposes `publish` scope although publishing is forbidden
### What happened?

 A namespaced API key created with the `administrator` role returns `publish` in its default `scopes` list. However, the same API key is not allowed to
  call the publish API and receives HTTP 403.

  This is inconsistent with namespaced Dashboard users, whose default scopes correctly exclude `publish`.

  A namespaced administrator API key returns publish in its default scopes, but publishing is denied at runtime.

  This makes the API key response misleading: clients may interpret publish as an effective permission, although it cannot be used.

  1. Create a managed namespace:

  ```http
  POST /api/v5/mt/ns/qa_scope_check

  2. Create a namespaced administrator API key:

  POST /api/v5/api_key
  Content-Type: application/json

  {
    "name": "qa_scope_key",
    "desc": "scope check",
    "enable": true,
    "role": "ns:qa_scope_check::administrator"
  }

  3. The response contains:

  {
    "role": "administrator",
    "namespace": "qa_scope_check",
    "scopes": [
      "connections",
      "publish",
      "data_integration",
      "access_control",
      "gateways",
      "monitoring",
      "cluster_operations",
      "system",
      "audit",
      "license"
    ]
  }

  4. Use the returned API key to call:

  POST /api/v5/publish

  Actual response:

  HTTP 403

  {
    "code": "UNAUTHORIZED_ROLE",
    "message": "You don't have permission to access this resource"
  }

_No response_

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
# paste output here
```

</details>

### OS version

  - EMQX: `6.1.4-rc.2`
  - Deployment: Enterprise
  - Endpoint: `http://10.42.3.91:18083`

### Log files

<details>

</details>

zmstone:
Release version:
6.0.4, 6.1.4, 6.2.3, 6.3.0

## Summary

Fixes #18220.

A namespaced administrator API key advertised the `publish` scope in its default scope list, but the `/publish` API is global-only (`namespace() -> undefined`) and namespaced actors are denied at runtime — that denial is by design and is unchanged here. The advertised default was also inconsistent with namespaced dashboard users, which already exclude `publish`.

Crucial changes:

* `emqx_mgmt_auth:role_default_scopes/1` is now namespace-aware (it parses the role string, mirroring `emqx_dashboard_admin:role_default_scopes/1`): a namespaced administrator key defaults to `?NS_ADMIN_COMMON_SCOPES` — the namespaced dashboard-user default (`?NS_ADMIN_ALLOWED_SCOPES`) minus the login-only scopes, exactly parallel to how the global API-key admin default is `?GENERIC_SCOPES` without the login-only scopes. Global administrator/viewer (`?GENERIC_SCOPES`) and publisher (`[publish]`) defaults are unchanged. The namespaced `role`-only bootstrap-file line now materialises the same namespace-aware default.
* An explicit scope list containing `publish` for a namespaced key is rejected with 400 (`emqx_mgmt_api_api_keys`) — on create, and on update whenever the list differs from the persisted one (a changed list is not a round-trip, so it must shed the unusable scope). Additionally, requesting a `[publish]`-only list on a namespaced key is rejected even as a verbatim round-trip, with a message explaining such a key cannot be used for anything and should be deleted and re-created without a namespace. Namespaced dashboard users already reject `publish` via the `?NS_ADMIN_ALLOWED_SCOPES` whitelist, so the two surfaces are consistent.

Backward compatibility (no migration, no read-time rewriting):

* Existing records are untouched; a stored list containing `publish` is served verbatim and keeps working (runtime behavior unchanged — publish was already denied for namespaced actors).
* A legacy `publish`-containing list (e.g. the previously materialised generic default, or a `publish`-only list) is not set-equal to the new namespaced default, so `write_scope_intent/2` keeps it as `{set, List}` — it never collapses to `unset` (which could broaden the key to the role default). `write_scope_intent/2` / `is_role_default_scopes/2` are untouched.
* Update path: a read-modify-write that re-submits the persisted scope list for the unchanged role/namespace is accepted verbatim without re-validation (`validate_update_scopes/3`), with one exception — a namespaced `[publish]`-only list, which is rejected with the delete-and-recreate hint since such a key can do nothing at all. The verbatim acceptance is needed because the legacy materialised default is a privilege-mixed list that would otherwise now trip the privilege-scope mutex (it used to escape validation by collapsing to `unset`). Updates that omit `scopes` (keep) also continue to work on legacy lists. Any role change or genuinely changed list is fully validated (including the namespaced `publish` rejection), so the publisher-demotion protection is preserved. No stored record is ever rewritten by any of this — a rejection leaves the key exactly as it was.

Drive-by test fix (pre-existing, not caused by this change): `t_ee_authorize_publisher` ended its POST `/publish` success check with the lazy eunit variant `?_assertMatch`, which builds a test fun and discards it — the publisher-can-publish assertion had never actually run in CT. Changed to `?assertMatch`; it passes.

The only observable changes: a newly created namespaced admin key's default no longer lists `publish`; an explicit `publish` for a namespaced role is rejected on create and on any scope-list change; and a namespaced `[publish]`-only update request gets a 400 with a delete-and-recreate hint. Everything already stored keeps working exactly as before.

## PR Checklist


apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
```diff
@@ -449,13 +469,34 @@ update_api_key(Name, Role, Body) ->
...
 validate_update_scopes(_Role, _Name, unset) ->
     ok;
 validate_update_scopes(Role, Name, keep) ->
     validate_scopes(Role, undefined, persisted_scopes(Name));
-validate_update_scopes(Role, _Name, {set, Scopes}) ->
-    validate_scopes(Role, Scopes, Scopes).
+validate_update_scopes(Role, Name, {set, Scopes}) ->
+    case is_unchanged_persisted_scopes(Role, Name, Scopes) of
+        true -> ok;
+        false -> validate_scopes(Role, Scopes, Scopes)
```
thalesmg: shouldn't also validate publish is not in the changed scope set if the caller inject such scope there?

apps/emqx_management/test/emqx_mgmt_api_api_keys_SUITE.erl
```diff
@@ -1546,6 +1567,121 @@ t_ee_api_key_unaffected_by_colliding_username_scopes(_Config) ->
...
+    delete_app(Name).
+
+-doc """
+Backward compatibility: a legacy namespaced administrator key stored with
+an explicit scope list containing `publish` (the pre-fix materialized
+default) is untouched — GET returns it verbatim, a read-modify-write that
+re-submits it succeeds and keeps it verbatim (neither stripped nor
+collapsed to the `unset` sentinel), and POST /publish with the key is
+still denied (403, by design).
+""".
```
thalesmg, line 1664: maybe also verify that attempting to change the scope list should reject publish in the new scope set?

apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
zmstone: Good catch — fixed in a883758781: a changed scope list on PUT now also rejects `publish` for a namespaced key, unless `publish` is already present in the persisted list (grandfathered so legacy keys keep round-tripping and may still shrink their list while keeping it).

apps/emqx_management/test/emqx_mgmt_api_api_keys_SUITE.erl
zmstone: Added t_ee_ns_key_update_rejects_new_publish_scope in a883758781: PUT introducing `publish` → 400 and record unchanged; changed list retaining a pre-existing `publish` → accepted.

apps/emqx_management/src/emqx_mgmt_api_api_keys.erl
```diff
@@ -469,20 +471,40 @@ update_api_key(Name, Role, Body) ->
...
+        ok -> validate_no_new_publish_for_namespaced(Role, Name, Scopes);
+        Error -> Error
+    end.
+
+validate_no_new_publish_for_namespaced(Role, Name, Scopes) ->
+    Persisted = persisted_scopes(Name),
+    HasPublish = is_list(Persisted) andalso lists:member(?SCOPE_PUBLISH, Persisted),
+    case HasPublish of
+        true -> ok;
+        false -> validate_no_publish_for_namespaced(Role, Scopes)
```
thalesmg: if the caller is changing the scope set, it's not a round trip.  so maybe that's a good opportunity to enforce it doesn't contain publish, instead of allowing it to persist in the list?

zmstone: this was to silently allow old scopes with only one scope `publish`.
however in this case, a better design is to return 400 with a clear message that the key should be deleted.

zmstone: fixing it

thalesmg (approved):

PR #18225, opened 2026-07-31
fix(mgmt): report why bootstrap-file API key scopes were dropped

zmstone:
Release version: 6.0.4

## Summary

When loading the API key bootstrap file, the loader drops scopes for three distinct reasons — an unknown scope name (typo), a scope not allowed for the `publisher` role, or a privilege scope that cannot be combined with other scopes — but logged all of them under a single `bootstrap_file_unknown_scopes_dropped` / "Unknown scope names ... dropped" warning. That is misleading for the latter two cases (the scope is a real, known scope, just not permitted in that combination).

Each dropped scope is now tagged with its reason, and the warning groups them under a `dropped` map (`unknown_scope`, `not_allowed_for_publisher_role`, `privilege_scope_conflict`) with an explanatory `info`. Behavior is otherwise unchanged: the key is still created/updated with the remaining valid scopes.

Main file: `apps/emqx_management/src/emqx_mgmt_auth.erl`.

## PR Checklist
- [ ] Schema changes are backward compatible or intentionally breaking (no schema changes)


apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -880,22 +882,71 @@ filter_publisher_scopes(?ROLE_API_PUBLISHER, Valid, Rejected) ->
...
+group_rejected_by_reason(Rejected) ->
+    lists:foldr(
+        fun({Scope, Reason}, Acc) ->
+            maps:update_with(Reason, fun(Scopes) -> [Scope | Scopes] end, [Scope], Acc)
+        end,
+        #{},
+        Rejected
+    ).
+
+-ifdef(TEST).
```
zmstone: move test to _tests.erl

apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -880,22 +883,45 @@ filter_publisher_scopes(?ROLE_API_PUBLISHER, Valid, Rejected) ->
...
+%% Group `{ScopeName, Reason}' pairs into `#{Reason => [ScopeName]}',
+%% preserving input order within each reason.
+group_rejected_by_reason(Rejected) ->
+    lists:foldr(
+        fun({Scope, Reason}, Acc) ->
+            maps:update_with(Reason, fun(Scopes) -> [Scope | Scopes] end, [Scope], Acc)
+        end,
+        #{},
+        Rejected
+    ).
```
thalesmg, line 923: nit

```suggestion
    maps:groups_from_list(
        fun({Scope, _Reason}) -> Scope end,
        fun({_Scope, Reason}) -> Reason end,
        Rejected
    ).
```

thalesmg (approved):

PR #18229, opened 2026-08-02
perf(resource): avoid per-message error formatting on non-action channel ids

zmstone:
Release version:
6.0.4, 6.1.5, 6.2.3, 6.3.0

## Summary

Since 6.0.0 (f810fd95989a), `emqx_resource:parse_connector_id_from_channel_id/1` builds an
`io_lib:format("Invalid action/source ID: ~p", [Id])` binary whenever the given id is not an
action/source channel id. On the buffer-worker send path this happened for **every message** of any
non-action resource — notably every cluster-link egress message — and the resulting string is always
discarded by the caller. Under high message volume the `~p` pretty-printing of the id binary shows up
as 240–305 ms `long_schedule` warnings in the buffer workers (`io_lib_pretty:printable_bin0`).

Commits (incremental):

1. **Make the invalid-id branch cheap.** Return the already-thrown `{invalid_id, Id}` term instead of
   formatting a message. Parsing semantics and the `{ok, ConnResId}` happy path are unchanged. The
   error-shape change is safe: all consumers (`emqx_resource_buffer_worker`, `emqx_resource_cache:split_channel_id/1`,
   `emqx_bridge_v2:extract_connector_id_from_bridge_v2_id/1` / `get_resource_ids/4`) only pattern-match
   `{ok, _}` vs `{error, _}`; the old message text appears nowhere else in the code base.

2. **Parse the connector id only once per query.** The parse actually ran *twice* per message: once in
   `emqx_resource_cache:get_runtime/1` (`split_channel_id`, to key the cache lookups) and once in
   `apply_query_fun` (`extract_connector_id`, to hand the connector callback its resource id). The
   connector id is now threaded from `call_query` down to `apply_query_fun`; `extract_connector_id/1`
   is removed.

3. **Split the channel id once per buffer worker, not per query.** A buffer worker serves a single
   immutable id, so the `{ConnResId, ChanId}` cache key is now computed once at worker init and kept in
   the worker state; `get_runtime/1` accepts the pre-split key. The buffer-worker hot path now performs
   no id parsing at all. The simple-query path (authn/authz etc.), which has no state to keep the key
   in, still splits once per call as before.

4. **Cleanup:** the connector id is the first element of the pre-split key `call_query` already holds,
   so it is derived there rather than duplicated in the `#rt{}` record.

## PR Checklist

### DIFF
--- a/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
+++ b/apps/emqx_resource/src/emqx_resource_buffer_worker.erl
@@ -1362,7 +1362,7 @@ handle_async_worker_down(Data0, Pid) ->
     end.
 
 -spec call_query(force_sync | async_if_possible, _, _, _, _, _, _) -> _.
-call_query(QM, Id, ChanKey, Index, Ref, Query, QueryOpts) ->
+call_query(QM, Id, {ConnResId, _} = ChanKey, Index, Ref, Query, QueryOpts) ->
     ?tp(call_query_enter, #{id => Id, query => Query, query_mode => QM}),
     case emqx_resource_cache:get_runtime(ChanKey) of
         %% This seems to be the only place where the `rm_status_stopped' state matters,
@@ -1377,8 +1377,7 @@ call_query(QM, Id, ChanKey, Index, Ref, Query, QueryOpts) ->
             st_err = #{status := Status},
             cb = Resource,
             query_mode = QueryMode,
-            channel_status = ChanSt,
-            conn_res_id = ConnResId
+            channel_status = ChanSt
         }} ->
             IsAlwaysSend = is_always_send(QueryMode),
             case Status =:= ?status_connected orelse IsAlwaysSend of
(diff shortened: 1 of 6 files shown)


apps/emqx_resource/src/emqx_resource_buffer_worker.erl
```diff
@@ -1367,12 +1377,13 @@ call_query(QM, Id, Index, Ref, Query, QueryOpts) ->
             st_err = #{status := Status},
             cb = Resource,
             query_mode = QueryMode,
-            channel_status = ChanSt
+            channel_status = ChanSt,
+            conn_res_id = ConnResId
```
thalesmg: since `ChanKey = {ConnResId, ChanResId | ?NO_CHANNEL}`, maybe there's no need to store the conn res id in the rt record?

thalesmg (dismissed):

thalesmg (approved):

PR #18231, opened 2026-08-03
chore(release): serve plugin download links from the emqx.com CDN

zmstone:
Internal change to the release download-link tooling (`scripts/rel/print-download-links.py`); no runtime or user-facing behavior.

Plugin package links now point at the emqx.com CDN instead of the S3 host:

- Before: `https://packages.emqx.io/emqx-plugins/6.0.3/emqx_backup_sync-0.1.1.tar.gz`
- After: `https://www.emqx.com/downloads/emqx-plugins/6.0.3/emqx_backup_sync-0.1.1.tar.gz`

Also updated the module docstring and the comment above `PLUGINS_BASE_URL`, which stated plugins were fetched straight from S3.

### DIFF
--- a/scripts/rel/print-download-links.py
+++ b/scripts/rel/print-download-links.py
@@ -12,8 +12,8 @@ It uses only the Python standard library so CI does not need jq or yq installed.
 
 Two families of URLs are printed:
   - EMQX packages, served from the emqx.com download CDN.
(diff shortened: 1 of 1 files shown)


thalesmg (dismissed):

zmstone:
removed the `/en` segment from the plugin download link, turns out to be unnecessary

PR #18232, opened 2026-08-03
feat(sec): use hardened security profile by default

savonarola:
Release version:
6.3.0

## Summary

Use the hardened security profile by default in EMQX 6.3 when `EMQX_SECURITY_PROFILE` is not set.

## PR Checklist


changes/ee/breaking-18232.en.md
```diff
@@ -0,0 +1 @@
+Changed the default EMQX security profile from `legacy` to `hardened`. Default MQTT and Dashboard listeners now bind to loopback; authentication, authorization, and access-control failures fail closed; unchanged Dashboard default credentials are rejected; outbound TLS verifies peers; missing JWTs are rejected; SAML signatures are required; and internal subscriptions are authorized. Set `EMQX_SECURITY_PROFILE=legacy` to retain the previous behavior during migration.
```
zmstone: we will put this to the head of the 6.3 release note to highlight.
please describe each breaking behavior with bit more information, before vs now.

changes/ee/breaking-18232.en.md
```diff
@@ -1 +1,16 @@
-Changed the default EMQX security profile from `legacy` to `hardened`. Default MQTT and Dashboard listeners now bind to loopback; authentication, authorization, and access-control failures fail closed; unchanged Dashboard default credentials are rejected; outbound TLS verifies peers; missing JWTs are rejected; SAML signatures are required; and internal subscriptions are authorized. Set `EMQX_SECURITY_PROFILE=legacy` to retain the previous behavior during migration.
+Changed the default EMQX security profile from `legacy` to `hardened`. This implies multiple incompatible defaults.
+
+- **MQTT listener exposure is restricted by default.** MQTT TCP, SSL, WS, and WSS listeners bind to loopback if no bind address is configured.
+- **Dashboard HTTP exposure is restricted by default.** The Dashboard HTTP listener now listens on loopback if no bind address is configured.
+- **Explicit authentication is required.** Clients are denied when no authentication, either global or listener-specific, is configured, or when all authenticators are disabled. To explicitly allow anonymous access, set the listener's `enable_authn = false`.
```
zmstone: we have demoted listener-specific auth chain, so no mention here.

changes/ee/breaking-18232.en.md
```diff
@@ -1 +1,16 @@
...
+- **Authentication failures are not ignored.** Authenticator backend errors, malformed responses, precondition failures, and unavailable verification keys **immediately deny** access instead of continuing to another authenticator. Set `authentication_settings.ignore_backend_failures = true` to allow fallback to later authenticators.
```
zmstone: need clarification on "precondition failures".

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -1231,16 +1231,13 @@ binfmt(Fmt, Args) ->
     iolist_to_binary(io_lib:format(Fmt, Args)).
 
 default_username() ->
-    binenv(default_username).
+    emqx:get_config([dashboard, default_username], "").
```
zmstone: emqx:get_config([dashboard, default_username], <<>>).

apps/emqx_gateway/test/emqx_gateway_test_utils.erl
```diff
@@ -239,22 +239,36 @@ collect_emqx_hooks_calls(Acc) ->
...
 disable_gateway_auth(Gateway) ->
+    ok = set_gateway_listeners_authn(Gateway, false),
     Path = io_lib:format("/gateways/~ts/authentication", [Gateway]),
     {204, _} = request(delete, Path),
     ok.
 
+set_gateway_listeners_authn(Gateway, Enable) ->
+    lists:foreach(
+        fun(#{type := Type, name := Name}) ->
+            ListenerConf = emqx:get_config([gateway, Gateway, listeners, Type, Name]),
```
zmstone: Question: this reads *checked* config (atom keys, resolved defaults, `bind` as a tuple) and feeds it back into `emqx_gateway_conf:update_listener/3`, while the suite-local helper this replaces used `emqx_conf:get_raw` with binary keys. It evidently passes today, but round-tripping checked values through `pre_config_update`/schema check is fragile — resolved fields don't always re-validate (and any converter runs again on already-converted values). Any reason not to keep the `get_raw` approach here?

changes/ee/breaking-18232.en.md
```diff
@@ -0,0 +1,16 @@
+Changed the default EMQX security profile from `legacy` to `hardened`. This implies multiple incompatible defaults.
+
+- **MQTT listener exposure is restricted by default.** MQTT TCP, SSL, WS, and WSS listeners bind to loopback if no bind address is configured.
```
zmstone: Two things this bullet under-states, both worth making explicit since this is the migration reference:

1. QUIC listeners get the same treatment — `base_listener/1` gives quic the bare-port default `14567`, and `listener_ip_port_converter` applies to every `base_listener` bind, so the default becomes `127.0.0.1:14567` too.
2. It's not only *unset* binds: an explicitly configured port-only bind (`bind = 1883`) is also rewritten to `127.0.0.1:1883` by `maybe_add_loopback_converter`. "binds to loopback if no bind address is configured" reads like only defaults are affected; suggest "...if the bind address is unset or specifies only a port".

savonarola: ✅

apps/emqx_gateway/test/emqx_gateway_test_utils.erl
savonarola: Raw config is more correct
✅

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
savonarola: ✅

changes/ee/breaking-18232.en.md
savonarola: ✅

changes/ee/breaking-18232.en.md
savonarola: ✅

zmstone (dismissed):

zmstone (approved):

Issue #18218, opened 2026-07-31
Support Username and Source IP Flapping Protection


zibuyu1995:
### What would you like to be added or enhanced?

# Support Username and Source IP Flapping Protection

## Background

Client ID-based flapping detection is insufficient when clients use random Client IDs. A malfunctioning or malicious client may repeatedly reconnect using the same username or source IP, creating excessive authentication requests and broker load.

## Requirement

Extend flapping protection to support:

* Username
* Source IP address

Detection should be based on disconnect frequency within a configurable time window. The threshold, detection window, and temporary blacklist duration should be configurable independently.

Node-local detection is sufficient; cluster synchronization is not required.

## Expected Behavior

When a username or source IP exceeds the configured threshold:

* Add it to a temporary blacklist.
* Reject new matching connection attempts.
* Perform the check before calling external authentication services when possible.
* Automatically expire the blacklist entry and support manual removal.

Adding an entry to the blacklist must not disconnect existing clients using the same username or source IP. It should only block subsequent connection attempts.

## Observability

Expose metrics and logs for:

* Current blacklisted usernames and source IPs
* Flapping trigger events
* Connections rejected by the blacklist

### Why is this needed?

This is not limited to external authentication. A leaked credential or a large number of abnormal clients behind the same source IP can also continuously connect and disconnect, potentially affecting broker availability.

PR #18239, opened 2026-08-04
feat(flapping): detect flapping by username and source IP address

zmstone:
Release version: 6.3.0
Introduced In: N/A

## Summary

Implements #18218: flapping protection by username and source IP address, in addition to the existing client ID dimension.

**Detection** (`apps/emqx/src/emqx_flapping.erl`): the flapping table now holds one counter row per dimension, keyed `{clientid | username | peerhost, Value}` (the table is in-memory and rebuilt on restart, so the key change has no compatibility impact). Each connect event is counted independently per enabled dimension against that dimension's own policy (`window_time`, `max_count`, `ban_time`). Counters remain node-local (ETS), as the issue deems sufficient. Anonymous connections (no username) are not counted towards the username dimension. The table GC match spec is now generated with `ets:fun2ms` instead of a hand-written positional pattern that would silently break on record changes.

**Enforcement**: reuses `emqx_banned` — on trigger, the offending value is banned with `who(username | peerhost, Value)`. `emqx_banned:check/1` already checks client ID, username, and peerhost, and `check_banned` already runs before authentication in the connect pipeline (`emqx_channel:process_connect/2`), so rejection happens before external auth services are called, banned entries auto-expire, and they can be listed/removed early via the existing `/banned` REST API ("current blacklisted usernames/IPs" observability). Adding a ban entry does not disconnect existing sessions — same semantics the client ID dimension always had. The resulting ban is cluster-replicated via mnesia, which is more than the issue's node-local requirement (a bonus, not a regression); a separate node-local blacklist with a new pre-auth check would have duplicated all of this.

**Schema** (`apps/emqx/src/emqx_schema.erl`): `flapping_detect` is now symmetric across dimensions — `by_clientid`, `by_username`, `by_peerhost`, each a union of `none` (disabled, the default for all three) and a policy struct (`window_time`, `max_count`, `ban_time`); configuring the struct enables the dimension, so there is no separate `enable` flag. The pre-6.3 flat fields (`enable`, `window_time`, `max_count`, `ban_time`) are deprecated (hidden) and a field converter lifts them into `by_clientid` wherever the struct is checked (global root, zones, global-zone config API), so existing configs and legacy-style API updates keep working unchanged: `enable = true` converts the flat params into the dimension struct, while `enable = false` — or flat params without an explicit `enable = true` — convert to `none`. A present `by_clientid` (struct or `none`) prevails over the deprecated fields unconditionally — converters run before schema defaults are filled, so a present value came from the input, and root config updates replace the root raw value, so pure legacy-layout API payloads never carry a stale `by_clientid`. This precedence also means a legacy `enable = true` pinned in `etc/emqx.conf` can be durably overridden from the dashboard: the `by_clientid = none` written to `cluster.hocon` survives restarts. The converter output drops the deprecated fields, so stored raw config converges to the new shape and `GET /configs/global_zone` (which fills defaults through the same converter) always serves the new shape — no separate API compatibility layer needed. Zone overrides of dimension structs are cloned into the zone schema without defaults, so dimension fields left unset in a zone inherit the global dimension config through the zone merge.

**Observability**: the `flapping_detected` log now carries the detecting dimension and value. New counters `flapping.detected.clientid`, `flapping.detected.username`, `flapping.detected.peerhost` (triggers per dimension) and `client.banned` (connection attempts rejected due to an active ban entry, counted in `check_banned`; a `RC_BANNED` CONNACK also counts towards the existing `packets.connack.error`). Exposed via `GET /metrics`; Prometheus export can be added as a follow-up (several existing counters, e.g. `client.post_authn`, are likewise not in the Prometheus meta lists).

## PR Checklist


apps/emqx/src/emqx_flapping.erl
```diff
@@ -85,12 +112,40 @@ get_policy(Zone) ->
...
+
+dimension_policy(Name, Policy) ->
+    maps:get(Name, Policy, #{enable => false}).
+
+%% The top-level policy is the client ID dimension.
+all_dimensions(Policy) ->
+    [Policy, dimension_policy(by_username, Policy), dimension_policy(by_peerhost, Policy)].
+
+enabled_window_times(Policy) ->
+    [W || #{enable := true, window_time := W} <- all_dimensions(Policy)].
```
zmstone: call lists:filtermap/2

apps/emqx/src/emqx_schema.erl
```diff
@@ -469,6 +469,60 @@ fields("flapping_detect") ->
                     default => <<"5m">>,
                     desc => ?DESC(flapping_detect_ban_time)
                 }
+            )},
+        {"by_username",
+            sc(
+                ref("flapping_detect_dimension"),
```
zmstone: union of 'none' and ref("flapping_detect_dimension"),
and for flapping_detect_dimension struct, there is no need for 'enable' field.

apps/emqx/src/emqx_schema.erl
```diff
@@ -469,6 +469,60 @@ fields("flapping_detect") ->
...
+                #{
+                    default => #{},
+                    importance => ?IMPORTANCE_MEDIUM,
+                    desc => ?DESC(flapping_detect_by_username)
+                }
+            )},
+        {"by_peerhost",
+            sc(
+                ref("flapping_detect_dimension"),
```
zmstone: same here.

apps/emqx/src/emqx_schema.erl
```diff
@@ -436,38 +436,98 @@ fields("topic_predicate_equals") ->
...
         {"enable",
             sc(
                 boolean(),
                 #{
-                    default => false,
-                    %% importance => ?IMPORTANCE_NO_DOC,
+                    required => false,
+                    importance => ?IMPORTANCE_HIDDEN,
                     desc => ?DESC(flapping_detect_enable)
                 }
```
thalesmg, line 450: Q: shouldn't it be `deprecated => ...`?

apps/emqx/src/emqx_schema.erl
```diff
@@ -436,38 +436,98 @@ fields("topic_predicate_equals") ->
...
+            )},
+        {"ban_time",
+            sc(
+                duration(),
+                #{
+                    required => false,
+                    importance => ?IMPORTANCE_HIDDEN,
+                    desc => ?DESC(flapping_detect_ban_time)
+                }
+            )},
```
thalesmg, line 481: if these had the same defaults as the ones below, then the default in emqx_flapping would not be needed?

> %% Must match the field defaults of the `flapping_detect_dimension`
%% struct in `emqx_schema`:

apps/emqx/src/emqx_schema.erl
```diff
@@ -2353,10 +2413,16 @@ desc("zone") ->
...
+    "This config limits how frequently clients may reconnect.\n"
+    "Connect events are counted per client ID (`by_clientid`), per username\n"
+    "(`by_username`), and per source IP address (`by_peerhost`), each with its\n"
+    "own detection window, threshold, and ban duration.\n"
+    "When a threshold is exceeded within the window, the offending value is\n"
+    "temporarily banned: new matching connection attempts are rejected before\n"
+    "authentication until the end of the time period defined by `ban_time`.";
+desc("flapping_detect_dimension") ->
+    "Flapping detection policy for a single dimension (client ID, username,\n"
+    "or source IP address), counted independently of the other dimensions.";
```
thalesmg, line 2429: q: shouldn't these use a hocon i18n file?

apps/emqx/src/emqx_schema.erl
```diff
@@ -4052,12 +4118,52 @@ mqtt_converter(#{<<"keepalive_backoff">> := Backoff} = Mqtt, _Opts) ->
...
+converted_by_clientid(Enable, LegacyParams, Existing) when
+    Enable =:= true; Enable =:= <<"true">>
+->
+    case Existing of
+        Dimension when is_map(Dimension) ->
+            %% Both shapes present: new-style params win over the
+            %% deprecated flat params.
+            maps:merge(LegacyParams, Dimension);
+        _ ->
+            LegacyParams
```
thalesmg: if by_clientid is explicitly none, I think it should prevail.  which would be consistent with the other case arm.

apps/emqx/src/emqx_schema.erl
zmstone: deprecated is not needed if it's hidden already.

apps/emqx/src/emqx_schema.erl
zmstone: left-overs from previous cleanup.
will fix in another pr.

apps/emqx/src/emqx_schema.erl
zmstone: there is unfortunately no way to tell if it's explicit or from default value.

will remove default from schema.

apps/emqx/src/emqx_schema.erl
zmstone: added for completeness.

apps/emqx/src/emqx_schema.erl
zmstone: actually, default is not filled yet when converter is evaluated.

apps/emqx/src/emqx_schema.erl
zmstone: thing is, the zone schema clones root schema without defaults.

apps/emqx/src/emqx_schema.erl
thalesmg: when fetching from the raw config some lines above, we could default to undefined to differentiate between absent or explicitly set.

thalesmg (approved):

Issue #18192, opened 2026-07-30
[6.1.4]When the database, username, and password are left blank in Datalayers, a function_clause error appears in the logs.


lina50:
### What happened?

When the database, username, and password are left blank in Datalayers, a function_clause error appears in the logs.

### What did you expect to happen?

Improve the log format to avoid function_clause errors.

### How can we reproduce it (as minimally and precisely as possible)?

<img width="1440" height="861" alt="Image" src="https://github.com/user-attachments/assets/a9e0596d-53a4-4fe9-bd01-1a481bf9dfda" />

### EMQX version

6.1.4-rc.1-g663ae3ef

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

[...]

PR #18242, opened 2026-08-04
fix(datalayers): handle blank database/credentials without function_clause

JimMoen:
Release version: 6.1.5

Introduced In: 6.0.0 (datalayers schema made `database` optional in PR #15544)

Fixes https://github.com/emqx/emqx/issues/18192

## Summary

When `database`, `username` and `password` are left blank in a Datalayers connector config, starting the connector crashed with `function_clause`:

- `emqx_bridge_datalayers_connector:on_start/2` dispatched to the influxdb driver, where `emqx_bridge_influxdb_connector:protocol_config/1` unconditionally matched `database := DB` in the parameters map, crashing when the key was absent.
- The Arrow Flight driver `emqx_bridge_datalayers_arrow_flight_connector:on_start/2` pattern-matched `username`/`password`/`database` as mandatory keys, so blank credentials crashed the same way.

Changes:

- `emqx_bridge_datalayers_connector.erl`: validate `database` in `on_start/2` before dispatching to a driver; missing or empty database returns `{error, {bad_config, database_required}}` instead of crashing. This covers both the influxdb and the Arrow Flight driver paths.
- `emqx_bridge_datalayers_arrow_flight_connector.erl`: fall back to empty binaries for `username`/`password` when absent, matching the underlying `datalayers` adapter which already uses `unwrap_or_default` on these options. Whether an empty credential is accepted is decided by the server side at connect time.
- Added eunit tests covering blank database (both drivers) and blank username/password (Arrow Flight driver).

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ce/fix-18242.en.md` files

### DIFF
--- a/apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
+++ b/apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
@@ -83,14 +83,12 @@ on_start(
     InstId,
     Config = #{
         server := Server,
-        parameters := #{
-            username := Username,
-            password := Password,
-            database := Database
-        } = _Parameters,
+        parameters := Parameters = #{database := Database},
         ssl := SSL
     }
 ) ->
+    Username = maps:get(username, Parameters, <<>>),
+    Password = maps:get(password, Parameters, <<>>),
     #{hostname := Host, port := Port} = emqx_schema:parse_server(
         Server, ?DATALAYERS_HOST_ARROW_OPTIONS
     ),
--- a/apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
+++ b/apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
@@ -83,12 +83,13 @@ on_start(
     InstId,
     Config = #{
         server := Server,
-        parameters := Parameters = #{database := Database},
+        parameters := Parameters,
         ssl := SSL
     }
 ) ->
     Username = maps:get(username, Parameters, <<>>),
     Password = maps:get(password, Parameters, <<>>),
+    Database = maps:get(database, Parameters, <<>>),
     #{hostname := Host, port := Port} = emqx_schema:parse_server(
         Server, ?DATALAYERS_HOST_ARROW_OPTIONS
     ),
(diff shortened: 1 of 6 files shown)


apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_flight_connector_tests.erl
```diff
@@ -0,0 +1,58 @@
...
+        fun() ->
+            meck:new(emqx_resource_pool, [passthrough, no_link]),
+            meck:expect(
+                emqx_resource_pool,
+                start,
+                fun(_, _, _) -> {error, {start_pool_failed, noproc, mocked}} end
+            ),
+            ok
+        end,
+        fun(_) -> meck:unload(emqx_resource_pool) end, fun(_) ->
```
zmstone, line 20:
```suggestion
        fun(_) -> meck:unload(emqx_resource_pool) end,
        fun(_) ->
```

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
```diff
@@ -83,14 +83,12 @@ on_start(
     InstId,
     Config = #{
         server := Server,
-        parameters := #{
-            username := Username,
-            password := Password,
-            database := Database
-        } = _Parameters,
+        parameters := Parameters = #{database := Database},
```
zmstone: why match database, but not maps:get(database, Parameters, <<>>), ?

apps/emqx_bridge_datalayers/test/emqx_bridge_datalayers_arrow_flight_connector_tests.erl
JimMoen: erlfmt keeps `fun(_) ->` at the end of the line.

apps/emqx_bridge_datalayers/src/emqx_bridge_datalayers_arrow_flight_connector.erl
JimMoen: Thanks. Considering schema compatibility, we won't change `database` to `required => true` here, and will handle it the same way as `username`/`password`.

zmstone (approved):

PR #18246, opened 2026-08-04
fix(authz): authorize delayed message replay

savonarola:
Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/111

Release version: 6.3.0

## Summary

The implementation appeared to be more extensive than it seemed because of current design limitations and several issues.

The major design limitation is:

- we must preserve the *exact* authorization context against the *exact* message that is later published
- however:
  - authorization happens against the raw packet, so there is no message yet, and it is impossible to "match" the context against the message
  - `message.publish`, the first hook where we are able to access a message, is *far* from the place where we perform authorization, so the message may undergo multiple transformations
 
### We solve this with the following pieces

- Change the order of constructing the message and performing authorization.
- We introduce a new fold callback, `message.ingress`, which we call after message construction and before authorization.
- We introduce the `emqx_authz_context` module, which builds the context, a projection of `clientinfo()`, for authorization.
- We introduce the `emqx_message_ingress` module, which encapsulates the `message.ingress` and `client.authorize` hooks (`emqx_access_control` invocation), mounting, and message publishing.

### Delayed authz design

- In `message.ingress`, `emqx_delayed` handles `$delayed/...` topics. It strips the prefix and puts the delayed metadata and authz context (former `clientinfo`) into the header. `emqx_access_control` does not know about delayed messages anymore.
- In `message.publish`, `emqx_delayed` sees the delayed metadata in the headers and stores the message together with the authz context.
- On timeout, `emqx_delayed` performs authorization and full ban checks using the saved authz context.

### Notes

* _Why do we need `emqx_authz_context` and an authz context?_ Client info is arbitrarily large, may contain sensitive data, and **contains a password**. In `emqx_delayed`, we store messages in Mnesia (on disk), so blindly saving client info is highly insecure.
* **Production behavior is unchanged** for performance. Regular authorization and its hooks continue to receive the full client info. But test builds project it to the restricted authorization context, so backend tests fail if authz depends on fields that delayed replay cannot persist.
* _Why do we need `emqx_message_ingress`?_ In `emqx_message_ingress`, we have methods that unite the `message.ingress` hook with access control, as well as mounting and publishing. By providing these functions, we localize important transformations, such as mounting, in a single place. By maximizing the use of `emqx_message_ingress` functions, we can be more certain that a message being published *correctly corresponds* to the one that was authorized. This was actually done incorrectly in some gateways.

### Incompatibilities.

They are mostly related to code that seems to be simply incorrect or insecure.

* Previously, `message.publish` hooks with priorities above the old delayed-hook priority ran both when a delayed message was scheduled and when it was replayed. The delayed hook now runs earlier and intercepts scheduling before those hooks, so they run only during replay.

* Several gateways had incorrect mounting logic: they mounted topics before authorization. Therefore, with `authorization.include_mountpoint = true`, ACL checks were run against topics mounted twice.
* Some gateways did not mount will messages correctly.
* Direct `emqx:publish/1` calls and management API publishes using `$delayed/...` bypass `message.ingress` and can fail to schedule the delayed message.
* Pending delayed messages created before the upgrade are dropped when replayed in hardened mode because they do not contain an authorization context. Legacy mode continues to replay them.

## PR Checklist


savonarola:
> please rebase it to dev-63
> also commits worth squash a bit?

✅ 

> will rule-engine republish to $delayed/ topic end up here?

Yes, and my understanding that it should 🤔 All the actual hooks should be run when the delayed message is published.

> why not #{result => false}?

✅ made ingress error similar to deny case

>need to double check if this covers all they keys used in all acl checks.
>my agent scan found now_time is used in emqx_authz_client_info.
>which I do not quite follow: it oddly takes auth_expire_at for NowMs.

`now_ms` is a way to reliably fix an issue where will message is not authorized after authz expiration and disconnect. In the PR, we pinned all the available keys for authz with tests. Also added a test for `now_time` and run additional scans.

> this is hot path building a new map for each publish.
> do I understand it correctly that we only build this context for external callback hooks?
> For internal (code in this monorepo), we can audit the code the ensure no unrestricted keys got abused ?

Now do not rebuild map for internal authzs in production, but do in tests. So if a backend tries to add a new field and test, the test will fail.

> fail-close here too? since breaking change anyway.

Implemented fail closed for `message.publish`

zmstone:
this means message republish to $delayed will not work anymore ?

changes/ee/fix-18246.en.md
```diff
@@ -0,0 +1,3 @@
+In the hardened security profile, delayed messages from MQTT and gateway clients now store a restricted authorization context and are checked against current publish authorization rules and bans when replayed.
+
+Message ingress now prepares messages before authorization and applies mountpoints consistently across MQTT and gateways. The legacy security profile retains the previous authorization context and delayed replay behavior.
```
zmstone: this is a breaking change?

also worth listing all config knobs for each gateway.

apps/emqx/src/emqx_channel.erl
```diff
@@ -2964,68 +2971,55 @@ check_pub_alias(_Packet, _Channel) ->
...
                 'authz.reason_code' => ?RC_NOT_AUTHORIZED
             }),
             ?EXT_TRACE_SET_STATUS_ERROR(),
-            {error, ?RC_NOT_AUTHORIZED}
+            {error, ?RC_NOT_AUTHORIZED, Channel};
+        {error, invalid_topic_name} ->
+            {error, ?RC_TOPIC_NAME_INVALID, Channel};
+        {error, Reason} ->
+            ?SLOG(warning, #{msg => "message_ingress_failed", reason => Reason}),
+            {error, ?RC_IMPLEMENTATION_SPECIFIC_ERROR, Channel}
```
zmstone: this error results in disconnect, right?
maybe make it respect `deny_action`, i.e. `{error, ?RC_NOT_AUTHORIZED, Channel}`

apps/emqx/src/emqx_authz_context.erl
```diff
@@ -0,0 +1,95 @@
...
+
+-doc """
+Create an authz context map from the client info.
+""".
+-spec make(emqx_types:clientinfo()) -> t().
+-ifdef(TEST).
+make(ClientInfo) ->
+    case emqx_security_profile:policy(authz_context) of
+        legacy -> ClientInfo;
+        restricted -> make_restricted(ClientInfo)
```
zmstone, line 72: this happens only when testing

PR description should be updated.

> Specifically, OnClientAuthorize receives an empty password string in hardened mode instead of the client's password.

savonarola: Good catch
~~we still need to cut password~~ no, it is not saved

changes/ee/fix-18246.en.md
savonarola: yes, now it is, after `hardened` as default profile

apps/emqx/src/emqx_authz_context.erl
savonarola: ✅

apps/emqx/src/emqx_channel.erl
savonarola: ✅

changes/ee/fix-18246.en.md
savonarola: :check:

savonarola:
>this means message republish to $delayed will not work anymore?

Made it work with `direct_dispatch` set to `false`.

changes/ee/breaking-18246.en.md
```diff
@@ -0,0 +1,14 @@
+Changed delayed publishing and message ingress behavior.
```
zmstone: sorry, we'll need to revert it back to feat- file name

changes/ee/breaking-18246.en.md
```diff
@@ -0,0 +1,14 @@
...
+
+- Delayed messages are reauthorized when replayed. Delayed messages from MQTT and gateway clients store a restricted authorization context. EMQX checks current publish authorization rules and ban records before replay. A message that was authorized when scheduled can be dropped when replayed.
+- Pending delayed messages can be dropped after an upgrade. In the default hardened security profile, EMQX drops pending delayed messages created before the upgrade because they do not contain an authorization context. The legacy profile continues to replay them.
```
zmstone:
```suggestion
- Pending delayed messages can be dropped after an upgrade. In the hardened security profile, EMQX drops pending delayed messages created before the upgrade because they do not contain an authorization context. The legacy profile continues to replay them.
```

changes/ee/fix-18246.en.md
```diff
@@ -0,0 +1,10 @@
+- Added delayed message authorization in hardened security profile.
+  - In the hardened security profile, delayed messages are reauthorized when replayed. Delayed messages from MQTT and gateway clients store a restricted authorization context. EMQX checks current publish authorization rules and ban records before replay. A message that was authorized when scheduled can be dropped when replayed.
+  - Pending delayed messages can be dropped after enabling the hardened security profile. In the hardened security profile, EMQX drops pending delayed messages created before the upgrade because they do not contain an authorization context. The legacy profile continues to replay them.
+- Fixed mountpoint handling in multiple gateways. Gateways now consistently pass logical, unmounted topics into authorization. When `authorization.include_mountpoint = false`, EMQX checks the logical topic. When it is `true`, EMQX applies the mountpoint once for the authorization check. EMQX applies the mountpoint once before publishing or subscribing.
```
zmstone, line 4:
```suggestion
- Fixed mountpoint handling in multiple gateways. Gateways now consistently pass logical, unmounted topics into authorization. When `authorization.include_mountpoint = false`, EMQX checks the logical topic. When it is `true`, EMQX applies the mountpoint once before publishing or subscribing.
```

zmstone (approved):

changes/ee/fix-18246.en.md
savonarola: The revised version changed the sense. 

>When it is `true`, EMQX applies the mountpoint once before publishing or subscribing.

In fixed version, EMQX applies the mountpoint **once** before publishing or subscribing regardless of `authorization.include_mountpoint` setting (which does not control mounting on publish or subscribe.)

PR #18252, opened 2026-08-05
test(crl_cache): assert per-process trace order in t_cache_overflow

zmstone:
Release versions:
5.8.12, 5.9.4, 5.10.5

Introduced in:
pre-5.8

## Summary

Test-only de-flake of `emqx_crl_cache_SUITE:t_cache_overflow`.

The post-condition asserted a single fixed interleaving of two trace
streams: the `mqtt_client_connection` span events (emitted by the test
client processes) and the cache events (`new_crl_url_inserted`,
`crl_cache_overflow`, `crl_cache_ensure_timer`, emitted by the single
`emqx_crl_cache` gen_server). The two streams interleave
non-deterministically, so the gen_server events sometimes landed after
the connection-complete span and broke the match.

The client-connect spans add little value, so drop them and assert only
the cache-event stream, whose global order is deterministic (single
gen_server). The final `emqtt:connect` returns before the last
re-insert / overflow / timer events are emitted, so block until the
second overflow lands before snapshotting the trace, avoiding a
truncated prefix.

Port of the v6/master fix (#17706 / #17774) back to dev-58/dev-59;
dev-510 already carries the earlier per-stream split (`dd8597582d`).

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_crl_cache_SUITE.erl
+++ b/apps/emqx/test/emqx_crl_cache_SUITE.erl
@@ -420,17 +420,16 @@ assert_successful_connection(Config, ClientNum) ->
         end
     ).
 
-trace_between(Trace0, Marker1, Marker2) ->
-    {Trace1, [_ | _]} = ?split_trace_at(#{?snk_kind := Marker2}, Trace0),
-    {[_ | _], [_ | Trace2]} = ?split_trace_at(#{?snk_kind := Marker1}, Trace1),
-    Trace2.
-
 of_kinds(Trace0, Kinds0) ->
     Kinds = sets:from_list(Kinds0, [{version, 2}]),
     lists:filter(
         fun(#{?snk_kind := K}) -> sets:is_element(K, Kinds) end,
         Trace0
     ).
+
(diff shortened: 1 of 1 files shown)


apps/emqx/test/emqx_crl_cache_SUITE.erl
```diff
@@ -699,157 +707,78 @@ t_cache_overflow(Config) ->
...
-                    },
-                    #{
-                        ?snk_kind := mqtt_client_connection,
-                        ?snk_span := {complete, ok},
-                        client_num := 2
-                    }
+                    #{?snk_span := start, client_num := 1},
+                    #{?snk_span := {complete, ok}, client_num := 1},
+                    #{?snk_span := start, client_num := 2},
+                    #{?snk_span := {complete, ok}, client_num := 2}
```
thalesmg: IIRC we had deleted those, as they don't add much to the test?  🤔 

https://github.com/emqx/emqx/pull/17706#discussion_r3467104179

thalesmg (approved):

PR #18253, opened 2026-08-05
feat(rule_engine): expose map_to_range and hash_to_range SQL functions

zmstone:
Release version:
6.2.3, 6.3.0

Introduced in:

## Summary

Exposes the existing variform bifs `map_to_range/3` and `hash_to_range/3` as Rule-Engine SQL functions.

Both functions already exist and are tested in `emqx_utils/src/emqx_variform_bif.erl`; this PR only adds thin pass-through wrappers (and exports) in `emqx_rule_funcs.erl` so they are callable from rule SQL — no new logic.

- `map_to_range(value, min, max)`: maps an integer (or a binary/atom, via `binary:decode_unsigned/1`) into the inclusive range `[min, max]`.
- `hash_to_range(value, min, max)`: sha256-hashes the value, then maps the hash into `[min, max]` — a stable, well-distributed bucket.

Use case: sharding a large device fleet across N rules/actions by deriving a shard index from a topic segment, e.g. `SELECT hash_to_range(nth(2, tokens(topic, '/')), 0, 3) as shard FROM "t/#"`, instead of enumerating hundreds of topic filters.

Tests cover the wrappers directly and end-to-end SQL evaluation via `emqx_rule_sqltester`.

Docs: the function reference lives in the emqx-docs repo — to be added there separately.

## PR Checklist

### DIFF
--- a/apps/emqx_rule_engine/test/emqx_rule_funcs_SUITE.erl
+++ b/apps/emqx_rule_engine/test/emqx_rule_funcs_SUITE.erl
@@ -1131,6 +1131,68 @@ prop_hash_fun() ->
         end
     ).
 
+-doc "Tests map_to_range/3 maps integer, binary and atom values into an inclusive range.".
+t_map_to_range(_) ->
+    ?assertEqual(3, apply_func(map_to_range, [7, 0, 3])),
+    ?assertEqual(0, apply_func(map_to_range, [4, 0, 3])),
+    ?assertEqual(7, apply_func(map_to_range, [7, 7, 7])),
+    %% binary values are converted with binary:decode_unsigned/1
+    ?assertEqual(1, apply_func(map_to_range, [<<1>>, 0, 3])),
+    %% atom values are converted via their utf8 binary representation ('a' -> 97)
+    ?assertEqual(1, apply_func(map_to_range, [a, 0, 3])),
+    ?assertThrow(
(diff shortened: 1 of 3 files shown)


apps/emqx_rule_engine/test/emqx_rule_funcs_SUITE.erl
```diff
@@ -1131,6 +1131,68 @@ prop_hash_fun() ->
...
+     || N <- lists:seq(1, 100)
+    ]),
+    ?assertEqual([0, 1, 2, 3], Buckets),
+    ?assertThrow(
+        #{reason := badarg, function := hash_to_range},
+        apply_func(hash_to_range, [<<"x">>, 3, 0])
+    ),
+    ?assertThrow(
+        #{reason := badarg, function := hash_to_range},
+        apply_func(hash_to_range, [<<>>, 0, 3])
```
thalesmg, line 1170: q: why is an empty binary invalid here?

thalesmg (approved):

apps/emqx_rule_engine/test/emqx_rule_funcs_SUITE.erl
zmstone: inherit from variform which treats empty string equivalent to null/undefined.

PR #18263, opened 2026-08-05
feat(listeners): add proxy_address_allow to restrict trusted sources for forwarded client-address header

zmstone:
Release version:
6.3.0

Introduced in:

## Summary

Rework how MQTT `ws`/`wss` listeners derive the client address and port from forwarded headers, with an intentionally breaking default.

**Breaking default change**: `proxy_address_header` and `proxy_port_header` now default to an empty string, meaning forwarded headers are not consulted and the socket source address/port are always used. Deployments behind load balancers or reverse proxies that rely on `x-forwarded-for` must configure the header names explicitly. An empty header name disables the lookup (test and docs for this behavior imported from #18265).

**New `proxy_address_allow` option**: a list of IPv4/IPv6 CIDR ranges specifying which connection sources are trusted to carry the forwarded headers once enabled. Defaults to `["0.0.0.0/0", "::/0"]` (all sources trusted) — since header consumption is now opt-in, the allow list only needs tightening when the listener port is reachable by clients directly as well as via the proxy.

Key points:

* Enforcement lives in `emqx_ws_connection:get_peer/2`, which is only reached when PROXY protocol is off; listeners using `proxy_protocol` are unaffected since esockd derives the client address from the PROXY protocol handshake there.
* **Multi-entry headers are scanned right to left**: trusted entries (intermediate proxies) are skipped, and the first entry outside the trusted ranges is used as the client address. If all entries are trusted (as under the default), the leftmost entry is used, matching the previous extraction behavior. An entry that does not parse as an IP address falls back to the socket peer address.
* CIDR parsing/matching reuses `esockd_cidr` (same machinery as listener `access_rules` and `allow_log_packet_data_from`), so both IPv4 and IPv6 ranges are supported. CIDR lists match by address family (e.g. `0.0.0.0/0` covers IPv4 sources only).

Follow-up (not in this PR):

* The gateway WebSocket connection (`emqx_gateway_conn_ws`) and OCPP read the same headers via the gateway `websocket` options — a separate schema surface to get the same treatment.

## PR Checklist


apps/emqx/src/emqx_schema.erl
```diff
@@ -1115,6 +1115,15 @@ fields("ws_opts") ->
                     desc => ?DESC(fields_ws_opts_proxy_port_header)
                 }
             )},
+        {"proxy_address_allow",
+            sc(
+                hoconsc:array(typerefl:alias("string", any())),
```
zmstone: union of array and `per_security_profile`

apps/emqx/src/emqx_schema.erl
```diff
@@ -1115,6 +1115,15 @@ fields("ws_opts") ->
...
+                #{
+                    default => ws_proxy_address_allow_default(),
```
zmstone: security profile is runtime, default value must be compile time static.
```suggestion
                    default => per_security_profile,
```

apps/emqx/src/emqx_security_profile.erl
```diff
@@ -109,6 +110,11 @@ policy(internal_subscription_checks) ->
     case profile() of
         legacy -> false;
         hardened -> true
+    end;
+policy(ws_proxy_address_allow) ->
+    case profile() of
+        legacy -> any_ipv4;
```
zmstone:
```suggestion
        legacy -> any;
```

zmstone:
Superseded: 6.3 is feature-frozen, so this is split into #18271 (dev-63: change proxy_address_header/proxy_port_header defaults to empty) and #18272 (dev-64: add proxy_address_allow with right-to-left multi-hop extraction).

Issue #4502, opened 2021-04-08
Get client list by multiple clientIds/userNames API


UglyTroLL:
**What would you like to be added/modified**:
A new HTTP API can take a list of `clientIds` / `userNames` and returns all matching clients

**Why is this needed**:
I believe batch query is needed in all circumstances

tigercl:
@UglyTroLL This is a good suggestion, but it may not be consistent with our current API design. In 5.0, we will make an adjustment to the API, and then we will seriously consider your needs.

zmstone:
Internal tracking reference: [EMQX-11120](https://emqx.atlassian.net/browse/EMQX-11120)

zmstone:
e.g.
```
GET /clients?clientid=foo
GET /clients?clientid=foo&clientid=bar&clienid=gee&clientid=dee
```

zmstone:
supported since 5.6. see: https://github.com/emqx/emqx/pull/12719
here i quote the changelog:


## Support multiple clientid and username Query string parameters in "/clients" API

Multi clientid/username queries examples:
 - "/clients?clientid=client1&clientid=client2
 - "/clients?username=user11&username=user2"
 - "/clients?clientid=client1&clientid=client2&username=user1&username=user2"

## Add an option to specify which client info fields must be included in the response

Request response fields examples:
 - "/clients?fields=all" (omitting "fields" Qs parameter defaults to returning all fields)
 - "/clients?fields=clientid,username"

zmstone:
https://github.com/emqx/emqx/pull/12719

PR #18267, opened 2026-08-05
perf(mgmt): direct lookup for exact client-id queries in /clients API

zmstone:
Release versions:
5.8.12, 5.9.4, 5.10.5

Introduced in:
pre-5.8

## Summary

`GET /clients?clientid=a&clientid=b` (one or more exact client IDs and no other filter) used to run the generic paginated match-spec scan: the client IDs end up in match-spec guards, so ETS walks the whole `emqx_channel_info` table on every node (measured: ~224 ms per node for a 1M-row table vs ~4 µs for a direct lookup).

Such queries are now short-circuited to direct per-ID lookups via the channel registry, with the durable-session fallback so disconnected persistent sessions still resolve. Any other filter (`username`, `conn_state`, `like_*`, `gte_*`/`lte_*`, `node`, ...) keeps the existing scan path unchanged; `page`/`limit`/`fields` work the same on both paths. The response shape (`data` + `meta`) is unchanged.

Behavior notes:
- Results are returned in query-parameter order (previously: table traversal order). Duplicate IDs are deduplicated as before.
- `meta.count` for such queries now equals the number of matched clients; previously, with durable sessions enabled, it could overshoot by including the total disconnected-session count.

Follow-up to #4502 (EMQX-11120): the planned `/clients_bulk` endpoint was dropped in favor of optimizing this existing path. A sibling PR applies the same optimization on `dev-60`. `/clients_v2` is out of scope here.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -931,19 +931,11 @@ inflight_msgs(get, #{
 
 list_clients(QString) ->
     Result =
-        case maps:get(<<"node">>, QString, undefined) of
-            undefined ->
-                Options = #{fast_total_counting => true},
-                list_clients_cluster_query(QString, Options);
-            Node0 ->
-                case emqx_utils:safe_to_existing_atom(Node0) of
-                    {ok, Node1} ->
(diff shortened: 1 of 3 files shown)


zmstone:
will not merge this as the query loop is overly complex

PR #18269, opened 2026-08-05
perf(mgmt): direct lookup for exact client-id queries in /clients API

### LINKED ISSUE #4502: Get client list by multiple clientIds/userNames API

zmstone:
Release version:
6.4.0

## Summary

`GET /clients` queries that filter on exact client IDs only (one or more `clientid` query-string parameters, optionally combined with `node`) previously went through the generic paginated match-spec scan. Client IDs end up in the match-spec **guards**, never in the match head, so ETS cannot derive a key range and every such query walks the whole `emqx_channel_info` table on every node (measured ~224 ms for 2 IDs on a 1M-row table, vs microseconds for direct lookups).

`list_clients/1` in `emqx_mgmt_api_clients` now detects this query shape up front and resolves each ID with a direct per-ID lookup, returning a single directly-computed page:

- without `node`: `emqx_mgmt:lookup_client/2` — registry-aware lookup with fallback to durable session storage, so disconnected persistent sessions are found too;
- with `node`: `emqx_mgmt:lookup_client/3` (node-scoped RPC), with the same invalid/down-node error responses as before.

Requested IDs are deduplicated and the bounded result set is sliced by `page`/`limit` into the same `data` + `meta` (`count`/`hasnext`) envelope; the `fields` projection is applied via the same `format_channel_info/3`. Any other filter (`username`, `conn_state`, `like_*`, `gte_*`/`lte_*`, ...) takes the generic scan path, which is untouched. `/clients_v2` is not changed.

Behavior notes:

- This PR also bumps minirest to 1.5.1, which fixes repeated query-string parameters arriving in reverse order. With that, result rows follow the order of the requested client IDs (first occurrence wins for duplicates) rather than ETS table order.
- With durable sessions enabled, an exact-ID query now returns only the requested clients, and `meta.count` is exact. Previously the v1 durable-session sweep did not apply query filters, so unrelated disconnected durable sessions could appear in the result, and `count` was inflated by the total disconnected-session count.

Fixes #4502 (EMQX-11120).

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -872,19 +872,11 @@ inflight_msgs(get, #{
 
 list_clients(QString) ->
     Result =
-        case maps:get(<<"node">>, QString, undefined) of
-            undefined ->
-                Options = #{fast_total_counting => true},
-                list_clients_cluster_query(QString, Options);
-            Node0 ->
-                case emqx_utils:safe_to_existing_atom(Node0) of
-                    {ok, Node1} ->
-                        QStringWithoutNode = maps:remove(<<"node">>, QString),
-                        Options = #{},
-                        list_clients_node_query(Node1, QStringWithoutNode, Options);
-                    {error, _} ->
-                        {error, Node0, {badrpc, <<"invalid node">>}}
-                end
+        case exact_clientid_query(QString) of
+            {ok, ClientIds} ->
+                list_clients_by_id(ClientIds, QString);
+            false ->
+                list_clients_query(QString)
         end,
     case Result of
         {error, page_limit_invalid} ->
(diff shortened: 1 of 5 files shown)


apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -904,6 +896,92 @@ list_clients(QString) ->
...
+Returns `{ok, ClientIds}` when the query filters on exact client IDs only
+(`node` is allowed), i.e. when it can be served by direct per-ID lookups
+instead of the generic table scan.
+""".
+exact_clientid_query(#{<<"clientid">> := ClientIds = [_ | _]} = QString) ->
+    HasOtherFilters = lists:any(
+        fun({Key, _Type}) ->
+            Key =/= <<"clientid">> andalso Key =/= <<"node">> andalso maps:is_key(Key, QString)
+        end,
+        ?CLIENT_QSCHEMA
```
zmstone: split this macro to `CLIENT_QSCHEMA_DIRECT_LOOKUP` and `CLIENT_QSCHEMA_NEED_SCAN`

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -904,6 +903,90 @@ list_clients(QString) ->
...
+            catch
+                throw:{badrpc, Reason} ->
+                    {error, Node, {badrpc, Reason}}
+            end
+    end.
+
+lookup_clientid_rows(undefined, ClientId) ->
+    emqx_mgmt:lookup_client({clientid, ClientId}, undefined);
+lookup_clientid_rows(Node, ClientId) ->
+    case emqx_mgmt:lookup_client(Node, {clientid, ClientId}, undefined) of
```
thalesmg, line 985: note that this does not return offline ds sessions (might want to assert that in the test case to document, as it might be confusing)

thalesmg (dismissed):

thalesmg (approved):

PR #18272, opened 2026-08-06
feat(listeners): add proxy_address_allow to restrict trusted sources for forwarded client-address header

zmstone:
Release version:
6.4.0

Introduced in:

## Summary

Add a new `proxy_address_allow` option to MQTT `ws`/`wss` listeners: a list of IPv4 or IPv6 CIDR ranges specifying which connection sources (typically load balancers or reverse proxies in front of EMQX) are trusted to carry the original client address and port in the `proxy_address_header` / `proxy_port_header` headers.

The address taken from these headers becomes the connection's source address used by IP-based authorization rules, banned-client matching, and logging. The headers are honored only when the socket peer address is within one of the trusted ranges; otherwise the socket peer address is used. This gives operators explicit, defense-in-depth control over where forwarded client-address information is accepted from.

Key points:

* Enforcement lives in `emqx_ws_connection:get_peer/2`, which is only reached when PROXY protocol is off; listeners using `proxy_protocol` are unaffected since esockd derives the client address from the PROXY protocol handshake there.
* **Multi-entry headers are scanned right to left**: trusted entries (intermediate proxies) are skipped, and the first entry outside the trusted ranges is used as the client address. If all entries are trusted — as under the default `["0.0.0.0/0", "::/0"]` (all sources) — the leftmost entry is used, matching the previous extraction behavior. An entry that does not parse as an IP address falls back to the socket peer address.
* CIDR parsing/matching reuses `esockd_cidr` (same machinery as listener `access_rules` and `allow_log_packet_data_from`), so both IPv4 and IPv6 ranges are supported. CIDR lists match by address family (e.g. `0.0.0.0/0` covers IPv4 sources only).
* This PR does not change the `proxy_address_header` / `proxy_port_header` defaults — the empty-by-default change ships in 6.3 via #18271 and reaches this branch through the regular dev-63 → dev-64 sync.

Replaces #18263 (closed: 6.3 is feature-frozen, so the new option moves to dev-64 while the default change lands via #18271).

Follow-up (not in this PR):

* The gateway WebSocket connection (`emqx_gateway_conn_ws`) and OCPP read the same headers via the gateway `websocket` options — a separate schema surface to get the same option.

## PR Checklist


apps/emqx/test/emqx_ws_connection_SUITE.erl
```diff
@@ -115,6 +154,203 @@ t_header(_) ->
...
+        }
+    ),
+    ?assertMatch(
+        #{
+            socktype := ws,
+            peername := {{127, 0, 0, 1}, 3456}
+        },
+        ConnInfo
+    ),
+    set_ws_opts(proxy_address_allow, default_proxy_address_allow()).
```
thalesmg, line 335: nit: maybe try an ipv6 source to match the description of "every source"?

thalesmg (dismissed):

apps/emqx/test/emqx_ws_connection_SUITE.erl
zmstone: Good point — extended the case to assert both an IPv4 and an IPv6 socket peer are unaffected: 63f356d737

thalesmg (approved):

PR #18273, opened 2026-08-06
feat(plugins): add emqx_maptabs mapping-table plugin for rule SQL

zmstone:
Release versions:
5.10.5

## Summary

Adds a new in-tree plugin `emqx_maptabs`: named mapping tables for rule SQL, seeded from JSON files and cached in ETS for hot-path lookups.

- New rule SQL functions `maptab_lookup(Table, Key)`, `maptab_lookup(Table, Key, Field)`, `maptab_lookup(Table, Key, Field, Default)`, registered via `emqx_rule_engine:register_external_functions/1` (first in-tree plugin to use this hook). Lookups never crash a rule: unknown table / missing key / type-mismatched key is a normal miss (`undefined`). Key matching is exact term equality with no coercion. The looked-up fields can drive the builtin `subbits/6` directly to decode binary payloads.
- Table files (one JSON array per table, file name = table name) are the source of truth, kept under `<data_dir>/plugins/emqx_maptabs/tables/`. Loading is fail-closed: invalid JSON, missing `key`, duplicate keys, or a float key rejects the whole file and keeps the previous version.
- Reads go straight to ETS (`read_concurrency`, per-table generation swap on reload, so readers never observe a partial table).
- Config: `max_tables` (default 100) and `max_rows_per_table` (default 10000), enforced fail-closed at load/replay time; replacing an existing table is always allowed, and a limit change never drops already-loaded tables.
- Management is CLI-only (5.10 has no plugin HTTP API / UI support): `emqx ctl maptabs list|status|load|reload|get|delete`. `load`/`delete` replicate through an `emqx_cluster_rpc` transaction (per-node: re-validate, atomic temp-file+rename write, cache reload), so a node that was down during an update replays it on rejoin. Before initiating the transaction, all running nodes are preflight-checked (plugin running, tables dir writable) so a broken node fails the request fast instead of lagging the cluster-rpc log. `maptabs reload` is the per-node reconcile fallback (re-reads local files on all nodes).

Notes:

- The plugin must be installed on all nodes of a cluster (usual plugin deployment); the preflight check enforces the plugin being operational before any replicated update.
- Per-table lookup hit/miss counters (stale-catalog observability) were considered and deliberately left out of this PR; they can be added later if drift detection is needed.
- Cluster replay is covered by a CT case that stops a node, updates a table, restarts the node from spec and asserts it catches up via the cluster-rpc replay.

## PR Checklist


plugins/emqx_maptabs/src/emqx_maptabs_loader.erl
```diff
@@ -0,0 +1,109 @@
...
+%%--------------------------------------------------------------------
+
+%% Parses and validates mapping-table JSON files.
+%%
+%% A table file is a JSON array of row objects. Each row must have a
+%% "key" field; the remaining fields are the row's value map. Validation
+%% is fail-closed: any invalid row rejects the whole file.
+-module(emqx_maptabs_loader).
+
+-feature(maybe_expr, enable).
```
zmstone, line 12: delete.

plugins/emqx_maptabs/src/emqx_maptabs.erl
```diff
@@ -0,0 +1,498 @@
...
+lookup(_Table, _Key) ->
+    undefined.
+
+-spec lookup(binary(), term(), binary()) -> term() | undefined.
+lookup(Table, Key, Field) ->
+    case lookup(Table, Key) of
+        undefined -> undefined;
+        Values when is_binary(Field) -> maps:get(Field, Values, undefined);
+        _Values -> undefined
+    end.
```
zmstone, line 99: Field is `binary()`, ensured by function spec.
```suggestion
    case lookup(Table, Key) of
        Values when is_map(Values) -> maps:get(Field, Values, undefined);
        _ -> undefined
    end.
```

plugins/emqx_maptabs/src/emqx_maptabs_cli.erl
```diff
@@ -0,0 +1,112 @@
...
+list_on_node(Node) when Node =:= node() ->
+    emqx_maptabs:list_local();
+list_on_node(Node) ->
+    try
+        erpc:call(Node, emqx_maptabs, list_local, [], 15000)
+    catch
+        _:_ -> []
+    end.
+
+print_tables(Node, Tables) ->
```
zmstone: CLI should always only print JSON

plugins/emqx_maptabs/src/emqx_maptabs_loader.erl
```diff
@@ -0,0 +1,109 @@
...
+        ok ?= check_duplicate(Key, Seen, N),
+        Values = maps:remove(<<"key">>, Row),
+        validate_rows(Rest, N + 1, Seen#{Key => true}, [{Key, Values} | Acc])
+    end;
+validate_rows([Row | _], N, _Seen, _Acc) ->
+    {error, #{reason => row_not_an_object, row_number => N, row => Row}}.
+
+row_key(Row, N) ->
+    case maps:find(<<"key">>, Row) of
+        {ok, Key} when is_integer(Key); is_binary(Key); is_boolean(Key) ->
```
zmstone: reject boolean key.

plugins/emqx_maptabs/src/emqx_maptabs.erl
```diff
@@ -0,0 +1,592 @@
...
+%% old or the new version, never a partial one.
+%%
+%% Updates are replicated with `emqx_cluster_rpc' (see `load_file/1' and
+%% `delete/1'): each node re-validates the content, writes the file
+%% atomically (temp file + rename) and reloads its cache. Nodes that were
+%% down during an update replay the transaction on rejoin. `reload/1' is
+%% the per-node reconcile fallback that re-reads the local files.
+-module(emqx_maptabs).
+
+-feature(maybe_expr, enable).
```
zmstone, line 19: delete.

zmstone: ah, we are on otp 27.

plugins/emqx_maptabs/src/emqx_maptabs.erl
zmstone: Done in f8242a1181.

plugins/emqx_maptabs/src/emqx_maptabs_cli.erl
zmstone: Done in f8242a1181.

plugins/emqx_maptabs/src/emqx_maptabs_loader.erl
zmstone: Done in f8242a1181.

plugins/emqx_maptabs/src/emqx_maptabs_cli.erl
```diff
@@ -0,0 +1,110 @@
...
+
+%%--------------------------------------------------------------------
+%% Internal functions
+%%--------------------------------------------------------------------
+
+list_on_node(Node) when Node =:= node() ->
+    emqx_maptabs:list_local();
+list_on_node(Node) ->
+    try
+        erpc:call(Node, emqx_maptabs, list_local, [], 15000)
```
zmstone: avoid using erpc calls, call remote gen_server emqx_maptabs

plugins/emqx_maptabs/src/emqx_maptabs.erl
```diff
@@ -0,0 +1,580 @@
...
+%% disk on the next plugin start.
+-spec do_load_v1(binary(), binary(), emqx_config:cluster_rpc_opts()) -> ok | {error, term()}.
+do_load_v1(Name, Bin, _ClusterRpcOpts) ->
+    maybe
+        ok ?= emqx_maptabs_loader:validate_name(Name),
+        {ok, Parsed} ?= emqx_maptabs_loader:parse(Bin),
+        ok ?= check_row_limit(Parsed),
+        ok ?= check_table_limit_on_disk(Name),
+        ok ?= write_table_file(Name, Bin),
+        call(#commit{name = Name, parsed = Parsed})
```
zmstone: only check limits when kind is ?KIND_INITIATE.
otherwise, if config (limit) changed during replication, the ?KIND_REPLICATE nodes will fail.

plugins/emqx_maptabs/priv/config.hocon
```diff
@@ -0,0 +1,2 @@
+max_tables = 100
+max_rows_per_table = 10000
```
zmstone, line 2: should not accept file size which is too large, committing such large files to cluret-rpc table is too much.
add:
`max_table_file_bytes = 10000000`

plugins/emqx_maptabs/src/emqx_maptabs_cli.erl
zmstone: Done in 4beb5a6ff8.

plugins/emqx_maptabs/src/emqx_maptabs.erl
zmstone: Done in 4beb5a6ff8.

plugins/emqx_maptabs/priv/config.hocon
zmstone: Done in 4beb5a6ff8.

plugins/emqx_maptabs/src/emqx_maptabs.erl
```diff
@@ -0,0 +1,609 @@
...
+
+%% Loads a table file from a local path and replicates it to all nodes.
+%% The content is validated and all nodes are preflight-checked (plugin
+%% running, tables dir writable) before the cluster_rpc transaction, so
+%% a broken node fails the request fast instead of lagging forever.
+-spec load_file(file:filename_all()) -> ok | {error, term()}.
+load_file(Path) ->
+    maybe
+        {ok, Name} ?= emqx_maptabs_loader:table_name_from_path(Path),
+        {ok, Bin} ?= read_source_file(Path),
```
zmstone, line 135: risk at reading 4GB file, check file size before read.

zmstone:
will release in v6.1
close this pr for v5.10

PR #18274, opened 2026-08-06
fix(tablestore): use O(1) health check probe instead of listing all tables

### LINKED ISSUE #17410: fix(tablestore): health check lists all timeseries tables on every tick
## Summary

The Aliyun Tablestore (OTS) connector calls `ots_ts_client:list_tables/1` on every `on_get_status/2` tick — returning the names of **every** timeseries table in the OTS instance. It's also re-run on every `on_start/2`, so a stale or large response stalls startup.

## Where

`apps/emqx_bridge_tablestore/src/emqx_bridge_tablestore_connector.erl:98-102` (status) and `:86` (start).

```erlang
on_get_status(_InstId, #{client := Client} = _State) ->
    case ots_ts_client:list_tables(Client) of
        {ok, _Tables} -> connected;
        ...
```

The underlying driver call (`deps/ots_erl/src/ots_ts_client.erl:227-235`) wraps the OTS `ListTimeseriesTable` RPC, which returns the full table list with no pagination wired through.

## Impact

* IoT workloads on Tablestore commonly create one table per device or per sensor — thousands to millions of tables is realistic.
* Every health-check tick pulls the entire list (no pagination passed by EMQX).
* Each list returns table metadata that grows with the table count — data egress cost and connector parse cost both scale linearly.
* Startup re-runs the same call, so a multi-thousand-table instance can stall `on_start/2` and time out before the connector ever becomes ready.

## Suggested fix

Two options, in preference order:

1. Replace with a `describe_timeseries_table` call against a known-probe table that the operator configures (or a sentinel table the connector itself creates on startup). The OTS `DescribeTimeseriesTable` API takes a single table name and returns O(1) metadata.
2. Fall back to a TCP probe of the OTS endpoint if no probe table is configured.

The current `list_tables` is the wrong shape for a health check on this service.

## Related

Part of the health-check audit following d94cbde4. Same class of bug — open-ended list on every tick.

JimMoen:
## Summary

- `on_start`/`on_get_status` used to call `ots_ts_client:list_tables/1`, pulling the names of **every** timeseries table in the OTS instance on every health check tick (default 15s). With IoT workloads (one table per device), this is unbounded cost per tick and can stall startup.
- Replace with a `DescribeTimeseriesTable` probe against an existing table (O(1) metadata), configured via new optional `probe_table_name` connector config.
- When `probe_table_name` is unset, fall back to a raw TCP probe of the configured endpoint (O(1), no auth check).
- Driver support: `ots_erl` `describe_table/2` was previously unusable (empty request body); fixed in emqx/ots_erl#8 (merged), pinned here via `branch: master`.
- Adds connector unit tests for both probe modes (25 assertions) + changelog.

Fixes #17410

## Test Plan

- [x] eunit: emqx_bridge_tablestore_connector_tests (25 assertions, all passed locally)
- [x] CI

### DIFF
--- a/apps/emqx_bridge_tablestore/src/emqx_bridge_tablestore_connector.erl
+++ b/apps/emqx_bridge_tablestore/src/emqx_bridge_tablestore_connector.erl
@@ -84,7 +84,7 @@ on_start(InstId, Config) ->
     ?LOG_T(info, #{msg => ots_start, ots_opts => BaseOpts}),
     {ok, ClientRef} = start_ots_ts_client(InstId, BaseOpts ++ SecretOpts),
     ProbeTableName = maps:get(probe_table_name, Config, undefined),
-    case probe_ots(ClientRef, ProbeTableName, BaseOpts) of
+    case probe_ots(ClientRef, ProbeTableName) of
         ok ->
             {ok, #{
                 client_ref => ClientRef,
@@ -101,11 +101,8 @@ on_stop(_InstId, #{client_ref := ClientRef} = State) ->
     ots_ts_client:stop(ClientRef),
     State.
 
-on_get_status(
-    _InstId,
-    #{client_ref := ClientRef, probe_table_name := ProbeTableName, ots_opts := OtsOpts}
-) ->
-    case probe_ots(ClientRef, ProbeTableName, OtsOpts) of
+on_get_status(_InstId, #{client_ref := ClientRef, probe_table_name := ProbeTableName}) ->
+    case probe_ots(ClientRef, ProbeTableName) of
         ok -> ?status_connected;
         _ -> ?status_connecting
     end.
@@ -166,10 +163,21 @@ send_batch_data(BatchDataList, ClientRef, ChannelId) ->
 start_ots_ts_client(_, OtsOpts) ->
     ots_ts_client:start(OtsOpts).
 
-probe_ots(_ClientRef, undefined, OtsOpts) ->
-    tcp_probe_ots(proplists:get_value(endpoint, OtsOpts));
-probe_ots(ClientRef, ProbeTableName, _OtsOpts) ->
+probe_ots(ClientRef, undefined) ->
+    list_ots_tables(ClientRef);
+probe_ots(ClientRef, ProbeTableName) ->
     describe_ots_table(ClientRef, ProbeTableName).
+
+list_ots_tables(ClientRef) ->
+    try
+        case ots_ts_client:list_tables(ClientRef) of
+            {ok, _} -> ok;
+            {error, Reason} -> {error, Reason}
+        end
+    catch
+        Err:CReason:CST ->
+            {error, #{error => Err, reason => CReason, stacktrace => CST}}
+    end.
 
 describe_ots_table(ClientRef, ProbeTableName) ->
     try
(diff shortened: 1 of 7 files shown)


apps/emqx_bridge_tablestore/src/emqx_bridge_tablestore_connector.erl
```diff
@@ -157,12 +166,44 @@ send_batch_data(BatchDataList, ClientRef, ChannelId) ->
...
+        end
+    catch
+        Err:CReason:CST ->
+            {error, #{error => Err, reason => CReason, stacktrace => CST}}
+    end.
+
+tcp_probe_ots(Endpoint) ->
+    case parse_endpoint(Endpoint) of
+        {ok, Host, Port} ->
+            case gen_tcp:connect(Host, Port, [binary, {active, false}], 5000) of
```
zmstone: `Host` is a binary here, make sure `gen_tcp:connect/4` accepts binary.

apps/emqx_bridge_tablestore/test/emqx_bridge_tablestore_connector_tests.erl
```diff
@@ -102,11 +116,159 @@ start_connector_failure_test_() ->
...
+        fun() ->
+            meck:new(ots_ts_client, [no_history]),
+            ok = meck:expect(ots_ts_client, start, fun(_OtsOpts) ->
+                {ok, dummy_client_ref}
+            end),
+            ok = meck:expect(ots_ts_client, stop, fun(_CRef) ->
+                ok
+            end),
+            meck:new(gen_tcp, [no_history, unstick]),
+            ok = meck:expect(gen_tcp, connect, fun(_Host, _Port, _Opts, _Timeout) ->
```
zmstone: this mock shadowed the bug binary() for Host bug?

apps/emqx_bridge_tablestore/src/emqx_bridge_tablestore_connector.erl
```diff
@@ -157,12 +166,44 @@ send_batch_data(BatchDataList, ClientRef, ChannelId) ->
 start_ots_ts_client(_, OtsOpts) ->
     ots_ts_client:start(OtsOpts).
 
-list_ots_tables(ClientRef) ->
+probe_ots(_ClientRef, undefined, OtsOpts) ->
+    tcp_probe_ots(proplists:get_value(endpoint, OtsOpts));
```
zmstone: fallback to old health check (tls, auth etc).
or check an unknown table, take 'unauthorized', 'unknown' errors as healthy.

zmstone (commented):
does probe failure result in logs with information to troublshoot ?

apps/emqx_bridge_tablestore/src/emqx_bridge_tablestore_connector.erl
JimMoen: Implemented: when `probe_table_name` is unset, the connector now falls back to the original `list_tables` health check, so TLS and credentials are still verified. The raw TCP probe fallback is dropped entirely.

apps/emqx_bridge_tablestore/src/emqx_bridge_tablestore_connector.erl
JimMoen: Moot now — the TCP probe is removed, so `gen_tcp:connect/4` is no longer called. (For the record, we confirmed on OTP 27 that it exits with badarg on a binary host, so the concern was valid.)

apps/emqx_bridge_tablestore/test/emqx_bridge_tablestore_connector_tests.erl
JimMoen: The gen_tcp mock is gone together with the TCP probe; the fallback tests now mock `list_tables` instead.

zmstone (approved):

PR #18275, opened 2026-08-06
feat(authz): include mountpoint by default in hardened mode

savonarola:
Release version: 6.3.0

Introduced in: N/A

## Summary

Make `authorization.include_mountpoint = true` in hardened mode.

## PR Checklist


changes/ee/breaking-18275.en.md
```diff
@@ -0,0 +1 @@
+Changed `authorization.include_mountpoint` to default to `true` under the hardened security profile.
```
zmstone: describe how users should update their authorization rules/records.

also describe how to make it compatible with old rules/recrods:

- change include_mountpoint config to false
- change to 'legacy' mode

apps/emqx/src/emqx_schema.erl
```diff
@@ -3063,7 +3063,7 @@ authz_fields() ->
             sc(
                 boolean(),
                 #{
-                    default => false,
+                    default => emqx_security_profile:policy(authz_default_include_mountpoint),
```
zmstone, line 3066: schema doc is static, default value should be static too.

```suggestion
                    default => true,
```

savonarola: Not quite getting, we already added quite a few hardening of _defaults_

savonarola: >schema doc is static

it is, given the default value for sec profile is used. 
once it is changed, the defaults change too.

changes/ee/breaking-18275.en.md
savonarola: ✅

zmstone (dismissed):

savonarola:
I think it's better to merge after https://github.com/emqx/emqx/pull/18246, because here we update tests to comply buggy behaviour fixed in that PR.

zmstone:
since we have reverted default profile to legacy, this one should not be a braking- changelog

changes/ee/breaking-18275.en.md
```diff
@@ -0,0 +1,10 @@
+Changed `authorization.include_mountpoint` to default to `true` under the hardened security profile. I.e., by default, the final topics including the mountpoint prefix are authorized.
```
zmstone, line 1: this is a breaking- for v7.
https://github.com/emqx/emqx/pull/18511

apps/emqx/src/emqx_security_profile.erl
```diff
@@ -164,6 +165,11 @@ policy(authn_builtin_accept_weak_password_hash) ->
         hardened -> false
     end;
 policy(authn_mnesia_mt_user_conflict_protection) ->
+    case profile() of
+        legacy -> false;
+        hardened -> true
+    end;
+policy(authz_default_include_mountpoint) ->
```
zmstone: Consideration on gating this behind the profile for **gateways** specifically. The gateway test changes show authz previously evaluated the *bare* gateway topic (e.g. `upstream/vlogin`, identical across every device) and now evaluates the mounted per-device topic (`gbt32960/<clientid>/upstream/vlogin`). For gateways the mountpoint *is* the per-device topic namespacing, so bare-topic authz cannot distinguish devices — a single `allow upstream/#` rule authorizes every device's upstream topic. Gating the fix behind `hardened` leaves legacy gateway deployments (the default until 7.0) with that coarse, device-blind authorization.

Is the profile-gating intended for the gateway case too, or should gateway authz include the mountpoint regardless of profile (bare-topic authz being arguably always wrong for a gateway whose mountpoint carries the device identity)? Related: https://github.com/emqx/emqx-dev-team-tasks/issues/329 (jt808 mountpoint not applied to routed topics). The core change and the changelog are good; this is a scope question, not a defect.

changes/ee/breaking-18275.en.md
zmstone: This adds a new hardened-profile behavior (`authz_default_include_mountpoint`). To keep the security-profile surface in sync, it should also be reflected in two places outside this PR: the security-profile documentation page in emqx-docs, and the default-flip changelog in https://github.com/emqx/emqx/pull/18511 (which enumerates the hardened behaviors that become active when the default switches to `hardened`). The changelog here is clear and the `org1/sensors/#` migration example is genuinely helpful.

thalesmg (approved):

PR #18287, opened 2026-08-07
fix: handle RPC failures consistently in BPAPI proto callers

zmstone:
Release version:

Introduced in:
pre-5.8 for most findings; `emqx_mgmt_data_backup_proto_v2:import_file/5` (erpc switch) and the bridge v2 per-node operation node check are 6.0.0 regressions.

## Summary

Follow-up to issue #17747 and PR #18107: a systematic audit of all 95 BPAPI proto modules on dev-60 for mismatches between an RPC transport's actual failure signaling and what the caller handles (`rpc:call` returns `{badrpc,_}`; `erpc:call` raises and never returns `{badrpc,_}`; `rpc:multicall` puts remote crashes *inside* the result list as `{badrpc,{'EXIT',_}}`; `erpc:multicall` wraps successes in `{ok,_}`).

Fixes by class:

* **erpc raises unhandled (class A)** — added/widened `try/catch` around erpc-backed proto calls, mapping raises onto each caller's existing error convention: data backup import, clients_v2 listing, file-transfer download + assembler, session-takeover conclusion, MQ consumer stop, Kinesis shared-limiter checks.
* **`{badrpc,_}` returns leaking as data (class C)** — added the missing match and mapped it to the surrounding convention (mirroring the nearest correct sibling handler): trace log streaming/download/list/detail, `GET /configs` (text/plain), delayed-message DELETE, topic-metrics reset, plugin list/describe, retainer reindex dispatch-wait.
* **erpc:multicall result shapes (class B)** — normalized `{throw,_}`/`{exit,_}` per-node failures to `{error,_}` in the bridge v2 and connector APIs.
* **Other** — `emqx_rule_engine_api:mk_format_fn/1` used `=` instead of `?=` in a `maybe` block, making the intended degrade-to-empty-summary branch unreachable (rule listing 500'd when any node was down); the bridge v2 per-node operation lost its cluster-membership check in the proto v8 migration (unknown node gave 400 instead of 404); the data-backup error responses invoked `?SERVICE_UNAVAILABLE(Reason)` (the 1-arity response macro) as the `code` field value, which is not JSON-serializable, and the data-backup schema did not declare the 500 responses at all, so minirest replaced the structured body with an "Unsupported code" placeholder.

No proto module is modified (frozen after GA); all fixes are caller-side.

Full audit report (bug-class table, per-module coverage, report-only findings left as follow-ups: DS monitor leak on subscribe error path, conf CLI badrpc crashes, plugins sync silent 204, action metrics rendering failed nodes as zeros, `emqx_mt` cluster-rpc badmatch, frozen-proto spec rot) is kept outside the repo; ask if you want a copy. A static check for "caller of an erpc-backed proto matching `{badrpc,_}`" (dev-58 report recommendation) remains a follow-up and is not implemented here.

Tests: meck-based CT cases assert structured error responses (instead of 500 crashes) for the delayed DELETE, configs text/plain, trace stream + download, topic-metrics reset, plugin listing, data-backup import, and retainer reindex paths. The ft download 503 path was already covered by `t_download_transfer`; the remaining fixes (cm takeover, MQ stop, Kinesis, clients_v2, rule listing, bridge/connector normalization) are one-clause changes on paths whose CT would require multi-node failure injection disproportionate to the change.

## PR Checklist


apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -530,8 +530,10 @@ validate_file_name(_Params, _Meta) ->
 %% API CallBack Begin
 list_plugins(get, _) ->
     Nodes = emqx:running_nodes(),
-    {Plugins, []} = emqx_mgmt_api_plugins_proto_v4:get_plugins(Nodes),
-    {200, format_plugins(Plugins)}.
+    {Plugins, BadNodes} = emqx_mgmt_api_plugins_proto_v4:get_plugins(Nodes),
+    BadNodes =/= [] andalso
+        ?SLOG(error, #{msg => "get_plugins_rpc_failed", bad_nodes => BadNodes}),
```
zmstone: warning

apps/emqx_bridge/src/emqx_bridge_v2_api.erl
```diff
@@ -1092,13 +1092,17 @@ handle_operation(Namespace, ConfRootKey, Id, Op) ->
...
+            {ok, TargetNode} ?= emqx_utils:safe_to_existing_atom(Node, utf8),
+            true ?= lists:member(TargetNode, emqx:running_nodes()),
+            {ProtoMod, OperFunc} = operation_func(Op),
+            BPAPIArgs = [[TargetNode], Namespace, ConfRootKey, BridgeType, BridgeName],
+            call_operation_if_enabled(
+                ProtoMod, OperFunc, Namespace, ConfRootKey, BridgeType, BridgeName, BPAPIArgs
+            )
+        else
+            false ->
+                ?NOT_FOUND(<<"Node not found: ", Node/binary>>);
```
thalesmg: nit: maybe use the same error as below to avoid an atom oracle?

apps/emqx_ft/src/emqx_ft_storage_exporter_fs_api.erl
```diff
@@ -123,7 +123,10 @@ fields(file_node) ->
                     'NOT_FOUND',
                     iolist_to_binary(["Invalid query parameter: ", Param])
                 )};
-        error:{erpc, noconnection} ->
+        error:{erpc, _} ->
+            {503, error_msg('SERVICE_UNAVAILABLE', <<"Service unavailable">>)};
+        error:{exception, Reason, _Stack} ->
+            ?SLOG(warning, #{msg => "get_ready_transfer_fail", error => Reason}),
```
thalesmg: maybe the log could contain the stack trace, since it's unexpected?

apps/emqx_management/src/emqx_mgmt_api_configs.erl
```diff
@@ -391,18 +391,32 @@ handle_configs_v1_result(_Node, Res) ->
 
 get_configs_v2(QueryStr) ->
     Node = maps:get(<<"node">>, QueryStr, node()),
```
thalesmg: not related here, but maybe we should change this param to a binary() 🙈

apps/emqx_management/src/emqx_mgmt_api_configs.erl
```diff
@@ -391,18 +391,32 @@ handle_configs_v1_result(_Node, Res) ->
...
                 emqx_conf_proto_v5:get_hocon_config(Node, ?global_ns, atom_to_binary(Key))
         end,
-    {
-        200,
-        #{<<"content-type">> => <<"text/plain">>},
-        iolist_to_binary(hocon_pp:do(Conf, #{}))
-    }.
+    case Res of
+        {badrpc, _} = Err ->
+            handle_configs_v1_result(Node, Err);
```
thalesmg: nit: this is inside a v2 fn, after calling a v5 proto

apps/emqx_management/src/emqx_mgmt_api_trace.erl
```diff
@@ -892,9 +904,20 @@ cluster_trace_details(Name) ->
...
+            ({badrpc, _}) -> false;
+            (_) -> true
+        end,
+        GoodRes0
+    ),
+    (BadNodes =/= [] orelse BadRes =/= []) andalso
+        ?SLOG(warning, #{
+            msg => "rpc_failed_to_read_trace",
+            bad_nodes => BadNodes,
+            errors => BadRes
```
thalesmg: nit: maybe strip the badrpc to ease reading/reduce chance of truncating output

apps/emqx_management/test/emqx_mgmt_api_plugins_SUITE.erl
```diff
@@ -143,6 +143,25 @@ t_plugins(_Config) ->
     ?assertMatch({ok, {{_, 403, _}, _, _}}, install_plugin(PackagePath)),
     ok.
 
+-doc """
+`GET /plugins` tolerates unreachable nodes and remote crashes in the
+cluster-wide plugin listing RPC instead of crashing with badmatch.
+""".
+t_list_plugins_rpc_failure(_Config) ->
+    ok = meck:new(emqx_mgmt_api_plugins_proto_v4, [passthrough, no_link]),
```
thalesmg: nit


```suggestion
    ok = meck:new(emqx_mgmt_api_plugins_proto_v4, [passthrough]),
```

apps/emqx_modules/src/emqx_delayed_api.erl
thalesmg: maybe change this?  🙈

thalesmg (approved):

PR #18292, opened 2026-08-07
fix(prometheus): omit namespaced metrics for unknown namespaces

### LINKED ISSUE #18288: [6.2.3]prometheus namespaced_stats emits session metric for a namespace when emqx_mt_ns is absent
## Summary

PR #18183 guards the Prometheus namespaced collector when emqx_mt and emqx_auth_mnesia are not running. The global collection path is safe, but the specific-namespace path still emits an emqx_sessions_count metric with value 0 when emqx_mt_ns is absent.

## Environment

- EMQX: 6.2.2-gdcba995d
- Endpoint: /api/v5/prometheus/namespaced_stats
- PR context: #18183, release targets 6.1.5, 6.2.3, 6.3.0

## Reproduction

1. Stop emqx_mt and remove the leftover emqx_mt_ns ETS table.
2. Request `/api/v5/prometheus/namespaced_stats?mode=node&format=prometheus&ns=pr18183-unknown` with an administrator bearer token.
3. Observe HTTP 200 and a line similar to:

```
emqx_sessions_count{namespace="pr18183-unknown"} 0
```

The same behavior is reproducible for the other supported modes.

## Expected behavior

When emqx_mt_ns does not exist, the requested namespace is not known and the namespaced session metric should be omitted, consistent with the global collection path and the PR summary: no namespaces exist when the owning feature is not running.

## Actual behavior

The PR guard makes `is_known_ns/1` return false and `count_clients/1` return `{error, not_found}`, but `gather_session_ns_data/1` handles that result by constructing `#{Namespace => [{emqx_sessions_count, 0}]}`. This creates a metric for a namespace that does not exist.

Relevant path:

```
gather_session_ns_data(Namespace)
  -> emqx_mt_state:count_clients(Namespace)
  -> {error, not_found}
  -> emqx_sessions_count = 0
```

The existing tests cover the direct `is_known_ns/1` false result and the global no-namespace scrape, but do not combine an absent table with a specific `ns` query.

zmstone:
Release version:
6.1.5, 6.2.3, 6.3.0, 6.4.0

Introduced in:
6.1.0

## Summary

Fixes #18288. Follow-up to #18183.

The specific-namespace path of `/prometheus/namespaced_stats` (`?ns=<ns>`) fabricated zero-valued samples for namespaces that do not exist, while the all-namespaces path correctly omits them. A fabricated `0` is worse than an omitted sample: for a real namespace hitting a transient read error it looks like all sessions vanished (false alerts), and for an unknown namespace it invents a series that never existed. Omission means "no data", which is what Prometheus consumers expect.

Changes in `emqx_prometheus`:

* `gather_session_ns_data/1` (the reported bug): when `emqx_mt_state:count_clients/1` returns an error for the requested namespace, omit the namespace instead of emitting `emqx_sessions_count 0`.
* The `all` clause's `{error, _}` branch (namespace deleted concurrently during the fold) now also skips the namespace instead of reporting `0`, for consistency.
* `gather_authz_ns_data/1` / `gather_authn_ns_data/1` had the same defect: `record_count/1` returns `0` both for an unknown namespace and when the `emqx_auth_mnesia` tables are absent, so the specific-ns path fabricated `..._record_count 0`. They are now gated on `emqx_mt_state:is_known_ns/1`; known namespaces keep reporting their (possibly zero) record count.
* `metric_data_ns/3` specific-ns clause also fabricated `0` for every packet/message/delivery metric (`emqx_metrics:all/1` returns `[]` for a namespace with no counters, and the sample builder defaults missing keys to `0`). It now omits the namespace when no metrics counters exist for it, mirroring the `all` clause which only enumerates namespaces present in `emqx_metrics:all_ns/0`.

The all-namespaces collection path is untouched. Unknown namespaces still respond HTTP 200 with the metrics omitted (matching the all-namespaces path); returning 404 would be a behavior change beyond the issue's scope.

## PR Checklist

### DIFF
--- a/apps/emqx_prometheus/src/emqx_prometheus.erl
+++ b/apps/emqx_prometheus/src/emqx_prometheus.erl
@@ -836,7 +836,15 @@ metric_data_ns(all, MetricSpecs, Mode) ->
     NsToMetrics = maps:remove(?global_ns, NsToMetrics0),
     do_metric_data_ns(NsToMetrics, MetricSpecs, Mode);
 metric_data_ns(Namespace, MetricSpecs, Mode) when is_binary(Namespace) ->
-    NsToMetrics = #{Namespace => emqx_metrics:all(Namespace)},
+    NsToMetrics =
+        case emqx_metrics:all(Namespace) of
+            [] ->
+                %% No metrics counters for this namespace: omit it, as the `all' clause
+                %% does for namespaces absent from `emqx_metrics:all_ns/0'.
+                #{};
+            Metrics ->
+                #{Namespace => Metrics}
+        end,
     do_metric_data_ns(NsToMetrics, MetricSpecs, Mode).
 
 do_metric_data_ns(NsToMetrics, MetricSpecs, Mode) ->
@@ -872,8 +880,8 @@ gather_session_ns_data(all) ->
             {ok, N} ->
                 Acc#{Namespace => [{emqx_sessions_count, N}]};
             {error, _} ->
-                %% race?
-                Acc#{Namespace => [{emqx_sessions_count, 0}]}
+                %% Namespace deleted concurrently: omit it.
+                Acc
         end
     end,
     InitAcc = #{},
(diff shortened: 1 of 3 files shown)


thalesmg (approved):
I'm thinking if it would make sense to change the api handler to return a 404 if the ns doesn't exist.

PR #18294, opened 2026-08-07
fix(sessds): tolerate concurrent session deletion on clean-start

zmstone:
Release version:

Introduced in:
6.0.0

## Summary

Fixes a race where a clean-start reconnect could crash the new connection while the previous durable session was being cleaned up concurrently.

`emqx_cm:do_open_session(_CleanStart = true, ...)` discards the old channel and destroys the session under the CM locker lock, but the discarded channel terminates asynchronously outside that lock and (with session expiry interval 0) drops the same DS session. Both drops open the session state and read the same collection guard; the winner's transaction deletes the guard, so the loser's `emqx_ds_pmap:tx_assert_guard/2` precondition fails, and the `{atomic, _, _}` badmatch in `emqx_persistent_session_ds_state_v2:delete/2` crashed the connecting client's connection process (client saw `tcp_closed`). Surfaced in CI as `emqx_takeover_SUITE` `persistence_enabled.mqttv5.t_takeover_willmsg_clean_session` hanging until the 120 s timetrap.

The fix tolerates `{error, unrecoverable, {precondition_failed, _}}` inside `emqx_persistent_session_ds_state_v2:delete/2`: a guard mismatch there means the state was concurrently deleted, or now belongs to a newer incarnation of the client whose data must not be deleted — in both cases not deleting is correct for any caller, so the source of the transaction is the precise place to handle it (rather than catching at the `session_drop/2` call site). This mirrors the existing idiom in `commit/3`, which tolerates `precondition_failed` for `lifetime => terminate` with a warning trace point. The tolerated path emits a new `?sessds_drop_conflict` trace point; other transaction errors now raise a descriptive `failed_to_delete_session` error instead of a bare badmatch.

New test `t_state_drop_conflict` in `emqx_persistent_session_ds_SUITE` reproduces the loser's view deterministically (two actors open the same state, both delete) and asserts idempotency plus the trace point.

Possible follow-up (out of scope here): `emqx_takeover_SUITE`'s `start_client_async_subscribe` helper hangs until the timetrap when the client crashes mid-case; it could fail fast instead.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state_v2.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state_v2.erl
@@ -330,7 +330,7 @@ delete(
         retries => trans_retries(),
         retry_interval => trans_retry_interval()
     },
-    {atomic, _, _} =
+    Result =
         emqx_ds:trans(
             Opts,
             fun() ->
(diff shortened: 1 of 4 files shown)


ieQu1:
This PR currently violates the principle of `_ds_state` module:

> It is completely devoid of business logic. Not even the default
values should be set in this module.

Translation of certain types of errors to `ok` is very much "business logic". The result of the transaction should be simply passed to the session, which will handle or ignore it.

PR #18296, opened 2026-08-07
feat(variform): add is_jwt/1 for authenticator preconditions

zmstone:
Release version:
6.2.3, 6.3.0

Introduced in: N/A

## Summary

Adds a never-throwing boolean variform function `is_jwt/1` (in `emqx_variform_bif`, next to `jwt_value/2`) so operators can set `precondition = "is_jwt(password)"` on a JWT authenticator.

Motivation: under the `hardened` security profile, a credential that is not a JWT at all exits the JWT authenticator through the verification-failure path and is denied, which breaks chains serving both JWT clients and legacy password/mTLS clients. With the precondition, structurally-non-JWT credentials skip the JWT authenticator (`precondition_not_met` → `ignore`, independent of the security profile) and fall through to the next authenticator, while real JWTs that fail verification stay strictly denied.

`is_jwt/1` returns true only when the value is structurally a JWS compact serialization: exactly 3 dot-separated segments, header base64url-decodable to a JSON object with an `"alg"` field, payload base64url-decodable. The signature is not verified and the payload is not inspected — the function classifies, it does not validate. The 5-segment JWE form returns false (EMQX JWT authentication verifies JWS only). Base64 padding in segments is tolerated, matching `jwt_value`'s decode mode; for borderline non-RFC tokens this errs toward running the verifier (deny) rather than skipping.

Unlike other string bifs (including `jwt_value`, which throws `invalid_jwt_token` / `badarg` on null), `is_jwt/1` never raises: null, undefined, empty, or any malformed input returns false. This matters specifically for preconditions: a render **error** maps to the backend-failure policy (deny under `hardened`), while a falsy render maps to a clean skip. An absent MQTT password binds to `undefined` and renders false, so passwordless (e.g. mTLS) clients also fall through cleanly.

Integration test in `emqx_authn_chains_SUITE` proves the skip-vs-deny split under `hardened`: non-JWT password → skipped, next authenticator decides; structurally valid JWT with bad signature → denied; valid JWT → authenticated.

Docs (emqx-docs) update to the authentication precondition page is a follow-up.

## PR Checklist


apps/emqx_utils/src/emqx_variform_bif.erl
```diff
@@ -766,6 +767,45 @@ jwt_value(Token, KeyPath) when is_binary(Token), is_binary(KeyPath) ->
             })
     end.
 
+-doc """
+Check whether a value is structurally a JWT (JWS compact serialization).
+
+Returns true only when the value is a string of exactly three dot-separated
+segments whose first segment base64url-decodes to a JSON object containing
+an "alg" field, and whose second segment is also base64url-decodable.
```
thalesmg: q: if the payload is decoded (although not inspected further), why not validate that the signature is also decodable?

apps/emqx_utils/src/emqx_variform_bif.erl
```diff
@@ -766,6 +767,45 @@ jwt_value(Token, KeyPath) when is_binary(Token), is_binary(KeyPath) ->
...
+  is_jwt(<<"a-plain-password">>) -> false
+  is_jwt(undefined) -> false
+""".
+-spec is_jwt(term()) -> boolean().
+is_jwt(Token) when is_binary(Token) ->
+    try
+        [Header, Payload, _Sig] = binary:split(Token, <<".">>, [global]),
+        HeaderBin = base64:decode(Header, #{mode => urlsafe, padding => false}),
+        _ = base64:decode(Payload, #{mode => urlsafe, padding => false}),
+        case emqx_utils_json:decode(HeaderBin, [return_maps]) of
```
thalesmg:
```suggestion
        case emqx_utils_json:decode(HeaderBin) of
```

thalesmg (dismissed):

apps/emqx_utils/src/emqx_variform_bif.erl
zmstone: good point.

thalesmg (approved):

PR #18300, opened 2026-08-08
fix(bigtable): apply configured ssl options to outbound gRPC connections

zmstone:
Release version:
6.3.0

Introduced in:
N/A (Bigtable connector has only shipped in 6.3.0 betas)

## Summary

Split out of #18299 (now retargeted at `dev-60` with the Snowflake fix only), from the audit of connectors whose schema exposes an `ssl` block that the implementation ignores.

- **Bigtable**: the connector schema exposes the standard `ssl` block, but the gRPC channel was started without any transport options — gun decided the transport by port and used its own TLS defaults, so every configured `ssl` field was silently ignored. The config is now threaded into gun's `transport`/`tls_opts`, keyed off the (hidden) `url` scheme: `https` forces TLS on and applies `emqx_tls_lib:to_client_opts/1`; `http` (emulator/testing) connects over TCP. The schema default for `verify` becomes `verify_peer`, codifying what the connector effectively did before the fix (gun injects the OS CA bundle; OTP defaults to peer verification); the endpoint is fixed at `bigtable.googleapis.com` (public CAs), so verification works with zero configuration. **Schema change**: default only; all fields remain user-settable.
- **Shared connector cert sanitation** (`emqx_connector_ssl`): blank `certfile`/`keyfile`/`cacertfile` values are now dropped as "not configured" regardless of the `verify` mode. Previously blank client-cert fields were rejected with `pem_file_path_or_string_is_required` when `verify = verify_peer`, although client certificates are orthogonal to peer verification (blank `cacertfile` was already allowed via `ALLOW_EMPTY_PEM`). This also makes the new Bigtable `verify_peer` default safe for dashboard-submitted configs that contain blank cert fields.

Validated with the module eunit tests and the full `emqx_bridge_bigtable_SUITE` docker CT run (34/34, gcp emulator via `http://` exercising the TCP path).

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_bigtable/src/emqx_bridge_bigtable_connector_schema.erl
+++ b/apps/emqx_bridge_bigtable/src/emqx_bridge_bigtable_connector_schema.erl
@@ -70,8 +70,11 @@ fields(connector_config) ->
 %% default: connections to the default endpoint are verified against the OS CA bundle.
 ssl_fields() ->
     [
-        {ssl, SC#{default => #{<<"enable">> => true, <<"verify">> => <<"verify_peer">>}}}
-     || {ssl, SC} <- emqx_connector_schema_lib:ssl_fields()
+        {ssl, #{
+            type => hoconsc:ref(emqx_schema, "ssl_client_opts"),
+            default => #{<<"enable">> => true, <<"verify">> => <<"verify_peer">>},
+            desc => ?DESC(emqx_connector_schema_lib, "ssl")
+        }}
     ].
 
 desc("config_connector") ->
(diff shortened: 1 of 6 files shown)


apps/emqx_bridge_bigtable/src/emqx_bridge_bigtable_connector_schema.erl
```diff
@@ -63,9 +63,17 @@ fields(connector_config) ->
...
+        ssl_fields() ++
         emqx_connector_schema:resource_opts().
 
+%% Same as `emqx_connector_schema_lib:ssl_fields/1', with peer verification on by
+%% default: connections to the default endpoint are verified against the OS CA bundle.
+ssl_fields() ->
+    [
+        {ssl, SC#{default => #{<<"enable">> => true, <<"verify">> => <<"verify_peer">>}}}
+     || {ssl, SC} <- emqx_connector_schema_lib:ssl_fields()
+    ].
```
keynslug, line 78: Looks a bit brittle: implicitly relying on the fact it has just one field. Maybe just reimplement it here instead?
```
    [
        {ssl, #{
            type => hoconsc:ref(emqx_schema, "ssl_client_opts"),
            default => #{<<"enable">> => true, <<"verify">> => <<"verify_peer">>},
            desc => ?DESC(emqx_connector_schema_lib, "ssl")
        }}
    ].
```

keynslug (dismissed):

keynslug (dismissed):

keynslug (approved):

PR #18302, opened 2026-08-08
fix(es): url-encode index and id when composing request path

zmstone:
Release version:

Introduced in:
pre-5.8

## Summary

The Elasticsearch action composes the outbound request path by concatenating the rendered `index` and `id` values verbatim. When these fields are templated (e.g. `index = "${payload.index}"`), a rendered value containing URL-significant characters such as `#`, `/`, `?` or whitespace reshaped the composed path, so the request could target a different endpoint than the configured action.

The request-path template is now rendered with a variable transform that percent-encodes each interpolated value (`uri_string:quote/1`), mirroring `emqx_auth_template:render_urlencoded_str/2`: literal template text (`/_doc/`, `/_update/`) is untouched and only interpolated values are encoded, so ordinary index names are unchanged on the wire and every currently-working configuration keeps working. Encoding was preferred over rejecting unexpected characters for that reason.

Since the same render function also renders the JSON request body (which must stay byte-identical), the path template is tagged with a literal prefix when handed to the HTTP connector, and the tag selects the URL-encoding transform at render time.

A value `uri_string:quote/1` cannot encode (not valid unicode) fails the render with a descriptive error instead of composing a corrupt path.

Covered by a new docker-free eunit module exercising the exact parse/render pipeline the HTTP connector uses: all four action variants (delete / update / create with and without id), plain and non-ASCII values, invalid-unicode failure, and body-rendering invariance.

Same change as #18301 (dev-58); the v5 and v6 lines are separate sync chains, so both base fixes are needed.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_es/src/emqx_bridge_es_connector.erl
+++ b/apps/emqx_bridge_es/src/emqx_bridge_es_connector.erl
@@ -38,7 +38,7 @@
     connector_example_values/0
 ]).
 
--export([render_template/2]).
+-export([render_template/3]).
 -export([convert_server/2]).
 
 %% emqx_connector_resource behaviour callbacks
@@ -71,13 +71,8 @@
 
 -define(CONNECTOR_TYPE, elasticsearch).
 
-%% Literal prefix tagging the request-path template handed to the HTTP
-%% connector, so `render_template/2' can tell it apart from the body
-%% template and URL-encode the interpolated values.
--define(PATH_TEMPLATE_TAG, "__path_template__:").
-
 -ifdef(TEST).
--export([path_template/1]).
+-export([path/1]).
 -endif.
 
 %%-------------------------------------------------------------------------------------
@@ -315,13 +310,13 @@ on_add_channel(
             {error, already_exists};
         _ ->
             Parameter1 = Parameter#{
-                path => path_template(Parameter),
+                path => path(Parameter),
                 method => method(Parameter),
                 body => get_body_template(Parameter)
             },
             ChannelConfig = #{
                 parameters => Parameter1,
-                render_template_func => fun ?MODULE:render_template/2
+                render_template_func => fun ?MODULE:render_template/3
             },
             {ok, State} = emqx_bridge_http_connector:on_add_channel(
                 InstanceId, State0, ChannelId, ChannelConfig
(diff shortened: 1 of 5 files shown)


apps/emqx_bridge_es/src/emqx_bridge_es_connector.erl
```diff
@@ -342,6 +351,12 @@ render_template([<<"update_without_doc_template">>], Msg) ->
     emqx_utils_json:encode(#{<<"doc">> => Msg});
 render_template([<<"create_without_doc_template">>], Msg) ->
     emqx_utils_json:encode(#{<<"doc">> => Msg, <<"doc_as_upsert">> => true});
+render_template([<<?PATH_TEMPLATE_TAG, First/binary>> | Template], Msg) ->
+    %% Request-path template: URL-encode interpolated values so each one
+    %% stays within its path segment; literal template text is untouched.
+    Opts = #{var_trans => fun to_urlencoded_string/2},
+    {String, _Errors} = emqx_template:render([First | Template], {emqx_jsonish, Msg}, Opts),
+    String;
```
keynslug: Nit. Relying on how exactly this "tag" is going to appear here looks brittle to me. As `emqx_bridge_http_connector` under our full control, perhaps it's simpler to just expand `render_template_func` contract, and feed it an extra atom describing what is being rendered (e.g. `path`, `body`, etc)?

apps/emqx_bridge_es/src/emqx_bridge_es_connector.erl
```diff
@@ -354,6 +364,13 @@ render_template(Template, Msg) ->
 
 to_string(Name, Value) ->
     emqx_template:to_string(render_var(Name, Value)).
+to_urlencoded_string(Name, Value) ->
+    try
+        uri_string:quote(to_string(Name, Value))
+    catch
+        throw:{error, Reason, _} ->
+            error({failed_to_urlencode_path_value, Name, Reason})
+    end.
```
keynslug, line 374: Nit.

```suggestion

to_urlencoded_string(Name, Value) ->
    try
        uri_string:quote(to_string(Name, Value))
    catch
        throw:{error, Reason, _} ->
            error({failed_to_urlencode_path_value, Name, Reason})
    end.

```

keynslug (dismissed):

keynslug (approved):

PR #18303, opened 2026-08-08
fix(schema_registry): scope Sparkplug B alias cache to MQTT client processes

zmstone:
Release version:

Introduced in:
6.0.2

## Summary

Fixes emqx/emqx-dev-team-tasks#325.

The Sparkplug B alias-to-name cache (`emqx_schema_registry_spb_state`) is kept in the process dictionary of the process running the `message.publish` hook, assuming that process is always the publishing client's own channel process. The assumption breaks for shared processes: an MQTT source-bridge ingress worker or a delayed-publish pool worker publishes on behalf of many clients from a single process, so one publisher's alias mapping could be applied to another publisher's decoded metrics.

This PR scopes the cache to genuine MQTT client channel processes:

- `emqx_schema_registry_spb_hookcb:on_message_publish/1` now also checks `is_client_process/0`, which inspects `proc_lib:get_label(self())` for the `{clientid, _}` label that `emqx_logger:set_metadata_clientid/1` sets on client channel processes. Shared/system workers carry no such label, so the hook neither populates nor reads the cache there; `get_current_alias_mapping/0` then returns `#{}` and `spb_decode` simply resolves no names.
- The gate lives in the hook so the state module stays a pure per-process cache.

Behavior change to note: SpB data messages ingested through an MQTT bridge no longer get aliases resolved to names by `spb_decode` (they previously resolved from whichever publisher's mapping happened to be cached in the shared ingress process). Direct client publishes are unchanged.

New tests cover the hook in a labeled process (mapping maintained), an unlabeled process (cache untouched), and per-process isolation.

## PR Checklist


apps/emqx_schema_registry/src/emqx_schema_registry_spb_hookcb.erl
```diff
@@ -55,6 +55,17 @@ on_message_publish(#message{} = Message) ->
...
+%% Alias mappings are per-publisher state kept in the process dictionary; only maintain
+%% them when the hook runs in an MQTT client's own channel process, identified by the
+%% `{clientid, _}' proc label set in `emqx_logger:set_metadata_clientid/1'.
+%% Bridge/delayed/system publishes share a process across publishers and must not
+%% populate or read the cache.
+is_client_process() ->
+    case proc_lib:get_label(self()) of
+        {clientid, _} -> true;
+        _ -> false
+    end.
```
zmstone, line 67: this is a bit ad-hoc, will refactor after the code reaches dev-63:

implement label set and get in emqx_channel module.

keynslug (approved):

PR #18312, opened 2026-08-10
fix(coap): defer UDP proxy takeover until token validation

hjianbo:
Release version: 6.3.0

Introduced in: 6.1.4

## Summary

- Scope plaintext CoAP UDP proxy IDs by peer so esockd cannot replace the current proxy before CoAP validation.
- Route candidate replies through a request-scoped socket and commit the persistent downlink handle only after the channel validates the client ID and token.
- Preserve the original downlink handle on rejection while allowing a validated peer change to migrate it.

Fix: emqx/emqx-dev-team-tasks#309

## PR Checklist

### DIFF
--- a/apps/emqx_gateway/src/bhvrs/emqx_gateway_conn.erl
+++ b/apps/emqx_gateway/src/bhvrs/emqx_gateway_conn.erl
@@ -79,7 +79,9 @@
     %% Channel Module
     chann_mod :: atom(),
     %% Listener Tag
-    listener :: listener() | undefined
+    listener :: listener() | undefined,
+    %% Defer UDP proxy ownership changes until the channel validates the packet.
+    defer_udp_proxy_takeover = false :: boolean()
 }).
 
 -type listener() :: {GwName :: atom(), LisType :: atom(), LisName :: atom()}.
@@ -327,7 +329,8 @@ init_state(WrappedSock, Peername, Options, FrameMod, ChannMod) ->
         oom_policy = OomPolicy,
         frame_mod = FrameMod,
         chann_mod = ChannMod,
-        listener = maps:get(listener, Options, undefined)
+        listener = maps:get(listener, Options, undefined),
+        defer_udp_proxy_takeover = maps:get(defer_udp_proxy_takeover, Options, false)
     }.
 
 run_loop(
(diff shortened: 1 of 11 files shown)


apps/emqx_gateway/src/bhvrs/emqx_gateway_conn.erl
```diff
@@ -894,6 +935,60 @@ with_channel(Fun, Args, State = #state{chann_mod = ChannMod, channel = Channel})
             shutdown(Reason, NState1)
     end.
 
+with_udp_proxy_channel(
```
zmstone, line 951: `with_udp_proxy_channel` duplicates all five return clauses of `with_channel` and differs only in the reply and outgoing handling. A later change to how channel return tuples are interpreted must be made in both, and missing one gives the deferred path stale behavior. Consider parameterizing the reply and outgoing handlers so one function decodes the channel return.

apps/emqx_gateway_coap/src/emqx_coap_proxy_conn.erl
```diff
@@ -53,6 +57,18 @@ get_connection_id(_Transport, Peer, State, Data) ->
 peer_id(Peer) ->
     {peer, Peer}.
 
+proxy_connection_id({peer, _} = CId, _Peer) ->
+    CId;
+proxy_connection_id(CId, Peer) ->
+    %% esockd takes over an existing proxy as soon as the same connection ID
+    %% is attached.  Keep candidates isolated until CoAP validates the token.
```
zmstone: i cannot find where the takeover happens.

apps/emqx_gateway/src/bhvrs/emqx_gateway_conn.erl
```diff
@@ -510,8 +518,20 @@ handle_msg(
...
+) ->
+    IdleTimer /= undefined andalso
+        emqx_utils:cancel_timer(IdleTimer),
+    NState = State#state{idle_timer = undefined},
+    handle_udp_proxy_incoming(Packet, NSock, NState);
+handle_msg({outgoing, NSock = {esockd_udp_proxy, _, _}, Data}, State) ->
+    handle_outgoing_via_socket(Data, NSock, State);
 handle_msg({outgoing, Data}, State) ->
     handle_outgoing(Data, State);
+handle_msg({udp_proxy, _Decision}, State) ->
```
zmstone: looks like an unnecessary leak to the common handler.

apps/emqx_gateway_coap/src/emqx_coap_proxy_conn.erl
hjianbo: The takeover now happens as a validated downlink handover. The new peer's UDP proxy looks up the existing connection using the logical client ID and dispatches the packet together with its NSock to that connection process. 
After the CoAP channel validates the token and returns commit, emqx_gateway_conn replaces the persistent State.socket with this NSock. Subsequent downlink packets are therefore sent through the new UDP proxy. On reject, State.socket remains unchanged.

zmstone (dismissed):

zmstone (dismissed):

zmstone (approved):

PR #18333, opened 2026-08-12
fix(plugins): tolerate emqx_plugins as a declared plugin dependency

zmstone:
Release version:

Introduced in:
pre-5.8

## Summary

Plugins are started from inside `emqx_plugins_app:start/2`, and each plugin app start is wrapped in a 10-second kill-timeout (`emqx_plugins_apps:start_app/1`). A plugin whose `.app` file declares `emqx_plugins` in its `applications` list makes `application:ensure_all_started/1` wait for `emqx_plugins` to finish starting, which cannot happen until the plugin start returns. The helper process is killed after 10 seconds and the plugin is left enabled but not running after every node restart. The only signal is one `failed_to_start_plugin_app` log line. The same deadlock applies transitively to any declared dependency that itself depends on `emqx_plugins` (for example `emqx_management`).

Verified on a real node restart (not `emqx_cth_cluster:restart`, which bypasses the boot path): a package declaring `emqx_plugins` fails on every boot with a 10-second stall; the same package without the declaration survives the restart.

Changes:

* `emqx_plugins_apps`: load the plugin app from its parsed spec and drop a declared `emqx_plugins` dependency, with a warning log. The dependency is satisfied by construction: `emqx_plugins` is running or starting whenever a plugin app is started. This also heals already-installed plugin packages without repackaging.
* `emqx_plugins_apps:start_app/1`: on timeout, the error now lists the declared dependencies that were not running and a hint naming the rule.
* In-tree plugins: remove `emqx_plugins` and `emqx_management` from the `.app.src` `applications` lists. Note that the mix-built packages do not carry the `.app.src` list (mix generates the `.app` from `mix.exs`), so the shipped packages of this repo's plugins are not affected; the trim keeps the sources correct for builds that honor `.app.src` (rebar3, as used by third-party plugins).

A dependency that transitively reaches `emqx_plugins` still fails at boot; it now fails with diagnostics naming the not-running dependencies. Dropping such dependencies automatically was considered and rejected: unlike `emqx_plugins`, they are genuinely not running when plugins start at boot. Starting plugins after `emqx_plugins_app:start/2` returns would fix this class at the root; tracked as a follow-up in emqx/emqx-dev-team-tasks#338.

Same fix shape as the dev-510 PR (v5 does not forward-sync into v6).

## PR Checklist

### DIFF
--- a/apps/emqx_plugins/src/emqx_plugins_apps.erl
+++ b/apps/emqx_plugins/src/emqx_plugins_apps.erl
@@ -323,7 +323,8 @@ do_load_plugin_app(AppName, Ebin) ->
     Modules = filelib:wildcard(filename:join([Ebin, "*.beam"])),
     maybe
         ok ?= load_modules(Modules),
-        ok ?= application:load(AppName)
+        {ok, AppSpec} ?= read_app_spec(AppName, Ebin),
+        ok ?= application:load(drop_self_dep(AppSpec))
     else
         {error, {already_loaded, _}} ->
             ok;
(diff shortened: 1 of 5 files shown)


apps/emqx_plugins/src/emqx_plugins_apps.erl
```diff
@@ -335,6 +336,40 @@ do_load_plugin_app(AppName, Ebin) ->
...
+    case lists:member(emqx_plugins, Deps) of
+        true ->
+            ?SLOG(warning, #{
+                msg => "plugin_app_declares_emqx_plugins_dependency",
+                name => AppName,
+                hint =>
+                    "remove emqx_plugins from the applications list"
+                    " in the plugin app's .app.src file"
+            }),
+            Deps1 = {applications, lists:delete(emqx_plugins, Deps)},
```
thalesmg, line 367: maybe we ought to generalize this to any umbrella app that depends on emqx_plugin?

e.g.: we could check that with machine_boot and remove any emqx_plugin dependents

thalesmg (approved):

apps/emqx_plugins/src/emqx_plugins_apps.erl
zmstone: there is another pr which moves emqx_plugin to the end of boot sequence.

thalesmg: adding this check would be future-proof, if there's ever an application that depends on emqx_plugins.

PR #18334, opened 2026-08-12
fix(plugins): tolerate emqx_plugins as a declared plugin dependency

zmstone:
Release versions:
5.10.5

Introduced in:
5.10.4

## Summary

Plugins are started from inside the `emqx_plugins` application's own `start/2` callback. A plugin that declares `emqx_plugins` in its `.app.src` applications list makes `application:ensure_all_started/1` wait for `emqx_plugins` to finish starting, which cannot happen until the plugin start returns. The plugin start then times out after 10 seconds on every node boot. The node comes up with the plugin enabled but not running, delayed by 10 seconds per affected plugin.

All three bundled plugins declared `emqx_plugins`; `emqx_backup_sync` also declared `emqx_management`, which depends on `emqx_plugins` and starts after it, so it fails the same way. Verified on a real `bin/emqx stop`/`start`: three serial 10-second timeouts, all three plugins left enabled but not running.

Changes:

- Drop a declared `emqx_plugins` dependency from the plugin app spec at load time and log a warning. The dependency is satisfied by construction, and dropping it also repairs already-installed plugin packages that declare it.
- When a plugin start times out, the error log now lists the declared applications that were not running at that moment.
- Remove `emqx_plugins` from the three bundled plugins' `.app.src`, and `emqx_management` from `emqx_backup_sync`. Bump `emqx_offline_messages` to 2.0.1.

The same fix for v6 is https://github.com/emqx/emqx/pull/18333. v5 does not forward-sync into v6, so the two PRs land independently.

## PR Checklist


thalesmg (approved):
don't the other 2 plugins also need version bumps?

zmstone:
> don't the other 2 plugins also need version bumps?

no, only changed should bump version

PR #18336, opened 2026-08-12
fix(api): redact secrets in listeners, exhook and audit read endpoints (5.8)

zmstone:
Release versions:
5.8.12, 5.9.4, 5.10.5

Introduced in:
pre-5.8

## Summary

v5 counterpart of emqx/emqx#18330. Fixes findings A2, A3 and A5 of emqx/emqx-dev-team-tasks#333 for the 5.8 / 5.9 / 5.10 lines.

- `GET /listeners` and `GET /listeners/{id}` now render the listener `ssl_options.password` as `******`.
- `GET /exhooks` and `GET /exhooks/{name}` now render the gRPC client `ssl.password` as `******`.
- Audit log entries for `POST /license` now record the request body as `******`, so the license key does not appear in `GET /audit` results.

The `PUT` handlers for listeners and exhook servers restore `******` placeholders from the stored config, so a body round-tripped from `GET` does not overwrite stored secrets.

A1 and A4 are out of scope; A4 is tracked in emqx/emqx-dev-team-tasks#335.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_listeners.erl
+++ b/apps/emqx/src/emqx_listeners.erl
@@ -123,7 +123,9 @@ do_list_raw() ->
             Key = <<"listeners">>,
             Raw = emqx_config:get_raw([Key], #{}),
             SchemaMod = emqx_config:get_schema_mod(Key),
-            #{Key := RawWithDefault} = emqx_config:fill_defaults(SchemaMod, #{Key => Raw}, #{}),
+            #{Key := RawWithDefault} = emqx_config:fill_defaults(SchemaMod, #{Key => Raw}, #{
+                obfuscate_sensitive_values => true
+            }),
             Listeners = maps:to_list(RawWithDefault),
             lists:flatmap(fun format_raw_listeners/1, Listeners);
         false ->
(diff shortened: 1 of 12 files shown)


apps/emqx/src/emqx_listeners.erl
```diff
@@ -123,7 +123,9 @@ do_list_raw() ->
             Key = <<"listeners">>,
             Raw = emqx_config:get_raw([Key], #{}),
             SchemaMod = emqx_config:get_schema_mod(Key),
-            #{Key := RawWithDefault} = emqx_config:fill_defaults(SchemaMod, #{Key => Raw}, #{}),
+            #{Key := RawWithDefault} = emqx_config:fill_defaults(SchemaMod, #{Key => Raw}, #{
+                obfuscate_sensitive_values => true
```
thalesmg, line 127: since this is a general module (not an http api module), maybe it would be better to either move the redaction to the api modules that use this value, because they are responsible for formatting data to be presented to users, or maybe rename `list_raw` to something like `list_raw_redacted`?

otherwise, if another module wants to get the raw config for internal usage, it'll break weirdly because potentially necessary data will be redacted.

thalesmg (dismissed):
it would be good to update to hocon 0.46.3 too, because there was a bug in the redaction for hoconsc:array values that was fixed there

apps/emqx/src/emqx_listeners.erl
zmstone: will improve in v6, but will keep v5 minimal.

thalesmg (approved):

PR #18337, opened 2026-08-12
fix(plugins): start plugin apps after all EMQX apps

zmstone:
Release version:

Introduced in:
pre-5.8

## Summary

Part of emqx/emqx-dev-team-tasks#338 (item 1 only).

Plugin applications used to start from inside `emqx_plugins_app:start/2`, in the middle of the boot sequence. `emqx_plugins_apps:start_app/1` wraps `application:ensure_all_started/1` in a 10-second timeout, so a plugin that declares an application which starts later in the boot sequence (for example `emqx_management`) stalled for 10 seconds and was left enabled but not running after every node restart.

This PR moves the plugin start to the end of the boot sequence:

* `emqx_plugins_app:start/2` no longer calls `emqx_plugins:ensure_started/0`. It still installs (extracts and loads) the configured plugins and registers the config handler.
* `emqx_machine_boot:ensure_apps_started/0` calls `emqx_plugins:ensure_started/0` after all reboot applications have started. This is the same function `stop_apps/0` mirrors with `emqx_plugins:ensure_stopped/0`.

Decisions:

* The plugin start goes into `ensure_apps_started/0`, not `post_boot/0`. The ekka cluster join/leave callback re-runs `ensure_apps_started/0` directly (`start_autocluster/0`), so plugins also come back after a cluster join. `post_boot/0` is not invoked on that path.
* Plugins start before the `emqx_machine_boot_apps_started` trace point, so a test that waits for it sees plugins running.
* A broken plugin cannot fail the boot: `emqx_plugins:ensure_started/0` catches per-plugin errors and aggregates them into an error log (`for_plugins_action_error_occurred`). A new test case covers this and the idempotency of the re-run on cluster join.

Known widening of the hook-registration window (item 2 of the issue): listeners already serve traffic before plugin hooks are registered, and starting plugins later widens that window. This is intentional here; a follow-up PR adds a readiness gate so listeners do not serve MQTT traffic until all enabled plugins are running.

Not reverted: the load-time `emqx_plugins` dependency strip and `not_running_deps` diagnostic from #18333 stay, for already-built plugin packages that declare `emqx_plugins`.

Follow-up: update the plugin development docs (emqx-docs) to state that a plugin may declare any EMQX application in `applications`.

Validation on a real release build (CT cannot catch this; `emqx_cth_cluster:restart` bypasses the real boot path):

* Before: a node with an enabled plugin declaring `emqx_management` stalls 10 seconds at the `emqx_plugins` boot slot; the plugin ends up not running.
* After: the same node starts clean and the plugin is running.

## PR Checklist

### DIFF
--- a/apps/emqx_machine/src/emqx_machine_boot.erl
+++ b/apps/emqx_machine/src/emqx_machine_boot.erl
@@ -102,6 +102,11 @@ stop_one_app(App) ->
 ensure_apps_started() ->
     ?SLOG(notice, #{msg => "(re)starting_emqx_apps"}),
     lists:foreach(fun start_one_app/1, sorted_reboot_apps()),
+    %% Start plugin applications only after all EMQX applications are up,
+    %% so a plugin may depend on any of them.
+    %% Note: this function is also the ekka cluster join/leave callback,
+    %% so plugins restart after a join as well (see `start_autocluster/0').
+    ok = emqx_plugins:ensure_started(),
     ?tp(emqx_machine_boot_apps_started, #{}).
 
 start_one_app(App) ->
--- a/apps/emqx_machine/src/emqx_machine_boot.erl
+++ b/apps/emqx_machine/src/emqx_machine_boot.erl
@@ -103,7 +103,9 @@ ensure_apps_started() ->
     ?SLOG(notice, #{msg => "(re)starting_emqx_apps"}),
     lists:foreach(fun start_one_app/1, sorted_reboot_apps()),
     %% Start plugin applications only after all EMQX applications are up,
-    %% so a plugin may depend on any of them.
+    %% so a plugin may depend on any of them.  Plugin apps are not part of
+    %% `sorted_reboot_apps/0'; `stop_apps/0' stops them explicitly via
+    %% `emqx_plugins:ensure_stopped/0'.
     %% Note: this function is also the ekka cluster join/leave callback,
     %% so plugins restart after a join as well (see `start_autocluster/0').
     ok = emqx_plugins:ensure_started(),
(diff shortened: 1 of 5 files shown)


apps/emqx_machine/src/emqx_machine_boot.erl
```diff
@@ -102,6 +102,11 @@ stop_one_app(App) ->
 ensure_apps_started() ->
     ?SLOG(notice, #{msg => "(re)starting_emqx_apps"}),
     lists:foreach(fun start_one_app/1, sorted_reboot_apps()),
+    %% Start plugin applications only after all EMQX applications are up,
+    %% so a plugin may depend on any of them.
+    %% Note: this function is also the ekka cluster join/leave callback,
+    %% so plugins restart after a join as well (see `start_autocluster/0').
+    ok = emqx_plugins:ensure_started(),
```
thalesmg, line 111: should also stop the plugins in `stop_apps/0`, since they are not listed in sorted_reboot_apps?

apps/emqx_plugins/test/emqx_plugins_SUITE.erl
```diff
@@ -213,6 +213,35 @@ t_demo_install_start_stop_uninstall(Config) ->
...
+    NameVsn = proplists:get_value(name_vsn, Config),
+    ok = emqx_plugins:ensure_installed(NameVsn),
+    Broken = #{name_vsn => <<"missing_plugin-1.0.0">>, enable => true},
+    Good = #{name_vsn => bin(NameVsn), enable => true},
+    ok = emqx_plugins:put_configured([Broken, Good]),
+    ?assertEqual(ok, emqx_plugins:ensure_started()),
+    ?assert(is_app_running(?EMQX_PLUGIN_APP_NAME)),
+    %% idempotent (cluster join re-runs the boot tail)
+    ?assertEqual(ok, emqx_plugins:ensure_started()),
+    ?assert(is_app_running(?EMQX_PLUGIN_APP_NAME)),
```
thalesmg: would need to call stop_apps to simulate join?

thalesmg (approved):

apps/emqx_machine/src/emqx_machine_boot.erl
zmstone: `emqx_machine_boot:stop_apps()` calls `emqx_plugins:ensure_stopped()`

PR #18341, opened 2026-08-12
ci: check emqx_cluster_rpc:multicall targets against an exemption list

zmstone:
## Summary

A node that boots after downtime does not replay the cluster_rpc transactions it missed: `emqx_conf_app:init_load/1` copies the cluster config from a peer and `emqx_cluster_rpc:maybe_init_tnx_id/2` fast-forwards the commit pointer past them. Any `emqx_cluster_rpc:multicall` target whose effect is not recoverable from the copied config or from a replicated mria table is silently lost on that node. Reproduced with a cluster-wide plugin delete: the rebooted node keeps the deleted package and it resurfaces in the cluster-wide plugins API.

This PR adds `scripts/check-cluster-rpc-mfas.escript`, a standalone checker run by `make static_checks` after the check profile is compiled. xref over the compiled release yields every caller of `emqx_cluster_rpc:multicall/3,5`; the caller's `bpapi_meta/0` (generated by `emqx_bpapi_trans`, which only accepts literal target atoms) yields the target MFA. In-tree plugins are in scope: the check profile builds without `plugins/*`, so the script compiles `plugins/*/src` itself (debug_info, against the analyzed lib dir) and adds the beams to the xref graph. The check fails when:
- a multicall target is neither a config update (`emqx:update_config/remove_config/reset_config`) nor an exempted function,
- a module that is not a BPAPI proto module calls `multicall` (this is what any plugin caller reports as — plugins must not use cluster_rpc at all, since their MFAs are not replayed on a rebooted node),
- an exemption has no remaining call site.

The exemption list documents the rules a new target must follow (durable state to config or mria before the multicall; the MFA only rebuilds node-local runtime state) and grandfathers the existing targets with per-entry reasons. The v6 counterpart is a mix task in #18342.


scripts/check-cluster-rpc-mfas.escript
```diff
@@ -0,0 +1,228 @@
...
+    Opts = [
+        debug_info,
+        report_errors,
+        {outdir, OutDir},
+        {i, filename:join(PluginDir, "src")},
+        {i, filename:join(PluginDir, "include")}
+    ],
+    lists:foreach(
+        fun(Src) ->
+            case compile:file(Src, Opts) of
```
thalesmg: maybe use the plugin's build tool?

thalesmg (dismissed):
maybe this could be added to d58, so that violations are caught earlier upstream?

scripts/check-cluster-rpc-mfas.escript
zmstone: changed to load from make emqx-enterprise-compile target

thalesmg (approved):

PR #18342, opened 2026-08-12
ci: check emqx_cluster_rpc:multicall targets against an exemption list

zmstone:
## Summary

Counterpart of #18341 (dev-510) for the v6 line.

A node that boots after downtime does not replay the cluster_rpc transactions it missed: `emqx_conf_app:init_load/1` copies the cluster config from a peer and `emqx_cluster_rpc:maybe_init_tnx_id/2` fast-forwards the commit pointer past them. Any `emqx_cluster_rpc:multicall` target whose effect is not recoverable from the copied config or from a replicated mria table is silently lost on that node. Reproduced with a cluster-wide plugin delete: the rebooted node keeps the deleted package and it resurfaces in the cluster-wide plugins API.

This PR adds the standalone mix task `emqx.check_cluster_rpc`, chained into `make static_checks`. xref over the compiled umbrella apps yields every caller of `emqx_cluster_rpc:multicall/3,5`; the caller's `bpapi_meta/0` (generated by `emqx_bpapi_trans`, which only accepts literal target atoms) yields the target MFA. In-tree plugins are in scope: they are separate mix projects outside the umbrella build, so the task compiles `plugins/*/src` itself (debug_info, against the loaded code path) and adds the beams to the xref graph. The check fails when:
- a multicall target is neither a config update (`emqx:update_config/remove_config/reset_config`) nor an exempted function,
- a module that is not a BPAPI proto module calls `multicall` (this is what any plugin caller reports as — plugins must not use cluster_rpc at all, since their MFAs are not replayed on a rebooted node),
- an exemption has no remaining call site.

The exemption list documents the rules a new target must follow (durable state to config or mria before the multicall; the MFA only rebuilds node-local runtime state) and grandfathers the existing targets with per-entry reasons. Verified against a full `make emqx-enterprise-compile` build: clean tree passes with all 4 in-tree plugins compiled into the graph; an injected proto module with a new target, a non-proto multicall caller, and a plugin module calling multicall all fail the task with exit 1.


Makefile
```diff
@@ -112,7 +110,8 @@ static_checks: $(ELIXIR_COMMON_DEPS)
 	@env BPAPI_BUILD_PROFILE=$(PROFILE:%-test=%) \
 	    $(MIX) do \
 	    emqx.xref, dialyzer --mode classic, \
-	    emqx.static_checks
+	    emqx.static_checks, \
```
thalesmg: since we're touching this, maybe we could also switch to using `+`  instead of `,`, so syncing to latter branches is simplified

apps/emqx_mix_utils/lib/mix/tasks/emqx.check_cluster_rpc.ex
```diff
@@ -0,0 +1,240 @@
...
+    opts = [
+      :debug_info,
+      :report_errors,
+      outdir: to_charlist(out_dir),
+      i: to_charlist(Path.join(plugin_dir, "src")),
+      i: to_charlist(Path.join(plugin_dir, "include"))
+    ]
+
+    for src <- Path.wildcard(Path.join(plugin_dir, "src/**/*.erl")) do
+      case :compile.file(to_charlist(src), opts) do
```
thalesmg: this unconditionally compiles every module every time this task is run, and in a simple way (i.e., if the plugin has something more complex in its rebar.config/mix.exs, it'll be missed).

maybe we could delegate to the plugin's build tool?

apps/emqx_mix_utils/lib/mix/tasks/emqx.check_cluster_rpc.ex
```diff
@@ -0,0 +1,240 @@
...
+  ]
+
+  @impl true
+  def run(args) do
+    if args != [] do
+      Mix.raise("Unknown options:\n  #{inspect(args, pretty: true)}")
+    end
+
+    callers = find_multicall_callers()
+    {violations, targets} = Enum.reduce(callers, {[], []}, &collect/2)
```
thalesmg, line 94:
```suggestion
    callers = find_multicall_callers()
    :xref.stop(@xref)
    {violations, targets} = Enum.reduce(callers, {[], []}, &collect/2)
```

thalesmg (dismissed):

thalesmg (approved):

PR #18354, opened 2026-08-13
feat(authn): implement safe password hashing defaults

savonarola:
Release version: 6.4.0

Introduced in: N/A

## Summary

- Apply security-profile-specific password hashing defaults to built-in database authenticators. The hardened profile generates high-entropy passwords by default and restricts manually supplied passwords.
- Store the password hashing algorithm with each built-in authentication user so existing users remain verifiable after authenticator configuration changes.
- Preserve built-in authentication users during rolling upgrades by migrating legacy records and supporting mixed-version reads and writes.

## Motivation

Most password hashing algorithms available in EMQX are insecure by default, while PBKDF2 and bcrypt do not enforce OWASP-recommended minimums. Applying expensive password hashing to every client does not fit high-rate MQTT connection workloads and increases DoS risk. The PR separates the two use cases to make possible safe defaults in hardened mode.

* assuming large automated deployments, by default we autogenerate password and use lightweight hashing 
* when switching to manually managed passwords, strong PBKDF2 defaults are used.

## PR Checklist


apps/emqx_auth_mnesia/src/emqx_authn_mnesia.erl
```diff
@@ -457,6 +594,7 @@ do_insert_user(UserInfoRecord, Namespace, UserGroup, UserId, Opts) ->
             ok = insert_user(UserInfoRecord),
             {success, Namespace};
         [UserInfoRecord] ->
+            delete_legacy_global_txn(Namespace, UserGroup, UserId),
```
zmstone: we must ensure migration does not happen when cluster has older version nodes ?

zmstone (commented):
please target dev-64

zmstone (approved):

PR #18356, opened 2026-08-14
fix(boot): gate connections until node boot completes (v5)

zmstone:
Release versions:
5.8.12, 5.9.4, 5.10.5

Introduced in:
pre-5.8

## Summary

Fixes emqx/emqx-dev-team-tasks#341.

MQTT listeners open before all applications have started. Plugins that install `client.authenticate` / `client.authorize` (or other) hooks start late in the boot sequence, so clients connecting in that window bypass the plugin's security checks.

This PR adds a node readiness gate:

- New `emqx_node_readiness` module holds a readiness flag in a `persistent_term`. The flag defaults to ready, so contexts that do not boot through `emqx_machine` (test suites) are unaffected.
- `emqx_machine_boot:ensure_apps_started/0` clears the flag before starting apps and sets it after all apps (including plugins) are up. This function is also the ekka cluster join callback, so the node is re-gated during a rejoin. If boot never completes, the node keeps refusing connections.
- Connection process inits check the flag and close the socket before reading any bytes while the node is not ready: `emqx_connection` (TCP/SSL), `emqx_ws_connection` (WS/WSS), `emqx_quic_connection` (QUIC, refused in the acceptor callback before the TLS handshake), and `emqx_gateway_conn` (gateway connections).
- `GET /status` reports HTTP 503 while the node is not ready, so load balancers and orchestration can route around the node during the startup window.

Refusal observability:

- TCP/SSL refusals exit with `{shutdown, node_not_ready}`, so esockd records them in the per-listener shutdown counter (visible in `emqx ctl listeners`). A test asserts the counter is bumped. The refusal also logs at `info` level with the transport module and peername.
- WS/WSS refusals cannot feed that counter: the cowboy path does not go through esockd's connection supervisor, and `emqx_listeners:shutdown_count/3` returns `[]` for ws/wss listeners. The WS refusal therefore logs a distinct `info` message (`ws_connection_refused_before_boot_complete`).
- QUIC refusals shut down the connection with a dedicated application error code (`MQTT_QUIC_CONN_ERROR_NOT_READY`) and log a distinct `info` message with the peername; QUIC listeners have no shutdown counter either.
- Gateway refusals log at `warning` level with the peername and channel module, since gateway listeners have no shutdown-counter observability.

Notes:

- Gateways without a per-connection process (connectionless UDP paths) are not covered by this gate.
- No schema change.
- The `emqx_access_control:default_authn_result()` = `ignore` behavior is kept as defense in depth.

Sibling dev-60 PR: same design and module names on the v6 line.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_connection.erl
+++ b/apps/emqx/src/emqx_connection.erl
@@ -308,9 +308,19 @@ init(Parent, Transport, RawSocket, Options) ->
         false ->
             %% Refuse to serve before node boot completes: authn/authz
             %% hooks (e.g. from plugins) may not be installed yet.
-            ?SLOG_THROTTLE(warning, #{msg => connection_refused_before_boot_complete}),
+            %% The exit reason feeds the listener's shutdown counter.
+            Peername =
+                case Transport:peername(RawSocket) of
+                    {ok, Peer} -> esockd:format(Peer);
+                    {error, _} -> unknown
+                end,
+            ?SLOG(info, #{
+                msg => connection_refused_before_boot_complete,
+                transport => Transport,
+                peername => Peername
+            }),
             ok = Transport:fast_close(RawSocket),
-            erlang:exit(normal)
+            erlang:exit({shutdown, node_not_ready})
     end.
 
 do_init(Parent, Transport, RawSocket, Options) ->
--- a/apps/emqx/src/emqx_connection.erl
+++ b/apps/emqx/src/emqx_connection.erl
@@ -302,6 +302,18 @@ stop(Pid) ->
 %%--------------------------------------------------------------------
 
 init(Parent, Transport, RawSocket, Options) ->
+    case emqx_node_readiness:is_ready() of
+        true ->
+            do_init(Parent, Transport, RawSocket, Options);
+        false ->
+            %% Refuse to serve before node boot completes: authn/authz
+            %% hooks (e.g. from plugins) may not be installed yet.
+            ?SLOG_THROTTLE(warning, #{msg => connection_refused_before_boot_complete}),
+            ok = Transport:fast_close(RawSocket),
+            erlang:exit(normal)
+    end.
+
+do_init(Parent, Transport, RawSocket, Options) ->
     case Transport:wait(RawSocket) of
         {ok, Socket} ->
             run_loop(Parent, init_state(Transport, Socket, Options));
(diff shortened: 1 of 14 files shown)


apps/emqx_gateway/src/bhvrs/emqx_gateway_conn.erl
```diff
@@ -279,6 +279,18 @@ is_datadram_socket({esockd_udp_proxy, _ProxyId, Sock}) -> erlang:is_port(Sock).
 %%--------------------------------------------------------------------
 
 init(Parent, WrappedSock, Peername0, Options, FrameMod, ChannMod) ->
+    case emqx_node_readiness:is_ready() of
+        true ->
+            do_init(Parent, WrappedSock, Peername0, Options, FrameMod, ChannMod);
+        false ->
+            %% Refuse to serve before node boot completes: authn/authz
+            %% hooks (e.g. from plugins) may not be installed yet.
+            ?SLOG_THROTTLE(warning, #{msg => connection_refused_before_boot_complete}),
```
zmstone: log peername, and ChannMod

apps/emqx/src/emqx_connection.erl
```diff
@@ -302,6 +302,18 @@ stop(Pid) ->
 %%--------------------------------------------------------------------
 
 init(Parent, Transport, RawSocket, Options) ->
+    case emqx_node_readiness:is_ready() of
+        true ->
+            do_init(Parent, Transport, RawSocket, Options);
+        false ->
+            %% Refuse to serve before node boot completes: authn/authz
+            %% hooks (e.g. from plugins) may not be installed yet.
+            ?SLOG_THROTTLE(warning, #{msg => connection_refused_before_boot_complete}),
```
zmstone: log Transport, and source IP:port

apps/emqx/src/emqx_ws_connection.erl
```diff
@@ -284,6 +284,17 @@ check_origin_header(Req, #{listener := {Type, Listener}} = Opts) ->
     end.
 
 websocket_init([Req, Opts]) ->
+    case emqx_node_readiness:is_ready() of
+        true ->
+            do_websocket_init(Req, Opts);
+        false ->
+            %% Refuse to serve before node boot completes: authn/authz
+            %% hooks (e.g. from plugins) may not be installed yet.
+            ?SLOG_THROTTLE(warning, #{msg => connection_refused_before_boot_complete}),
```
zmstone:
```suggestion
            ?SLOG_THROTTLE(warning, #{msg => "ws_connection_refused_before_boot_complete"}),
```

thalesmg (dismissed):
quic_connection also needs a gate?

qzhuyan (approved):

Issue #18370, opened 2026-08-18
[6.3.0] Backup file list empty when namespace contains wildcard metacharacters


Le9Wang:
### What happened?

When a namespace contains special characters that are also wildcard metacharacters
(e.g., `*`, `{`, `}`), backups created under that namespace are exported successfully
but do not appear in the backup file list on the dashboard.

Example namespace: `ns~%^&*-={}<>@#1`

### Reproduction Steps

1. Create a namespace with special characters in its name, e.g. `ns~%^&*-={}<>@#1`.
2. Log in to the dashboard as that namespace user.
3. Go to the backup page and create a backup.
4. Observe that the backup does not appear in the list.

### Expected Behavior

The backup should be visible in the file list after export.

### Actual Behavior

The backup file is created on disk (export succeeds), but the list API returns
no results for that namespace.

### Root Cause

`emqx_mgmt_data_backup:backup_files/1` constructs a wildcard pattern using
`filelib:wildcard/1` to list backup files:

```erlang
backup_files(Namespace) ->
    {ok, Pattern} = backup_path(Namespace, "*" ++ ?TAR_SUFFIX),
    filelib:wildcard(Pattern).
```

When the namespace contains `*`, `{`, or `}` (among others), `filelib:wildcard`
interprets them as glob metacharacters rather than literal path characters,
causing the pattern to match the wrong directory or fail to match entirely.

The export itself succeeds because `namespaced_backup_dir/1` uses
`filelib:safe_relative_path` + `ensure_path`, which treat these characters as
ordinary filename characters.

**Affected file:** `apps/emqx_management/src/emqx_mgmt_data_backup.erl`

### Suggested Fix

Either escape wildcard metacharacters in the namespace before passing to
`filelib:wildcard/1`, or replace `filelib:wildcard` with `file:list_dir/1`
and filter for `*.tar.gz` files directly.

### Environment

- **Software version / build**: 6.3.0-beta.2-ga1587007 (Enterprise)
- **OS / Platform**: N/A

zmstone:
fix by https://github.com/emqx/emqx/pull/18372

zmstone:
fixed by https://github.com/emqx/emqx/pull/18377

PR #18372, opened 2026-08-18
fix(mgmt): keep namespaced backups within the namespace's own directory

zmstone:

Release version:

Introduced in:
6.1.4 (namespaced backup directories)

## Summary

**Map each namespace to exactly one backup directory leaf under `<root>/ns/`.**
`filelib:safe_relative_path("ns/..", Root)` returns `[]` (not `unsafe`), and `filename:join(Root, [])` is `Root`, so `namespaced_backup_dir(<<"..">>)` returned the global backup root: a namespaced caller listed, downloaded and deleted the global administrator's archives through `list_files/1`, `read_file/2` and `delete_file/2`. A namespace named `.` similarly resolved to the shared `ns` container directory, and an interior `..` (e.g. `a/../b`) resolved to another namespace's directory.

`namespaced_backup_dir/1` now requires the resolved path to be `ns/<name>` verbatim and returns `{error, bad_namespace}` otherwise (400 on the REST API; listing returns an empty list). In other words: a namespace whose name cannot be used as a single directory component (`.`, `..`, names containing path separators or interior `..`) has no backup support on this release line. Such names also cannot appear as one tar entry path segment, which the namespaced archive format requires (`<base>/ns/<NS>/cluster.hocon`).

Namespace name validation at creation is intentionally **not** part of this PR: refusing names that are accepted today would change API behavior on released lines (6.0 to 6.2). It will be proposed separately against `dev-63` as a breaking change for 6.3.0.

Not changed here: the `filelib:wildcard/1` listing issue with glob metacharacters in file names (#18370) is a separate fix.

Most relevant file: `apps/emqx_management/src/emqx_mgmt_data_backup.erl`.

## PR Checklist

### DIFF
--- a/apps/emqx_mt/src/emqx_mt_config.erl
+++ b/apps/emqx_mt/src/emqx_mt_config.erl
@@ -508,13 +508,16 @@ validate_ns_name(Ns) when is_binary(Ns) ->
 invalid_ns_name_msg() ->
     <<
         "Invalid namespace name: only ASCII letters, digits and the"
-        " characters '.', '-', '_' and '~' are allowed,"
+        " characters '.', '-' and '_' are allowed,"
         " with length from 1 to 255; '.' and '..' are not valid names"
     >>.
 
 %% `\z' rather than `$': `$' also matches just before a trailing newline.
+%% `~' is excluded even though URLs allow it: shells expand it to a home
+%% directory, making a directory named after such a namespace hazardous to
+%% manage by hand.
 has_only_safe_ns_chars(Ns) ->
-    match =:= re:run(Ns, <<"^[A-Za-z0-9._~-]+\\z">>, [{capture, none}]).
+    match =:= re:run(Ns, <<"^[A-Za-z0-9._-]+\\z">>, [{capture, none}]).
 
 %% Bulk import both creates and updates namespace configs. Only names that
 %% would be created are validated: existing namespaces keep working.
--- a/apps/emqx_mt/src/emqx_mt_config.erl
+++ b/apps/emqx_mt/src/emqx_mt_config.erl
@@ -74,6 +74,10 @@
 
 -define(disabled, disabled).
 
+%% Maximum byte size of a managed namespace name; it must fit in a single
+%% directory name on common filesystems.
+-define(NS_NAME_MAX_LENGTH, 255).
+
 -type limiter_config() :: #{
     ?tenant => disabled | tenant_config(),
     ?client => disabled | client_config()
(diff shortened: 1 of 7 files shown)


apps/emqx_mt/src/emqx_mt_config.erl
```diff
@@ -472,12 +480,69 @@ limiter_update_side_effects(Ns, {?changed, OldConfig, NewConfig}) ->
...
+            Ns =/= <<"..">> andalso
+            byte_size(Ns) =< ?NS_NAME_MAX_LENGTH andalso
+            has_only_safe_ns_chars(Ns),
+    case Valid of
+        true -> ok;
+        false -> {error, invalid_ns_name_msg()}
+    end.
+
+invalid_ns_name_msg() ->
+    <<
```
savonarola: nit: `~b"""`

changes/ee/fix-18372.en.md
```diff
@@ -0,0 +1,5 @@
+Validate managed namespace names at creation, and keep every namespace's backup files in its own directory.
+
+A managed namespace name may now contain only ASCII letters, digits, and the characters `.`, `-`, and `_`, with a length of 1 to 255 bytes; the names `.` and `..` are not accepted. This ensures every namespace name works correctly in URLs and on the file system. Namespaces that already exist are not affected.
```
savonarola: IMO this exposes a leaky abstraction/interface. Conceptually, not quite clear why scoping mqtt clients has to be related to OS filename conventions anyhow.

savonarola: Maybe we could use something like `slug(NS)."-".base62(NS)` for filepaths (`foo/../bar/^%` -> `foo-bar-xX6AWhkSz`) to avoid adding validations for fs/url/other yet unknown uses

savonarola: We also have `emqx_ft_fs_util:escape_filename/1`

zmstone: there is no good reason to support those special chars as ns, especially ns can be used as topic words, supporting `/` opens a whole to ACL

zmstone: decided to fix it non-breaking in this pr

zmstone: that is, refuse to create backup if the namespace string is not safe.

savonarola: >there is no good reason to support those special chars as ns, especially ns can be used as topic words, supporting / opens a whole to ACL

the idea is not to support some concrete characters, but doing that consistently: if `.-.` ns is good, then `..` probably should be also; or we may discard these special chars completely. We do not need to support `/`.

zmstone: this pr now checks if path is safe, no checks for special chars. 
if this pr looks good, let’s move the special char validation discussion to https://github.com/emqx/emqx/pull/18377

zmstone: about base62, it creates a compatibility challenge.
given the odd ns strings are unlikely to happen in legitimate use cases i’m inclined to keep it as a defensive fix

savonarola (approved):

zmstone:
related fix: https://github.com/emqx/emqx/pull/18377

PR #18377, opened 2026-08-18
feat(mt): validate managed namespace names at creation

zmstone:
Follow-up to #18372.

Release version:
6.3.0

Introduced in:
N/A (behavior restriction for 6.3.0)

## Summary

Validate managed namespace names when they are created. A name may contain only ASCII letters, digits, and the characters `.`, `-` and `_`, with a length of 1 to 255 bytes; the names `.` and `..` are not accepted.

A managed namespace name is used as a URL path segment, as a directory name under the backup root, and as a tar entry path segment inside namespaced backup archives (`<base>/ns/<NS>/cluster.hocon`), so it must be a single safe path component in all of them. The allowed set is the URL-unreserved character set minus `~` (shells expand a leading `~` to a home directory, so directories named after such a namespace are hazardous to manage by hand).

Enforced in `emqx_mt_config:create_managed_ns/1` (behind `POST /mt/ns/:ns`, next to the existing deny-list check) and in bulk config import for namespaces the import would create. Validation applies only at creation:

* Bulk import validates only names that do not already exist, so configs of pre-existing namespaces with out-of-policy names can still be imported.
* `POST /mt/bulk_import_ns_configs` needs no validation: it refuses namespaces that are not already managed.
* Backup restore (`on_backup_table_imported`) writes tables directly and is not affected, so restoring a backup that contains legacy names keeps working.
* Names presented by clients at CONNECT (`client_attrs.tns`) are intentionally not validated: rejecting them would break existing deployments on upgrade.

This lands on `dev-63` only (a breaking change for 6.3.0). Released lines 6.0 to 6.2 keep accepting such names; there, #18372 makes backup operations unavailable for names that cannot be used as a directory name.

## PR Checklist

### DIFF
--- a/apps/emqx_mt/src/emqx_mt_config.erl
+++ b/apps/emqx_mt/src/emqx_mt_config.erl
@@ -75,6 +75,10 @@
 
 -define(disabled, disabled).
 
+%% Maximum byte size of a managed namespace name; it must fit in a single
+%% directory name on common filesystems.
+-define(NS_NAME_MAX_LENGTH, 255).
+
 -type limiter_config() :: #{
     ?tenant => disabled | tenant_config(),
     ?client => disabled | client_config()
@@ -173,7 +177,8 @@ update_managed_ns_config(Ns, Configs) ->
         ]
     }}
     | {error, {duplicated_nss, [emqx_mt:tns()]}}
-    | {error, {aborted, term()}}.
+    | {error, {aborted, term()}}
+    | {error, binary()}.
 bulk_import_configs(Entries) ->
     handle_bulk_import_configs(Entries).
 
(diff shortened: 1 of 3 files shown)


apps/emqx_mt/src/emqx_mt_config.erl
```diff
@@ -539,12 +546,68 @@ limiter_update_side_effects(Ns, {?changed, OldConfig, NewConfig}) ->
...
+    case Valid of
+        true -> ok;
+        false -> {error, invalid_ns_name_msg()}
+    end.
+
+invalid_ns_name_msg() ->
+    ~b"""
+    Invalid namespace name: only ASCII letters, digits and the characters
+    '.', '-' and '_' are allowed, with length from 1 to 255;
+    '.' and '..' are not valid names
```
savonarola, line 577: Same consideration as https://github.com/emqx/emqx/pull/18372#discussion_r3805072715

zmstone: dot is common, so we can’t drop it. 
then there is aslo the path of managed certs, there is uri (percent) encoding for ns there, but forgot that dot is not % encoded.

savonarola (approved):

PR #18378, opened 2026-08-18
fix(mgmt): keep namespaced managed certs within the namespace's own directory

zmstone:
Companion to #18372 (namespaced backups). Same path-containment class, in the managed certs feature.

Release version:
6.1.5, 6.2.3, 6.3.0, 6.4.0

Introduced in:
6.1.0 (managed certificate bundles)

## Summary

`emqx_managed_certs:base_dir/1` escapes the namespace with `uri_string:quote/1`, which percent-encodes path separators but not the special path components `.` and `..`. A namespace named `..` resolved `<data>/certs2/ns/..` to the shared `<data>/certs2` directory, so a namespaced caller could `list_bundles/1`, `list_managed_files/2`, `add_managed_files/3`, `delete_managed_file/3` and `delete_bundle/2` against the global administrator's (and other tenants') certificate bundles.

Every filesystem-touching entry point now calls `check_namespace/1`: a namespace whose escaped form is not a single path component distinct from `.` and `..` (this also covers the empty name) returns `{error, bad_namespace}`. The REST API maps that to 400; the multi-tenancy janitor treats it as nothing to clean. The namespace always comes from the authenticated caller or from config and is never recovered from the directory name, so the check does not need to be reversible.

This mirrors the backup-directory guard in #18372 and is needed independently of namespace-name validation, because implicitly created namespaces (`client_attrs.tns`) and namespaces created before validation existed are not checked at creation.

Most relevant file: `apps/emqx/src/emqx_managed_certs.erl`.

## PR Checklist

### DIFF
--- a/apps/emqx_management/test/emqx_mgmt_api_certs_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_api_certs_SUITE.erl
@@ -871,7 +871,7 @@ t_namespace_path_traversal_rejected(TCConfig) when is_list(TCConfig) ->
                 #{ns => Ns}
             )
         end,
-        [<<"..">>, <<".">>, <<>>]
+        [<<"..">>, <<".">>, <<>>, <<"../..">>, <<"./..">>]
     ),
     %% The global bundle survived every attempt with its content intact.
     ?assertMatch({200, [#{<<"name">> := Victim}]}, list_bundles_global()),
--- a/apps/emqx_management/test/emqx_mgmt_api_certs_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_api_certs_SUITE.erl
@@ -826,6 +826,59 @@ t_escape_namespace_in_dirs(TCConfig) when is_list(TCConfig) ->
     ok.
 
 -doc """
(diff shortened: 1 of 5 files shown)


apps/emqx_management/test/emqx_mgmt_api_certs_SUITE.erl
```diff
@@ -825,6 +825,59 @@ t_escape_namespace_in_dirs(TCConfig) when is_list(TCConfig) ->
...
+                emqx_managed_certs:delete_managed_file_v1(Ns, Victim, ?FILE_KIND_CA),
+                #{ns => Ns}
+            ),
+            ?assertEqual(
+                {error, bad_namespace},
+                emqx_managed_certs:delete_bundle_v1(Ns, Victim),
+                #{ns => Ns}
+            )
+        end,
+        [<<"..">>, <<".">>, <<>>]
```
thalesmg: to cover splits:

```suggestion
        [<<"..">>, <<".">>, <<>>, <<"../..">>, <<"./..">>]
```

thalesmg (dismissed):

thalesmg (approved):

Issue #344, opened 2015-10-15
Need to persist subscriptions?


erylee:

erylee:
The feature is supported in 0.14 release. 

etc/emqttd.config:

```
        %% PubSub and Router
        {pubsub, [
            %% Default should be scheduler numbers
            %% {pool_size, 8},

            %% Subscription: disc | ram | false
            {subscription, disc},

            ......
```

Issue #343, opened 2015-10-15
when will announce release version ?


kevinsui581:
when will announce release version ?now is beta version.

erylee:
The current release is stable enough. 1.0 will come soon.

PR #18387, opened 2026-08-19
feat(dashboard): add SCRAM challenge-response login

hjianbo:
Release version: 7.0

Introduced in: N/A

## Summary

Add an optional SCRAM-SHA-256 challenge-response login flow for the Dashboard JSON API.

The server-side policy defaults to `dashboard.password_login = both`, so the existing
`POST /api/v5/login` clients remain compatible. The embedded API Spec login page
uses SCRAM by default:

- `POST /api/v5/login/challenge`
- Browser-side PBKDF2-HMAC-SHA256, HMAC-SHA256, and SHA-256 proof derivation
- `POST /api/v5/login/verify`
- Server-signature verification before accepting the session token

The API specification explorer now uses the same SCRAM login flow. Session expiry
returns the user to the shared SCRAM login page instead of maintaining a second
legacy login implementation. The main Dashboard SPA frontend is not included in
this PR; its SCRAM migration is tracked in issue #344.

Browser-based SCRAM login requires HTTPS or another secure browser context because
the implementation uses Web Crypto. This is an operational requirement for browser
clients; the backend does not require EMQX itself to have an HTTPS listener. TLS may
terminate at a reverse proxy or load balancer. `http://localhost` and
`http://127.0.0.1` are normally treated as secure browser contexts.

During rolling upgrades, the browser falls back to the legacy password endpoint only
when the server-rendered page permits legacy fallback and the SCRAM challenge endpoint
returns a genuine `404 NOT_FOUND`. It does not fallback on authentication errors,
service errors, network errors, or missing Web Crypto. In `scram_only` mode the
legacy endpoint remains disabled, including when challenge returns 404.

Do not enable `scram_only` until all Dashboard clients, including the main SPA, and
all cluster nodes support SCRAM. Legacy password hashes are detected before enabling
`scram_only`, with an actionable migration message:

`emqx ctl admins passwd <username> <new-password>`

This PR covers the backend and embedded API-spec login portions of
[issue #343](https://github.com/emqx/emqx-dev-team-tasks/issues/343). The Dashboard
SPA follow-up is tracked in
[issue #344](https://github.com/emqx/emqx-dev-team-tasks/issues/344).

## PR Checklist

- [x] Schema changes are backward compatible or intentionally breaking (default remains `both`)

### DIFF
--- a/apps/emqx_dashboard/include/emqx_dashboard.hrl
+++ b/apps/emqx_dashboard/include/emqx_dashboard.hrl
@@ -67,6 +67,23 @@
     extra = #{} :: map()
 }).
 
+-record(emqx_dashboard_scram_challenge, {
+    id :: binary(),
+    username :: dashboard_username(),
+    client_nonce :: binary(),
+    server_nonce :: binary(),
+    salt :: binary(),
+    iterations :: pos_integer(),
+    credential_fingerprint :: binary() | undefined,
+    expires_at :: integer(),
+    extra = #{} :: map()
+}).
+
+-record(emqx_dashboard_scram_meta, {
+    key :: atom(),
+    value :: binary()
+}).
+
 -define(TAB_COLLECT, emqx_collect).
 
 -define(EMPTY_KEY(Key), ((Key == undefined) orelse (Key == <<>>))).
--- a/apps/emqx_dashboard/priv/api-spec-login.html
+++ b/apps/emqx_dashboard/priv/api-spec-login.html
@@ -1,0 +1,252 @@
+<!doctype html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <meta name="robots" content="noindex,nofollow">
+  <title>EMQX API - Sign in</title>
+  <style>
+    body{font:14px system-ui,sans-serif;background:#0e1116;color:#e6e6e6;display:flex;align-items:center;justify-content:center;min-height:100vh;margin:0}
+    form{background:#161b22;border:1px solid #2a313a;border-radius:8px;padding:24px;width:min(380px,calc(100vw - 32px))}
+    h1{margin:0 0 6px;font-size:16px}
+    p{margin:0 0 16px;color:#9aa4ae;font-size:12px;line-height:1.5}
+    label{display:block;font-size:12px;color:#9aa4ae;margin-bottom:4px}
+    input{width:100%;padding:8px 10px;margin-bottom:12px;border:1px solid #2a313a;border-radius:6px;background:#0e1116;color:#e6e6e6;font-size:13px;box-sizing:border-box}
+    button{width:100%;padding:9px 12px;border:0;border-radius:6px;background:#3b82f6;color:#fff;font-size:13px;cursor:pointer}
+    button:disabled{opacity:.6;cursor:wait}
+    .err{color:#ffb4b4;font-size:12px;margin-bottom:10px;min-height:14px;white-space:pre-wrap}
+    .warning{color:#ffd580}
+    .hidden{display:none}
+  </style>
+</head>
+<body>
+  <form id="login-form" autocomplete="off">
+    <h1>Sign in to view the API specification</h1>
+    <p id="login-hint">The browser login uses SCRAM-SHA-256. Use HTTPS or another secure browser context.</p>
+    <div class="err" id="login-error"></div>
+    <label for="username">Username</label>
+    <input id="username" name="username" autocomplete="username" required>
+    <label for="password">Password</label>
+    <input id="password" name="password" type="password" autocomplete="current-password" required>
+    <div id="mfa-field" class="hidden">
+      <label for="mfa-token">MFA token</label>
+      <input id="mfa-token" name="mfa_token" inputmode="numeric" autocomplete="one-time-code">
+    </div>
+    <button id="submit" type="submit">Sign in</button>
+  </form>
+  <script>
+    const legacyFallbackAllowed = __EMQX_LEGACY_FALLBACK_ALLOWED__;
+    const form = document.getElementById('login-form');
+    const usernameInput = document.getElementById('username');
+    const passwordInput = document.getElementById('password');
+    const mfaField = document.getElementById('mfa-field');
+    const mfaInput = document.getElementById('mfa-token');
+    const errorElement = document.getElementById('login-error');
+    const submitButton = document.getElementById('submit');
+
+    function updateSecureContextHint() {
+      if (!window.isSecureContext || !window.crypto || !window.crypto.subtle) {
+        const message = legacyFallbackAllowed
+          ? 'SCRAM login requires HTTPS or another secure browser context. An older-node compatibility fallback may be available.'
+          : 'SCRAM login requires HTTPS or another secure browser context. Legacy password login is disabled.';
+        document.getElementById('login-hint').textContent = message;
+        document.getElementById('login-hint').classList.add('warning');
+      }
+    }
+
+    updateSecureContextHint();
+
+    function utf8(value) {
+      return new TextEncoder().encode(value);
+    }
+
+    function base64ToBytes(value) {
+      const normalized = value.replace(/-/g, '+').replace(/_/g, '/');
+      const padded = normalized + '='.repeat((4 - normalized.length % 4) % 4);
+      const decoded = atob(padded);
+      return Uint8Array.from(decoded, (char) => char.charCodeAt(0));
+    }
+
+    function bytesToBase64(value) {
+      let binary = '';
+      for (const byte of value) binary += String.fromCharCode(byte);
+      return btoa(binary);
+    }
+
+    function bytesToBase64Url(value) {
+      return bytesToBase64(value).replace(/\+/g, '-').replace(/\//g, '_').replace(/=+$/g, '');
+    }
+
+    function randomNonce() {
+      return bytesToBase64Url(crypto.getRandomValues(new Uint8Array(24)));
+    }
+
+    function escapeScramUsername(username) {
+      return username.replace(/=/g, '=3D').replace(/,/g, '=2C');
+    }
+
+    async function sha256(value) {
+      return new Uint8Array(await crypto.subtle.digest('SHA-256', value));
+    }
+
+    async function hmac(key, value) {
+      const cryptoKey = await crypto.subtle.importKey(
+        'raw', key, { name: 'HMAC', hash: 'SHA-256' }, false, ['sign']
+      );
+      return new Uint8Array(await crypto.subtle.sign('HMAC', cryptoKey, value));
+    }
+
+    async function saltedPassword(password, salt, iterations) {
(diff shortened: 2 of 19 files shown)


zmstone:
aim dev-64 (dev-70)

apps/emqx_dashboard/src/emqx_dashboard_login.erl
```diff
@@ -0,0 +1,450 @@
...
+        [] ->
+            {error, not_found}
+    end.
+
+create_fake_challenge(_Username, _ClientNonce, ServerNonce) ->
+    %% Use the same challenge shape for unknown and legacy users to reduce
+    %% username/credential-format disclosure.  Fake challenges are stateless:
+    %% they never issue a token and therefore must not consume challenge-table
+    %% capacity.
+    Salt = crypto:strong_rand_bytes(16),
```
zmstone: Username-enumeration oracle: the fake-challenge salt is non-deterministic.

For a real user, `create_challenge` returns the stable salt from their stored verifier — identical across repeated challenge requests. Here the fake path returns `crypto:strong_rand_bytes(16)`, a fresh value every request. An attacker who requests two challenges for the same username and compares the returned `salt` (and, for migrated users, `iterations`) can distinguish "exists with a SCRAM verifier" (salt stable) from "unknown or legacy" (salt changes). That defeats the stated goal of this function ("reduce username/credential-format disclosure").

Suggest deriving a deterministic mock salt from the username under a server-side secret, e.g. `Salt = binary:part(hmac(ServerSecret, Username), 0, 16)`, and returning a fixed typical `iterations`, so a fake challenge is stable per username and indistinguishable from a real one. This is the standard SCRAM mock-salt approach (PostgreSQL `scram_mock_salt`, MongoDB). The secret can be a persisted per-cluster value; do not reuse the RPC/erlang cookie.

apps/emqx_dashboard/src/emqx_dashboard_login.erl
```diff
@@ -0,0 +1,450 @@
...
+            do_scram_challenge(Username, ClientNonce)
+    end.
+
+do_scram_challenge(Username, ClientNonce) ->
+    ServerNonce = nonce(),
+    case lookup_credential(Username) of
+        {ok, User, Verifier} ->
+            case emqx_dashboard_login_lock:verify(Username) of
+                {error, login_locked} ->
+                    {error, login_locked};
```
zmstone: Minor: returning `login_locked` at the challenge phase discloses account existence. A locked real user gets `{error, login_locked}`, while an unknown user gets a (fake) challenge — so an attacker who can drive an account into lockout can confirm it exists, and can tell locked-real from unknown. Consider returning a normal-shaped (fake) challenge here too and enforcing the lockout only at `scram_verify` time, where the failure is already generic. Low severity, but it's on the same enumeration-resistance seam as the salt note above.

apps/emqx_dashboard/src/emqx_dashboard_scram.erl
```diff
@@ -0,0 +1,130 @@
...
+        iterations => Iterations,
+        salt => Salt,
+        salted_password => SaltedPassword,
+        stored_key => crypto:hash(sha256, ClientKey),
+        server_key => hmac(SaltedPassword, <<"Server Key">>)
+    }.
+
+auth_message(Username, ClientNonce, ServerNonce, Salt, Iterations) ->
+    CombinedNonce = <<ClientNonce/binary, ServerNonce/binary>>,
+    ClientFirstBare = <<"n=", Username/binary, ",r=", ClientNonce/binary>>,
```
zmstone: Nit / frontend-coordination: the AuthMessage embeds the raw `Username` without RFC 5802 escaping (`,` -> `=2C`, `=` -> `=3D`) or SASLprep. It is internally consistent — server `verify` and the client both build the same fragment — so there is no server-side bypass. But if a dashboard username can contain `,` or `=`, the `n=<user>,r=<nonce>` boundary becomes ambiguous and must be escaped identically on the frontend (#344). Worth a comment here pinning the encoding contract so the two implementations cannot drift.

apps/emqx_dashboard/src/emqx_dashboard_schema.erl
```diff
@@ -67,6 +67,15 @@ fields("dashboard") ->
                     desc => ?DESC(password_expired_time)
                 }
             )},
+        {password_login,
+            ?HOCON(
+                hoconsc:enum([both, scram_only]),
+                #{
+                    default => both,
+                    desc => ?DESC(password_login),
+                    importance => ?IMPORTANCE_LOW
```
zmstone:
```suggestion
                    importance => ?IMPORTANCE_MEDIUM
```

apps/emqx_dashboard/src/emqx_dashboard_login.erl
zmstone: Username-enumeration oracle: the fake-challenge salt is non-deterministic.

For a real user, `create_challenge` returns the stable salt from their stored verifier — identical across repeated challenge requests. Here the fake path returns `crypto:strong_rand_bytes(16)`, a fresh value every request. An attacker who requests two challenges for the same username and compares the returned `salt` (and, for migrated users, `iterations`) can distinguish "exists with a SCRAM verifier" (salt stable) from "unknown or legacy" (salt changes). That defeats the stated goal of this function ("reduce username/credential-format disclosure").

Suggest deriving a deterministic mock salt from the username under a server-side secret, e.g. `Salt = binary:part(hmac(ServerSecret, Username), 0, 16)`, and returning a fixed typical `iterations`, so a fake challenge is stable per username and indistinguishable from a real one. This is the standard SCRAM mock-salt approach (PostgreSQL `scram_mock_salt`, MongoDB). The secret can be a persisted per-cluster value; do not reuse the RPC/erlang cookie.

apps/emqx_dashboard/src/emqx_dashboard_login.erl
zmstone: do not reveal locked state to attacker.

apps/emqx_dashboard/src/emqx_dashboard_scram.erl
zmstone: Nit / frontend-coordination: the AuthMessage embeds the raw `Username` without RFC 5802 escaping (`,` -> `=2C`, `=` -> `=3D`) or SASLprep. It is internally consistent — server `verify` and the client both build the same fragment — so there is no server-side bypass. But if a dashboard username can contain `,` or `=`, the `n=<user>,r=<nonce>` boundary becomes ambiguous and must be escaped identically on the frontend (https://github.com/emqx/emqx-dev-team-tasks/issues/344). Worth a comment here pinning the encoding contract so the two implementations cannot drift.

changes/ee/feat-18387.en.md
```diff
@@ -0,0 +1,4 @@
+Added optional SCRAM-SHA-256 challenge-response authentication for Dashboard
+JSON API login. The existing password login remains enabled by default through
+`dashboard.password_login = both`; set it to `scram_only` after migrating
```
zmstone: Consider noting the scripting-client impact here (this is user-facing behavior). Under `scram_only`, `POST /login` with username+password is rejected with `403 password_login_disabled` — the only password path left is the `/login/challenge` + `/login/verify` handshake, which a client must implement (client-side PBKDF2-HMAC-SHA256 at 600000 iterations + HMAC/XOR).

Automation should use API keys instead, which are a separate auth mechanism and are unaffected by `password_login = scram_only`. Worth one sentence steering operators: before enabling `scram_only`, migrate any script or third-party tool that logs in via `/login` (username/password) to an API key, or it will stop working.

rel/i18n/emqx_dashboard_api.hocon
```diff
@@ -42,6 +42,12 @@ login_failed_response400.desc:
 login_success.desc:
 """Dashboard Auth Success"""
 
+client_nonce.desc:
+"""SCRAM client nonce. Use 20 to 128 URL-safe characters: letters, digits, `_`, or `-`."""
```
zmstone: base62 ?

apps/emqx_dashboard/include/emqx_dashboard.hrl
```diff
@@ -67,6 +67,17 @@
...
 
+-record(emqx_dashboard_scram_challenge, {
+    id :: binary(),
+    username :: dashboard_username(),
+    client_nonce :: binary(),
+    server_nonce :: binary(),
+    salt :: binary(),
+    iterations :: pos_integer(),
+    credential_fingerprint :: binary() | undefined,
+    expires_at :: integer()
```
zmstone: add an extra field.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -1011,6 +1013,26 @@ admin_users() ->
     Admins = ets:select(?ADMIN, MS),
     lists:map(fun to_external_user/1, Admins).
 
+-spec scram_incompatible_users() -> [dashboard_username()].
+scram_incompatible_users() ->
```
zmstone: this check can be expensive.
maybe we do not check it. 
log `error` level if password is old format, but configured scram_only.

When password is not SCRAM-ready:
- if client calls `/login` directly (not scram), then log + response failure like a regular bad password.
- if client calls `/challenge`, then log + response failure message that it should change password first.

rel/i18n/emqx_dashboard_api.hocon
hjianbo: The nonce is generated as unpadded Base64URL on both the backend and the embedded API Spec client, using the URL-safe alphabet (A-Z, a-z, 0-9, -, _). Therefore, the description has been updated to explicitly say "unpadded Base64URL".

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
hjianbo: Removed the expensive startup and configuration preflight scan for incompatible users. When a legacy password hash is encountered during SCRAM challenge creation, the server logs an actionable migration message at error level when password_login is configured as scram_only. To avoid exposing username or credential-format information through the API, legacy and unknown users receive the same fake challenge response, and verification fails with the generic BAD_USERNAME_OR_PWD / Auth failed response.

apps/emqx_dashboard/priv/api-spec-login.html
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    async function scramProof(username, password, clientNonce, challenge) {
+      if (challenge.mechanism !== 'SCRAM-SHA-256') throw new Error('Unsupported SCRAM mechanism.');
+      const salt = base64ToBytes(challenge.salt);
+      const serverNonce = challenge.server_nonce;
+      const combinedNonce = clientNonce + serverNonce;
+      const iterations = challenge.iterations;
+      const authMessage = utf8(
+        'n=' + username + ',r=' + clientNonce +
```
zmstone: need escape for username.

zmstone (approved):

PR #18388, opened 2026-08-19
fix(rule_engine): mount republish topic under the rule's namespace

zmstone:
Release version:
6.1.5, 6.2.3, 6.3.0, 6.4.0

Introduced in: 6.1.0

## Summary

Internal issue: emqx/emqx-dev-team-tasks#350.

The Rule Engine `republish` action of a namespaced rule published to the raw rendered topic, while the rule's `SELECT` side is confined to the rule's namespace when `rule_engine.limit_selects_in_namespace` is enabled (the default). This change makes the republish side follow the same setting:

- `republish/3` reads the rule's namespace from the action metadata (`emqx_rule_runtime` adds `namespace` next to `rule_id`);
- when `rule_engine.limit_selects_in_namespace` is enabled and the rendered topic is not already under `<namespace>/`, the prefix is prepended with `emqx_mountpoint:mount/2`;
- a rendered topic already under `<namespace>/` is published unchanged, so republish templates that add the prefix themselves keep working;
- with the flag disabled, and for global rules, the rendered topic is published unchanged.

Notes:

- The mount is applied to the rendered topic before `safe_publish`, so the fix reaches 6.3/6.4 unchanged via the regular `dev-61` -> `dev-62` -> `dev-63` sync: the sync merge auto-applies cleanly (only the routine `mix.exs` version conflict). An earlier `dev-63` counterpart (#18389) was closed as redundant.
- The gate is the root `rule_engine.limit_selects_in_namespace` config; no zone config is consulted, the rule path has no client zone.

New tests in `emqx_rule_engine_SUITE` (all with `limit_selects_in_namespace = true`): a namespaced rule's republish is received under the namespace and not on the raw topic, the confinement holds with `namespace_as_mountpoint` disabled, and a template that already prefixes the namespace is not prefixed twice. Existing namespaced-rule suites (`t_namespaced_crud`, `t_namespaced_sources`, `t_namespaced_restart`, `t_limit_selects_in_namespace`, `t_fallback_actions_different_namespaces`) pass unchanged.

## PR Checklist


apps/emqx_rule_engine/src/emqx_rule_actions.erl
```diff
@@ -235,6 +235,21 @@ pre_process_args(Mod, Func, Args) ->
         false -> Args
     end.
 
+-doc """
+Mount the republish topic under the rule's namespace when
+`mqtt.namespace_as_mountpoint` is enabled, mirroring how a direct publish
+from a namespaced client is mounted.
+""".
+mount_rule_namespace(#{namespace := Namespace}, Topic) when is_binary(Namespace) ->
+    case emqx_config:get_zone_conf(default, [mqtt, namespace_as_mountpoint], false) of
```
zmstone: cannot hard-code zone.

thalesmg (approved):

PR #18392, opened 2026-08-19
fix(aggregated actions): separate different namespaces

thalesmg:
Fixes https://github.com/emqx/emqx-dev-team-tasks/issues/352

Release version:

## Summary

Makes aggregated actions with same type and name but different namespaces use distinct work dirs.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_aggregated_impl.erl
+++ b/apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_aggregated_impl.erl
@@ -750,7 +750,7 @@ start_aggregator(ConnResId, ActionResId, ActionConfig, ActionState0) ->
     #{http := HTTPClientConfig} = ActionState0,
     Type = ?ACTION_TYPE_AGGREG_BIN,
     AggregId = {Namespace, Type, Name},
-    WorkDir = work_dir(Namespace, Type, Name),
+    WorkDir = emqx_connector_aggregator:work_dir(Namespace, Type, Name),
     AggregOpts = #{
         max_records => MaxRecords,
         time_interval => TimeInterval,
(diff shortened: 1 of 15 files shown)


zmstone (approved):

apps/emqx_bridge_snowflake/src/emqx_bridge_snowflake_aggregated_impl.erl
```diff
@@ -748,8 +749,8 @@ start_aggregator(ConnResId, ActionResId, ActionConfig, ActionState0) ->
     } = ActionConfig,
     #{http := HTTPClientConfig} = ActionState0,
     Type = ?ACTION_TYPE_AGGREG_BIN,
-    AggregId = {Type, Name},
-    WorkDir = work_dir(Type, Name),
+    AggregId = {Namespace, Type, Name},
```
zmstone: Version bump likely missing for `emqx_bridge_snowflake`. This app's source changed here, but its `mix.exs` stays at `6.0.3`, while the sibling aggregated connectors (`emqx_bridge_s3`, `emqx_bridge_azure_blob_storage`, `emqx_bridge_s3tables`) were bumped `6.0.3 -> 6.0.4` in this PR. `./scripts/apps-version-check.exs` will likely require `emqx_bridge_snowflake -> 6.0.4` as well. (`emqx_bridge` and `emqx_connector_aggregator` are already at `6.0.4`, so they're probably fine — running the check confirms all three.)

apps/emqx_connector_aggregator/src/emqx_connector_aggregator.erl
```diff
@@ -102,6 +104,16 @@ is_empty(Name) ->
             true
     end.
 
+work_dir(?global_ns, Type, Name) ->
+    filename:join([emqx:data_dir(), bridge, Type, Name]);
+work_dir(Ns, Type, Name) when is_binary(Ns) ->
```
zmstone: Upgrade edge worth a deliberate call: for a **namespaced** action the work dir moves from the old shared `data/bridge/<Type>/<Name>` to the new `data/bridge/ns/<Ns>/<Type>/<Name>`. Any un-flushed buffer written under the old path before upgrade won't be found at the new path, so it's orphaned (never recovered or delivered) after upgrade. This is transient data and the old shared-dir behavior was the bug being fixed, so it's probably acceptable — but consider either noting it in the changelog/upgrade docs or adding a one-time cleanup of the orphaned pre-upgrade dirs. (Global actions keep the old path and are unaffected.)

PR #18393, opened 2026-08-19
feat: implement zerobus connector and action

thalesmg:
Release version:
7.0.0

## Summary

Ref: https://github.com/emqx/emqx/discussions/17920

## PR Checklist

- [x] The changes are covered with new or existing tests (also performed some manual testing, due to the lack of an emulator container)


apps/emqx_bridge_zerobus/test/emqx_bridge_zerobus_SUITE.erl
```diff
@@ -0,0 +1,2386 @@
...
+                }
+            }},
+            get_action_metrics_api(TCConfig)
+        )
+    ),
+    ok.
+
+%% -doc """
+%% When the schema used by an already-running channel changes, it needs to get the new
+%% descriptor proto and re-open the stream.
```
thalesmg, line 2487: since this will likely involve bigger changes (e.g.: running a hook to trigger finding the descriptor again) if we want to avoid finding it for every single batch, and this pr is already quite large, maybe it's best left for a follow up.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_impl.erl
```diff
@@ -0,0 +1,1089 @@
...
+        transport := #{
+            connect_timeout := ConnectTimeout,
+            pool_size := PoolSize
+        }
+    } = ConnConfig,
+    Opts = #{
+        url => ZerobusEndpoint,
+        grpc_opts => #{
+            pool_size => PoolSize,
+            gun_opts => #{tls_opts => emqx_tls_lib:to_client_opts(SSL)}
```
zmstone: TLS for the gRPC transport is not selected here. `grpc_client:do_connect` calls `gun:open(Host, Port, GunOpts)` and ignores the URL scheme; without a `transport` key gun uses `default_transport(Port)`, which is `tls` only for port 443.

Consequences:
- `zerobus_endpoint = "https://host:8443"` connects in plaintext; `tls_opts` are never used.
- `"http://host:443"` connects with TLS.
- `ssl.enable = false` with port 443 still uses TLS (gun default opts), so the `ssl` block has no effect on this transport either way.

The production endpoint `https://…:443` works by coincidence of the port. Suggest following `emqx_bridge_bigtable_client:gun_opts/2` (or `emqx_exhook_server:channel_opts/1`): derive `transport => ssl | tcp` from the scheme and pass `tls_opts` only for `https`. The suite does not reach this path (mock endpoint is `http://127.0.0.1:<port>`, `ssl.enable = false`).

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
```diff
@@ -0,0 +1,642 @@
...
+            ?tp("zerobus_writer_send_error", #{reason => Reason}),
+            emqx_resource:apply_reply_fun(
+                ReplyFnAndArgs, {error, {recoverable_error, stream_closed}}
+            ),
+            set_health(ActionResId, Idx, {?status_connecting, ~"reopening stream"}),
+            State2 = State1#{?stream := ?undefined},
+            {?reopen, State2};
+        {error, Reason} ->
+            ?tp("zerobus_writer_send_error", #{reason => Reason}),
+            emqx_resource:apply_reply_fun(ReplyFnAndArgs, {error, {unrecoverable_error, Reason}}),
```
zmstone: This catch-all replies `unrecoverable_error` for any send failure other than `closed`/`not_found`/`bad_stream`, and some of the remaining reasons are transient:

- `grpc_client` answers every call with `{error, Reason}` from `do_connect` when `gun_pid = undefined` (after the gun process exited), so `econnrefused` / `timeout` / `{down, _}` land here.
- `grpc_client:call/3` returns `{error, {deadline_exceeded, <<"Waiting for response timeout">>}}` when the client process does not answer within `request_ttl`. That happens while the client is inside `do_connect` → `gun:await_up` (up to 60 s) on behalf of the connector health check.

Sequence: server unreachable → gun gives up and exits → health check calls `grpc_client:health_check` → client blocks in `do_connect` → every batch the buffer worker sends meanwhile times out in the `send` call → replied unrecoverable → dropped, while the connector still shows `connected` because the health check has not returned yet.

The stream is reopened right after this reply, so replying `recoverable_error` for these reasons would make the buffer worker retry them on the new stream instead of dropping them. `t_ingest_grpc_send_errors` asserts `failed := 1` for `unexpected_mocked_error`; that expectation would need adjusting if the default flips.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
```diff
@@ -0,0 +1,642 @@
...
+    %% May be foreshadowed by a `CloseStreamSignal` message.
+    {error, {recoverable_error, Reason}};
+handle_insert_result({error, stream_closed = Reason}) ->
+    {error, {recoverable_error, Reason}};
+handle_insert_result({error, {Kind, Reason}}) when
+    Kind == unrecoverable_error;
+    Kind == recoverable_error
+->
+    {error, {Kind, Reason}};
+handle_insert_result({error, Reason}) ->
```
zmstone, line 671: When the server closes the stream with trailers, the receiver passes `{error, {Code, Msg}}` to `handle_errored/3`, which replies this result to every in-flight caller. Only `{unavailable, <<"Stream timeout reached." …>>}` is recoverable; every other status, including `unavailable` with a different message, `deadline_exceeded`, `resource_exhausted`, `aborted`, `cancelled`, `unknown`, and `stream_closed_without_any_response` (stream ended with no trailers), falls through to `{unrecoverable_error, _}`.

Scenario: 50 batches in flight, server restarts and closes the stream with `UNAVAILABLE "server shutting down"` → all 50 are dropped and counted as failed, although the writer immediately reopens the stream and could have retried them. Same for a load balancer returning `UNAVAILABLE` without that exact text.

Suggest classifying by status code: treat at least `unavailable`, `deadline_exceeded`, `resource_exhausted`, `aborted`, `cancelled`, and `stream_closed_without_any_response` as recoverable, and keep `invalid_argument`, `unauthenticated`, `permission_denied`, `not_found`, `failed_precondition` unrecoverable. The message-text match would then be unnecessary. `t_recover_stream_close_during_ingestion` only exercises the one matched message.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_impl.erl
```diff
@@ -0,0 +1,1089 @@
...
+                ?pool_size := PoolSize
+            }}
+    } = ConnState,
+    ok = register_oauth2(ConnResId, ChanResId, ChanConfig, ConnState),
+    Record =
+        case Record0 of
+            #{type := json} ->
+                ?json;
+            #{type := proto} ->
+                Proto = get_serde(Record0),
```
zmstone, line 740: Low severity. `register_oauth2/4` has already allocated `?oauth2(ChanResId)` and registered the channel with `emqx_connector_oauth2` when `get_serde/1` throws (schema missing, wrong type, message type not found). The throw is caught by `?SAFE_CALL` in `emqx_resource:call_add_channel`, not by the `else` branch below, so nothing unregisters it. The channel is then not in `installed_channels`, so `on_remove_channel` is a no-op when the action is deleted, and the registration lives until the connector stops.

Moving the `get_serde/1` call ahead of `register_oauth2/4` (or running it inside the `maybe` so the `else` cleanup applies) closes this. `t_serde_gone_before_create` & co. exercise the error but not the leftover registration.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
```diff
@@ -0,0 +1,642 @@
...
+        ?open_stream_retry_interval := OpenStreamRetryInterval
+    } = State0,
+    ?tp(warning, "zerobus_failed_to_open_stream", #{
+        action_res_id => ActionResId,
+        reason => Reason
+    }),
+    case Reason of
+        {?stream_closed, {unauthenticated, Details}} ->
+            Msg = append_error_detail(~"Invalid credentials", Details),
+            set_health(ActionResId, Idx, {?status_disconnected, {unhealthy_target, Msg}});
```
zmstone, line 372: Question, not a blocker. Once a writer reports `{unhealthy_target, _}` the manager stops health-checking the channel (`is_channel_apt_for_health_check/1` returns `false`), so a later successful open that sets the optvar to `connected` is never observed: the action stays `disconnected` until the user re-enables it. Meanwhile every writer keeps retrying the open every `grpc_open_stream_retry_interval` (8 workers × 5 s by default) for as long as the action exists.

If `unauthenticated` / `permission_denied` from Zerobus are always a configuration problem, this matches how other connectors use `unhealthy_target` and is fine. If the server can return `unauthenticated` transiently (token expiring between fetch and open), the action wedges without operator action. Is the former known to hold?

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
```diff
@@ -0,0 +1,642 @@
...
+        ?schema => Schema,
+        ?table => Table,
+        ?table_fqn => TableFQN
+    },
+    {ok, State, {continue, #open_stream{}}}.
+
+terminate(_Reason, State) ->
+    #{?action_res_id := ActionResId, ?pool := Pool, ?idx := Idx} = State,
+    gproc_pool:disconnect_worker(Pool, {Pool, Idx}),
+    unregister_stream(ActionResId, Idx),
```
zmstone, line 183: nit: neither `terminate/2` nor the reopen paths close the previous gRPC stream, so the server keeps it until its own idle timeout and `grpc_client` keeps the entry in its `streams` map until that remote close arrives. Each action restart/update leaves `pool_size` such streams behind for a few minutes. Bounded by the server timeout, so not a leak in the strict sense; noting it in case grpc-erl grows a cancel API.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
```diff
@@ -0,0 +1,372 @@
...
+do_recv_stream_once_async(Stream, Opts) ->
+    spawn_link(?MODULE, do_recv_stream_once_async1, [Stream, Opts]).
+
+-spec do_recv_stream_once_async1(_Stream, _Opts) -> no_return().
+do_recv_stream_once_async1(Stream, Opts) ->
+    Res = do_recv_stream_once(Stream, Opts),
+    exit({ok, Res}).
+
+abort_helper(#{?helper := Helper} = State0) when is_pid(Helper) ->
+    exit(Helper, kill),
```
zmstone, line 297: this causes the zerobus_receiver_helper_died warning log.
would be good to avoid warning on intended stops.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
```diff
@@ -0,0 +1,372 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_bridge_zerobus_async_receiver_worker).
```
zmstone, line 4: nit: moduledoc to document the purpose of this module would be nice.

zmstone (commented):
great work!
the lengthy comments are added by agents.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_impl.erl
```diff
@@ -0,0 +1,1089 @@
...
+        Body = emqx_utils_json:encode(Records0),
+        emqx_trace:rendered_action_template(ChanResId, #{body => Body}),
+        Request = {
+            Path,
+            [
+                {~"Authorization", <<"Bearer ", Token/binary>>},
+                {~"Content-Type", ~"application/json"}
+            ],
+            Body
+        },
```
zmstone: nit: worth extracting to a helper func.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
thalesmg: the line just below this cleans the down signal.  there is no such noise.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: there's no API in grpc-erl (grpc_client) to close/cancel a stream.

would need to see how to do so and introduce it there.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_impl.erl
thalesmg: a function that just assembles this small tuple, simple data?  :thinking:

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: keep retrying on any errors could make messages get forever stuck if request_ttl is infinity, if the error is not actually recoverable.  I don't know all possible errors that could occur here...

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: I don't know all the possible ways zerobus might behave, so I think staying conservative and dropping the messages might prevent a message from being forever stuck due to such newly discovered behaviors/returns, which can be improved afterwards.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_impl.erl
thalesmg: doesn't matter because it's registered.  once the action is removed or the connector stopped, it'll be deallocated.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: isn't that the goal of unhealthy target?

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
thalesmg: oh, never mind :melting_face: :see_no_evil:

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: https://github.com/emqx/grpc-erl/pull/41

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: anyways, changed to retry send errors, hopefully they are all reasonable

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: classified based on code anyways.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_impl.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_impl.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
thalesmg: 🏁

zmstone (approved):

PR #18403, opened 2026-08-20
fix(cli): read and write standard I/O as UTF-8 regardless of OS locale

### LINKED ISSUE #18398: [6.2.3]emqx ctl maptabs get emits invalid UTF-8 and JSON for Unicode values
## Problem

The command emqx ctl maptabs get name does not emit valid UTF-8 or JSON when a mapping table contains Chinese or special Unicode characters. Loading, Rule SQL lookup, and three-node replication work normally; the defect is in the CLI get output path.

## Environment

EMQX 6.2.3-rc.2, emqx_maptabs-0.1.1, three-node cluster.

## Reproduction

Create /tmp/maptab_utf8.json as UTF-8 with this content:

[{"key":"中文键","label":"温度°C / 压力","text":"中文、emoji 😀、引号 quoted、反斜杠 \\\\、换行\\n、符号 <>&/=?"}]

Run:

emqx ctl maptabs load /tmp/maptab_utf8.json
emqx ctl maptabs get maptab_utf8 > /tmp/maptab_get.out
python3 -c "open(\"/tmp/maptab_get.out\",\"rb\").read().decode(\"utf-8\")"

Load returns {"result":"ok"}, but strict UTF-8 decoding fails with: UnicodeDecodeError: utf-8 codec cannot decode byte 0xb0. The raw output also represents Chinese characters with Erlang-style escapes such as backslash-x-brace-4E2D-brace, which are not valid JSON escapes. The degree symbol is emitted as raw byte 0xB0.

For comparison, this Rule SQL query works and returns the original Unicode value:

SELECT maptab_lookup(maptab_utf8, 中文键, text) AS v FROM t/#

## Expected behavior

maptabs get should print the stored table as valid UTF-8 JSON parseable by standard JSON tools.

## Impact

CLI inspection, export, backup, and automation fail for tables containing Unicode or affected special characters.

Black-box evidence: TC-18353-024 in the EMQX QA repository.

zmstone:
Release version:
6.3.0, 6.4.0

Introduced in:

## Summary

Fixes #18398. All Erlang VMs started by the `emqx` script now read and write standard I/O as UTF-8, independent of the OS locale.

### Problem

OTP picks the standard I/O encoding from `LC_ALL`/`LC_CTYPE`/`LANG` at VM start. When none of them names a UTF-8 locale (non-interactive SSH, cron, `su -c`, minimal images, `LANG` unset), the device is `latin1`. `emqx ctl` commands run through nodetool and inherit its device, so `emqx_ctl:print/2` prints codepoints above 255 as `\x{4E2D}` escapes and U+00B0 as the raw byte `0xB0`. The node's own console output has the same problem.

### Change

Add `-kernel standard_io_encoding unicode` (OTP 26+) to every VM the script starts:

- `bin/emqx`: `call_nodetool` (`ctl`, `eval`, `hocon`, and so on), `relx_escript`, both `remsh` flavors, and `eval-ex`.
- `mix.exs` `copy_escript`: the `%%!` header injected into the packaged `bin/nodetool`. escript only honors the first `%%!` line, and this injected line sits above the one in `bin/nodetool`, so the flags in the repo file do not apply in a release. The repo file gets the flag too for direct `escript bin/nodetool` use.
- `apps/emqx/etc/vm.args.cloud` (next to `+pc unicode`), `rel/vm.args.eex`, `rel/remote.vm.args.eex`: the node itself and the Mix `remote.vm.args`.

`+pc unicode` is not enough: it only widens the printable-character heuristic for `~p`/`~tp`; the device encoding is controlled by `standard_io_encoding`.

### Verification

On a release build with `LANG`, `LC_ALL`, and `LC_CTYPE` unset:

| | before | after |
|---|---|---|
| `emqx eval 'io:getopts().'` (ctl path) | `latin1` | `unicode` |
| `emqx eval 'io:getopts(user).'` (node) | `latin1` | `unicode` |
| `emqx_ctl:print("~ts~n", [<<"温度°C 中文"/utf8>>])` | `\x{6E29}\x{5EA6}<0xB0>C ...` | correct UTF-8 bytes |
| `emqx ctl maptabs get` with the unfixed plugin 0.1.1 | invalid UTF-8 | valid UTF-8 JSON |

Checked with OTP 26, 27, and 28 that the kernel parameter is honored by `escript` and that stdout redirection to a pipe or file does not change the result.

## PR Checklist


mix.exs
```diff
@@ -997,6 +997,19 @@ defmodule EMQXUmbrella.MixProject do
...
+    ## reader, so once that is gone nothing in the release opens it.
+    [
+      [release.path, "bin", "emqx.bat"],
+      [release.version_path, "env.bat"],
+      [release.version_path, "elixir.bat"],
+      [release.version_path, "iex.bat"],
+      [release.version_path, "remote.vm.args"]
+    ]
+    |> Enum.map(&Path.join/1)
+    |> Enum.each(&File.rm_rf!/1)
```
thalesmg, line 1011: maybe this is not necessary now that the `rel/` files were removed

thalesmg (approved):

mix.exs
zmstone: mix generates them.

PR #18409, opened 2026-08-20
fix(cluster_link): connect through all addresses listed in server

### LINKED ISSUE #18406: [6.2.3]After filling multiple addresses in the server field for Cluster‑Linking, the resource cannot be updated or deleted, and it also blocks creation of new resources.
### What happened?

After filling multiple addresses in the server field for Cluster‑Linking, the resource cannot be updated or deleted, and it also blocks creation of new resources.

This issue also exists in 6.2.2.

Delete and update work correctly.

1. Create a Cluster‑Linking resource with multiple addresses filled in the server field; saving succeeds.Image
2. Editing the Cluster‑Linking fails.

Image

3. Deletion also fails.

Image

4. Creating new Cluster‑Linking resources also fails.

Image

_No response_

### EMQX version

6.2.3-rc.2

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

`{
    "code": "BAD_REQUEST",
    "message": "Update link failed",
    "path": "cluster.links",
    "kind": "validation_error",
    "expected_data_type": "array",
    "got": {
      "operation": "update",
      "clientid": "sdfsd",
      "enable": true,
      "max_inflight": 32,
      "name": "tets",
      "password": "******",
      "pool_size": 8,
      "retry_interval": "15s",
      "server": "10.42.4.62:1883,10.42.4.61:1883",
      "topics": [
        "sdfsdfs"
      ],
      "username": "sdfsd",
      "resource_opts": {
        "dispatch_strategy": "per_clientid",
        "health_check_interval": "15s",
        "health_check_interval_jitter": "0ms",
        "health_check_timeout": "60s",
        "inflight_window": 100,
        "max_buffer_bytes": "256MB",
        "request_ttl": "45s",
        "start_timeout": "5s",
        "worker_pool_size": 16
      },
      "ssl": {
        "ciphers": [],
        "depth": 10,
        "enable": false,
        "hibernate_after": "5s",
        "log_level": "notice",
        "middlebox_comp_mode": true,
        "reuse_sessions": false,
        "secure_renegotiate": true,
        "verify": "verify_peer",
        "versions": [
          "tlsv1.3",
          "tlsv1.2"
        ]
      },
      "tcp_opts": {
        "active_n": 10
      }
    }
  }`

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:
5.8.0

## Summary

Fixes #18406.

A cluster link `server` value with more than one address (for example `10.42.4.62:1883,10.42.4.61:1883`) passes the schema check but could not be used at runtime. After such a link was saved, no link could be created, updated or deleted.

### Root cause

- `emqx_cluster_link_schema` declares `server` with the plural `emqx_schema:servers_sc/2`, whose validator accepts a comma-separated list.
- `emqx_cluster_link_config:mk_emqtt_options/1` parsed it with the singular `emqx_schema:parse_server/2`, which throws `expecting_one_host_but_got: 2`.
- The throw crashed the route replication actor (`emqx_cluster_link_routerepl`) on every connect attempt. The restart loop escalated through the supervisors until the `emqx_cluster_link` application stopped, and `stop/1` removed the `cluster.links` config handler.
- With no handler, `emqx_config_handler` skips `pre_config_update/3` and hands the raw `{create, _}` / `{update, _}` tuple to the schema check. That is the reported `expected_data_type => array` error.

Reproduced on the unfixed branch: the node logs show 396 `expecting_one_host_but_got: 2` throws and 78 `reached_max_restart_intensity` reports before `create_link/1` returns the reported error.

### Changes

- emqtt is bumped from 1.15.4 to 1.16.0 ([emqx/emqtt#316](https://github.com/emqx/emqtt/pull/316)): it fixes the `hosts` try order (the list used to be tried in reverse) and adds `shuffle_hosts` (not used here).
- `mk_emqtt_options/1` parses every address with `parse_servers/2` and passes them as the emqtt `hosts` option. One connect attempt tries the addresses in the configured order until one accepts.
- `prefer_host/2` rotates the list per forwarding pool worker (`emqx_cluster_link_mqtt:connect/1`), so the connections of one link spread over the remote nodes while each of them can still fail over. The route replication client keeps the configured order.
- `emqx_cluster_link_app:start_configured_links/0` catches per link, so one link that fails to start cannot stop the application or unregister the config handler.
- `emqx_conf_cli:check_res/5` crashed with `config_not_found` when reporting any `cluster.links` load error, because it looked up `[<<"cluster.links">>]` as a config path. It now splits the dotted key.
- The `server` field description documents the list form and the failover behaviour.

A node that already stores a multi-address link (6.2.2, 6.2.3) starts connecting through it after the upgrade; no config change is needed.

### Tests

- `emqx_cluster_link_config_SUITE:t_multi_server`: a link with a dead address first and two live addresses on two remote nodes. The route replication bootstrap completes, the forwarding resource becomes `connected`, both remote nodes receive link connections, messages reach subscribers on both remote nodes, and the link keeps working after the addresses are reordered. Uses `emqx_cth_cluster:sync_routes/1` after subscribing, since route propagation inside the remote cluster is asynchronous.
- `emqx_cluster_link_config_SUITE:t_broken_link_crud`: a node boots with a stored link whose addresses are all unreachable; the application starts, the config handler stays registered, and the broken link can coexist with a new link, be updated, and be deleted.
- `emqx_cluster_link_api_SUITE:t_multi_server`: create and update through the REST API.
- `emqx_cluster_link_config_tests`: `hosts` order and per-worker rotation.

### Scope

The mismatch exists on every maintained branch since cluster linking was introduced (`dev-510` and every v6 line). This PR targets `dev-63` and reaches `dev-70`/`master` through the forward-sync chain.

## PR Checklist

### DIFF
--- a/apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
+++ b/apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
@@ -478,6 +478,173 @@ t_misconfigured_links(Config) ->
         10_000
     ).
 
+t_multi_server('init', Config) ->
+    NameA = fmt("~s_~s", [?FUNCTION_NAME, "a"]),
+    NameB = fmt("~s_~s", [?FUNCTION_NAME, "b"]),
+    ClusterA = emqx_cth_cluster:start(mk_cluster(1, NameA, 1, Config)),
+    ClusterB = emqx_cth_cluster:start(mk_cluster(2, NameB, 2, Config)),
+    ok = snabbkaffe:start_trace(),
+    [{cluster_a, ClusterA}, {cluster_b, ClusterB} | Config];
+t_multi_server('end', Config) ->
+    ok = snabbkaffe:stop(),
+    ok = emqx_common_test_helpers:call_janitor(),
+    ok = emqx_cth_cluster:stop(?config(cluster_a, Config)),
+    ok = emqx_cth_cluster:stop(?config(cluster_b, Config)).
+
+-doc """
+A link whose `server` lists several addresses uses all of them: a dead address
+is skipped in favour of a live one (the route replication client and the first
+forwarding client both start with the dead address), and the forwarding clients
+spread over the live addresses (the third forwarding client prefers the second
+node of the remote cluster).
+""".
+t_multi_server(Config) ->
+    ClusterA = [NodeA] = ?config(cluster_a, Config),
+    ClusterB = [NodeB1, NodeB2] = ?config(cluster_b, Config),
+    DeadPort = emqx_common_test_helpers:select_free_port(tcp),
+    PortB1 = emqx_cluster_link_cth:tcp_port(NodeB1, clink),
+    PortB2 = emqx_cluster_link_cth:tcp_port(NodeB2),
+    Server = fmt("localhost:~b,localhost:~b,localhost:~b", [DeadPort, PortB1, PortB2]),
+    LinkConfA = mk_link_conf_to(ClusterB, #{
+        <<"server">> => Server,
+        <<"topics">> => [<<"t/#">>],
+        <<"pool_size">> => 3
+    }),
+    LinkConfB = mk_link_conf_to(ClusterA, #{<<"topics">> => [<<"t/#">>]}),
+    ?assertMatch({ok, _}, update(api, NodeB1, [LinkConfB], Config)),
+    {{ok, _}, {ok, _}} = ?wait_async_action(
+        update(api, NodeA, [LinkConfA], Config),
+        #{?snk_kind := "cluster_link_routerepl_bootstrap_complete", ?snk_meta := #{node := NodeA}},
+        30_000
+    ),
+    #{<<"name">> := NameB} = LinkConfA,
+    ResId = emqx_cluster_link_mqtt:resource_id(NameB),
+    ?retry(
+        500,
+        40,
+        ?assertMatch({ok, _, #{status := connected}}, ?ON(NodeA, emqx_resource:get_instance(ResId)))
+    ),
+    %% Assert both cluster B nodes have at least one connection.
+    ?assertMatch([_ | _], ?ON(NodeB1, emqx_cm:all_client_ids())),
+    ?assertMatch([_ | _], ?ON(NodeB2, emqx_cm:all_client_ids())),
+    %% Cluster B's actors connect back once cluster A knows cluster B.
+    lists:foreach(
+        fun(NodeB) ->
+            ?assertMatch(
+                {ok, _},
+                ?block_until(
+                    #{
+                        ?snk_kind := "cluster_link_routerepl_bootstrap_complete",
+                        ?snk_meta := #{node := NodeB}
+                    },
+                    30_000
+                )
+            )
+        end,
+        ClusterB
+    ),
+    ClientA = start_client("t_multi_server_a", NodeA),
+    ClientB = start_client("t_multi_server_b", NodeB1),
+    on_exit(fun() -> catch emqtt:stop(ClientA) end),
+    on_exit(fun() -> catch emqtt:stop(ClientB) end),
+    {{ok, _, _}, {ok, _}} = ?wait_async_action(
+        emqtt:subscribe(ClientB, <<"t/multi">>, qos1),
+        #{?snk_kind := "cluster_link_route_sync_complete", ?snk_meta := #{node := NodeB1}},
+        10_000
+    ),
+    ok = emqx_cth_cluster:sync_routes(ClusterB),
+    {ok, _} = emqtt:publish(ClientA, <<"t/multi">>, <<"hello-1">>, qos1),
+    ?assertReceive(
+        {publish, #{topic := <<"t/multi">>, payload := <<"hello-1">>, client_pid := ClientB}},
+        5_000
+    ),
+    %% The link keeps working when the addresses are reordered.
+    Server1 = fmt("localhost:~b,localhost:~b", [PortB2, DeadPort]),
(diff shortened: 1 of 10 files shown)


apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
```diff
@@ -478,6 +478,176 @@ t_misconfigured_links(Config) ->
...
+    ),
+    #{<<"name">> := NameB} = LinkConfA,
+    ResId = emqx_cluster_link_mqtt:resource_id(NameB),
+    ?retry(
+        500,
+        40,
+        ?assertMatch({ok, _, #{status := connected}}, ?ON(NodeA, emqx_resource:get_instance(ResId)))
+    ),
+    %% Only link clients connect to the second node of cluster B.
+    ?assertMatch([_ | _], ?ON(NodeB2, emqx_cm:all_client_ids())),
```
thalesmg: not sure how this comment and assertion are connected;  seems like both B nodes should have at least one client.

apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
```diff
@@ -478,6 +478,176 @@ t_misconfigured_links(Config) ->
...
+    {{ok, _, _}, {ok, _}} = ?wait_async_action(
+        emqtt:subscribe(ClientB2, <<"t/multi">>, qos1),
+        #{?snk_kind := "cluster_link_route_sync_complete", ?snk_meta := #{node := NodeB2}},
+        10_000
+    ),
+    {ok, _} = emqtt:publish(ClientA, <<"t/multi">>, <<"hello-2">>, qos1),
+    ?assertReceive(
+        {publish, #{topic := <<"t/multi">>, payload := <<"hello-2">>, client_pid := ClientB2}},
+        5_000
+    ),
```
thalesmg: q: how do we know there's no "cross-node hop" (?) here?

apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
```diff
@@ -478,6 +478,176 @@ t_misconfigured_links(Config) ->
...
+    %% The subscriber on the first remote node depends on route replication
+    %% inside cluster B, which can lag under load; publish until it receives.
+    ?retry(
+        1_000,
+        15,
+        begin
+            {ok, _} = emqtt:publish(ClientA, <<"t/multi">>, <<"hello-3">>, qos1),
+            ?assertReceive(
+                {publish, #{
+                    topic := <<"t/multi">>, payload := <<"hello-3">>, client_pid := ClientB
```
thalesmg: since clientB is already subscribed to t/multi, shouldn't it also receive hello-2 from above (hence no need for another publish)?

thalesmg (dismissed):

apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
zmstone: because after the config update, cluster B is `[PortB2, DeadPort]`.
So all links from cluster A must go to node B2.

apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
zmstone: due the async route replication, the first publish may be lost.

apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
zmstone: fixed with assertMatch on both nodes, and updated comment.

apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
```diff
@@ -478,6 +478,177 @@ t_misconfigured_links(Config) ->
...
+    {ok, _} = emqtt:publish(ClientA, <<"t/multi">>, <<"hello-2">>, qos1),
+    ?assertReceive(
+        {publish, #{topic := <<"t/multi">>, payload := <<"hello-2">>, client_pid := ClientB2}},
+        5_000
+    ),
+    %% The subscriber on the first remote node depends on route replication
+    %% inside cluster B, which can lag under load; publish until it receives.
+    ?retry(
+        1_000,
+        15,
```
keynslug: There's `emqx_cth_cluster:sync_routes/1` now, perhaps call it after subscribing to simplify the rest of the testcase code? So it's reduced to asserting both clients get the `hello-2` message.

apps/emqx_cluster_link/test/emqx_cluster_link_config_SUITE.erl
```diff
@@ -478,6 +478,177 @@ t_misconfigured_links(Config) ->
...
+    ok = snabbkaffe:stop(),
+    ok = emqx_cth_cluster:stop(?config(cluster_a, Config)),
+    ok = emqx_cth_cluster:stop(?config(cluster_b, Config)).
+
+-doc """
+Simulates an upgrade of a node that already stores a link whose `server` lists
+several addresses, the first of them dead. The application starts, the link
+connects through the live address, and the link can still be updated and deleted.
+""".
+t_preexisting_multi_server_link(Config) ->
```
keynslug: Given the PR changes, I'd expect a testcase for "broken cluster link can be deleted" instead.

keynslug (approved):

PR #18423, opened 2026-08-20
fix(data_backup): skip global-only roots in namespaced imports

zmstone:
Release version:

Introduced in:
6.0.0

## Summary

Fixes emqx/emqx-dev-team-tasks#349.

A namespaced import dispatched every root found in `ns/<NS>/cluster.hocon` to the `emqx_config_backup` importers. Most importers ignore the namespace argument and write the global configuration, so a namespaced archive carrying a root outside the namespaced-config allow-list was applied globally.

`emqx_mgmt_data_backup:do_import_cluster_hocon/4` now drops every root that is not in `emqx_config:namespaced_config_allowed_roots/0` from a namespaced import before the schema check and the importer dispatch. The skipped roots are logged as a warning (`data_import_skipped_global_roots`) and printed on the CLI; the namespace-writable roots in the same archive are still imported. The import is not failed, so an archive produced by a node with a different allow-list still imports the roots this node can apply.

Test: `t_namespaced_import_skips_global_roots` in `emqx_mgmt_data_backup_SUITE`. It fails on the unfixed code.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_data_backup.erl
+++ b/apps/emqx_management/src/emqx_mgmt_data_backup.erl
@@ -67,6 +67,7 @@
 -include_lib("kernel/include/file.hrl").
 -include_lib("emqx/include/logger.hrl").
 -include_lib("emqx/include/emqx_config.hrl").
+-include_lib("snabbkaffe/include/snabbkaffe.hrl").
 
 -define(ROOT_BACKUP_DIR, "backup").
 %% Namespaced backups are isolated under `<root>/ns/<Namespace>/'.
@@ -1383,7 +1384,8 @@ do_import_cluster_hocon(Namespace, BackupDir, Filename, Opts) ->
             maybe
                 {ok, RawConf0} ?= hocon:files([Filename]),
                 RawConf1 = upgrade_raw_conf(emqx_conf:schema_module(), RawConf0),
-                {ok, RawConf} ?= validate_cluster_hocon(RawConf1),
+                RawConf2 = drop_global_only_roots(Namespace, RawConf1, Opts),
+                {ok, RawConf} ?= validate_cluster_hocon(RawConf2),
                 maybe_print(
                     "Importing cluster configuration for namespace ~s...~n", [Namespace], Opts
                 ),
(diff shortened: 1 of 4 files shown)


apps/emqx_management/test/emqx_mgmt_api_data_backup_SUITE.erl
```diff
@@ -768,6 +768,38 @@ t_namespaced_import_foreign_content_forbidden(Config) ->
...
     ok.
 
+-doc """
+A namespaced administrator must not be able to write global configuration
+through a backup import. When `ns/<NS>/cluster.hocon' carries a root that is
+not namespace-writable (`listeners' here), the import skips that root and the
+global configuration stays untouched, while the namespace-writable roots in
+the same archive are still imported into the namespace.
+""".
+t_namespaced_import_skips_global_roots(Config) ->
```
keynslug: Nit. Tests basically the same behavior as `emqx_mgmt_data_backup_SUITE:t_namespaced_import_skips_global_roots`? Maybe no need for this testcase, if there are no API-level-only observable effects?

apps/emqx_management/src/emqx_mgmt_data_backup.erl
```diff
@@ -1383,7 +1384,8 @@ do_import_cluster_hocon(Namespace, BackupDir, Filename, Opts) ->
             maybe
                 {ok, RawConf0} ?= hocon:files([Filename]),
                 RawConf1 = upgrade_raw_conf(emqx_conf:schema_module(), RawConf0),
-                {ok, RawConf} ?= validate_cluster_hocon(RawConf1),
+                RawConf2 = drop_global_only_roots(Namespace, RawConf1, Opts),
```
keynslug, line 1387: Q: Why not reject the whole import?

zmstone: good question.
there are two reasons to end up dropping some key.

1. in the future the namespace-only root keys changed, importing from a different version;
2. the .tar.gz is tampered, and a malicious namespaced admin wants to do harm;

for 1, drop makes sense.
2 is not possible when https://github.com/emqx/eip/pull/119 is in place, also, if one know tampered will be dropped, then they would not tamper it...

apps/emqx_management/test/emqx_mgmt_api_data_backup_SUITE.erl
zmstone: yeah, right.
this is a test from 1st iteration of the pr, later refactored to drop, ended up duplicated.

apps/emqx_management/src/emqx_mgmt_data_backup.erl
zmstone: done

keynslug (approved):

PR #18436, opened 2026-08-21
fix(nats): enforce account JWT expiry and revocations

hjianbo:
Release version: 6.2.4

Introduced in: 6.2.0

Fixes: emqx/emqx-dev-team-tasks#358

## Summary

Validate NATS account JWT lifecycle claims during JWT authentication.

- Enforce account JWT exp and nbf claims during connection authentication.
- Apply account-level nats.revocations to new user JWT authentication, including wildcard and normalized user NKeys.
- Validate resolver-preloaded account JWT structure, signatures, trust, and revocation data when gateway configuration is loaded or updated.
- Return account-specific expiry and revocation errors with the offending public revocation key for actionable diagnostics.
- Set the connection authentication deadline to the earliest user/account JWT expiration so existing connections are closed by the existing connection_expire_timer.
- Revocation updates do not cancel existing connections; existing sessions follow the expiration captured at authentication time.

## PR Checklist

- [x] Change log for changes visible by users has been added to changes/ee/fix-18436.en.md
- [x] Schema validation is intentionally stricter for malformed resolver-preloaded account JWT data; valid configurations are unchanged

### DIFF
--- a/apps/emqx_gateway_nats/src/emqx_nats_authn.erl
+++ b/apps/emqx_gateway_nats/src/emqx_nats_authn.erl
@@ -9,7 +9,8 @@
     is_auth_required/2,
     ensure_nkey_nonce/2,
     maybe_add_nkey_nonce/2,
-    authenticate/4
+    authenticate/4,
+    validate_jwt_config/1
 ]).
 
 -ifdef(TEST).
@@ -193,11 +194,11 @@ jwt_authenticate(JWT, NKey, Sig, ConnInfo, ClientInfo, Method) ->
     maybe
         {ok, JWTToken = #{claims := Claims}} ?= decode_jwt(JWT),
         JWTConfig = maps:get(config, Method, #{}),
-        ok ?= verify_jwt_signature_chain(JWTToken, JWTConfig),
+        {ok, AccountClaims} ?= verify_jwt_signature_chain(JWTToken, JWTConfig),
         ok ?= verify_jwt_claims_time(Claims),
         {ok, Username} ?= verify_jwt_nonce_signature(Claims, NKey, Sig, ConnInfo),
         JWTPerms = extract_jwt_permissions(Claims),
-        AuthExpireAt = jwt_claim_expire_at(Claims),
+        AuthExpireAt = jwt_claims_expire_at([Claims, AccountClaims]),
         {ok, ClientInfo#{
             username => Username,
             password => undefined,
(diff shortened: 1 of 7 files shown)


apps/emqx_gateway_nats/src/emqx_nats_authn.erl
```diff
@@ -500,6 +510,96 @@ verify_jwt_claims_time(Claims) ->
...
     end.
 
+verify_jwt_user_revocation(UserClaims, AccountClaims) ->
+    maybe
+        {ok, UserSubject} ?= jwt_claim_sub(UserClaims),
+        UserIssuedAt = map_get(UserClaims, <<"iat">>, undefined),
+        {ok, Revocations} ?= normalized_jwt_revocations(AccountClaims),
+        RevocationAt = max_revocation_time(
+            map_get_any(Revocations, [UserSubject], undefined),
+            map_get_any(Revocations, [<<"*">>, '*'], undefined)
```
zmstone: nit: 
```suggestion
            map_get_any(Revocations, [<<"*">>], undefined)
```

apps/emqx_gateway_nats/src/emqx_nats_authn.erl
```diff
@@ -300,23 +300,33 @@ verify_jwt_signature_chain(#{claims := Claims} = UserToken, Config) ->
...
             AccountJWT,
             Config
         )
     end.
 
-verify_jwt_account_chain(UserToken, UserIssuer, AccountPubKey, AccountJWT, Config) ->
+verify_jwt_account_chain(UserToken, UserClaims, UserIssuer, AccountPubKey, AccountJWT, Config) ->
+    maybe
+        {ok, AccountClaims} ?= verify_jwt_account_token(AccountPubKey, AccountJWT, Config),
+        ok ?= verify_jwt_claims_time(AccountClaims),
```
zmstone: Observability: account `exp`/`nbf` are checked here with the same `verify_jwt_claims_time/1` used for the user claims, so both return the same reasons (`jwt_expired` / `jwt_not_before`, lines ~507/521). After this change, a `jwt_auth_failed` log with `reason => jwt_expired` no longer tells the operator whether the **user** credential or the **account** JWT expired — and the remediation differs (reissue the user credential vs. have the operator republish the account JWT).

Suggest tagging the account-side result distinctly, e.g. return `account_jwt_expired` / `account_jwt_not_before` (or wrap as `{account, Reason}`) from the account chain so the log points at the right artifact.

apps/emqx_gateway_nats/src/emqx_nats_authn.erl
```diff
@@ -500,6 +510,96 @@ verify_jwt_claims_time(Claims) ->
...
+
+normalize_revocation_map(Revocations) ->
+    maps:fold(
+        fun
+            (Key, Value, {ok, Acc}) ->
+                case {normalize_revocation_key(Key), valid_revocation_time(Value)} of
+                    {{ok, NormalizedKey}, true} ->
+                        {ok, Acc#{NormalizedKey => Value}};
+                    _ ->
+                        {error, invalid_jwt_account}
```
zmstone: Two things here, both about diagnosing a malformed account JWT:

1. **`invalid_jwt_account` is heavily overloaded.** It is the reason for at least seven distinct failures — account decode/subject/operator-trust (lines ~345/353/361/363/371) plus all three revocation paths here (revocations not a map, a key that doesn't decode, a non-integer time). They all surface as the same `reason => invalid_jwt_account` in the `jwt_auth_failed` warning, and the offending revocation entry is not logged. An operator can't tell it was a revocation problem, let alone which entry. Suggest cause-specific reasons (`invalid_account_revocations`, `invalid_revocation_key`, `invalid_revocation_time`) and include the offending key — revocation keys are public NKeys, not secrets, so logging `#{revocation_key => Key}` is safe and directly actionable.

2. **One bad entry rejects the whole account.** A single malformed key/value fails `normalize_revocation_map`, so every user of that account is denied until the account JWT is fixed. That is correct fail-closed behavior for operator-signed data, but it is a config-level fault that currently surfaces as a stream of per-connection auth failures with different clientids. Consider validating the account JWT's revocation map when it is loaded into the resolver / on config change, so the operator sees one clear error at configuration time instead of discovering it lazily per connection.

zmstone (approved):

PR #18451, opened 2026-08-22
feat(packaging): install /etc/emqx/emqx.env and read it from the systemd unit

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:
N/A

## Summary

EMQX now reads boot-time environment variables from `etc/emqx.env`. The deb and rpm packages install it as `/etc/emqx/emqx.env`; the docker image carries it at `/opt/emqx/etc/emqx.env`; a tgz install has it at `etc/emqx.env`.

`EMQX_FEATURES` and `EMQX_SECURITY_PROFILE` are read before `emqx.conf` is parsed (`emqx_security_profile` decides schema defaults such as the default bind address), so they cannot be HOCON settings. Until now a systemd install had no discoverable place to set them: editing the unit is lost on upgrade, and `systemctl edit emqx` is folklore. Docker takes them as plain environment variables. This PR gives every install form the same file.

Changes:

- `apps/emqx_conf/etc/emqx.env` (new): both variables commented out at their current defaults in the `KEY="${KEY:-default}"` form (uncommenting a shipped line keeps a value already set in the environment; writing a literal value forces it), each with a comment on what it does and what changing it costs. The security profile comment enumerates the behaviours `hardened` flips, taken from `emqx_security_profile:policy/1` and `bin/emqx` (default cookie refusal), and states that `hardened` becomes the default in 7.0. It points at the docs page from emqx/emqx-docs#3768 rather than carrying a full policy table. `policy(authz_context)` is intentionally not listed: `emqx_authz_context:make/1` only honours it under `-ifdef(TEST)`, so it changes nothing in a release build.
- `bin/emqx`: source `$EMQX_ETC_DIR/emqx.env` (with `set -a`) right after `releases/emqx_vars`, on every command. The `EMQX_FEATURES` and `EMQX_SECURITY_PROFILE` handling moves below that point; nothing between the old and new position used either variable. The file is optional. Values in the file override the inherited environment, the same precedence systemd's `EnvironmentFile=` has, so the file can override the unit's `Environment=EMQX_DEFAULT_LOG_HANDLER=file`. Sourcing a file from `etc/` adds no new trust boundary: `releases/emqx_vars` is already sourced from a tree that `postinst` chowns to `emqx:emqx`.
- `mix.exs`: copy the file into the release `etc/`, same as `acl.conf`. The deb rules (`cp -R etc/*`) and the rpm spec (`%config(noreplace) %{_conf_dir}/*`) pick it up without changes.
- `deploy/packages/emqx.service` (the deb and rpm units are symlinks to it): a comment pointing at the file. No `EnvironmentFile=`, so there is one parser, and `emqx ctl` run from a shell sees the same variables (for example `EMQX_NODE__COOKIE`) as the service.
- `scripts/test/test_emqx_boot.py`: the shipped file must keep every assignment commented out (a shipped active value would override `-e` flags in docker); a profile set in the file is enforced; the file wins over the inherited environment; an invalid value in the file fails before boot.

Verified with the built deb (Ubuntu 20.04, systemd 249 in a privileged container) and an rpm built from the same release tarball (Rocky 9):

- `dpkg-deb -e` lists `/etc/emqx/emqx.env` in `conffiles`; rpm file flags are 17 (config + noreplace).
- Untouched file: node boots, `emqx_is_running ... security_profile: legacy`.
- `EMQX_SECURITY_PROFILE=hardened` with the default cookie: boot refused with `Cannot continue with default cookie and EMQX_SECURITY_PROFILE set to 'hardened'`. With `EMQX_NODE__COOKIE` added to the same file: `security_profile: hardened`.
- `EMQX_SECURITY_PROFILE=bogus`: boot fails with `Invalid security profile: bogus`.
- Upgrade to a version-bumped package (`dpkg -i`, `rpm -U`): the edited file keeps its md5 and the node boots `hardened` again.

## PR Checklist

Follow-up: #18471 collects `etc/emqx.env` (secrets obfuscated) in `node_dump`.

### DIFF
--- a/bin/emqx
+++ b/bin/emqx
@@ -191,12 +191,6 @@ if [ "${2:-}" = 'help' ]; then
     fi
 fi
 
-## starts only the slim apps list, and we flip code loading to interactive so
-## the skipped apps' .beam files don't sit resident in memory.
-if [ "${EMQX_FEATURES:-}" = 'ESSENTIAL' ]; then
-    export CODE_LOADING_MODE="${CODE_LOADING_MODE:-interactive}"
-fi
-
 ## IS_BOOT_COMMAND is set for later to inspect node name and cookie from hocon config (or env variable)
 case "${COMMAND}" in
     start|console|console_clean|foreground|check_config)
@@ -214,16 +208,6 @@ case "${COMMAND}" in
         IS_BOOT_COMMAND='no'
         ;;
 esac
-
-# Will be set to hardened since v7
-export EMQX_SECURITY_PROFILE="${EMQX_SECURITY_PROFILE:-legacy}"
-
-if [[ "$IS_BOOT_COMMAND" = 'yes' ]]; then
-    if [ "$EMQX_SECURITY_PROFILE" != 'legacy' ] && [ "$EMQX_SECURITY_PROFILE" != 'hardened' ]; then
-        logerr "Invalid security profile: $EMQX_SECURITY_PROFILE. Valid values are: legacy, hardened."
-        exit 1
-    fi
-fi
 
 RELUP_DIR="relup"
 BASE_RUNNER_ROOT_DIR="${BASE_RUNNER_ROOT_DIR:-$RUNNER_ROOT_DIR}"
@@ -248,6 +232,32 @@ export EMQX_ETC_DIR
 export REL_VSN
 export SCHEMA_MOD
 export IS_ENTERPRISE
+
+## Boot-time environment variables (EMQX_FEATURES, EMQX_SECURITY_PROFILE, ...).
+## The file is optional. Values in the file override the inherited environment.
+if [ -f "$EMQX_ETC_DIR/emqx.env" ]; then
+    set -a
+    # shellcheck disable=SC1090,SC1091
+    . "$EMQX_ETC_DIR/emqx.env"
+    set +a
+fi
+
+## starts only the slim apps list, and we flip code loading to interactive so
+## the skipped apps' .beam files don't sit resident in memory.
+if [ "${EMQX_FEATURES:-}" = 'ESSENTIAL' ]; then
+    export CODE_LOADING_MODE="${CODE_LOADING_MODE:-interactive}"
+fi
+
+# Will be set to hardened since v7
+export EMQX_SECURITY_PROFILE="${EMQX_SECURITY_PROFILE:-legacy}"
+
+if [[ "$IS_BOOT_COMMAND" = 'yes' ]]; then
+    if [ "$EMQX_SECURITY_PROFILE" != 'legacy' ] && [ "$EMQX_SECURITY_PROFILE" != 'hardened' ]; then
+        logerr "Invalid security profile: $EMQX_SECURITY_PROFILE. Valid values are: legacy, hardened."
+        exit 1
+    fi
+fi
+
 
 RUNNER_SCRIPT="$RUNNER_BIN_DIR/$REL_NAME"
 CODE_LOADING_MODE="${CODE_LOADING_MODE:-embedded}"
--- a/apps/emqx_conf/etc/emqx.env
+++ b/apps/emqx_conf/etc/emqx.env
@@ -1,15 +1,17 @@
 # EMQX boot-time environment variables.
 #
-# The emqx systemd service reads this file (EnvironmentFile=) before it starts
-# the node. EMQX reads these variables before it parses emqx.conf, so they
-# cannot be set in emqx.conf.
+# The emqx command (bin/emqx, /usr/bin/emqx on package installs) sources this
+# file on every invocation: service start, foreground start, `emqx ctl`, and
+# so on. EMQX reads these variables before it parses emqx.conf, so they cannot
+# be set in emqx.conf.
 #
+# Values in this file override variables inherited from the environment.
 # A package upgrade keeps your edits to this file.
-# Uncomment a line, set the value, then run: systemctl restart emqx
+# Uncomment a line, set the value, then restart the node.
 #
-# Syntax: one KEY=value per line. No `export`, no shell expansion.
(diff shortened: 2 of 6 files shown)


keynslug:
> and `systemctl edit emqx` is folklore

Hm, I believe this is pretty standard and easy to google way to do it.

Not a strong opinion, but I think adding one more configuration file to cover more-or-less-well-known systemd user story does not bring a lot of value.

zmstone:
> > and `systemctl edit emqx` is folklore
> 
> Hm, I believe this is pretty standard and easy to google way to do it.
> 
> Not a strong opinion, but I think adding one more configuration file to cover more-or-less-well-known systemd user story does not bring a lot of value.

The service file location is not so conventional, and usually hard to manage with IaC.
with a file in emqx’s etc dir, it’s a unified configuration experience no matter how emqx is installed.

keynslug:
> The service file location is not so conventional, and usually hard to manage with IaC.

FWIW systemd should always respect drop-in overrides located in `/etc/systemd/system/emqx.service.d/`, e.g. the usual `/etc/systemd/system/emqx.service.d/override.conf`. This is a convention outlined in the systemd manual. My brief research didn't find any distributions where systemd is patched and/or configured in a way that breaks this convention.

zmstone:
> > The service file location is not so conventional, and usually hard to manage with IaC.
> 
> FWIW systemd should always respect drop-in overrides located in `/etc/systemd/system/emqx.service.d/`, e.g. the usual `/etc/systemd/system/emqx.service.d/override.conf`. This is a convention outlined in the systemd manual. My brief research didn't find any distributions where systemd is patched and/or configured in a way that breaks this convention.

Right, i was not aware of the override.conf.
But there is also:

```
System Unit Search Path
/etc/systemd/system.control/*
/run/systemd/system.control/*
/run/systemd/transient/*
/run/systemd/generator.early/*
/etc/systemd/system/*
/etc/systemd/system.attached/*
/run/systemd/system/*
/run/systemd/system.attached/*
/run/systemd/generator/*
…
/usr/local/lib/systemd/system/*
/usr/lib/systemd/system/*
/run/systemd/generator.late/*
User Unit Search Path
~/.config/systemd/user.control/*
$XDG_RUNTIME_DIR/systemd/user.control/*
$XDG_RUNTIME_DIR/systemd/transient/*
$XDG_RUNTIME_DIR/systemd/generator.early/*
~/.config/systemd/user/*
$XDG_CONFIG_DIRS/systemd/user/*
/etc/systemd/user/*
$XDG_RUNTIME_DIR/systemd/user/*
/run/systemd/user/*
$XDG_RUNTIME_DIR/systemd/generator/*
$XDG_DATA_HOME/systemd/user/*
$XDG_DATA_DIRS/systemd/user/*
…
/usr/local/lib/systemd/user/*
/usr/lib/systemd/user/*
$XDG_RUNTIME_DIR/systemd/generator.late/*

```

keynslug:
>  But there is also: ...

Yeah. Systemd is very complex. I agree with this but that's another point. Note that all of these locations cover different use cases for different roles and modes of operation.

I was only arguing that there's a conventional, universally supported and more or less well known solution for the user story "I'm a system administrator and I want to configure this systemd service with few extra environment variables": there are `/etc/systemd/system/emqx.service.d/override.conf` and `systemctl edit emqx.service [--stdin]`. To me, this PR just duplicates existing functionality.

Again, not a strong opinion.

bin/emqx
```diff
@@ -249,6 +233,32 @@ export REL_VSN
...
+fi
+
+## starts only the slim apps list, and we flip code loading to interactive so
+## the skipped apps' .beam files don't sit resident in memory.
+if [ "${EMQX_FEATURES:-}" = 'ESSENTIAL' ]; then
+    export CODE_LOADING_MODE="${CODE_LOADING_MODE:-interactive}"
+fi
+
+# Will be set to hardened since v7
+export EMQX_SECURITY_PROFILE="${EMQX_SECURITY_PROFILE:-legacy}"
```
keynslug, line 252: Q: Btw, do we need to include them in the `node_dump` output? Or they are already included?

zmstone:
> > But there is also: ...
> 
> Yeah. Systemd is very complex. I agree with this but that's another point. Note that all of these locations cover different use cases for different roles and modes of operation.
> 
> I was only arguing that there's a conventional, universally supported and more or less well known solution for the user story "I'm a system administrator and I want to configure this systemd service with few extra environment variables": there are `/etc/systemd/system/emqx.service.d/override.conf` and `systemctl edit emqx.service [--stdin]`. To me, this PR just duplicates existing functionality.
> 
> Again, not a strong opinion.

Setting it in system unit file was actually the first iteration (lost after rebase).
Then when drafting the documentation, we'd need to document "if installed using rpm/deb, then do this", and "if running in docker or extracted tarball" do that.
-- motivated me to use a common way for both.

the doc pr: https://github.com/emqx/emqx-docs/pull/3768

keynslug (approved):

keynslug:
>  Then when drafting the documentation, we'd need to document "if installed using rpm/deb, then do this", and "if running in docker or extracted tarball" do that.

Perhaps the documentation should explain how to customize env variables in a chapter for each mode of installation? E.g. there's https://docs.emqx.com/en/emqx/latest/deploy/install-enterprise.html already anyway.

bin/emqx
zmstone: will send another pr

zmstone:
> > Then when drafting the documentation, we'd need to document "if installed using rpm/deb, then do this", and "if running in docker or extracted tarball" do that.
> 
> Perhaps the documentation should explain how to customize env variables in a chapter for each mode of installation? E.g. there's https://docs.emqx.com/en/emqx/latest/deploy/install-enterprise.html already anyway.

yes. will send a doc pr.

PR #18452, opened 2026-08-22
feat(mgmt): expose security_profile and feature_preset in node info

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:
N/A

## Summary

Add two boot-time node properties to the node info returned by `GET /nodes` and `GET /nodes/{node}`:

- `security_profile`: `legacy` | `hardened`
- `feature_preset`: `full` | `essential` | `custom`

Both come from per-node environment variables and are fixed until restart, so they belong with `edition` and `role` in `node_info` rather than in a new endpoint. Because `/nodes` returns one entry per node, a cluster with divergent settings is visible in the list view. During a rolling upgrade mixed profiles are expected transiently.

Changes:

- `apps/emqx_management/src/emqx_mgmt.erl`: `node_info/0` adds the two keys. Both are `persistent_term` reads (`emqx_security_profile:profile/0`, `emqx_machine_features:info/0`).
- `apps/emqx_management/src/emqx_mgmt_api_nodes.erl`: `fields(node_info)` adds the two fields as enums, so the API docs list the valid values.
- `rel/i18n/emqx_mgmt_api_nodes.hocon`: descriptions.

Design notes:

- Only the preset scalar is exposed. `GET /features` already returns the enabled/disabled arrays.
- Stopped nodes do not report the new keys, the same as `version` and `edition`. The fields are not marked required, so the stopped-node entry still matches the schema. The test validates this with `hocon_tconf:check_plain` against `fields(node_info)`.
- No new BPAPI version. `emqx_management_proto_v5:node_info/1` keeps its signature; an older node in a rolling upgrade answers without the new keys, which the schema tolerates.

The Dashboard indicator that reads these fields is a separate follow-up.

## PR Checklist

### DIFF
--- a/apps/emqx_management/test/emqx_mgmt_api_nodes_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_api_nodes_SUITE.erl
@@ -174,9 +174,7 @@ t_multiple_nodes_api(Config) ->
             get, #{bindings => #{node => Node1}}
         ]),
         ?assertMatch(#{node := Node1, security_profile := legacy, feature_preset := full}, Node11),
-        lists:foreach(fun assert_node_info_schema/1, NodesList),
         %% A stopped node never reports its profile or preset.
-        %% The entry must still validate against the schema.
         ok = emqx_cth_cluster:stop_node(Node2),
         {200, [RunningInfo, StoppedInfo]} = ?retry(
             _Interval = 200,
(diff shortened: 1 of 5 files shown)


qzhuyan (dismissed):

apps/emqx_management/test/emqx_mgmt_api_nodes_SUITE.erl
```diff
@@ -142,8 +173,62 @@ t_multiple_nodes_api(Config) ->
...
+    },
+    ?assertMatch(
+        #{<<"node_info">> := #{}},
+        hocon_tconf:check_plain(
+            Schema,
+            #{<<"node_info">> => Json},
+            %% same semantics as the swagger layer: only `required => true` is required
+            #{atom_key => false, required => false}
+        )
+    ).
```
keynslug: Nit. To me, making this kind of assertions too specific and lax makes the whole thing a bit pointless. I'd suggest to address this separately, perhaps by implementing and using validating API client in API-level app tests, and then work on the issues it finds.

keynslug (dismissed):

apps/emqx_management/test/emqx_mgmt_api_nodes_SUITE.erl
zmstone: dropped

keynslug (approved):

PR #18453, opened 2026-08-22
feat(alarm): raise security_profile_divergence when cluster nodes disagree

zmstone:
Release version:
6.3.0

Introduced in:
N/A

## Summary

Adds the `security_profile_divergence` alarm. A node running the `hardened` security profile raises it when another running node in the cluster runs `legacy`. The alarm message names the `legacy` nodes; the details keep the current `legacy_nodes` / `hardened_nodes` / `unknown_nodes` lists while the alarm is active. It clears on its own once no running peer runs `legacy` (peer restarted with `hardened`, left the cluster, or went down).

Design decisions:

* **Peer profiles are resolved through BPAPI announcements plus RPC, on every tick.** `emqx_security_profile_monitor` (new, in `emqx_sys_sup` right after `emqx_alarm`; `init/1` returns `ignore` on `legacy` nodes) resolves the profile of every running peer anew on each 30s tick, so a peer that restarts with another profile is picked up by the next tick. A peer that announces the `emqx_security_profile` v1 BPAPI is asked over RPC (new `emqx_security_profile_proto_v1`, `erpc:multicall` of `emqx_security_profile:profile/0`); a peer whose BPAPI announcement lacks it runs an older release without security profiles and counts as `legacy`; a peer with no announcements yet is still booting and stays `unknown` (retried next tick), as do peers whose RPC fails. `unknown` peers never raise the alarm.
* **Only `hardened` nodes raise it.** Alarms are node-local; raising on every node would produce N alarms for one condition. `hardened` nodes are the ones whose expectation a `legacy` peer undermines, so the alarm count tracks converged nodes. On `legacy` nodes the monitor does not start at all (`init/1` returns `ignore`); single nodes never raise it.
* **The message carries the node list** because `emqx_alarm`'s log action logs only `name` and `message`. The alarm is raised once and kept active while divergence persists; the details are updated as the legacy set changes, and it is not re-raised per step, so a rolling upgrade yields one activation per converged node rather than O(N²) log lines.
* **No grace period.** A stalled rollout is indistinguishable from a slow one; the message names a rolling upgrade as the expected benign cause.
* Security profile only — `EMQX_FEATURES` divergence is intentionally not covered (features degrade via BPAPI negotiation and feature topologies can be deliberate).

Test helper: `emqx_cth_cluster` nodespecs accept a per-node `env_vars` list (merged over the cluster-wide one) so a cluster can start with mixed profiles.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_security_profile_monitor.erl
+++ b/apps/emqx/src/emqx_security_profile_monitor.erl
@@ -1,0 +1,188 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+-module(emqx_security_profile_monitor).
+
+-moduledoc """
+Raises the `security_profile_divergence` alarm when nodes of one cluster run
+different security profiles.
+
+On every timer tick, the node resolves the security profile of every running
+peer in the cluster, so a peer that restarts with another profile is picked up
+by the next tick.
+
+A peer is resolved through its BPAPI announcement: a peer that announces
+`emqx_security_profile` support is asked for its profile over RPC, and a peer
+whose announcement lacks it runs a release without security profiles, which
+behaves as `legacy`. A peer that has not announced anything yet is still
+booting and stays `unknown` until a later tick, as does a peer whose RPC
+fails.
+
+The process only runs on nodes with the `hardened` profile; `init/1` returns
+`ignore` on `legacy` nodes. Hardened nodes name the running peers that run the
+`legacy` profile. The alarm clears on the node once no running peer runs
+`legacy`.
+""".
+
+-behaviour(gen_server).
+
+-include("logger.hrl").
+-include_lib("snabbkaffe/include/trace.hrl").
+
+-export([start_link/0]).
+
+-export([
+    init/1,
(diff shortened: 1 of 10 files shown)


apps/emqx/src/emqx_security_profile_monitor.erl
```diff
@@ -0,0 +1,197 @@
...
+%% The application environment key exists for tests.
+check_interval() ->
+    application:get_env(emqx, security_profile_check_interval, ?DEFAULT_CHECK_INTERVAL).
+
+-spec check(state()) -> state().
+check(#{profiles := Profiles0} = State) ->
+    Peers = lists:delete(node(), emqx:running_nodes()),
+    Known = maps:with(Peers, Profiles0),
+    Fetched = fetch_profiles(Peers -- maps:keys(Known)),
+    Profiles = maps:merge(Known, Fetched),
```
thalesmg: once a node name is consulted, it's never checked again.  so, if a node boots as hardened, is verified to be fine, and later reboots as legacy, it won't be checked here, since it's known.  is it intentional?  sounds more like all nodes should be constantly checked.

zmstone: good catch

qzhuyan (approved):

PR #18465, opened 2026-08-24
fix(bridges): implement context-aware SQL rendering

savonarola:
Release versions:
5.8.12

Introduced in:
pre-5.8

## Summary

We fix emqx/emqx-dev-team-tasks#306, a problem with SQL injections in templated SQL in bridges where db-side placeholders are not available.

To fix it, we introduce placeholder-aware SQL parsers. They parse a **limited but known** subset of each SQL dialect and safely render interpolated values according to their context.

The PR is somewhat large, but it mostly repeats the same approach four times.

**Motivation**

The existing "SQL templating" never seemed robust. Logically, there are two possible models for SQL templating:

- **A.** A template is first rendered as a string and then treated as SQL. In other words, interpolate first and interpret the SQL structure afterwards. This is unsafe because the rendered data can change the SQL structure.
- **B.** SQL is extended with special values (placeholders) in explicitly allowed contexts. The SQL structure is checked first, and placeholders are then rendered according to their parsed context.

The previous solution implemented neither model fully, although it was closer to the first. It performed some ad hoc parsing, such as checking the basic structure with regular expressions, but not enough parsing to determine the context of each placeholder.

There is no safe universal quotation working for any lexical and syntactic context. For example:

```sql
VALUES ('${a}', ${b})
```

`${a}` is content inside an existing string literal. Its value must be escaped without adding another pair of quotes. `${b}` is a complete SQL value. It must preserve numbers, booleans, and `NULL`, while adding and escaping quotes for strings.

There is no "simple" way to detect the context by inspecting the surroundings of a placeholder:

```sql
/* '${a}' */
/* ${a} */
'/* ${a} */'
```

The first two occurrences are inside comments and **must not** allow `*/`. The third occurrence is inside a string and must be rendered allowing `*/`. (In the new implementation we reject comments entirely). On the contrary, `'` must be forbidden (escaped) in the third variant but must be allowed without escaping in the first two. (N.B. we disallow comments entirely).

We now follow the *B* model, implementing placeholder-aware SQL parsers. The main guiding principle when choosing this way is that when security is concerned, we should **root out entire classes of vulnerabilities rather than address them on a case-by-case basis**. This follows the [Secure by Design](https://www.cisa.gov/news-events/news/categorically-unsafe-software) approach. We should not keep adding more ad hoc checks. We also should not handover to the user verification of query security.

The approach is not completely perfect. We still may receive some very tricky SQL that we misinterpret. However, the SQL is provided by an admin, not an attacker, so the chances are low. Additionally, even if we misinterpret some SQL, on render we still correctly produce some valid SQL in which the placeholders cannot escape.

As a result of the fix, we also:

- Remove [regular-expression parsing of `INSERT` statements](https://github.com/emqx/emqx/blob/4dea601c1c38196b1b7832fc160e5396262a2748/apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_connector.erl#L94), which was already a form of ad hoc SQL parsing.
- Remove SQL normalization that [possibly changed the values](https://github.com/emqx/emqx/blob/4dea601c1c38196b1b7832fc160e5396262a2748/apps/emqx_bridge_tdengine/src/emqx_bridge_tdengine_connector.erl#L450), including whitespace inside quoted strings.
- Do not reuse the common quoting functions from `emqx_template_sql`. Combining backend-specific quoting rules behind switches in a universal quotation function is more error-prone than keeping quotation local to each backend.

**Resulting inprovements** 

Besides fixing security.
- We are able to much better validate SQL on creation, not on actual requests.
- We now correctly and safely support interpolation inside string literals, for example:
```sql
VALUES ('xxx ${yyy} zzz')
```
- With the approach we are able to safely interpolate integers, booleans, nulls and any types with special syntax.

**Incompatibilities**

* This is a compatibility-breaking change because it significantly narrows the accepted SQL syntax. Templates using unknown or unsupported syntax are now rejected. The supported subsets cover common `INSERT ... VALUES` templates, but they do not cover every valid construction in each backend dialect. Hovewer, we expect most of the reasonable cases to be accepted.
* For MySQL correct quotation is particularly hard ([emqx code](https://github.com/emqx/emqx-dev-team-tasks/issues/306#issuecomment-5399584274), [1](https://nvd.nist.gov/vuln/detail/CVE-2026-44172), [2](https://nvd.nist.gov/vuln/detail/CVE-2026-48188), [3](https://nvd.nist.gov/vuln/detail/CVE-2026-33468), [4](https://nvd.nist.gov/vuln/detail/CVE-2026-33442), [5](https://nvd.nist.gov/vuln/detail/CVE-2019-10748), [6](https://nvd.nist.gov/vuln/detail/CVE-2006-2753)). See also  [2025 PDO parser research](https://www.slcyber.io/research/a-novel-technique-for-sql-injection-in-pdos-prepared-statements). MySQL has session/server setting that make interpret `'single quoted'` and `"double quoted"` strings differently. For ensuring seecure interpolation, we explicitly disable `ANSI_QUOTES` and `NO_BACKSLASH_ESCAPE` features in each connection (as by default server settings), and treat string tokens accordingly. This may break some obscure clients who have `NO_BACKSLASH_ESCAPE/ANSI_QUOTES` enabled in their server settings and who composed their templates relying on that.

**Rejected alternatives**

- Trying to limit the fix to tracing only the lexical context (whether we are inside a string, comment, or identifier), but it looked like one more ad hoc addition. Also, the chances of failing to detect some unknown syntax are high. With the current implementation, we reject any unknown syntax. 
- Introducing explicit placeholders like `'xxx $sq_str{a.b.c} yyy'`, `"xxx $dq_str{a.b.c} yyy"`, etc. This approach is not foolproof by default and shifts responsibility to the user.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_sql.erl
+++ b/apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_sql.erl
@@ -1,0 +1,924 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+%% @doc Compile and render restricted ClickHouse INSERT templates.
+-module(emqx_bridge_clickhouse_sql).
+
+-export([compile/1, render/3, render_batch/3, parse_placeholder/1]).
+-export_type([plan/0]).
+
+-type placeholder() :: emqx_template:placeholder().
+
+-record(tpl_text, {text :: binary()}).
+-record(tpl_placeholder, {placeholder :: placeholder()}).
+-record(part_text, {text :: binary()}).
+-record(part_value, {value :: term()}).
+
+-type template_part() :: #tpl_text{} | #tpl_placeholder{}.
+-type part() :: #part_text{} | #part_value{}.
+
+-record(raw, {sql :: binary()}).
+-record(sql_value, {placeholder :: placeholder()}).
+-record(json_value, {placeholder :: placeholder()}).
+-record(sql_string, {parts :: [template_part()]}).
+-record(json_string, {parts :: [template_part()]}).
+
+-type sql_render_op() :: #raw{} | #sql_value{} | #sql_string{}.
+-type json_render_op() :: #raw{} | #json_value{} | #json_string{}.
+-type sql_render_plan() :: [sql_render_op()].
+-type json_render_plan() :: [json_render_op()].
+
+-record(clickhouse_values_plan, {
+    insert_prefix :: binary(),
+    plan :: sql_render_plan()
+}).
+-record(clickhouse_json_plan, {
+    insert_prefix :: binary(),
+    plan :: json_render_plan()
+}).
+
+-opaque plan() :: #clickhouse_values_plan{} | #clickhouse_json_plan{}.
+
+-define(VALUES_BATCH_SEPARATOR, <<", ">>).
+-define(JSON_BATCH_SEPARATOR, <<>>).
+
+%%------------------------------------------------------------------------------
+%% API
+%%------------------------------------------------------------------------------
+
+-spec compile(unicode:chardata()) -> {ok, plan()} | {error, term()}.
+compile(SQL0) ->
+    SQL = unicode:characters_to_binary(SQL0),
+    try
+        case emqx_bridge_clickhouse_sql_lexer:string(binary_to_list(SQL)) of
+            {ok, Tokens, _EndLine} ->
+                case emqx_bridge_clickhouse_sql_parser:parse(Tokens) of
+                    {ok, AST} -> compile_ast(AST);
+                    {error, Reason} -> {error, {invalid_clickhouse_insert_template, Reason}}
+                end;
+            {error, Reason, _EndLine} ->
+                {error, {invalid_clickhouse_insert_template, Reason}}
+        end
+    catch
+        Class:CatchReason -> {error, {invalid_clickhouse_insert_template, {Class, CatchReason}}}
+    end.
+
+-spec render(plan(), map(), map()) -> {ok, iolist()} | {error, term()}.
+render(#clickhouse_values_plan{insert_prefix = Prefix, plan = Plan}, Data, Opts) ->
+    render_insert(Prefix, Plan, Data, Opts);
+render(#clickhouse_json_plan{insert_prefix = Prefix, plan = Plan}, Data, Opts) ->
+    render_insert(Prefix, Plan, Data, Opts).
+
+-spec render_batch(plan(), [map()], map()) -> {ok, iolist()} | {error, term()}.
+render_batch(
+    #clickhouse_values_plan{insert_prefix = Prefix, plan = Plan}, DataList, Opts
+) ->
+    render_batch_insert(Prefix, Plan, DataList, Opts, ?VALUES_BATCH_SEPARATOR);
+render_batch(
+    #clickhouse_json_plan{insert_prefix = Prefix, plan = Plan}, DataList, Opts
+) ->
+    render_batch_insert(Prefix, Plan, DataList, Opts, ?JSON_BATCH_SEPARATOR).
+
+-spec parse_placeholder(binary()) -> {ok, placeholder()} | {error, invalid_placeholder}.
+parse_placeholder(Source) ->
+    case valid_placeholder_source(Source) of
+        true ->
+            case emqx_template:parse(Source) of
+                [{var, _, _} = Placeholder] -> {ok, Placeholder};
+                _ -> {error, invalid_placeholder}
+            end;
+        false ->
+            {error, invalid_placeholder}
+    end.
+
+%%------------------------------------------------------------------------------
+%% Private funs
+%%------------------------------------------------------------------------------
+
+render_insert(Prefix, Template, Data, Opts) ->
+    case render_unit(Template, Data, Opts) of
+        {ok, Rendered} -> {ok, [Prefix, Rendered]};
+        {error, _} = Error -> Error
+    end.
+
+render_batch_insert(Prefix, Plan, DataList, Opts, Separator) ->
+    case render_batch_units(DataList, Plan, Opts, _Index = 1, _Acc = [], Separator) of
+        {ok, Rendered} -> {ok, [Prefix, Rendered]};
+        {error, _} = Error -> Error
+    end.
+
+render_batch_units([Data | Rest], Unit, Opts, Index, Acc, Separator) ->
+    case render_unit(Unit, Data, Opts) of
+        {ok, Rendered} ->
+            render_batch_units(Rest, Unit, Opts, Index + 1, [Rendered | Acc], Separator);
+        {error, Reason} ->
+            {error, {clickhouse_template_render_failed, #{batch_index => Index, reason => Reason}}}
+    end;
+render_batch_units([], _Unit, _Opts, _Index, Acc, Separator) ->
+    {ok, lists:join(Separator, lists:reverse(Acc))}.
+
+render_unit(Unit, Data, Opts) ->
+    render_plan(Unit, Data, Opts, #{}, []).
+
+render_plan([#raw{sql = SQL} | Rest], Data, Opts, Cache, Acc) ->
+    render_plan(Rest, Data, Opts, Cache, [SQL | Acc]);
+render_plan([#sql_value{placeholder = Placeholder} | Rest], Data, Opts, Cache0, Acc) ->
+    render_value_op(
+        Placeholder,
+        fun(Value) -> encode_value(Value, Opts) end,
+        Rest,
+        Data,
+        Opts,
+        Cache0,
+        Acc
+    );
+render_plan([#json_value{placeholder = Placeholder} | Rest], Data, Opts, Cache0, Acc) ->
+    render_value_op(
+        Placeholder,
+        fun(Value) -> encode_json_value(Value, Opts) end,
+        Rest,
+        Data,
+        Opts,
+        Cache0,
+        Acc
+    );
+render_plan([#sql_string{parts = Parts} | Rest], Data, Opts, Cache0, Acc) ->
+    render_template_op(
+        Parts,
+        fun(Resolved) -> encode_sql_string(render_text_parts(Resolved, Opts)) end,
+        Rest,
+        Data,
+        Opts,
+        Cache0,
+        Acc
+    );
+render_plan([#json_string{parts = Parts} | Rest], Data, Opts, Cache0, Acc) ->
+    render_template_op(
+        Parts,
+        fun(Resolved) -> emqx_utils_json:encode(render_text_parts(Resolved, Opts)) end,
+        Rest,
+        Data,
+        Opts,
+        Cache0,
+        Acc
+    );
+render_plan([], _Data, _Opts, _Cache, Acc) ->
+    {ok, lists:reverse(Acc)}.
+
+render_value_op(Placeholder, Encoder, Rest, Data, Opts, Cache0, Acc) ->
+    case resolve_placeholder(Placeholder, Data, Cache0) of
+        {ok, Value, Cache} ->
+            case encode_result(fun() -> Encoder(Value) end) of
+                {ok, Encoded} -> render_plan(Rest, Data, Opts, Cache, [Encoded | Acc]);
+                {error, Reason} -> {error, render_error(Placeholder, Reason)}
+            end;
+        {error, Reason} ->
+            {error, render_error(Placeholder, Reason)}
+    end.
+
+render_template_op(Parts, Encoder, Rest, Data, Opts, Cache0, Acc) ->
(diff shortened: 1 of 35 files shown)


rel/i18n/emqx_bridge_clickhouse.hocon
```diff
@@ -51,7 +51,7 @@ local_topic.label:
 """Local Topic"""
 
 sql_template.desc:
-"""The template string can contain ${field} placeholders for message metadata and payload field. Make sure that the inserted values are formatted and escaped correctly. [Prepared Statement](https://docs.emqx.com/en/enterprise/v5.0/data-integration/data-bridges.html#Prepared-Statement) is not supported."""
+"""A restricted ClickHouse INSERT template. It supports VALUES, FORMAT Values, and FORMAT JSONCompactEachRow. Placeholders are allowed only in value positions and string values. Comments, extra statements, and dynamic identifiers are not allowed."""
```
zmstone: quote so no need to change dict.
```suggestion
"""A restricted ClickHouse `INSERT` template. It supports `VALUES`, `FORMAT Values`, and ` JSONCompactEachRow`. Placeholders are allowed only in value positions and string values. Comments, extra statements, and dynamic identifiers are not allowed."""
```

apps/emqx_bridge_clickhouse/test/emqx_bridge_clickhouse_SUITE.erl
```diff
@@ -392,6 +259,74 @@ t_send_message_query(Config) ->
...
+        {0, Attack},
+        {1, <<16#FF>>},
+        {2, <<"a", 0, "b">>},
+        {3, <<"你好😀"/utf8>>}
+    ],
+    lists:foreach(
+        fun({EnableBatch, BaseKey}) ->
+            BridgeID = make_bridge(#{
+                enable_batch => EnableBatch,
+                batch_size => 1
```
thalesmg: what actually makes the action use batching or not is `batch_size /= 1`; so, set it to > 1 to actually enable batching.  enable_batch is long deprecated 😅

apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_sql.erl
```diff
@@ -0,0 +1,1134 @@
...
+%% Retain `${}` and `${.}`.
+%% See emqx_template:parse/1 in apps/emqx_utils/src/emqx_template.erl.
+valid_placeholder_source(<<"${}">>) ->
+    true;
+valid_placeholder_source(<<"${.}">>) ->
+    true;
+valid_placeholder_source(Source) ->
+    re:run(
+        Source,
+        <<"^\\$\\{\\.?[A-Za-z0-9_]+(?:\\.[A-Za-z0-9_]+)*\\}$">>,
```
thalesmg, line 637: Q: since we already use a regex to lex the placeholders, shouldn't we just adjust the regex from the lexer to require non-empty paths and then just use its output?

savonarola: We lex only standalone (value-level placeholders), but placeholders may occur inside e.g. sql/json strings: `VALUES ('x ${y} z')`. We lex the whole string then and still have to parse placeholders from it.

apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_sql_lexer.xrl
```diff
@@ -0,0 +1,116 @@
...
+ID_START        = [A-Za-z_\$]
+ID_CONTINUE     = [A-Za-z0-9_\$]
+IDENTIFIER      = {ID_START}{ID_CONTINUE}*
+%% emqx_template placeholder envelope:
+%% See emqx_template:parse/1 in apps/emqx_utils/src/emqx_template.erl.
+PLACEHOLDER     = \$\{[A-Za-z0-9_.]*\}
+%% ClickHouse decimal token scanning:
+%% https://github.com/ClickHouse/ClickHouse/blob/8dfb1700858195fa704221e360fa0798ac6ee9ed/src/Parsers/Lexer.cpp#L120-L223
+%% https://github.com/ClickHouse/ClickHouse/blob/8dfb1700858195fa704221e360fa0798ac6ee9ed/src/Parsers/Lexer.cpp#L250-L287
+NUMBER          = (([0-9]+(\.[0-9]*)?)|(\.[0-9]+))([eE][+-]?[0-9]+)?
```
thalesmg: I noticed that clickhouse [supports](https://clickhouse.com/docs/reference/syntax#numeric) `_` number separators and binary/hexadecimal literals, which are seemingly not handled here.  
Is it intentional?  If so, might add a note in the docstring?

apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_sql.erl
```diff
@@ -0,0 +1,1134 @@
...
+    ok = assert_no_identifier_placeholder(Source),
+    decode_quoted_identifier(strip_quotes(Source), $", []);
+decode_identifier(Source, backtick) ->
+    ok = assert_no_identifier_placeholder(Source),
+    decode_quoted_identifier(strip_quotes(Source), $`, []).
+
+assert_no_identifier_placeholder(Source) ->
+    case binary:match(Source, <<"${">>) of
+        nomatch -> ok;
+        _ -> error(dynamic_identifier_not_allowed)
```
thalesmg, line 339: Q: why don't we allow placeholders inside double-quotes/backticks?

thalesmg (commented):
still digesting; awesome work

apps/emqx_bridge_sqlserver/src/emqx_bridge_sqlserver_sql.erl
```diff
@@ -0,0 +1,741 @@
...
+    encode_string(<<"undefined">>, varchar);
+encode_value(undefined, _Opts) ->
+    <<"NULL">>;
+encode_value(null, _Opts) ->
+    <<"NULL">>;
+encode_value(true, _Opts) ->
+    encode_string(<<"true">>, varchar);
+encode_value(false, _Opts) ->
+    encode_string(<<"false">>, varchar);
+encode_value(Value, _Opts) when is_integer(Value) -> integer_to_binary(Value);
```
thalesmg: nano nit:

```suggestion
encode_value(Value, _Opts) when is_integer(Value) -> 
    integer_to_binary(Value);
```

apps/emqx_bridge_tdengine/src/emqx_bridge_tdengine_sql.erl
```diff
@@ -0,0 +1,895 @@
...
+    encode_string(<<"undefined">>);
+encode_value(undefined, _Opts) ->
+    <<"NULL">>;
+encode_value(null, _Opts) ->
+    <<"NULL">>;
+encode_value(true, _Opts) ->
+    <<"true">>;
+encode_value(false, _Opts) ->
+    <<"false">>;
+encode_value(Value, _Opts) when is_integer(Value) -> integer_to_binary(Value);
```
thalesmg: nano nit:


```suggestion
encode_value(Value, _Opts) when is_integer(Value) -> 
    integer_to_binary(Value);
```

apps/emqx_bridge_tdengine/src/emqx_bridge_tdengine_sql_lexer.xrl
```diff
@@ -0,0 +1,121 @@
...
+ID_CONTINUE     = [A-Za-z0-9_]
+IDENTIFIER      = {ID_START}{ID_CONTINUE}*
+%% emqx_template placeholder envelope; parse_placeholder/1 applies local validation:
+%% See emqx_template:parse/1 in apps/emqx_utils/src/emqx_template.erl.
+PLACEHOLDER     = \$\{[A-Za-z0-9_.]*\}
+%% Match an identifier template that contains at least one placeholder.
+%% It may start with an identifier, or with a placeholder followed by an identifier byte
+%% or another placeholder. The tail may mix identifier bytes and placeholders.
+%% A lone placeholder does not match, so the PLACEHOLDER rule handles it.
+IDENTIFIER_TEMPLATE = ({IDENTIFIER}{PLACEHOLDER}|{PLACEHOLDER}({ID_CONTINUE}|{PLACEHOLDER}))({ID_CONTINUE}|{PLACEHOLDER})*
```
thalesmg: maybe:

```suggestion
IDENTIFIER_TEMPLATE = ({IDENTIFIER}{PLACEHOLDER}|{PLACEHOLDER}({ID_CONTINUE}|{PLACEHOLDER})+)
```

?

thalesmg (dismissed):
🌶️ 💪

apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_sql.erl
savonarola: This allows placeholders in tables, which were mostly forbidden and only occasionally available previously, so I delayed that to be a separate feature.

apps/emqx_bridge_clickhouse/test/emqx_bridge_clickhouse_SUITE.erl
savonarola: ✅

apps/emqx_bridge_tdengine/src/emqx_bridge_tdengine_sql_lexer.xrl
savonarola: the new variant breaks `pre_${tenant}_${clientid}` into `pre_${tenant}`. I simplified regex a bit and added a comment.

zmstone (approved):

apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_sql_lexer.xrl
```diff
@@ -19,10 +19,18 @@ IDENTIFIER      = {ID_START}{ID_CONTINUE}*
 %% emqx_template placeholder envelope:
 %% See emqx_template:parse/1 in apps/emqx_utils/src/emqx_template.erl.
 PLACEHOLDER     = \$\{[A-Za-z0-9_.]*\}
-%% ClickHouse decimal token scanning:
+%% ClickHouse numeric token scanning:
 %% https://github.com/ClickHouse/ClickHouse/blob/8dfb1700858195fa704221e360fa0798ac6ee9ed/src/Parsers/Lexer.cpp#L120-L223
 %% https://github.com/ClickHouse/ClickHouse/blob/8dfb1700858195fa704221e360fa0798ac6ee9ed/src/Parsers/Lexer.cpp#L250-L287
-NUMBER          = (([0-9]+(\.[0-9]*)?)|(\.[0-9]+))([eE][+-]?[0-9]+)?
+DEC_DIGITS      = [0-9](_?[0-9])*
+HEX_DIGITS      = [0-9A-Fa-f](_?[0-9A-Fa-f])*
```
thalesmg, line 26: I think more than one separator in a row is allowed?  🤔

apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_sql.erl
```diff
@@ -1119,6 +1119,42 @@ reject_unsupported_syntax_test() ->
...
+            "INSERT INTO `t` VALUES (10_000, 1_000.25, 1_0e+1_0, .5_0, "
+            "0xc0_fe, 0XCA_FE, 0b11_01, 0B10_10)"
+        >>},
+        rendered_binary(render(Plan, #{}, null_opts()))
+    ).
+
+%% Checks rejection of separators outside digit groups and invalid base digits.
+malformed_numeric_literals_test() ->
+    Invalid = [
+        <<"1__0">>,
```
thalesmg, line 1141: strange, this doesn't seem like it should be invalid.  🤔

apps/emqx_bridge_clickhouse/src/emqx_bridge_clickhouse_sql_lexer.xrl
savonarola: https://github.com/ClickHouse/ClickHouse/blob/8dfb1700858195fa704221e360fa0798ac6ee9ed/src/Common/StringUtils/StringUtils.h#L160-L170
from the code is seems that `1__0` are not treated as numbers

thalesmg: yes, I saw that later and resolved the comment.  :speak_no_evil:

savonarola: I will check on a real DB and make a follow-up if support is needed. But being more strict IMO not a big problem here 🤔

thalesmg: I tried locally, and indeed it rejects double underscores, which is a bit surprising given the behavior of other languages with such separators.

PR #18468, opened 2026-08-24
fix(relup): reject bad target version before touching the node tree

zmstone:
Release version:

Introduced in:
6.0.3

## Summary

Fixes emqx/emqx-dev-team-tasks#353.

`emqx_relup_handler:check_and_unpack/3` deployed the unpacked package into `<root>/relup/<vsn>` before the upgrade-path catalog check ran. A package that ends up rejected (for example with `no_relup_entry`) had already deleted and overwritten files on disk. The target version string read from the package was also used as a path component without validation.

Changes in `plugins/emqx_relup/src/emqx_relup_handler.erl`:
* Validate the target version read from `releases/emqx_vars` as a plain version string. Reject anything else with `bad_target_vsn`.
* Refuse a deploy dir that resolves outside `<root>/relup` as a last-resort check in `get_deploy_dir/2`.
* Run the relup catalog lookup before `deploy_files/6`, so a rejected package does not mutate the filesystem.

Package authenticity verification (a signature, as opposed to the existing `.sha256` integrity check) is intentionally out of scope here and can be a follow-up.

## PR Checklist


plugins/emqx_relup/src/emqx_relup_handler.erl
```diff
@@ -330,11 +332,20 @@ read_rel_vsn(UnpackDir) ->
...
+                {match, [Vsn]} -> validate_rel_vsn(Vsn, File);
                 nomatch -> throw(make_error(rel_vsn_not_found, #{file => File}))
             end
     end.
 
+%% REL_VSN is later used as a path component (deploy dir, releases
+%% subdir); accept only a plain version string like "6.0.1" or
+%% "6.0.1-beta.1", never a path.
+validate_rel_vsn(Vsn, File) ->
+    case re:run(Vsn, "^[0-9]+\\.[0-9]+\\.[0-9]+([.+-][0-9A-Za-z.+-]+)?$", [{capture, none}]) of
```
thalesmg: should we really allow `+` in the version?  do we use that?  🤔

zmstone: never used it before.

thalesmg: maybe we should remove it from the regex, then?

thalesmg:
```elixir
iex(1)> :re.run("+", "[.+-]", capture: :none)
:match
```

thalesmg (approved):

PR #18471, opened 2026-08-24
feat(node_dump): collect etc/emqx.env with secrets obfuscated

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:
N/A

## Summary

Companion to #18451, which ships `etc/emqx.env` and makes `bin/emqx` source it.

`node_dump` now collects the boot-time settings from `etc/emqx.env` when the file exists, so support gets the settings (`EMQX_FEATURES`, `EMQX_SECURITY_PROFILE`) that shape the node but do not appear in `emqx.conf`.

Only these two known non-secret variables are collected (allowlist). The file can carry arbitrary `EMQX_*` overrides, including secrets such as `EMQX_NODE__COOKIE` or a `..._SERVICE_ACCOUNT_JSON`, and no obfuscation pattern matches all of them reliably, so nothing else is copied. The collected lines (`log/emqx.env.dump`) are left next to `app_env.dump` and `conf.hocon` for inspection, and the final warning names the file.

Mergeable independently of #18451: the collection is guarded by a file-existence check, so on a node without `etc/emqx.env` the dump is unchanged.

## PR Checklist

### DIFF
--- a/bin/node_dump
+++ b/bin/node_dump
@@ -82,10 +82,12 @@ done
 } > "${CONF_DUMP}"
 
 
-# Collect the boot-time environment file, with secret values obfuscated:
+# Collect the boot-time settings from the environment file.
+# Only these known non-secret variables; the file can carry arbitrary
+# EMQX_* overrides, including secrets.
 if [ -f "$EMQX_ETC_DIR/emqx.env" ]; then
-    sed -E 's/^([^#=]*(COOKIE|PASSWORD|PASSWD|SECRET|TOKEN|KEY)[^=]*)=.*/\1=********/' \
-        "$EMQX_ETC_DIR/emqx.env" > "${ENV_FILE_DUMP}"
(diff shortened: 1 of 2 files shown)


bin/node_dump
```diff
@@ -81,6 +82,12 @@ done
 } > "${CONF_DUMP}"
 
 
+# Collect the boot-time environment file, with secret values obfuscated:
+if [ -f "$EMQX_ETC_DIR/emqx.env" ]; then
+    sed -E 's/^([^#=]*(COOKIE|PASSWORD|PASSWD|SECRET|TOKEN|KEY)[^=]*)=.*/\1=********/' \
```
keynslug: If any configuration option can be overridden there, this pattern is (strictly speaking) not enough? E.g. `service_account_json`. Perhaps start with dumping just these 2 boot-time environment variables?

qzhuyan (approved):

Issue #18141, opened 2026-07-28
NATS data-integration connector (sink)


id:
### What would you like to be added or enhanced?

A data-integration connector and action for publishing messages to an external NATS server, in the same family as the Kafka and RabbitMQ sinks: connection pooling, health checks, TLS, the common NATS auth options (username/password, token, NKey), and subject templating from rule-engine fields.

Part of the protocol layer already exists in-tree: `apps/emqx_gateway_nats` contains a complete wire codec (`emqx_nats_frame.erl`) and the test suite has a working client (`test/emqx_nats_client.erl`) that runs the conformance tests against a real NATS server. The missing work is connector/action plumbing rather than protocol implementation.

Scope questions: core NATS publish vs JetStream publish, and whether request-reply is needed (probably not in a first iteration).

### Why is this needed?

The NATS gateway is inbound only: it lets NATS clients connect to EMQX. There is no path for EMQX to forward messages out to an existing external NATS deployment, and no NATS connector exists among the data-integration bridges.

The available workaround is the NATS server's own MQTT interface combined with the EMQX MQTT bridge, which requires JetStream and an explicitly enabled MQTT listener on the NATS side. A native sink covers deployments where the NATS configuration cannot be changed, and gives proper rule-engine integration (subject templating, payload transformation, buffering and retry semantics consistent with the other sinks).

zmstone:
desired GA timeline ? @id

id:
Two scope updates from requirements gathering:

Auth: add decentralized JWT authentication with .creds files to the planned auth options (username/password, token, NKey). A .creds file bundles a user JWT and an Ed25519 NKey seed; the client proves possession by signing the server nonce from the INFO frame and sends the signature together with the JWT in CONNECT. This is the standard scheme for NATS 2.x accounts. Implementation-wise it builds on plain NKey support: the nonce-signing handshake is the same, and .creds adds parsing the credentials file and presenting the JWT.

JetStream publish: an optional add-on to consider. A core publish is fire-and-forget at the protocol level, so the connector alone cannot offer at-least-once delivery into NATS. JetStream publish is a request-reply exchange: the client sets a reply-to inbox, waits for the PubAck, and retries on timeout, optionally with a Nats-Msg-Id header for server-side dedup. That gives an end-to-end acknowledgement path for deployments where the target subjects are JetStream-backed.

id:
[...]

PR #18475, opened 2026-08-25
feat: add NATS Core and JetStream sink connector

hjianbo:
Release version: 7.0.0

Introduced in: N/A

## Summary

Adds a standalone `enats_client` dependency and an EMQX NATS data-integration Connector/Action for Core NATS and JetStream publishing.

The connector supports pooled connections, TLS/TLS-first configuration, username/password, token, NKey, JWT and `.creds` authentication, subject/payload/header templates, Resource-managed batching through always-sync connector callbacks, Core batch publishing, and JetStream PubAck with optional stable message IDs. Core batches pre-render and validate messages, submit them through one `enats_client:publish_batch` call, and use one flush barrier. JetStream keeps per-message publishing and PubAck semantics. The dependency is pinned to `enats_client 0.1.12`.

The implementation includes Rule Engine coverage for Core templates, batching, JetStream deduplication, reconnect/error paths, and credential handling.

### Local benchmark (development reference)

A simple single-host benchmark used a Ryzen 5 5600G (6 cores / 12 threads), EMQX Enterprise (OTP 28), NATS Server 2.11.6, and 50 MQTT 3.1.1 clients (QoS 0, 256-byte payloads). Go publisher and subscriber ran separately, with MQTT connecting directly to the container IP. Latency covers MQTT publish to NATS receipt.

Each mode completed three 60-second runs and one five-minute run after a five-second warmup. The five-minute results below met the 200ms P95 threshold.

| Delivery mode | NATS storage / action settings | TPS (received) | P50 | P95 | P99 |
| --- | --- | ---: | ---: | ---: | ---: |
| Core NATS | `batch_size=1000`, `batch_time=20ms`, async, worker/connector pool `32/32`, `dispatch_strategy=random` | 100K/s | 17.3ms | 30.4ms | 36.4ms |
| JetStream | File storage, one replica, `batch_size=100`, `batch_time=5ms`, async, worker/connector pool `128/32`, `dispatch_strategy=random` | 40K/s | 8.6ms | 15.6ms | 20.1ms |

All 30.5M Core and 12.2M JetStream messages, including warmup, were received without action failures, drops or retries. JetStream measures PubAck publishing and live subject delivery, not durable consumption.

Development reference only, not production guarantees or capacity limits. Changed tooling, runtime and topology prevent direct comparison with earlier results; these numbers do not establish a code-level speedup.

Related issue: #18141

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ee/feat-18475.en.md`
- [x] Schema changes are backward compatible or intentionally breaking (new connector; authentication is represented as an explicit union)


apps/emqx_bridge_nats/src/emqx_bridge_nats_action_schema.erl
```diff
@@ -0,0 +1,197 @@
...
+    ];
+%%--------------------------------------------------------------------
+%% Action resource options and examples
+%%--------------------------------------------------------------------
+
+fields(action_resource_opts) ->
+    emqx_bridge_v2_schema:action_resource_opts_fields([
+        {batch_size, #{default => 1}},
+        {batch_time, #{default => <<"0ms">>}}
+    ]).
```
thalesmg, line 140: this is already the default batch size and time, so no need to manually specify this clause.

apps/emqx_bridge_nats/src/emqx_bridge_nats_action_schema.erl
```diff
@@ -0,0 +1,197 @@
...
+fields(?ACTION_TYPE) ->
+    emqx_bridge_v2_schema:make_producer_action_schema(
+        hoconsc:mk(
+            hoconsc:ref(?MODULE, action_parameters),
+            #{
+                required => true,
+                desc => ?DESC("parameters")
+            }
+        ),
+        #{resource_opts_ref => hoconsc:ref(?MODULE, action_resource_opts)}
```
thalesmg, line 64:
```suggestion
        )
```

apps/emqx_bridge_nats/src/emqx_bridge_nats_action_schema.erl
```diff
@@ -0,0 +1,197 @@
...
+    );
+%%--------------------------------------------------------------------
+%% Action parameters
+%%--------------------------------------------------------------------
+
```
thalesmg, line 69:
```suggestion
```

apps/emqx_bridge_nats/src/emqx_bridge_nats_action_schema.erl
```diff
@@ -0,0 +1,197 @@
...
+                #{
+                    required => true,
+                    desc => ?DESC("subject")
+                }
+            )},
+        {payload_template,
+            hoconsc:mk(
+                emqx_schema:template(),
+                #{
+                    default => <<"$", "{.payload}">>,
```
thalesmg, line 84: q: why is it written like this?

maybe:
```suggestion
                    default => <<"${.payload}">>,
```
?

apps/emqx_bridge_nats/src/emqx_bridge_nats_action_schema.erl
```diff
@@ -0,0 +1,197 @@
...
+                #{
+                    required => true,
+                    desc => ?DESC("header_value")
+                }
+            )}
+    ];
+%%--------------------------------------------------------------------
+%% Action resource options and examples
+%%--------------------------------------------------------------------
+
```
thalesmg, line 135:
```suggestion
```

apps/emqx_bridge_nats/src/emqx_bridge_nats_action_schema.erl
```diff
@@ -0,0 +1,197 @@
...
+        }
+    );
+example(put) ->
+    #{
+        enable => true,
+        description => <<"NATS action">>,
+        connector => <<"nats_connector">>,
+        parameters => #{
+            subject => <<"events">>,
+            payload_template => <<"$", "{.payload}">>,
```
thalesmg, line 186:
```suggestion
            payload_template => <<"${.payload}">>,
```

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector_schema.erl
```diff
@@ -0,0 +1,223 @@
...
+        <<"jwt">> ->
+            [hoconsc:ref(?MODULE, auth_jwt)];
+        _ ->
+            throw(#{
+                field_name => mechanism,
+                expected => "user_password | token | nkey | jwt"
+            })
+    end;
+authentication_selector({value, Value}) ->
+    throw(#{field_name => authentication, reason => {not_a_map, Value}}).
```
thalesmg, line 174:
```suggestion
    throw(#{field_name => authentication, reason => bad_value}).
```

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector_schema.erl
```diff
@@ -0,0 +1,223 @@
...
+                #{default => none, desc => ?DESC("authentication")}
+            )},
+        {tls_handshake,
+            hoconsc:mk(
+                hoconsc:enum([starttls, first]),
+                #{default => starttls, desc => ?DESC("tls_handshake")}
+            )}
+    ] ++
+        emqx_connector_schema:resource_opts() ++
+        emqx_connector_schema_lib:ssl_fields();
```
thalesmg, line 82: if the connection is usually tls in prod (probably is?), then perhaps:


```suggestion
        emqx_connector_schema_lib:ssl_fields(#{enable_by_default => true});
```

scripts/ct/run.sh
```diff
@@ -381,6 +381,19 @@ if [ "$STOP" = 'no' ] && [ "$PS" = 'no' ]; then
         exit 1
     fi
     set -e
+
+    if [[ " ${CT_DEPS} " == *" nats "* ]]; then
+        # The NATS bridge suite starts short-lived servers with custom configs
+        # (auth, TLS, JetStream, and restart cases), so it needs the server
+        # executable inside the Erlang test container, not only a NATS sidecar.
```
thalesmg, line 388: maybe use a shared volume for sharing the secrets, like kafka test container does?

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+on_start(InstanceId, Config) ->
+    case client_options(Config) of
+        {error, Reason} ->
+            {error, {invalid_config, Reason}};
+        ClientOpts ->
+            PoolOpts = [
+                {pool, InstanceId},
+                {pool_size, maps:get(pool_size, Config, 8)},
+                {config, ClientOpts}
+            ],
```
thalesmg, line 47: you probably want to set `auto_reconnect` so ecpool restarts crashed workers.

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+    emqx_resource_pool:common_health_check_workers(
+        InstanceId,
+        #{
+            check_fn => fun health_check/1,
+            is_success_fn => fun
+                (ok) -> false;
+                (_) -> true
+            end
+        }
+    ).
```
thalesmg, line 105: we need to honor the configured `resource_opts.health_check_timeout` from the connector by setting `timeout => HCTimeout` here.

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+        msg_id => emqx_placeholder:preproc_tmpl(maps:get(msg_id_template, Params, <<>>)),
+        request_ttl => RequestTTL
+    },
+    {ok, State#{channels => maps:put(ChannelId, Channel, Channels)}}.
+
+on_remove_channel(_InstanceId, #{channels := Channels} = State, ChannelId) ->
+    {ok, State#{channels => maps:remove(ChannelId, Channels)}}.
+
+on_get_channel_status(InstanceId, _ChannelId, State) ->
+    on_get_status(InstanceId, State).
```
thalesmg, line 93: if the health check for each channel is identical to the connector's health check, there's no use in repeating it here.


```suggestion
    ?status_connected.
```

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+
+on_add_channel(
+    _InstanceId,
+    #{channels := Channels} = State,
+    ChannelId,
+    #{parameters := Params} = ChannelConfig
+) ->
+    Subject = maps:get(subject, Params),
+    PayloadTemplate = emqx_template:parse(
+        maps:get(payload_template, Params, <<"$", "{.payload}">>)
```
thalesmg, line 68: not only it's written in this strange way, but the schema already enforces a default, right?

```suggestion
        maps:get(payload_template, Params)
```

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+    ),
+    HeaderTemplates = [
+        {
+            emqx_template:parse(maps:get(key, Header)),
+            emqx_template:parse(maps:get(value, Header))
+        }
+     || Header <- maps:get(headers, Params, [])
```
thalesmg, line 75: I suggest deconstructing the config map at the top and using the schema's default values instead of multiple map:get's.

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+    ],
+    ResourceOpts = maps:get(resource_opts, ChannelConfig, #{}),
+    RequestTTL = maps:get(request_ttl, ResourceOpts, 5000),
```
thalesmg, line 78: these also have default values.

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+            check_fn => fun health_check/1,
+            is_success_fn => fun
+                (ok) -> false;
+                (_) -> true
+            end
+        }
+    ).
+
+health_check(Client) ->
+    enats_client:flush(Client, 1000).
```
thalesmg, line 108: should also honor resource_opts.health_check_timeout.

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+    [classify_result(Result) || Result <- Results];
+classify_result({error, ecpool_empty}) ->
+    {error, {recoverable_error, disconnected}};
+classify_result({error, Reason}) ->
+    {error, classify_error(Reason)};
+classify_result(Result) ->
+    Result.
+
+classify_error(disconnected) ->
+    {recoverable_error, disconnected};
```
thalesmg, line 402: we're in the process of improving ecpool's errors here: https://github.com/emqx/emqx/pull/19055

with the newer ecpool version, this should be:


```suggestion
classify_error({disconnected, #{reason := Reason} = Error}) ->
    %% from ecpool_worker:client/1
    {Kind, _} = classify_error(Reason),
    {Kind, Error};
```

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+    SSL = maps:get(ssl, Config, #{}),
+    case auth_options(maps:get(authentication, Config, none)) of
+        {error, _} = Error ->
+            Error;
+        Auth ->
+            #{
+                servers => Servers,
+                connect_timeout => maps:get(connect_timeout, Config, 5000),
+                tls => maps:get(enable, SSL, false),
+                tls_handshake => maps:get(tls_handshake, Config, starttls),
```
thalesmg, line 359: again, IMO it would be far easier to read if the config was destructured at the beginning instead of a bunch of nested fn calls

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+    enats_client:flush(Client, maps:get(request_ttl, Channel, 5000));
+maybe_flush_core(_Client, _Channel, _FlushCore, Result) ->
+    Result.
+
+publish_rendered(Client, Channel, #{
+    subject := Subject, payload := Payload, headers := Headers, msg_id := MsgId
+}) ->
+    Options = #{
+        headers => Headers,
+        timeout => maps:get(request_ttl, Channel, 5000)
```
thalesmg, line 327: IMO, this should be taken without a hardcoded inline default, and enforce each channel state has the key at creation time.

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,456 @@
...
+        {
+            emqx_template:parse(maps:get(key, Header)),
+            emqx_template:parse(maps:get(value, Header))
+        }
+     || Header <- maps:get(headers, Params, [])
+    ],
+    ResourceOpts = maps:get(resource_opts, ChannelConfig, #{}),
+    RequestTTL = maps:get(request_ttl, ResourceOpts, 5000),
+    Channel = #{
+        subject => emqx_placeholder:preproc_tmpl(Subject),
```
thalesmg, line 80: nit: `emqx_template` is our newer, I believe better alternative for templating.

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector_schema.erl
hjianbo: We don't have enough evidence that NATS servers generally enable TLS by default. To stay consistent with the other connectors, we'll keep SSL disabled by default and let users opt in.

apps/emqx_bridge_nats/src/emqx_bridge_nats_action_schema.erl
hjianbo: Most files in the repository keep a blank line after section comments, so I kept that convention unchanged.

thalesmg: though this section is between clause of the same function, not the same as other sections...

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector_schema.erl
thalesmg: ok, though it would be really weird that the norm in production is not to use TLS.

thalesmg: or, thinking about the hardened profile we've been implementing, maybe it makes more sense to enable it by default and let the user disable it if they really want/need?

i.e.: opt out of a safe default, instead of opting in to it.

scripts/ct/run.sh
```diff
@@ -381,6 +381,17 @@ if [ "$STOP" = 'no' ] && [ "$PS" = 'no' ]; then
...
+        # The host destination below is mounted at /var/lib/secret in Erlang.
+        docker cp nats:/nats-server /tmp/emqx-ci/emqx-shared-secret/nats-server
+        docker exec -u root erlang bash -c \
+            "install -m 0755 /var/lib/secret/nats-server /usr/local/bin/nats-server"
+        docker exec erlang nats-server --version
+    fi
```
thalesmg, line 394: now that the shared secret volume is mounted, maybe there's no need for this?

just start the nats container similar to other docker-ct files do?

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,484 @@
...
+
+%%--------------------------------------------------------------------
+%% NATS connection and publishing
+%%--------------------------------------------------------------------
+
+connect(Options) ->
+    ClientOpts = proplists:get_value(config, Options),
+    case enats_client:start_link(ClientOpts#{owner => self(), reconnect => true}) of
+        {ok, Client} ->
+            case enats_client:connect(Client, maps:get(connect_timeout, ClientOpts, 5000)) of
```
thalesmg, line 185: nit: make the parameter required and expected, since we control it completely.

otherwise, it might always use the default due to a typo or refactoring.

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
```diff
@@ -0,0 +1,484 @@
...
+publish_batch(Client, #{delivery_mode := DeliveryMode} = Channel, Batch) ->
+    case DeliveryMode of
+        core ->
+            publish_core_batch(Client, Channel, Batch);
+        jetstream ->
+            publish_jetstream_batch(Client, Channel, Batch)
+    end.
+
+publish_core_batch(Client, Channel, Batch) ->
+    IndexedBatch = lists:zip(lists:seq(1, length(Batch)), Batch),
```
thalesmg, line 208: micro nit:


```suggestion
    IndexedBatch = lists:enumerate(Batch),
```

apps/emqx_bridge_nats/src/emqx_bridge_nats_connector.erl
thalesmg: this was marked as resolved, but nothing was changed?

i.e.: ecpool_worker:client/1 no longer returns this reason.

apps/emqx_bridge_nats/test/emqx_bridge_nats_SUITE.erl
```diff
@@ -0,0 +1,1392 @@
...
+    ActionName = proplists:get_value(action_name, Config),
+    Parent = self(),
+    Messages = [
+        {<<"emqx.batch.one">>, <<"batch-one">>},
+        {<<"bad subject">>, <<"batch-bad">>},
+        {<<"emqx.batch.two">>, <<"batch-two">>}
+    ],
+    _ = [
+        spawn(fun() ->
+            Result = emqx_bridge_v2:send_message(
```
thalesmg, line 411: please avoid calling those inner functions.  use instead HTTP API calls to manage the connector/action, and real emqtt clients to do the publishing.

calling those lower-level API functions make these tests much harder to fix and/or refactor if necessary, besides essentially mocking out stuff that might happen in real scenarios.

apps/emqx_bridge_nats/test/emqx_bridge_nats_SUITE.erl
```diff
@@ -0,0 +1,1392 @@
...
+            put(nats_test_reason, Reason),
+            ?assertEqual(Expected, emqx_bridge_nats_connector:on_query(test, {channel, #{}}, State))
+        end,
+        Cases
+    ).
+
+t_invalid_connector_config(_Config) ->
+    ?assertMatch(
+        {error, {invalid_config, {invalid_authentication, _}}},
+        emqx_bridge_nats_connector:on_start(
```
thalesmg, line 362: please avoid calling those inner functions.  use instead HTTP API calls to manage the connector/action, and real emqtt clients to do the publishing.

calling those lower-level API functions make these tests much harder to refactor if necessary, besides essentially mocking out stuff that might happen in real scenarios.

apps/emqx_bridge_nats/test/emqx_bridge_nats_SUITE.erl
```diff
@@ -0,0 +1,1392 @@
...
+            {jetstream, invalid_ack, bad_ack},
+            {error, {unrecoverable_error, {jetstream, invalid_ack, bad_ack}}}
+        },
+        {other_error, {error, {unrecoverable_error, other_error}}},
+        {ecpool_empty, {error, {recoverable_error, disconnected}}}
+    ],
+    lists:foreach(
+        fun({Reason, Expected}) ->
+            put(nats_test_reason, Reason),
+            ?assertEqual(Expected, emqx_bridge_nats_connector:on_query(test, {channel, #{}}, State))
```
thalesmg, line 354: please avoid calling those inner functions.  use instead HTTP API calls to manage the connector/action, and real emqtt clients to do the publishing.

calling those lower-level API functions make these tests much harder to refactor if necessary, besides essentially mocking out stuff that might happen in real scenarios.

apps/emqx_bridge_nats/test/emqx_bridge_nats_SUITE.erl
thalesmg, line 1: there are lots of direct calls to the inner, lower-level APIs like `bridge_v2:send` and several resource module callbacks.

please avoid calling those inner functions.  use instead HTTP API calls to manage the connector/action, and real emqtt clients to do the publishing.

calling those lower-level API functions make these tests much harder to refactor if necessary, besides essentially mocking out stuff that might happen in real scenarios.

Issue #18478, opened 2026-08-25
[6.3.0][fuzzing] MQTT 3 PUBREL flood causes unbounded PUBCOMP responses and TCP backpressure


Le9Wang:
### What happened?

EMQX continuously responds to a large number of repeated MQTT 3.1.1 PUBREL packets with PUBCOMP packets. This causes outbound response buildup and TCP backpressure. The connection eventually stops reading incoming data, and the client resets the connection.

<img width="2706" height="1382" alt="Image" src="https://github.com/user-attachments/assets/40dd53c2-e40f-4608-b74e-844f3594a6c1" />

### Reproduction

Send a valid PUBREL packet repeatedly:

```
62 02 00 01
```

The packet represents:

- **PUBREL**, Remaining Length: 2, Packet Identifier: 1

The fuzzing case appends approximately 8,388,352 anomaly bytes, consisting of repeated PUBREL packets.

### Expected Behavior

EMQX should protect the connection from unlimited responses caused by repeated PUBREL packets with an unknown Packet Identifier. Possible acceptable behaviors include:

- Rate-limiting PUBCOMP responses
- Closing the client connection after exceeding an invalid-PUBREL threshold
- Applying bounded packet processing and response buffering

EMQX should remain responsive and should not allow one client to create unbounded protocol response pressure.

### Actual Behavior

EMQX accepts and processes repeated PUBREL packets:

```
pubrel_packetId_not_found
packetId: 1
```

It sends a PUBCOMP response for each packet:

```
PUBCOMP(PacketId=1, ReasonCode=146)
```

Packet capture evidence:

- Client sent approximately 200 KB of fuzz data
- EMQX sent approximately 6,981 bytes of responses
- EMQX then advertised a TCP receive window of 0
- The client continued sending data
- The client eventually sent TCP RST at `2026-08-25 14:00:26.007612`

No EMQX crash or OOM was observed.

### Environment

- **EMQX revision**: 4aa0fc40f7
- **Protocol**: MQTT 3.1.1 over TCP
- **Broker**: 192.168.66.18:1883
- **Gateway/source address**: 192.168.66.4
- **Client ID**: Defensics
- **Username**: testuser
- **Max MQTT packet size**: 100 MB

### Suspected Cause

The incoming stream parser processes a large number of packets from one read and generates a response for every unknown PUBREL. The connection layer does not appear to enforce a bound on:

- The number of packets processed per read
- Repeated unknown Packet Identifier responses
[...]

PR #18482, opened 2026-08-25
fix(mqtt): lower log level for acks with unknown packet id

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:
pre-5.8

## Summary

Addresses [#18478](https://github.com/emqx/emqx/issues/18478). A client that repeatedly sends
PUBREL, PUBREC, or PUBCOMP packets carrying a Packet Identifier the broker has no state for could
fill the log with warnings, one line per packet. For MQTT 3.1.1 each such acknowledgement is 4
bytes and the broker's reply is also 4 bytes, so there is no response amplification, but a 4-byte
packet produced a 130–180 byte warning line on disk, and the default log level is `warning`.

These three events (`emqx_channel.erl`, in `process_pubrec/2`, `process_pubrel/2`, and
`process_pubcomp/2`) are already recorded by the `packets.pubrec.missed`, `packets.pubrel.missed`,
and `packets.pubcomp.missed` counters. Following the project convention that a counted event is
logged at `info` rather than `warning`, the three log calls are moved to `info`. The flood then
writes nothing at the default level, while the counters keep the aggregate observable and the
per-packet detail remains available at `info`.

Protocol behavior is unchanged. MQTT 3.1.1 requires the broker to answer a PUBREL with a PUBCOMP
carrying the same Packet Identifier ([MQTT-3.6.4-1], [MQTT-4.3.3-2]); an unknown Packet Identifier
is not a protocol violation, so the reply is still sent. No rate-limiting or disconnect is added,
because neither is warranted by the input and disconnecting on a valid packet would violate the
MUST-respond requirement.

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_channel_SUITE.erl
+++ b/apps/emqx/test/emqx_channel_SUITE.erl
@@ -494,6 +494,33 @@ t_handle_in_pubrel_not_found_error(_) ->
     {ok, {outgoing, ?PUBCOMP_PACKET(1, ?RC_PACKET_IDENTIFIER_NOT_FOUND)}, _Channel} =
         emqx_channel:handle_in(?PUBREL_PACKET(1, ?RC_SUCCESS), channel()).
 
+t_handle_in_pubrel_not_found_bounded_log(_) ->
+    %% A flood of PUBREL packets with an unknown Packet Identifier must not
+    %% produce warning-level log output: the event is recorded by the
+    %% packets.pubrel.missed counter and logged at info instead.
+    ok = meck:expect(
+        emqx_session,
+        pubrel,
+        fun(_, _PacketId, _Session) ->
+            {error, ?RC_PACKET_IDENTIFIER_NOT_FOUND}
+        end
+    ),
+    N = 100,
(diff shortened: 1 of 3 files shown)


apps/emqx/test/emqx_channel_SUITE.erl
```diff
@@ -494,6 +494,33 @@ t_handle_in_pubrel_not_found_error(_) ->
...
+    %% A flood of PUBREL packets with an unknown Packet Identifier must not
+    %% produce warning-level log output: the event is recorded by the
+    %% packets.pubrel.missed counter and logged at info instead.
+    ok = meck:expect(
+        emqx_session,
+        pubrel,
+        fun(_, _PacketId, _Session) ->
+            {error, ?RC_PACKET_IDENTIFIER_NOT_FOUND}
+        end
+    ),
```
keynslug: Q: Btw, why mocking if empty session should not have any awaiting rels anyway?

keynslug (dismissed):

keynslug (approved):

PR #18485, opened 2026-08-25
feat(mqtt): disable await_rel_timeout by default

keynslug:
Release version: 7.0.0

## Summary

This PR changes few things around default `await_rel_timeout`.
1. `await_rel_timeout` is now disabled by default.
2. `max_awaiting_rel` is now advertised in CONNACK, which better agrees with MQTT specification.
3. `max_awaiting_rel` is soft-clamped to 1.
4. Affected schema field descriptions are corrected.

See also: #18470, #16721.

## PR Checklist

    Defaults are changed however.

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -2257,7 +2257,7 @@ enrich_conninfo(
         username => Username,
         conn_props => ConnProps,
         expiry_interval => ExpiryInterval,
-        receive_maximum => receive_maximum(Zone, ConnProps)
+        receive_maximum => client_receive_maximum(Zone, ConnProps)
     },
     {ok, Channel#channel{conninfo = NConnInfo}}.
 
@@ -2279,7 +2279,7 @@ clamp_session_expiry(RequestedSec, infinity) ->
 clamp_session_expiry(RequestedSec, MaxMs) when is_integer(MaxMs) ->
     min(RequestedSec, MaxMs div 1000).
 
-receive_maximum(Zone, ConnProps) ->
+client_receive_maximum(Zone, ConnProps) ->
     MaxInflightConfig =
         case get_mqtt_conf(Zone, max_inflight) of
             0 -> ?RECEIVE_MAXIMUM_LIMIT;
@@ -3251,9 +3251,7 @@ enrich_connack_caps(AckProps, ?IS_MQTT_V5 = Channel) ->
         clientinfo = #{
             zone := Zone
         },
-        conninfo = #{
-            receive_maximum := ReceiveMaximum
-        }
+        session = Session
     } = Channel,
     #{
         max_packet_size := MaxPktSize,
@@ -3263,6 +3261,7 @@ enrich_connack_caps(AckProps, ?IS_MQTT_V5 = Channel) ->
         shared_subscription := Shared,
         wildcard_subscription := Wildcard
     } = emqx_mqtt_caps:get_caps(Zone),
+    ReceiveMaximum = server_receive_maximum(emqx_session:info(awaiting_rel_max, Session)),
     NAckProps = AckProps#{
         'Retain-Available' => flag(Retain),
         'Maximum-Packet-Size' => MaxPktSize,
@@ -3282,6 +3281,11 @@ enrich_connack_caps(AckProps, ?IS_MQTT_V5 = Channel) ->
     end;
 enrich_connack_caps(AckProps, _Channel) ->
     AckProps.
+
+server_receive_maximum(infinity) ->
+    ?RECEIVE_MAXIMUM_LIMIT;
+server_receive_maximum(MaxAwaitingRel) ->
+    min(MaxAwaitingRel, ?RECEIVE_MAXIMUM_LIMIT).
 
 %%--------------------------------------------------------------------
 %% Enrich server keepalive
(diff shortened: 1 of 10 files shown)


zmstone:
maybe dev-70

apps/emqx/src/emqx_schema.erl
```diff
@@ -4553,7 +4553,7 @@ mqtt_session() ->
             sc(
                 duration(),
                 #{
-                    default => <<"300s">>,
+                    default => 0,
```
zmstone, line 4556: is 0 a valid default value for duration() ? if not.
```suggestion
                    default => ~"0s",
```

apps/emqx/src/emqx_channel.erl
```diff
@@ -3263,6 +3261,7 @@ enrich_connack_caps(AckProps, ?IS_MQTT_V5 = Channel) ->
         shared_subscription := Shared,
         wildcard_subscription := Wildcard
     } = emqx_mqtt_caps:get_caps(Zone),
+    ReceiveMaximum = server_receive_maximum(emqx_session:info(awaiting_rel_max, Session)),
```
zmstone, line 3264: Before this change, `max_infight` is for client maximum, and server mirrors it.

Now it's no longer symmetric. EMQX by default allows 100 inflight messages published from client.

Should perhaps change the default value to 32 for `awaiting_rel_max` too ?

apps/emqx/src/emqx_channel.erl
```diff
@@ -2279,7 +2279,7 @@ clamp_session_expiry(RequestedSec, infinity) ->
 clamp_session_expiry(RequestedSec, MaxMs) when is_integer(MaxMs) ->
     min(RequestedSec, MaxMs div 1000).
 
-receive_maximum(Zone, ConnProps) ->
+client_receive_maximum(Zone, ConnProps) ->
```
zmstone, line 2282: the i18n desc for max_inflight should be updated.
current description is not so clear which direction.
now it means that the server will not publish more than such number of PUBLISH messages before acknowledgement (discuss qos1 and qos2 respectively).

rel/i18n/emqx_schema.hocon
```diff
@@ -511,7 +511,9 @@ base_listener_mountpoint.desc: """~
 base_listener_mountpoint.label: "Mountpoint"
 
 mqtt_max_awaiting_rel.desc:
-"""For each publisher session, the maximum number of outstanding QoS 2 messages pending on the client to send PUBREL. After reaching this limit, new QoS 2 PUBLISH requests will be rejected with `147(0x93)` until either PUBREL is received or timed out."""
+"""For each publisher session, the maximum number of QoS 2 messages for which EMQX is waiting to receive `PUBREL`. Once this limit is reached, EMQX rejects a further QoS 2 `PUBLISH` by disconnecting the client with "Receive Maximum Exceeded" error code (0x93). An entry is removed when its `PUBREL` is received or when `mqtt.await_rel_timeout` expires.
```
zmstone: this now also impacts qos1?

apps/emqx/src/emqx_channel.erl
keynslug: > Now it's no longer symmetric.

It was actually never supposed to be symmetric. _Receive Maximum_ in CONNACK should tell the client how many concurrent QoS > 0 publishes the broker is willing to process, and since EMQX does not limit client->broker QoS 1 concurrency in any way the only tangible limit we can communicate to the client is `max_awaiting_rel`.

> Should perhaps change the default value to 32 for `awaiting_rel_max` too?

You mean just to make sure user-visible / client-visible impact is minimized? E.g. so that well-behaving spec-adhering MQTT clients do not suddenly increase their publishing concurrency three times?

rel/i18n/emqx_schema.hocon
keynslug: Not sure how? QoS1 client->broker publishing is unaffected by this PR, i.e. clients publishing QoS1-only should never receive "Receive Maximum Exceeded", just as before. Unless I'm missing something.

apps/emqx/src/emqx_schema.erl
keynslug: Verified against a release, starts just fine both with default `emqx.conf` and with an explicit
```
mqtt {
  await_rel_timeout = 0
}
```

apps/emqx/src/emqx_channel.erl
zmstone: > You mean just to make sure user-visible / client-visible impact is minimized? E.g. so that well-behaving spec-adhering MQTT clients do not suddenly increase their publishing concurrency three times?

that's right

rel/i18n/emqx_schema.hocon
zmstone: isn't it sent as max receive in connack

rel/i18n/emqx_schema.hocon
```diff
@@ -511,7 +511,9 @@ base_listener_mountpoint.desc: """~
...
+
+Values below 1 are treated as 1. For MQTT 5.0 clients, the effective limit is advertised as Receive Maximum in `CONNACK`, capped at `65535`."""
```
zmstone: Receive Maximum also impacts QoS 1, so we should document QoS 1 in the line above too.

zmstone: that EMQX send PUBACK as soon as possible, so there is no effect limit, but this number should, at least for well behaving clients, help clients to flow control.

zmstone (approved):

PR #18492, opened 2026-08-25
feat(zerobus): follow ups

thalesmg:
Release version:
7.0.0

## Summary

- update test suite to actually run against real zerobus.
- notify stream writers when protobuf schema changes, so they may reopen their streams if needed.
- set gun connect retries to 0 so we may capture connection errors before calls time out.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
+++ b/apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
@@ -160,7 +160,12 @@ recover(ActionResId, Idx) ->
     maybe
         [?META_ROW(_, Val)] ?=
             ets:lookup(?META_TAB, ?META_STREAM_KEY(ActionResId, Idx)),
-        #{?writer := Writer} ?= Val,
+        #{
+            ?writer := Writer,
+            ?n_restarts := _,
+            ?stream := _,
+            ?opts := _
+        } ?= Val,
         true ?= is_pid(Writer) andalso is_process_alive(Writer),
         Val
     else
@@ -202,15 +207,6 @@ handle_follow_stream(FollowReq, State0) ->
     },
     {Reply, State}.
 
-clear_state(State0) ->
-    State0#{
-        ?writer := ?undefined,
-        ?n_restarts := ?undefined,
-        ?last_acked_seq := -1,
-        ?stream := ?undefined,
-        ?opts := #{}
-    }.
-
 handle_recv(#{?stream := ?undefined} = State0) ->
     ?tp(~"zerobus_receiver_recv_no_stream", #{}),
     State0;
@@ -240,13 +236,13 @@ handle_helper_down({ok, Res}, State0) ->
             Reason = maybe_format_grpc_reason(grpc_client:trailers_to_error(Trailers)),
             Error = {error, Reason},
             notify_errored(Error, State1),
-            State2 = clear_state(State1),
-            {?continue, State2};
+            {?continue, State1};
         {more, Results} ->
             LastAckedSeq = find_last_acked_seq(Results, LastAckedSeq0, State0),
             ?tp("zerobus_last_seq_scanned", #{}),
             State1 = State0#{?last_acked_seq := LastAckedSeq},
             maybe_notify_acked(LastAckedSeq0, State1),
+            ?tp("zerobus_last_seq_notified", #{}),
             {?nudge, State1};
         {error, {deadline_exceeded, _}} ->
             {?nudge, State0};
@@ -254,8 +250,7 @@ handle_helper_down({ok, Res}, State0) ->
             %% stream is gone
             Error = {error, stream_closed},
             notify_errored(Error, State0),
-            State1 = clear_state(State0),
-            {?continue, State1};
+            {?continue, State0};
         {error, Reason} ->
             ?tp(info, "zerobus_receiver_unexpected_error_response", #{
                 action_res_id => ActionResId,
@@ -266,8 +261,7 @@ handle_helper_down({ok, Res}, State0) ->
                 false ?= is_process_alive(Pid),
                 Error = {error, stream_closed},
                 notify_errored(Error, State0),
-                State1 = clear_state(State0),
-                {?continue, State1}
+                {?continue, State0}
             else
                 _ ->
                     {?nudge, State0}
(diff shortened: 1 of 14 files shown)


apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_impl.erl
```diff
@@ -550,7 +562,9 @@ do_fold_grpc_optvar_results(Keys) ->
         end,
         Results
     ),
+    %% {disconnected, unhealthy_target} > disconnected > connecting > connected
     ToStatus = fun
+        ({S, {unhealhy_target, _}}) -> {S, unhealthy_target};
```
zmstone: unhealhy_target -> unhealthy_target

missing test coverage?

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
```diff
@@ -202,15 +207,6 @@ handle_follow_stream(FollowReq, State0) ->
...
-clear_state(State0) ->
-    State0#{
-        ?writer := ?undefined,
-        ?n_restarts := ?undefined,
-        ?last_acked_seq := -1,
-        ?stream := ?undefined,
-        ?opts := #{}
-    }.
-
 handle_recv(#{?stream := ?undefined} = State0) ->
```
zmstone: Removing `clear_state/1` leaves `?stream` pointing at a dead stream on every `?continue` path (`{done, ...}`, `stream_closed`, and the dead-helper branch).

That is fine as long as nothing calls `handle_recv/1` again before a new `#follow_stream{}` arrives, since `?continue` does not nudge. But a `#recv{}` already sitting in the mailbox from an earlier `?nudge` would now reach the `#{?helper := ?undefined}` clause below and call `do_recv_stream_once_async/2` on the closed stream — previously `clear_state/1` set `?stream := ?undefined` and this first clause caught it.

Is that race actually reachable? If not, worth a comment saying why, because this clause and its `zerobus_receiver_recv_no_stream` probe are now only reachable from the initial state.

Related: `t_...` at SUITE `:2469` uses that probe as a `?force_ordering` anchor. If the clause becomes effectively initial-state-only, that ordering still works but for a different reason than before — worth confirming the test still exercises what it intends.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_utils.erl
```diff
@@ -0,0 +1,97 @@
...
+
+%%------------------------------------------------------------------------------
+%% API
+%%------------------------------------------------------------------------------
+
+-spec list_all_stream_writer_pids() -> [pid()].
+list_all_stream_writer_pids() ->
+    Key = emqx_bridge_zerobus_stream_writer_worker:gproc_name('_', '_'),
+    MS = [{{Key, '$1', '_'}, [], ['$1']}],
+    gproc:select({local, names}, MS).
```
zmstone: `gproc:select({local, names}, MS)` is node-local, so a writer only learns about a serde change if the hook runs on its own node.

`emqx_schema_registry_config:post_config_update/5` should run on every node via cluster_rpc, which would make this correct — but it is worth stating in a comment, since a reader of this module alone cannot tell whether cross-node notification is handled or missing.

If `post_config_update/5` turns out to run only on the originating node, writers elsewhere keep a stale descriptor indefinitely and there is no self-healing path.

apps/emqx_schema_registry/src/emqx_schema_registry_config.erl
```diff
@@ -173,6 +176,9 @@ post_config_update(
     end,
     case emqx_schema_registry:build_serdes([{NewName, NewSchema}]) of
         ok ->
+            %% run hook in worker pool to avoid blocking config update for long.
+            Ctx = #{name => NewName},
+            emqx_pool:async_submit(fun ?MODULE:serde_updated_hook/1, [Ctx]),
```
zmstone: `emqx_pool:async_submit/2` is fire-and-forget, so a rejected or dropped submission silently means some writers never reopen and keep encoding against the old descriptor.

Given the failure mode is silent and the recovery path is manual (restart the action), is it worth logging when the submit fails? The comment explains why it is async, which is the right call — it is only the unobserved failure that concerns me.

Also: `build_serdes/1` has already succeeded at this point, but the hook reads the serde back through `emqx_schema_registry:get_serde/1` on the pool worker. `handle_serde_updated/2` handles `schema_not_found` with a comment calling it a rare race — that race is presumably this asynchrony, so the two comments are worth cross-referencing.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
```diff
@@ -202,15 +207,6 @@ handle_follow_stream(FollowReq, State0) ->
     },
     {Reply, State}.
 
-clear_state(State0) ->
```
zmstone, line 205: why deleted?

thalesmg: commented in the commit https://github.com/emqx/emqx/pull/18492/commits/20cabb68ebdffcdfb265390df2c8e3dfc80d64ea

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_impl.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_async_receiver_worker.erl
zmstone: maybe keep n_restarts and clear others

or drain #recv records in mailbox

thalesmg: why keep others?  that goes against the comment.

and once it detects such errors, it no longer nudges itself with #recvs .

thalesmg: I think I see what you mean by "others".

thalesmg: :checkered_flag:

zmstone (approved):

PR #18493, opened 2026-08-25
feat(mgmt): add POST /certs/pem_cache_clean HTTP API

### LINKED ISSUE #11110: Is there an API interface that can perform a task similar to "emqx_ctl pem_cache clean all"

From WeChat Group

### Why is this needed?

-

zmstone:
Release version:
7.0.0

Introduced in:

## Summary

Closes #11110.

Adds `POST /certs/pem_cache_clean`: an HTTP API that cleans the x509 certificate (PEM) cache on all nodes, equivalent to the CLI command `emqx ctl pem_cache clean all`.

The endpoint lives in `emqx_mgmt_api_certs` and calls the existing `emqx_mgmt:clean_pem_cache_all/0` (over the existing `emqx_proto_v1:clean_pem_cache/1` BPAPI); no new RPC plumbing.

Decisions taken:

* Method and path: `POST /certs/pem_cache_clean`, mirroring the CLI subcommand name (`pem_cache clean`) so the two stay discoverable from each other. `POST` (rather than `DELETE` on a cache resource) because the operation is an action, not a plain eviction: clearing also refreshes the trusted CA database, which re-reads CA files still referenced by live TLS connections, and evicted entries are re-read from disk on the next handshake. The path sits under the module's `/certs` prefix but outside the bundle paths (`/certs/global/...`, `/certs/ns/...`), so it does not read as an operation on managed certificate bundles.
* Cluster-wide only. The issue asks for the `clean all` form; a per-node endpoint would be a second path to document and test for little gain, and the per-node form stays available via the CLI. On partial failure the response body lists the failing nodes, so the operator can retry the failed nodes from the CLI.
* Error status is 500 (`INTERNAL_ERROR` with a `node_errors` list), not the 400 the authz cache endpoint returns: a node RPC failure is a server-side error, not a malformed request.
* Scope: the endpoint inherits the module's `system` API-key scope. TLS certificate handling is system configuration (the listeners API is also `system`); the `access_control` scope of the authz analogue covers authn/authz, which the PEM cache is not part of.

The test populates the cache with a real TLS handshake against the default SSL listener and asserts the listener certificate entry is present in the `ssl_pem_cache` table before and gone after the call, on every node (single-node and 2-node cluster matrix).

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_certs.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_certs.erl
@@ -29,7 +29,8 @@
     '/certs/global/list'/2,
     '/certs/global/name/:name'/2,
     '/certs/ns/:namespace/list'/2,
-    '/certs/ns/:namespace/name/:name'/2
+    '/certs/ns/:namespace/name/:name'/2,
+    '/certs/pem_cache_clean'/2
 ]).
 -export([ns_bundle_filter/2]).
 
@@ -57,7 +58,8 @@ paths() ->
         "/certs/global/list",
         "/certs/global/name/:name",
         "/certs/ns/:namespace/list",
-        "/certs/ns/:namespace/name/:name"
+        "/certs/ns/:namespace/name/:name",
+        "/certs/pem_cache_clean"
     ].
 
 schema("/certs/global/list") ->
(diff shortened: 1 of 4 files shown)


thalesmg (approved):
not related to changes herein, but it's curious that this "multicall" is actually several serial calls.

Issue #13587, opened 2024-08-07
Auto subscribe based on client id pattern


Aposhian:
### What would you like to be added or enhanced?

As I understand, currently you can only create auto subscribes for all clients. You can use placeholders for the topics, but you can't have one set of auto subscribes for a certain set of clients, and another set of auto subscribes for another set of clients.

It would be nice if you could specify a client id list or regex for which a set of auto subscribes should be applied. Or even if you could write a custom rule to parse client ids to auto subscribes, like if you have client ids of the pattern A-B-C, and then you want to do different auto subscribes based on the contents of A, B and/or C.

### Why is this needed?

If you have different types of clients, you don't want them to subscribe to the same set of topics.

qzhuyan:
https://docs.emqx.com/en/enterprise/v5.7/admin/api-docs.html#tag/Clients/paths/~1clients~1%7Bclientid%7D~1subscribe/post

maybe you can use restapi as the workaround?

Aposhian:
That is a fair point, although one downside of that approach is that you can't subscribe client ids that don't have an active session. I would love to be able to use auto-subscribe so that a client can connect for the first time and not need to subscribe. If I use the REST API I still need to wait for the client to come online. My goal is to have the client receiving data on connect as soon as possible, without having to manage subscribe or resubscribe logic on the client. Not a huge deal, but it seems like auto-subscribe is close to being able to do what I want.

zmstone:
In EMQX 5.8 Enterprise edition, there is going to be a message transformation feature which can do more than auto-subscribe.
There is no public doc released yet, but here is the design doc for a quick glance: https://github.com/emqx/eip/blob/main/active/0027-message-transform.md (will be moved to `implemented` dir after 5.8 is relesed)

PR #18496, opened 2026-08-26
feat(auto_subscribe): support ${client_attrs.<key>} placeholders in topics

zmstone:
Release version:
7.0.0

Introduced in:
N/A

## Summary

Closes emqx/emqx#13587.

Auto-subscribe topics now support `${client_attrs.<key>}` placeholders, so a client can be auto-subscribed to topics built from the attributes set by `mqtt.client_attrs_init`.

`emqx_auto_subscribe_placeholder` used a hand-written tokeniser that recognised only `${clientid}`, `${username}`, `${host}` and `${port}`. Any other placeholder was silently kept as literal text in the subscription topic, so `${client_attrs.tenant}` produced a wrong subscription with no error. The module now parses topics with `emqx_template`, the same way `emqx_mountpoint:replvar/2` does.

Decisions taken:

* The allowed-vars list is module-local: `clientid`, `username`, `host`, `port` and the `client_attrs` namespace. `${host}` and `${port}` are legacy auto-subscribe names (the standard modules use `${peerhost}`) and resolve from the connection peername; they keep rendering byte-identical topics. `t_placeholder_compat` pins this and passes on the pre-change code.
* A topic referencing a client attribute the client does not have is skipped with an `auto_subscribe_ignored` warning, following the existing `${username}`-undefined behaviour. It does not render an empty string or leave literal placeholder text.
* Unknown placeholders such as `${nope}` are still kept as literal topic text (escaped via `emqx_template:escape_disallowed/2`, as in `emqx_mountpoint`), but a warning is now logged when the config is loaded. Rejecting them outright could prevent a node with an existing config from booting, so validation stays warn-only.

Most relevant files: `apps/emqx_auto_subscribe/src/emqx_auto_subscribe_placeholder.erl`, `apps/emqx_auto_subscribe/test/emqx_auto_subscribe_SUITE.erl`, `rel/i18n/emqx_auto_subscribe_schema.hocon`.

## PR Checklist


apps/emqx_auto_subscribe/src/emqx_auto_subscribe_placeholder.erl
```diff
@@ -51,39 +76,55 @@ to_topic_table(PHs, ClientInfo, ConnInfo) ->
...
+lookup([<<?VAR_CLIENTID>>], {#{clientid := ClientId}, _ConnInfo}) ->
+    {ok, ClientId};
+lookup([<<?VAR_USERNAME>>], {#{username := Username}, _ConnInfo}) when Username =/= undefined ->
+    {ok, Username};
+lookup([<<?VAR_HOST>>], {_ClientInfo, #{peername := {Host, _Port}}}) ->
+    {ok, list_to_binary(inet:ntoa(Host))};
+lookup([<<?VAR_PORT>>], {_ClientInfo, #{peername := {_Host, Port}}}) ->
+    {ok, integer_to_binary(Port)};
+lookup([<<"client_attrs">>, Name], {ClientInfo, _ConnInfo}) ->
+    case maps:get(client_attrs, ClientInfo, #{}) of
```
savonarola: Micro nit: may use a single nested match

apps/emqx_auto_subscribe/src/emqx_auto_subscribe_placeholder.erl
```diff
@@ -10,12 +10,37 @@
 
 -export([to_topic_table/3]).
 
+%% `emqx_template' access module callback
+-export([lookup/2]).
+
+%% Legacy auto-subscribe names. Unlike the standard `${peerhost}'/`${peerport}',
```
savonarola, line 16: Maybe support standard vars too

savonarola (dismissed):

apps/emqx_auto_subscribe/src/emqx_auto_subscribe_placeholder.erl
zmstone: can add

```
VAR_CERT_CN_NAME
VAR_CERT_SUBJECT 
VAR_ZONE
VAR_LISTENER
```

savonarola (approved):

PR #18504, opened 2026-08-26
fix(stomp): decode header escape sequences and accept CRLF line endings

### LINKED ISSUE #12917: After STOMP enables authentication, authentication fails when the account password carries a colon
### What happened?

After STOMP enables authentication, authentication fails when the account password carries a colon. 
MQTT doesn't have this problem

STOMP account and password could contain a colon

_No response_

_No response_

### EMQX version

sysdescr : EMQX
version : 5.5.1
datetime : 2024-04-22T19:25:59.961416581+08:00
uptime : 10 days, 9 hours, 35 minutes, 26 seconds

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

zmstone:
Release version:

Introduced in:
v5.0.0 (cce0b1ca34, #6650)

## Summary

Fixes #12917. Port of #18495 to `dev-60`; the parser code is identical on both branches.

The STOMP frame parser never decoded any header escape sequence and rejected CRLF line endings. A client that escapes a colon in its passcode as `\c`, as STOMP 1.2 requires, could not authenticate: the password arrived at the authenticator as `pa\css` instead of `pa:ss`. Clients that send a raw colon worked, which is why the bug depended on how spec-compliant the client library is.

### Root cause

Commit cce0b1ca34 (#6650, v5.0.0) fixed sticky TCP stream parsing by adding a catch-all clause to `emqx_stomp_frame:parse/2`:

```erlang
parse(Bytes, #{phase := Phase, state := State}) when Phase =/= none ->
    parse(Phase, Bytes, State);
```

This clause matches every parser state with a non-`none` phase, so the later `parse/2` clauses that handled CR, backslash escapes, and the partial-escape continuation became unreachable, and `unescape/1` became dead code. The guard hides the shadowing from the compiler.

### Fix

Bytes are consumed in the `parse/3` state machine, so escape and CR handling now lives in the `parse/3` clauses instead of being reordered in `parse/2`:

* `hdname` and `hdvalue` decode `\c`, `\r`, `\n`, `\\`. A chunk ending in a bare `\` or bare `\r` returns `{more, ...}` with the byte kept in `pre`, and the next chunk continues the sequence, preserving the sticky-stream behavior #6650 fixed. A test feeds a frame one byte at a time.
* `command`, `headers`, `hdname` and `hdvalue` accept CRLF wherever LF is accepted, and a standalone CRLF is a heartbeat. Previously a client using CRLF line endings could not connect at all.
* The `headers` phase now has its own continuation state, so "between header lines" and "in the middle of a header name" are distinct. This also fixes a corner where a chunk boundary inside a header name followed by LF jumped to body parsing.

Behavior decisions:

* **CONNECT and CONNECTED frames are exempt from unescaping.** STOMP 1.2 ("Value Encoding") excludes them from header escaping for backward compatibility with STOMP 1.0, so their headers pass through verbatim: a password containing a literal backslash is preserved, and sequences that are undefined escapes elsewhere are not an error there. A password containing a colon is sent raw in CONNECT, which the parser already accepts. The `STOMP` command, added in 1.1, is not exempt and decodes escapes.
* **In escaping frames, undefined escape sequences are a fatal frame error** (`\t` etc.), as STOMP 1.2 requires ("MUST be treated as a fatal protocol error"). Previously they passed through silently, but only because the escape handling was unreachable; the pre-#6650 parser also raised on them.
* **A space after the header colon is now part of the value.** The old parser stripped one space after the colon. STOMP 1.2 forbids trimming or padding header values. No test or shipped client relied on the strip.
* **Only header names and values are unescaped.** The command line and the body pass through untouched.

The serialize side (`escape/1`) is unchanged; for escaping frames, serialize-then-parse now round-trips exactly, including values containing `:`, `\`, CR and LF. Serialize still escapes CONNECTED headers even though the spec exempts them: the only such value EMQX emits that a client could influence is `session`, and escaping keeps a value containing LF from injecting a header line. In practice these headers never contain escapable octets.

### Measured behavior, before and after

Parsing a `SEND` frame with `emqx_stomp_frame:parse/2`, destination value shown:

| input | before | after |
|---|---|---|
| `a\cb` | `a\cb` | `a:b` |
| `a:b` (raw colon) | `a:b` | `a:b` |
| `a\\b` | `a\\b` | `a\b` |
| `a\nb` | `a\nb` | `a` LF `b` |
| `a\tb` (undefined) | `a\tb` | fatal frame error |
| CRLF line endings | `unexpected_linefeed` error | parses |
| chunk split at bare `\` | escape left undecoded | `a:b` |
| chunk split at bare `\r` | `unexpected_linefeed` error | parses |

`CONNECT` and `CONNECTED` headers are exempt from escaping (STOMP 1.0 compatibility) and pass through verbatim, before and after; a raw colon in a `CONNECT` passcode parses correctly. The `STOMP` command decodes escapes like any other frame.

## PR Checklist

### DIFF
--- a/apps/emqx_gateway_stomp/src/emqx_stomp_frame.erl
+++ b/apps/emqx_gateway_stomp/src/emqx_stomp_frame.erl
@@ -85,6 +85,10 @@
 -define(BSL, $\\).
 -define(COLON, $:).
 
+%% CONNECT and CONNECTED frames do not use header escaping, for backward
+%% compatibility with STOMP 1.0 (STOMP 1.2, "Value Encoding").
+-define(NO_HEADER_ESCAPE(Cmd), (Cmd =:= ?CMD_CONNECT orelse Cmd =:= ?CMD_CONNECTED)).
+
 -record(parser_state, {
     cmd,
     headers = [],
@@ -174,6 +178,10 @@ parse(hdname, <<?CR, _Rest/binary>>, _State) ->
     error(unexpected_linefeed);
 parse(hdname, <<?COLON, Rest/binary>>, State = #parser_state{acc = Acc}) ->
     parse(hdvalue, Rest, State#parser_state{hdname = Acc, acc = <<>>});
+parse(hdname, <<?BSL, Rest/binary>>, State = #parser_state{cmd = Cmd}) when
+    ?NO_HEADER_ESCAPE(Cmd)
+->
+    parse(hdname, Rest, acc(?BSL, State));
 parse(hdname, <<?BSL>>, State) ->
     {more, #{phase => hdname, pre => <<?BSL>>, state => State}};
 parse(hdname, <<?BSL, Ch:8, Rest/binary>>, State) ->
@@ -199,6 +207,10 @@ parse(hdvalue, <<?CR, ?LF, Rest/binary>>, State) ->
     parse(hdvalue, <<?LF, Rest/binary>>, State);
 parse(hdvalue, <<?CR, _Rest/binary>>, _State) ->
     error(linefeed_expected);
+parse(hdvalue, <<?BSL, Rest/binary>>, State = #parser_state{cmd = Cmd}) when
+    ?NO_HEADER_ESCAPE(Cmd)
+->
+    parse(hdvalue, Rest, acc(?BSL, State));
 parse(hdvalue, <<?BSL>>, State) ->
     {more, #{phase => hdvalue, pre => <<?BSL>>, state => State}};
 parse(hdvalue, <<?BSL, Ch:8, Rest/binary>>, State) ->
--- a/apps/emqx_gateway_stomp/src/emqx_stomp_frame.erl
+++ b/apps/emqx_gateway_stomp/src/emqx_stomp_frame.erl
@@ -126,58 +126,58 @@ g(Key, Opts, Val) ->
 -spec parse(binary(), parse_state()) -> parse_result().
 parse(<<>>, Parser) ->
     {more, Parser};
-%% treat the \n as a heartbeat frame
-parse(<<$\n>>, Parser = #{phase := none}) ->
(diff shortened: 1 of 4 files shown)


apps/emqx_gateway_stomp/src/emqx_stomp_frame.erl
```diff
@@ -193,6 +193,16 @@ parse(
...
+parse(hdvalue, <<?CR>>, State) ->
+    {more, #{phase => hdvalue, pre => <<?CR>>, state => State}};
+parse(hdvalue, <<?CR, ?LF, Rest/binary>>, State) ->
+    parse(hdvalue, <<?LF, Rest/binary>>, State);
+parse(hdvalue, <<?CR, _Rest/binary>>, _State) ->
+    error(linefeed_expected);
+parse(hdvalue, <<?BSL>>, State) ->
+    {more, #{phase => hdvalue, pre => <<?BSL>>, state => State}};
+parse(hdvalue, <<?BSL, Ch:8, Rest/binary>>, State) ->
+    parse(hdvalue, Rest, acc(unescape(Ch), State));
```
zmstone: Should this unescape path be command-aware and skip `CONNECT`/`CONNECTED`? STOMP 1.2 value encoding exempts those frames from header escaping for backward compatibility (https://stomp.github.io/stomp-specification-1.2.html#Value_Encoding). With unconditional decoding, a valid `CONNECT` credential containing a literal `\c` is parsed as `:`, and a literal `\t` becomes a fatal frame error even though `CONNECT` headers are not supposed to be escaped. Regular frames should still decode/reject escapes, but applying it to `CONNECT` can break passwords with backslashes.

hjianbo (dismissed):

hjianbo (approved):

Issue #18506, opened 2026-08-26
[6.3.0] mqueue_len mismatch — reports 4 messages but only 3 returned by mqueue_messages API


Le9Wang:
### What happened?

The `mqueue_len` field in the client info API reports a different count than the actual number of messages returned by the mqueue messages API. Specifically, `mqueue_len` shows 4, but only 3 messages are returned.

### Reproduction Steps

1. Subscribe with a shared subscription and persistent session:
   ```
   mqttx sub -h 10.42.6.6 -t \$share/group/t/# -q 1 -i sub-1 --no-clean -se 6000
   ```
2. Disconnect the subscriber
3. Publish 3 messages:
   ```
   mqttx pub -h 10.42.6.254 -t t/a1 -q 1 -S 1B
   mqttx pub -h 10.42.6.254 -t t/a1 -q 1 -S 1B
   mqttx pub -h 10.42.6.254 -t t/a1 -q 1 -S 1MB
   ```
4. Query client info:
   ```
   GET /api/v5/clients/sub-1
   ```
5. Query mqueue messages:
   ```
   GET /api/v5/clients/sub-1/mqueue_messages?limit=200&position=none
   ```

### Expected Behavior

`mqueue_len` should match the actual number of messages in the queue.

### Actual Behavior

Client info shows `mqueue_len: 4`, but the mqueue messages API returns only 3 messages:

```json
// GET /api/v5/clients/sub-1
{
  "mqueue_len": 4,
  "send_msg": 3,
  "send_cnt": 2,
  "mqueue_dropped": 0
}

// GET /api/v5/clients/sub-1/mqueue_messages
{
  "data": [
    { "payload": "Bw==" },   // 1B
    { "payload": "Nw==" },   // 1B
    { "payload": "fw==" }    // 1MB
  ]
}
```

### Environment

- **EMQX version**: 6.3.0-beta.2
- **Protocol**: MQTT 5
- **Subscription**: `$share/group/t/#`, QoS 1, no-clean session, session expiry 6000s
- **Payloads**: 1B, 1B, 1MB

zmstone:
@Le9Wang 
~If you call the `GET /api/v5/clients/sub-1` API again some time later, you'll likely observe consistent numbers.~

zmstone:
Reproduced. 
This is not a stale counter: mqueue_len: 4 is correct,  and the two numbers will not converge. 
The mqueue_messages API is what hides the fourth message.

[...]

PR #18509, opened 2026-08-26
fix(mgmt): do not skip client messages cut off by max_payload_bytes

zmstone:
Release version:

Introduced in:
pre-5.8

## Summary

`GET /clients/{clientid}/mqueue_messages` and `GET /clients/{clientid}/inflight_messages` bound each response page by the `max_payload_bytes` parameter (default 1MB). When this bound cut a page short, the returned `meta.position` pointed at the last message walked under the `limit` parameter, past the messages that were cut off. Requesting the next page from that position skipped the cut messages, so a client following the pager could never list them. Reported in #18506 as `mqueue_len` reporting more messages than the mqueue messages API returns.

The fix is in `emqx_mgmt_api_clients`: `format_msgs/4` now reports whether the page was cut short and which message was the last one included; `format_msgs_resp/4` then recomputes `meta.position` from that message, so the next page starts at the first cut-off message. The `max_payload_bytes` parameter description is updated to document the paging behavior.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -1895,7 +1895,7 @@ format_msgs(MsgType, [FirstMsg | Msgs], PayloadFmt, MaxBytes) ->
             Msgs
         ),
     Data = lists:reverse(Msgs1),
-    case length(Data) < 1 + length(Msgs) of
+    case length(Data) =< length(Msgs) of
         true -> {Data, {truncated, LastMsg}};
         false -> {Data, all}
     end;
--- a/apps/emqx_management/src/emqx_mgmt_api_clients.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_clients.erl
@@ -1859,8 +1859,9 @@ format_msgs_resp(MsgType, Msgs, Meta, QString) ->
         <<"payload">> := PayloadFmt,
         <<"max_payload_bytes">> := MaxBytes
     } = QString,
-    Meta1 = encode_msgs_meta(MsgType, Meta),
-    Resp = #{meta => Meta1, data => format_msgs(MsgType, Msgs, PayloadFmt, MaxBytes)},
+    {Data, Truncation} = format_msgs(MsgType, Msgs, PayloadFmt, MaxBytes),
+    Meta1 = encode_msgs_meta(MsgType, adjust_pager_position(MsgType, Truncation, Meta)),
+    Resp = #{meta => Meta1, data => Data},
     %% Make sure minirest won't set another content-type for self-encoded JSON response body
     Headers = #{<<"content-type">> => <<"application/json">>},
     case emqx_utils_json:safe_encode(Resp) of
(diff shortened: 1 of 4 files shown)


apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -1879,23 +1880,40 @@ format_msgs_resp(MsgType, Msgs, Meta, QString) ->
...
                         {halt, Acc}
                 end
             end,
-            {[FirstMsg1], PayloadSize0},
+            {[FirstMsg1], {FirstMsg, PayloadSize0}},
             Msgs
         ),
-    lists:reverse(Msgs1);
+    Data = lists:reverse(Msgs1),
+    case length(Data) < 1 + length(Msgs) of
```
zmstone:
```suggestion
    case length(Data) =< length(Msgs) of
```

rel/i18n/emqx_mgmt_api_clients.hocon
```diff
@@ -556,7 +556,7 @@ payload_encoding.label:
 """Payload Encoding"""
 
 max_payload_bytes.desc:
-"""Client's inflight/mqueue messages payload limit. The total payload size of all messages in the response will not exceed this value. Messages beyond the limit will be silently omitted in the response. The only exception to this rule is when the first message payload is already larger than the limit, in this case, the first message will be returned in the response."""
+"""Client's inflight/mqueue messages payload limit. The total payload size of all messages in the response will not exceed this value. When the limit is reached, the response may contain fewer messages than `limit`; `meta.position` points to the last returned message, so the omitted messages can be fetched by requesting the next page from that position. The only exception to this rule is when the first message payload is already larger than the limit, in this case, the first message will be returned in the response."""
```
zmstone:
```suggestion
"""Client's inflight/mqueue messages payload limit. The total payload size of all messages in the response will not exceed this value. When the limit is reached, the response may contain fewer messages than `limit`; `meta.position` points to the last returned message, so the omitted messages can be fetched by requesting the next page from that position. The only exception to this rule is when the first message payload is already larger than the limit, in this case, only the first message will be returned in the response."""
```

apps/emqx_management/src/emqx_mgmt_api_clients.erl
```diff
@@ -1879,23 +1880,40 @@ format_msgs_resp(MsgType, Msgs, Meta, QString) ->
...
                         {halt, Acc}
                 end
             end,
-            {[FirstMsg1], PayloadSize0},
+            {[FirstMsg1], {FirstMsg, PayloadSize0}},
             Msgs
         ),
-    lists:reverse(Msgs1);
+    Data = lists:reverse(Msgs1),
+    case length(Data) =< length(Msgs) of
```
thalesmg: micro suggestion: the fold could return a special value to denote it has been cut short, so we don't need to traverse (potentially?) long lists

thalesmg (dismissed):

apps/emqx_management/src/emqx_mgmt_api_clients.erl
zmstone: done

thalesmg (approved):

PR #18326, opened 2026-08-11
feat(sec): restore legacy sec profile as default

savonarola:
Release version: 6.3

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (describe the changes and the reasoning in the summary)

PR #18511, opened 2026-08-26
feat(sec): use hardened security profile by default

zmstone:
Release version:
7.0.0

Introduced in:
N/A

## Summary

Make `hardened` the default security profile in EMQX 7.0.

This redoes #18232 (reverted in #18326 because the team decided to switch in 7.0 rather than 6.3).
Since the revert, the test suites gained legacy/hardened matrix groups, so almost all of the test
adaptation from #18232 is already on `dev-70`. What remains:

* `apps/emqx/src/emqx_security_profile.erl`: flip `?PROFILE_DEFAULT` to `hardened`.
* `bin/emqx`: flip the boot script default (`EMQX_SECURITY_PROFILE:-hardened`). This copy of the
  default was added after #18232; both must agree, or packaged nodes stay on `legacy`.
* `apps/emqx_conf/etc/emqx.env`: rewrite the security-profile block for a release where `hardened`
  is the default. The advice now targets operators who need `legacy` to migrate. The behaviour list
  is re-checked against the current `emqx_security_profile:policy/1` set, adding ExHook
  server-unavailable denial, plugin SHA-256 binding, and the built-in database password policies.
* `apps/emqx_auth/etc/acl.conf`: correct a stale comment claiming the default flipped in 6.3.

Upgrade-safety fix found while running the suites under the new default:

* `apps/emqx_auth_mnesia/src/emqx_authn_mnesia_schema.erl`: the built-in database union selector
  treated an absent `autogenerate_password` field as the profile default. Under `hardened` that
  routed every pre-7.0 config (hash algorithm set, no `autogenerate_password` field) into the
  generated-password struct, and validation failed with
  `Expected: disable, got:suffix` at `password_hash_algorithm.salt_position`. Backup import
  (`emqx_mgmt_data_backup`) failed on such configs, and boot-time config validation would fail the
  same way after an upgrade. The selector now treats a config that carries a hash algorithm and no
  `autogenerate_password` field as a manual config, matching the intent already encoded in
  `builtin_db_converter/2`. Note: this keeps manual `sha256` configs valid; algorithms the
  hardened profile deliberately rejects (for example bcrypt) still fail closed, as designed
  in the safe-password-hashing feature.

Tests:

* `emqx_security_profile_SUITE:t_unset_default` now asserts the default is `hardened` when
  `EMQX_SECURITY_PROFILE` is unset or empty; new `t_env_var_selects_profile` asserts both explicit
  values still work. #18232 had no such test, which is why the revert was invisible to CI.
* `scripts/test/test_emqx_boot.py`: new `test_security_profile_defaults_to_hardened` guards the
  `bin/emqx` default independently of the Erlang define.
* `emqx_authn_mnesia_schema_SUITE`: new hardened case for the pre-7.0 manual-hash config shape.
* `emqx_authn_schema_SUITE:t_check_schema`: made profile-explicit; asserts bcrypt is rejected and
  manual sha256 stays valid under `hardened`.
* `emqx_mgmt_api_nodes_SUITE`: default-profile assertions flipped from `legacy` to `hardened`.
* `emqx_mgmt_api_plugins_SUITE`: three cluster cases now grant plugin installation through the
  profile-aware `allow_package/1` helper (binds the package sha256 under `hardened`).
* `emqx_mt_api_SUITE`: the built-in database authenticator fixture declares
  `autogenerate_password = false` so manual user passwords stay valid in both profiles.
* `emqx_ds_shared_sub_SUITE`: relaxed the cluster ping budgets (`heartbeat_interval` 100ms -> 500ms,
  `leader_timeout` 100ms -> 1s). The 200ms leader-side ping deadline could kick a live borrower
  during the QoS1 storm in `t_cluster_smoke`; hardened adds small per-message work in test builds
  and made this race fail often enough to block CI. Reallocation, checkpoint, and commit budgets
  are unchanged.

No test suite is pinned to `legacy` (0 pinned); the two spots that now run under an explicit
`legacy` profile (`emqx_authn_schema_SUITE:t_check_schema`, the flipped branch of
`emqx_mgmt_api_nodes_SUITE:t_node_info_non_default_profile_and_preset`) assert legacy-specific
behavior and also cover hardened.

Verified locally: `apps/emqx` full CT (107 suites), plus the `emqx_conf`, `emqx_dashboard`,
`emqx_management`, `emqx_auth`, `emqx_exhook`, gateway family, `emqx_mt`, `emqx_telemetry`,
`emqx_license`, and `emqx_bridge_mqtt` app CTs under the new default. Remaining failures are
pre-existing load flakes that fail identically in the legacy and hardened matrix groups
(`emqx_dashboard_monitor:t_randomize`, `emqx_mt_SUITE:t_namespaced_metrics`,
`emqx_bridge_mqtt_source:t_reconnect_with_session`) or PBKDF2-login-cost timetraps
(`emqx_dashboard_login_lock`, `emqx_mgmt_api_api_keys:t_create_failed`).

CI under the new default (all first exposed by this PR: at #18232's time `bin/emqx` still pinned
`legacy`, so that PR's packaging and docker CI never ran hardened):

* `fix(authn): add descriptions for hardened password hash structs`: the schema dump is generated
  under the default profile; the docs check failed on the `hardened_builtin_*` union members.
* `ci: configure what the hardened default profile requires`: the packaging, docker, compose, and
  helm smoke harnesses now configure what the migration guide tells operators to configure — a
  non-default Erlang cookie, a non-default dashboard password (passed to `emqx-smoke-test.sh` via
  the new `EMQX_SMOKE_PASSWORD`), `0.0.0.0` binds for listeners reached over a docker or k8s
  network, `enable_authn = false` where test clients connect anonymously, and sha256 instead of
  the (now rejected) plain hash for the essential-auth-smoke bootstrap users.
* Verified locally against the built release: boot with only cookie and password set, loopback
  listener binds, `emqx-smoke-test.sh` green, anonymous MQTT connect green with listener authn
  disabled.
* After rebasing onto #18557 (`node.default_listener_address`, merged into `dev-70` via the
  `dev-63` sync), the bind overrides are no longer needed in any container-based harness: the
  docker entrypoint sets `all`, and the helm chart runs the same entrypoint. They are removed
  here, so CI exercises the shipped image default rather than a CI-only workaround. What stays
  is what an operator must genuinely configure: a non-default cookie and dashboard password,
  `enable_authn=false` plus permissive authz where test clients connect anonymously, and sha256
  password hashing in authn fixtures.
* `emqx.env` and the changelog now point at `node.default_listener_address` as the way to serve
  remote clients without abandoning the rest of the hardened profile.

First-run notes for a fresh 7.0 install with defaults:

* The node refuses to boot with the shipped default Erlang cookie; `bin/emqx` dies with a message
  telling the operator to set `node.cookie` or `EMQX_NODE__COOKIE`.
* Dashboard login with the unchanged default password returns 401 with instructions to run
  `emqx ctl admins passwd` or set `dashboard.default_password`.
* Default MQTT and Dashboard listeners bind to loopback.
* Docker: the container still requires `EMQX_NODE__COOKIE`, but the listener binds are handled
  by #18557 — the image's entrypoint sets `EMQX_NODE__DEFAULT_LISTENER_ADDRESS=all`, so published
  ports keep working under the hardened default. This PR is rebased on that feature and drops the
  per-listener bind overrides it had added to the docker, compose, and helm CI harnesses.

## PR Checklist


.ci/docker-compose-file/conf.cluster.env
```diff
@@ -2,6 +2,10 @@ EMQX_NAME=emqx
 EMQX_LICENSE__KEY=evaluation
 EMQX_CLUSTER__DISCOVERY_STRATEGY=static
 EMQX_CLUSTER__STATIC__SEEDS="[emqx@node1.emqx.io, emqx@node2.emqx.io]"
+# The hardened profile (default since 7.0) refuses the default cookie.
```
zmstone: delete this comment

.github/actions/package-macos/action.yaml
```diff
@@ -64,6 +64,11 @@ runs:
         mkdir emqx
         unzip -d emqx $pkg_name > /dev/null
         set -e
+        # The hardened profile (default since 7.0) refuses the default cookie
```
zmstone: delete

.github/workflows/build_and_push_docker_images.yaml
```diff
@@ -103,9 +103,16 @@ jobs:
       - name: smoke test
         timeout-minutes: 5
         run: |
-          CID=$(docker run -d -p 18083:18083 $_EMQX_DOCKER_IMAGE_TAG)
+          # The hardened profile (default since 7.0) refuses the default
```
zmstone: delete

.github/workflows/build_and_push_docker_images.yaml
```diff
@@ -103,9 +103,16 @@ jobs:
...
+          # cookie and rejects login with the unchanged default password.
+          # The dashboard bind needs no override: the docker entrypoint sets
```
zmstone: delete

savonarola (approved):

PR #18514, opened 2026-08-26
test(pulsar): wait for action to connect instead of matching create

zmstone:
## Summary

De-flake `emqx_bridge_pulsar_SUITE` action setup:

```
%%% emqx_bridge_pulsar_SUITE ==> local.tcp.sync.t_send_when_down: FAILED
{{badmatch,{201, #{... <<"status">> => <<"connecting">>,
                   <<"status_reason">> => <<"Not connected for unknown reason">> ...}}}, ...}
```

Seen in: https://github.com/emqx/emqx/actions/runs/32962334187/job/98160575457?pr=18508 (r62→r63 sync PR, which does not touch pulsar)

Pulsar producers connect asynchronously, so the action create API can return 201 while the action is still `connecting` (first health check pending). Three sites hard-matched `{201, #{<<"status">> := <<"connected">>}}` on the create response: `do_t_send_with_failure/2` (feeds `t_send_when_down` and `t_send_when_timeout`) and the cluster/resilience cases.

Fix (test-only): add `create_action_connected/1` — create with `{201, _}`, then `?retry(500, 20, ...)` until `get_action_api` reports `connected`, the same pattern the suite already uses for the connector. All three sites use the helper.

Base `dev-62`: the suite blob is identical on dev-62/63/70, so the fix follows the sync chain. dev-60/61 carry an older variant of this suite and are left alone.

### DIFF
--- a/apps/emqx_bridge_pulsar/test/emqx_bridge_pulsar_SUITE.erl
+++ b/apps/emqx_bridge_pulsar/test/emqx_bridge_pulsar_SUITE.erl
@@ -482,6 +482,20 @@ get_connector_api(TCConfig) ->
 get_action_api(TCConfig) ->
     emqx_bridge_v2_testlib:get_action_api2(TCConfig).
 
+%% Pulsar producers connect asynchronously, so the create API can return while
+%% the action is still `connecting'. Create, then wait until it is connected.
+create_action_connected(TCConfig) ->
+    {201, _} = create_action_api(TCConfig, #{}),
+    ?retry(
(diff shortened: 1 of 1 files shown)


thalesmg (approved):

thalesmg:
maybe should port to d60?

zmstone:
> maybe should port to d60?

Base dev-62: the suite blob is identical on dev-62/63/70, so the fix follows the sync chain. dev-60/61 carry an older variant of this suite and are left alone.

PR #18357, opened 2026-08-14
fix(boot): gate connections until node boot completes (v6)

zmstone:
Release version:
6.0.4, 6.1.5, 6.2.3, 6.3.0

Introduced in:
pre-5.8

## Summary

Fixes emqx/emqx-dev-team-tasks#341 (part 2 of the closed emqx/emqx-dev-team-tasks#338).

MQTT listeners open before the managed boot sequence finishes. A client that connects in this window is served before plugin hooks (`client.authenticate`, `client.authorize` and others) are registered, so it bypasses a plugin's security logic. The part 1 fix (#18337) moved plugin start to the tail of `ensure_apps_started/0`, which widened this window on v6.

Changes:

* New module `emqx_node_readiness` in the `emqx` app: a `persistent_term` flag with three functions, `is_ready/0`, `mark_ready/0`, `mark_not_ready/0`. The flag defaults to ready. Only `emqx_machine_boot:ensure_apps_started/0` clears it (before starting any app) and sets it (after `emqx_plugins:ensure_started/0`). `ensure_apps_started/0` is also the ekka join/leave callback, so a cluster rejoin re-gates the window. If boot never completes, the node keeps refusing connections.
* Connection process inits check the flag and refuse the connection while the node is not ready:
  * `emqx_connection:init/4` closes the raw socket before the transport handshake and exits with `{shutdown, node_not_ready}`.
  * `emqx_ws_connection:init/2` replies HTTP 503 before the WebSocket upgrade.
  * `emqx_quic_connection:new_conn/3` shuts the incoming QUIC connection down with a dedicated `MQTT_QUIC_CONN_ERROR_NOT_READY` application error code, mirroring the overload-protection branch.
  * `emqx_gateway_conn:init/6` closes the socket and exits (TCP-based gateways). UDP/connectionless gateway paths (CoAP, MQTT-SN) have no per-connection process and are not covered by this gate.
* MQTT TCP and WS refusals log at `info` level; refusal volume is observable through the per-listener shutdown counter, fed by the `{shutdown, node_not_ready}` exit reason. Gateway refusals log at `warning` (with peername and channel module) since gateway listeners have no shutdown-counter observability. No new throttle-log message id.
* `GET /status` returns HTTP 503 while the node is not ready, so load balancers stop routing new connections to a starting node.
[...]

PR #18527, opened 2026-08-26
fix(hooks): tolerate missing topic index tables in publish hook lookups

### LINKED ISSUE #17776: Schema validation: on_message_publish crashes with badarg (ets:next) on $SYS topic messages in 6.2.1
## What happened?

After enabling a **Schema Validation** rule on EMQX Enterprise **6.2.1**, the `emqx_schema_validation` `on_message_publish` hook throws `badarg` whenever a `$SYS/...` system message is published (e.g. the `$SYS/brokers/<node>/clients/<clientid>/disconnected` event emitted on every client disconnect).

The crash originates in `emqx_schema_validation_registry:matching_validations/1`, which performs a topic-index search that calls `ets:next/2` with a malformed key (`{[<<"$SYS">>],{}}`) and fails with `badarg`.

The exception is caught by `emqx_hooks:safe_execute/2`, so the broker stays up, but:
- The `on_message_publish` hook fold is aborted for the affected message.
- The log is flooded with `hook_callback_exception` errors (one per client disconnect, since each produces a `$SYS .../disconnected` publish).
- More importantly, it raises doubt about whether the validation hook reliably evaluates **non-`$SYS`** messages, which is a problem when the rule is used as a security control (`failure_action: drop`).

This did **not** occur on 6.1.0 with the same configuration.

## Stacktrace

```
[error] clientid: <id>, msg: hook_callback_exception, username: webserver, reason: badarg,
stacktrace: [
  {ets,next,[emqx_schema_validation_index,{[<<"$SYS">>],{}}],[{error_info,#{cause => id,module => erl_stdlib_errors}}]},
  {emqx_topic_index,'-make_nextf/1-fun-0-',2,[{file,"emqx_topic_index.erl"},{line,97}]},
  {emqx_trie_search,search_new,4,[{file,"emqx_trie_search.erl"},{line,155}]},
  {emqx_trie_search,search,3,[{file,"emqx_trie_search.erl"},{line,201}]},
  {emqx_schema_validation_registry,matching_validations,1,[{file,"emqx_schema_validation_registry.erl"},{line,114}]},
  {emqx_schema_validation,on_message_publish,1,[{file,"emqx_schema_validation.erl"},{line,119}]},
  {emqx_hooks,safe_execute,2,[{file,"emqx_hooks.erl"},{line,213}]},
  {emqx_hooks,do_run_fold,3,...},
  {emqx_broker,eval_hook_and_publish,2,...},
  ...
],
callback_module: emqx_schema_validation, callback_function: on_message_publish
```

The triggering message topic was `$SYS/brokers/emqx@<node>/clients/<clientid>/disconnected`.

## Steps to reproduce

1. Run EMQX Enterprise 6.2.1 (single node, default RocksDB-backed mnesia).
2. Create a schema validation that matches a normal (non-`$SYS`) topic, e.g.:
   - name: `deny_drc_mode_enter`
   - topics: `["thing/product/+/services"]`
   - checks: a single `sql` check, e.g. `SELECT payload.method as method WHERE payload.method != 'drc_mode_enter' OR username IN ('webserver', 'uavia')`
   - failure_action: `drop`, enable: `true`
3. Connect any MQTT client and then disconnect it (this generates a `$SYS/.../disconnected` publish), or otherwise cause any `$SYS` topic publish.
4. Observe the `hook_callback_exception` / `badarg` error above in the broker log.

## Expected

`matching_validations/1` should safely return no matches for `$SYS/...` topics (and any topic that does not match a configured validation), without throwing. The publish hook should never raise on system topics.

## Environment

- EMQX **Enterprise 6.2.1** (Docker `emqx/emqx:6.2.1`)
- Single node, Docker
- Same config ran without error on 6.1.0
- One enabled schema validation rule with a single `sql` check, `failure_action: drop`

## Impact / workaround

- Workaround: disable schema validation, or pin to 6.1.0.
- Note: downgrading 6.2.1 -> 6.1.0 in place is not possible because 6.2.1 bumps the RocksDB `format_version` to 6, which 6.1.0 cannot open (`Corruption: unsupported format_version: 6`). A data-dir reset is required to roll back.

zmstone:
Release version:

Introduced in:
pre-5.8

## Summary

Fixes #17776.

The `message.publish` hook callbacks of schema validation, message transformation, and the rule engine look up their ETS topic index tables without checking that the tables exist. When a table is absent, every publish raises `badarg` and floods the log with `hook_callback_exception` errors.

This PR makes four changes:

1. Guard all three lookup sites (`emqx_schema_validation_registry:matching_validations/1`, `emqx_rule_engine:get_rules_for_topic/1`, `emqx_message_transformation_registry:matching_transformations/1`). Each site checks the table with `ets:whereis/1` and returns no matches when the table is absent.
2. Log a throttled `topic_index_table_missing` message when the guard fires, so the fault that destroyed the table stays visible: error level once the node is ready (`emqx_node_readiness:is_ready/0`), warning level during boot. Throttled per table name.
3. Move the schema validation and message transformation table creation from the registry gen_server into the supervisor init, matching the rule engine layout. The tables now exist before any child starts, are deleted only when the supervisor exits, and survive registry worker restarts.
4. Unregister the hooks in `prep_stop/1` instead of `stop/1` in all three applications, and move config unload there too.

Change 4 fixes a probable mechanism behind the reported flood. OTP calls `stop/1` only after the application's supervision tree is terminated, and wraps the call in a bare `catch`. For schema validation, `stop/1` started with config unload, which makes gen_server calls to the registry; the registry is already dead at that point, so with at least one validation configured `stop/1` crashed silently and the hook was never unregistered. Every publish after that (broker-internal publishers such as the `$SYS` timers keep running until the `emqx` application stops) hit the destroyed index table. Message transformation leaked its config handler the same way, and the rule engine unregistered its event hooks only after its tables died. `prep_stop/1` runs before the tree is terminated, so the hooks and the loaded config are removed while the registry and its tables are still alive.

Decisions taken:

- The guard is per caller, not inside `emqx_topic_index:matches/3`. The router (`emqx_router:match_routes/1`), the durable session router, cluster link, and the MQ registry share that module. For those consumers, a missing table silently returning no routes would turn a serious fault into silent message loss. For the three hook callbacks, "no matches" is the correct meaning.
- `ets:whereis/1` instead of `try ... catch error:badarg`. The catch would also swallow genuine badargs from other causes.
- The log is throttled (`?SLOG_THROTTLE`) because the hook runs on every publish; an unthrottled line per occurrence would recreate the flood this issue is about.
- The guard stays even with the lifecycle fixes: hook deregistration can never be atomic with in-flight `emqx_hooks:run_fold` calls, cluster join/leave restarts these applications at runtime, and the reporter's observation of an absent table on a running node is still not fully explained.

This complements #18357 (gate client connections until boot completes) and #18523 (stop listeners before applications on shutdown) rather than replacing them: #18523 removes client-driven publishes from the shutdown window, while broker-internal publishers (`emqx_sys` timers, delayed publish, rule engine republish) run the publish hook regardless of listeners.

## PR Checklist


apps/emqx_message_transformation/src/emqx_message_transformation_registry.erl
```diff
@@ -104,6 +104,16 @@ delete(Transformation, Pos) ->
 %% @doc Returns a list of matching transformation names, sorted by their configuration order.
 -spec matching_transformations(emqx_types:topic()) -> [transformation()].
 matching_transformations(Topic) ->
+    %% The index table is absent while the registry owner process is down; return no
+    %% matches instead of crashing the `message.publish' hook.
+    case ets:whereis(?TRANSFORMATION_TOPIC_INDEX) of
+        undefined ->
```
keynslug, line 126: Hard to see if there are regular circumstances where this is possible: `emqx_message_transformation_app` correctly sets hooks up at the very end of the boot sequence, and unregisters them at the very beginning of `stop/1`. And if it's about some unexpected system states / error conditions, wouldn't it be helpful instead to see errors in the logs?

zmstone: changed to log at error level (throttled).

zmstone: warning level if it's during bootup/shutdown

thalesmg:
maybe this is superseded by not closing the listeners when restarting: https://github.com/emqx/emqx/pull/18523

zmstone:
> maybe this is superseded by not closing the listeners when restarting: #18523

yes, that's the hope.
I had the impression that the features can be toggled (table create/delete) at runtime -- apparently that's not the case.

zmstone:
> > maybe this is superseded by not closing the listeners when restarting: #18523
> 
> yes, that's the hope. I had the impression that the features can be toggled (table create/delete) at runtime -- apparently that's not the case.

ah, forgot there is SYS message etc triggered internally.

thalesmg:
> > > maybe this is superseded by not closing the listeners when restarting: #18523
> > 
> > 
> > yes, that's the hope. I had the impression that the features can be toggled (table create/delete) at runtime -- apparently that's not the case.
> 
> ah, forgot there is SYS message etc triggered internally.

true, also integration sources

thalesmg (approved):

PR #18528, opened 2026-08-26
chore(otel schema): enforce scheme and port validation for endpoint

### LINKED ISSUE #18497: [6.3.0] OpenTelemetry exporter endpoint accepts string without URL scheme
### What happened?

The OpenTelemetry exporter `endpoint` field accepts any string without validating that it contains a URL scheme (e.g., `http://` or `https://`). An invalid value like `"not-a-valid-url"` is persisted successfully by the API.

In `emqx_otel_schema.erl`, the endpoint field is defined as:

```erlang
{endpoint,
    ?HOCON(
        emqx_schema:url(),  %% type url() :: binary() — no URL scheme validation
        #{
            default => <<"http://localhost:4317">>,
            ...
        }
    )}
```

The `emqx_schema:url()` type is simply `binary()`, so no URL format validation is performed.

### Reproduction Steps

1. Send a configuration to the API:
   ```
   PUT /api/v5/opentelemetry
   ```
2. Set the exporter endpoint to:
   ```json
   {
     "exporter": {
       "endpoint": "not-a-valid-url"
     }
   }
   ```
3. Query the configuration:
   ```
   GET /api/v5/opentelemetry
   ```

### Expected Behavior

The API should reject the configuration with HTTP 400 because the endpoint must be an absolute URL with a valid scheme (`http://` or `https://`). For example, `http://not-a-valid-url` is syntactically valid and may be accepted, while `not-a-valid-url` should be rejected.

### Actual Behavior

The API returns HTTP 200 and persists:
```json
"endpoint": "not-a-valid-url"
```

### Additional Observation

The Dashboard correctly rejects this value, indicating inconsistent validation between the Dashboard and the backend API schema.

### Impact

Invalid exporter configuration can be persisted through the API and may only fail later during telemetry export, resulting in delayed runtime errors and loss of logs/traces.

### Environment

- **EMQX version**: 6.3.0-beta.2
- **Component**: OpenTelemetry exporter
- **API**: `PUT /api/v5/opentelemetry`
- **Source**: `apps/emqx_opentelemetry/src/emqx_otel_schema.erl`

thalesmg:
Fixes https://github.com/emqx/emqx/issues/18497

Release version:

Introduced in:
pre-5.8

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -3570,6 +3570,12 @@ servers_validator(Opts, Required) ->
             Str when SingleServer ->
                 %% it's valid as long as it can be parsed
                 _ = parse_server(Str, Opts),
+                case string:find(Str, ",") of
+                    nomatch ->
+                        ok;
+                    _ ->
+                        throw("expecting_single_server_but_found_comma")
+                end,
                 ok;
             Str ->
                 %% it's valid as long as it can be parsed
--- a/apps/emqx/src/emqx_schema.erl
+++ b/apps/emqx/src/emqx_schema.erl
@@ -3504,12 +3504,13 @@ servers_sc(Meta0, ParseOpts) ->
     %% NOTE: maps:is_key is not the solution because #{default => undefined} is legit
     HasDefault = (maps:get(default, Meta0, undefined) =/= undefined),
     Required = maps:get(required, Meta0, not HasDefault),
+    Type = maps:get(type, Meta0, string()),
     Meta = #{
         required => Required,
         converter => fun convert_servers/2,
         validator => servers_validator(ParseOpts, Required)
     },
-    sc(string(), maps:merge(Meta, Meta0)).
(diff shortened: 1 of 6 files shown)


apps/emqx/src/emqx_schema.erl
```diff
@@ -3565,6 +3567,10 @@ servers_validator(Opts, Required) ->
                 throw("cannot_be_empty");
             "undefined" ->
                 ok;
+            Str when SingleServer ->
+                %% it's valid as long as it can be parsed
+                _ = parse_server(Str, Opts),
```
zmstone, line 3572: previously, `url()` converter rejected `http://localhost:4317,`.
now it's valid for parse_server/2, and keeping the raw will cause crash ?

thalesmg: good catch, added a test case to forbid commas in the validator

zmstone (approved):

PR #18535, opened 2026-08-27
fix(mt): follow namespace change when a session is resumed

### LINKED ISSUE #18533: [6.3.0] Namespace client list is not updated after a persistent session moves between namespaces
### What happened?

When the same MQTT client ID reconnects from namespace `ns.1` to namespace `ns.2`, the global client information is updated to `ns.2`, but the namespace client list for `ns.2` does not contain the client.

The client uses `clean_start=false`, so EMQX resumes the existing persistent session. The response shows that `connected_at` changes while `created_at` remains unchanged.

### Reproduction Steps

1. Create namespaces `ns.1` and `ns.2`.
2. Connect an MQTT client with:
   - Client ID: `conn-1`
   - Username / namespace: `ns.1`
   - `clean_start=false`
3. Verify that the client is listed by:

```text
GET /api/v5/mt/ns/ns.1/client_list
```

4. Disconnect the client.
5. Reconnect the same client ID with:
   - Client ID: `conn-1`
   - Username / namespace: `ns.2`
   - `clean_start=false`
6. Query the client:

```text
GET /api/v5/clients/conn-1
```

7. Query the namespace client list:

```text
GET /api/v5/mt/ns/ns.2/client_list
```

### Expected Behavior

After the client reconnects to `ns.2`, it should be listed in `ns.2/client_list` and should no longer be listed in `ns.1/client_list`.

### Actual Behavior

The global client endpoint reports the new namespace, for example:

```json
{
  "clientid": "conn-1",
  "username": "ns.2",
  "client_attrs": {
    "tns": "ns.2"
  },
  "mountpoint": "ns.2/",
  "clean_start": false
}
```

However, the following endpoint returns an empty list:

```text
GET /api/v5/mt/ns/ns.2/client_list?limit=20
```

### Technical Analysis

The namespace client index is populated from the `session.created` hook. When `clean_start=false`, the reconnect resumes the existing persistent session instead of creating a new one. The namespace index is not updated when the session's namespace changes, leaving the client associated with the original namespace in the namespace index.

### Environment

- **EMQX version**: 6.3.0
- **API**: `GET /api/v5/mt/ns/:ns/client_list`
- **Protocol**: MQTT 5
- **clean_start**: `false`
- **Namespaces**: `ns.1`, `ns.2`

zmstone:
Release version: 6.3.0, 7.0.0

Introduced in: 5.9.0

## Summary

Fixes #18533.

The multi-tenancy client index (`emqx_mt`) was populated only from the `session.created` hook (`emqx_mt_hookcb`). A client that resumed an existing session (`clean_start=false`) fired `session.resumed` instead, so the index was never updated. When the client's namespace changed between connections, `GET /api/v5/mt/ns/{new-ns}/client_list` stayed empty and the client remained listed under the old namespace, while `GET /api/v5/clients/{clientid}` correctly showed the new `client_attrs.tns`.

The fix registers the client on `session.resumed` as well, through the same code path as `session.created`.

Why adding on resume is safe (no double listing):

* The index record key is `{Ns, ClientId, Pid}` and removal is driven by the channel-process `DOWN` monitor in `emqx_mt_pool`, which deletes exactly the dead pid's entry. The old and the new connection are different processes, so a late `DOWN` for the old channel cannot remove the new entry.
* Resuming a session always terminates the previous channel process (takeover), so the stale entry is removed by the same mechanism the `clean_start=true` path already relies on. The transient window where both entries exist is the same as in the existing `clean_start=true` reconnect flow.

This also fixes durable sessions: their channel process dies at disconnect (removing the index entry), and resuming fired only `session.resumed`, so the client was missing from the namespace client list after every durable-session resume, even within the same namespace.

Manual verification of the reported steps on this branch (`mqtt.client_attrs_init = [{expression = username, set_as_attr = tns}]`; connect `conn-1` as `ns.1` with MQTT 5, `clean_start=false`, session-expiry 300; disconnect; reconnect as `ns.2`):

```
### GET /api/v5/mt/ns/ns.1/client_list (after first connect):
["conn-1"]
### GET /api/v5/clients/conn-1 (after reconnect under ns.2):
{"clientid":"conn-1","tns":"ns.2"}
### GET /api/v5/mt/ns/ns.2/client_list:
["conn-1"]
### GET /api/v5/mt/ns/ns.1/client_list:
[]
```

## PR Checklist

### DIFF
--- a/apps/emqx_mt/src/emqx_mt_hookcb.erl
+++ b/apps/emqx_mt/src/emqx_mt_hookcb.erl
@@ -9,6 +9,7 @@
     register_hooks/0,
     unregister_hooks/0,
     on_session_created/2,
+    on_session_resumed/2,
     on_authenticate/2,
     on_post_authn/1,
     on_api_actor_will_be_created/2,
@@ -25,6 +26,7 @@
 
 -define(TRACE(MSG, META), ?TRACE("MULTI_TENANCY", MSG, META)).
 -define(SESSION_HOOK, {?MODULE, on_session_created, []}).
+-define(SESSION_RESUMED_HOOK, {?MODULE, on_session_resumed, []}).
 -define(AUTHN_HOOK, {?MODULE, on_authenticate, []}).
 -define(POST_AUTHN_HOOK, {?MODULE, on_post_authn, []}).
 -define(LIMITER_HOOK(LEVEL), {emqx_mt_limiter, adjust_limiter, [LEVEL]}).
@@ -34,6 +36,7 @@
 
 register_hooks() ->
     ok = emqx_hooks:add('session.created', ?SESSION_HOOK, ?HP_HIGHEST),
+    ok = emqx_hooks:add('session.resumed', ?SESSION_RESUMED_HOOK, ?HP_HIGHEST),
     ok = emqx_hooks:add('client.authenticate', ?AUTHN_HOOK, ?HP_HIGHEST),
     ok = emqx_hooks:add('client.post_authn', ?POST_AUTHN_HOOK, ?HP_HIGHEST),
     ok = emqx_hooks:add('channel.limiter_adjustment', ?LIMITER_HOOK(channel), ?HP_HIGHEST),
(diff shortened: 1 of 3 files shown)


zmstone:
will be ported from 6.0 to 6.3 later after https://github.com/emqx/emqx/pull/18539 merge

PR #18542, opened 2026-08-27
fix(topic_metrics): remove fallible audit emission from reset cluster_rpc transaction

### LINKED ISSUE #18534: [6.3.0] Topic metrics reset returns success but fails during cluster RPC audit logging
### What happened?

When audit logging is enabled, resetting a topic metrics collection through the REST API returns success, but the cluster RPC reset fails while writing the audit log. As a result, topic metrics may only be reset on some nodes, leaving non-zero values in the cluster.

### Reproduction Steps

1. Run EMQX 6.3.0 as a cluster.
2. Enable audit logging.
3. Create a topic metrics collection named `test`.
4. Generate traffic matching the collection so that its counters contain non-zero values.
5. Reset the collection:

```bash
curl --request PUT \
  --url 'http://<emqx-host>:18083/api/v5/mqtt/topic_metrics2/test/reset' \
  --header 'Authorization: Bearer <token>'
```

6. Check the EMQX logs and query the topic metrics on all cluster nodes.

### Expected Behavior

- The reset API should return success only when the reset has been successfully applied across the cluster.
- All counters for the collection should be reset on every cluster node.
- Audit logging should not cause the reset operation to fail.

### Actual Behavior

The API returns success, but the cluster logs contain an error similar to:

```text
cluster_rpc_failed_to_init_transaction
result: {error,#{reason => {badmatch,#{
  args => [<<"test">>,global],
  time => ..., level => info,
  cmd => topic_metrics_reset,
  from => cluster_rpc,
  kind => initiate,
  duration_ms => 0
}}}}
```

The stack trace points to:

```text
emqx_audit:to_audit/1
emqx_audit:log_to_db/1
emqx_topic_metrics_registry:emit_reset_audit/2
```

The reset callback clears local counters before emitting the audit event. Because the audit event generated with `from => cluster_rpc` does not contain all fields expected by `emqx_audit:to_audit/1`, audit logging raises a `badmatch` and the cluster reset is not completed consistently.

The observed result is that some metrics, such as received-message counters, are reset, while other counters or values from other cluster nodes remain non-zero.

### Triggering Condition

The issue is reproducible when audit logging is enabled and the reset operation uses the cluster RPC path. It may not be reproducible when audit logging is disabled.

### Environment

- **EMQX version**: 6.3.0
- **Deployment**: EMQX cluster
- **API**: `PUT /api/v5/mqtt/topic_metrics2/:name/reset`
- **Topic metrics collection**: `test`
- **Audit logging**: enabled

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:
N/A

## Summary

Fixes #18534.

Resetting a topic-metrics collection returned HTTP success, but the cluster_rpc transaction died inside audit logging, so counters were reset on some nodes only.

Root cause: `emqx_topic_metrics_registry:do_reset_local/2` emitted a per-node audit event with `from => cluster_rpc` from inside the cluster_rpc transaction. No `emqx_audit:to_audit/1` clause matched such events, so the generic clause raised `badmatch` on the missing HTTP-request fields. The counters were cleared before the emission, so the local side effect landed and then the transaction failed: on the initiating node the abort meant the transaction was never committed, so the other nodes never applied it; a follower that still had the audit handler installed kept replaying and failing a committed transaction (`kind => replicate` in the report).

The reporter reproduced this after enabling and then disabling audit: disabling intentionally leaves the logger handler installed until the next audit event is logged (so the disable action itself is audited, see `emqx_config_logger:update_log_handler/1`), and the crash happened before that deferred removal, keeping the handler installed forever.

Changes:

* `emqx_topic_metrics`: remove the per-node audit emission from the reset cluster_rpc callback. The reset request is already audited at the REST API entrypoint (`emqx_dashboard_audit`) with the actor, source IP and operation id, and reset has no other entrypoint. The per-node event carried no actor context, and it put a fallible logging call inside the cluster_rpc transaction. The callback is side-effect-only now.
* `emqx_audit`: a `log_to_db/1` crash no longer propagates to the `?AUDIT` call site. Audit events are emitted from cluster_rpc transactions and CLI commands, where a raise breaks the audited operation itself, and the crash also skipped the deferred handler removal. The failure is logged with a stacktrace instead.

Tests: a REST regression case in `emqx_topic_metrics2_api_SUITE` (the suite now runs with audit logging enabled; the reset transaction must commit), and a case in `emqx_audit_api_SUITE` asserting a malformed audit event does not crash the `?AUDIT` caller.

No changelog: the affected code first ships in 6.3.0, which has no GA release yet.

## PR Checklist


apps/emqx_audit/src/emqx_audit_api.erl
```diff
@@ -64,7 +64,7 @@ schema("/audit") ->
                         desc => ?DESC(filter_node)
                     })},
                 {from,
-                    ?HOCON(?ENUM([dashboard, rest_api, cli, erlang_console]), #{
+                    ?HOCON(?ENUM([dashboard, rest_api, cli, erlang_console, cluster_rpc]), #{
```
zmstone: delete cluster_rpc.

thalesmg (approved):

PR #18557, opened 2026-08-27
feat(config): add node.default_listener_address

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:

## Summary

Adds `node.default_listener_address`, which sets the address of listener binds that have no explicit address, that is bare-port binds such as `bind = 1883`. Motivation: #18511 makes `hardened` the default in 7.0, which binds those listeners to loopback.

Values: `loopback` | `nodename` (the host part of the Erlang node name) | `all` | an IP address | a hostname. When not set, the security profile decides as before. An explicit `IP:port` bind always wins.

Applies to MQTT listeners, gateway listeners and the dashboard HTTP listener. The option is node-local and boot-only, and can be set with `EMQX_NODE__DEFAULT_LISTENER_ADDRESS`. The docker entrypoint sets `all`, because a container's loopback is unreachable through published ports.

Schema defaults stay static bare ports: the schema only validates the value, and the address is applied in runtime code when a listener starts. So the configuration keeps showing the configured bind, and each node resolves its own address.

## PR Checklist


apps/emqx_conf/etc/emqx.env
```diff
@@ -64,3 +64,25 @@
...
+#               no IPv4 address resolves. Note: on Debian-family systems
+#               /etc/hosts often maps the hostname to 127.0.1.1, so
+#               hostname_i binds that address; fix /etc/hosts or set a
+#               literal address in that case.
+#   all         bind 0.0.0.0, that is all IPv4 interfaces.
+#   <address>   a literal IPv4 or IPv6 address, for example 192.168.1.10 or
+#               ::1. The address :: accepts IPv4 and IPv6 connections on
+#               most systems; the OS bindv6only setting decides.
+# When the variable is not set, the security profile decides the default
+# address: legacy binds all interfaces, hardened binds loopback.
```
zmstone:
```suggestion
# address: 'legacy' profile binds all interfaces, 'hardened' binds loopback.
```

apps/emqx/src/emqx_default_address.erl
```diff
@@ -0,0 +1,125 @@
...
+%%--------------------------------------------------------------------
+
+-module(emqx_default_address).
+
+-moduledoc """
+This module resolves the `EMQX_DEFAULT_ADDRESS` boot environment variable,
+which sets the address of defaulted listener binds independently of the
+security profile.
+
+Valid values are `loopback`, `hostname_i`, `all`, or a literal IPv4/IPv6
```
zmstone:
```suggestion
Valid values are `loopback`, `nodename`, `all`, or a literal IPv4/IPv6
```

zmstone:
cannot support nodename, because the config is synced in cluster.

apps/emqx_conf/src/emqx_conf_schema.erl
```diff
@@ -564,6 +564,17 @@ fields("node") ->
                     desc => ?DESC(node_name)
                 }
             )},
+        {"default_listener_address",
+            sc(
+                string(),
```
zmstone: use binary() type whenever possible.

apps/emqx/src/emqx_default_address.erl
```diff
@@ -0,0 +1,220 @@
...
+-define(CONF_NAME, "node.default_listener_address").
+
+-export([resolve/1, listen_on/2, validate/1, clear/0]).
+
+-export_type([address/0, scope/0]).
+
+-type address() :: any | loopback | inet:ip_address().
+-type scope() :: mqtt | dashboard | gateway.
+-type value() ::
+    default | loopback | all | nodename | {hostname, string()} | inet:ip_address().
```
zmstone:
```suggestion
    per_security_policy | loopback | all | nodename | {hostname, string()} | inet:ip_address().
```

rel/i18n/emqx_conf_schema.hocon
```diff
@@ -149,6 +149,31 @@ node_name.desc:
...
+- <code>nodename</code>: take the host part of the Erlang node name (after
+the @). When it is an IP address, bind it; otherwise resolve it at boot and
+bind the first IPv4 address, or the first IPv6 address when no IPv4 address
+resolves.<br/>
+- <code>all</code>: bind 0.0.0.0, that is all IPv4 interfaces.<br/>
+- A literal IPv4 or IPv6 address, for example <code>192.168.1.10</code> or
+<code>::1</code>. The address <code>::</code> accepts IPv4 and IPv6
+connections on most systems; the OS bindv6only setting decides.<br/>
+- A hostname, for example <code>broker1.example.com</code>, resolved at boot
+like the nodename host.<br/>
```
zmstone:
```suggestion
like the node name host.<br/>
```

apps/emqx_gateway/src/emqx_gateway_utils_conf.erl
```diff
@@ -299,7 +299,7 @@ to_rt_listener_ids(GwName, GwConfig0) ->
     lists:map(
         fun({Type, Name, ListenerConfig}) ->
             ListenerId = listener_id(GwName, Type, Name),
-            ListenOn = maps:get(bind, ListenerConfig),
+            ListenOn = emqx_default_address:listen_on(gateway, maps:get(bind, ListenerConfig)),
```
zmstone, line 302: This normalizes the runtime IDs used by gateway start/stop, but `/gateways/:name/listeners` still builds its cluster-status `ListenOn` with `emqx_gateway_conf:get_bind/1` and then calls `esockd:get_current_connections({Id, ListenOn})`. With `node.default_listener_address = all` (now the Docker default), a bare TCP/UDP/etc gateway listener is registered under `{{0,0,0,0}, Port}` while that API still queries `Port`, so it reports `running=false`/`current_connections=0`. 

should either change `all` to just the port (no IP),
or normalize the API path too.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -819,6 +829,26 @@ ip_port(Port) when is_integer(Port) ->
...
+that is not running yet.
+""".
+listen_on(Id, Bind) ->
+    case find_listen_on(Id) of
+        {ok, ListenOn} -> ListenOn;
+        error -> emqx_default_address:listen_on(mqtt, Bind)
+    end.
+
+find_listen_on(Id) ->
+    case [ListenOn || {{I, ListenOn}, _Pid} <- esockd:listeners(), I =:= Id] of
```
zmstone: `lists:mapfilter` with `fun({{I, ListenOn}, _Pid})` pattern match, because `<-` silently dropps unmatched.

apps/emqx/src/emqx_default_address.erl
```diff
@@ -0,0 +1,233 @@
...
+%% Internal functions
+%%--------------------------------------------------------------------
+
+profile_policy(mqtt) ->
+    emqx_security_profile:policy(mqtt_default_bind);
+profile_policy(dashboard) ->
+    emqx_security_profile:policy(dashboard_http_default_bind);
+profile_policy(gateway) ->
+    %% The security profile does not cover gateway binds.
+    any.
```
thalesmg, line 117: q: shouldn't a clause be added for it?

apps/emqx_dashboard/test/emqx_dashboard_listener_config_SUITE.erl
```diff
@@ -16,16 +16,18 @@ all() ->
...
 
 init_per_suite(Config) ->
     emqx_common_test_helpers:clear_security_profile(),
+    emqx_common_test_helpers:clear_default_address(),
     Config.
 
 end_per_suite(_Config) ->
-    emqx_common_test_helpers:clear_security_profile().
+    emqx_common_test_helpers:clear_security_profile(),
+    emqx_common_test_helpers:clear_default_address().
```
thalesmg: probably should add this to emqx_cth_suite:clean_suite_state

apps/emqx/test/emqx_default_address_SUITE.erl
```diff
@@ -0,0 +1,301 @@
...
+        Class:Reason:Stacktrace ->
+            %% end_per_group does not run when init fails; do not leak the
+            %% profile into later test cases.
+            emqx_common_test_helpers:clear_security_profile(),
+            erlang:raise(Class, Reason, Stacktrace)
+    end.
+
+end_per_group(_Profile, Config) ->
+    emqx_cth_suite:stop(?config(apps, Config)),
+    emqx_common_test_helpers:clear_security_profile().
```
thalesmg, line 53: nit: this is done by emqx_cth_suite:clear_suite_state

thalesmg (dismissed):

apps/emqx/src/emqx_default_address.erl
zmstone: will do it in a follow-up pr.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -196,7 +189,8 @@ current_conns(Id, ListenOn) ->
     {ok, #{type := Type, name := Name}} = parse_listener_id(Id),
     current_conns(Type, Name, ListenOn).
 
-current_conns(Type, Name, ListenOn) when Type == tcp; Type == ssl ->
+current_conns(Type, Name, Bind) when Type == tcp; Type == ssl ->
+    ListenOn = emqx_default_address:listen_on(mqtt, Bind),
```
savonarola, line 193: ~~Wondering: If we change  the `default_address` setting on the fly, will we still be able to locate the listener?~~

zmstone: it's under the 'node' root -- all configs under it are read-only.
let me double check the `/config` api endpoint though.

zmstone: right, 

API:
https://github.com/emqx/emqx/blob/744e675b57d38bc892bb49fe3ed9fecb240c5386/apps/emqx_management/src/emqx_mgmt_api_configs.erl#L317-L336

CLI:
https://github.com/emqx/emqx/blob/716b604415adc093d5eb57b837914c0b80231681/apps/emqx_conf/src/emqx_conf_cli.erl#L697

apps/emqx/test/emqx_default_address_SUITE.erl
zmstone: this is end_per_group though.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -488,11 +509,15 @@ do_start_listener(quic, Name, Id, #{bind := Bind} = Opts) ->
...
-    Type = tcp, Name, OldConf, NewConf = #{bind := ListenOn, tcp_backend := Backend}
+    Type = tcp, Name, OldConf, NewConf = #{bind := Bind, tcp_backend := Backend}
 ) ->
     Id = listener_id(tcp, Name),
-    case OldConf of
-        #{bind := ListenOn, tcp_backend := Backend} ->
+    %% Compare the address the listener runs on with the one the new config
+    %% resolves to. The configured bind alone does not tell which address a
+    %% running listener uses, because a bare port gets the default address.
+    ListenOn = listen_on(Id, Bind),
```
savonarola: The logic is a bit hard to grasp. Here we are already assume that the listener _is_ running, but we still call `listen_on(Id, Bind)` which handles not running listeners. Maybe call `find_listen_on` here.
This will also handle a (very low possible) situation when listener concurrently stopped before we reach this point.

savonarola (dismissed):

apps/emqx/test/emqx_default_address_SUITE.erl
thalesmg: and yet, stop is being called here, which cleans the state.

zmstone: emqx_cth_suite:stop is not emqx_cth_suite:clean_suite_state

apps/emqx/src/emqx_default_address.erl
zmstone: actually, no need to fix, the profile is only needed for listeners started by default.
for gateway listeners, it's admin's responsibility to bind a secure address no matter which profile it is running on.

apps/emqx/test/emqx_default_address_SUITE.erl
thalesmg: stop calls it...

https://github.com/emqx/emqx/blob/398ec0cde73b3defe69898d6c1f91a593dd0375e/apps/emqx/test/emqx_cth_suite.erl#L468-L470

zmstone: oh right.

then there is 
`os:unsetenv(?SECURITY_PROFILE_ENV_VAR)` not covered in suite end.

https://github.com/emqx/emqx/blob/a39a46b7812f94eab71abd93bc69c62a14af5319/apps/emqx/test/emqx_common_test_helpers.erl#L1406-L1409

thalesmg: indeed, makes sense it should be added there as well

thalesmg (approved):

PR #18560, opened 2026-08-27
test(ps ds): attempt to stabilize flaky test

thalesmg:
See also https://github.com/emqx/emqx/pull/17882#issuecomment-5444250268


zmstone (approved):

thalesmg:
D'oh, this "fix" actually does not affect the object of the sharding...  :melting_face: :see_no_evil:

thalesmg:
maybe the reason is actually the subscription taking a while to set in, so we miss the first message?

PR #18564, opened 2026-08-27
test(ps ds): attempt to stabilize flaky test

thalesmg:
https://github.com/emqx/emqx/actions/runs/33096586357/job/98626235934#step:5:537

### DIFF
--- a/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_sup.erl
+++ b/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_sup.erl
@@ -63,7 +63,7 @@ init(top) ->
             type => worker,
             start => {emqx_persistent_message, start_link, []},
             restart => transient,
-            shutdown => 100
+            shutdown => 1000
         }
     ],
     {ok, {SupFlags, Children}};
--- a/apps/emqx/test/emqx_persistent_session_ds_SUITE.erl
+++ b/apps/emqx/test/emqx_persistent_session_ds_SUITE.erl
@@ -1280,6 +1280,8 @@ t_session_replay_retry(_Config) ->
             ClientId = mk_clientid(?FUNCTION_NAME, sub),
             NClients = 10,
             ClientSubOpts = #{
+                %% important to use v5 as we rely on the reason code
+                proto_ver => v5,
                 clientid => ClientId,
                 auto_ack => never
             },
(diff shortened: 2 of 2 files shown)


apps/emqx/test/emqx_persistent_session_ds_SUITE.erl
```diff
@@ -1306,17 +1313,46 @@ t_session_replay_retry(_Config) ->
...
+            }),
 
             ok = emqtt:stop(ClientSub),
 
             %% 4. Make `emqx_ds` believe that roughly half of the
             %% shards are unavailable.
             ?tp(notice, test_restart_client, #{}),
             meck:new(emqx_ds, [passthrough, no_history]),
+            %% we need to ensure _at least one_ iterator fails.  otherwise, the test might
+            %% get flaky and fail bellow with length(Pubs1) == length(Pubs0).
```
thalesmg, line 1347: observed this locally while trying to reproduce the failure above

apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_sup.erl
```diff
@@ -63,7 +63,7 @@ init(top) ->
             type => worker,
             start => {emqx_persistent_message, start_link, []},
             restart => transient,
-            shutdown => 100
+            shutdown => 1000
```
thalesmg, line 66: I haven't yet found the actual cause for the missing messages...  

~~But since making this shutdown timeout adjustment, I have run the test case [several times in a row in ci](https://github.com/emqx/emqx/actions/runs/33534715114?pr=18564) and it has not failed yet.  I wonder if the 100 ms timeout was too tight for CI, which is quite constrained, and then the teardown followed by start apps again ended up brutally killing `emqx_persistent_message`, which in turn didn't clear the `?optvar_ready` optvar, and the db was "ready" before actually setting everything up?~~

Apparently, even though I didn't update my branch, it was running in a merge commit and it picked up the flaky test retry patch...

thalesmg: Even after disabling the auto retries, I could not observe `t_session_replay_retry` break [after 48 consecutive runs](https://github.com/emqx/emqx/actions/runs/33665684290?pr=18564).

Maybe it actually worked?

PR #18576, opened 2026-08-28
fix(sso): do not redact oidc client_jwks when it is `none`

### LINKED ISSUE #18569: [6.3.0] OIDC API incorrectly masks client_jwks = none as ******
## Bug Title

OIDC API incorrectly masks client_jwks = none as ******

## Description

The Dashboard OIDC configuration API incorrectly redacts the non-sensitive value client_jwks = none.

When the OIDC configuration is queried through the REST API, client_jwks is returned as ******, while the CLI correctly returns the configured value none.

## Environment

- EMQX branch/version: release-63
- API endpoint: GET /api/v5/sso/oidc

## Steps to Reproduce

1. Configure Dashboard OIDC with:

```hocon
dashboard.sso.oidc {
  enable = false
  client_jwks = none
}
```

2. Query the configuration through the REST API:

```sh
curl --url 'http://<emqx-host>:18083/api/v5/sso/oidc' \
  -H 'Authorization: Bearer <token>'
```

3. Query the configuration through the CLI:

```sh
emqx ctl conf show dashboard
```

## Actual Result

REST API:

```json
{
  "client_jwks": "******"
}
```

CLI:

```hocon
client_jwks = none
```

## Expected Result

The REST API should return the actual non-sensitive value:

```json
{
  "client_jwks": "none"
}
```

Sensitive values such as `secret` should remain masked.

## Impact

This causes inconsistent OIDC configuration results between the REST API and CLI and prevents clients from determining whether `client_jwks` is disabled or configured with a JWKS source.

zmstone:
Release version:

Introduced in:
6.1.3

## Summary

NOTE: the reveal of `none` is not considered a vulnerability.

Fixes #18569.

`GET /api/v5/sso/oidc` returned `"client_jwks": "******"` even when the value was the atom `none` (the default, meaning no client JWKS is configured). The CLI (`emqx ctl conf show dashboard`) shows `client_jwks = none`, so the API and CLI disagreed and a caller could not tell whether a client JWKS is configured.

#17739 added `client_jwks` to the sensitive keys in `emqx_utils_redact`, and redaction there is key-based: once the key matches, the value is masked whatever it is. But the schema makes `client_jwks` a union of the atom `none` and a `client_file_jwks` object; only the object holds key material.

This PR narrows the rule instead of removing it: `redact_v/2` passes the value through when the key is `client_jwks` and the value is `none` (atom, or the raw-config string forms). Any other value, including a configured file JWKS object, stays masked. The exemption is scoped to the `client_jwks` key, so redaction behaviour for every other sensitive key is unchanged.

Dropping the blanket key in favour of the inner fields was rejected: `client_file_jwks` has only `type` and `file`, neither is a sensitive key, and `file` carries the raw JWKS content in the update request body before it is written to disk, so the outer key must stay sensitive.

Relevant files:
- `apps/emqx_utils/src/emqx_utils_redact.erl` — key-scoped `none` pass-through in `redact_v/2`.
- `apps/emqx_utils/test/emqx_utils_redact_tests.erl` — `none` is not masked; a configured JWKS (checked and raw shapes) still is.
- `apps/emqx_dashboard_sso/test/emqx_dashboard_sso_oidc_SUITE.erl` — API-level test: GET returns `none` for the default config; a configured file JWKS and the client secret stay masked in GET and update responses.

## PR Checklist

### DIFF
--- a/apps/emqx_utils/src/emqx_utils_redact.erl
+++ b/apps/emqx_utils/src/emqx_utils_redact.erl
@@ -22,6 +22,10 @@
 
 -define(REDACT_VAL, "******").
 -define(IS_KEY_HEADERS(K), (K == headers orelse K == <<"headers">> orelse K == "headers")).
+-define(IS_KEY_CLIENT_JWKS(K),
+    (K == client_jwks orelse K == <<"client_jwks">> orelse K == "client_jwks")
+).
+-define(IS_VAL_NONE(V), (V == none orelse V == <<"none">> orelse V == "none")).
 
 redacted_value() -> <<?REDACT_VAL>>.
 
@@ -141,7 +145,7 @@ do_redact({Headers, Value}, _Checker) when ?IS_KEY_HEADERS(Headers) ->
 do_redact({Key, Value}, Checker) ->
     case Checker(Key) of
         true ->
-            {Key, redact_v(Value)};
+            {Key, redact_v(Key, Value)};
         false ->
             {do_redact(Key, Checker), do_redact(Value, Checker)}
     end;
@@ -162,7 +166,7 @@ do_redact(Headers, V, _Checker) when ?IS_KEY_HEADERS(Headers) ->
 do_redact(K, V, Checker) ->
     case Checker(K) of
         true ->
-            redact_v(V);
+            redact_v(K, V);
         false ->
             do_redact(V, Checker)
     end.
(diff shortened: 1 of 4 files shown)


apps/emqx_utils/src/emqx_utils_redact.erl
```diff
@@ -225,6 +229,13 @@ is_sensitive_header("x-auth-token") ->
 is_sensitive_header(_Any) ->
     false.
 
+%% `client_jwks' is a union of the atom `none' and a JWKS object; only the
+%% object holds key material. `none' (the default) must stay readable.
+redact_v(K, V) when ?IS_KEY_CLIENT_JWKS(K), ?IS_VAL_NONE(V) ->
+    V;
+redact_v(_K, V) ->
+    redact_v(V).
```
keynslug, line 237: Nit. Weird that this kind of knowledge gets on this level.

That said, maybe better to pair `is_sensitive_key/1` with `is_sensitive_value/2` to better contain redacting decisions?

keynslug (approved):

apps/emqx_utils/src/emqx_utils_redact.erl
zmstone: will pay the debt with a more elegant solution for secret redactions in general.

zmstone: for this one specifically, if we reveal file path instead, there is no need for redaction at all.

PR #18581, opened 2026-08-28
feat(listener): log when bind address is host-specific

zmstone:
Release version: 7.0.0

Introduced in:

## Summary

When a config change sets a listener `bind` to a host-specific IP address (not the wildcard and not loopback), log a portability notice:

- **warning** when the address is on the local node's interfaces and the cluster has one node: the listener works here, but the config breaks when migrated (config copy or backup restore) to another host.
- **error** when the address is not on the local interfaces, or the cluster has more than one node: the listener cannot bind the address on this node, or on some nodes of the cluster.

The check only logs and never fails the config change. `post_config_update` runs on every node that applies a replicated config change; rejecting there would block config replication across the cluster, which is a much worse failure than one listener that cannot bind. Each node evaluates the address against its own interfaces, so exactly the node that cannot bind it logs the error.

The check is hooked at the `post_config_update` level (per-listener create/update/action clauses and the root-key diff) rather than inside `start_listener`/`update_listener`: this logs exactly once per applied config change, and avoids repeating the warning on every boot and double-logging through the update → restart → start funnel. All of `127.0.0.0/8` counts as generic (loopback on every host).

Gateway and dashboard listener binds are out of scope; possible follow-up.

## PR Checklist


apps/emqx/test/emqx_listeners_update_SUITE.erl
```diff
@@ -458,8 +459,127 @@ t_delete_default_conf(Config) ->
...
     ok.
 
+-doc """
+Binding a listener to a real local non-loopback IP on a single-node cluster
+logs a portability warning; the config change succeeds and the listener runs.
+""".
+t_bind_portability_warning(_Config) ->
+    case local_non_loopback_addr() of
+        undefined ->
+            {skip, no_non_loopback_interface};
```
thalesmg: nit: this test case will report as ok rather than skipped when running without any non loopback addrs

apps/emqx/test/emqx_listeners_update_SUITE.erl
thalesmg, line 1: missing coverage for ipv6?

apps/emqx/test/emqx_listeners_update_SUITE.erl
```diff
@@ -460,16 +460,18 @@ t_delete_default_conf(Config) ->
...
+cluster logs a portability warning; the config change succeeds and the
+listener runs.
 """.
 t_bind_portability_warning(_Config) ->
-    case local_non_loopback_addr() of
+    case local_non_loopback_addr(inet) of
         undefined ->
-            {skip, no_non_loopback_interface};
+            ct:pal("host has no non-loopback IPv4 address, nothing to check"),
+            ok;
```
thalesmg: this test case will still report ok and print nothing, since it didn't fail

maybe make it skip in init_per_testcase so it's more clear when the environment is improper for it?

zmstone: changed to print a message and return ok

thalesmg: yes, that was what I said...

thalesmg (approved):

Issue #18363, opened 2026-08-17
Dashboard: make default_mfa a true global MFA enforcement (users cannot disable, rotation allowed)


zmstone:
### What would you like to be added or enhanced?

Make `dashboard.default_mfa` a **true account-wide enforcement** rather than a one-time default. When it is set (not `none`), every Dashboard user must have MFA and **cannot disable it**; only **rotation** (re-enrolling a new authenticator secret) stays allowed. An admin may still exempt specific accounts explicitly.

This is a behavior change to the meaning of `default_mfa`, so it targets **7.0.0**.

#### Current behavior (the gap)

Today `default_mfa` is a *soft default*, not a mandate:

- It only forces **initial** TOTP setup for users that have **no** MFA state. Once a user has any state, it is never re-applied.
- A user can **self-disable** their TOTP even after enrolling successfully (`do_disable_mfa` sets `mfa_state = disabled` for any non-disabled state, including a verified one), and because the global default skips users that already have a state, they are **not re-forced** at the next login. They stay off.
- Disable authorization never consults `default_mfa` — the only durable block is a per-user `admin_override = mfa_required`, and even that is bypassable by a self-caller holding the `mfa_management` scope (treated as self-exempt).

Net: setting `default_mfa = totp` globally does not actually guarantee that users keep MFA. Anyone can opt out permanently by disabling once. That makes it unsuitable as a compliance control.

#### Proposed behavior (7.0.0)

When `default_mfa != none` (global enforcement active):

1. **Users cannot disable MFA.** Self-disable is rejected regardless of the `mfa_management` scope self-exempt shortcut. The global enforcement must be checked independently of the existing per-user lock, so the `mfa_management` self-exempt path does not bypass it.
2. **Rotation stays allowed.** Re-enrolling a new authenticator secret (`reinit_mfa` / setup on an existing state) keeps MFA on, so it is permitted and encouraged (lost/replaced device).
3. **Re-force previously-disabled users, except exempted ones.** A user in `disabled` state is re-forced to enroll at next login, **unless** they carry `admin_override = mfa_exempted`. This closes the "disable once, escape forever" hole while honoring admin exemptions. Concretely, `maybe_init_mfa_state` must re-init `disabled` users when global enforcement is on, but skip those whose `admin_override_of/1` is `mfa_exempted`.
4. **Explicit admin exemption remains the only escape** — and it already exists (see below). No self-service exemption exists under enforcement.

Interaction with the existing per-user override stays coherent:

[...]

Issue #18362, opened 2026-08-17
Dashboard: encourage self-enabling TOTP 2FA (post-login status hint + one-click enrollment)


zmstone:
### What would you like to be added or enhanced?

Make Dashboard TOTP two-factor authentication (2FA) **discoverable and easy to self-enable**, without forcing it on globally.

The Dashboard has supported authenticator-app (TOTP) 2FA since 5.9.0 (#14584), but it is off by default and there is no in-product cue that it exists. Turning it on today means either editing `dashboard.default_mfa` config (which *forces* enrollment for everyone at next login — abrupt on upgrade) or knowing the `/users/:username/mfa` API exists. This proposal adds a light path in between: surface each user's MFA status after they log in, and let them switch TOTP on for themselves with one click — while leaving the global default unchanged.

This needs no new global config tier. A logged-in user can already self-enroll via `POST /users/:username/mfa` regardless of `dashboard.default_mfa`; the missing pieces are (1) telling the frontend the user's current MFA status and (2) a frontend affordance to act on it.

#### Proposed MFA status (derived, not a new config)

Expose a per-user status computed from existing state — user `mfa_state`, TOTP `first_verify_ts`, global `default_mfa`/security-profile enforcement, and the per-user admin-required override:

| Status | Meaning |
|---|---|
| `complete` | Enrolled and at least one successful TOTP verification. |
| `pending_enforced` | Enrollment required (global default or admin-required) and not yet complete — login is blocked until setup. |
| `pending_voluntary` | Not enrolled, not enforced — the user *may* enable it. |
| `disabled` | Explicitly opted out (and not enforced). |

`default_mfa` stays binary (`none | totp`); "enforced" may also come from a hardened security profile or the per-user admin-required override that already exists. No `optional`/`offered` config value is introduced.

#### Enforcement semantics — `default_mfa` is a soft default, not a durable mandate

This shapes the status derivation, so it is worth stating explicitly. The global `dashboard.default_mfa` setting only forces **initial** TOTP setup for users that have **no** MFA state yet. Once a user has any state — including an explicit `disabled` — the global default is not re-applied, so the user can self-disable and will not be re-forced at the next login. Disabling is also never gated on the global setting; the only per-user block on self-disable is an admin-set `admin_override = mfa_required` (and even that is bypassable by a self-caller who holds the `mfa_management` scope, who is treated as self-exempt).

Consequences:

[...]

PR #18586, opened 2026-08-28
feat(dashboard): enforce default_mfa for every account

MorganaFuture:
Release version: 7.0.0

Introduced in: N/A

## Summary

Fixes https://github.com/emqx/emqx/issues/18363

`dashboard.default_mfa` only seeded MFA for users that had none. `maybe_init_mfa_state/2`
matched `{ok, _}`, so a `disabled` state counted as "already handled", and a self-disable
writes exactly that state. Disabling once was permanent, which is what makes the setting
unusable as a compliance control.

`emqx_dashboard_sso_mfa:check_user_mfa_state/2` already draws the line for an SSO backend's
`force_mfa`: `disabled` bypasses only with `admin_override = mfa_exempted`, otherwise the
user is enrolled again. The local login path did not. Both now share one predicate,
`mfa_enforced_for/1`:

* `maybe_init_mfa_state/2` enrolls a `disabled` user again at the next login. Password
  login and SCRAM login both reach it through `verify_mfa_token/3`.
* `authorize_mfa_change/3` denies a self `disable` with 403 `MFA_ENFORCED` before the
  decision matrix, since `IsSelf AND HasMfaMgmt => allow` would otherwise let any
  `mfa_management` holder opt out. `setup` is untouched, so rotation stays available.

An admin disabling another user's MFA still writes `mfa_exempted` and that account is left
alone. `emqx ctl admins mfa <User> disable` never goes through the HTTP authorization
layer, so break-glass recovery keeps working while enforcement is on.

`t_enable_by_config` asserted the hole ("should be able to login without TOTP even though
default MFA is configured") and now asserts the mandate instead.

Left out on purpose, happy to add here or separately:

* step-up re-auth to disable. It needs a request body on an endpoint that has none today,
  and the issue says it can ship on its own.
* `pending_enforced` status. It adds a value to the `mfa` field of the users API that the
  dashboard has to understand, and #18362 is where the status vocabulary is being reworked.
* `default_mfa` at SSO login. `check_sso_mfa/2` still decides from the backend's own
  `force_mfa`. Whether a dashboard-level mandate should override a backend's
  `force_mfa = false` is your call, not something to settle here.

Tests: `emqx_dashboard_mfa_SUITE` 10/10, `emqx_dashboard_user_scopes_SUITE` 64/64,
`emqx_dashboard_rbac_SUITE` 32/32.

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ee/breaking-18586.en.md`
- [x] Schema changes are backward compatible or intentionally breaking (the meaning of `default_mfa` changes; the schema itself is unchanged, its description is updated)


rel/i18n/emqx_dashboard_schema.hocon
```diff
@@ -174,8 +174,11 @@ default_mfa.label:
 default_mfa.desc:
 """Default Multifactor authentication settings for Dashboard.
 When set as `none`, MFA can be enabled per user from the dashboard.
-If set as `{mechanism = totp}`, the user will be initialized with
-a random TOTP secret when logging in for the first time."""
+If set as `{mechanism = totp}`, MFA is required for every user: a user
```
zmstone: This says "MFA is required for every user" (and the changelog says it "forces every non-exempted user"). That holds for password and SCRAM login — both reach `maybe_init_mfa_state/2` via `verify_mfa_token/3` — but not for SSO login: `emqx_dashboard_sso_mfa:check_sso_mfa/2` decides purely from the backend's `force_mfa` and never consults `dashboard.default_mfa`. So an operator who uses SSO and sets `default_mfa` would reasonably believe MFA is now mandatory for everyone, which it is not. Either scope the wording (e.g. "required for every user who signs in with a Dashboard password; SSO logins follow the SSO backend's force_mfa") or extend the mandate to the SSO path. The PR description flags the SSO decision as open, so the user-facing text should not claim universal coverage until it is settled.

zmstone:
Thank you very much for the PR @MorganaFuture 

This one may take a while though.
We need to implement https://github.com/emqx/emqx/issues/18607 first.
If you want to take a stab, please go ahead, otherwise wait for it then rebase this one.

MorganaFuture:
Took a stab at it: https://github.com/emqx/emqx/pull/18608

rel/i18n/emqx_dashboard_schema.hocon
MorganaFuture: scoped the wording rather than extending the mandate, since the SSO decision is still open.

schema text now says "required for every user who signs in with a Dashboard password" and adds that SSO logins follow the backend force_mfa instead. changelog carries the same caveat.

worth noting the gap is wider than the text: with force_mfa = true and default_mfa = none, an SSO user who self-disabled reads back as disabled but is stopped for setup at the next login. that is on the sso side, not this pr, but the two paths never meet.

zmstone:
@MorganaFuture 
Now we can rebase ?

MorganaFuture:
done

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -1206,16 +1207,36 @@ do_verify_mfa_token(Username, MfaToken, IsPwdOk) ->
             ok
     end.
 
+%% @doc Whether `dashboard.default_mfa' requires MFA for this account.
+%% It applies to every account except the ones an admin has exempted by
+%% setting `admin_override' to mfa_exempted.
+-spec mfa_enforced_for(dashboard_username()) -> boolean().
+mfa_enforced_for(Username) ->
```
zmstone: This predicate does not distinguish SSO accounts, but `DELETE /current_user/mfa` runs it for them too (`with_caller/2` resolves the `?SSO_USERNAME` tuple, and `emqx_dashboard_sso_mfa:check_user_mfa_state/2` documents that SSO users self-DELETE their MFA).

With `default_mfa = {mechanism = totp}` and an SSO backend with `force_mfa = false`: an SSO user who self-enrolled gets `403 MFA_ENFORCED` ("MFA is required for every account by dashboard.default_mfa"), while the schema text now says the setting does not cover SSO logins, and their next SSO login indeed does not require MFA (`emqx_dashboard_sso_mfa:mfa_required_for_user/2` reads only the override and `force_mfa`). The login path and the disable path disagree about the same account.

Suggest scoping the predicate to local accounts so it matches the documented scope:

```erlang
mfa_enforced_for(?SSO_USERNAME(_, _)) ->
    %% SSO accounts follow the backend's force_mfa and the admin override.
    false;
mfa_enforced_for(Username) ->
    default_mfa() =/= none andalso
        admin_override_of(Username) =/= ?ADMIN_MFA_EXEMPTED.
```

plus a case asserting an SSO user can still self-disable while `default_mfa` is set. If the intent is instead that the mandate blocks self-disable for SSO accounts as well, then the schema text, the changelog and the 403 message should say so.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -1206,16 +1207,36 @@ do_verify_mfa_token(Username, MfaToken, IsPwdOk) ->
             ok
     end.
 
+%% @doc Whether `dashboard.default_mfa' requires MFA for this account.
```
zmstone: Nit: new documentation in this repo should use the `-doc` attribute rather than a `%% @doc` comment. The rest of this file predates the rule, so this is only for the new function.

apps/emqx_dashboard/src/emqx_dashboard_admin.erl
zmstone, line 1214: please confrim if `SSO_USERNAME(_, _)` reaches here.
if yes, we need to return false for it because SSO's MFA is controlled by `emqx_dashboard_sso_mfa:mfa_required_for_user/2`.

with a test coverage please.

zmstone:
Hi @MorganaFuture 
Sorry for the slow review process.
Any chance you can fix my latest comments soon.
I can takeover from here otherwise.

MorganaFuture:
@zmstone hi! I'm on vacation till 26, don't have access to laptop. I'm continuing work on this after vacation

PR #18589, opened 2026-08-28
feat(dashboard): report an MFA enrollment status

MorganaFuture:
Release version: 7.0.0

Introduced in: N/A

## Summary

Fixes https://github.com/emqx/emqx/issues/18362

> **Stacked on https://github.com/emqx/emqx/pull/18586.** The first four commits are that PR
> (`39717be` … `6fcfc0e`); review the last three (`d183d15`, `ffd6ab7`, `c28b461`).
> `changes/ee/breaking-18586.en.md` in the file list belongs to #18586. Once #18586 merges,
> this branch rebases onto `dev-70` and carries only its own three commits.

Nothing in the API says whether an account uses TOTP, so the frontend cannot offer it.
`format_mfa/1` reports what is stored (`none | disabled | totp`), not what the user has to
do about it. `mfa_status/1` answers that, from state that already exists:

* `complete` — enrolled, and one code has been verified
* `pending_enforced` — not complete while MFA is required, so the next login stops for setup
* `pending_voluntary` — no MFA and nothing requires it
* `disabled` — opted out, and nothing requires them to opt back in

Two derivations are worth a look. A secret issued and never verified reads as
`pending_enforced` even when the enrollment was voluntary, because login stops on it either
way and the banner must not offer to dismiss it. And a `disabled` account reads as
`pending_enforced` once the global mandate covers it, since the login path enrolls it again;
the admin exemption (`admin_override = mfa_exempted`) is what keeps it `disabled`. That
second one is why this is stacked: #18362 was written before #18363 and says a disabled user
reports `disabled` even when `default_mfa` is set. #18586 turns `default_mfa` into a mandate
that re-enrolls a `disabled` account at its next login, so the resolver follows #18586
rather than the issue text.

Returned only where the first factor is already verified: both login responses
(`POST /login`, `POST /login/verify`) and `GET /current_user`. `GET /current_user` already
exists on `dev-70` and is `?SCOPE_PUBLIC`, so no new endpoint is added. `format_mfa/1` and
the `/users` response are untouched: widening that enum would change a response the
dashboard already consumes, and adding a field is easier to undo.

Item 4 of the issue needs no code on `dev-70`. Self-enrollment is `POST /current_user/mfa`,
which `emqx_dashboard_rbac:do_check_rbac/3` allows for any authenticated Dashboard user
without a scope, so a viewer can self-enroll today. `/users/:username/mfa` is admin-only and
rejects the caller's own account (`reject_self_target/2`), so the `mfa_management` scope is
not involved in the voluntary path.

One thing left for you to decide:

* **Item 5, the login enumeration.** `do_verify_mfa_token/3` ignores `IsPwdOk` on purpose
  ("so to protect from brute password being forced"), and `check_mfa_pwd/2` converts back to
  `password_error` only when the error carries a `secret`. So for an account that has
  verified a code, a wrong password answers `BAD_MFA_TOKEN`, which does reveal that the
  account exists and has MFA. Closing it means picking which oracle to expose instead, and
  `t_login_with_mfa_setting` asserts the current behaviour. Happy to do it either way.

Also noticed: `admin_override = mfa_required` does not force setup at a local login —
`maybe_init_mfa_state/2` acts on `default_mfa` only, while
`emqx_dashboard_sso_mfa:mfa_required_for_user/2` does honour the override. The resolver
counts the override as required, per the issue, so for a local account in that state the
status describes intent rather than what login does. Noted in a comment on
`mfa_required_for/1`, not changed here.

Tests: `emqx_dashboard_mfa_SUITE`, `emqx_dashboard_login_scram_SUITE`,
`emqx_dashboard_current_user_SUITE`, `emqx_dashboard_admin_SUITE`,
`emqx_dashboard_api_spec_SUITE`, `emqx_dashboard_user_scopes_SUITE`,
`emqx_dashboard_rbac_SUITE`.

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ee/feat-18589.en.md`
- [x] Schema changes are backward compatible or intentionally breaking (additive: a new `mfa_status` field on the login responses and on `GET /current_user`)


MorganaFuture:
Rebased. `GET /account` is gone — #18608 shipped `GET /current_user`, so the derived status is a field on that instead of a second endpoint returning the same record. Tests and changelog follow it.

This branch still sits on top of #18586, so its diff shows both until that one merges.

Issue #18226, opened 2026-07-31
feat(license): duration-based (sustained) triggering for the message-throughput (TPS) alarm


zmstone:
## Summary

Add an option to trigger the **message-throughput / TPS limit alarm** (`license_tps`, shown to operators as the throughput-limit alert) only when the limit is exceeded for a **sustained duration**, instead of firing immediately on the first over-limit sample. The goal is to alarm on genuine sustained overload rather than on brief transient spikes.

## Motivation / use case

Some workloads produce occasional **short TPS spikes** that momentarily exceed the license `max_tps` while the **average/steady-state TPS stays well below the limit**. Today each such spike raises the throughput-limit alarm, which operators experience as alarm noise — the dashboard/Grafana shows low average TPS, yet the alert fires on the peaks.

Operators would like to raise the alarm only when TPS stays above the limit **continuously for a configurable period** (e.g. ~1 hour), so transient bursts are ignored and the alarm reflects a real, sustained condition.

Note: **alarm throttling** only limits how often a notification for an already-active alarm is repeated — it does **not** change the activation condition, so it does not address this. This request is specifically about the *activation* condition.

## Current behaviour (dev-63)

- `apps/emqx_license/src/emqx_license_resources.erl` samples cluster TPS every `?CHECK_INTERVAL` (5s) and calls `max_tps_alarm/1`.
- `max_tps_alarm/1` **activates `license_tps` as soon as a single sampled cluster TPS exceeds `max_tps`** (message: `"License: TPS limit (N) exceeded."`). There is no minimum sustained-breach duration and no hysteresis on the TPS side.
- For contrast, the **connection-quota** alarm in the same module already has hysteresis via `license.connection_low_watermark` / `connection_high_watermark`; the TPS alarm has no equivalent.

So duration-based (sustained-breach) triggering is **not currently supported** for the throughput alarm.

## Proposed behaviour

Add a configurable **sustained-breach duration** (a.k.a. "for"/dwell time) for the TPS alarm, e.g.:

- `license.tps_alarm.trigger_duration` (duration, default `0s` = today's immediate behaviour for backward compatibility): activate `license_tps` only after observed cluster TPS has stayed **above the limit for at least this long continuously** (i.e. across consecutive samples spanning the duration).
- Optional companion for deactivation/anti-flapping: a low-watermark or a clear-duration so the alarm deactivates only after TPS drops below the limit (or a fraction of it) for a period — mirroring the existing connection-quota watermark pattern.

[...]

PR #18592, opened 2026-08-28
feat(license): add a sustained-breach duration for the TPS alarm

MorganaFuture:
Release version: 7.0.0

Introduced in: N/A

## Summary

Fixes https://github.com/emqx/emqx/issues/18226

`max_tps_alarm/1` activates `license_tps` from a single over-limit sample of the 5s cluster
TPS poll, so a workload with brief spikes over an otherwise comfortable average alarms on
the peaks.

Worth noting while reading this: the alarm is sticky. The `MaxTps > Limit` decision reads
`maps:get(max_tps, AlarmDetails)`, and for an existing alarm those details carry the peak
seen while it was active, not the latest sample, so it only clears when a license with a
higher limit arrives. One transient spike therefore leaves the alarm up indefinitely, which
makes the noise the issue describes worse than "it fires and then goes away".

`license.tps_alarm_trigger_duration` holds activation back until the limit has been exceeded
without interruption for that long. The window start lives in the module's existing
ephemeral table next to the TPS cache, and any sample at or below the limit clears it, so
the duration is a continuous window rather than a total. Default `0` keeps today's
behaviour. The gate is on the `activate` branch only — updating an active alarm's peak, and
deactivating it when a new license raises the limit, are unchanged.

I left the deactivation half out. The issue offers a clear-duration or low-watermark as an
optional companion, but the alarm not clearing on its own is deliberate (the `@private`
comment says so, and `t_alarm_if_tps_over_limit` asserts it), so adding it changes that
contract rather than extending this one. Happy to do it here if you want; the natural shape
is `tps_alarm_clear_duration` next to this one.

On the name: the issue gives `license.tps_alarm.trigger_duration` as an example. I went flat
because every field in this schema is flat, `connection_low_watermark` /
`connection_high_watermark` is the existing precedent for a pair of knobs tuning one alarm,
and `default_setting/0` is consumed as a flat map (`emqx_license_http_api_SUITE:reset_license/0`
does `atom_to_binary` over `maps:to_list`). Trivial to nest if you prefer.

`emqx_license_cli_SUITE:t_conf_update` compares the whole `license` config map with
`assertEqual`, so it lists the new key. The other suites match the keys they care about and
are untouched.

Milestone on the issue is 6.4.0, but there is no `dev-64` branch, so this targets `dev-70`.

Tests: `emqx_license_resources_SUITE` 3/3, `emqx_license_http_api_SUITE` 26/26,
`emqx_license_SUITE` 12/12, `emqx_license_cli_SUITE` 10/10, `emqx_license_checker_SUITE`
15/15, `emqx_license_session_hwm_SUITE` 9/9.

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ee/feat-18592.en.md`
- [x] Schema changes are backward compatible or intentionally breaking (additive, default `0s` preserves current behaviour)


apps/emqx_license/src/emqx_license_schema.erl
```diff
@@ -148,6 +157,9 @@ default(connection_high_watermark) ->
     <<"80%">>;
 default(high_watermark_timezone) ->
     system;
+default(tps_alarm_trigger_duration) ->
+    %% Fire on the first over-limit sample, as before this option existed.
+    <<"0s">>;
```
zmstone:
```suggestion
    ~"30m";
```

changes/ee/feat-18592.en.md
```diff
@@ -0,0 +1,3 @@
+Added `license.tps_alarm_trigger_duration`, which holds back the license message-throughput alarm until the limit has been exceeded continuously for that long.
+
+Workloads with brief traffic spikes raised the alarm on a single over-limit sample, even when the steady-state throughput stayed well below the licensed limit. Setting a duration, for example `1h`, alarms only on sustained overload; a sample at or below the limit restarts the count. The default `0s` keeps the previous behaviour of alarming on the first sample above the limit. The check runs on the license resource sampling interval, currently 5s, so the duration takes effect in units of that interval.
```
zmstone:
```suggestion
Added `license.tps_alarm_trigger_duration`, which holds back the license message-throughput alarm until the limit has been exceeded continuously for that long.

Workloads with brief traffic spikes raised the alarm on a single over-limit sample, even when the steady-state throughput stayed well below the licensed limit. Setting a duration, for example `1h`, alarms only on sustained overload; a sample at or below the limit restarts the count. The default value is 30 minutes.
```

apps/emqx_license/src/emqx_license_resources.erl
```diff
@@ -194,6 +238,12 @@ cached_latest_cluster_tps() ->
 cached_max_tps() ->
     ?SAFE_CACHE_LOOKUP(max_cluster_tps, 0).
 
+cached_tps_over_limit_since() ->
+    ?SAFE_CACHE_LOOKUP(tps_over_limit_since, undefined).
+
+now_ms() ->
```
zmstone: `erlang:system_time/1` is wall-clock time, so it can step. `tps_breach_sustained/0` subtracts two
samples of it to measure elapsed time, and an NTP correction, a manual clock change, or a VM
suspend/resume moves the result. A backward step delays the alarm by the size of the step. A forward
step fires it early, which is the case this option exists to prevent.

Use `erlang:monotonic_time(millisecond)` here. It is the clock for measuring a duration, and it cannot
step.

The existing `erlang:system_time(millisecond)` in `update_resources/0` is fine as it is, because that
value is a timestamp that is reported, not a difference that is compared against a configured window.

This matters more the longer the window. With the `30m` default under discussion, a clock correction
inside the window is entirely plausible.

apps/emqx_license/src/emqx_license_resources.erl
```diff
@@ -166,15 +167,58 @@ max_tps_alarm({ok, #{max_tps := Limit}}) ->
...
+%% Held in the same ephemeral table as the TPS cache, so a restart of this
+%% process starts the window over. That only delays an alarm by the configured
+%% duration, and the sampling itself restarts with the process anyway.
+track_tps_breach(_IsOverLimit = false) ->
+    _ = ?OK(ets:delete(?MODULE, tps_over_limit_since)),
+    ok;
+track_tps_breach(_IsOverLimit = true) ->
+    case cached_tps_over_limit_since() of
+        undefined ->
+            _ = ?OK(ets:insert(?MODULE, {tps_over_limit_since, now_ms()})),
```
zmstone: `?OK/1` catches every exception and returns `ok`. It exists in this module to keep alarm API calls
from taking the sampling process down, which is the right call for a remote-ish API. Applying it to
this module's own table is different: `?MODULE` is created in `init/1` and written from the process
that owns it, so a failure here is a programming error, not an expected condition.

Wrapped this way, a future rename or a table that is missing turns into a silent no-op. The window
never advances, `tps_breach_sustained/0` keeps returning `false`, and the alarm never fires again —
with nothing in the log to say why. That is the worst failure mode for an alarm.

Write to the table directly, or log if you want to keep it non-fatal.

Same for the `ets:delete/2` on line 195.

apps/emqx_license/src/emqx_license_resources.erl
```diff
@@ -166,15 +167,58 @@ max_tps_alarm({ok, #{max_tps := Limit}}) ->
             _ = emqx_alarm:update_details(license_tps, AlarmDetails),
             ok;
         true when Action =:= activate ->
-            Message = iolist_to_binary(io_lib:format("License: TPS limit (~w) exceeded.", [Limit])),
-            ?OK(emqx_alarm:activate(license_tps, AlarmDetails, Message));
+            case tps_breach_sustained() of
```
zmstone, line 197: Nothing clears `tps_over_limit_since` when the alarm activates, so after activation the marker stays
set for as long as the samples stay over the limit.

That is harmless today, because the alarm is sticky and only clears when a license with a higher limit
arrives — and that same condition also drives `track_tps_breach(false)`, which clears the marker.

It stops being harmless the moment the alarm can clear on its own. The PR description proposes exactly
that as a follow-up (`tps_alarm_clear_duration`). With a self-clearing alarm and the marker still set,
the next sample takes the `activate` branch, finds the window already satisfied, and re-raises
immediately — so the clear duration would produce a flapping alarm rather than a quiet one.

Worth clearing the marker on activation, or at least a comment here recording the constraint, so the
follow-up does not inherit it silently.

apps/emqx_license/test/emqx_license_resources_SUITE.erl
```diff
@@ -220,6 +220,61 @@ t_alarm_if_tps_over_limit(Config) when is_list(Config) ->
...
+        meck:expect(
+            emqx_license_proto_v3,
+            stats,
+            fun(_Nodes, _Now) -> [{ok, #{sessions => 21, tps => Tps}}] end
+        ),
+        ok = update_now()
+    end,
+    %% Over the limit, but the breach has only just started.
+    MockTps(11),
+    ?assertNotReceive({alarm_activated, _, _}, 100),
```
zmstone, line 271: This pair of assertions has less than a second of slack against wall-clock time.

The window opens at the first `MockTps(11)` on line 261. The only thing between it and the second one
on line 263 is a 100 ms `?assertNotReceive`. If the runner stalls for the remaining ~900 ms — GC, a
noisy neighbour on CI, a slow `meck:expect` — the second sample satisfies `now_ms() - Since >= 1000`,
the alarm activates, and this assertion fails.

The failure would read as a real regression, which is the expensive kind of flake.

Raising the configured duration well above any plausible stall fixes the early phase, at the cost of
longer sleeps later. Making the clock injectable is the better fix: with `now_ms/0` behind something
the test can control, this case asserts the logic exactly and needs no `timer:sleep/1` at all. That
also removes the two bare sleeps on lines 267 and 271, which the house test style discourages.

apps/emqx_license/src/emqx_license_resources.erl
```diff
@@ -166,15 +167,58 @@ max_tps_alarm({ok, #{max_tps := Limit}}) ->
...
+tps_breach_sustained() ->
+    case emqx_conf:get([license, tps_alarm_trigger_duration], 0) of
+        Duration when Duration =< 0 ->
+            true;
+        Duration ->
+            case cached_tps_over_limit_since() of
+                undefined ->
+                    false;
+                Since ->
+                    now_ms() - Since >= Duration
```
zmstone:
```suggestion
                    now_ms() - Since > Duration
```

apps/emqx_license/src/emqx_license_schema.erl
```diff
@@ -68,6 +68,14 @@ fields(key_license) ->
             example => default(connection_high_watermark),
             desc => ?DESC(connection_high_watermark_field)
         }},
+        {tps_alarm_trigger_duration, #{
```
zmstone: maybe misleading in the issue description,
i think the better name is `tps_alarm_min_interval`

apps/emqx_license/src/emqx_license_resources.erl
MorganaFuture: good catch, done. monotonic_time(millisecond) for the window now.

left the system_time in update_resources/0 as it is, that one is a timestamp
being reported rather than a difference compared against the window.

apps/emqx_license/src/emqx_license_resources.erl
MorganaFuture: agreed, dropped it on both. the writes are `true = ets:insert(...)` and
`true = ets:delete(...)` now.

put the reasoning in a comment above track_tps_breach/1 as well, since the
silent-freeze failure mode is not obvious from the call site.

apps/emqx_license/src/emqx_license_schema.erl
MorganaFuture: done, 30m.

worth a decision though: this changes what existing clusters see. a deployment
with max_tps set that alarms today will alarm half an hour later after the
upgrade, with nobody touching the config. t_alarm_if_tps_over_limit did not set
the option at all and broke on the new default, so it asks for 0 explicitly now.

should the changelog be breaking- rather than feat- ?

changes/ee/feat-18592.en.md
MorganaFuture: applied as suggested.

i also updated the schema description, it still said the default was 0s. while
there i dropped the sentence about the check running on the 5s sampling
interval. the issue asked for that to be documented, but at a 30m default it
reads like noise. say the word if you want a short version kept.

apps/emqx_license/src/emqx_license_resources.erl
MorganaFuture: good spot, and it is worse than it looks. clearing the marker on activation is
not enough on its own: track_tps_breach/1 runs before the decision, so the very
next over-limit sample re-arms it, and it then ages for as long as the alarm is
up. a self-clearing alarm would still re-raise at once.

so the window now only runs while there is no alarm to raise, and the reason is
in a comment above it.

apps/emqx_license/test/emqx_license_resources_SUITE.erl
MorganaFuture: you are right, and the sleeps are gone.

added backdate_tps_breach/1 as a test seam, so the case ages the window instead
of waiting on it. no sleeps left in the case and it now asserts the logic
exactly. it goes through the gen_server rather than writing the table directly,
since the table is protected.

apps/emqx_license/src/emqx_license_resources.erl
MorganaFuture: applied. the 0 case is unaffected, it returns early before the comparison.

apps/emqx_license/src/emqx_license_schema.erl
MorganaFuture: i would push back once on this one. the issue has a paragraph saying alarm
throttling, how often an already-active alarm is repeated, is a different thing
and that the request is about the activation condition. `min_interval` reads
like the gap between alarms, which is the meaning it warns against.

if you still prefer it after that i will rename, it is your call.

apps/emqx_license/src/emqx_license_schema.erl
zmstone: it’s fine and not breaking

apps/emqx_license/src/emqx_license_schema.erl
zmstone: true, misleading in a different way 

how about tps_alarm_sustain_duration

zmstone: clarify: the combination of the instantaneous-action word “trigger” with “duration”, which implies something continuous, reads a little awkwardly.

MorganaFuture: Renamed to `tps_alarm_sustain_duration`, agreed on the reading.

The i18n label still said "TPS Alarm Trigger Duration" as well, so that went with it, along with the test case and helpers that carried the word.

apps/emqx_license/src/emqx_license_resources.erl
```diff
@@ -160,21 +178,77 @@ max_tps_alarm({ok, #{max_tps := Limit}}) ->
...
+            ok;
+        _AlreadyTracking ->
+            ok
+    end.
+
+%% @private Whether the current run of over-limit samples has lasted at least
+%% `license.tps_alarm_sustain_duration'. The default of 0 keeps the alarm firing
+%% on the first over-limit sample.
+tps_breach_sustained() ->
+    case emqx_conf:get([license, tps_alarm_sustain_duration], 0) of
```
zmstone: The in-code fallback is `0` but the schema default is now `30m`, so the same setting has two
different defaults depending on which one is reached.

`emqx_config:fill_defaults/2` normally guarantees the key is present, in which case this argument is
dead and the disagreement is invisible. When it is *not* present — a config path that skips the fill,
or a test that puts a partial `license` map — the node silently alarms on the first over-limit sample
while the schema and the changelog both say it waits 30 minutes.

Either drop the default argument so a missing key fails loudly, or make it `default(tps_alarm_sustain_duration)`
so the two cannot drift apart. The first is probably better: a silently-wrong alarm threshold is worse
than a crash at the one call site that reads it.

rel/i18n/emqx_license_schema.hocon
```diff
@@ -12,6 +12,14 @@ connection_low_watermark_field.desc:
 connection_low_watermark_field.label:
 """Connection low watermark"""
 
+tps_alarm_sustain_duration {
+    label: "TPS Alarm Sustain Duration"
+    desc: """~
+        Minimum duration for which the cluster message throughput must stay above the license TPS limit before the throughput alarm is activated.
+        A sample at or below the limit restarts the count, so the limit has to be exceeded continuously.
+        Set to `0s` to activate the alarm on the first sample above the limit."""
```
zmstone: Worth restoring the sentence about the sampling interval that an earlier revision had here.

The window is only evaluated when `max_tps_alarm/1` runs, which is on the license resource poll
(`?CHECK_INTERVAL`, 5s). So the duration takes effect in units of that interval, and a value finer
than it — `2s`, say — is silently rounded up to the next sample. Without that note an operator has no
way to know why a short duration does not behave as written.

One line is enough, e.g. "The check runs on the license resource sampling interval (5s), so the
duration takes effect in units of that interval."

changes/ee/feat-18592.en.md
```diff
@@ -1 +1,3 @@
-`bin/node_dump` now collects node information through a dedicated `emqx ctl node_dump` command instead of `emqx eval`. Audit records for the collected system info, application environment, and configuration dump now identify the `node_dump` command instead of the raw Erlang expression that produced them.
+Added `license.tps_alarm_sustain_duration`, which holds back the license message-throughput alarm until the limit has been exceeded continuously for that long.
+
+Workloads with brief traffic spikes raised the alarm on a single over-limit sample, even when the steady-state throughput stayed well below the licensed limit. Setting a duration, for example `1h`, alarms only on sustained overload; a sample at or below the limit restarts the count. The default value is 30 minutes.
```
zmstone: This paragraph reads as advice to opt in — "Setting a duration, for example `1h`, alarms only on
sustained overload" — and then mentions the new default almost in passing at the end. An operator
skimming the release notes will not realise their existing deployment's behaviour has changed.

It has: an installation that today raises `license_tps` on the first over-limit sample will, after
upgrading, wait 30 minutes. Anyone using that alarm for capacity alerting sees it 30 minutes later
than before, and anyone who wants the old behaviour has to find the option and set it to `0s`.

Suggest stating the change directly — that the alarm now requires 30 minutes of continuous overload
by default, and that `0s` restores alarming on the first sample.

MorganaFuture:
@zmstone looks like flaky?

apps/emqx_license/src/emqx_license_resources.erl
zmstone: schema has default (which is 30m), there is no reason for this config to fail.
```suggestion
    case emqx_conf:get([license, tps_alarm_sustain_duration]) of
```

changes/ee/feat-18593.en.md
```diff
@@ -0,0 +1 @@
+`bin/node_dump` now collects node information through a dedicated `emqx ctl node_dump` command instead of `emqx eval`. Audit records for the collected system info, application environment, and configuration dump now identify the `node_dump` command instead of the raw Erlang expression that produced them.
```
zmstone, line 1: this file should not be added in this pr.

apps/emqx_license/src/emqx_license_schema.erl
```diff
@@ -68,6 +68,14 @@ fields(key_license) ->
             example => default(connection_high_watermark),
             desc => ?DESC(connection_high_watermark_field)
         }},
+        {tps_alarm_sustain_duration, #{
+            type => emqx_schema:duration_ms(),
+            default => default(tps_alarm_sustain_duration),
+            example => <<"1h">>,
```
zmstone:
```suggestion
            example => <<"2m">>,
```

apps/emqx_license/src/emqx_license_resources.erl
zmstone: we need to clamp this value with hardcoded range.

`[0, timer:minutes(10)]`

so one cannot configure a infinit large number to avoid getting alarmed.
the purpose of this alarm is to annoy people so they should upgrade their license tier.

apps/emqx_license/src/emqx_license_schema.erl
```diff
@@ -148,6 +157,8 @@ default(connection_high_watermark) ->
     <<"80%">>;
 default(high_watermark_timezone) ->
     system;
+default(tps_alarm_sustain_duration) ->
+    ~"30m";
```
zmstone: sorry for going back-n-forth on this one.
maybe 30m is too long to classify as a "burst"

```suggestion
default(tps_alarm_sustain_duration) ->
    ~"1m";
```

changes/ee/feat-18592.en.md
zmstone:
```suggestion
By default workloads with brief TPS burst raises the alarm on a single over-limit sample, even when the steady-state throughput stayed well below the licensed limit. Setting a duration, for example `5m`, alarms only on sustained overload; a sample at or below the limit resets the counter.
```

rel/i18n/emqx_license_schema.hocon
zmstone: describe the clamped range too.

zmstone:
> @zmstone looks like flaky?

looks like so. i can retry in the next round.

apps/emqx_license/src/emqx_license_resources.erl
MorganaFuture: both done. the read is `emqx_conf:get([license, tps_alarm_sustain_duration])` now, no fallback.

for the range i put it in the schema as a validator, so an out-of-range value is rejected with a message instead of being silently capped, and kept a `min/2` where the window is read. the test suite sets this through `emqx_config:put` and skips validation, so the second one is not redundant.

MorganaFuture: capped at 10m, `?MAX_TPS_ALARM_SUSTAIN_DURATION` in emqx_license.hrl. `t_conf_update_tps_alarm_sustain_duration_range` covers 11m and 3650d being rejected and 10m and 0s being accepted.

changes/ee/feat-18593.en.md
MorganaFuture: this one i would keep. `changes/ee/feat-18592.en.md` on dev-70 is the node_dump entry from #18593, which merged yesterday under the wrong number. dropping the new file makes that release note disappear, and check-changes-filename-pattern.sh only looks at the shape so nothing would catch it.

happy to do it the other way round if you prefer: leave 18593 text where it is and file this one under another id.

apps/emqx_license/src/emqx_license_schema.erl
MorganaFuture: done, 2m.

apps/emqx_license/src/emqx_license_schema.erl
MorganaFuture: done, 1m. also updated t_conf_update, it asserted the old default.

changes/ee/feat-18592.en.md
MorganaFuture: applied, with the first clause in past tense. with the default at 1m, "by default ... raises the alarm on a single over-limit sample" would read as what still happens. added the default and the accepted range at the end.

rel/i18n/emqx_license_schema.hocon
MorganaFuture: done, range is in the desc now.

apps/emqx_license/test/emqx_license_resources_SUITE.erl
MorganaFuture: came back to this one. the first two phases were still racing: the window was 1000 ms and only a 100 ms assertNotReceive sat between the two samples. window is 5m now and every phase ages it with backdate_tps_breach/1, so nothing depends on how fast the runner is.

changes/ee/feat-18593.en.md
zmstone: right. my bad not noticing file 18592.md has the old content.

apps/emqx_license/src/emqx_license_resources.erl
```diff
@@ -160,21 +180,86 @@ max_tps_alarm({ok, #{max_tps := Limit}}) ->
...
+                    monotonic_ms() - Since > Duration
+            end
+    end.
+
+%% @private The schema rejects anything outside the permitted range, but the
+%% cap is re-applied here so the alarm cannot be silenced by a value that
+%% reached the config without passing validation.
+sustain_duration() ->
+    min(
+        emqx_conf:get([license, tps_alarm_sustain_duration]),
```
zmstone: sorry for coming back to this.
it's safer so the process won't crash every 5s if the config for some reason is missing.
```suggestion
        emqx_conf:get([license, tps_alarm_sustain_duration], timer:seconds(60)),
```

changes/ee/feat-18592.en.md
```diff
@@ -1 +1,3 @@
-`bin/node_dump` now collects node information through a dedicated `emqx ctl node_dump` command instead of `emqx eval`. Audit records for the collected system info, application environment, and configuration dump now identify the `node_dump` command instead of the raw Erlang expression that produced them.
+Added `license.tps_alarm_sustain_duration`, which holds back the license message-throughput alarm until the limit has been exceeded continuously for that long.
+
+Workloads with brief TPS bursts used to raise the alarm on a single over-limit sample, even when the steady-state throughput stayed well below the licensed limit. Setting a duration, for example `5m`, alarms only on sustained overload; a sample at or below the limit resets the counter. The default is 1 minute, and the accepted range is `0s` to `10m`.
```
zmstone:
```suggestion
Workloads with brief TPS bursts used to raise the alarm on a single over-limit sample (samples every 5 seconds), even when the steady-state throughput stayed well below the licensed limit. Setting a duration, for example `5m`, alarms only on sustained overload; a sample at or below the limit resets the counter. The default is 1 minute, and the accepted range is `0s` to `10m`.
```

zmstone:
thank you so much, last two nitpicking comments. 
once addressed and ci turns green, i'll approve and merge.

apps/emqx_license/src/emqx_license_resources.erl
MorganaFuture: Done. Put the fallback in a macro next to the existing cap rather than inline, so the number lives in one place: the schema keeps `1m` because that is what the generated docs and config files show, and the fallback needs milliseconds. Nothing tied the two together, so there is a test that fails if they drift.

changes/ee/feat-18592.en.md
MorganaFuture: Applied. Checked the interval against the code before writing it in: `?CHECK_INTERVAL` is 5000.

MorganaFuture:
Both addressed in the threads. CI is sitting at `action_required` on the new head, so it needs an approve-and-run from your side before it can turn green.

The previous run was 25 passed, 6 skipped and one failure that was `curl: (6) Could not resolve host` while downloading rebar3, so nothing from the change itself.

zmstone (approved):

zmstone:
Thank you for the PR @MorganaFuture

PR #18596, opened 2026-08-28
fix(authz): isolate namespaced built-in database rules

savonarola:
Release version: 7.0.0

Introduced in: 6.0.0

Fixes emqx/emqx-dev-team-tasks#67

Authz counterpart of https://github.com/emqx/emqx/pull/18512

## Summary

In the hardened security profile, namespaced rule lookup is now as follows, in order:
- If matching global client ID or username rules exist, theb deny. 
- If namespace is not empty, apply only namespace rules. Do not fallback to global rules even if the particular user there are no rules.
- Fall back to global rules when the namespace is empty.

## PR Checklist


changes/ee/breaking-18596.en.md
```diff
@@ -0,0 +1 @@
+Added built-in database authorization rule conflict protection for the hardened security profile. EMQX now rejects namespaced username and client ID rules that conflict with explicit global rules, and prevents configured namespaces from falling back to global rules.
```
zmstone, line 1: this would be a breaking- for v7

apps/emqx_auth_mnesia/src/emqx_authz_mnesia.erl
```diff
@@ -281,6 +312,56 @@ read_rules(Namespace, Key) ->
...
+        {atomic, {error, _} = Error} ->
+            Error;
+        {aborted, Reason} ->
+            {error, Reason}
+    end.
+
+store_explicit_rules_tx(_Strict, ?global_ns, Who, Rules) ->
+    Record = #?ACL_TABLE{who = Who, rules = Rules},
+    ok = mnesia:write(?ACL_TABLE, Record, write),
+    {ok, false};
```
thalesmg, line 349: q: if we write a global rule that does shadow an existing ns rule, shouldn't we also fail or at least log a warning?

thalesmg (dismissed):

apps/emqx_auth_mnesia/src/emqx_authz_mnesia.erl
savonarola: this is the same as for authn: it may be too costly to verify that no NS has a shadowed user. So we assume global admins to have more power and we do not allow ns admins to "squat" user identities.

zmstone (approved):

PR #18597, opened 2026-08-28
feat(tls): add self-signed certificate generation library

### LINKED ISSUE #4636: Example Security Certificates in this Repository and the Docker Image
Hi!

We recognized that there are example security certificates, including their private keys, in this repository (/etc/certs). These certificates are also copied into the EMQX Broker Docker Image. 

The distribution of example certificates allows users to use these certificates for their deployments directly. However, as the private key is public, these security certificates are basically compromised and give a false sense of security. As everybody in possession of the private key can authenticate as the user's MQTT broker, attackers could easily perform impersonation attacks.

To prevent users from using these compromised security certificates, a significantly more secure way would be to provide a script that allows users to generate their own security certificates. Hence, we would recommend removing the example certificates and provide such a script.

Let me know what you think.

Markus

zmstone:
Release version:
7.0.0

Introduced in:
N/A

## Summary

First PR in a multi-PR split of #18129 (resolves #4636). Adds `emqx_utils_certs`, a self-signed X.509
generation library built on OTP `public_key` only, with no external tools. **Nothing in this PR
calls it** — it is a pure addition and cannot change any behavior. Later PRs in the series build on
this to actually generate and serve a default certificate.

Lives in `apps/emqx_utils/src` rather than `apps/emqx/src`: it has no dependency on `emqx`, and a
later PR needs it reachable from test code across roughly 30 other apps, which is a much lighter
reach through `emqx_utils` than through `emqx`. `apps/emqx_utils/mix.exs` gained `:public_key` in
`extra_applications` as part of this move — it needs it to resolve `-include_lib("public_key/...")`,
same as `apps/emqx/mix.exs` already declares for the same reason.

The library exposes three functions:

- `generate_ca/1` — a root CA certificate and key.
- `generate_cert/2` — a certificate signed by a given CA.
- `self_signed_bundle/1` — a complete bundle: a one-off CA, a leaf certificate signed by it, and
  **no CA private key**. The CA key is generated, used once to sign the leaf, and discarded. No
  further certificate can be issued under that CA afterwards. This is the property that will make
  `verify_peer` safe with no configured CA once a later PR serves this bundle from a listener.

Key parameters, stated in the module doc: `ec` (`secp256r1`) by default, `rsa` (2048 bits) also
available, 3650-day validity starting one day in the past to tolerate clock skew. `basicConstraints`
is `CA:TRUE` + `keyUsage` `[keyCertSign, cRLSign]` on the CA, `CA:FALSE` + key-type-appropriate
`keyUsage` + `extKeyUsage [serverAuth, clientAuth]` on the leaf.

The module has no dependency on managed certs at all: `self_signed_bundle/1` returns a plain
`#{ca, cert, key}` map, not the `?FILE_KIND_*` vocabulary from `emqx_managed_certs.hrl` — that
vocabulary implies a value ready to drop into a managed-certs bundle, which this isn't. A later PR
that wires this into managed certs maps the fields explicitly.

`cn` is validated before it reaches the certificate subject: control characters and `#`/`+`/`/` are
rejected (the latter because EMQX can substitute a certificate's CN into topic filters as a client
attribute), raising an error rather than emitting a certificate with an unsafe subject.

### Certificate dump

Generated with `self_signed_bundle(#{cn => "localhost", sans => [{dns, "localhost"}, {ip, {127,0,0,1}}, {ip, {0,0,0,0,0,0,0,1}}]})`
(default key type, `ec`), dumped with `openssl x509 -text -noout`:

**Leaf**

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            8f:39:4b:34:2b:ba:7a:79
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: CN = EMQX Self-Signed CA, O = EMQX
        Validity
            Not Before: Aug 27 00:00:00 2026 GMT
            Not After : Aug 25 00:00:00 2036 GMT
        Subject: CN = localhost, O = EMQX
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            X509v3 Basic Constraints: critical
                CA:FALSE
            X509v3 Key Usage:
                Digital Signature, Key Agreement
            X509v3 Extended Key Usage:
                TLS Web Server Authentication, TLS Web Client Authentication
            X509v3 Subject Alternative Name:
                DNS:localhost, IP Address:127.0.0.1, IP Address:0:0:0:0:0:0:0:1
    Signature Algorithm: ecdsa-with-SHA256
```

**CA**

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 2222408309938311821 (0x1ed7949543bfd28d)
        Signature Algorithm: ecdsa-with-SHA256
        Issuer: CN = EMQX Self-Signed CA, O = EMQX
        Validity
            Not Before: Aug 27 00:00:00 2026 GMT
            Not After : Aug 25 00:00:00 2036 GMT
        Subject: CN = EMQX Self-Signed CA, O = EMQX
        Subject Public Key Info:
            Public Key Algorithm: id-ecPublicKey
                Public-Key: (256 bit)
                ASN1 OID: prime256v1
                NIST CURVE: P-256
        X509v3 extensions:
            X509v3 Basic Constraints: critical
                CA:TRUE
            X509v3 Key Usage: critical
                Certificate Sign, CRL Sign
    Signature Algorithm: ecdsa-with-SHA256
```

(Public key and signature bytes trimmed from both for length; full dumps reproduce with the command
above against this branch. `rsa` produces the same extension shapes with `sha256WithRSAEncryption`
signatures instead.)

## PR Checklist

No changelog: this module has no callers and no user-visible effect. A changelog entry lands with a
later PR in the series once something actually generates and serves certificates.

New test suite: `apps/emqx_utils/test/emqx_utils_certs_SUITE.erl`.

### DIFF
--- a/apps/emqx/src/emqx_default_cert.erl
+++ b/apps/emqx/src/emqx_default_cert.erl
@@ -1,0 +1,278 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_default_cert).
+
+-moduledoc """
+Self-signed X.509 certificate generation using the OTP `public_key`
+application, with no external tools required.
+
+`self_signed_bundle/1` generates a one-off root CA and a leaf
+certificate signed by it, and returns only the leaf key, the leaf
+certificate and the CA certificate: the CA private key is discarded
+after signing. No further certificate can be issued under that CA, so
+a caller that trusts only this CA is safe even though the CA's key was
+never kept anywhere.
+
+Key parameters:
+
+* RSA keys are 2048 bits; EC keys use `secp256r1`. Both are the
+  current minimum widely accepted for a server certificate.
+* Certificates are valid for 3650 days (about ten years) starting one
+  day in the past, so a small clock skew between the generating node
+  and a client does not make the certificate look not-yet-valid.
+
+This module has no callers. It is a pure cryptography library.
+""".
+
+%% ASN.1 record fields and values (e.g. `namedCurve`, `generalTime`)
+%% are camel-cased by the public_key application.
+-elvis([{elvis_style, atom_naming_convention, disable}]).
+
+-include_lib("public_key/include/public_key.hrl").
+-include("emqx_managed_certs.hrl").
+
+-export([
+    generate_ca/1,
+    generate_cert/2,
+    self_signed_bundle/1
+]).
+
+-export_type([pem_pair/0, cert_bundle/0, san/0, key_type/0]).
+
+-define(VALIDITY_DAYS, 3650).
+-define(CURVE, secp256r1).
+-define(RSA_KEY_SIZE, 2048).
+
+-type pem() :: binary().
+-type pem_pair() :: #{cert_pem := pem(), key_pem := pem()}.
+-type san() :: {dns, string()} | {ip, inet:ip_address()}.
+-type key_type() :: rsa | ec.
+-type cert_bundle() :: #{
+    ?FILE_KIND_KEY := pem(),
+    ?FILE_KIND_CHAIN := pem(),
+    ?FILE_KIND_CA := pem()
+}.
+
+%%--------------------------------------------------------------------
+%% API
+%%--------------------------------------------------------------------
+
+-doc """
+Generates a root CA certificate and its private key.
+
+Options: `cn` (required) becomes the subject common name; `org` is the
+subject organization name (defaults to `"EMQX"`); `key_type` is `rsa`
+(default) or `ec`.
+""".
+-spec generate_ca(#{cn := string(), org => string(), key_type => key_type()}) -> pem_pair().
+generate_ca(#{cn := _} = Opts) ->
+    Key = gen_key(maps:get(key_type, Opts, rsa)),
+    Subject = subject(Opts),
+    TBS = tbs_certificate(Subject, Subject, Key, Key, ca_extensions()),
+    Der = public_key:pkix_sign(TBS, Key),
+    #{cert_pem => cert_to_pem(Der), key_pem => key_to_pem(Key)}.
+
+-doc """
+Generates a certificate signed by the given CA.
+
+Options: `cn` (required) becomes the subject common name; `org` is the
+subject organization name (defaults to `"EMQX"`); `sans` is the list of
+subject alternative names (defaults to `[]`); `key_type` is `rsa`
+(default) or `ec`.
+""".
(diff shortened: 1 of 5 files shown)


apps/emqx/src/emqx_default_cert.erl
```diff
@@ -0,0 +1,278 @@
...
+%%--------------------------------------------------------------------
+%% API
+%%--------------------------------------------------------------------
+
+-doc """
+Generates a root CA certificate and its private key.
+
+Options: `cn` (required) becomes the subject common name; `org` is the
+subject organization name (defaults to `"EMQX"`); `key_type` is `rsa`
+(default) or `ec`.
```
zmstone: default to ec

apps/emqx/src/emqx_default_cert.erl
```diff
@@ -0,0 +1,281 @@
...
+-define(RSA_KEY_SIZE, 2048).
+
+-type pem() :: binary().
+-type pem_pair() :: #{cert_pem := pem(), key_pem := pem()}.
+-type san() :: {dns, string()} | {ip, inet:ip_address()}.
+-type key_type() :: rsa | ec.
+-type cert_bundle() :: #{
+    ?FILE_KIND_KEY := pem(),
+    ?FILE_KIND_CHAIN := pem(),
+    ?FILE_KIND_CA := pem()
```
zmstone: let's keep it distinguished for this interface,
so not to be confused (as if it can be used as a drop-in bundle in etc/certs2).
```suggestion
    ca := pem(),
    cert := pem(),
    key := pem()
```

apps/emqx_utils/src/emqx_utils_certs.erl
```diff
@@ -0,0 +1,281 @@
...
+
+-doc """
+Generates a root CA certificate and its private key.
+
+Options: `cn` (required) becomes the subject common name; `org` is the
+subject organization name (defaults to `"EMQX"`); `key_type` is `ec`
+(default) or `rsa`.
+""".
+-spec generate_ca(#{cn := string(), org => string(), key_type => key_type()}) -> pem_pair().
+generate_ca(#{cn := _} = Opts) ->
```
zmstone, line 75: fail fast.
raise error exception if cn has control chars, e.g. CRLF, and since EMQX supports using CN as client attributes which often also used as topic words, we should reject # / and + too. 
see is_mqtt_safe_utf8

thalesmg (approved):

Issue #16307, opened 2025-11-18
Add client attributes to database authentication


lugrinder:
### What would you like to be added or enhanced?

For all database password-based authentication integrations, add selected columns as client_attrs, similar to the approach used in HTTP authentication.

For example:
```sql
SELECT password_hash, salt, is_superuser, sn as client_attrs.sn FROM mqtt_user WHERE username = ${username} LIMIT 1
```

### Why is this needed?

This expands the authentication possibilities to the same level as HTTP authentication, without the need for an external authentication service/endpoint.

id:
<img width="927" height="443" alt="Image" src="https://github.com/user-attachments/assets/9a2064a0-465d-4866-86a7-77e7ccc305a6" />

> The SQL supports the following placeholders：${clientid}, ${username}, ${password}, ${peerport}, ${zone}, ${listener}, ${client_attrs.\<attribute\>}

zmstone:
Hi @lugrinder 

Client attribute injection is only supported in HTTP auth results, but not others.
We do not plan to support this any time soon.
But we'll keep this feature request open.

Issue #18140, opened 2026-07-28
LDAP authenticator: map directory attributes into client_attrs


id:
### What would you like to be added or enhanced?

A configurable mapping in the LDAP authenticator that copies attributes from the authenticated user's directory entry into `client_attrs`, in the same spirit as the existing `is_superuser_attribute` and `clientid_override_attribute` fields. Example: map `memberOf` to `client_attrs.groups`.

Touch points: a new schema field in `emqx_authn_ldap_schema.erl`, and populating `client_attrs` in the authn result in `emqx_authn_ldap_hash.erl` and `emqx_authn_ldap_bind.erl`. Today only the HTTP, JWT and SCRAM-over-HTTP authenticators can return `client_attrs`.

Design points to settle:

- Multi-valued attributes: `memberOf` typically returns several DNs, so the representation needs to be defined (joined string, list, or first value) in a way `{client_attr, ...}` ACL conditions can match against.
- Values are DNs, so matching is by substring or regex unless we offer CN extraction.
- Directories that do not serve `memberOf` on the user entry expose membership only through a group search (`(member=<user DN>)`), which returns one entry per group; `emqx_ldap_connector` currently rejects multi-entry results (`ldap_query_found_more_than_one_match`). A separate group-search query is probably out of scope for the first iteration, but the limitation should be documented.

### Why is this needed?

Authorization by LDAP group membership with the group-to-topic mapping kept in broker configuration. Corporate directories are often read-only for the team operating the broker, which rules out the current LDAP authorization source: it reads permissions from attributes stored in the directory itself (`publish_attribute`, `subscribe_attribute`, `all_attribute`, `acl_rule_attribute`).

The authorization side needs no new code. `acl.conf` already supports `{client_attr, Name, Value}` and `{client_attr, Name, {re, Regex}}` conditions plus `${client_attrs.NAME}` topic placeholders. The gap is only getting group membership out of the directory and into `client_attrs` at CONNECT time.

Workarounds today require replacing LDAP authentication entirely with an HTTP authenticator backed by a custom service, or a plugin on `client.post_authn`. Chaining does not help, since the authentication chain stops at the first authenticator that returns a decision, so LDAP cannot be combined with a second authenticator that only supplies attributes.

zmstone:
please add desired GA timeline.

id:
Design input on multi-valued attributes:

- Matching semantics: a condition against a multi-valued attribute (memberOf being the common case) should match when any single value matches.
[...]

PR #18605, opened 2026-08-29
feat(authn): return client attributes from database backends

MorganaFuture:
Release version: 7.0.0

Introduced in: N/A

## Summary

Fixes https://github.com/emqx/emqx/issues/16307

Only the HTTP backend could put `client_attrs` in an authentication result, so attributes
stored next to the credentials in a database were unreachable without standing up an external
authentication service.

Nothing downstream was missing: `emqx_channel:merge_auth_result/2` takes `client_attrs` from
whatever the authenticator returned and merges it into `clientinfo`, whichever backend
produced it. The gap was that `authn_result/1` in the database backends never put the key
there, so this stays local to the backends plus one helper.

`emqx_authn_utils:client_attrs/1` accepts two shapes, because the backends cannot all express
the same one: `client_attrs` holding a map, which is what an HTTP body or a Mongo subdocument
gives, and `client_attrs.<name>`, one key per attribute, which is what a SQL column alias or a
Redis hash field gives. A per-attribute key wins over the same name inside the map. Which of
the two you would rather standardise on is worth saying, though dropping either one leaves a
backend unable to report attributes at all.

Values are no longer required to be strings. `drop_invalid_attr/1` ran `iolist_to_binary/1`
over the value, which raises badarg on an integer; `emqx_authn_chains` catches that into an
`authenticator_error` and refuses the client, so a numeric column would have failed the whole
login rather than the one attribute. Numbers and booleans are converted, and a value with no
string representation, a NULL column arriving as `undefined` or `null`, is dropped with a log
line. That also removes the same latent failure for an HTTP backend returning a JSON number.

The backends use `maybe_client_attrs/1`, which omits the key when there is nothing to report,
the way `clientid_override/1` already does, so a result without attributes keeps exactly the
shape it had. `emqx_authn_redis_SUITE:t_authenticate` caught the first version, which grew a
`client_attrs => #{}` on every result.

Redis validates its HMGET fields against an allowlist and rejected `client_attrs.tier` at
configuration time, so `allowed_fields` now takes a `{prefix, P}` entry alongside exact names.
The SQL backends take a free-form query and Mongo reads the document, so neither needs one.

HTTP and JWT keep `client_attrs/1` and so keep returning the key unconditionally; they also
gain the per-attribute shape, since it is the same helper.

LDAP is untouched. It builds its result differently and #18140 covers it.

Tests: `emqx_authn_utils_tests` 15/15 (eunit), and against real servers
`emqx_authn_mysql_SUITE` 8/8, `emqx_authn_postgresql_SUITE` 11/11, `emqx_authn_redis_SUITE`
9/9, `emqx_authn_mongodb_SUITE` 9/9, plus `emqx_authn_http_SUITE` 41/41 and
`emqx_authn_jwt_SUITE` 31/31 since they share the modified helper.

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ce/feat-18605.en.md`
- [x] Schema changes are backward compatible or intentionally breaking (additive; a result without attributes is unchanged)


apps/emqx_auth/src/emqx_authn/emqx_authn_utils.erl
```diff
@@ -309,10 +353,40 @@ do_drop_invalid_attr([]) ->
 do_drop_invalid_attr([{K, V} | More]) ->
     case emqx_utils:is_restricted_str(K) of
         true ->
-            [{iolist_to_binary(K), iolist_to_binary(V)} | do_drop_invalid_attr(More)];
+            case attr_value(V) of
+                {ok, Value} ->
+                    [{iolist_to_binary(K), Value} | do_drop_invalid_attr(More)];
+                error ->
+                    ?SLOG(debug, #{msg => "invalid_client_attr_value_dropped", attr_name => K}, #{
```
zmstone: collect all drops and report at once.
also change log level to warning

apps/emqx_auth/src/emqx_authn/emqx_authn_utils.erl
```diff
@@ -309,10 +353,40 @@ do_drop_invalid_attr([]) ->
...
+%% `emqx_authn_chains' turns into an `authenticator_error' and the client is
+%% refused - a whole failed login over one unusable attribute.
+attr_value(V) when is_binary(V) -> {ok, V};
+attr_value(V) when is_integer(V) -> {ok, integer_to_binary(V)};
+attr_value(V) when is_float(V) -> {ok, float_to_binary(V, [short])};
+attr_value(true) ->
+    {ok, <<"true">>};
+attr_value(false) ->
+    {ok, <<"false">>};
+attr_value(V) when is_list(V) ->
```
zmstone: i wonder if this is reachable in practical cases.

because for strings, we expect binary.

a list of integers can be mistakenly formatted into garbled attribute string otherwise.

suggestion: drop this clause

zmstone: `iolist_to_binary/1` succeeds on a list of small integers, so this clause converts rather than
rejects the one value shape a database is most likely to produce that has no string meaning.

A MongoDB array field `[1, 2, 3]` reaches here as a list and becomes `<<1, 2, 3>>` — three C0 control
characters, accepted as an attribute value. A BSON array of strings fares no better: `[<<"a">>,
<<"b">>]` silently becomes `<<"ab">>`, which looks like a value the operator chose.

That contradicts what the comment directly above promises — *"Convert what has an obvious
representation and drop the rest"*. A list is exactly the case where there is no obvious
representation, and it is the only clause here that can produce a value the operator never wrote.

The values matter downstream: `client_attrs` feed authorization rules and topic templates. #18496
had to add a control-character guard on the auto-subscribe side for rendered topics, which is the
same class of value arriving by a different route.

The pre-existing `iolist_to_binary(V)` had the same hole, so this is not a regression — but the
change moves database values onto this path, where arrays are common in a way they are not in an
HTTP JSON body, so it is much more reachable now.

Suggest validating the produced binary rather than trusting `iolist_to_binary/1` — at minimum reject
what is not printable, e.g. `emqx_utils:is_mqtt_safe_utf8/1`, which exists for vetting bytes from
non-MQTT sources flowing into `ClientInfo`.

apps/emqx_auth/src/emqx_authn/emqx_authn_utils.erl
```diff
@@ -176,12 +177,55 @@ is_superuser(#{<<"is_superuser">> := Value}) ->
...
+maybe_client_attrs(Selected) ->
+    case client_attrs(Selected) of
+        #{client_attrs := Attrs} = Result when map_size(Attrs) > 0 ->
+            Result;
+        _ ->
+            #{}
+    end.
+
+attr_map(Attrs) when is_map(Attrs) -> Attrs;
+attr_map(_NotAMap) -> #{}.
```
zmstone: This drops a malformed `client_attrs` silently, while the value path three functions down logs
`invalid_client_attr_value_dropped` for the same class of problem. Worth a log line here too.

It is also a behaviour change for the HTTP and JWT backends that the PR body does not mention.
Before, a non-map `client_attrs` reached `drop_invalid_attr/1`, which has only an `is_map` clause, so
it raised `function_clause`; `emqx_authn_chains` turns that into an `authenticator_error` and the
client is refused. After this change the login succeeds with no attributes at all.

Failing the login over a malformed backend response was arguably the wrong behaviour, and this PR is
consistent in moving away from it, so I am not arguing for the crash. The concern is that an
authentication server which starts returning `"client_attrs": "{\"sn\":\"s1\"}"` — a JSON string
rather than an object, an easy mistake — used to fail loudly and now silently authorizes clients with
none of the attributes their authz rules expect. Silent is the worst outcome for that; the log line
makes it diagnosable.

The eunit case at `emqx_authn_utils_tests` pins the new behaviour, so the intent is clear — worth
saying so in the PR body as an HTTP/JWT-visible change rather than leaving it to be discovered.

changes/ee/feat-18605.en.md
```diff
@@ -0,0 +1,5 @@
+Database authentication backends can now return client attributes, which previously only the HTTP backend could do.
```
zmstone, line 1: move to ee

we do not release ce any more

apps/emqx_auth/src/emqx_authn/emqx_authn_utils.erl
MorganaFuture: Done, warning level and one line per authentication instead of one per attribute. Each name is tagged with whether the name or the value was the problem, so a row with several bad columns still says which is which.

apps/emqx_auth/src/emqx_authn/emqx_authn_utils.erl
MorganaFuture: Dropped it. Agreed on the reasoning: a list here is more likely a driver returning a column as a list of integers, and iolist_to_binary would render those bytes as whatever they spell. There was a test pinning the old flattening behaviour; it now asserts the drop instead, for a mixed iolist and an all-integer list.

changes/ee/feat-18605.en.md
MorganaFuture: Moved to changes/ee/.

Issue #18601, opened 2026-08-28
Durable timer worker discards every callback exception without logging


zmstone:
### What happened?

`emqx_durable_timer_worker:handle_timeout/5` wraps every durable-timer callback in a catch-all that discards the exception and logs nothing:

https://github.com/emqx/emqx/blob/c859d42a6f134224ab5041cc35d8a3367a264db3/apps/emqx_durable_timer/src/emqx_durable_timer_worker.erl#L469-L477

When a callback raises, the worker survives, no log line is produced, and the timer is treated as fired, so it is not retried. The failure leaves no trace at any log level.

Two callback modules implement the `emqx_durable_timer` behaviour, and a silent failure in either loses work an operator cannot observe:

- `emqx_persistent_session_ds_gc_timer` — deletes an expired durable session. A swallowed failure leaves the session state in place.
- `emqx_durable_will` — publishes a delayed will message. A swallowed failure drops the will message.

### What did you expect to happen?

A failing durable-timer callback should be logged with the exception, the stacktrace, the timer type and the key. Whether the timer is then retried or dropped is a separate decision, but the failure must be visible.

### How can we reproduce it (as minimally and precisely as possible)?

A reachable instance, confirmed by instrumented reproduction on `dev-60`:

`emqx_persistent_session_ds_state_v2:delete/2` matches the transaction result with `{atomic, _, _} =` after asserting the collection guard:

https://github.com/emqx/emqx/blob/c859d42a6f134224ab5041cc35d8a3367a264db3/apps/emqx/src/emqx_persistent_session_ds/emqx_persistent_session_ds_state_v2.erl#L320-L344

When the delete loses a race to a concurrent guard rewrite, `emqx_ds_pmap:tx_assert_guard/2` fails, the transaction returns `{error, unrecoverable, {precondition_failed, _}}`, and the match raises `{badmatch, {error, unrecoverable, {precondition_failed, _}}}`. Reached through `emqx_persistent_session_ds:session_drop/2` running inside the GC timer callback, that badmatch is swallowed by the catch-all above.

Reproduction used in the original investigation: connect an MQTT 5.0 client with Session Expiry Interval 0 so its disconnect arms a 0 ms session GC timer, block the timer's `session_drop` (meck on `emqx_persistent_session_ds_subs:on_session_drop/2`), let a new connection for the same client ID commit and rewrite the guard, then release the block. Observed: the badmatch is raised, the worker pid stays alive, and nothing is logged.

[...]

PR #18606, opened 2026-08-29
fix(durable_timer): log a failing timer callback

MorganaFuture:

Introduced in: pre-6.0

## Summary

Fixes https://github.com/emqx/emqx/issues/18601

`handle_timeout/5` caught every exception from a durable-timer callback with `_:_ -> ok`,
discarding the class, the reason and the stacktrace. The worker survived, the timer counted
as fired and was not retried, and nothing was written to the log at any level, so the lost
work left no trace at all.

The failure is now reported at error level with the timer type, the callback module, the key,
and the exception with its stacktrace. The value is deliberately left out: it is opaque
callback data, and for `emqx_durable_will` it is a will message payload.

Whether the timer should then be retried or dropped is a separate decision and is unchanged
here, as the issue asks.

Worth knowing for review: this app logs through `?tp/3` and has no `?SLOG` at all. Only the
three-argument form reaches the log in a release build, since snabbkaffe's `trace_prod.hrl`
compiles `?tp/2` down to `begin _ = EVT, ok end`. A two-argument trace point would have
produced nothing outside tests.

`t_080_callback_failure` asserts the event carries the type, key, exception and stacktrace,
and that the worker keeps firing later timers. With `emqx_durable_timer_worker.erl` reverted
and the case kept, it fails: the trace point never arrives.

Commits touching this app have not carried a changelog before, so say the word if you would
rather drop `changes/ce/fix-18606.en.md`.

Tests: `emqx_durable_timer_SUITE` 8/8, run in
ghcr.io/emqx/emqx-builder/6.1-8:1.18.3-27.3.4.2-8-ubuntu24.04.

Unrelated, but seen while running the suite: `t_030_cancellation` failed once in four runs
at the `emqx_ds:dirty_read` assertion on line 392, finding the dead hand record still there.
It passed in the other three, including with this change reverted, so it looks like the read
racing the delete rather than anything here.

## PR Checklist

- [x] Change log added: `changes/ce/fix-18606.en.md`

### DIFF
--- a/apps/emqx_durable_timer/src/emqx_durable_timer_worker.erl
+++ b/apps/emqx_durable_timer/src/emqx_durable_timer_worker.erl
@@ -472,7 +472,19 @@ handle_timeout(Type, CBM, Time, Key, Value) ->
         emqx_durable_timer:handle_durable_timeout(CBM, Key, Value),
         ok
     catch
-        _:_ ->
+        Class:Reason:Stacktrace ->
+            %% The timer is still treated as fired and is not retried, so this
+            %% log is the only trace the failure leaves. The value is left out:
+            %% it is opaque callback data, and for `emqx_durable_will' it is a
+            %% will message payload.
+            ?tp(error, ?tp_callback_failed, #{
+                type => Type,
+                cbm => CBM,
+                key => Key,
+                exception => Class,
+                reason => Reason,
+                stacktrace => Stacktrace
+            }),
             ok
     end.
 
--- a/apps/emqx_durable_timer/src/emqx_durable_timer_worker.erl
+++ b/apps/emqx_durable_timer/src/emqx_durable_timer_worker.erl
@@ -473,10 +473,10 @@ handle_timeout(Type, CBM, Time, Key, Value) ->
         ok
     catch
         Class:Reason:Stacktrace ->
-            %% The timer is still treated as fired and is not retried, so this
-            %% log is the only trace the failure leaves. The value is left out:
-            %% it is opaque callback data, and for `emqx_durable_will' it is a
-            %% will message payload.
+            %% A failing callback does not prevent the timer from being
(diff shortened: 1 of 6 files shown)


zmstone:
@MorganaFuture thank you for the pr. This one i will need help from team.
@ieQu1 please help to review both the issue and this PR.

apps/emqx_durable_timer/src/emqx_durable_timer_worker.erl
```diff
@@ -472,7 +472,19 @@ handle_timeout(Type, CBM, Time, Key, Value) ->
...
+            %% consumed: it is not retried. The value is not logged, because it
+            %% is opaque callback data and may hold user payload, such as an
+            %% `emqx_durable_will' will message.
+            ?tp(error, ?tp_callback_failed, #{
+                type => Type,
+                cbm => CBM,
+                key => Key,
+                exception => Class,
+                reason => Reason,
+                stacktrace => Stacktrace
```
ieQu1, line 486:
```suggestion
            ?tp(error, ?tp_callback_failed, #{
                Class => Reason,
                type => Type,
                cbm => CBM,
                key => Key,
                stacktrace => Stacktrace
```

apps/emqx_durable_timer/test/emqx_durable_timer_SUITE.erl
```diff
@@ -913,6 +913,59 @@ db_dump(Node) ->
 
 suite() -> [{timetrap, {minutes, 1}}].
 
+%% A callback that raises is reported with its exception, and the timer is
```
ieQu1, line 916: This PR merely adds a log message. I don't think we need a dedicated testcase to verify the log message (or tracepoint). Please remove this test and revert the related changes in `test_timer`.

Issue #18607, opened 2026-08-29
Split Dashboard self-service APIs from admin user-management (introduce /current_user/*)


zmstone:
## Summary

Introduce a dedicated **self-service** Dashboard API namespace — `/current_user/*` — for a user managing **their own** account (password, MFA, profile), separate from the admin endpoints that manage **other** users under `/users/:username/*`.

Today both self-service and admin actions share the `/users/:username/*` routes and are told apart by "is this me?" logic spread across four layers (RBAC viewer-self clauses + a path-parsing helper, a 7-row MFA authorization matrix, an `Authorization`-header re-parser, and a separate caller-resolution path). Splitting the two lets each side be authorized by one simple rule:

- **Self endpoints** — authorized by the authenticated identity alone (the caller *is* the subject). No scope, no role matrix.
- **Admin endpoints** — authorized by scope + administrator role; the target is always another user. No `IsSelf` special-casing.

This removes the MFA decision matrix and collapses the RBAC. It also simplifies the follow-up MFA-enforcement work in https://github.com/emqx/emqx/pull/18586, which then needs a single check instead of a fifth self-vs-other branch.

**Target: 7.0 (breaking API change).** This should land **before** https://github.com/emqx/emqx/pull/18586 so that PR rebases onto the clean split. Related: https://github.com/emqx/emqx/issues/18363, https://github.com/emqx/emqx/issues/18362.

## Proposed API

New self-service endpoints — identity-authorized, **no scope required**, operationIds/handlers prefixed `current_user_`:

- `GET    /current_user` — own profile: username, role, effective scopes, MFA status.
- `POST   /current_user/change_pwd` — change own password (`old_pwd` + `new_pwd`).
- `POST   /current_user/mfa` — self setup / rotate MFA.
- `DELETE /current_user/mfa` — self disable MFA.
- `GET    /current_user/mfa` — own MFA status (optional; may be folded into `GET /current_user`).

`/current_user/*` is a top-level path outside `/users/`, so it cannot collide with a username (`self`, `me`, `current_user` are all legal usernames — there is no reserved-word check).

Admin endpoints become purely admin (scope + administrator role; the target is always another user, self is routed to `/current_user/*`):

- `POST` / `DELETE /users/:username/mfa` — require `mfa_management` **and** administrator role. No self handling.
- `DELETE /users/:username` — the "cannot delete self" guard becomes trivially `target =/= caller`.
[...]

PR #18608, opened 2026-08-29
feat(dashboard): split self-service out of the admin user API

MorganaFuture:
Fixes https://github.com/emqx/emqx/issues/18607

Self-service and administration shared `/users/:username/*` and were told apart by an "is this me?" test spread across four layers: the viewer-self and namespaced-self RBAC clauses plus `is_self_service_endpoint/2`, a seven-row MFA matrix, an `Authorization` header re-parser (`is_self_auth/2`), and a separate caller-resolution path. Self-service moves to `/current_user/*`, where the caller is the subject, so each side gets one rule.

**New**

| | |
|---|---|
| `GET /current_user` | own profile: username, role, effective scopes, MFA status |
| `POST /current_user/change_pwd` | own password |
| `POST`/`DELETE /current_user/mfa` | own MFA |

Authorized by the authenticated identity alone: RBAC allows any authenticated dashboard user, and the paths are `?SCOPE_PUBLIC`, so a viewer with `scopes = []` still reaches its own account.

**Removed / narrowed**

* `POST /users/:username/change_pwd` stays as a **deprecated self-only shim** (per the [refinement](https://github.com/emqx/emqx/issues/18607#issuecomment-5461632246)): it asserts the path target is the caller, then delegates to the canonical route, and answers `403` otherwise. The old route accepted any target and merely failed the `old_pwd` check, so a cross-user call was refused by accident; now it is refused by rule. Resetting another user's password stays CLI-only.
* `/users/:username/mfa` is administrator-only, target is always another user. The handler refuses a self target so a self-change cannot be recorded as an `admin_override` decision.
* `mfa_management` becomes administrator-only. Its non-admin meaning was a self-exemption key over the holder's own MFA, and self-MFA now needs no scope at all.
* The seven-row matrix collapses to two rules: first-time setup always allowed, administrator-set MFA not changeable by the account holder.

**Three points the issue does not mention**

1. `emqx_mgmt_auth:authorize/4` denies API keys **by handler function name**, not by path. The new routes are `?SCOPE_PUBLIC`, so the scope layer would have let an API key through. Clauses added, and pinned in `emqx_mgmt_api_api_keys_SUITE`. This now covers the shim too, which no longer carries a scope either.
2. `change_password/3` is guarded on a binary username. The caller identity now comes from the token, which for SSO is a `{Backend, Name}` tuple, so this would have been a `function_clause` and a 500. `/current_user/change_pwd` returns 400 for SSO accounts instead.
3. Replacing `is_self_auth/2` with a comparison against the resolved admin key incidentally fixes the delete-self guard for SSO: the old header parser returned `false` for every SSO caller, so an SSO administrator could delete its own account.

`/current_user` is top-level, outside `/users/`, so a user named `current_user`, `self` or `me` neither shadows nor is shadowed by it (asserted).

**One fix outside the issue's scope**

Dropping `user_management` from the shim's route was not sufficient on its own. `classify_path/1` matches exact paths first, so a `?SCOPE_PUBLIC` entry that is a *template* was reachable only through segment matching, and `match_template_iter/2` skipped public entries there. `/users/:username/change_pwd` therefore classified as `not_found`, which `check_login_user_scopes_strict/2` fails closed for any user carrying an explicit scope list, so a viewer with `scopes = []` got a 403 from the scope layer before the handler ran. `/sso/login/:backend` has the same shape and is public too; it never surfaced because it is a pre-login route the scope layer does not run for.

`fix(mgmt): let a ?SCOPE_PUBLIC path template cover its concrete paths` makes scoped templates a first pass and public ones a second, so a scoped template still wins over a public one (`/sso/oidc` keeps resolving to `sso_management`, already asserted by `t_wildcard_does_not_claim_public_path`). Happy to split it out if you'd rather it did not ride along.

The umbrella has exactly **three** public templates, and all three move from `not_found` to `public`: `/users/:username/change_pwd`, `/sso/login/:backend`, and `/schemas/:name` (declared through the whole-module form `scopes() -> ?SCOPE_PUBLIC`). The third means a login user pinned to an explicit scope list can now read the static hocon/OpenAPI schema text that `emqx_dashboard_schema_api` already meant to serve "any authenticated Dashboard user". No scoped path is downgraded, and API keys are unaffected in either direction because `path_to_scope/1` collapses `public` and `not_found` to `undefined`.

**Tests**

New `emqx_dashboard_current_user_SUITE` (17 cases) covers all eight rows of the issue's coverage table, including the username-collision case and an assertion that the generated OpenAPI spec carries the split. Five of them are the shim: self-change works for viewer, global admin and namespaced admin (including with `scopes = []`); a non-self target is `403` for all three and leaves both passwords untouched; `old_pwd` is still verified; an SSO caller gets the "no local password" `400`; and the shim is marked `deprecated` in the spec while the canonical route is not. The old viewer-self / `is_self_service_endpoint` RBAC cases are deleted, not edited.

Green in `ghcr.io/emqx/emqx-builder/6.1-8:1.19.1-28.4.1-4-ubuntu24.04`: `current_user` 18/18, `user_scopes` 62/62, `rbac` 32/32, `mfa` 8/8, `sso_mfa` 41/41, `api_keys` 26/26, `key_scopes` 31/31, `api_spec` 19/19.

Four failures on this checkout are environmental, not from this branch:

* `emqx_dashboard_SUITE:t_security_response_headers` needs `priv/www/index.html` (downloaded by `make`).
* `emqx_swagger_parameter_SUITE:t_summary_i18n` needs `priv/desc.zh.hocon` (downloaded by `pre-compile.sh`).
* `sso_oidc` 4/13 and `sso_ldap` 6/10 need the compose-file Dex and LDAP servers. Both produce identical counts on an untouched `dev-70` tree.

The full CT run needs the compose stack and was not attempted locally.

**Follow-up outside this repo:** the Dashboard SPA and emqx-docs call the removed and narrowed routes and need to migrate with the cutover.

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_api.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_api.erl
@@ -30,8 +30,10 @@
     users/2,
     user_scopes/2,
     user/2,
-    change_pwd/2,
-    change_mfa/2
+    change_mfa/2,
+    current_user/2,
+    current_user_change_pwd/2,
+    current_user_mfa/2
 ]).
 -export([scopes/0]).
 
@@ -48,6 +50,7 @@
 -define(PASSWORD_LOGIN_DISABLED, 'PASSWORD_LOGIN_DISABLED').
 -define(SCRAM_CHALLENGE_INVALID, 'SCRAM_CHALLENGE_INVALID').
 -define(SERVICE_UNAVAILABLE, 'SERVICE_UNAVAILABLE').
+-define(MFA_LOCKED, 'MFA_LOCKED').
 
 namespace() -> "dashboard".
 
@@ -63,6 +66,14 @@ api_spec() ->
 %%   * /login -- pre-login (security => []).
 %%   * /logout -- any authenticated role may log itself out.
 %%   * /user_scopes -- static catalog endpoint, no tenant data.
+%%   * /current_user* -- self-service; the caller is the subject, so the
+%%     authenticated identity alone authorizes the operation. A scope
+%%     check here would let an explicit `scopes = []' lock a user out of
+%%     their own password and MFA, which is what the scope layer exists
+%%     to gate for OTHER users, not for the holder's own account. None of
+%%     these operations can widen the caller's own privileges:
+%%     change_pwd verifies `old_pwd', and the MFA routes only ever touch
+%%     the caller's own record.
 scopes() ->
     #{
         <<"/login">> => ?SCOPE_PUBLIC,
@@ -70,9 +81,11 @@ scopes() ->
         <<"/login/verify">> => ?SCOPE_PUBLIC,
         <<"/logout">> => ?SCOPE_PUBLIC,
         <<"/user_scopes">> => ?SCOPE_PUBLIC,
+        <<"/current_user">> => ?SCOPE_PUBLIC,
+        <<"/current_user/change_pwd">> => ?SCOPE_PUBLIC,
+        <<"/current_user/mfa">> => ?SCOPE_PUBLIC,
         <<"/users">> => ?SCOPE_USER_MGMT,
         <<"/users/:username">> => ?SCOPE_USER_MGMT,
-        <<"/users/:username/change_pwd">> => ?SCOPE_USER_MGMT,
         <<"/users/:username/mfa">> => ?SCOPE_MFA_MGMT
     }.
 
@@ -82,9 +95,11 @@ paths() ->
         "/login/challenge",
         "/login/verify",
         "/logout",
+        "/current_user",
+        "/current_user/change_pwd",
+        "/current_user/mfa",
         "/users",
         "/users/:username",
-        "/users/:username/change_pwd",
         "/users/:username/mfa",
         "/user_scopes"
     ].
@@ -165,6 +180,74 @@ schema("/logout") ->
             }
         }
     };
+%% Self-service routes. `/current_user' is a top-level path outside
+%% `/users/', so it cannot be shadowed by -- or shadow -- a user whose
+%% name happens to be `current_user', `self' or `me'; those are all
+%% legal usernames and there is no reserved-word check on them.
+%%
+%% Every operationId is prefixed `current_user_' because minirest
+%% requires globally unique operationIds and the admin routes already
+%% own `change_pwd' / `change_mfa'.
+schema("/current_user") ->
+    #{
+        'operationId' => current_user,
+        get => #{
+            tags => [<<"dashboard">>],
+            desc => ?DESC(current_user_api),
+            security => [#{'bearerAuth' => []}],
+            responses => #{
+                200 => current_user_fields(),
+                404 => response_schema(404)
+            }
+        }
+    };
+schema("/current_user/change_pwd") ->
+    #{
+        'operationId' => current_user_change_pwd,
+        post => #{
+            tags => [<<"dashboard">>],
+            desc => ?DESC(current_user_change_pwd_api),
+            security => [#{'bearerAuth' => []}],
+            'requestBody' => fields([old_pwd, new_pwd]),
+            responses => #{
+                204 => <<"Password is updated">>,
+                404 => response_schema(404),
+                400 => emqx_dashboard_swagger:error_codes(
+                    [?BAD_REQUEST, ?ERROR_PWD_NOT_MATCH, ?NOT_ALLOWED],
+                    ?DESC(login_failed_response400)
+                )
+            }
+        }
+    };
+schema("/current_user/mfa") ->
+    #{
+        'operationId' => current_user_mfa,
+        post => #{
+            tags => [<<"dashboard">>],
+            desc => ?DESC(current_user_mfa_api),
+            security => [#{'bearerAuth' => []}],
+            'requestBody' => emqx_dashboard_schema:mfa_fields(),
+            responses => #{
+                204 => <<"MFA setting is updated">>,
+                403 => emqx_dashboard_swagger:error_codes(
+                    [?MFA_LOCKED], ?DESC(current_user_mfa_locked)
+                ),
+                404 => response_schema(404)
+            }
+        },
+        delete => #{
+            tags => [<<"dashboard">>],
+            desc => ?DESC(delete_current_user_mfa_api),
+            security => [#{'bearerAuth' => []}],
+            responses => #{
+                204 => <<"MFA setting is disabled">>,
+                403 => emqx_dashboard_swagger:error_codes(
+                    [?MFA_LOCKED], ?DESC(current_user_mfa_locked)
+                ),
+                404 => response_schema(404)
+            }
+        }
+    };
 schema("/users") ->
     #{
         'operationId' => users,
@@ -237,24 +320,8 @@ schema("/users/:username") ->
             }
         }
     };
-schema("/users/:username/change_pwd") ->
-    #{
-        'operationId' => change_pwd,
-        post => #{
-            tags => [<<"dashboard">>],
-            desc => ?DESC(change_pwd_api),
-            parameters => fields([username_in_path]),
-            'requestBody' => fields([old_pwd, new_pwd]),
-            responses => #{
-                204 => <<"Update user password successfully">>,
-                404 => response_schema(404),
-                400 => emqx_dashboard_swagger:error_codes(
-                    [?BAD_REQUEST, ?ERROR_PWD_NOT_MATCH],
-                    ?DESC(login_failed_response400)
-                )
-            }
-        }
-    };
+%% Administrator-only: the target is always another user. A caller
+%% managing its own MFA uses `/current_user/mfa'.
 schema("/users/:username/mfa") ->
     #{
         'operationId' => change_mfa,
@@ -265,6 +332,9 @@ schema("/users/:username/mfa") ->
             'requestBody' => emqx_dashboard_schema:mfa_fields(),
             responses => #{
                 204 => <<"MFA setting is updated">>,
+                400 => emqx_dashboard_swagger:error_codes(
+                    [?NOT_ALLOWED], ?DESC(login_failed_response400)
+                ),
                 404 => response_schema(404)
             }
         },
@@ -274,6 +344,9 @@ schema("/users/:username/mfa") ->
             parameters => sso_parameters(fields([username_in_path])),
             responses => #{
                 204 => <<"MFA setting is disabled">>,
+                400 => emqx_dashboard_swagger:error_codes(
+                    [?NOT_ALLOWED], ?DESC(login_failed_response400)
+                ),
                 404 => response_schema(404)
             }
         }
@@ -295,6 +368,15 @@ fields(List) ->
 
 user_fields() ->
     fields([username, role, description, backend, scopes_response]) ++ ee_user_fields().
+
+%% Same shape as `user_fields/0', but `scopes' carries the EFFECTIVE
+%% list rather than the raw tri-state one. The admin routes report the
+%% raw value so that a read-modify-write of `PUT /users/:username' does
+%% not sediment the role default into an explicit list; this endpoint
+%% has no write counterpart, and its caller is asking what it may
+%% actually do, so the expanded list is the useful answer.
+current_user_fields() ->
+    fields([username, role, description, backend, effective_scopes_response]) ++ ee_user_fields().
 
 ee_user_fields() ->
     [
@@ -373,6 +455,15 @@ field(scopes_response) ->
             desc => ?DESC(user_scopes_response),
             required => false,
             example => [?SCOPE_USER_MGMT, ?SCOPE_MFA_MGMT]
+        })};
+field(effective_scopes_response) ->
+    %% Always a concrete list -- the role default is expanded, so the
+    %% `unset' sentinel of `scopes_response' cannot appear here.
+    {scopes,
+        mk(hoconsc:array(binary()), #{
+            desc => ?DESC(effective_user_scopes_response),
+            required => true,
+            example => [?SCOPE_MONITORING, ?SCOPE_CONNECTIONS]
         })};
 field(backend) ->
     {backend, mk(binary(), #{desc => ?DESC(backend), example => <<"local">>})};
@@ -758,9 +849,15 @@ is_default_admin(_NonLocalTarget) ->
     %% `dashboard.default_username'.
     false.
 
-handle_delete_user(#{bindings := #{username := Username0}, headers := Headers} = Req) ->
+handle_delete_user(#{bindings := #{username := Username0}} = Req) ->
     Username = username(Req, Username0),
-    case is_self_auth(Username0, Headers) of
+    %% The caller is identified by the authenticated token, not by
+    %% re-parsing the `Authorization' header: the header form varies
+    %% (basic / bearer, either capitalisation) and only the token
+    %% resolution knows the SSO key. `caller_key/1' returns the same
+    %% admin-record key `username/2' produces, so the two are directly
+    %% comparable for both local and SSO targets.
+    case caller_key(Req) =:= Username of
         true ->
             {400, ?NOT_ALLOWED, <<"Cannot delete self">>};
         false ->
@@ -775,40 +872,47 @@ handle_delete_user(#{bindings := #{username := Username0}, headers := Headers} =
             end
     end.
 
-is_self_auth(?SSO_USERNAME(_, _), _) ->
-    false;
-is_self_auth(Username, #{<<"authorization">> := Token}) ->
-    is_self_auth(Username, Token);
-is_self_auth(Username, #{<<"Authorization">> := Token}) ->
-    is_self_auth(Username, Token);
-is_self_auth(Username, <<"basic ", Token/binary>>) ->
-    is_self_auth_basic(Username, Token);
-is_self_auth(Username, <<"Basic ", Token/binary>>) ->
-    is_self_auth_basic(Username, Token);
-is_self_auth(Username, <<"bearer ", Token/binary>>) ->
-    is_self_auth_token(Username, Token);
-is_self_auth(Username, <<"Bearer ", Token/binary>>) ->
-    is_self_auth_token(Username, Token).
+%%--------------------------------------------------------------------
+%% Self-service handlers (`/current_user/*')
+%%
+%% The subject is the authenticated identity itself. There is no
+%% `:username' in the path, so there is nothing to spoof and nothing to
+%% compare: `caller_key/1' IS the target. Authorization is therefore
+%% complete once the request is authenticated -- RBAC lets any
+%% authenticated dashboard user through, and the scope layer treats
+%% these paths as public.
+%%--------------------------------------------------------------------
 
-is_self_auth_basic(Username, Token) ->
-    UP = base64:decode(Token),
-    case binary:match(UP, Username) of
-        {0, N} ->
-            binary:part(UP, {N, 1}) == <<":">>;
-        _ ->
-            false
-    end.
+current_user(get, Req) ->
+    with_caller(Req, fun(#?ADMIN{username = Username} = Admin) ->
+        Profile = emqx_dashboard_admin:to_external_user(Admin),
+        {200, Profile#{scopes => emqx_dashboard_admin:effective_scopes_of(Username)}}
+    end).
 
-is_self_auth_token(Username, Token) ->
-    case emqx_dashboard_token:owner(Token) of
-        {ok, Owner} ->
-            Owner == Username;
-        {error, _NotFound} ->
-            false
-    end.
+current_user_change_pwd(post, #{body := Params} = Req) ->
+    with_caller(Req, fun(#?ADMIN{username = Username}) ->
+        do_change_pwd(Username, Params)
+    end).
 
-change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
-    LogMeta = #{msg => "dashboard_change_password", username => binary_to_list(Username)},
+%% An SSO user has no local password -- the identity provider owns the
+%% credential -- so there is nothing here to change. Reject explicitly:
+%% `emqx_dashboard_admin:change_password/3' is guarded on a binary
+%% username and would otherwise raise a function_clause on the
+%% `?SSO_USERNAME' tuple, turning this into a 500.
+do_change_pwd(?SSO_USERNAME(Backend, Name), _Params) ->
+    ?SLOG(warning, #{
+        msg => "dashboard_change_password",
+        username => Name,
+        backend => Backend,
+        result => denied,
+        reason => "sso_user_has_no_local_password"
+    }),
+    {400, ?NOT_ALLOWED, <<
+        "This account signs in through an SSO backend and has no local "
+        "password. Change it with the identity provider instead."
+    >>};
+do_change_pwd(Username, Params) when is_binary(Username) ->
+    LogMeta = #{msg => "dashboard_change_password", username => Username},
     OldPwd = maps:get(<<"old_pwd">>, Params),
     NewPwd = maps:get(<<"new_pwd">>, Params),
     case ?EMPTY(OldPwd) orelse ?EMPTY(NewPwd) of
@@ -832,23 +936,99 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
             end
     end.
 
+current_user_mfa(post, #{body := Settings} = Req) ->
+    Mechanism = maps:get(<<"mechanism">>, Settings),
+    with_caller(Req, fun(#?ADMIN{username = Username}) ->
+        LogMeta = #{msg => "dashboard_user_mfa_setup", username => Username},
+        case authorize_self_mfa(Username, setup) of
+            ok ->
+                %% Never `ByAdmin': a self-initiated (re)init must not
+                %% touch the admin_override decision.
+                mfa_result(emqx_dashboard_admin:reinit_mfa(Username, Mechanism, false), LogMeta);
+            {deny, Code, ErrCode, Msg} ->
+                ?SLOG(warning, LogMeta#{result => denied, reason => ErrCode}),
+                {Code, ErrCode, Msg}
+        end
+    end);
+current_user_mfa(delete, Req) ->
+    with_caller(Req, fun(#?ADMIN{username = Username}) ->
+        LogMeta = #{msg => "dashboard_user_mfa_disable", username => Username},
+        case authorize_self_mfa(Username, disable) of
+            ok ->
+                mfa_result(emqx_dashboard_admin:disable_mfa(Username, false), LogMeta);
+            {deny, Code, ErrCode, Msg} ->
+                ?SLOG(warning, LogMeta#{result => denied, reason => ErrCode}),
+                {Code, ErrCode, Msg}
+        end
+    end).
+
+mfa_result(ok, LogMeta) ->
+    ?SLOG(info, LogMeta#{result => success}),
+    {204};
+mfa_result({error, <<"username_not_found">>}, LogMeta) ->
+    ?SLOG(error, LogMeta#{result => failed, reason => "username not found"}),
+    {404, ?USER_NOT_FOUND, <<"User not found">>};
+mfa_result({error, Reason}, LogMeta) ->
+    ?SLOG(error, LogMeta#{result => failed, reason => Reason}),
+    {400, ?BAD_REQUEST, Reason}.
+
+%% Self-MFA policy, in full:
+%%
+%%   first-time setup           => allow  (deadlock prevention: a user
+%%                                         that has never enrolled must
+%%                                         always be able to)
+%%   admin_override = required  => deny   (an administrator reset this
+%%                                         account's MFA; only another
+%%                                         administrator or the CLI can
+%%                                         lift it)
+%%   otherwise                  => allow
+%%
+%% `admin_override' is only ever written when an administrator acts on
+%% ANOTHER user (`emqx_dashboard_api:change_mfa/2' passes ByAdmin), so a
+%% user cannot lock themselves out by rotating their own MFA.
+authorize_self_mfa(Username, Op) ->
+    case is_first_time_setup(Username, Op) orelse not self_mfa_locked(Username) of
+        true ->
+            ok;
+        false ->
+            {deny, 403, ?MFA_LOCKED, <<
+                "MFA for this account was set by an administrator and "
+                "cannot be changed here. Ask an administrator to reset it."
+            >>}
+    end.
+
+self_mfa_locked(Username) ->
+    emqx_dashboard_admin:admin_override_of(Username) =:= ?ADMIN_MFA_REQUIRED.
+
+is_first_time_setup(_Username, disable) ->
+    false;
+is_first_time_setup(Username, setup) ->
+    case emqx_dashboard_admin:get_mfa_state(Username) of
+        {ok, _} -> false;
+        _ -> true
+    end.
+
+%% Resolve the caller's own admin record from the bearer token's
+%% `source' -- the key `emqx_dashboard:authorize/2' already resolved
+%% (a `?SSO_USERNAME' tuple for SSO users, a plain binary otherwise).
+with_caller(Req, Fun) ->
+    case caller_admin(Req) of
+        #?ADMIN{} = Admin ->
+            Fun(Admin);
+        undefined ->
+            %% Unreachable in practice: these routes are bearer-only and
+            %% API keys are refused for them in `emqx_mgmt_auth:authorize/4'.
+            %% Reachable only if the account is deleted between login and
+            %% this request.
+            {404, ?USER_NOT_FOUND, <<"User not found">>}
+    end.
+
 change_mfa(delete, #{bindings := #{username := Username0}} = Req) ->
     Username = username(Req, Username0),
     LogMeta = #{msg => "dashboard_user_mfa_disable", username => Username},
-    case authorize_mfa_change(Req, Username, disable) of
+    case reject_self_target(Req, Username) of
         ok ->
-            ByAdmin = caller_is_admin_acting_on_other(Req, Username),
-            case emqx_dashboard_admin:disable_mfa(Username, ByAdmin) of
-                ok ->
-                    ?SLOG(info, LogMeta#{result => success}),
-                    {204};
-                {error, <<"username_not_found">>} ->
-                    ?SLOG(error, LogMeta#{result => failed, reason => "username not found"}),
-                    {404, ?USER_NOT_FOUND, <<"User not found">>};
-                {error, Reason} ->
-                    ?SLOG(error, LogMeta#{result => failed, reason => Reason}),
-                    {400, ?BAD_REQUEST, Reason}
-            end;
+            mfa_result(emqx_dashboard_admin:disable_mfa(Username, true), LogMeta);
         {deny, Code, ErrCode, Msg} ->
             ?SLOG(warning, LogMeta#{result => denied, reason => ErrCode}),
             {Code, ErrCode, Msg}
@@ -857,23 +1037,28 @@ change_mfa(post, #{bindings := #{username := Username0}, body := Settings} = Req
     Username = username(Req, Username0),
     Mechanism = maps:get(<<"mechanism">>, Settings),
     LogMeta = #{msg => "dashboard_user_mfa_setup", username => Username},
-    case authorize_mfa_change(Req, Username, setup) of
+    case reject_self_target(Req, Username) of
         ok ->
-            ByAdmin = caller_is_admin_acting_on_other(Req, Username),
-            case emqx_dashboard_admin:reinit_mfa(Username, Mechanism, ByAdmin) of
-                ok ->
-                    ?SLOG(info, LogMeta#{result => success}),
-                    {204};
-                {error, <<"username_not_found">>} ->
-                    ?SLOG(error, LogMeta#{result => failed, reason => "username not found"}),
-                    {404, ?USER_NOT_FOUND, <<"User not found">>};
-                {error, Reason} ->
-                    ?SLOG(error, LogMeta#{result => failed, reason => Reason}),
-                    {400, ?BAD_REQUEST, Reason}
-            end;
+            mfa_result(emqx_dashboard_admin:reinit_mfa(Username, Mechanism, true), LogMeta);
         {deny, Code, ErrCode, Msg} ->
             ?SLOG(warning, LogMeta#{result => denied, reason => ErrCode}),
             {Code, ErrCode, Msg}
+    end.
+
+%% The admin MFA routes act on other users only; the caller's own
+%% account is served by `/current_user/mfa'. Without this guard an
+%% administrator could route a self-change through the admin path,
+%% where it would be recorded as an administrator decision
+%% (`admin_override') and would skip the self policy entirely.
+reject_self_target(Req, TargetUsername) ->
+    case caller_key(Req) =:= TargetUsername of
+        false ->
+            ok;
+        true ->
+            {deny, 400, ?NOT_ALLOWED, <<
+                "This endpoint manages other users' MFA. "
+                "Use /current_user/mfa to manage your own."
+            >>}
     end.
 
 register_unsuccessful_login(Username, <<"password_error">>) ->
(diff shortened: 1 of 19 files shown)


apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -82,9 +95,11 @@ paths() ->
         "/login/challenge",
         "/login/verify",
         "/logout",
+        "/current_user",
+        "/current_user/change_pwd",
+        "/current_user/mfa",
         "/users",
         "/users/:username",
-        "/users/:username/change_pwd",
```
zmstone, line 87: i regret the original decision.
removing is a bit too aggressive.
changed the request here:
https://github.com/emqx/emqx/issues/18607#issuecomment-5461632246

rel/i18n/emqx_dashboard_api.hocon
```diff
@@ -140,4 +135,30 @@ change_mfa.label:
...
+current_user_mfa_api.label:
+"""Change own MFA"""
+
+delete_current_user_mfa_api.desc:
+"""Disable MFA for the signed-in user's own account. To reset or re-key TOTP, call the `POST` method."""
+delete_current_user_mfa_api.label:
+"""Delete own MFA"""
+
+current_user_mfa_locked.desc:
+"""MFA for this account was configured by an administrator and cannot be changed by the account holder. An administrator must reset it."""
```
zmstone:
```suggestion
"""MFA for this account was configured by an administrator and cannot be changed by the account holder."""
```

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -758,9 +849,15 @@ is_default_admin(_NonLocalTarget) ->
...
-handle_delete_user(#{bindings := #{username := Username0}, headers := Headers} = Req) ->
+handle_delete_user(#{bindings := #{username := Username0}} = Req) ->
     Username = username(Req, Username0),
-    case is_self_auth(Username0, Headers) of
+    %% The caller is identified by the authenticated token, not by
+    %% re-parsing the `Authorization' header: the header form varies
+    %% (basic / bearer, either capitalisation) and only the token
+    %% resolution knows the SSO key. `caller_key/1' returns the same
+    %% admin-record key `username/2' produces, so the two are directly
+    %% comparable for both local and SSO targets.
```
zmstone:
```suggestion
```

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: Done, the shim is in.

One thing worth a look: dropping `user_management` from the route was not enough on its own. `classify_path/1` matches exact paths first, so a `?SCOPE_PUBLIC` entry that is a template was only reachable through segment matching, and `match_template_iter/2` skipped public entries there. `/users/:username/change_pwd` therefore classified as `not_found`, which `check_login_user_scopes_strict/2` fails closed for anyone carrying an explicit scope list, so a viewer with `scopes = []` got a 403 from the scope layer before the handler ran. `/sso/login/:backend` has the same shape and is public too, it just never showed because it is pre-login.

Fixed in a separate commit: scoped templates are tried first, public ones only if nothing matched, so a scoped template still wins over a public one. Say the word if you would rather I kept that out of this PR.

The 403 for a non-self target is asserted for viewer, global admin and namespaced admin, and an SSO caller now reaches the delegate and gets the "no local password" 400 instead of a misleading 403.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -832,23 +1008,99 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
...
+authorize_self_mfa(Username, Op) ->
+    case is_first_time_setup(Username, Op) orelse not self_mfa_locked(Username) of
+        true ->
+            ok;
+        false ->
+            {deny, 403, ?MFA_LOCKED, <<
+                "MFA for this account was set by an administrator and "
+                "cannot be changed here. Ask an administrator to reset it."
+            >>}
+    end.
```
zmstone: split it into two sub functions (not nested case clauses for readability), and return different 403 messages.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -832,23 +1008,99 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
...
+
+self_mfa_locked(Username) ->
```
zmstone: please double check, if admin_overreide = required means all changes are blocked -- which would be an odd design.

if not, we should check Op here too.
that is, if AMDIN_FMA_REQUIRED, then rolling TOTP should be allowed, but not disable.

maybe wording is the source of confusion.
amdin_override = required, means this user cannot turn it off.
MFA_LOCKED implies all changes are blocked.
so maybe change it to MFA_ENFORCED.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -832,23 +1012,99 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
...
+%%                                         that has never enrolled must
+%%                                         always be able to)
+%%   admin_override = required  => deny   (an administrator reset this
+%%                                         account's MFA; only another
+%%                                         administrator or the CLI can
+%%                                         lift it)
+%%   otherwise                  => allow
+%%
+%% `admin_override' is only ever written when an administrator acts on
+%% ANOTHER user (`emqx_dashboard_api:change_mfa/2' passes ByAdmin), so a
```
zmstone, line 1026: this function routes both setup/rotate and disable through the same check, 
so when `admin_override = required` a user can't change their MFA at all
not even roll their TOTP. That's stricter than it needs to be: 
rotation keeps MFA enabled, so it doesn't weaken the "MFA required" state the lock is protecting. 

let's gate on Op here, when locked, allow setup/rotate and deny only disable.

Loosening it is safe: a self-initiated rotate is reinit_mfa(ByAdmin=false), which doesn't touch admin_override, so the lock persists and MFA stays on. It also matches the default_mfa behavior in https://github.com/emqx/emqx/pull/18586, which denies self-disable but keeps rotation available.

On the error code: MFA_LOCKED reads as "all changes blocked," which won't be true once this is disable-only, rename it to `MFA_ADMIN_REQUIRED` would be less confusing.

zmstone (commented):
will continue review later.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
zmstone: not longer relevant, see the next comment.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -857,25 +1113,30 @@ change_mfa(post, #{bindings := #{username := Username0}, body := Settings} = Req
...
             {Code, ErrCode, Msg}
     end.
 
+%% The admin MFA routes act on other users only; the caller's own
+%% account is served by `/current_user/mfa'. Without this guard an
+%% administrator could route a self-change through the admin path,
+%% where it would be recorded as an administrator decision
+%% (`admin_override') and would skip the self policy entirely.
+reject_self_target(Req, TargetUsername) ->
+    case caller_key(Req) =:= TargetUsername of
```
zmstone: `reject_self_target` returns `400 NOT_ALLOWED` for "you targeted your own account on the admin MFA route," but the analogous case in the `change_pwd` shim — targeting *another* user on the self route — returns `403 NOT_ALLOWED` (same `?NOT_ALLOWED` code). Same error code, two HTTP statuses for the same "wrong route for this target" class. Suggest aligning them; `403` reads better here (the request is well-formed and the caller is authenticated — the target is simply not permitted on this route).

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -775,40 +913,82 @@ handle_delete_user(#{bindings := #{username := Username0}, headers := Headers} =
...
+
+%% Deprecated shim: assert the path target is the caller, then run the
+%% canonical self operation. It always acts on `Username', the caller's
+%% own resolved key, never on the path segment, so the path cannot steer
+%% the write even if the match below were ever loosened.
+change_pwd(post, #{bindings := #{username := Target}, body := Params} = Req) ->
+    with_caller(Req, fun(#?ADMIN{username = Username}) ->
+        case is_self_target(Username, Target) of
+            true ->
+                do_change_pwd(Username, Params);
```
zmstone, line 918: On a denied cross-user password-change attempt this logs `username => Target` (the account someone tried to change) but not the caller who attempted it. A denied attempt to change *another* user's password is exactly the event an audit trail wants attributed — add the caller's resolved key, e.g. `attempted_by => Username`. Without it the log shows which account was targeted but not who tried.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: done. setup is no longer gated at all, only disable checks admin_override, and the code is now MFA_ADMIN_REQUIRED.

one deviation: instead of matching on Op inside the two-arg form i dropped it and kept `authorize_self_mfa_disable/1`. with the setup clause always returning ok the deny branch in the post handler is unreachable and dialyzer flags it, and the 403 in the post route schema becomes a status that cannot happen, so both are gone.

is_first_time_setup/2 went with it, the exemption only existed because rotate was blocked. t_self_can_rotate_when_admin_override_required asserts the rotate succeeds and leaves admin_override in place, then that the disable is still refused.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: added, `attempted_by => Username` next to the target.

rel/i18n/emqx_dashboard_api.hocon
MorganaFuture: took the wording but not verbatim. with the gate now on Op only the disable is refused, so "cannot be changed by the account holder" would be wrong. it reads:

MFA is required on this account by an administrator, so the account holder cannot turn it off. Re-keying it is still allowed.

also renamed the key to current_user_mfa_admin_required so it matches the error code.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: removed.

zmstone (commented):
look good to me
@savonarola should also have a look. 

thank you both.

zmstone:
The two failing cases in emqx_dashboard_user_scopes_SUITE are a test-code issue, 
not a regression in the PR's production code

both cases fail with {badmatch, {ok, 401, <<"BAD_TOKEN ...">>}}

the request is rejected at token validation, before the MFA/RBAC logic runs.

cause:
`POST /current_user/mfa` (setup/rotate) calls `reinit_mfa(Username, _, ByAdmin=false)`, 
which runs `emqx_dashboard_token:destroy_by_username(Username)`. 

Re-keying MFA deliberately invalidates the account's sessions, 
so the caller's bearer token is dead after `setup_own_mfa/1`. 
Both cases then reuse that same token for delete_own_mfa/1, 
so they get 401 BAD_TOKEN instead of the expected 403.

The tests should login again to get a new token.

MorganaFuture:
CI caught two of my own cases: `t_self_can_rotate_when_admin_override_required` and `t_self_with_mfa_mgmt_still_required` reused the token that authorized the rotate for the disable after it. `reinit_mfa/3` destroys the account's tokens, so those answered 401 instead of 403. Fixed by taking a fresh token, 62/62 locally.

savonarola:
Thanks! Still reviewing.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -63,16 +67,33 @@ api_spec() ->
...
 %%   * /logout -- any authenticated role may log itself out.
 %%   * /user_scopes -- static catalog endpoint, no tenant data.
+%%   * /current_user* -- self-service; the caller is the subject, so the
+%%     authenticated identity alone authorizes the operation. A scope
+%%     check here would let an explicit `scopes = []' lock a user out of
+%%     their own password and MFA, which is what the scope layer exists
+%%     to gate for OTHER users, not for the holder's own account. None of
+%%     these operations can widen the caller's own privileges:
+%%     change_pwd verifies `old_pwd', and the MFA routes only ever touch
+%%     the caller's own record.
```
savonarola, line 72:
```suggestion
%%   * /current_user* -- self-service; the caller is the subject, so the
%%     authenticated identity alone authorizes the operation. These routes only ever touch
%%     the caller's own record.
```

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -63,16 +67,33 @@ api_spec() ->
...
+%%   * /users/:username/change_pwd -- the deprecated self-only shim. It
+%%     carries no scope for the same reason as `/current_user/change_pwd':
+%%     the handler asserts the target is the caller and then acts on the
+%%     caller's own record, so there is nothing here to gate per-scope.
+%%     Dropping `user_management' is what lets a viewer keep using its
+%%     old password-change call.
```
savonarola:
```suggestion
%%   * /users/:username/change_pwd -- the deprecated self-only shim. It
%%     carries no scope for the same reason as `/current_user/change_pwd':
%%     the handler asserts the target is the caller and then acts on the
%%     caller's own record, so there is nothing here to gate per-scope.
```

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -165,6 +189,71 @@ schema("/logout") ->
...
         }
     };
+%% Self-service routes. `/current_user' is a top-level path outside
+%% `/users/', so it cannot be shadowed by -- or shadow -- a user whose
+%% name happens to be `current_user', `self' or `me'; those are all
+%% legal usernames and there is no reserved-word check on them.
+%%
+%% Every operationId is prefixed `current_user_' because minirest
+%% requires globally unique operationIds and the admin routes already
+%% own `change_pwd' / `change_mfa'.
```
savonarola:
```suggestion
```

No need to copy motivational reasoning into the code; it's self-evident that different routes do not shadow each other.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -237,24 +326,40 @@ schema("/users/:username") ->
...
     };
+%% Deprecated self-only shim for `/current_user/change_pwd'. Kept so the
+%% heavily-integrated password-change call keeps working; scheduled for
+%% removal a release later.
+%%
+%% This is not the old endpoint with a new name. The old one was reachable
+%% for any target and merely failed on the `old_pwd' check, so a cross-user
+%% call was refused by accident. The shim asserts the target is the caller
+%% and answers 403 otherwise, which also makes the namespaced-administrator
+%% invariant deliberate rather than incidental.
```
savonarola:
```suggestion
%% Deprecated self-only shim for `/current_user/change_pwd'. Kept so the
%% heavily-integrated password-change call keeps working; scheduled for
%% removal a release later.
%% The shim asserts the target is the caller and answers 403 otherwise.
```

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -296,6 +407,15 @@ fields(List) ->
 user_fields() ->
     fields([username, role, description, backend, scopes_response]) ++ ee_user_fields().
 
+%% Same shape as `user_fields/0', but `scopes' carries the EFFECTIVE
+%% list rather than the raw tri-state one. The admin routes report the
+%% raw value so that a read-modify-write of `PUT /users/:username' does
+%% not sediment the role default into an explicit list; this endpoint
+%% has no write counterpart, and its caller is asking what it may
+%% actually do, so the expanded list is the useful answer.
```
savonarola:
```suggestion
%% Same fields as `user_fields/0', except `scopes' contains the effective
%% scope list. Admin endpoints return the stored value so a `PUT
%% /users/:username' round-trip does not replace implicit role defaults
%% with an explicit list. This endpoint is read-only and reports the
%% permissions the current user actually has.
```

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -857,25 +1090,30 @@ change_mfa(post, #{bindings := #{username := Username0}, body := Settings} = Req
...
         {deny, Code, ErrCode, Msg} ->
             ?SLOG(warning, LogMeta#{result => denied, reason => ErrCode}),
             {Code, ErrCode, Msg}
     end.
 
+%% The admin MFA routes act on other users only; the caller's own
+%% account is served by `/current_user/mfa'. Without this guard an
+%% administrator could route a self-change through the admin path,
+%% where it would be recorded as an administrator decision
+%% (`admin_override') and would skip the self policy entirely.
```
savonarola, line 1080: @zmstone accessing own account through administrator route _is_ an administrative action, do we really should to forbid it?

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -1203,10 +1354,15 @@ caller_admin(#{auth_meta := #{auth_type := jwt_token, source := Username}}) ->
 caller_admin(_) ->
     undefined.
 
-caller_is_admin_acting_on_other(Req, TargetUsername) ->
+%% The caller's own admin-record key, in the same shape `username/2'
+%% builds for a path target, so the two compare directly. `undefined'
+%% for a non-bearer caller or a deleted account; it never equals a real
+%% target key, so a self-check against it is false, which is the safe
+%% answer for both call sites (delete and the admin MFA routes).
```
savonarola: This is quite indirectly enforced (through minirest & cowboy), I would not claim the safety of comparison and better add expicit check for `undefined` values on the call sites like these:
```erlang
handle_delete_user(#{bindings := #{username := Username}} = Req) ->
   ...
    case caller_key(Req) of
        undefined -> ....
        Username when is_binary(Username)  -> ...
        {_Backend, Username} when is_binary(Username) ->  ...
        ...
```

apps/emqx_dashboard/test/emqx_dashboard_current_user_SUITE.erl
```diff
@@ -0,0 +1,519 @@
...
+%% HTTP coverage for the self-service / administrator split:
+%%
+%%   /current_user            GET     own profile
+%%   /current_user/change_pwd POST    own password
+%%   /current_user/mfa        POST    own MFA setup / rotate
+%%                            DELETE  own MFA disable
+%%
+%% Self endpoints are authorized by the authenticated identity alone --
+%% no scope, no role, no `:username' in the path. The administrator
+%% routes under `/users/:username/*' are the mirror image: scope plus
```
savonarola: Please avoid comparing administrative actions with self-service, as they are separate entities and are authorized independently. There is no symmetry.
We should test them independently, and neither test should require the other to behave in the opposite way.
Also, please avoid using metaphors and keep the comments in plain English.

apps/emqx_dashboard/test/emqx_dashboard_current_user_SUITE.erl
```diff
@@ -0,0 +1,519 @@
...
+-include_lib("emqx_utils/include/emqx_api_key_scopes.hrl").
+-include("emqx_dashboard.hrl").
+
+-define(HOST, "http://127.0.0.1:18083").
+-define(BASE_PATH, "/api/v5").
+
+-define(PASSWORD, <<"P@ssw0rd_1">>).
+-define(NEW_PASSWORD, <<"P@ssw0rd_2">>).
+
+-define(EE_ONLY(EXPR, NON_EE),
```
savonarola: We don't need this in v7, the version is always EE.

apps/emqx_dashboard/test/emqx_dashboard_current_user_SUITE.erl
```diff
@@ -0,0 +1,519 @@
...
+%%--------------------------------------------------------------------
+%% Routing
+%%--------------------------------------------------------------------
+
+%% `current_user', `self' and `me' are all legal usernames and there is
+%% no reserved-word check on them. `/current_user' is a top-level path
+%% outside `/users/', so such a user neither shadows nor is shadowed by
+%% the self-service routes, and an administrator still manages it by
+%% name.
+t_username_colliding_with_self_route(_Config) ->
```
savonarola: Probably no need to test routing, this test basically tests `minirest`, I suggest removing

apps/emqx_dashboard/test/emqx_dashboard_current_user_SUITE.erl
```diff
@@ -0,0 +1,519 @@
...
+    ),
+    ?assertMatch(
+        #{<<"post">> := _, <<"delete">> := _}, maps:get(<<"/current_user/mfa">>, Paths)
+    ),
+    %% The administrator MFA route stays, and so does the deprecated
+    %% change_pwd shim (see t_shim_is_marked_deprecated).
+    ?assertMatch(
+        #{<<"post">> := _, <<"delete">> := _}, maps:get(<<"/users/{username}/mfa">>, Paths)
+    ),
+    ?assertMatch(#{<<"post">> := _}, maps:get(<<"/users/{username}/change_pwd">>, Paths)).
```
savonarola: These are also testing minirest, I suggest removing.

apps/emqx_dashboard/test/emqx_dashboard_user_scopes_SUITE.erl
```diff
@@ -481,9 +475,12 @@ t_viewer_cannot_hold_user_management(_Config) ->
...
     ).
 
-%% Viewer CAN hold mfa_management — non-admin self-exemption rule
-%% (viewer self-exemption rule).
-t_viewer_can_hold_mfa_management(_Config) ->
+%% Viewer cannot hold mfa_management: the scope means "manage another
+%% user's MFA" and is administrator-only. Its former non-administrator
+%% meaning (a self-exemption key) is gone -- managing one's own MFA is
+%% an identity-authorized operation on /current_user/mfa and needs no
+%% scope.
```
savonarola:
```suggestion
%% Viewer cannot hold mfa_management: the scope means "manage another
%% user's MFA" and is administrator-only.
```
No need to track changes in the comments unless they explain compatibility shims.

apps/emqx_dashboard/test/emqx_dashboard_user_scopes_SUITE.erl
```diff
@@ -1040,49 +1040,37 @@ t_role_demotion_with_compatible_persisted_scopes_succeeds(_Config) ->
     ?assertEqual(?ROLE_VIEWER, Admin#?ADMIN.role).
 
 %%--------------------------------------------------------------------
-%% MFA self-lock matrix (the 7-line decision table)
+%% Self-MFA policy (`/current_user/mfa')
+%%
+%% What used to be a seven-row matrix over {IsFirstSetup, IsSelf,
```
savonarola: Pls do not reference removed logic, describe the actual state.

apps/emqx_dashboard/test/emqx_dashboard_user_scopes_SUITE.erl
```diff
@@ -1127,69 +1112,48 @@ t_self_cannot_rotate_when_admin_override_required_locked(_Config) ->
...
-%% mfa_management scope self-exempts from both.
-t_self_with_mfa_mgmt_can_rotate_under_admin_override_required_lock(_Config) ->
+    {ok, 403, RespBody} = delete_own_mfa(Token),
+    ?assertEqual(<<"MFA_ADMIN_REQUIRED">>, error_code(RespBody)).
+
+%% `mfa_management' is no longer a self-exemption key. It is now an
+%% administrator-only scope, so the holder must be an administrator --
+%% and even then it does not lift the requirement on their own account.
+%% Replaces the two former "self with mfa_management can rotate/delete
+%% under lock" rows.
```
savonarola: Pls document the current state of things.

apps/emqx_dashboard/test/emqx_dashboard_user_scopes_SUITE.erl
```diff
@@ -1127,69 +1112,48 @@ t_self_cannot_rotate_when_admin_override_required_locked(_Config) ->
...
-
-%% Self-DELETE under admin_override=mfa_required without mfa_management — denied.
-t_self_cannot_delete_when_admin_override_required_locked(_Config) ->
+    ?assertMatch({ok, 204, _}, setup_own_mfa(jwt(<<"u">>, test_password()))),
+    %% Fresh token: the rotate above invalidated the account's sessions.
+    {ok, 403, DeleteBody} = delete_own_mfa(jwt(<<"u">>, test_password())),
+    ?assertEqual(<<"MFA_ADMIN_REQUIRED">>, error_code(DeleteBody)).
+
+%% A viewer with an explicitly emptied scope list still reaches its own
+%% MFA: self-service is not scope-gated.
```
savonarola, line 1133: Pls document actual state of things, no need to mention how things were.

apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl
```diff
@@ -256,38 +256,52 @@ t_logout(TCConfig) when is_list(TCConfig) ->
     {ok, #{actor := Username}} = emqx_dashboard_admin:verify_token(FakeReq, FakeHandlerInfo, Token),
     ok.
 
-t_change_pwd(_) ->
-    Viewer1 = <<"viewer1">>,
-    Viewer2 = <<"viewer2">>,
-    SuperUser = <<"super_user">>,
+%% Self-service moved to `/current_user/*'. RBAC's rule for those routes
```
savonarola: Pls do not describe code changes in comments.

apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl
```diff
@@ -296,19 +310,17 @@ t_setup_mfa(_) ->
...
+%% Descendant of the #17122 regression (SSO usernames may contain `@`,
+%% which the HTTP layer percent-encodes as `%40`, and RBAC used to match
+%% the decoded path segment against the logged-in actor).
+%%
+%% `/current_user/mfa' carries no username segment at all, so there is
+%% nothing to encode, decode or compare -- the class of bug is gone
+%% rather than fixed. What this test still pins is the part that
+%% survives: the self-MFA lock is driven by the per-user
+%% `admin_override' field, not by the SSO backend's live `force_mfa'
+%% flag, and it applies to an SSO identity whose name needs escaping.
```
savonarola: Pls describe current state of things.

apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl
```diff
@@ -438,66 +450,74 @@ t_check_login_user_scopes_undefined_falls_back(_) ->
         emqx_dashboard_rbac:check_login_user_scopes(Username, <<"/clients">>)
     ).
 
-%% Self-targeted user endpoints (own change_pwd, own MFA) bypass the
-%% scope check — they are gated by RBAC's self-check and, for MFA, by
-%% emqx_dashboard_api:authorize_mfa_change/3.
-t_check_login_user_scopes_self_user_endpoints_bypass(_) ->
+%% Self-service is unscoped because `/current_user/*' is declared
+%% ?SCOPE_PUBLIC, not because of a path-parsing exception -- the old
+%% `is_self_service_endpoint/2' whitelist is deleted. A user with an
```
savonarola: Pls do not describe or reference deleted code.

apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl
```diff
@@ -438,66 +450,74 @@ t_check_login_user_scopes_undefined_falls_back(_) ->
...
-%% (modifying one's own record) MUST still be subject to the scope
-%% check — otherwise a user with explicit `scopes = []' could PUT
-%% itself to add scopes back, defeating the self-restriction.
-t_check_login_user_scopes_self_user_record_not_bypassed(_) ->
-    Username = <<"login_user_scopes_self_put">>,
+%% A user literally named `current_user' does not gain the self-service
+%% exemption for its own record: `/users/current_user' is still scope-
+%% checked. Guards against a prefix/segment confusion between the
+%% top-level `/current_user' route and a username of the same text.
+t_check_login_user_scopes_username_named_current_user(_) ->
```
savonarola: This test looks redundant too. Probably no need to test that different routes are independent.

apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl
```diff
@@ -116,37 +117,47 @@ classify(Path, PathMap) ->
...
-%% Entries whose value is ?SCOPE_PUBLIC are skipped: they are kept in
-%% the cache as sentinels (so exact-match lookup can distinguish
-%% "intentionally public" from "genuinely unmapped"), but they must
-%% not claim a sibling concrete path via wildcard segment match.
-%% Without this skip, e.g. `/sso/running' (public) would be claimed
-%% by the sibling template `/sso/:backend' (sso_management).
+%% Scoped templates are tried first and ?SCOPE_PUBLIC ones only if none
+%% matched, so a scoped template always wins over a public one. That
+%% ordering is what the previous unconditional skip of public entries
+%% was protecting (a public template must not claim a path some other
```
savonarola: Pls do not document what have been removed. Describe and explain the current logic.

apps/emqx_management/test/emqx_mgmt_api_key_scopes_SUITE.erl
```diff
@@ -731,6 +739,50 @@ create_app(Name, Extra) ->
         Error -> Error
     end.
 
+-doc """
+A ?SCOPE_PUBLIC entry that is itself a TEMPLATE must classify the
+concrete paths it covers as `public', not `not_found'.
+
+`classify_path/1' answers an exact-match lookup first, so a public
+template is only ever reached through segment matching. Public entries
+used to be skipped there unconditionally, which made a public template
```
savonarola: Pls do not describe how things used to be, document the correct current expectation instead.

changes/ee/breaking-18608.en.md
```diff
@@ -0,0 +1,16 @@
...
+- `POST` and `DELETE /users/:username/mfa` now require the administrator role and manage other users only. Use `/current_user/mfa` for your own account. A namespaced (multi-tenant) administrator can no longer reach another user's MFA, not even within its own namespace.
+- The `mfa_management` scope becomes administrator-only and means "manage other users' MFA". It previously also acted as a self-exemption key that let a non-administrator bypass an MFA lock on their own account. That meaning is gone, and assigning the scope to a non-administrator is now rejected with `400`. An account whose MFA an administrator has reset can be unlocked by another administrator or from the node console.
+
+Dashboard clients and any scripts calling the removed or narrowed endpoints must be updated.
```
savonarola:
```suggestion
Dashboard account APIs are now separate from user administration APIs.

Users manage their own accounts through `/current_user`:

- `GET /current_user` returns the signed-in user's username, role, effective scopes, and MFA status.
- `POST /current_user/change_pwd` changes the signed-in user's password. The request must include `old_pwd` and `new_pwd`.
- `POST /current_user/mfa` sets up or rotates the signed-in user's MFA.
- `DELETE /current_user/mfa` disables the signed-in user's MFA.

Breaking changes:

- `POST /users/:username/change_pwd` is deprecated and restricted to the signed-in user. EMQX returns `403` when the path identifies another user. Use `POST /current_user/change_pwd` instead. Administrators can reset another local user's password only with `emqx ctl admins passwd`.
- `POST` and `DELETE /users/:username/mfa` now manage other users and require the administrator role. Use `/current_user/mfa` to manage your own MFA. Namespaced administrators cannot manage another user's MFA.
- The `mfa_management` scope is now administrator-only and grants permission to manage another user's MFA. It no longer lets non-administrators override an MFA requirement on their own accounts. EMQX returns `400` when this scope is assigned to a non-administrator. Another administrator or the node console must remove an administrator-set MFA requirement.
```

savonarola:
@zmstone Looks good to me too. There are some minor issues, plus lots of comments describing what has changed. Don't think we should track all that transitions in the comments as code unless we need to explain existance of some compatibility shims.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
zmstone: yes, to keep the rules simple.

a currently logged in user (admin or not) has no reason to call the non-self management API to `delete` or `post` self mfa.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
```diff
@@ -832,23 +1004,84 @@ change_pwd(post, #{bindings := #{username := Username}, body := Params}) ->
             end
     end.
 
+current_user_mfa(post, #{body := Settings} = Req) ->
+    Mechanism = maps:get(<<"mechanism">>, Settings),
+    with_caller(Req, fun(#?ADMIN{username = Username}) ->
+        LogMeta = #{msg => "dashboard_user_mfa_setup", username => Username},
+        %% Never `ByAdmin': a self-initiated (re)init must not touch the
+        %% admin_override decision.
+        mfa_result(emqx_dashboard_admin:reinit_mfa(Username, Mechanism, false), LogMeta)
```
zmstone, line 990: for an SSO caller, `Username` here is `{Backend, Name}`. `reinit_mfa/3` passes that value to `emqx_dashboard_token:destroy_by_username/1`, but JWT rows store `username = Name` and `backend = Backend` separately, and `jwt_by_username_spec/1` matches only the `username` field. Consequently `POST /current_user/mfa` returns 204 without deleting any SSO JWTs, so the bearer token used for the re-key remains valid (unlike the local-user path). Please make token invalidation match the resolved SSO key, preserving backend isolation, and add a regression assertion that the same SSO token is rejected after rotation.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: fixed

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: fixed

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: fixed

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: fixed

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: fixed

apps/emqx_dashboard/test/emqx_dashboard_current_user_SUITE.erl
MorganaFuture: fixed

apps/emqx_dashboard/test/emqx_dashboard_user_scopes_SUITE.erl
MorganaFuture: fixed

apps/emqx_dashboard/test/emqx_dashboard_user_scopes_SUITE.erl
MorganaFuture: fixed

apps/emqx_dashboard/test/emqx_dashboard_user_scopes_SUITE.erl
MorganaFuture: fixed

apps/emqx_dashboard/test/emqx_dashboard_user_scopes_SUITE.erl
MorganaFuture: fixed

apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl
MorganaFuture: fixed

apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl
MorganaFuture: fixed

apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl
MorganaFuture: fixed

apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl
MorganaFuture: fixed

apps/emqx_management/test/emqx_mgmt_api_key_scopes_SUITE.erl
MorganaFuture: fixed

apps/emqx_dashboard/test/emqx_dashboard_current_user_SUITE.erl
MorganaFuture: removed

apps/emqx_dashboard/test/emqx_dashboard_current_user_SUITE.erl
MorganaFuture: removed

apps/emqx_dashboard/test/emqx_dashboard_current_user_SUITE.erl
MorganaFuture: removed

apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl
MorganaFuture: removed

changes/ee/breaking-18608.en.md
MorganaFuture: fixed, with one line added for the SSO token fix in this round.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: fixed. `is_caller/2` now lists every key shape explicitly and returns false only for `undefined`; an unrecognised shape raises rather than answering false, which both call sites read as "acting on somebody else".

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: no change here, following @zmstone's answer in this thread.

apps/emqx_dashboard/src/emqx_dashboard_api.erl
MorganaFuture: fixed. The cause is in `emqx_dashboard_token`, untouched by this PR: `format/5` splits the key on write (bare name in `username`, backend in `extra`) while `jwt_by_username_spec/1` matched the whole key, so `{Backend, Name}` matched no row. `reinit_mfa/3`, `remove_user/1`, the role change in `update_user/4` and `change_password_hash/2` all went through it.

Selection now matches on the stored name and keeps the rows whose `full_admin_key/2` equals the requested key, so backends stay isolated — ending a local account's sessions no longer ends those of an SSO account of the same name.

Regression cases: `t_own_mfa_write_ends_sso_session` and `t_own_mfa_write_keeps_other_backend_session` over HTTP, plus `t_lookup_by_username_jwt_sso` and `t_lookup_by_username_jwt_backend_isolation` on the token module. All four fail without the fix.

MorganaFuture:
@savonarola all addressed. The comment narration was not limited to the lines you marked — I swept every file this branch touches and rewrote 35 of them to describe the current state; the only ones left mentioning a prior state describe runtime facts (an account deleted between login and the request), not code changes. Removed the three cases that exercise minirest, and the `?EE_ONLY` guard with them.

savonarola (approved):

zmstone:
@MorganaFuture 
thank you very much for the contribution.
we are a bit slow in review. please bear with us.

PR #18609, opened 2026-08-29
feat(listener): report the resolved bind address in listener views

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:
N/A

## Summary

Follow-up to #18557 (`node.default_listener_address`), which is groundwork for #18511 (`hardened` becomes the default security profile).

#18557 deliberately keeps schema defaults as static bare ports and applies the address in runtime code when a listener starts, so the configuration keeps showing the configured bind, and each node resolves its own address. The consequence: with `bind = 1883` under the `hardened` profile, `GET /api/v5/listeners/:id` reports `"bind": ":1883"` while the socket is on `127.0.0.1:1883`, and nothing reports the difference.

This PR adds two read-only fields to the listener views, using `emqx_listeners:listen_on/2`, which every call site already computes to key an esockd lookup and then discards:

* `resolved_address` — the IP a listener is actually bound to on this node, without the port (the port never differs from `bind`, so repeating it would be redundant).
* `resolved_address_from` — why it has that value: `bind` when the listener's own `bind` already sets an explicit address, `0.0.0.0` for all interfaces, `127.0.0.1` for loopback, `nodename`, or the literal `node.default_listener_address` value when it is a hostname or IP address.

Both are node-local by nature — the security profile and `node.default_listener_address` both resolve independently on each node — so they are reported accordingly:

* `GET /api/v5/listeners/:id` (`emqx_listeners:do_list_raw/0`, `emqx_mgmt_api_listeners:listeners_info/1`) reports them as flat fields for the node handling the request.
* `GET /api/v5/listeners` (the cluster-status list) reports them per node under `node_status`, not in the cluster-wide `status` — a listener with the same id can legitimately resolve to different addresses on different nodes (e.g. under `node.default_listener_address = nodename`), and unlike `running` or the connection counts, there's no meaningful cluster-wide rollup of an address.
* `emqx_listeners:list/0`, used by `emqx ctl listeners`, and the gateway listener API follow the same shape.

`bind` keeps reporting the static configured value unchanged, port included; both new fields are read-only and stripped from an update body before it reaches the config (`parse_listener_conf/2`), so a GET-then-PUT round trip (what the dashboard does on every listener save) keeps succeeding.

The dashboard's own HTTP listener resolves its bind the same way (`emqx_dashboard:ip_port/2`), but there is no existing read API exposing it; this PR does not add one.

## Manual validation

Started a node on non-default ports with the `hardened` profile, `node.default_listener_address = nodename`, and a bare-port bind (`bind = 21883`):

`GET /api/v5/listeners/tcp:default`:
```json
{
  "bind": ":21883",
  "resolved_address": "127.0.0.1",
  "resolved_address_from": "nodename",
  "running": true,
  ...
}
```

`GET /api/v5/listeners` (cluster status list):
```json
{
  "id": "tcp:default",
  "status": { "running": true, "max_connections": "infinity", "current_connections": 0 },
  "node_status": [
    {
      "node": "emqx@...",
      "status": { "running": true, ..., "resolved_address": "127.0.0.1", "resolved_address_from": "nodename" }
    }
  ]
}
```

`emqx ctl listeners`:
```
tcp:default
  listen_on             : :21883
  resolved_address      : 127.0.0.1
  resolved_address_from : nodename
  ...
```

`ss -lntp`:
```
LISTEN  0  1024  127.0.0.1:21883  0.0.0.0:*  users:(("beam.smp",...))
```

GET-then-PUT the same listener body back unchanged returns 200 with both fields unchanged.

## PR Checklist


rel/i18n/emqx_mgmt_api_listeners.hocon
```diff
@@ -104,6 +104,11 @@ listener_running_status {
     label: "Running Status"
 }
 
+listener_resolved_address {
+    desc: "The address the listener is actually bound to on this node. Read-only. Can differ from 'bind' when the address comes from the security profile or 'node.default_listener_address', which are resolved locally on each node. Can be 'inconsistent' when nodes resolve to different addresses."
```
zmstone: expand on it:

- `inconsistent` when nodes in the cluster resolve to different address when default_listener_address is configured to nodename or a FQDN/hostname
- IP Address when all nodes in the cluster resolves to the same address, e.g. 127.0.0.1, 0.0.0.0. If `bind` is configured with IP:Port (not just the port number), then resolved address should be the IP in this ocnfig.

apps/emqx_gateway/src/emqx_gateway_api_listeners.erl
```diff
@@ -307,26 +321,49 @@ ensure_integer_or_infinity(I) when is_integer(I) ->
...
-
-aggregate_running(R, R) -> R;
-aggregate_running(R, undefined) -> R;
-aggregate_running(_, _) -> inconsistent.
+    NRunning = aggregate_agreement(Running, RunningAcc),
+    %% Nodes with byte-identical config can resolve different addresses
+    %% (security profile / node.default_listener_address are node-local),
+    %% which is exactly the disagreement this field exists to surface, so
+    %% mark it `inconsistent` instead of picking one node's value silently.
+    NResolvedAddress = aggregate_agreement(ResolvedAddress, ResolvedAddressAcc),
```
keynslug: Maybe there's no need to report the cluster-wide "aggregate" at all if it's node-specific most of the time?

apps/emqx_management/src/emqx_mgmt_api_listeners.erl
```diff
@@ -668,14 +720,36 @@ format_status(Key, Node, Listener, Acc) ->
...
+            %% Nodes with byte-identical config can resolve different
+            %% addresses (security profile / node.default_listener_address
+            %% are node-local), which is exactly the disagreement this field
+            %% exists to surface, so mark it `inconsistent` instead of
+            %% picking one node's value silently.
+            NResolvedAddress =
+                case ResolvedAddress == ResolvedAddress0 of
+                    true -> ResolvedAddress0;
+                    _ -> inconsistent
+                end,
```
keynslug: Similar question: seems there's no need to report the cluster-wide "aggregate" at all if it's inherently node-specific?

keynslug (approved):

Issue #17896, opened 2026-07-07
[5.8.x] Nodes API emits Internal Error if queried for unreachable/stopping node info


keynslug:
### What happened?

Nodes API crashes and responds with 500 Internal Error in some circumstances.

As evidenced in K8s events, produced by emqx-operator.

```
  9m          Normal    Killing                   pod/emqx-core-778db64d58-1                  Stopping container emqx
  9m          Warning   Unhealthy                 pod/emqx-core-778db64d58-1                  Readiness probe failed: Get "http://10.244.0.22:18083/status": dial tcp 10.244.0.22:18083: connect: connection refused
  9m          Warning   ReconcilerFailed          emqx/emqx                                   failed to get API http://10.244.0.18:18083/api/v5/nodes/emqx@emqx-core-778db64d58-1.emqx-headless.default.svc.cluster.local, status : 500 Internal Server Error, body: {"code":"INTERNAL_ERROR","message":"error, function_clause, [{emqx_mgmt_api_nodes,format,[{error,badarg}],[{file,\"/emqx/apps/emqx_management/src/emqx_mgmt_api_nodes.erl\"},{line,293}]},{emqx_mgmt_api_nodes,'-to_ok_result_fun/1-fun-0-',2,[{file,\"/emqx/apps/emqx_management/src/emqx_mgmt_api_nodes.erl\"},{line,310}]},{emqx_utils_api,with_node,2,[{file,\"/emqx/apps/emqx_utils/src/emqx_utils_api.erl\"},{line,37}]},{minirest_handler,apply_callback,4,[{file,\"/emqx/deps/minirest/src/minirest_handler.erl\"},{line,177}]},{minirest_handler,handle,2,[{file,\"/emqx/deps/minirest/src/minirest_handler.erl\"},{line,71}]},{minirest_handler,init,2,[{file,\"/emqx/deps/minirest/src/minirest_handler.erl\"},{line,32}]},{cowboy_handler,execute,2,[{file,\"/emqx/deps/cowboy/src/cowboy_handler.erl\"},{line,41}]},{cowboy_stream_h,execute,3,[{file,\"/emqx/deps/cowboy/src/cowboy_stream_h.erl\"},{line,318}]},{cowboy_stream_h,request_process,3,[{file,\"/emqx/deps/cowboy/src/cowboy_stream_h.erl\"},{line,302}]},{proc_lib,init_p_do_apply,3,[{file,\"proc_lib.erl\"},{line,241}]}]"}
```

[...]

PR #18619, opened 2026-08-29
fix(mgmt): return 400 instead of 500 for an unreachable node in Nodes API

zmstone:
Fixes: #17896
Introduced in: pre-5.8

## Summary

`GET /nodes/{node}` (and the `/metrics`, `/stats` sub-paths) returned a 500
Internal Server Error instead of a 400 Bad Request when the target node
became unreachable between `emqx_mgmt_api_lib:with_node/2`'s liveness check
(`lists:member(Node, mria:running_nodes())`) and the RPC that fetches its
info. The most common trigger is a concurrent `cluster leave`: the node
still shows up as running in the caller's cached membership view, but the
`erpc` call to it fails.

`emqx_mgmt:lookup_node/1` already surfaces that failure as `{error, Reason}`
(via `emqx_rpc:unwrap_erpc/1`), and `to_ok_result/1` already knows how to
turn `{error, _}` into a proper `400 Bad Request` — the sibling
`node_metrics`/`node_stats` handlers rely on exactly this path today.
`emqx_mgmt_api_nodes:format/1` was the one place that didn't: it only had
clauses for a map, so an `{error, Reason}` tuple raised a `function_clause`.

The fix adds one clause: `format({error, _} = Error) -> Error.` No other
behavior changes; a successful lookup is unaffected.

Reproduced with a real 2-node cluster: `emqx1` executing `cluster leave`
while `emqx2`'s Nodes API was concurrently queried for `emqx1`'s info
produced the exact crash reported in the issue (`{error,badarg}` reaching
`format/1`). With the fix, the same race returns a normal 400 response.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_nodes.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_nodes.erl
@@ -283,7 +283,12 @@ format(Info = #{memory_total := Total, memory_used := Used}) ->
         memory_used := emqx_mgmt_util:kmg(Used)
     };
 format(Info) when is_map(Info) ->
-    Info.
+    Info;
+%% `emqx_mgmt:lookup_node/1` reaches this as `Info` when the target node
+%% becomes unreachable between the caller's liveness check and the RPC,
+%% e.g. a concurrent `cluster leave`.
+format({error, _} = Error) ->
+    Error.
 
 to_ok_result({error, _} = Error) ->
     Error;
--- a/apps/emqx_management/test/emqx_mgmt_api_nodes_SUITE.erl
+++ b/apps/emqx_management/test/emqx_mgmt_api_nodes_SUITE.erl
@@ -73,6 +73,24 @@ t_nodes_api(_) ->
         emqx_mgmt_api_test_util:request_api(get, BadNodePath)
     ).
 
+-doc """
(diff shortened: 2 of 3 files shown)


apps/emqx_management/src/emqx_mgmt_api_nodes.erl
keynslug, line 271: I think there hides somewhat related issue: as `emqx_mgmt:list_nodes/0` does `[... || #{node := Node} = Info <- node_info(Running)]`, nodes that are stopping or leaving "disappear" from the response.

keynslug (approved):

apps/emqx_management/src/emqx_mgmt_api_nodes.erl
zmstone: > I think there hides somewhat related issue: as `emqx_mgmt:list_nodes/0` does `[... || #{node := Node} = Info <- node_info(Running)]`, nodes that are stopping or leaving "disappear" from the response.

depends on who consumes the API.
if it's purely for dashboarding, missing a node is probably harmless.
if it supports cluster management (e.g. scaling), then we should probably avoid silent drops.

zmstone: created #18640

PR #18620, opened 2026-08-29
fix(mqueue): enforce max_len across the whole queue, not per priority lane

### LINKED ISSUE #13409: The message queue size may exceed the maximum limit after setting topic priority
### What happened?

The corresponding code in EMQX is as follows:

```erlang
-spec in(message(), mqueue()) -> {option(message()), mqueue()}.
in(Msg = #message{qos = ?QOS_0}, MQ = #mqueue{store_qos0 = false}) ->
    {_Dropped = Msg, MQ};
in(
    Msg = #message{topic = Topic},
    MQ =
        #mqueue{
            default_p = Dp,
            p_table = PTab,
            q = Q,
            len = Len,
            max_len = MaxLen,
            dropped = Dropped
        } = MQ
) ->
    Priority = get_priority(Topic, PTab, Dp),
    PLen = ?PQUEUE:plen(Priority, Q),
    Msg1 = with_ts(Msg),
    case MaxLen =/= ?MAX_LEN_INFINITY andalso PLen =:= MaxLen of
        true ->
            %% reached max length, drop the oldest message
            {{value, DroppedMsg}, Q1} = ?PQUEUE:out(Priority, Q),
            Q2 = ?PQUEUE:in(Msg1, Priority, Q1),
            {without_ts(DroppedMsg), MQ#mqueue{q = Q2, dropped = Dropped + 1}};
        false ->
            {_DroppedMsg = undefined, MQ#mqueue{len = Len + 1, q = ?PQUEUE:in(Msg1, Priority, Q)}}
    end.
```

`PLen` is the length of a priority queue. The following code uses `PLen` and `MaxLen` to compare, which may cause the actual message queue length to exceed the maximum queue length I set.

For example, if I set priorities for 10 topics and set the maximum message queue length to 1000, the actual maximum message queue length will be 10 * 1000 = 10000.

Below is my test result, showing that the current queue length 2 has exceeded the maximum queue length 1:

image-20240704141634027

And here is http response returned by EMQX:

image-20240704141443943

The total length of all priority queues should not exceed the maximum queue length.

_No response_

_No response_

### EMQX version

EMQX 5.7.1

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

zmstone:
Release version:
6.2.4, 6.3.0, 7.0.0

Introduced in:
pre-5.8

## Summary

`emqx_mqueue.max_len` was enforced per priority lane instead of across the whole
queue: `in/2` compared the incoming message's own lane length (`emqx_pqueue:plen/2`)
against `max_len`. A session with N distinct message priorities could therefore
queue up to N x `max_len` messages, contradicting both the module's own
documentation and `len/1`/`info(len, _)`, which already reported the true total.

Fixes #13409.

- `apps/emqx/src/emqx_mqueue.erl`: `in/2` now compares `emqx_pqueue:len/1` (the
  total) against `max_len`.
- `apps/emqx/src/emqx_pqueue.erl`: added `lowest/1`, mirroring the existing
  `highest/1`, returning the lowest non-empty priority.
- Dropping now targets the oldest message in the lowest non-empty priority lane
  (via `lowest/1`), not the incoming message's own lane. That lane can be empty
  (a message arriving at a brand-new priority) or hold a higher priority than
  another lane, so dropping from it could evict a more important message than
  necessary, or drop the message just enqueued when a lower-priority one was
  already queued elsewhere. A lower-priority message must never survive at the
  expense of a higher-priority one.

Test changes in `emqx_mqueue_SUITE`:
- `t_priority_mqueue` asserted the old per-lane growth (`len` reaching 5 with
  `max_len = 3`); updated to assert the new global cap and that the lowest
  non-empty lane is the one dropped from.
- `t_priority_order`/`t_priority_order2` mixed dequeue-order testing with the old
  per-lane retention (`max_len = 5` with 10 messages per topic, relying on each
  lane independently keeping its own last 5). Set `max_len` to unbounded so they
  test only the round-robin/credit dequeue order, unaffected by this change; the
  expected sequences were regenerated from the (unchanged) scheduling logic.
- Added `t_max_len_bounds_total_not_per_priority` (regression test for #13409:
  asserts total length never exceeds `max_len` across several priorities) and
  `t_drop_from_lowest_priority_lane` (asserts the dropped message comes from the
  lowest non-empty lane, including when the incoming message's own lane was
  empty).
- Added `t_lowest` to `emqx_pqueue_SUITE` for the new function.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_mqueue.erl
+++ b/apps/emqx/src/emqx_mqueue.erl
@@ -308,15 +308,19 @@ in(
         end,
     Msg1 = with_ts(Msg),
     Q1 = emqx_pqueue:in(Msg1, Prio, Class, Q0),
-    case MaxLen =/= ?MAX_LEN_INFINITY andalso emqx_pqueue:plen(Prio, Q1) > MaxLen of
+    case MaxLen =/= ?MAX_LEN_INFINITY andalso emqx_pqueue:len(Q1) > MaxLen of
         false ->
             {_DroppedMsg = undefined, MQ#mqueue{
                 q = Q1,
                 payload_bytes = PayloadBytes + emqx_message:payload_size(Msg1)
             }};
         true ->
-            %% Reached max length: drop the oldest least important message.
-            {{value, DroppedMsg}, Q2} = emqx_pqueue:drop(Prio, Q1),
+            %% Reached max length: drop the oldest message from the lowest-priority
+            %% non-empty lane, so a lower-priority message never survives at the
+            %% expense of a higher-priority one. This may not be the lane the
+            %% incoming message was just added to.
+            DropPrio = emqx_pqueue:lowest(Q1),
+            {{value, DroppedMsg}, Q2} = emqx_pqueue:drop(DropPrio, Q1),
             PayloadBytes1 =
                 PayloadBytes - emqx_message:payload_size(DroppedMsg) +
                     emqx_message:payload_size(Msg1),
--- a/apps/emqx/src/emqx_mqueue.erl
+++ b/apps/emqx/src/emqx_mqueue.erl
@@ -319,8 +319,7 @@ in(
             %% non-empty lane, so a lower-priority message never survives at the
             %% expense of a higher-priority one. This may not be the lane the
             %% incoming message was just added to.
-            DropPrio = emqx_pqueue:lowest(Q1),
-            {{value, DroppedMsg}, Q2} = emqx_pqueue:drop(DropPrio, Q1),
+            {{value, DroppedMsg}, Q2} = emqx_pqueue:drop_lowest(Q1),
             PayloadBytes1 =
                 PayloadBytes - emqx_message:payload_size(DroppedMsg) +
                     emqx_message:payload_size(Msg1),
--- a/apps/emqx/test/emqx_mqueue_SUITE.erl
+++ b/apps/emqx/test/emqx_mqueue_SUITE.erl
@@ -106,6 +106,7 @@ t_infinity_simple_mqueue(_) ->
     {{value, V}, _Qy} = ?Q:out(Qx),
     ?assertEqual(<<1>>, V#message.payload).
 
+-doc "max_len caps the total queue length, not each priority lane independently.".
 t_priority_mqueue(_) ->
     Opts = #{
         max_len => 3,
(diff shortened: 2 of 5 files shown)


changes/ee/fix-18620.en.md
```diff
@@ -0,0 +1,3 @@
+Fixed `mqueue.max_len` being enforced per priority lane instead of across the whole message queue.
+
+A session with prioritised topics could hold up to (number of priorities) x `max_len` queued messages, instead of `max_len`. `mqueue.max_len` now limits the total number of queued messages, so such a session no longer holds more than the configured maximum. Deployments that relied on the old behaviour will see a smaller effective queue.
```
zmstone:
```suggestion
Before this fix, a session with prioritised topics could hold up to (number of priorities) x `max_len` queued messages, instead of `max_len`. `mqueue.max_len` now limits the total number of queued messages, so such a session no longer holds more than the configured maximum. Deployments that relied on the old behaviour will see a smaller effective queue.
```

apps/emqx/test/emqx_mqueue_SUITE.erl
```diff
@@ -124,21 +125,32 @@ t_priority_mqueue(_) ->
...
+    ?assertEqual(<<"t3">>, Msg#message.topic),
+    ?assertEqual(2, ?Q:len(Q6)).
 
 t_priority_mqueue_conservation(_) ->
     true = proper:quickcheck(conservation_prop()).
 
+-doc """
+Dequeue order across priority lanes (round-robin with priority-weighted
+credits). max_len is unbounded here so the global cap from t_priority_mqueue
+does not interact with the scheduling being tested.
```
keynslug: Nit. Unclear why mention separate testcase here, and "interaction with scheduling" sounds obscure. Same with the next testcase.

apps/emqx/test/emqx_mqueue_SUITE.erl
```diff
@@ -289,6 +338,76 @@ t_length_priority_mqueue(_) ->
...
+    ?assertEqual(MaxLen, ?Q:len(Q)),
+    %% t5 has the highest priority: a lower-priority message must never
+    %% survive at the expense of a higher-priority one.
+    ?assert(lists:any(fun(#message{topic = T}) -> T =:= <<"t5">> end, ?Q:to_list(Q))).
+
+-doc """
+The message dropped when the queue is full comes from the lowest-priority
+non-empty lane, even when the incoming message's own lane was empty before
+this enqueue (so a naive "drop from the incoming lane" rule would have
+dropped the message just enqueued, or refused it, instead).
```
keynslug: Nit. These details look irrelevant to the testcase itself. Perhaps this should be part of the commit message.

apps/emqx/src/emqx_mqueue.erl
```diff
@@ -308,15 +308,19 @@ in(
         end,
     Msg1 = with_ts(Msg),
     Q1 = emqx_pqueue:in(Msg1, Prio, Class, Q0),
-    case MaxLen =/= ?MAX_LEN_INFINITY andalso emqx_pqueue:plen(Prio, Q1) > MaxLen of
+    case MaxLen =/= ?MAX_LEN_INFINITY andalso emqx_pqueue:len(Q1) > MaxLen of
```
keynslug, line 311: Perhaps optimize `emqx_pqueue:len/1` for _pqueues_ case, so that it doesn't allocate one more list?

apps/emqx/src/emqx_mqueue.erl
```diff
@@ -308,15 +308,19 @@ in(
...
                 payload_bytes = PayloadBytes + emqx_message:payload_size(Msg1)
             }};
         true ->
-            %% Reached max length: drop the oldest least important message.
-            {{value, DroppedMsg}, Q2} = emqx_pqueue:drop(Prio, Q1),
+            %% Reached max length: drop the oldest message from the lowest-priority
+            %% non-empty lane, so a lower-priority message never survives at the
+            %% expense of a higher-priority one. This may not be the lane the
+            %% incoming message was just added to.
+            DropPrio = emqx_pqueue:lowest(Q1),
```
keynslug: Q: Perhaps `drop_lowest/1`? Otherwise, `lowest/1` + `drop/2` do some of the work twice?

keynslug (commented):
Q: Need to mention that this change might incur small performance impact? Or better to confirm first?

keynslug (approved):

Issue #18622, opened 2026-08-30
Add 'emqx ctl actions' command to inspect action status on the local node (JSON output)


zmstone:
### What would you like to be added or enhanced?

An `emqx ctl` command to inspect data integration **action** status **on the local node**, with JSON output.

Proposed interface:

```
emqx ctl actions show   [--name <type:name>] [--ns <namespace>]
emqx ctl actions status [--name <type:name>] [--ns <namespace>]
```

- **`show`** prints the same information as `GET /api/v5/actions/{id}`, but for the local node only — no cluster aggregation, no RPC to other nodes.
- **`status`** prints a compact JSON array of `{"<name>": "<status>"}` objects.
- `--name` selects one action. When omitted, all actions are printed.
- `--ns` selects the namespace. When omitted, the global namespace is used.

Sketch:

```console
$ emqx ctl actions status
[{"kafka_producer:my_action":"connected"},{"http:my_webhook":"connecting"}]

$ emqx ctl actions status --name kafka_producer:my_action
[{"kafka_producer:my_action":"connected"}]

$ emqx ctl actions show --name kafka_producer:my_action
{"name":"my_action","type":"kafka_producer","status":"connected","error":"", ...}
```

Both commands print JSON so they can be consumed by `jq` in a health check or probe script.

The same machinery backs sources and connectors, so `emqx ctl sources ...` and
`emqx ctl connectors ...` would be natural follow-ups, but actions are what matter for the
use case below.

### Why is this needed?

**1. There is no CLI for this today.** Action, source and connector status is REST-only.
Neither `emqx_bridge*`, `emqx_connector*` nor `emqx_resource*` registers an `emqx_ctl`
command (checked on `release-62` and `dev-63`).

**2. The REST top-level `status` field is cluster-aggregated and cannot answer "is this node
ready?"** `emqx_bridge_v2_api:aggregate_status/1` returns the common status only when every
node agrees, and `inconsistent` otherwise — and it returns the same value no matter which node
serves the request. Gating a per-pod readiness probe on that field would remove *every* pod
from the load balancer as soon as a *single* pod became unhealthy. Callers must instead read
`node_status[]` and filter it by the local node's name, which is awkward inside a probe.

**3. The motivating use case: Kubernetes readiness probes.** A node that restarts accepts MQTT
connections before its connectors have finished connecting to the external system. A message
routed by a rule to a not-yet-installed action fails with an unrecoverable error
[...]

PR #18624, opened 2026-08-30
feat(bridge): add 'emqx ctl actions show/status' commands

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:

## Summary

Adds `emqx ctl actions show` and `emqx ctl actions status`, closing
[#18622](https://github.com/emqx/emqx/issues/18622). Both report action status for the
**local node only** — no RPC, no cluster aggregation — so a Kubernetes readiness probe can
check whether the local node's own actions are usable without REST credentials or an HTTP
call. This is the gap the issue describes: the REST `status` field is cluster-aggregated
(`emqx_bridge_v2_api:aggregate_status/1`) and answers "do all nodes agree", not "is this
node ready", which is the wrong question for a per-pod probe.

New module `apps/emqx_bridge/src/emqx_bridge_v2_cli.erl`, loaded from `emqx_bridge_app`
(mirrors `emqx_bridge_v2:load/0`/`unload/0`). No RPC, no new bpapi module — everything it
calls (`emqx_bridge_v2:lookup/4`, `emqx_bridge_v2:list/2`) is already local-only.

- `status` prints a compact JSON array of `{"<type>:<name>": "<status>"}` objects.
- `show` prints the same information as `GET /api/v5/actions/{id}`, minus the cluster-only
  fields (`node_status`, `rules`) that require aggregation, for the local node only.
- `--name <type:name>` selects one action; omitted, every action in the namespace is shown.
- `--ns <namespace>` selects a namespace; omitted, the global namespace is used.

**Secrets.** `lookup/4`'s `raw_config` can carry a secret set directly on the action (e.g.
an HTTP action's `parameters.headers` can hold a bearer token, independent of the
connector). `show` redacts it with `emqx_utils:redact/1`, the same function
`emqx_bridge_v2_api:format_resource/3` uses on the REST path. `resource_data` (the
connector's live resource state — pids, refs) is left out of `show` entirely, same as
`format_resource/3` does; it isn't JSON-encodable and a probe has no use for it beyond
`status`/`error`, which `lookup/4` already surfaces.

**Option parsing.** `--name`/`--ns` are flags, which most `emqx_ctl` commands don't use.
No dedicated flag-parsing helper exists in `emqx_ctl` or `emqx_mgmt_cli` (checked); this
follows the house pattern used by the few commands that do take flags — e.g.
`emqx_mgmt_cli:collect_session_top_args/2` — a small recursive accumulator.

**Exit status.** `emqx_ctl` doesn't use the process exit code for business-logic signaling
anywhere in this codebase today — `emqx_ctl:usage/1` and every command handler return `ok`
regardless of outcome, and `nodetool`'s `rpc_infinity` doesn't propagate the callee's return
value either. This command follows that convention: it always exits 0 on a successful
lookup. A `--name` that matches nothing is not an error — it's valid, quiet input for a
probe — so it prints `[]` (`status`) or `null` (`show`) instead. Both values are falsy to
`jq -e`, so `jq -e` alone distinguishes "ready" from "not ready or not found" without the
command needing a second signaling channel. Verified below.

**Scope.** Actions only, per the issue. `sources` and `connectors` share the same
`emqx_bridge_v2` primitives and are natural follow-ups, but are not in this PR. The module
takes a `ConfRootKey` internally the same way `emqx_bridge_v2` itself does, so adding
`emqx ctl sources ...` later is a new top-level clause, not a rewrite.

### Manual verification

Ran on a throwaway single node (`6.3.0-beta.3-gbae79a4c`) with two `http` actions: one on a
connector pointing at a reachable local HTTP server, one at a closed port. The healthy
action's connector has a secret (`Authorization: Bearer super-secret-token-xyz`) configured
directly on the action's `parameters.headers`, to prove redaction.

```console
$ emqx ctl actions status
[
  {
    "http:action_up" : "connected"
  },
  {
    "http:action_down" : "disconnected"
  }
]

$ emqx ctl actions show --name http:action_up
{
  "type" : "http",
  "status" : "connected",
  "parameters" : {
    "path" : "/",
    "method" : "post",
    "headers" : {
      "Authorization" : "******"
    },
    "body" : "${payload}"
  },
  "node" : "emqx1@127.0.0.1",
  "namespace" : null,
  "name" : "action_up",
  "last_modified_at" : 1788083482439,
  "error" : "",
  "enable" : true,
  "created_at" : 1788083482439,
  "connector" : "conn_up"
}

$ emqx ctl actions status --name http:no_such_action
[

]

$ emqx ctl actions show --name http:no_such_action
null
```

The `jq -e` one-liner from the issue, working against this node:

```console
$ emqx ctl actions status --name http:action_up | jq -e '.[0] | to_entries[0].value == "connected"'
true
$ echo $?
0

$ emqx ctl actions status --name http:action_down | jq -e '.[0] | to_entries[0].value == "connected"'
false
$ echo $?
1

$ emqx ctl actions status --name http:no_such_action | jq -e '.[0] | to_entries[0].value == "connected"'
jq: error (at <stdin>:3): null (null) has no keys
$ echo $?
5
```

A readiness probe using `jq -e` gets exit 0 only when the named action is actually
connected; a missing action or a down connector both fail the check, distinctly from a
connected one, without the CLI command itself needing to change its own exit code.

## PR Checklist

### DIFF
--- a/apps/emqx_bridge/src/emqx_bridge_v2_cli.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_v2_cli.erl
@@ -66,11 +66,18 @@ run(Cmd, Args) ->
     case collect_opts(Args, #{}) of
         {ok, Opts} ->
             Result = query(Cmd, Opts),
-            emqx_ctl:print("~ts~n", [emqx_utils_json:encode(Result, [pretty, force_utf8])]);
+            emqx_ctl:print("~ts~n", [format_json(Result)]);
         {error, Reason} ->
             emqx_ctl:warning("~ts~n", [Reason]),
             usage()
     end.
+
+%% Pretty-printing an empty array renders as three lines (`[`, a blank line, `]`); an
+%% empty result carries no structure worth spreading out, so print it compactly.
+format_json([]) ->
+    <<"[]">>;
+format_json(Result) ->
+    emqx_utils_json:encode(Result, [pretty, force_utf8]).
 
 %%--------------------------------------------------------------------------------
 %% Option parsing
--- a/apps/emqx_bridge/src/emqx_bridge_v2_cli.erl
+++ b/apps/emqx_bridge/src/emqx_bridge_v2_cli.erl
@@ -1,0 +1,173 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2020-2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+
+%% `emqx ctl actions ...`: inspect action status on the local node, in JSON.
+%%
+%% Unlike `GET /api/v5/actions/{id}`, this never RPCs to other nodes and never
+%% aggregates status across the cluster (see `emqx_bridge_v2_api:aggregate_status/1`).
+%% It answers "is *this* node's action usable", which is what a per-pod readiness
+%% probe needs and the cluster-aggregated REST `status` field cannot provide.
+%%
+%% Scope: actions only. Sources and connectors share the same `emqx_bridge_v2`
+%% primitives (see `?ROOT_KEY_SOURCES`), so adding `emqx ctl sources ...` later is
+%% a matter of adding another top-level clause with its own root key, not a rewrite.
+-module(emqx_bridge_v2_cli).
+
+-include_lib("emqx/include/emqx_config.hrl").
+
+-export([load/0, unload/0]).
+-export([actions/1, actions_audit_args/1]).
+
+-define(ROOT_KEY_ACTIONS, actions).
+
+%%--------------------------------------------------------------------------------
+%% Loading and unloading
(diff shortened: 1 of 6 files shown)


apps/emqx_bridge/src/emqx_bridge_v2_cli.erl
```diff
@@ -0,0 +1,182 @@
...
+actions(["show" | Args]) ->
+    run(show, Args);
+actions(["status" | Args]) ->
+    run(status, Args);
+actions(_) ->
+    usage().
+
+usage() ->
+    emqx_ctl:usage([
+        {"actions show [--name <type:name>] [--ns <namespace>]",
```
savonarola, line 52: Nit: `~S"""..."""`

savonarola (approved):

PR #18628, opened 2026-08-30
feat(backup): reject legacy-profile backup import into hardened node

zmstone:
Release version:
6.3.0, 7.0.0

Introduced in:

## Summary

A cluster running the `legacy` security profile takes a data backup (export). Restoring that backup onto a node running the `hardened` profile can carry over data and configuration that behaves differently, because `hardened` applies several stricter defaults (see [emqx-docs#3768](https://github.com/emqx/emqx-docs/pull/3768) for the full picture):

- Bare-port MQTT, gateway, and Dashboard HTTP listener binds resolve to loopback instead of all interfaces (`apps/emqx/src/emqx_default_address.erl`, #18557).
- An authenticator chain left empty or disabled starts denying every client instead of allowing them.
- A restored Dashboard account still on the default password can no longer log in.
- Authentication and authorization backend failures that `legacy` ignored now deny the operation.

Nothing in the backup file changes across these, so there is no config diff to review; the outage or lockout is discovered only after the restore. Related but independent from #18557 and #18609, which surfaced this gap in the same discussion thread but affect a different subsystem (listener bind resolution, not backups).

This PR:
1. Records the exporting node's security profile (`legacy` | `hardened`) in the backup's `META.hocon`.
2. On import, refuses a backup recorded (or missing the field, which predates this change and can't be assumed safe) as anything other than `hardened` into a node currently running `hardened`, unless the caller opts in.
3. Never blocks import into a `legacy` node, regardless of the backup's recorded profile — `legacy` is the permissive profile.

The override is a CLI flag (`--allow-security-profile-mismatch`) or an API body field (`allow_security_profile_mismatch`), threaded through the existing `Opts :: map()` RPC argument — no BPAPI version bump needed.

The check lives in `do_import/2`, not `validate_backup/1`: that function is also called from `do_upload/3` at upload/staging time, before any override is in scope, so staging a `legacy` backup on a `hardened` node must never be blocked.

This PR only gates the import; it does not add per-risk detection (e.g. inspecting whether the restored config actually has an empty authenticator chain). The error message and API field description enumerate the risks so an operator can review before overriding.

## Manual verification

Built a release, exported while running under the default `legacy` profile, restarted with `EMQX_SECURITY_PROFILE=hardened`, and attempted import:

```
=== import WITHOUT override ===
Importing data from "data/backup/emqx-export-2026-08-31-08-47-49.941.tar.gz"...
[error] Data import failed, reason: "backup was exported under the \"legacy\" security profile, but this node runs the \"hardened\" profile, whose stricter defaults can behave differently for restored data: bare-port listener binds resolve to loopback, an authenticator chain left empty or disabled denies every client, a restored dashboard account on the default password can no longer log in, and authentication/authorization backend failures now deny the operation. Override with allow_security_profile_mismatch to import anyway.\n".
```

```
=== import WITH override ===
Importing data from "data/backup/emqx-export-2026-08-31-08-47-49.941.tar.gz"...
Importing cluster configuration for namespace global...
Importing built-in database...
...
Data has been imported successfully.
```

## PR Checklist


apps/emqx_management/src/emqx_mgmt_data_backup.erl
```diff
@@ -584,6 +585,19 @@ format_error({unsupported_version, ImportVersion}) ->
             [str(ImportVersion), str(emqx_release:version())]
         )
     );
+format_error({security_profile_mismatch, BackupProfile}) ->
+    str(
+        io_lib:format(
+            "backup was exported under the \"~s\" security profile, but this node runs the "
```
keynslug, line 591: Nit. Triple-quoted string?

changes/ee/feat-18628.en.md
```diff
@@ -0,0 +1,10 @@
+Data backup exports now record the exporting node's security profile.
+
+Restoring a backup that was exported under the `legacy` security profile onto a node running the `hardened` security profile can carry over data and configuration that behaves differently once restored:
+
+- Bare-port MQTT, gateway, and Dashboard HTTP listener binds resolve to loopback instead of all interfaces.
+- An authenticator chain left empty or disabled starts denying every client instead of allowing them.
+- A restored Dashboard account still on the default password can no longer log in.
+- Authentication and authorization backend failures that were previously ignored now deny the operation.
+
+Importing such a backup now needs the `--allow-security-profile-mismatch` CLI flag, or the `allow_security_profile_mismatch` API parameter, so an operator gets a chance to review these differences instead of discovering them after clients stop connecting or logins start failing. A backup that predates this change is treated the same as one exported under `legacy`. Restoring into a node running `legacy` is never affected.
```
keynslug, line 10: Not a strong opinion: from security standpoint, a hardened profile should allow either importing hardened backups, or importing a pre-6.3 backup and hardening it, and do not provide such flags as they are too easy to use without thinking.

keynslug (approved):

changes/ee/feat-18628.en.md
zmstone: this is mostly to raise awareness of the compatibility implication.

a better solution is actually to check compatibility,
but that'll require parsing the configs before importing.

e.g. 

- for port number listener, if node.default_listener_address is explicitly configured in local node, it should be allowed

- auth chain being NOT empty in the configs being imported should not raise any concern

- A restored Dashboard account still on the default password can no longer log in -- there is no way to check the password hash, but this is not a big issue, user just change password after login

- Authentication and authorization backend failures, and similar error conditions that `legacy` ignored, now deny the operation
  - check config `authentication_settings.ignore_backend_failures = true`
  - check config `authorization.ignore_backend_failures = true`

keynslug: > but that'll require parsing the configs before importing.

Yeah this is roughly what I was thinking about "hardening". On the other hand, introducing a lot of new logic for moving backups between profiles is a recipe for introducing security issues, so I think this should ideally be restricted to pre-6.3 backups only. Just a thought.

PR #18630, opened 2026-08-30
fix(mgmt): enforce namespaced-role scope allowlist for API keys

zmstone:
Fixes: https://github.com/emqx/emqx-dev-team-tasks/issues/412
Introduced in: 6.0.3

## Summary

A namespaced administrator (`ns:<ns>::administrator`) could mint, update, or bootstrap an API key carrying scopes outside `?NS_ADMIN_ALLOWED_SCOPES` — notably `gateways` and, worse, `audit`, which reaches the **global** audit log. The dashboard login-user path already rejects this via `emqx_dashboard_api:validate_role_scope_compat/2`. https://github.com/emqx/emqx/pull/18222 closed only the `publish` leg of this gap on the API-key side; `gateways`, `audit`, and any other out-of-allowlist scope remained mintable.

This PR enforces the same `?NS_ADMIN_ALLOWED_SCOPES` allowlist across all three API-key scope-granting paths in `apps/emqx_management/src/emqx_mgmt_api_api_keys.erl` and `apps/emqx_management/src/emqx_mgmt_auth.erl`:

- **Create** (`validate_explicit_scopes/2` → new `validate_namespaced_scope_allowlist/2`): rejects any scope outside the allowlist for a namespaced role with a 400 naming the disallowed scopes. This subsumes the old publish-only check.
- **Update**: the `{set, Scopes}` leg routes through the same validation whenever the requested list genuinely differs from the persisted one. A verbatim round-trip of an already-persisted list is still accepted without re-validation (see design note below).
- **Bootstrap**: `parse_bootstrap_scopes_lenient/2` is now `/3` — the namespace is threaded through from both bootstrap-tail parsers so a namespaced line can drop disallowed scopes with a new `namespaced_scope_not_allowed` reason, consistent with the existing lenient drop-and-warn style (`unknown_scope`, `not_allowed_for_publisher_role`, `privilege_scope_conflict`).

Also corrected a stale comment in `emqx_utils/include/emqx_api_key_scopes.hrl` claiming the scope layer is only "defense-in-depth" behind RBAC. That is not true for GET endpoints: RBAC's catch-all grants a namespaced administrator blanket read access, so there the scope list is the sole gate.

### Design decision: grandfathered keys (please confirm)

This is create/update/bootstrap-time validation only. It stops new over-grants but does **not** retroactively strip scopes from API keys already minted through the gap — an existing namespaced key still holding `audit`/`gateways` keeps working until rotated, matching the "existing records untouched" precedent from #18222. The changelog tells operators to rotate such keys.

A more thorough alternative would additionally intersect a namespaced principal's effective scopes with `?NS_ADMIN_ALLOWED_SCOPES` at the single runtime choke point (`emqx_mgmt_auth:check_rbac_and_scopes/6`, which already has the key's `Namespace` in scope), neutering already-minted over-granted keys immediately instead of only on rotation. That changes runtime behavior for existing keys, so it leans `breaking`. The linked issue's fix direction did not ask for it, so I left it out of this PR — flagging it here as a follow-up candidate if the grandfathering window is a concern.

### Out of scope (tracked separately)

`GET /api/v5/audit` itself returns the global log with no namespace filter (`emqx_audit_api.erl`). Namespace-scoping that endpoint is a separate defense-in-depth follow-up, not implemented here.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_auth.erl
+++ b/apps/emqx_management/src/emqx_mgmt_auth.erl
@@ -65,7 +65,7 @@
 -export([trans/2, force_create_app/1]).
 -export([init_bootstrap_file/1]).
 %% Exported for emqx_mgmt_auth_tests.
--export([parse_bootstrap_scopes_lenient/2, group_rejected_by_reason/1]).
+-export([parse_bootstrap_scopes_lenient/3, group_rejected_by_reason/1]).
 -endif.
 
 -define(APP, emqx_app).
@@ -787,7 +787,7 @@ parse_simple_tail(ApiKey, ApiSecret, Tail) ->
             end);
         [Role, ScopesStr] ->
             with_valid_role(Role, fun(R) ->
-                {Scopes, Rejected} = parse_bootstrap_scopes_lenient(R, ScopesStr),
+                {Scopes, Rejected} = parse_bootstrap_scopes_lenient(R, ?global_ns, ScopesStr),
                 bootstrap_entry(ApiKey, ApiSecret, R, ?global_ns, Scopes, Rejected)
             end);
         _ ->
@@ -830,7 +830,7 @@ parse_role_and_scopes(ApiKey, ApiSecret, Namespace, RoleAndScopes) ->
             end);
         [Role, ScopesStr] ->
             with_valid_role(Role, fun(R) ->
-                {Scopes, Rejected} = parse_bootstrap_scopes_lenient(R, ScopesStr),
+                {Scopes, Rejected} = parse_bootstrap_scopes_lenient(R, Namespace, ScopesStr),
                 bootstrap_entry(ApiKey, ApiSecret, R, Namespace, Scopes, Rejected)
             end);
         _ ->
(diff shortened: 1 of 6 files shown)


apps/emqx_management/src/emqx_mgmt_auth.erl
```diff
@@ -896,6 +900,18 @@ drop_mixed_privilege_scopes(Valid, Rejected) ->
...
+%% A namespaced bootstrap key (`Namespace' is a binary) may only hold
+%% scopes in `?NS_ADMIN_ALLOWED_SCOPES'; a global line (`Namespace' is
+%% `?global_ns') is unaffected. Returns updated `{Valid, Rejected}'.
+drop_disallowed_namespaced_scopes(Namespace, Valid, Rejected) when is_binary(Namespace) ->
+    {Keep, Drop} = lists:partition(
+        fun(S) -> lists:member(S, ?NS_ADMIN_ALLOWED_SCOPES) end,
+        Valid
+    ),
+    {Keep, Rejected ++ [{S, namespaced_scope_not_allowed} || S <- Drop]};
+drop_disallowed_namespaced_scopes(_Namespace, Valid, Rejected) ->
```
savonarola:
```suggestion
drop_disallowed_namespaced_scopes(?global_ns, Valid, Rejected) ->
```

savonarola (dismissed):

apps/emqx_management/src/emqx_mgmt_auth.erl
savonarola: We better not create a permissive fallback

savonarola (approved):

PR #18631, opened 2026-08-30
test(client): retry throttled reconnects in t_sock_closed_reason_normal

zmstone:
## Summary

De-flake `emqx_client_SUITE:t_sock_closed_reason_normal`:

```
%%% emqx_client_SUITE ==> others.t_sock_closed_reason_normal: FAILED
{'EXIT',{shutdown,server_unavailable}}
```

Seen in: https://github.com/emqx/emqx/actions/runs/33312708342/job/99261541000?pr=18629

The case connects and disconnects the *same* clientid once per protocol version (`[v3, v4, v5]`). Since 743e2e510b, `emqx_cm:open_session/4` refuses a clientid while a previous local channel is still being cleaned up, so the next connect in the loop can be rejected. The failing run logs `clientid_registration_throttled` with `local_alive: 0, local_down: 1` immediately before the failure, which is that refusal.

Retry the connect on `server_busy`/`server_unavailable`, the same treatment `emqx_message_transformation_http_api_SUITE` got in d58b1448e9 and `emqx_prometheus_SUITE` in #18179.

Full suite passes locally: 22 ok, 0 failed.

Base `dev-58`: the case is present unchanged from dev-58 through dev-70.

### DIFF
--- a/apps/emqx/test/emqx_client_SUITE.erl
+++ b/apps/emqx/test/emqx_client_SUITE.erl
@@ -462,8 +462,7 @@ t_sock_closed_reason_normal(_) ->
     [
         ?check_trace(
             begin
-                {ok, C} = emqtt:start_link([{proto_ver, Ver}, {clientid, ClientId}]),
-                {ok, _} = emqtt:connect(C),
+                C = connect_retry_busy([{proto_ver, Ver}, {clientid, ClientId}]),
                 ?wait_async_action(
                     emqtt:disconnect(C),
                     #{?snk_kind := sock_closed_normal},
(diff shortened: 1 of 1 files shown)


apps/emqx/test/emqx_client_SUITE.erl
```diff
@@ -462,8 +462,7 @@ t_sock_closed_reason_normal(_) ->
     [
         ?check_trace(
             begin
-                {ok, C} = emqtt:start_link([{proto_ver, Ver}, {clientid, ClientId}]),
-                {ok, _} = emqtt:connect(C),
+                C = connect_retry_busy([{proto_ver, Ver}, {clientid, ClientId}]),
```
keynslug: Maybe simpler option is to allocate different / semi-random ClientIDs for each of `ProtoVers`?

keynslug (approved):

hjianbo (approved):

Issue #18638, opened 2026-08-30
MQTT-SN gateway: socket close is silently ignored while the channel is asleep


zmstone:
### What happened?

`emqx_mqttsn_channel:handle_info({sock_closed, Reason}, Channel)` has clauses for
`conn_state` in `idle`, `connecting`, `connected`, `awake` and `disconnected`, but **not for
`asleep`**.

`apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl`:

```erlang
handle_info({sock_closed, Reason}, Channel = #channel{conn_state = idle})       -> shutdown(Reason, Channel);
handle_info({sock_closed, Reason}, Channel = #channel{conn_state = connecting}) -> shutdown(Reason, Channel);
handle_info({sock_closed, Reason}, Channel = #channel{conn_state = connected, ...}) -> handle_active_sock_closed(...);
handle_info({sock_closed, Reason}, Channel = #channel{conn_state = awake})      -> handle_active_sock_closed(...);
handle_info({sock_closed, Reason}, Channel = #channel{conn_state = disconnected}) ->
    ?SLOG(error, #{msg => "unexpected_sock_closed", reason => Reason}),
    {ok, Channel};
```

A close delivered while the channel is `asleep` matches none of them and falls through to the
catch-all at the end of the function:

```erlang
handle_info(Info, Channel) ->
    ?SLOG(error, #{msg => "unexpected_info", info => Info}),
    {ok, Channel}.
```

So the connection process logs `unexpected_info` and **stays alive**. The socket-close side effects
that every other state performs — `ensure_disconnected/2`, will-message publication via
`mabye_publish_will_msg/1`, and the `maybe_shutdown/2` decision in `handle_active_sock_closed/2` —
are all skipped.

`asleep` is a first-class state in this module: `conn_state()` includes it (`:84`), it is entered from
the DISCONNECT-with-duration path (`:1014`, `asleep(Duration, Channel)`), and several other handlers
match on it explicitly (`:601`, `:989`, `:1006`, `:1504`).

### What did you expect to happen?

A socket close while asleep should be handled like the other active states — publish the will message
if one is configured, mark the session disconnected, and let `maybe_shutdown/2` decide whether the
process terminates based on the session expiry interval.

Whether `asleep` should behave exactly like `awake`/`connected` or needs its own semantics is the
design question. `asleep` exists so a sleeping client's session survives without a live socket, so
[...]

PR #18652, opened 2026-08-31
fix(mqttsn): handle asleep channel lifecycle

hjianbo:
Fixes: #18638
Release version: 5.8.12, 5.9.4, 5.10.5
Introduced in: pre-5.8

## Summary

Fix MQTT-SN channel lifecycle handling around the `asleep` state.

The channel now distinguishes a client-initiated normal `DISCONNECT` from transport and UDP proxy close notifications. Normal disconnects from `connected`, `asleep`, or `awake` move the channel to `disconnected` without publishing the Will. Transport or proxy closes during `asleep` preserve the sleeping session, while connection loss during `connected` or `awake` performs lost-client cleanup and publishes the Will. Sleep timeout also performs lost-client cleanup and publishes the Will before applying session expiry. Reconnecting from `asleep` cancels stale sleep and keepalive timers so keepalive supervision restarts.

This keeps the behavior compatible with both the ownerless UDP proxy callbacks used by this release line and the owner-aware callbacks used by newer branches.

## PR Checklist

### DIFF
--- a/apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
+++ b/apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
@@ -118,6 +118,7 @@
 -define(REGISTER_INFLIGHT(TopicId, TopicName), #channel{register_inflight = {TopicId, _, TopicName}}).
 
 -define(MAX_RETRY_TIMES, 3).
+-define(CLIENT_DISCONNECT, client_disconnect).
 
 % 5s
 -define(REGISTER_TIMEOUT, 5000).
@@ -596,10 +597,13 @@ handle_in(
         previous_state => ConnState,
         clientid => ClientId
     }),
+    %% A keepalive timeout may fire while asleep without removing its timer entry.
+    %% Clear it so CONNACK can arm fresh keepalive supervision.
+    Channel1 = cancel_timer(keepalive, cancel_timer(expire_asleep, Channel)),
     handle_out(
         connack,
         ?SN_RC_ACCEPTED,
-        Channel#channel{conn_state = connected}
+        Channel1#channel{conn_state = connected}
     );
 %% new connection
 handle_in(
@@ -1594,6 +1598,11 @@ handle_out(
             }),
             {ok, Channel#channel{register_awaiting_queue = NRAQueue}}
     end;
+handle_out(disconnect, normal, Channel = #channel{conn_state = ConnState}) when
+    ConnState =:= connected; ConnState =:= asleep; ConnState =:= awake
+->
+    DisPkt = ?SN_DISCONNECT_MSG(undefined),
+    {ok, [{outgoing, DisPkt}, {close, ?CLIENT_DISCONNECT}], Channel};
 handle_out(disconnect, RC, Channel) ->
     DisPkt = ?SN_DISCONNECT_MSG(undefined),
     Reason =
@@ -1845,6 +1854,15 @@ handle_info(
 ) ->
     shutdown(Reason, Channel);
 handle_info(
+    {sock_closed, ?CLIENT_DISCONNECT},
+    Channel = #channel{conn_state = ConnState}
+) when
+    ConnState =:= connected;
+    ConnState =:= asleep;
+    ConnState =:= awake
+->
+    handle_client_disconnect(Channel);
+handle_info(
     {sock_closed, Reason},
     Channel = #channel{
         conn_state = connected,
@@ -1855,19 +1873,30 @@ handle_info(
     %% How to get the flapping detect policy ???
     %emqx_zone:enable_flapping_detect(Zone)
     %    andalso emqx_flapping:detect(ClientInfo),
-    NChannel = ensure_disconnected(Reason, mabye_publish_will_msg(Channel)),
-    case maybe_shutdown(Reason, NChannel) of
-        {ok, NChannel1} -> {ok, {event, disconnected}, NChannel1};
-        Shutdown -> Shutdown
-    end;
+    handle_connection_lost(Reason, Channel);
 handle_info(
     {sock_closed, Reason},
-    Channel = #channel{conn_state = disconnected}
+    Channel = #channel{conn_state = awake}
 ) ->
-    ?SLOG(error, #{
-        msg => "unexpected_sock_closed",
+    handle_connection_lost(Reason, Channel);
+handle_info(
+    {sock_closed, Reason},
+    Channel = #channel{
+        conn_state = asleep,
+        clientinfo = #{clientid := ClientId}
+    }
+) ->
+    %% A sleeping MQTT-SN client has no live transport to lose.
+    %% Preserve its session until it wakes up or the sleep timer expires.
+    ?tp(debug, mqttsn_asleep_sock_closed_ignored, #{
+        clientid => ClientId,
         reason => Reason
     }),
+    {ok, Channel};
+handle_info(
+    {sock_closed, _Reason},
+    Channel = #channel{conn_state = disconnected}
+) ->
     {ok, Channel};
 handle_info(clean_authz_cache, Channel) ->
     ok = emqx_authz_cache:empty_authz_cache(),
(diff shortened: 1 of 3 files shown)


changes/ee/fix-18652.en.md
```diff
@@ -0,0 +1 @@
+MQTT-SN now sends configured Will messages when a sleeping client remains inactive beyond its sleep duration and correctly completes normal disconnects from sleep mode.
```
zmstone: The changelog covers the sleep-mode paths but omits the change that will reach the most users: a normal DISCONNECT no longer publishes the Will in **any** state, including plain `connected`.

Before this PR, `handle_in(?SN_DISCONNECT_MSG(undefined), Channel)` fell through to the generic `handle_out(disconnect, RC, Channel)`, which emitted `{close, normal}`. `emqx_gateway_conn` turns that into `{sock_closed, normal}`, which matched the `connected` clause and called `mabye_publish_will_msg/1`. So a clean client disconnect published the Will.

After this PR the same path emits `{close, ?CLIENT_DISCONNECT}` and lands in `handle_client_disconnect/1`, which does not publish.

Not publishing is correct — MQTT reserves the Will for abnormal termination — and `t_asleep_test07_to_connected` asserts it with `?assertNotReceive`, so the intent is clear. But it is a visible behavior change for anyone whose clients disconnect cleanly and who has come to rely on the Will firing. Worth one more sentence here saying so.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -1844,6 +1851,27 @@ handle_info(
...
+    handle_client_disconnect(Channel);
+handle_info(
+    {sock_closed, ?CLIENT_DISCONNECT},
+    Channel = #channel{
+        conn_state = asleep
+    }
+) ->
+    handle_client_disconnect(Channel);
+handle_info(
+    {sock_closed, ?CLIENT_DISCONNECT},
```
zmstone: This adds `awake` handling for a client disconnect, but `awake` still has no clause for any other `sock_closed` reason.

After this PR the reasons reaching `handle_info/2` are `client_disconnect` (handled for connected, asleep and awake) and everything else (handled for connected, ignored for asleep and disconnected). A transport error, or `udp_proxy_closed` — which `emqx_gateway_conn` delivers as `{sock_closed, normal}` — arriving while `awake` matches no clause. It falls through to the catch-all at the end of `handle_info/2`, logs `unexpected_info`, leaves the process alive and publishes no Will.

That is the same shape as the bug this PR fixes for `asleep`. It is **not** a regression — the base on `dev-58` has no `awake` clause either, so the issue's description of the existing clauses does not match this branch. But reaching into `awake` for one reason and leaving the other to fall through is hard to read as deliberate, and `awake` is reachable with a live socket (entered from `asleep` on PINGREQ).

Either handle it here or say in the description why `awake` is intentionally left out.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -1855,19 +1883,16 @@ handle_info(
...
     %    andalso emqx_flapping:detect(ClientInfo),
-    NChannel = ensure_disconnected(Reason, mabye_publish_will_msg(Channel)),
-    case maybe_shutdown(Reason, NChannel) of
-        {ok, NChannel1} -> {ok, {event, disconnected}, NChannel1};
-        Shutdown -> Shutdown
-    end;
+    handle_connection_lost(Reason, Channel);
 handle_info(
-    {sock_closed, Reason},
+    {sock_closed, _Reason},
```
zmstone: Two things about this clause.

**It needs a comment.** Ignoring a socket close is defensible for MQTT-SN over UDP — a sleeping client has no live socket to lose, and `asleep` exists so the session outlives the transport. But the *absence* of a clause here is precisely what #18638 reported as a bug, and this PR resolves it by ignoring the close deliberately rather than acting on it. Without a line saying why that is correct, the next reader redoes the same investigation and may well "fix" it back.

**It is the one behavior change with no test.** The issue's reproduction is a transport close while asleep. The new tests cover a client DISCONNECT while asleep (`t_asleep_test10_client_disconnect`) and the sleep timeout (`t_asleep_test01_timeout`), but nothing asserts that a transport close while asleep leaves the session intact and publishes no Will. Since this clause is the direct answer to the reported issue, and it does the opposite of what the issue asked for, it is the behavior most worth pinning.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -1844,6 +1851,27 @@ handle_info(
     Channel = #channel{conn_state = connecting}
 ) ->
     shutdown(Reason, Channel);
+handle_info(
+    {sock_closed, ?CLIENT_DISCONNECT},
```
zmstone, line 1857: These three clauses have identical bodies and differ only in `conn_state`. One clause with a guard would do:

```erlang
handle_info(
    {sock_closed, ?CLIENT_DISCONNECT},
    Channel = #channel{conn_state = ConnState}
) when
    ConnState =:= connected; ConnState =:= asleep; ConnState =:= awake
->
    handle_client_disconnect(Channel);
```

That also puts the state list in the same form as the one in `handle_out(disconnect, normal, ...)`, which enumerates the same three states. The two lists have to stay in sync: `{close, ?CLIENT_DISCONNECT}` is emitted only for those states, so a state added to one list and not the other falls through to the catch-all and is silently ignored.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -596,10 +597,11 @@ handle_in(
         previous_state => ConnState,
         clientid => ClientId
     }),
+    Channel1 = cancel_timer(keepalive, cancel_timer(expire_asleep, Channel)),
```
zmstone, line 602: This line does more than it looks, and is worth a comment.

`ensure_timer/2` arms a timer only when the name is absent from `#channel.timers`:

```erlang
TRef = maps:get(Name, Timers, undefined),
case TRef == undefined andalso is_integer(Time) andalso Time > 0 of
```

While asleep, a keepalive timeout is swallowed by the `ConnState =:= disconnected; ConnState =:= asleep` clause in `handle_timeout/3`, which returns `{ok, Channel}` without clearing the entry. The dead `TRef` stays in the map. So the `ensure_keepalive/1` that `handle_out(connack, ?SN_RC_ACCEPTED, ...)` performs on the way out of this function is a no-op, and the resumed connection runs with no keepalive supervision at all.

Cancelling here is what makes that `ensure_keepalive/1` actually arm a timer. Good fix, and `t_connect_from_asleep_restarts_keepalive` covers it — but the mechanism spans three functions, and a later reader trimming this as redundant would silently reintroduce the gap.

apps/emqx_gateway_mqttsn/test/emqx_sn_protocol_SUITE.erl
```diff
@@ -1929,6 +1946,87 @@ t_asleep_test08_to_disconnected(_) ->
...
+        send_willtopic_msg(Socket, WillTopic, QoS),
+        ?assertEqual(<<2, ?SN_WILLMSGREQ>>, receive_response(Socket)),
+        send_willmsg_msg(Socket, WillPayload),
+        ?assertEqual(<<3, ?SN_CONNACK, 0>>, receive_response(Socket)),
+
+        send_disconnect_msg(Socket, SleepDuration),
+        ?assertEqual(<<2, ?SN_DISCONNECT>>, receive_response(Socket)),
+
+        %% Let the original keepalive timer expire while the client is asleep.
+        timer:sleep(1500),
```
zmstone: This sleep is load-bearing — the test needs the original keepalive timer to fire while the client is asleep — but it pins the test to a timing assumption rather than an observable event.

Better tools are already in scope: `?retry` is used a few lines below, and `snabbkaffe` is included in the suite. Waiting on the keepalive timeout through a trace point would remove the fixed delay.

Same for the `timer:sleep(1500)` at line 1977 and the `timer:sleep(3000)` in `t_asleep_test01_timeout`. Not blocking, and the suite is already full of these — but new sleeps in a gateway suite are reliable candidates for the next flaky-test report, and this PR is already rewriting these functions.

zmstone (approved):

Issue #18640, opened 2026-08-30
GET /nodes silently omits nodes whose node_info RPC fails


zmstone:
### What happened?

`GET /api/v5/nodes` silently omits any node whose `node_info` RPC does not return a map. A node that
is stopping, leaving, or merely slow to answer disappears from the response entirely — it is not
reported as `stopped`, not reported as failed, just absent.

`apps/emqx_management/src/emqx_mgmt.erl`:

```erlang
list_nodes() ->
    Running = emqx:cluster_nodes(running),
    %% all stopped core nodes
    Stopped = emqx:cluster_nodes(stopped),
    DownNodes = lists:map(fun stopped_node_info/1, Stopped),
    [{Node, Info} || #{node := Node} = Info <- node_info(Running)] ++ DownNodes.
```

`node_info/1` is `emqx_rpc:unwrap_erpc(emqx_management_proto_v5:node_info(Nodes))`, i.e.
`erpc:multicall(Nodes, emqx_mgmt, node_info, [], 30000)`. On failure `unwrap_erpc/1` yields
`{error, Reason}`, which does not match the `#{node := Node} = Info` generator pattern, so the list
comprehension **drops the element**.

The window is between taking `emqx:cluster_nodes(running)` and the multicall landing — up to the
30 s RPC timeout.

### What did you expect to happen?

A node that was believed running but did not answer should still appear, with a status distinguishing
"did not respond" from `stopped`. The module already has the precedent for representing a node it
cannot reach:

```erlang
stopped_node_info(Node) ->
    {Node, #{node => Node, node_status => 'stopped', role => core}}.
```

### How can we reproduce it (as minimally and precisely as possible)?

1. Form a cluster of two or more nodes.
2. Begin stopping or leaving one node, or otherwise make its `emqx_mgmt:node_info/0` RPC fail or
   time out.
3. Call `GET /api/v5/nodes` during that window.
4. The node is missing from the response rather than listed with a status.

### Anything else we need to know?

**Impact is bounded and mostly cosmetic**, but the direction of the error is unhelpful: in a cluster
view a vanished row is indistinguishable from a *removed* node, which is the condition an operator is
watching for. Three nodes becoming two says nothing about whether the third left, died, or simply did
not answer within the timeout. The next poll corrects it once the node appears under
`emqx:cluster_nodes(stopped)`.

**Note on scope.** `/nodes` is a per-node statistics endpoint — it fans out `emqx_mgmt:node_info/0`
[...]

PR #18656, opened 2026-08-31
fix(mgmt): report a node that did not answer instead of dropping it

MorganaFuture:
Fixes: #18640

## Summary

`list_nodes/0` collected the multicall results with

```erlang
[{Node, Info} || #{node := Node} = Info <- node_info(Running)]
```

so a node whose `node_info` RPC failed produced `{error, Reason}`, missed the
generator pattern, and left the response entirely. In a cluster view a vanished
row is indistinguishable from a removed node, which is the condition an operator
is watching for.

`erpc:multicall/5` answers in request order and `unwrap_erpc/1` maps over that
list, so zipping the request list with the results recovers which node produced
which answer. A failure becomes an entry with `node_status` `unreachable`, a new
value in the schema. `role` is left out of it: `stopped_node_info/1` can say
`core` because `cluster_nodes(stopped)` yields core nodes only, but a running
node that stopped answering may be either, and `role` comes from the node's own
`mria_rlog:role()`. `format/1` already passes a partial map through, which is how
the `stopped` entry works today.

`node_info_entry/2` keeps a catch-all rather than matching only `{error, _}`.
The comprehension it replaces dropped every shape it did not match, so a
narrower match would turn an unexpected answer from a missing row into a 500 for
the whole endpoint.

`GET /nodes/{node}` had the same gap and is fixed with it. `with_node/2` admits
any node in `mria:running_nodes/0`, the same set the list endpoint fans out to,
and there `lookup_node/1` answers `{error, Reason}` while `format/1` takes a map
only, so it raised `function_clause` and reached the client as a 500. The fix is
in `get_node/1`, not `lookup_node/1`, because `t_lookup_node` pins the
`{error, _}` return deliberately.

The dashboard SPA is a separate artifact (`scripts/get-dashboard.sh`), so
rendering the new status is not something this PR can carry. The i18n text now
spells out all three values.

Local run: `emqx_mgmt_SUITE` 22/22, `emqx_mgmt_api_nodes_SUITE` 6/6,
`emqx_mgmt_api_cluster_SUITE` 5/5, `emqx_mgmt_api_status_SUITE` 10/10. Reverting
only `emqx_mgmt.erl` fails the new case with the silent node absent from the list.

## PR Checklist

- [x] Schema changes are backward compatible or intentionally breaking (new `node_status` value, additive)

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt.erl
+++ b/apps/emqx_management/src/emqx_mgmt.erl
@@ -22,6 +22,7 @@
     lookup_broker/1,
     node_info/0,
     node_info/1,
+    unreachable_node_info/1,
     broker_info/0,
     broker_info/1
 ]).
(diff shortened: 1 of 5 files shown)


MorganaFuture:
CI failure here is `emqx_ds_builtin_raft_SUITE:t_leader_otx_supervion`, same signature as #18553 which was a stabilization attempt for it. Nothing in this PR touches `emqx_ds_builtin_raft`; the change is in `emqx_mgmt:list_nodes/0`. The other 96 jobs passed.

zmstone (approved):

Issue #18653, opened 2026-08-31
[6.3.0] Audit logs cannot distinguish topic-metrics collections with the same name across namespaces


Le9Wang:
### What happened?

When two topic-metrics collections have the same name but belong to different namespaces, audit logs do not record the target namespace for reset operations.

For example, both `global/test` and `ns2/test` can exist. A global administrator can reset both collections successfully, but the corresponding audit records only contain the collection name `test`. It is impossible to determine which namespace was targeted.

### Reproduction Steps

1. Create a topic-metrics collection named `test` in namespace `ns2`.
2. Create another topic-metrics collection named `test` in the global namespace.
3. As a global administrator, reset the global collection:

   ```
   PUT /api/v5/mqtt/topic_metrics2/test/reset
   ```

4. Reset the `ns2` collection by specifying the namespace:

   ```
   PUT /api/v5/mqtt/topic_metrics2/test/reset?ns=ns2
   ```

5. Query the audit logs:

   ```
   GET /api/v5/audit
   ```

The issue is reliably reproducible.

### Expected Behavior

Each audit record should contain enough information to uniquely identify the target collection, for example:

- Collection name: `test`
- Namespace: `global` or `ns2`

The `ns` query parameter should be preserved in the audit log, or the resolved target namespace should be recorded as a dedicated field.

### Actual Behavior

Both audit records only identify the collection as `test`. The namespace is missing, so the records cannot distinguish between:

- `global/test`
- `ns2/test`

The reset operations themselves are executed against the correct collections.

### Test Case

```bash
# Reset the global collection
curl -X PUT \
  -H "Authorization: Bearer <global-admin-token>" \
  http://localhost:18083/api/v5/mqtt/topic_metrics2/test/reset

# Reset the ns2 collection
curl -X PUT \
  -H "Authorization: Bearer <global-admin-token>" \
  "http://localhost:18083/api/v5/mqtt/topic_metrics2/test/reset?ns=ns2"

# Query audit logs
curl -H "Authorization: Bearer <global-admin-token>" \
  http://localhost:18083/api/v5/audit
```

### Environment

- **Software version / build**: EMQX 6.3.0-beta.3-g81f6b7d8 (Enterprise)
- **Configuration**: Topic-metrics v2 enabled; audit logging enabled
- **OS / Platform**: Ubuntu 22.04

PR #18664, opened 2026-08-31
fix(audit): record target namespace and query params on audit entries

zmstone:
Fixes: #18653
Release version: 6.3.1, 7.0.0
Introduced in: 6.3.0

## Summary

`emqx_dashboard_audit:http_request/1` recorded only `method`, `headers`, `bindings` and `body`.
A namespace passed as a query parameter (`PUT /mqtt/topic_metrics2/:name/reset?ns=ns2`) was
therefore dropped before the audit record was written, so resetting `global/test` and `ns2/test`
produced indistinguishable audit entries.

Query parameters alone are not enough either: a namespaced admin's target namespace comes from
their dashboard token, not from `?ns=`, so a request with no query parameter at all still has a
definite target that a query-string-only fix would miss (see the `OwnerNs` comment in
`emqx_topic_metrics2_api.erl`).

This PR does both, and both are needed to close the issue:

- `emqx_dashboard_audit` records the parsed query string on the audit request
  (`http_request.query_string`), redacted through the same `emqx_utils:redact/1` key-based check
  already applied to headers and bodies. It's already a parsed map by the time it reaches the audit
  layer (see `minirest_handler:parse_qs/1`), not a raw query string, so redaction can see the keys.
  This is a generic, audit-layer-only change: it covers every endpoint that takes a query
  parameter, with no per-module changes.
- `emqx_topic_metrics2_api` calls `minirest_handler:update_log_meta/1` at the point it resolves the
  operation's target namespace (create, delete, delete-all, reset), so `http_request.namespace`
  records the *resolved* namespace — present whether or not `?ns=` was passed.

**Note on the diff since the last review:** the `emqx_dashboard_audit` capture logic and the two new
`emqx_audit_api:fields(http_request)` entries have since reached `dev-63` through the sync chain
(they were part of #18677 on `dev-60`), so they no longer appear here. What remains in
`emqx_dashboard_audit.erl` is only the simplification requested in review — folding `namespace` into
the existing `maps:with/2` call instead of a separate `maybe_put/3`.

**Survey of the wider issue** (the `ns` query parameter pattern is not unique to topic-metrics):
`emqx_a2a_registry_api`, `emqx_prometheus_api`, `emqx_bridge_v2_api`, `emqx_mt_api`,
`emqx_rule_engine_api` and `emqx_connector_api` all declare an `ns` query parameter the same way.
The generic `query_string` capture covers all of them — an explicit `?ns=` on any of those endpoints
is recorded. Recording the *resolved* namespace (needed for the implicit no-`?ns=` case) is
per-module; #18677 (`dev-60`, data backup) and #18678 (`dev-62`, A2A registry) cover two more.

`emqx_topic_metrics_registry`'s module doc (from #18534) says the reset request is audited "at the
REST API entrypoint like any other mutating request" — this issue is exactly that entrypoint
coverage being incomplete for namespaced resources. No per-node audit emission is re-added; the
entrypoint record is what's made sufficient.

### Manual verification (issue's exact repro)

Before: both records showed only `{"bindings":{"name":"test"}, ...}` — no way to tell `global/test`
from `ns2/test` apart. After, on a local build:

```
PUT /mqtt/topic_metrics2/test/reset            -> 204
PUT /mqtt/topic_metrics2/test/reset?ns=ns2     -> 204
GET /audit?operation_id=/mqtt/topic_metrics2/:name/reset
```

```json
{
  "http_request": {
    "bindings": {"name": "test"},
    "method": "put",
    "query_string": {"ns": "ns2"},
    "namespace": "ns2"
  }
},
{
  "http_request": {
    "bindings": {"name": "test"},
    "method": "put",
    "namespace": "global"
  }
}
```

The two records are now distinguishable, and the no-`?ns=` record correctly carries no
`query_string` key while still naming its namespace explicitly.

## PR Checklist

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_audit.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_audit.erl
@@ -138,8 +138,15 @@ operation_type(Meta) ->
     end.
 
 http_request(Meta) ->
+    %% `namespace' is the resolved target namespace, set by handlers that call
+    %% `minirest_handler:update_log_meta/1' (see emqx_topic_metrics2_api for
+    %% an example). It is not always the same as the `ns' query param: a
+    %% namespaced admin's namespace comes from their dashboard token, not
+    %% from the request, so recording it here also covers requests with no
+    %% `ns' query param at all. `maps:with/2' already omits it when absent
+    %% from `Meta', same as it does for `method'/`headers'/`bindings'/`body'.
     Request0 =
-        case maps:with([method, headers, bindings, body], Meta) of
+        case maps:with([method, headers, bindings, body, namespace], Meta) of
             #{body := Body} = Request when is_binary(Body) ->
                 Request#{body => <<"******">>};
             #{body := _} = Request ->
@@ -153,17 +160,10 @@ http_request(Meta) ->
     %% `query_string' is the parsed request query params (a map, not a raw
     %% binary), so the `emqx_utils:redact/1' call at the end of `log_meta/3'
     %% can see the keys and redact sensitive ones by name, same as it does
-    %% for `headers' and `body' above.
-    Request1 = maybe_put(
-        query_string, non_empty_map(maps:get(query_string, Meta, undefined)), Request0
-    ),
-    %% `namespace' is the resolved target namespace, set by handlers that
-    %% call `minirest_handler:update_log_meta/1' (see emqx_topic_metrics2_api
-    %% for an example). It is not always the same as the `ns' query param:
-    %% a namespaced admin's namespace comes from their dashboard token, not
-    %% from the request, so recording it here also covers requests with no
-    %% `ns' query param at all.
-    maybe_put(namespace, maps:get(namespace, Meta, undefined), Request1).
+    %% for `headers' and `body' above. Handled separately from the `with/2'
+    %% above because an empty query string must be omitted, not recorded as
+    %% `query_string => #{}'.
+    maybe_put(query_string, non_empty_map(maps:get(query_string, Meta, undefined)), Request0).
 
 maybe_put(_Key, undefined, Map) -> Map;
 maybe_put(Key, Value, Map) -> Map#{Key => Value}.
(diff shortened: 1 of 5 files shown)


apps/emqx_dashboard/src/emqx_dashboard_audit.erl
```diff
@@ -117,17 +134,38 @@ operation_type(Meta) ->
...
+    Request1 = maybe_put(
+        query_string, non_empty_map(maps:get(query_string, Meta, undefined)), Request0
+    ),
+    %% `namespace' is the resolved target namespace, set by handlers that
+    %% call `minirest_handler:update_log_meta/1' (see emqx_topic_metrics2_api
+    %% for an example). It is not always the same as the `ns' query param:
+    %% a namespaced admin's namespace comes from their dashboard token, not
+    %% from the request, so recording it here also covers requests with no
+    %% `ns' query param at all.
+    maybe_put(namespace, maps:get(namespace, Meta, undefined), Request1).
```
savonarola: Why not just add `namespace` to the `with` key list above (`maps:with([method, headers, bindings, body], Meta)`) 🤔

apps/emqx_dashboard/test/emqx_dashboard_audit_tests.erl
```diff
@@ -0,0 +1,69 @@
+%%--------------------------------------------------------------------
+%% Copyright (c) 2026 EMQ Technologies Co., Ltd. All Rights Reserved.
+%%--------------------------------------------------------------------
+-module(emqx_dashboard_audit_tests).
+
+-include_lib("eunit/include/eunit.hrl").
+
+%% `log_meta/3' only needs a plain map shaped like a cowboy request (in
```
savonarola, line 12: Nit: although it is clear from further reading, maybe add some general description what are these constants for, not only explain why they have the given shape 🤔

apps/emqx_dashboard/test/emqx_dashboard_audit_tests.erl
```diff
@@ -0,0 +1,69 @@
...
+        code => 204,
+        method => put,
+        auth_type => jwt_token,
+        operation_id => <<"/mqtt/topic_metrics2/:name/reset">>,
+        bindings => #{name => <<"test">>},
+        headers => #{},
+        body => #{},
+        req_start => 0,
+        req_end => 1000
+    }.
```
savonarola, line 33: Nit: why `REQ` is a macro but `base_meta` is a fun?

savonarola (dismissed):

savonarola (dismissed):

savonarola (approved):

PR #18677, opened 2026-09-01
fix(audit): record target namespace and query params on audit entries (data backup)

zmstone:
Fixes: N/A
Introduced in: 6.0.0

## Summary

Same class of gap as #18653 / PR #18664 (base `dev-63`): `emqx_dashboard_audit:http_request/1`
only copied `method`, `headers`, `bindings` and `body` into the stored audit request. A namespace
passed as a query parameter (`?namespace=`), or resolved implicitly from a namespaced admin's own
dashboard token with no query parameter at all, never reached `GET /audit`.

This branch (`dev-60`) does not yet have #18664's fix — it hasn't synced back from `dev-63` yet
(that PR is still open). This PR ports the same foundation to `dev-60` and applies it to
`apps/emqx_management/src/emqx_mgmt_api_data_backup.erl`, the one module on this branch with the
same bug shape (found via a follow-up survey after #18664 — see the audit note referenced below).

- `emqx_dashboard_audit` now also records the parsed query string on the audit request
  (`http_request.query_string`), redacted through the same `emqx_utils:redact/1` key-based check
  already applied to headers and bodies.
- `emqx_mgmt_api_data_backup` now calls `minirest_handler:update_log_meta/1` at the point each
  mutating handler resolves its target namespace (`POST /data/export`, `POST /data/import`,
  `POST /data/files` upload, `DELETE /data/files/:filename`), so `http_request.namespace` records
  the *resolved* namespace — present whether or not `?namespace=` was passed. `POST /data/export`
  in particular has no query override at all (only the implicit actor-namespace case), so it needed
  this even though it never had a raw query parameter to log.

`http_request.namespace` and `http_request.query_string` are new, optional fields on the
`GET /audit` response (already added to `emqx_audit_api:fields(http_request)` by this same PR, as
part of the ported foundation). No existing field changed shape.

### Local validation note

This worktree hit unrelated environment issues in this session: the multi-node CT cluster for
`emqx_mgmt_api_data_backup_SUITE` failed to boot nodes (`not_alive`) under heavy load on the shared
dev box (load average ~22 on 20 cores at the time) — confirmed this affects a pre-existing,
unmodified test case in the same suite (`t_export_backup`) identically, so it is not caused by this
change. `make fmt-diff` and `./scripts/elvis-check.sh` both pass. The equivalent fix + test pattern
already passed full CT (and a live manual repro) on the `dev-63` PR (#18664) this ports from.
Please let CI be the gate here.

## PR Checklist

### DIFF
--- a/apps/emqx_management/src/emqx_mgmt_api_data_backup.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_data_backup.erl
@@ -244,6 +244,7 @@ field_filename(IsRequired, Meta) ->
 
 data_export(post, #{body := Params0} = Req) ->
     Namespace = emqx_dashboard:get_namespace(Req),
+    ok = log_ns(Namespace),
     Params1 = emqx_utils_maps:put_if(Params0, <<"namespace">>, Namespace, is_binary(Namespace)),
     Params = maybe_filter_export_params(Params1, auth_meta(Req)),
     maybe
@@ -277,6 +278,7 @@ data_export(post, #{body := Params0} = Req) ->
 
 data_import(post, #{body := #{<<"filename">> := Filename} = Body, query_string := QS} = Req) ->
     Namespace = op_namespace(Req, QS),
+    ok = log_ns(Namespace),
     Nodes = emqx_bpapi:nodes_supporting_bpapi_version(?BPAPI_NAME, 2),
     case safe_parse_node(Body, Nodes) of
         {error, Msg} ->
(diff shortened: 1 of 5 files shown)


apps/emqx_management/src/emqx_mgmt_api_data_backup.erl
```diff
@@ -484,6 +489,17 @@ auth_meta(_) -> #{}.
 %% ignored for them. A global administrator may pass `namespace' to inspect or
 %% clean up a specific namespace's backups; without it, they operate on the
 %% global (non-namespaced) backups.
+%% Record the resolved target namespace on the audit log for this request
+%% (same class of gap as emqx/emqx#18653: the URL alone does not distinguish
+%% a global backup operation from a namespaced one). See
+%% emqx_dashboard_audit:http_request/1.
+log_ns(?global_ns) ->
+    _ = minirest_handler:update_log_meta(#{namespace => <<"global">>}),
```
thalesmg, line 497: "global" (binary) is only reserved/forbidden after 6.0, right?

thalesmg (approved):

apps/emqx_management/src/emqx_mgmt_api_data_backup.erl
zmstone: from 6.3.0 and it’s configurable

Issue #18673, opened 2026-09-01
[6.3.0] MQTT 5 CONNECT crashes when Will User-Property is present with OpenTelemetry enabled


Le9Wang:
### What happened?

When OpenTelemetry / external tracing is enabled, an MQTT 5 client connection containing Will User-Property entries causes the EMQX socket connection process to crash.

The MQTT packet is accepted and parsed, but EMQX fails while serializing the Will Properties into JSON.

### Reproduction Steps

1. Start EMQX with OpenTelemetry or external tracing enabled.
2. Connect an MQTT 5 client with Will Properties containing User-Property entries, for example:
   - `Will-Property1 = Will-Value1`
   - `Will-Property2 = Will-Value2`
3. Observe the EMQX log.

Example Python reproduction:

```python
import time
import paho.mqtt.client as mqtt
from paho.mqtt.properties import Properties
from paho.mqtt.packettypes import PacketTypes

client = mqtt.Client(protocol=mqtt.MQTTv5)

will_properties = Properties(PacketTypes.WILLMESSAGE)
will_properties.UserProperty = [
    ("Will-Property1", "Will-Value1"),
    ("Will-Property2", "Will-Value2"),
]

client.will_set(
    topic="test/will",
    payload="offline",
    qos=0,
    retain=False,
    properties=will_properties,
)

client.connect("127.0.0.1", 1883, 60)
client.loop_start()
time.sleep(3)
```

The issue is reliably reproducible when tracing is enabled.

### Expected Behavior

EMQX should successfully process the valid MQTT 5 CONNECT packet and establish the client connection.

If an internal tracing attribute cannot be serialized, EMQX should handle the error gracefully without crashing the socket connection process.

### Actual Behavior

The connection process crashes with:

```text
{invalid_ejson,{<<"Will-Property2">>,<<"Will-Value2">>}}
```

Relevant stack trace:

```text
jiffy:finish_encode/2
emqx_utils_json:encode/1
emqx_external_trace:connect_attrs/2
emqx_channel:handle_in/2
emqx_socket_connection:with_channel/3
```

The relevant code directly encodes the Will Properties:

```erlang
'client.will_props' => json_encode(WillProps)
```

However, MQTT 5 User-Property values are represented as a list of key-value tuples, which is not valid input for direct `jiffy:encode/1` encoding in this context.

The connection is then terminated because the socket connection process crashes.

### Suggested Fix

Use the proplist-aware encoder, consistent with connection properties:

```erlang
[...]

PR #18684, opened 2026-09-01
fix(otel): fix crash on CONNECT with will user-properties when tracing is on

zmstone:
Fixes: #18673
Introduced in: 5.8.3

## Summary

With OpenTelemetry tracing enabled, an MQTT 5 CONNECT whose will message carries `User-Property` entries crashed the connection process:

```
{invalid_ejson,{<<"Will-Property2">>,<<"Will-Value2">>}}
  jiffy:finish_encode/2
  emqx_utils_json:encode/1
  emqx_external_trace:connect_attrs/2
```

`connect_attrs/2` encoded `will_props` with `emqx_utils_json:encode/1`, which sends MQTT 5's `'User-Property'` (a list of `{Key, Value}` tuples) straight to jiffy. `conn_props` hit the exact same shape and was fixed in 1d5e783664 by switching to `encode_proplist/1`, which converts the tuple list to a JSON object first. `will_props` was left on the old encoder — the crash is that missed twin, 19 months later.

This also closes the second half of #18673: `connect_attrs/2` was the only attrs function in the module not wrapped in `?wrap(...)`, so any other unencodable CONNECT field (not just will/conn props) could still kill the connection. Wrapping the whole clause in `?wrap(...)` would have been a one-line fix, but `?fallback_attrs()` discards every connect attribute on one bad field, and the connect span is the richest one EMQX emits. Instead:

- `will_props` and `conn_props` now go through `safe_json_encode_proplist/1`, which degrades only the offending attribute to a `"encode_error"` marker on failure.
- The whole clause is also wrapped in `?wrap(...)` as a last-resort net, so nothing in `connect_attrs/2` can take down the connection process.

Will properties have no `'Authentication-Data'` field, so there's nothing to mirror from `redact_conn_props/1`.

Verified end to end against a real Jaeger/otel-collector stack (`.ci/docker-compose-file/docker-compose-otel.yaml`): connect with a will message and two will user-properties, kill the socket without DISCONNECT, and confirm both the `client.connect` span's `client.will_props` tag and the delivered will message carry the properties intact (`emqx_otel_SUITE:t_e2e_will_message_user_properties/1`).

### Manual repro (issue's Python script, node on non-default ports)

Before fix, client disconnects immediately and the node logs the exact crash from the issue:

```
$ python3 repro_18673.py 21883
connect() rc: 0
is_connected: False
```
```
crasher: initial call: emqx_connection:init/4, ...
error: {{invalid_ejson,{<<"Will-Property1">>,<<"Will-Value1">>}},
        [{jiffy,encode,2,...},
         {emqx_utils_json,encode,1,...},
         {emqx_external_trace,connect_attrs,2,...},
         {emqx_channel,handle_in,2,...},
         {emqx_connection,with_channel,3,...}]}
```

After fix, same script, same node, tracing still enabled (`trace_all: true`):

```
$ python3 repro_18673.py 21883
connect() rc: 0
is_connected: True
```

No crash in the node log.

## PR Checklist


apps/emqx/src/emqx_external_trace.erl
```diff
@@ -399,6 +399,23 @@ json_encode(Term) ->
...
+    try
+        json_encode_proplist(Properties)
+    catch
+        Class:Reason ->
+            ?SLOG(debug, #{
+                msg => "failed_to_encode_trace_attr",
+                class => Class,
+                reason => Reason
+            }),
+            <<"\"encode_error\"">>
```
zmstone: why escape quotes?

apps/emqx/test/emqx_external_trace_SUITE.erl
```diff
@@ -0,0 +1,178 @@
...
+-doc """
+An unencodable will-property value degrades only `client.will_props',
+without killing the connect attrs.
+""".
+t_connect_attrs_unencodable_will_prop_degrades(_) ->
+    WillProps = #{
+        'User-Property' => [{<<"Will-Property1">>, <<"Will-Value1">>}],
+        %% A bare 3-tuple is not valid ejson and is not converted by
+        %% `emqx_utils_json:to_ejson/1', so jiffy rejects it.
+        'Unencodable' => {1, 2, 3}
```
thalesmg, line 117: is this even possible, since we allow-list the valid properties?  this test case seems only possible because the channel initialization is heavily mocked?

thalesmg (approved):

apps/emqx/test/emqx_external_trace_SUITE.erl
zmstone: not from any MQTT packet parse result.
This is to exercise the try `catch` clause.

Issue #18626, opened 2026-08-30
Allow apps to inject readiness gate checks; gate listeners on named actions being ready


zmstone:
### What would you like to be added or enhanced?

Two related things.

**1. Let applications inject readiness checks into the node readiness gate.**

Today `emqx_node_readiness` holds a single boolean in a `persistent_term`, set once by
`emqx_machine_boot:ensure_apps_started/0`. It is consumed by `emqx_connection`,
`emqx_socket_connection`, `emqx_quic_connection` and `emqx_cluster`. Its meaning is therefore
fixed at "all applications have started".

Proposal: allow applications to register a predicate, so the gate can mean "this node is ready
to serve traffic":

```erlang
emqx_node_readiness:register_check(Name, fun() -> boolean() end).
emqx_node_readiness:deregister_check(Name).

%% is_ready/0 returns true only when boot has completed
%% AND every registered check returns true.
emqx_node_readiness:is_ready().
```

**2. First consumer: gate listeners on named actions being ready.**

```
rule_engine.service_readiness_gate_actions = ["name1", "name2"]
```

Listeners do not accept connections until every listed action reports ready on the local node.
If a listed action does not exist, log an error and carry on rather than blocking on it.

Local action status is already available cheaply through `emqx_bridge_v2:lookup/4` — the same
function the REST layer uses as its RPC target — so the check is a local lookup with no RPC.

### Why is this needed?

**1. "All applications started" is not the same as "this node can serve traffic".** Connectors
are started asynchronously by `emqx_resource_manager`: a resource enters `?status_connecting`
and only reaches `?status_connected` on a later health check. A node can therefore report ready
while its actions are not yet installed.

**2. The consequence today is silent message loss on startup.** A node that boots while the
external system is briefly unreachable accepts MQTT connections immediately. A message routed
by a rule to a not-yet-installed action fails with an unrecoverable error
(`emqx_resource_buffer_worker:error_if_channel_is_not_installed/2`, which is unrecoverable for
internal-buffer actions such as Kafka and Pulsar), the rule engine discards the action result,
and the publishing client still receives a success acknowledgement.

Measured on `dev-63`: a node started while its Kafka broker was unreachable acknowledged 60 of
60 published messages and delivered none of them. Publish latency was 0 ms — no write was
attempted — and the messages never arrived, even after the broker recovered. From the client's
point of view the publishes succeeded.

**3. This is the unfixed mirror of #18523.** That PR closed the same window on shutdown, by
[...]

PR #18690, opened 2026-09-01
feat(readiness): let applications register readiness checks

MorganaFuture:
Release version: 7.0.0

Introduced in: N/A

## Summary

Part 1 of https://github.com/emqx/emqx/issues/18626 — the readiness-check registry
only. `emqx_node_readiness:register_check/2` and `deregister_check/1`; `is_ready/0`
returns true only once boot has completed and every registered check returns true.

The registry lives in the same `persistent_term` the flag already uses, so `is_ready/0`
stays lock-free. It is read once per incoming connection — `emqx_connection`,
`emqx_ws_connection`, `emqx_quic_connection`, `emqx_socket_connection` and
`emqx_gateway_conn` all call it at init — so an empty registry short-circuits on
`map_size/1`, and the moduledoc says a check has to be cheap. Registration is a
read-modify-write, so it is documented as safe only from a serialized context (an
application's `start/2`/`stop/1`); adding a process to serialize it looked heavier than
the problem.

A check that raises, or returns a non-boolean, holds the gate closed — it did not
"return true". That is the opposite of the module's existing
`persistent_term:get(?KEY, true)` default, so say the word if you would rather have it
fail open. Either way it is reported through a throttled `readiness_check_failed`, which
is the one atom added to `?LOG_THROTTLING_MSGS`.

No consumer here, so nothing user-visible changes and there is no changelog entry; that
belongs with part 2.

Part 2 is not in this PR: it needs the three open questions answered, and @keynslug's
objection on the thread is unanswered. One thing from the query path that bears on it —
`is_channel_apt_for_queries/1` already passes `connecting`, and a channel is
`disconnected` only until `on_add_channel/4` returns, so the loss window is "channel not
installed" rather than "action not connected". Gating on installed closes the same window
without deriving readiness from external reachability.

Tests: `emqx_node_readiness_SUITE` 12/12 (6 new), `emqx_log_throttler_SUITE` 6/6,
`emqx_machine_SUITE` 6/6, `emqx_conf_schema_tests` 50/50.

## PR Checklist

- [ ] Change log — none; no user-visible behaviour without a consumer
- [x] Schema changes are backward compatible or intentionally breaking (additive: one atom added to the `log.throttling.msgs` enum and its default)

### DIFF
--- a/apps/emqx/src/emqx_node_readiness.erl
+++ b/apps/emqx/src/emqx_node_readiness.erl
@@ -12,24 +12,22 @@ refuse to serve until the node is ready, so no client can connect
 before authentication, authorization and plugin hooks are installed.
 The `GET /status` REST API and cluster join checks read it too.
 
-The gate is closed while the node is booting and re-opened by the
-managed boot sequence (`emqx_machine_boot:ensure_apps_started/0`) once
-all applications (including plugins) are started.  Contexts that do
-not boot through `emqx_machine` (test suites) are therefore always
-ready.
+The readiness flag defaults to `true`.  The managed boot sequence
+(`emqx_machine_boot:ensure_apps_started/0`) clears it while the node
+boots and sets it back once all applications (including plugins) are
+started.  Contexts that do not boot through `emqx_machine` (test
+suites) never clear it and are therefore always ready.
 
-"All applications started" is not always the same as "this node can
-serve traffic", so applications can narrow the gate further by
-registering a predicate with `register_check/2`.  `is_ready/0` returns
-`true` only once boot has completed *and* every registered check
-returns `true`.
+An application can add a further condition by registering a predicate
+with `register_check/2`.  `is_ready/0` returns `true` only once boot
+has completed and every registered check returns `true`.
 
 Registered checks are evaluated on every `is_ready/0` call, and that
 happens once per incoming connection.  A check must therefore be
 cheap: poll in your own process and let the predicate read the cached
 answer.  A check that raises, or returns anything other than a
-boolean, holds the gate closed and is reported through a throttled
-`readiness_check_failed` log.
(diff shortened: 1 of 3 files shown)


apps/emqx/src/emqx_node_readiness.erl
```diff
@@ -38,3 +66,67 @@ mark_ready() ->
...
+Register a readiness check under `Name`, replacing any check already
+registered under it.  While the check returns anything but `true` the
+node reports itself as not ready.
+""".
+-spec register_check(check_name(), check()) -> ok.
+register_check(Name, Check) when is_function(Check, 0) ->
+    persistent_term:put(?CHECKS, maps:put(Name, Check, checks())).
+
+-doc "Remove the readiness check registered under `Name`, if any.".
+-spec deregister_check(check_name()) -> ok.
```
zmstone, line 80: 2nd thought 
maybe no need for apps to deregister 

machine should reduce the register-injected map to a empty list once all are ready.
the list is repopulated at app start callbacks.

this reduces the hot path to a single persistent term lookup without list fold

apps/emqx/src/emqx_node_readiness.erl
```diff
@@ -38,3 +66,67 @@ mark_ready() ->
...
     persistent_term:put(?KEY, false).
+
+-doc """
+Register a readiness check under `Name`, replacing any check already
+registered under it.  While the check returns anything but `true` the
+node reports itself as not ready.
+""".
+-spec register_check(check_name(), check()) -> ok.
+register_check(Name, Check) when is_function(Check, 0) ->
+    persistent_term:put(?CHECKS, maps:put(Name, Check, checks())).
```
zmstone, line 77: use list for deterministic order

Issue #18687, opened 2026-09-01
GET /api/v5/audit returns 500 for records created by SSO users


hjianbo:
#### Environment

- EMQX version: 5.10.4 (reported customer environment)
- SSO: LDAP (observed; the same issue is expected for other SSO backends)
- Audit logging: enabled
- API: `GET /api/v5/audit`

#### What happened?

When an SSO-authenticated dashboard user performs an audited operation, querying the audit log can return HTTP 500 instead of JSON.

The captured response contains audit records with an SSO user identifier represented as an Erlang tuple:

```text
500: invalid json term: #{data => [... source => {ldap,<<"<redacted-user>">>}, ...], meta => #{...}}
```

The captured response contained six records with this shape. Affected operation IDs included:

```text
/users/:username
/actions_probe
/nodes/:node/actions/:id/:operation
```

#### Expected behavior

`GET /api/v5/audit` should return HTTP 200 with a valid JSON response. The `source` field should be serialized as a JSON-compatible string or another documented JSON value.

#### How can we reproduce it?

1. Enable audit logging and configure an SSO backend such as LDAP.
2. Log in through SSO and obtain a dashboard Bearer token.
3. Use that token to perform any authenticated, audited operation, for example:

   ```http
   PUT /api/v5/users/<target-username>
   Authorization: Bearer <sso-token>
   Content-Type: application/json

   {"description":"updated by repro","role":"administrator"}
   ```

4. Query the audit log:

   ```http
   GET /api/v5/audit?limit=20
   Authorization: Bearer <sso-token>
   ```

5. If the response page contains the audit record created in step 3, the request fails with `500: invalid json term`.

A minimal deterministic reproduction of the failing serialization seam is:

```erlang
Bad = #{
    data => [#{source => {ldap, <<"user">>}, operation_id => <<"/users/:username">>}],
    meta => #{count => 1, page => 1, limit => 20, hasnext => false}
},
minirest_json_encoder:encode(Bad).
```

This returns a 500 response containing `invalid json term`. Replacing `{ldap, <<"user">>}` with `<<"user">>` returns valid JSON.

#### Impact

- The Dashboard audit log page cannot load pages containing affected records.
- Any client querying `/api/v5/audit` is affected, not only the SSO user who created the records.
- A single invalid record makes the whole response page fail.
- Local-user and API-key requests generally do not trigger the issue because their source is a binary value.

#### Affected versions

The following release tags were inspected in the repository:

[...]

PR #18696, opened 2026-09-01
fix(audit): stringify SSO admin identity in audit API responses

zmstone:
Fixes: #18687
Release version: 6.2.4, 6.3.0, 7.0.0
Introduced in: pre-5.8

## Summary

`GET /api/v5/audit` returned HTTP 500 (`invalid json term`) whenever a returned page contained an audit record created by an SSO-authenticated admin. `emqx_audit_api:format/1` passed the `source` field straight through to the response map, but for SSO admins `source` is a `{Backend, Name}` tuple (`?SSO_USERNAME/2`), while the swagger schema for this field already declares `binary()`. One bad record failed JSON encoding for the whole page, for every caller, not just the SSO user.

Fix: stringify `source` as `"backend:name"` (e.g. `"ldap:some_user"`) only at the audit-record to JSON-response boundary, in `emqx_audit_api:format/1`. The tuple shape is left untouched everywhere else it's relied upon, such as admin lookup in `emqx_dashboard_api:caller_admin/1`.

Also tightened `emqx_dashboard:auth_meta()`'s `source` type from `binary()` to `dashboard_username()`, since a tuple was already stored there for SSO sessions — the old type annotation was inaccurate.

This is a reported 6.3.0 release-blocking bug (#18687). Sibling PRs are landing independently on `dev-60` and `dev-510` for the same issue, so this fix shows up more than once by design.

## PR Checklist


apps/emqx_audit/test/emqx_audit_api_SUITE.erl
```diff
@@ -108,6 +109,53 @@ t_http_api(_) ->
     ),
     ok.
 
+-doc """
+GET /audit must not 500 when a page includes a record written by an
+SSO-authenticated user, whose `auth_meta.source` is `{Backend, Name}`
+instead of a plain binary.
+""".
+t_http_api_sso_source(_) ->
+    process_flag(trap_exit, true),
```
thalesmg, line 118: why is this needed?

apps/emqx_audit/src/emqx_audit_api.erl
```diff
@@ -326,7 +326,7 @@ format(Audit) ->
         created_at => emqx_utils_calendar:epoch_to_rfc3339(CreatedAt, microsecond),
         node => Node,
         from => From,
-        source => Source,
+        source => emqx_dashboard_admin:format_username(Source),
```
thalesmg, line 329: maybe we could format _as we emit the audit entry_ instead of here, so we avoid emqx_audit declaring a dependency on emqx_dashboard?  not sure if feasible.

thalesmg (dismissed):

apps/emqx_audit/src/emqx_audit_api.erl
zmstone: for existing deployments, there are already entries in the datbase.

apps/emqx_audit/test/emqx_audit_api_SUITE.erl
zmstone: copy-paste from other tests.

zmstone: probably not needed.

PR #18697, opened 2026-09-01
fix(audit): stringify SSO source in GET /audit responses

zmstone:
Fixes: #18687
Introduced in: 6.0.3

## Summary

`GET /api/v5/audit` returned HTTP 500 (`invalid json term`) whenever a page included an audit record created by an SSO-authenticated dashboard user. For such users, `auth_meta.source` is the tuple `{Backend, Name}` instead of a binary, and `emqx_audit_api:format/1` passed it straight into the response map even though the swagger schema for that field already declares `binary()`. A single such record failed the whole page for every caller.

Fix:
- Add `emqx_dashboard_admin:format_username/1`, which folds `{Backend, Name}` into the binary `"Backend:Name"` and passes plain binaries through unchanged.
- Call it from `emqx_audit_api:format/1` at the audit-record-to-JSON boundary, replacing `source => Source` with `source => emqx_dashboard_admin:format_username(Source)`.

`auth_meta.source` keeps its tuple shape everywhere else — other code (e.g. admin lookup in `emqx_dashboard_api:caller_admin/1`) depends on that shape.

This is the dev-60 follow-up to the same fix landing on dev-63 (flagged as a 6.3.0 release blocker) and dev-510.

## PR Checklist

### DIFF
--- a/apps/emqx_audit/test/emqx_audit_api_SUITE.erl
+++ b/apps/emqx_audit/test/emqx_audit_api_SUITE.erl
@@ -5,7 +5,7 @@
 -compile(export_all).
 -compile(nowarn_export_all).
 
--include("../../emqx_dashboard/include/emqx_dashboard.hrl").
+-include_lib("emqx_dashboard/include/emqx_dashboard.hrl").
 -include_lib("eunit/include/eunit.hrl").
 -include_lib("common_test/include/ct.hrl").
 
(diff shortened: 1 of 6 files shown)


apps/emqx_audit/test/emqx_audit_api_SUITE.erl
```diff
@@ -5,6 +5,7 @@
 -compile(export_all).
 -compile(nowarn_export_all).
 
+-include("../../emqx_dashboard/include/emqx_dashboard.hrl").
```
zmstone: include_lib

hjianbo (approved):

PR #18706, opened 2026-09-01
fix(cli): redact secrets in debug logs

zmstone:
Fixes:
Release version: 6.3.0, 7.0.0
Introduced in: pre-5.8

## Summary

The `bin/emqx` script enables shell xtrace (`set -x`) when `DEBUG=1` or `DEBUG=2` is set. The trace output printed the Erlang distribution cookie at every point where the cookie flows through a shell variable:

- the cookie resolution block (`EMQX_NODE__COOKIE`, `node.cookie`, `file://` sources);
- the `remsh` and boot command building blocks (`-setcookie` / `--cookie` arguments);
- `relx_nodetool` (`ERL_FLAGS` assignment);
- the `eval-ex` command;
- `get_boot_config`, which echoes the whole `EMQX_BOOT_CONFIGS` string (it carries `node.cookie` and `license.key`);
- `maybe_warn_default_cookie` and `diagnose_boot_failure_and_die` (cookie in the test expression and `ps` output);
- `check_license` (license key in the `key_license` assignment and the nodetool argument).

The `DEBUG=1` `logdebug` lines also printed `PS_LINE` and `EMQX_BOOT_CONFIGS`, which carry the cookie.

Independently of `DEBUG`, every node start wrote the full boot command, cookie included, to syslog via `logger`. This applies only to this branch and later: the cookie moved from the generated vm.args file to the boot command line during the 6.3 development cycle, so no GA release is affected.

The fix turns tracing off around each secret-bearing region and re-arms it after, following the `set +x` / `set -x` pattern already used in the script. The `logdebug` lines and the syslog `EXEC:` line now mask the cookie and license key values as `******`. Tracing everywhere else is unchanged.

The dev-60 counterpart is #18708 (the debug-trace half only; the pre-6.3 script has no cookie on the boot command line, and its `generate_config` vm.args merge loop needs a wrap this branch does not).

Manual validation (all print 0 cookie occurrences after the fix, nonzero before):

```sh
DEBUG=1 EMQX_NODE__COOKIE=secret ./bin/emqx console 2>&1 | grep -c secret
```

Verified for `EMQX_NODE__COOKIE`, `node.cookie` in `emqx.conf`, and `file://` cookie sources, with both `DEBUG=1` and `DEBUG=2`, for the `console`, `start`, `stop`, `ctl`, `ping`, and `remote_console` commands. Verified the syslog entry shows `-setcookie ******`. Verified a normal (non-DEBUG) `start`/`ctl status`/`stop` cycle still works and that tracing is re-armed after each masked region.

## PR Checklist

### DIFF
--- a/bin/emqx
+++ b/bin/emqx
@@ -50,6 +50,20 @@ die() {
     exit "$errno"
 }
 
+# Print the given arguments space-joined, with the value following
+# any -setcookie/--cookie flag masked.
+redact_cookie_args() {
+    local prev='' tok out=''
+    for tok in "$@"; do
+        case "$prev" in
+            -setcookie|--cookie) out="${out}****** " ;;
+            *) out="${out}${tok} " ;;
+        esac
+        prev="$tok"
+    done
+    printf '%s' "${out% }"
+}
+
 THIS_SCRIPT="$0"
 
 usage() {
@@ -410,6 +424,10 @@ remsh() {
     # transparently
     id="remsh$(gen_node_id)-${NAME}"
 
+    # Turn off debug (and keep it off through the exec below) so the
+    # cookie in the command arguments isn't traced by 'set -x'.
+    set +x
+
     # shellcheck disable=SC2086
     # Setup remote shell command to control node
     if [ "${EMQX_CONSOLE_FLAVOR:-erl}" = 'erl' ] ; then
@@ -455,9 +473,14 @@ call_nodetool() {
 
 # Control a node
 relx_nodetool() {
+    local ret=0
     command="$1"; shift
+    # Turn off debug so the cookie value isn't traced by 'set -x'.
+    set +x
     ERL_FLAGS="${ERL_FLAGS:-} $EPMD_ARGS -setcookie $COOKIE" \
-             call_nodetool "$NAME_TYPE" "$NAME" "$command" "$@"
+             call_nodetool "$NAME_TYPE" "$NAME" "$command" "$@" || ret=$?
+    if [ "$DEBUG" -gt 0 ]; then set -x; fi
+    return $ret
 }
 
 call_hocon() {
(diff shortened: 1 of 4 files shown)


bin/emqx
```diff
@@ -556,13 +580,16 @@ else
         EMQX_BOOT_CONFIGS="$(call_hocon -s "$SCHEMA_MOD" -c "$EMQX_ETC_DIR"/emqx.conf multi_get "${CONF_KEYS[@]}")"
     fi
 fi
-logdebug "EMQX_BOOT_CONFIGS: $EMQX_BOOT_CONFIGS"
+logdebug "EMQX_BOOT_CONFIGS: $(echo -e "$EMQX_BOOT_CONFIGS" | sed -E 's/^(node\.cookie|license\.key)=.*/\1=******/')"
```
thalesmg: since EMQX_BOOT_CONFIGS is the output of call_hocon, I wonder if we could use hocon itself to do the redacting, so other sensitive keys are also handled?

thalesmg (dismissed):

bin/emqx
zmstone: cookie cannot be redacted for EMQX_BOOT_CONFIGS
this redaction is for boot script logdebug.

bin/emqx
```diff
@@ -235,12 +249,16 @@ export IS_ENTERPRISE
 
 ## Boot-time environment variables (EMQX_FEATURES, EMQX_SECURITY_PROFILE, ...).
 ## The file is optional. Values in the file override the inherited environment.
+# Turn off debug: the file may set secrets such as EMQX_NODE__COOKIE,
+# which must not be traced by 'set -x'.
+set +x
```
zmstone: [P1] Disable xtrace before sourcing `releases/emqx_vars`

This `set +x` still comes after `. "$RUNNER_ROOT_DIR"/releases/emqx_vars`. That file explicitly tells operators not to replace it during upgrades and to append overrides under `updated vars here`; it also directs them to configure `EMQX_NODE_COOKIE`. With `DEBUG=1`/`2`, a cookie or license key appended there is printed by Bash while the file is sourced (for example, `++ EMQX_NODE_COOKIE=secret`) before execution reaches this boundary. Please move the boundary above both source operations and extend the smoke test to cover this persisted override location.

hjianbo (approved):

PR #18708, opened 2026-09-01
fix(cli): do not print the Erlang cookie in debug trace

zmstone:
Fixes:
Introduced in: pre-5.8

## Summary

The `bin/emqx` script enables shell xtrace (`set -x`) when `DEBUG=1` or `DEBUG=2` is set. The trace output printed the Erlang distribution cookie at every point where the cookie flows through a shell variable:

- the cookie resolution block (`EMQX_NODE__COOKIE`, `node.cookie`);
- the `remsh` command building block (`-setcookie` / `--cookie` arguments);
- `relx_nodetool` (`ERL_FLAGS` assignment);
- the `generate_config` vm.args merge loop (on this branch the generated args carry the `-setcookie` line);
- the `eval-ex` command;
- `get_boot_config`, which echoes the whole `EMQX_BOOT_CONFIGS` string (it carries `node.cookie` and `license.key`);
- `maybe_warn_default_cookie` and `diagnose_boot_failure_and_die` (cookie in the test expression and `ps` output);
- `check_license` (license key in the `key_license` assignment and the nodetool argument).

The `DEBUG=1` `logdebug` lines also printed `PS_LINE` and `EMQX_BOOT_CONFIGS`, which carry the cookie.

The fix turns tracing off around each secret-bearing region and re-arms it after, following the `set +x` / `set -x` pattern already used in the script. The `logdebug` lines now mask the cookie and license key values as `******`. Tracing everywhere else is unchanged; in particular the main boot command building block stays traced, because on this branch the boot command line does not carry the cookie (it goes through the generated vm.args file).

The dev-63 counterpart is #18706 (the script diverged there: the cookie moved to the boot command line, which adds sites this branch does not have).

Manual validation (all print 0 cookie occurrences after the fix, nonzero before):

```sh
DEBUG=1 EMQX_NODE__COOKIE=secret ./bin/emqx console 2>&1 | grep -c secret
```

Verified with both `DEBUG=1` and `DEBUG=2` for the `console`, `start`, `stop`, `ctl`, `ping`, and `remote_console` commands, including non-boot commands reading the cookie back from the running node's `ps` output. Verified a normal (non-DEBUG) `start`/`ctl status`/`stop` cycle still works and that tracing is re-armed after each masked region.

## PR Checklist

### DIFF
--- a/bin/emqx
+++ b/bin/emqx
@@ -494,7 +494,7 @@ if [ -n "${EMQX_NODE_COOKIE:-}" ]; then
 fi
 
 PS_LINE="$(find_emqx_process)"
-logdebug "PS_LINE=$(echo -e "$PS_LINE" | sed -E 's/(-setcookie) +[^ ]+/\1 ******/g')"
+logdebug "PS_LINE=$(printf '%s' "$PS_LINE" | sed -E 's/(-setcookie) +[^ ]+/\1 ******/g')"
 RUNNING_NODES_COUNT="$(echo -e "$PS_LINE" | sed '/^\s*$/d' | wc -l)"
 [ "$RUNNING_NODES_COUNT" -gt 1 ] && logdebug "More than one running node found: count=$RUNNING_NODES_COUNT"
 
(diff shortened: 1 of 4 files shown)


bin/emqx
```diff
@@ -556,13 +566,16 @@ else
         EMQX_BOOT_CONFIGS="$(call_hocon -s "$SCHEMA_MOD" -c "$EMQX_ETC_DIR"/emqx.conf multi_get "${CONF_KEYS[@]}")"
     fi
 fi
-logdebug "EMQX_BOOT_CONFIGS: $EMQX_BOOT_CONFIGS"
+logdebug "EMQX_BOOT_CONFIGS: $(echo -e "$EMQX_BOOT_CONFIGS" | sed -E 's/^(node\.cookie|license\.key)=.*/\1=******/')"
```
thalesmg: same musing about using hocon itself to redact all sensitive values as in the other pr

thalesmg (dismissed):

hjianbo (approved):

Issue #18715, opened 2026-09-02
[6.3.0]OIDC metadata loader reports bad_certificate when managed_certs bundle is configured with verify_none


lina50:
## Environment

- EMQX Dashboard/API: 10.42.4.242:18083
- OIDC issuer: https://<oidc-host>:8443/realms/emqx
- Cluster: 3 nodes
- Related context: PR #18624

## Configuration

```hocon
dashboard.sso.oidc {
  enable = true
  issuer = "https://<oidc-host>:8443/realms/emqx"
  ssl {
    enable = false
    verify = verify_none
    managed_certs { bundle_name = fd }
  }
}
```

The managed certificate bundle `fd` contains a `chain.pem` entry only; no client `cert` or `key` is configured.

<img width="1124" height="837" alt="Image" src="https://github.com/user-attachments/assets/6504f463-4165-49ad-be82-a2aec6ad867b" />

<img width="1369" height="880" alt="Image" src="https://github.com/user-attachments/assets/7bd14787-7c99-4582-9b88-5f155acbc72e" />

## Steps to reproduce

1. Enable Dashboard OIDC with an HTTPS issuer.
2. Configure `dashboard.sso.oidc.ssl.managed_certs` to reference bundle `fd`.
3. Keep `ssl.verify = verify_none` and `ssl.enable = false`.
4. Restart EMQX.
5. Observe the EMQX error log.
6. Remove `managed_certs`, restart EMQX, and observe that the repeated errors stop.

## Actual behavior

After restart, EMQX repeatedly logs:

```text
Metadata load failed for issuer https://<oidc-host>:8443/realms/emqx. Retrying in <interval> ms. Error Details: {configuration_load_failed,{failed_connect,[{to_address,{<oidc-host>,8443}},{inet,[inet],{tls_alert,{bad_certificate,...}}}]}}
```

The decoded TLS detail is similar to:

```text
TLS client: In state wait_cert_cr at ssl_handshake.erl:2219 generated CLIENT ALERT: Fatal - bad certificate
```

The OIDC metadata endpoint itself is reachable and returns HTTP 200 when accessed with a client that accepts the self-signed server certificate.

When the `managed_certs` reference is removed and EMQX is restarted, the repeated `bad_certificate` messages stop. Re-adding the bundle may not print immediately until the next restart, suggesting the failure occurs during OIDC TLS worker initialization or metadata refresh.

## Expected behavior

[...]

PR #18724, opened 2026-09-02
fix(sso): reject oidc https issuer with ssl.enable=false at config time

zmstone:
Fixes: #18715
Release version:
Introduced in: pre-5.8

## Summary

Dashboard SSO OIDC's config accepted an `https` issuer with `ssl.enable = false`. At runtime, `emqx_tls_lib:to_client_opts/2` treats `enable =/= true` the same as "no TLS options at all", so the request ends up with no `ssl` httpc option and falls back to OTP's default `verify_peer` with no configured CA. Against a self-signed or private-CA issuer this fails every request with an opaque `bad_certificate`-style error, with no indication that `ssl.enable` was the cause.

This PR is a config-time validation fix, not a change to the underlying TLS-option-construction behavior. `emqx_dashboard_sso_oidc:create/1` now rejects `https` issuer + `ssl.enable = false` up front, mirroring the existing `check_ssl_opts/1` pattern in `emqx_auth_http/src/emqx_authn_http.erl`:

```erlang
check_ssl_opts(#{issuer := Issuer, ssl := #{enable := false}}) ->
    case is_https(Issuer) of
        true -> {error, {invalid_ssl_opts, <<"it's required to enable the TLS option to establish a https connection">>}};
        false -> ok
    end;
check_ssl_opts(_Config) -> ok.
```

`create/1` is the single entry point used by both `create/1` and `update/2` (which calls `create/1` internally), so both config-write paths get the check. An `https` issuer with the `ssl` field entirely absent is left unchanged (not rejected) — this matches how many working configurations (public CAs, no custom `ssl` block) function today via OTP's default `verify_peer`, and is not the combination this issue reports.

The `emqx_dashboard_sso` behaviour has no `validations/0`-style HOCON-schema-level hook (unlike `emqx_authn_schema`), so this couldn't be wired as a HOCON struct-level validation without extending that behaviour — a larger change than this fix warrants. `create/1` gives the same real error instead of the current silent runtime failure.

Out of scope for this PR (left for later discussion):
- The underlying `to_client_opts/2` behavior of discarding all SSL options (including `verify`) when `ssl.enable =/= true` — present unchanged since OIDC's configurable-SSL-options feature shipped in 6.0.1/5.9.3, and reproducible on every branch from dev-58 through dev-70.
- A structurally similar issue in `emqx_auth_jwt/src/emqx_authn_jwks_client.erl` (JWKS fetch over https).
- `emqx_dashboard_sso_saml.erl`'s IdP metadata fetch, which has no configurable `ssl`/`verify` field at all (different shape of problem).
- Backporting this validation to dev-58/59/510/61/62/63/70 — the same root mechanism exists on all of them, but that's a separate decision.

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ee/fix-18724.en.md`

### DIFF
--- a/apps/emqx_dashboard_sso/test/emqx_dashboard_sso_oidc_tests.erl
+++ b/apps/emqx_dashboard_sso/test/emqx_dashboard_sso_oidc_tests.erl
@@ -115,6 +115,14 @@ check_ssl_opts_test_() ->
                     ssl => #{enable => false}
                 })
             )},
+        {"http issuer with ssl enabled is rejected",
+            ?_assertMatch(
+                {error, {invalid_ssl_opts, _}},
+                emqx_dashboard_sso_oidc:check_ssl_opts(#{
+                    issuer => <<"http://example.com">>,
+                    ssl => #{enable => true}
+                })
+            )},
         {"https issuer without an ssl config is allowed",
             ?_assertEqual(
                 ok,
@@ -129,6 +137,17 @@ create_rejects_https_issuer_with_ssl_disabled_test() ->
         name_var => <<"${sub}">>,
         issuer => <<"https://example.com">>,
         ssl => #{enable => false}
+    },
+    ?assertMatch(
+        {error, {invalid_ssl_opts, _}},
+        emqx_dashboard_sso_oidc:create(Config)
+    ).
+
+create_rejects_http_issuer_with_ssl_enabled_test() ->
+    Config = #{
+        name_var => <<"${sub}">>,
+        issuer => <<"http://example.com">>,
+        ssl => #{enable => true}
     },
     ?assertMatch(
         {error, {invalid_ssl_opts, _}},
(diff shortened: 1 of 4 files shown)


apps/emqx_dashboard_sso/test/emqx_dashboard_sso_oidc_tests.erl
```diff
@@ -88,3 +88,49 @@ issuer_validation_test_() ->
...
+                })
+            )},
+        {"http issuer with ssl disabled is allowed",
+            ?_assertEqual(
+                ok,
+                emqx_dashboard_sso_oidc:check_ssl_opts(#{
+                    issuer => <<"http://example.com">>,
+                    ssl => #{enable => false}
+                })
+            )},
```
thalesmg, line 117: should we also disallow http + ssl.enabled = true?

thalesmg (dismissed):

apps/emqx_dashboard_sso/test/emqx_dashboard_sso_oidc_tests.erl
zmstone: will need to test out this odd combination.

zmstone: fixed, now url scheme must match ssl config.

thalesmg (approved):

Issue #18711, opened 2026-09-02
[6.3.0]audit hook crashes on SSO login source atom after PR #18696


lina50:
## Summary

PR #18696 introduces a regression when Dashboard audit is enabled and an SSO login is initiated.

The request process crashes in the audit hook because the SSO login endpoint sets `log_source` to the backend atom `oidc`, while the new `emqx_dashboard_admin:format_username/1` only handles a `{Backend, Name}` tuple or binary.

## Environment

- EMQX fixed-build environment: 6.3.0-rc.1-g0cbc3b65
- Dashboard: `http://10.42.4.242:18083`
- Keycloak: 26.3.2, HTTPS OIDC provider
- Audit: enabled
- PR: #18696

## Steps to reproduce

1. Enable Dashboard audit.
2. Configure an OIDC SSO backend.
3. Send `POST /api/v5/sso/login/oidc` with `{"backend":"oidc"}`.
4. Inspect the EMQX log.

## Actual result

The request process exits with:

```text
{function_clause,[{emqx_dashboard_admin,format_username,[oidc],[]},
 {emqx_dashboard_audit,log_meta,3,...}]}
```

The relevant call path is:

```text
emqx_dashboard_audit:source/1
 -> emqx_dashboard_admin:format_username(oidc)
 -> function_clause
```

In the live test, the OIDC redirect and token exchange still completed, but the audit hook crashed and the login audit event may be lost. Depending on timing this can also result in an interrupted or 500 response.

## Expected result

SSO login should not crash any request process, and the login audit event should be recorded or safely skipped when only the backend identity is available.

## Suggested fix

Handle atom backend values in `format_username/1`, or only call it when the source is a complete `dashboard_username()` tuple/binary. Add a regression test for `POST /api/v5/sso/login/:backend` with audit enabled.

## Log

[...]

PR #18730, opened 2026-09-02
fix(audit): handle bare backend atom in format_username/1

zmstone:
Fixes: #18711
Introduced in: N/A (regression from #18697, merged into dev-60 on 2026-09-01, not yet released)

## Summary

Port of #18725 (dev-63) to dev-60.

On dev-63, `POST /sso/login/:backend` crashed the request process directly, because that branch's audit write path (`emqx_dashboard_audit:source/1`) calls `format_username/1` at write time. dev-60 took a different approach in its sibling PR (#18697, this branch's counterpart of #18696): it only fixed the *read* side, `emqx_audit_api:format/1`, which calls `emqx_dashboard_admin:format_username/1` when a client fetches `GET /api/v5/audit`. The write path here stores `source` unformatted.

That means on dev-60 the crash doesn't happen on the login request itself — `log_source` is set to the bare backend atom (e.g. `oidc`) at pre-authentication time and stored as-is — but the *next* `GET /api/v5/audit` call that includes that record 500s, because `format_username/1` still only handles a `{Backend, Name}` tuple or a plain binary, not a bare atom.

The fix is the same as #18725: add a clause to `format_username/1` for a bare atom, and widen its `-spec` to match.

## PR Checklist

- [x] No changelog entry — the regressed behavior was never released (#18697 merged into dev-60 the day before this was reported against its dev-63 counterpart)
- [x] Schema changes are backward compatible or intentionally breaking (N/A — no schema change)

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard_admin.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard_admin.erl
@@ -1162,11 +1162,16 @@ flatten_username(#{username := ?SSO_USERNAME(Backend, Name)} = Data) ->
 flatten_username(#{username := Username} = Data) when is_binary(Username) ->
     Data#{backend => ?BACKEND_LOCAL}.
 
--spec format_username(dashboard_username()) -> binary().
+%% `Username' is normally a `dashboard_username()', but during SSO
+%% pre-authentication (before any user identity is known) the audit hook
+%% may pass the bare backend atom instead, e.g. `oidc'.
+-spec format_username(dashboard_username() | dashboard_sso_backend()) -> binary().
 format_username(?SSO_USERNAME(Backend, Name)) ->
     iolist_to_binary([atom_to_binary(Backend), <<":">>, Name]);
 format_username(Username) when is_binary(Username) ->
-    Username.
+    Username;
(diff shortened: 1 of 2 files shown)


apps/emqx_dashboard/src/emqx_dashboard_admin.erl
```diff
@@ -1162,11 +1162,16 @@ flatten_username(#{username := ?SSO_USERNAME(Backend, Name)} = Data) ->
...
+%% pre-authentication (before any user identity is known) the audit hook
+%% may pass the bare backend atom instead, e.g. `oidc'.
+-spec format_username(dashboard_username() | dashboard_sso_backend()) -> binary().
 format_username(?SSO_USERNAME(Backend, Name)) ->
     iolist_to_binary([atom_to_binary(Backend), <<":">>, Name]);
 format_username(Username) when is_binary(Username) ->
-    Username.
+    Username;
+format_username(Backend) when is_atom(Backend) ->
+    atom_to_binary(Backend).
```
keynslug, line 1174: Going forward, probably needs a way to tell it apart from a regular username?

keynslug (approved):

Issue #18727, opened 2026-09-02
[6.3.0] TCP listener regresses with buffer=0KB and active_n=0 failed


Le9Wang:
### What happened?

Two TCP listener `tcp_options` combinations regress on EMQX 6.3.0 failed.

1. `buffer=0KB` with `active_n=1000` is accepted and shown in the effective listener configuration, but a new MQTT client cannot connect.
2. `active_n=0` with `buffer=1KB`, `4KB`, or `1MB` is accepted and shown in the effective listener configuration, but the client connects successfully. The existing listener test expects the MQTT CONNECT operation to time out for this configuration.

The scenarios succeed on EMQX 6.2.3 with the same test setup.

### Reproduction Steps

1. Create a TCP listener.
2. Update the listener through the Management API with either configuration:

   ```json
   {
     "tcp_options": {
       "active_n": 1000,
       "buffer": "0KB"
     }
   }
   ```

   or:

   ```json
   {
     "tcp_options": {
       "active_n": 0,
       "buffer": "1KB"
     }
   }
   ```

   Repeat the second configuration with `buffer` set to `4KB` and `1MB`.
3. Verify that the update succeeds and query the listener configuration.
4. Establish a new MQTT/TCP connection and publish a QoS 1 message:

   ```bash
   python -m pytest \
     regs/tests_emqx/src/tests_emqx/integration/management/listeners/test_update_parameters_listeners.py \
     --urls <EMQX_NODE_IP> \
     -k 'test_update_parameters_buffer_and_activen'
   ```

### Expected Behavior

- With `buffer=0KB` and `active_n=1000`, the listener update succeeds and new MQTT clients can connect and publish, as observed on EMQX 6.2.3.
- With `active_n=0` and `buffer=1KB`, `4KB`, or `1MB`, the MQTT CONNECT operation times out, matching the existing regression test expectation associated with EMQX-11122.

### Actual Behavior

- `buffer=0KB`, `active_n=1000`:
  - The Management API returns `200`.
  - The listener reports the applied configuration: `buffer => 0`, `active_n => 1000`.
  - A new MQTT client times out during connection establishment.
  - No MQTT connection debug log is emitted, indicating that the connection does not reach the MQTT connection handler.

- `active_n=0`, `buffer=1KB/4KB/1MB`:
  - The Management API returns `200`.
  - The listener reports `active_n => 0`.
  - The MQTT client connects successfully instead of timing out.
  - Pytest fails with:

    ```text
[...]

PR #17854, opened 2026-07-02
feat(emqx): switch TCP listener to socket backend by default

keynslug:
Release version: 6.3.0

## Summary

This PR:
1. Switches the default `tcp_backend` for TCP listeners to `socket`.
2. Updates various test suites to adopt the changed default.
3. Restructures few test suites for better coverage.

## PR Checklist

- [x] The changes are covered with new or existing tests
- [x] Change log for changes visible by users has been added to `changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible
    Note that schema defaults are changing.

PR #18734, opened 2026-09-02
fix(listener): clamp tcp_options active_n and buffer to safe ranges

zmstone:
Fixes: #18727
Release version: 6.2.4, 6.3.0, 7.0.0
Introduced in: N/A (pre-release; the underlying gap has existed since `emqx_socket_connection.erl` was added, but only became reachable once #17854 switched the TCP listener default to the `socket` backend, which has not GA'd yet)

## Summary

Two `tcp_options` combinations made a plain TCP listener unusable on the `socket` backend (default on Linux since #17854):

- `buffer=0`: maps to `socket:setopt(Sock, {otp, rcvbuf}, 0)`, and OTP's `socket` module rejects `rcvbuf=0` outright (`{error,{invalid,{socket_option,{otp,rcvbuf},0}}}`, `esock_decode_bufsz` fails on `val == 0`), so `esockd` resets the connection right after accept, before it reaches the MQTT protocol handler. The `inet` driver instead floors `buffer` at 1 byte on its own (`INET_BUFFER_MIN`), which is why `gen_tcp` never had this problem.
- `active_n=0`: on `gen_tcp` this is `{active, 0}` and the socket stays passive forever; on the `socket` backend `active_n` is only the `#deliver{}` batch size and the number of sent packets between `check_oom` calls, so 0 does something else entirely. The two backends disagreed on the same setting.

Both reproduced independently (raw TCP client, A/B against `tcp_backend=gen_tcp` on the same binary) and root-caused; full analysis, the alternatives considered, and the buffer-floor study: `/mnt/data/emqx/_out/260902-tcp-listener-active-n-buffer-rca.md` (internal, not part of this diff).

Fix: clamp both values, silently, at the single point each connection module reads them from config — no config value is rejected, so `GET` on a listener still round-trips the value exactly as configured.

- `active_n` is clamped to `[1, 1000]` (`emqx_listeners:clamp_active_n/1`, used by `emqx_connection` for `tcp`/`ssl`, `emqx_socket_connection` for the `tcp` `socket` backend, and `emqx_ws_connection` for `ws`/`wss`). The cap bounds how many packets either backend handles between OOM checks.
- `buffer` is floored to 5 bytes, the largest MQTT fixed header (1 type byte + up to 4 remaining-length bytes), in `emqx_listeners:tcp_opt/1` — the single point where `tcp_options` becomes the option list handed to `esockd` (`tcp`/`ssl`) and `ranch` (`ws`/`wss`). Nothing in OTP or the kernel requires more than 1 byte (`buffer` is a user-space read chunk size; the kernel only bounds `recbuf`/`SO_RCVBUF`, which it clamps itself), so this is a parser-motivated floor, not a driver one.

This was chosen over implementing real `gen_tcp`-style passive mode in the `socket` backend's recv loop, and over a floor in `esockd` (a separate git-dependency repo): neither `active_n=0` passive mode nor sub-5-byte read buffers have a realistic use, so clamping at the config-read site is the smaller change. One consequence: `active_n=0` on `ssl` listeners (unaffected by the original regression) also loses real passive-mode support, for consistency across listener types that share the same `tcp_options` schema struct.

## PR Checklist
- [x] Change log for changes visible by users has been added to `changes/ee/fix-18734.en.md`

https://claude.ai/code/session_01AD34QW2ZJGJMRwuxho9mxy

### DIFF
--- a/apps/emqx/src/emqx_listeners.erl
+++ b/apps/emqx/src/emqx_listeners.erl
@@ -23,7 +23,8 @@
     default_max_conn/0,
     shutdown_count/2,
     tcp_opts/1,
-    ip_port/1
+    ip_port/1,
+    clamp_active_n/1
 ]).
 
 -export([
@@ -86,6 +87,22 @@
 
 -define(ESOCKD_LISTENER(T), (T == tcp orelse T == ssl)).
 -define(COWBOY_LISTENER(T), (T == ws orelse T == wss)).
+
+%% `tcp_options.active_n' and `tcp_options.buffer' are clamped at the point of
+%% use. Out-of-range values are not rejected, so existing configs keep working.
+%%
+%% `active_n': `emqx_connection' passes it as `{active, N}', and N = 0 leaves
+%% the socket passive forever. `emqx_socket_connection' uses it only as the
+%% `#deliver{}' batch size and as the number of sent packets between OOM checks.
+%% The cap bounds how many packets either backend handles between OOM checks.
+-define(MIN_ACTIVE_N, 1).
+-define(MAX_ACTIVE_N, 1000).
+%% `buffer': the inet driver floors it at 1 byte (INET_BUFFER_MIN), while the
+%% `socket' module rejects `{otp, rcvbuf}' = 0 and esockd then drops the
+%% accepted connection. Floor it at the largest MQTT fixed header (1 type byte
+%% + up to 4 remaining-length bytes) on both backends, so a single read can
+%% hold a complete fixed header.
+-define(MIN_TCP_BUFFER, 5).
 
 -define(ROOT_KEY, listeners).
 -define(CONF_KEY_PATH, [?ROOT_KEY, '?', '?']).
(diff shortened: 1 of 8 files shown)


apps/emqx/src/emqx_listeners.erl
```diff
@@ -87,6 +88,15 @@
 -define(ESOCKD_LISTENER(T), (T == tcp orelse T == ssl)).
 -define(COWBOY_LISTENER(T), (T == ws orelse T == wss)).
 
+%% `tcp_options.active_n' and `tcp_options.buffer' accept values that make the
+%% listener unusable (0 disables read gating on the `socket' backend instead of
```
keynslug: What's "read gating"? 🤔 I believe for `emqx_socket_connection` this config now only drives batching of `#deliver{}`s and frequency of `emqx_utils:check_oom/1`.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -87,6 +88,15 @@
...
+%% enabling passive mode; a `buffer' below 1KB is rejected outright by the OTP
+%% `socket' module on accept). Clamp both silently, at the point of use, rather
```
keynslug: Are we sure this is true? Again, for `emqx_socket_connection` having small-ish buffers may be a good idea for some situations, so I'd recommend against setting such "high" lower bound.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -87,6 +88,15 @@
...
+%% than rejecting the config, so existing out-of-range configs keep working.
+-define(MIN_ACTIVE_N, 1).
+-define(MAX_ACTIVE_N, 100).
```
keynslug: I think this bound also needs an explanation / motivation.

apps/emqx/src/emqx_listeners.erl
zmstone: sonet sucks.
in the future, agents will not mark pr ready for review by default

keynslug (approved):

PR #14329, opened 2024-12-03
Add peerport placeholder to authn/authz

savonarola:
Fixes [EMQX-13390](https://emqx.atlassian.net/browse/EMQX-13390)

Release version: v/e5.9.0

## Summary

Note: the most important commit is https://github.com/emqx/emqx/pull/14329/commits/f4deaf038c7df476ef601c446d1d6e10288ade02, the others are some kind of housekeeping.

## PR Checklist
Please convert it to a draft if any of the following conditions are not met. Reviewers may skip over until all the items are checked:

- [x] Added tests for the changes
- [x] Change log has been added to `changes/(ce|ee)/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] For internal contributor: there is a jira ticket to track this change
- [x] Created PR to [emqx-docs](https://github.com/emqx/emqx-docs) if documentation update is required, or link to a follow-up jira ticket: https://github.com/emqx/emqx-docs/pull/2757 
- [x] Schema changes are backward compatible

[EMQX-13390]: https://emqx.atlassian.net/browse/EMQX-13390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #18737, opened 2026-09-02
fix(auth): bind peerport and add peername placeholder in authn/authz templates

zmstone:
Fixes: N/A
Introduced in: 5.8.5

## Summary

Reported issue: `${peerport}` renders as an empty string in HTTP authn/authz request templates, and `${peername}` is not recognized at all (passed through as the literal text `${peername}`), even though the documentation lists it as a supported placeholder.

**This revives previously closed work.** `${peerport}` was fixed the same way in #16914 (release-58), #16915 (release-510), and a release-60 port referenced by #16918, all closed without merging. The closing rationale on #16914 was that binding a new template variable is feature work, not a pure bugfix, on an already-released minor: *"We can revisit this in a later minor release."* This PR is that revisit, targeting `dev-60` per current branch policy.

Two distinct problems, two commits:

### `fix(auth):` `${peerport}` — a regression, not a missing feature

`${peerport}` worked when it shipped, then silently broke:

| When | Commit | Effect |
|---|---|---|
| 2023-12-06 | 68da627b4d | added `peerport` to `ClientInfo` in `emqx_channel.erl` |
| 2024-12-03 | f4deaf038c (#14329) | added the `${peerport}` placeholder — on top of a binding that already existed, so it genuinely worked |
| 2025-01-13 | 8c9970121e | *"rely on common `peername` present in all channel impls"* removed `peerport => PeerPort` as redundant with `peername` |

The removal left the placeholder allow-listed and its render clause in place, but with nothing bound. In `emqx_auth_template.erl` the first `render_var/2` clause is `render_var(_, undefined) -> <<>>`, which matches before the `?VAR_PEERPORT` clause is reached — so an allow-listed but unbound variable renders as an empty string rather than failing. That is the reported behavior.

This also explains why CI stayed green: the same commit deleted the `peerport := 3456` assertions from `emqx_channel_SUITE.erl`, and `emqx_authn_http_SUITE`'s `?CREDENTIALS` map hardcodes `peerport => 12345` on a hand-built credential that never passes through `emqx_channel:init/2`, so the render clause kept being exercised against a fixture value.

The fix restores the binding. Regression coverage is added at the level that was missing: `t_init_binds_peerport` calls `emqx_channel:init/2` directly and asserts on the resulting `ClientInfo`, rather than on a fixture map that the channel never produced.

Since the bug first shipped in 5.8.5, this half is a regression fix on every currently maintained line, not new capability.

### `feat(auth):` `${peername}` — genuinely new

`${peername}` was never a registered template variable — no `?VAR_PEERNAME` macro existed anywhere in the repo. It is therefore classified as a disallowed placeholder and re-escaped to literal text, which is why it reaches the auth server unsubstituted instead of rendering empty. `ClientInfo` already carries the raw `peername` tuple (copied in for other authn/authz contexts), so this is purely an allow-list plus render-clause registration, reusing the same `IP:PORT` formatting the rule engine's `client.connected` event already uses (`emqx_utils:ntoa/1`).

### Scope

Both halves share the same allow-list/render machinery (`emqx_auth_template.erl`, `emqx_authn.hrl`, `emqx_authz.hrl`, `emqx_authz_http.erl`), so HTTP authn and authz are fixed together. Authn has no HTTP-specific allow-list — it parses templates through `emqx_authn_utils`, which passes `?AUTHN_DEFAULT_ALLOWED_VARS` — so the change reaches every authn backend, and likewise `?AUTHZ_DEFAULT_ALLOWED_VARS` on the authz side.

Verified manually against a local build: rendering `${peerhost}`/`${peername}`/`${peerport}` for a client at `192.168.65.1:51544` now produces `peerhost="192.168.65.1"`, `peername="192.168.65.1:51544"`, `peerport="51544"` — the exact values from the original report.

**Out of scope, deliberately:**

- The client-info authenticator (`emqx_auth_cinfo`) uses a separate Variform-expression binding mechanism that exposes only `peerhost`. It shares no code with the template renderer touched here.
- LDAP authz (`emqx_authz_ldap.erl`) keeps its own `?ALLOWED_VARS`, which lists `peerhost` but neither `peerport` nor `peername`. It renders through the same `emqx_auth_template:render_var/2`, so adding them is a two-line change — left out as it is outside the reported HTTP scope.

## PR Checklist

https://claude.ai/code/session_01HbkZAmyxcGPxwfdVyMhBco


apps/emqx/src/emqx_channel.erl
```diff
@@ -260,6 +260,7 @@ init(
             listener => ListenerId,
             protocol => Protocol,
             peerhost => PeerHost,
+            peerport => PeerPort,
```
keynslug, line 263: I think original motivation was that this is just a dead weight: two map pairs per each client that are just copies of existing information. My rough estimation suggests that this will take maybe 24 more bytes so I guess that's fine.

When it comes to `emqx_auth_template`, one alternative is to customize [access module](https://github.com/emqx/emqx/blob/94ffdd7f7c106c2e315ce3ba01887e4ae5e17998/apps/emqx_utils/src/emqx_template.erl#L92-L93) and pick `${peerhost}` / `${peerport}` out of `peername` value. Not sure if it worth the trouble.

keynslug (approved):

apps/emqx/src/emqx_channel.erl
zmstone: right, we can try to optimize it in v7.

PR #18740, opened 2026-09-02
fix(streams):  fix stream write hook priority

savonarola:
Fixes emqx/emqx-dev-team-tasks#330
Release version: 7.0.0
Introduced in: 6.1.0

## Summary

Run Message Streams writes after Schema Validation and Message Transformation hooks.

## PR Checklist

### DIFF
--- a/apps/emqx_streams/src/emqx_streams.erl
+++ b/apps/emqx_streams/src/emqx_streams.erl
@@ -43,7 +43,9 @@ unregister_hooks() ->
 
 -spec register_stream_hooks() -> ok.
 register_stream_hooks() ->
-    _ = emqx_hooks:add('message.publish', {?MODULE, on_message_publish_stream, []}, ?HP_HIGHEST),
+    _ = emqx_hooks:add(
+        'message.publish', {?MODULE, on_message_publish_stream, []}, ?HP_RETAINER + 1
+    ),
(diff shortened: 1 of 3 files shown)


apps/emqx_streams/src/emqx_streams.erl
```diff
@@ -43,7 +43,9 @@ unregister_hooks() ->
 
 -spec register_stream_hooks() -> ok.
 register_stream_hooks() ->
-    _ = emqx_hooks:add('message.publish', {?MODULE, on_message_publish_stream, []}, ?HP_HIGHEST),
+    _ = emqx_hooks:add(
+        'message.publish', {?MODULE, on_message_publish_stream, []}, ?HP_RETAINER + 1
```
thalesmg, line 47: if the goal is to run after message transformation and schema validation, maybe it could instead be defined in those terms?

e.g.:


```suggestion
        'message.publish', {?MODULE, on_message_publish_stream, []}, ?HP_MESSAGE_TRANSFORMATION - 1
```

?

thalesmg (approved):

apps/emqx_streams/src/emqx_streams.erl
savonarola: The intention was "if a message is good enough to be stored by retainer, it is good for streams and mqs too"

PR #18767, opened 2026-09-03
fix(connector api): stop rocketmq namespace shadowing the EMQX namespace

zmstone:
Fixes:
Release version: 6.1.5, 6.2.4, 6.3.0, 7.0.0
Introduced in: 6.1.0

## Summary

A RocketMQ connector configured with a `namespace` (the Aliyun instance namespace, e.g.
`rmq-cn-fzh4bmq240a`) runs fine, but cannot be managed from the Dashboard. The Dashboard shows
"Only the administrator of namespace rmq-cn-fzh4bmq240a can perform operations on the connector",
and opening the connector returns `400 BAD_REQUEST: Managed namespace not found`.

### Cause

`emqx_connector_api:format_resource/3` merges the EMQX namespace into the raw config under the
atom key `namespace`, while the raw config uses binary keys. RocketMQ is the only connector type
with its own top-level `namespace` config field, so the response map holds both `namespace` and
`<<"namespace">>`. These are distinct Erlang terms, so nothing collapses them, but both encode to
the JSON key `"namespace"`. The response body carries the key twice:

```
"namespace":null, ... ,"namespace":"rmq-cn-fzh4bmq240a"
```

Every JSON parser keeps the last occurrence, so clients read the RocketMQ value. The Dashboard
concludes the connector belongs to a managed namespace and requests it with
`?ns=rmq-cn-fzh4bmq240a`, which does not exist.

Only the response is affected. The stored config is correct: the EMQX namespace is part of the
`emqx_config` mnesia key (`{Namespace, RootKey}`), never a field inside the value.

### Changes

`format_resource/3` removes the connector's own `namespace` field from the response, so the single
`"namespace"` key always holds the EMQX namespace.

Dropping the field alone would be unsafe. The Dashboard reads a connector, changes one field and
writes the whole body back, so an update would no longer carry the RocketMQ namespace and would
clear it. `restore_own_namespace/2`, called from the existing `deobfuscate/3` on the update path,
puts the stored value back when the request omits it or sends it empty. The value can still be
changed by sending a non-empty `namespace`.

I checked the alternative of keeping both and letting the EMQX namespace win the key. It is worse:
a read-modify-write then stores the EMQX namespace as the RocketMQ namespace, silently pointing the
connector at the wrong RocketMQ namespace.

### Scope

I expanded the schema registry on a running node and checked every registered type against all
metadata keys injected by the connector and action APIs (`namespace`, `node`, `status`,
`status_reason`, `node_status`, `actions`, `sources`, `error`, `rules`). Across 47 connector types
(including both `iotdb` union members), 44 action types and 4 source types, the RocketMQ connector's
`namespace` is the only collision. The `namespace` field in the S3Tables action sits under
`parameters`, so it is nested and unaffected.

`emqx_bridge_v2_api.erl:1624` injects metadata the same way and has the same latent problem, but no
action or source schema currently defines a colliding field, so it is left alone here.

### Follow-up

The injected `namespace` is not declared in the response schema, so it does not appear in the API
docs even though the Dashboard depends on it. The durable fix is to rename the injected key to
something that cannot collide (e.g. `managed_namespace`) at both injection sites and declare it in
the schema. That is a breaking response change needing a coordinated Dashboard update, so it is not
suitable for a patch release and is not attempted here.

### Tests

`t_own_namespace_field_does_not_shadow_emqx_namespace` asserts the response carries `"namespace"`
exactly once, that it holds the EMQX namespace, and that a read-modify-write of an unrelated field
keeps the stored RocketMQ namespace. Verified that it fails on the duplicate-key assertion without
the fix. The full `emqx_connector_api_SUITE` single group passes (31/31).

## PR Checklist

Response shape change: the RocketMQ connector's `namespace` config field is no longer returned by
`GET /connectors`. It remains stored, keeps working, and is preserved across updates. Reverting this
would require the Dashboard change described above.

https://claude.ai/code/session_01Hgjpqnpr9PKXzAXkCVhZdQ

### DIFF
--- a/apps/emqx_connector/test/emqx_connector_api_SUITE.erl
+++ b/apps/emqx_connector/test/emqx_connector_api_SUITE.erl
@@ -161,7 +161,8 @@ groups() ->
         t_actions_field,
         t_update_with_failed_validation,
         t_create_with_failed_root_validation,
-        t_create_or_update_with_unexpected_error_term
+        t_create_or_update_with_unexpected_error_term,
+        t_own_namespace_field_does_not_shadow_emqx_namespace
     ],
     ClusterOnlyTests = [
         t_inconsistent_state,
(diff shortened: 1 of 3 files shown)


apps/emqx_connector/test/emqx_connector_api_SUITE.erl
```diff
@@ -1891,3 +1892,62 @@ t_namespaced_load_on_restart(TCConfig) ->
...
     ok.
+
+-doc """
+The rocketmq connector schema defines its own top-level `namespace' config
+field, which used to collide with the EMQX namespace injected into API
+responses: both encoded to the JSON key "namespace" and clients kept the last
+one.  Checks that the response carries the key exactly once, holding the EMQX
+namespace, and that an update built from a `GET' body keeps the stored value.
+""".
+t_own_namespace_field_does_not_shadow_emqx_namespace(TCConfig) ->
```
thalesmg, line 1903: maybe this is best place in the rocketmq suite

thalesmg (approved):
I wonder if renaming the rocketmq config key to something else (e.g.: `rocketmq_namespace`) and setting namespace as an alias would also help?

zmstone:
> I wonder if renaming the rocketmq config key to something else (e.g.: `rocketmq_namespace`) and setting namespace as an alias would also help?

that's probably the way out.
will draft a followup (prob with a hocon schema alias).

apps/emqx_connector/test/emqx_connector_api_SUITE.erl
zmstone: will move it in the followup pr

Issue #18579, opened 2026-08-28
[6.3.0]Disabling the queue and stream APIs results in very slow responses and network errors.


lina50:
### What happened?

Enabling queue and stream via the Dashboard works fine, but disabling them via API returns extremely slowly and results in network errors.

### What did you expect to happen?

Disabling the API should respond normally.

### How can we reproduce it (as minimally and precisely as possible)?

1. Deploy a 3‑node cluster environment.
2. Enable queue or stream from the Dashboard.
3.Disable queue or stream from the Dashboard.
Observed result: The disable API responds extremely slowly, and network errors occur.

### EMQX version

6.3.0-beta.3-gd8a8f4e7

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Log files

<details>

</details>

lina50:
https://github.com/user-attachments/assets/cf22506f-e9a4-4fbd-83dc-efd7f03aad93

PR #18769, opened 2026-09-03
fix(mq,streams): make lifecycle config updates non-blocking

savonarola:
Fixes: #18579
Release version: 7.0.0
Introduced in: 6.3.0

## Summary

Run Message Queue and Message Stream lifecycle operations in linked workers so cluster configuration updates return without waiting for database readiness or shutdown.

## PR Checklist


apps/emqx_mq/src/emqx_mq_controller.erl
```diff
@@ -94,152 +91,152 @@ status() ->
...
     ok = emqx_mq:register_hooks(),
-
-    ok = set_status(started),
     ?tp(debug, mq_controller_start_mqs_done, #{}),
 
-    State#state{status = started}.
+    ok.
 
-do_stop_mqs(State = #state{status = Status}) ->
+do_stop_mqs() ->
```
zmstone, line 245: `do_stop_mqs/0` has lost the guard that used to skip teardown when the subsystem was not fully started:

```erlang
case Status of
    started -> ok = emqx_topic:enable_queue_alias_to_share(true), ok = emqx_mq:unregister_hooks(), ...;
    _ -> %% Not fully started: only metrics and DB were started
         ok
end
```

That guard arguably matters more now, not less. `terminate/2` calls `stop_worker/1`, which can kill a start worker mid-flight, and then calls `do_stop_mqs()` regardless of how far the start actually got. So `ok = emqx_mq:unregister_hooks()` and `ok = emqx_mq_sup:stop_gc_scheduler()` can run against a system where the matching start step never happened — and both are strict-matched, so a non-`ok` return crashes `terminate/2`.

Are those two idempotent when the corresponding start step never ran? If not, the guard needs to come back, keyed off the worker's operation rather than the removed `status` field. Same question for `emqx_streams_controller.erl`.

apps/emqx_mq/src/emqx_mq_controller.erl
```diff
@@ -94,152 +91,152 @@ status() ->
...
+    State = #state{worker = #{pid := Pid, operation := Operation}}
+) ->
+    Status = operation_result(Operation),
+    ok = set_status(Status),
+    {noreply, operation_complete(Status, State#state{worker = undefined})};
+handle_info(
+    {'EXIT', Pid, Reason},
+    State = #state{worker = #{pid := Pid, operation := Operation}}
+) ->
+    {stop, {lifecycle_operation_failed, Operation, Reason}, State#state{worker = undefined}};
```
zmstone: This turns a failed lifecycle operation into a controller crash. The supervisor restarts it, `init/1` sees `need_start() =:= true`, spawns another start worker, and it fails the same way — so a deterministic failure (DB readiness never arriving being the obvious candidate) becomes a restart loop that ends at max restart intensity and takes the parent supervisor down with it.

Before this change a start that could not complete parked in `starting` and left the rest of the node alone. Is the escalation deliberate? If it is, worth a comment saying so — the behaviour change is easy to miss and the blast radius is wider than this module.

apps/emqx_mq/src/emqx_mq_controller.erl
```diff
@@ -94,152 +91,152 @@ status() ->
...
+run_operation(start) -> do_start_mqs();
+run_operation(stop) -> do_stop_mqs().
+
+reply_waiters(Status, State = #state{waiters = Waiters}) ->
+    lists:foreach(fun(From) -> gen_server:reply(From, Status) end, Waiters),
+    State#state{waiters = []}.
+
+stop_worker(#state{worker = undefined}) ->
+    ok;
+stop_worker(#state{worker = #{pid := Pid}}) ->
```
zmstone, line 221: Unbounded `receive` on the shutdown path. The worker does not trap exits, so `exit(Pid, shutdown)` should kill it promptly and this should never actually hang — but it runs inside `terminate/2`, where being wrong costs the supervisor's shutdown timeout and ends in a brutal kill. An `after` clause is cheap insurance here.

apps/emqx_mq/src/emqx_mq_controller.erl
```diff
@@ -94,152 +91,152 @@ status() ->
...
+set_target(TargetStatus, State = #state{target_status = TargetStatus, worker = undefined}) ->
+    State;
+%% Not in desired state and no transition in progress, start transition
+set_target(TargetStatus, State = #state{worker = undefined}) ->
+    start_operation(target_operation(TargetStatus), State#state{target_status = TargetStatus});
+%% Transition in progress, update target status
+set_target(TargetStatus, State) ->
+    State#state{target_status = TargetStatus}.
+
+operation_complete(Status, State = #state{target_status = Status}) ->
```
zmstone, line 187: Minor, and possibly intended: if `set_target(stopped, ...)` arrives while a start worker is in flight, the worker still runs to completion, `operation_complete/2` calls `set_status(started)`, and only then does the stop operation begin. So anything polling `status()` — a persistent_term other code reads — can observe `started` after a stop was already requested.

Probably acceptable, but worth confirming it is deliberate rather than incidental, and perhaps noting it in the module doc since `status/0` is a public read.

apps/emqx_streams/src/emqx_streams_controller.erl
```diff
@@ -94,144 +91,143 @@ status() ->
...
-    State.
-
-status_reached(#state{status = Status, target_status = TargetStatus}) ->
-    Status =:= TargetStatus.
-
-%% Phase 1 (starting): Open DB, start metrics.
-do_start_streams(State) ->
-    ?tp(debug, streams_controller_start_streams, #{}),
-    ok = set_status(starting),
+set_target(TargetStatus, State = #state{target_status = TargetStatus, worker = undefined}) ->
```
zmstone, line 178: Nit: `emqx_mq_controller.erl` carries explanatory comments on these three `set_target/2` clauses ("Already in the desired state…", "Not in desired state and no transition in progress…", "Transition in progress…") and this one has none, though the clauses are identical.

The two modules are otherwise byte-identical after name normalisation. Since diffing them against each other is how future drift gets caught, gratuitous differences cost more here than they normally would.

apps/emqx_mq/src/emqx_mq_controller.erl
savonarola: >Before this change a start that could not complete parked in starting and left the rest of the node alone. 

The behaviour is unchanged, the change is not an escalation. Previously, errors in start just directly crashed the controller.

We may restart an op, however.

apps/emqx_mq/src/emqx_mq_controller.erl
savonarola: >Are those two idempotent 

yes

apps/emqx_mq/src/emqx_mq_controller.erl
savonarola: ✅

zmstone (approved):

PR #18771, opened 2026-09-03
ci: speed up macos package builds

zmstone:
## Summary

- split macOS package building from publish-time finalization so build jobs do not wait on Apple notarization
- notarize only official release tags such as `6.3.0` and `6.3.0-patch.1`; skip alpha, beta, rc, and milestone tags
- keep macOS smoke coverage in the package action and add GNU coreutils PATH setup for `jq` checksum validation

## Testing

- `make fmt-diff`
- `scripts/shelltest/run_tests.sh`
- `bash -n scripts/parse-git-ref.sh scripts/rel/macos-notarize-package.sh scripts/rel/macos-sign-binaries.sh`
- PyYAML parse for changed workflow/action YAML
- Dockerized `actionlint -shellcheck=` for `_push-entrypoint.yaml` and `build_packages.yaml`
- `git diff --check`

### DIFF
--- a/.github/workflows/build_packages.yaml
+++ b/.github/workflows/build_packages.yaml
@@ -13,6 +13,10 @@ on:
       publish:
         required: true
         type: boolean
+      notarize:
+        required: false
+        type: boolean
+        default: false
       otp_vsn:
         required: true
         type: string
@@ -53,6 +57,10 @@ on:
         required: true
         default: 'emqx-enterprise'
       publish:
+        required: false
+        type: boolean
+        default: false
+      notarize:
         required: false
         type: boolean
         default: false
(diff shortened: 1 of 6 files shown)


.github/workflows/build_packages.yaml
```diff
@@ -176,19 +192,31 @@ jobs:
...
         pkg_name=$(find . -mindepth 1 -maxdepth 1 -iname \*.zip)
-        mkdir emqx
-        unzip -d emqx $pkg_name > /dev/null
-        ./emqx/bin/emqx start || cat emqx/log/erlang.log.1
-        ./scripts/test/emqx-smoke-test.sh 127.0.0.1 18083
-        ./emqx/bin/emqx ping
-        ./emqx/bin/emqx help
-        ./emqx/bin/emqx ctl status
-        ./emqx/bin/emqx eval 'emqx_conf:check_dynamic_desc()'
-        ./emqx/bin/emqx stop || cat emqx/log/erlang.log.1
```
zmstone, line 191: this is deleted because test is done actions/package-macos

Issue #18747, opened 2026-09-03
[6.3.0] Creating a gateway listener with an overlong ID returns an unmapped system_limit error


Le9Wang:
### What happened?

Creating a NATS Gateway listener with a listener ID longer than 255 characters reliably fails with a low-level `system_limit` error.

The REST API appears to return `UNKNOWN_ERROR`, but that code is not declared in the endpoint's response schema, resulting in this message:

```text
Unsupported code 'UNKNOWN_ERROR', message <<"system_limit">>, schema def ['NOT_FOUND',
                                                                          'RESOURCE_NOT_FOUND',
                                                                          'BAD_REQUEST']
```

The listener ID used in this reproduction is 257 characters long (`nats:tcp:` plus 248 ASCII characters).

### Reproduction Steps

1. Start EMQX 6.3.0-rc.2-ga79302ea (Enterprise).
2. Authenticate to the Dashboard/API.
3. Send a `POST` request to `/api/v5/gateways/nats/listeners` with a listener ID longer than 255 characters:

```bash
curl -X POST 'http://<host>:18083/api/v5/gateways/nats/listeners' \
  -H 'Authorization: Bearer <token>' \
  -H 'Content-Type: application/json' \
  --data-raw '{
    "type": "tcp",
    "id": "nats:tcp:0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ",
    "name": "test-listener",
    "bind": "12345",
    "acceptors": 16,
    "max_connections": 102400,
    "max_conn_rate": 1000,
    "proxy_protocol": false,
    "proxy_protocol_timeout": "3s",
    "tcp_options": {
      "nodelay": false,
      "reuseaddr": true,
      "send_timeout_close": true,
      "active_n": 100,
      "backlog": 1024,
      "buffer": "4KB",
      "send_timeout": "15s",
      "keepalive": "none",
      "nolinger": false
    }
  }'
```

4. Observe the API response.

This is reliably reproducible.

### Expected Behavior

[...]

PR #18789, opened 2026-09-04
fix(listener): validate listener names before atom conversion

hjianbo:
Fixes: #18747
Introduced in: pre-5.8

## Summary

Reject newly created MQTT and Gateway listener names longer than 64 bytes or outside the restricted-name format before atom conversion. REST listener IDs remain binary until they are safely parsed and matched to an existing listener, preventing invalid requests from creating atoms. Existing listeners with longer names remain editable and removable.

## PR Checklist

- [x] Change log for changes visible to users has been added to changes/ee/fix-18789.en.md
- [x] Schema changes are backward compatible or intentionally breaking (new listener names are restricted to 64 bytes and the restricted-name format)

### DIFF
--- a/apps/emqx/src/emqx_listeners.erl
+++ b/apps/emqx/src/emqx_listeners.erl
@@ -38,7 +38,9 @@
 
 -export([
     listener_id/2,
+    validate_listener_name/1,
     parse_listener_id/1,
+    parse_listener_id_without_atom/1,
     esockd_access_rules/1
 ]).
 
@@ -83,6 +85,7 @@
 -define(COWBOY_LISTENER(T), (T == ws orelse T == wss)).
 
 -define(ROOT_KEY, listeners).
+-define(MAX_LISTENER_NAME_BYTES, 64).
 -define(CONF_KEY_PATH, [?ROOT_KEY, '?', '?']).
 -define(TYPES_STRING, ["tcp", "ssl", "ws", "wss", "quic"]).
 -define(MARK_DEL, ?TOMBSTONE_CONFIG_CHANGE_REQ).
@@ -576,7 +579,12 @@ resume_cowboy_listener(Id, Retries) ->
 pre_config_update([?ROOT_KEY, Type, Name], {create, NewConf}, V) when
     V =:= undefined orelse V =:= ?TOMBSTONE_VALUE
 ->
-    {ok, convert_certs(Type, Name, NewConf)};
+    case validate_listener_name(Name) of
+        ok ->
+            {ok, convert_certs(Type, Name, NewConf)};
+        {error, _} = Error ->
+            Error
+    end;
 pre_config_update([?ROOT_KEY, _Type, _Name], {create, _NewConf}, _RawConf) ->
     {error, already_exist};
 pre_config_update([?ROOT_KEY, _Type, _Name], {update, _Request}, undefined) ->
@@ -591,8 +599,13 @@ pre_config_update([?ROOT_KEY, _Type, _Name], ?MARK_DEL, _RawConf) ->
     {ok, ?TOMBSTONE_VALUE};
 pre_config_update([?ROOT_KEY], RawConf, RawConf) ->
     {ok, RawConf};
-pre_config_update([?ROOT_KEY], NewConf, _RawConf) ->
-    {ok, convert_certs(NewConf)}.
+pre_config_update([?ROOT_KEY], NewConf, OldConf) ->
+    case validate_new_listener_ids(NewConf, OldConf) of
+        ok ->
+            {ok, convert_certs(NewConf)};
+        {error, _} = Error ->
+            Error
+    end.
 
 post_config_update([?ROOT_KEY, Type, Name], {create, _Request}, NewConf, OldConf, _AppEnvs) when
     OldConf =:= undefined orelse OldConf =:= ?TOMBSTONE_TYPE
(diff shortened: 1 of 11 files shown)


zmstone:
what about the regular listeners

apps/emqx_gateway/src/emqx_gateway_utils.erl
```diff
@@ -418,6 +420,15 @@ parse_listenon(Str) when is_list(Str) ->
             end
     end.
 
+validate_listener_id(GwName, Type, LisName) ->
```
zmstone: Cover regular MQTT listener IDs too

This validator is only wired into gateway paths. Regular MQTT listener configuration still reaches `emqx_listeners:listener_id/2` -> `list_to_atom(Name)` for dynamic `listeners.<type>.<name>` keys, without an equivalent length check, so an overlong regular listener name can still raise the same `system_limit` failure during config load/update. Since this fix is meant to prevent unsafe atom conversion, please either apply the validation to regular listener creation/configuration too or document why that path is out of scope.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -890,7 +890,16 @@ validate_listener_id(Id) ->
...
+    TypeBin = listener_id_part_to_binary(Type),
+    NameBin = listener_id_part_to_binary(Name),
+    validate_listener_id(<<TypeBin/binary, ":", NameBin/binary>>).
+
+listener_id_part_to_binary(Value) when is_binary(Value) ->
+    Value;
+listener_id_part_to_binary(Value) when is_atom(Value) ->
+    atom_to_binary(Value, utf8);
+listener_id_part_to_binary(Value) when is_list(Value) ->
+    unicode:characters_to_binary(Value).
```
zmstone, line 910: should probably reject non ascii chars.

zmstone (commented):
i see no reason to support non-ascii chars in listener id.
also, while we are here, we should also validate the id to not-contain special chars, probably limit to `A-Za-z0-9_`.
Unsure about `-`.

Below chars should absolutely not be allowed:

- `:` - it's delimiter 
- `/` - probably contaminates url
- `#` - turns into a comment in hocon, and a url fragment
- `\` - escape

apps/emqx/src/emqx_listeners.erl
```diff
@@ -863,6 +876,40 @@ format_bind(Bin) when is_binary(Bin) ->
...
+        true ->
+            {error, {listener_id_too_long, ?MAX_LISTENER_ID_BYTES}}
+    end.
+
+is_ascii_listener_id(<<>>) ->
+    true;
+is_ascii_listener_id(<<Byte, Rest/binary>>) when Byte < 128 ->
+    is_ascii_listener_id(Rest);
+is_ascii_listener_id(_) ->
+    false.
```
zmstone: when it's binary, check size `<= MAX_LISTENER_ID_BYTES`
when size is not over limit, run re `^[A-Za-z0-9-_]+$`

apps/emqx/src/emqx_listeners.erl
```diff
@@ -897,11 +924,29 @@ is_ascii_listener_id(_) ->
...
+            {error, {listener_id_too_long, ?MAX_LISTENER_ID_BYTES}};
+        true ->
+            case valid_listener_id_part(TypeBin) andalso valid_listener_id_part(NameBin) of
+                true ->
+                    ok;
+                false ->
+                    case is_ascii_listener_id(IdBin) of
+                        true -> {error, listener_id_invalid_chars};
+                        false -> {error, listener_id_non_ascii}
+                    end
```
zmstone: {error, listener_id_invalid_chars}
no need to check ascii or not.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -877,18 +878,44 @@ listener_id(Type, ListenerName) ->
...
+validate_listener_id_chars(IdBin) ->
+    Parts = binary:split(IdBin, <<":">>, [global]),
+    case valid_listener_id_parts(Parts) of
+        true ->
+            ok;
+        false ->
+            case is_ascii_listener_id(IdBin) of
+                true -> {error, listener_id_invalid_chars};
+                false -> {error, listener_id_non_ascii}
+            end
```
zmstone: return {error, listener_id_invalid_chars}, no need to check ascii or not.

zmstone (commented):
there is actually emqx_utils:is_restricted_str/1 for validation.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -863,6 +877,84 @@ format_bind(Bin) when is_binary(Bin) ->
...
+    IdBin = listener_id_part_to_binary(Id),
+    case byte_size(IdBin) =< ?MAX_LISTENER_ID_BYTES of
+        false ->
+            {error, {listener_id_too_long, ?MAX_LISTENER_ID_BYTES}};
+        true ->
+            validate_listener_id_chars(IdBin)
+    end.
+
+validate_listener_id_chars(IdBin) ->
+    Parts = binary:split(IdBin, <<":">>, [global]),
```
zmstone: this part is duplicating logic in parse_listener_id/1

should abstract a new function to parse_listener_id without list_to_atom or list_to_existing_atom to serve both parsing and validation.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -863,6 +877,84 @@ format_bind(Bin) when is_binary(Bin) ->
...
+                true -> {error, listener_id_invalid_chars};
+                false -> {error, listener_id_non_ascii}
+            end
+    end.
+
+valid_listener_id_part(Part) ->
+    re:run(Part, ?LISTENER_ID_PATTERN, [{capture, none}]) =:= match.
+
+valid_listener_id_parts([Type, Name]) ->
+    lists:member(Type, [<<"tcp">>, <<"ssl">>, <<"ws">>, <<"wss">>, <<"quic">>]) andalso
```
zmstone: this part is duplicating the TYPES_STRING macro, again the new parse_listener_id (without converting to atom) should validate it.

zmstone (approved):

PR #18794, opened 2026-09-04
fix(bpapi): read BPAPI dumps from the emqx_bpapi application

zmstone:
Fixes:
Introduced in: pre-5.8

## Summary

`make static_checks` has not compared the BPAPI surface against any previous release since
2025-07-01.

`emqx_bpapi_static_checks:dumps_dir/0` joined `emqx_app_dir/0`, which returns `apps/emqx`, so the
compatibility check globbed `apps/emqx/test/emqx_static_checks_data`. That directory only ever
holds the dump the current build just wrote, so `check_compat/1` compared that dump against itself
and always passed. The checked-in per-release snapshots live in `apps/emqx_bpapi/test/
emqx_static_checks_data` and were never loaded. `7a425c0dd3` moved the module and the snapshots
into `apps/emqx_bpapi` and rewrote `emqx_app_dir/0` to hop back to `apps/emqx`;
`versions_file/0` and `project_root_dir/0` need that hop, `dumps_dir/0` does not.

The first four commits below are the fix and the findings it surfaced. Read them separately.

### `73f42b1894` — the path

`dumps_dir/0` now resolves the directory of the application this module belongs to. `emqx_app_dir/0`,
`project_root_dir/0` and `versions_file/0` are unchanged. `.gitignore` follows the generated
`master.bpapi2` to its new location, and two references to the pre-refactor README path are corrected.

### `0900638b11` — versions dropped while the check was inert

`emqx_bridge` v1 … v7, `emqx_license` v1, `emqx_mgmt_trace` v1, `emqx_ds_beamsplitter` v1 and v2,
`emqx_mgmt_api_relup` v1. All were removed deliberately in commits that have shipped, so they are
listed in `?FORCE_DELETED_MODULES` and `?FORCE_DELETED_APIS`.

`emqx_ds_beamsplitter` was already meant to be exempt, but its three entries read `beamspliiter`.
No such module ever existed.

`typecheck_apis/2` now skips `?EXPERIMENTAL_APIS` the way `check_api_immutability/2` does. Without
that the list only suppresses immutability errors, and `{emqx_ds_beamformer, 1}` still failed on
`emqx_ds:shard()` narrowing from `term()` to `binary()` in 6.0. DS subscriptions were not generally
available before 6.0.

### `eab474a356` — 24 proto versions no peer can select

A proto module is a local RPC wrapper; a peer calls the target function, never our wrapper. So the
only constraint on deleting one is that no local code still calls it.

`emqx_bpapi` announces one version per API per node and the cluster selects the lowest announced
version. The oldest release a node can roll over from is 5.8, so any version below the maximum 5.8
announced is never selected. That marks 26 of the 97 proto modules dead. 39 others have an
`introduced_in` below 5.8.0 but are still the highest version their API announces; deleting those
would break rolling upgrade from 5.8.

Two of the 26 stay, because the newer module is not equivalent:

- `emqx_proto_v1:get_metrics/1` passes a 5 second timeout and v2 passes none, so moving
  `emqx_mgmt:get_metrics/1` would let a metrics read hang on an unresponsive node.
- `emqx_mgmt_cluster_proto_v2:connected_replicants/1` uses `erpc:multicall` and returns one
  `{ok, _}` per node; v3 uses `rpc:multicall` and returns `{Results, BadNodes}`, which drops the
  per-node error detail `cluster_topology/2` reports and misaligns the following `lists:zip/2`.

The callers that pointed at a superseded module move to the current one. Every function involved is
identical there, except `emqx_eviction_agent`, where v3's `evict_session_channel` takes a will
message v2 did not; the caller passes `undefined`, so behaviour is unchanged.

The four `emqx_resource` entries in the cluster-rpc exemption list covered targets only
`emqx_resource_proto_v1` reached, and that task fails on a stale exemption, so they go too.

### `e02b8ea151` — a function added to a frozen proto module

`emqx_mgmt_data_backup_proto_v2` froze at 6.0.0. `60724f8fe9` added `peek_sensitive_table_sets/3`
to it afterwards and that shipped in 6.0.3. A node running 6.0.0 through 6.0.2 has no
`emqx_mgmt_data_backup:peek_sensitive_table_sets/1`, so during a rolling upgrade the call answers
`badrpc` and an API-key caller cannot download a backup from that node. The path fails closed, so
no archive is exposed.

The function moves to `emqx_mgmt_data_backup_proto_v4` and the caller follows it.
`emqx_mgmt_data_backup_proto_v3` already exists but appears in the 6.3 snapshot, so it is frozen on
`dev-63`; a new v4 is the only version that lands cleanly on every branch in the chain. Moving the
declaration changes nothing a user sees, so there is no changelog entry.

### `3c7d8bc789` — the 6.0 baseline was 5.10

`6.0.bpapi2` was added in `b15e6dc719` and its `api` and `signatures` maps compare equal to
`5.10.bpapi2`. It lists `emqx_bridge_proto_v1` … `v7`, which `544d229e35` deleted a month before the
6.0.0 bump, and records `emqx_ds:shard()` as `term()` when 6.0.0 already declared `binary()`. It was
dumped from a stale build tree. Regenerated from the 6.0.0 tag.

## Verification

`make static_checks` passes end to end: xref, dialyzer, the BPAPI compatibility check and the
cluster-rpc target check. The compatibility check now reports errors naming `Old release: [5, 10]`
and `[6, 0]` when it should; before this branch it compared `master` against `master`. `make fmt-diff`
is clean and elvis reports no violations. CT for `apps/emqx_eviction_agent` and `apps/emqx_modules`
passes.

## PR Checklist


thalesmg (approved):
🙈

PR #18795, opened 2026-09-04
refactor(zerobus): use new grpc_client:recv_async api

thalesmg:
Release version: 7.0.0
Introduced in: N/A

## PR Checklist


zmstone:
grpc_client PR: https://github.com/emqx/grpc-erl/pull/42

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
```diff
@@ -226,9 +224,13 @@ handle_cast(#serde_updated{serde_name = SerdeName}, State0) ->
 handle_cast(_Cast, State) ->
     {noreply, State}.
 
-handle_info({'EXIT', Helper, Res}, #{?helper := Helper} = State0) ->
-    State1 = State0#{?helper := ?undefined},
-    State = handle_helper_down(Res, State1),
+handle_info({'DOWN', Handle, _, _, Reason}, #{?recv_handle := {_, Handle}} = State0) ->
```
zmstone: [P1] Handle `DOWN` from the active receiver

After `recv_async_active/4`, `recv_handle` is a bare monitor reference (line 668), so this tuple-only match handles only the one-shot receiver used while opening. `grpc_client` reports death of its client process with a `DOWN` message even when `reply_fn` is supplied; the callback cannot run after that process itself dies. This `DOWN` therefore reaches the catch-all, leaving the writer with the dead stream and any in-flight callers unresolved until another send happens to discover the dead PID. Please handle the bare active handle by failing pending callers and reopening the stream.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
```diff
@@ -312,65 +333,66 @@ do_reply_callers_up_to(Reply, LastAckedSeq, Callers0) ->
...
-    unregister_stream(ActionResId, Idx),
     Opts = grpc_opts(State0),
     Req = create_stream_req(State0),
     maybe
         {ok, Meta} ?= grpc_meta(State0),
         {ok, Stream} ?= do_create_stream_impl(Meta, Opts),
         ok ?= grpc_send(Stream, Req, Opts),
-        Helper = grpc_recv_single_resp(Stream, Opts),
-        State0#{?helper := Helper}
+        Handle = grpc_client:async_install_receiver(Stream, #{mode => once}),
```
zmstone: [P1] Retain a deadline for the initial stream response

`async_install_receiver(Stream, #{mode => once})` is implemented as `recv` with an infinite timeout. This replaces `grpc_client:recv(Stream, Opts)`, where `Opts` carried `request_ttl`. If the endpoint accepts the stream/request but never sends the create-stream response, this handle remains installed forever, no reopen timer is scheduled, and the writer never obtains a usable stream. Please preserve the finite `request_ttl` here, for example with an explicit timeout that cancels and reopens the stream.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
```diff
@@ -682,7 +741,7 @@ ensure_stream_closed(#{?stream := ?undefined} = State0) ->
     State0;
 ensure_stream_closed(#{?stream := Stream} = State0) ->
     grpc_client:close_async(Stream),
-    State0#{?stream := ?undefined}.
+    State0#{?stream := ?undefined, ?recv_handle := ?undefined}.
```
zmstone: [P2] Demonitor retired active receive handles

The handle stored by `recv_async_active/4` is a process monitor, and the grpc client API explicitly requires active-mode callers to demonitor it when they stop receiving. This path discards the reference without doing so, retaining one monitor for every retired stream until the writer or grpc client dies. Since Zerobus closes streams periodically, those monitors grow without bound during normal operation. Please call `erlang:demonitor(Handle, [flush])` before clearing the active handle.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
zmstone, line 372: q: is it possible for grpc_client to never send a reply?

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
zmstone, line 770: here is where we should demonitor ?

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
```diff
@@ -612,46 +639,78 @@ grpc_send(Stream, Req, Opts) ->
...
+        ?n_restarts => NRestarts,
+        ?writer => self(),
+        ?stream => Stream
+    },
+    ReplyFn = {fun ?MODULE:grpc_reply_callback/3, [Ctx]},
+    Handle = grpc_client:async_install_receiver(Stream, #{mode => active, reply_fn => ReplyFn}),
+    State0#{
+        ?atomics_ref := AtomicsRef,
+        ?stream := Stream,
+        ?recv_handle := Handle,
```
zmstone, line 694: this Handle can be the reference in the DOWN ?
the handle info clause matches {_, Handle} though.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: I don't think there's any need.  The stream is already discarded at this point.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: good question.  the server _should_ reply to this request, but it could in theory never reply.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: yes, that's when we're waiting for the stream creation once-request.

but there should be another clause for the active recv mode, indeed.

thalesmg: :checkered_flag:

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
zmstone: one possibility is the ttl.
maybe start a timer here too

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
zmstone: Handle is set to ?undefined here, so there could be a DOWN in the mailbox a demonitor with [flush] option will consume it.

thalesmg: since we clear the handle here, the unknown down signal will simply be ignored since it doesn't match.

apps/emqx_bridge_zerobus/src/emqx_bridge_zerobus_stream_writer_worker.erl
thalesmg: :checkered_flag:

zmstone (approved):

PR #18799, opened 2026-09-05
feat(tls): add the node's default certificate bundle

zmstone:
Release version:
7.0.0

Introduced in:
N/A

## Summary

Following #18597. Adds
`emqx_default_cert:ensure_localhost_bundle/0`, which returns the managed-certs bundle named
`localhost` — `CN=localhost`, SANs `DNS:localhost`, `IP:127.0.0.1`, `IP:::1` — generating it if this
node does not have one, using the generation library from #18597.

**Nothing calls it yet.** The caller is the serving fallback in the next PR of the chain: a TLS
server that finds itself without a configured certificate. Generation is therefore *on demand*, not
at boot, which buys two things:

- **Deleting the bundle keeps it deleted.** An operator who configures their own certificates and
  removes this one is not fighting a node that recreates it every restart — nothing asks for it
  again. No "already generated" flag to persist, and no state to get out of step with the files.
- **The dependency points the right way.** Each TLS server calls into core when it needs a default
  certificate. Core never has to know what a gateway, dashboard or Prometheus listener config looks
  like in order to decide whether one is needed.

An operator can also supply this bundle themselves: a complete bundle already stored under this name
is used as it is and never overwritten, so seeding `localhost` is a supported way to choose the
node's default certificate.

## Details worth a reviewer's attention

**The bundle has a key and a chain, and deliberately no `ca` file.** The `ca` slot resolves to
`cacertfile`, the anchor for verifying a *peer*'s certificate. A node has no reason to trust
certificates issued by its own one-off CA, and could not be presented one anyway since that CA's
private key is discarded at generation time. The CA instead travels in the chain, after the leaf it
signed, so a client that wants to trust this node can take it from there.

**The bundle never leaves the node.** It is written through the local managed-certs API rather than
the clustered one, *and* `emqx_conf` now excludes it from the `data/certs2` copy a joining node
receives. Without that exclusion a joining node would inherit its peer's bundle — including the
private key that peer generated for itself, which is the shared-private-key problem #4636 exists to
remove, merely narrowed to one cluster. Everything else under the synced directories still travels.

**Concurrent callers are safe.** New `emqx_managed_certs:create_bundle/3` writes a bundle into a
temporary directory and moves it into place with a single rename, so a bundle is never observed
half-written and two callers cannot interleave into a mix of each other's files. It reports
`{error, exists}` rather than deciding what to do about a bundle already present; `emqx_default_cert`
takes that to mean another caller won the race and keeps theirs.

`create_bundle/3` is deliberately **not** a replacement for `add_managed_files/3`, and the natural
question — "why keep a non-atomic write at all?" — has an answer worth stating: they are different
operations. `add_managed_files/3` *merges* the given kinds into whatever the bundle already holds,
which the REST API and ACME rely on (`handle_upload_files/3` supports uploading only a `ca` while
ACME-managed `key` and `chain` stay in place); a whole-bundle write would silently delete the files
the caller did not send. That said, its file-by-file writes do mean a listener resolving certificates
*during* an upload can observe a half-updated bundle. Fixing that properly means keeping merge
semantics and adding atomicity (read existing, merge, temp dir, rename) on a v1 RPC target used
product-wide — a separate change with its own test surface, not something to smuggle into a PR that
has no callers.

## Verification

- `emqx_default_cert_SUITE` (new, 11 cases): 11/11. Generation when absent, CN and SANs, the key,
  leaf and CA belonging to each other, the chain shape and absence of a `ca` file, no CA private key
  anywhere, idempotency, an operator-supplied bundle being kept, an incomplete bundle being replaced,
  a deleted bundle staying deleted, eight concurrent callers still yielding one self-consistent
  bundle, and two clustered nodes holding different bundles.
- `emqx_conf_app_SUITE:t_default_cert_bundle_is_not_synced` (new): the bundle is left out of the
  collected data files while a user-created bundle and an `authz` file still travel.
- `emqx_conf_app_SUITE:t_copy_new_data_dir` (existing, real cluster sync): passes, so the exclusion
  did not disturb the sync it lives in.
- `make fmt-diff` and `./scripts/elvis-check.sh` clean.

## PR Checklist

No changelog: nothing calls this, so there is no behavior change to describe. The changelog lands
with the PR that flips the schema default.


apps/emqx/src/emqx_default_cert.erl
```diff
@@ -0,0 +1,110 @@
...
+                {ip, {127, 0, 0, 1}},
+                {ip, {0, 0, 0, 0, 0, 0, 0, 1}}
+            ]
+        }),
+        %% `emqx_utils_certs' speaks a plain ca/cert/key map and knows
+        %% nothing about managed certs; translate to file kinds here.
+        write_bundle(#{
+            ?FILE_KIND_KEY => Key,
+            ?FILE_KIND_CHAIN => Cert,
+            ?FILE_KIND_CA => Ca
```
zmstone: makes no sense for the serving bundle to trust itself unless the tls client uses the same key & chain - which is usually not the case, to avoid confusion, i recommend we save the ca in the full chain, and leave ca in the serving bundle empty

apps/emqx/src/emqx_default_cert.erl
```diff
@@ -0,0 +1,110 @@
...
+                reason => Reason,
+                stacktrace => Stacktrace
+            }),
+            ok
+    end.
+
+write_bundle(Files) ->
+    %% Strictly local, never the clustered `add_managed_files/3': see
+    %% the module doc for why each node keeps its own bundle.
+    case emqx_managed_certs:add_managed_files_v1(?global_ns, ?DEFAULT_CERT_BUNDLE_NAME, Files) of
```
zmstone: good to bypass cluster rpc, but there is cluster data sync before boot, it copies all version in data/certs2 , we should exclude localhost dir

apps/emqx/src/emqx_managed_certs.erl
```diff
@@ -208,6 +209,80 @@ add_managed_files(Namespace, BundleName, Files) ->
...
+        ok ?= write_files_to_dir(TmpDir, Namespace, BundleName, Files),
+        ok ?= filelib:ensure_dir(filename:join(Dir, "dummy")),
+        move_into_place(TmpDir, Dir)
+    end.
+
+tmp_dir() ->
+    Unique = binary:encode_hex(crypto:strong_rand_bytes(8)),
+    %% Sibling of the namespace directories rather than inside one: a directory
+    %% under a namespace would be listed as a bundle while it exists.
+    filename:join([emqx:data_dir(), certs2, ".tmp", Unique]).
```
hjianbo: Could we keep the temporary directory outside the directories synchronized by `emqx_conf`, or explicitly exclude `certs2/.tmp/`?

`sync_data_from_node/0` recursively collects all regular files under `data/certs2`. If it runs while `create_bundle/3` is between writing this directory and renaming it, the generated private key can be included in the archive under `certs2/.tmp/...`. If the rename happens after collection but before `zip:zip/3` reads the file, the synchronization fails with `enoent` instead.

This breaks the stated guarantee that the generated private key never leaves its node.

apps/emqx/src/emqx_default_cert.erl
```diff
@@ -0,0 +1,135 @@
...
+%%--------------------------------------------------------------------
+
+list_bundle() ->
+    emqx_managed_certs:list_managed_files(?global_ns, ?NODE_DEFAULT_CERT_BUNDLE_NAME).
+
+generate_localhost_bundle() ->
+    %% Anything already stored under this name is incomplete (`ensure_localhost_bundle/0'
+    %% checked), and would block the rename. Clear it first, locally: a bundle
+    %% this node cannot use is not worth keeping, and other nodes keep theirs.
+    _ = emqx_managed_certs:delete_bundle_v1(?global_ns, ?NODE_DEFAULT_CERT_BUNDLE_NAME),
```
hjianbo: This delete makes concurrent `ensure_localhost_bundle/0` calls unsafe.

For example:

1. Caller A observes an absent or incomplete bundle and is descheduled.
2. Caller B generates and installs a complete bundle and returns its paths.
3. Caller A resumes here and deletes B’s bundle before installing its own.

During that interval, B has returned paths which no longer exist. B may also fail in its final `list_bundle/0` if A deletes the bundle before B reaches it.

Please serialize the whole check/delete/create operation, or use an atomic conditional replacement which cannot delete a bundle installed after the initial check.

apps/emqx/src/emqx_default_cert.erl
```diff
@@ -0,0 +1,135 @@
...
+install(Files) ->
+    case emqx_managed_certs:create_bundle(?global_ns, ?NODE_DEFAULT_CERT_BUNDLE_NAME, Files) of
+        ok ->
+            ?SLOG(info, #{
+                msg => "default_tls_certificate_generated",
+                bundle => ?NODE_DEFAULT_CERT_BUNDLE_NAME,
+                dir => emqx_managed_certs:dir(?global_ns, ?NODE_DEFAULT_CERT_BUNDLE_NAME)
+            }),
+            ok;
+        {error, exists} ->
```
hjianbo, line 167: Please propagate deletion failures other than `enoent`, and do not assume every `{error, exists}` means another caller installed a complete bundle.

If an incomplete bundle exists and its deletion fails, `create_bundle/3` returns `{error, exists}`. `install/1` currently converts that to `ok`, after which `list_bundle/0` can return `{ok, IncompleteFiles}`. The API therefore reports success without the promised key-and-chain bundle.

Rechecking that the existing bundle is complete after `{error, exists}` would also make this path safer.

zmstone: removed the delete

hjianbo (approved):

PR #18806, opened 2026-09-05
perf(channel): build the limiter container on first use

zmstone:
Fixes:
Release version: 6.3.1, 7.0.0
Introduced in: 5.9.0

## Summary

An idle connection carries a fully built limiter client container in the channel state
(~2.2 KB of live heap per connection) even when no rate limit is configured. The container is
only consumed on PUBLISH and SUBSCRIBE, so connect-only clients pay for state they never
read. At 20k idle connections this is one BEAM heap-size step per connection process.

This PR builds the container on first use:

* At connect, `emqx_channel` runs only the `'channel.limiter_adjustment'` hook. A hook
  callback (multi-tenant limiters) may still install a container at that point; for everyone
  else the `quota` field stays `undefined`.
* `check_quota_exceeded/2` and `check_subscribe_quota_exceeded/2` materialize the container
  on first use via `ensure_quota/1`.
* A rate limit configured at runtime still applies: materialization happens on the consume
  path, and limiter clients re-read options from the limiter registry on every consume.
* Zone change resets the field back to the connect-time state, so the container is rebuilt
  with the new zone on next use.

Late materialization starts the exclusive buckets full at that moment, which equals the state
a connect-time bucket would have refilled to while idle, so no client gains tokens it could
not have had.

## Measured effect

10,000 idle MQTT v5 connections (`emqtt-bench conn -c 10000 -R 1000 -k 300`, connect only,
default config, socket TCP backend), single node, 6 minutes idle, after a forced full GC of
every channel process. Before = `dev-63` @ f8746e1359, after = this branch; identical
procedure and timing on the same host.

|---|---:|---:|---:|
| `#channel.quota` live size (`erts_debug:flat_size`) | 270 words | 0 (atom) | -270 w |
| Connection `#state{}` live size | 628 words | 358 words | -43% |
| Process memory, hibernated channels (99+% of them) | 5,840 B | 4,280 B | -27% |
| Process memory, all channels | 5,934 B | 4,293 B | -28% |
| `erlang:memory(processes)` at 10k connections | 130 MB | 114 MB | -16 MB |

The -16 MB per 10k connections scales to ~1.6 GB per million idle connections.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -240,6 +240,8 @@ info(alias_maximum, #channel{alias_maximum = Limits}) ->
     Limits;
 info(timers, #channel{timers = Timers}) ->
     Timers;
+info(quota, #channel{quota = Quota}) ->
+    Quota;
 info(session_state, #channel{session = Session}) ->
     Session;
 info(impl, #channel{session = Session}) ->
@@ -667,7 +669,7 @@ post_process_connect(
     }
 ) ->
     Channel = Channel0#channel{
-        quota = create_limiter(ClientInfo)
+        quota = adjust_limiter(ClientInfo)
     },
     case emqx_cm:open_session(CleanStart, ClientInfo, ConnInfo, MaybeWillMsg) of
         {ok, #{session := Session, present := false}} ->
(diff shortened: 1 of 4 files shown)


keynslug (approved):
🚀

PR #18807, opened 2026-09-05
perf(sessmem): build the delivery limiter container on first use

zmstone:
Fixes:
Release version: 6.3.1, 7.0.0
Introduced in: 6.1.2

## Summary

Every in-memory MQTT session carries a delivery limiter client container (64 words of live
heap) even when no delivery rate limit is configured. The container is only consumed on
message delivery.

This PR stores a compact `{lazy, ListenerId}` placeholder in `#session.quota` instead:

* `emqx_session_mem:create_limiter/2` runs the `'session.limiter_adjustment'` hook over the
  placeholder. A hook callback (multi-tenant limiters) may still install a real container at
  session creation; the MQTT-SN `enable_quota => false` path is unchanged.
* `try_consume_delivery_rate_limit/2` checks the limiter registry when it sees the
  placeholder: with no finite delivery limit it grants and keeps the placeholder; once a
  finite limit is configured (also at runtime, without reconnect) it builds the container via
  `emqx_limiter:create_session_client_container/1` and consumes from it. The registry read
  matches what a materialized client already performs on every consume.
* New helper `emqx_limiter:session_limits_configured/1` reports whether any session
  (delivery) limiter of a listener has a finite limit.

Related: #18806 applies the same treatment to the channel-level limiter container.

## Measured effect

10,000 idle MQTT v5 connections (`emqtt-bench conn -c 10000 -R 1000 -k 300`, connect only,
default config, socket TCP backend), single node, 6 minutes idle, after a forced full GC of
every channel process. Before = `dev-63` @ 96e1af4f78 (includes #18806), after = this branch
rebased on it; identical procedure and timing on the same host.

|---|---:|---:|---:|
| `#session.quota` live size (`erts_debug:flat_size`) | 64 words | 3 words (placeholder) | -61 w |
| Connection `#state{}` live size | 358 words | 297 words | -17% |
| Process memory, hibernated channels | 4,216 B | 3,888 B | -8% |
| Process memory, resident channels | 8,128 B | 6,171 B | -24% |

The resident saving is larger than the raw 61 words because 297 live words fit one BEAM
heap-size step lower (377 instead of 610). The whole-population mean depends on the
hibernated fraction at sample time (95% vs 79% in these runs), so the split means are the
comparable figures.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_session_mem.erl
+++ b/apps/emqx/src/emqx_session_mem.erl
@@ -219,8 +219,17 @@ create(
 create_limiter(_ClientInfo, #{enable_quota := false}) ->
     false;
 create_limiter(#{listener := ListenerId} = ClientInfo, _Conf) ->
-    Limiter = emqx_limiter:create_session_client_container(ListenerId),
-    emqx_hooks:run_fold('session.limiter_adjustment', [ClientInfo], Limiter).
+    %% A hook callback may install a limiter container here (e.g.
+    %% multi-tenant limiters). Otherwise the compact placeholder stays
+    %% and the container is built on first delivery; see
+    %% try_consume_delivery_rate_limit/2.
+    %%
+    %% The hook must run here, not at materialization: materialization
+    %% is gated by emqx_limiter:session_limits_configured/1, which only
+    %% sees listener-level limits. A hook-installed container (e.g. a
+    %% tenant with delivery limits under a listener with none) must not
+    %% depend on that gate.
+    emqx_hooks:run_fold('session.limiter_adjustment', [ClientInfo], {lazy, ListenerId}).
 
 get_mqueue_conf(Zone) ->
     #{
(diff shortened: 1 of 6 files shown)


apps/emqx/src/emqx_session_mem.erl
```diff
@@ -219,8 +219,11 @@ create(
 create_limiter(_ClientInfo, #{enable_quota := false}) ->
     false;
 create_limiter(#{listener := ListenerId} = ClientInfo, _Conf) ->
-    Limiter = emqx_limiter:create_session_client_container(ListenerId),
-    emqx_hooks:run_fold('session.limiter_adjustment', [ClientInfo], Limiter).
+    %% A hook callback may install a limiter container here (e.g.
+    %% multi-tenant limiters). Otherwise the compact placeholder stays
+    %% and the container is built on first delivery; see
+    %% try_consume_delivery_rate_limit/2.
+    emqx_hooks:run_fold('session.limiter_adjustment', [ClientInfo], {lazy, ListenerId}).
```
savonarola, line 232: Why mt limiter path is not lazy? Why didn't we move the hook call into `try_consume_delivery_rate_limit(Msg, {lazy, ListenerId})` clause?

zmstone: will make mt limiter containers lazy in follow-up pr.

savonarola (approved):

PR #18808, opened 2026-09-05
perf(limiter): keep container entries lazy until a finite limit is set

zmstone:
Fixes:
Release version: 6.3.1, 7.0.0
Introduced in: 5.9.0

## Summary

#18806 and #18807 stop idle connections from building limiter client containers. A
connection that publishes or subscribes still materializes the full container (~270 words of
live heap for the channel container) even when every limit is at its default (unlimited).

This PR makes the container entries themselves lazy:

* `emqx_limiter_client_container` entries may now be `{lazy, LimiterIds}` specs beside
  connected clients. On consume, a lazy entry checks the limiter registry: while every
  limiter of the entry is unlimited it grants and stays lazy (one persistent_term read, the
  same read a connected client performs on every consume); once a finite limit is configured
  — also at runtime, without reconnect — it connects into a real client (a composite for
  zone + listener pairs) and consumes from it.
* `emqx_limiter:create_channel_client_container/2` and
  `create_session_client_container/1` emit lazy entries instead of connecting eagerly.
* A vanished group or name is treated as unlimited, matching how connected clients fail
  open on a missing limiter.
* Multi-tenant containers (`emqx_mt_limiter`) still build connected clients; mixed
  containers work unchanged.

Also makes `emqx_limiter_client_container_SUITE` boot without binding listener ports, so it
can run beside a running broker.

With default settings an active connection's channel container drops from 270 words to 59;
combined with #18806/#18807, connect-only clients carry none at all.

## Measured effect

10,000 MQTT v5 clients that each SUBSCRIBE once and then idle (`emqtt-bench sub -c 10000
-R 1000 -k 300 -t 'bench/%i'`; SUBSCRIBE materializes the channel container), default
config, socket TCP backend, single node, 6 minutes idle, after a forced full GC of every
channel process. Before = `dev-63` @ a4f0a6304b (includes #18806 and #18807), after = this
branch on top of it; identical procedure and timing on the same host.

|---|---:|---:|---:|
| `#channel.quota` live size (`erts_debug:flat_size`) | 270 words | 59 words | -78% |
| Connection `#state{}` live size | 591 words | 380 words | -36% |
| Process memory, hibernated channels (91-99% of them) | 6,023 B | 4,887 B | -19% |

The whole-population mean depends on the hibernated fraction at sample time (99.8% vs
90.9% in these runs), so the hibernated mean is the comparable figure. Idle connect-only
clients are unaffected by this PR (they carry no container since #18806).

## PR Checklist

### DIFF
--- a/apps/emqx/test/emqx_limiter_client_container_SUITE.erl
+++ b/apps/emqx/test/emqx_limiter_client_container_SUITE.erl
@@ -18,7 +18,17 @@ all() ->
     emqx_common_test_helpers:all(?MODULE).
 
 init_per_suite(Config) ->
-    Apps = emqx_cth_suite:start([emqx], #{work_dir => emqx_cth_suite:work_dir(Config)}),
+    %% No listener is needed; not binding ports lets the suite run
+    %% beside a running broker.
+    ListenerConf =
+        "listeners.tcp.default.enable = false\n"
+        "listeners.ssl.default.enable = false\n"
+        "listeners.ws.default.enable = false\n"
+        "listeners.wss.default.enable = false",
+    Apps = emqx_cth_suite:start(
+        [{emqx, ListenerConf}],
+        #{work_dir => emqx_cth_suite:work_dir(Config)}
+    ),
     [{apps, Apps} | Config].
 
 end_per_suite(Config) ->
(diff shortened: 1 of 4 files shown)


apps/emqx/test/emqx_limiter_client_container_SUITE.erl
```diff
@@ -18,7 +18,17 @@ all() ->
     emqx_common_test_helpers:all(?MODULE).
 
 init_per_suite(Config) ->
-    Apps = emqx_cth_suite:start([emqx], #{work_dir => emqx_cth_suite:work_dir(Config)}),
+    %% No listener is needed; not binding ports lets the suite run
+    %% beside a running broker.
+    ListenerConf =
+        "listeners.tcp.default.enable = false\n"
```
savonarola, line 24: Nit: `"""..."""`

thalesmg (approved):

PR #18809, opened 2026-09-05
perf(limiter): drop copied static data from shared client state

zmstone:
Fixes:
Release version: 6.3.1, 7.0.0
Introduced in: 5.9.0

## Summary

A shared limiter client copies per-limiter constants into every holding process:
`bucket_ref` (40 words), `options` (6) and `mode` (4) beside the 6-word `limiter_id` — 76
words per client, and channels hold one shared client per zone limiter. All of it except the
identity is derivable at consume time:

* `options` is already re-read from `emqx_limiter_registry` (persistent_term) on every
  `try_consume`; the copy was only used by `put_back` and `inspect`.
* `bucket_ref` lived in the bucket registry, an ETS table.
* `mode` is arithmetic over the options.

This PR keeps only the limiter id as the client state (76 -> 6 words per shared client) and
derives the rest per call:

* `emqx_limiter_bucket_registry` keeps its role as the registry of shared limiter
  *instances* (`insert_buckets/2`, `delete_buckets/1`, `find_bucket/1`), but is backed by
  persistent_term instead of ETS — one key per group holding `Name => bucket_ref`. A consumer
  reads its bucket as a literal reference: no ETS copy, no garbage. The group registry stays a
  registry of prototypes; it is not touched.
* Per consume, `emqx_limiter_shared` reads options from the group registry and the bucket
  from the bucket registry (two persistent_term reads) and derives the mode arithmetically.
  The unlimited (default) configuration short-circuits on the options before touching the
  bucket, so it gains no cost. A vanished bucket fails open, as before.
* persistent_term writes: creating a shared group inserts two new keys (options, buckets) —
  new keys never trigger the literal-area GC, so creation stays as cheap as base. Option
  updates still reset the atomics in place without rewriting either key. Deleting a shared
  group erases two keys instead of one (one extra global GC sweep per deleted shared group);
  deletion is the rarest of the three operations.

**Measured cost** (micro-benchmark, 1M `try_consume` ops, 4 schedulers, finite limit
configured, three rounds each): one consume takes 1.57-1.60 us (regular capacity) and
1.55-1.57 us (exhausted tight limit), versus 1.55-1.63 us and 1.55-1.57 us on the base
commit — parity; the two literal reads cost the same as the old cached-state path. (An
earlier revision that resolved the bucket with `ets:lookup_element` per consume measured
+0.25-0.38 us.) With default (unlimited) settings there is no change.

Also makes `emqx_limiter_shared_SUITE` and `emqx_limiter_composite_SUITE` boot without
binding listener ports, so they can run beside a running broker.

Related: #18806, #18807, #18808 (lazy limiter containers; independent of this change).

## Measured effect

Single node, default config plus finite zone limits high enough never to throttle
(`mqtt.limiter.messages_rate = "1000000/s"`, `bytes_rate = "1GB/s"`), so the zone shared
clients materialize. Before = `dev-63` @ dbe4425ea0 (this PR's merge base, includes
#18806-#18808), after = this branch; identical procedure and timing on the same host.

**Memory** — 10,000 MQTT v5 publishers, each publishing one 32 B QoS 0 message per minute
(`emqtt-bench pub -c 10000 -R 1000 -I 60000 -s 32`), sampled 6 minutes in after a forced
full GC of every channel process (numbers from the previous revision; the client state and
the container are byte-for-byte the same in this revision):

|---|---:|---:|---:|
| `#channel.quota` live size (`erts_debug:flat_size`) | 261 words | 141 words | -46% |
| Connection `#state{}` live size | 558 words | 438 words | -22% |
| Process memory, hibernated channels | 5,881 B | 5,143 B | -13% |
| Process memory, resident channels | 14,008 B | 13,057 B | -7% |
| `erlang:memory(processes)` at 10k connections | 200 MB | 188 MB | -12 MB |

The 120-word drop is the two zone shared clients (`messages`, `bytes`) at ~62 words each.

**Throughput** — same node and zone limits, with the 10k slow publishers still connected,
plus 200 publishers at 100 msg/s each (`emqtt-bench pub -c 200 -I 10 -s 64`, QoS 0, no
subscribers). Two independent passes per side, two 60 s windows per pass; per-message cost
is beam CPU time over `messages.received`: before 143-166 us/msg at 20.8k msg/s, after
154-174 us/msg at 20.8-21.0k msg/s (previous revision). The per-pass spread on this shared
host is ~10%, larger than any difference between the sides; the publish path shows no
measurable change, consistent with the micro-benchmark.

**persistent_term writes** — 1,000 namespaces' worth of groups (one shared tenant group + one
exclusive client group each) created, updated and deleted the way emqx_mt does, with 10k idle
connections present; wall time, emulator CPU during the phase, and CPU in the following 5 s
(the asynchronous literal-area GC):

| phase (2,000 groups, 10k conns) | base `dev-63` (ETS buckets) | this PR (buckets in persistent_term) |
|---|---:|---:|
| create: wall | 78 ms | 88 ms |
| create: CPU during + after | 0.06 s + 0.5 s | 0.08 s + 0.6 s |
| update: wall | 1.8 s | 1.8 s |
| update: CPU during + after | 5.1 s + 12.7 s | 5.1 s + 12.9 s |
| delete: wall | 1.8 s | 2.7 s |
| delete: CPU during + after | 4.8 s + 13.5 s | 7.5 s + 13.6 s |
| persistent_term keys after create | 2,459 | 3,466 |

Creation and update match base: new keys are plain inserts, and updates rewrite only the
options key as before. Deletion erases the buckets key as well as the options key, hence the
extra caller-side time for 1,000 shared groups; the asynchronous sweep work is unchanged
because the two erases of a group are collected together. Namespaces without a limiter
config create no groups and no keys.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_shared.erl
+++ b/apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_shared.erl
@@ -59,12 +59,11 @@
     last_burst_time := atomic_value()
 }.
 
--type client_state() :: #{
-    limiter_id := emqx_limiter:id(),
-    bucket_ref := bucket_ref(),
-    mode := mode(),
-    options := emqx_limiter:options()
-}.
+%% The client state is just the limiter identity. The options come from
+%% the limiter registry and the bucket from the bucket registry on every
+%% use; both are persistent_term reads. This keeps the per-process copy
+%% small: channels hold a client per limiter.
+-type client_state() :: emqx_limiter:id().
 
 -type reason() :: emqx_limiter_client:reason().
 
@@ -98,7 +97,7 @@
 ]) -> ok | {error, term()}.
 create_group(Group, LimiterConfigs) when length(LimiterConfigs) > 0 ->
     Size = length(LimiterConfigs),
-    %% Factor 3 is because we have 4 atomics per limiter:
+    %% Factor 4 is because we have 4 atomics per limiter:
     %% mini_tokens
     %% last_time
     %% burst_tokens
@@ -117,7 +116,7 @@ delete_group(Group) ->
 update_group(Group, LimiterConfigs) ->
     NowUs = now_us_monotonic(),
     Buckets = [
-        emqx_limiter_bucket_registry:find_bucket({Group, Name})
+        find_bucket({Group, Name})
      || {Name, _} <- LimiterConfigs
     ],
     ok = reset_buckets(LimiterConfigs, Buckets, NowUs),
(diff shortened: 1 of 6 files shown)


apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_shared.erl
```diff
@@ -59,11 +59,12 @@
     last_burst_time := atomic_value()
 }.
 
+%% The client state carries only the limiter identity. The bucket ref,
+%% options and mode are shared per-limiter data, fetched from the bucket
+%% registry and the limiter registry on use. This keeps the per-process
+%% copy small: channels hold a client per limiter.
 -type client_state() :: #{
```
savonarola: to save more memory, why not just
```suggestion
-type client_state() :: emqx_limiter:id().
```

apps/emqx/src/emqx_limiter/src/emqx_limiter_shared/emqx_limiter_shared.erl
savonarola:
```suggestion
    %% Factor 4 is because we have 4 atomics per limiter:
```

apps/emqx/test/emqx_limiter_composite_SUITE.erl
```diff
@@ -18,7 +18,17 @@ all() ->
     emqx_common_test_helpers:all(?MODULE).
 
 init_per_suite(Config) ->
-    Apps = emqx_cth_suite:start([emqx], #{work_dir => emqx_cth_suite:work_dir(Config)}),
+    %% No listener is needed; not binding ports lets the suite run
+    %% beside a running broker.
+    ListenerConf =
+        "listeners.tcp.default.enable = false\n"
```
savonarola: Nit: """..."""

savonarola:
Although, technically, the change works, it breaks the logic & the model: limiter registry is a registry of prototypes while bucket registry is a registry of instances of a specific (shared) kind of limiter. Now this is blurred.

Also, shared limiter puts its buckets into registry which also holds e.g. local limiters for which buckets make no sense. 

I would prefer to keep bucket registry for shared limiter but just move it to PT.

Retaining registry also should fix creation regression addressed by #18865

savonarola:
If we decide not to use multiple limiter instances, we may need to refactor things more heavily to simplify the design.

Name part of limiter becomes redundant since limiters are identified by the group. So the group concept becomes redundant too -- it just becomes the limiter itself. This however requires a new concept for local limiters. There is a unique one for each channel so we cannot ignore the instance -- prototype relationship 🤔

zmstone:
> If we decide not to use multiple limiter instances, we may need to refactor things more heavily to simplify the design.
> 
> Name part of limiter becomes redundant since limiters are identified by the group. So the group concept becomes redundant too -- it just becomes the limiter itself. This however requires a new concept for local limiters. There is a unique one for each channel so we cannot ignore the instance -- prototype relationship 🤔

keeping them separated. switched from ets to persistent_term.

savonarola (approved):

PR #18810, opened 2026-09-05
fix(sso): bind OIDC and SAML login round-trips to the initiating browser

zmstone:
Fixes: emqx/emqx-dev-team-tasks#421
Release version: 6.3.2, 7.0.0
Introduced in: pre-5.8

## Summary

Dashboard SSO now binds each browser-redirect login to the browser that started it.

`emqx_dashboard_sso_browser_binding` (new) mints a random value at login start, returns it to the
browser in an `HttpOnly` cookie scoped to `Path=/api/v5/sso`, and reads it back at the callback.
Each backend sends the same value to the identity provider in the field the provider echoes back:

* OIDC (`emqx_dashboard_sso_oidc:login/2`, `emqx_dashboard_sso_oidc_api:code_callback/2`) — the
  existing `state`. `ensure_oidc_state/3` checks the cookie before `lookup_all_nodes/1`, so a
  callback that arrives without it cannot consume the state of a pending login. On mismatch the
  endpoint answers `403` with code `FORBIDDEN` and leaves the state in place.
* SAML (`emqx_dashboard_sso_saml:login/2,callback/2`) — `RelayState`, which the AuthnRequest
  carried as an empty binary until now. `callback/2` checks the cookie before the assertion reaches
  the XML parser and the replay cache.

Each login gets its own cookie, named after a hash of its value (`emqx_sso_oidc_<16 hex>`,
`emqx_sso_saml_<16 hex>`). The callback derives the name from the echoed `state` or `RelayState`,
so logins started in several tabs of one browser keep separate cookies and each callback
succeeds. A successful callback deletes only its own cookie. An abandoned login's cookie lasts
only its Max-Age: the OIDC `session_expiry` (30 s by default), or 5 minutes for SAML.

### SameSite

* OIDC: `SameSite=Lax`. Its callback is a top-level GET, which carries a Lax cookie.
* SAML: `SameSite=None` plus `Secure` when `dashboard_addr` is `https`. Its assertion consumer
  service is a cross-site top-level POST, on which browsers send no Lax cookie. Chrome's two-minute
  Lax+POST grace is an intervention, not a guarantee, so it is not relied on.
* Both add `Secure` when `dashboard_addr` is `https`.

### SAML over plain HTTP

The check stays on by default, and the login fails with a clear message. No cookie attribute makes
a browser send a cookie on a cross-site POST to an `http` address. The changelog states the
requirement, and `skip_login_cookie_check` below lets an operator opt out.

### `skip_login_cookie_check`

A new boolean on the OIDC and SAML backends, default `false`, documented as "Enable at your own
risk". When enabled, the callback does not check the cookie. The single-use, expiring OIDC state
and the SAML signature and replay checks still apply. It exists so that an operator whose browsers
cannot keep the cookie is not locked out of SSO. The login start still sets the cookie. For SAML, which keeps no login state on the server, a
login already in flight still completes after the check is turned back on. For OIDC it does not
help: any change to the OIDC backend config, this flag included, goes through `update/2` and
`destroy/2`, which stop `emqx_dashboard_sso_oidc_session` and drop every pending `state`. A login
in flight then fails whatever its cookie says. That is how every OIDC config change already
behaves, and this PR does not change it. The message of commit `943515da5d` states the in-flight
claim without this qualification. An enabled backend that skips the check
logs `sso_login_cookie_check_skipped` at warning level when it starts or its config changes.

### Behaviour changes

* The browser must open the Dashboard at `dashboard_addr`, because that is the origin the callback
  returns to. A login started at a different origin has no cookie at the callback.
* IdP-initiated SAML login no longer completes, because an unsolicited assertion carries no
  `RelayState` this server issued. It completes again with `skip_login_cookie_check` enabled.
  Making it an explicit opt-in needs `InResponseTo`, which the vendored esaml cannot express:
  `#esaml_authnreq{}` has no `id`, and `#esaml_assertion{}` has no `in_response_to`. Follow-up.

### Known limitation

Logins one after another in one browser all complete. Overlapping logins by the same user can
still collide after the callback. The one-time SSO code and the MFA pending token (TOTP enrollment
and the login challenge) each have one slot per user in the `emqx_admin` record, and a new login
overwrites it. When a second login's callback lands before the first tab has exchanged its code,
or while the first tab is still on its TOTP step, the first tab fails and the user logs in again.
This storage predates this PR and is left as is. The message of commit `943515da5d` says each tab's
login can complete; that holds for the callbacks, not for this later step.

### API

* `GET /api/v5/sso/oidc/callback` gains a `403` response with code `FORBIDDEN`.
  `POST /api/v5/sso/saml/acs` already answered `403` on a failed login and keeps doing so.
* The OIDC and SAML backend config (`GET` and `PUT` on `/api/v5/sso/oidc` and `/api/v5/sso/saml`)
  gains the optional boolean `skip_login_cookie_check`, default `false`. Existing request bodies
  stay valid.

### Dependency

`cowboy_req:reply/4` exits when the response header map carries `set-cookie`. A response may carry
several of them and a map holds one value per name, so cowboy keeps cookies in the request's
`resp_cookies` field and fills the map key itself at reply time. minirest passed the handler's
header map straight through and never touched `resp_cookies`, so a handler had no route to a
cookie at all. emqx/minirest#139 adds that translation. It is released as minirest 1.5.2, which
this PR bumps to. Without it every OIDC login start answers `500`.

### Tests

* `emqx_dashboard_sso_browser_binding_tests` — cookie names per login and backend, cookie
  attributes per backend and scheme, clearing, two logins in one cookie header, and absent or
  malformed cookie headers.
* `emqx_dashboard_sso_oidc_SUITE`:
  * `t_reject_callback_from_another_browser` — replays the callback URL with no cookie and with a
    foreign cookie, asserts `403`, asserts that no dashboard user was created and that the pending
    login survived, then completes it in the original browser.
  * `t_concurrent_logins_in_one_browser` — two logins in one cookie jar. The callback of the one
    started last succeeds first and deletes only its own cookie, then the first one's callback
    succeeds. The test does not exchange the one-time codes; see the known limitation.
  * `t_skip_login_cookie_check` — with the switch on, a callback without the cookie completes.
  * The existing flows thread the cookie through.
* `emqx_dashboard_sso_saml_SUITE`:
  * `t_callback_requires_matching_browser_cookie` — no cookie, a foreign cookie, and a missing
    `RelayState`, with signature verification off so that only the cookie check can reject.
  * `t_callback_concurrent_logins_in_one_browser` — the SAML version of the two-tab case.
  * `t_callback_skip_login_cookie_check` — the default is `false`. With the switch on, assertions
    without the cookie complete, with and without `RelayState`.
  * The Keycloak end-to-end flows thread the cookie and assert that the same assertion fails
    without it.

Full app CT in docker against dex and Keycloak: 102 cases, 0 failed
(`PROFILE=emqx-enterprise-test ./scripts/ct/run.sh --ci --app apps/emqx_dashboard_sso`).

## PR Checklist


apps/emqx_dashboard_sso/src/emqx_dashboard_sso_browser_binding.erl
```diff
@@ -0,0 +1,138 @@
...
+    end.
+
+%% Hashing first keeps the comparison constant time for values of different
+%% lengths, which `crypto:hash_equals/2' alone does not accept.
+equals(A, B) ->
+    crypto:hash_equals(crypto:hash(sha256, A), crypto:hash(sha256, B)).
+
+is_https(Url) ->
+    case uri_string:parse(Url) of
+        #{scheme := <<"https">>} -> true;
```
zmstone: [P2] Treat the HTTPS scheme case-insensitively

URI schemes are case-insensitive, and `dashboard_addr` is only typed as a binary; it is not normalized to lowercase. `uri_string:parse(<<"HTTPS://example.com">>)` preserves `<<"HTTPS">>`, so this classifies a valid HTTPS callback as insecure. For SAML that emits a non-Secure `SameSite=Lax` cookie, which the browser omits on the cross-site ACS POST, so every login fails; OIDC also silently loses the `Secure` attribute. Please case-fold the parsed scheme and cover a mixed/uppercase HTTPS address in the tests.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_browser_binding.erl
```diff
@@ -0,0 +1,138 @@
...
+%% endpoints (`/api/v5/sso/oidc/callback', `/api/v5/sso/saml/acs') live under
+%% this prefix, so the cookie never reaches any other endpoint.
+-define(COOKIE_PATH, <<"/api/v5/sso">>).
+-define(VALUE_LEN, 32).
+
+-doc "Random value to bind one login round-trip to one browser.".
+-spec new_value() -> binary().
+new_value() ->
+    emqx_utils_conv:bin(emqx_utils:gen_id(?VALUE_LEN)).
+
```
zmstone: [P2] Keep concurrent login transactions independently bound

There is one fixed cookie name per backend, so starting a second OIDC or SAML login in the same browser overwrites the value required by the first. Two Dashboard tabs (or two EMQX instances on the same host but different ports, since cookies are not port-scoped) therefore make an otherwise valid callback fail with the browser-binding error. Please key the cookie by the transaction value, or encode enough outstanding bindings to let each callback verify and clear only its own login; a two-overlapping-logins test would guard this behavior.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc.erl
```diff
@@ -130,6 +130,11 @@ fields(oidc) ->
                     desc => ?DESC(require_pkce),
                     default => false
                 })},
+            {skip_login_cookie_check,
```
zmstone, line 133: [P3] Do not promise that an OIDC login survives toggling this flag

The PR body and commit message say that continuing to set the cookie means turning this check back on does not break logins already in flight. Changing this field goes through `emqx_dashboard_sso_oidc:update/2`, however, which calls `destroy/2`; that stops `emqx_dashboard_sso_oidc_session` and drops the ETS table containing every pending `state`. Those callbacks fail even though their cookies exist. Please remove or qualify the in-flight-login claim (or preserve the OIDC session state for a flag-only update). The claim is valid for the stateless SAML path, but not for OIDC today.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_browser_binding.erl
```diff
@@ -0,0 +1,186 @@
...
+skips_check(Config) ->
+    maps:get(skip_login_cookie_check, Config, false) =:= true.
+
+name_prefix(oidc) -> <<"emqx_sso_oidc_">>;
+name_prefix(saml) -> <<"emqx_sso_saml_">>.
+
+set_cookie(Name, Value, Opts) ->
+    #{<<"set-cookie">> => iolist_to_binary(cow_cookie:setcookie(Name, Value, Opts))}.
+
+%% The OIDC callback is a top-level GET, which carries a `Lax' cookie. The SAML
```
keynslug, line 147: Q: What's "top-level" here?

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc.erl
keynslug, line 133: Q: Are situations where this option might help really worth introducing this? If so, as this downgrades security, perhaps make it hidden / undocumented?

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_browser_binding.erl
```diff
@@ -0,0 +1,186 @@
...
+        true ->
+            ?SLOG(warning, #{
+                msg => "sso_login_cookie_check_skipped",
+                backend => Backend,
+                reason => "SSO logins are not bound to the browser that started them"
+            }),
+            ok;
+        false ->
+            ok
+    end;
```
keynslug, line 130: Nit. I'd suggest moving this out to implementation, so this library-ish module is not concerned with logging.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc.erl
```diff
@@ -300,7 +308,13 @@ login(
             of
                 {ok, [Base, Delimiter, Params]} ->
                     RedirectUri = <<Base/binary, Delimiter/binary, Params/binary>>,
-                    Redirect = {302, ?REDIRECT_HEADERS(RedirectUri), ?REDIRECT_BODY},
+                    %% The cookie binds `State' to this browser. The callback
+                    %% rejects a `state' that arrives without it.
+                    Cookie = emqx_dashboard_sso_browser_binding:set_cookie_header(
+                        oidc, State, #{max_age => SessionExpiry, url => CallbackUrl}
+                    ),
+                    Headers = maps:merge(?REDIRECT_HEADERS(RedirectUri), Cookie),
```
keynslug, line 316: Nit on naming. Technically, `set_cookie_header` returns a set of headers, and with current naming it's slightly confusing the result can be `maps:merge`d.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc_api.erl
```diff
@@ -76,14 +79,28 @@ schema("/sso/oidc/callback") ->
...
+                emqx_dashboard_sso_browser_binding:clear_cookie_header(?BACKEND, State)
+            ),
+            {302, Headers, ?REDIRECT_BODY};
+        {error, browser_binding_mismatch} ->
+            %% Keep the server side state: the browser that started this login
+            %% may still complete it.
+            ?SLOG(info, #{
+                msg => "dashboard_sso_login_rejected",
+                reason => browser_binding_mismatch
+            }),
```
keynslug, line 102: Q: Why logging here? The outcome is already communicated clearly to the user behind an HTTP client.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_browser_binding.erl
zmstone: as in contrast to `` or iframe.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc.erl
zmstone: this was added after the initial draft,
because some (hard to say how many) people may block ALL cookies in the browser.

also a clarification: this issue is not that serious after all.

apps/emqx_dashboard_sso/src/emqx_dashboard_sso_oidc_api.erl
zmstone: good point.

keynslug (approved):

Issue #18797, opened 2026-09-04
EMQX sends topic alias when topic alias maximum is 0


julian-graf:
### What happened?

Introduced in 6.3.0, I am not observing this with 6.2.2.

I connect a client with an omitted topic alias maximum property because the default value is 0. EMQX then sends a topic alias in an incoming QoS 0 PUBLISH packet.

### What did you expect to happen?

The server must not send a topic alias.

> If the Topic Alias Maximum property is absent, the default value is 0.
> …
> A value of 0 indicates that the Client does not accept any Topic Aliases on this connection. If Topic Alias Maximum is absent or zero, the Server MUST NOT send any Topic Aliases to the Client [MQTT-3.1.2-27].

### How can we reproduce it (as minimally and precisely as possible)?

Let me know if you cannot reproduce this and need the exact setup.

### EMQX version

<details>

```console
$ ./bin/emqx_ctl broker
sysdescr  : EMQX Enterprise
version   : 6.3.0
datetime  : 2026-09-04T21:36:57.672688503+00:00
uptime    : 35 seconds
```

</details>

### OS version

<details>

```console
# On Linux:
$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 13 (trixie)"
NAME="Debian GNU/Linux"
VERSION_ID="13"
VERSION="13 (trixie)"
VERSION_CODENAME=trixie
DEBIAN_VERSION_FULL=13.6
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
$ uname -a
Linux 5f12025949b3 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 GNU/Linux
```

</details>

### Log files

<details>

</details>

zmstone:
Hi @julian-graf 
is this after scram auth ?

julian-graf:
Yes, SCRAM-SHA-256

zmstone:
> Yes, SCRAM-SHA-256

thanks, this helps to locate the bug.

Issue #8886, opened 2022-09-02
LWT is sent when a connection is rejected 


jsantacruz87:
### What happened?

I try to connect a client that doesn't have permissions, setting the last will message option, and althought the validation works perfectly, the broker publish the LWT message.
I think that it is a bug, because the client wasn't connected to the broker.

### What did you expect to happen?

I expect that the broker doesn't send LWT message because the client was not connected.

### How can we reproduce it (as minimally and precisely as possible)?

Try to connect to a broker with the MQTTX.app , set the LWT options ( topic:any that has a subscriber to could see the LWT that is published, payload:any, etc.. ) and set wrong client's options (username, clientid, password).
The subscriber will receive a LWT message of the client that is not connected.

### EMQX version

sysdescr  : EMQX
version   : 5.0.2
datetime  : 2022-08-29T11:37:37.570495507+00:00
uptime    : 2 days, 12 hours, 46 minutes, 34 seconds

### OS version

```console
$ cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

$ uname -a
Linux 69ee644bc5ed 5.10.104-linuxkit #1 SMP Thu Mar 17 17:08:06 UTC 2022 x86_64 GNU/Linux
```

### Log files

<details>

</details>

PR #18811, opened 2026-09-05
fix(mqtt): do not send topic alias to enhanced-auth clients

zmstone:
Fixes: #18797
Release version: 6.3.1, 7.0.0
Introduced in: 6.3.0

## Summary

A client that omits the `Topic Alias Maximum` property in CONNECT, or sets it to 0, does not
accept topic aliases:

> If Topic Alias Maximum is absent or zero, the Server MUST NOT send any Topic Aliases to the
> Client [MQTT-3.1.2-27].

EMQX 6.3.0 still sent a `Topic-Alias` property to such a client when the client authenticated
with enhanced authentication, for example SCRAM. The reporter of #18797 confirmed SCRAM is in
use. Plain authentication is not affected.

### Cause

`emqx_channel:process_connect/2` set `#channel.alias_maximum` only on the branch where
`authenticate/2` succeeds from the CONNECT packet alone. Enhanced authentication returns
`{continue, ...}` on that call and completes later in `handle_in(?AUTH_PACKET(...))`, which
reaches `post_process_connect/2` on its own and never had the connect packet to derive the
value from. Such a channel kept `alias_maximum = undefined`, and `packing_alias/2` reads
`undefined` as "no limit":

```erlang
case (Limits =:= undefined) orelse (AliasId =< maps:get(outbound, Limits, 0)) of
```

so it assigned an alias and sent it.

`alias_maximum` was set before authentication until 6.2, which is why 6.2.2 behaves correctly.
It was moved into the success branch by c5db54554b, a delivery-limiter refactor.

### Change

`apps/emqx/src/emqx_channel.erl` — set `alias_maximum` before `authenticate/2` again, so both
paths into `post_process_connect/2` get it. `will_msg` stays where it is; it must only be
stored after successful authentication (#8886).

### Tests

- `emqx_channel_SUITE:t_handle_in_extended_auth_no_topic_alias` — drives a full enhanced-auth
  exchange and asserts the channel ends with `alias_maximum` `#{outbound := 0}` and that
  `packing_alias/2` leaves the packet untouched. Fails on the parent commit.
- `emqx_mqtt_protocol_v5_SUITE:t_subscribe_no_topic_alias` — end-to-end, for both an absent
  `Topic Alias Maximum` and an explicit 0, checking the second delivery to the same topic as
  well as the first. Runs in every transport group of that suite.

## PR Checklist

https://claude.ai/code/session_01AHCNBuYeHB7xatu1hb1KCP

### DIFF
--- a/apps/emqx/src/emqx_channel.erl
+++ b/apps/emqx/src/emqx_channel.erl
@@ -641,12 +641,17 @@ process_connect(?CONNECT_PACKET(ConnPkt) = Packet, Channel0) ->
     of
         {ok, NConnPkt, Channel1 = #channel{clientinfo = ClientInfo}} ->
             ?TRACE("MQTT", "mqtt_packet_received", #{packet => Packet}),
-            case authenticate(?CONNECT_PACKET(NConnPkt), Channel1) of
+            %% NOTE: Set alias_maximum before authentication, because enhanced
+            %% authentication completes in `handle_in(?AUTH_PACKET(...))', which
+            %% reaches `post_process_connect/2' without the connect packet.
+            Channel2 = Channel1#channel{
+                alias_maximum = init_alias_maximum(NConnPkt, ClientInfo)
+            },
+            case authenticate(?CONNECT_PACKET(NConnPkt), Channel2) of
                 {ok, Properties, Channel3} ->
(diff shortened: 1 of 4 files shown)


keynslug (approved):
🫠

PR #18812, opened 2026-09-06
chore(bpapi): drop pre-5.8 baselines and detect exemptions that suppress nothing

zmstone:
Fixes:
Introduced in: N/A

## Summary

Follow-up to #18794. Three changes to what the BPAPI compatibility check covers
and to whether it can tell you when its coverage shrinks.

### Drop the baselines older than 5.8

5.8 is the earliest release a node can roll over from into v6, so comparing the
current surface against 5.5, 5.6 and 5.7 asserts compatibility with releases no
supported upgrade starts from. Nothing is lost: every API those three describe
that still matters is also in 5.8, 5.9 or 5.10.

`5.7.bpapi2` was in any case a byte-identical copy of `5.6.bpapi2` — api map and
signatures both — so it never described 5.7.0. The six APIs introduced there,
`emqx_cm` v3, `emqx_eviction_agent` v3, `emqx_metrics` v2,
`emqx_mgmt_api_plugins` v3 and `emqx_plugins` v2 among them, were missing from
it. That makes three of the nine baselines copies rather than dumps, after
`6.0.bpapi2` in #18794 and `6.1.bpapi2` in #18800.

### Remove nine force-deleted entries that suppress nothing

An entry in `?FORCE_DELETED_APIS` or `?FORCE_DELETED_MODULES` only ever
suppresses an error about an API or a module some dump still holds.

`{emqx_node_rebalance_purge, 1}` was described only by the pre-5.8 baselines and
goes with them. The other eight were already inert: `emqx_statsd`,
`{emqx_plugin_libs, 1}`, `{emqx_persistent_session, 1}`,
`{emqx_ds_shared_sub, 2}` and the four matching module entries appear in no
baseline at all.

### Fail on entries that suppress nothing

The lists had no staleness check, which is why those eight sat there. The
exemptions above were found by hand rather than by the check that reads them.

`check_no_stale_exemptions/1` collects every API key and every caller and callee
module the loaded dumps describe, and reports the entries matching neither.
`mix emqx.check_cluster_rpc` already does this for its own exemption list, so
the shape is not new to the repo.

### State the right release in a baseline, and check it

A dump is compared as the release its *file name* states: `check_compat/1`
overwrites the `release` field with `filename_to_release/1`. The field the file
carries is never read, so it drifts unnoticed — 5.8, 5.9, 5.10 and 6.1 all said
`"master"`, because the README step that replaces it at a cut was skipped.

That makes the field the one trace a copied baseline leaves. All three baselines
now known to be copies kept the source's value: `6.0.bpapi2` held 5.10,
`6.1.bpapi2` held 6.0.0, `5.7.bpapi2` held 5.6 — and all three said `"master"`.
Comparing the field against the file name would have reported every one of them
on the commit that added it.

The three drifted fields are corrected here. Only the field changes; the API and
signature data is untouched.

### Freeze the 6.0 line at the current dev-60 tip

`6.0.bpapi2` described the 6.0.0 tag, so the nine APIs the 6.0.x patches have
added since were frozen by no baseline on this branch. Eight of them are already
in `6.3.bpapi2` — frozen by a release cut downstream that dev-60 cannot see,
because baselines only travel forward. `emqx_mgmt_data_backup` v4 is newer than
that cut.

The file now describes the 6.0 *line* rather than the 6.0.0 tag, which is what
the name has to mean once a patch may add a version. Its `release` field stays
`"6.0"` to match the file name.

```
ADDED   (9)  emqx_broker_heal v1, emqx_dashboard_sso_oidc v1, emqx_gateway_cm v2,
             emqx_mgmt_api_plugins v5, emqx_mgmt_data_backup v3, emqx_mgmt_data_backup v4,
             emqx_plugins v4, emqx_plugins v5, emqx_session_tool v1
REMOVED (26) deleted proto versions already listed in ?FORCE_DELETED_APIS
CHANGED (0)
```

All twenty-six removals are still described by 5.8, 5.9 or 5.10, so no exemption
is left matching nothing and no coverage is lost. That no key *changed* is the
property that matters: removing `peek_sensitive_table_sets` from
`emqx_mgmt_data_backup_proto_v2` restored it to the shape 6.0.0 shipped.

A freeze leaves the file equal to the generated dump apart from the `release`
field. That is expected right after one, and stops being true as soon as the
next proto version lands.

## Verification

`make static_checks` passes end to end. The new check was verified to fail, not
just to pass: adding `{emqx_never_existed, 42}` and `emqx_never_existed_proto_v9`
reports both and returns a non-zero exit.

```
Stale FORCE_DELETED_APIS entries: [emqx_never_existed: 42].
Stale FORCE_DELETED_MODULES entries: [:emqx_never_existed_proto_v9].
```

The release-field check likewise: setting 5.9 back to `"master"` reports it and
fails the run.

```
5.9.bpapi2 states release "master". The file name says "5.9".
```

`make fmt-diff` is clean and elvis reports no violations. No changelog: internal
tooling with no user-visible behaviour change.

## PR Checklist


keynslug (approved):

ieQu1 (approved):

zmstone:
will fix the broken base .bpapi files in the next pr.
merge this as a cleanup for now.

Issue #18043, opened 2026-07-22
[6.1.4]  PR #17895 does not remove stale `managed_certs` when Dashboard sends `managed_certs: null`


lina50:
**What happened?

When switching a WSS listener from a managed certificate bundle back to file-based certificates, the Dashboard
  sends:

  json
  {
    "ssl_options": {
      "certfile": "${EMQX_ETC_DIR}/certs/cert.pem",
      "keyfile": "${EMQX_ETC_DIR}/certs/key.pem",
      "managed_certs": null
    }
  }

 PR #17895 checks only whether the managed_certs key exists:

  RequestSetsManaged = maps:is_key(<<"managed_certs">>, RequestSSL)

  Therefore, managed_certs: null is incorrectly treated as an active managed certificate source. Because the
  listener update uses deep_merge, the previous managed certificate bundle reference can remain in the resulting
  configuration.

### What did you expect to happen?

Update without error

## Steps to Reproduce

  1. Configure a WSS listener with a managed certificate bundle named ff.
  2. Delete the bundle, or make the bundle unavailable.
  3. In the Dashboard, switch the listener back to file-based certificates.
  4. Submit the listener configuration containing:

```
curl 'http://10.42.3.115:18083/api/v5/listeners/wss:default' \
  -X 'PUT' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Accept-Language: zh-CN,zh;q=0.9' \
  -H 'Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE3ODQ2ODg3NDkzOTIsImlzcyI6IkVNUVgifQ.zYLUd7g8wnr0B3VSXqwv7rNzke57Ys6_PQK8fKXcznc' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -b 'emqx_auth=' \
  -H 'Origin: http://10.42.3.115:18083' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://10.42.3.115:18083/' \
  -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36' \
[...]

PR #18813, opened 2026-09-06
feat(tls): serve a generated certificate when none is configured

zmstone:
Release version:
7.0.0

Introduced in:
N/A

## Summary

Finishes what #18597 and #18799 built: a TLS listener with no certificate configured now serves a
certificate the node generated for itself, instead of the example certificate every installation
shipped. This is the user-facing end of #4636.

The shipped `.pem` files stay for now. They go once the test suites stop depending on them.

## How it is wired

`emqx_tls_lib:ensure_default_certs/1` points a configuration with no certificate of its own at the
node's `localhost` bundle, generating it at that moment if the node has none. It is applied to
**checked** config at the point a server actually starts: MQTT `ssl`/`wss`/`quic`, the dashboard's
HTTPS listener, and gateway TLS/DTLS. Removing the schema defaults for `cacertfile`, `certfile` and
`keyfile` is what activates it — until now every listener had a certificate from the schema, so the
fallback never fired.

Three deliberate limits, each of which cost a bug to learn:

- **Never on raw config.** Raw config is what gets persisted and what later REST updates deep-merge
  against, so a bundle injected there would read back as a certificate the operator had configured.
- **Never on inspection.** `to_server_opts/2` stays a plain conversion. The config a running
  listener is being updated *away* from resolves through `default_certs_if_present/1`, which never
  generates: it is only compared against and rolled back to, so generating there created a key as a
  side effect of an update that did not need one.
- **`emqx_prometheus` is not wired.** Its `to_server_opts/2` call sits in `resolve_listener_certfile/3`,
  which feeds the cert-expiry metric by reading *other* listeners' certificates; applying the
  fallback there would generate a certificate as a side effect of a scrape. The cost is that a
  listener relying on the fallback reports no expiry metric.

The bundle a node generates holds a key and a chain. An operator can install their own under the
same name, and theirs may carry a `ca` file; that file is served as the `cacertfile`, so such a
listener can verify peers. A `cacertfile` the configuration names still wins, and one holding no
certificate is ignored — `ssl` refuses to start a listener whose `cacertfile` decodes to nothing.

## What removing the defaults exposed

Four bugs that the schema defaults had been masking. Each is a separate commit.

**Switching a listener off a managed bundle never worked end to end.** The update API drops the
residual `managed_certs`, then `pre_config_update/3` deep-merges against the stored config a second
time and a deep merge cannot drop keys, so the reference came back — and the request's `null` was
gone by then. A round-tripped body always carried `certfile`/`keyfile` from the defaults and so took
the file-certs branch instead, which is why the `null` clause added in #18043 had never fired. The
clear is now marked with `null`, which survives both merges and which the schema check drops.

That reconciliation is an artefact of how the *API* updates a listener, so it now lives in
`emqx_mgmt_api_listeners` rather than `emqx_listeners`, which loses ~75 lines and whose
`pre_config_update/3` just merges. A request may spell the clear as `null`, an empty string or an
empty array; all normalise to `null`. An empty string cannot be the marker: `hocon` passes it
through the schema check unchecked and unconverted, so it would reach the listener as a bundle
reference.

**A configured `cacertfile` was dropped.** Pointing `managed_certs` at the bundle replaced the whole
certificate triple, because resolving a managed bundle returns only that bundle's files. A listener
with `verify_peer` and a CA but no certificate of its own then failed to start with no trust anchor.
The fallback now names the bundle's key and chain directly and leaves the rest alone.

**The dashboard required `keyfile` and `certfile` unconditionally**, so updating the dashboard config
with no HTTPS certificate failed with `ssl_file_option_not_found`. They are required only once the
configuration names something; naming one without the other is still an error.

**A request that named no certificate could not be told from a partial update.** It is left alone,
so a client that PUTs a body without cert fields does not silently clear a bundle.

## Behaviour changes beyond the obvious

- `GET /listeners/:id` no longer returns `certfile`/`keyfile`/`cacertfile` for a listener that has
  none configured. Nothing is defaulted into the response any more.
- An HTTPS dashboard listener with no certificate is accepted rather than rejected. A certificate
  that is configured but cannot be resolved is still an error.
- `is_cert_configured/1` does not count `cacertfile`: it names the authority used to verify peers,
  not the server's own identity, so a listener setting only a `cacertfile` still gets the default
  certificate to present.
- OCSP stapling on a listener with no `certfile` is rejected with "Server certificate must be
  defined when using OCSP stapling". It was previously satisfied by the shipped default, which meant
  stapling against the certificate every installation shared.

## Test fallout

Suites that set `verify_peer` without naming a certificate, or whose clients present or verify the
shipped example certificates, now name them explicitly through
`emqx_common_test_helpers:listener_example_certs/0,1`. Most `init_per_suite` failures in a CI run
are cascade: one suite failing to start `ssl:default` leaves the node dirty and the next suites trip
`emqx_cth_suite`'s clean-state check.

Three cases assert the new behaviour instead of being restored to the old one: the dashboard serving
the node's own bundle, the OCSP validation ordering, and a config round-trip where a deep merge
cannot drop a `cacertfile` the case set earlier.

## Verification

Run locally, all green:

- `emqx_mgmt_api_certs_SUITE` 22/22, `emqx_default_cert_SUITE` 18/18, `emqx_listeners_SUITE` 33/33,
  `emqx_dashboard_https_SUITE` 11/11, `emqx_mgmt_api_listeners_SUITE` 27/27,
  `emqx_tls_auth_ext_SUITE` 4/4.
- 305 eunit tests across `emqx_tls_lib_tests`, `emqx_schema_tests`, `emqx_conf_schema_tests` and the
  new `emqx_mgmt_api_listeners_tests`.
- `emqx_default_cert_SUITE` includes an end-to-end case: a real MQTT TLS listener configured with no
  certificate accepts a TLS connection, serves `CN=localhost`, and the served DER is byte-identical
  to the leaf in this node's own bundle.
- `make static_checks`, `make fmt-diff` and `./scripts/elvis-check.sh` clean.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_listeners.erl
+++ b/apps/emqx/src/emqx_listeners.erl
@@ -326,19 +326,10 @@ start_listener(Type, Name, #{enable := false}) ->
     ),
     ok.
 
-%% `ensure_default_certs/1' throws when a listener needs the node's own
-%% certificate and it cannot be had. Report that the way every other start
-%% failure is reported, rather than letting it escape as a crash. The try covers
-%% only that call: throws from starting the listener itself keep reaching their
-%% existing handlers unchanged.
+%% A listener that needs the node's own certificate and cannot be given one does
+%% not start: the failure is reported the way every other start failure is.
 start_listener_with_certs(Type, Name, ListenerId, Conf) ->
-    case
-        try
-            {ok, ensure_default_certs(Conf)}
-        catch
-            throw:Reason -> {error, Reason}
-        end
-    of
+    case ensure_default_certs(Conf) of
         {ok, ConfWithCerts} -> do_start_listener(Type, Name, ListenerId, ConfWithCerts);
         {error, _} = Error -> Error
     end.
@@ -382,18 +373,18 @@ update_listener(Type, Name, OldConf, NewConf) ->
     end.
 
 do_update_running_listener(Type, Name, OldConf, NewConf) ->
-    case
-        do_update_listener(
-            Type, Name, default_certs_if_present(OldConf), ensure_default_certs(NewConf)
-        )
-    of
-        ok ->
-            ok = maybe_unregister_ocsp_stapling_refresh(Type, Name, NewConf),
-            ok;
-        {skip, Error} when Type =:= quic ->
-            {error, {rollbacked, Error}};
-        {error, _Reason} ->
-            restart_listener(Type, Name, OldConf, NewConf)
+    maybe
+        {ok, NewConfWithCerts} ?= ensure_default_certs(NewConf),
+        OldConfWithCerts = default_certs_if_present(OldConf),
+        case do_update_listener(Type, Name, OldConfWithCerts, NewConfWithCerts) of
+            ok ->
+                ok = maybe_unregister_ocsp_stapling_refresh(Type, Name, NewConf),
+                ok;
+            {skip, Error} when Type =:= quic ->
+                {error, {rollbacked, Error}};
+            {error, _Reason} ->
+                restart_listener(Type, Name, OldConf, NewConf)
+        end
     end.
 
 restart_listener(Type, Name, OldConf, NewConf) ->
@@ -1144,9 +1135,12 @@ Applied to both sides of an update so the comparison that decides whether the
 transport has to be torn down stays symmetric.
 """.
 ensure_default_certs(#{ssl_options := SSLOpts} = Conf) ->
-    Conf#{ssl_options => emqx_tls_lib:ensure_default_certs(SSLOpts)};
+    case emqx_tls_lib:ensure_default_certs(SSLOpts) of
+        {ok, SSLOpts1} -> {ok, Conf#{ssl_options => SSLOpts1}};
+        {error, _} = Error -> Error
+    end;
 ensure_default_certs(Conf) ->
-    Conf.
+    {ok, Conf}.
 
 %% The config a listener is being updated away from is only compared against and
 %% rolled back to, never served.  Resolving it must not generate a certificate.
--- a/apps/emqx/src/emqx_listeners.erl
+++ b/apps/emqx/src/emqx_listeners.erl
@@ -47,7 +47,6 @@
 
 -export([pre_config_update/3, post_config_update/5]).
 -export([post_zone_config_update/2, update_listener_for_zone_changes/3]).
--export([reconcile_cert_source/2]).
 
 -export([format_bind/1, format_bind_ip/1]).
 
@@ -282,7 +281,7 @@ start_listener(Type, Name, #{bind := Bind0, enable := true} = Conf0) ->
     Conf = Conf0#{bind := Bind},
     ListenerId = listener_id(Type, Name),
     ok = emqx_limiter:create_listener_limiters(ListenerId, Conf),
-    case do_start_listener(Type, Name, ListenerId, Conf) of
+    case start_listener_with_certs(Type, Name, ListenerId, Conf) of
         {ok, {skipped, Reason}} when
             Reason =:= quic_app_missing
         ->
@@ -327,6 +326,23 @@ start_listener(Type, Name, #{enable := false}) ->
     ),
     ok.
 
+%% `ensure_default_certs/1' throws when a listener needs the node's own
+%% certificate and it cannot be had. Report that the way every other start
+%% failure is reported, rather than letting it escape as a crash. The try covers
+%% only that call: throws from starting the listener itself keep reaching their
+%% existing handlers unchanged.
+start_listener_with_certs(Type, Name, ListenerId, Conf) ->
+    case
+        try
+            {ok, ensure_default_certs(Conf)}
+        catch
+            throw:Reason -> {error, Reason}
+        end
+    of
+        {ok, ConfWithCerts} -> do_start_listener(Type, Name, ListenerId, ConfWithCerts);
+        {error, _} = Error -> Error
+    end.
+
 %% @doc Restart all listeners
 -spec restart() -> ok.
 restart() ->
@@ -366,7 +382,11 @@ update_listener(Type, Name, OldConf, NewConf) ->
     end.
 
 do_update_running_listener(Type, Name, OldConf, NewConf) ->
-    case do_update_listener(Type, Name, OldConf, NewConf) of
+    case
+        do_update_listener(
+            Type, Name, default_certs_if_present(OldConf), ensure_default_certs(NewConf)
+        )
+    of
         ok ->
             ok = maybe_unregister_ocsp_stapling_refresh(Type, Name, NewConf),
             ok;
@@ -642,9 +662,8 @@ pre_config_update([?ROOT_KEY, _Type, _Name], {update, _Request}, undefined) ->
     {error, not_found};
 pre_config_update([?ROOT_KEY, Type, Name], {update, Request}, RawConf) ->
     RawConf1 = emqx_utils_maps:deep_merge(RawConf, Request),
-    RawConf2 = reconcile_cert_source(Request, RawConf1),
-    ok = assert_zone_exists(RawConf2),
-    {ok, convert_certs(Type, Name, RawConf2)};
+    ok = assert_zone_exists(RawConf1),
+    {ok, convert_certs(Type, Name, RawConf1)};
 pre_config_update([?ROOT_KEY, _Type, _Name], {action, _Action, Updated}, RawConf) ->
     {ok, emqx_utils_maps:deep_merge(RawConf, Updated)};
 pre_config_update([?ROOT_KEY, _Type, _Name], ?MARK_DEL, _RawConf) ->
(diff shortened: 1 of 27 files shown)


apps/emqx/src/emqx_tls_lib.erl
```diff
@@ -771,6 +773,61 @@ to_client_opts(Type, Opts = #{enable := true}) ->
...
+
+-doc """
+Whether this configuration names a certificate at all, by file or by managed
+bundle. Empty values do not count: the schema can leave a key present and unset.
+""".
+-spec is_cert_configured(map()) -> boolean().
+is_cert_configured(Opts) ->
+    lists:any(
+        fun(Key) -> is_configured(conf_get_opt(Key, Opts)) end,
+        [certfile, keyfile, cacertfile, managed_certs]
```
zmstone:
```suggestion
        [certfile, keyfile, managed_certs]
```

apps/emqx/src/emqx_schema.erl
```diff
@@ -2584,7 +2584,7 @@ filter(Opts) ->
 %% @private This function defines the SSL opts which are commonly used by
 %% SSL listener and client.
 -spec common_ssl_opts_schema(map(), server | client) -> hocon_schema:field_schema().
-common_ssl_opts_schema(Defaults, Type) ->
+common_ssl_opts_schema(Defaults, _Type) ->
```
zmstone: delete  _Type from the API ?

apps/emqx/src/emqx_tls_lib.erl
```diff
@@ -771,6 +774,104 @@ to_client_opts(Type, Opts = #{enable := true}) ->
...
+
+%% Names the bundle's files directly rather than pointing `managed_certs' at it.
+%% Resolving a managed bundle replaces the whole certificate triple, which would
+%% drop a `cacertfile' the configuration sets to verify peers with.
+use_default_certs(Opts, Files) ->
+    #{
+        ?FILE_KIND_CHAIN := #{path := Chain},
+        ?FILE_KIND_KEY := #{path := Key}
+    } = Files,
+    Opts#{certfile => Chain, keyfile => Key}.
```
zmstone: should use cacertfile too if it's valid.

apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -424,7 +424,8 @@ jwt_token_bearer_authorize(Req, HandlerInfo, Token) ->
             {403, 'UNAUTHORIZED_ROLE', Msg}
     end.
 
-ensure_ssl_cert(Listeners = #{https := Https0 = #{ssl_options := SslOpts}}) ->
+ensure_ssl_cert(Listeners = #{https := Https0 = #{ssl_options := SslOpts0}}) ->
+    SslOpts = emqx_tls_lib:ensure_default_certs(SslOpts0),
```
zmstone: [P1] Remove the Dashboard base-layer example certificates

`apps/emqx_conf/etc/base.hocon` still supplies `${EMQX_ETC_DIR}/certs/cert.pem` and `key.pem` under `dashboard.listeners.https.ssl_options`. That file is copied into the release, so enabling HTTPS by setting only `bind` makes `SslOpts0` already look configured. `ensure_default_certs/1` returns it unchanged and the Dashboard continues serving the shared example private key, contrary to the main security behavior of this PR. Please remove those two values from `base.hocon` as part of this change (and adjust the config test fixture that mirrors them).

apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -424,7 +424,8 @@ jwt_token_bearer_authorize(Req, HandlerInfo, Token) ->
             {403, 'UNAUTHORIZED_ROLE', Msg}
     end.
 
-ensure_ssl_cert(Listeners = #{https := Https0 = #{ssl_options := SslOpts}}) ->
+ensure_ssl_cert(Listeners = #{https := Https0 = #{ssl_options := SslOpts0}}) ->
```
zmstone, line 451: [P2] Do not generate a certificate for disabled HTTPS

`start_listeners/1` calls this before `listeners/1` filters out `#{bind := 0}`. Once the base-layer example certificate is removed, an ordinary Dashboard startup with HTTPS disabled will still create the `localhost` bundle, even when no TLS server needs it; deleting the bundle also will not survive a Dashboard restart. Please guard the disabled bind here, or move injection after filtering, to preserve the stated generate-only-when-starting behavior.

apps/emqx/src/emqx_tls_lib.erl
```diff
@@ -771,6 +774,137 @@ to_client_opts(Type, Opts = #{enable := true}) ->
...
+                    }),
+                    Opts
+            end
+    end;
+use_default_cacertfile(Opts, _Files) ->
+    Opts.
+
+holds_certificate(Path) ->
+    case file:read_file(Path) of
+        {ok, PEM} -> is_pem(PEM);
```
zmstone: [P2] Check for a certificate entry, not just arbitrary PEM

`is_pem/1` only checks that `public_key:pem_decode/1` returns a non-empty list, so a private-key PEM in the `ca` slot passes this test. Managed-certificate upload accepts arbitrary binary content, and the resulting injected `cacertfile` makes OTP fail the listener with `{options, {cacertfile, {Path, no_certs}}}`. Please use the certificate-specific validation here (for example `try_validate_pem(PEM, cacertfile, undefined) =:= ok`) and cover a non-certificate PEM in the test.

apps/emqx/src/emqx_listeners.erl
```diff
@@ -366,7 +365,11 @@ update_listener(Type, Name, OldConf, NewConf) ->
     end.
 
 do_update_running_listener(Type, Name, OldConf, NewConf) ->
-    case do_update_listener(Type, Name, OldConf, NewConf) of
+    case
+        do_update_listener(
+            Type, Name, default_certs_if_present(OldConf), ensure_default_certs(NewConf)
```
keynslug: If I'm reading this correctly, this is where "create localhost bundle for the first time" happens. This looks misplaced because the loss of "issuing failed" error condition, and this weird `default_certs_if_present/1` vs `ensure_default_certs/1` disconnect also hints at it. Ideally I think "create localhost bundle" and "use localhost bundle" operations should be used separately.

So the questions:
* What about issuing localhost bundle in `start/0` unconditionally and exit on error?
* What about issuing localhost bundle in `start/0` + `pre_config_update/3` if autogenerated certs are requested?

zmstone: this is because we want the bundle creation to be lazy.
if we issue localhost bundle in start/0, it means we'd need to parse listener configs there so to skip if no ssl listener is enabled, or without cert configred already.

zmstone: now failures to ensure default certs are surfaced as listener start error like other common error returns.
for pre_config_update, failures may cause cluster-config replication to crash and stall.

keynslug: > for pre_config_update, failures may cause cluster-config replication to crash and stall.

Not sure I follow, why? This mechanism is pretty commonly used to reject invalid or (closer to this case) inconsistent configuration update. E.g. in `emqx_cluster_link_config` for example, and in `emqx_listeners` to signal `{error, already_exist}` and `{error, not_found}`.

1. Do we need to move off this mechanism?
2. How else a cluster-wide configuration update can be safely accepted or rejected? Assuming we need to base our decision on a stable and consistent snapshot of cluster configuration.

zmstone: the current chain:
```
post_config_update([listeners, Type, Name], {update,_}, NewConf, OldConf, _)
  -> update_listener(Type, Name, #{enable := false}, #{enable := true})
  -> start_listener/3
  -> start_listener_with_certs/4
  -> ensure_default_certs/1        <- generates here
  -> do_start_listener/4
```

And post_config_upgrae error produces a critical error and move on:

https://github.com/emqx/emqx/blob/63b49afced0552d4d8a5ed729fa27f947c800106/apps/emqx/src/emqx_config_handler.erl#L374-L377

the `pre_` call should ideally never fail (idempotent).

apps/emqx/test/emqx_common_test_helpers.erl
```diff
@@ -33,6 +33,8 @@
     client_ssl/0,
     client_ssl/1,
     client_mtls/0,
+    listener_example_certs/0,
+    listener_example_certs/1,
```
keynslug, line 37: Nit. Functions have the same name but produce different shape of config.

keynslug (approved):

apps/emqx/src/emqx_listeners.erl
keynslug: Right, idempotency in `pre_config_update` is important. Makes sense, thanks for clarification.

apps/emqx/test/emqx_common_test_helpers.erl
zmstone: will be fixed in the next big bang which cleans up the test cases

PR #18822, opened 2026-09-07
fix(dsraft): record OTX leadership before schema propagation

keynslug:
Release version: 6.2.4, 6.3.1, 7.0.0
Introduced in: 6.1.1

## Summary

This PR refines DS Raft OTX leader startup.
1. When local leadership is obtained for a shard, Raft command `otx_new_leader(Pid)` is now applied first, before the local schema propagation. This helps to avoid situations when OTX leader startup advances shard timestamp before claiming leadership, causing transient `non_monotonic_timestamp` errors. This reordering appears to be safe: both operations must be appended to the log before OTX leader is announced to the cluster.
2. Raft machine now considers `not_the_leader` errors as recoverable.
3. Storage-layer RocksDB column families orphaned by incomplete schema initialization are now handled properly on re-initialization, making the whole operation idempotent. This is similar to the "incomplete creation of new generation" case.

## PR Checklist

### DIFF
--- a/apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_machine.erl
+++ b/apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_machine.erl
@@ -388,7 +388,7 @@ apply(
         _ ->
             %% Leader mismatch:
             State = State0,
-            Result = ?err_unrec({not_the_leader, #{got => From, expect => Leader}}),
+            Result = ?err_rec({not_the_leader, #{got => From, expect => Leader}}),
             Effects = []
     end,
     {State, Result, Effects};
--- a/apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
+++ b/apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
@@ -539,7 +539,7 @@ init({ShardId, Options}) ->
                 db = DB,
                 cf_refs = CFRefs
             },
-            {ok, S};
+            {ok, S, {continue, clean_orphans}};
         Schema ->
             Shard = open_shard(ShardId, DB, CFRefs, Schema),
             CurrentGenId = maps:get(current_generation, Schema),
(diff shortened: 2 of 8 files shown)


apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
```diff
@@ -557,6 +557,21 @@ init({ShardId, Options}) ->
             {ok, S, {continue, clean_orphans}}
     end.
 
+handle_continue(
```
ieQu1: This looks rather unsafe. I don't think that this should be done automatically, on every start. I'd prefer if this was exposed as an API that can be called explicitly.

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_machine.erl
```diff
@@ -388,7 +388,7 @@ apply(
         _ ->
             %% Leader mismatch:
             State = State0,
-            Result = ?err_unrec({not_the_leader, #{got => From, expect => Leader}}),
+            Result = ?err_rec({not_the_leader, #{got => From, expect => Leader}}),
```
ieQu1, line 391: While treating `not_the_leader` error as recoverable is ok for `dirty_append` (which doesn't have preconditions), it may break the transactions, since this error is reported during the *commit*. With the current `emqx_ds:trans` logic, DS will retry ONLY the commit operation, instead of restarting the whole transaction flow. This will likely lead to a commit retry, leading to another unrecoverable error, since the new leader starts with an empty OTX conflict trie, meaning it will surely reject all transactions that have reads that happened during the old leader's reign. I don't think can lead to an actual retry loop, but a lot of transactions will retry 1 extra time for no good reason.

Proper fix should look smth. like one of these options:
1. Translate this error to recoverable in `dirty_append` only
2. Adjust the transaction logic accordingly, so it restarts the whole flow from the beginning when commit returns a recoverable error. This may work badly for async transactions, so I can't say from the top of my head what the full implications are.

apps/emqx_durable_storage/src/emqx_ds_storage_layer.erl
keynslug: This is recovery procedure, similar to what happens in the next clause. Storage layer will not be able to start without this, as the attempt to initialize the schema will fail with "column family already exists".

ieQu1: Can we gracefully handle this error instead?

apps/emqx_ds_builtin_raft/src/emqx_ds_builtin_raft_machine.erl
ieQu1: P.S. Upon re-reading the code, this appears to be wrong, transaction fully restarts on any recoverable error.

keynslug: > DS will retry ONLY the commit operation, instead of restarting the whole transaction flow

Could you please explain where it happens? Both here and on `dev-63` I'm currently seeing that `trans_maybe_retry/4` retries whole `trans_inner/3` that does:
1. Instantiates new context with `new_tx(DB, Opts)`.
2. Evaluates transaction function `Ret = Fun()`.
3. Commits non-empty transaction with `commit_tx(DB, Ctx, Tx)`.

ieQu1: Yes. Please see the above comment. I was trying to remember why this error was made unrecoverable in the first place. Perhaps at the time it was introduced the `trans` function behavior was different from the present one. Please disregard this comment.

ieQu1 (approved):

PR #18828, opened 2026-09-07

zmstone:
Fixes:
Release version: 6.3.1, 7.0.0
Introduced in: pre-5.8

## Summary

A client that has not sent CONNECT could send any packet type. The frame parser
buffered the announced body, up to `mqtt.max_packet_size` (1 MB by default,
256 MB at most), before `emqx_channel` rejected the packet on the fully parsed
frame. An unauthenticated client could tie up that much memory per connection.

`emqx_frame` now carries an `expect_connect` flag in its parse options. While
the flag is set, a fixed header whose type is not CONNECT throws
`unexpected_packet_before_connect`, so the remaining length is never read and no
body byte is buffered. The check runs in both `parse/2` (stream parser) and
`parse_complete/2` (whole-frame parser). The parser clears the flag when a
CONNECT parses, so a second CONNECT on an established connection is unaffected.

`emqx_connection`, `emqx_socket_connection` and `emqx_ws_connection` set the flag
when they build the parser, and carry its value across the parser re-init that
`init_zone_specific_state/3` performs on a zone change. Without that, the
re-init would re-arm the fence on a connection that had already sent CONNECT.

`emqx_channel` needs no change: a frame error in `idle` state already shuts the
connection down under the `invalid_connect_packet` counter and sends no CONNACK.

The check lives in the parser rather than in the three connection modules. The
parser already owns the record of what it has seen on this connection, and one
copy covers both parse modes and every transport. Gateways use their own frame
modules and are out of scope.

The flag defaults to `false`, so `emqx_frame:parse/1` and the standalone
round-trip tests keep working.

Existing unit tests that fed a non-CONNECT packet into a fresh parser now send a
CONNECT first, through new `st_after_connect/0` helpers.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_connection.erl
+++ b/apps/emqx/src/emqx_connection.erl
@@ -1556,18 +1556,8 @@ init_zone_specific_state(Zone, Opts, #state{conf = Conf} = State0) ->
             undefined ->
                 init_parser_and_serializer(FrameOpts0, State0);
             Parser1 ->
-                case emqx_frame:describe_state(Parser1) of
-                    #{state := Clean, proto_ver := ProtoVer, expect_connect := ExpectConnect} when
-                        Clean == frame; Clean == clean
-                    ->
-                        FrameOpts = FrameOpts0#{
-                            version => ProtoVer, expect_connect => ExpectConnect
-                        },
-                        init_parser_and_serializer(FrameOpts, State0);
-                    _ ->
-                        %% Keep state
-                        {State0#state.parser, State0#state.serialize}
-                end
+                {ok, Parser2, Serialize2} = emqx_frame:update_opts(Parser1, FrameOpts0),
+                {Parser2, Serialize2}
         end,
     GcThresholds =
         case emqx_config:get_zone_conf(Zone, [force_gc]) of
--- a/apps/emqx/src/emqx_connection.erl
+++ b/apps/emqx/src/emqx_connection.erl
@@ -1547,7 +1547,9 @@ init_zone_specific_state(Zone, Opts, #state{conf = Conf} = State0) ->
         strict_mode => emqx_config:get_zone_conf(Zone, [mqtt, strict_mode]),
         %% N.B.: when the listener's `parse_unit = frame`, `max_packet_size` from the new
         %% zone will **not** take effect after the override.
-        max_size => emqx_config:get_zone_conf(Zone, [mqtt, max_packet_size])
+        max_size => emqx_config:get_zone_conf(Zone, [mqtt, max_packet_size]),
+        %% Any packet received before CONNECT is rejected by the parser.
+        expect_connect => true
     },
     {Parser, Serialize} =
         case State0#state.parser of
@@ -1555,8 +1557,12 @@ init_zone_specific_state(Zone, Opts, #state{conf = Conf} = State0) ->
                 init_parser_and_serializer(FrameOpts0, State0);
             Parser1 ->
                 case emqx_frame:describe_state(Parser1) of
-                    #{state := Clean, proto_ver := ProtoVer} when Clean == frame; Clean == clean ->
-                        FrameOpts = FrameOpts0#{version => ProtoVer},
+                    #{state := Clean, proto_ver := ProtoVer, expect_connect := ExpectConnect} when
+                        Clean == frame; Clean == clean
+                    ->
+                        FrameOpts = FrameOpts0#{
+                            version => ProtoVer, expect_connect => ExpectConnect
+                        },
                         init_parser_and_serializer(FrameOpts, State0);
                     _ ->
                         %% Keep state
--- a/apps/emqx/src/emqx_frame.erl
+++ b/apps/emqx/src/emqx_frame.erl
@@ -29,7 +29,7 @@
     serialize/3
 ]).
 
--export([describe_state/1]).
+-export([describe_state/1, update_opts/2]).
 
 -export_type([
     options/0,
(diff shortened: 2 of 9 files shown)


zmstone:
related: https://github.com/emqx/emqx-dev-team-tasks/issues/369

apps/emqx/src/emqx_connection.erl
```diff
@@ -1547,16 +1547,22 @@ init_zone_specific_state(Zone, Opts, #state{conf = Conf} = State0) ->
...
             Parser1 ->
                 case emqx_frame:describe_state(Parser1) of
-                    #{state := Clean, proto_ver := ProtoVer} when Clean == frame; Clean == clean ->
-                        FrameOpts = FrameOpts0#{version => ProtoVer},
+                    #{state := Clean, proto_ver := ProtoVer, expect_connect := ExpectConnect} when
+                        Clean == frame; Clean == clean
+                    ->
+                        FrameOpts = FrameOpts0#{
+                            version => ProtoVer, expect_connect => ExpectConnect
+                        },
```
keynslug: Nit. Perhaps `emqx_frame` just needs an API to update `max_size` / `strict_mode`?

keynslug (dismissed):

apps/emqx/src/emqx_connection.erl
zmstone: exported an api maybe_update_opts

apps/emqx/src/emqx_connection.erl
```diff
@@ -1547,19 +1547,20 @@ init_zone_specific_state(Zone, Opts, #state{conf = Conf} = State0) ->
...
-                case emqx_frame:describe_state(Parser1) of
-                    #{state := Clean, proto_ver := ProtoVer} when Clean == frame; Clean == clean ->
-                        FrameOpts = FrameOpts0#{version => ProtoVer},
-                        init_parser_and_serializer(FrameOpts, State0);
-                    _ ->
-                        %% Keep state
+                case emqx_frame:maybe_update_opts(Parser1, FrameOpts0) of
+                    {ok, Parser2, Serialize2} ->
+                        {Parser2, Serialize2};
+                    false ->
```
keynslug: Q: This should be unreachable? Or there are legitimate circumstances where this clause is possible? The consequence in this case is that the new zone config is silently applied only partially, right?

apps/emqx/src/emqx_frame.erl
```diff
@@ -102,6 +106,49 @@ describe_state(#body{hdr = Hdr, need = Need, acc = Acc, opts = Options}) ->
...
+version and whether a CONNECT is still expected. Callers holding a parse state
+across a config change use this instead of building a fresh state, which would
+drop that.
+
+Returns `false' while a frame is partially parsed: the options that frame is
+being read under must not change halfway through, so the caller keeps what it
+has until the frame is complete.
+""".
+-spec maybe_update_opts(parse_state(), options()) ->
+    {ok, parse_state(), serialize_opts()} | false.
```
keynslug: Nit. Perhaps?

```suggestion
-spec update_opts(parse_state(), options()) ->
    {ok, parse_state(), serialize_opts()} | {error, between_frames}.
```

keynslug (dismissed):

apps/emqx/src/emqx_connection.erl
zmstone: good point, zone change can only happen between frames (after auth).

zmstone: fix, but not in an assertive way.
in `chunk` mode, if CONNECT and PUBLISH (partial) arrived in the same chunk,
there the parser could actually be in a non-clean state.

apps/emqx/src/emqx_socket_connection.erl
```diff
@@ -1539,19 +1539,21 @@ init_zone_specific_state(Zone, Opts, #state{conf = Conf0} = State0) ->
...
-                        FrameOpts = FrameOpts0#{version => ProtoVer},
-                        init_parser_and_serializer(FrameOpts);
-                    _ ->
-                        %% Keep state
+                case emqx_frame:update_opts(Parser1, FrameOpts0) of
+                    {ok, Parser2, Serialize2} ->
+                        {Parser2, Serialize2};
+                    {error, mid_frame} ->
+                        %% Keep the frame being parsed; the new options apply
+                        %% from the next one.
```
keynslug: Q: What does "options apply from the next one" mean in this context?

apps/emqx/src/emqx_frame.erl
```diff
@@ -102,6 +106,48 @@ describe_state(#body{hdr = Hdr, need = Need, acc = Acc, opts = Options}) ->
...
+""".
+-spec update_opts(parse_state(), options()) ->
+    {ok, parse_state(), serialize_opts()} | {error, mid_frame}.
+update_opts(ParseState, NewOpts) ->
+    case between_frames(ParseState) of
+        {true, #options{version = Version}} ->
+            SerializeOpts = initial_serialize_opts(NewOpts#{version => Version}),
+            {ok, do_update_opts(ParseState, NewOpts), SerializeOpts};
+        false ->
+            {error, mid_frame}
```
keynslug: For a followup: maybe wrap in a special `#update{inner = ParseState, opts = NewOpts}` up until "state machine" transitions into `#options{}`?

keynslug (dismissed):

apps/emqx/src/emqx_socket_connection.erl
zmstone: good catch, it never applies

apps/emqx/src/emqx_frame.erl
zmstone: will fix in current

keynslug (approved):

PR #18834, opened 2026-09-08
feat(bin): switch `emqx` script to new HOCON CLI

keynslug:
Release version: 6.3.2, 7.0.0
Introduced in: N/A

## Summary

This PR switches `bin/emqx` to the new HOCON CLI, introduced in HOCON 1.0.0. See https://github.com/emqx/hocon/pull/324 for reference.

Few responsibilities moved into the `bin/emqx` as a result:
* Temporary file naming and conflict avoidance.
* Pruning (including pre-pruning) of older generations of published config files.
* Pruning of temporary file leftovers from interrupted script invocations.
* Basic rollbacks on errors.
* Fsyncing of staged files.

Verified:
1. Locally, with a set of contained test scenarios with failure injection.
2. Under `emqx/emqx:6.3.0` and a 1 MiB tmpfs, exercising `enospc`-focused scenarios.

## PR Checklist

### DIFF
--- a/bin/emqx
+++ b/bin/emqx
@@ -555,7 +555,7 @@ if [ "$IS_BOOT_COMMAND" = 'yes' ]; then
     [ -f "$EMQX_ETC_DIR"/emqx.conf ] || die "emqx.conf is not found in $EMQX_ETC_DIR" 1
     maybe_use_portable_dynlibs
     if [ "${EMQX_BOOT_CONFIGS:-}" = '' ]; then
-        EMQX_BOOT_CONFIGS="$(call_hocon -s "$SCHEMA_MOD" -c "$EMQX_ETC_DIR"/emqx.conf multi_get "${CONF_KEYS[@]}")"
+        EMQX_BOOT_CONFIGS="$(call_hocon get -s "$SCHEMA_MOD" -c "$EMQX_ETC_DIR"/emqx.conf "${CONF_KEYS[@]}")"
         ## export here so the 'console' command recursively called from
         ## 'start' command does not have to parse the configs again
         export EMQX_BOOT_CONFIGS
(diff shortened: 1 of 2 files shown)


zmstone:
if disk space is full writes to configs may results in 0 bytes files,
this was fixed here: https://github.com/emqx/hocon/pull/318
should consider the same situation here

bin/emqx
```diff
@@ -733,51 +810,69 @@ generate_config() {
...
     if [ "$DEBUG" -gt 0 ]; then set -x; fi
-    echo "$name_type $node_name" >> "$TMP_ARG_FILE"
-    echo "-mnesia dir '\"$DATA_DIR/mnesia/$NAME\"'" >> "$TMP_ARG_FILE"
-    ## rename the generated vm.<time>.args file
-    mv -f "$TMP_ARG_FILE" "$ARGS_FILE"
+    echo "$name_type $node_name" >> "$staged_vm_args_file"
+    echo "-mnesia dir '\"$DATA_DIR/mnesia/$NAME\"'" >> "$staged_vm_args_file"
+
+    ## Flush both completed outputs before publishing either filename.
+    call_nodetool fsync "$generated_app_config_file" "$staged_vm_args_file" || {
```
keynslug, line 857: On a second thought: I'm starting to doubt that this brings any perceptible benefit.

Note that each `bin/emqx start` invocation produces new pair of config artifacts. So the idea is that this protects BEAM from reading partial or empty config artifacts that can happen when all of the following happens:
1. There is no space on the storage device backing the filesystem.
2. Filesystem does not let the writer know about it during `write(2)`, e.g. write succeeds without hitting block allocation and is only persisted in the writeback bufferes. Note that this is **not** the case for the overwhelming majority of deployment scenarios, e.g. most local filesystems, AWS EBS, etc.
3. Between the moment script finishes all the pre-start work and the moment BEAM actually starts, kernel tries to flush the writeback buffers and fails due to lack of disk space. Pages are also evicted at the same time, presumably due to memory pressure.
4. BEAM reads empty or partially written config, because writes are neither on the storage device nor in the writeback buffers.

mix.exs
```diff
@@ -177,7 +177,7 @@ defmodule EMQXUmbrella.MixProject do
   def common_dep(:ekka), do: {:ekka, github: "emqx/ekka", tag: "1.0.1", override: true}
   def common_dep(:esockd), do: {:esockd, github: "emqx/esockd", tag: "5.17.4", override: true}
   def common_dep(:gproc), do: {:gproc, "1.0.0", override: true}
-  def common_dep(:hocon), do: {:hocon, github: "emqx/hocon", tag: "0.46.5", override: true}
+  def common_dep(:hocon), do: {:hocon, github: "emqx/hocon", tag: "1.0.0", override: true}
```
zmstone: 1.0.1

bin/emqx
zmstone: agree to drop

bin/emqx
```diff
@@ -720,11 +771,37 @@ generate_config() {
...
+
+    ## Remove staging files older than one hour, preserving files from recent generators.
+    ## Note on poratability: `-mmin` and `-maxdepth` are supported by GNU, macOS, and typical
+    ## BusyBox find, but are not specified by POSIX.
+    if ! find "$CONFIGS_DIR" -maxdepth 1 -type f -name '*.generated.tmp' -mmin "+30" -delete; then
+        logerr "Could not delete stale generated config files"
+    fi
+
+    ## Reserve one history slot before staging so a full filesystem can recover.
+    prune_generated_configs 2
```
zmstone, line 808: overkill to prune here if the assumption is filesystem being full will not cause 0 byte files left on disk.

keynslug: This is from https://github.com/emqx/hocon/pull/318, there the idea is actually to preemptively free up space, to increase chances that disk being nearly-full prevents new config generation.

bin/emqx
keynslug: Dropped, added pruning of empty files as a cheap countermeasure.

bin/emqx
zmstone: yes, i know.

zmstone (approved):

PR #18840, opened 2026-09-08
docs: group 5.10.5 changes by topic

zmstone:
## Summary

- add the EMQX 5.10.5 English release notes
- group enhancements and bug fixes by topic to match earlier releases
- consolidate duplicate entries while retaining all PR references

## Testing

- `./scripts/apps-version-check.sh`
- `git diff --cached --check`

### DIFF
--- a/changes/e5.10.5.en.md
+++ b/changes/e5.10.5.en.md
@@ -1,0 +1,264 @@
+# e5.10.5
+
+## Enhancements
+
+### Plugins
+
+- [#17449](https://github.com/emqx/emqx/pull/17449) Added the EMQX Backup Sync plugin to periodically synchronize selected configuration from a primary cluster to a secondary cluster by using the Data Backup APIs. The plugin supports configurable TLS options for HTTPS calls to the primary cluster.
+
+- [#17887](https://github.com/emqx/emqx/pull/17887) Added the `emqx_sync_request` plugin for synchronous MQTT request/response flows through the EMQX REST API. It also provides node-local CLI diagnostics for request counters and current pending state.
+
+### Performance
+
+- [#17964](https://github.com/emqx/emqx/pull/17964) No longer spawn buffer worker pools for authenticator and authorizer resources.  These resources always query through `simple_sync_query`, which bypasses the buffer workers, so the previously spawned workers were idle and never used.
+
+- [#18138](https://github.com/emqx/emqx/pull/18138) [#18186](https://github.com/emqx/emqx/pull/18186) Improved deep-page queries in the subscriptions HTTP API by accumulating in-memory subscription rows on each target node, avoiding one RPC per pagination batch.
+
+### Access Control
+
+- [#17864](https://github.com/emqx/emqx/pull/17864) The Dashboard user and API-key endpoints now reject scope lists that mix privilege scopes (`system`, `user_management`, `api_key_management`, `sso_management`) with other scopes. Each of the four privilege scopes is administrator-equivalent in effect, so combining them with a restricted scope list cannot meaningfully restrict the account. Use either a privilege-only scope list or a non-privilege-only scope list, depending on whether the account should have administrator-equivalent capability. Pre-existing records with a mixed scope set continue to function until the next update; the next update must split the list to succeed.
+
+### Data Integration
+
+- [#17948](https://github.com/emqx/emqx/pull/17948) Added AWS IAM role credential support to DynamoDB connectors.
+
+  When both the access key ID and secret access key are omitted, EMQX obtains temporary credentials from an ECS task role or EC2 instance metadata and refreshes them before they expire.
+
+- [#17472](https://github.com/emqx/emqx/pull/17472) Reduced the overhead of IoTDB REST API connector health checks by using a bounded version query instead of listing all databases on each check.
+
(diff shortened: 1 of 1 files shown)


changes/e5.10.5.en.md
```diff
@@ -0,0 +1,292 @@
...
+
+- [#18301](https://github.com/emqx/emqx/pull/18301) Elasticsearch action `index` and `id` values are now URL-encoded when composing the request path, so characters such as `#` or `/` in a templated value are treated as literal text within a single path segment instead of altering the request target. The JSON request body is not affected.
+
+- [#18317](https://github.com/emqx/emqx/pull/18317) When reading GCP Connectors (GCP PubSub Producer/Consumer) that use JSON Service Account authentication via the HTTP API, now the values are redacted.
+
+- [#18328](https://github.com/emqx/emqx/pull/18328) The Snowflake connector now applies its configured `ssl` options when connecting to Snowflake endpoints. Previously the connector ignored the `ssl` settings and never verified the server certificate.
+
+- [#18465](https://github.com/emqx/emqx/pull/18465) Fixed handling of templated INSERT SQL statements in the ClickHouse, TDengine, SQL Server, and MySQL bridges (when batch insert is enabled).
```
terry-xiaoyu: This entry (and a few others in this file) reads like a breaking change, and 5.10.5 has no `## Breaking Changes` section while 5.10.0–5.10.3 all have one.

For #18465 specifically: SQL templates are now fully parsed when an action is created, so templates that used to be accepted can now be rejected; comments in SQL are rejected; the MySQL bridge unconditionally disables `ANSI_QUOTES` and `NO_BACKSLASH_ESCAPES` for all connections; and the ClickHouse bridge now ignores the configured `batch_value_separator`. Users with existing templates/bridges can be affected on upgrade.

Other candidates in this file:

- #17593 — `emqx ctl relup upgrade` now refuses to run when `data/patches/` holds `*.beam` files unless `--force` is passed.
- #18823 — the entry itself says "Scripts that parse this output must be updated to match the corrected name" (`enbale` → `enable`).
- #17864 — dashboard user / API-key endpoints now reject mixed privilege scope lists; "the next update must split the list to succeed".

For comparison, e5.10.2's Breaking Changes section contains #16062, which is the same shape of change (a fix that may require users to update their templates), so the `fix-` fragment prefix alone probably shouldn't keep these out of that section.

Suggestion: add `## Breaking Changes` after `## Bug Fixes` (as in 5.10.0–5.10.3) and move these entries there, or at least link them from it so users upgrading from 5.10.4 see them.

changes/e5.10.5.en.md
```diff
@@ -0,0 +1,292 @@
...
+  authentication and authorization backend query traces, JWT signing key
+  material (carried inside `jose_jwk` records), HTTP connector request
+  headers, Prometheus configuration API responses, and authenticator
+  creation responses.
+
+- [#18835](https://github.com/emqx/emqx/pull/18835) Stopped including the result of a successful configuration change in the cluster configuration sync debug logs.
+
+  The result could carry compiled runtime state, such as the HTTP authenticator header templates, which held secrets that log redaction did not cover.
+
+- [#18336](https://github.com/emqx/emqx/pull/18336) Read-only REST endpoints no longer return secrets in cleartext:
```
terry-xiaoyu, line 111: Minor: entries are otherwise ordered by ascending PR number, but a few in the `changes/ee` group are out of place:

- Security Hardening: #18336 (here) after #18835 (line 96)
- Access Control: #18963 after #18997
- Data Integration: #18301 after #18920, and #18762 after #18846
- Gateway: #18652 after #18825
- Enhancements / Data Integration: #17472 after #17948

Separately, all `changes/ce` entries precede the `changes/ee` ones (e.g. #17729 before #17522, #17875 before #17710, #18044 before #17395, #17538/#17627/#17701/#17954 before #17414), which reads as arbitrary in the published notes. If that split is intentional, ignore that part; otherwise sorting each section by PR number would make it uniform.

Cosmetic: #17593, #17652, #17857, #17538 and #17627 keep the fragments' hard line wraps while the rest of the file uses one line per paragraph.

terry-xiaoyu (commented):
Reviewed the notes against the changelog fragments on this branch — content is in good shape.

Verified:
- Every `changes/{ce,ee}/*.en.md` fragment added since `e5.10.4` is represented, and the two duplicate pairs (`#18138`/`#18186`, `#18816`/`#18820`) are correctly merged into single entries with all links retained.
- Spot-checked a few claims against `dev-510`: `emqx_packets_publish_quota_exceeded` (#17886), `socket_opts.ip_family` (#18917) and the 500 rows/s + 10-minute sweep cadence (#17522) all match the code.

Two inline comments below: one on breaking changes, one minor on ordering.

Also: `apps/emqx/include/emqx_release.hrl` and `deploy/charts/emqx-enterprise/Chart.yaml` still read `5.10.5-rc.3`. The equivalent 5.10.4 commit (7470f0b70d) dropped the `-rc.N` suffix in the same commit that added `changes/e5.10.4.en.md` — is that intentionally left to `scripts/rel/cut.sh` this time, or still to come?

terry-xiaoyu (approved):

PR #18842, opened 2026-09-08
fix(gateway): warn on unsupported client ID overrides

hjianbo:
Improve: https://github.com/emqx/emqx-dev-team-tasks/issues/365
Release version: 6.0.4
Introduced in: 6.0.0

## Summary

Gateway protocols preserve their protocol-defined client IDs. If an authentication backend returns `clientid_override`, `emqx_gateway_ctx` removes the unsupported field, emits a warning, and continues authentication without changing the protocol client ID.

As a secondary code organization improvement, mountpoint evaluation is centralized in `emqx_gateway_ctx` after authentication results are merged. The protocol-specific `fix_mountpoint` helpers were removed.

## PR Checklist

### DIFF
--- a/changes/ee/fix-18842.en.md
+++ b/changes/ee/fix-18842.en.md
@@ -1,3 +1,3 @@
-Gateway connections now reject authentication results containing `clientid_override`, which is not supported by Gateway protocols. A warning is logged when this occurs.
+Gateway connections now ignore `clientid_override` values returned by authentication backends, which is not supported by Gateway protocols. A warning is logged when this occurs.
 
 Mountpoint templates for Gateway connections are now evaluated in the shared Gateway authentication flow after authentication results are merged.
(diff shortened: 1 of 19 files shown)


changes/ee/fix-18842.en.md
```diff
@@ -0,0 +1,3 @@
+Gateway connections now reject authentication results containing `clientid_override`, which is not supported by Gateway protocols. A warning is logged when this occurs.
```
zmstone: if we start rejecting now, it brakes for one who reuse an existing mqtt authenticator for gateway e.g. http, jwt.

since it never taken effect we can simplify the implementation to explicitly drop it from the authentication result and log a warning (unlikely to happen so don’t bother throttling)

zmstone (approved):

PR #18843, opened 2026-09-08
feat(api): add plugin_api scope for plugin-extended API calls

zmstone:
Fixes:
Introduced in:

## Summary

Adds `plugin_api`, an API scope that grants the plugin-extended API gateway and nothing else.

### The problem

All plugin-extended API calls go through one catch-all route in
`apps/emqx_plugins/src/emqx_plugins_api_endpoint.erl`:

```
"/plugin_api/:plugin/[...]"
```

It required the `system` scope. `system` is administrator-equivalent — it covers `/configs`
and `/data`, which can rewrite the whole node configuration. So anyone who needed to call a
plugin's API had to be handed the keys to the node.

**This is the plugin-extended API gateway only — the endpoints a plugin itself publishes.**
It is not `emqx_mgmt_api_plugins` (`/plugins`, `/plugins/:name`). Plugin administration —
install, start, stop, configure — stays on `system` and is untouched by this PR. The two are
easy to conflate.

### Backward compatibility

Existing API keys and Dashboard users that hold `system` must keep reaching the gateway. The
path therefore has to admit either scope, which the previous one-scope-per-path model could
not express. Two commits, so the mechanism change reads apart from the policy change.

**`refactor(api): let an API path declare several acceptable scopes`**

A `scopes/0` callback may now return a list of scope names, meaning a holder of any one of
them may call the path. A single binary keeps working, and the collector normalizes every
declaration to a list. Both scope checks — the API-key check in `emqx_mgmt_auth` and the
login-user check in `emqx_dashboard_rbac` — now test for a non-empty intersection through the
same `emqx_mgmt_api_key_scopes:any_scope_granted/2`, so a multi-scope path cannot be enforced
one way for keys and another for login users.

The alternative was to keep one scope per path and grant `plugin_api` implicitly to `system`
holders. The any-of declaration was chosen because an implication is invisible where a scope
list is read, and would have to be applied identically in API-key auth, login-user RBAC, the
role defaults and the bootstrap-file loader; missing one place breaks compatibility there
silently. The any-of form states the overlap where the path is declared, and one predicate
enforces it.

One lookup fix comes with it. The catch-all route needs it, and it is the only route of that
shape today:

* A trailing `[...]` template segment now matches the remainder of a concrete request path.
  Without it a concrete path resolved to no scope while its route template resolved to one,
  so the login-user check (which sees concrete paths) and the API-key check (which sees
  templates) disagreed about the same endpoint.

Handling of unmapped paths is unchanged. API keys are allowed through an unmapped path. Login users with an explicit scope list are denied it.

**`feat(api): add plugin_api scope for plugin-extended API calls`**

* `plugin_api` is a **restricted** scope, deliberately not in `?PRIVILEGE_SCOPES` — the whole
  point is a key that reaches a plugin without administering the node, and a privilege scope
  cannot be combined with restricted ones.
* It is in `?GENERIC_SCOPES` and `?NS_ADMIN_COMMON_SCOPES`, so accounts on the role defaults
  keep the access they have today. Namespaced callers reach the gateway through `system`
  today, and the gateway forwards the caller's namespace to the plugin.
* It is not in `?LOGIN_ONLY_SCOPES` or `?ADMIN_ONLY_SCOPES`, so API keys and Dashboard login
  users can both hold it.
* The gateway declares `[plugin_api, system]`. Keys and users created before this change hold
  `system` and keep working.

### Limitation

The scope says nothing about what a plugin does behind the gateway. A plugin can publish a
dangerous endpoint, and holding `plugin_api` reaches it. The scope bounds which EMQX surface
the caller reaches, not what the plugin exposes there.

### REST API impact

`plugin_api` is a new value in the `/api_key_scopes` and `/user_scopes` catalogs, and an
accepted value in the `POST /api_key`, `PUT /api_key/:name`, `POST /users` and
`PUT /users/:name` request bodies. Added enum value only; no endpoint, field or response shape
changed. Backward compatible.

### Tests

`emqx_plugins_api_endpoint_SUITE` (all through real HTTP against the gateway):

* `t_gateway_reachable_with_system_scope_key` — a key holding **only** `system` reaches the
  gateway. This is the compatibility regression guard: it fails if someone later removes the
  overlap.
* `t_gateway_reachable_with_plugin_api_scope_key` — a key holding only `plugin_api` reaches
  the gateway and is rejected (403) on `/configs`.
* `t_gateway_denied_without_either_scope` — a key holding neither is rejected.
* `t_gateway_login_user_scopes` — the login-user path, which runs through
  `emqx_dashboard_rbac:check_login_user_scopes/2` rather than
  `emqx_mgmt_auth:check_path_in_scopes/2`, for `plugin_api`, `system` and a third scope.
* `t_gateway_declares_both_scopes` — the route template and concrete request paths resolve to
  the same pair, including empty and multi-segment remainders.

`emqx_mgmt_api_key_scopes_SUITE`: `t_multi_scope_path` and `t_catch_all_template_matches_remainder` cover the mechanism directly against a synthetic
cache. `t_all_endpoints_covered_by_scopes` still guards against unmapped paths.

Suites run locally on the dev-60 base (OTP 27): `emqx_plugins_api_endpoint_SUITE` (13/13),
`emqx_mgmt_api_key_scopes_SUITE` (28/28), `emqx_dashboard_user_scopes_SUITE` (63/63),
`emqx_dashboard_rbac_SUITE` (31/31), `emqx_mgmt_api_api_keys_SUITE` (57/57).

## PR Checklist

### DIFF
--- a/apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
+++ b/apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
@@ -73,11 +73,15 @@ parse_dashboard_role(Role) ->
 %%                                          point), so allow.
 %%   * scopes = [...]  (list)            -> the path's scopes must
 %%                                          overlap the listed scopes;
-%%                                          unmapped paths fail-open
-%%                                          (allow).
+%%                                          unmapped paths are denied
+%%                                          (fail closed).
 %%
-%% The unmapped-path fail-open is consistent with API key scope
-%% semantics (emqx_mgmt_auth:check_path_in_scopes/2). CT
+%% Public paths are allowed for every user. An unmapped path is allowed
+%% only for a user with no explicit scope list, so a catalog gap cannot
+%% grant a deliberately restricted user an endpoint nobody scoped.
+%%
+%% This differs from API keys: emqx_mgmt_auth:check_path_in_scopes/2
+%% allows unmapped paths even for a key with an explicit scope list. CT
 %% t_all_endpoints_covered_by_scopes guards against accidentally
 %% leaving a non-public path unmapped.
 %%
--- a/apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
+++ b/apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
@@ -71,9 +71,10 @@ parse_dashboard_role(Role) ->
 %%   * scopes absent  (undefined)        -> fall back to RBAC default
 %%                                          (already passed at this
 %%                                          point), so allow.
-%%   * scopes = [...]  (list)            -> path must map to one of
-%%                                          the listed scopes; unmapped
-%%                                          paths fail-open (allow).
+%%   * scopes = [...]  (list)            -> the path's scopes must
+%%                                          overlap the listed scopes;
+%%                                          unmapped paths fail-open
+%%                                          (allow).
 %%
 %% The unmapped-path fail-open is consistent with API key scope
 %% semantics (emqx_mgmt_auth:check_path_in_scopes/2). CT
@@ -142,13 +143,15 @@ check_login_user_scopes_strict(Username, Path) ->
         %% governed by role-based RBAC alone, so they are not locked out.
         not_found ->
             emqx_dashboard_admin:scopes_of(Username) =:= undefined;
-        {scope, PathScope} ->
+        {scopes, PathScopes} ->
             %% Work on the effective scope list (role-default expanded) so
             %% administrators with no explicit scopes implicitly hold the
             %% full catalog and viewers implicitly hold the common scopes.
             %% Explicit [] is honoured as "no permissions".
+            %% A path may declare more than one acceptable scope; holding
+            %% any one of them grants access.
             Scopes = emqx_dashboard_admin:effective_scopes_of(Username),
-            lists:member(PathScope, Scopes)
+            emqx_mgmt_api_key_scopes:any_scope_granted(PathScopes, Scopes)
     end.
 
 %% Whitelist of self-service paths that may skip the login-user
--- a/apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl
@@ -193,21 +193,7 @@ split_segments(Path) ->
 %% after splitting so an encoded separator stays within its segment
 %% and never introduces an extra boundary.
 normalize_segments(Segments) ->
-    remove_dot_segments([urldecode(S) || S <- Segments], []).
-
-%% `cow_uri:urldecode/1' raises on a byte that is not valid in a URI
-%% path. Cowboy validates a request path before it dispatches, so such
-%% a byte can only reach here from a value that did not come from the
-%% router — a route template carrying `[...]', for one. Keep the
-%% segment verbatim in that case: it then matches only a template
-%% segment that is literally equal, and the authorisation path returns
-%% a decision instead of raising.
-urldecode(Segment) ->
-    try
-        cow_uri:urldecode(Segment)
-    catch
-        _:_ -> Segment
-    end.
+    remove_dot_segments([cow_uri:urldecode(S) || S <- Segments], []).
 
 remove_dot_segments([], Acc) ->
     lists:reverse(Acc);
--- a/apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl
+++ b/apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl
@@ -8,12 +8,19 @@
 API Key scope management.
 
 Each minirest_api module declares its scope via a `scopes/0` callback
-that returns either a scope name binary (all paths share the same
-scope) or a `#{Path => ScopeName}` map (for modules whose endpoints
-span multiple scopes).
+that returns either a scope declaration (all paths share it) or a
+`#{Path => ScopeDeclaration}` map (for modules whose endpoints span
+multiple scopes).
 
-This module collects those declarations, builds a path → scope cache,
-and exposes the user-visible scope catalog.
+A scope declaration is either a single scope name binary or a
+non-empty list of scope names. A list means the path is reachable by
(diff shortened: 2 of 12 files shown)


apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
```diff
@@ -71,9 +71,10 @@ parse_dashboard_role(Role) ->
...
 %%                                          (already passed at this
 %%                                          point), so allow.
-%%   * scopes = [...]  (list)            -> path must map to one of
-%%                                          the listed scopes; unmapped
-%%                                          paths fail-open (allow).
+%%   * scopes = [...]  (list)            -> the path's scopes must
+%%                                          overlap the listed scopes;
+%%                                          unmapped paths fail-open
+%%                                          (allow).
 %%
```
zmstone, line 78: [P3] Describe unmapped login-user paths as fail-closed

For a user with any explicit scope list, the `not_found` branch below returns `false`; only users whose scopes are unset are allowed. This edited comment still says unmapped paths fail open, and the following paragraph says the semantics match API keys even though API keys always fail open. Please update both statements so this security-sensitive behavior is documented accurately.

apps/emqx_utils/include/emqx_api_key_scopes.hrl
```diff
@@ -97,7 +119,8 @@
     ?SCOPE_CLUSTER_OPERATIONS,
     ?SCOPE_SYSTEM,
     ?SCOPE_AUDIT,
-    ?SCOPE_LICENSE
+    ?SCOPE_LICENSE,
```
zmstone, line 122: [P3] Update the hard-coded default-scope counts

This makes `GENERIC_SCOPES` contain 11 entries and `NS_ADMIN_COMMON_SCOPES` contain 8, but comments in `emqx_mgmt_auth.erl`, `emqx_dashboard_user_scopes_SUITE.erl`, and `emqx_mgmt_api_key_scopes_SUITE.erl` still describe 10/7. Please update those references as part of adding the scope; otherwise the documented default shape is immediately stale.

savonarola:
My remark is the same as before: we put more and more logic into the same problematic path of parallel dispatch implementation. 

What Astra noted:

> The scope matcher still does not implement the same route-selection process as Cowboy:
> - Route precedence: it returns the first matching entry from `maps:iterator(PathMap)`. That does not preserve Cowboy’s dispatch order. Overlapping templates can therefore select different scope declarations.
> - Normalization: Cowboy drops the final empty segment from a trailing slash; the scope classifier’s binary:split/3 retains it. There are already differences beyond the catch-all issue.
> - Public-path handling: exact raw-path hits get special treatment, while public entries are skipped during template matching. Equivalent encoded paths can take a different classification path.
> 
> These are existing properties of the matcher, rather than newly introduced bypasses. **Adding a catch-all increases the matching behavior that must stay consistent with Cowboy.**

We are increasing the risk of dispatch divergence and the resulting security issues.  Tying the scope to `{emqx_plugins_api_endpoint, gateway}`  would be much more robust.

savonarola:
👀

apps/emqx_dashboard_rbac/src/emqx_dashboard_rbac.erl
```diff
@@ -130,53 +132,51 @@ serialize_role(#{?role := Role, ?namespace := ?global_ns}) ->
...
-            Decoded = uri_string:percent_decode(SelfSeg),
-            is_same_user(Decoded, Username);
-        _ ->
-            false
-    end;
-is_self_service_endpoint(_Path, _Username) ->
+%% The target user is the `username' path binding, already decoded by
+%% cowboy. It must be the caller, whether Username is a bare binary
+%% (local) or a ?SSO_USERNAME(Backend, Name) tuple (SSO; the binding
+%% is just Name).
```
savonarola: new version looks so much more robust

apps/emqx_dashboard_rbac/test/emqx_dashboard_rbac_SUITE.erl
```diff
@@ -652,11 +632,11 @@ t_check_login_user_scopes_explicit_empty_denies_generic_paths(_) ->
     {ok, ok} = emqx_dashboard_admin:set_user_scopes(Username, []),
     ?assertEqual(
         false,
-        emqx_dashboard_rbac:check_login_user_scopes(Username, <<"/clients">>)
+        login_scope_check(Username, clients)
```
savonarola: nit: maybe not in this PR since it follows the existing patter, but in all `login_scope_check`-related tests I'd prefer having not forged `HandlerInfo` and `Bindings`, but actual ones, obatained from actual requests done to cowboy.

apps/emqx_management/src/emqx_mgmt_api_key_scopes.erl
```diff
@@ -360,51 +311,115 @@ collect_module_scopes(Module, Acc) ->
...
+    end.
+
+%% Map form: per-path scope assignment. The sentinel ?SCOPE_PUBLIC marks
+%% paths that are intentionally unscoped (pre-login entry points and
+%% static catalog endpoints). Such paths ARE inserted into the cache,
+%% carrying `[?SCOPE_PUBLIC]', so that lookups can tell "explicitly
+%% public" from "genuinely unmapped". Genuinely missing paths warn.
+declared_scopes(Module, Path, ScopeMap) when is_map(ScopeMap) ->
+    PathBin = path_to_binary(Path),
+    case maps:get(PathBin, ScopeMap, maps:get(Path, ScopeMap, undefined)) of
```
savonarola: I think that when evaluating `ScopeSpec = apply(Module, scopes, [])`, we should strictly validate it, not allowing any fuzziness and normalization. 

We should strictly forbid `ScopeMap` contain a `Path` that a) is not in `Module:paths()` b) returns missing clause on `Module:schema(Path)`. 

Alternatively, instead of map form of scopes, we may move path-specific scopes to schema.

savonarola (dismissed):

savonarola:
👍

zmstone:
@savonarola thank you for the review. fixed the nits.

savonarola (approved):

PR #18849, opened 2026-09-08
sync r63 to r70

zmstone:
Forward sync of `dev-63` into `dev-70`. No 7.0 baseline to pin yet, so there is
no freeze commit — the baselines simply arrive from dev-63.

```
5.8.bpapi2  5.9.bpapi2  5.10.bpapi2  6.0.bpapi2  6.1.bpapi3  6.2.bpapi3  6.3.bpapi3
```

Five conflicts. Two needed more than picking a side.

## `mix.exs` — each side held the newer tag for a different dep

```
dev-70   ekka 1.0.2   esockd 5.17.3
dev-63   ekka 1.0.1   esockd 5.17.4
```

Taking either side wholesale downgrades one of them. Resolved to the newer of
each: ekka 1.0.2 and esockd 5.17.4. The lines that auto-merged were all dev-70
supersessions — grpc-erl 0.7.10, minirest 1.5.1, and gpb gaining
`manager: :make` — so nothing regressed there either.

## `build_packages.yaml` — the smoke test was running twice

The macOS smoke test lived in *both* the `package-macos` composite action and
inline in the workflow, so it ran twice on every macOS package build.
`3da7219e38 ci: speed up macos package builds` deleted the inline copy on dev-63;
dev-70 still had both.

```
                      action.yaml   build_packages.yaml
merge-base                  1                1
dev-63                      1                0
dev-70                      1                1
merged                      1                0
```

Taking dev-63's side carries that de-duplication forward. The two blocks were
step for step identical — same cookie and password exports, same
`ping` / `help` / `ctl status` / `emqx eval 'emqx_conf:check_dynamic_desc()'` —
and the composite action's copy is the better one: it sets `set -e` and lets
`start` and `stop` fail hard instead of swallowing the error with
`|| cat emqx/log/erlang.log.1`. So dev-70 loses a redundant run and gains
notarization.

## The rest

- `emqx_dashboard_api.erl` and `emqx_mgmt_auth.erl`: dev-63's side is empty in
  both; kept dev-70's MFA macro and authorize clauses. The resolution drops
  dev-63's standalone `change_pwd` clause, which is safe because dev-70's guard
  clause already covers `Fn == change_pwd`.
- `emqx_mgmt_api_clients_SUITE`: `t_persistent_sessions_exact_clientid_query`
  (dev-70) and `t_persistent_sessions_filters` (dev-63) are different tests. Both
  kept. The `-doc """` opener was shared context ahead of the marker, so the
  second test needed its own — `make fmt` caught that.

## Verification

`make static_checks` passes end to end, `make fmt-diff` clean, elvis clean,
`apps-version-check.exs` clean after the bumps in the second commit.

Those bumps read 6.3.1 because `EMQX_RELEASE_EE` on dev-70 is still
"6.3.0-rc.1", which the check derives the target from. Unrelated to this sync,
but worth knowing before the 7.0 cut.


thalesmg:
could we also pack this one here?  :crossed_fingers: 

https://github.com/emqx/emqx/pull/18848

zmstone:
> could we also pack this one here? 🤞
> 
> #18848

done

hjianbo (dismissed):

keynslug (approved):

Issue #366, opened 2015-10-30
Subscribe timeout exception


erylee:
```
2015-10-27 17:02:21 =CRASH REPORT====
  crasher:
    initial call: emqttd_client:init/1
    pid: <0.10953.8024>
    registered_name: []
    exception exit: {{timeout,{gen_server2,call,[<0.391.7942>,{subscribe,[{<<"znews/global">>,1},{<<"znews/03604e144aef0755">>,1}]},60000]}},[{gen_server,terminate,7,[{file,"gen_server.erl"},{line,804}]},{proc_lib,wake_up,3,[{file,"proc_lib.erl"},{line,247}]}]}
    ancestors: [<0.305.0>,<0.304.0>,esockd_sup,<0.121.0>]
    messages: [{inet_async,#Port<0.219727177>,17713,{error,closed}}]
    links: [<0.305.0>]
    dictionary: []
    trap_exit: false
    status: running
    heap_size: 2586
    stack_size: 27
    reductions: 2312
  neighbours:
```

erylee:
Fixed. 0.14 release could handle 4000+ SUBSCRIBE/second.

PR #18855, opened 2026-09-09
fix(mqttsn): validate cross-peer session wake-up

hjianbo:
Fixes: emqx/emqx-dev-team-tasks#366
Release version: 6.0.4
Introduced in: N/A

## Summary

MQTT-SN PINGREQ carries only a Client ID and cannot authenticate a new network association by itself. This change preserves the existing ClientId-only wake-up behavior for plaintext UDP and DTLS connections without a client certificate, including when the source IP or port changes.

For DTLS connections authenticated with a verified client certificate:

- Bind the session to the peer certificate presented during the original connection.
- Allow a PINGREQ from a new association to resume the session only when it presents the same peer certificate.
- Reject a missing, different, or reissued certificate, even when the reissued certificate has the same subject DN and CN.
- Do not use the source IP or port as the session identity; certificate rotation requires CONNECT and the normal authentication/session takeover flow.

The MQTT-SN resume coordinator now:

- Sends only the peer certificate in the authorized takeover-begin request.
- Uses the standard `{takeover, 'begin', Request}` and `{takeover, 'end'}` interaction.
- Keeps the new connection as the takeover owner and restores the old channel with a watchdog if resume processing is interrupted.
- Preserves queued deliveries and subscriptions during a successful resume.
- Handles malformed replies and connection/RPC failures without crashing the channel or connection manager.

Configure the DTLS listener with `verify = verify_peer` and `fail_if_no_peer_cert = true` when every client must present a verified certificate. Plaintext UDP and server-only DTLS retain the legacy ClientId-only PINGREQ behavior.

## PR Checklist

- [x] Change log for changes visible by users has been added to changes/ee/(feat|perf|fix|breaking)-<PR-id>.en.md files


apps/emqx_gateway/src/proto/emqx_gateway_cm_proto_v1.erl
```diff
@@ -78,7 +78,7 @@ kick_session(GwName, Action, ClientId, ChanPid) ->
     emqx_gateway_cm:gateway_name(),
     emqx_types:clientid(),
     pid()
-) -> atom() | {badrpc, _}.
+) -> atom() | undefined | {badrpc, _}.
```
zmstone, line 81: atom() covers 'undefined', so no change.
if it's really a new contract, there should be a v2
and ideally atom() should be expanded

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -171,6 +177,7 @@ init(
```
zmstone: **[P1] Do not retain unauthenticated cross-peer resume by default**

This branch still returns `true` for `nossl` and `undefined` peer certificates, so the default plaintext UDP listener and server-only/optional-client-certificate DTLS listener continue to let any sender who knows the ClientId resume the sleeping channel and drain its queued messages. The new test `t_asleep_pingreq_from_new_peer_resumes_legacy_udp` explicitly preserves the exploit that issue #366 reports. That may be a deliberate compatibility choice, but it means this PR does not fix the vulnerability for the default listener and the changelog title/PR summary overstate the coverage. Please reject source changes for unauthenticated transports, or gate the legacy behavior behind an explicit opt-in with a prominent security warning.

apps/emqx_gateway/src/emqx_gateway_cm.erl
```diff
@@ -400,6 +403,239 @@ open_session(
...
+        {ok, Pendings} when is_list(Pendings) ->
+            {ok, Pendings};
+        {ok, {error, Reason}} ->
+            {error, Reason};
+        ok ->
+            ?SLOG(warning, #{
+                msg => "mqttsn_resume_takeover_end_unexpected_reply",
+                reply => ok,
+                chan_pid => ChanPid
+            }),
```
zmstone, line 628: **[P1] Do not report takeover success after losing pending deliveries**

When `request_stepdown({takeover, 'end'}, ...)` fails (for example timeout/noproc), this branch returns `{ok, []}` after killing or losing the old channel. `resume_session_locked/7` then treats the resume as successful and the new channel proceeds, but the old channel's `pendings` were never returned in `TakeoverData` and are not merged into the new session. Any queued QoS delivery accumulated during the takeover can therefore disappear. Please return an error and keep/rollback the old channel, or transfer the pending list before force-killing it; add a test that injects an end timeout with a queued delivery and verifies it is delivered exactly once after the next wake-up.

apps/emqx_gateway/src/emqx_gateway_cm.erl
```diff
@@ -400,6 +403,239 @@ open_session(
...
+    emqx_types:clientinfo(),
+    emqx_types:conninfo(),
+    module()
+) ->
+    {ok, resume_result()} | {error, term()}.
+resume_session(
+    GwName,
+    ClientInfo = #{clientid := ClientId},
+    ConnInfo,
+    SessionMod
```
zmstone, line 431: **[P2] Do not discard connected duplicate channels during resume**

`OtherPids` contains every registry entry except the selected candidate, and `discard_other_channels/3` immediately discards all of them before the new session is registered. `select_resume_candidate/2` does not filter those entries by `asleep`/`awake` state. If stale duplicate registry entries include a currently connected channel while an asleep candidate is selected, a PINGREQ can terminate that active channel even though it is not the channel being resumed. Filter the extra PIDs by live resumable state, or make the candidate/cleanup decision state-aware and atomic.

apps/emqx_gateway_mqttsn/src/emqx_mqttsn_channel.erl
```diff
@@ -514,6 +560,48 @@ handle_in(
...
+                asleep_timer_duration = SleepDuration,
+                conn_state = asleep
+            },
+            ResumedChannel = schedule_connection_expire(ResumedChannel0),
+            awake(
+                ClientId,
+                ResumedChannel
+            );
+        {error, _Reason} ->
+            handle_out(disconnect, normal, Channel)
```
zmstone, line 608: should add a log/trace for resumption failure?

apps/emqx_gateway/src/emqx_gateway_cm.erl
hjianbo: This behavior is intentional and follows the existing `emqx_gateway_cm:takeover_session` semantics.

The resume path selects one channel from the registry. Once the selected channel has successfully authorized the takeover, the remaining registry entries are treated as stale duplicate channels and discarded. Changing this behavior to preserve connected duplicates would introduce a different takeover policy and is outside the scope of this PR.

To make duplicate registry state observable, this change adds the `more_than_one_channel_found` warning when multiple channel PIDs are found during resume.

zmstone (approved):

Issue #18761, opened 2026-09-03
[6.3.0] HTTP authentication headers are lost when changing header-name casing


Le9Wang:
### What happened?

Updating an HTTP authenticator from the Dashboard can silently remove a sensitive HTTP header when only the header name's casing changes.

For example, an existing `authorization` header is returned as `******` by the API. If the user changes its name to `Authorization` in the Dashboard and saves without re-entering the secret, the updated authenticator ends up with an empty `headers` object.

### Reproduction Steps

1. Create a Password-Based HTTP authenticator with:
   - `hostname_resolution: dynamic`
   - method: `get`
   - headers: `Authorization: abcd1234`
2. Open the authenticator in the Dashboard. The API returns the header name normalized as `authorization` and its value as `******`.
3. Change only the header name from `authorization` to `Authorization`.
4. Save the authenticator. The Dashboard sends:

   ```json
   {
     "headers": {
       "Authorization": "******"
     }
   }
   ```

5. Query `GET /api/v5/authentication`.

### Expected Behavior

HTTP header names are case-insensitive. Changing only a header name's casing must preserve the existing sensitive value, or the UI/API should require an explicit replacement value. The saved configuration should retain the header.

### Actual Behavior

The saved authenticator configuration contains:

```json
{
  "headers": {}
}
```

The Authorization header is silently removed, which can cause HTTP authentication requests to fail.

### Root Cause

HTTP authenticator schema conversion normalizes header names to lowercase. However, redacted-value restoration during an update matches header map keys case-sensitively before that normalization.

Therefore, the old key `authorization` does not match the submitted key `Authorization`. Since `******` is recognized as a redacted sensitive value, it is dropped instead of being restored from the old configuration.

### Suggested Fix

Normalize header names before deobfuscating/redacted-value restoration, or make the header-key lookup case-insensitive.

### Environment

- **Software version**: EMQX 6.3.0
- **Module**: HTTP Authentication / Dashboard
- **Configuration**:

  ```json
  {
    "backend": "http",
    "hostname_resolution": "dynamic",
    "method": "get",
    "url": "http://127.0.0.1:8080"
  }
  ```

PR #18859, opened 2026-09-09
fix(redact): preserve masked HTTP headers across case changes

hjianbo:
Fixes: #18761
Introduced in: pre-5.8

## Summary

HTTP header names are case-insensitive, but sensitive-value deobfuscation matched header map keys case-sensitively before authentication schema normalization. A casing-only edit could therefore drop a masked header instead of restoring its stored value.

This change preserves exact-key round trips and falls back to case-insensitive lookup only when the stored value is unambiguous. Ambiguous casing variants remain fail-closed, and case-insensitive matching is limited to direct HTTP header entries so ordinary nested configuration maps keep exact-key semantics.

## PR Checklist

### DIFF
--- a/apps/emqx_utils/src/emqx_utils_redact.erl
+++ b/apps/emqx_utils/src/emqx_utils_redact.erl
@@ -224,6 +224,37 @@ check_is_sensitive_header(Key) ->
     Key2 = string:trim(emqx_utils_conv:str(Key1)),
     is_sensitive_header(string:lowercase(Key2)).
 
+normalize_header_name(Key) ->
+    Key1 =
+        try iolist_to_binary(Key) of
+            Bin -> Bin
+        catch
+            _:_ -> Key
+        end,
+    list_to_binary(string:lowercase(emqx_utils_conv:str(Key1))).
+
+find_header(Key, Headers) ->
+    case maps:find(Key, Headers) of
+        {ok, _} = Found ->
+            Found;
+        error ->
+            NormalizedKey = normalize_header_name(Key),
+            MatchingValues = [
+                Value
+             || {OldKey, Value} <- maps:to_list(Headers),
+                normalize_header_name(OldKey) =:= NormalizedKey
+            ],
+            find_unambiguous_header(MatchingValues)
+    end.
+
+find_unambiguous_header([]) ->
+    error;
+find_unambiguous_header([Value | Values]) ->
+    case lists:all(fun(V) -> V =:= Value end, Values) of
+        true -> {ok, Value};
+        false -> error
+    end.
+
 is_sensitive_header("authorization") ->
     true;
 is_sensitive_header("api-key") ->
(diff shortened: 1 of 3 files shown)


apps/emqx_utils/src/emqx_utils_redact.erl
```diff
@@ -224,6 +224,37 @@ check_is_sensitive_header(Key) ->
     Key2 = string:trim(emqx_utils_conv:str(Key1)),
     is_sensitive_header(string:lowercase(Key2)).
 
+normalize_header_name(Key) ->
```
zmstone, line 217: This module now has two normalisers for the same header key, and they disagree.

`check_is_sensitive_header/1` (line 214) does the same `try iolist_to_binary ... catch` dance, then `string:trim`, then lowercase. `normalize_header_name/1` skips the trim. So a key such as `<<" Authorization ">>` is *classified* sensitive — `GET` masks it — but is never *matched* on `PUT`, and is dropped. The `HeaderConf3` case in `deobfuscate_test` pins exactly that outcome, so it is deliberate; but it means "is this sensitive" and "does this match the stored one" follow different rules for the same input, which is the shape of bug this PR fixes.

The schema converter is the reference: `emqx_auth_http_utils:transform_header_name/1` does `string:to_lower(to_list(K0))` — lowercase only, no trim. `normalize_header_name/1` mirrors that correctly. `check_is_sensitive_header/1`'s trim is the odd one out.

Suggest one helper used by both — `check_is_sensitive_header(Key) -> is_sensitive_header(normalize_header_name(Key))` — with a one-line comment naming `transform_header_name/1` as the rule it must stay in step with. That is the real fix for #18761: the two rules live in different apps and drifted apart with nothing to say they are supposed to agree.

apps/emqx_utils/src/emqx_utils_redact.erl
```diff
@@ -224,6 +224,37 @@ check_is_sensitive_header(Key) ->
...
+                Value
+             || {OldKey, Value} <- maps:to_list(Headers),
+                normalize_header_name(OldKey) =:= NormalizedKey
+            ],
+            find_unambiguous_header(MatchingValues)
+    end.
+
+find_unambiguous_header([]) ->
+    error;
+find_unambiguous_header([Value | Values]) ->
```
zmstone, line 243: When the stored `headers` holds two casing variants with *different* values and the request sends a third casing, this returns `error`, `is_redacted/3` is true, and the key is dropped — the request ends up with `headers => #{}`. That is the same silent-removal symptom #18761 reports, kept for a rarer input.

Fail-closed is the safe default, and the `AmbiguousHeadersObs` test pins it, so this is a question rather than a defect: is dropping actually better than picking one? The checked config has already been through `transform_header_name/1` and holds exactly one value under the lowercase key, so the duplicate in the raw config is an artefact of how it was written, not two live headers. Preferring the entry whose key is already lowercase — the one the schema would have used — would resolve the ambiguity the same way the running system already does.

If you keep the drop, a `?SLOG(warning, ...)` at this point would at least make the removal visible; today it is indistinguishable from the user having cleared the header on purpose.

apps/emqx_utils/src/emqx_utils_redact.erl
```diff
@@ -301,21 +332,32 @@ do_redact_v(F) ->
             ?REDACT_VAL
     end.
 
+deobfuscate_headers(NewHeaders, OldHeaders) ->
```
zmstone, line 328: Worth a comment here on the scoping: case-insensitive lookup applies to the direct entries of the `headers` map only. A map-valued header recurses through the `{ok, OldV} when is_map(V), is_map(OldV)` clause, which calls `deobfuscate/3`, which reverts to `maps:find/2` — so nested keys are exact-match again. The `NestedHeaderConf` test asserts it and the PR description states it, but in the code the property is a side effect of which arity the recursion happens to call, and a later refactor that threads `FindOldValue` through the recursion would silently widen the rule.

Otherwise the `deobfuscate/4` injection is a clean way to do this — one new argument, default `maps:find/2`, every existing caller untouched.

zmstone (commented):
comments from agent.

apps/emqx_utils/src/emqx_utils_redact.erl
hjianbo: We decided to keep the fail-closed behavior here. For the HTTP authenticator path, the checked configuration is normalized and the normal casing-change case is unambiguous. However, `emqx_utils:deobfuscate/2` is also used on raw connector and bridge configurations before schema-specific normalization, so there is not always an authoritative lowercase entry.

Exact-key matches still take precedence, and fallback restoration succeeds only when the matching values are unique or all equal. When conflicting values remain, selecting either stored secret would be arbitrary.

We also kept this shared utility free of logging side effects because it does not have the configuration path or caller context needed for an actionable warning. The behavior is documented and covered by exact-match, equal-value, and conflicting-value tests.

zmstone (approved):

PR #18864, opened 2026-09-09
ci: fail the build on CT cases listed in no group

zmstone:
Fixes:
Introduced in: N/A

## Summary

A CT case that appears in neither `all/0` nor any group in `groups/0` never
runs, and nothing reports it. The compiler sees an exported function, elvis
sees nothing, and CT never looks at the case. It rots silently until the code
it covered has been removed under it — two cases here called
`emqx_mgmt_auth:authorize/3`, which became `authorize/4` returning
`{ok, AuthMeta}` long ago, and one looked up `emqx_cluster_cleaner`, a module
name that has never existed.

### The check — `scripts/check-unlisted-ct-cases.escript`

It evaluates each `*_SUITE.beam` rather than reading its source: it calls
`all/0` and `groups/0`, expands group references recursively, and reports every
exported `t_*/1` the expansion does not reach. A source scan cannot do this —
many suites build their lists from helpers (`emqx_common_test_helpers:all/1`,
`groups_with_matrix/1`, `emqx_authz_test_lib:table_groups/2`) that collect case
names through `module_info/1`.

The expansion handles every CT form: bare atoms, `{group, G}` with and without
properties, `{testcase, TC, Props}`, inline group definitions, and lists nested
inside `all/0`. Group names that are legitimately reused under several parents
(`with_batch` under both `sync` and `async`) expand under each, and a group name
defined twice contributes both definitions, matching `ct_groups:find/7`.

A suite whose `all/0` or `groups/0` raises is a hard error naming the suite, not
a skip. There is no allowlist and no skip mechanism. A function that is not a
case, or a case disabled on purpose, is renamed so it does not start with `t_`
(for example `disabled__t_foo`), with the reason written next to it.

It runs in the compile job next to the API scope check, where the test-profile
beams exist. When it finds no suite at all, it fails with
`no CT suites found under <dir>`, so it cannot pass by evaluating nothing.

### The 25 cases it found

The check ships strict, so every one is resolved. One commit per suite.

| suite | cases | outcome |
|---|---|---|
| `emqx_frame_SUITE` | 1 | listed, passes unchanged |
| `emqx_client_SUITE` | 4 | listed, pass unchanged |
| `emqx_cluster_rpc_SUITE` | 2 | listed, both repaired |
| `emqx_exproto_SUITE` | 1 | listed, repaired |
| `emqx_mgmt_api_api_keys_SUITE` | 4 | listed, 2 repaired |
| `emqx_rule_engine_SUITE` | 5 | listed, 1 repaired |
| `emqx_quic_multistreams_SUITE` | 7 | 3 listed, 4 deleted |
| `emqx_ds_builtin_raft_SUITE` | 1 | renamed, stays disabled |

Rot repaired, all detailed in the individual commits:

- `emqx_cluster_rpc_SUITE:t_cleaner_unexpected_msg` looked up
  `emqx_cluster_cleaner`. The module was renamed `emqx_cluster_rpc_handler` ->
  `emqx_cluster_rpc_cleaner` and the rename mistyped the new name here. It is
  also unregistered now, so take the pid from `emqx_conf_sup`, and
  `handle_call/3` replies `ignored`, not `ignore`.
- `emqx_cluster_rpc_SUITE:t_catch_up_status_handle_next_commit` asserted a
  two-element `{ok, 2}`; the reply has been `{ok, TnxId, Result}` for years. It
  also drove the commit through a failure-injecting MFA that fails on every odd
  call per process, so the outcome depended on how many times catch-up retried
  rather than on the behaviour the name describes.
- `emqx_exproto_SUITE:t_idle_timeout` mecked `emqx_exproto_gcli:async_call/3`,
  replaced by `maybe_shoot/3` when the unary handler landed. Meck the generated
  unary handler client instead, so the gcli queue and inflight bookkeeping stays
  real. Its trace assertion also matched any `conn_process_terminated` event,
  and the preceding case in the sequence emits one; match the reason too.
- `emqx_mgmt_api_api_keys_SUITE`: the two bootstrap-override cases called
  `emqx_mgmt_auth:authorize/3`. Route them through the suite's
  `auth_authorize/3` helper and match `{ok, _}`. `t_bootstrap_file_dup_override`
  also asserted the order `mnesia:match_object/1` returns two records with the
  same api key; that order is not promised, so compare sorted names.
- `emqx_rule_engine_SUITE:t_function_clause_errors` expected the SQL runtime to
  wrap a `function_clause` from `upper/1` as `bad_sql_function_argument` with
  the argument list. `emqx_variform_bif` now rejects a missing column in
  `upper/1` itself and throws `#{reason => badarg, function => upper}`, which
  is not a `function_clause` error and so passes through `do_apply_func/4`
  unwrapped. Note that the user-facing error lost the `arguments` field in that
  change; that is a separate question, not touched here.

Deleted, with reasons:

- `emqx_quic_multistreams_SUITE`: `t_quic_sock`, `t_0_rtt` and `t_0_rtt_fail`
  test the `emqtt_quic` client transport against `quic_server`, a helper in the
  emqtt dependency's own test directory. EMQX does not compile that directory,
  so they can only crash with `undef`; `t_quic_sock_fail` is the negative
  counterpart of `t_quic_sock`. emqtt covers its client transport in its own
  suite.

Kept disabled:

- `emqx_ds_builtin_raft_SUITE:t_crash_restart_recover` was subtracted from
  `all/0` by a `Broken` list, with the rework it needs written next to it. It is
  renamed to `disabled__t_crash_restart_recover`, so `all/0` does not collect it
  and the check does not report it. The reason stays next to the function.

### Why `dev-60`

The dead cases exist on every 6.x branch, and the check is only useful if each
branch is clean when it arrives. Landing here carries it through
`dev-61` -> `dev-62` -> `dev-63` -> `dev-70` -> `master` by forward sync.

### Expected sync conflict

`dev-63` PR #18852 already resurrected `t_bootstrap_file_override` and
`t_bootstrap_file_dup_override` in the api-keys suite, ported to `authorize/4`
and with the same order-independent assertion. The `dev-60` -> `dev-61` -> ... ->
`dev-63` sync will conflict on those two cases; keep the `dev-63` version. The
other two api-keys cases were not touched by #18852.

## Verification

- `make test-compile`, then `./scripts/check-unlisted-ct-cases.escript
  _build/emqx-enterprise-test/lib`: 25 cases before the fixes, `OK: 448 CT
  suite(s)` after.
- An empty or missing lib dir fails with `no CT suites found under <dir>` and
  exit 1. `disabled__t_crash_restart_recover` is exported but absent from
  `emqx_ds_builtin_raft_SUITE:all/0`.
- Every touched suite run in full locally, all green: `emqx_frame_SUITE`,
  `emqx_client_SUITE` (mqttv4 and others groups), `emqx_cluster_rpc_SUITE`
  (11/11), `emqx_exproto_SUITE` (76/76), `emqx_mgmt_api_api_keys_SUITE`
  (61/61), `emqx_rule_engine_SUITE` (91/91), `emqx_quic_multistreams_SUITE`
  (misc and takeover groups).
- `make fmt-diff`, `./scripts/elvis-check.sh $(git merge-base origin/dev-60 HEAD)`.

## PR Checklist


apps/emqx_ds_builtin_raft/test/emqx_ds_builtin_raft_SUITE.erl
```diff
@@ -15,6 +15,11 @@
 
 -define(DB, testdb).
 
+%% Disabled on purpose, not forgotten -- see `Broken' in all/0. Keep
+%% the two in step so scripts/check-unlisted-ct-cases.escript does not
+%% report the case as an oversight.
+-ct_lint_skip_cases([t_crash_restart_recover]).
```
keynslug: Nit. Introducing special module attributes sounds excessive. Maybe just rename the function to `disabled__t_crash_restart_recover`?

.github/workflows/compile.yaml
```diff
@@ -66,6 +66,23 @@ jobs:
...
+        run: |
+          ## Same lib dir as the scope check above: the suite beams only
+          ## exist in the test profile.
+          OUTPUT="$(./scripts/check-unlisted-ct-cases.escript _build/${PROFILE}-test/lib)"
+          echo "$OUTPUT"
+          ## Assert the check evaluated something. A check that silently
+          ## looks at no suite passes without testing anything, which is
+          ## the failure mode it exists to catch.
+          COUNT="$(echo "$OUTPUT" | sed -n 's/^OK: \([0-9]\{1,\}\) CT suite(s).*/\1/p')"
+          if [ -z "$COUNT" ] || [ "$COUNT" -eq 0 ]; then
```
keynslug: Nit. Maybe turn this into a script-level error condition (e.g. `No suites were found in <...>`), and drop odd `maybe_hint_test_profile/0` to simplify / generalize?

qzhuyan (dismissed):

keynslug (approved):

Issue #18753, opened 2026-09-03
[6.3.0] SAML GET API masks an empty SP private key as configured


Le9Wang:
### What happened?

After SAML request signing is disabled, the effective configuration correctly clears
`sp_private_key` and `sp_public_key`. However, `GET /api/v5/sso/saml` still returns
`"sp_private_key": "******"`.

This response is ambiguous: it is identical to the response for a configured private
key, so Dashboard/API clients cannot determine that the key has been cleared.

Source analysis shows that the SSO API reads the effective configuration and applies
the generic `emqx_utils:redact/1` function. `sp_private_key` is a sensitive key, and
the redaction function replaces every binary value, including an empty binary (`<<>>`),
with `<<"******">>`.

### Reproduction Steps

1. Configure SAML SSO with `sp_sign_request` enabled and provide SP public/private keys.
2. Update the SAML configuration to set `sp_sign_request` to `false`.
3. Run:

   ```bash
   emqx ctl conf show dashboard | grep key
   ```

4. Verify that the effective configuration shows:

   ```hocon
   sp_private_key = ""
   sp_public_key = ""
   ```

5. Query the SAML backend:

   ```bash
   curl --insecure \
     -H "Authorization: Bearer <token>" \
     http://<dashboard-host>:18083/api/v5/sso/saml
   ```

### Expected Behavior

When request signing is disabled and the effective private key is empty, the GET API
should return an empty value for `sp_private_key`, allowing clients to distinguish
“no private key configured” from “private key configured but redacted”.

### Actual Behavior

The GET API returns:

```json
{
  "sp_sign_request": false,
  "sp_private_key": "******"
}
```

The CLI simultaneously reports:

```hocon
sp_private_key = ""
sp_public_key = ""
```

### Environment

- **Software version / build**: EMQX Enterprise `6.3.0-rc.2-ga79302ea`
- **Configuration**:

  ```hocon
  dashboard.sso.saml {
    sp_sign_request = false
    sp_private_key = ""
    sp_public_key = ""
  }
  ```

zmstone:
similar to
https://github.com/emqx/emqx/issues/18757

Issue #18757, opened 2026-09-03
[6.3.0] Cluster Linking GET API masks an unset password as configured


Le9Wang:
### What happened?

Creating a Cluster Linking connection without providing a password succeeds, but
`GET /api/v5/cluster/links/link/:name` returns `"password": "******"`.

This response is indistinguishable from a link with a configured password. It causes
Dashboard/API clients to treat an unset optional password as configured.

The issue is reliably reproducible immediately after creating a Cluster Link without
a password.

### Reproduction Steps

1. Create a Cluster Link without setting `password`.

   ```json
   {
     "name": "link-without-password",
     "enable": true,
     "server": "mqtt://<remote-broker>:1883",
     "username": ""
   }
   ```

2. Query the created link:

   ```bash
   curl --insecure \
     -H "Authorization: Bearer <token>" \
     http://<dashboard-host>:18083/api/v5/cluster/links/link/link-without-password
   ```

### Expected Behavior

For an optional password that was not configured, the GET API should represent the
unset state (for example, with an empty value, `null`, or by omitting the field).

### Actual Behavior

The GET API returns:

```json
{
  "name": "link-without-password",
  "username": "",
  "password": "******"
}
```

`******` should mean “configured but redacted”; it should not be returned for an
unset password.

### Environment

- **Software version / build**: EMQX Enterprise `6.3.0-rc.2-ga79302ea`
- **Configuration**:

  ```hocon
  cluster.links = [
    {
      name = "link-without-password"
      enable = true
      server = "mqtt://<remote-broker>:1883"
      username = ""
    }
  ]
  ```

zmstone:
The fact that "this peer cluster requires no MQTT password" is in a sense secret information too.
i.e. it makes sense to obfuscate.

PR #18872, opened 2026-09-10
fix(saml): preserve empty SP private key in API response

hjianbo:
Fixes: #18753
Introduced in: N/A

## Summary

When SAML request signing was disabled, the effective configuration contained an empty `sp_private_key`, but `GET /api/v5/sso/saml` returned the redaction placeholder `******`.

The fix keeps the empty value readable only for SAML API configuration responses. Non-empty SAML private keys remain redacted, and all other SSO backends and common redaction behavior are unchanged.

## PR Checklist


zmstone:
suggest to reject, similar: https://github.com/emqx/emqx/issues/18757

PR #18875, opened 2026-09-10
perf(mt): build tenant limiter containers lazily

zmstone:
Fixes:
Release version: 6.3.1, 7.0.0
Introduced in: 5.9.0

## Summary

#18806-#18808 made the limiter containers of ordinary MQTT clients lazy, but clients of a
multi-tenant namespace (`tns` attribute) still got fully connected containers from the
`'channel.limiter_adjustment'` / `'session.limiter_adjustment'` hooks, because
`emqx_mt_limiter` builds its containers with eager `emqx_limiter:connect/1,2` calls. A tenant
client therefore carried the full per-limiter client state at connect regardless of which
limits were configured.

This PR makes those containers lazy the same way, without changing when the hooks run or
what they decide:

* `emqx_limiter_client_container` lazy specs can now carry client options:
  `{lazy, [Id | {Id, #{not_found_mode => close}}]}`. On consume, a close-mode spec whose
  group is gone **denies** (`{limiter_not_found, Id}`) instead of counting as unlimited, so
  namespace deletion keeps its "clients stop being admitted" semantics
  (`emqx_mt_limiter:cleanup_configs/2`). Open-mode specs whose group is gone are skipped
  when the rest of the entry connects, instead of crashing the process on `connect`.
* `emqx_mt_limiter:create_limiter/4` emits `{lazy, Specs}` entries: the zone limiter,
  the tenant limiter (close mode) when the namespace has a tenant config, and the client
  limiter (close mode) or the listener's channel limiter otherwise. The config-presence
  decisions stay at hook time; only the `connect` is deferred, and the `mt_limiter_adjusted`
  trace still fires there.

Semantics: a tenant client materializes its clients on the first PUBLISH / SUBSCRIBE /
delivery after a finite limit exists, with exclusive buckets starting full at that moment —
the same behavior #18806-#18808 gave non-tenant clients.

## PR Checklist

### DIFF
--- a/apps/emqx/src/emqx_limiter/src/emqx_limiter_client_container.erl
+++ b/apps/emqx/src/emqx_limiter/src/emqx_limiter_client_container.erl
@@ -22,11 +22,15 @@
 %% Type declarations
 %%------------------------------------------------------------------------------
 
--type entry() :: emqx_limiter_client:t() | {lazy, [emqx_limiter:id()]}.
+%% A lazy spec names a limiter to connect on first use, optionally with
+%% client options (`not_found_mode => close` denies instead of failing
+%% open when the limiter's group is gone).
+-type lazy_spec() :: emqx_limiter:id() | {emqx_limiter:id(), emqx_limiter:client_options()}.
+-type entry() :: emqx_limiter_client:t() | {lazy, [lazy_spec()]}.
 -type t() :: #{emqx_limiter:name() => entry()}.
 -type reason() :: emqx_limiter_client:reason().
 
--export_type([t/0, entry/0, reason/0]).
+-export_type([t/0, entry/0, lazy_spec/0, reason/0]).
 
 %%--------------------------------------------------------------------
 %% API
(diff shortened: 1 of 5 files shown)


apps/emqx/src/emqx_limiter/src/emqx_limiter_client_container.erl
```diff
@@ -22,11 +22,15 @@
 %% Type declarations
 %%------------------------------------------------------------------------------
 
--type entry() :: emqx_limiter_client:t() | {lazy, [emqx_limiter:id()]}.
+%% A lazy spec names a limiter to connect on first use, optionally with
+%% client options (`not_found_mode => close` denies instead of failing
+%% open when the limiter's group is gone).
+-type lazy_spec() :: emqx_limiter:id() | {emqx_limiter:id(), emqx_limiter:client_options()}.
```
savonarola, line 28:
```suggestion
-type spec() :: emqx_limiter:id() | {emqx_limiter:id(), emqx_limiter:client_options()}.
```
(like `entry`, not always lazy)

savonarola (approved):

apps/emqx/src/emqx_limiter/src/emqx_limiter_client_container.erl
zmstone: can change it to `limiter_spec` in the next pr, but `spec` for type spec is a bit confusing.

PR #18877, opened 2026-09-10
fix(audit): name the actor and target of an RBAC-denied request

### LINKED ISSUE #18821: [5.10.5]Dashboard audit record for RBAC-denied gateway update loses actor and target
## What happened?

An authenticated local Dashboard user with the `viewer` role was correctly denied when attempting to update the JT/T 808 gateway. The request returned HTTP 403 and did not change the gateway configuration.

However, the corresponding audit record cannot identify the authenticated user or the target gateway:

```json
{
  "operation_id": "/gateways/:name",
  "operation_type": "gateways",
  "http_method": "put",
  "http_status_code": 403,
  "operation_result": "failure",
  "from": "dashboard",
  "source": "",
  "source_ip": "<internal-client-ip>",
  "http_request": {
    "method": "put"
  }
}
```

The Dashboard consequently displays an empty value after `User:`. The `http_request` object also lacks `bindings.name = "jt808"`, so the target resource cannot be determined from the record.

This was reproduced consistently during the black-box test of PR #17427. Successful requests from a Dashboard administrator and an API key contained the expected actor and target information.

## What did you expect to happen?

The audit record for an authenticated RBAC-denied request should retain enough information for incident investigation:

- `source` should contain the authenticated local Dashboard username.
- The request target should identify the JT808 gateway, for example through `http_request.bindings.name = "jt808"`.
- The record should continue to contain the operation ID, time, source IP, HTTP status, and failure result without exposing credentials or sensitive request data.

## How can we reproduce it (as minimally and precisely as possible)?

1. Enable audit logging with `log.audit.enable = true`.
2. Create a local Dashboard user with role `viewer`.
3. Log in as that user and obtain a bearer token.
4. Send a valid `PUT /api/v5/gateways/jt808` request with that bearer token.
5. Confirm that the request returns HTTP 403 and does not change the gateway configuration.
6. As an administrator, call `GET /api/v5/audit` and locate the 403 record for `/gateways/:name`.
7. Observe that `from` is `dashboard`, but `source` is empty and `http_request` contains only `method`, without the `name=jt808` binding.

Observed record time in the reproducing run: `2026-09-07T07:49:00.170651+00:00`.

Image

## Anything else we need to know?

- The 403 authorization decision itself is correct.
- Gateway configuration remains unchanged after the denied request.
- Audit JSON serialization and collection queries return HTTP 200.
- A bad API-secret request is separately recorded as HTTP 401 and `from=rest_api`; this report concerns an already authenticated local Dashboard user whose identity should be known.
- Test evidence: EMQX PR #17427 black-box case `TC-17427-010`.

## EMQX version

EMQX Enterprise `5.10.4-rc.1-g298c83e6`, three-node cluster (two core nodes and one replicant).

## OS version

Not collected; the issue is reproducible through the public Dashboard and audit APIs.

## Log files

No server archive attached. The relevant sanitized audit record is included above.

zmstone:
Release versions:
5.8.12, 5.9.4, 5.10.6

Introduced in:
pre-5.8

## Summary

Fixes #18821. The v6 counterpart is #18879; the token and RBAC code differs between the two lines, so that one is a hand port, not a cherry-pick.

An authenticated request that is denied by role-based access control (HTTP 403 `UNAUTHORIZED_ROLE`) was audited without an actor and without a target: `source` was empty and `http_request` held only `method`. A `PUT /gateways/jt808` denied for a `viewer` could not be attributed to anyone.

minirest produces the auth meta and parses the request parameters only after a successful authorization, so nothing on the deny path ever wrote `source` or `bindings`. The identity is known at that point — the token has been verified, or the API key secret has been checked — and was discarded.

Crucial changes:

* `emqx_dashboard:authorize/2` and `api_key_authorize/4` write the actor and the routing bindings into the audit meta through `minirest_handler:update_log_meta/1`, the same hook the dashboard API modules already use. Only the routing bindings are added — no headers and no body.
* `emqx_dashboard_token:verify/2` returns `{error, {unauthorized_role, Username}}`. Widening the return keeps the audit side effect in `authorize/2` next to the other meta production, instead of putting minirest process-dictionary calls into the token module. The API key path needs no such change: the key is already bound at the deny site.
* The audit write is wrapped so that a missing per-request meta cannot change the authorization result. `authorize/2` is also called directly from tests, where minirest never initialised that meta.

401 records are unchanged. Those requests are not authenticated, so there is no actor to name.

REST API impact: `GET /api/v5/audit` records for 403 responses now carry a non-empty `source` and a populated `http_request.bindings`. Additive and backward compatible.

## PR Checklist

https://claude.ai/code/session_01DoGvpRh1mJGj9UonfY3m86

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard.erl
@@ -270,6 +270,8 @@ audit_log_fun() ->
 
 %% dialyzer complains about the `unauthorized_role' clause...
 -dialyzer({no_match, [authorize/2, api_key_authorize/4]}).
+%% ...and about the helper that only those clauses call.
+-dialyzer({nowarn_function, [audit_rbac_denied/3]}).
 
 -endif.
 
@@ -286,7 +288,8 @@ authorize(Req, HandlerInfo) ->
                     {401, 'TOKEN_TIME_OUT', <<"Token expired, get new token by POST /login">>};
                 {error, not_found} ->
                     {401, 'BAD_TOKEN', <<"Get a token by POST /login">>};
-                {error, unauthorized_role} ->
+                {error, {unauthorized_role, Username}} ->
+                    ok = audit_rbac_denied(Req, jwt_token, Username),
                     {403, 'UNAUTHORIZED_ROLE',
                         <<"You don't have permission to access this resource">>}
             end;
@@ -296,6 +299,29 @@ authorize(Req, HandlerInfo) ->
                 <<"Support authorization: basic/bearer ">>
             )
     end.
+
+%% RBAC denies the request after the caller has been authenticated, so the
+%% actor is known. minirest never reaches its own meta-producing code on a
+%% failed authorization, so write the actor and the path parameters into the
+%% audit meta here. Only the routing bindings are added - no headers, no body.
+%%
+%% `update_log_meta/1' merges into the per-request meta that minirest
+%% initialises before it calls this authorizer. Tests call `authorize/2'
+%% directly, without that meta; `badmap' then means there is no request to
+%% audit, which must not change the authorization result.
+audit_rbac_denied(Req, AuthType, Source) ->
+    _ =
+        try
+            minirest_handler:update_log_meta(#{
+                auth_type => AuthType,
+                source => Source,
+                bindings => cowboy_req:bindings(Req)
+            })
+        catch
+            error:{badmap, undefined} ->
+                ok
+        end,
+    ok.
 
 return_unauthorized(Code, Message) ->
     {401,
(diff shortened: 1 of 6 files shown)


apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -297,6 +300,29 @@ authorize(Req, HandlerInfo) ->
...
 
+%% RBAC denies the request after the caller has been authenticated, so the
+%% actor is known. minirest never reaches its own meta-producing code on a
+%% failed authorization, so write the actor and the path parameters into the
+%% audit meta here. Only the routing bindings are added - no headers, no body.
+%%
+%% `update_log_meta/1' merges into the per-request meta that minirest
+%% initialises before it calls this authorizer. Tests call `authorize/2'
+%% directly, without that meta; `badmap' then means there is no request to
+%% audit, which must not change the authorization result.
```
zmstone, line 311: this is a bit hack-ish,
will fix it properly in v6.

savonarola (approved):

PR #18879, opened 2026-09-10
fix(audit): name the actor and target of an RBAC-denied request

zmstone:
Fixes: #18821
Introduced in: pre-5.8

## Summary

This is the v6 counterpart of #18877 (based on `dev-58`). The token and RBAC code differs between the two lines, so the fix was ported by hand rather than cherry-picked.

An authenticated request that is denied by role-based access control (HTTP 403 `UNAUTHORIZED_ROLE`) was audited without an actor and without a target: `source` was empty and `http_request` held only `method`. A `PUT /gateways/jt808` denied for a `viewer` could not be attributed to anyone.

minirest produces the auth meta and parses the request parameters only after a successful authorization, so nothing on the deny path ever wrote `source` or `bindings`. The identity is known at that point — the token has been verified, or the API key secret has been checked — and was discarded.

Crucial changes:

* `emqx_dashboard:jwt_token_bearer_authorize/3` and `api_key_authorize/4` write the actor and the routing bindings into the audit meta through `minirest_handler:update_log_meta/1`, the same hook the dashboard API modules already use. Only the routing bindings are added — no headers and no body.
* `emqx_dashboard_token:verify/3` returns `{error, {unauthorized_role, Message, AdminKey}}`. The admin key is what the success path already records as `source`, so a denied record names the caller the same way an allowed one does. The API key path needs no such change: the key is already bound at the deny site.
* The audit write is wrapped so that a missing per-request meta cannot change the authorization result. `emqx_dashboard:authorize/2` is also called directly from tests, where minirest never initialised that meta.

401 records are unchanged. Those requests are not authenticated, so there is no actor to name.

REST API impact: `GET /api/v5/audit` records for 403 responses now carry a non-empty `source` and a populated `http_request.bindings`. Additive and backward compatible.

## PR Checklist

### DIFF
--- a/apps/emqx_dashboard/src/emqx_dashboard.erl
+++ b/apps/emqx_dashboard/src/emqx_dashboard.erl
@@ -385,6 +385,7 @@ api_key_authorize(Req, HandlerInfo, Key, Secret) ->
                     " path is not permitted">>
             );
         {error, {unauthorized_role, Msg}} when is_binary(Msg) ->
+            ok = audit_rbac_denied(Req, api_key, Key),
             {403, 'UNAUTHORIZED_ROLE', Msg};
         {error, _} ->
             return_unauthorized(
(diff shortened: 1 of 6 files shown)


apps/emqx_dashboard/src/emqx_dashboard.erl
```diff
@@ -415,10 +416,34 @@ jwt_token_bearer_authorize(Req, HandlerInfo, Token) ->
...
+        try
+            minirest_handler:update_log_meta(#{
+                auth_type => AuthType,
+                source => Source,
+                bindings => cowboy_req:bindings(Req)
+            })
+        catch
+            error:{badmap, undefined} ->
+                ok
+        end,
```
keynslug: Nit. Looks a bit brittle, especially with this `error:{badmap, ...}`. Shouldn't this be addressed on the `minirest` side?

zmstone: https://github.com/emqx/minirest/pull/140

keynslug (approved):

Issue #18857, opened 2026-09-09
[5.10.5] Backup Sync accepts invalid interval and timeout values and silently uses defaults


lina50:
### What happened?

The Backup Sync plugin configuration API accepts invalid and non-positive duration values for both `sync.interval` and `sync.timeout`.

For example, the following values are accepted by `PUT /api/v5/plugins/emqx_backup_sync-0.1.0/config` with HTTP 204:

```json
{
  "sync": {
    "interval": "-5s",
    "timeout": "5ssss"
  }
}
```

The invalid values are persisted/returned by the configuration API, but they are not actually used. Runtime code silently substitutes defaults:

- invalid or non-positive `sync.interval` becomes 5 minutes;
- invalid or non-positive `sync.timeout` becomes 30 seconds.

This makes a successful HTTP 204 misleading: an administrator sees the submitted values in configuration while the worker runs with different, undocumented effective values. Typographical errors can therefore go unnoticed and scheduling/request behavior differs from the saved configuration.

Source inspection confirms that `emqx_backup_sync_client:validate_config/1` validates required primary fields and SSL `verify`, but does not validate either duration. `emqx_backup_sync:interval_ms/1` and `emqx_backup_sync_client:timeout_ms/1` independently fall back to constants when parsing fails or the parsed value is not positive. The existing unit test also explicitly expects an invalid interval to become 300000 ms.

### What did you expect to happen?

The configuration API should validate both fields before saving the configuration:

- reject malformed values such as `5ssss`;
- reject zero and negative values such as `0s` and `-5s`;
- return HTTP 400 with a field-specific validation error;
- keep the previously valid configuration unchanged.

If fallback behavior is intentionally retained for backward compatibility, the API should normalize the stored/returned value to the effective default and clearly report that substitution. Silently storing an invalid value while using another runtime value should be avoided.

### How can we reproduce it (as minimally and precisely as possible)?

1. Install and start `emqx_backup_sync-0.1.0` on a secondary cluster and configure valid primary connection details.
2. Read and save the current plugin configuration.
3. Set `sync.interval` to `-5s` and send the complete configuration to:

```text
PUT /api/v5/plugins/emqx_backup_sync-0.1.0/config
```

4. Observe HTTP 204 instead of a validation error.
5. Repeat with `sync.interval = "5ssss"`, `sync.timeout = "-5s"`, and `sync.timeout = "5ssss"`; each request returns HTTP 204.
[...]

PR #18886, opened 2026-09-10
fix(backup_sync): validate sync config and TLS options

### LINKED ISSUE #18856: [5.10.5] Backup Sync logs cert_file_not_found for empty TLS paths while HTTP sync succeeds
### What happened?

On a secondary cluster running the `emqx_backup_sync-0.1.0` plugin from PR #17449, updating the plugin configuration with an HTTP primary URL and `primary.ssl.enable = true` succeeds, but each synchronization emits three error-level logs:

```text
[error] msg: cert_file_not_found, path: []
[error] msg: cert_file_not_found, path: []
[error] msg: cert_file_not_found, path: []
```

The same cycle then completes successfully:

```text
[info] msg: backup_sync_finished, result: #{..., cleanup => #{local => skipped, remote => ok}}, stage: finished
```

The three errors correspond to the empty `cacertfile`, `certfile`, and `keyfile` values. They are false-positive operational errors: the configured URL uses HTTP, no TLS handshake is attempted, and the backup is successfully exported, downloaded, imported, and cleaned up.

Source inspection shows that `emqx_backup_sync_client:client_ssl_options/1` always includes all three file options, converting empty binaries to empty lists. `emqx_tls_lib:to_client_opts/1` then tries to resolve each empty path and logs `cert_file_not_found`.

This creates misleading error-level noise for a successful synchronization. The configuration also accepts the inconsistent combination of an `http://` URL and `ssl.enable = true` without warning or validation.

- Empty optional certificate paths should be omitted before calling `emqx_tls_lib:to_client_opts/1` and should not produce `cert_file_not_found` errors.
- When `base_url` uses `http://`, TLS options should either be ignored without error logs or the API should reject/warn about `ssl.enable = true`.
- A successful synchronization should not emit error-level TLS messages.

1. Install and start `emqx_backup_sync-0.1.0` on a secondary cluster.
2. Configure a reachable HTTP primary and valid API credentials. Credentials below are placeholders:

```json
{
  "primary": {
    "base_url": "http://PRIMARY_HOST:18083/api/v5",
    "api_key": "REDACTED",
    "api_secret": "REDACTED",
    "ssl": {
      "enable": true,
      "verify": "verify_none",
      "server_name_indication": "disable",
      "cacertfile": "",
      "certfile": "",
      "keyfile": ""
    }
  },
  "sync": {
    "interval": "5m",
    "root_keys": ["rule_engine"],
    "table_sets": [],
    "timeout": "30s",
    "retain_remote_backup": false,
    "retain_backup_after_import": true
  }
}
```

3. `PUT` the JSON to `/api/v5/plugins/emqx_backup_sync-0.1.0/config`.
4. Observe three `cert_file_not_found` errors with `path: []` followed by `backup_sync_finished`.

- Tested PR: #17449
- Merge commit: `15c1515e33468a94c8a70e78455a86386dcdc603`
- Branch: `release-510`
- Candidate fix: filter empty `cacertfile`, `certfile`, and `keyfile` values before calling `emqx_tls_lib:to_client_opts/1`; optionally validate URL scheme against `ssl.enable`.
- No Dashboard token, API key, or API secret is included in this report.

### EMQX version

```console
e5.10.4-50-g15c1515e33
```

### OS version

Kubernetes-based shared QA environment; OS details not required to reproduce the configuration-path issue.

### Log files

Relevant sanitized log lines are included above.

hjianbo:
Release versions:
5.10.6

## Summary

Closes #18856 and #18857.

This change fixes two Backup Sync configuration and runtime inconsistencies:

- Validate `sync.interval` and `sync.timeout` as positive durations before the plugin configuration is persisted. Invalid values now return field-specific configuration errors instead of being stored and silently replaced by runtime defaults.
- Omit empty TLS certificate paths before TLS option conversion, and ignore TLS options for `http://` primary URLs. Successful HTTP synchronizations no longer produce false-positive `cert_file_not_found` errors.

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ce/(feat|perf|fix|breaking)-<PR-id>.en.md` files
- [x] Schema changes are backward compatible or intentionally breaking (no schema format change)


plugins/emqx_backup_sync/src/emqx_backup_sync_client.erl
```diff
@@ -362,42 +363,78 @@ http_options(Conf, Timeout) ->
...
+    case {maps:get(<<"enable">>, SSL, false), BaseUrl} of
+        {true, <<"http://", _/binary>>} ->
+            [];
+        {true, _} ->
+            emqx_tls_lib:to_client_opts(client_ssl_options(SSL));
+        {false, _} ->
+            []
     end.
 
 client_ssl_options(SSL) ->
```
zmstone, line 376: emqx_tls_lib:to_client_opts doesn’t handle all these for you?

hjianbo: `to_client_opts/1` is not sufficient here because its empty-value filtering happens after the extractor runs. For the certificate fields, `conf_extract_opt/3` first calls `conf_resolve_path_strict/2`, which calls `resolve_cert_path_for_read_strict/1`. An input `""` is treated as an empty path, emits `cert_file_not_found`, and returns `undefined`; only then does `conf_extract_opt/3` drop the value. I reproduced three error logs for empty `cacertfile`, `certfile`, and `keyfile`, while passing `undefined` produces no logs. Therefore Backup Sync must normalize these optional empty paths before calling the shared TLS helper.

zmstone (commented):
maybe take this opportunity to make the error returns of config checks structured.
For example,
`{error, #{cause => "invalid_config", filed => Filed, value => ...}}`

zmstone (approved):

PR #18910, opened 2026-09-11
test(client): assert on this client's sock_closed trace event

zmstone:
Backport of a06f31c257 (#17590) from dev-510.

`emqx_client_SUITE:t_sock_closed_reason_normal` failed on #18909 (run 34587140860). Its trace held two `sock_closed_normal` events. The extra one came from the previous case, `t_client_attr_from_user_property`: the server closed that client's socket after the next case had started.

The check now filters the trace by this iteration's client id. `t_sock_closed_force_closed_by_client` gets the same filter.

The dev-58 → dev-59 sync carries this to dev-59. dev-510 already has it.


apps/emqx/test/emqx_client_SUITE.erl
```diff
@@ -501,8 +512,16 @@ t_sock_closed_force_closed_by_client(_) ->
                     )
                 end,
                 fun(Trace0) ->
+                    %% Filter by this iteration's unique client id, see
+                    %% t_sock_closed_reason_normal.
                     ?assertMatch(
-                        [#{clientid := ClientId}], ?of_kind(sock_closed_with_other_reason, Trace0)
+                        [#{clientid := ClientId} | _],
```
keynslug: Nit. This also allows more than 1 event to match for unclear reasons. I feel it should be easier to just supply randomness through `per_version_clientid/2`, and keep the assertion dumb.

keynslug (dismissed):

Issue #18442, opened 2026-08-21
Expose a gauge for disconnected session holders instead of requiring connections.count - live_connections.count


zmstone:
## Problem

A session whose client has disconnected but whose `session_expiry_interval` has not yet elapsed is held by its connection process, which stays alive with the socket closed until the client reconnects or the session expires. See `emqx_channel:maybe_shutdown/2`:

```erlang
maybe_shutdown(Reason, _Intent = idle, Channel = #channel{conninfo = ConnInfo}) ->
    case maps:get(expiry_interval, ConnInfo) of
        ?EXPIRE_INTERVAL_INFINITE -> {ok, Channel};
        I when I > 0 -> {ok, ensure_timer(expire_session, I, Channel)};
        _ -> shutdown(Reason, Channel)
    end;
```

This is intended behaviour for persistent sessions. The population of such detached holders is roughly `disconnect rate x time until reconnect or expiry`, so a fleet with flaky connectivity and a long expiry interval can hold a large number of them.

There is no metric that reports this number. Operators can only derive it:

```
detached holders = connections.count - live_connections.count
```

That is not discoverable, is easy to get wrong, and does not appear in any dashboard or alert by default. In [#14482](https://github.com/emqx/emqx/issues/14482) a user reported reaching 24-32% of total connections in this state before identifying the cause, having first had to work out the subtraction themselves.

## Proposal

Export the count directly as a gauge, alongside the existing connection stats.

The two tables that back the existing gauges already give the answer, and `?CHAN_LIVE_TAB` is a strict subset of `?CHAN_CONN_TAB` — an entry is inserted into `?CHAN_CONN_TAB` at registration and removed at unregistration, while `mark_channel_disconnected/1` removes only the `?CHAN_LIVE_TAB` entry and leaves the process and its `?CHAN_CONN_TAB` row in place. So the value is the difference of two sizes that `stats_fun/0` already computes.

`emqx_cm.erl` `?CHAN_STATS`:

```erlang
-define(CHAN_STATS, [
    {?CHAN_TAB, 'channels.count', 'channels.max'},
    {?CHAN_TAB, 'sessions.count', 'sessions.max'},
    {?CHAN_CONN_TAB, 'connections.count', 'connections.max'},
    {?CHAN_LIVE_TAB, 'live_connections.count', 'live_connections.max'},
    {?CHAN_REG_TAB, 'cluster_sessions.count', 'cluster_sessions.max'}
]).
```

[...]

PR #18918, opened 2026-09-11
feat(stats): expose disconnected session holders as a gauge

terry-xiaoyu:
## Summary

A session whose client has disconnected but whose `session_expiry_interval` has not yet elapsed is held by its connection process. The number of such sessions could previously only be derived as `connections.count - live_connections.count`.

This publishes the value directly:

- `apps/emqx/src/emqx_cm.erl`: `stats_fun/0` now also derives `disconnected_connections.count` from the sizes of `?CHAN_CONN_TAB` and `?CHAN_LIVE_TAB`, the two tables already read for the `connections`/`live_connections` gauges. The derived value is clamped at zero: during a session takeover the connection row is deleted before the live row, so the raw difference can be transiently negative.
- `apps/emqx/src/emqx_stats.erl`: registers `disconnected_connections.count` and `emqx_disconnected_connections_count`.
- `apps/emqx_prometheus/src/emqx_prometheus.erl`: exposes it as the Prometheus gauge `emqx_disconnected_connections_count`.
- `apps/emqx_management/src/emqx_mgmt_api_stats.erl` + `rel/i18n/emqx_mgmt_api_stats.hocon`: exposes it on `GET /stats`.

Because the stat is registered in `emqx_stats`, it is also published on `$SYS` and becomes an OpenTelemetry gauge automatically.

## Testing

- New `t_disconnected_connections_stat` in `apps/emqx/test/emqx_cm_SUITE.erl`: register (not counted) -> `mark_channel_disconnected` (+1) -> unregister (back to baseline). It also guards the `stats_fun/0` return value contract.
- `emqx_cm_SUITE`: 40 ok, 0 failed
- `emqx_stats_SUITE`: 5 ok, 0 failed
- `emqx_mgmt_api_stats_SUITE`: 1 ok, 0 failed
- `./scripts/erlfmt -c` clean

Fixes https://github.com/emqx/emqx/issues/18442


apps/emqx/src/emqx_stats.erl
```diff
@@ -71,7 +71,10 @@
         %% Count of connected clients
         'live_connections.count',
         %% Maximum number of connected clients
-        'live_connections.max'
+        'live_connections.max',
+        %% Count of sessions whose client has disconnected but whose
+        %% session expiry interval has not yet elapsed
+        'disconnected_connections.count'
```
zmstone: make it right this time
`disconnected_sessions.count`

terry-xiaoyu: Renamed to `disconnected_sessions.count` (Prometheus `emqx_disconnected_sessions_count`), grouped with the session stats, in `d43e182782`. Also fixed the `emqx_prometheus` vsn bump that `sanity-checks` flagged.

zmstone (approved):

Issue #17878, opened 2026-07-06
Observable metric for PUBLISH messages rejected/dropped by the rate limiter


zmstone:
### Problem

When using publish-side rate limiting, it is hard to verify from metrics that the limiter is actually taking effect.

The incoming publish message rate reflects everything the broker received **before** the limiter is applied, so it does not drop even while the limiter is actively suppressing publishes. For example, with an incoming rate of ~500 msg/s that the limiter reduces to a much lower effective rate, the publish/received rate still reads ~500 msg/s — only the downstream (subscriber) delivery rate visibly drops. That makes it look like the limiter is not working, even when it is.

Digging in, the messages the limiter stops are only **partially** observable:

- For **QoS 0**, they are counted as dropped and are visible in Prometheus (`emqx_messages_dropped_quota_exceeded`).
- For **QoS 1/2**, the publishes are **rejected** (the broker returns a "quota exceeded" reason code), not "dropped". These rejections are counted in a broker packet counter (`packets.publish.quota_exceeded`) that is **not** exported to Prometheus, so they are effectively invisible to Prometheus-based dashboards/alerts.

So there is no straightforward, Prometheus-visible way to answer "how many publishes is the rate limiter rejecting/dropping?" or "what is the effective throughput after limiting vs. the intake before it?"

### Desired outcome

A clear, Prometheus-exported way to observe the rate limiter's effect, for example:

- the count of PUBLISH messages rejected or dropped by the rate limiter, **across all QoS levels** (not just QoS 0); and/or
- an **effective post-limiter throughput** metric (messages actually accepted after limiting) that can be compared against the incoming rate to quantify how much the limiter suppressed.

This would let operators confirm that rate limiting is working and alert on limiter activity directly, instead of inferring it indirectly from subscriber-side rates.

### Environment

- EMQX 5.10.x

zmstone:
workaround: with https://github.com/emqx/emqx/pull/17886

`messages_routed = emqx_messages_received - emqx_packets_publish_quota_exceeded`

PR #18919, opened 2026-09-11
fix(channel): count rate-limited QoS 1/2 publishes as rejected

terry-xiaoyu:
Fixes: #17878
Release version: 7.0.0
Introduced in: 5.9.0

## Summary

Publish-side rate limiting was hard to verify from metrics because the limiter's
rejections were only partly observable:

* QoS 0 publishes stopped by the limiter were counted in
  `messages.dropped.quota_exceeded` (Prometheus:
  `emqx_messages_dropped_quota_exceeded`).
* QoS 1 and QoS 2 publishes are **rejected** with a `quota exceeded` reason code,
  not dropped, and were only counted in `packets.publish.quota_exceeded`, so no
  message-level metric reflected that the limiter was active.

QoS 1 and QoS 2 rejections are now counted in a new
`messages.rejected.quota_exceeded` metric, exported to Prometheus as
`emqx_messages_rejected_quota_exceeded`. QoS 0 keeps using the existing
`messages.dropped.quota_exceeded` metric, and `packets.publish.quota_exceeded`
still counts every QoS level and remains the all-QoS total.

This makes the rate limiter's effect observable per QoS level (dropped for QoS 0,
rejected for QoS 1/2), instead of looking like the limiter is not working while
the incoming publish rate stays flat.

Files touched:

* `apps/emqx/src/emqx_channel.erl` — increment the new metric on the quota-exceeded
  publish path.
* `apps/emqx/src/emqx_metrics.erl`, `rel/i18n/emqx_metrics.hocon` — define the metric.
* `apps/emqx_prometheus/src/emqx_prometheus.erl` — export it to Prometheus
  (`apps/emqx_prometheus` version bumped accordingly).

## Test plan

* `apps/emqx/test/emqx_channel_SUITE.erl`: `t_quota_qos1` and `t_quota_qos2`
  assert that the two rate-limited publishes in each case increment
  `messages.rejected.quota_exceeded` and leave `messages.dropped.quota_exceeded`
  unchanged.
* `t_quota_qos0` still covers the dropped path.
* Ran `apps/emqx/test/emqx_channel_SUITE.erl`: all cases pass.
* Ran `apps/emqx/test/emqx_limiter_SUITE.erl` case `t_metrics_quota_exceeded`: pass.
* Ran `scripts/elvis-check`, `scripts/erlfmt -c`, `scripts/check-i18n-style.sh` and
  `scripts/apps-version-check.exs` locally: pass.

## PR Checklist


apps/emqx/src/emqx_channel.erl
```diff
@@ -778,10 +778,13 @@ handle_publish_error(QoS, Topic, PacketId, RC = ?RC_NOT_AUTHORIZED, Channel) ->
     end;
 handle_publish_error(QoS, _Topic, PacketId, RC = ?RC_QUOTA_EXCEEDED, Channel) ->
     ok = inc_metrics('packets.publish.quota_exceeded', Channel),
+    %% Count the rate-limited message for every QoS, not only QoS 0, so that
+    %% `emqx_messages_dropped_quota_exceeded` answers "how many publishes did
+    %% the rate limiter reject or drop?" regardless of the QoS level.
+    ok = inc_metrics('messages.dropped', Channel),
+    ok = inc_metrics('messages.dropped.quota_exceeded', Channel),
```
zmstone: this is different from the issue description (as requirement)
and different from the PR description too.

QoS > 0 messages are not dropped.
they are rejected.
bumping `dropped` would be confusing.

terry-xiaoyu: Fixed — thanks for the catch. QoS 1/2 publishes are now counted as *rejected*, not dropped:

- QoS 0 stays in `messages.dropped.quota_exceeded` (silently dropped, there is no way to signal the client).
- QoS 1/2 now bump a new `messages.rejected.quota_exceeded` metric, exported to Prometheus as `emqx_messages_rejected_quota_exceeded`.
- `packets.publish.quota_exceeded` still counts every QoS level, so it remains the all-QoS total.

`t_quota_qos1`/`t_quota_qos2` now assert that the rejected metric increments while `messages.dropped.*` stays flat. Happy to collapse this into a single neutral all-QoS metric instead if you'd prefer that over the per-QoS split.

rel/i18n/emqx_metrics.hocon
```diff
@@ -167,6 +167,9 @@ emqx_metrics {
     messages_dropped_receive_maximum {
         desc: "Number of messages dropped due to receive maximum being reached"
     }
+    messages_rejected_quota_exceeded {
+        desc: "Number of messages rejected due to quota being exceeded"
```
zmstone:
```suggestion
        desc: "Number of QoS 1 and QoS 2 messages rejected due to quota being exceeded"
```

terry-xiaoyu: Applied in d839734983 — thanks.

terry-xiaoyu:
Note on `ci(docker): raise the image size limit to 170 MB` (4302184980): the docker size gate started failing on every dev-70 PR after Debian published a trixie point release and `debian:13-slim` was not yet refreshed on Docker Hub. The image grew to amd64 167.9 MB / arm64 166.3 MB against the 160 MB gate. This is the same change already merged as #18938 and pending sync to dev-70 in #18942; it's included here only so this PR is not blocked by the base image drift. It can be dropped once #18942 lands.

Also re-ran the `apps/emqx-3_10` shard, which had failed on an unrelated flaky `emqx_ds_shared_sub_SUITE:declare_implicit.t_cluster_smoke`; it passed on retry. All checks are green now.

zmstone (approved):

PR #18923, opened 2026-09-11
fix(tls): default-certificate follow-ups

zmstone:
Fixes: 
Release version: 7.0.0
Introduced in: N/A

## Summary

Four small changes left over from the default-certificate series (#18813, #18884, #18906).

**Prometheus reports `emqx_cert_expiry_at` for listeners serving the node's default certificate.** No changelog: the default certificate is new in 7.0, not released. `resolve_listener_certfile/3` looked for `certfile` in the listener's own configuration, which such a listener does not have, so no point was produced. It now resolves the configuration with `emqx_tls_lib:default_certs_if_present/1` first, which names the bundle's files when the bundle exists and never generates one. Test: `t_cert_expiry_default_cert`.

**Gateways resolve the configuration they are updated away from without generating.** Every gateway module builds runtime listener configurations for both the old and the new configuration on update and diffs them; resolving the old one went through `ensure_default_certs/1`, which generates the bundle when it is missing. `emqx_gateway_utils_conf:to_rt_listener_configs/5` takes `generate_default_certs => false`, and each of the eight gateway modules passes it at the old-config call site. Same fix `emqx_listeners` got in #18813. Test: `t_remove_listener_does_not_regenerate_default_cert` (verified to fail without the change).

**`emqx_utils_certs` falls back to RSA on a crypto library without `secp256r1`.** `default_key_type/0` returns `ec` when the curve is in `crypto:supports(curves)` and `rsa` otherwise, and is the default for callers that do not choose a key type; the node's default certificate is generated that way, and its `default_tls_certificate_generated` log records `key_type`. A caller that asks for `ec` explicitly on such a library gets `{curve_unsupported, secp256r1}` instead of the bare `badarg` `public_key` raises. A reported-unsupported curve is a capability the node lacks and an RSA certificate serves; a failure during generation is something else and still raises. Test: `t_default_key_type`.

**`emqx:cert_file/1` is deleted.** It built `etc/certs/<file>` paths and had no callers.

## PR Checklist

### DIFF
--- a/apps/emqx_gateway/src/emqx_gateway_utils_conf.erl
+++ b/apps/emqx_gateway/src/emqx_gateway_utils_conf.erl
@@ -70,6 +70,7 @@ This module provides functions make such transformations.
 
 -export([
     to_rt_listener_configs/4,
+    to_rt_listener_configs/5,
     to_rt_listener_ids/2,
     rt_listener_id/1,
     rt_listener_id/2,
@@ -90,6 +91,7 @@ This module provides functions make such transformations.
 -type listener_name() :: atom().
 
 -type esockd_type() :: tcp | ssl | udp | dtls.
+-type rt_opts() :: #{generate_default_certs => boolean()}.
 -type cowboy_type() :: ws | wss.
 
 -type listener_type() :: esockd_type() | cowboy_type().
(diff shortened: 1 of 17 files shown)


apps/emqx_gateway/src/emqx_gateway_utils_conf.erl
```diff
@@ -171,6 +173,19 @@ Transform gateway config + module config + runtime context into runtime listener
...
 to_rt_listener_configs(GwName, GwConfig0, ModConfig0, Ctx) ->
+    to_rt_listener_configs(GwName, GwConfig0, ModConfig0, Ctx, #{}).
+
+-doc """
+Same as `to_rt_listener_configs/4`, with options.
+
+`generate_default_certs` (default `true`): when `false`, a TLS listener with no
+certificate configured is given the node's default certificate only if that
+bundle already exists; nothing is generated. Use it for the configuration a
+gateway is being updated away from, which is read, not served.
```
savonarola, line 184: Difficult to understand

savonarola (approved):

apps/emqx_gateway/src/emqx_gateway_utils_conf.erl
zmstone: clarified here: https://github.com/emqx/emqx/pull/18924/changes#diff-b18ae72c8678fa5d1631cda25e38e33f73c6a36c117bf2cd94420e5811cd9e38R181-R191

PR #18924, opened 2026-09-11
ci: generate the docker-compose test certificates instead of committing them

zmstone:
## Summary

`.ci/docker-compose-file/certs/` held a CA with its private key, a server key pair, a client key pair and three Java keystores, all committed. Anyone with the repository could mint certificates the 17 TLS test services trust. Harmless in CI, but the same class of thing #18906 removed from the release.

- `scripts/ct/gen-compose-certs.sh` generates the set into the same directory under the names `scripts/gen-test-certs.sh` uses everywhere: `cacert.pem`, `cert.pem`, `key.pem`, `client-cert.pem`, `client-key.pem` (the committed set mixed `ca.crt`, `server.crt`, `server.key`, `client.pem`, `client.key`). The host side of every mount, the `COPY` lines of the openldap and pgsql images, and the configs of the two services that mount the whole directory (nats, otel collector) follow the new names; **container-side paths of file mounts are unchanged**. The server certificate carries every TLS service hostname on the compose network as a SAN, plus `toxiproxy` (the committed one had no SANs at all). The CA private key is not kept.
- `scripts/ct/run.sh` runs it before `docker compose up`; `run-emqx.sh` (used by `run_docker_tests`) does the same for its haproxy stack. A set is regenerated when the generator changes, or with `--force`.
- `scripts/gen-test-certs.sh` gains `EXTRA_DNS_SANS` and writes PKCS#1 keys on OpenSSL 3 too, matching the files being replaced.
- The keystores (`server.jks`, `server.p12`, `truststore.jks`) had no consumer: Kafka and Cassandra build their own with the `ssl_cert_gen` service. `ca.key` had none either.
- **Copies of the old set in test fixtures are deleted** (found by certificate fingerprint, not by name): `apps/emqx_auth/test/data/certs`, `apps/emqx_bridge/test/data/certs`, `apps/emqx_exhook/test/data/certs`, `apps/emqx/test/data/certs` — 15 files incl. five private keys. Suites that used them as trust anchor / client cert for a compose service (`emqx_authn_test_lib:client_ssl_cert_opts/0` shared by the mongo/mysql/pgsql/redis TLS auth suites and `emqx_bridge_redis_SUITE`; `emqx_s3_test_helpers:cert_path/1`; the `emqx_offline_messages` plugin's redis TLS group) resolve the generated compose set by repository path, like `emqx_bridge_pulsar`, `emqx_bridge_rabbitmq` and `emqx_dashboard_sso_oidc` already did. Suites that ran a mock HTTPS server on them (`emqx_authn_https_SUITE`, `emqx_authn_jwt_SUITE`, `emqx_bridge_snowflake_streaming_SUITE`) get a runtime certificate from the new `emqx_common_test_helpers:mock_server_certs/2` — RSA CA and key (so `ECDHE-RSA-*` cipher pins keep working and every TLS client accepts the chain) and CN-only for a name of the suite's choosing (so `emqx_authn_https_SUITE` can still expect verification against the connected-to host to fail when no SNI is set). Content-only cases (`emqx_exhook_SUITE`, `emqx_bridge_v2_SUITE`, `emqx_mgmt_api_listeners_SUITE`) read the generic `test_cert/1` set. `apps/emqx_s3/test/certs/ca.crt` deleted too. License, JWT, CRL and OCSP fixtures are unrelated and stay.
- `scripts/test/influx/certs` (orphaned set for a manual script, two private keys) is deleted; the script runs `gen-test-certs.sh`.
- The `to_rt_listener_configs/5` doc in `emqx_gateway_utils_conf` is rewritten after the review comment on #18923.
- `pgsql-tls` `COPY`s the set at image build, after generation.

- Datalayers had a per-service leaf (`datalayers/server.crt` + committed key) signed by the old shared CA, minted because the old shared cert had no SANs and the Arrow Flight client checks the hostname. The generated cert names `datalayers_tls`, so datalayers mounts the shared pair now and `emqx_bridge_datalayers_arrow_SUITE` trusts the shared CA instead of the old-CA copy it kept in its data dir; four files deleted.

Per-service sets (`greptimedb/`, `elastic/`, `keycloak/`) are out of scope here.

CI-only change, no user impact.

## Verification

Local `scripts/ct/run.sh --ci` over every app whose `docker-ct` pulls an affected service (26 apps) plus the `emqx_offline_messages` plugin: all green; after the rename, 11 apps covering each rename path (image `COPY`, directory mounts, in-container client cert, mTLS helper, path-reading suites) re-run green, nats included; datalayers re-run green on the shared pair (127/127); after the fixture removal, `emqx_auth_http` (105 cases), `emqx_auth_jwt` (55), `emqx_exhook` (63), `emqx_bridge_v2_SUITE` (42), `emqx_bridge_snowflake_streaming_SUITE` (7) and `emqx_mgmt_api_listeners_SUITE` (27) re-run green, including the mTLS auth suites, `emqx_bridge_redis`, `emqx_s3`/`emqx_bridge_s3`, pulsar, rabbitmq, dashboard SSO (OIDC via dex, SAML), otel, ldap, datalayers. Two apps could not run on my host for reasons unrelated to this change (`influxdb3` image build dies with SIGILL; Docker Hub rate limit on the nats pull); CI covers them.


.ci/docker-compose-file/certs/README.md
```diff
@@ -1,23 +1,23 @@
...
 
-### How to convert server PEM to JKS Format
+The set, named as `scripts/gen-test-certs.sh` names its output everywhere:
 
-1. Convert server.crt and server.key to server.p12
+- `cacert.pem`: the CA. Test suites that verify a service's certificate trust
+  this.
+- `cert.pem`, `key.pem`: the certificate every TLS service presents. Its
+  subject alternative names cover every service hostname on the compose network
+  and `toxiproxy`; the list lives in `gen-compose-certs.sh`.
```
savonarola: Maybe put the list of hostnames here (in the `certs` dir) as a `.txt` file? To make additions more obvious.

scripts/ct/gen-compose-certs.sh
```diff
@@ -0,0 +1,91 @@
...
+redis-cluster-tls-3
+redis-cluster-tls-4
+redis-cluster-tls-5
+redis-cluster-tls-6
+redis-sentinel-tls
+redis-sentinel-tls-master
+redis-sentinel-tls-slave
+redis-tls
+toxiproxy
+'
```
savonarola: Are the `_`-containing hosts actually used? They seem invalid.

zmstone:
@savonarola thank you for the review.
had to rebase for ci
see the last two commits for changes since your last review.

ieQu1 (approved):

PR #18930, opened 2026-09-12
perf(channel): drop will_msg, conn_props and subscriptions from the channel-info cache

zmstone:
Fixes: 
Release version: 7.0.0
Introduced in: pre-5.8

## Summary

The `emqx_channel_info` ETS row cached three attributes nobody reads from the table:

- `will_msg` (the full will `#message{}`, converted to a map),
- `conninfo.conn_props` (the raw MQTT 5 CONNECT properties),
- `session.subscriptions` (the whole subscriptions map).

The clients API strips all three (`emqx_mgmt_api_clients:format_channel_info/3`), and every other reader
(`emqx_mgmt_cli`, `emqx_cm:all_channels_stream/1` → eviction agent, `emqx_cm_takeover`, durable-session offline
info, `emqx_setopts`, `emqx_mgmt:do_call_client/2`) takes a handful of scalars from `conninfo` / `clientinfo`.

The row is rewritten on every SUBACK and UNSUBACK (`return_sub_unsub_ack/2`) and on each deliver while
disconnected, so the `subscriptions` copy cost O(K²) entries for a client subscribing to K topics, and durable
sessions rebuilt the map from DS state on each refresh.

Changes:

- The connection modules insert `cached_info/1` into the table. `emqx_channel:cached_info/1` omits the three
  attributes; `emqx_session:cached_info/1` is `info/1` without `subscriptions` (durable sessions no longer rebuild
  the map from DS state on every refresh). `emqx_channel:handle_cast({keepalive, _}, _)` and
  `emqx_eviction_agent_channel` use the cached form for the rows they write.
- `emqx_channel:info/1` and `ConnMod:info(Pid)` are unchanged and still return the complete map, so a plugin that
  needs `will_msg`, `conn_props` or `subscriptions` for one client can read them from the channel process.
- Tests: `emqx_channel_SUITE:t_chan_info` asserts the cached shape and that `info/1` is complete;
  `emqx_client_SUITE:t_chan_info_cache_omits_large_attrs` checks the same end to end through
  `emqx_cm:get_chan_info/1` and `ConnMod:info(Pid)`.

**Breaking for plugins** that read the table through `emqx_cm:get_chan_info/1,2`, `emqx_cm:lookup_client/1` or
`emqx_mgmt:lookup_client/2` and match on `will_msg`, `conninfo.conn_props` or `session.subscriptions`. Hooks, the
REST API, `emqx_broker:subscriptions/1` and `ConnMod:info(Pid)` are unaffected. Targeted at 7.0.0 for that reason,
with a `breaking-18930` changelog entry.

Not changed on purpose: `conn_props` stays in the channel state because the `client.ping` hook passes the live
`conninfo` to the rule engine (`$events/client_ping` has a `conn_props` column) and enhanced re-auth reads
`'Authentication-Method'` from it. `will_msg` stays a full `#message{}` in the state; the publish path refreshes
the timestamp, and a compact form would save ~250 B per client with a will.

### Measurements

10 000 idle connections, one node, measured on dev-63 @ f6ac260d00 before / with this change after (the cached-row code is identical on dev-70). Bytes per connection.
"ETS B/row" is the compressed table size divided by row count; "flat" is `erts_debug:flat_size` of one row's
Info map; "heap" is `process_info(memory)` of one hibernated channel process (unchanged by the PR by design).

| scenario | ETS B/row before | ETS B/row after | row Info flat before → after | dropped from row (flat) | channel heap |
|---|---|---|---|---|---|
| MQTT 3.1.1, no props, no will | 937 | 913 | 1608 → 1512 | session.subscriptions 40 | 9032 |
| MQTT 5, no props, no will | 937 | 914 | 1608 → 1512 | — | 9032 |
| MQTT 5, 10 User-Properties (32 B values) | 1409 | 922 | 2752 → 1512 | conn_props 1168 | 13920 |
| MQTT 5, 100 User-Properties | 5641 | 922 | 12832 → 1512 | conn_props 11248 | 55312 |
| MQTT 5, will with 4 KB payload | 1114 | 922 | 2152 → 1512 | will_msg 544 (payload is a shared refc binary) | 13920 |
| MQTT 5, will with 32 KB payload | n/a | 921 | n/a → 1512 | will_msg (same as 4 KB: payload not copied) | 13920 |
| MQTT 5, will 4 KB + 5 will User-Properties | 1369 | 921 | 2736 → 1512 | will_msg 1128 (props 608) | 13920 |
| MQTT 5, 50 subscriptions | 3887 | 924 | 11888 → 1512 | session.subscriptions 10368 (~207 B each) | 34696 |

Node-level: at 20 k connections the table goes from 18.7 MB to 18.3 MB for plain clients, from 77.7 MB to 18.5 MB for
clients with 50 subscriptions, and from 112.8 MB to 18.4 MB for clients sending 100 User-Properties. The remaining
~920 B/row is `conninfo` (520 B flat: peername/sockname tuples, clientid, username, conn_mod, proto, timers),
`clientinfo` (~480 B), `sockinfo` (224 B, peername/sockname again), `session` (176 B) and the 1520 B stats proplist,
all compressed by ETS to roughly half.

## PR Checklist


apps/emqx/src/emqx_channel.erl
```diff
@@ -184,11 +185,27 @@
...
 
+-doc """
+Channel attributes cached in the `emqx_channel_info` ETS table.
+
+The cache omits `will_msg`, `conninfo.conn_props` and `session.subscriptions`.
+They are the largest attributes, they grow with client input, and no reader of the
+table uses them. Read them from the channel process with `info/1` or `info/2`.
+""".
+-spec cached_info(channel()) -> emqx_types:infos().
+cached_info(#channel{conninfo = ConnInfo, session = Session} = Channel) ->
```
keynslug: Q: Why not just replace the current `info/1` implementation with this?

apps/emqx/src/emqx_session.erl
```diff
@@ -671,6 +683,16 @@ run_terminate_hooks(ClientInfo, Reason, Session) ->
...
     maps:from_list(info(?INFO_KEYS, Session)).
 
+-doc """
+Session attributes cached in the `emqx_channel_info` ETS table.
+
+Same as `info/1` without `subscriptions`: building and copying that map costs
+O(number of subscriptions) on every cache refresh, and no reader of the table uses it.
+""".
+-spec cached_info(t()) -> emqx_types:infos().
+cached_info(Session) ->
```
keynslug: Q: Same here, perhaps just make `info/1` return small but useful amount of information? E.g.:
* `subscriptions_cnt`
* `inflight_cnt`
* `awaiting_rel_cnt`
* `mqueue_len`
* ...

apps/emqx/src/emqx_channel.erl
zmstone: the intention is to keep public api compatible. gen call based info retrieval still gets all fields

apps/emqx/src/emqx_session.erl
zmstone: good point, will fix this

apps/emqx/src/emqx_channel.erl
keynslug: I understand the intention, but if it's considered a breaking change anyway why not break `info/1` as well, for the sake of simplicity? All those fields that are now excluded from ETS table are still accessible through `info/2` anyway.

zmstone: got it.

apps/emqx/test/emqx_client_SUITE.erl
```diff
@@ -455,6 +456,46 @@ assert_receive_maximum(MaxAwaitingRel, ServerReceiveMaximum, Config) ->
     ?assertMatch(#{'Receive-Maximum' := ServerReceiveMaximum}, Props),
     ok = emqtt:disconnect(C).
 
+-doc """
+The `emqx_channel_info` table omits the will message, the CONNECT properties and the
+subscriptions map; the channel still returns them per key.
+""".
+t_chan_info_cache_omits_large_attrs(Config) ->
```
keynslug: Nit. Now I think it mostly repeats `t_chan_info`. If we're to keep, I'd suggest making it more future-proof: make more-or-less strictly match on the expected structure of chaninfo, not on what is expected to be missing.

keynslug (dismissed):

keynslug (dismissed):

zmstone:
@keynslug trimming only channel-info cache in this pr.

zmstone:
> @keynslug trimming only channel-info cache in this pr.

emqx_session:info/1 stays complete

keynslug (approved):

PR #18949, opened 2026-09-14
fix(rocketmq): report a missing topic instead of dropping messages with case_clause

zmstone:
Release versions:
5.8.12, 5.9.4, 5.10.5

Introduced in:
pre-5.8

## Summary

Port of emqx/emqx#18945 (dev-60) to dev-58; the connector code is the same on both lines.

A RocketMQ action whose topic does not exist on the broker, and whose broker does not auto-create topics (for example an Alibaba Cloud instance, where topics are created per namespace in the console), dropped every message with `{unrecoverable_error, {case_clause, {error, {{shutdown, {start_producer_failed, ...}}, ...}}}}` and logged the whole query, payload included, for each message. The same deployment showed the connector alarm raised and cleared every two minutes.

Root causes, all in the client library (fixed in emqx/rocketmq-client-erl#39, released as 0.7.3):

- `rocketmq_producers_sup:ensure_present/4` had no clause for a failed child start, so a producer whose topic has no route raised `case_clause`; the connector's catch-all turned it into `unrecoverable_error`.
- With a namespace configured, the auto-create default topic was requested as `<namespace>%TBW102`, which never exists. Auto-create could not work in a namespace, and the failure remark named `TBW102` instead of the user's topic.
- The name server closes a connection that carried no request for 120 s (`serverChannelMaxIdleTimeSeconds`). The client sends nothing on its own, so with no producer running the socket dropped every two minutes, `get_connection_state/1` answered `connecting` for one poll, and the resource manager raised an alarm and cleared it at the next poll.

Connector changes (`emqx_bridge_rocketmq_connector`):

- `get_producers/5` returns `{ok, _} | {error, _}`. A start error becomes an `unrecoverable_error` with the readable reason, the same policy as the Kafka action for a templated topic that resolves to an unknown topic.
- `on_get_channel_status/3` runs `rocketmq:check_topic/2` for a topic without placeholders on every health check. A missing topic makes the action `disconnected` with `Topic "T" not found in namespace "N", and the broker does not auto-create topics. Create the topic on the broker. Name server reply: ...`. The buffer worker then holds and retries messages, and the action recovers on the first health check after the topic is created. The check also keeps the name server connection busy. The error is a plain binary on purpose: `{unhealthy_target, _}` would stop further health checks of the channel.
- `rocketmq_connector_do_query_failed` logs action, topic and reason, not the query.

Tests: the CI broker auto-creates topics, so `rocketmq_client:get_routeinfo_by_topic/2` is mocked to answer "no route". A new `topic_not_found` group holds `t_topic_not_found_static_topic` (status, message, the message is held and delivered after the mock is removed) and `t_topic_not_found_templated_topic` (`topic_not_found` unrecoverable error, action stays connected).

The client is pinned at `v0.7.3` (emqx/rocketmq-client-erl#39) in `apps/emqx_bridge_rocketmq/rebar.config` and `mix.exs`. In 0.7.3 only `TOPIC_NOT_EXIST` (17) counts as a missing topic; any other name server error for the topic keeps the action `connecting` and the send fails with `route_lookup_failed`.

## PR Checklist


zmstone:
non-critical, no backport to v5.
fixed in v6: https://github.com/emqx/emqx/pull/18945

Issue #18944, opened 2026-09-14
[6.3.1]Audit log stores SCRAM client_proof without redaction


lina50:
### What happened?

When audit logging is enabled, a successful `POST /api/v5/login/verify` SCRAM-SHA-256 request is persisted with the exact `client_proof` value in `http_request.body.client_proof`.

The proof is not a reusable bearer credential because the SCRAM challenge is one-time and nonce-bound. However, retaining the proof together with the public SCRAM transcript (`username`, client/server nonce, salt, and iteration count) can support offline password guessing. It should therefore be treated as sensitive authentication material.

This was found while testing #18777 on a real EMQX Enterprise environment. The full `GET /api/v5/audit` response remained valid JSON, and passwords, API secrets, and JWTs were not exposed. The leak was limited to the SCRAM client proof in this test.

### What did you expect to happen?

Audit records should redact or omit `http_request.body.client_proof`, for example:

```json
{
  "client_proof": "******"
}
```

The audit entry should otherwise retain its operation ID, actor/source, timestamp, result, HTTP status, and other non-sensitive diagnostic fields.

### How can we reproduce it (as minimally and precisely as possible)?

1. Enable audit logging by setting `log.audit.enable = true`.
2. Call `POST /api/v5/login/challenge` with an existing Dashboard username and a valid URL-safe client nonce.
3. Calculate the SCRAM-SHA-256 client proof from the returned salt, iteration count, and server nonce.
4. Save the exact proof locally and call `POST /api/v5/login/verify`:

```json
{
  "challenge_id": "<challenge-id>",
  "combined_nonce": "<client-nonce><server-nonce>",
  "client_proof": "<base64-client-proof>"
}
```

5. Confirm that login succeeds.
6. Call `GET /api/v5/audit?page=1&limit=100` and locate the record whose `operation_id` is `/login/verify`.
7. Compare `http_request.body.client_proof` with the proof saved in step 4.

Actual result: the values are identical.

Automated reproduction from the QA workspace:

```console
python3 tests/emqx/pull_requests.emqx/18777/testscritps.py \
  --host <emqx-host> \
  --only TC-18777-014
```

The test reports:

```text
CaseFailure: sensitive value category leaked in audit JSON: SCRAM client proof
```

The reproduction deliberately does not print or attach the real proof.

### Anything else we need to know?

- Related feature PR: #18777
- Audit records were tested through both a local Dashboard user and an administrator API key.
[...]

PR #18953, opened 2026-09-14
fix(audit): redact SCRAM verification body

hjianbo:
Fixes: #18944
Release version: 6.3.2, 7.0.0
Introduced in: N/A

## Summary

Redact the entire `/login/verify` request body before audit events are written to the audit file or database. This keeps SCRAM proofs, MFA tokens, and the rest of the authentication exchange out of persisted audit records while preserving operation metadata.

Add regression coverage for the audit boundary, the stored Mnesia record, and the serialized `/audit` response.

## PR Checklist
- [x] Change log is not required because this fixes the unreleased SCRAM login feature in 6.3.1


apps/emqx_dashboard/src/emqx_dashboard_audit.erl
```diff
@@ -173,6 +173,8 @@ non_empty_map(_) -> undefined.
 %% before the generic key-name redaction runs.
 redact_request_body(#{operation_id := <<"/license">>}, _Body) ->
     <<"******">>;
+redact_request_body(#{operation_id := <<"/login/verify">>}, _Body) ->
+    <<"******">>;
```
zmstone, line 177: [P2] Also redact credentials copied into validation failures

Masking only `http_request.body` leaves another copy on the schema-validation failure path. I reproduced this at a963deee with a `/login/verify` body containing `client_proof: {"proof":"review-synthetic-proof"}` and string values for `challenge_id` and `combined_nonce`. The request returns 400, but its audit event contains both `body = <<"******">>` and `failure.message = <<"{\"value\":{\"proof\":\"review-synthetic-proof\"},...}">>`.

The [request validator serializes the rejected value into its error message](https://github.com/emqx/emqx/blob/a963deee4633e36ef6564acb7b11b8202b66b19a/apps/emqx_dashboard/src/emqx_dashboard_swagger.erl#L371-L376). Minirest copies that message into `failure`. [log_meta/3 retains this field](https://github.com/emqx/emqx/blob/a963deee4633e36ef6564acb7b11b8202b66b19a/apps/emqx_dashboard/src/emqx_dashboard_audit.erl#L73-L89), and the generic redactor cannot inspect credentials embedded in a binary. [Both persistence sinks receive it](https://github.com/emqx/emqx/blob/a963deee4633e36ef6564acb7b11b8202b66b19a/apps/emqx_audit/src/emqx_audit.erl#L88-L106).

Sanitize the audit failure details for this credential-bearing operation as well, while retaining the status and error code. Add a malformed-request regression that checks the entire stored record and serialized /audit response for the submitted proof; checking only `http_request.body` misses this copy.

zmstone:
need rebase.

zmstone (approved):

Issue #18714, opened 2026-09-02
[6.3.0] Dashboard plugin upload returns ALREADY_INSTALLED for an unauthorized package


Le9Wang:
### What happened?

When a plugin package exists in the plugin installation directory but has not been successfully installed, the Dashboard upload API returns `ALREADY_INSTALLED` before checking whether installation has been authorized.

This prevents the Dashboard from showing the required `plugins allow` instruction.

### Reproduction Steps

1. Copy `emqx_plugin_template_a-5.0.0.tar.gz` to `/opt/emqx/plugins/`.
2. Run the CLI install command without granting permission:

   ```bash
   emqx ctl plugins install emqx_plugin_template_a-5.0.0
   ```

3. The CLI returns:

   ```json
   {
     "result": "not_ok",
     "name_vsn": "emqx_plugin_template_a-5.0.0",
     "cause": "not_allowed",
     "action": "ensure_installed"
   }
   ```

4. Open the Dashboard plugin management page.
5. Upload the same plugin package and click Install.

### Expected Behavior

The API should check the installation permission and return a clear `403 FORBIDDEN` response instructing the user to run:

```bash
emqx ctl plugins allow emqx_plugin_template_a-5.0.0
```

Alternatively, the API should distinguish between an installed plugin and an orphan `.tar.gz` package.

### Actual Behavior

The API returns:

```json
{
  "code": "ALREADY_INSTALLED",
  "message": "[\"plugins/emqx_plugin_template_a-5.0.0.tar.gz\"] already installed"
}
```

However, the plugin is not actually installed and does not appear in the Dashboard.

### Test Case

```python
await ssh.arun(
    "emqx ctl plugins install emqx_plugin_template_a-5.0.0",
    expected_exited=0,
)

# Dashboard upload/install returns ALREADY_INSTALLED
```

### Environment

- **Software version / build**: EMQX 6.3.0
- **Affected component**: Plugin management HTTP API / Dashboard plugin upload
- **Related change**: PR #17884
- **Plugin**: `emqx_plugin_template_a-5.0.0`
- **OS / Platform**: N/A

PR #18957, opened 2026-09-14
fix(plugins): apply the install allow gate to orphan packages

terry-xiaoyu:
Release versions:
5.9.4, 5.10.6
5.10.5

Introduced in:
pre-5.8

## Summary

Fixes https://github.com/emqx/emqx/issues/18714

`POST /plugins/install` replied `ALREADY_INSTALLED` when a `.tar.gz` was already present in the plugin install directory but had not been unpacked (for example copied by hand, or left behind by a failed install). That check ran **before** the installation allow-list, so the Dashboard could not show the `emqx ctl plugins allow ...` instruction and the package could never be installed. Getting the state of the install directory right needed three parts, all of them in this PR:

1. **An orphan tarball is not an installation.** Anything that is not a usable installation now goes through the same allow gate as any other install: without a grant the upload is rejected with `403 FORBIDDEN` and the `emqx ctl plugins allow <name-vsn>` hint; with a grant the uploaded bytes overwrite the stale tarball and the package is installed.
2. **A manifest without its application files is not an installation either.** A valid `release.json` only proves that the metadata is readable: the unpack may have been interrupted after the manifest had been written. Such a directory used to be reported as installed (`400 ALREADY_INSTALLED`, and `ensure_installed_from_tar/2` returned `ok` without unpacking the missing applications). Completeness is now checked against the package content: every application in `rel_apps` must have its `<app>.app` resource file, and the beam files that resource file declares.
3. **The files of a running plugin are never deleted.** A failed metadata read does not prove that the directory is an interrupted installation: if a *running* plugin's `release.json` becomes unreadable, purging its directory would leave the plugin running from replaced or missing code. The runtime state is checked independently of `release.json` (the applications whose code directory is under the plugin's install directory), and the replacement is refused with `400 BAD_PLUGIN_INFO`, message `plugin_is_in_use: <name-vsn>. stop the plugin first`. `emqx ctl plugins stop <name-vsn>` now also works when `release.json` can not be read (the applications to stop are found in the install directory instead of in the metadata), so the way out of a refused upload is: stop the plugin, upload the package again.

A genuinely installed plugin still returns `400 ALREADY_INSTALLED`.

Where the bug came from, and which branches are affected, is described in the comment below.

**Behaviour changes:**

- `emqx_plugins:ensure_started/1` (and the `ensure_existed` RPC) reach `ensure_installed_from_tar/2`, so a configured plugin whose install dir is manifest-only is now repaired (purged and re-unpacked from the local package, or copied from a peer node) instead of failing to load. Without a package anywhere the start fails with `plugin_not_found` and the leftovers are gone; the plugin configuration (cluster config and `data_dir`) is not affected.
- Applications that are merely *loaded* (stopped, not running) are unloaded before the installation is replaced, so the package that is unpacked afterwards is loaded cleanly.
- `ensure_existed/1` still folds every install error into its existing `404 Plugin Not Found`, including the new refusal; left unchanged.

**Tests:**

- EUnit `emqx_plugins_tests` (no external dependencies, **30/30**): leftover and broken metadata are purged and unpacked again; a manifest-only installation is detected and repaired; an installation whose declared beams are missing is repaired; a *running* plugin's files are left untouched (verified to fail on the unfixed code, where the directory was purged); a loaded-but-stopped application is unloaded before the replacement; a plugin with an unreadable `release.json` can be stopped and then repaired.
- CT `emqx_plugins_SUITE` (**25/25**) and `emqx_plugins_config_SUITE` (**2/2**): `t_install_recovers_from_leftover_dir`, `t_install_recovers_from_incomplete_install`, `t_install_refuses_in_use_plugin` (refused, then stop → install), plus the pre-existing suite.
- CT `emqx_mgmt_api_plugins_SUITE` (**17/17**): `t_install_orphan_package_requires_allow`, `t_install_over_leftover_install_dir`, `t_install_over_broken_install_dir`, `t_install_over_complete_install_dir` (a complete installation still yields `ALREADY_INSTALLED`, checked against the real package), `t_install_over_manifest_only_install_dir`, and `t_install_running_plugin_broken_metadata` (refused with the hint, files untouched, then stop → upload → installed).

## PR Checklist

- [x] Change log for changes visible by users has been added to `changes/ce/fix-18957.en.md`
- [x] Schema changes are backward compatible or intentionally breaking (no schema changes in this PR)


zmstone (dismissed):

apps/emqx_plugins/src/emqx_plugins_fs.erl
```diff
@@ -213,11 +213,32 @@ install_from_local_tar(NameVsn, InstallValidator) ->
...
+            %% installation: a directory without a readable `release.json'.
+            %% That is not an installation, so purge it and unpack the package
+            %% into a clean directory instead of letting stale files shadow the
+            %% new content.
+            ?SLOG(warning, #{
+                msg => "purging_incomplete_plugin_installation",
+                name_vsn => NameVsn,
+                reason => Reason
+            }),
+            case purge_installed(NameVsn) of
```
zmstone, line 232: [P2] Check the runtime state before purging a broken installation

A failed metadata read does not establish that this is an interrupted installation. If a running plugin's `release.json` becomes unreadable or corrupt, the new HTTP error branch permits an allowed upload and this branch deletes its directory while its application remains running. I reproduced this with a started synthetic application, a malformed `release.json`, and a resource file in its plugin directory: the real validator returned an error, `ensure_installed_from_tar/2` removed the resource and replaced the directory, and `running_status/1` still returned `running` afterward.

[The fresh-install entry point explicitly relies on callers to exclude installed/running plugins](https://github.com/emqx/emqx/blob/bf2d5ebe03568aa0c0fd4a4d8f70527fe0322a32/apps/emqx_plugins/src/emqx_plugins.erl#L252-L257). It does not stop or unload the application. [The loader also skips reloading an already-loaded application with the same version](https://github.com/emqx/emqx/blob/bf2d5ebe03568aa0c0fd4a4d8f70527fe0322a32/apps/emqx_plugins/src/emqx_plugins_apps.erl#L171-L181), so old code can continue running against replaced or missing files.

Check the running/loaded state independently of `release.json` before deleting the directory on each target node. Refuse automatic recovery until the plugin is stopped and unloaded, or use an explicit lifecycle that safely stops it before replacement. Add a regression asserting that a failed metadata read cannot cause a running plugin's files to be purged.

apps/emqx_management/src/emqx_mgmt_api_plugins.erl
```diff
@@ -533,24 +533,20 @@ get_plugins() ->
...
-                {true, TarGzs} ->
-                    %% TODO
-                    %% What if a tar file is present but is not unpacked, i.e.
-                    %% the plugin is not fully installed?
-                    {400, #{
-                        code => 'ALREADY_INSTALLED',
-                        message => iolist_to_binary(io_lib:format("~p already installed", [TarGzs]))
-                    }}
-            end;
         {ok, _} ->
```
zmstone, line 536: [P2] Recognize partial extraction even when release.json is valid

The interrupted-install recovery still misses an extraction that writes `release.json` before stopping while writing the application files. `describe/2` succeeds for a directory containing only a valid manifest: [its checks validate the metadata and app-name strings, not the presence of the declared applications](https://github.com/emqx/emqx/blob/bf2d5ebe03568aa0c0fd4a4d8f70527fe0322a32/apps/emqx_plugins/src/emqx_plugins_info.erl#L120-L160). This branch therefore still returns `ALREADY_INSTALLED` and prevents the recovery upload.

I reproduced this with a manifest-only plugin directory and a complete tarball containing its `.app` file. The upload returned 400 `ALREADY_INSTALLED`; calling `ensure_installed_from_tar/2` with the real installation validator also returned `ok` without extracting the missing app. The new filesystem check reuses [the same metadata-only validator](https://github.com/emqx/emqx/blob/bf2d5ebe03568aa0c0fd4a4d8f70527fe0322a32/apps/emqx_plugins/src/emqx_plugins.erl#L707-L713), so both layers consider the partial installation complete.

Extend the recovery checks to establish that extraction completed, such as a completion marker written after successful installation or checks for the required application files. Add a case where `release.json` survives but an application file is missing. Keep the running-plugin protection separate from this completeness check.

zmstone (commented):
reviewed by agent

zmstone (dismissed):

terry-xiaoyu:
### Root cause and affected branches

Two separate states of the plugin install directory are mishandled, and they were introduced at different points.

**1. Orphan tarball / manifest-only directory — present since dev-58 (5.8), hence `Introduced in: pre-5.8`.**

- dev-58: `emqx_mgmt_api_plugins:upload_install/2` decides with
  `filelib:wildcard(filename:join(install_dir(), AppName) ++ "-*.tar.gz")`
  (`?VSN_WILDCARD` is `"-*.tar.gz"`). That pattern matches the orphan tarball
  `<install_dir>/<AppName>-<vsn>.tar.gz`, so the upload is answered with `400 ALREADY_INSTALLED` before
  `emqx_plugins:is_allowed_installation/2` is ever called. `emqx_plugins:ensure_exists_and_installed/1`
  also treats any existing plugin directory as installed (`filelib:is_dir(plugin_dir(NameVsn))`), so a
  manifest whose applications were never unpacked is never repaired either.
- dev-59 and later: `ef3b0ea251 refactor(plugins): improve module interfaces` reshaped exactly the same
  checks — `emqx_plugins:is_package_present/1` in `upload_install/2` (orphan tarball) and
  `emqx_plugins_fs:is_installed/1` in `ensure_installed_from_tar/2` (directory exists). Same behaviour,
  different code.

**2. Purging a running plugin's directory — introduced by the earlier commits of this PR (dev-59 only).**

The recovery branch added to `emqx_plugins_fs:ensure_installed_from_tar/2` purges the install directory
whenever the metadata can not be read, without checking whether the plugin's applications are still
running. The runtime check (`emqx_plugins_apps:load/2` skips an already loaded application with the same
version, so old code keeps running against replaced or missing files) is what the review comment was
about, and it is fixed in this PR.

| branch | orphan tarball | manifest-only dir | running plugin files |
| --- | --- | --- | --- |
| dev-58 | `400 ALREADY_INSTALLED` | `400 ALREADY_INSTALLED` | not purged by the fresh-install path |
| dev-59 .. dev-63 | `400 ALREADY_INSTALLED` | `400 ALREADY_INSTALLED` | purged (this PR's earlier commits) |
| this PR (dev-59) | `403` / install via allow gate | `403` / install via allow gate | refused, stop first |

Both states of (1) are reachable on 5.8+ without the CLI allow gate: a `.tar.gz` copied into the plugin
install directory, or a left-over package from a failed install, is enough. The `emqx ctl plugins install`
step in the issue refers to a 6.x CLI, but the Dashboard upload path shows the same symptom on 5.8+.

Verified by reading `origin/dev-58` (`upload_install/2`, `ensure_exists_and_installed/1`,
`pkg_file_path/1`, `read_file_fun/3`) and by running the wildcard above against a directory holding only
`<AppName>-<vsn>.tar.gz`.

zmstone (approved):

apps/emqx_plugins/src/emqx_plugins_fs.erl
terry-xiaoyu: Fixed in `45534e0e53`.

- `emqx_plugins_fs:prepare_replacement/1` now checks the runtime state before the recovery purges anything, **independently of `release.json`**: it looks at the applications whose code directory lives under the plugin's install directory (`emqx_plugins_apps:running_apps_from/1` / `loaded_apps_from/1`, resolved with `code:lib_dir/1`), so it keeps working when the metadata is unreadable. It does not assume that the plugin name is the name of one of its applications.
- If any of those applications is **running**, the replacement is refused: `{error, #{msg => "plugin_is_in_use", hint => <<"stop the plugin first">>}}` (logged as `refusing_to_replace_running_plugin`), and the HTTP API answers `400 BAD_PLUGIN_INFO` with `plugin_is_in_use: <name-vsn>. stop the plugin first`. Nothing is deleted.
- Applications that are merely **loaded** (stopped, not running) are unloaded before the purge, so the package unpacked afterwards is loaded cleanly instead of the old modules staying in the code server (`load_plugin_app/4` skips an already loaded application of the same version).
- The same guard protects the recovery path of `emqx_plugins:ensure_installed/1` (boot, `emqx ctl plugins install`, `ensure_existed` RPC).
- Way out: `emqx ctl plugins stop <name-vsn>` now also works when `release.json` is unreadable — `emqx_plugins:do_ensure_stopped/1` falls back to `emqx_plugins_apps:stop_and_unload_loaded/1`, which finds the applications in the install directory. Stop the plugin, then upload the package again.
- Regression tests: EUnit `ensure_installed_from_tar_refuses_in_use_plugin_test` (a really started synthetic application + broken `release.json`: the call is refused and neither the resource file nor the `ebin` is touched), plus `ensure_installed_from_tar_unloads_loaded_app_test` and `ensure_stopped_without_readable_metadata_test`; CT `emqx_plugins_SUITE:t_install_refuses_in_use_plugin` and `emqx_mgmt_api_plugins_SUITE:t_install_running_plugin_broken_metadata` (refused with the hint, files untouched, then stop → upload → installed). All of them were verified to fail on the unfixed code, where the running plugin's directory was purged.

apps/emqx_management/src/emqx_mgmt_api_plugins.erl
terry-xiaoyu: Fixed in `45534e0e53`.

- `emqx_plugins_fs:install_state/2` classifies the install directory as `installed | incomplete | absent`. `installed` requires the metadata to be readable **and** every application declared in `rel_apps` to have been extracted: `is_extraction_complete/1` checks the `<app>.app` resource file of each declared application **and** the beam files that resource file declares.
- `ensure_installed_from_tar/2` dispatches on that state, so a manifest-only directory is `incomplete` and is purged and unpacked again instead of returning `ok` without extracting the missing application.
- `emqx_mgmt_api_plugins:upload_install/2` uses `emqx_plugins:install_state/1` instead of `describe/2`, so a manifest-only directory goes through the allow gate (`403 FORBIDDEN` with the `emqx ctl plugins allow ...` hint, or installs with a grant) instead of `400 ALREADY_INSTALLED`. A complete installation still returns `400 ALREADY_INSTALLED` (`t_install_over_complete_install_dir` checks that against the real plugin package).
- `emqx_plugins:ensure_installed/1` (normal mode: boot, CLI `plugins install`, `ensure_existed` RPC) uses the same classification, so those paths repair a manifest-only installation too.
- The running-plugin protection is kept separate from this completeness check (`is_in_use/1` / `prepare_replacement/1` vs `is_extraction_complete/1`), as suggested.
- Tests: EUnit `ensure_installed_from_tar_recovers_manifest_only_install_test` and `ensure_installed_from_tar_recovers_missing_beam_test`; CT `emqx_plugins_SUITE:t_install_recovers_from_incomplete_install` and `emqx_mgmt_api_plugins_SUITE:t_install_over_manifest_only_install_dir`. Both EUnit cases fail on the pre-fix code (which returned `ok` without unpacking).

PR #18958, opened 2026-09-14

zmstone:
Release versions:
5.8.12, 5.9.4, 5.10.6

Introduced in:
pre-5.8

## Summary

A client that has not sent CONNECT can send any other packet type. The frame parser buffered the announced body, up to `mqtt.max_packet_size`, before the channel rejected the packet. This PR rejects such a packet at its fixed header, before any body byte is buffered.

This is the same fix as https://github.com/emqx/emqx/pull/18828 on `dev-63`, adapted to the v5 parser:

- `emqx_frame:expect_connect/0` sets a flag in the connection process dictionary. `emqx_connection` and `emqx_ws_connection` call it where they build the parser.
- The `parse/2` header clause throws `unexpected_packet_before_connect` for a non-CONNECT type while the flag is set. The parser erases the flag when a CONNECT parses.
- v5 uses the process dictionary so that the parse state and its options map keep their shape.
- QUIC data streams parse in their own processes and never carry CONNECT, so they do not set the flag.
- `emqx_channel` needs no change. A map frame error in `idle` state already shuts the connection down without a CONNACK.

The `update_opts/2` refactor from https://github.com/emqx/emqx/pull/18828 is not ported. On this branch the connection modules build the parser once at connection init and never re-initialize it on a zone change, so there is no parser state to carry across.

## PR Checklist


apps/emqx/src/emqx_frame.erl
```diff
@@ -161,6 +176,21 @@ parse(Bin, {
...
+    ok;
+validate_connect_first(Type) ->
+    case erlang:get(?PD_EXPECT_CONNECT) of
+        undefined ->
+            ok;
+        true ->
+            ?PARSE_ERR(#{
+                cause => unexpected_packet_before_connect,
+                header_type => emqx_packet:type_name(Type)
+            })
```
thalesmg: nit for consistency:


```suggestion
        true ->
            ?PARSE_ERR(#{
                cause => unexpected_packet_before_connect,
                header_type => emqx_packet:type_name(Type)
            });
        _ ->
            ok
```

thalesmg (dismissed):

thalesmg (dismissed):

zmstone:
rebased to resolve conflict.

PR #18959, opened 2026-09-14

zmstone:
Release versions:
6.0.4, 6.1.5, 6.2.4, 6.3.1, 7.0.0

Introduced in:
pre-5.8

## Summary

A client that has not sent CONNECT can send any other packet type. The frame parser buffered the announced body, up to `mqtt.max_packet_size`, before the channel rejected the packet. This PR rejects such a packet at its fixed header, before any body byte is buffered.

This is the same fix as https://github.com/emqx/emqx/pull/18828 on `dev-63`. The v5 line gets it in https://github.com/emqx/emqx/pull/18958.

- The parser options carry an `expect_connect` flag. While it is set, a non-CONNECT fixed header throws `unexpected_packet_before_connect`. The check runs in `parse/2` and in the whole-frame `parse_complete/2`.
- `emqx_connection`, `emqx_socket_connection` and `emqx_ws_connection` set the flag where they build the parser. The parser clears it when a CONNECT parses.
- `emqx_channel` needs no change. A frame error in `idle` state already shuts the connection down without a CONNACK.

The `update_opts/2` refactor from https://github.com/emqx/emqx/pull/18828 is not ported. On this branch the connection modules build the parser once at connection init and never re-initialize it on a zone change, so there is no parser state to carry across.

## PR Checklist


thalesmg (approved):
do we need to update quic_connection too?

zmstone:
> do we need to update quic_connection too?

quic streams are tricky, maybe for single stream yes.
worth fixing separately anyways.

PR #18962, opened 2026-09-14
fix(resource): wait 6 s for a manual health check

zmstone:
`emqx_resource_manager:health_check/1` is the manual, test-only resource health check. It waited `?T_OPERATION` (5 s) for the resource manager process.

greptimedb-ingester-erl `v0.2.5-emqx.3` sends its health check straight on the gRPC channel with a 5 s deadline. With a stalled TLS connection, the manual call timed out first, and `health_check/1` returned `{error, timeout}` instead of the resource status. `emqx_bridge_greptimedb_SUITE` `grpcv1_tls.t_get_status` failed this way in CI on the dev-510 sync PR #18952, three attempts in a row.

- Add `?T_MANUAL_HEALTH_CHECK` (6 s) and use it in `health_check/1` only.
- `channel_health_check/2` and all other operations keep `?T_OPERATION` (5 s). Production code calls `channel_health_check/2`, so its timeout does not change.

Local `emqx_resource_SUITE` on the verification host fails some timing-sensitive query-counter cases with timetraps with and without this change (61/64 with it, 60/64 without it on dev-58). Those cases do not call `health_check/1`.


thalesmg (approved):

zmstone:
docker size will be fixed here: https://github.com/emqx/emqx/pull/18960

PR #18963, opened 2026-09-14
fix(api_key): default optional desc and enable when creating a key

### LINKED ISSUE #18713: [6.3.0]POST /api/v5/api_key returns 500 when required desc is missing
## Summary

`POST /api/v5/api_key` returns HTTP 500 `INTERNAL_ERROR` when the request omits the required `desc` field. This should be rejected by request validation with HTTP 400 `BAD_REQUEST`.

## Environment

- EMQX: 6.3.0-rc.1-g0cbc3b65
- Endpoint: `POST /api/v5/api_key`

## Steps to reproduce

Send an API key request with `description` instead of the supported `desc` field, or omit the description field:

```json
{
  "name": "pr18696_api_key",
  "description": "test",
  "role": "viewer",
  "enable": true
}
```

## Actual result

```text
HTTP/1.1 500 Internal Server Error
```

```json
{
  "code": "INTERNAL_ERROR",
  "message": "error, {badmatch,#{<<\\"enable\\">> => true,<<\\"expired_at\\">> => infinity,<<\\"name\\">> => <<\\"pr18696_api_key\\">>,<<\\"role\\">> => <<\\"viewer\\">>}}, ..."
}
```

The log contains:

```text
2026-09-02T02:32:00.399408+00:00 [warning] pid: <0.22664.0>, reason: {badmatch,#{<<"enable">> => true,<<"expired_at">> => infinity,<<"name">> => <<"pr18696_api_key">>,<<"role">> => <<"viewer">>}}, path: /api_key, stacktrace: [{emqx_mgmt_api_api_keys,api_key,2,[{file,"emqx_mgmt_api_api_keys.erl"},{line,316}]},{minirest_handler,apply_callback,4,[{file,"/emqx/deps/minirest/src/minirest_handler.erl"},{line,192}]},{minirest_handler,handle,2,[{file,"/emqx/deps/minirest/src/minirest_handler.erl"},{line,73}]},{minirest_handler,init,2,[{file,"/emqx/deps/minirest/src/minirest_handler.erl"},{line,32}]},{cowboy_handler,execute,2,[{file,"/emqx/deps/cowboy/src/cowboy_handler.erl"},{line,37}]},{cowboy_stream_h,execute,3,[{file,"/emqx/deps/cowboy/src/cowboy_stream_h.erl"},{line,310}]},{cowboy_stream_h,request_process,3,[{file,"/emqx/deps/cowboy/src/cowboy_stream_h.erl"},{line,299}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,333}]}], exception: error
```

The handler pattern-matches `<<\\"desc\\">> := Desc0` after the schema has discarded the unknown `description` field.

## Expected result

Return HTTP 400 `BAD_REQUEST` with a clear validation message indicating that `desc` is required or that `description` is unsupported. The request must not produce an internal `badmatch` or HTTP 500.

Using the correct field works:

```json
{
  "name": "pr18696_api_key",
  "desc": "test",
  "role": "viewer",
  "enable": true
}
```

This returns HTTP 200.

terry-xiaoyu:
### What changed

`POST /api/v5/api_key` no longer returns HTTP 500 when the request body omits the optional `desc` (or `enable`) field.

### Why

The request body schema is a plain field list, so `emqx_dashboard_swagger:check_request_body/5` checks each field as an independent root: unknown fields are silently dropped, and an optional field without a `default` is absent from the checked body. The handler pattern-matched `#{<<"desc">> := Desc0, <<"enable">> := Enable}` and crashed with `badmatch` (reported in #18713).

`desc` and `enable` are optional by contract:

- the schema declares both as `required => false`;
- the Dashboard create form initialises `desc: ""` / `enable: true` and has no required rule for the note;
- the documented REST example omits `desc`.

### How

- `desc` defaults to `<<>>` (the same value the Dashboard sends when the note is left empty) and `enable` defaults to `true`; `name` stays mandatory in the schema.
- `PUT /api_key/:name` is untouched and keeps its partial-update semantics.

### Tests

- `t_create_optional_fields`: body with only `name`; the issue's exact `description` typo body; a missing `name` is still rejected with 400.
- `t_update_optional_fields`: a PUT omitting `desc`/`enable` preserves the persisted values.
- Full `emqx_mgmt_api_api_keys_SUITE` passes 49/49; the negative control (fix stashed) reproduces the exact `badmatch` → HTTP 500.

Fixes #18713


terry-xiaoyu:
**Re: the review finding above — "The unsupported `description` field is still accepted and discarded instead of returning HTTP 400"**

Thanks. We did evaluate switching `POST /api_key` to strict unknown-field validation and decided to keep the current behaviour in this fix. Rationale:

1. **It is the framework-wide contract for this request-body form.** `'requestBody'` here is a field list, so `emqx_dashboard_swagger:check_request_body/5` validates each field as an independent root and `hocon_tconf:filter_by_roots/3` (`maps:with/2`) drops everything else; only the few `hoconsc:ref/1`-style bodies (e.g. `POST /alarms/force_deactivate`) run `hocon_tconf:check_unknown_fields/3` and answer 400. Making this one endpoint the exception is a user-visible contract change on a stable release branch, and is out of scope for the #18713 crash fix.
2. **A strict shape also has to rule on the read-only fields of the same object** (`api_key`, `api_secret`, `created_at`, `expired`), which clients can currently round-trip into a create request. Rejecting those is a second, separate behaviour change.
3. **The 500 is gone and the declared contract is preserved.** `desc` has been optional since the endpoint was introduced (`required => false`, the Dashboard posts `desc: ""`, and the documented example omits it), so an omitted note is legitimately served as an empty note. The body from the issue (`{"name", "description", "role", "enable"}`) no longer takes the endpoint down.
4. The changelog now states explicitly that request-body fields outside the API key schema (such as `description`) are ignored by validation, so the missing note is not a surprise.

Verified on a real `emqx/emqx:5.10.4` instance: the issue's body returned `500 {badmatch, ...}` before the fix and returns `200` with `desc: ""` now, and the payload the shipped dashboard sends (`{name, desc, enable, role, scopes[, expired_at]}`) is unaffected.

If strict unknown-field rejection is wanted as a deliberate contract change, it is better done separately (and then applied to the other write endpoints as well); the `PUT` body is intentionally left lenient because it doubles as a partial update.

Also: the branch was rebased onto `dev-510` (`9e0ce1fbc7`) to pick up `07d1e9774f` (*ci(docker): raise the image size limit to 350 MB*), which was the only failing check in the previous run.

Unrelated, but found while reviewing this endpoint: `PUT /api_key/:name` silently resets `role` to `administrator` when `role` is omitted — filed as #18968.

zmstone (approved):

Issue #18752, opened 2026-09-03
[6.3.0] Dashboard submits unsupported oauth2 field for HTTP SCRAM authentication


Le9Wang:
### What happened?

When creating an HTTP authentication provider with the SCRAM mechanism in the Dashboard, the generated request includes an `oauth2` field.

The backend schema for `authn:scram_restapi_post` does not support this field, so the creation request is rejected with `BAD_REQUEST`.

### Reproduction Steps

1. Start EMQX 6.3.0-rc.2-ga79302ea (Enterprise).
2. Open Dashboard and navigate to **Access Control > Authentication > Create Authentication**.
3. Select **SCRAM** as the mechanism.
4. Select **HTTP Server** as the backend.
5. Configure the HTTP authentication provider and enable/configure OAuth2 in the displayed form.
6. Submit the configuration.

### Expected Behavior

The generated request should pass the backend schema validation.

### Actual Behavior

The Dashboard submits an `oauth2` object in the request body:

```json
{
  "backend": "http",
  "mechanism": "scram",
  "oauth2": {
    "enable": true
  }
}
```

The API rejects the request:

```json
{
  "code": "BAD_REQUEST",
  "message": "{\"unmatched\":\"enable,max_retries,...\",\"unknown\":\"oauth2\",\"reason\":\"unknown_fields\",\"path\":\"root\",\"matched_type\":\"authn:scram_restapi_post\",\"kind\":\"validation_error\"}"
}
```

### Test Case

Manual Dashboard reproduction using the steps above.

### Environment

- **Software version / build**: EMQX 6.3.0-rc.2-ga79302ea (Enterprise)
- **Affected module**: Dashboard / Authentication
- **OS / Platform**: N/A

PR #18964, opened 2026-09-14
fix(authn): accept oauth2 for scram_restapi authenticators

terry-xiaoyu:
Fixes https://github.com/emqx/emqx/issues/18752

### Background

The Dashboard lets users enable OAuth2 (client credentials) for the **SCRAM + HTTP** authenticator and sends an `oauth2` object in the create request. The backend schemas `authn:scram_restapi_get` / `authn:scram_restapi_post` did not declare an `oauth2` field, so the request was rejected with `BAD_REQUEST`:

```json
{"unknown":"oauth2","reason":"unknown_fields","matched_type":"authn:scram_restapi_post"}
```

### Root cause

OAuth2 client-credentials support (#17944, commit `1a6034e75f`) added `emqx_connector_oauth2_schema:oauth2_field()` to `emqx_authn_http_schema` and `emqx_authz_http_schema` but missed `emqx_authn_scram_restapi_schema`. The runtime was already fully wired: `emqx_authn_scram_restapi:create_state/2` delegates to `emqx_authn_http:create_state/2`, which validates the OAuth2/header combination and forwards `oauth2` to `emqx_bridge_http_connector`, where the token manager is registered and `Authorization: Bearer <token>` is injected into the user lookup request.

### Changes

- `apps/emqx_auth_http/src/emqx_authn_scram_restapi_schema.erl`: add `oauth2_field()` right after `headers` in both `scram_restapi_get` and `scram_restapi_post`, matching the HTTP authn/authz schemas.
- `apps/emqx_auth_http/test/emqx_authn_scram_restapi_SUITE.erl`:
  - `t_oauth2_client_credentials/1` runs the full MQTT SCRAM-SHA-512 handshake for both GET and POST lookups with OAuth2 enabled, against a local HTTP server serving the token endpoint, and asserts the `Authorization: Bearer <token>` header reaches the user lookup.
  - `t_create_oauth2_authorization_header_conflict/1` asserts a manually configured `authorization` header is rejected at runtime.
- `apps/emqx_auth_http/test/emqx_authn_scram_restapi_test_server.erl`: also serve an `/auth/token` route.

### Tests

`env SUITES=emqx_authn_scram_restapi_SUITE make apps/emqx_auth_http-ct` → 11 ok, 0 failed.

No hot-upgrade impact: additive optional schema field; no process state, mnesia table or exported function changes.


apps/emqx_auth_http/src/emqx_authn_scram_restapi_schema.erl
```diff
@@ -46,12 +46,14 @@ select_union_member(_Value) ->
 fields(scram_restapi_get) ->
     [
         {method, #{type => get, required => true, desc => ?DESC(emqx_authn_http_schema, method)}},
-        {headers, fun emqx_authn_http_schema:headers_no_content_type/1}
+        {headers, fun emqx_authn_http_schema:headers_no_content_type/1},
+        emqx_connector_oauth2_schema:oauth2_field()
```
terry-xiaoyu, line 50: Nit (non-blocking): placement right after `headers` matches `emqx_authn_http_schema` / `emqx_authz_http_schema` and the convention documented in `emqx_connector_oauth2_schema:oauth2_field/0`. I also checked that `common_fields()` (which appends `proplists:delete(pool_type, emqx_bridge_http_connector:fields(config))`) does not already carry an `oauth2` field, so there is no duplicate field here.

apps/emqx_auth_http/test/emqx_authn_scram_restapi_SUITE.erl
```diff
@@ -399,6 +453,66 @@ set_user_handler(Username, Password, Extra0) ->
...
+%% the user lookup endpoint on the same test server, and reports the
+%% `authorization' header seen on the user lookup request back to the test.
+oauth2_handler(Method, Username, Password, Token, TestPid) ->
+    Oauth2Path = emqx_authn_scram_restapi_test_server:oauth2_token_path(),
+    fun(Req0, State) ->
+        case cowboy_req:path(Req0) of
+            Oauth2Path ->
+                {ok, Body, Req1} = cowboy_req:read_body(Req0),
+                Form = uri_string:dissect_query(Body),
+                ?assert(lists:member({<<"grant_type">>, <<"client_credentials">>}, Form)),
```
terry-xiaoyu, line 466: Nit (non-blocking): these `?assert`s execute in the cowboy request process, so on failure the handler crashes and the test only surfaces the failure via the 1s `receive ... after` timeout in `test_oauth2_client_credentials/1`, losing the assertion context. This matches the existing handler style in this suite (match expressions there crash the same way), so it is fine as-is; an alternative is to send the check result back to `TestPid` and assert in the test process.

zmstone (approved):

PR #18965, opened 2026-09-14
feat: implement gcp pubsub consumer using grpc

### LINKED ISSUE #18563: Consider gRPC StreamingPull for the GCP PubSub consumer

Add a gRPC `StreamingPull` transport for the GCP PubSub **consumer** (source), as an alternative to the current REST unary pull.

### Why is this needed?

The consumer currently drives a unary REST loop. `emqx_bridge_gcp_pubsub_consumer_worker:do_pull_async/1` issues `POST /v1/{subscription}:pull` and re-issues it on every reply, with a separate `POST /v1/{subscription}:acknowledge` for acks.

Three properties of that design are worth improving.

**Long polling is emulated by a request timeout tied to the ack deadline.** From `do_pull_async/1`:

```erlang
%% We should link the request TTL to the ack deadline: after the latter, the pull
%% request _should_ be disconnected by the server as expected.  But, if the former is
%% shorter than the ack deadline used in the pull lease, we'll start a new pull
%% request that might timeout because the older request is still the "owner" of the
%% lease.
AckDeadlineMS = erlang:convert_time_unit(AckDeadlineSeconds, second, millisecond),
ReqOpts = #{request_ttl => AckDeadlineMS},
```

The comment describes the coupling well: the pull request TTL cannot be chosen independently of the ack deadline, because the two are entangled through lease ownership. `StreamingPull` removes the entanglement — the stream stays open, the server pushes as messages arrive, and the ack deadline is a per-message property rather than a bound on the transport.

**Every pull is a fresh request/response cycle.** Latency is bounded below by request setup, and throughput is shaped by `pull_max_messages` (default 100) times the round-trip rate. `StreamingPull` delivers messages as they are published, without a poll interval.

**Acks are a second, independent request.** `StreamingPull` carries `ack_ids` back on the same bidirectional stream, which removes the separate `:acknowledge` call and the failure modes where a pull succeeds and its ack does not.

`StreamingPull` is also what Google's own client libraries use by default for subscribers; the REST pull is documented as the simpler, lower-throughput option.

### Suggested implementation

The infrastructure already exists in the tree — this would not be introducing gRPC to EMQX.

- `grpc`, `grpcbox` and `gpb` are already dependencies.
- `emqx_bridge_zerobus` runs a long-lived **bidirectional** gRPC stream with reopen-on-error, sequence tracking and health reporting: see `emqx_bridge_zerobus_stream_writer_worker` (writer side) and `emqx_bridge_zerobus_async_receiver_worker` (receive side, `grpc_client:recv/2` in a helper process). That worker pair is close to the shape a `StreamingPull` subscriber needs, in reverse.
- `emqx_bridge_bigtable` also consumes a gRPC stream via `grpc_client:recv/2`.
- Authentication can be reused as-is: `emqx_bridge_gcp_pubsub_client` already mints service-account JWTs and handles attached-SA tokens via `emqx_connector_jwt`. gRPC would carry the same bearer token in call metadata instead of an HTTP header.

Rough shape:

1. Generate the `google.pubsub.v1.Subscriber` service stubs with `gpb`, as zerobus does for its own service.
2. A per-subscription worker holding an open `StreamingPull` stream, modelled on the zerobus writer/receiver split: one process owning the stream, a helper doing the blocking `recv`.
3. Map `StreamingPullResponse` messages onto the existing message-handling path, so rule-engine integration and the source schema stay unchanged.
4. Send acks as `StreamingPullRequest.ack_ids` on the same stream, replacing the separate `:acknowledge` call.
5. Honour the server's flow-control fields, and surface stream errors through the existing health-check/status reporting.

### Open questions

- **Transport selection.** New connector type, or a `transport = rest | grpc` field on the existing GCP PubSub consumer connector? The latter keeps one connector and lets existing sources switch, but the two transports do not have identical configuration surfaces.
- **Config compatibility.** `pull_max_messages`, `pull_retry_interval` and `ack_retry_interval` are pull-loop concepts with no direct `StreamingPull` equivalent. Decide whether they are ignored (and warned about), repurposed as flow-control hints, or hidden when gRPC is selected.
- **Ordering and redelivery.** Confirm behaviour matches the REST path for message ordering keys and for redelivery after a stream break, so switching transport is not a silent semantic change.
- **Emulator/CI coverage.** The existing consumer tests run against the PubSub emulator over REST. Check the emulator supports `StreamingPull` well enough for CI, or plan how the new path gets covered.
- **Scope.** Producer side is out of scope here; this is about the consumer only.

thalesmg:
Fixes https://github.com/emqx/emqx/issues/18563

Release version: 7.0.0

## PR Checklist


apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
```diff
@@ -0,0 +1,860 @@
...
+                details => Msg
+            }),
+            ?keep_state_actions(
+                Data0,
+                [?state_timeout(?RETRY_PULL_TIMEOUT, #retry_pull{})]
+            );
+        {permission_denied, Details} ->
+            %% unhealthy
+            Msg = append_error_detail(~"Permission denied", Details),
+            set_health(SourceResId, {?status_disconnected, {unhealthy_target, Msg}}),
```
zmstone, line 592: no retry can change health state back?

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
```diff
@@ -0,0 +1,860 @@
...
+
+client_id(Data) ->
+    #{?source_res_id := SourceResId, ?idx := Idx} = Data,
+    IdxBin = integer_to_binary(Idx),
+    NodeBin = atom_to_binary(node(), utf8),
+    <<NodeBin/binary, ":", IdxBin/binary, ":", SourceResId/binary>>.
+
+grpc_meta(State) ->
+    #{?auth_ctx := #{auth_config := AuthConfig}} = State,
+    Token = emqx_bridge_gcp_pubsub_client:get_authorization_token(AuthConfig),
```
zmstone: if this call raise (badmatch error tuple), it crashes supervisor.

handle token error, and enter disconnected state.

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
```diff
@@ -0,0 +1,860 @@
...
+
+append_error_detail(Msg, ~"") ->
+    Msg;
+append_error_detail(Msg, Details) when is_binary(Details) ->
+    <<Msg/binary, ": ", Details/binary>>;
+append_error_detail(Msg, _Details) ->
+    Msg.
+
+pull_recv_async_active(Stream, Data0) ->
+    Handle0 = grpc_client:async_install_receiver(Stream, #{mode => active}),
```
zmstone, line 668: should we care to demonitor

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
```diff
@@ -0,0 +1,860 @@
...
+        ?request_ttl => RequestTTL,
+        ?source_res_id => SourceResId,
+        ?subscription_resource => SubscriptionResource,
+        ?topic_resource => TopicResource
+    },
+    {ok, ?s_update_subscription, Data}.
+
+terminate(_Reason, _State, Data) ->
+    #{?pool := Pool, ?idx := Idx} = Data,
+    gproc_pool:disconnect_worker(Pool, {Pool, Idx}),
```
zmstone, line 163: care to call grpc_client:close_async here?

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
```diff
@@ -0,0 +1,860 @@
...
+    } = Data,
+    %% remember to update `update_mask` in `update_subscription_req` if we ever start
+    %% using other fields here.
+    #{
+        name => SubscriptionResource,
+        topic => TopicResource,
+        ack_deadline_seconds => AckDeadline
+    }.
+
+streaming_pull_req(Data) ->
```
zmstone, line 779: agent suggesttion:
The StreamingPullRequest sets neither max_outstanding_messages nor max_outstanding_bytes. Google documents a
value of 0 as no limit, and gun does not apply HTTP/2 flow control on behalf of the consumer, so a subscription
backlog lands in the worker mailbox as fast as the network delivers it while process_msg/2 runs the hooks
synchronously. The HTTP consumer bounded this with pull_max_messages. Suggest setting both fields from
configuration with a conservative default. I could not demonstrate the effect with the emulator, which delivers
about 60 messages per second to one stream.

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_sup.erl
```diff
@@ -0,0 +1,249 @@
...
+        grpc_opts := GRPCOpts,
+        url := URL0
+    } = Opts,
+    SupFlags = #{
+        strategy => rest_for_one,
+        intensity => 10,
+        period => 5
+    },
+    URL = str(URL0),
+    %% TODO: handle url error more gracefully (or in the schema)
```
zmstone: fix this todo?

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_connector_schema.erl
```diff
@@ -0,0 +1,139 @@
...
+    Field == "put_connector";
+    Field == "post_connector"
+->
+    emqx_connector_schema:api_fields(Field, ?CONNECTOR_TYPE, fields(connector_config));
+fields("config_connector") ->
+    emqx_connector_schema:common_fields() ++ fields(connector_config);
+fields(connector_config) ->
+    [
+        {url,
+            emqx_schema:servers_sc(
```
zmstone, line 56: this schema allows multi-host, but we can only accept single for gcp pubsub

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_connector_schema.erl
```diff
@@ -0,0 +1,139 @@
...
+                default => <<"10s">>, desc => ?DESC("connect_timeout")
+            })},
+        {pool_size,
+            mk(pos_integer(), #{
+                default => 1,
+                importance => ?IMPORTANCE_HIDDEN,
+                desc => ?DESC("pool_size")
+            })}
+    ] ++
+        emqx_connector_schema_lib:ssl_fields(#{enable_by_default => true}) ++
```
zmstone, line 76: validate http scheme + ssl.enable flag consistency and reject early.

reject at config check is preferred, reject at runtime is acceptable.

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_connector_schema.erl
zmstone: we previously received reports pointing out the risk of sending tokens to controlled endpoints (wif in this case)

if possible, we should limit the url pattern (like limiting the token aud)

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
```diff
@@ -0,0 +1,860 @@
...
+    maybe
+        {value, {eos, Trailers}, Rest} ?= lists:keytake(eos, 1, Resp),
+        {true, Rest, Trailers}
+    end.
+
+do_grpc_open_impl(Def, Metadata, Opts) ->
+    try
+        grpc_client:open(Def, Metadata, Opts)
+    catch
+        exit:{noproc, _} ->
```
zmstone: dead code.
grpc-erl exits with `noproc`

changes/ee/feat-18965.en.md
```diff
@@ -0,0 +1 @@
+Added new Connector and Source that ingests data from GCP PubSub via gRPC.
```
zmstone: name both types of the rest / grpc connectors.

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
thalesmg: it will.  and then likely will switch back to the same status, being only in a short time window connected again.  unless it's really fixed.  there are a few tests observing that.

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
thalesmg: I don't think so, same as zerobus.

we keep it open for as long as it may live, and only match on the active handle.

changes/ee/feat-18965.en.md
thalesmg: the change here is only the grpc one.

zmstone: i mean name the new type, and as a reference (in addition to `old-type`) should make it easy to understand.

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
zmstone: `state_timeout` with `#retry_pull{}` calls do_pull,

but set health is called before do_pull here: https://github.com/emqx/emqx/pull/18965/changes#diff-968952d6eda7b74651e429843a903a1df11b0ee16951581f450a21ab3a1f8090R520

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_connector_schema.erl
thalesmg: the token audience is already hardcoded for gcp

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
thalesmg: yes, that was what I said.

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_sup.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_connector_schema.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_connector_schema.erl
thalesmg: :checkered_flag:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
thalesmg: :checkered_flag:

changes/ee/feat-18965.en.md
thalesmg: :checkered_flag:

apps/emqx_bridge_gcp_pubsub/src/emqx_bridge_gcp_pubsub_consumer_grpc_worker.erl
thalesmg: as previously mentioned, the time window where the source would be set to connected to then find out again that it has no permissions should be very tiny and really rare to observe.

still, changed it.  by starting the subscription check again we eliminate it.

:checkered_flag:

zmstone (approved):

PR #18973, opened 2026-09-15
feat(mqtt): limit CONNECT user properties

zmstone:
Release versions:
5.8.12, 5.9.4, 5.10.6

Introduced in:
pre-5.8

## Summary

This backports the CONNECT User Property count limit from emqx/emqx#18829.

- Add `mqtt.max_connect_user_properties` with a default of 100. Set it to `infinity` to disable the limit.
- Count CONNECT properties and Will properties separately. Reject an excess pair before parsing it.
- Keep non-CONNECT property parsing and CONNECT packet-size behavior unchanged.
- Keep dev-58 hot-patch compatibility. Missing configuration uses the default of 100 for new connections. Existing parser state without the option remains unlimited and unchanged.

PR #18958 merged into `dev-58` before this PR was opened. The rebase combined three overlap blocks:

- `emqx_connection` and `emqx_ws_connection`: initialize the parser with the User Property limit, then call `emqx_frame:expect_connect/0`.
- `emqx_frame`: keep both the CONNECT-first process-dictionary key and the independent temporary User Property counter keys.

The configuration schema addition is backward compatible, while the new finite default is an intentional behavior change documented in a breaking changelog entry. The schema adds a field to `/api/v5/configs` responses.

## PR Checklist


rel/i18n/emqx_schema.hocon
```diff
@@ -1393,6 +1393,16 @@ mqtt_max_packet_size.desc:
 mqtt_max_packet_size.label:
 """Max Packet Size"""
 
+mqtt_max_connect_user_properties.desc:
+"""Maximum number of 'User-Property' pairs accepted in a CONNECT packet. Default: 10.
```
keynslug: Nit. Default is already in the schema, looks odd to duplicate it here.

apps/emqx/src/emqx_schema.erl
```diff
@@ -3620,6 +3621,14 @@ mqtt_general() ->
                     desc => ?DESC(mqtt_max_packet_size)
                 }
             )},
+        {"max_connect_user_properties",
+            sc(
+                hoconsc:union([infinity, non_neg_integer()]),
+                #{
+                    default => ?DEFAULT_MAX_CONNECT_USER_PROPERTIES,
```
keynslug, line 3628: Q: This is a breaking change, right?

zmstone: changed to breaking-
and raised default limit to 100 so it's less likely to break any real use case.

keynslug (approved):

PR #18976, opened 2026-09-15
fix(greptimedb): reject invalid typed point values

hjianbo:
Release versions:
5.10.6

## Summary

Related report: https://emqx.atlassian.net/wiki/spaces/ECS/pages/3234922524/2026-09-15+5.10.4+emqx+GreptimeDB+Connector+Failure+Test

GreptimeDB point rendering could raise `badarg` when an integer or unsigned-integer suffix received a value of the wrong type, or when the final Unicode conversion rejected malformed input.  The resource buffer treated that exception as retriable, retained the whole batch in inflight, and blocked subsequent messages assigned to the worker.

This change propagates invalid rendered values through the existing `greptimedb_trans_point_failed` path.  A malformed batch keeps the existing all-or-nothing behavior, but now returns an explicit `unrecoverable_error` and is dropped instead of entering blocked retry processing.  The Unicode fallback catches only `error:badarg`; unrelated exceptions remain visible.

## PR Checklist

- [x] Schema changes are backward compatible or intentionally breaking (no schema format change)


zmstone:
should port to 6.0?

hjianbo:
yes

changes/ee/fix-18976.en.md
```diff
@@ -0,0 +1 @@
+Fixed GreptimeDB actions to treat invalid integer, unsigned-integer, and Unicode point values as unrecoverable transformation errors. Malformed batches are now dropped instead of entering blocked retry processing and preventing subsequent messages from being sent.
```
zmstone: **suggestion:** This text describes the implementation. A user sees a GreptimeDB action that stops sending data. Consider:

> Fixed an issue where a GreptimeDB action stopped writing data after it received a message whose value did not match the type in the write syntax, for example a float in a field with the `i` suffix. Such a message is now dropped and counted as failed. When batching is enabled, the other messages in the same batch are dropped with it.

The last sentence matters. With the default batch size of 100, one bad message drops up to 99 valid ones.

apps/emqx_bridge_greptimedb/src/emqx_bridge_greptimedb_connector.erl
```diff
@@ -673,12 +673,16 @@ parse_timestamp([TsBin]) ->
...
-        {_, [#{fields := Fields}]} when map_size(Fields) =:= 0 ->
+        {ok, {_, [#{fields := Fields}]}} when map_size(Fields) =:= 0 ->
             %% greptimedb client doesn't like empty field maps...
             ErrorPointsAcc1 = [{error, no_fields} | ErrorPointsAcc],
             lines_to_points(Data, DbName, Rest, ResultPointsAcc, ErrorPointsAcc1);
-        Point ->
-            lines_to_points(Data, DbName, Rest, [Point | ResultPointsAcc], ErrorPointsAcc)
+        {ok, Point} ->
+            lines_to_points(Data, DbName, Rest, [Point | ResultPointsAcc], ErrorPointsAcc);
+        {error, Reason} ->
```
zmstone, line 742: **question:** The batch callbacks wrap this error as `{error, {unrecoverable_error, points_trans_failed}}`. `on_query/3` and `on_query_async/4` still return a plain `{error, ErrorPoints}`. The buffer worker drops that today only because `emqx_resource_buffer_worker:is_unrecoverable_error/1` treats every plain `{error, _}` as unrecoverable, and that clause is marked `TODO: delete this clause`. If it is removed, a single bad message goes back into the retry loop that this PR fixes. Could the single-message callbacks return `{error, {unrecoverable_error, ErrorPoints}}` as well?

apps/emqx_bridge_greptimedb/src/emqx_bridge_greptimedb_connector.erl
```diff
@@ -692,17 +696,26 @@ line_to_point(
...
-    TableName = emqx_placeholder:proc_tmpl(Measurement, Data),
-    Metric = #{dbname => DbName, table => TableName, timeunit => ToPrecision},
-    {Metric, [
-        maps:without([precision, measurement], Item#{
-            tags => EncodedTags,
-            fields => EncodedFields,
-            timestamp => maybe_convert_time_unit(Ts, Precision)
-        })
-    ]}.
+    case config_to_data(Data, Tags) of
```
zmstone: **suggestion:** The two nested `case` expressions could be one `maybe` block, which the project style prefers for sequential checks. On OTP 26 the module needs `-feature(maybe_expr, enable).`.

```erlang
maybe
    {ok, EncodedTags} ?= config_to_data(Data, Tags),
    {ok, EncodedFields} ?= config_to_data(Data, Fields),
    TableName = emqx_placeholder:proc_tmpl(Measurement, Data),
    Metric = #{dbname => DbName, table => TableName, timeunit => ToPrecision},
    {ok, {Metric, [ ... ]}}
end
```

apps/emqx_bridge_greptimedb/src/emqx_bridge_greptimedb_connector.erl
```diff
@@ -712,61 +725,80 @@ time_unit(ms) -> millisecond;
...
 value_type([<<"FALSE">>]) ->
-    greptimedb_values:boolean_value(false);
+    {ok, greptimedb_values:boolean_value(false)};
 value_type([<<"False">>]) ->
-    greptimedb_values:boolean_value(false);
+    {ok, greptimedb_values:boolean_value(false)};
 value_type([Float]) when is_float(Float) ->
-    Float;
+    {ok, Float};
 value_type(Val0) ->
```
zmstone, line 873: **question:** On 5.10 a field without a suffix whose value is an integer also ends in this clause. `unicode:characters_to_binary([470], utf8)` returns `<<"ǖ"/utf8>>`, so `temp=${t}` with `t = 470` writes the string `ǖ`. An integer above U+10FFFF is now dropped as `non_utf8_string_value`. Release 6.2 and later cast a bare integer to a float with `value_type([Int]) when is_integer(Int) -> greptimedb_values:float64_value(Int)`, added in #16647. That was a feature upstream, so a port may be out of scope here. This PR rewrites the same function, though. Should this be decided now?

apps/emqx_bridge_greptimedb/src/emqx_bridge_greptimedb_connector.erl
```diff
@@ -712,61 +725,80 @@ time_unit(ms) -> millisecond;
...
-                throw({unrecoverable_error, {non_utf8_string_value, Val0}})
-        end,
-    greptimedb_values:string_value(Val).
+    try unicode:characters_to_binary(Val0, utf8) of
+        Val1 when is_binary(Val1) ->
+            {ok, greptimedb_values:string_value([Val1])};
+        _Error ->
+            {error, {non_utf8_string_value, Val0}}
+    catch
+        error:badarg ->
```
zmstone, line 880: **suggestion:** `non_utf8_string_value` does not describe everything that reaches this point. The new test passes `[470.5, <<"x">>]`, which is a float in a mixed template. An integer above U+10FFFF also ends here, through the `_Error` clause above. An operator who reads the log will look for invalid bytes that do not exist. A name such as `invalid_string_value` fits all of these cases.

apps/emqx_bridge_greptimedb/src/emqx_bridge_greptimedb_connector.erl
```diff
@@ -825,6 +857,55 @@ is_auth_key_test_() ->
         ?_assertNot(is_auth_key(89))
     ].
 
+integer_point_validation_test() ->
```
zmstone, line 944: **suggestion:** This test checks return values. The reported symptom was a worker that stayed blocked, so later messages were never sent. A case in `emqx_bridge_greptimedb_SUITE` would cover that directly. Publish one message with a float in the `i` field, then several valid messages. Assert that the valid messages reach GreptimeDB, that the bad one counts as failed, and that `queuing` and `inflight` return to 0. Run it in both the batch and the non-batch groups.

